PR #4094, opened 2025-06-07
feat: template version display

yiqun12:
Fixes #4006 

Key Changes

Version Fetching: In aboutPanelStore.ts, the templateVersionStore is used to asynchronously fetch the version of workflow_templates.

UI Update: A new badge is added to the coreBadges computed property in aboutPanelStore.ts to display the fetched version.

Link Addition: The badge not only shows the version but also links directly to the [Comfy-Org/workflow_templates](https://github.com/Comfy-Org/workflow_templates) GitHub repo for quick access to the source code and changelog.
image (3)

┆Issue is synchronized with this [Notion page] by [Unito](https://www.unito.io)


webfiltered:
Can you please check the failed test, and make a (very brief) note of what caused it?

yiqun12:
> Can you please check the failed test, and make a (very brief) note of what caused it?

I think my Playwright test failed because it couldn’t find the element in the about settings. Let me rewrite the test.

webfiltered (commented):
The code style looks great.  My only concern is that I think this can be done with much less code by extending an existing API (mentioned on core review).

yiqun12:
reopen

src/stores/aboutPanelStore.ts
```diff
@@ -11,9 +11,14 @@ export const useAboutPanelStore = defineStore('aboutPanel', () => {
   const frontendVersion = __COMFYUI_FRONTEND_VERSION__
   const extensionStore = useExtensionStore()
   const systemStatsStore = useSystemStatsStore()
+
   const coreVersion = computed(
     () => systemStatsStore?.systemStats?.system?.comfyui_version ?? ''
   )
+  const workflowsTemplatesVersion = computed(
+    () =>
+      systemStatsStore?.systemStats?.system?.workflows_templates_version ?? ''
```
webfiltered: It looks like this is making two API calls now.

webfiltered (changes requested):

src/stores/aboutPanelStore.ts
yiqun12: hmm. I will double check and fix it later.

webfiltered (dismissed):

vite.config.mts
```diff
@@ -31,6 +31,9 @@ export default defineConfig({
   server: {
     host: VITE_REMOTE_DEV ? '0.0.0.0' : undefined,
     proxy: {
+      '/templates_version': {
+        target: DEV_SERVER_COMFYUI_URL
+      },
```
christian-byrne: Why is this still here?

christian-byrne (changes requested):

vite.config.mts
yiqun12: Just deleted

Issue #4061, opened 2025-06-03
integrate forge into load3d


jtydhr88:
Forge: https://forge.dev/, designed by world labs, Gaussian Splatting renderer for THREE.js
I try a bit to include it into load3d node, but it failed by
![Image](https://github.com/user-attachments/assets/adabbc37-ba76-4cbb-9605-64de20195d42)
seems we need to enable wasm on ComfyUI server side. but I am not sure we are allowed to do.

If we are not allowed on core side, I might write custom node to support it

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4061-integrate-forge-into-load3d-2076d73d365081d4b515ddfa249fc63e) by [Unito](https://www.unito.io)

jtydhr88:
finished

PR #4095, opened 2025-06-07
Improve bg color image logic

jtydhr88:
no feature changed, but a logic improvement.

Previously the logic of setting background color and image are two different parts, this PR merged them into one.

in fact, this is a prerequisite of while I invesigating integrate Forge https://github.com/Comfy-Org/ComfyUI_frontend/issues/4061, the previous logic has conflict to Forge render system.

Also included one bugfix in this PR for preview camera caused by previous changes 

No releated to Frontend core.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4095-Improve-bg-color-image-logic-20b6d73d365081069e4bcf6b92b0af79) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/load3d/Load3d.ts
+++ b/src/extensions/core/load3d/Load3d.ts
@@ -104,9 +104,7 @@ class Load3d {
       this.getActiveCamera.bind(this),
       this.getControls.bind(this),
       () => this.renderer,
-      this.eventManager,
-      this.sceneManager.backgroundScene,
-      this.sceneManager.backgroundCamera
+      this.eventManager
     )
 
     this.modelManager = new ModelManager(
@@ -183,7 +181,6 @@ class Load3d {
     this.renderer.setScissor(0, 0, width, height)
     this.renderer.setScissorTest(true)
 
-    this.renderer.clear()
     this.sceneManager.renderBackground()
     this.renderer.render(
       this.sceneManager.scene,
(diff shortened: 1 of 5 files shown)


christian-byrne (approved):
LGTM

Also asked for Claude review:

>   No major bugs or breaking issues found. The PR appears to be a solid refactoring that consolidates
>   background rendering logic. Here's what I observed:
> 
>   1. Main improvement: The PR merges the previously separate background color and image handling into a
>   unified system. This is cleaner architecture.
>   2. Bug fix included: Fixed a preview camera issue caused by previous changes - the preview camera now
>   properly clones and syncs with the main camera type (perspective vs orthographic).
>   3. Key changes that look safe:
>     - Background rendering now uses a single approach with material swapping instead of separate paths
>     - Preview background handling is more consistent with the main scene
>     - Proper cleanup of materials when switching between color/image backgrounds
>     - Better memory management with explicit disposal of textures and materials
>   4. Minor observations (not bugs):
>     - The syncWithMainCamera() method was removed and its logic moved inline into renderPreview(), which is
>    cleaner
>     - The clear color is now always black with alpha 0, relying on background mesh for color/image display
>     - Better handling of orthographic camera zoom calculations for the preview
> 
>   The changes follow good Three.js practices for material disposal and memory management. The refactoring
>   makes the code more maintainable without changing any user-facing features.

webfiltered (approved):

PR #4100, opened 2025-06-08
[fix] resolve @ symbol parsing errors in extension tooltips

christian-byrne:
Fixes extension settings tooltips containing @ symbols causing Vue i18n "Invalid linked format" errors by using the safe translation function that bypasses i18n parsing for fallback text.

Fixes #4063

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4100-fix-resolve-symbol-parsing-errors-in-extension-tooltips-20c6d73d365081ad8ccccfadf77bb2f7) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/setting/SettingItem.spec.ts
+++ b/src/components/dialog/content/setting/SettingItem.spec.ts
@@ -17,6 +17,11 @@ vi.mock('@/utils/formatUtil', () => ({
   normalizeI18nKey: vi.fn()
 }))
 
+vi.mock('@/i18n', () => ({
+  st: vi.fn((_key: string, fallback: string) => fallback),
+  t: vi.fn((key: string) => key)
+}))
+
 describe('SettingItem', () => {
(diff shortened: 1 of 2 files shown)


src/components/dialog/content/setting/SettingItem.spec.ts
```diff
@@ -17,6 +17,11 @@ vi.mock('@/utils/formatUtil', () => ({
   normalizeI18nKey: vi.fn()
 }))
 
+vi.mock('@/i18n', () => ({
+  st: vi.fn((_key: string, fallback: string) => fallback),
+  t: vi.fn((key: string) => key)
+}))
```
webfiltered: My first impression is that this voids the intent of the new test - did not investigate fully.

christian-byrne: True. Seems like this mock is actually not even needed in the first place.

webfiltered (approved):

PR #4109, opened 2025-06-08
feat: ✨ support frontend node description

melMass:
I'm not sure if it's enough as I did not test it yet but I think we should be able to set custom description for frontend nodes.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4109-feat-support-frontend-node-description-20c6d73d3650819d9318d19f4046bf5c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -832,7 +832,8 @@ export class ComfyApp {
           output_is_list: [],
           output_node: false,
           python_module: 'custom_nodes.frontend_only',
-          description: `Frontend only node for ${name}`
+          // @ts-expect-error description doesn't exist in core LT types
+          description: node.description || `Frontend only node for ${name}`
         } as ComfyNodeDefV1
       ])
     )
--- a/src/types/litegraph-augmentation.d.ts
+++ b/src/types/litegraph-augmentation.d.ts
@@ -124,6 +124,9 @@ declare module '@comfyorg/litegraph' {
 
     comfyClass?: string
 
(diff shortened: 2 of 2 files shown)


webfiltered (commented):
Hrm, could be a little bit more complicated.  The `description` declared in the augmentations file is for the instance, but the description being called in app.ts is actually a static class prop.

We can push this up to litegraph, but it would be good to have this come from the python side, originally.  I thought we already had a way to do this, but I might just be thinking of hte tooltips.

melMass:
Oh right that's static. Where would it be best?

I think you got confused about one thing, this is for frontend only nodes, python nodes do work like you described already. But FE ones get overridden by the text in the screenshot

webfiltered:
Haha, I thought you were adding something completely custom, outside this repo.. disregard!

Anyway, can you point me at where you're declaring your node, if it's published?  Or something similar?  Might be better for me to look at this in reverse.  (edit: Will fall back to the rgthree one, otherwise)

melMass:
> Anyway, can you point me at where you're declaring your node, if it's published? Or something similar? Might be better for me to look at this in reverse. (edit: Will fall back to the rgthree one, otherwise)

Yep `Note+` in [mtb](https://github.com/melMass/comfy_mtb/blob/7e89e96e9d83d7753a64a528df0828eb82a5f20d/web/note_plus.js#L867-L871)

DrJKL:
Closing in favor of https://github.com/Comfy-Org/ComfyUI_frontend/pull/5187
@melMass 🫡

Issue #2154, opened 2025-01-04
[Bug]: After adding nested groups from the inside out, the inner groups cannot be operated


lldacing:
### Frontend Version

1.7.1

### Expected Behavior

After adding another group outside the group, can directly operate on the internal group.
在一个分组外面再添加一个分组后，可以对内部分组做局部修改（移动，修改标题，忽略节点）

### Actual Behavior

After adding another group outside the group, cannot directly operate on the internal group, it works after do something for the outermost group.
在一个分组外面再添加一个分组后，无法对内部的分组单独做操作，通过移动外层分组或对内部节点操作后才恢复正常
![PixPin_2025-01-04_19-25-38](https://github.com/user-attachments/assets/d04612e7-8db7-45be-97c0-5b837b2dcb3f)

### Steps to Reproduce

1. Add a group A for some nodes
2. Add a group B containing group A
3. Try move group A, rename title of group A

### Debug Logs

```powershell
N/A
```

### Browser Logs

N/A

### Setting JSON

N/A

### What browsers do you use to access the UI ?

Microsoft Edge

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-2154-Bug-After-adding-nested-groups-from-the-inside-out-the-inner-groups-cannot-be-oper-1716d73d36508146b69dd273c6921a42) by [Unito](https://www.unito.io)

webfiltered:
Technical detail:
- Known issue due to stacking
- Intention was to resolve when impl. litegraph group overhaul (new design resolves organically)
- Workaround is possible: recalc group stacking on add

a-und-b:
Hi @webfiltered – could you please give a hint/starting point for this?
> Workaround is possible: recalc group stacking on add

webfiltered:
Litegraph - LGraphGroup.  Detect stacking when adding a group by checking overlaps of `boundingRect`.  Create a list, reorder the groups so the parent group is lowest on the stack.

a-und-b:
Great, thanks a lot!

yiqun12:
https://github.com/user-attachments/assets/a4692e6b-c3ed-4c11-99eb-839a023fe19e I think simply calling group.recomputeInsideNodes() whenever a new group is added could resolve this issue.

robinjhuang:
[...]

PR #4156, opened 2025-06-14
Bug Fix issue where inner groups become unresponsive after nesting (#2154)

yiqun12:
Description
This PR fixes the issue described in #2154 where inner groups become unresponsive after adding nested groups from the outside.
Fix:
Simply calling group.recomputeInsideNodes() whenever a new group is added could resolve this issue.

https://github.com/user-attachments/assets/d8a5e0b2-1979-4b4a-a89c-4523612338cd

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4156-Bug-Fix-issue-where-inner-groups-become-unresponsive-after-nesting-2154-2126d73d3650819d94c9f1605344e641) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useCoreCommands.ts
+++ b/src/composables/useCoreCommands.ts
@@ -446,6 +446,9 @@ export function useCoreCommands(): ComfyCommand[] {
         )
         group.resizeTo(canvas.selectedItems, padding)
         canvas.graph?.add(group)
+
+        group.recomputeInsideNodes()
+
         useTitleEditorStore().titleEditorTarget = group
       }
     },
--- a/src/extensions/core/groupOptions.ts
+++ b/src/extensions/core/groupOptions.ts
@@ -42,6 +42,8 @@ app.registerExtension({
               this.graph.add(group)
               // @ts-expect-error fixme ts strict error
               this.graph.change()
+
+              group.recomputeInsideNodes()
             }
           })
         }


christian-byrne (approved):
## Technical Analysis Summary

### Root Cause
Groups created via hotkey (Ctrl+G) or context menu don't call `recomputeInsideNodes()`, leaving their `_children` and `_nodes` collections empty even when they have correct visual bounds. This breaks interaction with nested groups because the parent group doesn't know about its contained children.

### Effect
When a group containing other groups is created interactively, inner groups become unresponsive until the outer group is interacted with first, forcing manual recomputation of the parent-child relationships.

### Solution Analysis
This PR correctly adds `group.recomputeInsideNodes()` calls after group creation in both interactive scenarios:
- `useCoreCommands.ts` (line 450): After hotkey group creation
- `groupOptions.ts` (line 46): After context menu group creation

### Performance Considerations
Performance concerns are mitigated because:
- Only affects user-initiated operations (hotkey/context menu), not batch operations
- Doesn't impact workflow loading or programmatic group creation
- O(n) operation during interactive creation is acceptable UX trade-off

### Code Quality
- Minimal, focused changes that directly address the root cause
- Follows existing patterns (similar calls already exist in other creation paths)
- No breaking changes to existing APIs

**Recommendation: Approve** - This is a well-targeted fix that resolves the specific issue without introducing broader side effects.

AustinMroz (approved):

PR #4163, opened 2025-06-14
Update the frontend to support async nodes.

guill:
This is the frontend change associated with the core change at https://github.com/comfyanonymous/ComfyUI/pull/8527

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4163-Update-the-frontend-to-support-async-nodes-2126d73d365081ceb942eaeb8a6f0bf9) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -71,7 +71,6 @@ import { useWorkflowAutoSave } from '@/composables/useWorkflowAutoSave'
 import { useWorkflowPersistence } from '@/composables/useWorkflowPersistence'
 import { CORE_SETTINGS } from '@/constants/coreSettings'
 import { i18n, t } from '@/i18n'
-import type { NodeId } from '@/schemas/comfyWorkflowSchema'
 import { UnauthorizedError, api } from '@/scripts/api'
 import { app as comfyApp } from '@/scripts/app'
 import { ChangeTracker } from '@/scripts/changeTracker'
@@ -189,22 +188,46 @@ watch(
   }
 )
 
-// Update the progress of the executing node
+// Update the progress of executing nodes
 watch(
-  () =>
-    [executionStore.executingNodeId, executionStore.executingNodeProgress] as [
-      NodeId | null,
-      number | null
-    ],
-  ([executingNodeId, executingNodeProgress]) => {
+  () => executionStore.nodeProgressStates,
+  (nodeProgressStates) => {
+    // Clear progress for all nodes first
     for (const node of comfyApp.graph.nodes) {
-      if (node.id == executingNodeId) {
-        node.progress = executingNodeProgress ?? undefined
-      } else {
-        node.progress = undefined
+      node.progress = undefined
+    }
+
+    // Then set progress for nodes with progress states
+    for (const nodeId in nodeProgressStates) {
+      const progressState = nodeProgressStates[nodeId]
+      const node = comfyApp.graph.getNodeById(progressState.display_node_id)
+
+      if (node && progressState) {
+        // Only show progress for running nodes
+        if (progressState.state === 'running') {
+          if (node.progress === undefined || node.progress === 0.0) {
+            console.log(
+              `${Date.now()} Setting progress for node ${node.id} to ${progressState.value / progressState.max}=${progressState.value}/${progressState.max} due to ${nodeId}`
+            )
+            node.progress = progressState.value / progressState.max
+          } else {
+            // Update progress if it was already set
+            console.log(
+              `${Date.now()} Setting progress for node ${node.id} to Math.min(${node.progress}, ${progressState.value / progressState.max}=${progressState.value}/${progressState.max}) due to ${nodeId}`
+            )
+            node.progress = Math.min(
+              node.progress,
+              progressState.value / progressState.max
+            )
+          }
+        }
       }
     }
-  }
+
+    // TODO - Do we need to force canvas redraw here?
+    // comfyApp.graph.setDirtyCanvas(true, true)
+  },
+  { deep: true }
 )
 
 // Update node slot errors
--- a/src/composables/useBrowserTabTitle.ts
+++ b/src/composables/useBrowserTabTitle.ts
@@ -1,6 +1,7 @@
 import { useTitle } from '@vueuse/core'
 import { computed } from 'vue'
 
+import { t } from '@/i18n'
 import { useExecutionStore } from '@/stores/executionStore'
 import { useSettingStore } from '@/stores/settingStore'
 import { useWorkflowStore } from '@/stores/workflowStore'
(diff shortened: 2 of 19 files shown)


src/components/graph/GraphCanvas.vue
```diff
@@ -189,22 +188,46 @@ watch(
...
+      const progressState = nodeProgressStates[nodeId]
+      const node = comfyApp.graph.getNodeById(progressState.display_node_id)
+
+      if (node && progressState) {
+        // Only show progress for running nodes
+        if (progressState.state === 'running') {
+          if (node.progress === undefined || node.progress === 0.0) {
+            console.log(
+              `${Date.now()} Setting progress for node ${node.id} to ${progressState.value / progressState.max}=${progressState.value}/${progressState.max} due to ${nodeId}`
+            )
```
webfiltered, line 211: This for testing - remove before merge?

src/extensions/core/groupNode.ts
```diff
@@ -1178,9 +1179,10 @@ export class GroupNodeHandler {
     node.onDrawForeground = function (ctx) {
       // @ts-expect-error fixme ts strict error
       onDrawForeground?.apply?.(this, arguments)
+      const executionStore = useExecutionStore()
```
webfiltered, line 1182:
```suggestion
      const progressState = useExecutionStore().nodeProgressStates[this.id]
```

src/extensions/core/groupNode.ts
```diff
@@ -1178,9 +1179,10 @@ export class GroupNodeHandler {
...
       if (
-        // @ts-expect-error fixme ts strict error
-        +app.runningNodeId === this.id &&
+        executionStore.nodeProgressStates[this.id] &&
+        executionStore.nodeProgressStates[this.id].state === 'running' &&
```
webfiltered, line 1185:
```suggestion
        progressState &&
        progressState.state === 'running' &&
```

nit: Requires previous.  May need prettier re-format.

src/scripts/api.ts
```diff
@@ -103,7 +104,17 @@ interface BackendApiCalls {
...
   b_preview: Blob
+  /** Binary preview with metadata (node_id, prompt_id) */
+  b_preview_with_metadata: {
+    blob: Blob
+    nodeId: string
+    parentNodeId: string
+    displayNodeId: string
+    realNodeId: string
+    promptId: string
+  }
```
webfiltered, line 115: Mr Blob Preview with Metadata, I presume...

Please ignore this - I just need it permanently committed to GH public record.  =)

src/scripts/app.ts
```diff
@@ -634,10 +636,6 @@ export class ComfyApp {
 
     api.addEventListener('executing', () => {
       this.graph.setDirtyCanvas(true, false)
-      // @ts-expect-error fixme ts strict error
-      this.revokePreviews(this.runningNodeId)
-      // @ts-expect-error fixme ts strict error
-      delete this.nodePreviewImages[this.runningNodeId]
```
webfiltered, line 640: ~~I will probably find the answer later, but if it's super concise to point out why this is unnecessary now, that would be appreciated!~~

Just confirming - looks like this is now handled in `executionStore`?

src/scripts/domWidget.ts
```diff
@@ -245,7 +246,9 @@ export class ComponentWidgetImpl<
     })
     this.component = obj.component
     this.inputSpec = obj.inputSpec
-    this.props = obj.props
+    this.props = obj.componentProps
+      ? ({ ...obj.props, ...obj.componentProps } as P)
+      : obj.props
```
webfiltered, line 251: 😲 TYPE ASSERTION!

Should `componentProps` be a different type to `props`, or can it just be `P` (and remove the `as P`, here)?

src/services/litegraphService.ts
```diff
@@ -96,7 +97,12 @@ export const useLitegraphService = () => {
        */
       #setupStrokeStyles() {
         this.strokeStyles['running'] = function (this: LGraphNode) {
-          if (this.id == app.runningNodeId) {
+          const nodeProgressStates = useExecutionStore().nodeProgressStates
+          const nodeId = String(this.id)
+          if (
+            nodeProgressStates[nodeId] &&
+            nodeProgressStates[nodeId].state === 'running'
+          ) {
```
webfiltered, line 105:
```suggestion
          const state = useExecutionStore().nodeProgressStates[nodeId]?.state
          if (state === 'running') {
```

nit: Simplify code

src/composables/useBrowserTabTitle.ts
```diff
@@ -36,11 +37,35 @@ export const useBrowserTabTitle = () => {
...
+      else {
+        const [nodeId, state] = runningNodes[0]
+        const progress = Math.round((state.value / state.max) * 100)
+        const nodeType =
+          executionStore.activePrompt?.workflow?.changeTracker?.activeState?.nodes.find(
+            (n) => String(n.id) === nodeId
+          )?.type || 'Node'
+
+        return `${executionText.value}[${progress}%] ${nodeType}`
+      }
```
webfiltered, line 64:
```suggestion
      const [nodeId, state] = runningNodes[0]
      const progress = Math.round((state.value / state.max) * 100)
      const nodeType =
        executionStore.activePrompt?.workflow?.changeTracker?.activeState?.nodes.find(
          (n) => String(n.id) === nodeId
        )?.type || 'Node'

      return `${executionText.value}[${progress}%] ${nodeType}`
```

nit: Early returns are more common, simply to avoid the cluster flap that is JavaScript brace nesting.

src/components/graph/GraphCanvas.vue
```diff
@@ -189,22 +188,46 @@ watch(
...
+              progressState.value / progressState.max
+            )
+          }
+        }
       }
     }
-  }
+
+    // TODO - Do we need to force canvas redraw here?
+    // comfyApp.graph.setDirtyCanvas(true, true)
```
webfiltered, line 228: This is probably a very good idea; when a caller dirties the graph, it is it is expected that the caller flags this.

The first param is foreground (nodes, reroutes, sugbraphery).  Second param is links & litegraph groups (probably false here, but not certain).

src/stores/executionStore.ts
```diff
@@ -42,7 +44,22 @@ export const useExecutionStore = defineStore('execution', () => {
...
-  const executingNodeId = ref<NodeId | null>(null)
+  // This is the progress of all nodes in the currently executing workflow
+  const nodeProgressStates = ref<Record<string, NodeProgressState>>({})
+
+  // Easily access all currently executing node IDs
+  const executingNodeIds = computed<NodeId[]>(() => {
+    return Object.entries(nodeProgressStates)
+      .filter(([_, state]) => state.state === 'running')
+      .map(([nodeId, _]) => nodeId)
+  })
```
webfiltered, line 55: Could this be converted to `executingNodes`, and make this ones `executingNodes.map(x => x.id)`)?

To replace duplication here:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/8fcde239de15a656833e931e47e20d05492b93a8/src/composables/useBrowserTabTitle.ts#L42-L47

src/stores/executionStore.ts
```diff
@@ -138,19 +157,46 @@ export const useExecutionStore = defineStore('execution', () => {
...
       activePromptId.value = null
     }
   }
 
+  function handleProgressState(e: CustomEvent<ProgressStateWsMessage>) {
+    const { nodes } = e.detail
+
+    // Revoke previews for nodes that are starting to execute
+    for (const nodeId in nodes) {
+      const nodeState = nodes[nodeId]
```
webfiltered, line 174:
```suggestion
    for (const [nodeId, nodeState] of Object.entries(nodes)) {
```

nit: `for..in` is generally frowned upon in TypeScript, because it walks the object prototype chain (instead of "own properties"), and can do different things with different inputs, that aren't always obvious.

My personal take on it has more expletives.

webfiltered (changes requested):
Aware you probably don't need the context / attaboy, but just so it is clear to any observers: this looks like a lot of changes, but is mostly code style nits and I am very much liking this PR.

Will also likely try this locally once I'm done w/the execution store changes for subgraph - see how they play!

guill:
Closing in favor of this rebased PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/4382
All comments on this PR have been addressed in that PR (in a single commit to make it easier to find).

PR #3997, opened 2025-05-27
load image from output need to change this to upload image correctly

LAOGOU-666:
┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-3997-load-image-from-output-need-to-change-this-to-upload-image-correctly-2006d73d36508169aecde0a4f3ebb048) by [Unito](https://www.unito.io)

PR #4186, opened 2025-06-15
Fix bug: drag-and-drop, copy-paste, and upload don't work in nodes that specify upload folder that isn't 'input'

christian-byrne:
This PR adds support for specifying the upload folder ('input', 'output', or 'temp') for image upload widgets.

## Context

This is the third part of the image upload type refactoring that was split after PR #4185 was reverted in PR #4190:

1. **PR #4192** (merged): Introduced a frontend type augmentation pattern for runtime-injected properties
2. **PR #4200** (merged): Added the `ResultItemType` enum and updated existing code to use it
3. **This PR**: Adds the `ImageUploadFormFields` interface and implements folder option support

### 1. Adds `ImageUploadFormFields` interface
- Created interface in `useNodeImageUpload.ts` with a `type` field for specifying the upload folder
- Updated `uploadFile` function to accept optional `formFields` parameter
- Appends the `type` field to FormData when uploading if provided

### 2. Implements folder option support
- Added `folder?: ResultItemType` to `ImageUploadOptions` interface
- Updated `handleUpload` to pass the folder option to `uploadFile`
- Modified `useImageUploadWidget` to extract the `image_folder` from input options and pass it to `useNodeImageUpload`

## Use Case

This enables image uploads to different folders (not just 'input'), particularly to support nodes like "Load Image (from Outputs)" where users can manually add images via drag-and-drop, copy-paste, or upload button to the 'output' folder.

This completes the full image upload type refactoring by adding the actual folder selection functionality on top of the type definitions from PR #4200.


webfiltered (commented):
Have you read the conversation on #3997? Can discuss next week.

christian-byrne:
Yes, I read it. The route handler already verifies the `type` form field is valid (one of the three possible strings)

christian-byrne:
Uploading to the output folder is part of the public API for quite some time. This node has also existed for a while now, so we should fix the drag-and-drop and copy-paste on it, with the added beenfit of makign cloud `type` param addition easier in near future.

webfiltered:
Probably better just to flag it for design; the UX feels really bizarre.

christian-byrne:
Would like to merge this. There have been multiple requests from community. Also, @comfyui-wiki requested this when preparing materials for edit model release. WDYT @webfiltered?

webfiltered:
Did we get the nod from design?  If the UX has already been approved should be good to go, but I hadn't seen anything.

christian-byrne:
It's been in design channel for multiple days with no objection.

webfiltered:
This gives me so much faith in our design review process hahaha.  Let's ping tomorrow morning and get a yay/nay.  afaik it hasn't even been looked at (has been bigger fish to fry).

christian-byrne:
I'm not toally sure how it's a design thing. If I was to ask design about this, I would explain it like, "it fixes issue in the load image from outputs node where the upload button doesn't work."

christian-byrne:
Correction: there was never a Slack message generated at all. It only works for issues.

webfiltered:
Load image, load image from outputs, picking different folder - this is all UX, and it requires explanation to figure out what everything does.

I'd like to have the UX reviewed before we fix it - right now, part of it has never worked, so can be changed easily.

christian-byrne:
Ok sounds good. In case other readers are confused, the folder selection is not exposed to the user, it's "option" in the sense that it can be passed as an option in the function call -- not an "option" for the user.

comfyui-wiki:
https://comfy-organization.slack.com/archives/C08NYSK0K0A/p1749139931659849
https://comfy-organization.slack.com/archives/C08NYSK0K0A/p1749178525080129

I reported the issue with the `Load image (from output)`  node in Slack. What makes me confused is that this node has an interface like ` load image` , it even has same upload button. But it didn't work as I expected. I chose a file and uploaded it, but it didn’t show in the preview, seems it can't worked normally. 

I don't know how it works, but the only thing I think is that this node is broken.

As a user, I don't mind if it is uploaded to the `input` or ` output`  folder, since most of the input images will only be used once.

I think it's not hard to learn that ` Load image` will upload the image to the ` Input`  folder, and ` Load image (from output)`  will upload it to the ` Output`  folder. We can clarify this in the built-in node docs.

And for the edit model and other multi-round editing workflows, if the ` Load image (from output)`  will work as this tweet:https://x.com/c__byrne/status/1934595314108739856 , I think it will improve the UX a lot.

christian-byrne:
Nice, thanks for the approval!

PR #4192, opened 2025-06-16
[refactor] introduce frontend type augmentation pattern

christian-byrne:
Establishes a clear pattern for separating backend API types from frontend runtime augmentations. This allows frontend code to add properties at runtime while maintaining type safety through type guards instead of type assertions.

Part 1 of replacement for https://github.com/Comfy-Org/ComfyUI_frontend/pull/4190

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4192-refactor-introduce-frontend-type-augmentation-pattern-2146d73d365081f8b776c348ffbf8f84) by [Unito](https://www.unito.io)

### DIFF
--- a/src/types/nodeDefAugmentation.ts
+++ b/src/types/nodeDefAugmentation.ts
@@ -1,0 +1,38 @@
+/**
+ * Frontend augmentations for node definitions.
+ *
+ * This module defines type extensions that augment the backend node definition
+ * types with frontend-specific properties. These augmentations are applied at
+ * runtime and are not part of the backend API contract.
+ */
+import type { ComboInputOptions, InputSpec } from '@/schemas/nodeDefSchema'
+
+/**
+ * Frontend augmentation for image upload combo inputs.
+ * This extends ComboInputOptions with properties injected by the uploadImage extension.
+ */
+export interface ImageUploadComboOptions extends ComboInputOptions {
+  /**
+   * Reference to the associated filename combo widget.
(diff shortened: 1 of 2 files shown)


src/types/nodeDefAugmentation.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Frontend augmentation for image upload combo inputs.
+ * This extends ComboInputOptions with properties injected by the uploadImage extension.
+ */
+export interface ImageUploadComboOptions extends ComboInputOptions {
+  /**
+   * Reference to the associated filename combo widget.
+   * Injected by uploadImage.ts to link upload buttons with their combo widgets.
+   *
+   * @frontend-only This property exists only in the frontend runtime.
```
webfiltered: I'd avoid creating new JSDoc tags for things like this.  We will either end up with partial coverage and wonder what they're for, or do busywork keeping things updated.  The standard tags cover nearly everything - some minor concepts from other doc systems like @remarks and @internal aren't present.

Also, pretty sure dashes aren't valid, but could be wrong!

webfiltered (approved):
One ultra-nit.

src/types/nodeDefAugmentation.ts
christian-byrne: Thanks.

christian-byrne: I love ultra-nits :D

PR #4196, opened 2025-06-16
[Manager] Impletent “Install All” button

viva-jinyi:
## Summary

This PR implements the "Install All" functionality for missing node packs in both the missing nodes dialog and manager dialog's missing tab.

## Behavior

- **Missing Nodes Dialog (`LoadWorkflowWarning.vue`)**:
  - Displays "Install All" button next to existing "Manager" button
  - Button only appears when there are missing node packs detected
  - One-click installation of all missing dependencies

- **Manager Dialog (`ManagerDialogContent.vue`)**:
  - Shows "Install All" button in top-right when viewing "missing" tab
  - Button integrates seamlessly with existing manager UI
  - Automatically detects and installs all missing packs from current workflow

- **Pack Detection Logic**:
  - `useMissingNodes` composable automatically identifies missing node packs
  - Filters out already installed packs from workflow requirements
  - Handles loading states and error conditions gracefully

## Components

- **`PackInstallAllButton.vue`**: Reusable install all button component
- **`useMissingNodes.ts`**: Core composable for missing pack detection and management
- Enhanced existing components with install all integration
- Updated styling in `RegistrySearchBar.vue` and `PackActionButton.vue`

## Notes

- i18n support added for install all button text

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4196-Manager-Impletent-Install-All-button-2146d73d365081bf9c80fa754726e51c) by [Unito](https://www.unito.io)


src/components/dialog/content/LoadWorkflowWarning.vue
```diff
@@ -30,8 +30,18 @@
...
   </ListBox>
-  <div v-if="isManagerInstalled" class="flex justify-end py-3">
-    <Button label="Open Manager" size="small" outlined @click="openManager" />
+  <div class="flex justify-end py-3">
+    <PackInstallAllButton
+      :disabled="isLoading || !!error || missingNodePacks.length === 0"
+      :node-packs="missingNodePacks"
+    />
+    <Button
+      v-if="isManagerInstalled"
```
christian-byrne: Can we move the `v-if="isManagerInstalled"` check to the parent div that wraps both buttons?

For context: This conditional check ensures we only show these buttons when the user has ComfyUI-Manager installed. Since Manager hasn't been migrated to core ComfyUI yet, these buttons are non-functional without the ComfyUI-Manager custom node.

Keep in mind though that when using the vite dev server, the frontend extensions are not loaded. So the `isManagerInstalled` will always be false. You can just hardcode the variable to `true` while developing features like this.

christian-byrne (changes requested):
Nice work on this! Before merging, let's handle two things:

1. **Reuse check**: Can we investigate whether the existing `PackInstallButton.vue` would work here instead of creating a new component?
2. **Rebase**: Please rebase onto main when you get a chance.

Thanks!

webfiltered (commented):
@christian-byrne Do you have capacity to review & merge this?  It could go straight into 1.23.0.

Edit: The timing of this was absolutely ridiculous.  >.>

src/components/dialog/content/LoadWorkflowWarning.vue
viva-jinyi: kk! Thank you!

christian-byrne (approved):
LGTM!

christian-byrne:
To clean up the commit history, you can rebase instead of merging. Here's how:

```bash
git checkout migration/feat/installall
git rebase main
```

If there are conflicts, resolve them and run `git rebase --continue`. Once done, you'll need to force push: `git push --force-with-lease`

This will remove the merge commits and create a cleaner linear history.

PR #4200, opened 2025-06-16
[refactor] Add ResultItemType and improve image upload typing

christian-byrne:
This PR adds the ResultItemType enum for proper typing of image folder values, completing part of the image upload type refactoring that was split after PR #4185 was reverted in PR #4190.

## Context

PR #4185 originally attempted to improve type safety for image upload functionality but was reverted because it mixed backend API types with frontend-only runtime properties in the schema. The refactor was then split into multiple parts:

1. **PR #4192** (merged): Introduced a frontend type augmentation pattern to handle runtime-injected properties like `imageInputName` without modifying the schema
2. **This PR**: Adds the `ResultItemType` enum and updates existing code to use it
3. **PR #4186**: Will add the `ImageUploadFormFields` interface and update `useNodeImageUpload.ts` (to be rebased after this PR merges)

### 1. Adds `ResultItemType` enum
- Created a proper enum in `apiSchema.ts` for the `type` field values: `'input'  < /dev/null |  'output' | 'temp'`
- These values represent which folder uploaded files should be stored in
- Exported both the zod enum (`resultItemType`) and TypeScript type (`ResultItemType`)

### 2. Updates existing code to use proper types
- `uploadAudio.ts`: Replaced local `FolderType` with shared `ResultItemType`
- `imagePreviewStore.ts`: Updated function signatures to use `ResultItemType` instead of `string`
- `nodeDefSchema.ts`: Updated `image_folder` to use the imported `resultItemType` zod enum

## Why this approach is better

- **Type safety**: Ensures folder types are constrained to valid values at compile time
- **Single source of truth**: The `ResultItemType` enum is defined once and reused everywhere
- **Clear separation**: Backend API types are separate from frontend augmentations
- **No schema pollution**: Only properties that exist in the backend are included in the schema

## Testing

- ✅ All TypeScript checks pass (`npm run typecheck`)
- ✅ All linting checks pass (`npm run lint`)
- ✅ All unit tests pass (`npm run test:unit`)
- ✅ No regressions in existing functionality

Note: The changes to `useNodeImageUpload.ts` have been moved to PR #4186 to keep this PR focused on just the type definitions.

### DIFF
--- a/src/extensions/core/uploadAudio.ts
+++ b/src/extensions/core/uploadAudio.ts
@@ -5,14 +5,13 @@ import { useNodeDragAndDrop } from '@/composables/node/useNodeDragAndDrop'
 import { useNodeFileInput } from '@/composables/node/useNodeFileInput'
 import { useNodePaste } from '@/composables/node/useNodePaste'
 import { t } from '@/i18n'
+import type { ResultItemType } from '@/schemas/apiSchema'
 import type { ComfyNodeDef } from '@/schemas/nodeDefSchema'
 import type { DOMWidget } from '@/scripts/domWidget'
 import { useToastStore } from '@/stores/toastStore'
 
 import { api } from '../../scripts/api'
 import { app } from '../../scripts/app'
-
-type FolderType = 'input' | 'output' | 'temp'
 
 function splitFilePath(path: string): [string, string] {
   const folder_separator = path.lastIndexOf('/')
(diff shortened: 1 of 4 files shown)


src/composables/node/useNodeImageUpload.ts
webfiltered, line 1: If we get design feedback on [the other](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4186), this can just go in, it's not worth mucking about.

If we don't, we can merge this PR if the changes to this file are reverted (looks like they should be part of the next PR?).

christian-byrne:
Changes:

- This PR now only contains typing of `ResultType` in various locations
- Changes to upload image will be in original PR, which will be rebased after and discussed with design

webfiltered (approved):

PR #4201, opened 2025-06-16
[refactor] File handler registration using Pinia store pattern

christian-byrne:
## Summary

This PR re-lands the file handling refactor from #3955 with a fix for the NaN parsing issue. The original PR was reverted because it broke loading of PNG files containing NaN values in metadata.

## Background

PR #3955 refactored file handling to centralize logic into a new `fileHandlers.ts` module. It was reverted due to issues:
- comfyanonymous/ComfyUI#6915 - "NaN causes workflows not to load from PNG"
- comfyanonymous/ComfyUI#8375 - Related workflow loading issue

## Root Cause

When ComfyUI's backend encounters exceptions in a node's `IS_CHANGED` method, it sets `node["is_changed"] = float("NaN")`, which creates invalid JSON:

```json
{
  "prompt": {"16": {"inputs": {...}, "is_changed": NaN}},
  "workflow": {"id": "...", "nodes": [...]}
}
```

The original refactor eagerly parsed both workflow and prompt fields, causing `JSON.parse()` to fail on the invalid prompt JSON containing literal `NaN`.

## Solution

This PR implements lazy evaluation for metadata parsing:
- Metadata fields are returned as functions that parse on demand
- Preserves the original if-else loading priority: if workflow exists, only parse workflow
- Allows valid workflows to load successfully even when prompt contains invalid JSON

## Testing

Added comprehensive test coverage including:
- Specific case where workflow is valid but prompt contains NaN
- Various edge cases for malformed JSON and missing metadata
- Lazy evaluation behavior verification

Fixes #4199

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -1355,27 +1355,37 @@ export class ComfyApp {
       }
     }
 
+    // Try to load workflow first
     if (metadata.workflow) {
       const workflowData = metadata.workflow()
       if (workflowData) {
         await this.loadGraphData(workflowData, true, true, fileName)
+        return
       }
-    } else if (metadata.prompt) {
+    }
+
+    // If no workflow, try prompt
+    if (metadata.prompt) {
       const promptData = metadata.prompt()
       if (promptData) {
         this.loadApiJson(promptData, fileName)
+        return
       }
-    } else if (metadata.parameters) {
-      // A1111 import
+    }
+
+    // If no workflow or prompt, try A1111 parameters
+    if (metadata.parameters) {
       useWorkflowService().beforeLoadNewGraph()
       importA1111(this.graph, metadata.parameters)
       useWorkflowService().afterLoadNewGraph(
         fileName,
         this.graph.serialize() as unknown as ComfyWorkflowJSON
       )
-    } else {
-      this.showErrorOnFileLoad(file)
+      return
     }
+
+    // No valid data found
+    this.showErrorOnFileLoad(file)
   }
 
   isApiJson(data: unknown) {
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -54,7 +54,8 @@ import type { ComfyExtension, MissingNodeType } from '@/types/comfy'
 import { ExtensionManager } from '@/types/extensionTypes'
 import { ColorAdjustOptions, adjustColor } from '@/utils/colorUtil'
 import { graphToPrompt } from '@/utils/executionUtil'
-import { getFileHandler } from '@/utils/fileHandlers'
+import { useFileHandlerStore } from '@/stores/fileHandlerStore'
+import { registerCoreFileHandlers } from '@/extensions/core/coreFileHandlers'
 import {
   executeWidgetsCallback,
   fixLinkInputSlots,
@@ -749,6 +750,9 @@ export class ComfyApp {
     await useWorkspaceStore().workflow.syncWorkflows()
     await useExtensionService().loadExtensions()
 
+    // Register core file handlers
+    registerCoreFileHandlers()
+
     this.#addProcessKeyHandler()
     this.#addConfigureHandler()
     this.#addApiUpdateHandlers()
@@ -1322,7 +1326,8 @@ export class ComfyApp {
     }
     const fileName = removeExt(file.name)
 
-    const handler = getFileHandler(file)
+    const fileHandlerStore = useFileHandlerStore()
+    const handler = fileHandlerStore.getHandlerForFile(file)
     if (!handler) {
       this.showErrorOnFileLoad(file)
       return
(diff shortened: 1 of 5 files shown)


christian-byrne:
This relands the file handler refactoring, which should make it easier to add more file formats in the future.

The original "bug" is fixed. The bug was that some custom nodes caused invalid JSON saved in the "prompt" tag of images (API format workflow). In the original code, the "prompt" tag JSON string is parsed lazily on the condition that the "workflow" field is undefined -- the refactor made both fields be parsed eagerly which exposed the invalid json issue. 

There are tests added this time, including coverage for the bug.

Can we target 1.24 for this refactor?

webfiltered:
I had a quick read but it looks like the mentions of bugs and solutions are just what the original refactor caused.  Because it was reverted, this is just 100% refactor right - no actual changes to functionality?

src/utils/fileHandlers.ts
```diff
@@ -0,0 +1,334 @@
...
+  return new Promise((resolve, reject) => {
+    const reader = new FileReader()
+    reader.onload = () => {
+      try {
+        const readerResult = reader.result as string
+        resolve({
+          workflow: () => JSON.parse(readerResult),
+          prompt: () => JSON.parse(readerResult),
+          jsonTemplateData: () => JSON.parse(readerResult)
+        })
```
webfiltered, line 140: Can this be simplified?

src/scripts/app.ts
```diff
@@ -1284,163 +1273,71 @@ export class ComfyApp {
...
+        this.graph.serialize() as unknown as ComfyWorkflowJSON
+      )
+      return
+    }
+
+    // No valid data found
+    this.showErrorOnFileLoad(file)
   }
 
   isApiJson(data: unknown) {
```
webfiltered, line 1396: Is this redundant, now?

src/utils/fileHandlers.ts
```diff
@@ -0,0 +1,334 @@
...
+// Register image format handlers
+IMAGE_WORKFLOW_FORMATS.mimeTypes.forEach((mimeType) => {
+  if (mimeType === 'image/png') {
+    mimeTypeHandlers.set(mimeType, handlePngFile)
+  } else if (mimeType === 'image/webp') {
+    mimeTypeHandlers.set(mimeType, handleWebpFile)
+  } else if (mimeType === 'image/svg+xml') {
+    mimeTypeHandlers.set(mimeType, handleSvgFile)
+  }
+})
```
webfiltered: I am really confused about why this has been written this way.  Can you please cast a critical eye over this PR?

webfiltered (changes requested):

src/utils/fileHandlers.ts
christian-byrne: Simplified to not use functions? It was changed to use functions to make the parsing lazy, which will fix the bug mentioned above.

src/scripts/app.ts
christian-byrne: This code executes if no comfy tags are found in the metadata.

src/utils/fileHandlers.ts
christian-byrne: Do you mean as opposed to writing it in a more declarative way with a big config object?

src/utils/fileHandlers.ts
webfiltered: This is the same thing repeated three times - why not just set a flag and use one of the existing two repeats?

src/utils/fileHandlers.ts
webfiltered:
```suggestion
mimeTypeHandlers.set('image/png', handlePngFile)
mimeTypeHandlers.set('image/webp', handleWebpFile)
mimeTypeHandlers.set('image/svg+xml', handleSvgFile)
```

This is what the code does.  It just does it in a loop for absolutely no reason.  It's already hard-coded and static, because it checks for these exact values in every loop.

src/utils/fileHandlers.ts
webfiltered: (or just not repeat it at all, if that works in the design.. which I'm not sure it does)

src/scripts/app.ts
webfiltered: That's GitHub being helpful with surrounding lines again haha.  `isApiJson` is now a function in two places.  Is it used in both - what's the difference?

src/utils/fileHandlers.ts
christian-byrne: This is just implementing the `WorkflowFileHandler` signature. It makes it easier if all registered handlers implement a common type, then consumers don't have to worry about the specifics and flags of each how each handler works.

src/scripts/app.ts
christian-byrne: Good point

PR #4213, opened 2025-06-17
fIx: side toolbar tab tooltip not reactive when changing locale

hyq2015:
In current comfyui, after I changed my locale from 中文 to English , I found the side toobar title/label not up to date, it seems the side toolbar button text is not a reactive value regarding the locale change

image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4213-fix-side-toolbar-tab-text-not-up-to-date-issue-when-changing-locale-2156d73d3650814ebd40d392ee850c0e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/sidebar/SidebarIcon.vue
+++ b/src/components/sidebar/SidebarIcon.vue
@@ -1,6 +1,10 @@
 <template>
   <Button
-    v-tooltip="{ value: tooltip, showDelay: 300, hideDelay: 300 }"
+    v-tooltip="{
+      value: computedTooltip,
+      showDelay: 300,
+      hideDelay: 300
+    }"
     text
     :pt="{
       root: {
@@ -9,7 +13,7 @@
             ? 'p-button-primary side-bar-button-selected'
             : 'p-button-secondary'
         }`,
-        'aria-label': tooltip
+        'aria-label': computedTooltip
       }
     }"
     @click="emit('click', $event)"
@@ -27,16 +31,20 @@
 import Button from 'primevue/button'
 import OverlayBadge from 'primevue/overlaybadge'
 import { computed } from 'vue'
+import { useI18n } from 'vue-i18n'
 
+const { t } = useI18n()
 const {
   icon = '',
   selected = false,
   tooltip = '',
+  tooltipSuffix = '',
   iconBadge = ''
 } = defineProps<{
   icon?: string
   selected?: boolean
   tooltip?: string
+  tooltipSuffix?: string
   iconBadge?: string | (() => string | null)
 }>()
 
@@ -47,6 +55,7 @@ const overlayValue = computed(() =>
   typeof iconBadge === 'function' ? iconBadge() ?? '' : iconBadge
 )
 const shouldShowBadge = computed(() => !!overlayValue.value)
+const computedTooltip = computed(() => t(tooltip) + tooltipSuffix)
 </script>
 
 <style>
--- a/src/components/sidebar/SideToolbar.vue
+++ b/src/components/sidebar/SideToolbar.vue
@@ -27,7 +27,9 @@
 </template>
 
 <script setup lang="ts">
+import { watchImmediate } from '@vueuse/core'
 import { computed } from 'vue'
+import { useI18n } from 'vue-i18n'
 
 import ExtensionSlot from '@/components/common/ExtensionSlot.vue'
 import { useKeybindingStore } from '@/stores/keybindingStore'
@@ -44,6 +46,7 @@ import SidebarThemeToggleIcon from './SidebarThemeToggleIcon.vue'
 const workspaceStore = useWorkspaceStore()
 const settingStore = useSettingStore()
 const userStore = useUserStore()
+const { locale } = useI18n()
 
 const teleportTarget = computed(() =>
   settingStore.get('Comfy.Sidebar.Location') === 'left'
@@ -67,6 +70,11 @@ const getTabTooltipSuffix = (tab: SidebarTabExtension) => {
   )
   return keybinding ? ` (${keybinding.combo.toString()})` : ''
 }
+
+// Update sidebar tab's text when locale changes
+watchImmediate(locale, () => {
+  workspaceStore.updateSidebarText()
+})
 </script>
 
 <style scoped>
--- a/src/components/sidebar/SideToolbar.vue
+++ b/src/components/sidebar/SideToolbar.vue
@@ -6,7 +6,8 @@
         :key="tab.id"
         :icon="tab.icon"
         :icon-badge="tab.iconBadge"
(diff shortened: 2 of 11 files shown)


webfiltered (changes requested):
Hey - thank you for the PR!

So looking this over, it feels like updateSidebarText is a very internal function - it has no params and should only be called be e.g. a watcher. (edit: because `workspaceStore` is a public API for extensions)

Instead, we should probably just make the sidebar text reactive.

If this is something you'd like to look into, please feel free!  If not, no stress - we can look into it when time permits.

hyq2015:
Thanks. Digged a bit, I suggest to follow the pattern in `<SidebarThemeToggleIcon />` and `<SidebarSettingsToggleIcon />`, can use `$t` in the element directly no? And when registerSidebarTab, we can just pass a locale key for both label and tooltip.
and btw the label of tabs seems not used at all, can be cleand?

christian-byrne (commented):
## PR Review: fix: side toolbar tab tooltip not reactive when changing locale (#4213)

### Summary
- Files changed: 7
- Lines added/removed: +22/-22
- Review scope: all

### Review Comments:

1. **[src/stores/workspace/sidebarTabStore.ts:31]** - [SEVERITY: Medium]
   ```typescript
   label: tab.title,
   ```
   **Issue**: The command label is now using the translation key instead of translated text
   **Impact**: Commands in the command palette will show translation keys (e.g., "sideToolbar.workflows") instead of translated labels (e.g., "Workflows")
   **Suggestion**: The command registration needs access to i18n to translate the title:
   ```typescript
   import { useI18n } from 'vue-i18n'
   
   // Inside the registerSidebarTab function
   const { t } = useI18n()
   useCommandStore().registerCommand({
     // ...
     label: t(tab.title),
     // ...
   })
   ```

### Positive Observations:
- ✅ Elegant fix that properly makes tooltips reactive to locale changes
- ✅ Maintains backward compatibility for extensions using plain strings
- ✅ Clean separation of concerns
- ✅ Minimal code changes with maximum effect

christian-byrne:
Tests will re-run when new changes are made :+1:

src/components/sidebar/SidebarIcon.vue
```diff
@@ -9,7 +13,7 @@
             ? 'p-button-primary side-bar-button-selected'
             : 'p-button-secondary'
         }`,
-        'aria-label': tooltip
+        'aria-label': computedTooltip
```
hyq2015, line 16: Hello @christian-byrne , the translate now moved to here, and I tested locally, it seems works well. so I don't really understand your question/fedback :)
image

it could be nice if you can provide some screenshot (or steps to reproduce the sidebar tooltip not been well translated) so that I can find where I missed, thanks a lot!

hyq2015:
> ## PR Review: fix: side toolbar tab tooltip not reactive when changing locale (#4213)
> ### Summary
> * Files changed: 7
> * Lines added/removed: +22/-22
> * Review scope: all
> 
> ### Review Comments:
> 1. **[src/stores/workspace/sidebarTabStore.ts:31]** - [SEVERITY: Medium]
>    ```ts
>    label: tab.title,
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    **Issue**: The command label is now using the translation key instead of translated text
>    **Impact**: Commands in the command palette will show translation keys (e.g., "sideToolbar.workflows") instead of translated labels (e.g., "Workflows")
>    **Suggestion**: The command registration needs access to i18n to translate the title:
>    ```ts
>    import { useI18n } from 'vue-i18n'
>    
>    // Inside the registerSidebarTab function
>    const { t } = useI18n()
>    useCommandStore().registerCommand({
>      // ...
>      label: t(tab.title),
>      // ...
>    })
>    ```
> 
> ### Positive Observations:
> * ✅ Elegant fix that properly makes tooltips reactive to locale changes
> * ✅ Maintains backward compatibility for extensions using plain strings
> * ✅ Clean separation of concerns
> * ✅ Minimal code changes with maximum effect

christian-byrne:
Hi @hyq2015, thanks for working on this\! I've been testing your changes and discovered something interesting.

Your implementation is actually working correctly for the built-in sidebar tabs\! The tooltips for Queue, Node Library, Model Library, and Workflows do translate properly when switching languages. Great work on making them reactive to locale changes.

However, I found an issue with the Theme Toggle and Settings icons. These components (`SidebarThemeToggleIcon` and `SidebarSettingsToggleIcon`) are passing pre-translated strings to `SidebarIcon` using `:tooltip="$t('...')"` in their templates. This means they're passing the already-translated text (like "设置" in Chinese) instead of the translation key.

When `SidebarIcon` receives these pre-translated strings and tries to translate them again with `t()`, it can't find those strings as translation keys, so it shows warnings and falls back to displaying the original string.

To fix this, we have two options:

1. **Update `SidebarThemeToggleIcon` and `SidebarSettingsToggleIcon`** to pass translation keys instead of pre-translated strings:
   ```vue
   // Change from:
   :tooltip="$t('g.settings')"
   // To:
   tooltip="g.settings"
   ```

2. **Or make `SidebarIcon` smarter** to detect if a string is already translated (though this is more complex and error-prone).

I think option 1 is cleaner and more consistent with how the other sidebar tabs work.

Regarding the command labels issue - you're right that we should use `i18n.global.t()` in the store. That's a good catch\!

Would you like to update your PR with these fixes? Let me know if you need any help\!

christian-byrne (approved):
Hi @hyq2015, apologies for the confusion in my previous comments.

I've tested your PR thoroughly:
- Changed language from English to Chinese and back
- All sidebar tooltips (Queue, Node Library, Model Library, Workflows) now properly translate and update when changing locales
- Theme toggle and settings icons also work correctly

Your fix successfully makes the sidebar tooltips reactive to locale changes. This was the issue - tooltips were stuck in English before your changes.

LGTM! Thanks for fixing this.

Let's just see if the automated tests pass and then fix any failures, then we can merge!

christian-byrne:
Just a heads up - the component tests for `SidebarIcon` will fail after this change because the component now uses `useI18n()`. 

To fix the tests, you'll need to update `src/components/sidebar/SidebarIcon.spec.ts` to include the i18n plugin:

```typescript
const i18n = createI18n({
  legacy: false,
  locale: 'en',
  messages: {
    en: {}
  }
})

// In mountSidebarIcon function, add i18n to plugins:
global: {
  plugins: [PrimeVue, i18n],
  // ... rest of config
}
```

You can verify the tests locally with:
```bash
npm run test:component -- src/components/sidebar/SidebarIcon.spec.ts
```

hyq2015:
@christian-byrne oh thanks for the good catch, I will be more careful next time! btw, I fixed the component test error.

christian-byrne (approved):

PR #4222, opened 2025-06-20
[Manager] Modal UI Adjustment (Align with Design)

viva-jinyi:
## Summary

Enhanced Manager Dialog interface and fixed ContentDivider component styling issues. Added responsive layout optimization and proper size constraints for better cross-device experience.

Key improvements:
- Fixed Manager Dialog oversizing on large displays (max-width: 1724px)
- Resolved ContentDivider double-line rendering issue  
- Enhanced responsive layout for mobile/tablet devices
- Improved component isolation and maintainability

## Behavior

**Before:**
- Manager Dialog could become oversized on large screens
- ContentDivider showed double lines due to browser default hr styling conflicts
- Fixed pixel widths caused poor responsiveness
- Sidebar was too wide (320px) for smaller screens

**After:**
- Manager Dialog constrained to max-width: 1724px, max-height: 1026px
- ContentDivider displays clean single lines with proper theme colors
- Responsive layout using relative units (w-3/12, w-4/12)
- Compact sidebar design optimized for all screen sizes

## Components

**ContentDivider.vue** - Complete rewrite
- Replaced `<hr>` with `<div>` to eliminate browser style conflicts
- Used background-color instead of border-color for cleaner rendering
- Added CSS v-bind for dynamic sizing and scoped styles

**GlobalDialog.vue** - Manager-specific styling
- Added `.manager-dialog` class with size constraints

**ManagerDialogContent.vue** - Layout improvements
- Converted fixed widths to responsive units
- Added background color and optimized spacing
- Improved main content area layout

**ManagerNavSidebar.vue** - Responsive sidebar
- Changed from w-80 fixed to max-w-[250px] w-3/12 responsive
- Reduced padding and text sizes for compact design
- Switched from absolute to flex-based positioning

**InfoPanel.vue** - Flexible panel sizing
- Removed fixed width constraints for better responsiveness
- Optimized spacing and positioning

**dialogService.ts** - Manager dialog configuration  
- Added manager-dialog class and improved header styling
- Enhanced content area configuration

[As-is]
스크린샷 2025-06-20 오후 7 01 05

[To-be]
스크린샷 2025-06-20 오후 6 58 55

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4222-Manager-Modal-UI-Adjustment-Align-with-Design-2186d73d365081079bfbe26233855f66) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/GlobalDialog.vue
+++ b/src/components/dialog/GlobalDialog.vue
@@ -50,4 +50,9 @@ const dialogStore = useDialogStore()
   @apply p-2 2xl:p-[var(--p-dialog-content-padding)];
   @apply pt-0;
 }
+
+.manager-dialog {
+  max-width: 1724px;
+  max-height: 1026px;
+}
 </style>
--- a/src/components/dialog/GlobalDialog.vue
+++ b/src/components/dialog/GlobalDialog.vue
@@ -52,6 +52,7 @@ const dialogStore = useDialogStore()
 }
 
 .manager-dialog {
+  height: 80vh;
   max-width: 1724px;
   max-height: 1026px;
 }
@@ -59,7 +60,7 @@ const dialogStore = useDialogStore()
 @media (min-width: 3000px) {
   .manager-dialog {
     max-width: 2200px;
-    max-height: 1300px;
+    max-height: 1320px;
   }
 }
 </style>
--- a/src/components/dialog/GlobalDialog.vue
+++ b/src/components/dialog/GlobalDialog.vue
@@ -55,4 +55,11 @@ const dialogStore = useDialogStore()
   max-width: 1724px;
   max-height: 1026px;
 }
+
+@media (min-width: 3000px) {
+  .manager-dialog {
+    max-width: 2200px;
+    max-height: 1300px;
+  }
+}
 </style>
--- a/src/components/dialog/content/manager/ManagerDialogContent.vue
+++ b/src/components/dialog/content/manager/ManagerDialogContent.vue
@@ -1,6 +1,6 @@
 <template>
   <div
-    class="flex flex-col mx-auto overflow-hidden h-[83vh] max-h-[956px]"
+    class="h-full flex flex-col mx-auto overflow-hidden"
     :aria-label="$t('manager.title')"
   >
     <ContentDivider :width="0.3" />
--- a/src/components/dialog/content/manager/ManagerDialogContent.vue
+++ b/src/components/dialog/content/manager/ManagerDialogContent.vue
@@ -1,14 +1,16 @@
 <template>
   <div
-    class="flex flex-col mx-auto overflow-hidden h-[83vh] relative"
+    class="flex flex-col mx-auto overflow-hidden h-[83vh]"
     :aria-label="$t('manager.title')"
   >
+    <ContentDivider :width="0.3" />
     <Button
       v-if="isSmallScreen"
       :icon="isSideNavOpen ? 'pi pi-chevron-left' : 'pi pi-chevron-right'"
-      text
+      severity="secondary"
+      filled
       class="absolute top-1/2 -translate-y-1/2 z-10"
-      :class="isSideNavOpen ? 'left-[19rem]' : 'left-2'"
+      :class="isSideNavOpen ? 'left-[12rem]' : 'left-2'"
       @click="toggleSideNav"
     />
     <div class="flex flex-1 relative overflow-hidden">
@@ -18,14 +20,12 @@
         :tabs="tabs"
       />
       <div
-        class="flex-1 overflow-auto pr-80"
+        class="flex-1 overflow-auto bg-[#F7F7F7]"
         :class="{
-          'transition-all duration-300': isSmallScreen,
-          'pl-80': isSideNavOpen || !isSmallScreen,
-          'pl-8': !isSideNavOpen && isSmallScreen
+          'transition-all duration-300': isSmallScreen
         }"
       >
-        <div class="px-6 pt-6 flex flex-col h-full">
+        <div class="px-6 flex flex-col h-full">
           <RegistrySearchBar
             v-model:searchQuery="searchQuery"
             v-model:searchMode="searchMode"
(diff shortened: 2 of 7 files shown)


christian-byrne (commented):
Starting review...

src/components/dialog/GlobalDialog.vue
```diff
@@ -50,4 +50,9 @@ const dialogStore = useDialogStore()
   @apply p-2 2xl:p-[var(--p-dialog-content-padding)];
   @apply pt-0;
 }
+
+.manager-dialog {
+  max-width: 1724px;
+  max-height: 1026px;
```
christian-byrne, line 57: Should we use a percentages or aspect ratios? On my 4k monitor, it seems somewhat small.

src/components/dialog/content/manager/ManagerDialogContent.vue
```diff
@@ -18,14 +20,12 @@
         :tabs="tabs"
       />
       <div
-        class="flex-1 overflow-auto pr-80"
+        class="flex-1 overflow-auto bg-[#F7F7F7]"
```
christian-byrne: Can we remove this and use the default color? Otherwise, we need to adjust for dark theme manually using `dark-theme:` prefix.

src/components/dialog/content/manager/ManagerDialogContent.vue
```diff
@@ -77,9 +77,9 @@
           </div>
         </div>
       </div>
-      <div class="w-80 border-l-0 absolute right-0 top-0 bottom-0 flex z-20">
+      <div class="max-w-[306px] w-4/12 border-l-0 flex z-20">
```
christian-byrne: It might be hard to use the info panel at 33% on small screens. How about a lower bound clamp like `w-[clamp(250px,33%,306px)]` or similar.

src/services/dialogService.ts
```diff
@@ -137,8 +137,15 @@ export const useDialogService = () => {
       dialogComponentProps: {
         closable: true,
         pt: {
-          header: { class: '!p-0 !m-0' },
-          content: { class: '!px-0 h-[83vh] w-[90vw] overflow-y-hidden' }
+          pcCloseButton: {
+            root: { class: 'bg-gray-500 w-9 h-9 p-1.5 rounded-full text-white' }
```
christian-byrne: Would you like to add a color for dark theme as well? You can use `dark-theme:` prefix.

christian-byrne (changes requested):
This is a good PR. Great improvements to responsiveness and the divider. I've added some comments.

src/services/dialogService.ts
viva-jinyi: I added background color for dark theme. Thanks!

src/components/dialog/content/manager/ManagerDialogContent.vue
viva-jinyi: oh, Great Idea. Thanks.

src/components/dialog/GlobalDialog.vue
viva-jinyi: I think we should adjust the max-width and max-height using media queries to better support users with ultra-wide monitors.

as-is
스크린샷 2025-06-23 오후 3 38 56

to-be
스크린샷 2025-06-23 오후 3 38 45

src/components/dialog/content/manager/ManagerDialogContent.vue
viva-jinyi: I added background color for dark theme. Thanks!

src/services/dialogService.ts
viva-jinyi: I added. Thanks!

src/components/dialog/GlobalDialog.vue
christian-byrne: @viva-jinyi Can we get closer to the screen proportions used in the [design spec](https://www.figma.com/design/sRL0jz7UP97n6ye2hwAGJs/Manager?node-id=211-4553&t=IqGfIb75tPNRTbEY-4)? I'm worried we are too far off from what the designers worked hard on.

I think it's okay to use more of the screen and it's more accomodating for users with unconventional screen sizes.

christian-byrne:
FYI: Left a comment [here](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4222#discussion_r2160938917)

src/components/dialog/GlobalDialog.vue
viva-jinyi: Ah, I see now that you were referring to the screen proportions — I hadn’t checked that part properly, so thanks for pointing it out.

I was mainly focused on the fact that the modal version specifies a max-width of **1724px**.
Previously, the modal would expand infinitely based on the browser size, which is what I wanted to prevent.

As you mentioned, it’s important that the modal’s proportions don’t get distorted.
That issue was caused by how I initially handled the styles.

**Before:**
Inside the manager-dialog, the content height was calculated based on the browser height.
I was using a fixed max-height in px or vh to make sure the content didn’t overflow the modal.
But this meant it ignored the height set by the .manager-modal class — which caused the proportions to break.

**After:**
I changed it so the content height is now always based on the modal size — using height: 100% for the content,
excluding any fixed-height sections (like the header).
Now the modal dimensions (both width and height) are determined entirely by the .modal-manager class.

However, there’s **one exception:**
When the browser window becomes too small, it’s impossible to maintain the full proportions — especially the width.
In those cases, **only the height is preserved, while the width shrinks responsively,**
which results in some distortion of the original modal proportions.
This is a necessary compromise to maintain usability on smaller screens.

I’ll attach screenshots sorted by monitor size from largest to smallest.

스크린샷 2025-06-24 오후 12 58 48
스크린샷 2025-06-24 오후 12 58 56
스크린샷 2025-06-24 오후 12 59 04
스크린샷 2025-06-24 오후 12 59 13
스크린샷 2025-06-24 오후 12 59 21

christian-byrne (approved):
LGTM

Issue #4183, opened 2025-06-14
Implement cascading failover for registry search services


christian-byrne:
## Goal

Since we distribute a native application, hard-coding dependencies on external services like Algolia creates vendor lock-in risk. If their service is discontinued, users on older versions would lose functionality until they update.

The best solution would be to do client service discovery (client queries a service registry API at runtime which returns currently available service instances with their endpoints => client selects from the live list and handles load balancing/failover)

For the sake of simplicity (and since we won't expect external services to register themselves automatically in some service registry) we can instead implement client-side service abstraction with cascading failover. We maintain a configured list of search providers (Algolia, Elasticsearch, our own API) that implement a common interface. The client fails over through providers at runtime using circuit breaker logic, with our own API as the guaranteed final fallback.

This eliminates vendor lock-in without the complexity of dynamic service discovery, allowing us to add/remove providers through configuration while ensuring service availability.

## Implementation Steps

1. Create an interface that dictates all necessary attributes a node pack search service needs to implement such that it will work with the current frontend

  ```typescript
  export interface SearchService {
    /**
     * Search for node packs with caching support
     */
    searchPacksCached(
      query: string,
      params: SearchNodePacksParams
    ): Promise<SearchPacksResult>
  
    /**
     * Convert from search result format to Comfy Registry format
     */
    toRegistryPack(searchResult: any): RegistryNodePack
  
    /**
     * Clear the search cache
     */
    clearSearchPacksCache(): void
  }
  
  export interface SearchServiceOptions {
    /**
     * Maximum number of search results to store in the cache
     * @default 64
     */
    maxCacheSize?: number
    /**
     * Minimum number of characters for suggestions
     * @default 2
     */
    minCharsForSuggestions?: number
  }
  ```

2. Make the current algolia search service implement the interface
3. Make a new search service that uses api.comfy.org endpoints and parameters. Make that also implement the interface
4. In all places where the current search service is initialized, creating cascading pattern to try each search service until one of them works without error

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4183-Implement-cascading-failover-for-registry-search-services-2126d73d365081d7ba02eec594e912da) by [Unito](https://www.unito.io)

PR #4249, opened 2025-06-23
[Manager] Fix: When using registry search provider, results not properly paginated'

christian-byrne:
Fixes bug introduced in https://github.com/Comfy-Org/ComfyUI_frontend/issues/4183 in which the registry search provider's results are not properly paginated (keeps getting the first page of results over and over).

The `offset` parameter was being used in search query, but the API expects `page` (`offset` is not specified in the [spec](https://api.comfy.org/openapi) at all). See also [generated types](https://github.com/Comfy-Org/ComfyUI_frontend/blob/3a1bd1829a06518dde21be80614be69fd129d52a/src/types/comfyRegistryTypes.ts#L10838-L10907)

https://github.com/user-attachments/assets/3c255993-294a-4acf-96bc-c6835efc1361

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4249-Manager-Fix-When-using-registry-search-provider-results-not-properly-paginated-21b6d73d3650812f8423cd850538d7f8) by [Unito](https://www.unito.io)

### DIFF
--- a/src/services/providers/registrySearchProvider.ts
+++ b/src/services/providers/registrySearchProvider.ts
@@ -32,7 +32,7 @@ export const useComfyRegistrySearchProvider = (): NodePackSearchProvider => {
       search: isNodeSearch ? undefined : query,
       comfy_node_search: isNodeSearch ? query : undefined,
       limit: pageSize,
-      offset: pageNumber * pageSize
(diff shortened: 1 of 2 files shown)


viva-jinyi (approved):
Looks good to me 👍

PR #4260, opened 2025-06-24
fix: node migration TypeError

viva-jinyi:
**Context**
[Slack Threads](https://comfy-organization.slack.com/archives/C075ANWQ8KS/p1750756579849109)

Added a guard clause for cases where a custom node defines RETURN_TYPES = ("BOOLEAN") instead of a proper tuple like ("BOOLEAN",).
Without the trailing comma, it’s interpreted as a string rather than an array, which causes issues on the frontend.
This fix ensures the frontend can handle such cases more gracefully.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4260-fix-node-migration-TypeError-21c6d73d365081c7aee4f1b125924744) by [Unito](https://www.unito.io)

### DIFF
--- a/src/schemas/nodeDef/migration.ts
+++ b/src/schemas/nodeDef/migration.ts
@@ -45,23 +45,27 @@ export function transformNodeDefV1ToV2(
   // Transform outputs
   const outputs: OutputSpecV2[] = []
 
-  if (nodeDefV1.output && Array.isArray(nodeDefV1.output)) {
-    nodeDefV1.output.forEach((outputType, index) => {
-      const outputSpec: OutputSpecV2 = {
-        index,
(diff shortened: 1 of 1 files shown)


webfiltered (changes requested):
To my eyes, it looks like this will now silently discard the error instead of throwing - but without any indication as to why nothing happened, or how to fix it.  I think we should instead do a nested if instead of `&&` - and probably log the error condition.

viva-jinyi:
@webfiltered Thank you for pointing out. I'll make the changes accordingly.

webfiltered (approved):

webfiltered:
Awesome - thank you!

PR #4263, opened 2025-06-24
[feat] Vue-Based Rendering System for the ComfyUI Node Graph

christian-byrne:
# Vue-Based Node System

### Overview 

Adds a rendering system for nodes, widgets, slots, and links in Vue 3 using PrimeVue components. PrimeVue prop APIs are exposed to Python via the [v3 Custom Node Schema](https://blog.comfy.org/p/dependency-resolution-and-custom).

This marks the beginning of decoupling the layout from the graph data model, with both now migrating into a centralized CRDT-backed state system.

Rendering design draws from xyflow, tldraw, rete.js, and Figma, emphasizing GPU compositing, minimizing layerization, LOD (level of detail) management, repaint throttling, and viewport culling. Following xyflow/tldraw patterns, zoom and pan occur within a single transform container. Nodes temporarily get their own translation layers during interaction, then collapse back into a single repaint/de-layerization pass once settled.

With the centralized state, graph observation begins its shifts from a poll/diff model on serialized interactions to push-based reactivity. Spatial queries for culling and interaction now leverage a non-authoritative QuadTree derived from the central state, enabling O(log n) lookups.

Takeaway for Users:

- Performance testing shows ~20% paint speed reduction and ~50% heap increase.
- Practical graph limit: ~1,000 nodes before noticeable degradation (prior was probably around 10k).
- Re-emphasized requirement of ~75 MB reserved VRAM and browser hardware acceleration (same as prior versions).
- Feature is off by default; enable via Settings → Keybindings → Search “Vue Nodes” → Add Keybinding → Close → Press Keybind → Refresh.

### TODO (Next Development Phases)

- Interaction System
- Link Management
- Improved LOD
- Compatibility Layers
- Fix hit-testing and teleport quirks with some PrimeVue components

### Architecture Overview 

```mermaid
graph TB
    subgraph "Vue Node Rendering Architecture"
        Canvas[LiteGraph Canvas]
        TransformPane[TransformPane Container]
        NodeManager[useGraphNodeManager]
        SpatialIndex[QuadTree Spatial Index]
        
        subgraph "Node Components"
            LGraphNode[LGraphNode.vue]
            NodeHeader[NodeHeader.vue] 
            NodeSlots[NodeSlots.vue]
            NodeWidgets[NodeWidgets.vue]
            NodeContent[NodeContent.vue]
        end
        
        subgraph "Widget System (15 Components)"
            direction TB
            InputText[Input Text]
            Textarea[Textarea]
            Slider[Slider]
            Toggle[Toggle Switch]
            Select[Select Dropdown]
            MultiSelect[Multi Select]
            SelectButton[Select Button]
            TreeSelect[Tree Select]
            ColorPicker[Color Picker]
            Image[Image Display]
            ImageCompare[Image Compare]
            Galleria[Image Gallery]
            Chart[Chart Display]
            Button[Action Button]
            FileUpload[File Upload]
        end
        
        subgraph "Performance Systems"
            LOD[Level of Detail System]
            Culling[Viewport Culling]
            RAF[RAF Batching]
            Memory[Memory Management]
        end
    end
    
    Canvas --> NodeManager
    NodeManager --> TransformPane
    NodeManager --> SpatialIndex
    TransformPane --> LGraphNode
    
    LGraphNode --> NodeHeader
    LGraphNode --> NodeSlots  
    LGraphNode --> NodeWidgets
    LGraphNode --> NodeContent
    
    NodeWidgets --> InputText
    NodeWidgets --> Select
    NodeWidgets --> Slider
    NodeWidgets --> ColorPicker
    NodeWidgets --> Button
    
    SpatialIndex --> Culling
    LOD --> NodeWidgets
    RAF --> TransformPane
    Memory --> NodeManager
    
    classDef core fill:#2563eb,stroke:#1d4ed8,color:#ffffff
    classDef component fill:#059669,stroke:#047857,color:#ffffff
    classDef widget fill:#7c3aed,stroke:#6d28d9,color:#ffffff
    classDef perf fill:#dc2626,stroke:#b91c1c,color:#ffffff
    
    class Canvas,TransformPane,NodeManager,SpatialIndex core
    class LGraphNode,NodeHeader,NodeSlots,NodeWidgets,NodeContent component
    class InputText,Textarea,Slider,Toggle,Select,MultiSelect,SelectButton,TreeSelect,ColorPicker,Image,ImageCompare,Galleria,Chart,Button,FileUpload widget
    class LOD,Culling,RAF,Memory perf
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4263-feat-Vue-widget-system-foundation-21c6d73d365081c29d32dc71efef3ddb) by [Unito](https://www.unito.io)


benceruleanlu:
template cards are dynamically sized and responsive is a failing test that can be updated IMO

image

src/assets/css/style.css
```diff
@@ -7,6 +7,66 @@
...
+@layer tailwind-utilities {
+  /* Set default values to prevent some styles from not working properly. */
+  *, ::before, ::after {
+    --tw-border-spacing-x: 0;
+    --tw-border-spacing-y: 0;
+    --tw-translate-x: 0;
+    --tw-translate-y: 0;
+    --tw-rotate: 0;
+    --tw-skew-x: 0;
+    --tw-skew-y: 0;
```
StoneCypher, line 19: i worry that setting these may cause tailwind to add some properties which can seriously slow rendering down.  is this just a reset, or do all of these properties have known causes?

i say this because i see we're fastidiously disabling drop shadows and text effects and so forth; 3d can cause expensive rendering

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -0,0 +1,813 @@
...
+
+    if (positionUpdates > 0 || sizeUpdates > 0) {
+      performanceMetrics.rafUpdateCount++
+    }
+  }
+
+  /**
+   * Main RAF change detection function
+   */
+  const detectChangesInRAF = () => {
```
StoneCypher, line 562: surprised this isn't just a dirty draw flag

src/extensions/core/groupNodeManage.ts
```diff
@@ -121,7 +121,7 @@ export class ManageGroupDialog extends ComfyDialog<HTMLDialogElement> {
   getGroupData() {
     this.groupNodeType = LiteGraph.registered_node_types[
       `${PREFIX}${SEPARATOR}` + this.selectedGroup
-    ] as LGraphNodeConstructor<LGraphNode>
+    ] as unknown as LGraphNodeConstructor<LGraphNode>
```
StoneCypher, line 124: why hard cast the type checker away?

(to be fair, as this was six months ago, i don't think anyone's going to remember the answer.  but i was asked to review this, so...)

src/lib/litegraph/src/widgets/SelectButtonWidget.ts
```diff
@@ -0,0 +1,50 @@
...
+    ctx.fillStyle = this.background_color
+    ctx.fillRect(15, y, width - 30, height)
+
+    ctx.strokeStyle = this.outline_color
+    ctx.strokeRect(15, y, width - 30, height)
+
+    ctx.fillStyle = this.text_color
+    ctx.font = '11px monospace'
+    ctx.textAlign = 'center'
+    ctx.textBaseline = 'middle'
```
StoneCypher, line 33: i keep seeing this over and over in every widget.  it seems like it ought to be extracted

src/renderer/core/layout/slots/useDomSlotRegistration.ts
```diff
@@ -0,0 +1,229 @@
+/**
```
StoneCypher, line 1: So, i think this approach might be a little bit performance problematic

Fundamentally, in order to do a render, you need to go query where things are.  Sure, those locations are cached, but you still have to query (and maintain) the cache.

What I'm saying here might be nonsense.  I'm good at React and Angular, but I'm new to Vue.

In both React and Angular, you have several performance concerns regarding the reconciliation algorithm.  One of them is that the value passed into the control needs to be same-by-identity - that is, in c speech, it needs to live at the same pointer address.  If you pass in a different structure with the same shape and values, you take an enormous performance penalty.  Pulling these locations from the cache guarantees that it will be a different structure every single time.

I made a quick throwaway draft renderer in react, and I made a simple test that compares in-place prop changes to prop replacement.  On a stock macbook, it can tolerate several million active nodes on in place prop changes, but only a few hundred thousand on prop replacement.  If the dataset and the DOM control disagree, I guess my opinion is that the dataset should win.

From that viewpoint, the workload of managing the locations of the nodes is basically free given respect to what the DOM library is already doing otherwise.

Fundamentally, I also sort of look at this differently.  I don't want to ask a control where it lives; I want to just tell it where it lives.  Having the position be managed as part of an exterior dataset rather than interior to every control is going to remove an enormous amount of traversal, and would probably speed this up quite a bit for a relatively low cost.

Not a big deal; this merged six months ago.  Was just asked to give commentary

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+  private yreroutes: Y.Map<Y.Map<unknown>> // Maps rerouteId -> Y.Map containing reroute data
+  private yoperations: Y.Array<LayoutOperation> // Operation log
+
+  // Vue reactivity layer
+  private version = 0
+  private currentSource: LayoutSource =
+    ACTOR_CONFIG.DEFAULT_SOURCE as LayoutSource
+  private currentActor = `${ACTOR_CONFIG.USER_PREFIX}${Math.random()
+    .toString(36)
+    .substring(2, 2 + ACTOR_CONFIG.ID_LENGTH)}`
```
StoneCypher, line 66: if we use a uuid we can quasi-guarantee non-collision

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+      nodeIds: [],
+      timestamp: operation.timestamp,
+      source: operation.source,
+      operation
+    }
+
+    // Use Yjs transaction for atomic updates
+    this.ydoc.transact(() => {
+      // Add operation to log
+      this.yoperations.push([operation])
```
StoneCypher, line 746: okay so i'm not super confident of this.  i haven't touched y in five years, so maybe the way i think about it is out of date, or maybe i'm misremembering

but i think this is an unbounded change log?  and i think that's fundamentally a memory leak that's going to end up crashing the browser tab

i think ynodes, ylinks, and yreroutes are also leaking

probably two steps would fix this: find a good time to clear it (switching the workflow seems like a good time), and cap the size when you're in-workflow. 

if i'm reading this correctly, clearing is straightforward, and approximately this:

```ts
this.ydoc.transact(() => {
  this.ynodes.clear();
  this.ylinks.clear();
  this.yreroutes.clear();
  this.yoperations.delete(0, this.yoperations.length);
  // ... what's already there
});
```

as far as capping the operational transformation log, y isn't screamingly expensive, and we do still want sync to be easy.  i feel like maybe capping at five thousand might be a good place to start?

i thiiiiiink it's this but i don't have testing set up yet (yay first day)

```ts
const MAX_OPTRAN_OPERATIONS = 1000;
this.yoperations.push([operation]);
if (this.yoperations.length > MAX_OPTRAN_OPERATIONS) {
  this.yoperations.delete(0, this.yoperations.length - MAX_OPTRAN_OPERATIONS);
}
```

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+  }
+
+  /**
+   * Initialize store with existing nodes
+   */
+  initializeFromLiteGraph(
+    nodes: Array<{ id: string; pos: [number, number]; size: [number, number] }>
+  ): void {
+    this.ydoc.transact(() => {
+      this.ynodes.clear()
```
StoneCypher, line 860: i think this also needs to clear ylinks, yoperations, and yreroutes

otherwise you're going to have stale data from a prior workflow and the same three boundless growths mentioned prior

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+class LayoutStoreImpl implements LayoutStore {
+  // Yjs document and shared data structures
+  private ydoc = new Y.Doc()
+  private ynodes: Y.Map<Y.Map<unknown>> // Maps nodeId -> Y.Map containing NodeLayout data
+  private ylinks: Y.Map<Y.Map<unknown>> // Maps linkId -> Y.Map containing link data
+  private yreroutes: Y.Map<Y.Map<unknown>> // Maps rerouteId -> Y.Map containing reroute data
+  private yoperations: Y.Array<LayoutOperation> // Operation log
+
+  // Vue reactivity layer
+  private version = 0
```
StoneCypher, line 61: this should be a `ref`.  it's not getting recomputed.  it's coincidentally working because the computed stuff is being replaced every call, but that will change and then this will break

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+
+  private yNodeToLayout(ynode: Y.Map<unknown>): NodeLayout {
+    return {
+      id: ynode.get('id') as string,
+      position: ynode.get('position') as Point,
+      size: ynode.get('size') as { width: number; height: number },
+      zIndex: ynode.get('zIndex') as number,
+      visible: ynode.get('visible') as boolean,
+      bounds: ynode.get('bounds') as Bounds
+    }
```
StoneCypher, line 1278: during drag this is going to generate and discard hundreds of frames at full screen framerate.  if we can flyweight them to prevent garbage collector allocation thrash, that'll improve performance a lot.

on the other hand, if the position is being pushed down from the top, this can just go away wholesale, which is even faster

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+   * Get current version for change detection
+   */
+  getVersion(): ComputedRef<number> {
+    return computed(() => this.version)
+  }
+
+  /**
+   * Query node at point (non-reactive for performance)
+   */
+  queryNodeAtPoint(point: Point): NodeId | null {
```
StoneCypher, line 290: i don't understand why this is full scanning the nodelist when there's a quadtree available

maybe i'm missing something though

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+  onChange(callback: (change: LayoutChange) => void): () => void {
+    this.changeListeners.add(callback)
+    return () => this.changeListeners.delete(callback)
+  }
+
+  /**
+   * Set the current operation source
+   */
+  setSource(source: LayoutSource): void {
+    this.currentSource = source
```
StoneCypher, line 829: this is global state and several systems are calling it in parallel.  this sort of structure often leads to bugs, though i haven't really studied this one enough to be confident

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+  }
+
+  /**
+   * Apply operation within a transaction
+   */
+  private applyOperationInTransaction(
+    operation: LayoutOperation,
+    change: LayoutChange
+  ): void {
+    switch (operation.type) {
```
StoneCypher, line 763: it would be nice if this and similar things had an `never any` failure guard.  eventually this list will change and the compiler should be defending us

src/renderer/extensions/vueNodes/components/widgets/LOD_IMPLEMENTATION_GUIDE.md
```diff
@@ -0,0 +1,295 @@
...
+Level of Detail is a technique used to optimize performance by showing different amounts of detail based on how zoomed in the user is. Think of it like Google Maps - when you're zoomed out looking at the whole country, you only see major cities and highways. When you zoom in close, you see street names, building details, and restaurants.
+
+For ComfyUI nodes, this means:
+- **Zoomed out** (viewing many nodes): Show only essential controls, hide labels and descriptions
+- **Zoomed in** (focusing on specific nodes): Show all details, labels, help text, and visual polish
+
+## Why LOD Matters
+
+Without LOD optimization:
+- 1000+ nodes with full detail = browser lag and poor performance
```
StoneCypher, line 14: this number is very surprisingly low to me.  a [naive react implementation](https://github.com/StoneCypher/comfyui_cached), ([try](https://stonecypher.github.io/comfyui_cached/proto.html)), (I need your GH user to hand out access) scales up into the millions easily.  granted my nodes are trivial, but i don't think that'd be the source of a performance problem like that.

thousands is not a lot for a modern machine.  this should be setting off skepticism alarms.

i did run one react implementation through a standard flux / stores centralization mindset, and it barely eked out 50,000.  i think going pure topdown may eliminate a massive traversal for you.  it does in react, at least.  (i'm not good at vue yet; give me a few days.)

hooks, in bulk, are performance murder in react.  i suspect the same to be true in vue.

i want to look into this some more.

src/renderer/extensions/vueNodes/widgets/components/WidgetSlider.vue
```diff
@@ -0,0 +1,170 @@
...
+})
+
+// Format a number according to the widget's precision
+const formatNumber = (value: number): string => {
+  if (precision.value === undefined) {
+    // No precision specified, return as-is
+    return String(value)
+  }
+  // Use toFixed to ensure correct decimal places
+  return value.toFixed(precision.value)
```
StoneCypher, line 101: given that the ui is internationalized, i think this should probably be `.toLocaleString({ maximumFractionDigits: prevision.value })`

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -0,0 +1,271 @@
+<template>
+  <div v-if="renderError" class="node-error p-2 text-red-500 text-sm">
+    {{ $t('Node Render Error') }}
+  </div>
+  <div
+    v-else
+    :data-node-id="nodeData.id"
+    :class="
+      cn(
+        'bg-white dark-theme:bg-[#15161A]',
```
StoneCypher, line 10: hard coded colors seem out of place

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -0,0 +1,813 @@
...
+  }
+
+  /**
+   * Sets up event listeners - now simplified with extracted handlers
+   */
+  const setupEventListeners = (): (() => void) => {
+    // Store original callbacks
+    const originalOnNodeAdded = graph.onNodeAdded
+    const originalOnNodeRemoved = graph.onNodeRemoved
+    const originalOnTrigger = graph.onTrigger
```
StoneCypher, line 716: three places in this pr, we do a swap and return pattern for this trigger handler.  thing is, it looks like their cleanup order is undefined, meaning a null handler could get swapped back in

this should be replaced with a stack imo

src/renderer/core/layout/sync/useSlotLayoutSync.ts
```diff
@@ -0,0 +1,163 @@
...
+        if (node) {
+          computeAndRegisterSlots(node)
+        }
+      }
+    })
+
+    // LiteGraph event hooks
+    const origNodeAdded = graph.onNodeAdded
+    const origNodeRemoved = graph.onNodeRemoved
+    const origTrigger = graph.onTrigger
```
StoneCypher, line 90: part two of [defect described here](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4263/changes#diff-f980db6f42cef913c3fe92669783b255d617e40b9ccef9a1ab9cc8e326ff1790R716)

src/renderer/core/layout/sync/useLinkLayoutSync.ts
```diff
@@ -0,0 +1,365 @@
...
+          }
+          break
+        case 'moveReroute':
+          recomputeLinksForReroute(change.operation.rerouteId)
+          break
+      }
+    })
+
+    // Hook collapse events
+    const origTrigger = graph.onTrigger
```
StoneCypher, line 313: part three of [defect described here](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4263/changes#diff-f980db6f42cef913c3fe92669783b255d617e40b9ccef9a1ab9cc8e326ff1790R716)

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -0,0 +1,813 @@
...
+      originalOnNodeRemoved || undefined,
+      originalOnTrigger || undefined
+    )
+  }
+
+  // Set up event listeners immediately
+  const cleanup = setupEventListeners()
+
+  // Process any existing nodes after event listeners are set up
+  if (graph._nodes && graph._nodes.length > 0) {
```
StoneCypher, line 783: these were already processed by `GraphCanvas.vue` when it called `layoutStore.initializeFromLiteGraph(nodes)`, weren't they?

i believe this is redundant

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -0,0 +1,813 @@
...
+
+    // Add/update existing nodes
+    graph._nodes.forEach((node) => {
+      const id = String(node.id)
+
+      // Store non-reactive reference
+      nodeRefs.set(id, node)
+
+      // Set up widget callbacks BEFORE extracting data (critical order)
+      setupNodeWidgetCallbacks(node)
```
StoneCypher, line 429: i think this is making a redundant wrapper for every node, on every flush (so, they're set for 4ms, but it's really hard to get something to fire that fast; they're probably triggering at the screenframe rate.)  i think this should only be done once at setup, rather than every frame.

this is probably a nontrivial memory leak, because it also prevents those callbacks from being garbage collected when their associated node vanishes

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+    // This is needed because Yjs observers don't fire for property changes
+    change.nodeIds.forEach((nodeId) => {
+      const trigger = this.nodeTriggers.get(nodeId)
+      if (trigger) {
+        trigger()
+      }
+    })
+
+    // Notify listeners (after transaction completes)
+    setTimeout(() => this.notifyChange(change), 0)
```
StoneCypher, line 814: the ref will update synchronously, but the change listener fires asynchronously.  this means that after the ref, later actions and watchers will see new data, but onChange listeners will see old data.  Given that drag is sending in hundreds of operations per screenframe, this will actually cause problems in the real world.

i assume that the `, 0)` is to allow other things to complete before this does.  in particular, i think this is about y.js updates.  the thing is, this is already after the `transact()` block, so that's unnecessary.

just calling this synchronously, without the `setTimeout`, would fix the bug.

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -0,0 +1,107 @@
...
+  <div v-if="renderError" class="node-error p-1 text-red-500 text-xs">⚠️</div>
+  <div
+    v-else
+    class="lg-slot lg-slot--input flex items-center cursor-crosshair group rounded-r-lg"
+    :class="{
+      'opacity-70': readonly,
+      'lg-slot--connected': connected,
+      'lg-slot--compatible': compatible,
+      'lg-slot--dot-only': dotOnly,
+      'pr-6 hover:bg-black/5 hover:dark:bg-white/5': !dotOnly
```
StoneCypher, line 11: the claudefile makes a point to say "never use dark, only dark-theme", so I assume this is a bug

src/renderer/extensions/vueNodes/components/OutputSlot.vue
```diff
@@ -0,0 +1,106 @@
...
+  <div v-if="renderError" class="node-error p-1 text-red-500 text-xs">⚠️</div>
+  <div
+    v-else
+    class="lg-slot lg-slot--output flex items-center cursor-crosshair justify-end group rounded-l-lg"
+    :class="{
+      'opacity-70': readonly,
+      'lg-slot--connected': connected,
+      'lg-slot--compatible': compatible,
+      'lg-slot--dot-only': dotOnly,
+      'pl-6 hover:bg-black/5 hover:dark:bg-white/5': !dotOnly,
```
StoneCypher, line 11: the claudefile makes a point to say "never use dark, only dark-theme", so I assume this is a bug

src/renderer/extensions/vueNodes/widgets/components/WidgetSlider.vue
```diff
@@ -0,0 +1,170 @@
...
+        @update:model-value="onChange"
+      />
+      <InputText
+        v-model="inputDisplayValue"
+        :disabled="readonly"
+        type="number"
+        :min="widget.options?.min"
+        :max="widget.options?.max"
+        :step="stepValue"
+        class="w-[4em] text-center text-xs px-0 !border-none !shadow-none !bg-transparent"
```
StoneCypher, line 22: claude.md says to never use the `!` prefix in vue, so i assume this is three bugs

src/utils/spatial/QuadTree.ts
```diff
@@ -0,0 +1,302 @@
...
+    const success = this.root.remove(item)
+    if (success) {
+      this.itemMap.delete(id)
+    }
+    return success
+  }
+
+  update(id: string, newBounds: Bounds): boolean {
+    const item = this.itemMap.get(id)
+    if (!item) return false
```
StoneCypher, line 269: i think that if you're dragging an item and you drag it outside of stage bounds, this means it just disappears forever, and all the data is lost and all tree queries just won't find it

src/utils/spatial/QuadTree.ts
```diff
@@ -0,0 +1,302 @@
...
+      }
+    }
+  }
+
+  private contains(itemBounds: Bounds): boolean {
+    return (
+      itemBounds.x >= this.bounds.x &&
+      itemBounds.y >= this.bounds.y &&
+      itemBounds.x + itemBounds.width <= this.bounds.x + this.bounds.width &&
+      itemBounds.y + itemBounds.height <= this.bounds.y + this.bounds.height
```
StoneCypher, line 192: this is `[ ]` inclusive inclusive, but `boundsIntersect` is `[ )` inclusive exclusive.  this means that there's a single pixel where these two disagree.  i ... don't think it causes any problems?  but it's gross.  

it does mean that a node could be inserted then never found by quadtree search, but the regional specificity to make that happen is quite unlikely

`boundsIntersect`'s inclusive exclusive is the preferable structure.

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+  ): void {
+    const ynode = this.ynodes.get(operation.nodeId)
+    if (!ynode) {
+      return
+    }
+
+    const size = ynode.get('size') as { width: number; height: number }
+    const newBounds = {
+      x: operation.position.x,
+      y: operation.position.y,
```
StoneCypher, line 908: calls to operation can return undefined if the far client wrote something that isn't fully pulled over in the operational transform (it can ship partial versions of objects which may be missing the properties you expect)

these need to be `?.` or the creation of `newBounds` could throw

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+    operation: ResizeNodeOperation,
+    change: LayoutChange
+  ): void {
+    const ynode = this.ynodes.get(operation.nodeId)
+    if (!ynode) return
+
+    const position = ynode.get('position') as Point
+    const newBounds = {
+      x: position.x,
+      y: position.y,
```
StoneCypher, line 937: same as at 908, these need to be `?.`

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -0,0 +1,1356 @@
...
+    linkData.set('sourceNodeId', operation.sourceNodeId)
+    linkData.set('sourceSlot', operation.sourceSlot)
+    linkData.set('targetNodeId', operation.targetNodeId)
+    linkData.set('targetSlot', operation.targetSlot)
+
+    this.ylinks.set(String(operation.linkId), linkData)
+
+    // Link geometry will be computed separately when nodes move
+    // This just tracks that the link exists
+    change.type = 'create'
```
StoneCypher, line 1028: i think this needs to populate `change.nodeIds`

src/renderer/extensions/vueNodes/layout/useNodeLayout.ts
```diff
@@ -0,0 +1,168 @@
...
+    // Capture pointer
+    const target = event.target as HTMLElement
+    target.setPointerCapture(event.pointerId)
+  }
+
+  /**
+   * Handle drag movement
+   */
+  const handleDrag = (event: PointerEvent) => {
+    if (!isDragging || !dragStartPos || !dragStartMouse || !transformState) {
```
StoneCypher, line 80: should transformState be in this list?  It can be undefined, and then dragging won't work

src/renderer/extensions/vueNodes/layout/useNodeLayout.ts
```diff
@@ -0,0 +1,168 @@
...
+
+    // Computed styles for Vue templates
+    nodeStyle: computed(() => ({
+      position: 'absolute' as const,
+      left: `${position.value.x}px`,
+      top: `${position.value.y}px`,
+      width: `${size.value.width}px`,
+      height: `${size.value.height}px`,
+      zIndex: zIndex.value,
+      cursor: isDragging ? 'grabbing' : 'grab'
```
StoneCypher, line 165: this needs to be a `ref` or it won't update, i think

but i'm new to vue so that might be wrong

StoneCypher (commented):
hi, i know this was merged months ago, but i was asked to leave commentary on this pr as if it was new

PR #4265, opened 2025-06-25
Vue expose

jtydhr88:
re-sent and enable Vue expose for custom nodes
Also as practiced in a real vue-based project, it is necessary to expose defineStore to allow pinia.for custom nodes, which reverted in https://github.com/Comfy-Org/ComfyUI_frontend/pull/4027 before

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4265-Vue-expose-21d6d73d36508187aa89c65b7e4c87fc) by [Unito](https://www.unito.io)


src/stores/workspaceStore.ts
```diff
@@ -100,6 +100,7 @@ export const useWorkspaceStore = defineStore('workspace', () => {
     colorPalette,
     dialog,
     bottomPanel,
+    defineStore,
```
webfiltered: I think the best way to ensure this isn't an issue is simply to namespace it.  We can wrap defineStore and simply prefix the store names.

It would be nice if this were transparent, but that doesn't seem feasible.  Unless pinia has its own namespacing API?

webfiltered (changes requested):
Just addressed the direct defineStore expose - can come back and review properly once we have a solution for that!

src/stores/workspaceStore.ts
jtydhr88: I think we could use "project-name" as namepace as while implemented this feature https://github.com/comfyanonymous/ComfyUI/pull/8357 and it is already merged

webfiltered: Definitely!  Would be good to also ensure extensions are namespaced in a way that prevents clashing with frontend - e.g. `extension:` prefix or the like.  That's the part where I was really hoping for some kind of namespace API from pinia (but we can just use a string if not).

jtydhr88: @webfiltered I remove this line for now, lets focus on Vue self expose in this PR.
we could do defineStore expose in later PR as it needs some additional work

christian-byrne (approved):
The past bugs have been confirmed resolved. We should merge this so we can get some coverage and see if any new bugs exist.

This PR is necessary for various things such as vue extensions.

PR #4283, opened 2025-06-27
[bugfix] Handle backend error messages with appended content

christian-byrne:
Updates error detection to use .includes() instead of exact string matching, allowing proper handling of auth/payment errors when backend appends additional information like OOM tips.

Fixes comfyanonymous/ComfyUI#8574

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4283-bugfix-Handle-backend-error-messages-with-appended-content-21f6d73d365081bd84b2e94c158d19d0) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -667,15 +667,18 @@ export class ComfyApp {
     })
 
     api.addEventListener('execution_error', ({ detail }) => {
+      console.log('execution_error', detail)
       // Check if this is an auth-related error or credits-related error
       if (
-        detail.exception_message ===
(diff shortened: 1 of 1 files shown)


webfiltered (approved):
LGTM either way, but does `startsWith` make more sense?

christian-byrne:
Wanted to be more future-proof, incase there's more formatting changes in the future to error messages (e.g., wrapping in decoration, adding timestamp, etc.). The one thing that won't change is the actual message set in the api nodes module, so this seems the most future-proof. Conversely, the potential downside is quite low.

PR #4288, opened 2025-06-28
Add API usage log table

jtydhr88:
added usage log table inside credit page:
image
please notice, some code come from existing website repo

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4288-usage-log-table-2206d73d365081c6b8d4c104eb5c6857) by [Unito](https://www.unito.io)


src/services/customerService.ts
```diff
@@ -0,0 +1,51 @@
...
+  | 'api_usage_started'
+  | 'api_usage_completed'
+
+export interface AuditLog {
+  event_type: AuditLogType
+  event_id: string
+  params: Record<string, any> // Using Record since it's additionalProperties: true
+  createdAt: string
+}
+
```
christian-byrne: Types that come from the comfy-api are auto-generated from the openapi spec. See:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/7620bb9063093ab2c70c1f68d81ac1f4652a8900/src/types/comfyRegistryTypes.ts#L3399-L3413

src/services/customerService.ts
```diff
@@ -0,0 +1,51 @@
...
+interface CustomerEventsResponse {
+  events: AuditLog[]
+  total: number
+  page: number
+  limit: number
+  totalPages: number
+}
```
christian-byrne: Same as above:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/7620bb9063093ab2c70c1f68d81ac1f4652a8900/src/types/comfyRegistryTypes.ts#L9806-L9818

src/services/customerService.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export async function getMyEvents({
+  page = 1,
+  limit = 10
+}: {
+  page?: number
+  limit?: number
+} = {}): Promise<CustomerEventsResponse> {
```
christian-byrne: The arguments can be typed using the generated types as well

https://github.com/Comfy-Org/ComfyUI_frontend/blob/7620bb9063093ab2c70c1f68d81ac1f4652a8900/src/types/comfyRegistryTypes.ts#L9785-L9792

src/services/customerService.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  const response = await fetch(url.toString(), {
+    headers
+  })
+
+  if (!response.ok) {
+    throw new Error(`Failed to fetch events: ${response.status}`)
+  }
+
+  return await response.json()
```
christian-byrne: Can we structure it following the patterns of the other services? See https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/services/README.md

E.g.

```typescript
import axios, { AxiosError } from 'axios'
import { ref } from 'vue'

const customerApiClient = axios.create({
   baseURL: COMFY_API_BASE_URL,
   headers: {
      'Content-Type': 'application/json'
   }
})

export const useCustomerService = () => {
   const isLoading = ref(false)
   const error = ref<string | null>(null)
   
   // Follow existing patterns...
}
```

Alternatively, we can use a composable similar to the composables in the `src/composables/auth` folder.

Ideally, usage of the auth headers should be centralized.

src/components/dialog/content/setting/UsageLogsTable.vue
```diff
@@ -0,0 +1,236 @@
...
+  }
+}
+
+const formatAmount = (amountMicros?: number) => {
+  if (!amountMicros) return '0.00'
+  return (amountMicros / 100).toFixed(2)
+}
+
+const formatDate = (dateString: string) => {
+  return new Date(dateString).toLocaleString(undefined, {
```
christian-byrne: Do we want to use the function provided by i18n for date formatting? It has some additional benefits and is the established pattern thus far in the codebase.

src/components/dialog/content/setting/UsageLogsTable.vue
```diff
@@ -0,0 +1,236 @@
...
+            <template v-if="data.event_type === 'credit_added'">
+              <div class="text-green-500 font-semibold">
+                {{ $t('credits.added') }} ${{
+                  formatAmount(data.params?.amount)
+                }}
+              </div>
+            </template>
+
+            <!-- Account Created -->
+            <template v-else-if="data.event_type === 'account_created'">
```
christian-byrne: Can we create an enum for the event types?

src/components/dialog/content/setting/UsageLogsTable.vue
christian-byrne, line 1: I would suggest moving more of the internal state management logic into a composable. If we extract the API logic into a composable, it's easier to test the logic and reuse it in other components.

In general, we have an established style of separating business logic and view/presentation logic. This might apply to `formatEventType`, `getEventSeverity`, `formatAmount`, and so on.

Also, I think we want to persist the customer events in a store while also ensuring there are not direct API calls in components -- so as to avoid tight coupling between component and API layers.

src/components/dialog/content/setting/UsageLogsTable.vue
```diff
@@ -0,0 +1,236 @@
...
+    <DataTable
+      v-else
+      :value="events"
+      :paginator="true"
+      :rows="pagination.limit"
+      :total-records="pagination.total"
+      :first="dataTableFirst"
+      :lazy="true"
+      class="p-datatable-sm custom-datatable"
+      responsive-layout="scroll"
```
christian-byrne: I don't see `responsive-layout` in the API docs for DataTable on our version of PrimeVue. Is this deprecated?

https://primevue.org/datatable/#api.datatable

For reference, we are using PrimeVue 4.2.5

src/components/dialog/content/setting/UsageLogsTable.vue
```diff
@@ -0,0 +1,236 @@
...
+          {{ formatDate(data.createdAt) }}
+        </template>
+      </Column>
+      <Column field="params" :header="$t('credits.additionalInfo')">
+        <template #body="{ data }">
+          <Button
+            v-if="hasAdditionalInfo(data)"
+            v-tooltip.top="{
+              escape: false,
+              value: getTooltipContent(data),
```
christian-byrne: I think making `getTooltipContent(data)` computed is best practice -- to avoid recalculating on every render and making the code more testable.

src/components/dialog/content/setting/UsageLogsTable.vue
```diff
@@ -0,0 +1,236 @@
...
+  })
+}
+
+const hasAdditionalInfo = (event: AuditLog) => {
+  const { amount, api_name, model, ...otherParams } = event.params || {}
+  return Object.keys(otherParams).length > 0
+}
+
+const getTooltipContent = (event: AuditLog) => {
+  const { amount, api_name, model, ...otherParams } = event.params || {}
```
christian-byrne: Why are `amount`, `api_name`, and `model` explicitly excluded from additional info? Consider documenting the reasoning or making it configurable. To me, it's not obvious just by reading the code.

christian-byrne (changes requested):

src/services/customerService.ts
jtydhr88: fixed

src/services/customerService.ts
jtydhr88: fixed

src/services/customerService.ts
jtydhr88: fixed

src/services/customerService.ts
jtydhr88: fixed

src/services/customerService.ts
```diff
@@ -0,0 +1,110 @@
...
+  operations['GetCustomerEvents']['parameters']['query']
+
+const customerApiClient = axios.create({
+  baseURL: COMFY_API_BASE_URL,
+  headers: {
+    'Content-Type': 'application/json'
+  }
+})
+
+export const useCustomerService = () => {
```
christian-byrne: Maybe we should name it something like "useCustomerEventsService" or "useApiUsageService". The term `useCustomerService` is definitely confusing.

src/components/dialog/content/setting/UsageLogsTable.vue
jtydhr88: this logic is basing on existing logic **hasAdditionalInfo()** from website, as we already do this on website, I think we should keep same logic here

jtydhr88: cc @benceruleanlu

src/components/dialog/content/setting/UsageLogsTable.vue
jtydhr88: I removed it here, and checked website is using PrimeVue 4.3.3, might not have it as well and should remove it too.
cc @benceruleanlu

src/components/dialog/content/setting/UsageLogsTable.vue
jtydhr88: fixed

src/components/dialog/content/setting/UsageLogsTable.vue
jtydhr88: fixed

src/components/dialog/content/setting/UsageLogsTable.vue
jtydhr88: fixed

src/services/customerService.ts
jtydhr88: fixed

src/components/dialog/content/setting/UsageLogsTable.vue
jtydhr88: fixed

christian-byrne (changes requested):
Looks good. Can we just add tests for the service and the `UsageLogsTable.vue`?

- Examples of service  unit tests: https://github.com/Comfy-Org/ComfyUI_frontend/tree/main/tests-ui/tests/services
- Guide for writing component tests: https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/tests-ui/component-testing.md
- Example of component test: https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/components/common/TreeExplorerTreeNode.spec.ts

christian-byrne (approved):
LGTM!

webfiltered:
Tests failed once this was merged to main: https://github.com/Comfy-Org/ComfyUI_frontend/actions/runs/15981619587/job/45077170846

jtydhr88:
@webfiltered I checked the test log, I don't see any error related to this changes, it is saying something:
[chromium] › browser_tests/tests/widget.spec.ts:181:3 › Image widget › Can change image by changing the filename combo value 

    Test timeout of 15000ms exceeded.

    Error: locator.click: Test timeout of 15000ms exceeded.
    Call log:
      - waiting for getByRole('menuitem', { name: 'image32x32.webp' })


      194 |       name: 'image32x32.webp'
      195 |     })
    > 196 |     await comboEntry.click({ noWaitAfter: true })
          |                      ^
      197 |
      198 |     // Expect the image preview to change automatically
      199 |     await expect(comfyPage.canvas).toHaveScreenshot(
        at /home/runner/work/ComfyUI_frontend/ComfyUI_frontend/ComfyUI_frontend/browser_tests/tests/widget.spec.ts:196:22

christian-byrne:
For [Viewport settings › Keeps viewport settings when changing tabs](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4288#?testId=967c1c643b6ca86a362c-03d59149c6b72bf356f0), it is unrelated to this PR. The release notes will popup but the timing is naturally inconsistent in test environment

Selection_1632

christian-byrne:
Let me handle the test errors. They are actually all from the popup in the bottom left, which sometimes appears and sometimes does not:

  - Viewport test: Screenshot shows "ComfyUI v0.3.43 Release" popup not present in expected image
  - Workflows test: Error explicitly states `<h1>ComfyUI v0.3.43 Release</h1>` intercepts pointer events
  - Node library test: Drag-and-drop fails because popup is blocking the bottom-left area

PR #4289, opened 2025-06-28
record audio node support

jtydhr88:
Bounty task - audio input node, requested in https://www.notion.so/drip-art/Audio-input-node-1706d73d365080eeb8e5e39790584dff?source=copy_link
Need BE change here https://github.com/comfyanonymous/ComfyUI/pull/8716

Please notice I introdcued two new dependencies to support record audio:
1. extendable-media-recorder https://www.npmjs.com/package/extendable-media-recorder
2. extendable-media-recorder-wav-encoder https://www.npmjs.com/package/extendable-media-recorder-wav-encoder

the reason of why I need this two is, as practised, the native browser MediaRecorder API has significant limitations for our use case:
1. Format Incompatibility: Chrome produces WebM format, Firefox produces OGG format, even when requesting audio/wav
2. Backend Processing Issues: The generated files cannot be properly read by torchaudio.load() in our Python backend
3. Cross-browser Inconsistency: Different browsers generate different audio formats, causing compatibility issues

so we need to use this extendable-media-recorder could allow us record WAV directly.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4289-record-audio-node-support-2206d73d365081e0a40ac6ed1c353018) by [Unito](https://www.unito.io)


src/extensions/core/uploadAudio.ts
```diff
@@ -241,3 +246,146 @@ app.registerExtension({
     }
   }
 })
+
+app.registerExtension({
+  name: 'Comfy.RecordAudio',
+
+  async setup() {
+    try {
+      await register(await connect())
```
christian-byrne: I think this should be done lazily. 99.99% of users will not be using this node (at least currently), so we don't want to introduce the memory overhead and startup blocker for them if we don't have to.

christian-byrne (changes requested):
Could you add a unit test and demo recording if possible?

src/extensions/core/uploadAudio.ts
```diff
@@ -241,3 +246,146 @@ app.registerExtension({
...
+          })
+
+          if (resp.status !== 200) {
+            const err = `Error uploading temp file: ${resp.status} - ${resp.statusText}`
+            useToastStore().addAlert(err)
+            throw new Error(err)
+          }
+
+          const tempAudio = await resp.json()
+          return `threed/${tempAudio.name} [temp]`
```
guill: Isn't the `threed` folder mostly for `3D` stuff? I think maybe the default directory for this should be different.

jtydhr88:
demo is
https://github.com/user-attachments/assets/62cf2cdd-5ab0-4873-a998-6b0a52a4b5d0

src/extensions/core/uploadAudio.ts
jtydhr88: fixed

src/extensions/core/uploadAudio.ts
jtydhr88: I moved this logic into nodeCreated, which will be called only this node created

christian-byrne (commented):
Great work on overall!

Left some comments that need to be addressed before merging, primarily around resource management and cleanup.

christian-byrne (commented):
Memory leak - Old blob URLs are never revoked when creating new recordings. Each recording creates a blob URL that persists in memory.

**Suggestion**: Revoke previous blob URL before creating new one:
```typescript
if (audioUIWidget.element.src && audioUIWidget.element.src.startsWith('blob:')) {
  URL.revokeObjectURL(audioUIWidget.element.src)
}
audioUIWidget.element.src = URL.createObjectURL(audioBlob)
```

christian-byrne:
To test:

```bash
git remote add jtydhr88
https://github.com/jtydhr88/ComfyUI.git && git
fetch jtydhr88 && git checkout -b
record-audio-node jtydhr88/record-audio-node
```

src/extensions/core/uploadAudio.ts
```diff
@@ -241,3 +243,122 @@ app.registerExtension({
...
+                mediaRecorder.start()
+                isRecording = true
+                recordWidget.label = t('g.stopRecording')
+              } catch (err) {
+                console.error('Error accessing microphone:', err)
+                useToastStore().addAlert(t('g.micPermissionDenied'))
+
+                useAudioService().stopAllTracks(currentStream)
+                isRecording = false
+                recordWidget.label = t('g.startRecording')
```
christian-byrne: Error handling doesn't properly clean up media stream on all error paths. Media stream might remain active if errors occur, keeping microphone access indicator on.

**Suggestion**: Ensure cleanup happens in all error cases:
```typescript
} catch (err) {
  console.error("Error accessing microphone:", err)
  useToastStore().addAlert(t("g.micPermissionDenied"))
  
  // Cleanup any partially initialized resources
  if (mediaRecorder) {
    try { mediaRecorder.stop() } catch {}
  }
  useAudioService().stopAllTracks(currentStream)
  currentStream = null
  isRecording = false
  recordWidget.label = t("g.startRecording")
}

File: src/extensions/core/uploadAudio.ts [hunk 4]
```diff
@@ -241,3 +243,122 @@ app.registerExtension({
...
+                mediaRecorder.ondataavailable = (event) => {
+                  audioChunks.push(event.data)
+                }
+
+                mediaRecorder.onstop = async () => {
+                  const audioBlob = new Blob(audioChunks, { type: 'audio/wav' })
+
+                  useAudioService().stopAllTracks(currentStream)
+
+                  audioUIWidget.element.src = URL.createObjectURL(audioBlob)
```
christian-byrne, line 345: **Memory Leak**: Old blob URLs are never revoked when creating new recordings. Each recording creates a blob URL that persists in memory.

**Suggestion**: Revoke previous blob URL before creating new one:
```typescript
if (audioUIWidget.element.src && audioUIWidget.element.src.startsWith("blob:")) {
  URL.revokeObjectURL(audioUIWidget.element.src)
}
audioUIWidget.element.src = URL.createObjectURL(audioBlob)
```

src/extensions/core/uploadAudio.ts
christian-byrne, line 344: **Incomplete Error Cleanup**: Media stream might remain active if errors occur, keeping microphone access indicator on.

**Suggestion**: Ensure cleanup happens in all error cases:
```typescript
} catch (err) {
  console.error("Error accessing microphone:", err)
  useToastStore().addAlert(t("g.micPermissionDenied"))
  
  // Cleanup any partially initialized resources
  if (mediaRecorder) {
    try { mediaRecorder.stop() } catch {}
  }
  useAudioService().stopAllTracks(currentStream)
  currentStream = null
  isRecording = false
  recordWidget.label = t("g.startRecording")
}

File: src/extensions/core/uploadAudio.ts [hunk 5]
```diff
@@ -241,3 +243,122 @@ app.registerExtension({
...
+        audio.setAttribute('name', 'media')
+
+        const audioUIWidget: DOMWidget<HTMLAudioElement, string> =
+          node.addDOMWidget(inputName, /* name=*/ 'audioUI', audio)
+
+        let mediaRecorder: MediaRecorder | null = null
+        let isRecording = false
+        let audioChunks: Blob[] = []
+        let currentStream: MediaStream | null = null
+        let recordWidget: any = null
```
christian-byrne: **Type Safety**: Using `any` type for recordWidget reduces type safety.

**Suggestion**: Define proper widget type:
```typescript
let recordWidget: IWidget
```

src/extensions/core/uploadAudio.ts
```diff
@@ -241,3 +243,122 @@ app.registerExtension({
...
+              }
+            } else if (mediaRecorder && isRecording) {
+              mediaRecorder.stop()
+            }
+          },
+          { serialize: false }
+        )
+
+        recordWidget.label = t('g.startRecording')
+        return { widget: recordWidget }
```
christian-byrne, line 418: **Missing Node Cleanup**: Active recordings continue even after node is deleted.

**Suggestion**: Add cleanup on node removal:
```typescript
const originalOnRemoved = node.onRemoved;
node.onRemoved = function() {
  if (isRecording && mediaRecorder) {
    mediaRecorder.stop();
  }
  useAudioService().stopAllTracks(currentStream);
  if (audioUIWidget.element.src?.startsWith("blob:")) {
    URL.revokeObjectURL(audioUIWidget.element.src);
  }
  originalOnRemoved?.call(this);
}
```

src/services/audioService.ts
```diff
@@ -0,0 +1,84 @@
...
+import { useToastStore } from '@/stores/toastStore'
+
+export interface AudioRecordingError {
+  type: 'permission' | 'not_supported' | 'encoder' | 'recording' | 'unknown'
+  message: string
+  originalError?: unknown
+}
+
+export const useAudioService = () => {
+  let isEncoderRegistered: boolean = false
```
christian-byrne: **Encoder Registration Scope**: `isEncoderRegistered` is in function scope, so each call to `useAudioService()` creates a new instance with its own variable. This means the encoder could be registered multiple times unnecessarily.

**Suggestion**: Move to module scope since WAV encoder registration is a global operation:
```typescript
// Move this outside the function
let isEncoderRegistered: boolean = false

export const useAudioService = () => {
  // Remove the let declaration from here
```

christian-byrne:
For anyone else: To test this feature, you need to do this in ComfyUI folder:

```bash
git remote add jtydhr88 https://github.com/jtydhr88/ComfyUI.git
git fetch jtydhr88
git checkout -b record-audio-node jtydhr88/record-audio-node
```

src/extensions/core/uploadAudio.ts
jtydhr88: fixed

src/extensions/core/uploadAudio.ts
jtydhr88: fixed

src/extensions/core/uploadAudio.ts
jtydhr88: fixed

src/extensions/core/uploadAudio.ts
jtydhr88: fixed

christian-byrne (commented):
Nice, this looks ready to merge. The only two minor things I would say are:

1. Is it possible to replace timeout with hard-coded 500ms with somthing that resolves on `mediaRecorder.onStop` (or similar)?
2. Could we remove the `console.log`s?

jtydhr88:
@christian-byrne fixed your feedback, please double check

jtydhr88:
there is one failed test about locale, likely not caused by this change

PR #4290, opened 2025-06-28
Add preview to workflow tabs

pythongosssss:
Adds a preview of the graph using the minimap when you hover over the tab

image

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4290-Add-preview-to-workflow-tabs-2206d73d365081a4b6fbf27edec22bed) by [Unito](https://www.unito.io)


christian-byrne:
Claude Code review

I reviewed this thoroughly and found that most potential concerns are already properly handled:

**Memory management with blob URLs** - The code correctly calls `URL.revokeObjectURL()` in `clearThumbnail()` and `clearAllThumbnails()`, properly cleaning up blob URLs when workflows are closed.

**Manual DOM positioning** - The manual positioning addresses a PrimeVue bug with scrolled tabs, as noted in the code comments. The workaround is reasonable.

**Removed tooltip** - The v-tooltip removal is intentional, replaced by a richer popover with visual preview that mirrors browser tab behavior.

**Device pixel ratio** - While DPR support could make thumbnails crisper on high-DPI displays, the current implementation works well with browser image scaling.

**CSS variables with Tailwind** - Using CSS variables for dynamic popover width that's referenced in multiple places is a practical pattern here.

The implementation is solid with good test coverage and follows Vue 3 best practices.

src/components/topbar/WorkflowTab.vue
```diff
@@ -86,6 +102,27 @@ const shouldShowStatusIndicator = computed(() => {
...
+})
+
+// Event handlers that delegate to the popover component
+const handleMouseEnter = (event: Event) => {
+  popoverRef.value?.showPopover(event)
+}
+
+const handleMouseLeave = () => {
+  popoverRef.value?.hidePopover()
+}
```
christian-byrne, line 120: Should we also call `hidePopover` during unmount?

src/composables/useWorkflowThumbnail.ts
```diff
@@ -0,0 +1,129 @@
...
+      if (sourceAspect > targetAspect) {
+        sourceWidth = graphCanvas.height * targetAspect
+        sourceX = (graphCanvas.width - sourceWidth) / 2
+      }
+      // If source is taller than target, crop vertically
+      else if (sourceAspect < targetAspect) {
+        sourceHeight = graphCanvas.width / targetAspect
+        sourceY = (graphCanvas.height - sourceHeight) / 2
+      }
+
```
christian-byrne: I wonder if there's a good way to get the low quality / zoomed-out render of litegraph before drawing.

src/composables/useWorkflowThumbnail.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  /**
+   * Store a thumbnail for a workflow
+   */
+  const storeThumbnail = async (workflow: ComfyWorkflow) => {
+    const thumbnail = await createMinimapPreview()
+    if (thumbnail) {
+      workflowThumbnails.value.set(workflow.key, thumbnail)
```
christian-byrne, line 53: I think here if a thumbnail exists already at this key, we need to revoke before updating the Map, right?

src/services/workflowService.ts
christian-byrne, line 1: Is it necessary to handle renaming by updating the thumbnails Map?

christian-byrne (approved):
Nice it works really well and looks amazing

PR #4291, opened 2025-06-28
fix(float-precision): correct float widget rounding

bigcat88:
This PR addresses a long-standing floating-point precision issue that causes float values like `0.1` to be saved incorrectly as `0.10000000000000002` in workflows.

Closes: 
 * https://github.com/comfyanonymous/ComfyUI/issues/8515
 * https://github.com/comfyanonymous/ComfyUI/issues/8658
 
 
---

When a float widget's rounding value was calculated, it was using `Math.pow(0.1, precision)`. Since `0.1` cannot be perfectly represented in binary, this introduced a tiny error that propagated into the final widget value.

The original code tried to mitigate this with a multiplication/division trick, but the error was already present before that trick could be effective.

_This PR was not fully tested, I run tests myself only on a few workflows to prove that there is no such error anymore as in the issues._

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4291-fix-float-precision-correct-float-widget-rounding-2206d73d365081cb8407f95ab34816f3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/widgets/useFloatWidget.ts
+++ b/src/composables/widgets/useFloatWidget.ts
@@ -62,7 +62,7 @@ export const useFloatWidget = () => {
         max: inputSpec.max ?? 2048,
         round:
           enableRounding && precision && !inputSpec.round
-            ? (1_000_000 * Math.pow(0.1, precision)) / 1_000_000
+            ? Math.pow(10, -precision)
             : (inputSpec.round as number),
         /** @deprecated Use step2 instead. The 10x value is a legacy implementation. */
         step: step * 10.0,
--- a/src/composables/widgets/useFloatWidget.ts
+++ b/src/composables/widgets/useFloatWidget.ts
@@ -10,14 +10,19 @@ import { type ComfyWidgetConstructorV2 } from '@/scripts/widgets'
 import { useSettingStore } from '@/stores/settingStore'
 
 function onFloatValueChange(this: INumericWidget, v: number) {
-  this.value = this.options.round
-    ? _.clamp(
-        Math.round((v + Number.EPSILON) / this.options.round) *
-          this.options.round,
-        this.options.min ?? -Infinity,
-        this.options.max ?? Infinity
-      )
-    : v
+  const round = this.options.round
+  if (round) {
+    const precision =
+      this.options.precision ?? Math.max(0, -Math.floor(Math.log10(round)))
+    const rounded = Math.round(v / round) * round
+    this.value = _.clamp(
+      Number(rounded.toFixed(precision)),
+      this.options.min ?? -Infinity,
+      this.options.max ?? Infinity
+    )
+  } else {
+    this.value = v
+  }
 }
 
 export const _for_testing = {


webfiltered (commented):
Looks great at first glance - thanks for tracking that down.  I'll give it a bit of a thrashing and see if  there's impact elsewhere.

christian-byrne (commented):
Tested the fix and confirmed it resolves the floating-point precision issue. Values like 0.1 and 1.0 now serialize correctly to JSON without the trailing digits (e.g., no more 0.10000000000000002). The mathematical approach using `Math.pow(10, -precision)` is cleaner and more accurate than the previous implementation. LGTM 👍

christian-byrne (approved):

christian-byrne (commented):
⚠️ **Found a regression**: While testing locally, I discovered that setting precision to 3 now causes 0.036 to save as `0.036000000000000004` instead of the previous `0.03600000000000001`. This is still a precision error, just a different one.

The root issue is that `Math.pow(10, -3)` returns `0.0010000000000000000208` instead of exactly `0.001`, causing floating-point precision errors at precision 3 and higher.

Should we consider a more robust approach for high precision values (3+), such as using string-based decimal arithmetic or clamping precision to safe values?

christian-byrne (commented):
🔍 **Clarification on the regression**: The precision issue occurs at **precision 3**, not 6. At precision 3,  now saves as `0.036000000000000004`.

**Test results:**
- Precision 1-2: ✅ Fixed correctly
- Precision 3: ❌ **NEW REGRESSION** ( → )
- Precision 4+: ✅ Works better than original

The root cause is  returning  instead of exactly . You can reproduce this by setting 'Float Rounding Precision' to 3 in ComfyUI settings and entering 0.036 in any float widget.

christian-byrne (commented):
🔍 **Clarification on the regression**: The precision issue occurs at **precision 3**, not 6. At precision 3, 0.036 now saves as `0.036000000000000004`.

**Test results:**
- Precision 1-2: ✅ Fixed correctly  
- Precision 3: ❌ **NEW REGRESSION** (0.036 becomes 0.036000000000000004)
- Precision 4+: ✅ Works better than original

The root cause is Math.pow(10, -3) returning an imprecise value. You can reproduce this by setting 'Float Rounding Precision' to 3 in ComfyUI settings and entering 0.036 in any float widget.

christian-byrne (commented):
Testing process:

  1. Basic Float Values in KSampler
    - Added KSampler node, set denoise to 0.1
    - Saveed workflow and checked JSON - was 0.1 not 0.10000000000000002
    - Tested other values: 0.5, 0.25, 0.75, 1.0
  2. Workflow Save/Load Cycle
    - Created workflow with multiple float values
    - Saved to file
    - Reloaded and verified all float values remain exact
  4. StyleModelApply Node Test
    - Added StyleModelApply node
    - Set strength to exactly 1.0
    - Export and verified it saves as 1 or 1.0, not 1.0000000000000002
  5. Different Precision Settings
    - Went to Settings → Comfy → Float Rounding Precision
    - Test with precision 1, 2, 3, 4, 5, 6
    - Verify rounding works correctly at each level (except with 6, serializes incorrectly after using buttons)
  6. Edge Case Values
    - Very small: 0.001, 0.0001
        - fails at precision 6 => e.g. 0.000010 serializes to 0.000009999999999999999
    - Large values: 100.1, 1000.1
    - Values that round badly in binary: 0.1, 0.2, 0.3, 0.7
        - fails on precision 6 => e.g. 0.1 serializes to 0.09999999999999999
  7. Multiple Node Types
    - Tested float widgets in different nodes:
        - LoRA nodes (model_weight, clip_weight)


Found a separate issue while testing: https://github.com/Comfy-Org/ComfyUI_frontend/issues/4350

christian-byrne:
It seems there's some new issues e.g. the serialization of `0.1` at precision 6 is now

0.09999999999999999

before it was 

0.10000000000000003

bigcat88:
> It seems there's some new issues e.g. the serialization of `0.1` at precision 6 is now

I tried setting the precision to `6` and exporting the workflow but couldn't reproduce it. Do you have any advice on how I can reproduce it?

Screenshot from 2025-07-05 03-47-57

christian-byrne:
Can you try after toggling this one?

Selection_1656

Then maybe try with using the buttons on the widget.

bigcat88:
wdyt about slightly modifying `onFloatValueChange` with adding final rounding to fix this?

I will test it, but the idea is to do this:

```js

function onFloatValueChange(this: INumericWidget, v: number) {
  let value = this.options.round
    ? _.clamp(
        Math.round((v + Number.EPSILON) / this.options.round) *
          this.options.round,
        this.options.min ?? -Infinity,
        this.options.max ?? Infinity
      )
    : v

  // THIS PART: final rounding to the desired precision to eliminate floating-point representation errors.
  if (this.options.precision != null && this.options.precision >= 0) {
    const p = Math.pow(10, this.options.precision)
    value = Math.round(value * p) / p
  }

  this.value = value
}
```

bigcat88:
Hope that this fix(use `toFixed(precision)` and convert back to `number`) is enough for these edge cases.

_Found another strange situation during testing: when we set `precision=1` and try from UI change the `denoise` it does not change at all, cause `step=0.01`. Looks weird. I have local version with demo of the fix. But that is for another PR as this is not related to the current scope of problem._


<details>
<summary>version with logging for fast check</summary>

```js
function onFloatValueChange(this: INumericWidget, v: number) {
  const round = this.options.round
  const min = this.options.min ?? -Infinity
  const max = this.options.max ?? Infinity

  if (round) {
    const precision =
      this.options.precision ?? Math.max(0, -Math.floor(Math.log10(round)))

    const adjusted = v / round
    const rounded = Math.round(adjusted) * round

    const fixedValue = Number(rounded.toFixed(precision))

    const clamped = _.clamp(fixedValue, min, max)

    console.log('[onFloatValueChange debug]')
    console.log('input value =', v)
    console.log('round =', round)
    console.log('precision =', precision)
    console.log('adjusted = v / round =', adjusted)
    console.log(
      'rounded (intermediate, possibly imprecise) = Math.round(adjusted) * round =',
      rounded
    )
    console.log(
      'fixedValue (final corrected value) = Number(rounded.toFixed(precision)) =',
      fixedValue
    )
    console.log('clamped =', clamped)

    this.value = clamped
  } else {
    console.log('[onFloatValueChange debug] no rounding, value =', v)
    this.value = v
  }
}
```
</details>

christian-byrne (approved):
Tested again and works well. The unwanting floating-point noise is successfuly removed and all edge cases are handled.

LGTM, nice work!

Issue #3996, opened 2025-05-27
[Bug]: Escape closes Settings dialog if login dialog open


webfiltered:
### Frontend Version

1.20.6

### Expected Behavior

Pressing escape closes the top-most dialog

### Actual Behavior

Pressing escape with the login dialog open never closes the login dialog - it closes whatever is behind it.

https://github.com/user-attachments/assets/2b6d418a-0b05-494f-a975-ef9871180c25

### Steps to Reproduce

1. Open settings
2. Click User
3. Click Log in / create account
4. Press Escape

### Debug Logs

```powershell
N/A
```

### Browser Logs

N/A

### Setting JSON

Default

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-3996-Bug-Escape-closes-Settings-dialog-if-login-dialog-open-2006d73d3650813ab4dfd97afb0b6dd0) by [Unito](https://www.unito.io)

robinjhuang:
This Issues has been set to be bounty, here is the link on how to sign up for this bounty: https://comfyorg.notion.site/ComfyUI-Bounty-Tasks-1fb6d73d36508064af76d05b3f35665f or [click here to sign up](mailto:bounty@comfy.org?subject=I%20can%20help%20%5B%5BBug%5D%3A%20Escape%20closes%20Settings%20dialog%20if%20login%20dialog%20open%5D&body=I%20can%20help%20with%20%5B%5BBug%5D%3A%20Escape%20closes%20Settings%20dialog%20if%20login%20dialog%20open%5D(%20https%3A%2F%2Fgithub.com%2FComfy-Org%2FComfyUI_frontend%2Fissues%2F3996%20)%0A%0AMy%20approach%20is%20...%0A%0AMy%20timeline%20is%20...)

Iflaqbhat:
@webfiltered assign it to me..

BenraouaneSoufiane:
Hello, can I work on this for 100$?

christian-byrne:
Follow Robin's instructions in the comment above please.

felmonon:
Hi! I'd like to work on this issue.

**My understanding of the bug:**
When the login dialog is open on top of the Settings dialog, pressing Escape closes the Settings dialog (behind) instead of the login dialog (front). This is a dialog z-index/focus management issue where the Escape key handler isn't properly checking which dialog is on top.

**Proposed fix:**
- Ensure the Escape key handler only closes the topmost/focused dialog
[...]

PR #4299, opened 2025-06-29
[WIP] Fix: Escape closes Settings dialog if login dialog open #3996

ssidharth010:
Fix for: https://github.com/Comfy-Org/ComfyUI_frontend/issues/3996

Problem:

The ESC button  closes the dialog opened below rather than top.

Root Cause:

Since there is not close button configured on the login modal, but on the settings modal it have close button configuration.
Prime vue dialog takes only the one first one hence closing that dialog only.
https://github.com/primefaces/primevue/issues/5138

Solution:

Current temporary solution is simply to add close button to Login modal.

Other solutions:

**CloseOnEsc** in the **DialogStore** config will always be kept false and a function can be written in the **KeyBindingService** to handle close on esc by closing only if the top most dialog has **closable** set to true

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4299-WIP-Fix-Escape-closes-Settings-dialog-if-login-dialog-open-3996-2216d73d36508193a00cee6bfe60fe50) by [Unito](https://www.unito.io)


webfiltered (changes requested):
Currently, pressing escape closes both the user and settings dialogs.

Would it be possible to have escape close the user dialog, then close settings, to bring this in line with stacked dialogs elsewhere in the app?

ssidharth010:
Currently from what i have checked with prime vue dialog doesn't have the option to configure that way.
The only options is to update our current flows to update the closeOnEsc to false for all the dialogs below the active dialog. This is the only work around that I can find at the moment.
This will also be a major change.

christian-byrne:
> The only options is to update our current flows to update the closeOnEsc to false for all the dialogs below the active dialog. This is the only work around that I can find at the moment.

In the dialogStore or dialogService, we can expose a function to close the "highest" dialog, while respecting the ranking system, then add the esc keyhandler to all dialogs that have close on escape prop enabled currently, then remove the close on escape prop. What do you think?

ssidharth010:
> > The only options is to update our current flows to update the closeOnEsc to false for all the dialogs below the active dialog. This is the only work around that I can find at the moment.
> 
> In the dialogStore or dialogService, we can expose a function to close the "highest" dialog, while respecting the ranking system, then add the esc keyhandler to all dialogs that have close on escape prop enabled currently, then remove the close on escape prop. What do you think?

Sure, i think that's the way to go.
I'll make the necessary changes.

ssidharth010:
@webfiltered I'll close this MR. Please continue the review on the below MR, I had some rebase issues with current branch.
https://github.com/Comfy-Org/ComfyUI_frontend/pull/4364

webfiltered:
Hey @ssidharth010, as your replacement PR is merged, I'm closing this one.  Please ping back if you want it reopened (and it's locked)!  Thanks again!

PR #4310, opened 2025-06-30
[fix] Mock release API in browser tests to prevent UI interference

christian-byrne:
Fixes intermittent test failures where release notification popups block test interactions with UI elements in the bottom-left area. The fix mocks the /releases endpoint by default while allowing tests to disable mocking when needed for release-specific functionality.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4310-fix-Mock-release-API-in-browser-tests-to-prevent-UI-interference-2226d73d365081768473e708698278cd) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/fixtures/ComfyPage.ts
+++ b/browser_tests/fixtures/ComfyPage.ts
@@ -268,8 +268,35 @@ export class ComfyPage {
     return this._history
   }
 
-  async setup({ clearStorage = true }: { clearStorage?: boolean } = {}) {
+  async setup({
+    clearStorage = true,
+    mockReleases = true
+  }: {
+    clearStorage?: boolean
+    mockReleases?: boolean
+  } = {}) {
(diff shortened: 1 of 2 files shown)


browser_tests/fixtures/ComfyPage.ts
```diff
@@ -1086,7 +1113,7 @@ export const comfyPageFixture = base.extend<{
   },
   comfyMouse: async ({ comfyPage }, use) => {
     const comfyMouse = new ComfyMouse(comfyPage)
-    use(comfyMouse)
+    await use(comfyMouse)
```
webfiltered, line 1116: Future reader note: Unrelated but required for linter pass; this should have no effect, as there is no cleanup for this fixture.

webfiltered (approved):

Issue #4142, opened 2025-06-12
[Bug] Minor: When the menu bar is displayed at the bottom, clicking on the menu causes a display issue.


comfyui-wiki:
When sizing the ComfyUI window to a very narrow width, the menu bar will collapse, and when the menu bar is at the bottom, clicking the menu button will display the menu under the menu bar as well. 

<img width="860" alt="Image" src="https://github.com/user-attachments/assets/66fb7af1-39d8-4bbf-a1d5-4493661ab84f" />

https://github.com/user-attachments/assets/d4071921-1e04-4280-ac83-7ccb0a2f0a3b

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4142-Bug-Minor-When-the-menu-bar-is-displayed-at-the-bottom-clicking-on-the-menu-causes-2106d73d365081e8b152cc33a7608801) by [Unito](https://www.unito.io)

efrancisworks:
Looking into this issue and it appears PrimeVue doesn't support this option in it's mobile view and is overriding the css to allow it. I'm exploring a custom way to get around it. I could also create a pop up menu similar to settings instead if that works?

christian-byrne:
It might be difficult to create a new popup menu that still retains the nesting behavior. Perhaps it is easier to just force the menubar to be at the top while on mobile? WDYT?

efrancisworks:
I like that idea better

christian-byrne:
Fixed by https://github.com/Comfy-Org/ComfyUI_frontend/pull/4312

PR #4312, opened 2025-07-01
Menu bar mobile behavior change

efrancisworks:
Address issue #4142, forces the menubar to be at the top while on mobile view.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4312-Menu-bar-mobile-behavior-change-2236d73d365081dc967aed01bf2ac36f) by [Unito](https://www.unito.io)


src/views/GraphView.vue
```diff
@@ -70,6 +70,8 @@ const settingStore = useSettingStore()
 const executionStore = useExecutionStore()
 const colorPaletteStore = useColorPaletteStore()
 const queueStore = useQueueStore()
+const breakpoints = useBreakpoints({ md: 961 })
```
christian-byrne, line 73: Is PrimeVue using 961 for Menubar mobile threshold?

christian-byrne (commented):
What about editing the setting as well like

```typescript
  {
    id: 'Comfy.UseNewMenu',
    category: ['Comfy', 'Menu', 'UseNewMenu'],
    defaultValue: 'Top',
    name: 'Use new menu',
    type: 'combo',
    options: ['Disabled', 'Top', 'Bottom'],
    tooltip: 'Menu bar position. On mobile devices, the menu is always shown at the top.',  // ADD THIS
    migrateDeprecatedValue: (value: string) => {
      if (value === 'Floating') {
        return 'Top'
      }
      return value
    }
  }
```

We could also use computed values for clarity:

```typescript
const showTopMenu = computed(() => isMobile.value || useNewMenu.value === 'Top')
const showBottomMenu = computed(() => !isMobile.value && useNewMenu.value === 'Bottom')
```

What do you think?

src/views/GraphView.vue
efrancisworks: Yes, this is the value which the Menubar changes.

christian-byrne (approved):
LGTM, thanks so much!

PR #4314, opened 2025-07-01
Style and type Vue WidgetToggleSwitch

benceruleanlu:
The first widget PR before the rest get updated with proper styling, functionality, and typings.

This PR updates the PrimeVue component with the Figma stylings, and adds some typings for the widget options.

As for the constructor, do we implement a feature flag or just replace it entirely?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4314-Style-and-type-Vue-WidgetToggleSwitch-2236d73d3650812a841ac5b55f08027a) by [Unito](https://www.unito.io)


benceruleanlu:
Needs component testing (?)

src/components/graph/vueWidgets/index.ts
christian-byrne, line 1: Do we need this?

src/components/graph/vueWidgets/WidgetToggleSwitch.vue
```diff
@@ -3,7 +3,27 @@
...
+            'hover:outline hover:outline-1 hover:outline-[#5B5E7D] rounded-full'
+        },
+        slider: ({ props }) => ({
+          style: {
+            backgroundColor: props.modelValue ? '#0b8ce9' : '#0e0e12'
+          }
+        }),
+        handle: ({ props }) => ({
+          style: {
+            backgroundColor: props.modelValue ? '#ffffff' : '#5b5e7d'
```
christian-byrne, line 22: Are our theme's colors different from the design?

christian-byrne (changes requested):
Can you include a screenshot or video and link to the Figma?

src/components/graph/vueWidgets/index.ts
benceruleanlu: We don't, I think it was one of the knowledge base markdown files that requested it.

christian-byrne:
> Needs component testing (?)

We can add tests later on. The implementations are still subject to lots of change.

benceruleanlu:
Figma reference: https://www.figma.com/design/CmhEJxo4oZSuYpqG1Yc39w/Project-V3?node-id=987-4107&m=dev

src/components/graph/vueWidgets/index.ts
benceruleanlu: Removed it.

src/components/graph/vueWidgets/WidgetToggleSwitch.vue
benceruleanlu: They are, unfortunately.

When the colors are reused, I think we can integrate it into our theme system.

benceruleanlu:
enabled, hovered
image

disabled, not hovered
image

in all of its glory...

the figma reference

image

src/components/graph/vueWidgets/WidgetToggleSwitch.vue
christian-byrne: Oh, are the differences very minor? I was under impression we are using the PrimeVue figma kit but not sure how it works for color themes. I think we may be able to just use the default colors and then adjust the color theme in tailwind config later if necessary.

PR #4328, opened 2025-07-02
Add unused i18n keys detection to pre-commit hook

christian-byrne:
## Summary
- Adds TypeScript script to detect unused i18n keys in staged locale files
- Integrates with pre-commit hook to block commits with unused keys
- Only checks NEW keys added in commits (ignores existing unused keys)
- Smart filtering to avoid false positives for dynamically constructed keys

## Test plan
- [x] Tested script detects unused keys and blocks commits
- [x] Tested script allows commits when no unused keys present
- [x] Verified integration with existing pre-commit workflow

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4328-Add-unused-i18n-keys-detection-to-pre-commit-hook-2246d73d3650816f8c8cd60939a690bd) by [Unito](https://www.unito.io)

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -3,3 +3,6 @@ if [[ "$OS" == "Windows_NT" ]]; then
 else
   npx lint-staged
 fi
+
+# Check for unused i18n keys in staged files
+npx tsx scripts/check-unused-i18n-keys.ts
--- a/scripts/check-unused-i18n-keys.ts
+++ b/scripts/check-unused-i18n-keys.ts
@@ -1,0 +1,183 @@
+#!/usr/bin/env tsx
+import { execSync } from 'child_process'
+import * as fs from 'fs'
+import { globSync } from 'glob'
+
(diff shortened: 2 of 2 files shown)


.husky/pre-commit
```diff
@@ -3,3 +3,6 @@ if [[ "$OS" == "Windows_NT" ]]; then
 else
   npx lint-staged
 fi
+
+# Check for unused i18n keys in staged files
+npx tsx scripts/check-unused-i18n-keys.ts
```
webfiltered, line 8: This breaks on windows; please mimic above (ridiculous) if/else workaround.

christian-byrne:
This is just a pre-commit script so let's merge this for now. If it is giving false positives or causing trouble we can disable later.

It was tested as described in the PR description.

.husky/pre-commit
christian-byrne: got it, making PR now!

Issue #4073, opened 2025-06-04
[Feature Request]: Implement versioned default settings system


christian-byrne:
**Is there an existing issue for this?**
I have searched the existing issues and checked the recent builds/commits

**What would your feature do ?**
Implement a versioned default settings system that allows changing default values for new users without affecting existing users who haven't explicitly set a preference. This feature would enable the team to improve the default user experience for new installations while respecting existing users' workflows.

**Proposed workflow**
1. Add a `defaultsByInstallVersion` field to SettingParams that tracks different defaults based on installation version:
   ```typescript
   {
     id: 'Comfy.LinkRelease.Action',
     defaultValue: LinkReleaseTriggerAction.CONTEXT_MENU,
     defaultsByInstallVersion: {
       '1.21.3': LinkReleaseTriggerAction.SEARCH_BOX,
       '1.40.3': LinkReleaseTriggerAction.FUTURE_FEATURE
     }
   }
   ```

2. Store an `installedVersion` setting when a user first runs the application. For refernce how to do this, search how `Comfy.TutorialCompleted` setting is set. You may also need to update tests to set this value, unless a test case specifically wants to test things for a new user. Again just search how it is handled for `Comfy.TutorialCompleted`

3. Installed versions can be found in systemStatsStore or as a global variable on window object. When getting a setting's default value:
   - If user has `installedVersion` >= specified version, use the versioned default
   - If user has no `installedVersion` (existing user), use the original `defaultValue`
   - This ensures backward compatibility for existing users

4. The setting store's `getDefaultValue` function would be updated to check `defaultsByInstallVersion` based on the user's installation version

**Additional information**
This approach was discussed in PR #3652 as a solution for changing the default link release action from context menu to search box for new users only. The implementation would:
- Maintain clean separation between user preferences and system defaults
- Allow progressive enhancement of defaults without disrupting existing users
- Provide a clear audit trail of when defaults changed
- Support A/B testing of defaults in the future if needed

Related discussion: https://github.com/Comfy-Org/ComfyUI_frontend/pull/3652#issuecomment-2447681991

christian-byrne:
Resolved by https://github.com/Comfy-Org/ComfyUI_frontend/pull/4337

jtydhr88:
@christian-byrne https://github.com/Comfy-Org/ComfyUI_frontend/pull/4337 is the first part of this feature, I will send another PR to finish this

jtydhr88:
[...]

PR #4337, opened 2025-07-03
add installedVersion

jtydhr88:
first part for https://github.com/Comfy-Org/ComfyUI_frontend/issues/4073, which is the foundation for next, when we should set Comfy.InstalledVersion, or in other words, how to determine whether a user is **a new user** or **an existing user**.
As research, the user settings will be stored in Comfy Core/BE folder - user/defalut/comfy.settings.json, and new instance of ComfyUI will have no settings. will return empty settings to frontend.
if this settings file is not empty, it means the user has already used ComfyUI before, and we should not set the Comfy.InstalledVersion, according to https://github.com/Comfy-Org/ComfyUI_frontend/issues/4073

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4337-add-installedVersion-2256d73d365081bba74ec3566423a2bb) by [Unito](https://www.unito.io)

### DIFF
--- a/src/services/newUserService.ts
+++ b/src/services/newUserService.ts
@@ -39,9 +39,6 @@ export const newUserService = () => {
     if (isNewUserDetermined) return
 
     isNewUserCached = checkIsNewUser(settingStore)
-
-    console.log(`New user status determined: ${isNewUserCached}`)
-
     isNewUserDetermined = true
 
     if (!isNewUserCached) {
--- a/src/services/newUserService.ts
+++ b/src/services/newUserService.ts
@@ -1,0 +1,77 @@
+import type { useSettingStore } from '@/stores/settingStore'
+
+export const newUserService = () => {
+  let isNewUserDetermined = false
+  let isNewUserCached: boolean | null = null
+  let pendingCallbacks: Array<() => Promise<void>> = []
+
+  function checkIsNewUser(
+    settingStore: ReturnType<typeof useSettingStore>
+  ): boolean {
+    const isNewUserSettings =
+      Object.keys(settingStore.settingValues).length === 0 ||
+      !settingStore.get('Comfy.TutorialCompleted')
+    const hasNoWorkflow = !localStorage.getItem('workflow')
+    const hasNoPreviousWorkflow = !localStorage.getItem(
+      'Comfy.PreviousWorkflow'
+    )
+
+    return isNewUserSettings && hasNoWorkflow && hasNoPreviousWorkflow
+  }
+
+  async function registerInitCallback(callback: () => Promise<void>) {
+    if (isNewUserDetermined) {
+      if (isNewUserCached) {
+        try {
+          await callback()
+        } catch (error) {
+          console.error('New user initialization callback failed:', error)
+        }
+      }
+    } else {
+      pendingCallbacks.push(callback)
+    }
+  }
+
+  async function initializeIfNewUser(
+    settingStore: ReturnType<typeof useSettingStore>
+  ) {
+    if (isNewUserDetermined) return
+
+    isNewUserCached = checkIsNewUser(settingStore)
+
+    console.log(`New user status determined: ${isNewUserCached}`)
+
+    isNewUserDetermined = true
+
+    if (!isNewUserCached) {
+      pendingCallbacks = []
+      return
+    }
+
+    await settingStore.set(
+      'Comfy.InstalledVersion',
+      __COMFYUI_FRONTEND_VERSION__
+    )
+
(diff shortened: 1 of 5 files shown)


christian-byrne:
Thanks for starting this implementation! I know I was the one who wrote the intial spec, but I have a suggestion to make this more robust and extensible so we can extend new user initialization in the future.

### Extract to a Service with Improved Detection

I recommend creating a dedicated service that:
1. Uses multiple indicators for more reliable new user detection
2. Allows extensions to register callbacks for new user initialization

```typescript
// src/services/newUserService.ts

let isNewUserDetermined = false
let isNewUserCached: boolean | null = null
let pendingCallbacks: Array<() => Promise<void>> = []

function checkIsNewUser(settingStore: ReturnType<typeof useSettingStore>): boolean {
  const hasNoSettings = Object.keys(settingStore.settingValues).length === 0
  const hasNoWorkflow = !localStorage.getItem('workflow')
  const hasNoPreviousWorkflow = !localStorage.getItem('Comfy.PreviousWorkflow')
  
  return hasNoSettings && hasNoWorkflow && hasNoPreviousWorkflow
}

export const newUserService = {
  async registerInitCallback(callback: () => Promise<void>) {
    if (isNewUserDetermined) {
      if (isNewUserCached) {
        try {
          await callback()
        } catch (error) {
          console.error('New user initialization callback failed:', error)
        }
      }
    } else {
      pendingCallbacks.push(callback)
    }
  },

  async initializeIfNewUser(settingStore: ReturnType<typeof useSettingStore>) {
    if (isNewUserDetermined) return

    isNewUserCached = checkIsNewUser(settingStore)
    isNewUserDetermined = true

    if (\!isNewUserCached) {
      pendingCallbacks = []
      return
    }

    await settingStore.set('Comfy.InstalledVersion', __COMFYUI_FRONTEND_VERSION__)

    // using a `while` loop here with `pop` is better though
    for (const callback of pendingCallbacks) {
      try {
        await callback()
      } catch (error) {
        console.error('New user initialization callback failed:', error)
      }
    }

    pendingCallbacks = []
  },

  isNewUser(): boolean | null {
    return isNewUserDetermined ? isNewUserCached : null
  }
}
```

Then in GraphCanvas.vue:
```typescript
import { newUserService } from '@/services/newUserService'

// After settings load
await newUserService.initializeIfNewUser(settingStore)
```

And update the setting to use `defaultValue: ''` instead of `defaultValue: null` in coreSettings.ts.

This approach:
- Uses multiple indicators (settings, workflow, previousWorkflow) for more reliable detection
- Provides a clean API for extensions to hook into new user initialization
- Handles timing issues by queuing callbacks if needed
- Has minimal memory overhead

Using an empty string as the default value maintains consistency with other string settings in the codebase and avoids potential null handling issues.

christian-byrne:
Actually, this line is not too robust:

```typescript
const hasNoSettings = Object.keys(settingStore.settingValues).length === 0
```

Because on Desktop, we pre-load some settings values before the python server starts.

We can check for the `Comfy.TutorialCompleted` for that case like


```typescript
const isNewUserSettings = Object.keys(settingStore.settingValues).length === 0 || settingsStore.get('Comfy.TutorialCompleted') !== true
```

webfiltered (commented):
For new user detection, we really should be setting a simple tri-state flag that explicitly says they're not a new user, immediately on first load.

Creating a new user (moving forward) should explicitly set this to false.

We can run the many-check workaround to cover the fact that we never set this for pre-existing users, but eventually it would be nice to remove it entirely.

src/services/newUserService.ts
```diff
@@ -0,0 +1,77 @@
+import type { useSettingStore } from '@/stores/settingStore'
+
+export const newUserService = () => {
+  let isNewUserDetermined = false
+  let isNewUserCached: boolean | null = null
+  let pendingCallbacks: Array<() => Promise<void>> = []
```
christian-byrne: I think these should be module-scoped so all instances of `newUserService` access the same variables. We can also make this a singleton but that has more memory overhead potentially.

jtydhr88: done, and also fixed and added some tests for this

christian-byrne: Wow, thank you!

christian-byrne:
Just pushed a small improvement to clarify the setting description. Changed from "Installed frontend version" to "The frontend version that was running when the user first installed ComfyUI" to better explain what this setting tracks.

christian-byrne:
I also removed the console.log, but didn't realize one of the unit tests was using it to verify something. I can fix it.

christian-byrne (approved):

PR #4341, opened 2025-07-03
[System Pop Up] Fix settings search showing hidden items and release toast logic bug

cloudmemox:
### Settings Search
- Fixed search showing hidden/deprecated settings that shouldn't be visible to users
- Now properly filters out `hidden` and `deprecated` settings during search, matching the behavior of the normal settings tree

### Release Toast Logic  
- Fixed inverted logic in `shouldShowToast` that was hiding toasts when it should show them
- The condition was checking if user had NOT skipped/seen changelog, but then returning false (hiding toast)
- Removed the negation operator so it correctly hides the toast when user has already skipped or seen the changelog

## Testing
- Search for settings - should only show user-configurable settings now
- Release toasts should now properly respect user's skip/seen status

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4341-System-Pop-Up-Fix-settings-search-showing-hidden-items-and-release-toast-logic-bug-2256d73d36508126a8d0e9e1534cfa8e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/setting/useSettingSearch.ts
+++ b/src/composables/setting/useSettingSearch.ts
@@ -53,6 +53,11 @@ export function useSettingSearch() {
     const queryLower = query.toLocaleLowerCase()
     const allSettings = Object.values(settingStore.settingsById)
     const filteredSettings = allSettings.filter((setting) => {
+      // Filter out hidden and deprecated settings, just like in normal settings tree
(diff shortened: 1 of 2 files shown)


christian-byrne:
Can we split these into separate PRs and add Playwright test for each?

PR #4135, opened 2025-06-11
Feat/version default searchbox

yiqun12:
issue #4073 
issue #3652 
### Summary

Implement versioned default settings system to allow changing default values for new users while preserving existing users' preferences.

### What this PR does
1.Introduced Comfy.InstalledVersion to record each user's initial install version.
image

2.When a user first loads the app, if no InstalledVersion exists, write the current app version to settings; otherwise default to 0.0.1(to identify existing users. or should i just empty??????) for existing users to ensure backward compatibility.

3.Added defaultsByInstallVersion field to SettingParams to define versioned default values.

4.When reading default values:
image

Case 1 : If user's InstalledVersion >= version threshold, use the corresponding versioned default. (E.g. 1.21.3)

image

Case 2 : Otherwise, fallback to the original defaultValue to preserve existing users' preferences. (E.g. 1.22.3)
image
image

3.Updated version check logic to ensure new defaults only apply to new users while protecting legacy users from config changes.
4.Added helper logic to simulate versioning scenarios for testing by modifying getCurrentVersion().

## Reproduction Logic

Start the app in multi-user mode and log in with a brand new user 

For those who does not have Comfy.InstalledVersion yet. The system will automatically send a request to store the current version as the user's initial installed version.

You can simulate different versions by temporarily modifying the return value of getCurrentVersion() in /src/utils/versioning.ts.

For example, set it to 1.20.1 and then log in as a new user. You will see that Comfy.InstalledVersion is now recorded as 1.20.1.

In this scenario, the Link Release setting will default to CONTEXT_MENU. Otherwsie if did not modify to old version, it would use the corresponding versioned default. 

After restoring getCurrentVersion() back to normal and restarting the app, the stored Comfy.InstalledVersion will remain 1.20.1 for that user — it won't be overwritten.

Even as the app version upgrades in the future, existing users will continue to use the default value CONTEXT_MENU, ensuring backward compatibility and preventing config changes for legacy users. This behavior fulfills the design goal: Store an installedVersion setting when a user first runs the application.

image
image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4135-Feat-version-default-searchbox-20f6d73d36508145a496e054173526e9) by [Unito](https://www.unito.io)

PR #4354, opened 2025-07-05
add defaultsByInstallVersion

jtydhr88:
second part of https://github.com/Comfy-Org/ComfyUI_frontend/issues/4073 to add versionedDefalutSettings. 
Main logic comes from parts of https://github.com/Comfy-Org/ComfyUI_frontend/pull/4135 but implemented more unit tests.

For each cases, please refer to unit tests

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4354-add-defaultsByInstallVersion-2276d73d365081c6bffdcc0bfa61e5f9) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/settingStore.ts
+++ b/src/stores/settingStore.ts
@@ -85,7 +85,23 @@ export const useSettingStore = defineStore('setting', () => {
   function getDefaultValue<K extends keyof Settings>(key: K): Settings[K] {
     const param = settingsById.value[key]
 
-    if (param?.defaultsByInstallVersion) {
+    const versionedDefault = getVersionedDefaultValue(key, param)
+
+    if (versionedDefault) {
+      return versionedDefault
+    }
+
+    return typeof param?.defaultValue === 'function'
+      ? param.defaultValue()
+      : param?.defaultValue
+  }
+
+  function getVersionedDefaultValue<K extends keyof Settings>(
+    key: K,
+    param: SettingParams
+  ): Settings[K] | null {
+    // get default versioned value, skipping if the key is 'Comfy.InstalledVersion' to prevent infinite loop
+    if (param?.defaultsByInstallVersion && key !== 'Comfy.InstalledVersion') {
       const installedVersion = get('Comfy.InstalledVersion')
 
       if (installedVersion) {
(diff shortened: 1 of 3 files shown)


src/stores/settingStore.ts
```diff
@@ -83,6 +84,26 @@ export const useSettingStore = defineStore('setting', () => {
...
+
+        for (const version of sortedVersions.reverse()) {
+          if (compareVersions(installedVersion, version) >= 0) {
+            const versionedDefault = param.defaultsByInstallVersion[version]
+            return typeof versionedDefault === 'function'
+              ? versionedDefault()
+              : versionedDefault
+          }
+        }
+      }
```
christian-byrne: Can we move it to its own function?

src/stores/settingStore.ts
```diff
@@ -83,6 +84,26 @@ export const useSettingStore = defineStore('setting', () => {
    */
   function getDefaultValue<K extends keyof Settings>(key: K): Settings[K] {
     const param = settingsById.value[key]
+
+    if (param?.defaultsByInstallVersion) {
```
christian-byrne: Potentially we also add condition here: `&& key !== 'Comfy.InstalledVersion'` with comment explaining it is to prevent infinite recursion.

src/types/settingTypes.ts
```diff
@@ -35,6 +35,7 @@ export interface Setting {
 export interface SettingParams extends FormItem {
   id: keyof Settings
   defaultValue: any | (() => any)
+  defaultsByInstallVersion?: Record<string, any | (() => any)>
```
christian-byrne: What about using template literal type like

```typescript
Record<`${number}.${number}.${number}`, any | (() => any)>
```

it's rare we get an opportunity to use them :smile:

christian-byrne (approved):
LGTM with minor nits.

src/stores/settingStore.ts
jtydhr88: fixed

src/stores/settingStore.ts
jtydhr88: fixed

src/types/settingTypes.ts
jtydhr88: fixed

jtydhr88: btw, if we do this, I need to add type check method in the main logic, otherwise it has type check error

christian-byrne (approved):
LGTM

PR #2921, opened 2025-03-08
Changed the MaskEditor to an Image Canvas

brucew4yn3rp:
https://github.com/Comfy-Org/ComfyUI_frontend/issues/2888

Revamped the MaskEditor UI and changed it to an Image Canvas. The context menu name has also been renamed accordingly.

The interface now supports drawing on the image itself (on the image layer) in addition to covering the area with a mask. Users can select a color or use an eyedropper tool for color matching. As indicated in the Feature Request, this is a powerful tool to guide the image generation process post-masking.

The tools on the left sidebar have been updated accordingly. The top one, is the existing mask brush, and has a matching mask logo. The second one is the image painting brush.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-2921-Changed-the-MaskEditor-to-an-Image-Canvas-1b06d73d3650816f85fbc0721e303231) by [Unito](https://www.unito.io)

PR #4361, opened 2025-07-06
[Feature] Enhanced MaskEditor to an Image Canvas

brucew4yn3rp:
@trsommer

Revamped the MaskEditor UI and changed it to an Image Canvas. The context menu name has also been renamed accordingly.

Cleaner PR of [2921](https://github.com/Comfy-Org/ComfyUI_frontend/pull/2921)

The interface now supports drawing on the image itself (on the image layer) in addition to covering the area with a mask. Users can select a color or use an eyedropper tool for color matching. This is a powerful tool to guide the image generation process post-masking (e.g., inpainting). Speaking with ComfyUI users for months, this has been highly desired addition.

The tools on the left sidebar have been updated accordingly. The top one, is the existing mask brush, and has a matching mask logo. The second one is the image painting brush.

image

image

Some relevant context from the old PR

> I need a way to only erase the drawing or the mask part. The easiest solution would be to make the layer clickable to select which layer to work on.

Layer buttons have been implemented. Also, right click will only apply to erasing whichever tool is activated. Additionally, the eraser tool will only erase on the active layer. Selecting the mask tool automatically activates the mask layer, and selecting the Paint tool automatically activates the Paint layer.

Thank you very much for your help in getting this over the line @duckcomfy. I had some issues with the prior PR so decided to close it and initiate a new, cleaner PR.


brucew4yn3rp:
@christian-byrne @webfiltered Tagging you in case you have a chance to take a look.

trsommer:
I just tried it. There seems to be a problem where the resulting layers are still split into 2 images in the load image preview. I tried it in both chrome and firefox and just painting on the canvas with the paint tool results in 2 images in the load image node
image
image

after reloading, it seems to be combined:
image

Also, if you run the workflow, the correct image seems to be handed over. The core process seems to work, but the previews in the load image node are wrong

UI fixes, not that important:
when the user selects the mask or paint tool, auto select the mask or image layer. It does not make sense for the user paint/mask without a selected layer

trsommer (changes requested):
I just tried it. There seems to be a problem where the resulting layers are still split into 2 images in the load image preview. I tried it in both chrome and firefox and just painting on the canvas with the paint tool results in 2 images in the load image node
image
image

after reloading, it seems to be combined:
image

If you run the workflow, the correct image seems to be handed over. The core process seems to work, just the previews in the load image node are wrong

UI fixes, not that important:
when the user selects the mask or paint tool, auto select the mask or image layer. It does not make sense for the user paint/mask without a selected layer

brucew4yn3rp:
Thanks @trsommer. I think with the current infrastructure across the frontend, for the user to be able to edit the painting after closing out and re-opening the interface, the node has to have the images separate. 

@duckcomfy Changing the rendering of the LoadImage node doesn't seem to work well. Let me know if you have any other ideas.

duckcomfy:
Thank you for your quick review @trsommer.
@brucew4yn3rp Please note, if it helps understanding the issue at hand, that when refreshing the page, although the correct preview is displayed, the paint layer is no longer editable.
I'm going to have a look and try to understand how the Load Image node knows what to preview, how the Mask Editor loads the different layers, and why the Load Image node contents are different when refreshing the page. If you have any insight to share on these topics please do.

brucew4yn3rp:
@duckcomfy As it stands, in app.ts, we are pushing RGBCanvas and CombinedCanvas to node.imgs. This is important because (if I remember correctly) when the MaskEditor is initialized, it accesses its images and adds the relevant painted pixels to the paint layer. That functionality is, as far as I can see with the current infrastructure, the only way to allow for continued editing. We cannot leverage the mask’s logic because it would not work across all RGB colors. 

It is interesting that upon refresh of the page, the node defaults to the combined, flattened image. My guess is that it is displaying the last saved image in the input/clipspace folder. 

I’m happy to discuss on Discord for easier communication if you’d like (@batman.76)

asagi4:
I gave this a quick try as well. Seems to work well enough and definitely solves a pain point with ComfyUI.

I did find a bug: the brush shape seems to be stuck to circular. If I select the rectangular brush, it'll show a rectangular outline but still make circular marks.

brucew4yn3rp:
Huge thanks to @duckcomfy for taking the charge with smoothing the rough edges that were remaining with this PR. Below I've added the summary and explanation of changes made from their child PR for convenience:

## Summary of Changes

- Ensures the node preview of a loaded image is that of the combined canvas (base image + mask + paint), with no other previewable image. [3e259c9](https://github.com/brucew4yn3rp/ComfyUI_frontend_ImageEditor/pull/2/commits/3e259c9ea60014bf5b3113063f63100674340857)
- Improves layer-switching UX (details below) [fbff95b](https://github.com/brucew4yn3rp/ComfyUI_frontend_ImageEditor/pull/2/commits/fbff95b8eac8d9cbeff8da5be6e570719dfbded4)
- Fixes the regression where square brush shapes became always round (NOTE: when hardness<1, brush shape becomes round, but this is true in main branch as well) [118522f](https://github.com/brucew4yn3rp/ComfyUI_frontend_ImageEditor/pull/2/commits/118522f1c9a2056ed3d758296b37dcb71d469286)
- Adds [ and ] as keyboard shortcuts to decrease/increase brush size [c744eeb](https://github.com/brucew4yn3rp/ComfyUI_frontend_ImageEditor/pull/2/commits/c744eeb774083bedbf69778fd6ac60350521376d)

## Preview image bug and fix explanation

The previewed images are pulled from the `LGraphNode` object's `imgs` (`HTMLImageElement[]`) property, and the way the code works, MaskEditor is initialized expecting the different layers to also be in that `imgs` array, and refactoring that would be difficult and high-risk. As a simple fix, I added a `data-nopreview` data attribute to the HTMLImageElement objects, to be able to supply MaskEditor with the discrete layers while only previewing the combined canvas.

Note: after refreshing the page, the paint layer is no longer editable. I'd suggest not considering this bug as a blocker for the main repo PR, as it could be a deep rabbit hole for a very minor bug.

## Layer UX improvements

This PR improves the layer UX in [fbff95b](https://github.com/brucew4yn3rp/ComfyUI_frontend_ImageEditor/pull/2/commits/fbff95b8eac8d9cbeff8da5be6e570719dfbded4). Here is an explanation copied from the commit of the UX improvements. We should copy this explanation to the parent PR after this child PR, for the benefit of reviewers.

In order to explain the new UX let me first explain what I'll call the different tools, as the names we previously used are now ambiguous.

- "Mask Pen": Previously named "Pen" internally, this is an existing tool lets you paint a mask on the image.
- "Paint Pen": New tool which lets you paint colored pixels on the image.
- "Eraser": Exising tool which lets you erase parts of your mask. As of @brucew4yn3rp's PR, you can also erase parts of your painted pixels, depending on which layer is active.
- "Mask Bucket": Previously named "Paint Bucket" internally, this is an existing tool which lets you take a mask that's just an outline, and fill in the inside of your mask outline with a full masked area.
- "Mask Color Fill": Previously named "Color Select" internally, this is an existing tool which can turn pixels into a masked area, based on color similarity in contiguous pixels on the base image. 

New UX:

- The "Mask" layer is now selected when you open the Mask Editor, instead of no layer selected.
- Selecting the "Mask Pen", "Mask Bucket", or "Mask Color Fill" tool automatically selects the Mask layer.
- Selecting the "Paint Pen" tool automatically selects the Paint layer.
- Selecting the "Mask" layer while the "Paint Pen" tool is selected automatically selects the "Mask Pen" tool.
- Selecting the "Paint" layer while the "Mask Pen", "Mask Bucket", or "Mask Color Fill" tool is selected automatically selects the "Paint Pen" tool.
- Changing the selected layer while the "Eraser" tool is selected has no additional effect, because Eraser supports both layers.
- Switching to the "Eraser" tool has no additional effect, because Eraser supports both layers.

christian-byrne:
I can review if needed but will defer to @trsommer!

trsommer:
Thank you for addressing the previous issues!  I’ve tested the updated version, and it’s very close to perfect - great work! I did notice one small issue: When a mask overlaps a painted area and the user reopens the mask editor, the mask is subtracted from the painted area.

I read your comment about the editing after reloading merging the paint layer. I think this issue is a separate problem, but let me know if not. I do think this issue is more critical as the other one, but if it is a deep rabbit hole as well, let me know

image

reopen the mask editor and save

image

brucew4yn3rp:
Thanks again for the quick review @trsommer. 

@duckcomfy fixed the mask intersection issue in the latest PR I’ve just merged in by correcting the ordering in which each layer was being pushed to the imgs array. Also ran prettier to align with formatting standard.

trsommer (dismissed):
thank you for fixing the issues I mentioned

brucew4yn3rp:
Seems like we are in the final stage of introducing a more comprehensive image editing canvas to ComfyUI! Especially with this being an edge that some other platforms have over ComfyUI, I think this solves a pain point and will be a huge value-add / attraction point for users. 

@christian-byrne As @trsommer has completed his review, would you mind taking a look as well, whenever you have a chance? I think we need a code owner review from comfy_frontend_devs before it can be merged in.

package.json
```diff
@@ -69,7 +69,7 @@
     "vite-plugin-html": "^3.2.2",
     "vite-plugin-vue-devtools": "^7.7.6",
     "vitest": "^2.0.0",
-    "vue-tsc": "^2.1.10",
+    "vue-tsc": "^2.2.12",
```
webfiltered: Can you quickly note why this is in this PR?  Is there a feature added in `vue-tsc` 2.2 that was required?

src/scripts/app.ts
```diff
@@ -349,13 +351,25 @@ export class ComfyApp {
...
 
+    var paintedIndex = 1
+    if (selectedIndex != 0) {
+      paintedIndex = selectedIndex + 1
+    }
+
+    var combinedIndex = 2
+    if (selectedIndex != 0) {
+      combinedIndex = selectedIndex + 2
+    }
```
webfiltered:
```suggestion
    const paintedIndex = selectedIndex + 1
    const combinedIndex = selectedIndex + 2
```

nits:
- Prefer `const` or `let`
- Simplifies code

src/scripts/app.ts
```diff
@@ -378,8 +394,9 @@ export class ComfyApp {
             node.images = ComfyApp.clipspace.images
           }
 
-          if (app.nodeOutputs[node.id + ''])
+          if (app.nodeOutputs[node.id + '']) {
             app.nodeOutputs[node.id + ''].images = node.images
+          }
```
webfiltered: nit: I get it, however this already a large PR.  Pure style changes being included make more difficult to review / merge / work around.  We can 100% address this in a follow-up style PR or as part of a smaller change.

tsconfig.json
```diff
@@ -15,8 +15,8 @@
 
     /* Linting */
     "strict": true,
-    "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedLocals": false,
+    "noUnusedParameters": false,
```
webfiltered: Please revert this and fix any resultant errors.

webfiltered (changes requested):
Keen to take a look at this properly!  Looks like a great idea.

Have paused my review partway through though - found the `tsconfig` changes.  Once the below have been addressed, please ping~

PabloWiedemann:
Hey @brucew4yn3rp , this is great! I only have one feedback from a UX perspective: 
Assuming the "Activate Layer" buttons do what I think they do,  namely activate the respective layer, I'd propose to remove the buttons and make it so that the layers can be activated simply by clicking on the component itself. Also, if not already implemented, we'd need an active state of the component to indicate which one is currently active (see below)
Screenshot 2025-07-11 at 7 05 03 PM

brucew4yn3rp:
Thanks, @webfiltered and @PabloWiedemann. The vue-tsc version, reversions requested, and nits have been fixed. 

In terms of UX, please see a more comprehensive description of changes below. To preface, it is important to have the Activate Layer buttons specifically for the eraser tool. As you mentioned, it automatically updates the layer based on the selected tool, but when using the eraser tool, the user must be able to discern which layer they are erasing. For convenience, there is the added functionality of right clicking while using either paint or mask to erase on that layer.   

For better coherence of use, the Mask Blending options have been moved to their own container, similar to the Mask Opacity slider. 

**Layer Interface Reorganization:**

The structure of the different layer containers is now standardized. There is a checkbox to hide/show the associated pixels, and an activation button for each of the mask and paint layers.

- Moved "Activate Mask Layer" button into the Mask Layer container alongside the mask icon
- Created separate "Mask Blending Options" container with centered dropdown (black, white, negative options)
- Reorganized layer order: Mask Layer → Mask Blending Options → Mask Opacity → Paint Layer → Base Image Layer
- Renamed "Image Layer" to "Base Image Layer" for clarity

**Activate Button Behavior:**

- Changed activate buttons from hiding/showing to greying out (50% opacity) and disabling when that layer is active
- Buttons now are only visible when Eraser tool is selected and remain visible but disabled for the currently active layer

**Active Layer Visual Feedback:**

- Added blue border (2px solid #007acc) around the currently active layer container
- Implemented updateActiveLayerHighlight() method to manage border highlighting

**Scrollbar**

- Given the enhancements and additions made to the side panel, we've implemented a scrollbar for resolutions that aren't tall enough for the entire menu


https://github.com/user-attachments/assets/bb4e2f1a-9acb-4f6a-a0e7-46c1dd957026

src/extensions/core/load3d/ControlsManager.ts
```diff
@@ -9,6 +9,7 @@ import {
 
 export class ControlsManager implements ControlsManagerInterface {
   controls: OrbitControls
+
```
webfiltered:
```suggestion
```

Only change to this file is now a new line.

src/extensions/core/load3d/SceneManager.ts
```diff
@@ -21,6 +21,7 @@ export class SceneManager implements SceneManagerInterface {
   private renderer: THREE.WebGLRenderer
 
   private getActiveCamera: () => THREE.Camera
+
```
webfiltered:
```suggestion
```

Only change to this file is now a new line.

package-lock.json
```diff
@@ -4919,30 +4919,30 @@
       }
     },
     "node_modules/@volar/language-core": {
-      "version": "2.4.10",
-      "resolved": "https://registry.npmjs.org/@volar/language-core/-/language-core-2.4.10.tgz",
-      "integrity": "sha512-hG3Z13+nJmGaT+fnQzAkS0hjJRa2FCeqZt6Bd+oGNhUkQ+mTFsDETg5rqUTxyzIh5pSOGY7FHCWUS8G82AzLCA==",
+      "version": "2.4.15",
+      "resolved": "https://registry.npmjs.org/@volar/language-core/-/language-core-2.4.15.tgz",
+      "integrity": "sha512-3VHw+QZU0ZG9IuQmzT68IyN4hZNd9GchGPhbD9+pa8CVv7rnoOZwo7T8weIbrRmihqy3ATpdfXFnqRrfPVK6CA==",
```
webfiltered: Please revert all changes to `package-lock.json`.

src/extensions/core/load3d/ViewHelperManager.ts
```diff
@@ -11,6 +11,7 @@ export class ViewHelperManager implements ViewHelperManagerInterface {
   private getActiveCamera: () => THREE.Camera
   private getControls: () => OrbitControls
   private nodeStorage: NodeStorageInterface
+
```
webfiltered:
```suggestion
```

Only change to this file is now a new line.

src/scripts/app.ts
```diff
@@ -448,6 +479,10 @@ export class ComfyApp {
         }
       }
 
+      node.imgs?.forEach((imgEl) => {
+        if (imgEl.src === combinedImgSrc) return
+        imgEl.setAttribute('data-nopreview', 'true')
+      })
```
webfiltered: This feels like it might be an unnecessary workaround - is there a reason we need to use node.imgs as storage?   Dual-purposing things in this way can often lead to maintainability issues.

src/extensions/core/maskeditor.ts
```diff
@@ -956,181 +982,149 @@ class MaskEditorDialog extends ComfyDialog {
...
+    refinedMaskCanvasCtx.globalCompositeOperation =
+      CompositionOperation.SourceOver
+    refinedMaskCanvasCtx.putImageData(refinedMaskOnlyData, 0, 0)
+
+    const filenames = {
+      maskedImage: 'clipspace-mask-' + performance.now() + '.png',
+      paint: 'clipspace-paint-' + performance.now() + '.png',
+      paintedImage: 'clipspace-painted-' + performance.now() + '.png',
+      paintedMaskedImage:
+        'clipspace-painted-masked-' + performance.now() + '.png'
```
webfiltered:
```suggestion
    const timestamp = performance.now()

    const filenames = {
      maskedImage: `clipspace-mask-${timestamp}.png`,
      paint: `clipspace-paint-${timestamp}.png`,
      paintedImage: `clipspace-painted-${timestamp}.png`,
      paintedMaskedImage: `clipspace-painted-masked-${timestamp}.png`
```

Fix all files have different timestamps.

src/extensions/core/maskeditor.ts
```diff
@@ -2450,15 +2551,35 @@ class BrushTool {
...
+    // Remove # if present
+    hex = hex.replace(/^#/, '')
+
+    // Parse r, g, b values
+    const r = parseInt(hex.substring(0, 2), 16)
+    const g = parseInt(hex.substring(2, 4), 16)
+    const b = parseInt(hex.substring(4, 6), 16)
+
+    return `rgba(${r}, ${g}, ${b}, ${alpha})`
+  }
```
webfiltered: nit: duplicate code - `hexToRgb` from `colourUtil.ts`

src/constants/coreKeybindings.ts
```diff
@@ -181,5 +181,7 @@ export const CORE_KEYBINDINGS: Keybinding[] = [
       shift: true
     },
     commandId: 'Comfy.Graph.ConvertToSubgraph'
-  }
+  },
+  { combo: { key: ']' }, commandId: 'Comfy.MaskEditor.BrushSize.Increase' },
+  { combo: { key: '[' }, commandId: 'Comfy.MaskEditor.BrushSize.Decrease' }
```
webfiltered: Can we remove the default key bindings for the initial PR?  Can add them back later.  We're working on improving keyboard UX atm so adding new _default_ bindings requires checking with a few people, right now.

src/composables/widgets/useImagePreviewWidget.ts
```diff
@@ -35,7 +35,9 @@ const renderPreview = (
     node.pointerDown = null
   }
 
-  const imgs = node.imgs ?? []
+  const imgs = (node.imgs ?? []).filter(
+    (img) => img.getAttribute('data-nopreview') !== 'true'
+  )
```
webfiltered: Can probably just revert this if we don't use `node.imgs`, right?

webfiltered (changes requested):
Looks pretty good!  Have some follow-ups but nothing major.  The biggest is a comment on `app.ts` about `node.imgs`.

Have also added suggestions which you can just click commit on if that works for you.  If any tests fail because of them, feel free to ping and I can resolve.

src/scripts/app.ts
duckcomfy: At first, I thought the `node` object was the only sensible place to get layer data from, and then the `node.imgs` the only semi-sensible existing property to store that in. After your comment, I first tried an approach that extends LGraphNode (after I noticed the frontend codebase was already doing that) to have a new `maskEditorLayers` prop, but then I realised that the only approach that'll work completely is to derive the layer URLs from the main image URL. The reason why is that this approach survives page refreshes, unlike storing layer data in the node object.

You can see how I implemented this new approach in: https://github.com/brucew4yn3rp/ComfyUI_frontend_ImageEditor/pull/6/commits/ab3ebe384c3c79247edddaa79273fa5814cd026e

I addressed all the other review items in the [stacked PR](https://github.com/brucew4yn3rp/ComfyUI_frontend_ImageEditor/pull/6) which @brucew4yn3rp told me on Discord he will merge when he next is available.

src/constants/coreKeybindings.ts
asagi4: Those hotkeys are also not the best for people with non-US keyboard layouts. The brush sizes could use a number input (precise control of size is almost impossible currently) but that's not for this PR

brucew4yn3rp:
Hi @webfiltered, thanks for the attentiveness on this - I know there's lots going on. 

All of your comments have been addressed in the latest commits. We discussed offline a way to circumvent the need to store images within the node. Therefore, we no longer need the `data-nopreview` workaround either. 

Originally, I had implemented the storing of images in `node.imgs` as it was the only way I'd found to be able to persist which correct files to reference for each layer upon closing and re-opening of the canvas. 

Now, the timestamp has been unified across all the related files, allowing us to derive the names of each layer's saved file from the final filename that gets saved to the node/workflow. As a byproduct, this allows editability to persist through refreshes because deriving the filenames is static (as opposed to `node.imgs` which did not persist through app refresh),

This seems to work well, and without the additional clog in retaining images within the node, which you mentioned may be a strain for future maintainability. To that end, @duckcomfy has also added a `maskeditor.test.ts` file that will assist in ensuring the new functionality does not break in the future.

PabloWiedemann (approved):
LGTM

PabloWiedemann (approved):

kevinpaik1:
@brucew4yn3rp once your PR gets merged, would you please write into me at bounty@comfy.org? i'd love to compensate you as a thank you for your significant effort here!

webfiltered (approved):

kevinpaik1:
Hi @trsommer — we’d like to propose adding @brucew4yn3rp as a CODEOWNER for the MaskEditor component, based on:

1. The recent refactoring work in this PR
2. He’s expressed interest in helping review future contributions in this area

@christian-byrne mentioned you're currently listed as the CODEOWNER here, so I wanted to check with you. Would you be open to adding @brucew4yn3rp alongside you? Thanks in advance for considering!

avc:
I found a bug I think may be related to this implementation https://github.com/Comfy-Org/ComfyUI_frontend/issues/5251. The mask editor is now anti-aliasing even at max hardness, making fills not fill completely and cleanly.

silveroxides:
@brucew4yn3rp You did not notice how this completely broke the right side tool configuration menu for now 2 months? It is supposed to change when selecting the color select tool with the dropper icon to the left. Code is still there, just that this was so sloppily implemented and completely lacked quality control and even the mere basic manual testing.

trsommer:
> @brucew4yn3rp You did not notice how this completely broke the right side tool configuration menu for now 2 months? It is supposed to change when selecting the color select tool with the dropper icon to the left. Code is still there, just that this was so sloppily implemented and completely lacked quality control and even the mere basic manual testing.

I have fixed paint bucket and color select tool settings not showing up in #5733 , I think that was the issue you were talking about. I have no idea why @brucew4yn3rp removed this. I did not notice because I only tested the new features implemented

silveroxides:
@trsommer
Thank you. This was most likely an oversight and the lack of manual testing led to it not being detected.
I don't know how much @brucew4yn3rp actually uses the mask editor himself when using ComfyUI but I would expect that if someone is set as codeowner for a feature, then they ought to be dogfooding said feature.

brucew4yn3rp:
Thank you @trsommer for the quick fix. That was indeed an oversight during the UI redesign. The PR went through multiple iterations covering both new functionalities and the UI (including the sidebar), so given the scope it’s unfortunate but not surprising that this slipped through during the rebuild. Again, given the scope and difficulty of implementing the new functions, I was also admittedly focused on making sure the new features worked well and missed the regression which probably happened early on.

>just that this was so sloppily implemented and completely lacked quality control and even the mere basic manual testing.

@silveroxides, instead of assuming “sloppy implementation” or lack of testing, it would be more constructive to check the PR history, which clearly shows extensive review and collaboration over several months. In a refactor of this size, oversights are inevitable, which is why collaboration exists. If you run into issues in the future, the most helpful approach is to either open an issue or submit a PR. And if you remain convinced this reflects “sloppy” work, you’re always free to continue using the old mask editor.

Issue #4356, opened 2025-07-05
[Feature Request]: Display warning when frontend version mismatches backend requirements


christian-byrne:
## Is there an existing issue for this?
- [x] I have searched the existing issues and checked the recent builds/commits

## What would your feature do?

Add a warning notification system that alerts users when there's a version mismatch between the ComfyUI frontend and the backend's required frontend version. This will help prevent compatibility issues and improve user experience by proactively notifying users when they need to update their frontend.

Currently:
- The frontend version is available via `app.config.app_version` (from `__COMFYUI_FRONTEND_VERSION__`)
- The backend ComfyUI version is available through `systemStatsStore.comfyui_version`
- There's no mechanism to check if these versions are compatible

## Proposed workflow

1. Backend would need to expose a new field in the system stats API (e.g., `required_frontend_version` or `frontend_version_range`)
2. Frontend would check this value against its own version on app initialization or when systemStatsStore is refreshed
3. If versions don't match:
   - Display a warning message using the existing Message/Toast system
   - Show current frontend version vs. required version
   - Provide action buttons to update or dismiss

Example warning scenarios:
- "Frontend version 1.24.0 is outdated. Backend requires version 1.25.0 or higher."
- "Frontend version 1.26.0 may not be compatible with backend version 1.24.0."

### Implementation Options:

1. **Persistent Warning Banner** (recommended):
   - Use PrimeVue's Message component like existing `FirstTimeUIMessage` or `MissingModelsWarning`
   - Display at the top of the UI when version mismatch is detected
   - Include "Update Now" and "Dismiss" actions

2. **Toast Notification**:
   - Use the existing toast system for a less intrusive notification
   - Could show on app startup or when backend connection is established

3. **Modal Dialog**:
   - For critical incompatibilities that might break functionality
   - Similar to `LoadWorkflowWarning` implementation

### Technical Details:

The frontend already has all the necessary infrastructure:
- Version info available in `app.config.app_version`
- System stats store for backend communication
- Multiple warning/notification UI patterns to choose from
- Existing examples of version-aware features (e.g., migration toasts)

This feature would improve user experience by preventing issues caused by version mismatches and guide users to keep their frontend updated with the backend requirements.

shivansh-gupta4:
Can I get this issue assigned?

christian-byrne:
Would you like to do this as part of the bounty program? If so, can you email bounty@comfy.org?

[...]

PR #4363, opened 2025-07-06
Feature Implemented: Warning displayed when frontend version mismatches

shivansh-gupta4:
# Issue: [Feature Request] Display warning when frontend version mismatches backend requirements #4356

- Created a store for automatic version compatibility check.
- The dismissal state is persisted using hidden settings
- A warning message component at the top of UI to show version mismatch warning with dismiss and update actions
- Backend /system_stats API exposes new field required_frontend_version to compute version mismatches
- Tests file for testing all the cases for the version compatibility.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4363-Feature-Implemented-Warning-displayed-when-frontend-version-mismatches-2286d73d36508118916dc52f6c282ddc) by [Unito](https://www.unito.io)


SrijitaBanerjee2002:
Hi Shivansh!
I was actually reviewing your code just now and have a suggestion for you. I saw your implementation for the version compatibility in your commit, however, there is already a good backend implementation of the same in the ComfyUI workspace. 

Path is: _ComfyUI/app/frontend_management.py_

I would highly recommend checking the function _check_frontend_version()_ and try to incorporate that into your implementation, for better code reusability. 

Hope that helps!

Best,
Srijita

christian-byrne:
This looks like a very high quality PR. I can review fully this week.

Should we make a corresponding PR to the backend to change this code so that it returns `required_frontend_version` https://github.com/comfyanonymous/ComfyUI/blob/1205afc708d963d160f38c1d6613a384ddf6c564/server.py#L526-L548. We could use the module mentioned by SrijitaBanerjee2002.

shivansh-gupta4:
Thank you for the detailed review and thoughtful feedback — I really appreciate it!

I'm glad to hear the implementation aligns with the expectations and integrates well with the existing architecture. I’ve also taken note of Srijita’s suggestion regarding reusing the check_frontend_version() logic from the backend. If the team decides to incorporate that, I’d be happy to refactor accordingly.

Please let me know if there are any adjustments or changes you'd like me to make — I’ll be ready to update the PR as needed.

Thanks again for the opportunity!

src/components/dialog/content/VersionMismatchWarning.vue
```diff
@@ -0,0 +1,80 @@
...
+
+const handleUpdate = () => {
+  // Open ComfyUI documentation or update instructions
+  window.open('https://docs.comfy.org/get_started/introduction', '_blank')
+}
+</script>
+
+<style scoped>
+/* Custom styles if needed */
+</style>
```
christian-byrne: Could we remove that?

src/components/dialog/content/VersionMismatchWarning.vue
```diff
@@ -0,0 +1,80 @@
...
+
+const versionStore = useVersionCompatibilityStore()
+
+const handleDismiss = () => {
+  void versionStore.dismissWarning()
+}
+
+const handleUpdate = () => {
+  // Open ComfyUI documentation or update instructions
+  window.open('https://docs.comfy.org/get_started/introduction', '_blank')
```
christian-byrne: Can we link to https://docs.comfy.org/installation/update_comfyui#missing-or-outdated-frontend%2C-workflow-templates%2C-node-after-updates? That page indicates how to solve the problem.

src/stores/versionCompatibilityStore.ts
```diff
@@ -0,0 +1,125 @@
...
+    const hasVersionMismatch = computed(() => {
+      return isFrontendOutdated.value || isFrontendNewer.value
+    })
+
+    const shouldShowWarning = computed(() => {
+      if (!hasVersionMismatch.value || isDismissed.value) {
+        return false
+      }
+
+      const currentVersionKey = `${frontendVersion.value}-${backendVersion.value}-${requiredFrontendVersion.value}`
```
christian-byrne: What do you think about using a computed for the `currentVersionKey` in the store? Maybe we don't want that memory overhead though? In that case should we use a helper function to generate the key?

src/stores/versionCompatibilityStore.ts
```diff
@@ -0,0 +1,125 @@
...
+
+    const isFrontendOutdated = computed(() => {
+      if (!frontendVersion.value || !requiredFrontendVersion.value) {
+        return false
+      }
+      return (
+        compareVersions(requiredFrontendVersion.value, frontendVersion.value) >
+        0
+      )
+    })
```
christian-byrne: Should we also test if the version is a valid semver? We might be able to avoid showing this dialog in the case that there's some error there.

christian-byrne (commented):
Looking good!

Should we edit this code to include `required_frontend_version` https://github.com/comfyanonymous/ComfyUI/blob/1205afc708d963d160f38c1d6613a384ddf6c564/server.py#L526-L548? I can also review that PR once you make it!

src/components/dialog/content/VersionMismatchWarning.vue
shivansh-gupta4: Sure, will make the changes to redirect to the relevant URL.

src/components/dialog/content/VersionMismatchWarning.vue
shivansh-gupta4: Sure, will remove this in the upcoming commit with the suggested changes.

src/stores/versionCompatibilityStore.ts
shivansh-gupta4: Thank you for pointing this out. I just checked the codebase and found that in usePackUpdateStatus.ts we are checking the semver validation before moving to the compareVersions function. We can similarly integrate the logic here and before we move to compareVersions function in both the conditions isFrontendNewer and isFrontendOutdated we perform the semver validation for both the version values to be compared. Please let me know if this approach aligns with your expectations.

src/stores/versionCompatibilityStore.ts
shivansh-gupta4: Thank you so much for the recommendation. We are currently calculating this key three times in the same file which can be optimised. We have two approaches, one, we can use a computed for the currentVersionKey and use the cached value each time, secondly to avoid memory overhead we can simply use a function to get the currentVersionKey at all the places. After thorough research on both the approaches, I came to a conclusion that using computed will have a small memory overhead but increased performance since it won't calculate the value each time and use the cached value. Also, looking at our application UI, which have high resolution animations, i would suggest using computed to maintain better performance since at a scale of greater than 1000 refreshes of the UI at a particular moment, the helper function though will need minute milliseconds to calculate value each time but that is also higher when compared to using computed at a bigger scale where performance is preferred. Please let me know if I can go forward with this approach or am I missing some important comparing factor.

shivansh-gupta4:
> Looking good!
> 
> Should we edit this code to include `required_frontend_version` https://github.com/comfyanonymous/ComfyUI/blob/1205afc708d963d160f38c1d6613a384ddf6c564/server.py#L526-L548? I can also review that PR once you make it!

@christian-byrne I have raised the PR https://github.com/comfyanonymous/ComfyUI/pull/8875 to include the required_frontend_version parameter. Also, I will be waiting for a heads-up on the approach shared for the issues in the previously raised PR to shape it perfectly.

shivansh-gupta4:
@christian-byrne I’ve committed the suggested improvements.

Additionally, I’ve used a `computed` property for `currentVersionKey` to prioritize performance. Though it have a small memory overhead but have better performance compared to recalculating via a helper function. Please let me know if this approach is correct or if I might be overlooking any important considerations.

Let me know if the current implementation aligns with your expectations — I’d be happy to make further adjustments if needed.

src/stores/versionCompatibilityStore.ts
christian-byrne: That makes sense, thanks for investigating both options thoroughly.

src/stores/versionCompatibilityStore.ts
christian-byrne: Yep, that looks great, thanks!

shivansh-gupta4:
@christian-byrne Can we run the workflow for this PR as well, so that if anything is breaking, I'll fix it.

shivansh-gupta4:
@christian-byrne Thank you for running the workflows. Can you please guide me on how to solve the chromium errors. Will fix all the errors and commit the changes ASAP.

christian-byrne:
It's possible that the chromium tests will fail if the backend PR is not merged yet and the frontend relies on it.

For ESLint and Prettier, please refer to:

- General dev setup for ComfyUI_frontend: https://github.com/Comfy-Org/ComfyUI_frontend?tab=readme-ov-file#development, make sure to run `npm run prepare` to enable pre-commit
- The dev scripts we have in package.json: https://github.com/Comfy-Org/ComfyUI_frontend/blob/ed1d944e0e4f851cfd1d90c8b2c68bda0fd9cdbb/package.json#L10-L29
  - In particular, `format`, `test:unit`, `test:component`, `lint:fix`, `build`
- Our CLAUDE.md if you use LLMs for developemnt, especially lines like these https://github.com/Comfy-Org/ComfyUI_frontend/blob/ed1d944e0e4f851cfd1d90c8b2c68bda0fd9cdbb/CLAUDE.md?plain=1#L3

shivansh-gupta4:
@christian-byrne I have corrected the workflow errors encountered. Also, I am attaching the image for the warning banner to show how it will actually look, please ignore the backend version upgradation to 1.27.4, that was staged to display the update frontend button as well. I have tested it locally and the dismissal state is properly persisting across tab and window refreshes along with the proper fetching of particular version parameters from backend. I have made a commit https://github.com/Comfy-Org/ComfyUI_frontend/pull/4363/commits/561bb008507b727aa94710a736b4029895cb5fd1 which contains all the final corrections incorporating all the actions stated.

image
image

shivansh-gupta4:
> @christian-byrne I have corrected the workflow errors encountered. Also, I am attaching the image for the warning banner to show how it will actually look, please ignore the backend version upgradation to 1.27.4, that was staged to display the update frontend button as well. I have tested it locally and the dismissal state is properly persisting across tab and window refreshes along with the proper fetching of particular version parameters from backend. I have made a commit [561bb00](https://github.com/Comfy-Org/ComfyUI_frontend/commit/561bb008507b727aa94710a736b4029895cb5fd1) which contains all the final corrections incorporating all the actions stated.
> 
> image image

@christian-byrne Awaiting for your feedback. Please let me know if this aligns with your expectations. Also, it's been some time that I have been working on the same issue and I badly want to take new issues to contribute as well. So, can we have a final testing and close both the PRs. Thank you.

shivansh-gupta4:
@christian-byrne I understand you might have been busy, but please squeeze some time to review this. I have shared more details about the solution implemented. Awaiting for your feedback.

kevinpaik1:
Thank you for your contribution @shivansh-gupta4 !

shivansh-gupta4:
@christian-byrne Wanted to know what are the problems that are causing the CI playwright tests to fail. Is there anything more that needs to be fixed in this PR? Would love to help and make further changes if needed. Thank you.

christian-byrne:
Nope, this is all good, just going to switch to using the toast API and add a test!

christian-byrne (approved):
LGTM!!

PR #4364, opened 2025-07-06
Fix: Escape closes Settings dialog if login dialog open

ssidharth010:
Fix for: https://github.com/Comfy-Org/ComfyUI_frontend/issues/3996

Problem:

The ESC button closes the dialog opened below rather than top.

Root Cause:

Since there is not close button configured on the login modal, but on the settings modal it have close button configuration.
Prime vue dialog takes only the one first one hence closing that dialog only.
https://github.com/primefaces/primevue/issues/5138

Solution:

Based on discussion in
https://github.com/Comfy-Org/ComfyUI_frontend/pull/4299, I will track the active dialog and based on the **closable** property of the top most dialog, closeOnEsc will also be updated.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4364-WIP-Fix-Escape-closes-Settings-dialog-if-login-dialog-open-2286d73d365081ffbe84cf883ea4eceb) by [Unito](https://www.unito.io)

### DIFF
--- a/src/services/dialogService.ts
+++ b/src/services/dialogService.ts
@@ -272,7 +272,7 @@ export const useDialogService = () => {
           onSuccess: () => resolve(true)
         },
         dialogComponentProps: {
-          closable: false,
+          closable: true,
           onClose: () => resolve(false)
         }
       })
--- a/src/stores/dialogStore.ts
+++ b/src/stores/dialogStore.ts
@@ -26,6 +26,8 @@ interface CustomDialogComponentProps {
   modal?: boolean
   position?: DialogPosition
   pt?: DialogPassThroughOptions
+  closeOnEscape?: boolean
+  dismissableMask?: boolean
 }
 
 type DialogComponentProps = InstanceType<typeof GlobalDialog>['$props'] &
@@ -130,7 +132,7 @@ export const useDialogStore = defineStore('dialog', () => {
       dialogStack.value.shift()
     }
 
-    const dialog: DialogInstance = {
+    const dialog = {
       key: options.key,
       visible: true,
       title: options.title,
--- a/src/stores/dialogStore.ts
+++ b/src/stores/dialogStore.ts
@@ -61,6 +61,11 @@ export interface ShowDialogOptions {
 
 export const useDialogStore = defineStore('dialog', () => {
   const dialogStack = ref<DialogInstance[]>([])
+
+  /**
+   * The key of the currently active (top-most) dialog.
+   * Only the active dialog can be closed with the ESC key.
+   */
   const activeKey = ref<string | null>(null)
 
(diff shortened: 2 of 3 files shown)


src/stores/dialogStore.ts
```diff
@@ -156,10 +171,29 @@ export const useDialogStore = defineStore('dialog', () => {
...
   }
 
+  /**
+   * Ensures only the top-most dialog in the stack can be closed with the Escape key.
+   * This is necessary because PrimeVue Dialogs do not handle `closeOnEscape` prop
+   * correctly when multiple dialogs are open.
+   */
+  function updateCloseOnEscapeStates() {
+    const topDialog = dialogStack.value.find((d) => d.key === activeKey.value)
+    const topClosable = topDialog?.dialogComponentProps.closable
```
christian-byrne, line 189: Should it be checking for `closable` or `closeOnEscape` here?

src/services/dialogService.ts
```diff
@@ -272,7 +272,7 @@ export const useDialogService = () => {
           onSuccess: () => resolve(true)
         },
         dialogComponentProps: {
-          closable: false,
+          closable: true,
```
christian-byrne, line 275: Do we need to change `closable` or `closeOnEscape`?

christian-byrne (commented):
This looks great. I feel like this PrimeVue name `closable` is causing me confusion

Selection_1675

It seems like `closable` is more just about whether to show the x in the top right rather than whether or not the dialog is physically closable whatsoever. Is that accurate? Forgive me if I misunderstood.

ssidharth010:
@christian-byrne 
The naming of the closable prop in PrimeVue can be a bit misleading. To truly prevent the dialog from being dismissed, all three props need to be set to false: closable, closeOnEscape, and dismissableMask.

Since closeOnEscape isn’t explicitly set anywhere in the app, I’m using the existing closable flag passed to the DialogStore as the single source of truth. This flag controls both closable and closeOnEscape on the PrimeVue <Dialog>, ensuring consistent and predictable behavior.

If you'd prefer a more Vue-specific approach with full prop control, I can update the implementation accordingly.

christian-byrne:
Thanks for the clarification, let me test now a bit.

christian-byrne:
Tested and looks good. Can merge this week on 1.25. Great work.

BTW, you can follow this part of the README to setup pre-commit hooks: https://github.com/Comfy-Org/ComfyUI_frontend?tab=readme-ov-file#git-pre-commit-hooks. Then it will run formatting and linting before commits (just convenient).

Once you have a PR merged, the tests will run without needing approval from maintainer :+1:

kevinpaik1:
Thank you for your contribution @ssidharth010 !

PR #4370, opened 2025-07-07
[fix] use getter functions for sidebar tab command labels to resolve i18n collection issues

christian-byrne:
Fixes i18n collection script issues caused by PR #4213's change to use raw i18n keys in sidebar tab command labels. The collection script was generating incorrect entries with i18n keys as values instead of translated text, causing CI failures in the update-locales workflow.

**Root cause:** When sidebar tabs use raw i18n keys like `'sideToolbar.queue'` as command labels, the collection script treats these as literal strings and generates incorrect locale entries like `"menuLabels.sideToolbar_queue": "sideToolbar.queue"`.

**Solution:** Use getter functions `() => t(key)` for command labels/tooltips instead of raw i18n keys. This maintains the reactive functionality from #4213 while providing properly translated strings to the collection script, preventing incorrect locale file generation.

**Changes:**
- Modified `sidebarTabStore.ts` to use `te()` to detect i18n keys and convert them to getter functions
- Maintains backward compatibility with existing extensions using plain strings


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
-if [[ "$OS" == "Windows_NT" ]]; then
+if [ "$OS" = "Windows_NT" ]; then
```
webfiltered: Should probably revert this - it is not related.

webfiltered (changes requested):

src/locales/en/commands.json
```diff
@@ -231,19 +231,19 @@
...
-    "label": "Toggle Queue Sidebar",
-    "tooltip": "Queue"
+    "label": "sideToolbar.queue",
+    "tooltip": "sideToolbar.queue"
   },
   "Workspace_ToggleSidebarTab_workflows": {
-    "label": "Toggle Workflows Sidebar",
-    "tooltip": "Workflows"
+    "label": "sideToolbar.workflows",
+    "tooltip": "sideToolbar.workflows"
```
christian-byrne: This is happening because the i18n collection scripts expects the commands to have tooltips/labels already translated (emulating runtime). But https://github.com/Comfy-Org/ComfyUI_frontend/pull/4213 changed the system from translating in the command declaration to translating on the component, which is better but requires adjusting this script a bit now.

christian-byrne: This doesn't really have downsides though besides adding a small bit of clutter to locales files and causing previously added "unused keys" check to fail.

Pragmatic solution might be to accept the minor cosmetic issue rather than over-engineer a complex fix for something that doesn't actually break anything.

src/locales/en/commands.json
```diff
@@ -231,19 +231,19 @@
     "label": "Toggle Focus Mode"
   },
   "Workspace_ToggleSidebarTab_model-library": {
-    "label": "Toggle Model Library Sidebar",
-    "tooltip": "Model Library"
+    "label": "sideToolbar.modelLibrary",
+    "tooltip": "sideToolbar.modelLibrary"
```
webfiltered: Do these need to be `sideToolbar.modelLibrary`?  Could we just revert these to the original English?

webfiltered (changes requested):
~~Can we add an exception in the unused key check, so it also passes?~~  Oops.

src/locales/en/commands.json
webfiltered: I take back the question mark - these definitely need to be set to original english:

image

christian-byrne: Good catch. Going to push a solution that will hopefully keep the #4213 fix and restore correct i18n collection script behavior for command definitions.

christian-byrne:
**Changes:**
- Modified `sidebarTabStore.ts` to generate "Toggle X Sidebar" format for command labels
- Added comprehensive ignore patterns in `check-unused-i18n-keys.ts` for all auto-generated categories
- Maintains reactive functionality from PR #4213 while fixing locale generation

PR #4371, opened 2025-07-07
[feat] Enhanced LOD system with component-driven approach

christian-byrne:
## Summary

Enhances the Level of Detail (LOD) system to be more flexible and component-driven while keeping it simple and zoom-based.

### Key Changes

- **Added continuous `lodScore` (0-1)** to `useLOD` composable for smooth transitions instead of discrete jumps
- **Minimal global CSS optimizations** - only essential performance optimizations (disable text-shadow at distance, GPU acceleration for dragging)
- **Comprehensive developer documentation** - detailed implementation guide for widget developers and designers
- **Fixed TransformPane tests** to match current event listener implementation with options

### Benefits

- Components can now make their own LOD decisions using the continuous score
- Backwards compatible - existing discrete level system still works
- Extensive documentation helps developers implement LOD correctly
- Global CSS only optimizes expensive effects that are invisible at distance

### Documentation Added

- `LOD_IMPLEMENTATION_GUIDE.md` - Complete guide with examples for developers and designers
- `widgets/README.md` - Quick reference for widget development patterns

The enhanced system maintains the zoom-based approach while providing more granular control for individual components to implement their own LOD strategies.

Fixes need for component-driven LOD decisions in Vue nodes system.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4371-feat-Enhanced-LOD-system-with-component-driven-approach-2296d73d365081dcb20bd7e68dc42892) by [Unito](https://www.unito.io)


christian-byrne:
The idea here is to use less global CSS classes for LOD and instead expose the `useLOD` so that each component can get the current LOD level or score and decide how to render.

We might get design help on this later.

PR #4376, opened 2025-07-08
[fix] Handle missing length widget in API node pricing

christian-byrne:
Updates pricing logic to show 5s pricing when length widget is removed from nodes, while maintaining future-proof support for when the widget is added back.

The nodes now always default to 5s (128 frames) and the pricing reflects this with the appropriate costs per run.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4376-fix-Handle-missing-length-widget-in-API-node-pricing-22a6d73d36508143a4fddb680446944e) by [Unito](https://www.unito.io)


webfiltered (approved):
😂

christian-byrne:
what happened here...

christian-byrne:
Accidentally included this in https://github.com/Comfy-Org/ComfyUI_frontend/pull/4378 then rebased.

PR #4382, opened 2025-07-08
Update the frontend to support async nodes.

guill:
Also fixes a number of bugs with subgraphs

This should NOT be merged until the relevant PR to core is merged: https://github.com/comfyanonymous/ComfyUI/pull/8830

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4382-Update-the-frontend-to-support-async-nodes-22a6d73d365081eb8d58de3af670eae7) by [Unito](https://www.unito.io)


tests-ui/tests/store/executionStore.test.ts
```diff
@@ -1,11 +1,22 @@
...
+import { useWorkflowStore } from '@/stores/workflowStore'
+
+// Mock the workflowStore
+vi.mock('@/stores/workflowStore', () => ({
+  useWorkflowStore: vi.fn(() => ({
+    hierarchicalIdToNodeLocatorId: vi.fn(),
+    nodeIdToNodeLocatorId: vi.fn(),
+    nodeLocatorIdToHierarchicalId: vi.fn()
+  }))
+}))
```
guill: It's a little weird that we're mocking this here -- let me know if there's a better way to do this.

webfiltered: For unit tests, this is pretty much expected.  vitest hoists mocks to top of file.

src/types/nodeIdentification.ts
```diff
@@ -0,0 +1,127 @@
...
+ * - For root graph nodes: `<local-node-id>`
+ *
+ * Examples:
+ * - "a1b2c3d4-e5f6-7890-abcd-ef1234567890:123" (node in subgraph)
+ * - "456" (node in root graph)
+ *
+ * Unlike hierarchical IDs which change based on the instance path,
+ * NodeLocatorId remains the same for all instances of a particular node.
+ */
+export type NodeLocatorId = string & { __brand: 'NodeLocatorId' }
```
webfiltered: This is not a great TS/JS pattern, unfortunately.

Below the type guards are asserting that they are both objects and strings, when that is not the case.

src/extensions/core/groupNode.ts
```diff
@@ -1275,6 +1277,45 @@ export class GroupNodeHandler {
...
+              ...d.nodes[innerNodeId],
+              node: id
+            };
+          }
+        }
+        
+        return newProgressState;
+      }
+    );
+    */
```
webfiltered: We have a "remove commented code before merging" practice in the frontend repos.  Work for any future PRs can just be moved to a branch for the PR.

src/schemas/apiSchema.ts
```diff
@@ -113,6 +129,8 @@ const zLogRawResponse = z.object({
   entries: z.array(zLogEntry)
 })
 
+const zFeatureFlagsWsMessage = z.record(z.string(), z.any())
+
```
webfiltered: Can we split the entire feature flags feature into its own PR?  It should be completely isolated from the async feature, right?

webfiltered (changes requested):
Thassalottacode.  Looks great so far, but it is a very large PR that could be split up into its distinct features.

Will need to come back to finish up, but have paused and just left the raw notes made as I flicked through.

src/types/nodeIdentification.ts
guill: Oops, I definitely meant to remove the second part after we talked about it earlier and forgot -- my bad 😅

src/schemas/apiSchema.ts
guill: Async needs feature flags, but I'm pretty sure that feature flags don't need async -- I can take a look at turning that into a separate PR and getting it merged first. The only weird part will be that there won't actually be any features we're flagging yet, but that shouldn't harm anything.

webfiltered: Historically we just don't worry about this - version X needs min. core version Y, and that's it.  Frontend couldn't even detect the backend version.. it just threw an exception on load if it wasn't high enough.

That's why my gut feels they're distinctly isolated features - I definitely want feature flags as a step to reducing that pain, but 95% of that is for extensions.  The reality is that we're just trying to support a tiny number of edge cases - there's just not a huge benefit to new frontend working with older versions of core.

src/stores/executionStore.ts
```diff
@@ -46,7 +51,102 @@ export const useExecutionStore = defineStore('execution', () => {
...
+          mergedState.value = newState.value
+          mergedState.max = newState.max
+        } else if (newPerc < oldPerc) {
+          mergedState.value = newState.value
+          mergedState.max = newState.max
+        }
+      } else {
+        mergedState.value = newState.value
+        mergedState.max = newState.max
+      }
```
christian-byrne: So for async, we will show the slowest/bottleneck node's progress right? I think all three of these code paths do the same

```typescript
mergedState.value = newState.value
mergedState.max = newState.max
```

src/stores/executionStore.ts
```diff
@@ -310,9 +425,13 @@ export const useExecutionStore = defineStore('execution', () => {
      */
     lastExecutionError,
     /**
-     * The id of the node that is currently being executed
+     * The id of the node that is currently being executed (backward compatibility)
```
christian-byrne, line 422: Should we add `@deprecated` markers?

src/components/graph/GraphCanvas.vue
```diff
@@ -192,22 +191,26 @@ watch(
...
       } else {
         node.progress = undefined
       }
     }
-  }
+
+    // Force canvas redraw to ensure progress updates are visible
+    canvas.graph.setDirtyCanvas(true, false)
+  },
+  { deep: true }
```
christian-byrne, line 213: I think all the dependencies here are [`shallowRef`](https://vuejs.org/api/reactivity-advanced.html#shallowref) or are entirely re-assigned when changed rather than mutated (e.g., from `computed` or the merge execution progress function). We may be able to get good performance gain by removing deep if it's possible.

guill: Hmm, it *does* seem like there's a distinct difference in behavior if I don't set `deep: true`. Specifically, when `deep` is true, we get an update when leaving the subraph view. When it's false, we don't.
Reproduction step is:
1. Create a subraph with a KSampler
2. Go to the top-level graph
3. Execute the workflow
4. Once the progress bar is halfway through the subgraph node, enter the subgraph
5. Wait until the graph execution entirely completes.
6. Zoom back out to the top-level view
7. --> If `deep: true`, it looks right.
8. --> If `deep: false`, the progress bar is stuck at the state it was at when we last viewed this node -- halfway done.

webfiltered: ~~Removing the deep line entirely should have no effect on the current code.~~ (I blame lack of caffeine)

`nodeLocationProgressStates` is an array; so watching it `deep: false` without replacing the array itself would explain your symptoms.

Solution: use `deep: 1` - or whatever level you need.

christian-byrne: Got it, makes sense!

webfiltered:
p.s. you can just revert the `Update locales` commit *before rebasing - it will regenerate and fix its own merge conflicts. 👍

src/types/nodeIdentification.ts
```diff
@@ -0,0 +1,123 @@
...
+    return {
+      subgraphUuid: null,
+      localNodeId: isNaN(Number(id)) ? id : Number(id)
+    }
+  }
+
+  const [subgraphUuid, localNodeId] = parts
+  return {
+    subgraphUuid,
+    localNodeId: isNaN(Number(localNodeId)) ? localNodeId : Number(localNodeId)
```
webfiltered, line 86:
```suggestion
    localNodeId
```

nit: Remove redundant code

src/types/nodeIdentification.ts
```diff
@@ -0,0 +1,123 @@
...
+): { subgraphUuid: string | null; localNodeId: NodeId } | null {
+  if (!isNodeLocatorId(id)) return null
+
+  const parts = id.split(':')
+
+  if (parts.length === 1) {
+    // Simple node ID (root graph)
+    return {
+      subgraphUuid: null,
+      localNodeId: isNaN(Number(id)) ? id : Number(id)
```
webfiltered, line 79:
```suggestion
      localNodeId: id
```

nit: Remove redundant code

src/types/nodeIdentification.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Parse a NodeExecutionId into its component node IDs
+ * @param id The NodeExecutionId to parse
+ * @returns Array of node IDs from root to target, or null if not an execution ID
+ */
+export function parseNodeExecutionId(id: string): NodeId[] | null {
+  if (!isNodeExecutionId(id)) return null
+
+  return id
+    .split(':')
+    .map((part) => (isNaN(Number(part)) ? part : Number(part)))
```
webfiltered, line 113:
```suggestion
  return id.split(':')
```

nit: Remove redundant code

src/stores/workflowStore.ts
```diff
@@ -488,6 +505,136 @@ export const useWorkflowStore = defineStore('workflow', () => {
...
+      targetUuid: string,
+      path: NodeId[] = []
+    ): NodeId[] | null => {
+      if (isSubgraph(graph) && graph.id === targetUuid) {
+        return path
+      }
+
+      for (const node of graph._nodes) {
+        if (node.isSubgraphNode?.() && (node as any).subgraph) {
+          const result = findSubgraphPath((node as any).subgraph, targetUuid, [
```
webfiltered:
```suggestion
        if (node.isSubgraphNode() && node.subgraph) {
          const result = findSubgraphPath(node.subgraph, targetUuid, [
```

nit: Remove redundant type assertions / null check.

src/stores/workflowStore.ts
```diff
@@ -488,6 +505,136 @@ export const useWorkflowStore = defineStore('workflow', () => {
...
+    targetSubgraph?: Subgraph
+  ): NodeExecutionId | null => {
+    const parsed = parseNodeLocatorId(locatorId)
+    if (!parsed) return null
+
+    const { subgraphUuid, localNodeId } = parsed
+
+    // If no subgraph UUID, this is a root graph node
+    if (!subgraphUuid) {
+      return String(localNodeId) as NodeExecutionId
```
webfiltered:
```suggestion
      return String(localNodeId)
```

nit: Remove redundant type assertion

src/stores/workflowStore.ts
```diff
@@ -488,6 +505,136 @@ export const useWorkflowStore = defineStore('workflow', () => {
...
+   * Convert an execution ID to a NodeLocatorId
+   * @param nodeExecutionId The execution node ID (e.g., "123:456:789")
+   * @returns The NodeLocatorId or null if conversion fails
+   */
+  const nodeExecutionIdToNodeLocatorId = (
+    nodeExecutionId: NodeExecutionId | string
+  ): NodeLocatorId | null => {
+    // Handle simple node IDs (root graph - no colons)
+    if (!nodeExecutionId.includes(':')) {
+      return nodeExecutionId as NodeLocatorId
```
webfiltered:
```suggestion
      return nodeExecutionId
```

nit: Remove redundant type assertion

src/stores/workflowStore.ts
```diff
@@ -488,6 +505,136 @@ export const useWorkflowStore = defineStore('workflow', () => {
...
+
+    const parts = parseNodeExecutionId(nodeExecutionId)
+    if (!parts || parts.length === 0) return null
+
+    const nodeId = parts[parts.length - 1]
+    const subgraphNodeIds = parts.slice(0, -1)
+
+    if (subgraphNodeIds.length === 0) {
+      // Node is in root graph, return the node ID as-is
+      return String(nodeId) as NodeLocatorId
```
webfiltered:
```suggestion
      return String(nodeId)
```

nit: Remove redundant type assertion

src/stores/workflowStore.ts
```diff
@@ -488,6 +505,136 @@ export const useWorkflowStore = defineStore('workflow', () => {
...
+   * @returns The NodeLocatorId (for root graph nodes, returns the node ID as-is)
+   */
+  const nodeIdToNodeLocatorId = (
+    nodeId: NodeId,
+    subgraph?: Subgraph
+  ): NodeLocatorId => {
+    const targetSubgraph = subgraph ?? activeSubgraph.value
+    if (!targetSubgraph) {
+      // Node is in the root graph, return the node ID as-is
+      return String(nodeId) as NodeLocatorId
```
webfiltered:
```suggestion
      return String(nodeId)
```

nit: Remove redundant type assertion

src/stores/workflowStore.ts
```diff
@@ -488,6 +505,136 @@ export const useWorkflowStore = defineStore('workflow', () => {
...
+  ): NodeLocatorId | null => {
+    // Handle simple node IDs (root graph - no colons)
+    if (!nodeExecutionId.includes(':')) {
+      return nodeExecutionId as NodeLocatorId
+    }
+
+    const parts = parseNodeExecutionId(nodeExecutionId)
+    if (!parts || parts.length === 0) return null
+
+    const nodeId = parts[parts.length - 1]
```
webfiltered, line 543:
```suggestion
    const nodeId = parts.at(-1)
```

nit: Prefer idiomatic `at(-1)` ([existing use here](https://github.com/Comfy-Org/ComfyUI_frontend/blob/7fcc39efe3ef5bb709fd114db813b71a4d218305/src/stores/workflowStore.ts#L502))

src/stores/executionStore.ts
```diff
@@ -46,7 +51,97 @@ export const useExecutionStore = defineStore('execution', () => {
...
+   * @returns The NodeLocatorId
+   */
+  const executionIdToNodeLocatorId = (
+    nodeId: string | number
+  ): NodeLocatorId => {
+    const nodeIdStr = String(nodeId)
+
+    if (!nodeIdStr.includes(':')) {
+      // It's a top-level node ID
+      return nodeIdStr as NodeLocatorId
```
webfiltered:
```suggestion
      return nodeIdStr
```

nit: Remove redundant type assertion

src/stores/executionStore.ts
```diff
@@ -46,7 +51,97 @@ export const useExecutionStore = defineStore('execution', () => {
...
+    return mergedState
+  }
+
+  const nodeLocationProgressStates = computed<
+    Record<NodeLocatorId, NodeProgressState>
+  >(() => {
+    const result: Record<NodeLocatorId, NodeProgressState> = {}
+
+    const states = nodeProgressStates.value // Apparently doing this inside `Object.entries` causes issues
+    for (const [_, state] of Object.entries(states)) {
```
webfiltered:
```suggestion
    for (const state of Object.values(states)) {
```

Simplify code

src/types/nodeIdentification.ts
```diff
@@ -0,0 +1,123 @@
...
+}
+
+/**
+ * Create a NodeExecutionId from an array of node IDs
+ * @param nodeIds Array of node IDs from root to target
+ * @returns A properly formatted NodeExecutionId
+ */
+export function createNodeExecutionId(nodeIds: NodeId[]): NodeExecutionId {
+  return nodeIds.join(':') as NodeExecutionId
```
webfiltered:
```suggestion
  return nodeIds.join(':')
```

nit: Remove redundant type assertion

src/types/nodeIdentification.ts
```diff
@@ -0,0 +1,123 @@
...
+ * Create a NodeLocatorId from components
+ * @param subgraphUuid The UUID of the immediate containing subgraph
+ * @param localNodeId The local node ID within that subgraph
+ * @returns A properly formatted NodeLocatorId
+ */
+export function createNodeLocatorId(
+  subgraphUuid: string,
+  localNodeId: NodeId
+): NodeLocatorId {
+  return `${subgraphUuid}:${localNodeId}` as NodeLocatorId
```
webfiltered:
```suggestion
  return `${subgraphUuid}:${localNodeId}`
```

nit: Remove redundant type assertion

src/scripts/domWidget.ts
```diff
@@ -245,7 +246,9 @@ export class ComponentWidgetImpl<
     })
     this.component = obj.component
     this.inputSpec = obj.inputSpec
-    this.props = obj.props
+    this.props = obj.componentProps
+      ? ({ ...obj.props, ...obj.componentProps } as P)
+      : obj.props
```
webfiltered: A valid compiler failure has been type-asserted away here.

```ts
// Assuming
P: { required: string }
```

But the input `obj` can have **both** an undefined `props` and `componentProps` can be `{}`.

`this.props` is now `{}` - does not have `required` property.

Can be fixed by removing `Partial` wrapper from `componentProps`.  But this does not feel like the intent of this code.

src/composables/useBrowserTabTitle.ts
```diff
@@ -36,11 +37,34 @@ export const useBrowserTabTitle = () => {
...
+    // If multiple nodes are running
+    if (runningNodes.length > 1) {
+      return `${executionText.value}[${runningNodes.length} ${t('g.nodesRunning', 'nodes running')}]`
+    }
+
+    // If only one node is running
+    const [nodeId, state] = runningNodes[0]
+    const progress = Math.round((state.value / state.max) * 100)
+    const nodeType =
+      executionStore.activePrompt?.workflow?.changeTracker?.activeState?.nodes.find(
```
webfiltered:
```suggestion
      executionStore.activePrompt?.workflow?.changeTracker?.activeState.nodes.find(
```

Very informative; I left with one less question.

src/components/graph/GraphCanvas.vue
```diff
@@ -192,22 +191,26 @@ watch(
   }
 )
 
-// Update the progress of the executing node
+// Update the progress of executing nodes
 watch(
   () =>
```
webfiltered, line 196:
```suggestion
```

Part 1...

src/components/graph/GraphCanvas.vue
```diff
@@ -192,22 +191,26 @@ watch(
...
-    [
-      executionStore.executingNodeId,
-      executionStore.executingNodeProgress
-    ] satisfies [NodeId | null, number | null],
-  ([executingNodeId, executingNodeProgress]) => {
-    for (const node of comfyApp.graph.nodes) {
-      if (node.id == executingNodeId) {
-        node.progress = executingNodeProgress ?? undefined
+    [executionStore.nodeLocationProgressStates, canvasStore.canvas] as const,
```
webfiltered, line 197:
```suggestion
  [executionStore.nodeLocationProgressStates, canvasStore.canvas] as const,
```

Part 2...

Subtle difference - this is telling Vue to watch multiple objects, rather than asking it to watch a new array of reactive objects every time.

webfiltered (changes requested):
This is 90% just removing type assertions.  Using 'as' does much more than e.g. specifying a function return type.  The latter holds you to the contract - `as` skips a lot of type checking.

All except one of these should be click-to-commit batchable, for your convenience.

src/stores/workflowStore.ts
guill: For some reason, the TypeScript type checker isn't smart enough to know that the result of `parts.at(-1)` is non-null when we have an array of `string | int` of size > 0, so it throws a type-checking error in the pre-commit checks. (It IS smart enough when we use indexing).

Is it worth changing to `.at(-1)` if that means I need to add a type assertion?

src/components/graph/GraphCanvas.vue
guill: For some reason, making this change causes things to not work. Specifically, we stop receiving updates when the `nodeLocationProgressStates` gets updated.

src/types/nodeIdentification.ts
guill: It doesn't look like this code is redundant. It's what causes a local node id of `"12345"` (the string) to be evaluated to `12345` (the number). I tried making this change, but that's now causing some unit tests to fail.

I could just update the unit tests to accept that all node ids are strings (which seems reasonable), but there's a lot of explicit type hinting that node ids can be strings or numbers, so I feel like this was intentional 🤔

src/components/graph/GraphCanvas.vue
webfiltered: It's an array.  Sorry!

It can probably stay as-is if it's not hurting anything.

I think the actual clean-up here would be to just not (deep) watch LGraphCanvas:

```ts
() => executionStore.nodeLocationProgressStates
```

You can just assign it / check nullability inside the watcher.

src/types/nodeIdentification.ts
webfiltered: The reason we accept strings and numbers is because old frontend code was written in vanilla JS, and took advantage of JS' implicit type casting ("coercion") in many places.

e.g. *some* JS functions will take either `"12"` or `12` and do the same thing (treat as number).  Others will not.  And those number op functions will *output* a number instead of a string - so ID got set to number in places.  The same happens in reverse for strings (which iirc, was the original problem).  And so, `NodeId` type was born.

The problem here is that if we have this function return NodeId, and it's maybe a string or maybe a number - it feels logical that we should be able to search for it:

```ts
const node = nodes.find(x => x.id === localNodeId)
```

#### TL;DR

Please disregard & leave this as number.  It is more expected to be a number than a string.

If you feel that `isNaN` is clear enough here, you can remove the `Number()` nested inside it; it's gonna do that whether you want it to or not.

p.s. Optional reading:

`isNan` itself is a steaming pile that should _usually_ be avoided.  Unless of course you're explicitly testing for something that matches its [bonky behaviour](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN#description) - even then, it's usually clearer to use other things.

webfiltered: p.p.s. `parseInt("123", 10)` is how I'd usually approach numeric parsing in this hellscape, but is not strictly necessary / can be too verbose.

src/stores/workflowStore.ts
webfiltered: > It IS smart enough when we use indexing

Untrue!  It just literally never checks this.  Try giving it an index of `-1` - or `Infinity`.  =)

There's a modern compiler option to enable "unchecked index" access, that would catch this (makes it behave exactly like `at()`).  We have not enabled it because of the hundreds of errors that would need to be fixed.

webfiltered: This is the cleanest way to refactor this to cover failure cases, imo:

```ts
const nodeId = parts?.at(-1)
if (nodeId === undefined) return null
```

The only real differences is if you were to do either of the below, it will return null here instead of returning `"undefined"` as the node ID later:

```ts
// Bad example of difference in edge case
parts[parts.length - 1] = undefined
// Or
delete parts[parts.length - 1]
```

src/types/nodeIdentification.ts
guill: 👍 I'll change it to `parseInt`. Now that you mention it, with `Number` I think you could technically make a workflow with nodes with ids `"10"`, `"1e1"`, and `"0xa"` and cause all sorts of... unexpected features.

src/scripts/domWidget.ts
guill: This part was changed in order to be able to pass the `parentNodeId` into the `TextPreviewWidget` within `useProgressTextWidget.ts`. Is there a better way to do this in Vue while maintaining type safety?

src/types/nodeIdentification.ts
guill: Hmm, actually, `parseInt` has its own issues. If you call `parseInt("25:3:9", 10)`, you just get `25`. I searched a bit through custom nodes, and it looks like most of them just use `Number`, so if it's compatibility we're concerned about, that's probably the best path.
Long term, I think we should consider just saying that node ids are opaque strings. With the addition of subgraphs, people will already have to handle non-number ids anyway.

src/components/graph/GraphCanvas.vue
christian-byrne: Shall we just table this for now and revisit later?

src/scripts/domWidget.ts
christian-byrne: Created a PR targeting this PR's branch to fix this:

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/4499

christian-byrne (approved):
LGTM!

webfiltered (approved):

PR #4360, opened 2025-07-05
Vue Node Widgets

benceruleanlu:
Vue widgets for the vue node system.

Color picker still doesn't have alpha support
Colors don't match design yet, coming in its own design system PR later
Fileupload widget only supports images, needs changes in core for others

... <= [bl-body](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4387) <= [bl-widgets](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4360) <= [bl-slots](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4515)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4360-Layout-and-formatting-of-widgets-2276d73d365081d9ad52f35e33a90fec) by [Unito](https://www.unito.io)

PR #4387, opened 2025-07-08
Vue Node Body

benceruleanlu:
Adds node header title editing and collapse, synced properly with litegraph.

Also has some formatting changes.

... <= [bl-body](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4387) <= [bl-widgets](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4360) <= [bl-slots](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4515)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4387-feat-Vue-based-node-rendering-system-with-TransformPane-22a6d73d365081a79889dc2e98501853) by [Unito](https://www.unito.io)


src/components/graph/vueNodes/NodeHeader.vue
```diff
@@ -4,40 +4,27 @@
...
+    class="lg-node-header flex items-center justify-between p-2 rounded-t-lg cursor-move -mt-[30px]"
     :style="{
       backgroundColor: headerColor,
       color: textColor
     }"
     @dblclick="handleDoubleClick"
   >
+    <!-- Collapse/Expand Button -->
+    <button
+      v-if="!readonly"
```
christian-byrne: If the node is readonly, should we make the button disabled or hidden?

benceruleanlu: disabled IMO, made the change

benceruleanlu: Actually, I'm not too sure

I changed my mind, and now think hidden would lead to a cleaner UI, especially with a ton of nodes around

But it's not terribly important to have to ask design about it

christian-byrne: True, hard to decide. Either way seems good to me.

src/components/graph/vueNodes/NodeHeader.vue
```diff
@@ -4,39 +4,33 @@
   </div>
   <div
     v-else
-    class="lg-node-header flex items-center justify-between px-3 py-2 rounded-t cursor-move"
+    class="lg-node-header flex items-center justify-between p-2 rounded-t-lg cursor-move -mt-[30px]"
```
christian-byrne: Shall we use the properties from litegraph style like `NODE_TITLE_HEIGHT`?

See also

https://github.com/Comfy-Org/ComfyUI_frontend/blob/bl-body/src/components/graph/TitleEditor.vue#L76-L77

and

https://github.com/Comfy-Org/litegraph.js/blob/3db44707faeb681c946f8deffdc37aae1468655e/src/LiteGraphGlobal.ts#L43

src/components/graph/vueNodes/NodeHeader.vue
```diff
@@ -73,8 +69,14 @@ onErrorCaptured((error) => {
   return false
 })
 
+// Editing state
+const isEditing = ref(false)
+
 const nodeInfo = computed(() => props.nodeData || props.node)
 
+// Local state for title to provide immediate feedback
+const displayTitle = ref(nodeInfo.value?.title || 'Untitled')
```
christian-byrne, line 83: Should we also sync this with the nodeInfo? I.e., watch for changes to nodeInfo.value?.title and on change write back to `displayTitle`?

src/components/graph/vueNodes/LGraphNode.vue
```diff
@@ -143,6 +146,9 @@ onErrorCaptured((error) => {
 // Track dragging state for will-change optimization
 const isDragging = ref(false)
 
+// Track collapsed state
+const isCollapsed = ref(props.nodeData.flags?.collapsed ?? false)
```
christian-byrne, line 147: Same as other comment, can this be reactive to nodeData changes?

src/components/graph/GraphCanvas.vue
```diff
@@ -458,6 +460,29 @@ const handleNodeSelect = (event: PointerEvent, nodeData: VueNodeData) => {
...
+// Handle node collapse state changes
+const handleNodeCollapse = (nodeId: string, collapsed: boolean) => {
+  if (!nodeManager) return
+
+  const node = nodeManager.getNode(nodeId)
+  if (!node) return
+
+  // Sync collapsed state back to LiteGraph node
+  node.flags = node.flags || {}
+  node.flags.collapsed = collapsed
```
christian-byrne: Would it be better to call `node.collapse()` method instead of setting flag? I think it might be a bit more maintainable at least. WDYT?

Code for reference:

https://github.com/Comfy-Org/litegraph.js/blob/3db44707faeb681c946f8deffdc37aae1468655e/src/LGraphNode.ts#L3132-L3138

src/components/graph/vueNodes/NodeHeader.vue
```diff
@@ -110,7 +112,18 @@ const handleCollapse = () => {
...
+    isEditing.value = true
+  }
+}
+
+const handleTitleEdit = (newTitle: string) => {
+  isEditing.value = false
+  const trimmedTitle = newTitle.trim()
+  if (trimmedTitle && trimmedTitle !== displayTitle.value) {
+    // Update local state immediately for instant feedback
+    displayTitle.value = trimmedTitle
```
christian-byrne: Nit: might be better to let the "parent" have authority over state and the displayTitle just rects to it. I.e., could we remove this line and just emit => react to litegraph's changes after receiving the emit?

src/components/graph/vueNodes/NodeHeader.vue
benceruleanlu: Sure, but this will only catch external changes vue-side, which I don't know if it ever happens yet.

benceruleanlu:
@christian-byrne Fixed based on your review comments, but do note that there is currently no reactivity between litegraph and Vue.

Are we going to have a pseudo fix for that soon, or should I update this PR to manually sync with litegraph so litegraph -> vue changes are synced properly?

webfiltered:
Did someone say litegraph?  👂

If you want a quick architectural sounding-board chat on how to make some interop work, I have one good eye and am down for discussion.

christian-byrne:
> If you want a quick architectural sounding-board chat on how to make some interop work, I have one good eye and am down for discussion.

That would be very helpful. We are looking to potentially add some new event emits on litegraph. How's Monday evening after v3 meeting?

PR #4389, opened 2025-07-08
[api] update getHistory to call history_v2

ric-yu:
## Summary
  Updates the history API to use a cleaner array format where each history item contains its `prompt_id` as a property rather than being wrapped in an
  object with the `prompt_id` as a key.

Backend PR: https://github.com/comfyanonymous/ComfyUI/pull/8844

  ## Changes Made

  ### API Response Format Update
  - **Before**: `{history: [{"prompt_id_1": {...}}, {"prompt_id_2": {...}}]}`
  - **After**: `{history: [{prompt_id: "prompt_id_1", ...}, {prompt_id: "prompt_id_2", ...}]}`

  ### Schema Changes (`src/schemas/apiSchema.ts`)
  - Added `prompt_id` field to `zRawHistoryItem` schema
  - Updated `zHistoryResponse` to expect a direct array of history items instead of an array of objects with dynamic keys
  - Added proper TypeScript types for the new response format

  ### API Implementation (`src/scripts/api.ts`)
  - Updated `getHistory()` method to use the new `/history_v2` endpoint
  - Simplified data extraction logic to directly map the history array
  - Removed unnecessary object key extraction since `prompt_id` is now a direct property
  - Removed unused `RawHistoryItem` import

  ### Test Coverage (`tests-ui/tests/scripts/api.test.ts`)
  - Added comprehensive test suite for the new history API format
  - Tests cover:
    - Array format handling from `/history_v2`
    - Empty history arrays
    - Error handling
    - Correct API endpoint usage
    - Default parameter behavior

  ## Test Results
  All tests pass successfully, confirming the API correctly handles the new format while maintaining backward compatibility for error scenarios.

image

### DIFF
--- a/src/scripts/api.ts
+++ b/src/scripts/api.ts
@@ -743,13 +743,13 @@ export class ComfyApi extends EventTarget {
       const json = await res.json()
 
       // The /history_v2/{prompt_id} endpoint returns data for a specific prompt
-      // The response format is: { prompt_id: { prompt: [...], outputs: {...}, status: {...} } }
+      // The response format is: { prompt_id: { prompt: {priority, prompt_id, extra_data}, outputs: {...}, status: {...} } }
       const historyItem = json[prompt_id]
       if (!historyItem) return null
 
-      // Extract workflow from the prompt array
-      // prompt[3] contains extra_data which has extra_pnginfo.workflow
-      const workflow = historyItem.prompt?.[3]?.extra_pnginfo?.workflow
+      // Extract workflow from the prompt object
+      // prompt.extra_data contains extra_pnginfo.workflow
+      const workflow = historyItem.prompt?.extra_data?.extra_pnginfo?.workflow
       return workflow || null
     } catch (error) {
       console.error(`Failed to fetch workflow for prompt ${prompt_id}:`, error)
--- a/src/scripts/api.ts
+++ b/src/scripts/api.ts
@@ -11,6 +11,7 @@ import type {
   ExecutionStartWsMessage,
   ExecutionSuccessWsMessage,
   ExtensionsResponse,
+  HistoryResponse,
   HistoryTaskItem,
   LogsRawResponse,
   LogsWsMessage,
@@ -23,6 +24,7 @@ import type {
   StatusWsMessage,
   StatusWsMessageStatus,
   SystemStats,
+  TaskPrompt,
   User,
   UserDataFullInfo
 } from '@/schemas/apiSchema'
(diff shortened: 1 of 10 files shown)


christian-byrne:
We can definitely upgrade vite. @jtydhr88 were you able to test vite 7 yet?

ric-yu:
> We can definitely upgrade vite. @jtydhr88 were you able to test vite 7 yet?

Makes sense, I'll leave it outside the scope of this PR for now, since it required a change in vite.config.mts

jtydhr88:
@christian-byrne previously I tested on vite 6.3.5, and it needs some dependencies upgrade as well, I could send a draft PR for your reference and testing

src/scripts/api.ts
```diff
@@ -710,13 +712,21 @@ export class ComfyApi extends EventTarget {
     max_items: number = 200
   ): Promise<{ History: HistoryTaskItem[] }> {
     try {
-      const res = await this.fetchApi(`/history?max_items=${max_items}`)
-      const json: Promise<HistoryTaskItem[]> = await res.json()
+      const res = await this.fetchApi(`/history_v2?max_items=${max_items}`)
```
christian-byrne, line 715: Will this endpoint be added to open-source ComfyUI core server?

ric-yu: Yessir https://github.com/comfyanonymous/ComfyUI/pull/8844

christian-byrne: Okay nice. For the discovery, you can also access the version of ComfyUI via `/system_stats` or the `systemStatsStore` in our codebase. In general we don't worry about supporting old versions of the backend (relative to frontend version) since comfyui_frontend_package is a dependency of ComfyUI and not used with any other backends. However, I may be missing some context.

src/locales/en/settings.json
```diff
@@ -232,10 +232,6 @@
   "Comfy_NodeBadge_ShowApiPricing": {
     "name": "Show API node pricing badge"
   },
-  "Comfy_Notification_ShowVersionUpdates": {
-    "name": "Show version updates",
-    "tooltip": "Show updates for new models, and major new features."
-  },
```
webfiltered: @ric-yu Feel free to mark as resolved if already resolved - I'm sure I saw this brought up somewhere else.

If not, @christian-byrne can you please confirm this can be ignored after your CI changes?

webfiltered (dismissed):
LGTM.

The only feedback I have is that it wasn't immediately obvious what the actual goal of the PR was - but the link to core PR took care of that. 👍

christian-byrne:
Backend change needs to happen first

webfiltered:
> Backend change needs to happen first

And this does higlight the need for integration tests again.  Issue'd.

- https://github.com/Comfy-Org/ComfyUI_frontend/issues/4402

ric-yu:
Closing - will just sort history by exec start

christian-byrne:
Stop force-pushing

christian-byrne:
We use `/jobs` now so this is no longer relevant.

PR #4390, opened 2025-07-08
[System Pop Up] Hide hidden and deprecated settings from search results

cloudmemox:
## Summary
Fixes setting search to exclude hidden and deprecated settings from results, maintaining consistency with the normal settings tree behavior.

## Changes
- Added filter in `useSettingSearch.ts` to exclude settings with `type: 'hidden'` or `deprecated: true`
- Added comprehensive unit tests and playwright tests for setting search functionality

## Before
Hidden and deprecated settings appeared in search results even though they're not shown in the normal settings tree.

## After
Search results now only show visible, non-deprecated settings, matching the expected behavior.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4390-System-Pop-Up-Hide-hidden-and-deprecated-settings-from-search-results-22a6d73d365081329065f068e9632f4a) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/useSettingSearch.spec.ts
+++ b/browser_tests/tests/useSettingSearch.spec.ts
@@ -1,0 +1,289 @@
+import { expect } from '@playwright/test'
+
+import { comfyPageFixture as test } from '../fixtures/ComfyPage'
+
+test.describe('Settings Search functionality', () => {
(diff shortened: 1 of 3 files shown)


christian-byrne (approved):
Amazing! LGTM!

PR #4392, opened 2025-07-09
Revert subgraphs

webfiltered:
Reverts subgraphs, including litegraph to 15.5.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4392-Revert-subgraphs-22b6d73d3650810b8bb5d35721a1d0ef) by [Unito](https://www.unito.io)


webfiltered:
Closing; instead we will have several `@latest` versions where group nodes do not work.  This will be fixed before stable is released.

PR #4393, opened 2025-07-09
Add prompt ID to interrupt API call

deepanjanroy:
The Comfy UI backend currently ignores the post body.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4393-Add-prompt-ID-to-interrupt-API-call-22b6d73d365081d4b9b1e82dcc3b3c27) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/api.ts
+++ b/src/scripts/api.ts
@@ -689,7 +689,8 @@ export class ComfyApi extends EventTarget {
         Running: data.queue_running.map((prompt: Record<number, any>) => ({
           taskType: 'Running',
           prompt,
-          remove: { name: 'Cancel', cb: () => api.interrupt() }
+          // prompt[1] is the prompt id
+          remove: { name: 'Cancel', cb: () => api.interrupt(prompt[1]) }
         })),
         Pending: data.queue_pending.map((prompt: Record<number, any>) => ({
           taskType: 'Pending',
@@ -770,10 +771,15 @@ export class ComfyApi extends EventTarget {
   }
 
   /**
-   * Interrupts the execution of the running prompt
+   * Interrupts the execution of the running prompt. If runningPromptId is provided,
+   * it is included in the payload as a helpful hint to the backend.
+   * @param {string | null} [runningPromptId] Optional Running Prompt ID to interrupt
    */
-  async interrupt() {
-    await this.#postItem('interrupt', null)
+  async interrupt(runningPromptId: string | null) {
+    await this.#postItem(
+      'interrupt',
+      runningPromptId ? { prompt_id: runningPromptId } : undefined
+    )
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/scripts/api.ts
```diff
@@ -689,7 +689,8 @@ export class ComfyApi extends EventTarget {
         Running: data.queue_running.map((prompt: Record<number, any>) => ({
           taskType: 'Running',
           prompt,
-          remove: { name: 'Cancel', cb: () => api.interrupt() }
+          // prompt[1] is the prompt id
+          remove: { name: 'Cancel', cb: () => api.interrupt(prompt[1]) }
```
deepanjanroy, line 693: prompt[1] is not the most readable, but this bit of code is only used by the legacy UI and it's somewhat complicated to have nicer types here, so I did a low-effort thing.

christian-byrne:
Hi @deepanjanroy, will this change be later supported by a change to the `/interrupt` endpoint on the HTTP server? Or will the change be dependent on the backend type?

Currently, the `/queue` and `/history` endpoint allow cancelling or deleting items by id when passing `delete` key:

- cancel queue item by id: https://github.com/comfyanonymous/ComfyUI/blob/181a9bf26d4445e160645f6c81dc2ee29e7b6a08/server.py#L677-L681
- delete history item by id: https://github.com/comfyanonymous/ComfyUI/blob/181a9bf26d4445e160645f6c81dc2ee29e7b6a08/server.py#L707-L710

On the frontend, we create abstracted `deleteItem` method on our public api module: https://github.com/Comfy-Org/ComfyUI_frontend/blob/f5b03f323df0b9d3a0ef5f14999765527bfd17bb/src/scripts/api.ts#L755-L762

Is there a known limitation for these approaches? Or is this something specific for cloud?

deepanjanroy:
Hi @christian-byrne! We needed it for cloud yes, don't think we need to support it right now on regular Comfy backend. The regular comfy backend only supports running one task at a time so it doesn't need to know the prompt id. (cc @robinjhuang if I got anything wrong)

Re: why not use cancel queue / delete history: I think none of those cancel a currently executing task. That's exclusively handled by interrupt.

robinjhuang:
Yeah echoing what Deep said. Interrupt is meant to cancel currently running jobs, so the intention is different from cancel (cancel items in the queue). Is this a harmless change on FE (even if ComfyUI backend doesn't use it for now)? Or are there potential side effects? @christian-byrne 

Edit: FYI this is not urgent at all so it can wait.

robinjhuang:
Looks like there's a PR to implement this in core too. https://github.com/comfyanonymous/ComfyUI/pull/8308/files

christian-byrne:
Okay makes sense, thanks for explanation.

I don't forsee side effects. This could be merged currently if necessary.

deepanjanroy:
@christian-byrne / any other code owners: would you mind approving the PR and merging it if there are no other blockers?

webfiltered (approved):

PR #4394, opened 2025-07-09
[fix] Sync subgraph node title changes with breadcrumb navigation

christian-byrne:
Fixes an issue where the breadcrumb navigation didn't update when subgraph node titles were changed via double-click editing. The fix ensures title changes are synchronized across all data structures for proper reactivity and persistence.

https://github.com/user-attachments/assets/cb357502-f263-49b1-898e-8e5acb85efb6

closes https://github.com/Comfy-Org/ComfyUI_frontend/issues/4178#

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4394-fix-Sync-subgraph-node-title-changes-with-breadcrumb-navigation-22b6d73d3650818eb4b7cf40de47c646) by [Unito](https://www.unito.io)


christian-byrne:
Forgot to commit the test workflow

webfiltered (changes requested):
Current design is directly editing the internal state of `changeTracker`, bypassing the API.  Is there a limitation that requires this?  We should not need something this low-level for a simple title change.

src/locales/en/settings.json
```diff
@@ -322,6 +322,7 @@
   },
   "Comfy_UseNewMenu": {
     "name": "Use new menu",
+    "tooltip": "Menu bar position. On mobile devices, the menu is always shown at the top.",
```
webfiltered: So.. i18n CI - fixed?  >.>

src/components/graph/TitleEditor.vue
```diff
@@ -41,7 +41,15 @@ const previousCanvasDraggable = ref(true)
 
 const onEdit = (newValue: string) => {
   if (titleEditorStore.titleEditorTarget && newValue.trim() !== '') {
-    titleEditorStore.titleEditorTarget.title = newValue.trim()
+    const trimmedTitle = newValue.trim()
+    titleEditorStore.titleEditorTarget.title = trimmedTitle
+
+    // If this is a subgraph node, sync the runtime subgraph name for breadcrumb reactivity
+    const target = titleEditorStore.titleEditorTarget
+    if (target instanceof LGraphNode && target.isSubgraphNode?.()) {
```
webfiltered, line 49:
```suggestion
    if (target instanceof LGraphNode && target.isSubgraphNode()) {
```

nit - remove redundant code.  We should auto-lint these.

webfiltered (approved):

webfiltered (changes requested):
Looks like tests failed.

src/locales/en/settings.json
christian-byrne: Haha... mistakes were made.

This tooltip was just changed in recent PR but the PR was made when the `update-locales` action was exiting early because pre-commit was still failing excessively. Should be fixed for any PR with the final commit made after yesterday

src/components/graph/TitleEditor.vue
christian-byrne: Apparently this rule will catch these https://typescript-eslint.io/rules/no-unnecessary-condition/

Should we add it?

webfiltered: Appears it isn't for optional chaining?  But a good rule regardless.

webfiltered (dismissed):

src/components/graph/TitleEditor.vue
christian-byrne: I'm not sure, I just read it on stackoverflow.

christian-byrne: They have it in their examples on the docs for the rule, but i can verify before PR

Selection_1684

webfiltered: Nope; you're right.  I just stopped reading after the bullet points, thinking the rule criteria would contain.. you know.. 🤦

christian-byrne:
Ah, tests were failing because of the classic "creating new browser tests file and forgetting to set `UseNewMenu: Top`."

I need to add a line in CLAUDE.md and cursorrules for that.

src/components/graph/TitleEditor.vue
christian-byrne: I tested adding the rule and there's 391 linting errors violating it in current code.

They are mostly auto-fixable, but then I realized: in many places, OUR code can definitively say that something is never null, but the code from extensions is not compiled against our types. So, the null checks are unnecesary from naive perspective of our compilation stage, but often added in reality because of custom nodes.

I wonder if there's realistically anything we can do about that? WDYT?

christian-byrne:
Going to add the update to nodeDefStore in a followup PR.

webfiltered (approved):

src/components/graph/TitleEditor.vue
webfiltered: If we keep using optional chaining without documenting it via type or even in comments, we are definitely going to end up with unexpected bugs in extensions later on - from seemingly-innocent, minor changes.

My (very strong) preference is that we throw very large, very loud exceptions whenever that happens - so anyone doing dev gets an explicit "something is wrong, here is exacltly where".

A great way to throw such exceptions is by removing optional chaining.  😄

I am leaning towards not "ripping the bandaid off" here - 391 instances is .. a lot.

PR #4407, opened 2025-07-10
Sort history by execution start

ric-yu:
## Context 

/history currently does not support pagination in its response. In addition, JSON objects are generally expected to not be ordered.

## Changes made

- Sort histories by execution_start if available
- Histories without an execution_start time appear at the start, they are most likely recently queued
- Added unit tests for history sorting

## Test plan

1. Shuffle outputs in local comfyUI backend
image

2. Queue is still in chronological order despite /history response being out of order

image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4407-Sort-history-by-execution-start-22c6d73d36508111bbe7f97784334a9b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/api.ts
+++ b/src/scripts/api.ts
@@ -712,11 +712,26 @@ export class ComfyApi extends EventTarget {
     try {
       const res = await this.fetchApi(`/history?max_items=${max_items}`)
       const json: Promise<HistoryTaskItem[]> = await res.json()
+      const historyItems = Object.values(json).map((item) => ({
+        ...item,
+        taskType: 'History'
+      }))
+
+      // Sort by execution_start timestamp in descending order (newest first)
+      historyItems.sort((a, b) => {
+        const getExecutionStartTimestamp = (item: HistoryTaskItem) => {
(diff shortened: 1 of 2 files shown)


src/scripts/api.ts
```diff
@@ -712,11 +712,26 @@ export class ComfyApi extends EventTarget {
...
       const json: Promise<HistoryTaskItem[]> = await res.json()
+      const historyItems = Object.values(json).map((item) => ({
+        ...item,
+        taskType: 'History'
+      }))
+
+      // Sort by execution_start timestamp in descending order (newest first)
+      historyItems.sort((a, b) => {
+        const getExecutionStartTimestamp = (item: HistoryTaskItem) => {
+          if (!item.status?.messages) return 0
```
webfiltered, line 723: Do we expect any items to be 0 here, or below?  If there are, will they all live at one end of the list, in some non-deterministic order?

If that's not an issue, this looks alright as far as workarounds go, but it does seem like a bandaid to a problem we should address at the source.

PR #4428, opened 2025-07-11
refactor: address some `noUncheckedIndexedAccess` errors

duckcomfy:
Tentative refactor addressing ~10% of errors arising from enabling the TypeScript compiler option [`noUncheckedIndexedAccess`](https://www.typescriptlang.org/tsconfig/noUncheckedIndexedAccess.html), with the tentative goal of enabling it permanently, after gradually fixing the remaining errors. This PR makes the error count go from 567 to 506. The idea is to continue in small chunks. We can adjust the size and frequency.

Enabling `noUncheckedIndexAccess` also allows for safely enabling the ESLint rule [`no-unnecessary-condition`](https://typescript-eslint.io/rules/no-unnecessary-condition/) which will eliminate undefined/boolean checks for values that are statically known to never be nullish/falsy.

Idea originally proposed in #4366

This is an experiment, if you find that my methodology is too high-risk or requires too much supervision, feel free to not continue.

### General methodology

- Changes are purely to fix `noUncheckedIndexedAccess`-related compiler errors.
- Sometimes, the code can be made to typecheck with no logic changes, only narrower type annotations
   - e.g. from `const names: string[] = ["John", "Mary"]; const name = names[0]`
   - to `const names: [string, ...string[]] = ["John", "Mary"]; const name = names[0]`
- Sometimes, the code logic is safe but the guards need to be swapped to something equivalent that enables type narrowing
   - e.g. from `arr.length && foo(arr[0])`
   - to `arr[0] && foo(arr[0])`
- Sometimes, values which `noUncheckedIndexedAccess` reveals to be undefinable must be coalesced to `null` if they're used in places that were typed to only accept `null`
- Sometimes, the code is trivially safe but the type system needs a little unsafe casting
   - e.g. `arr.length === 1 && arr[0]!`
   - or destructuring litegraph `Point`, `Size`, and `Rect`, which have fixed length but are a type system blind spot due to the fact they can be `TypedArray`s
- Sometimes adding a new guard with a default value or early return is necessary, in which case I make sure that the previous behaviour of throwing an exception would not be preferable due to other pieces of code relying on it
- If a default value, additional guard or early return doesn't make sense, and a large refactor is advisable to explicitly handle the undefined case, I will leave the error untouched in order to keep the commit as simple as possible. At the very end of the process, these errors can be addressed, or a `// @ts-expect-error fixme noUncheckedIndexedAccess error` comment can be added before permanently enabling the compiler option

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4428-refactor-address-some-noUncheckedIndexedAccess-errors-22d6d73d3650811f9822c6e6b8ce5e03) by [Unito](https://www.unito.io)

### DIFF
--- a/src/utils/treeUtil.ts
+++ b/src/utils/treeUtil.ts
@@ -15,7 +15,7 @@ export function buildTree<T>(items: T[], key: (item: T) => string[]): TreeNode {
     const keys = key(item)
     let parent = root
     for (let i = 0; i < keys.length; i++) {
-      const k = keys[i]
+      const k = keys[i]!
       // 'a/b/c/' represents an empty folder 'c' in folder 'b' in folder 'a'
       // 'a/b/c/' is split into ['a', 'b', 'c', '']
       if (k === '' && i === keys.length - 1) break
--- a/src/components/common/ElectronFileDownload.vue
+++ b/src/components/common/ElectronFileDownload.vue
@@ -87,7 +87,7 @@ import { formatSize } from '@/utils/formatUtil'
 const props = defineProps<{
   url: string
   hint?: string
-  label?: string
+  label: `${string}/${string}`
   error?: string
 }>()
 
@@ -101,8 +101,7 @@ const fileSize = computed(() =>
   download.fileSize.value ? formatSize(download.fileSize.value) : '?'
 )
 const electronDownloadStore = useElectronDownloadStore()
-// @ts-expect-error fixme ts strict error
-const [savePath, filename] = props.label.split('/')
+const [savePath, filename] = props.label.split('/') as [string, string]
 
 electronDownloadStore.$subscribe((_, { downloads }) => {
   const download = downloads.find((download) => props.url === download.url)
--- a/src/components/common/InputKnob.vue
+++ b/src/components/common/InputKnob.vue
@@ -77,9 +77,7 @@ const updateValue = (newValue: number | null) => {
 const displayValue = (value: number): string => {
   updateValue(value)
   const stepString = (props.step ?? 1).toString()
-  const resolution = stepString.includes('.')
-    ? stepString.split('.')[1].length
-    : 0
+  const resolution = stepString.split('.')[1]?.length ?? 0
   return value.toFixed(props.resolution ?? resolution)
 }
 
--- a/src/components/common/SystemStatsPanel.vue
+++ b/src/components/common/SystemStatsPanel.vue
@@ -30,7 +30,10 @@
           <DeviceInfo :device="device" />
         </TabPanel>
       </TabView>
-      <DeviceInfo v-else :device="props.stats.devices[0]" />
+      <DeviceInfo
+        v-else-if="props.stats.devices[0]"
+        :device="props.stats.devices[0]"
+      />
     </div>
   </div>
 </template>
--- a/src/components/dialog/content/MissingModelsWarning.vue
+++ b/src/components/dialog/content/MissingModelsWarning.vue
@@ -92,7 +92,7 @@ const missingModels = computed(() => {
       downloading: false,
       completed: false,
       progress: 0,
-      error: null,
+      error: undefined,
       name: model.name,
       directory: model.directory,
       url: model.url,
--- a/src/components/dialog/content/manager/ManagerDialogContent.vue
+++ b/src/components/dialog/content/manager/ManagerDialogContent.vue
@@ -149,7 +149,7 @@ const {
   toggle: toggleSideNav
 } = useResponsiveCollapse()
 
-const tabs = ref<TabItem[]>([
+const tabs = ref<[TabItem, ...TabItem[]]>([
   { id: ManagerTab.All, label: t('g.all'), icon: 'pi-list' },
   { id: ManagerTab.Installed, label: t('g.installed'), icon: 'pi-box' },
   {
@@ -368,7 +368,7 @@ const resultsWithKeys = computed(
 
 const selectedNodePacks = ref<components['schemas']['Node'][]>([])
 const selectedNodePack = computed<components['schemas']['Node'] | null>(() =>
-  selectedNodePacks.value.length === 1 ? selectedNodePacks.value[0] : null
+  selectedNodePacks.value.length === 1 ? selectedNodePacks.value[0]! : null
 )
 
 const getLoadingCount = () => {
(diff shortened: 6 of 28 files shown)


src/utils/treeUtil.ts
```diff
@@ -15,7 +15,7 @@ export function buildTree<T>(items: T[], key: (item: T) => string[]): TreeNode {
     const keys = key(item)
     let parent = root
     for (let i = 0; i < keys.length; i++) {
-      const k = keys[i]
+      const k = keys[i]!
```
webfiltered, line 18: Non-null assertions are only acceptable as documented, temporary workarounds to get a feature in.  For this PR (being type fixes), they block it completely.

Using `// @ts-expect-error` highlights the issue, is full-text searchable.

webfiltered (changes requested):
Unfortunately this PR is taking "the LLM approach" - fixing individual compiler errors without understanding what's actually going on.

Where currently the app with throw and include a stack trace, this PR silently discards error states and would make debugging a nightmare.

webfiltered:
I'm just going to close this one - it's not a direction we can go in.

duckcomfy:
@webfiltered I have no problem with ending this experiment, but I want a chance to clear up the idea that I've done any of these changes thoughtlessly, and that therefore perhaps I wouldn't be able to understand and execute the actual desired conventions and direction. I'm not expecting a debate, your time is obviously much better spent than arguing over this, so if you don't want to read this, or  if this doesn't change your mind about the changes being made "not understanding what's actually going on", I'm fine with ending the conversation here. But here is my commentary on each change:

`src/components/common/ElectronFileDownload.vue`
`src/components/dialog/content/manager/ManagerDialogContent.vue L152`
`src/components/load3d/Load3DControls.vue`
Turned any possible exception into compiler errors. No error state masked.

`src/components/common/InputKnob.vue`
`src/components/dialog/content/MissingModelsWarning.vue`
`src/components/dialog/content/manager/infoPanel/InfoPanelHeader.vue`
`src/components/dialog/content/manager/infoPanel/tabs/DescriptionTabPanel.vue`
`src/components/dialog/content/setting/KeybindingPanel.vue`
`src/components/graph/selectionToolbox/ColorPickerButton.vue`
`src/components/searchbox/NodeSearchBox.vue`
`src/composables/setting/useSettingSearch.ts L104`
Equivalent code, refactored in a way that typechecks. No error state masked.

`src/components/dialog/content/manager/ManagerDialogContent.vue L371`
`src/components/dialog/content/manager/infoPanel/MarkdownText.vue`
`src/components/graph/TitleEditor.vue`
`src/components/load3d/controls/RecordingControls.vue`
`src/composables/useCoreCommands.ts`
`src/utils/treeUtil.ts`
Trivially safe casting/assertions (meaning I've verified that the current related code/value definitions shows the casting will never throw).

I'm aware code can change and the compiler won't say if previously trivially safe casting no longer is; I don't use any unsafe casting in my own projects, and extract the unsafety into unit-tested helpers (or use libs like ts-pattern) to fill in the type system blind spots.

However, it seems widespread conventional practice in the TypeScript ecosystem is to do some unsafe casting like this. This is why, for the tentative first proposal, I decided to go with this convention, making the incorrect assumption that because the unsafe behavior was deemed acceptable then, it is also considered acceptable going forward.

If I had known better, I'd've been happy to make small safe refactors where possible and leave the more dicey ones with `// @ts-expect-error` in the final PR.

`src/components/graph/GraphCanvas.vue`
`src/components/graph/NodeTooltip.vue`
`src/composables/node/useNodeAnimatedImage.ts`
`src/composables/node/useNodeCanvasImagePreview.ts`
`src/composables/node/useNodeChatHistory.ts`
`src/composables/node/useNodeProgressText.ts`
`src/composables/setting/useSettingSearch.ts L107`
`src/stores/workflowStore.ts`

"Trivially unnecessary" guards (meaning I've verified that the related code means the guards never come up). Same comments as the above, with the caveat that in these cases, if the code becomes unsafe, yes, error states would become hidden. I could make all similar guards throw instead. Looking back, you're right, throwing is definitely superior to doing nothing, which added debugging burden. I'm too used to making all non-IO errors compiler errors. But I didn't make any of these changes without understanding them.

`src/components/searchbox/NodeSearchFilter.vue`

I decided to go with that because I saw that the way this value is used (fuseutil.ts L51) would have behaved the same (no exception would have been thrown in case of incorrect state), and of course the PR can't add new exceptions in case existing code relies on the incorrect state being handled silently. I suppose, optimally, this should have been left for the end with a `// @ts-expect-error` for future refactoring.

`src/stores/settingStore.ts`

This made sense to me for the reasons we spoke about in #4366 - if extension settings are somehow causing settingsById[id] to be undefined, their category must be "Other", a default case that was already provided by the existing code in case of missing category or badly-formed setting id (name).

`src/components/common/SystemStatsPanel.vue`

Point taken. The absence of devices is an abnormal state. But I didn't add the guard thoughtlessly. I actually went ahead and looked at what the parser said, incorrectly trusted that it was correct, and didn't fix this error in isolation. In this case the `devices` parser should be `z.array(zDeviceStats).nonEmpty()` instead of `z.array(zDeviceStats)`. This way, the exception is thrown in the most debugging-friendly location, and it is clear to everyone that lack of device is an abnormal state.

src/utils/treeUtil.ts
duckcomfy: I agree, as the PR said, I would ignore cases that require a `// @ts-expect-error` until the very last PR. The reason why (I didn't explicitly say), is because you cannot use `// @ts-expect-error` until you enable the relevant rule.

webfiltered:
Apologies, intention was not to infer negativity; was just writing as I went and not really thinking about broader context.  Whilst we could spend time going through this and fixing it, it's just not a priority atm.  The PR would need endless rebases before we're remotely ready to get into it, so I closed it instead.

For a slightly expanded explanation, I am strongly against using type assertions.  When I add them, each gets an explanation, and I only use them as a last resort (or a temporary measure).

The reason is quite simple: the assertion itself is probably fine, when it is written.  People often do their homework and confirm that it's doing what it's supposed to, and it is just needed to avoid unnecessary runtime changes / some kind of TS limitation.

The issue is when someone changes something else that _should_ be caught by `tsc`, but is not because of the assertion.  It can lead to hours of debugging, and is usually just not worth the pain.

duckcomfy:
All good, I understand that too this requires too much time to dedicate to.

duckcomfy:
I'm also strongly against unsafe assertions and casting. But I also try to adapt to the conventions I see in surrounding code.

PR #4439, opened 2025-07-13
Add support for Feature Flags

guill:
This change implements feature flag support on the frontend. When the frontend
establishes a websocket connection to the backend, it will send a feature flag message
informing the backend of available functionality. This allows the backend to be smarter
about sending messages that the frontend will understand.
We also receive a feature flag message from the backend and provide functions to query
available functionality on the backend. This allows custom nodes (and built-in code)
to gate functionality depending on whether the backend actually supports it.

This change was factored out of https://github.com/Comfy-Org/ComfyUI_frontend/pull/4382

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4439-Add-support-for-Feature-Flags-22f6d73d365081e08d73cc9f31172dc7) by [Unito](https://www.unito.io)

A video showing that ComfyUI still works (no visible changes result from this PR.)
https://github.com/user-attachments/assets/d79b01f0-df5f-426c-b69a-b8a56e4a5453


src/scripts/api.ts
```diff
@@ -234,6 +237,30 @@ export class ComfyApi extends EventTarget {
...
+  /**
+   * Alias for serverFeatureFlags for test compatibility.
+   */
+  get feature_flags() {
+    return this.serverFeatureFlags
+  }
+
+  set feature_flags(value: Record<string, any>) {
+    this.serverFeatureFlags = value
+  }
```
webfiltered: What does this do?  If it's removed and the refs in the unit test replaced with `serverFeatureFlags`, everything still passes.

webfiltered (dismissed):
Can you please note the need for the new interfaces to be `any` typed and merge?

If there isn't one (anymore/yet?), can we make them `unknown` instead?  It makes the eventual enabling of "no implicit/explicit any type" rules easier.

src/scripts/api.ts
guill: I've removed it. Claude suggested it and I assumed it was related to how TypeScript stubs out functionality in unit tests. Not really sure though now 🤔

guill:
> Can you please note the need for the new interfaces to be `any` typed and merge?
> 
> If there isn't one (anymore/yet?), can we make them `unknown` instead? It makes the eventual enabling of "no implicit/explicit any type" rules easier.

Had no idea `unknown` was a thing! Fixed 👍

guill:
https://github.com/user-attachments/assets/d0370c88-ab63-4ac7-bd95-52b2edef2dc0

A video showing that ComfyUI still works.

webfiltered (approved):
The blurriness took me a second, but I managed to read your test prompt.  Megapproved.  🤣

PR #4441, opened 2025-07-13
[test] Add docker-compose to faciliate browser test execution

luke-mino-altherr:
# Summary

Improves Playwright browser test infrastructure by adding Docker support and fixing cross-platform compatibility issues.

## Problem

The existing browser test setup was complex and had several issues:

- **Platform-specific screenshot issues**: Tests generated Linux-specific screenshots that failed on macOS due to font rendering differences
- **Complex local setup**: Required manually installing and configuring ComfyUI backend, managing dependencies, and setting up test environment  
- **User context errors**: Backend threw "Unknown user: default" errors in multi-user mode when tests made API calls without proper user context
- **Inconsistent test environment**: Different developers had varying local setups leading to flaky tests

### Docker Infrastructure

- **`Dockerfile`**: Multi-stage build for ComfyUI with built frontend, includes system dependencies for PyTorch
- **`docker-compose.test.yml`**: Orchestrates ComfyUI backend and Playwright test containers with proper networking and health checks
- **`Dockerfile.playwright`**: Dedicated container for running Playwright tests with all browsers pre-installed
- **`npm run test:browser:docker`**: Single command to run entire test suite in containers

### Cross-Platform Compatibility  

- **Screenshot tolerance**: Added 3% pixel difference tolerance for local testing to handle font rendering variations
- **Gitignore patterns**: Ignore platform-specific screenshot files (e.g., `*-darwin.png`) to prevent commit conflicts

### User Context Fixes

- **Default user creation**: Global setup now creates a "default" user in `users.json` to fix errors logged in the backend API logs.

## Benefits

- **One-command testing**: `npm run test:browser:docker` runs complete test suite
- **Platform consistency**: Tests run in identical Linux containers regardless of host OS  
- **Easier onboarding**: New developers don't need complex local ComfyUI setup
- **Reliable CI/CD**: Consistent containerized environment eliminates "works on my machine" issues
- **Faster debugging**: HTML reports and screenshots automatically generated and accessible via browser

## Usage

```bash
# Run all browser tests in Docker
npm run test:browser:docker

# View test results at http://localhost:9323 (if report server starts)
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4441-test-Add-docker-compose-to-faciliate-browser-test-execution-22f6d73d36508110bf53fb9c04ade703) by [Unito](https://www.unito.io)

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -28,6 +28,13 @@ export default defineConfig({
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: 'on-first-retry'
   },
+  /* Global expect configuration for screenshot tolerance */
+  expect: {
+    toHaveScreenshot: {
+      /* Allow 3% difference in pixel values during local testing */
+      maxDiffPixelRatio: process.env.CI ? undefined : 0.03
+    }
+  },
   /* Path to global setup file. Exported function runs once before all the tests */
   globalSetup: './browser_tests/globalSetup.ts',
   /* Path to global teardown file. Exported function runs once after all the tests */
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,56 @@
+# Dependencies
+node_modules
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# Build outputs
+dist
+.next
+.nuxt
+
+# Development files
+.env
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
+
+# IDE files
+.vscode
+.idea
+*.swp
+*.swo
+*~
+
+# OS files
+.DS_Store
+Thumbs.db
+
+# Git
+.git
+.gitignore
+
+# Test outputs
+test-results
+playwright-report
+coverage
+
+# Documentation
+README.md
+*.md
+docs
+
+# Docker files (don't copy into themselves)
+Dockerfile*
+docker-compose*.yml
+.dockerignore
+
+# CI/CD
+.github
+.gitlab-ci.yml
+
+# Temporary files
+tmp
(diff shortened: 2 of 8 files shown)


playwright.config.ts
christian-byrne, line 1: If the primary issue with local testing is screenshots due to things like FreeType vs. DirectWrite, isn't this change the only thing that's necessary?

webfiltered: I actually think this is not something we can merge. 😅

At only 1024x768, this ignores anything that changes less than 23,000 pixels.  It would cripple the usefulness of playwright.

webfiltered (changes requested):

playwright.config.ts
luke-mino-altherr: I am running on a mac. Since the only screenshots committed to the repository are linux based screenshots, i need to execute these tests in a linux based execution environment in order to properly validate there are no visual regressions. 

Also as a new developer, i found the initial setup complex -- Docker cleanly abstracts that initial setup and server startup.

luke-mino-altherr: > I actually think this is not something we can merge. 😅
> 
> At only 1024x768, this ignores anything that changes less than 23,000 pixels.  It would cripple the usefulness of playwright.

Ah fair point. We could lower the tolerance? I was most often seeing differences at less than 1%. Also note, this is only in effect during local testing. Github Actions will still check for exact matches

webfiltered: I'm super curious about the discrepancies, and wonder if it's just a difference between the image GH actions uses and the docker image?  As the display should be virtualised too, we may be able to get pixel-perfect.

I can take a look on mac/windows and see if anything stands out (likely within a day or two).

PR #4443, opened 2025-07-14
Manager Conflict Nofitication

viva-jinyi:
### **Summary**

  This PR implements a comprehensive compatibility warning system for the ComfyUI-Manager that detects and notifies users about potential
  compatibility conflicts with custom node packages. The system provides multiple levels of user notification while maintaining a non-intrusive user
  experience.

### **Conflict Detection and Notification Design**

  **1. On App Mount: Only Detection (No UI)**

  When ComfyUI launches, all installed custom nodes are scanned in the background for conflicts.
  No dialog is shown at this point.

  **2. When Is the Modal Dialog Shown? (Only One Case)**

  After a ComfyUI update, the "What's New" modal appears first.
  Once dismissed, a conflict modal will appear once if conflicts are found.
  If the user closes this conflict modal, it's marked as acknowledged in localStorage and won't be shown again for the same issue.
  Other than this case, the modal never appears automatically.
  

  **3. Warning Badges (Node Manager UI)**

  Node packs with known conflicts show a warning badge in the Node Manager UI.
  If the user proceeds to install or enable such a node pack despite the warning,
  we treat it as user acknowledged the conflict, and no modal will be shown.

  **4. Red Dot Notification (Passive Alert on App Mount)**

  If any conflicts are detected during app mount, a red dot appears on the Help Center.
  After the user opens the Help Center and clicks through to the Manager Extension, and views the yellow conflict banner at the top of the screen, →
  the red dot is cleared.
  This is saved in localStorage so the red dot and modal won't show again for the same conflict.

### **Key Features Implemented**

  **Compatibility Detection System**

  - Multi-dimensional conflict checking: OS, accelerators, ComfyUI version, frontend version, Python version, banned packages, and security status
  - Centralized detection logic: useConflictDetection composable with consistent checking across the application
  - Background scanning: Automatic detection on app launch without blocking UI

  **Manager Dialog Enhancements**

  - Warning Tab: Auto-appears in InfoPanel when compatibility issues are detected
  - Conflict Banner: Yellow warning banner at top of manager with link to documentation
  - Version Selector Improvements:
    - Display format: "Latest (1.0.3)" for clarity
    - Compatibility warnings per version with tooltips
    - No compatibility indicators for nightly versions (uncertain compatibility)
    - Correct version resolution for installation (sends "1.0.3" instead of "latest")

  **Visual Indicators**

  - Warning badges on package cards with compatibility conflicts
  - Red dot notification on Help Center when conflicts detected
  - Consistent iconography across all warning states

### **Technical Implementation**

  **Core Components Added/Modified**

  - useConflictDetection.ts - Centralized compatibility checking logic
  - WarningTabPanel.vue - New tab for displaying conflict details
  - InfoTabs.vue - Auto-switching to warning tab when conflicts detected
  - PackVersionSelectorPopover.vue - Enhanced version selection with compatibility warnings
  - ManagerDialogContent.vue - Version-specific API data fetching and conflict banner

  **Conflict Types Supported**

  - Operating System compatibility (Windows, macOS, Linux)
  - Accelerator requirements (CUDA, CPU, Metal, etc.)
  - ComfyUI version constraints
  - Frontend version constraints
  - Python version compatibility
  - Banned package detection
  - Security status checking

  **API Enhancements**

  - Version-specific data fetching: For installed nodes, fetch detailed version information
  - Data merge priority: Ensure API compatibility data takes precedence over search data
  - Consistent data sources: "Latest" and actual latest version use identical compatibility data

  **Test Coverage**

  - 25 conflict detection tests passing with comprehensive edge case coverage
  - Component tests updated for new warning UI elements
  - Integration tests for user workflow scenarios
  - vitest.setup.ts initialized with global variables to simulate production-like conditions during tests
  
 
스크린샷 2025-07-19 오후 11 13 09
스크린샷 2025-07-19 오후 11 13 52
스크린샷 2025-07-19 오후 11 13 42
스크린샷 2025-07-19 오후 11 13 17

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4443-Manager-Conflict-Nofitication-2306d73d3650815aab0de900a9b21c65) by [Unito](https://www.unito.io)


src/components/dialog/content/manager/ManagerDialogContent.vue
```diff
@@ -464,7 +549,33 @@ whenever(selectedNodePack, async () => {
...
+const handleDialogClose = () => {
+  if (hasConflicts.value) {
+    localStorage.setItem('comfy_help_center_conflict_seen', 'true')
+    localStorage.setItem('comfy_manager_conflict_banner_dismissed', 'true')
+
+    // Also try to update refs
+    hasSeenConflicts.value = true
+    isConflictBannerDismissed.value = true
+  }
+}
```
christian-byrne: Would it work to simplify this and just use `onBeforeUnmount`?

christian-byrne:
https://vueuse.org/core/useStorage/ might be helpful to use.

src/components/sidebar/SidebarHelpCenterIcon.vue
```diff
@@ -57,22 +58,57 @@
...
+const forceCheckStorage = () => {
+  const currentValue =
+    localStorage.getItem('comfy_help_center_conflict_seen') === 'true'
+  if (hasSeenConflicts.value !== currentValue) {
+    hasSeenConflicts.value = currentValue
+  }
+}
+
+// Check every 500ms for localStorage changes
+setInterval(forceCheckStorage, 500)
```
christian-byrne: I think `useStorage` will observe storage changes even from other parts of the code (or other tabs), even without using the ref. I could be wrong.

christian-byrne: If not, then what do you think about using https://vueuse.org/shared/useInterval/? Otherwise we will have to clear the interval in the unmount hook.

src/components/dialog/content/manager/ManagerDialogContent.vue
```diff
@@ -150,6 +177,19 @@ const {
...
+  CONFLICT_BANNER_DISMISSED_KEY,
+  false
+)
+
+// Help center conflict seen state
+const hasSeenConflicts = useStorage('comfy_help_center_conflict_seen', false)
+
+// Computed properties for conflicts
+const hasConflicts = computed(() => conflictDetectionStore.hasConflicts)
+
```
christian-byrne: Do you think we should make a new component hook/composable for this?

src/components/dialog/content/manager/ManagerDialogContent.vue
```diff
@@ -441,11 +488,49 @@ whenever(selectedNodePack, async () => {
...
+    }
+  }
+
+  // For uninstalled nodes or if version-specific data fetch failed, use default API
+  if (!data) {
+    data = await getPackById.call(pack.id)
+  }
+
   // If selected node hasn't changed since request, merge registry & Algolia data
-  if (data?.id === pack.id) {
```
christian-byrne, line 446: I think this check was important to handle condition where the pack that's selected in the UI has changed since the request started (timing issue / race condition)

src/components/dialog/content/manager/ManagerDialogContent.vue
```diff
@@ -480,9 +591,31 @@ onBeforeUnmount(() => {
...
+  // Cleanup
+  window.removeEventListener('beforeunload', handleDialogClose)
+  const interval = (window as any).__managerDialogInterval
+  if (interval) {
+    clearInterval(interval)
+    delete (window as any).__managerDialogInterval
+  }
+
+  // Final attempt to set localStorage
+  handleDialogClose()
```
christian-byrne: I believe we can remove the event listener, the `__managerDialogInterval` system, and the interval. The `onUnmounted` will handle when the component is unmounted, and we probably don't need to handle page navigation. WDYT?

src/components/dialog/content/manager/PackVersionSelectorPopover.vue
```diff
@@ -173,16 +210,120 @@ const handleSubmit = async () => {
...
+        latestVersionData?.supported_comfyui_frontend_version ??
+        nodePack.supported_comfyui_frontend_version,
+      supported_python_version:
+        (latestVersionData as any)?.supported_python_version ??
+        (nodePack as any).supported_python_version,
+      is_banned:
+        (latestVersionData as any)?.is_banned ?? (nodePack as any).is_banned,
+      has_registry_data:
+        (latestVersionData as any)?.has_registry_data ??
+        (nodePack as any).has_registry_data
```
christian-byrne: After we updated the registry API types and rebased, is the `as any` type assertion still necessary?

src/components/dialog/content/manager/PackVersionSelectorPopover.vue
```diff
@@ -173,16 +210,120 @@ const handleSubmit = async () => {
...
+  // Fallback to nodePack data
+  return {
+    supported_os: nodePack.supported_os,
+    supported_accelerators: nodePack.supported_accelerators,
+    supported_comfyui_version: nodePack.supported_comfyui_version,
+    supported_comfyui_frontend_version:
+      nodePack.supported_comfyui_frontend_version,
+    supported_python_version: (nodePack as any).supported_python_version,
+    is_banned: (nodePack as any).is_banned,
+    has_registry_data: (nodePack as any).has_registry_data
```
christian-byrne: Same question as above

src/components/dialog/content/manager/ManagerDialogContent.vue
```diff
@@ -441,11 +488,49 @@ whenever(selectedNodePack, async () => {
...
   if (lastFetchedPackId.value === pack.id) return
-  const data = await getPackById.call(pack.id)
+
+  let data = null
+
+  // For installed nodes only, fetch version-specific information
+  if (comfyManagerStore.isPackInstalled(pack.id)) {
+    const installedPack = Object.values(comfyManagerStore.installedPacks).find(
+      (installed) => (installed.cnr_id || installed.aux_id) === pack.id
+    )
```
christian-byrne: Should we make a `installedPacksById` in the store?

src/composables/useConflictDetection.ts
```diff
@@ -571,18 +912,116 @@ export function useConflictDetection() {
...
+        // Check if this conflict type already exists
+        return (
+          array.findIndex(
+            (c) =>
+              c.type === conflict.type &&
+              c.current_value === conflict.current_value &&
+              c.required_value === conflict.required_value
+          ) === index
+        )
+      })
```
christian-byrne: Have you used `uniqBy` before?

```typescript
const uniqueConflicts = _.uniqBy(
  [...existing.conflicts, ...conflict.conflicts],
  conflict => `${conflict.type}-${conflict.current_value}-${conflict.required_value}`
)
```

It's nice to use and goes from O(n^2) to O(n).

src/components/sidebar/SidebarHelpCenterIcon.vue
viva-jinyi: You’re right — I replaced the manual interval with useStorage directly, and it’s working as expected across tabs.

src/components/dialog/content/manager/ManagerDialogContent.vue
viva-jinyi: Already implemented! Added race condition checks at lines 491-494 and 503-506 to handle exactly this scenario. The checks ensure we don't apply stale data if the selected pack changes during async operations.

src/components/dialog/content/manager/ManagerDialogContent.vue
viva-jinyi: Created `useConflictBannerState` composable to centralize this logic and eliminate duplication across the 3 components that were
  managing conflict state separately.

src/components/dialog/content/manager/ManagerDialogContent.vue
viva-jinyi: Replaced the complex event listener + interval system with simple onBeforeUnmount().

src/components/dialog/content/manager/ManagerDialogContent.vue
viva-jinyi: Removed all the event listener + __managerDialogInterval complexity and just used onBeforeUnmount/onUnmounted.

src/components/dialog/content/manager/PackVersionSelectorPopover.vue
viva-jinyi: Removed both as any assertions and replaced with proper type guards using 'property' in object checks.

src/components/dialog/content/manager/PackVersionSelectorPopover.vue
viva-jinyi: Removed both as any assertions and replaced with proper type guards using 'property' in object checks.

src/components/dialog/content/manager/ManagerDialogContent.vue
viva-jinyi: Added getInstalledPackByCnrId getter to the store and updated ManagerDialogContent to use it instead of the Object.values().find()
  pattern.

src/composables/useConflictDetection.ts
viva-jinyi: Used uniqBy exactly as you suggested to optimize from O(n²) to O(n).

src/components/dialog/content/manager/ManagerDialogContent.vue
christian-byrne: Well done!

src/components/dialog/content/manager/NodeConflictDialogContent.vue
```diff
@@ -0,0 +1,175 @@
...
+
+const allConflictDetails = computed(() => {
+  const details: ConflictDetail[] = []
+  conflictData.value.forEach((conflictResult) => {
+    conflictResult.conflicts.forEach((conflict) => {
+      details.push(conflict)
+    })
+  })
+  return details
+})
```
christian-byrne: nit: How about flatMap?

```
    const allConflictDetails = computed(() =>
      conflictData.value.flatMap(result => result.conflicts)
    )
```

Up to you! Since this will be recomputed on deep changes, it might be worth it to optimize.

src/composables/useConflictDetection.ts
```diff
@@ -388,50 +382,165 @@ export function useConflictDetection() {
...
+              }
+
+              console.warn(
+                `[ConflictDetection] Failed to check import failure for ${packageName}:`,
+                error
+              )
+            }
+            return null
+          })
+        )
```
christian-byrne, line 471: Do we plan on updating the Manager handler to support batches after this PR lands or before?

src/components/dialog/content/manager/NodeConflictDialogContent.vue
christian-byrne: Also, would this be beneficial to have in the store? Or is it only used here?

christian-byrne:
BTW, is the plan to remove `console.log` when the `manager/compatibility` is being merged?

src/composables/useConflictDetection.ts
```diff
@@ -388,50 +382,165 @@ export function useConflictDetection() {
...
+              }
+            } catch (error) {
+              // If API returns 400, it means no import failure info available
+              // This is normal for packages that imported successfully
+              if (error && typeof error === 'object' && 'response' in error) {
+                const axiosError = error as any
+                if (axiosError.response?.status === 400) {
+                  return null // No failure info available (normal case)
+                }
+              }
```
christian-byrne, line 462: I wonder if we should just adjust the Manager server to serve all failed imports by node pack ID, instead of us having to include IDs in the request at all?

src/components/dialog/content/manager/button/PackEnableToggle.vue
```diff
@@ -61,9 +82,41 @@ const handleDisable = () => {
...
+        return
+      }
+    }
+  }
+
+  // No conflicts or conflicts acknowledged - proceed with toggle
+  await performToggle(enable)
+}
+
+const performToggle = async (enable: boolean) => {
```
christian-byrne, line 119: It seems like `handleToggle` calls `onToggle` which calls `handleToggle` in the debounce function (I could be wrong).

src/components/dialog/content/manager/packCard/PackCardFooter.vue
```diff
@@ -1,13 +1,25 @@
...
+          :has-conflict="!!packageConflicts"
+        />
+      </template>
+      <template v-else>
+        <PackEnableToggle
+          :node-pack="nodePack"
+          :has-conflict="!!packageConflicts"
+        />
+      </template>
+    </div>
```
christian-byrne: For this component, do we only need a boolean to determine whether conflicts exist or not?

In that case, we may be able to optimize the computed we are using a bit (e.g., return early if we find a single conflict). We could also consider splitting the computed into multiple properties for memoization like

```typescript
    const exactPackageMatch = computed(() =>
      conflictStore.getConflictsForPackage(nodePack.id)
    )
    const packageConflicts = computed(() =>
      exactPackageMatch.value || fallbackPackageMatch.value
    )
```

Could this also be something provided by a composable / component hook? It might be worth it if the functionality is going to be used elsewhere.

src/components/dialog/content/manager/infoPanel/InfoTabs.vue
```diff
@@ -44,4 +60,14 @@ const nodeNames = computed(() => {
...
+
+// Watch for compatibility issues and automatically switch to warning tab
+watchEffect(() => {
+  if (hasCompatibilityIssues) {
+    activeTab.value = 'warning'
+  } else if (activeTab.value === 'warning') {
+    // If currently on warning tab but no issues, switch to description
+    activeTab.value = 'description'
+  }
+})
```
christian-byrne: Shuold this have `{ flush: 'post' }`?

src/composables/useConflictBannerState.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    const currentDismissedValue =
+      localStorage.getItem(MANAGER_CONFLICT_BANNER_DISMISSED_KEY) === 'true'
+    if (isConflictBannerDismissed.value !== currentDismissedValue) {
+      isConflictBannerDismissed.value = currentDismissedValue
+    }
+  }
+
+  // Set up periodic storage checking (500ms interval)
+  useIntervalFn(forceCheckStorage, 500)
```
christian-byrne: Are you sure it's necessary? I thought `useStorage` would be able to handle it by itself. I could be wrong!

src/components/dialog/content/manager/NodeConflictDialogContent.vue
```diff
@@ -0,0 +1,175 @@
...
+import type {
+  ConflictDetail,
+  ConflictDetectionResult
+} from '@/types/conflictDetectionTypes'
+import { getConflictMessage } from '@/utils/conflictMessageUtil'
+
+interface Props {
+  conflicts?: ConflictDetectionResult[]
+  conflictedPackages?: ConflictDetectionResult[]
+  onDismiss?: () => void
```
christian-byrne: Would it be difficult to use `emit` here?

src/components/dialog/content/manager/packCard/PackCardFooter.vue
christian-byrne: `hasConflict` might already be in the store.

src/components/dialog/content/manager/NodeConflictDialogContent.vue
viva-jinyi: It is only used here. So I think using flatMap is benetical but having in the store is not necessary for now.

src/composables/useConflictDetection.ts
viva-jinyi: Yes, I'm planning to update this `import_failed_info` API to bulk api

viva-jinyi:
@christian-byrne
> BTW, is the plan to remove console.log when the manager/compatibility is being merged?

Yes, I will remove all the `console.log` before **manager/compatibility** merged to base branch.

src/composables/useConflictDetection.ts
viva-jinyi: I tried to create the `import_failed_info` API in bulk, so I asked Data for help. With Data's help on Saturday, I think I can try this. It's not a big change, so it should be done quickly.

src/components/dialog/content/manager/button/PackEnableToggle.vue
viva-jinyi: This part can really lead to errors. I'll fix it.

src/components/dialog/content/manager/packCard/PackCardFooter.vue
viva-jinyi: The main issue with current conflict detection is that package IDs include version suffixes (e.g., `img2colors-comfyui-node@1_0_3`), but the
  UI searches without versions (`img2colors-comfyui-node`). This causes `getConflictsForPackage` to fail at exact matching.

  To resolve this issue, the following improvements are needed:

  1. **Enhance getConflictsForPackage in the store**: Add smart matching logic that handles version suffixes
  2. **Normalize package IDs**: Standardize how version information from API data is processed
  3. **Version-specific compatibility management**: Same package may have different conflict status across versions

  Currently using complex fallback logic as a temporary workaround, but once these improvements are implemented, simple boolean checks will be
  sufficient for accurate conflict detection.

  I'll create a dedicated branch to address this issue immediately after this PR is merged.

src/components/dialog/content/manager/infoPanel/InfoTabs.vue
viva-jinyi: Since this watchEffect changes activeTab.value which triggers UI updates, using { flush: 'post' } would ensure the effect runs after DOM updates are complete. This could prevent potential timing issues with tab transitions.

christian-byrne (approved):
LGTM!

src/composables/useConflictBannerState.ts
viva-jinyi: You're absolutely right that useStorage should handle localStorage synchronization automatically! However, we discovered a timing issue during testing.

  **The Problem:**
  When `markConflictsAsSeen()` is called, useStorage updates the reactive ref values immediately (`hasSeenConflicts.value = true`), but the localStorage
  synchronization happens asynchronously. This created a timing gap where:

  1. Dialog closes → `markConflictsAsSeen()` called → refs updated
  2. Dialog reopens immediately → `shouldShowManagerBanner` computed runs
  3. useStorage hasn't synced to localStorage yet → banner shows again

  **The Solution:**
  I've added direct `localStorage.setItem()` calls as a backup alongside the useStorage refs:

  ```javascript
  hasSeenConflicts.value = true           // useStorage ref (async to localStorage)
  isConflictBannerDismissed.value = true  // useStorage ref (async to localStorage)

  // Immediate localStorage backup for timing-critical scenarios
  localStorage.setItem(HELP_CENTER_CONFLICT_SEEN_KEY, 'true')
  localStorage.setItem(MANAGER_CONFLICT_BANNER_DISMISSED_KEY, 'true')
```


This ensures the state is persisted immediately while still leveraging useStorage for reactivity and cross-tab synchronization. The redundancy is minimal but eliminates the race condition.

PR #4461, opened 2025-07-15
[feat] sync node library when subgraph titles change

christian-byrne:
Ensures subgraph title changes are immediately reflected in the node library sidebar. When users rename a subgraph node, the node library shows the updated display name without requiring a reload.

Implementation preserves object references while updating both nodeDefsByName and nodeDefsByDisplayName mappings, plus LiteGraph registered node type titles for new instance creation.

https://github.com/user-attachments/assets/fee3e476-0cbb-4226-86ed-f12db5337b30

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4461-feat-sync-node-library-when-subgraph-titles-change-2316d73d36508134be5bdf40586ffb95) by [Unito](https://www.unito.io)


browser_tests/tests/sidebar/nodeLibrary.spec.ts
```diff
@@ -336,4 +336,54 @@ test.describe('Node library sidebar', () => {
     await comfyPage.page.waitForTimeout(1000)
     expect(await tab.getNode('KSampler (Advanced)').count()).toBe(2)
   })
+
+  test('Subgraph node display name updates in library when renamed', async ({
```
christian-byrne, line 340: We will skip this test later and then re-enable when subgraphs are re-added to sidebar.

christian-byrne:
This is outdated and the beahvior has been replaced by

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/5139

PR #4472, opened 2025-07-18
Add the copy URL button for the missing models dialog to Desktop

comfyui-wiki:
Terry only add the Copy url to Portable, so I bring this feature to Desktop in order to solved this [issue](https://github.com/comfyanonymous/ComfyUI/issues/8958)

image

https://github.com/user-attachments/assets/53d4ae33-4229-41c0-8379-0a864b68d37b

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4472-Add-the-copy-url-button-for-Desktop-2346d73d3650816889e2e227f9d797b0) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/common/ElectronFileDownload.vue
+++ b/src/components/common/ElectronFileDownload.vue
@@ -23,6 +23,14 @@
           icon="pi pi-download"
           @click="triggerDownload"
         />
+        <Button
+          v-if="status === null || status === 'error'"
+          :label="$t('g.copyURL')"
+          size="small"
+          outlined
+          :disabled="!!props.error"
+          @click="copyURL"
+        />
       </div>
     </div>
     <div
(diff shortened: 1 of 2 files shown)


src/components/common/ElectronFileDownload.vue
```diff
@@ -23,6 +23,14 @@
...
           @click="triggerDownload"
         />
+        <Button
+          v-if="status === null || status === 'error'"
+          :label="$t('g.copyURL')"
+          size="small"
+          outlined
+          :disabled="!!props.error"
+          @click="copyURL"
+        />
```
christian-byrne, line 32: I don't think we need to disable the button if `props.error`. The error occurs when the link is not from an allowed domain. The user can decide what to do -- they will already get a warning message that the link is potentially unsafe.

src/components/common/ElectronFileDownload.vue
```diff
@@ -23,6 +23,14 @@
           icon="pi pi-download"
           @click="triggerDownload"
         />
+        <Button
+          v-if="status === null || status === 'error'"
```
christian-byrne: We should add condition `!!props.url`

comfyui-wiki:
Updated now.

christian-byrne:
Should we also add margin/gap between the buttons?

christian-byrne (approved):

PR #4481, opened 2025-07-20
Fix i18n workaround breaks node configure & subgraphs

webfiltered:
### Current

A stack of workarounds over time for i18n (and other fixes) resulted in a monkey patch in frontend that was:
- Reading the *current* inputs of an existing node on the graph
- (Good) Applying i18n to the *new* node configure data, if the input existed
- (Bad) If not, adding inputs to the configure data as-is (the original, unserialised input objects)
  - Node inputs *can not* be removed by using `configure`

### Proposed

- Copy *only* the i18n values from the node into the configure data
- Do not replace the inputs / outputs properties of the configure data
- Do not pass existing input / output objects to the parent `configure` method
- Removes need for extensions to workaround this peculiar behaviour
- Drastically simplifies code

### Issues

Any extensions that worked around the former behaviour may now be reliant on it.  This should be verified before merging.

- @rgthree reported an issue caused by this (iirc, it was a property being retained because the slot itself was passed - the workaround was removing the prop before configure ran, and adding it back afterwards...)
- Root cause of: https://github.com/Comfy-Org/litegraph.js/issues/1130


jtydhr88:
as it has conflict and no update since last August, I would like to close for now, please rebase and resend if we still need it

PR #4499, opened 2025-07-22
Update component widget props type to exclude reserved props 

christian-byrne:
## Summary
- Removed the `componentProps` parameter from ComponentWidgetImpl
- Added ComponentWidgetStandardProps type to clearly define which props are reserved by the framework
- Updated widget implementations to use local type extraction pattern for better type safety

This approach provides better type safety while maintaining the ability to pass custom props to Vue component widgets.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4499-Update-component-widget-props-type-to-exclude-reserved-props-2386d73d36508100bf2fd7854142fc15) by [Unito](https://www.unito.io)


guill (approved):
👍

PR #4500, opened 2025-07-23
[feat] Update node output system to use NodeLocatorIds

christian-byrne:
Updates the global node outputs store to use node locator IDs instead of execution IDs using foundation established in https://github.com/Comfy-Org/ComfyUI_frontend/pull/4382.

The `executed` WS message sends messages mapping hierarchical node IDs (based on execution context) to outputs (file references). This PR changes to using locator IDs (immediate parent graph ID : node ID).

During rendering logic, we won't know the execution context of node instances, but we know the node instance's parents graph. So, we can query the store using the active parent graph ID prefixed with the node instance ID.

https://github.com/user-attachments/assets/6225b9dd-ffbb-4b03-99e3-b0dcb10ac65c

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4500-feat-Update-node-output-system-to-use-NodeLocatorIds-2396d73d365081af8ceff78e31497af2) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/imagePreviewStore.ts
+++ b/src/stores/imagePreviewStore.ts
@@ -133,19 +133,17 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
   ) {
     if (!filenames || !node) return
 
-    const nodeLocatorId = nodeIdToNodeLocatorId(node.id)
-
     if (typeof filenames === 'string') {
-      setOutputsByLocatorId(
-        nodeLocatorId,
+      setNodeOutputsByNodeId(
+        node.id,
         createOutputs([filenames], folder, isAnimated)
       )
     } else if (!Array.isArray(filenames)) {
-      setOutputsByLocatorId(nodeLocatorId, filenames)
+      setNodeOutputsByNodeId(node.id, filenames)
     } else {
       const resultItems = createOutputs(filenames, folder, isAnimated)
       if (!resultItems?.images?.length) return
-      setOutputsByLocatorId(nodeLocatorId, resultItems)
+      setNodeOutputsByNodeId(node.id, resultItems)
     }
   }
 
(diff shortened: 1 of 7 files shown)


src/stores/imagePreviewStore.ts
```diff
@@ -96,24 +133,67 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
+   * @param options.merge - If true, merge with existing outputs (arrays are concatenated)
+   */
+  function setNodeOutputsByNodeId(
+    nodeId: string | number,
+    outputs: ExecutedWsMessage['output'] | ResultItem,
+    options: SetOutputOptions = {}
+  ) {
+    const nodeLocatorId = nodeIdToNodeLocatorId(nodeId)
+    setOutputsByLocatorId(nodeLocatorId, outputs, options)
+  }
```
webfiltered: This is unused - is there a plan to use it?

webfiltered (commented):
Please pause on merging this; I am investigating the use of locator ID here.

src/stores/imagePreviewStore.ts
christian-byrne: Yeah, this should be getting called by `setNodeOutputs`.

webfiltered: It's unused.

webfiltered (commented):
My design concern here is that we are silently discarding the execution ID, which doesn't make a lot of sense to me.

My general design for storing execution results (progress, output, everything) was to always keep the ref to the execution ID, as it is the source of truth.  Not suggesting we don't use locator IDs for the reverse lookup - suggesting both be in the store.

Previews are currently still bust:

image

christian-byrne:
> My general design for storing execution results (progress, output, everything) was to always keep the ref to the execution ID, as it is the source of truth. Not suggesting we don't use locator IDs for the reverse lookup - suggesting both be in the store.

We can add executionId->nodeOutput to the store but since it won't be used yet it might belong in followup PR. Currently, the output store is also used for things like LoadImage which don't have access to the execution Id.

christian-byrne:
> Previews are currently still bust:

~This PR is for outputs, I'll update mid-execution previews in separate PR.~ Actually let me just update here

PR #4519, opened 2025-07-24
Fixed Square Brush, Improve Brush Hardness and Smoothing Precision

brucew4yn3rp:
This PR does the following:

- Improves the scaling and calculation of brush hardness
  - Historically, a radial gradient was being applied to both the circle and square pens with a hardness < 1, which was causing the square pen to revert into a circle. This has been fixed with the use of a helper, drawSoftSquare function. 
- Improves the scale of smoothing precision so that there is a meaningful difference in step count based on the smoothing precision.

https://github.com/user-attachments/assets/b117595e-3a74-4f13-8a36-01f0430e02aa

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4519-Fixed-Square-Brush-Improve-Brush-Hardness-and-Smoothing-Precision-23a6d73d365081a4b7efd05b4d601300) by [Unito](https://www.unito.io)


brucew4yn3rp:
@trsommer @webfiltered Quickly following up; this is a much smaller, addendum, PR that improves functionality to [4361](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4361)

brucew4yn3rp:
@christian-byrne @webfiltered Can you please review this PR when you have a chance?

brucew4yn3rp:
Hi @webfiltered @christian-byrne 

Following up on the review of this patch as it fixes and improves functionality of the new Image Canvas. Shouldn’t be too much in terms of changes to review.

src/extensions/core/maskeditor.ts
```diff
@@ -2435,82 +2449,114 @@ class BrushTool {
...
+    // Helper function to draw soft square brush
+    const drawSoftSquare = (
+      ctx: CanvasRenderingContext2D,
+      color: string,
+      centerOpacity: number
+    ) => {
+      const steps = 10
+      const hardRadius = brushRadius * hardness
+
+      for (let i = 0; i < steps; i++) {
```
christian-byrne: **Performance Consideration**: The loop creates 10 fill operations per brush stroke which could impact drawing performance on high-DPI screens or during rapid brush movements.

**Suggestion**: Consider using a single gradient with multiple color stops or pre-rendering to an off-screen canvas for better performance:
```typescript
// Alternative approach using gradient
const gradient = ctx.createRadialGradient(x, y, hardRadius, x, y, brushRadius)
for (let i = 0; i <= steps; i++) {
  const progress = i / steps
  const opacity = centerOpacity * (1 - progress)
  gradient.addColorStop(progress, color.replace(/[\d.]+\)$/, `${opacity})`))
}
```

*Note: This is a performance optimization suggestion - the current implementation works correctly.*

src/extensions/core/maskeditor.ts
```diff
@@ -2435,82 +2449,114 @@ class BrushTool {
...
+        const progress = i / (steps - 1)
+        const currentRadius = hardRadius + (brushRadius - hardRadius) * progress
+        const currentOpacity = centerOpacity * (1 - progress)
+
+        ctx.fillStyle = color.replace(/[\d.]+\)$/, `${currentOpacity})`)
```
christian-byrne: **Code Quality**: Consider using existing colorUtil functions for safer color manipulation instead of regex replacement.

**Suggestion**: Use the robust parsing functions from `src/utils/colorUtil.ts`:
```typescript
import { adjustColor } from '@/utils/colorUtil'

// Instead of:
ctx.fillStyle = color.replace(/[\d.]+\)$/, `${currentOpacity})`)

// Use:
ctx.fillStyle = adjustColor(color, { opacity: currentOpacity })
```

The colorUtil functions have test coverage and handle edge cases better than regex manipulation.

src/extensions/core/maskeditor.ts
```diff
@@ -2285,9 +2285,23 @@ class BrushTool {
...
+    // Convert 1-100 range to 0-1 range
+    const smoothing = Math.min(
+      Math.max(this.brushSettings.smoothingPrecision, 1),
+      100
+    ) // clamp 1-100
+    const normalizedSmoothing = (smoothing - 1) / 99 // Convert to 0-1 range
+
+    // Optionality to use exponential curve
+    const stepNr = Math.round(
+      minSteps + (maxSteps - minSteps) * Math.pow(normalizedSmoothing, 1)
```
christian-byrne: **Code Quality**: `Math.pow(normalizedSmoothing, 1)` is redundant since any number to the power of 1 equals itself.

**Suggestion**: 
```typescript
// Remove Math.pow entirely:
const stepNr = Math.round(minSteps + (maxSteps - minSteps) * normalizedSmoothing)

// Or use a meaningful exponent for curve shaping:
const stepNr = Math.round(minSteps + (maxSteps - minSteps) * Math.pow(normalizedSmoothing, 0.5))
```

src/extensions/core/maskeditor.ts
```diff
@@ -2285,9 +2285,23 @@ class BrushTool {
       totalLength += Math.sqrt(dx * dx + dy * dy)
     }
 
-    const distanceBetweenPoints =
-      (this.brushSettings.size / this.brushSettings.smoothingPrecision) * 6
-    const stepNr = Math.ceil(totalLength / distanceBetweenPoints)
+    const maxSteps = 30
```
christian-byrne: **Maintainability**: Magic numbers (30, 2, 10) scattered throughout make the code harder to maintain and understand.

**Suggestion**: Define constants at class level:
```typescript
private static readonly SMOOTHING_MAX_STEPS = 30
private static readonly SMOOTHING_MIN_STEPS = 2
private static readonly SOFT_BRUSH_STEPS = 10
```

src/extensions/core/maskeditor.ts
```diff
@@ -2435,82 +2449,114 @@ class BrushTool {
...
+          x - currentRadius,
+          y - currentRadius,
+          currentRadius * 2,
+          currentRadius * 2
+        )
+        ctx.fill()
+      }
+    }
+
+    // RGB brush logic
```
christian-byrne, line 2530: **Code Quality**: Similar gradient creation logic exists for both RGB and mask brushes, leading to maintenance overhead.

**Suggestion**: Extract common gradient creation logic:
```typescript
private createBrushGradient(
  ctx: CanvasRenderingContext2D, 
  x: number, y: number,
  brushRadius: number, 
  hardness: number, 
  color: string, 
  opacity: number
) {
  // Common gradient logic here
  return gradient
}
```

src/extensions/core/maskeditor.ts
```diff
@@ -2285,9 +2285,23 @@ class BrushTool {
...
     }
 
-    const distanceBetweenPoints =
-      (this.brushSettings.size / this.brushSettings.smoothingPrecision) * 6
-    const stepNr = Math.ceil(totalLength / distanceBetweenPoints)
+    const maxSteps = 30
+    const minSteps = 2
+
+    // Convert 1-100 range to 0-1 range
+    const smoothing = Math.min(
```
christian-byrne: **Code Organization**: Input validation mixed with business logic makes the code harder to read and test.

**Suggestion**: Extract validation to a separate method:
```typescript
private clampSmoothingPrecision(value: number): number {
  return Math.min(Math.max(value, 1), 100)
}

// Then use:
const smoothing = this.clampSmoothingPrecision(this.brushSettings.smoothingPrecision)
```

christian-byrne (commented):
Great work fixing the square brush hardness issue! The solution correctly addresses the core problem where square brushes were becoming circular when hardness < 1.

**Key Improvements:**

- Fixes square brush rendering with proper hardness control
- Improves smoothing precision algorithm with better granular control  
- Maintains backward compatibility
- Clear separation between RGB and mask brush logic

I've added specific inline comments focusing on performance optimizations and code quality improvements. The most important suggestions are:

1. **Performance**: Consider optimizing the 10-fill loop in `drawSoftSquare`
2. **Code Quality**: Leverage existing `colorUtil` functions for safer color manipulation

The changes work correctly as-is - these are optimization suggestions for better maintainability and performance.

src/extensions/core/maskeditor.ts
brucew4yn3rp: Thanks, nice catch. I was testing alternate exponents previously so the Math.pow was leftover

src/extensions/core/maskeditor.ts
brucew4yn3rp: Thanks @christian-byrne. Due to the createRadialGradient method creating circular falloffs, for square brushes I believe the current 10-fill approach is the most straightforward solution. Given the use of this function only when square brushes are being utilized at a hardness < 1, the potential performance impact might be acceptable. With 10 fill operations per stroke, I don't anticipate it to be problematic from my testing.

src/extensions/core/maskeditor.ts
brucew4yn3rp: The logic may have similarities, but I believe it is best kept separate, due to the rectangle brush not being able to use createRadialGradient. The actual code duplication is in the radial gradient creation logic only, and it only applies to:

- Circle brushes (any hardness)
- Rectangle brushes with hardness = 1

Keeping the logic separate maintains clarity about when each rendering method is used

src/extensions/core/maskeditor.ts
brucew4yn3rp: Following up - I've implemented a tempCtx to render the square gradient more effectively by manipulating pixel opacities.

brucew4yn3rp:
Thank you for your review @christian-byrne. I've replied to your comments and pushed some updates accordingly.

christian-byrne (approved):
LGTM

christian-byrne (changes requested):
Performance concern with the drawSoftSquare function - is it creating a new canvas and manipulating ImageData for every brush stroke when using soft square brushes?


That seems liek it would be significantly slower than the existing gradient-based rendering for other brush types. Could we consider using cached brush textures or optimizing the ImageData approach to avoid creating new canvases on each stroke?

The fix for square brush hardness is great, but the performance impact could cause lag during drawing for users of soft rectangular brushes.

christian-byrne (changes requested):
```javascript
const tempCanvas = document.createElement('canvas')
const tempCtx = tempCanvas.getContext('2d')
// ... manual pixel manipulation loop
```

brucew4yn3rp:
Thanks @christian-byrne. I've fixed brush rendering anti-aliasing and performance issues. 

- Added solid fills for max hardness brushes to eliminate gradient anti-aliasing and achieve true binary mask edges.
- Implemented texture caching for soft square brushes to avoid creating new canvases on every stroke. The cache is self-managing in terms of memory to avoid bloat.

Additionally, It seems that @trsommer has reacted with a thumbs up to Kevin's message [here](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4361#issuecomment-3137711915) in regards to adding me as codeowner. Could you / the team please do so when you have a chance?

christian-byrne:
> Additionally, It seems that @trsommer has reacted with a thumbs up to Kevin's message https://github.com/Comfy-Org/ComfyUI_frontend/pull/4361#issuecomment-3137711915 in regards to adding me as codeowner. Could you / the team please do so when you have a chance?

Sure, could you make a PR to add yourself to the CODEOWNERS file?

brucew4yn3rp:
Done - can you please approve changes if they look good to you?

src/extensions/core/maskeditor.ts
```diff
@@ -2049,9 +2050,15 @@ class BrushTool {
   rgbCtx: CanvasRenderingContext2D | null = null
   initialDraw: boolean = true
 
+  private static brushTextureCache = new Map<string, HTMLCanvasElement>()
```
christian-byrne: What about using an LRU cache? We have a library we use in some other places, it can be used like:

```typescript
import QuickLRU from '@alloc/quick-lru'

private static brushTextureCache = new QuickLRU<string, HTMLCanvasElement>({
  maxSize: 8 // Reasonable limit for brush texture variations?
})
```

christian-byrne (commented):
This is looking good. I tested it and it works nicely. What are your thoughts on adding a unit test :smile:? You don't have to.

brucew4yn3rp:
I can leverage QuickLRU but I'm getting the following error when doing `pnpm run build`. Have you seen this before? I git clone'd the main Comfy repo and the error is happening with that one too:

```
> vue-tsc --noEmit

error TS2688: Cannot find type definition file for 'dompurify'.
  The file is in the program because:
    Entry point for implicit type library 'dompurify'
   ```

**Comment by christian-byrne** [comment]
Does it still happen when using node 24?

`nvm use 24`

**Comment by brucew4yn3rp** [comment]
It does still happen when using v24.4.0

I bypassed the dompurify issue by going to ComfyUI_frontend\src\types and creating a file `dompurify.d.ts` that contains: `declare module "dompurify"`

But there are still issues with the packages. This is likely stemming from the switch to pnpm in the latest frontend updates. For example, the `primeuix` `forms`, `styles`, and `themes` are all giving errors like the following

```
 ERROR  error during build:
[generate-import-map-plugin] ENOENT: no such file or directory, copyfile 'C:\Users\BATMAN\Desktop\Testing\ComfyUI_frontend\node_modules\@primeuix\forms\index.mjs' -> 'C:\Users\BATMAN\Desktop\Testing\ComfyUI_frontend\dist\assets\lib\@primeuix\forms\index.mjs'
    at Object.copyFileSync (node:fs:3100:11)
    at Object.generateBundle (file:///C:/Users/BATMAN/Desktop/Testing/ComfyUI_frontend/vite.config.mts.timestamp-1756603068573-7b2cee086facf8.mjs:140:14)
    at file:///C:/Users/BATMAN/Desktop/Testing/ComfyUI_frontend/node_modules/.pnpm/rollup@4.49.0/node_modules/rollup/dist/es/shared/node-entry.js:22395:40
```

Are you not seeing the errors on your end?
It might have to do with pnpm’s behavior on Windows

> Switching to pnpm breaks things because it doesn’t use a flat node_modules like npm/yarn — instead it creates symlinks into a hidden store. Tools and type resolvers (like TypeScript looking for @types/dompurify or your import-map plugin looking for @primeuix/styles) often assume the old flat layout, so they can’t find files where they expect them and throw ENOENT or “cannot find type definition” errors.

Might be easier to chat on discord to not sidetrack this thread? I sent you a DM and also posted in the frontend channel of the Comfy server.

**Comment by christian-byrne** [comment]
> Are you not seeing the errors on your end?

No, I am not. Have you tried deleeting the node_modules folder and reinstalling?

```bash
rm -rf node_modules pnpm-lock.yaml
pnpm install 
```

Can also try 

```bash
rm -rf node_modules pnpm-lock.yaml
pnpm install --shamefully-hoist 
```

brucew4yn3rp:
@christian-byrne Fixed the issue. The frontend’s switch to pnpm resulted in much longer file paths due to its symlinked `node_modules` structure. On my system, `pnpm install` was silently failing because Windows long paths were not enabled in the group policy:

Just pushed an update that uses QuickLRU for the brushTextureCache. 

If the changes look good to you, can you approve instead of @trsommer? I'm not sure that he is active on this PR.

src/extensions/core/maskeditor.ts
```diff
@@ -2049,9 +2051,17 @@ class BrushTool {
...
+  private static brushTextureCache = new QuickLRU<string, HTMLCanvasElement>({
+    maxSize: 8 // Reasonable limit for brush texture variations?
+  })
+
   brushStrokeCanvas: HTMLCanvasElement | null = null
   brushStrokeCtx: CanvasRenderingContext2D | null = null
 
+  private static readonly SMOOTHING_MAX_STEPS = 30
+  private static readonly SMOOTHING_MIN_STEPS = 2
+  //private static readonly SOFT_BRUSH_STEPS = 20
```
christian-byrne: Should this be removed?

brucew4yn3rp: Right, the SOFT_BRUSH_STEPS should be removed but   SMOOTHING_MAX_STEPS and SMOOTHING_MIN_STEPS are used.

christian-byrne (approved):
LGTM!

PR #4475, opened 2025-07-19
minimap

jtydhr88:
As part of New Gen, this PR implements a minimap, making it convenient for users to preview the entire workflow from a global perspective.

**Minimap Feature Summary**
This PR introduces a minimap component that provides users with a bird's-eye view of their entire workflow. Here are the key features for review:

**Core Functionality**

Global Overview: Displays a scaled-down representation of the entire workflow graph, allowing users to see the big picture at a glance
Interactive Navigation: Users can click and drag within the minimap to quickly pan to different areas of the main canvas
Viewport Indicator: Shows a highlighted rectangle representing the currently visible area of the main canvas

**User Controls**

Toggle Visibility: Can be shown/hidden via a dedicated button in the canvas menu or keyboard shortcut (Alt+M)
Zoom Support: Mouse wheel interactions allow zooming in/out directly from the minimap
Persistent Settings: Remembers the user's visibility preference across sessions

**Integration**

Real-time Updates: Automatically reflects changes when nodes are added, removed, or connections are modified
Responsive Design: Adapts to different canvas sizes and maintains proper scaling
Non-intrusive Positioning: Fixed in the bottom-right corner with appropriate z-indexing

Demo video

https://github.com/user-attachments/assets/8cbb64cc-048e-41b9-acc4-653f23cab5f1

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4475-minimap-2356d73d365081babee2fd2e39eeda24) by [Unito](https://www.unito.io)

PR #4520, opened 2025-07-24
minimap

jtydhr88:
(replacement for https://github.com/Comfy-Org/ComfyUI_frontend/pull/4475 as there are too many conflicts hard to fix in previous one)
As part of New Gen, this PR implements a minimap, making it convenient for users to preview the entire workflow from a global perspective.

**Minimap Feature Summary**
This PR introduces a minimap component that provides users with a bird's-eye view of their entire workflow. Here are the key features for review:

**Core Functionality**

Global Overview: Displays a scaled-down representation of the entire workflow graph, allowing users to see the big picture at a glance
Interactive Navigation: Users can click and drag within the minimap to quickly pan to different areas of the main canvas
Viewport Indicator: Shows a highlighted rectangle representing the currently visible area of the main canvas

**User Controls**

Toggle Visibility: Can be shown/hidden via a dedicated button in the canvas menu or keyboard shortcut (Alt+M)
Zoom Support: Mouse wheel interactions allow zooming in/out directly from the minimap
Persistent Settings: Remembers the user's visibility preference across sessions

**Integration**

Real-time Updates: Automatically reflects changes when nodes are added, removed, or connections are modified
Responsive Design: Adapts to different canvas sizes and maintains proper scaling
Non-intrusive Positioning: Fixed in the bottom-right corner with appropriate z-indexing

Demo video

https://github.com/user-attachments/assets/8cbb64cc-048e-41b9-acc4-653f23cab5f1

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4520-minimap-23a6d73d3650813689bdcaec02dc615b) by [Unito](https://www.unito.io)


src/composables/useMinimap.ts
```diff
@@ -0,0 +1,631 @@
...
+
+    ds.offset[0] = -(worldX - viewportWidth / 2)
+    ds.offset[1] = -(worldY - viewportHeight / 2)
+
+    updateFlags.value.viewport = true
+
+    c.setDirty(true, true)
+  }
+
+  let originalCallbacks: any = {}
```
christian-byrne: Should we just define a type for this object?

src/composables/useMinimap.ts
```diff
@@ -0,0 +1,631 @@
...
+
+  const calculateScale = () => {
+    if (bounds.value.width === 0 || bounds.value.height === 0) {
+      return 1
+    }
+
+    const scaleX = width / bounds.value.width
+    const scaleY = height / bounds.value.height
+
+    return Math.min(scaleX, scaleY) * 0.9
```
christian-byrne, line 180: Why 0.9?

src/components/graph/MiniMap.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+  <div
+    v-if="visible"
```
christian-byrne: Should we also wait for `initialized`?

src/composables/useMinimap.ts
```diff
@@ -0,0 +1,631 @@
...
+import type { NodeId } from '@/schemas/comfyWorkflowSchema'
+import { api } from '@/scripts/api'
+import { useCanvasStore } from '@/stores/graphStore'
+import { useSettingStore } from '@/stores/settingStore'
+
+const globalState = {
+  visible: ref(true),
+  minimapRef: ref<any>(null),
+  initialized: false
+}
```
christian-byrne: I think the state can be bound to the instance. It will be mounted / cleaned up when the parent component (GraphCanvas) is mounted / unmounted. In the future won't we possibly want to re-use across multiple instances anyway?

christian-byrne:
Sorry forgot to push the button to send the review earlier today.

src/composables/useMinimap.ts
christian-byrne, line 1: I also think we will want to split this into separate modules. At least, we will want a module that can render a preview of the graph -- so we can re-use it for other preview systems (tab preview, subgraph preview, workflow thumbnail).

src/composables/useMinimap.ts
jtydhr88: if there is no this factor (or 1), it will be no padding (gap) between node and boards

christian-byrne:
Changes:

- Added type for callbacks object
- Removed global state singleton pattern and convert to stateful composable
- Don't render component if minimap not initialized
- Add documentation for the 0.9 magic number
- Visual design to match Figma design: Blue nodes (#0B8CE999), orange connections/slots (#F99614), charcoal background (#15161C), 8px border radius
- Create minimum viewport dimensions to prevent small number of nodes being zoomed too much
- Performance: Fast path rendering for 0 nodes case
- Merged with main to resolved conflict in apiSchema

christian-byrne:
Updated unit tests to mock `arc` and `fill` and expect new colors.

christian-byrne:
Changed toggle logic.

christian-byrne (approved):
LGTM!

PR #4527, opened 2025-07-24
Refactor canvas transform sync to composable

christian-byrne:
Adds composable to sync transform state of canvas. Call with a `sync` function that gets called on the canvas only when the transform state has actually changed.

This utility will be used by future components, and having the syncing happen in a single place will be much more efficient because we won't need to have multiple observers each diffing the state themselves.

This also changes the implementation from calling `updateViewport` 60 times per second to only calling on change.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4527-Refactor-canvas-transform-sync-to-composable-23a6d73d3650816a993bc69b9abb2279) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/canvas/useCanvasTransformSync.ts
+++ b/src/composables/canvas/useCanvasTransformSync.ts
@@ -1,7 +1,21 @@
+import { LGraphCanvas } from '@comfyorg/litegraph'
 import { onUnmounted, ref } from 'vue'
 
-export interface TransformSyncOptions {
+import { useCanvasStore } from '@/stores/graphStore'
+
+interface CanvasTransformSyncOptions {
+  /**
+   * Whether to automatically start syncing when canvas is available
+   * @default true
+   */
+  autoStart?: boolean
+  /**
+   * Called when sync starts
+   */
   onStart?: () => void
+  /**
+   * Called when sync stops
+   */
   onStop?: () => void
 }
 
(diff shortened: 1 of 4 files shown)


src/composables/canvas/useCanvasTransformSync.ts
```diff
@@ -0,0 +1,132 @@
...
+  offsetY: number
+}
+
+/**
+ * Manages requestAnimationFrame-based synchronization with LiteGraph canvas transforms.
+ *
+ * This composable provides a clean way to sync Vue transform state with LiteGraph canvas
+ * on every frame. It handles RAF lifecycle management, and ensures proper cleanup.
+ *
+ * The sync function typically reads canvas.ds (draw state) properties like offset and scale
```
christian-byrne: oh, `ds` stands for "Drag and Scale" not "draw state" actually

src/composables/canvas/useCanvasTransformSync.ts
```diff
@@ -0,0 +1,132 @@
...
+ * ```ts
+ * const { isActive, startSync, stopSync } = useCanvasTransformSync(
+ *   canvas,
+ *   (canvas) => syncWithCanvas(canvas),
+ *   {
+ *     onStart: () => emit('rafStatusChange', true),
+ *     onUpdate: (time) => emit('transformUpdate', time),
+ *     onStop: () => emit('rafStatusChange', false)
+ *   }
+ * )
```
christian-byrne: this example is outdated

jtydhr88 (approved):
LGTM

PR #4535, opened 2025-07-25
[Manager] Fix banned status logic to only check Registry status

viva-jinyi:
## Summary
- Fixed `is_banned` logic in conflict detection to only check `NodeVersionStatusBanned` from Registry
- Removed local enabled/disabled state from banned determination  
- Updated tests to reflect correct banned status behavior

## Changes
- Modified `useConflictDetection.ts` to set `is_banned` based only on Registry status
- Updated test cases to verify locally disabled packages are NOT treated as banned
- Added test coverage for `NodeVersionStatusPending` and `NodeVersionStatusFlagged` states

## Context
Previously, packages were incorrectly marked as "banned" when they were simply disabled locally (`enabled: false`). This caused confusion as normal packages appeared with banned status in the UI.

Now the logic correctly distinguishes between:
- **Registry banned**: `versionData.status === 'NodeVersionStatusBanned'` - Actually banned by Registry
- **Locally disabled**: User disabled the package locally (not a ban)

## Test plan
- [x] Verified locally disabled packages no longer show as banned
- [x] Confirmed Registry-banned packages still correctly identified

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4535-Manager-Fix-banned-status-logic-to-only-check-Registry-status-23b6d73d365081999948c673f0e1759a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useConflictDetection.ts
+++ b/src/composables/useConflictDetection.ts
@@ -280,14 +280,7 @@ export function useConflictDetection() {
             // Status information
             registry_status: undefined, // Node status - not critical for conflict detection
             version_status: versionData.status,
-            is_banned:
-              versionData.status === 'NodeVersionStatusBanned' || !isEnabled,
-            ban_reason:
-              versionData.status === 'NodeVersionStatusBanned'
-                ? 'Version is banned in Registry'
-                : !isEnabled
-                  ? 'Package is disabled locally'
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):
LGTM!

PR #4538, opened 2025-07-25
[Manager] Optimize conflict detection with bulk API and version support

viva-jinyi:
## Summary
- Add `installedPacksWithVersions` computed property to useInstalledPacks to include version information
- Replace individual version API calls with bulk API in conflict detection
- Add `getBulkNodeVersions` API method to comfyRegistryService for efficient batch requests
- Use actual installed package versions instead of latest versions for compatibility checks

## Changes
- **src/composables/nodePack/useInstalledPacks.ts**: Added installedPacksWithVersions computed property that combines nodepack id with version info
- **src/composables/useConflictDetection.ts**: Refactored to use bulk API calls instead of individual requests, significantly reducing API overhead
- **src/services/comfyRegistryService.ts**: Added getBulkNodeVersions method for batch version fetching

## Why this change?
The existing `installedPacks` doesn't include version information, which was needed for accurate conflict detection. This change improves performance by reducing API calls from N individual requests to N/30 bulk requests while ensuring compatibility checks use actually installed versions.

Performance improvement: Reduces API calls significantly for conflict detection operations.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4538-Manager-Optimize-conflict-detection-with-bulk-API-and-version-support-23b6d73d3650819ebe2cdf9c47ed4e4b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useConflictDetection.ts
+++ b/src/composables/useConflictDetection.ts
@@ -210,27 +210,19 @@ export function useConflictDetection() {
       // Step 3: Setup abort controller for request cancellation
       abortController.value = new AbortController()
 
-      // Step 4: Use bulk API to fetch version data efficiently
-      const chunkSize = 30 // Process in chunks to avoid overwhelming the API
+      // Step 4: Use bulk API to fetch all version data in a single request
       const versionDataMap = new Map<
         string,
         components['schemas']['NodeVersion']
       >()
 
-      // Use installedPacksWithVersions which has the actual installed versions from Manager API
-      for (
-        let i = 0;
-        i < installedPacksWithVersions.value.length;
-        i += chunkSize
-      ) {
-        const chunk = installedPacksWithVersions.value.slice(i, i + chunkSize)
+      // Prepare bulk request with actual installed versions from Manager API
+      const nodeVersions = installedPacksWithVersions.value.map((pack) => ({
+        node_id: pack.id,
+        version: pack.version
+      }))
 
-        // Prepare bulk request with actual versions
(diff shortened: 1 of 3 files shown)


src/composables/useConflictDetection.ts
```diff
@@ -203,58 +204,67 @@ export function useConflictDetection() {
...
       }
 
       // Step 5: Combine local installation data with Registry version data
-      // Use installedPacks from useInstalledPacks composable
       const requirements: NodePackRequirements[] = []
 
+      // Create a map for quick access to version info
+      const versionInfoMap = new Map(
+        installedPacksWithVersions.value.map((pack) => [pack.id, pack.version])
+      )
```
christian-byrne, line 262: What about moving this Map to the store and making it a `computed` dependent on installedPacksIds like

```typescript
      const installedNodePackVersionsById = computed(() => new Map(
        installedPackIds.value.map((packId) => installedPacksWithVersions[packId] )
      ))
```

or similar

viva-jinyi: @christian-byrne The `versionInfoMap` is a temporary data structure that's only needed within the conflict detection logic, so I don't think it's necessary to move it to a computed property. It's used locally just for this specific operation and doesn't need to be reactive or shared with other components.

christian-byrne: SGTM

christian-byrne (approved):
LGTM!

PR #4547, opened 2025-07-27
[fix] Detect missing nodes in subgraphs

### LINKED ISSUE #2032: [Bug/Oversight] Custom node suites are not picked up from subgraphs
I am currently testing quite a large workflow at the moment that I intend to publish publicly, which currently I have a near-complete version here:

[noobai txt2img wf v0.9.json](https://github.com/user-attachments/files/21441922/noobai.txt2img.wf.v0.9.json)

The issue however, is that I tested this on another install on ComfyUI. Practically _almost_ all the custom nodes were able to be picked up (with an issue of [giriss' comfy-image-saver suite](https://github.com/giriss/comfy-image-saver) being picked up instead of [alexopus' comfy-image saver suite](https://github.com/alexopus/ComfyUI-Image-Saver) which causes an issue due to one node from alexopus' suite being used that doesn't exist on giriss'), however, I noticed that the widgets on the Detailer Models subgraph was not being picked up.

What it should look like, versus how it shows up on the other install:
Image
Image

And when I checked the subgraph on the new install:
Image

Comfy-Manager does not identify that the impact-subpack is missing, not in the main workflow, or even in the subgraph itself. This is a flaw that the group nodes had that I thought would've been corrected with subgraphs. It should be easier to be able to detect missing custom node suites from subgraphs than from group nodes, so this issue ought to be addressed before subgraphs are made a part of future releases of the frontend that are listed in the requirements.txt of ComfyUI itself.

viva-jinyi:
## Summary
- Use `collectAllNodes` utility to recursively scan subgraphs for missing custom nodes
- Update both `useMissingNodes` and `useWorkflowPacks` to include nodes from all graph levels
- This ensures Manager can detect and list custom nodes that are inside subgraphs

## Test plan
- [x] Create a workflow with custom nodes inside a subgraph
- [x] Load the workflow in a ComfyUI instance without those custom nodes installed
- [x] Open Manager → "Install Missing Custom Nodes"
- [x] Verify that custom nodes from inside subgraphs are now listed

Fixes [#2032](https://github.com/Comfy-Org/ComfyUI-Manager/issues/2032)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4547-fix-Detect-missing-nodes-in-subgraphs-23d6d73d36508153b861d8111af319d2) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/nodePack/useMissingNodes.ts
+++ b/src/composables/nodePack/useMissingNodes.ts
@@ -8,6 +8,7 @@ import { app } from '@/scripts/app'
 import { useComfyManagerStore } from '@/stores/comfyManagerStore'
 import { useNodeDefStore } from '@/stores/nodeDefStore'
 import type { components } from '@/types/comfyRegistryTypes'
+import { collectAllNodes } from '@/utils/graphTraversalUtil'
 
 /**
  * Composable to find missing NodePacks from workflow
@@ -56,7 +57,8 @@ export const useMissingNodes = () => {
   }
 
   const missingCoreNodes = computed<Record<string, LGraphNode[]>>(() => {
-    const missingNodes = app.graph.nodes.filter(isMissingCoreNode)
+    const allNodes = collectAllNodes(app.graph)
+    const missingNodes = allNodes.filter(isMissingCoreNode)
     return groupBy(missingNodes, (node) => String(node.properties?.ver || ''))
   })
 
--- a/src/composables/nodePack/useMissingNodes.ts
+++ b/src/composables/nodePack/useMissingNodes.ts
@@ -57,8 +57,7 @@ export const useMissingNodes = () => {
   }
 
   const missingCoreNodes = computed<Record<string, LGraphNode[]>>(() => {
-    const allNodes = collectAllNodes(app.graph)
-    const missingNodes = allNodes.filter(isMissingCoreNode)
+    const missingNodes = collectAllNodes(app.graph, isMissingCoreNode)
     return groupBy(missingNodes, (node) => String(node.properties?.ver || ''))
   })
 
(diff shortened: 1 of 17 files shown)


src/composables/nodePack/useMissingNodes.ts
```diff
@@ -56,7 +57,8 @@ export const useMissingNodes = () => {
   }
 
   const missingCoreNodes = computed<Record<string, LGraphNode[]>>(() => {
-    const missingNodes = app.graph.nodes.filter(isMissingCoreNode)
+    const allNodes = collectAllNodes(app.graph)
+    const missingNodes = allNodes.filter(isMissingCoreNode)
```
christian-byrne: We can pass a filter function to `collectAllNodes`

https://github.com/Comfy-Org/ComfyUI_frontend/blob/68f50670d363f30791c052d4b3354bf08442ba8d/src/utils/graphTraversalUtil.ts#L103-L113

Then, we only need to traverse a single time.

viva-jinyi: Thanks for pointing that out!

src/composables/nodePack/useWorkflowPacks.ts
```diff
@@ -109,11 +110,13 @@ export const useWorkflowPacks = (options: UseNodePacksOptions = {}) => {
...
-   * Get the node packs for all nodes in the workflow.
+   * Get the node packs for all nodes in the workflow (including subgraphs).
    */
   const getWorkflowPacks = async () => {
-    if (!app.graph?.nodes?.length) return []
-    const packs = await Promise.all(app.graph.nodes.map(workflowNodeToPack))
+    if (!app.graph) return []
+    const allNodes = collectAllNodes(app.graph)
+    if (!allNodes.length) return []
+    const packs = await Promise.all(allNodes.map(workflowNodeToPack))
```
christian-byrne, line 119: Thanks for catching this one

christian-byrne (approved):
LGTM, thanks for fixing!

PR #4584, opened 2025-07-30
Scroll templates better

### LINKED ISSUE #3736: [Feature Request] Make workflow templates searchable
With the growing number of workflow templates. A search functionality is now necessary to better nativate the content.

Image

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-3736-Feature-Request-Make-workflow-templates-searchable-1e76d73d36508180948ef1eb58d528d4) by [Unito](https://www.unito.io)

### LINKED ISSUE #4488: [Bug]: Why does it take a long time to load every time I click on the workflow template after launching the page? Is it due to network issues?
### Custom Node Testing

### Frontend Version

/

### Expected Behavior

Why does it take a long time to load every time I click on the workflow template after launching the page? Is it due to network issues?

### Actual Behavior

/

### Steps to Reproduce

/

### Debug Logs

```powershell
/
```

### Browser Logs

/

### Setting JSON

/

Microsoft Edge

### Other Information

/

### LINKED ISSUE #4496: [Feature Request]: Add lazy loading (with infinite scroll) to the templates container
## Is there an existing issue for this?

## What would your feature do?

This feature would implement lazy loading and/or pagination for the workflow templates dialog to significantly improve performance when loading template galleries, especially for users with many custom nodes installed.

Currently, the templates dialog eagerly loads all images, videos, and audio files for every template, which causes performance issues as reported in #4488. This is particularly problematic when:
- The "All" tab is selected (which combines all core and custom node templates)
- Many custom nodes are installed, each potentially adding their own templates with media

The feature would:
1. **Lazy load media**: Only load images/videos/audio when templates are visible or about to become visible in the viewport
2. **Virtual scrolling**: Render only visible templates in the DOM when dealing with large lists
3. **Progressive loading**: Show placeholders/skeletons while media loads
4. **Intelligent caching**: Cache loaded media to avoid re-fetching

## Proposed workflow

1. User opens the workflow templates dialog
2. Only templates visible in the viewport (plus a small buffer) have their media loaded
3. As user scrolls, new templates entering the viewport begin loading their media
4. Previously loaded media remains cached for instant display when scrolling back
5. For categories with many templates (like "All"), virtual scrolling ensures only visible items are rendered in the DOM

### Technical implementation approach:

These are possible things to add but the implementation does not need to follow this exactly.

- Use Intersection Observer API for viewport-based lazy loading
- Implement virtual scrolling for grid/list views with large numbers of templates
- Add skeleton loaders or blur-up placeholders during media loading
- Create a media caching service to prevent redundant fetches
- Consider pagination as an alternative/addition for extremely large collections

### Existing components/utilities to leverage:

- **PrimeVue Skeleton**: Already integrated for loading states throughout the app
- **Skeleton component helper**: Existing skeleton utilities can be used for template card placeholders
- PrimeVue offers some things like **VirtualScroller** and a datatable component with virtual scrolling. Vueuse offers `useInfiniteScroll`. However, neither of these support components with dynamically sized CSS grids that can be resized. So, we have to use our [VirtualGrid](https://github.com/Comfy-Org/ComfyUI_frontend/blob/61611fb0cba3fa14e57081a0c502df232465c1c8/src/components/common/VirtualGrid.vue#L4) (most likely).

## Additional information

This addresses the performance regression noted in #4488 where the templates dialog has become slow due to the growing number of templates. The original implementation worked well with a small number of templates but doesn't scale.

Related optimizations that could be considered:

- Default to a category other than "All" to reduce initial load
- Request lower resolution thumbnails initially, load full resolution on hover/selection
- Implement HTTP/2 push or prefetching for likely-to-be-viewed templates
- Add a search/filter feature (#3736) to reduce the number of visible templates

Performance patterns from other parts of the codebase (like the Transform Pane implementation) could be adapted, including:
- RAF-based update batching
- CSS containment for performance
- `content-visibility: auto` for off-screen elements

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4496-Feature-Request-Add-lazy-loading-and-or-pagination-to-the-templates-container-2386d73d3650819f8a45f3ea053efb6f) by [Unito](https://www.unito.io)

Myestery:
fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/4496
fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/3736

This PR helps to fix the issue with the templates page loading all images at once
This is done by implementing 
- skeleton loaders
-  frontend pagination 
- lazy loading using intersection observers
- adding a search bar to the templates page
- caching images using media cache service 
Screenshot 2025-07-30 at 01 23 35


Myestery:
Will have to fix the playwright test

src/components/common/LazyImage.vue
christian-byrne, line 1: This is a smart approach

src/components/common/LazyImage.vue
```diff
@@ -0,0 +1,113 @@
...
+    class="relative overflow-hidden w-full h-full flex items-center justify-center"
+  >
+    <Skeleton
+      v-if="!isImageLoaded"
+      width="100%"
+      height="100%"
+      class="absolute inset-0"
+    />
+    <img
+      v-show="isImageLoaded"
```
christian-byrne, line 13: Do you think it is worth it to hide when it goes back out of view?

Myestery: I think for this use case, the image is already in memory at this point and loading is no longer a problem.
So No, we should only hide when it has not been seen

christian-byrne: SGTM.

The counter-argument: When the image is rendered, it will get rasterized to a bitmap then put in a layer. The bitmap is typically considerably larger that the binary jpg/png/webm data. So tearing it down will free up a lot of memory. However, the images don't typically need re-paint because they are just transformed in compositing stage after initial render, so there's cost of tear down and re-render as well!

Myestery:
new behaviour

https://github.com/user-attachments/assets/6634fedb-386f-4542-bdb9-d4603f6c26ab

Myestery:
We may need to update test expectations for the playwright tests to work. However this should be okay for merging

src/services/mediaCacheService.ts
```diff
@@ -0,0 +1,221 @@
...
+      return errorEntry
+    }
+  }
+
+  getCacheStats() {
+    const cacheMap = this.cache.value
+    return {
+      size: cacheMap.size,
+      maxSize: this.maxSize,
+      entries: Array.from(cacheMap.keys())
```
christian-byrne: **Issue**: Race condition - Object URLs might be revoked while still in use by LazyImage components
**Severity**: High
**Impact**: Components displaying revoked URLs will show broken images

**Scenario**: 
- T1: LazyImage A starts loading image.jpg
- T2: MediaCache creates blob:url-123 for image.jpg  
- T3: LazyImage A receives blob:url-123 and renders
- T4: User scrolls quickly, cache cleanup runs
- T5: Cache calls URL.revokeObjectURL(blob:url-123)
- T6: LazyImage A still visible but now shows broken image

**Suggestion**: Track URL usage by components:
```typescript
// In LazyImage.vue
onUnmounted(() => {
  if (cachedSrc.value?.startsWith('blob:')) {
    mediaCacheService.releaseUrl(src)
  }
})

// In mediaCacheService.ts
private urlRefCount = new Map<string, number>()

releaseUrl(src: string) {
  const entry = this.cache.get(src)
  if (entry?.objectUrl) {
    const count = (this.urlRefCount.get(entry.objectUrl) || 1) - 1
    if (count <= 0) {
      URL.revokeObjectURL(entry.objectUrl)
      this.urlRefCount.delete(entry.objectUrl)
    } else {
      this.urlRefCount.set(entry.objectUrl, count)
    }
  }
}
```

src/components/common/LazyImage.vue
```diff
@@ -0,0 +1,116 @@
...
+
+import { useIntersectionObserver } from '@/composables/useIntersectionObserver'
+import { useMediaCache } from '@/services/mediaCacheService'
+
+const {
+  src,
+  alt = '',
+  imageClass = '',
+  imageStyle,
+  rootMargin = '50px'
```
christian-byrne: **Issue**: rootMargin of '50px' may be too small for template cards
**Severity**: Low
**Impact**: Images might still be loading as they enter the viewport, causing visible loading states

**Suggestion**: Increase rootMargin to be larger than a typical card height (~300px):
```typescript
rootMargin = '300px' // Load images before they're visible
```

This ensures images are fully loaded before cards scroll into view, providing a smoother experience.

src/services/mediaCacheService.ts
```diff
@@ -0,0 +1,221 @@
...
+      src,
+      isLoading: true,
+      lastAccessed: Date.now()
+    }
+
+    // Update cache with loading entry
+    const newCache = new Map(cacheMap)
+    newCache.set(src, entry)
+    this.cache.value = newCache
+
```
christian-byrne: **Issue**: Creating new Map instances on every update is inefficient
**Severity**: Medium
**Impact**: Performance degradation with frequent cache updates

**Suggestion**: Use reactive() for automatic reactivity:
```typescript
import { reactive } from 'vue'

class MediaCacheService {
  public cache = reactive(new Map<string, CachedMedia>())
  
  // Now mutations are automatically tracked
  this.cache.set(src, entry) // Vue detects this change\!
}
```

This eliminates the need to create new Map instances for every update.

src/components/templates/TemplateSearchBar.vue
```diff
@@ -0,0 +1,64 @@
...
+    </div>
+
+    <div class="flex items-center gap-4 mt-2">
+      <small
+        v-if="searchQuery && filteredCount !== null"
+        class="text-color-secondary"
+      >
+        {{ $t('g.resultsCount', { count: filteredCount }) }}
+      </small>
+      <Button
```
christian-byrne, line 34: **Issue**: Missing translation key 'g.resultsCount'
**Severity**: Low
**Impact**: Will display raw translation key to users

**Suggestion**: Add to all locale files:
```json
// src/locales/en/main.json (and other locales)
"g": {
  "resultsCount": "{count} results found"
}
```

src/components/templates/TemplateWorkflowView.spec.ts
```diff
@@ -53,10 +54,46 @@ vi.mock('@/components/templates/TemplateWorkflowList.vue', () => ({
...
 
+vi.mock('@/components/templates/TemplateSearchBar.vue', () => ({
+  default: {
+    template: '<div class="mock-search-bar"></div>',
+    props: ['searchQuery', 'filteredCount'],
+    emits: ['update:searchQuery', 'clearFilters']
+  }
+}))
+
+vi.mock('@/components/templates/TemplateWorkflowCardSkeleton.vue', () => ({
```
christian-byrne, line 65: **Issue**: Mock doesn't match actual component v-model interface
**Severity**: Low
**Impact**: Tests may pass incorrectly with wrong prop/emit names

**Suggestion**: Since TemplateSearchBar uses `defineModel`, the mock should use standard v-model props:
```typescript
vi.mock('@/components/templates/TemplateSearchBar.vue', () => ({
  default: {
    template: '<div class="mock-search-bar"></div>',
    props: ['modelValue', 'filteredCount'],  // v-model uses modelValue
    emits: ['update:modelValue', 'clearFilters']
  }
}))
```

src/components/templates/TemplateSearchBar.vue
Myestery: this already exists

src/components/templates/TemplateWorkflowView.spec.ts
Myestery: actually the prop is correct, the component is using a v-model parameter to bind to the defineModel(searchQuery) so no need to have a special modelValue prop
https://vuejs.org/guide/components/v-model#v-model-arguments

christian-byrne:
The test failures we can fix after as long as they are just screenshots and the changes are expected, which seems to be the case:

```
  3 failed
    [chromium] › browser_tests/tests/templates.spec.ts:148:3 › Templates › template cards are dynamically sized and responsive 
    [chromium] › browser_tests/tests/templates.spec.ts:176:3 › Templates › hover effects work on template cards 
    [chromium] › browser_tests/tests/templates.spec.ts:195:3 › Templates › template cards descriptions adjust height dynamically 
```

christian-byrne (approved):
LGTM, excellent work

PR #4613, opened 2025-07-31
[feat] Viewport persistence for subgraph navigation

### LINKED ISSUE #4275: [Feature Request]: Subgraphs could maintain their own viewport

When entering a subgraph, the zoom and pan seem to be those of the main graph. So the subgraph nodes may not even be visible.

It would make more sense for a subgraph to maintain it's own zoom and pan.

### Proposed workflow

When entering a subgraph, I'd like the view to be the same as when I last exited that subgraph. And if I pan or zoom in the subgraph, when I exit I'd like the view of the main graph to be the same as when I entered the subgraph.

### Additional information

The `DragAndScale` object is a property of `LGraphCanvas`. But as `app.canvas` refers to the same object regardless of which graph or subgraph is being viewed, that means there is only one `DragAndScale` object, `app.canvas.ds`.

Suggest `ds` should be a property of `LGraph` instead.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4275-Feature-Request-Subgraphs-could-maintain-their-own-viewport-21e6d73d365081cfa813f9a3f17a7996) by [Unito](https://www.unito.io)

christian-byrne:
Saves and restores canvas viewport state (zoom/pan) when navigating between subgraphs using QuickLRU cache with 32-item limit.

Resolves https://github.com/Comfy-Org/ComfyUI_frontend/issues/4275.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4613-feat-Viewport-persistence-for-subgraph-navigation-2416d73d3650814f8d44e45247fe710b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/subgraphNavigationStore.ts
+++ b/src/stores/subgraphNavigationStore.ts
@@ -1,10 +1,13 @@
+import QuickLRU from '@alloc/quick-lru'
 import type { Subgraph } from '@comfyorg/litegraph'
+import type { DragAndScaleState } from '@comfyorg/litegraph/dist/DragAndScale'
 import { defineStore } from 'pinia'
 import { computed, shallowReactive, shallowRef, watch } from 'vue'
 
 import { app } from '@/scripts/app'
 import { isNonNullish } from '@/utils/typeGuardUtil'
 
+import { useCanvasStore } from './graphStore'
 import { useWorkflowStore } from './workflowStore'
 
 /**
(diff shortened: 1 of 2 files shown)


christian-byrne:
If following codebase convention, the more standard approach would be to use changeTracker. However, there is a general goal to make changeTracker track the data model only and not be conscious of the view.

PR #4624, opened 2025-07-31
[feat] Move partial execution to the backend and make work with subgraphs

### LINKED ISSUE #2046: [BUG] Incorrect Output Node Execution with "Execute Selected Output Nodes" Feature
**Python version**
3.12.10 (tags/v3.12.10:0cc8128, Apr  8 2025, 12:21:36) [MSC v.1943 64 bit (AMD64)]

**ComfyUI version**
0.3.40

**ComfyUI frontend version**
1.21.7

**Additional note about resource**
- Installing the requirements into the system-wide Python environment instead of within the venv

**Problem**
I’m using or have installed ComfyUI Manager, and I encountered an issue: when I click the 'Execute to selected output nodes' button on one of the output nodes (such as preview images, save images, etc.) among multiple output nodes, the execution flow goes wrong, it ends up executing a different output node instead

_Image 1_
Image

_Image 2_
Image

This can be seen in Image 1 (where I attempt to execute a specific flow), and in Image 2 (where execution incorrectly runs on a different KSampler). The two KSamplers are not connected linearly at all, this workflow was freshly created for testing, to confirm that the issue still occurs even in a new workflow

This behavior is completely normal when ComfyUI Manager is not installed. That means, before I installed ComfyUI Manager and after I deleted the ComfyUI Manager folder (from the custom nodes folder), the issue does not occur at all and everything works perfectly fine

**Additional note about problem**
- From what I've observed, the incorrect target it executes always points to the KSampler that I last added to the workflow
- This issue occurs when there are multiple KSamplers, at least 3 are needed for the problem to become noticeable. However, no matter how many output nodes are present, the issue does not happen at all if ComfyUI Manager is not installed

### LINKED ISSUE #4581: [Bug]: Execute button throws error when used inside subgraph
## Custom Node Testing
☑️ I have tried disabling custom nodes and the issue persists

## Frontend Version
1.25.2

## Expected Behavior
When selecting output nodes inside a subgraph and clicking the execute button in the selection toolbox, the partial execution should work correctly, executing only the selected nodes and their dependencies within the subgraph context.

## Actual Behavior
When using the execute button on selected nodes while inside a subgraph, it throws an error. The error toast displays:
"cannot read properties of undefined (reading 'inputs')"

## Steps to Reproduce
1. Create a workflow with a subgraph node
2. Enter the subgraph (double-click or right-click → "Enter Subgraph")
3. Add some nodes inside the subgraph, including at least one output node (e.g., SaveImage)
4. Select one or more output nodes inside the subgraph
5. Click the execute button (play icon) in the selection toolbox
6. Observe the error toast message

## Debug Logs
The error occurs in the browser, no terminal logs are generated for this issue.

## Setting JSON
Using default settings. The issue is reproducible with a fresh ComfyUI installation.

## What browsers do you use to access the UI?
- Google Chrome
- Mozilla Firefox

### Root Cause Analysis
The issue appears to be in the  function in . When executing from within a subgraph, the function tries to traverse node inputs but encounters undefined nodes because the node IDs in the subgraph context don't match those in the root graph's output object.

The partial execution system currently assumes all nodes are in the root graph context, but when inside a subgraph, nodes have different execution IDs that follow a hierarchical format (e.g., "parentId:subgraphId:nodeId").

### Related Code
-  -  function (line 28-44)
-  - UI component
-  -  command

This is a significant usability issue for the new subgraph feature, as users cannot use the convenient partial execution functionality when working inside subgraphs.

christian-byrne:
This PR updates the frontend to use the backend partial execution API introduced in https://github.com/comfyanonymous/ComfyUI/pull/9123.

https://github.com/user-attachments/assets/8bc9496b-8580-4b3d-a2d6-874b73a544dc

Fixes #4581
Fixes https://github.com/Comfy-Org/ComfyUI-Manager/issues/2046
Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/4581
Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/3987

### Previous Implementation of Partial Execution

1. UI (ExecuteButton.vue) - Filters selected nodes to only show button when output nodes selected
2. Command Handler - Filters selected nodes to only output nodes: `filter((node) => node.constructor.nodeData?.output_node)`
3. `app.queuePrompt` - Receives already-filtered output node IDs
4. `graphToPrompt` - Passes these IDs to recursiveAddNodes
5. `recrusiveAddNodes` - In single traversal, discovers the node set (finds transitive closure of "is-input-to" relation on output nodes) and builds the subgraph structure (induced subgraph). Then sends that subgraph as `prompt` (full workflow stil sent in `extra_data`)

### New Implementation of Partial Execution

Now, the frontend simply collects the execution IDs of all selected nodes (including all nodes within selected subgraphs) and passes them as a `partial_execution_targets` parameter, letting the backend handle the actual partial execution logic far more efficiently (excluding output nodes not in targets) and with access to cache info.

### Added Utilities

The implementation adds efficient graph traversal utilities using iterative depth-first search to collect execution IDs in the hierarchical format required for nested subgraph nodes (e.g., "123:456:789"). The generic traversal functions `traverseNodesDepthFirst` and `collectFromNodes` provide a reusable foundation for graph operations, while `getExecutionIdsForSelectedNodes` specifically handles the execution ID collection. This approach ensures proper handling of subgraphs where selecting a subgraph node executes all nodes within it recursively.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4624-feat-Move-partial-execution-to-the-backend-and-make-work-with-subgraphs-2416d73d365081989d2de336d16b7f69) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useCoreCommands.ts
+++ b/src/composables/useCoreCommands.ts
@@ -33,7 +33,7 @@ import {
   getAllNonIoNodesInSubgraph,
   getExecutionIdsForSelectedNodes
 } from '@/utils/graphTraversalUtil'
-import { getSelectedOutputNodes } from '@/utils/nodeFilterUtil'
+import { filterOutputNodes } from '@/utils/nodeFilterUtil'
 
 const moveSelectedNodesVersionAdded = '1.22.2'
 
@@ -368,7 +368,7 @@ export function useCoreCommands(): ComfyCommand[] {
       function: async () => {
         const batchCount = useQueueSettingsStore().batchCount
         const selectedNodes = getSelectedNodes()
-        const selectedOutputNodes = getSelectedOutputNodes(selectedNodes)
+        const selectedOutputNodes = filterOutputNodes(selectedNodes)
 
         if (selectedOutputNodes.length === 0) {
           toastStore.add({
--- a/src/composables/useCoreCommands.ts
+++ b/src/composables/useCoreCommands.ts
@@ -29,7 +29,10 @@ import { useBottomPanelStore } from '@/stores/workspace/bottomPanelStore'
 import { useColorPaletteStore } from '@/stores/workspace/colorPaletteStore'
 import { useSearchBoxStore } from '@/stores/workspace/searchBoxStore'
 import { useWorkspaceStore } from '@/stores/workspaceStore'
-import { getAllNonIoNodesInSubgraph } from '@/utils/graphTraversalUtil'
+import {
+  getAllNonIoNodesInSubgraph,
+  getExecutionIdsForSelectedNodes
+} from '@/utils/graphTraversalUtil'
 
 const moveSelectedNodesVersionAdded = '1.22.2'
 
@@ -363,10 +366,11 @@ export function useCoreCommands(): ComfyCommand[] {
       versionAdded: '1.19.6',
       function: async () => {
         const batchCount = useQueueSettingsStore().batchCount
-        const queueNodeIds = getSelectedNodes()
-          .filter((node) => node.constructor.nodeData?.output_node)
-          .map((node) => node.id)
-        if (queueNodeIds.length === 0) {
+        const selectedOutputNodes = getSelectedNodes().filter(
+          (node) => node.constructor.nodeData?.output_node
+        )
+
+        if (selectedOutputNodes.length === 0) {
           toastStore.add({
             severity: 'error',
             summary: t('toastMessages.nothingToQueue'),
@@ -375,7 +379,11 @@ export function useCoreCommands(): ComfyCommand[] {
           })
           return
         }
-        await app.queuePrompt(0, batchCount, queueNodeIds)
+
+        // Get execution IDs for all selected output nodes and their descendants
+        const executionIds =
+          getExecutionIdsForSelectedNodes(selectedOutputNodes)
+        await app.queuePrompt(0, batchCount, executionIds)
       }
     },
     {
--- a/src/composables/useCoreCommands.ts
+++ b/src/composables/useCoreCommands.ts
@@ -33,6 +33,7 @@ import {
   getAllNonIoNodesInSubgraph,
   getExecutionIdsForSelectedNodes
 } from '@/utils/graphTraversalUtil'
+import { getSelectedOutputNodes } from '@/utils/nodeFilterUtil'
 
 const moveSelectedNodesVersionAdded = '1.22.2'
 
@@ -366,9 +367,8 @@ export function useCoreCommands(): ComfyCommand[] {
       versionAdded: '1.19.6',
       function: async () => {
         const batchCount = useQueueSettingsStore().batchCount
-        const selectedOutputNodes = getSelectedNodes().filter(
-          (node) => node.constructor.nodeData?.output_node
-        )
+        const selectedNodes = getSelectedNodes()
+        const selectedOutputNodes = getSelectedOutputNodes(selectedNodes)
 
         if (selectedOutputNodes.length === 0) {
           toastStore.add({
--- a/src/scripts/api.ts
+++ b/src/scripts/api.ts
@@ -35,12 +35,13 @@ import type {
   NodeId
 } from '@/schemas/comfyWorkflowSchema'
 import type { ComfyNodeDef } from '@/schemas/nodeDefSchema'
+import type { NodeExecutionId } from '@/types/nodeIdentification'
 import { WorkflowTemplates } from '@/types/workflowTemplateTypes'
 
 interface QueuePromptRequestBody {
   client_id: string
   prompt: ComfyApiWorkflow
-  partial_execution_targets?: string[]
+  partial_execution_targets?: NodeExecutionId[]
   extra_data: {
     extra_pnginfo: {
       workflow: ComfyWorkflowJSON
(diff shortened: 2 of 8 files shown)


src/scripts/app.ts
```diff
@@ -1281,11 +1277,17 @@ export class ComfyApp {
             executeWidgetsCallback(subgraph.nodes, 'beforeQueued')
           }
 
-          const p = await this.graphToPrompt(this.graph, { queueNodeIds })
+          // Don't pass queueNodeIds to graphToPrompt anymore - let backend handle partial execution
```
christian-byrne: remove this comment

src/scripts/app.ts
```diff
@@ -1281,11 +1277,17 @@ export class ComfyApp {
...
+          const p = await this.graphToPrompt(this.graph)
           try {
             api.authToken = comfyOrgAuthToken
             api.apiKey = comfyOrgApiKey ?? undefined
-            const res = await api.queuePrompt(number, p)
+            // Pass queueNodeIds as partial_execution_targets to backend
```
christian-byrne: remove this comment

src/scripts/api.ts
```diff
@@ -611,17 +612,22 @@ export class ComfyApi extends EventTarget {
...
    * @param {number} number The index at which to queue the prompt, passing -1 will insert the prompt at the front of the queue
    * @param {object} prompt The prompt data to queue
+   * @param {string[]} partialExecutionTargets Optional list of node IDs to execute (partial execution)
    * @throws {PromptExecutionError} If the prompt fails to execute
    */
   async queuePrompt(
     number: number,
-    data: { output: ComfyApiWorkflow; workflow: ComfyWorkflowJSON }
+    data: { output: ComfyApiWorkflow; workflow: ComfyWorkflowJSON },
+    partialExecutionTargets?: string[]
```
christian-byrne: possible known issue: `queuePrompt` is heavily monkey-patched by custom nodes and some rudely only return the two parameters `number` and `data` when invoking original `queuePrompt`.

src/scripts/api.ts
```diff
@@ -611,17 +612,22 @@ export class ComfyApi extends EventTarget {
    * Queues a prompt to be executed
    * @param {number} number The index at which to queue the prompt, passing -1 will insert the prompt at the front of the queue
    * @param {object} prompt The prompt data to queue
+   * @param {string[]} partialExecutionTargets Optional list of node IDs to execute (partial execution)
```
christian-byrne: This comment should specify that they are execution IDs. Also, we should use the ExecutionId type defined in the nodeIdentification types file.

src/composables/useCoreCommands.ts
```diff
@@ -363,10 +366,11 @@ export function useCoreCommands(): ComfyCommand[] {
       versionAdded: '1.19.6',
       function: async () => {
         const batchCount = useQueueSettingsStore().batchCount
-        const queueNodeIds = getSelectedNodes()
-          .filter((node) => node.constructor.nodeData?.output_node)
-          .map((node) => node.id)
-        if (queueNodeIds.length === 0) {
+        const selectedOutputNodes = getSelectedNodes().filter(
+          (node) => node.constructor.nodeData?.output_node
+        )
```
christian-byrne: Consider refactoring this to a utility file and adding a unit test.

src/scripts/api.ts
christian-byrne: We should define an `options` object instead of continuing to add position arguments that are more difficult to read and maintain.

src/utils/graphTraversalUtil.ts
```diff
@@ -351,3 +351,105 @@ export function mapSubgraphNodes<T>(
...
+/**
+ * Collects execution IDs for selected nodes and all their descendants.
+ * Uses the generic DFS traversal with optimized string building.
+ *
+ * @param selectedNodes - The selected nodes to process
+ * @returns Array of execution IDs for selected nodes and all nodes within selected subgraphs
+ */
+export function getExecutionIdsForSelectedNodes(
+  selectedNodes: LGraphNode[]
+): string[] {
```
christian-byrne: Similar to above, use the ExecutionID from nodeIdentification types file.

tests-ui/tests/utils/graphTraversalUtil.test.ts
christian-byrne, line 1: Consider adding a few more edge cases to the unit tests added for graphTraversalUtils.

src/utils/graphTraversalUtil.ts
```diff
@@ -351,3 +351,105 @@ export function mapSubgraphNodes<T>(
...
+  while (stack.length > 0) {
+    const { node, context } = stack.pop()!
+
+    // Visit node and get updated context for children
+    const childContext = visitor(node, context)
+
+    // If it's a subgraph and we should expand, add children to stack
+    if (expandSubgraphs && node.isSubgraphNode?.() && node.subgraph) {
+      // Add in reverse order to maintain left-to-right processing
+      const children = Array.from(node.subgraph.nodes).reverse()
```
christian-byrne: reversing the array on every step down is definitely sub-optimal

PR #4639, opened 2025-08-01
[fix] Detect missing nodes in subgraphs

### LINKED ISSUE #4618: [Bug] When there is a missing node inside the subgraph, there will be no missing node promo
https://github.com/user-attachments/assets/237b7d99-ae9b-439c-8f03-77ab5cccf0c8

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4618-Bug-When-there-is-a-missing-node-in-the-subgraph-there-will-be-no-missing-node-prom-2416d73d36508103beeac6c614e1f6ea) by [Unito](https://www.unito.io)

viva-jinyi:
## Summary
This PR fixes an issue where missing nodes inside subgraphs were not being detected and reported to users. The missing node warning dialog would only appear for missing nodes at the top level of the workflow, but not for nodes within subgraphs.

## Changes
- Modified `loadGraphData` in `app.ts` to recursively check nodes in subgraphs for missing node types
- Added context to missing node error messages showing the subgraph location (e.g., "MISSING_NODE (in subgraph 'MySubgraph')")
- Created test workflow `missing_nodes_in_subgraph.json` with a missing node inside a subgraph
- Added Playwright test to verify the warning appears for missing nodes in subgraphs

## Testing
- Created new Playwright test that verifies missing nodes in subgraphs are detected
- All existing tests pass
- Manual testing confirms the warning dialog now shows missing nodes from subgraphs with proper context

## Related Issues
Fixes #4618

## Screenshots
[screen-capture.webm](https://github.com/user-attachments/assets/494ae0d4-edd9-4f36-b2c0-03612a6490ba)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4639-fix-Detect-missing-nodes-in-subgraphs-2426d73d36508160ac5ce8cb558fcb01) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -1092,23 +1092,42 @@ export class ComfyApp {
 
     const embeddedModels: ModelFile[] = []
 
-    for (let n of graphData.nodes) {
-      // Patch T2IAdapterLoader to ControlNetLoader since they are the same node now
-      if (n.type == 'T2IAdapterLoader') n.type = 'ControlNetLoader'
-      if (n.type == 'ConditioningAverage ') n.type = 'ConditioningAverage' //typo fix
-      if (n.type == 'SDV_img2vid_Conditioning')
-        n.type = 'SVD_img2vid_Conditioning' //typo fix
+    // Helper function to process nodes including those in subgraphs
+    const processNodesRecursively = (nodes: any[], path: string = '') => {
+      for (let n of nodes) {
+        // Patch T2IAdapterLoader to ControlNetLoader since they are the same node now
+        if (n.type == 'T2IAdapterLoader') n.type = 'ControlNetLoader'
+        if (n.type == 'ConditioningAverage ') n.type = 'ConditioningAverage' //typo fix
+        if (n.type == 'SDV_img2vid_Conditioning')
+          n.type = 'SVD_img2vid_Conditioning' //typo fix
 
-      // Find missing node types
-      if (!(n.type in LiteGraph.registered_node_types)) {
-        missingNodeTypes.push(n.type)
-        n.type = sanitizeNodeName(n.type)
+        // Find missing node types
+        if (!(n.type in LiteGraph.registered_node_types)) {
+          // Include context about subgraph location if applicable
+          const nodeTypeWithContext = path
+            ? `${n.type} (in subgraph '${path}')`
+            : n.type
+          missingNodeTypes.push(nodeTypeWithContext)
+          n.type = sanitizeNodeName(n.type)
+        }
+
+        // Collect models metadata from node
+        const selectedModels = getSelectedModelsMetadata(n)
+        if (selectedModels?.length) {
+          embeddedModels.push(...selectedModels)
+        }
       }
+    }
 
-      // Collect models metadata from node
-      const selectedModels = getSelectedModelsMetadata(n)
-      if (selectedModels?.length) {
-        embeddedModels.push(...selectedModels)
+    // Process nodes at the top level
+    processNodesRecursively(graphData.nodes)
+
+    // Process nodes in subgraphs
+    if (graphData.definitions?.subgraphs) {
+      for (const subgraph of graphData.definitions.subgraphs as any[]) {
+        if (subgraph.nodes) {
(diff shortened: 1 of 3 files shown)


src/scripts/app.ts
```diff
@@ -1092,23 +1092,42 @@ export class ComfyApp {
...
+          n.type = 'SVD_img2vid_Conditioning' //typo fix
+
+        // Find missing node types
+        if (!(n.type in LiteGraph.registered_node_types)) {
+          // Include context about subgraph location if applicable
+          const nodeTypeWithContext = path
+            ? `${n.type} (in subgraph '${path}')`
+            : n.type
+          missingNodeTypes.push(nodeTypeWithContext)
+          n.type = sanitizeNodeName(n.type)
```
christian-byrne: For the missingNodeTypes, it is a union type:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/47e18088611f905588deb6912524ea233daab491/src/types/comfy.ts#L33-L43

It might be better to not modify the `type` and instead use the secondary format like

```ts
missingNodeTypes.push({
  type: n.type
  hint: nodeTypeWithContext
})
```

or something similar.

The type of missingNodeTypes is defined in the public API for extension lifecycle hook `beforeConfigureGraphData` so there is an added obligation to adhere to it.

src/scripts/app.ts
```diff
@@ -1092,23 +1092,42 @@ export class ComfyApp {
...
+    processNodesRecursively(graphData.nodes)
 
-      // Collect models metadata from node
-      const selectedModels = getSelectedModelsMetadata(n)
-      if (selectedModels?.length) {
-        embeddedModels.push(...selectedModels)
+    // Process nodes in subgraphs
+    if (graphData.definitions?.subgraphs) {
+      for (const subgraph of graphData.definitions.subgraphs as any[]) {
+        if (subgraph.nodes) {
```
christian-byrne: it seems like `nodes` is not defined on `graphData.definitions.subgraphs` in the schema https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/schemas/comfyWorkflowSchema.ts

But in the actual serialized workflows, the `nodes` property exists. I think this might be a mistake in the schema.

christian-byrne: Looking into it now

christian-byrne: I found the issue: https://github.com/Comfy-Org/ComfyUI_frontend/issues/4650.

src/scripts/app.ts
viva-jinyi: Thank you for the review! I've updated the code to use the proper MissingNodeType format as you suggested.

src/scripts/app.ts
viva-jinyi: Removed the as any[] cast - replaced with a defensive type guard 'nodes' in subgraph &&  Array.isArray(subgraph.nodes) to ensure type safety

christian-byrne: I am adding a typeguard to use so we don't need to cast to `any`: https://github.com/Comfy-Org/ComfyUI_frontend/pull/4651

src/scripts/app.ts
```diff
@@ -1092,23 +1092,47 @@ export class ComfyApp {
...
 
-      // Collect models metadata from node
-      const selectedModels = getSelectedModelsMetadata(n)
-      if (selectedModels?.length) {
-        embeddedModels.push(...selectedModels)
+    // Process nodes at the top level
+    processNodesRecursively(graphData.nodes)
+
+    // Process nodes in subgraphs
+    if (graphData.definitions?.subgraphs) {
```
christian-byrne, line 1133: Can you use this typeguard so that we don't need to use `any` in line 1096?

https://github.com/Comfy-Org/ComfyUI_frontend/blob/61c9341450f56f892cc7673b33de8a1fb799df3e/src/schemas/comfyWorkflowSchema.ts#L463-L478

This will cast to the correct type. It is necessary because of zod `lazy` on circular subgraph schema.

christian-byrne: I'll take care of this so we can release subgraphs quickly

christian-byrne (approved):
LGTM thanks for adding browser test

Issue #4604, opened 2025-07-30
Use transform to position floating selection toolbox


christian-byrne:
When the user zooms or pans, the [`SelectionToolbox`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/components/graph/SelectionOverlay.vue) updates its position using layout. Naturally, this causes quite a lot of reflow while dragging/panning/zooming. Although it's true that those updates are batched in RAF and there is not necessarily thrashing occurring, these reflows would actually not be happening at all if not for the `SelectionToolbox` (If you test our app currently (with no custom extensions that add floating elements), you will see 0 laytouts/sec while panning/zooming). In general, it's bad to rely on cached layout and having just one element recalc layout can cause cascading reflow effect.

If we change the element to be positioned via `transform`, it will skip layout entirely and prevent any cascading. It will also be considerably more performant and feel smoother.

In terms of implementation:

- There is a utility for syncing with the Canvas's transform state: https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/composables/canvas/useCanvasTransformSync.ts, example usage in Minimap. We can simply add the initial position offset from when the element was mounted/shown.
- Use the `afterTransformSettled` from Vue Nodes branch to force a singular reflow after panning/zooming has stopped for some ~200ms to fix pixel stretch.

christian-byrne:
Great task for a new team member to learn about `transform` and syncing it across components + browser rendering pipeline in general.

Helpful links:

- https://webperf.tips/tip/browser-rendering-pipeline/
- https://webperf.tips/tip/layers-and-compositing/
- https://webperf.tips/tip/layout-thrashing/
- https://webperf.tips/tip/animate-on-compositor-thread/

christian-byrne:
@simula-r WDYT?

simula-r:
@christian-byrne I think this is definitely worth optimizing. Looking closer, there
   are a few surface areas I want to understand first.

  I'll come back with a more detailed plan once I've gotten familiar with how the
  selection system works.

simula-r:
Actually, decided to go further and found a clean solution that achieves 0 layouts/sec by rendering the selection border directly on canvas instead of using DOM elements. 

Full explanation and implementation: #5158

PR #4641, opened 2025-08-01
Fix performance issue with selection overlay updates during canvas interactions

Jackpkn:
The positionSelectionOverlay function was being called every frame during
canvas interactions (dragging, zooming, panning), causing excessive DOM
updates and performance degradation.

This change implements requestAnimationFrame-based throttling to limit
updates to browser refresh rate (max 60fps), dramatically reducing CPU
usage while maintaining visual correctness.

- Add throttledPositionUpdate function with RAF scheduling
- Replace direct canvas.ds.onChanged assignment with throttled version
- Use throttled function in drag handling for consistency
- Add proper cleanup with cancelAnimationFrame on component unmount
- Maintain all existing functionality and visual behavior

Fixes #4631

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4641-Fix-performance-issue-with-selection-overlay-updates-during-canvas-interactions-2426d73d3650812e843ac928a1aa8426) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/SelectionOverlay.vue
+++ b/src/components/graph/SelectionOverlay.vue
@@ -15,7 +15,7 @@
 <script setup lang="ts">
 import { createBounds } from '@comfyorg/litegraph'
 import { whenever } from '@vueuse/core'
-import { ref, watch } from 'vue'
+import { onUnmounted, ref, watch } from 'vue'
 
 import { useSelectedLiteGraphItems } from '@/composables/canvas/useSelectedLiteGraphItems'
 import { useAbsolutePosition } from '@/composables/element/useAbsolutePosition'
@@ -39,12 +39,24 @@ const positionSelectionOverlay = () => {
 
   visible.value = true
   const bounds = createBounds(selectableItems)
+
   if (bounds) {
     updatePosition({
       pos: [bounds[0], bounds[1]],
       size: [bounds[2], bounds[3]]
     })
   }
+}
+
+let rafId: number | null = null
+
+const throttledPositionUpdate = () => {
+  if (rafId !== null) return // Already scheduled - skip redundant calls
+
+  rafId = requestAnimationFrame(() => {
+    positionSelectionOverlay()
+    rafId = null
+  })
 }
 
 whenever(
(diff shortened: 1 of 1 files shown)


huchenlei (commented):
There are existing VueUse hooks that could be used for this throttling:

## `useRafFn` - Perfect Match
This is the most appropriate hook for this use case. It's specifically designed for `requestAnimationFrame`-based operations with pause/resume controls, which matches exactly what this PR is implementing.

```typescript
import { useRafFn } from '@vueuse/core'

const { pause, resume } = useRafFn(() => {
  positionSelectionOverlay()
})
```

## `useThrottleFn` - Alternative Option
This provides time-based throttling (default 200ms) which could work but isn't as optimal as RAF-based throttling for canvas operations:

```typescript
import { useThrottleFn } from '@vueuse/core'

const throttledPosition = useThrottleFn(() => {
  positionSelectionOverlay()
}, 16) // ~60fps
```

**Recommendation**: Use `useRafFn` since this PR specifically implements RAF-based throttling for canvas performance. It provides the same functionality with built-in pause/resume controls and proper cleanup.

christian-byrne (commented):
This PR implements RAF throttling to reduce selection overlay updates to 60fps, which will improve performance. However, issue #4604 outlines a more effective solution: using transform-based positioning to eliminate layout operations entirely. 

The current approach still triggers layout/reflow on each update, while transform-based positioning would run on the compositor thread and avoid layout thrashing completely (see [layout thrashing](https://webperf.tips/tip/layout-thrashing/) and [compositor thread animation](https://webperf.tips/tip/animate-on-compositor-thread/)). 

**Suggestion**: Consider using the existing `useAbsolutePosition({ useTransform: true })` option with `useCanvasTransformSync` (similar to Minimap implementation) for zero-layout performance during canvas interactions.

The RAF solution is a solid improvement, but the transform approach from #4604 would provide even smoother performance by eliminating reflows entirely.

**Minor**: Add defensive checks around `canvas.ds` access and consider storing/restoring the original `onChanged` handler in cleanup.

PR #4679, opened 2025-08-04
minimap improve

### LINKED ISSUE #4548: [Feature Request]: Improve minimap internal layout

- **Hide wires** 
The wires are too prevalent in the minimap and it's a bit distracting. 
The shape of the wire it's a bit different, so it does not help with visual recognition.   
Minimaps in similar programs also don't show the wires. 

- **Show groups**
Normally, large workflows, ( the ones which need the minimap ), have groups. 
And in this case, the "shape" of the workflow it's mostly dictated by groups.
These large workflows are mostly recognizable by the group's layout.

- **User option to add color**
I agree that if you use too many colors, they won't look good in a small size, and it would be distracting.
But if you use colors well, they could be useful in the mini map. 
I rarely use color, but when I do, it has a meaning, so it would be easily recognizable in the minimap.

### Proposed workflow

.

### Additional information

.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4548-Feature-Request-Improve-minimap-internal-layout-23d6d73d365081c4add3c2bdad150743) by [Unito](https://www.unito.io)

jtydhr88:
resolve https://github.com/Comfy-Org/ComfyUI_frontend/issues/4548
key changes:
1. add options to allow user customize minimap style:
- node code
- show group
- show line
- bypass state
- error state

https://github.com/user-attachments/assets/c9efcfa4-d19e-43fa-a28a-0d58d63c452a

┆Issue is synchronized with this
 [Notion page](https://www.notion.so/PR-4679-minimap-improve-2456d73d365081f6a83dc4d8651b843d) by [Unito](https://www.unito.io)


src/composables/useMinimap.ts
```diff
@@ -203,8 +223,20 @@ export function useMinimap() {
       const w = node.size[0] * scale.value
       const h = node.size[1] * scale.value
 
+      let color = nodeColor.value
+
+      if (node.color) {
+        color = adjustColor(node.color, { lightness: 0.1 })
+      }
```
christian-byrne: What's the reasoning for this?

src/composables/useMinimap.ts
```diff
@@ -203,8 +223,20 @@ export function useMinimap() {
...
+
+      if (node.mode === LGraphEventMode.BYPASS) {
+        color = isLightTheme.value ? '#DBDBDB' : '#441B44'
+      } else if (isLightTheme.value) {
+        color = adjustColor(color, { lightness: 0.5 })
+      }
```
christian-byrne: We can also check for `node.has_errors` and draw the red outline?

https://github.com/search?q=repo%3AComfy-Org%2Flitegraph.js%20has_errors&type=code

christian-byrne:
Claude won't review until the tests pass.

src/components/graph/MiniMapPanel.vue
```diff
@@ -0,0 +1,95 @@
...
+  panelStyles: any
+  nodeColors: boolean
+  showLinks: boolean
+  showGroups: boolean
+  renderBypass: boolean
+  renderError: boolean
+}>()
+
+defineEmits<{
+  updateOption: [key: any, value: boolean]
```
christian-byrne: We should type these and not use `any`

src/composables/useMinimap.ts
```diff
@@ -27,6 +28,27 @@ export function useMinimap() {
...
+  )
+  const renderBypass = computed(() =>
+    settingStore.get('Comfy.Minimap.RenderBypassState')
+  )
+  const renderError = computed(() =>
+    settingStore.get('Comfy.Minimap.RenderErrorState')
+  )
+
+  const updateOption = async (key: any, value: boolean) => {
+    await settingStore.set(key, value)
```
christian-byrne: Can use same type instead of `any`

christian-byrne (changes requested):

christian-byrne (approved):
LGTM!

Issue #1193, opened 2024-10-09
[Command] Make convert selected node to group node functionality a command


huchenlei:
![image](https://github.com/user-attachments/assets/5466d390-7d99-49c7-ae9a-57c6f8a365c2)

PR #4688, opened 2025-08-05
Fix inconsistent subgraphNode usage

AustinMroz:
Prior to this commit, subgraphNode inconsistently refers to either the parent graph, or to indicate the current node is a subgraph.

This corrects the usage of subgraphNode to consistently refer to the subgraph instance as defined in the constructor.

This requires the corresponding comfy-org/litegraph.js#1193 commit to function, and solves a bug where graph serialization fails due to an incorrectly reported infinite loop.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4688-Fix-inconsistent-subgraphNode-usage-2466d73d3650812399d5ffc1d7b4f7cc) by [Unito](https://www.unito.io)

### DIFF
--- a/src/utils/executionUtil.ts
+++ b/src/utils/executionUtil.ts
@@ -3,11 +3,7 @@ import type {
   ExecutionId,
   LGraph
 } from '@comfyorg/litegraph'
-import {
-  ExecutableNodeDTO,
-  LGraphEventMode,
-  SubgraphNode
-} from '@comfyorg/litegraph'
+import { ExecutableNodeDTO, LGraphEventMode } from '@comfyorg/litegraph'
 
 import type {
   ComfyApiWorkflow,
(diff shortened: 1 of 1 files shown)


christian-byrne:
Changing `find` to `some` per ESlint suggestion.

christian-byrne (approved):
This PR correctly removes the incorrect 4th parameter being passed to ExecutableNodeDTO constructor.

The removed code was passing 'node instanceof SubgraphNode ? node : undefined' as the subgraphNode parameter, which is semantically incorrect. The subgraphNode parameter is meant to reference the parent SubgraphNode that contains the current node, not the node itself when it happens to be a SubgraphNode.

This confusion stems from the ambiguous naming where 'subgraphNode' could be interpreted as either 'the parent subgraph containing this node' (correct) or 'this node which is a subgraph' (incorrect). The frontend code was using the latter interpretation.

By removing this parameter and letting it default to undefined, the code now correctly indicates that nodes at the root level of the graph are not contained within any subgraph. The ExecutableNodeDTO constructor's 4th parameter should only be set when creating DTOs for nodes that are actually inside a subgraph, which would be done during the recursive flattening process in getInnerNodes().

This fix works in conjunction with the litegraph PR #1193 to resolve the serialization infinite loop bug. Both PRs address different manifestations of the same conceptual confusion about what subgraphNode represents.

PR #4797, opened 2025-08-07
[docs] Add feature flags documentation

christian-byrne:
## Summary
- Added comprehensive documentation for the feature flags system in `docs/FEATURE_FLAGS.md`

## Details
This documentation explains:
- How the bidirectional feature flag negotiation works between frontend and backend
- The WebSocket handshake flow with mermaid diagrams
- Component architecture and implementation details
- Current usage examples (preview metadata, max upload size)
- Naming conventions and structure for feature flags
- How to add new feature flags in both frontend and backend

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4797-docs-Add-feature-flags-documentation-2486d73d3650819c9d30c9effe2d2b13) by [Unito](https://www.unito.io)

### DIFF
--- a/docs/FEATURE_FLAGS.md
+++ b/docs/FEATURE_FLAGS.md
@@ -60,46 +60,28 @@ graph TB
 
 ## Feature Flag Structure
 
-Feature flags are organized as nested dictionaries with package/namespace as the top-level key:
-
-```mermaid
-graph LR
-    A[Feature Flags] --> B[Core Features]
-    A --> C[Extension Features]
-    A --> D[Package Features]
-    
-    B --> E["core"]
-    C --> F["extension/manager"]
-    D --> G["package/custom-nodes"]
-    
-    E --> H["{ async_execution: true, version: '0.2.0' }"]
-    F --> I["{ supports_v4: true, version: '3.0.0' }"]
-    G --> J["{ custom_capability: true }"]
-```
+Feature flags are organized as a flat dictionary at the top level, with extensions nested under an `extension` object:
 
 ### Naming Convention
 
-- **Core features**: Use `"core"` namespace
-- **Extensions**: Use `"extension/{name}"` format
-- **Packages**: Use `"package/{name}"` format
+- **Core features**: Top-level keys (e.g., `"async_execution"`, `"supports_batch_queue"`)
+- **Client features**: Top-level keys (e.g., `"supports_preview_metadata"`)
+- **Extensions**: Nested under `"extension"` object (e.g., `extension.manager`)
 
 ### Structure Example
 
 ```json
 {
-  "core": {
-    "async_execution": true,
-    "version": "0.2.0",
-    "supports_batch_queue": false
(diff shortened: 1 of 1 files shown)


docs/FEATURE_FLAGS.md
```diff
@@ -0,0 +1,376 @@
...
+    "supports_batch_queue": false
+  },
+  "extension/manager": {
+    "supports_v4": true,
+    "version": "3.0.0",
+    "supports_ai_search": false
+  },
+  "client": {
+    "supports_preview_metadata": true,
+    "supports_websocket_v2": false
```
christian-byrne: This should be changed. Instead of `version` flags, we will aim to make the flags communicate functionality which may or may not be associated with a specific version.

Also, the `core` and `client` ones shuold just be at the top level, as that pattern has already been established.

The `manager` should be a sub-object inside of the `extension` object.

docs/FEATURE_FLAGS.md
```diff
@@ -0,0 +1,376 @@
...
+    A --> D[Package Features]
+    
+    B --> E["core"]
+    C --> F["extension/manager"]
+    D --> G["package/custom-nodes"]
+    
+    E --> H["{ async_execution: true, version: '0.2.0' }"]
+    F --> I["{ supports_v4: true, version: '3.0.0' }"]
+    G --> J["{ custom_capability: true }"]
+```
```
christian-byrne: We don't need a diagram for this.

docs/FEATURE_FLAGS.md
```diff
@@ -0,0 +1,376 @@
...
+
+### Naming Convention
+
+- **Core features**: Use `"core"` namespace
```
christian-byrne: Not true, see comment below.

docs/FEATURE_FLAGS.md
```diff
@@ -0,0 +1,376 @@
...
+    "max_upload_size": args.max_upload_size * 1024 * 1024,  # Convert MB to bytes
+}
+```
+
+**Frontend Usage:**
+```typescript
+const { getServerFeature } = useFeatureFlags()
+const maxUploadSize = getServerFeature('max_upload_size', 100 * 1024 * 1024) // Default 100MB
+```
+
```
christian-byrne, line 212: This can be presented as "Examples" that are also currently used. We don't want to frame this as the "Current Feature Flags" because then we have to continually update this doc.

docs/FEATURE_FLAGS.md
```diff
@@ -0,0 +1,376 @@
...
+// Check feature support
+if (serverSupports('supports_preview_metadata')) {
+    // Use enhanced previews
+}
+
+// Use convenience properties
+if (manager.supportsV4.value) {
+    // Use V4 manager API
+}
+```
```
christian-byrne: For number 2, should clarify that the composable makes the flags reactive.

PR #4798, opened 2025-08-07
Fix widget display on subgraph nodes

### LINKED ISSUE #4692: [Bug] When creating a nested subgraph inside subgraph, the widgets will switch to inputs.
FE version 1.24.3

Image

1. The node's widgets inside the subgraph are already linked to input slots to expose them to the first level.
2. When converting them to nested subgraph, it seems not worked right

Image

## Full screen recording

https://github.com/user-attachments/assets/90690e8d-ab69-424e-b38c-f5d2394a50d6

In this screen recording, I have demonstrated two parts.
1. How to reproduce the same issues (before 00:45)
2. How I expect it to work (after 00:45)

AustinMroz:
Resolves #4692

This adds a workaround for widget data being clobbered when converting to subgraph and adds a path for widget data to still bubble upwards when a graph is loaded before dependent graphs with connected widgets.

Additionally, it recalculates input positions so that links are correctly displayed after conversion.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4798-Fix-widget-display-on-subgraph-nodes-2486d73d3650817b8414ff45456f9cba) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraph.ts
+++ b/src/lib/litegraph/src/LGraph.ts
@@ -1554,6 +1554,9 @@ export class LGraph
       outputs: structuredClone(outputs)
     })
     if (!subgraphNode) throw new Error('Failed to create subgraph node')
+    for (let i = 0; i < inputs.length; i++) {
+      Object.assign(subgraphNode.inputs[i], inputs[i])
+    }
 
     // Resize to inputs/outputs
     subgraphNode.setSize(subgraphNode.computeSize())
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):
LGTM

christian-byrne:
Don't worry about doing the cherry pick. Unlikely we will release 1.24 patch version before 1.25. Cherry pick definitely won't work across litegraph-merge commits.

PR #4800, opened 2025-08-07
Fix disconnection from subgraph inputs

AustinMroz:
Fixes the inability to disconnect a link from a subgraph input node by dragging it off of the target side and dropping it onto the canvas.

Making ToInputFromIoNodeLink inherit from MovingLinkBase was considered, but the latter has quite a few other required functions and implementations that are undesirable.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4800-Fix-disconnection-from-subgraph-inputs-2486d73d365081a4bb69ce6de7927dac) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/canvas/LinkConnector.ts
+++ b/src/lib/litegraph/src/canvas/LinkConnector.ts
@@ -795,7 +795,10 @@ export class LinkConnector {
    */
   disconnectLinks(): void {
     for (const link of this.renderLinks) {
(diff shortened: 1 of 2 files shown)


benceruleanlu (approved):

benceruleanlu:
lint-and-format action seems bonked

probably because it's a fork, but it's blocking

everything else is green so I'll bypass

PR #4826, opened 2025-08-07
Fix TransformPane pos/size

christian-byrne:
Was being offset causing the culling to be miscalculated. Doesn't cause reflow.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4826-Fix-TransformPane-pos-size-2486d73d365081909e84ceaeee83e359) by [Unito](https://www.unito.io)


christian-byrne:
The performance tests fail expectedly in CI.

The playwright tests also naturally fail.

Issue #4816, opened 2025-08-07
[Bug] Desktop user guide appears in non-desktop build


huchenlei:
<img width="669" height="451" alt="Image" src="https://github.com/user-attachments/assets/0461bff9-12fa-4661-8849-dc03080c522a" />

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4816-Bug-Desktop-user-guide-appears-in-non-desktop-build-2486d73d3650810b9bded55b405dfa1f) by [Unito](https://www.unito.io)

PR #4837, opened 2025-08-08
fix: hide More menu when no submenu items are visible

vivekchavan14:
- Added visible: isElectron() to desktop-guide submenu item
- Added hasVisibleMoreItems computed property to check submenu visibility
- Added visible property to More menu item that depends on hasVisibleMoreItems
- This prevents empty More menu from appearing in web builds where all submenu items are electron-only

Desktop:
image

web:
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4837-fix-hide-More-menu-when-no-submenu-items-are-visible-2496d73d365081dc9d41fccb4a2531bf) by [Unito](https://www.unito.io)

follow up fix #https://github.com/Comfy-Org/ComfyUI_frontend/issues/4816

### DIFF
--- a/src/components/helpcenter/HelpCenterMenuContent.vue
+++ b/src/components/helpcenter/HelpCenterMenuContent.vue
@@ -188,17 +188,8 @@ const showVersionUpdates = computed(() =>
   settingStore.get('Comfy.Notification.ShowVersionUpdates')
 )
 
-const moreMenuItem = computed(() =>
-  menuItems.value.find((item) => item.key === 'more')
-)
-
-const hasVisibleMoreItems = computed(() => {
-  // Check if there are any visible items in the more submenu
-  return isElectron() // Since all moreItems depend on isElectron()
-})
-
-const menuItems = computed<MenuItem[]>(() => {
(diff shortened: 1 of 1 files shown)


src/components/helpcenter/HelpCenterMenuContent.vue
```diff
@@ -192,12 +192,18 @@ const moreMenuItem = computed(() =>
   menuItems.value.find((item) => item.key === 'more')
 )
 
+const hasVisibleMoreItems = computed(() => {
+  // Check if there are any visible items in the more submenu
+  return isElectron() // Since all moreItems depend on isElectron()
```
christian-byrne: What about moving `moreItems` to a separate computed, then `hasVisibleMoreItems` can just be `!!moreItems.value.length`

vivekchavan14:
That's a good suggestion, we can do that

vivekchavan14:
@christian-byrne pushed updates according to your suggestion

christian-byrne (approved):
LGTM!

PR #4840, opened 2025-08-08
Implement subgraph unpacking

### LINKED ISSUE #4866: Don't show "Convert to Subgraph" button when selection is empty group or single subgraph node
Don't include the "Convert to Subgraph" item in the selection toolbox under conditions where it doesn't make sense, such as:

- The selection only includes groups (no other types of selectable items)
- The selection only includes a single node and it is a `SubgraphNode`
- Potentially others?

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4866-Don-t-show-Convert-to-Subgraph-button-when-selection-is-empty-group-or-single-subgra-24a6d73d3650815a8c8feb8cc7af5234) by [Unito](https://www.unito.io)

AustinMroz:
unpacking

Known Issues:
- [x] ~~Reroutes on links connecting to the subgraph node are left dangling~~
- [x] ~~Widget related errors when destorying a subgraph node nested inside another subgraph.~~

Resolves #4866 by changing the convert to subgraph button to an unpack button

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4840-Implement-subgraph-unpacking-2496d73d365081668119e83072e8211f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/selectionToolbox/ConvertToSubgraphButton.vue
+++ b/src/components/graph/selectionToolbox/ConvertToSubgraphButton.vue
@@ -1,6 +1,20 @@
 <template>
   <Button
-    v-show="isVisible"
+    v-if="isUnpackVisible"
+    v-tooltip.top="{
+      value: t('commands.Comfy_Graph_UnpackSubgraph.label'),
+      showDelay: 1000
+    }"
+    severity="secondary"
+    text
+    @click="() => commandStore.execute('Comfy.Graph.UnpackSubgraph')"
+  >
+    <template #icon>
+      <i-lucide:expand />
+    </template>
+  </Button>
+  <Button
+    v-else-if="isConvertVisible"
     v-tooltip.top="{
       value: t('commands.Comfy_Graph_ConvertToSubgraph.label'),
       showDelay: 1000
@@ -20,6 +34,7 @@ import Button from 'primevue/button'
 import { computed } from 'vue'
 import { useI18n } from 'vue-i18n'
 
+import { SubgraphNode } from '@/lib/litegraph/src/litegraph'
 import { useCommandStore } from '@/stores/commandStore'
 import { useCanvasStore } from '@/stores/graphStore'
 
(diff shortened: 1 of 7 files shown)


JorgeR81:
> Re-arrange nodes in parent graph to make room for unpacked nodes.

This would be nice, but it may be hard to make it work correctly every time. 
We could have too many nodes close together, in a large workflow.
Or they could be inside groups, and these would need to be re-scaled too.
Or we may just not like the new automatic node placement. 

Could it be possible to show some sort of preview ( e.g. translucid blue ) of the unpacked nodes ?
This would allow the user to make room manually, before unpacking.

AustinMroz:
I agree on both points. My original thought for "making room" was to always push existing parent graph nodes left or  right and ignore height. It would be comparatively simple to implement, but rather disruptive. I had also failed to consider how to handle groups (the cosmetic colored background rectangles). I now lean towards not including functionality for making room.

Selecting unpacked nodes has been implemented and at least gives the option of dragging the new nodes to a new corner of the graph for manually arrange later. Similarly, you can Ctrl+z undo the unpack operation if it winds up being too much of a mess.

I had considered and like the idea of previewing where nodes will unpack to. I don't currently have a solid grasp of how to implement it in a performant or intuitive manner. Maybe as a future PR.

Thanks for the feedback!

JorgeR81:
> I had considered and like the idea of previewing where nodes will unpack to. I don't currently have a solid grasp of how to implement it in a performant or intuitive manner. Maybe as a future PR.

Some thoughts for when you have time to work on this.

- Maybe it's not necessary to show the actual nodes, just a single rectangular outline, of the space they would take. 

- The outline could show up when we mouseover the right-click option to unpack.

christian-byrne:
Small conflict because I also added a command in a separate PR.

AustinMroz:
Looks like github really didn't like the merge direction.  I can reverse the direction and force-push if need be.

christian-byrne:
Could you rebase? If it causes too many issues then it's not too big of a deal.

christian-byrne (approved):
LGTM, very nice!

PR #4845, opened 2025-08-08
[fix] Adjust performance test timing constraints for CI environments

christian-byrne:
Performance tests were failing in CI due to slower execution environments. This PR adds CI detection using `process.env.CI` and applies a 3x multiplier to all timing constraints when running in CI, while maintaining strict timing requirements for local development.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4845-fix-Adjust-performance-test-timing-constraints-for-CI-environments-2496d73d365081309dc5c14ebb7e13a9) by [Unito](https://www.unito.io)


christian-byrne:
Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/4843

PR #4864, opened 2025-08-09
Fix execution breaks on multi/any-type slots

webfiltered:
### Current

Bypassing a node with inputs that are not an identical match on the bypassed slot will break execution.  This includes:
- `*` ("any" types)
- `INT,FLOAT` (multi-types)
- `Int` vs `INT`
- `0` vs `"0"`

If the destination input is lazy-loading, this always fails - even if the input is optional.

Using core nodes (`PreviewAny` is `*` type):

image

### Proposed

- Add a short-circuit for any types
- More explicit short-circuit for same-index matches (e.g. the third input & output)
- Preserve existing behaviour - iterate over all remaining slots, checking exact match
- Add a fall back to the standard `isValidConnection` check (only do this if we would otherwise fail)

A more complex example, with a switch, now working:

{3400E1B5-E9AB-492A-9FB8-106B4305F33B}

### Caveat

In some _extremely specific_ circumstances using custom nodes, this PR could alter workflow output.  The new behaviour is, however, intended.

An example would be a node with one IMAGE input, and two outputs.  First output is `IMAGE,MASK`, second output is `IMAGE`.  Previously, the first output would be `NoneType` when bypassed.  Now, both outputs will bypass to the `IMAGE` input.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4864-Fix-execution-breaks-on-multi-any-type-slots-24a6d73d36508169ac9bcf4e93b47cd8) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/subgraph/ExecutableNodeDTO.ts
+++ b/src/lib/litegraph/src/subgraph/ExecutableNodeDTO.ts
@@ -9,7 +9,7 @@ import type {
   CallbackReturn,
   ISlotType
 } from '@/lib/litegraph/src/interfaces'
-import { LGraphEventMode } from '@/lib/litegraph/src/litegraph'
+import { LGraphEventMode, LiteGraph } from '@/lib/litegraph/src/litegraph'
 
 import { Subgraph } from './Subgraph'
 import type { SubgraphNode } from './SubgraphNode'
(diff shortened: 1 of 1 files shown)


guill (approved):
Looks good to me! I feel a little weird about prioritizing the slot with the same index as the output (since that means the execution result can change if we ever allow inputs to be reordered or "collapsed"), but I guess that's been the behavior for ages and isn't actually changing in this PR.

webfiltered:
Yep!  It feels kind of wrong programmatically, but also very intuitive from the UI.  When we add re-ordering, we should maybe think about the impact here.

I would in most(?) scenarios, expect it to work the way it does (as a user).  However if the inputs and outputs were e.g. input_1, output_1 and I reordered them... then I'd expect them to still bypass `1 <-> 1`.

But ... I would never do that, because gross.

My initial thought on this was adding an API / override to the LGraphNode instance, that it calls* when bypassed.  But that is a bit more than a "quick fix" PR.

webfiltered:
Oh!  A faint visual indicator to show which slot goes to which when bypassed (like a mini-link inside the node body).

It would only show when links are connected on both sides.

PR #4865, opened 2025-08-09
Fix Simplified Chinese Translation

MakkiShizu:
Revise some traditional Chinese to simplified Chinese.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4865-Fix-Simplified-Chinese-Translation-24a6d73d365081d38c83cb65b54d3c9a) by [Unito](https://www.unito.io)


christian-byrne (approved):
The translation changes from Traditional Chinese to Simplified Chinese are correct and valid. All character conversions maintain the proper meanings.

PR #4879, opened 2025-08-09
Fix Alt-Click-Drag-Copy of Subgraph Nodes

### LINKED ISSUE #4854: [Bug]: Widget order changes when copying subgraph node with alt+click+drag
## Prerequisites
- [x] I have tested with all custom nodes disabled

## What happened?

When copying a subgraph node using alt+click+drag, the widget order gets rearranged. DOM-promoted widgets that were originally in the middle of the widget list get moved to the bottom after all other widgets in the copied node.

## Steps to Reproduce

1. Create or load a subgraph node that has the following structure:
   - 3 input slots (regular connections)
   - 2 DOM-promoted widgets (e.g., from nodes like LoadImage with preview)
   - 2 normal widgets (e.g., number or text inputs)
2. Observe the original widget order on the subgraph node (DOM widgets should be in the middle)
3. Hold Alt and click+drag the subgraph node to create a copy
4. Observe the copied node: the DOM widgets are now at the bottom, below all other widgets
5. The widget functionality remains intact, but the visual order has changed

## How is this affecting you?

## ComfyUI Frontend Version

1.26.0

## Browser

Chrome/Chromium

## Console Errors

No console errors are generated - this is a widget ordering issue.

## Additional Context

This appears to be related to how widgets are serialized and deserialized during the copy process:

1. During serialization, the system creates a sparse array where non-serializable widgets leave gaps
2. During deserialization, it filters to get only serializable widgets (creating a dense array) and maps values by index
3. This causes index mismatches if any widgets have 

Additionally, SubgraphNodes rebuild their widget arrays based on input slot order during configuration (), which may contribute to the reordering.

The issue specifically affects DOM widgets which have special handling and are stored in a separate domWidgetStore with custom  implementations.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4854-Bug-Widget-order-changes-when-copying-subgraph-node-with-alt-click-drag-24a6d73d36508131b9ced17d21fe15d7) by [Unito](https://www.unito.io)

christian-byrne:
Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/4854 by using same logic as the normal paste path:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/e70b127f2af4bb47708679eff29ea7878858c0eb/src/lib/litegraph/src/LGraphCanvas.ts#L3770-L3774

which ensures the node is instantiated and configured against the correct subgraph ID. Previously, the ID would be the same.

**Before**:

https://github.com/user-attachments/assets/351cf95e-bf37-48b0-97d3-99b8b1f526a8

**After**:

https://github.com/user-attachments/assets/d78caf24-e3eb-4ade-9066-94225a316d28

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4879-Fix-Alt-Click-Drag-Copy-of-Subgraph-Nodes-24a6d73d365081d491e7ef5da86460f4) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/subgraph.spec.ts
+++ b/browser_tests/tests/subgraph.spec.ts
@@ -372,6 +372,68 @@ test.describe('Subgraph Operations', () => {
       const deletedNode = await comfyPage.getNodeRefById('2')
       expect(await deletedNode.exists()).toBe(false)
     })
+
+    test.describe('Subgraph copy and paste', () => {
+      test('Can copy subgraph node by dragging + alt', async ({
+        comfyPage
+      }) => {
+        await comfyPage.loadWorkflow('basic-subgraph')
+
+        const subgraphNode = await comfyPage.getNodeRefById('2')
+
+        // Get position of subgraph node
+        const subgraphPos = await subgraphNode.getPosition()
+
+        // Alt + Click on the subgraph node
+        await comfyPage.page.mouse.move(subgraphPos.x + 16, subgraphPos.y + 16)
+        await comfyPage.page.keyboard.down('Alt')
+        await comfyPage.page.mouse.down()
+        await comfyPage.nextFrame()
+
+        // Drag slightly to trigger the copy
+        await comfyPage.page.mouse.move(subgraphPos.x + 64, subgraphPos.y + 64)
+        await comfyPage.page.mouse.up()
+        await comfyPage.page.keyboard.up('Alt')
+
+        // Find all subgraph nodes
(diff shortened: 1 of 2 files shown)


christian-byrne:
I should probably add a test for this.

christian-byrne:
Added browser tests for alt+drag for copying and ensuring alt+drag copying creates new instances rather than linked subgraphs.

PR #4909, opened 2025-08-11
fix: Handle missing _listenerController.abort when loading workflows

### LINKED ISSUE #4907: [Bug]  missing _listenerController.abort when loading workflows
Issue from https://github.com/Comfy-Org/litegraph.js/pull/1134 (reverted) occurs still.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4907-Bug-missing-_listenerController-abort-when-loading-workflows-24c6d73d36508107ac7bd9c86afab208) by [Unito](https://www.unito.io)

vivekchavan14:
- Added proper type checking before calling abort() on _listenerController
- Applied in 3 locations: #addSubgraphInputListeners, configure, and onRemoved
- Fixed TypeError that occurs when loading workflows with promoted DOM widgets
- The issue occurred because _listenerController is not properly initialized on deserialized input slots

fix #4907

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4909-fix-Handle-missing-_listenerController-abort-when-loading-workflows-24c6d73d3650810ab316d1772c61c7d7) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/subgraph/SubgraphNode.ts
+++ b/src/lib/litegraph/src/subgraph/SubgraphNode.ts
@@ -171,7 +171,9 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
     subgraphInput: SubgraphInput,
     input: INodeInputSlot & Partial<ISubgraphInput>
   ) {
-    input._listenerController?.abort()
+    if (input._listenerController) {
+      input._listenerController.abort()
+    }
     input._listenerController = new AbortController()
     const { signal } = input._listenerController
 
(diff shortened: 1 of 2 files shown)


webfiltered (changes requested):
This PR is implementing the same workaround that the original, reverted PR tried.  Instead, we need to address the underlying state corruption so that this does not cause very strange and extremely difficult to trace bugs.

The code -should- be throwing an error here; it is correctly letting us know that the wrong type of object was passed in to the constructor, but because of the `Object.assign` workaround, it is not being caught.

vivekchavan14:
You're right, I'll try looking at the core issue, will close this for now, this PR masks it instead of solving it

christian-byrne:
I think it may be fixed with something like this https://github.com/Comfy-Org/ComfyUI_frontend/compare/fix/subgraph-serialization-issue

jtydhr88:
as discussed, this one should be closed, please rebase and resend if we still need this.
I would like to close it for now

PR #4913, opened 2025-08-11
Add automatic trackpad / mouse sensing

webfiltered:
### Automatic trackpad detection & scaling

- Automatically senses trackpad-style input based on sensible threshold
- Applies appropriate zoom scaling for the detected device
- Supports "flick" gestures
- Tested with Apple, HP, and Microsoft trackpads
- Tested with high and low end 3rd party mice (with and without companion apps)
- Uses built-in shift-wheel control (on MacBook at least, this sends 1/3 the increment of a wheel event when not using shift)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4913-Add-automatic-trackpad-mouse-sensing-24c6d73d36508144909ce3f09091d22d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/CanvasPointer.ts
+++ b/src/lib/litegraph/src/CanvasPointer.ts
@@ -43,6 +43,19 @@ export class CanvasPointer {
   /** {@link maxClickDrift} squared.  Used to calculate click drift without `sqrt`. */
   static #maxClickDrift2 = this.#maxClickDrift ** 2
 
+  /** Assume that "wheel" events with both deltaX and deltaY less than this value are trackpad gestures. */
+  static trackpadThreshold = 60
+
+  /**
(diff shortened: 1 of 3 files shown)


webfiltered:
Expecting test failures on zoom; will use results as baseline to fix.

Edit: Or I just got gaslit by a previous PR test failure.

christian-byrne (approved):
LGTM

lihaoyun6:
This doesn't work at all! Tested on a MacBook Pro M1, I can't move the canvas with two fingers; any two-finger gesture is recognized as a mouse wheel.

lihaoyun6:
image

By checking the code, I found that this feature is missing an important description: "You need to set the Navigation Mode to [Standard (New)] in ComfyUI settings to use this feature." But the default setting used to be "Legacy"😂

PR #4927, opened 2025-08-11
Add vue node feature flag

benceruleanlu:
goal is to add a feature flag that properly separates vue node rendering from the classic rendering, which will allow for vue nodes to continue in development for the next few weeks without needing to constantly rebase

also serves as an actual feature flag for beta testing/QA later

will be replaced by a system to properly plug and play rendering systems as part of #4661 whenever that gets implemented

also, important to note, vue nodes can be broken is when going into main, but it cannot break existing functionality
this is important because there are some known bugs that are breaking litegraph right now, like the image preview widget, and possibly the fileupload widget as well. it's known that we'd need more QA before even having vue nodes in main, even after this feature flag PR, as all this PR does is add a feature flag, not fix existing bugs

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4927-Add-vue-node-feature-flag-24c6d73d365081e994e4f276ad610913) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/DomWidgets.vue
+++ b/src/components/graph/DomWidgets.vue
@@ -16,7 +16,6 @@ import { computed } from 'vue'
 
 import DomWidget from '@/components/graph/widgets/DomWidget.vue'
 import { useChainCallback } from '@/composables/functional/useChainCallback'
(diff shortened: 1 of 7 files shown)


christian-byrne (approved):
LGTM

PR #4929, opened 2025-08-11
[fix] Resolve workflow corruption from subgraph serialization

christian-byrne:
This commit fixes a critical issue where saving a workflow with complex subgraphs could lead to a corrupted file, causing errors upon loading.

The root cause was twofold:

1.  **Unsafe Object Cloning:** The `LiteGraph.cloneObject` function used `JSON.stringify`, which cannot handle circular references that can exist in a nodes properties, especially in complex graphs with promoted widgets. This has been replaced with `_.cloneDeep` from lodash for robust and safe deep cloning.

2.  **Unsafe Node Configuration:** The `LGraphNode.configure` method used a generic loop that would overwrite the `inputs` and `outputs` arrays with plain objects from the serialized data. This created a temporary but dangerous state where class instances were replaced by plain objects, leading to `TypeError` exceptions when methods or private fields were accessed on them. The `configure` method has been refactored to handle `inputs`, `outputs`, and `properties` explicitly and safely, ensuring they are always proper class instances.

This fix makes the node loading process more robust and predictable, preventing data corruption and ensuring the stability of workflows with subgraphs.

## Changes

- **What**: 
- **Breaking**: 
- **Dependencies**: 

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4929-fix-Resolve-workflow-corruption-from-subgraph-serialization-24c6d73d365081a1937af8e477be85bf) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphNode.ts
+++ b/src/lib/litegraph/src/LGraphNode.ts
@@ -769,12 +769,7 @@ export class LGraphNode
       this.graph._version++
     }
     for (const j in info) {
-      if (j == 'properties') {
-        // i don't want to clone properties, I want to reuse the old container
-        for (const k in info.properties) {
-          this.properties[k] = info.properties[k]
-          this.onPropertyChanged?.(k, info.properties[k])
-        }
+      if (j === 'properties' || j === 'inputs' || j === 'outputs') {
         continue
       }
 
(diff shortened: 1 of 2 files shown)


christian-byrne:
So, it seems the exact same tests fail as the other fix (https://github.com/Comfy-Org/ComfyUI_frontend/pull/4481)

christian-byrne:
The original issue seems to be fixed. Will add note to look into underlying system mentioned below

> The LGraphNode.configure method used a generic loop that would overwrite the inputs and outputs arrays with plain objects from the serialized data. This created a temporary but dangerous state where class instances were replaced by plain objects, leading to TypeError exceptions when methods or private fields were accessed on them. The configure method has been refactored to handle inputs, outputs, and properties explicitly and safely, ensuring they are always proper class instances.

PR #4936, opened 2025-08-12
[bugfix] Fix queue not updating with completed task images

weberjc:
## Summary

Fixes bug where task item in queue would sometimes not update after completing and stay as a spinner

## Changes

Add execution_success event listener to trigger queue updates when tasks complete. This ensures that completed task images are displayed in the queue UI after execution finishes in development mode.

Fixes the issue where WebSocket execution events would store image data correctly but the queue UI would never refresh to show the completed status and images.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4936-bugfix-Fix-queue-not-updating-with-completed-task-images-24d6d73d365081528b35faffb12c85f7) by [Unito](https://www.unito.io)

### DIFF
--- a/src/views/GraphView.vue
+++ b/src/views/GraphView.vue
@@ -189,6 +189,10 @@ const onStatus = async (e: CustomEvent<StatusWsMessageStatus>) => {
   await queueStore.update()
 }
 
+const onExecutionSuccess = async () => {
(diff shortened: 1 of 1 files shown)


christian-byrne (approved):
LGTM!

PR #4955, opened 2025-08-13
[bugfix]: Clicking textarea DOM widget triggers selection update causing toolbar remount animation

### LINKED ISSUE #4953: [Bug]: Clicking textarea DOM widget triggers selection update causing toolbar remount animation
## Prerequisites

- [x] I have tested with all custom nodes disabled

## What happened?

When clicking on a textarea DOM widget on a node, it triggers a selection update which in turn re-triggers the toolbar mounting animation. This creates an unwanted visual effect where the toolbar appears to animate/remount when the user is simply trying to interact with the textarea.

## Steps to Reproduce

1. Load a workflow with a node that has a textarea DOM widget
2. Click on the textarea to focus it
3. Observe that the toolbar remounts with animation
4. The animation is triggered unnecessarily

## How is this affecting you?

Visual/UI issue only

## ComfyUI Frontend Version

Latest

## Browser

Chrome/Chromium

## Additional Context

This appears to be a UI state management issue where the selection system is overly sensitive to DOM widget interactions. The textarea click event is propagating or triggering selection change events that cause the toolbar to re-render with its mount animation. The expected behavior would be that clicking into a textarea should not trigger any toolbar state changes.

huchenlei:
## Summary

- Fixed toolbar remount animation when clicking on textarea DOM widgets
- Prevents redundant selection updates for already-selected nodes
- Maintains proper selection behavior for unselected nodes

## Problem

When clicking on a textarea DOM widget on a node, it was triggering a selection update which caused the toolbar to re-animate unnecessarily. This happened because the widget's focus/click event handlers were calling `selectNode` even when the node was already selected.

## Solution

Modified the `DomWidget.vue` component to check if the node is already selected before calling `selectNode`. This prevents redundant selection changes while still allowing proper node selection when clicking on unselected nodes.

## Test plan

1. Load a workflow with textarea widgets (e.g., CLIP Text Encode nodes)
2. Select a node
3. Click on its textarea - toolbar should not re-animate
4. Click on an unselected node's textarea - node should be selected with toolbar animation
5. Verify `bringToFront` still works when clicking on selected nodes

Fixes #4953

## Future work

`LGraphCanvas.selectItems` should not cause `selectionChange` if the final selection state is same as initial selection state.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4955-Bug-Clicking-textarea-DOM-widget-triggers-selection-update-causing-toolbar-remount-ani-24e6d73d3650812c9815f3b0dcf1cfc7) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/widgets/DomWidget.vue
+++ b/src/components/graph/widgets/DomWidget.vue
@@ -140,8 +140,13 @@ const setupDOMEventListeners = () => {
   for (const evt of widget.options.selectOn ?? ['focus', 'click']) {
     useEventListener(widget.element, evt, () => {
       const lgCanvas = canvasStore.canvas
-      lgCanvas?.selectNode(widget.node)
-      lgCanvas?.bringToFront(widget.node)
+      if (!lgCanvas) return
+
+      const node = widget.node
+      if (!node.is_selected) {
+        lgCanvas.selectItems([node])
+      }
+      lgCanvas.bringToFront(node)
     })
   }
 }
--- a/src/locales/ar/commands.json
+++ b/src/locales/ar/commands.json
@@ -131,6 +131,9 @@
   "Comfy_Graph_GroupSelectedNodes": {
     "label": "تجميع العقد المحددة"
   },
+  "Comfy_Graph_UnpackSubgraph": {
+    "label": "فك تجميع الرسم الفرعي المحدد"
+  },
   "Comfy_GroupNode_ConvertSelectedNodesToGroupNode": {
     "label": "تحويل العقد المحددة إلى عقدة مجموعة"
   },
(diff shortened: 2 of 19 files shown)


christian-byrne:
I noticed a double flash sometimes when changing selection. What about using `Transition` component like https://github.com/Comfy-Org/ComfyUI_frontend/compare/main...feature/hide-convert-to-subgraph (ignore changes related to conditional visibility on this branch).

huchenlei:
> I noticed a double flash sometimes when changing selection. What about using `Transition` component like [main...feature/hide-convert-to-subgraph](https://github.com/Comfy-Org/ComfyUI_frontend/compare/main...feature/hide-convert-to-subgraph) (ignore changes related to conditional visibility on this branch).

The transition based solution was not picked previously as
- it remounts the toolbar
- it has issue triggering animation when selection changes

huchenlei:
Screenshot 2025-08-13 at 11 59 49 AM

Note: This PR actually introduce extra bugs. the selection overlay won't be updated after multiple nodes are collapsed to a single node, which in turn will cause wrong positioning of the toolbar. I think the fundamental fix might be more complicated.

Reproduction:
- Convert 2 CLIP text encode to subgraph
- Expose the input of the second CLIP text encode node
- Exit subgraph
- Select the exposed widget
- Observe that the toolbar is rendered on the original CLIP text encode node's location

PR #4964, opened 2025-08-13
Bundled subgraph fixes

AustinMroz:
### Group support for subgraph unpacking
The unpacking code would silently delete groups (the cosmetic colored rectangles). They are now correctly transferred.
### Fix subgraph node position on conversion to subgraph
Converting to subgraph will no longer cause nodes to inch upwards
subgraph-conversion-positioning
### Make unpacking use same positioning calcs as conversion
Non trivial, but unpacking is now a proper inverse for conversion.
subgraph-conversion-inverse
### Clean up old output links when unpacking
Unpacked nodes were left with dangling outputs. This would cause cascading issues later, such as when consecutively unpacking nested subgraphs.
### Minor refactoring for code clarity

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4964-Bundled-subgraph-fixes-24e6d73d365081d3a043ef1531d9d38a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraph.ts
+++ b/src/lib/litegraph/src/LGraph.ts
@@ -1735,17 +1735,16 @@ export class LGraph
         }
       }
     }
-
-    const newLinks: [
-      NodeId,
-      number,
-      NodeId,
-      number,
-      LinkId,
-      RerouteId | undefined,
-      RerouteId | undefined,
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):
LGTM

Issue #4962, opened 2025-08-13
Run litegraph tests as part of CI


christian-byrne:
Run all litegraph unit tests as part of the CI process

christian-byrne:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/5072

PR #4971, opened 2025-08-13
feat: add litegraph tests to CI pipeline

vivekchavan14:
added a new npm script called ```test:litegraph ```to the package.json file that specifically runs the litegraph tests, and then updated the GitHub Actions workflow to include this new script alongside the existing component and unit tests. 

fix #https://github.com/Comfy-Org/ComfyUI_frontend/issues/4962


vivekchavan14:
This PR has quite a few circular dependency issues - might need some refactoring to break those cycles

PR #4988, opened 2025-08-14
Fix inconsistency on bypass from context menu

AustinMroz:
When a node is bypassed from the selection toolbox or by pressing a keybind for bypass, it will also recursively bypass the contents of a subgraph. This effect was not applied when clicking the bypass button from the context menu. The context menu option has been updated to perform the same action as the others so that behaviour is consistent.

See comfyanonymous/ComfyUI#9282

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4988-Fix-inconsistency-on-bypass-from-context-menu-24f6d73d3650818195afcb0c49378ff8) by [Unito](https://www.unito.io)


christian-byrne (approved):
nice, LGTM!

PR #4991, opened 2025-08-14
Fix widget value corruption when input order changes (#4983)

### LINKED ISSUE #4983: Node widget values corrupted when input order differs due to JSON key ordering
### Problem Description

When loading workflows, widget values can appear in the wrong positions if the backend returns node inputs in a different order than when the workflow was saved. This occurs because JSON object key ordering is not guaranteed to be preserved (especially in languages like Go), and the frontend doesn't handle this variation correctly.

For example, a KSampler node might have its seed, steps, and cfg values swapped around when loading a workflow, resulting in a seed value appearing where cfg should be, etc.

### Example

Here's how the same KSampler node's inputs can be returned in different orders from different backends:

**Backend A (alphabetical order):**
```json
{
  "cfg": ["FLOAT", { "default": 8, "max": 100, "min": 0, "round": 0.01, "step": 0.1 }],
  "denoise": ["FLOAT", { "default": 1, "max": 1, "min": 0, "step": 0.01 }],
  "latent_image": ["LATENT", {}],
  "model": ["MODEL", {}],
  "negative": ["CONDITIONING", {}],
  "positive": ["CONDITIONING", {}],
  "sampler_name": [["euler", "euler_cfg_pp", "..."], {}],
  "scheduler": [["simple", "sgm_uniform", "..."], {}],
  "seed": ["INT", { "default": 0, "max": 18446744073709552000, "min": 0 }],
  "steps": ["INT", { "default": 20, "max": 10000, "min": 1 }]
}
```

**Backend B (different order):**
```json
{
  "model": ["MODEL", {}],
  "seed": ["INT", { "default": 0, "min": 0, "max": 18446744073709551615 }],
  "steps": ["INT", { "default": 20, "min": 1, "max": 10000 }],
  "cfg": ["FLOAT", { "default": 8.0, "min": 0.0, "max": 100.0, "step": 0.1, "round": 0.01 }],
  "sampler_name": [["euler", "euler_cfg_pp", "..."], {}],
  "scheduler": [["simple", "sgm_uniform", "..."], {}],
  "positive": ["CONDITIONING", {}],
  "negative": ["CONDITIONING", {}],
  "latent_image": ["LATENT", {}],
  "denoise": ["FLOAT", { "default": 1.0, "min": 0.0, "max": 1.0, "step": 0.01 }]
}
```

The widget values in a workflow are stored as an array like this:
```json
{
  "3": {
    "class_type": "KSampler",
    "inputs": { ... },
    "widgets_values": [156680208700286, 1234, "euler", "normal", 20, 8, 1]
  }
}
```

If the widget order changes between save and load, these values get misaligned - the seed value might end up where steps should be, etc.

### Technical Context

- Widget values are serialized as an array indexed by widget position
- The backend provides an `input_order` field in node definitions that specifies the correct order
- The issue relates to how the frontend transforms node definitions and creates widgets from inputs
- This affects the workflow serialization/deserialization process

### Key Areas to Investigate

- How node definitions are transformed when received from the backend
- The widget creation process during node registration
- How widget values are serialized to and deserialized from workflow JSON
- Where and how the `input_order` field could be utilized

### Expected Behavior

- Widget values should maintain their correct associations regardless of JSON key ordering from the backend
- The solution should gracefully handle cases where `input_order` might not exist (backwards compatibility)
- Workflows should be portable between different ComfyUI backends without value corruption

vivekchavan14:
The Problem:
•  Widget values were stored as a simple array by index position
•  Widget creation order depended on JSON key ordering from backend
•  When JSON key order changed, widgets got created in different order
•  Same indexed values got assigned to wrong widgets = corruption

The Fix:
•  Changed from index-based to name-based widget value storage
•  Now stores {name: "seed", value: 12345} instead of just 12345
•  On load, matches values by widget name instead of position
•  Fully backward compatible - old workflows still work fine

fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/4983

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4991-Fix-widget-value-corruption-when-input-order-changes-4983-24f6d73d365081ca9dcdebba5bed9819) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphNode.ts
+++ b/src/lib/litegraph/src/LGraphNode.ts
@@ -849,10 +849,38 @@ export class LGraphNode
         const widgetsWithValue = this.widgets.filter(
           (w) => w.serialize !== false
         )
-        for (let i = 0; i < info.widgets_values.length; ++i) {
-          const widget = widgetsWithValue[i]
-          if (widget) {
-            widget.value = info.widgets_values[i]
+
+        // Enhanced widget values restoration with name-based mapping
+        // Check if widgets_values contains name-based mapping (new format)
+        if (
+          Array.isArray(info.widgets_values) &&
+          info.widgets_values.length > 0 &&
+          info.widgets_values[0] &&
+          typeof info.widgets_values[0] === 'object' &&
+          'name' in info.widgets_values[0] &&
+          'value' in info.widgets_values[0]
+        ) {
+          // New format: name-based mapping
+          const valuesByName = new Map()
+          for (const item of info.widgets_values as Array<{
+            name: string
+            value: any
+          }>) {
+            valuesByName.set(item.name, item.value)
+          }
+
+          for (const widget of widgetsWithValue) {
+            if (valuesByName.has(widget.name)) {
+              widget.value = valuesByName.get(widget.name)
+            }
+          }
+        } else {
+          // Legacy format: index-based mapping for backward compatibility
+          for (let i = 0; i < info.widgets_values.length; ++i) {
+            const widget = widgetsWithValue[i]
+            if (widget) {
+              widget.value = info.widgets_values[i]
+            }
           }
         }
       }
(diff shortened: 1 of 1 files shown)


jtydhr88:
the original issue has been fixed, I don't think we still need this, close it for now

PR #4992, opened 2025-08-14
Fix user.css loading in multi-user mode (#4978)

### LINKED ISSUE #4978: [Bug]: user.css fetched from default user when Comfy started with --multi-user
### Prerequisites

### What happened?

When activating --multi-user in comfy, the front end still tries to fetch `user.css` from default user.

It should instead probably fetch the actual user one.

### Steps to Reproduce

1. start comfy with `--mutli-user`
2. open comfy and setup your username
3. check network logs and see the 404 on user.css
4. add a `default` user with a 'user.css`
5. reload
6. notice the error is not there anymore

Minor inconvenience

### ComfyUI Frontend Version

1.24.4

### Browser

Chrome/Chromium

### Console Errors

```javascript
GET http://localhost:8190/api/userdata/user.css net::ERR_ABORTED 404 (Not Found)
```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4978-Bug-user-css-fetched-from-default-user-when-Comfy-started-with-multi-user-24f6d73d36508158a0aae989af5e1235) by [Unito](https://www.unito.io)

vivekchavan14:
Problem:
•  Frontend had hardcoded <link href="api/userdata/user.css"> in index.html
•  This ignored the current user context and caused 404s in multi-user mode
•  No way to load user-specific CSS files

Implemented Solution:
•  Removed hardcoded CSS link from HTML
•  Created dynamic CSS loading system that respects user context
•  Uses proper Comfy-User headers for multi-user API calls
•  Gracefully handles missing CSS files without errors
•  Works in both single-user and multi-user modes

Testing:
•  Verified with curl commands for different user contexts
•  Handles 404s gracefully without breaking the app
•  CSS switches correctly when users login/logout

fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/4978

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4992-Fix-user-css-loading-in-multi-user-mode-4978-24f6d73d3650819b9326f857bdf6de33) by [Unito](https://www.unito.io)

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,7 +6,6 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
     <link rel="stylesheet" type="text/css" href="materialdesignicons.min.css" />
     <link rel="stylesheet" type="text/css" href="user.css" />
-    <link rel="stylesheet" type="text/css" href="api/userdata/user.css" />
     
     <!-- Fullscreen mode on iOS -->
     <meta name="apple-mobile-web-app-capable" content="yes">
--- a/src/stores/userStore.ts
+++ b/src/stores/userStore.ts
@@ -3,6 +3,7 @@ import { computed, ref, watchEffect } from 'vue'
 
 import type { User as UserConfig } from '@/schemas/apiSchema'
 import { api } from '@/scripts/api'
+import { CSSLoader } from '@/utils/cssLoader'
 
 export interface User {
   userId: string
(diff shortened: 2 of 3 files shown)


viva-jinyi:
It looks like this PR mainly avoids breaking the app when user.css is missing, but it doesn’t really solve the root cause.

Even when logged in, /api/userdata/user.css still returns 404 — likely because the file doesn’t exist by default, or because <link> requests can’t include the Comfy-User header.

jtydhr88:
as there is not update since last Aug, and has conflict now, close it for now. please rebase and resend if we still need it.

PR #5003, opened 2025-08-15
Deep copy subgraphs to clipboard, update nested ids on paste

### LINKED ISSUE #5002: copyToClipboard only shallow copies subgraphs
The copyToClipboard method of LGraphCanvas only walks the current graph for determining which subgraphs need to be serialized.

Getting to an actual visible error requires more steps than expected
- Double nest a subgraph
- Copy from the top level
- Paste into a new workflow
  - At this point, the workflow verifiably only contains the single subgraph, but the nested one can still be viewed
- Isolate from the parent workflow (closing it is sufficient)
- Reload, and an error is thrown for non-existant subgraph

Image

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-5002-copyToClipboard-only-shallow-copies-subgraphs-2506d73d365081d58f87ce3554c2b81f) by [Unito](https://www.unito.io)

AustinMroz:
Resolves #5002

The copyToClipboard function wasn't walking subgraphs and leaving nested subgraphs unserialized. This has now been fixed. Note that this implementation involves adding subgraphs to the set mid iteration.

This requires that equivalent support be added to _pasteFromClipboard to update the ids of nested subgraphs which are pasted.

EDIT: Removed the now resolved TODO

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5003-Deep-copy-subgraphs-to-clipboard-update-nested-ids-on-paste-2506d73d36508141b6f9f1cf4a4ea3dc) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -3646,8 +3646,12 @@ export class LGraphCanvas
     }
 
     // Add unique subgraph entries
-    // TODO: Must find all nested subgraphs
     for (const subgraph of subgraphs) {
+      for (const node of subgraph.nodes) {
+        if (node instanceof SubgraphNode) {
+          subgraphs.add(node.subgraph)
+        }
+      }
       const cloned = subgraph.clone(true).asSerialisable()
       serialisable.subgraphs.push(cloned)
     }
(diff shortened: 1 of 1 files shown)


webfiltered (commented):
First glance says this is only going one level deep - should it be recursive like the flattening in execution?

AustinMroz:
Will double check, but that's the tricky part I draw notice to. The set is being modifed as it's being iterated over.

EDIT: I can confirm a 5-deep nested subgraph is copying as expected.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -3646,8 +3646,12 @@ export class LGraphCanvas
     }
 
     // Add unique subgraph entries
-    // TODO: Must find all nested subgraphs
     for (const subgraph of subgraphs) {
+      for (const node of subgraph.nodes) {
+        if (node instanceof SubgraphNode) {
+          subgraphs.add(node.subgraph)
+        }
+      }
```
christian-byrne, line 3656: This is pretty sleek.

Do you think we should add a comment specifying that `subgraphs` must not be changed from a `Set` to an `Array`? It's possible a future maintainer doesn't understand this nuance on first glance.

christian-byrne (dismissed):
LGTM with possible suggestion.

src/lib/litegraph/src/LGraphCanvas.ts
AustinMroz: Good call. I've added a note outside the loop, and at the definition for subgraphs.

christian-byrne (approved):

PR #5005, opened 2025-08-15
fix: Correct traditional Chinese to simplified Chinese in translations

comfyui-wiki:
Corrected inconsistent translations that were using traditional Chinese characters instead of simplified Chinese in localization files.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5005-fix-Correct-traditional-Chinese-to-simplified-Chinese-in-translations-2506d73d365081f797f0e68c9042438a) by [Unito](https://www.unito.io)


christian-byrne (approved):

christian-byrne:
I will do the backport

PR #5024, opened 2025-08-15
Fix Connection of Primitive nodes to Subgraph node

AustinMroz:
Had a clever idea which greatly simplifies the logic of nesting here. This patch is now entirely usable, and, in my opinion, worth releasing, but I've noticed a little spottyness with testing of swapped connections to partially matching types (FLOAT to FLOAT with different max/min)

subgraph-primitives

Resolves #5000

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5024-Fix-Connection-of-Primitive-nodes-to-Subgraph-node-2506d73d365081afaf1cc58656aee460) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/subgraph/SubgraphNode.ts
+++ b/src/lib/litegraph/src/subgraph/SubgraphNode.ts
@@ -369,10 +369,11 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
       widget: promotedWidget,
       subgraphNode: this
     })
-
-    const backingInput =
-      (widget as { node?: LGraphNode }).node?.findInputSlot(widget.name, true)
-        ?.widget ?? {}
+    const hasNode = (w: any): w is { node: LGraphNode } =>
+      w && typeof w.node === 'object' && w.node.findInputSlot
+    const backingInput = hasNode(widget)
+      ? widget.node.findInputSlot(widget.name, true)?.widget ?? {}
+      : {}
     input.widget ??= { name: subgraphInput.name }
     input.widget.name = subgraphInput.name
     Object.setPrototypeOf(input.widget, backingInput)
--- a/src/lib/litegraph/src/subgraph/SubgraphNode.ts
+++ b/src/lib/litegraph/src/subgraph/SubgraphNode.ts
@@ -313,9 +313,8 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
     widget: Readonly<IBaseWidget>
   ) {
     // Use the first matching widget
-    const promotedWidget = toConcreteWidget(widget, this).createCopyForNode(
-      this
-    )
+    const targetWidget = toConcreteWidget(widget, this)
+    const promotedWidget = targetWidget.createCopyForNode(this)
 
     Object.assign(promotedWidget, {
       get name() {
@@ -369,11 +368,9 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
       widget: promotedWidget,
       subgraphNode: this
     })
-    const hasNode = (w: any): w is { node: LGraphNode } =>
-      w && typeof w.node === 'object' && w.node.findInputSlot
-    const backingInput = hasNode(widget)
-      ? widget.node.findInputSlot(widget.name, true)?.widget ?? {}
-      : {}
+
+    const backingInput =
+      targetWidget.node.findInputSlot(widget.name, true)?.widget ?? {}
     input.widget ??= { name: subgraphInput.name }
     input.widget.name = subgraphInput.name
     Object.setPrototypeOf(input.widget, backingInput)
--- a/src/lib/litegraph/src/subgraph/SubgraphNode.ts
+++ b/src/lib/litegraph/src/subgraph/SubgraphNode.ts
@@ -370,7 +370,10 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
       subgraphNode: this
     })
 
-    input.widget = { name: subgraphInput.name }
+    const backingInput =
+      (widget as { node?: LGraphNode }).node?.findInputSlot(widget.name, true)
+        ?.widget ?? {}
+    input.widget = { ...backingInput, name: subgraphInput.name }
     input._widget = promotedWidget
   }
 
--- a/src/lib/litegraph/src/subgraph/SubgraphNode.ts
+++ b/src/lib/litegraph/src/subgraph/SubgraphNode.ts
@@ -373,7 +373,9 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
     const backingInput =
       (widget as { node?: LGraphNode }).node?.findInputSlot(widget.name, true)
         ?.widget ?? {}
-    input.widget = { ...backingInput, name: subgraphInput.name }
+    input.widget ??= { name: subgraphInput.name }
+    input.widget.name = subgraphInput.name
+    Object.setPrototypeOf(input.widget, backingInput)
     input._widget = promotedWidget
   }
 
--- a/src/lib/litegraph/src/subgraph/SubgraphNode.ts
+++ b/src/lib/litegraph/src/subgraph/SubgraphNode.ts
@@ -369,11 +369,15 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
       subgraphNode: this
     })
 
+    // NOTE: This code creates linked chains of prototypes for passing across
+    // multiple levels of subgraphs. As part of this, it intentionally avoids
+    // creating new objects. Have care when making changes.
     const backingInput =
       targetWidget.node.findInputSlot(widget.name, true)?.widget ?? {}
     input.widget ??= { name: subgraphInput.name }
     input.widget.name = subgraphInput.name
     Object.setPrototypeOf(input.widget, backingInput)
+
     input._widget = promotedWidget
   }
 


src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -370,7 +370,12 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
       subgraphNode: this
     })
 
-    input.widget = { name: subgraphInput.name }
+    const backingInput =
+      (widget as { node?: LGraphNode }).node?.findInputSlot(widget.name, true)
```
christian-byrne: This type assertion is potentially unsafe and could cause runtime errors if the widget doesn't have the expected structure. What about using a type guard instead of casting like

```typescript
const hasNode = (w: any): w is { node: LGraphNode } => w && typeof w.node === 'object' && w.node.findInputSlot
const backingInput = hasNode(widget) ? widget.node.findInputSlot(widget.name, true)?.widget ?? {} : {}
```

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -370,7 +370,12 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+        ?.widget ?? {}
+    input.widget ??= { name: subgraphInput.name }
+    input.widget.name = subgraphInput.name
+    Object.setPrototypeOf(input.widget, backingInput)
```
christian-byrne, line 379: I think `Object.setPrototypeOf` is pretty slow and creates complex prototype chains that are hard to debug. If you need inheritance behavior, what about object composition or a more explicit pattern with proper validation.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -370,7 +370,12 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
       subgraphNode: this
     })
 
-    input.widget = { name: subgraphInput.name }
+    const backingInput =
+      (widget as { node?: LGraphNode }).node?.findInputSlot(widget.name, true)
+        ?.widget ?? {}
+    input.widget ??= { name: subgraphInput.name }
```
christian-byrne, line 377: nit: The name property is being set twice here which is redundant. You could simplify this by just setting input.widget ??= {} and then assigning the name once.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -370,7 +370,12 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
       subgraphNode: this
     })
 
-    input.widget = { name: subgraphInput.name }
+    const backingInput =
```
christian-byrne, line 375: nit: Consider adding some error handling here in case findInputSlot returns null or the widget is missing

src/lib/litegraph/src/subgraph/SubgraphNode.ts
AustinMroz: This is objectively better. :+1:

src/lib/litegraph/src/subgraph/SubgraphNode.ts
AustinMroz: Unfortunately, that's the key ugly crux of this PR. Swapping prototypes on existing objects is required to avoid the very complex (and from my time attempting implementation, error prone) event bubbling approach.  If two subgraphs are nested and the inner one has a connection change, this propagates the changes upwards without requiring link resolution.

The speed cost here *should* be negligible as the input.widget property is only accessed when connecting a primitive widget (verified by grep). If the concept of the implementation is otherwise agreeable, I'm willing to sit down and do some in depth profiling to make sure I'm not misunderstanding things.

I don't propose this with expectation of acceptance; I want the proposal to be kicked back to the drawing board if it leaves a bad feeling, but it's currently required baggage.

To that end, an implementation that adds an optional  "child reference" to input.widget  and instead changes the primitive node code to query for it would be near equivalent, but potentially break compatibility with custom nodes.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
AustinMroz: This was how I originally implemented it as well, but it causes type errors since {} can't be assigned to input.widget.

Do you think it would be nicer if a temp variable was introduced and it was only assigned back after name is set and the type satisfies?

src/lib/litegraph/src/subgraph/SubgraphNode.ts
AustinMroz: That case is already handled. The prototype is set to an empty object, clearing any prior primitive widgetDef. As an empty dict doesn't contain a "GET_CONFIG symbol" element, it is functionally equivalent to the prior implementation of not containing a 

But adding some angry danger comments would be a good idea. Both to clarify this and the other areas.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne: Thanks for the context - the nested event propagation requirement makes sense.

Given this only happens during primitive connections (not a hot path), the performance impact is probably negligible. The child reference alternative sounds interesting, but this also works fine for now IMO.

Maybe it could benefit from a comment explaining the rationale/purpose.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne: This is totally fine in that case!

christian-byrne (dismissed):
LGTM, remaining comments are just nits

christian-byrne:
Do you know about this github feature Austin https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests?

AustinMroz:
I do! I'm quite a fan of it and have made use of it in most of the PRs I open. In this case, I didn't have a full  fix when I opened the PR as draft and forgot to edit it when I returned to the issue. I try to err on the side of caution with declaring an issue fixed when a PR is only a partial solution.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -369,8 +369,14 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
       widget: promotedWidget,
       subgraphNode: this
     })
-
-    input.widget = { name: subgraphInput.name }
+    const hasNode = (w: any): w is { node: LGraphNode } =>
```
DrJKL: Something like this?

```suggestion
    function hasNode(w: unknown): w is { node: LGraphNode } {
      return typeof w === 'object' &&
        typeof w['node'] === 'object' &&
        w.node.findInputSlot !== undefined
```

DrJKL: Though it's still kind of a lie unless you also validate the full shape of LGraphNode 🫤

AustinMroz: The `node` property is provided by `BaseWidget` which is part of litegraph, so the the property is guaranteed for anything that isn't a custom node. The real question becomes how much we  trust/distrust that a custom node devs could have a `node` property  that isn't actually an `LGraphNode`.

christian-byrne: I think inferring that `node` is a `LGraphNode` is fair if those conditions are true, based on our knowledge of the world. One thing for certain though is that we should be narrowing from the top type `unknown`

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -369,8 +369,14 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+      w && typeof w.node === 'object' && w.node.findInputSlot
```
webfiltered:
```suggestion
    function hasNode(
      w: typeof widget
    ): w is typeof w & { node: Pick<LGraphNode, 'findInputSlot'> } {
      return (
        'node' in w &&
        typeof w.node === 'object' &&
        w.node !== null &&
        'findInputSlot' in w.node &&
        typeof w.node.findInputSlot === 'function'
      )
    }
```

typescript4eva.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
webfiltered: Upcasting from interfaced widgets back to concrete was obviously.. not a thing that was planned hah.  But I left a strongly-typed guard that should cover all concerns below.

(.. please don't use it)

christian-byrne:
Anything blocking on this PR?

src/lib/litegraph/src/subgraph/SubgraphNode.ts
webfiltered: PR'd an actual (non-joke) solve for the type assertion.  @AustinMroz if you're happy with it, all yours~
#5493

AustinMroz:
All the feedback was nonblocking nits. With webfiltered's beautiful PR (and a forgotten warning comment), it's solidly GTM.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -370,7 +369,15 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
 
-    input.widget = { name: subgraphInput.name }
+    // NOTE: This code creates linked chains of prototypes for passing across
+    // multiple levels of subgraphs. As part of this, it intentionally avoids
+    // creating new objects. Have care when making changes.
+    const backingInput =
+      targetWidget.node.findInputSlot(widget.name, true)?.widget ?? {}
+    input.widget ??= { name: subgraphInput.name }
+    input.widget.name = subgraphInput.name
+    Object.setPrototypeOf(input.widget, backingInput)
```
webfiltered, line 379: Sorry just need to pile on to the `setPrototypeOf` debate, as it hasn't been mentioned: this bypasses the constructor.  So long as we're not doing anything interesting in the constructor (this *includes* props with initialisers), it should be fine right now.. but is extremely brittle.

```ts
class TestClass {
  a = 123;
  b;
  constructor() {
    this.b = 456;
  }
}

const good = new TestClass();
const bad = {};
Object.setPrototypeOf(bad, TestClass);

good.a // 123
good.b // 456

bad.a // undefined
bad.b // undefined
```

webfiltered: (I clicked review with the intention of rubber stamping to enable you to merge... sorry!)

AustinMroz: All feedback on the section of actual concern is great!

Here specifically, I consider the properties desirable. The prototype isn't being set to a class, but an already existing javascript object. Extending off your example,
```javascript
const ok = {}
Object.setPrototypeOf(ok, good)

ok.a // 123
ok.b // 456
```

Even if the type for `input.widget` changes extensively in the future, no change will be required here unless `{ name: subgraphInput.name }` ceases to be a valid type. At that point typescript will still mark this code as invalid.

webfiltered: HAHA!  Christ.  Not a class.  I literally did not read to the end of the code line.  Sorry!  >.<

webfiltered (approved):

christian-byrne (approved):
LGTM!

PR #5029, opened 2025-08-16
fix: issue #4121

arjansingh:
## Summary

Search popover wasn't adding nodes. Now it adds the selected node.

## Changes

- **What**: We were early exiting in the absence of a canvas `triggerEvent`. We were just using that event to add the node where a users's mouse was if the added via right-click. Now instead we're using the fallback value which is the center of the canvas.

## Review Focus

Fixes #4121 

https://github.com/user-attachments/assets/8e8b65ec-08bd-4c06-a28f-e9a3b31961da

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5029-fix-issue-4121-2516d73d3650813eb666eb94a5be1d36) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM!

PR #5041, opened 2025-08-16
Fix menu commands categorization and complete i18n localization

vivekchavan14:
Completed missing translations and localization support for menu commands, ensuring that all labels and categorization elements are properly internationalized.

fix #https://github.com/Comfy-Org/ComfyUI_frontend/issues/5039

Screenshot from 2025-08-17 02-17-29

image

### DIFF
--- a/src/components/LiteGraphCanvasSplitterOverlay.vue
+++ b/src/components/LiteGraphCanvasSplitterOverlay.vue
@@ -63,9 +63,14 @@ const unifiedWidth = computed(() =>
   settingStore.get('Comfy.Sidebar.UnifiedWidth')
 )
 
-const sidebarPanelVisible = computed(
-  () => useSidebarTabStore().activeSidebarTab !== null
-)
+const sidebarPanelVisible = computed(() => {
+  const sidebarTabStore = useSidebarTabStore()
+  // Show sidebar if there's an active tab OR if there are tabs available (fallback)
+  return (
+    sidebarTabStore.activeSidebarTab !== null ||
+    sidebarTabStore.sidebarTabs.length > 0
+  )
+})
 const bottomPanelVisible = computed(
   () => useBottomPanelStore().bottomPanelVisible
 )
--- a/src/components/bottomPanel/BottomPanel.vue
+++ b/src/components/bottomPanel/BottomPanel.vue
@@ -13,8 +13,8 @@
               <span class="font-bold">
                 {{
                   shouldCapitalizeTab(tab.id)
-                    ? tab.title.toUpperCase()
-                    : tab.title
+                    ? getTabTitle(tab).toUpperCase()
+                    : getTabTitle(tab)
                 }}
               </span>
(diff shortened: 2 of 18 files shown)


src/components/LiteGraphCanvasSplitterOverlay.vue
```diff
@@ -63,9 +63,14 @@ const unifiedWidth = computed(() =>
...
 
-const sidebarPanelVisible = computed(
-  () => useSidebarTabStore().activeSidebarTab !== null
-)
+const sidebarPanelVisible = computed(() => {
+  const sidebarTabStore = useSidebarTabStore()
+  // Show sidebar if there's an active tab OR if there are tabs available (fallback)
+  return (
+    sidebarTabStore.activeSidebarTab !== null ||
+    sidebarTabStore.sidebarTabs.length > 0
```
viva-jinyi, line 71: The condition `sidebarTabStore.sidebarTabs.length > 0` makes the sidebar always visible when tabs are registered, even when the user intentionally closes all tabs. This breaks the expected UX behavior where users should be able to hide the sidebar completely.

src/components/bottomPanel/BottomPanel.vue
```diff
@@ -13,8 +13,8 @@
               <span class="font-bold">
                 {{
                   shouldCapitalizeTab(tab.id)
-                    ? tab.title.toUpperCase()
-                    : tab.title
+                    ? getTabTitle(tab).toUpperCase()
```
viva-jinyi, line 16: This will throw an error if tab.title is a function

src/stores/workspace/sidebarTabStore.ts
```diff
@@ -28,6 +28,11 @@ export const useSidebarTabStore = defineStore('sidebarTab', () => {
   const registerSidebarTab = (tab: SidebarTabExtension) => {
     sidebarTabs.value = [...sidebarTabs.value, tab]
 
+    // Auto-select first tab if none is active
+    if (!activeSidebarTabId.value && sidebarTabs.value.length === 1) {
```
viva-jinyi, line 32: This automatically opens the first tab when registering, which might not be desired behavior. Consider these scenarios:

1. User intentionally closes all sidebar tabs to maximize canvas space
2. On app initialization, the sidebar immediately opens without user action
3. When extensions register tabs dynamically, they auto-open unexpectedly

viva-jinyi (changes requested):

src/components/LiteGraphCanvasSplitterOverlay.vue
benceruleanlu: if something is expected UX, we should probably have a test case affirming that so any PR authors would understand from reading the failed test

not sure if this exists though

jtydhr88:
as the orignial issue has been fixed, and this PR has huge confilct, please rebase and resend if we still need this.
close it for now

PR #5051, opened 2025-08-17
Add support for custom icons in menu

pythongosssss:
## Summary

Add support for Vue component icons in menu and change Browse Templates menu item to use the same custom icon as the sidebar tab icon as there was a mismatch

## Changes
- **What**:  Support custom icon components in command menu using  `<component :is>` to either render an `i` element or the component

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5051-Add-support-for-custom-icons-in-menu-2526d73d36508113b020cf752a3c0e34) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/menu.spec.ts
+++ b/browser_tests/tests/menu.spec.ts
@@ -178,6 +178,79 @@ test.describe('Menu', () => {
       await comfyPage.menu.topbar.triggerTopbarCommand(['ext', 'foo-command'])
       expect(await comfyPage.getVisibleToastCount()).toBe(1)
     })
+
+    test('Browse Templates custom icon is visible and matches sidebar icon', async ({
+      comfyPage
+    }) => {
+      // Open the top menu
+      await comfyPage.menu.topbar.openTopbarMenu()
+      const menu = comfyPage.page.locator('.comfy-command-menu')
+
+      // Find the Browse Templates menu item
+      const browseTemplatesItem = menu.locator(
+        '.p-menubar-item-label:text-is("Browse Templates")'
+      )
+      await expect(browseTemplatesItem).toBeVisible()
+
+      // Check that the Browse Templates item has an icon
+      const menuIcon = browseTemplatesItem
+        .locator('..')
+        .locator('.p-menubar-item-icon')
+        .first()
+      await expect(menuIcon).toBeVisible()
+
+      // Get the icon's tag name and class to verify it's a component (not a string icon)
+      const menuIconType = await menuIcon.evaluate((el) => {
+        // If it's a Vue component, it will not have pi/mdi classes
+        // and should be an SVG or custom component
+        const tagName = el.tagName.toLowerCase()
+        const classes = el.className || ''
+        return {
+          tagName,
+          classes,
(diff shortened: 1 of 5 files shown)


christian-byrne:
I wonder if it will be better to just bite the bullet and setup the plugin for the dynamic CSS icon selectors https://iconify.design/docs/usage/css/tailwind/tailwind3/#plugins.

Possible downsides:

- worse bundle size; unplugin-icons compatibility?
- more upfront initialization I think
 
Benefits:

- better DX
- simpler code and types

src/components/topbar/CommandMenubar.vue
```diff
@@ -66,18 +66,20 @@
...
-        <span
-          v-else-if="
-            item.icon && item.comfyCommand?.id !== 'Comfy.NewBlankWorkflow'
-          "
+        <component
+          :is="getIconComponent(item)"
+          v-else-if="getIconLocation(item) === 'left'"
           class="p-menubar-item-icon"
-          :class="item.icon"
+          :class="typeof item.icon === 'string' ? item.icon : undefined"
```
DrJKL, line 73: Since this is repeated a few times, could it be extracted?

src/components/graph/selectionToolbox/ExtensionCommandButton.vue
```diff
@@ -7,7 +7,13 @@
...
     severity="secondary"
     text
-    :icon="typeof command.icon === 'function' ? command.icon() : command.icon"
+    :icon="
+      typeof command.icon === 'function'
+        ? command.icon()
+        : typeof command.icon === 'string'
+          ? command.icon
+          : undefined
+    "
```
DrJKL, line 16: Could you pull this into the script?

src/components/topbar/CommandMenubar.vue
```diff
@@ -254,7 +266,8 @@ const translatedItems = computed(() => {
       : [])
   )
 
-  return items
+  // Cast to PrimeVue type - our template overrides icon handling
+  return items as PrimeMenuItem[]
```
DrJKL, line 270: Is there a type guard we can apply to guarantee the asserted type?

DrJKL: Or can it be typed on instantiation?

browser_tests/tests/menu.spec.ts
```diff
@@ -178,6 +178,79 @@ test.describe('Menu', () => {
...
+        // and should be an SVG or custom component
+        const tagName = el.tagName.toLowerCase()
+        const classes = el.className || ''
+        return {
+          tagName,
+          classes,
+          hasStringIcon:
+            typeof classes === 'string' &&
+            (classes.includes('pi ') || classes.includes('mdi '))
+        }
```
DrJKL, line 214: Since you're doing this twice, could you make it a utility in here?

Also, I think instanceof and classList would give better fidelity for the checks.

pythongosssss:
Abandoning this will re-implement using the tailwind/iconify plugin as it's a better approach

PR #5054, opened 2025-08-17
[fix] Resolve group node execution error when connecting to external nodes

christian-byrne:
Fixes a bug where group node children would fail to execute when connected to nodes outside the group, throwing 'Failed to get input node' errors.

The issue was in ExecutableGroupNodeChildDTO.resolveInput which assumed all node IDs follow a colon-separated format, but external nodes have simple IDs. Now properly handles both external connections (trying full ID first) and internal group node connections (falling back to shortened ID).

Includes comprehensive unit tests to prevent regression.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5054-fix-Resolve-group-node-execution-error-when-connecting-to-external-nodes-2526d73d3650816db779fbc080c10579) by [Unito](https://www.unito.io)

### DIFF
--- a/src/utils/executableGroupNodeChildDTO.ts
+++ b/src/utils/executableGroupNodeChildDTO.ts
@@ -27,6 +27,13 @@ export class ExecutableGroupNodeChildDTO extends ExecutableNodeDTO {
   }
 
   override resolveInput(slot: number) {
+    // Check if this group node is inside a subgraph (unsupported)
+    if (this.id.split(':').length > 2) {
+      throw new Error(
+        'Group nodes inside subgraphs are not supported. Please convert the group node to a subgraph instead.'
+      )
+    }
+
     const inputNode = this.node.getInputNode(slot)
     if (!inputNode) return
 
--- a/src/utils/executableGroupNodeChildDTO.ts
+++ b/src/utils/executableGroupNodeChildDTO.ts
@@ -33,19 +33,28 @@ export class ExecutableGroupNodeChildDTO extends ExecutableNodeDTO {
     const link = this.node.getInputLink(slot)
     if (!link) throw new Error('Failed to get input link')
 
-    const id = String(inputNode.id).split(':').at(-1)
-    if (id === undefined) throw new Error('Invalid input node id')
+    const inputNodeId = String(inputNode.id)
 
-    const inputNodeDto = this.nodesByExecutionId?.get(id)
+    // Try to find the node using the full ID first (for nodes outside the group)
+    let inputNodeDto = this.nodesByExecutionId?.get(inputNodeId)
+
+    // If not found, try with just the last part of the ID (for nodes inside the group)
+    if (!inputNodeDto) {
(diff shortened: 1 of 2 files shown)


AustinMroz:
It's an improvement, but things still break if a group node is placed inside a subgraph. Doing some quick testing to see if there's a saner fix than just iteratively trying combinations of id chunks

I'm also seeing errors on execution caused by failing to resolve progress bar display. This seems to be entirely cosmetic and can be pushed to a future fix.

AustinMroz:
The issue is more complex than I initially presumed.  While you can calculate the full id of the resolved input from the id of the current ExecutableNodeDTO, it just creates cascading errors when trying to resolve links across the subgraph boundry to the (not actually real) group node.

I think it's fine to merge as is with optional minor nit:
- If a group node is inside a subgraph (I think `this.id.split(':').length > 2` is sufficient?), instead print an error about groupnodes inside subgraphs being unsupported.

christian-byrne:
@AustinMroz Thanks for the feedback\! You're right that this doesn't handle group nodes inside subgraphs.

I was thinking we could leverage the existing `parseExecutionId` utility from `graphTraversalUtil.ts` for a more robust solution:

```typescript
// Use parseExecutionId to handle all nesting scenarios
const inputNodeIdParts = parseExecutionId(inputNodeId)
let inputNodeDto = this.nodesByExecutionId?.get(inputNodeId)

if (!inputNodeDto && inputNodeIdParts) {
  // Try last segment (nodes in same group)
  const lastSegment = inputNodeIdParts[inputNodeIdParts.length - 1]
  inputNodeDto = this.nodesByExecutionId?.get(lastSegment)
  
  // If in subgraph, try constructing proper execution ID
  if (!inputNodeDto && this.subgraphNodePath.length > 0) {
    const sameSubgraphId = [...this.subgraphNodePath, lastSegment].join(':')
    inputNodeDto = this.nodesByExecutionId?.get(sameSubgraphId)
  }
}
```

This would handle more complex nesting cases. What do you think about this approach vs the iterative solution you mentioned?

AustinMroz:
My initial suggestion was misguided and not viable. The solution I eventually saw partial success with was functionally equivalent to your new suggestion. :+1: 

However, this exposed errors in litegraph's ExecutableNodeDTO.ts trying to resolve inputs to the group node. These issues are solvable, but I do not think them a good use of development time:
- As group nodes have never worked inside subgraphs, trying to fix this becomes a new feature instead of patching a regression.
- If users are creating new workflows with group nodes, we should instead be encouraging them to convert those group nodes to subgraphs instead of creating a high risk patchwork of interconnected fixes.

christian-byrne:
Changes:

- Added validation in `resolveInput()` that detects when `this.id.split(':').length > 2` and throws a clear error message directing users to convert to subgraphs instead.

The fix now:

- Resolves the original group node execution error for main graph
- Provides clear error for unsupported subgraph nesting

AustinMroz (approved):

PR #4924, opened 2025-08-11
[feat] Add manager capability feature flags

christian-byrne:
## Summary
- Added `managerApiVersion` and `supportsManagerV4` feature flags to support manager capability detection
- Updated tests to verify the new flags work correctly
- Added documentation explaining the pattern of using command-line args for configuration and feature flags for capabilities

## Manager Compatibility Pattern
The manager uses a two-tier approach:
1. **Command-line args** control user configuration:
   - `--disable-manager` - Manager is disabled
   - `--enable-manager-legacy-ui` - Use legacy UI

2. **Feature flags** report backend capabilities:
   - `api_version`: 'v1'  < /dev/null |  'v2' - Which API version the manager supports
   - `supports_v4`: boolean - Whether v4 features are available

## Implementation
Frontend can now check capabilities via:
```typescript
const { flags } = useFeatureFlags()
if (flags.managerApiVersion === 'v2' && flags.supportsManagerV4) {
  // Use new manager features
}
```

## Backend Requirements
The backend needs to send these flags in the WebSocket feature_flags message:
```json
{
  "extension": {
    "manager": {
      "api_version": "v1" or "v2",
      "supports_v4": true/false
    }
  }
}
```

Fixes #4823

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4924-feat-Add-manager-capability-feature-flags-24c6d73d36508161b85cefeaa2ea6dfd) by [Unito](https://www.unito.io)

PR #5063, opened 2025-08-17
[feat] Add manager capability feature flags

christian-byrne:
## Summary
- Adds support for manager-specific feature flags in the `useFeatureFlags` composable
- Follows the proper pattern for extension feature flags using `extension.manager.*` namespace

## Changes
- Added `MANAGER_SUPPORTS_V4` flag to check if manager supports v4 features
- Added `MANAGER_API_VERSION` flag to get the manager API version
- Added corresponding reactive accessors in the composable
- Added comprehensive tests for the new flags

## Context
This is a clean recreation of PR #4924 which had rebase issues. The functionality is identical but applied cleanly to the current `manager/menu-items-migration` branch.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5063-feat-Add-manager-capability-feature-flags-2526d73d365081a0b2e8ee72cfa082a3) by [Unito](https://www.unito.io)


src/stores/managerStateStore.ts
```diff
@@ -0,0 +1,59 @@
...
+    const systemStats = useSystemStatsStore().systemStats
+    const { flags } = useFeatureFlags()
+    const clientSupportsV4 =
+      api.getClientFeatureFlags().supports_manager_v4_ui ?? false
+
+    // Check command line args first
+    if (systemStats?.system?.argv?.includes('--disable-manager')) {
+      managerUIState.value = ManagerUIState.DISABLED
+    } else if (
+      systemStats?.system?.argv?.includes('--enable-manager-legacy-ui')
```
christian-byrne, line 44: Here we still need to check if the legacy manager UI is even available to account for scenario:

User updates such that this arg `--enable-manager-legacy-ui` exists but has not installed dependencies so the manager pip package is not installed. In that case, go to disabled.

src/views/GraphView.vue
```diff
@@ -250,6 +251,13 @@ void nextTick(() => {
   versionCompatibilityStore.initialize().catch((error) => {
     console.warn('Version compatibility check failed:', error)
   })
+
+  // Initialize manager state after system stats are loaded
+  useManagerStateStore()
+    .initializeManagerState()
+    .catch((error) => {
+      console.warn('Manager state initialization failed:', error)
+    })
```
christian-byrne: This should not block the startup

package.json
```diff
@@ -41,6 +41,7 @@
     "@trivago/prettier-plugin-sort-imports": "^5.2.0",
     "@types/dompurify": "^3.0.5",
     "@types/fs-extra": "^11.0.4",
+    "@types/lodash": "^4.17.20",
```
viva-jinyi: Don't we use es-toolkit instead of lodash?

viva-jinyi (changes requested):

christian-byrne:
Thanks for the review @viva-jinyi\! You're absolutely right - we use es-toolkit instead of lodash. I've removed the unnecessary @types/lodash dependency that was added by mistake.

Also addressed the previous review comments from @christian-byrne:
- Added check for legacy manager availability when --enable-manager-legacy-ui is set
- Implemented lazy initialization to avoid blocking startup

The PR should be ready for another review now.

christian-byrne:
This was merged into https://github.com/Comfy-Org/ComfyUI_frontend/pull/5291

PR #5069, opened 2025-08-18
[refactor] Migrate minimap to domain-driven renderer architecture

christian-byrne:
## Summary
- Refactored minimap implementation to follow domain-driven design principles
- Moved from monolithic composable to modular renderer architecture
- Added unit tests for each file

Selection_2006


christian-byrne:
Merging this since it is a major refactor and I don't want to cause conflicts.

PR #5077, opened 2025-08-18
fix: Make bottom panel tab titles reactive to language changes

benceruleanlu:
## Summary
- Fixed bottom panel tab titles and command menu bar `extraMenuItems` not updating when switching languages
- Tab titles and command menu bar now properly reflect the selected language immediately upon switching

## Changes
- Added `titleKey` field to `BottomPanelExtension` interface to store i18n keys instead of pre-translated strings
- Updated core tab definitions (`useTerminalTabs`, `useShortcutsTab`) to use `titleKey` 
- Modified `BottomPanel.vue` to dynamically translate titles using `$t()` function
- Added locale key to `Tabs` component to force proper re-rendering of tab underlines on language change
- Maintained backward compatibility for extensions that still use static `title` strings
- Converted extraMenuItems to computed

https://github.com/user-attachments/assets/fa2edc96-bc47-4bcf-a3d1-1ed8b3e1ba16

fixes #5039 

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5077-fix-Make-bottom-panel-tab-titles-reactive-to-language-changes-2536d73d365081edab98cc93055fecfb) by [Unito](https://www.unito.io)


src/components/bottomPanel/BottomPanel.vue
```diff
@@ -13,8 +16,13 @@
               <span class="font-bold">
                 {{
                   shouldCapitalizeTab(tab.id)
-                    ? tab.title.toUpperCase()
-                    : tab.title
+                    ? (tab.titleKey
```
Myestery: We could use a computed value for better performance

benceruleanlu:
Thanks for the review @Myestery! I've implemented your suggestion in commit 5a85d8c1a.

I extracted the complex nested ternary logic into a `getTabDisplayTitle` helper method. I went with a simple helper method rather than a computed property since:
- The calculation is inexpensive (just string manipulation and i18n lookup)
- Vue's template reactivity already handles re-rendering when dependencies change
- It keeps the solution simple while significantly improving code readability

The template is now much cleaner with just `{{ getTabDisplayTitle(tab) }}` instead of the nested ternaries.

*Reply generated with Claude Code*

src/components/bottomPanel/BottomPanel.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <div class="flex flex-col h-full">
-    <Tabs v-model:value="bottomPanelStore.activeBottomPanelTabId">
+    <Tabs
+      :key="$i18n.locale"
```
christian-byrne, line 4: that's clever

christian-byrne: Although I feel a very small number of readers will understand what is happening there.

christian-byrne (approved):

PR #5106, opened 2025-08-19
Fix/widget ordering consistency

### LINKED ISSUE #4983: Node widget values corrupted when input order differs due to JSON key ordering

simula-r:
## Summary

Problem: Different backends can serialize node inputs in different orders when sending node definitions. Since the frontend previously created widgets based on the order inputs were positioned in the response, the same node could have widgets in different orders across backends. This breaks workflow compatibility because widgets_values is an array indexed by position.

Solution: Use the input_order field from node definitions to enforce consistent widget ordering across all backends. When input_order is not provided, fall back to the previous behavior for backward compatibility.

## Changes

  1. nodeSchema.ts (src/schemas/nodeDefSchema.ts:241)
    - Added input_order as optional field to track backend-specified ordering
  2. nodeDefStore.ts (src/stores/nodeDefStore.ts:69,137)
    - Store and pass through input_order from backend node definitions
    - Made available in both V1 and V2 formats
  3. nodeDefOrderingUtil.ts (src/utils/nodeDefOrderingUtil.ts)
    - Created getOrderedInputSpecs() to order V2 format inputs according to input_order
    - Handles backward compatibility when input_order is missing
    - Created sortWidgetValuesByInputOrder() for reordering existing widget values
  4. litegraphService.ts (src/services/litegraphService.ts)
    - Updated #addInputs() method to use getOrderedInputSpecs()
    - Ensures widgets are always created in deterministic order
    - Applied to both regular nodes and subgraph nodes
  5. Tests (tests-ui/tests/utils/nodeDefOrderingUtil.test.ts)
    - Comprehensive test coverage including real KSampler node example
    - Tests backward compatibility and edge cases

Fixes: https://github.com/Comfy-Org/ComfyUI_frontend/issues/4983

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5106-Fix-widget-ordering-consistency-2546d73d3650812687d3c556feaa55cb) by [Unito](https://www.unito.io)

### DIFF
--- a/src/utils/nodeDefOrderingUtil.ts
+++ b/src/utils/nodeDefOrderingUtil.ts
@@ -1,86 +1,54 @@
-import { ComfyNodeDef } from '@/schemas/nodeDefSchema'
+import type { InputSpec } from '@/schemas/nodeDef/nodeDefSchemaV2'
+import type { ComfyNodeDefImpl } from '@/stores/nodeDefStore'
 
 /**
- * Sorts the inputs of a node definition according to the input_order property.
- * If input_order is not provided, returns inputs unchanged (maintains backward compatibility).
+ * Gets an ordered array of InputSpec objects based on input_order.
+ * This is designed to work with V2 format used by litegraphService.
  *
- * @param nodeDef - The node definition containing inputs and potentially input_order
- * @returns A new object with sorted inputs, or the original if no sorting needed
+ * @param nodeDefImpl - The ComfyNodeDefImpl containing both V1 and V2 formats
+ * @param inputs - The V2 format inputs (flat Record<string, InputSpec>)
+ * @returns Array of InputSpec objects in the correct order
  */
-export function sortNodeInputsByOrder(nodeDef: ComfyNodeDef): ComfyNodeDef {
-  // If no input_order or no inputs, return as-is
-  if (!nodeDef.input_order || !nodeDef.input) {
-    return nodeDef
+export function getOrderedInputSpecs(
+  nodeDefImpl: ComfyNodeDefImpl,
+  inputs: Record<string, InputSpec>
+): InputSpec[] {
+  const orderedInputSpecs: InputSpec[] = []
+
+  // If no input_order, return default Object.values order
+  if (!nodeDefImpl.input_order) {
+    return Object.values(inputs)
   }
 
-  const sortedNodeDef = { ...nodeDef }
-
-  // Sort each input category (required, optional, hidden)
-  if (sortedNodeDef.input) {
-    sortedNodeDef.input = { ...sortedNodeDef.input }
-
-    // Process each category that has an order defined
-    for (const category of ['required', 'optional', 'hidden'] as const) {
-      const order = nodeDef.input_order[category]
-      const inputs = sortedNodeDef.input[category]
-
-      if (order && inputs) {
-        // Create a new sorted object based on the order
-        const sortedInputs: Record<string, any> = {}
-
-        // First, add inputs in the specified order
-        for (const inputName of order) {
-          if (inputName in inputs) {
-            sortedInputs[inputName] = inputs[inputName]
-          }
-        }
-
-        // Then add any remaining inputs not in the order (for safety)
-        for (const [inputName, inputSpec] of Object.entries(inputs)) {
-          if (!(inputName in sortedInputs)) {
-            sortedInputs[inputName] = inputSpec
-          }
-        }
-
-        sortedNodeDef.input[category] = sortedInputs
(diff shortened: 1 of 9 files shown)


src/utils/nodeDefOrderingUtil.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Reorders widget values based on the input_order to match expected widget order.
+ * This is used when widgets were created in a different order than input_order specifies.
+ *
+ * @param widgetValues - The current widget values array
+ * @param currentWidgetOrder - The current order of widget names
+ * @param inputOrder - The desired order from input_order
+ * @returns Reordered widget values array
+ */
+export function sortWidgetValuesByInputOrder(
+  widgetValues: any[],
```
christian-byrne: Does it still compile if using `TWidgetValue` in place of all the `any` types?

christian-byrne (commented):
Looks good, just one minor question about types

src/utils/nodeDefOrderingUtil.ts
simula-r: yes indeed, just had to export TWidgetValue from lib/litegraph

christian-byrne: @simula-r In that case we may be able to use it for the other `any` in this file as well

simula-r: ah yeah i had a straggler, fixed all anys now.

christian-byrne (approved):
Nice!

Selection_2018

PR #5109, opened 2025-08-19
Fix CopyToClipboard Issue

### LINKED ISSUE #5036: [Bug]: copy URL button doesn't work when ComfyUI is hosted on Linux
### Prerequisites

### What happened?

TLDR:
ComfyUI frontend is v1.23.4 in both systems. When served/hosted on Linux, the copy button fails. When served hosted on Windows, the copy button works. Python 3.12.10 in both cases. Because these buttons are not actual hyperlinks, you cannot right-click to copy the URLs, either.

Longer:
On Windows 11 running Comfy natively with the standalone download, Python 3.12.10 and Comfy v0.3.48 still have a functional copy URL button. I recently built a dedicated rig for running Comfy, and that is running Fedora. This issue is being reported with Python 3.12.11 being run via uv in a venv.

I then downgraded Linux Python to 3.12.10 and comfy v0.3.48 with frontend v1.23.4 and still have the same issue with trying to use the "copy URL" button --

The toast notification pops up, but pasting to the clipboard contains only a single empty character.

When I click the button, I notice in the browser console that I see this message:
Autofocus processing was blocked because a document already has a focused element.

When using the Windows-python-hosted ComfyUI, where the copy button works, that error does not persist.

So something is wonky specifically with ComfyUI running on Linux and being accessed from Windows.

Tested with Chrome on Linux and Comfy running on Linux with the same result -- the copy button pops the toast notification but does not get anything into the clipboard. The browser console spits the same message:
Autofocus processing was blocked because a document already has a focused element.

### Steps to Reproduce

1. Open a comfy workflow with missing models
2. The modal pops up with buttons to "copy URL"
3. Click the button

### ComfyUI Frontend Version

1.23.4

### Browser

Chrome/Chromium

### Console Errors

```javascript
`Autofocus processing was blocked because a document already has a focused element.`
```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-5036-Bug-copy-URL-button-doesn-t-work-when-ComfyUI-is-hosted-on-Linux-2516d73d3650811ca008ff7633735a54) by [Unito](https://www.unito.io)

Myestery:
## Summary

Enhanced clipboard functionality with fallback support to address clipboard API failures in secure contexts and various browser environments.

## Changes

- **What**: Improved `useCopyToClipboard` composable with fallback mechanism using `document.execCommand('copy')` when VueUse clipboard API fails or is unsupported
- **Dependencies**: None - uses existing VueUse and PrimeVue dependencies

## Review Focus

- **Fallback Strategy**: The implementation now gracefully falls back to the legacy `document.execCommand('copy')` method when the modern Clipboard API fails
- **Error Handling**: Comprehensive error handling with user-friendly toast notifications for both success and failure scenarios
- **Cross-browser Compatibility**: Ensures clipboard functionality works across different browsers and security contexts

## Technical Details

The enhancement addresses scenarios where:
1. VueUse clipboard API is not supported in the current environment
2. VueUse copy operation fails (e.g., in insecure contexts or older browsers)
3. Permission issues with the Clipboard API

The fallback method creates a temporary textarea element, selects the text, and uses the legacy copy command, providing a reliable alternative when modern APIs fail.

## Review Focus

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/5036

https://github.com/user-attachments/assets/1da022b4-d55e-4679-99a0-9ab0f62f3e9c


src/composables/useCopyToClipboard.ts
```diff
@@ -4,32 +4,68 @@ import { useToast } from 'primevue/usetoast'
...
+    }
+  }
+
+  const copyToClipboard = async (text: string) => {
+    if (isSupported) {
+      try {
+        await copy(text)
+
+        // Check if copy was successful
+        if (copied.value) {
```
arjansingh: To avoid duplication can we do the toast in a finally block and key off the success for failure of the try catch.

That way we only have two toast  branches (success and error) instead of four.

src/composables/useCopyToClipboard.ts
```diff
@@ -4,32 +4,68 @@ import { useToast } from 'primevue/usetoast'
...
-        detail: t('clipboard.errorNotSupported')
+        detail: t('clipboard.errorMessage')
       })
+    } finally {
+      document.body.removeChild(textarea)
+    }
+  }
+
+  const copyToClipboard = async (text: string) => {
+    if (isSupported) {
```
arjansingh: If we're doing a try/catch, do we still need this if/else?

You can simplify this to be one try catch or one if.

arjansingh:
@Comfy-Org/comfy_frontend_devs Reviewing this.

src/composables/useCopyToClipboard.ts
```diff
@@ -4,32 +4,68 @@ import { useToast } from 'primevue/usetoast'
...
+      const successful = document.execCommand('copy')
+      if (successful) {
         toast.add({
           severity: 'success',
           summary: t('g.success'),
           detail: t('clipboard.successMessage'),
           life: 3000
         })
-      } catch (err) {
+      } else {
```
arjansingh, line 43: again if we move toasts to the `finally` in `copyToClipboard` we can remove a lot of these duplicate toast blocks.

src/composables/useCopyToClipboard.ts
```diff
@@ -4,32 +4,68 @@ import { useToast } from 'primevue/usetoast'
...
 
-  const copyToClipboard = async (text: string) => {
-    if (isSupported) {
-      try {
-        await copy(text)
+  function fallbackCopy(text: string) {
+    const textarea = document.createElement('textarea')
+    textarea.value = text
+    textarea.style.position = 'fixed'
+    textarea.style.opacity = '0'
```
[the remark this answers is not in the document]
arjansingh: I agree with Claude here. that's the proper way to do it. I was mentioning it in a separate review comment but i'll second Claude here.

src/composables/useCopyToClipboard.ts
```diff
@@ -4,32 +4,68 @@ import { useToast } from 'primevue/usetoast'
...
+    document.body.appendChild(textarea)
```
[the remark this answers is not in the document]
arjansingh: Ditto. Agree with claud on this.

src/composables/useCopyToClipboard.ts
```diff
@@ -4,32 +4,64 @@ import { useToast } from 'primevue/usetoast'
...
+  const showErrorToast = () => {
+    toast.add({
+      severity: 'error',
+      summary: t('g.error'),
+      detail: t('clipboard.errorMessage')
+    })
+  }
+
+  function fallbackCopy(text: string) {
+    const textarea = document.createElement('textarea')
```
arjansingh, line 26: `textarea.setAttribute('readonly', '')` we need this also to not accidentally trigger virtual keyboards.

AustinMroz:
At least on my linux system, I've tried multiple browsers and was unable to reproduce the issue on either. I'm still digging into things, but the issue has me quite perplexed.

The original report mentions that they get a toast indicating success and the contents of the clipboard does change. The wording isn't the most precise, but I'm not sure if the fallback handler implemented here would even run for the person reporting this issue.

arjansingh (changes requested):
@Myestery Got a couple suggestions for cleaning up the code. Please see feedback.

Before you look at my suggestions though, could you check out a similar error message discussed in the Nuxt repo? 

We might just need to make sure the element getting copied has focus before copying:
https://github.com/nuxt/ui/issues/1350

@christian-byrne would tests be good for this?

Myestery:
> At least on my linux system, I've tried multiple browsers and was unable to reproduce the issue on either. I'm still digging into things, but the issue has me quite perplexed.
> 
> The original report mentions that they get a toast indicating success and the contents of the clipboard does change. The wording isn't the most precise, but I'm not sure if the fallback handler implemented here would even run for the person reporting this issue.

I was able to replicate the issue on a linux container and it got fixed there

Myestery:
> @Myestery Got a couple suggestions for cleaning up the code. Please see feedback.
> 
> Before you look at my suggestions though, could you check out a similar error message discussed in the Nuxt repo?
> 
> We might just need to make sure the element getting copied has focus before copying: [nuxt/ui#1350](https://github.com/nuxt/ui/issues/1350)
> 
> @christian-byrne would tests be good for this?

I'm not able to replicate the focus issue however relying on the copied.value, we are able to check for all scenarios.
For my test, it didn't copy but fail silently. This may have a different behavior across browsers

arjansingh:
> relying on the copied.value, we are able to check for all scenarios.
> For my test, it didn't copy but fail silently. This may have a different be

Alright then don't worry about it. Let's just clean up your existing implementation then.

src/composables/useCopyToClipboard.ts
Myestery: since finally runs in failure and success, the current implementation is fine

src/composables/useCopyToClipboard.ts
Myestery: having that we might still have to do some checks in finally, this should suffice

arjansingh (approved):
Good to go!

christian-byrne (commented):
I don't think the element needs to be inserted into the DOM or have style attributes for this to work. Did you try?

Myestery:
> I don't think the element needs to be inserted into the DOM or have style attributes for this to work. Did you try?

Just tested, doesn't work without insertion

PR #5115, opened 2025-08-20
Fix: Shift+Click+Drag from outputs with Subgraph outputs

DrJKL:
## Summary

A patch for now that addresses an issue in synchronizing the output slot's internal links and the Subgraph (as LinkNetwork)'s links map when connecting to or disconnecting from the Subgraph's outputs.

## Changes

- **What**: Node outputs that were connected to subgraph outputs before lost support for shift+click+drag. Now they behave as expected.

## Review Focus

Fixes #4877 

## Demonstration

https://github.com/user-attachments/assets/9bd1e719-4808-46a8-9de8-2ac16279efb3

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5115-Fix-Shift-Click-Drag-from-outputs-with-Subgraph-outputs-2556d73d3650817fb0bdefedd68cba75) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -2585,9 +2585,21 @@ export class LGraphCanvas
               e.shiftKey &&
               (output.links?.length || output._floatingLinks?.size)
             ) {
-              linkConnector.moveOutputLink(graph, output)
-              this.#linkConnectorDrop()
-              return
+              const outputLinks = [
+                ...(output.links ?? []),
+                ...[...(output._floatingLinks ?? new Set())]
+              ]
+              if (
+                outputLinks.some(
+                  (linkId) =>
+                    typeof linkId === 'number' &&
+                    graph.getLink(linkId) !== undefined
+                )
+              ) {
+                linkConnector.moveOutputLink(graph, output)
+                this.#linkConnectorDrop()
+                return
+              }
             }
 
             // New output link
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -36,6 +36,7 @@ import type {
   INodeSlot,
   INodeSlotContextItem,
   ISlotType,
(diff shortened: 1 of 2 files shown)


christian-byrne (dismissed):

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -2575,16 +2576,27 @@ export class LGraphCanvas
...
+        network: LinkNetwork
+      ): boolean {
+        const outputLinks = [
+          ...(output.links ?? []),
+          ...[...(output._floatingLinks ?? new Set())]
+        ]
+        return outputLinks.some(
+          (linkId) =>
+            typeof linkId === 'number' && network.getLink(linkId) !== undefined
+        )
```
webfiltered, line 2590: This change appears to discard errors caused by link corruption.  Is this intentional?  Not sure the state this is resolving, but it seems like there's a root cause elsewhere.

DrJKL: Yeah, the code that it mostly copies is for a single new connection, where we'd want to error, but in here, we have some state management issues that lead to stale references.
By continuing instead of noping out, we can move the remaining valid links.

DrJKL: The root cause is that we lack a single source of truth for links. The output slots have references, but so does the graph itself.

webfiltered: What happens to the stale references, though?  If we just ignore them, the stale ID stays on the output slot right?  Unless we adopt a "ignore stale links everywhere" approach (including a requirement for any extensions), it is likely we'll end up just pushing the issue somewhere else.

DrJKL: Yeah, we definitely need to address the links' state issue more fundamentally. But for now, this just stems the bleeding.

christian-byrne (approved):
LGTM!

PR #5128, opened 2025-08-20
Cleanup: Potpourri PR of small changes that reduce the warning noise

DrJKL:
## Summary

Potpourri PR of small changes that reduce the warning noise

## Changes

- **What**:
  - A compiler macro that doesn't need to be imported
  - Two components with the same name that need to be differentiated for Vite
  - LF (\n) EOL sequences for the qualified versions of JS/TS files.

## Review Focus

Let me know what other warnings you are too used to seeing that could be addressed.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5128-Cleanup-Potpourri-PR-of-small-changes-that-reduce-the-warning-noise-2556d73d365081b1a984ee7f89e5defd) by [Unito](https://www.unito.io)

### DIFF
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -136,7 +136,8 @@ export default defineConfig({
       ],
       dirs: ['src/components', 'src/layout', 'src/views'],
       deep: true,
-      extensions: ['vue']
+      extensions: ['vue'],
+      directoryAsNamespace: true
     })
   ],
 
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,9 +2,13 @@
 * text=auto
 
 # Force TS to LF to make the unixy scripts not break on Windows
+*.cjs text eol=lf
+*.js text eol=lf
+*.json text eol=lf
+*.mjs text eol=lf
+*.mts text eol=lf
 *.ts text eol=lf
 *.vue text eol=lf
-*.js text eol=lf
 
 # Generated files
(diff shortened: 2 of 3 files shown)


arjansingh (approved):

christian-byrne:
Rebasing to apply test fixes!

vite.config.mts
```diff
@@ -136,7 +136,8 @@ export default defineConfig({
       ],
       dirs: ['src/components', 'src/layout', 'src/views'],
       deep: true,
-      extensions: ['vue']
+      extensions: ['vue'],
+      directoryAsNamespace: true
```
christian-byrne, line 140: I guess this would be a major change if not for the fact that we don't use uplugin-components and typically still import explicitly.

christian-byrne: I'm wondering if there's any cases of us not importing a component though and if there's a way to check/verify with one of our linter rules.

DrJKL: If we're not using it, should we get rid of it?

christian-byrne: I'm not 100% positive we are not using it, there may be some locations where it's used.

DrJKL: No new build errors, all the vitest tests pass, App seems to be working as expected.
If I missed something and it breaks later, I will accept the shame.

PR #5137, opened 2025-08-20
[fix] Correct WhatsNew popup arrow alignment with help center icon

christian-byrne:
Fixes the WhatsNew popup arrow positioning that was misaligned after new sidebar icons were added below the help center icon. The arrow now correctly points to the help center icon by accounting for the additional icons in its positioning calculation.

Fixes #5126

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5137-fix-Correct-WhatsNew-popup-arrow-alignment-with-help-center-icon-2556d73d36508188b9c4ca88e71c8d1d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/helpcenter/WhatsNewPopup.vue
+++ b/src/components/helpcenter/WhatsNewPopup.vue
@@ -186,7 +186,8 @@ defineExpose({
 .whats-new-popup-container.sidebar-left.small-sidebar .help-center-arrow {
   left: -14px; /* Overlap with popup outline */
   bottom: calc(
-    var(--sidebar-width) * 2 + 10px - var(--whats-new-popup-bottom)
+    var(--sidebar-width) * 2 + var(--sidebar-icon-size) / 2 -
+      var(--whats-new-popup-bottom)
   ); /* Position to center of help center icon (2 icons below + half icon height for center - whats new popup bottom position ) */
 }
 
--- a/src/components/helpcenter/WhatsNewPopup.vue
+++ b/src/components/helpcenter/WhatsNewPopup.vue
@@ -159,8 +159,10 @@ defineExpose({
 <style scoped>
 /* Popup container - positioning handled by parent */
 .whats-new-popup-container {
+  --whats-new-popup-bottom: 1rem;
+
   position: absolute;
-  bottom: 1rem;
+  bottom: var(--whats-new-popup-bottom);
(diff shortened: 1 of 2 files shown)


benceruleanlu (dismissed):

src/components/helpcenter/WhatsNewPopup.vue
```diff
@@ -171,8 +169,8 @@ defineExpose({
 .help-center-arrow {
   position: absolute;
   bottom: calc(
-    var(--sidebar-width, 4rem) + 0.25rem
-  ); /* Position toward center of help center icon */
+    var(--sidebar-width, 4rem) * 2 + 0.25rem
+  ); /* Position toward center of help center icon (accounting for 2 icons below) */
```
viva-jinyi: `.whats-new-popup-container.sidebar-left.small-sidebar .help-center-arrow` The CSS further down has a higher priority, so this one doesn’t get applied first. And you also need to take the --p-button-padding-y padding value into account to find the correct position.
스크린샷 2025-08-21 오후 1 01 59

viva-jinyi (changes requested):

src/components/helpcenter/WhatsNewPopup.vue
christian-byrne: Thank you for the help, let me fix tomorrow

christian-byrne:
@viva-jinyi Can you finish this?

viva-jinyi:
스크린샷 2025-08-26 오후 1 40 07
스크린샷 2025-08-26 오후 1 40 13
@christian-byrne Since the WhatsNew popup was teleported to #graph-canvas-container, it couldn’t access the sidebar CSS variable (--sidebar-width). Moved the --sidebar-width CSS variable to :root so it can be accessed globally.

viva-jinyi (dismissed):

christian-byrne:
Do we actually need to pollute global namespace to do that? Can't it be a prop or injected?

viva-jinyi:
@christian-byrne Since the component is teleported elsewhere, we can’t use props or injections. In the current situation, handling it with a CSS variable seems like the fastest and most manageable approach. And since the sidebar width variable is something that can be used at the layout level anyway, I think it’s fine to make it global. WDYT?

PR #5139, opened 2025-08-21
Implement subgraph publishing

AustinMroz:
Adds a system for saving blueprints of subgraphs for reuse later.  When a subgraph is selected, a new publish icon is added to the selection toolbar. Subgraph blueprints can by added by the node library or the node searchbox. When accessed from the library, options are provided for editing or deleting a subgraph.

These blueprints are saved to the user data folder and will persist across different browsers.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5139-Implement-subgraph-publishing-2566d73d365081b7a51ac6ed2345f215) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -3728,6 +3730,14 @@ export class LGraphCanvas
    */
   _pasteFromClipboard(
     options: IPasteFromClipboardOptions = {}
+  ): ClipboardPasteResult | undefined {
+    const data = localStorage.getItem('litegrapheditor_clipboard')
```
DrJKL, line 3860: https://vueuse.org/core/useLocalStorage/
You might be able to use the composable instead. Should help prevent typo based key misalignment.

DrJKL: I think it'd also help with the [serialize/deserialize](https://vueuse.org/core/useStorage/#custom-serialization)

src/scripts/app.ts
```diff
@@ -796,6 +797,7 @@ export class ComfyApp {
     this.resizeCanvas()
 
     await useWorkspaceStore().workflow.syncWorkflows()
+    await useSubgraphStore().fetchSubgraphs()
     await useExtensionService().loadExtensions()
```
DrJKL, line 802: Do these conflict with one another? If not, could we batch them with [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled)?

src/services/litegraphService.ts
```diff
@@ -948,6 +949,20 @@ export const useLitegraphService = () => {
...
+      const bp = useSubgraphStore().getBlueprint(nodeDef.name)
+      const items = { nodes: bp.nodes, subgraphs: bp.definitions?.subgraphs }
+      //FIXME
+      const strippedItems = JSON.parse(JSON.stringify(items))
+      const results = canvas._deserializeItems(strippedItems, {
+        position: options.pos
+      })
+      if (!results) throw new Error('Failed to add subgraph blueprint')
+      const node = results.nodes.values().next().value
+      if (!node) throw new Error('Failed to add subgraph blueprint')
```
DrJKL: It might be helpful to differentiate between the ways this is failing, if a consumer only looks at the message and not the stacktrace.

src/services/litegraphService.ts
```diff
@@ -948,6 +949,20 @@ export const useLitegraphService = () => {
     options: Record<string, any> = {}
   ): LGraphNode {
     options.pos ??= getCanvasCenter()
+    if (nodeDef.name.startsWith('SubgraphBlueprint')) {
+      const canvas = canvasStore.getCanvas()
+      const bp = useSubgraphStore().getBlueprint(nodeDef.name)
+      const items = { nodes: bp.nodes, subgraphs: bp.definitions?.subgraphs }
+      //FIXME
```
DrJKL: //FIXME

src/stores/nodeDefStore.ts
```diff
@@ -291,8 +292,14 @@ export const useNodeDefStore = defineStore('nodeDef', () => {
   const showDeprecated = ref(false)
   const showExperimental = ref(false)
   const nodeDefFilters = ref<NodeDefFilter[]>([])
+  const subgraphStore = useSubgraphStore()
 
-  const nodeDefs = computed(() => Object.values(nodeDefsByName.value))
+  const nodeDefs = computed(() => {
+    return [
+      ...Object.values(nodeDefsByName.value),
```
DrJKL, line 299:
```suggestion
      ...Object.values(nodeDefsByName),
```

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+  hasPromptedSave: boolean = false
+
+  validateSubgraph() {
+    if (!this.activeState?.definitions)
+      throw new Error(
+        'The root graph of a subgraph blueprint must consist of only a single subgraph node'
+      )
+    const subgraphs = this.activeState.definitions.subgraphs
+    const nodes = this.activeState.nodes
+    //Instanceof doesn't funciton as nodes are serialized
```
DrJKL: Could we use Zod to validate?

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+
+  hasPromptedSave: boolean = false
+
+  validateSubgraph() {
+    if (!this.activeState?.definitions)
+      throw new Error(
+        'The root graph of a subgraph blueprint must consist of only a single subgraph node'
+      )
+    const subgraphs = this.activeState.definitions.subgraphs
+    const nodes = this.activeState.nodes
```
DrJKL: If you're feeling fancy...
```suggestion
    const { subgraphs } = this.activeState.definitions
    const { nodes } = this.activeState
```

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+    throw new Error(
+      'The root graph of a subgraph blueprint must consist of only a single subgraph node'
+    )
+  }
+
+  override async save(): Promise<UserFile> {
+    this.validateSubgraph()
+    if (!this.hasPromptedSave) {
+      //Swap to saveAs?
+      const newname = await useDialogService().prompt({
```
DrJKL: `newName`?

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+    useSubgraphStore().updateDef(this as LoadedComfyWorkflow)
+    return super.save()
+  }
+
+  //TODO: reconsider semantics here with promptonce
+  //Should update current name, but keep true?
+  override async saveAs(path: string) {
+    this.validateSubgraph()
+    this.hasPromptedSave = true
+    useSubgraphStore().updateDef(this as LoadedComfyWorkflow)
```
DrJKL: Why do you need to cast?

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+      res.map(async (f) => {
+        const name = f.path.slice(0, -5)
+        f.path = SubgraphBlueprint.basePath + f.path
+        const bp = await new SubgraphBlueprint(f).load()
+        useWorkflowStore().attachWorkflow(bp)
+        const nodeDef = convertToNodeDef(bp)
+
+        subgraphDefCache.value.set(name, nodeDef)
+        subgraphCache[name] = bp
+      })
```
DrJKL: It might be cleaner to extract a function here.

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+    )
+  }
+  function convertToNodeDef(workflow: LoadedComfyWorkflow): ComfyNodeDefImpl {
+    const name = workflow.filename
+    const subgraphNode = workflow.changeTracker.initialState.nodes[0]
+    if (!subgraphNode) throw new Error('Invalid Subgraph Blueprint')
+    const uuid = subgraphNode.type
+    if (!uuid) throw new Error('')
```
DrJKL: Empty Error message

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+    subgraphNode.inputs ??= []
+    subgraphNode.outputs ??= []
+    //NOTE: Types are cast to string. This is only used for input coloring on previews
+    const inputs = Object.fromEntries(
+      subgraphNode.inputs.map((i) => [
+        i.name,
+        ['' + i.type, undefined] satisfies InputSpec
```
DrJKL: Very nit
```suggestion
        [`${i.type}`, undefined] satisfies InputSpec
```

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+      ])
+    )
+    const nodedefv1: ComfyNodeDefV1 = {
+      input: { required: inputs },
+      output: subgraphNode.outputs.map((o) => '' + o.type),
```
DrJKL: Same here, unless there's a style guide rule that string casts are done by concatenating `''`

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+    }
+    //prompt name
+    const name = await dialogService.prompt({
+      title: t('subgraphStore.saveBlueprint'),
+      message: t('subgraphStore.blueprintName') + ':',
+      defaultValue: subgraphNode.title
+    })
+    if (!name) return
+    if (subgraphDefCache.value.has(name))
+      if (!(await confirmOverwrite(name)))
```
DrJKL:
```suggestion
    if (subgraphDefCache.value.has(name) && !(await confirmOverwrite(name)))
```

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,254 @@
...
+    void useWorkflowService().openWorkflow(subgraphCache[name])
+  }
+  function getBlueprint(nodeType: string): ComfyWorkflowJSON {
+    const name = nodeType.slice(typePrefix.length)
+    if (!(name in subgraphCache))
+      //As loading is blocked on in startup, this can likely be changed to invalid type
+      throw new Error('not yet loaded')
+    return subgraphCache[name].changeTracker.initialState
+  }
+  function deleteBlueprint(nodeType: string) {
```
DrJKL: Would you mind spacing out the functions in here?

src/stores/workflowStore.ts
```diff
@@ -309,6 +310,20 @@ export const useWorkflowStore = defineStore('workflow', () => {
     const fullPath = getUnconflictedPath(
       ComfyWorkflow.basePath + (path ?? 'Unsaved Workflow.json')
     )
+    //TODO: Reconsider. Referencing subgraph constants outside
```
DrJKL: Agreed.

src/stores/workflowStore.ts
```diff
@@ -357,7 +372,10 @@ export const useWorkflowStore = defineStore('workflow', () => {
   }
 
   const persistedWorkflows = computed(() =>
-    Array.from(workflows.value).filter((workflow) => workflow.isPersisted)
+    Array.from(workflows.value).filter(
```
DrJKL, line 375:
```suggestion
    Array.from(workflows).filter(
```

DrJKL (commented):
Looks good, mostly just code style comments 👍🏻

src/scripts/app.ts
AustinMroz: I don't think extensions can access the workflowStore or the subgraphStore, but it's not something I'd want to risk.
On the other hand, I hadn't realized Promise.All was sequential. Will have to fix the incorrect usage of it inside fetchSubgraphs()

src/stores/subgraphStore.ts
AustinMroz: There's an unfortunate multiple inheritance issue here. I would be most happy if I could use isLoaded as a type guard, but I can safely swap this to
```ts
useSubgraphStore().updateDef(await this.load())
```
to minimize the number of casts.

src/stores/subgraphStore.ts
AustinMroz: Just adding an empty newline, yeah? Same for all the functions?

src/stores/subgraphStore.ts
AustinMroz: I don't think Zod has a way of checking  the indirection required here. There's a couple other places with saving and loading blueprints where checks couldd be rolled into Zod,  but it kinda comlicates things with zod workflow validation being an optional setting. I wouldn't want the important checks to be skipped if a user has the workflow validation disabled.

christian-byrne:
We can merge this in 1.27 after the manager is merged, if that works.

AustinMroz:
Sounds like an appropriate goal. I got some additional feedback on menu design from Pablo earlier today which I'm working on, but everything else should be solid.

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    validateSubgraph() {
+      if (!this.activeState?.definitions)
+        throw new Error(
+          'The root graph of a subgraph blueprint must consist of only a single subgraph node'
+        )
+      const { subgraphs } = this.activeState.definitions
+      const { nodes } = this.activeState
+      //Instanceof doesn't funciton as nodes are serialized
```
christian-byrne: nit: Typo in word "function"

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,313 @@
...
+      function isSubgraphNode(node: ComfyNode) {
+        return node && subgraphs.some((s) => s.id == node.type)
```
christian-byrne: The loose equality `==` is intentional correct?

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,313 @@
...
+async function confirmOverwrite(name: string): Promise<boolean | null> {
+  return await useDialogService().confirm({
+    title: t('subgraphStore.overwriteBlueprintTitle'),
+    type: 'overwriteBlueprint',
+    message: t('subgraphStore.overwriteBlueprint'),
+    itemList: [name]
+  })
+}
+
+const subgraphCache: Record<string, LoadedComfyWorkflow> = {}
```
christian-byrne: nit: Is it necessary for this to be in the module scope? The pinia store should also be a singleton but has benefit of being lazily loaded, better encapsulate for testing, accessible in vue dev tools. That said, this is also perfectly fine in my opinion.

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,313 @@
...
+    }
+
+    const res = (
+      await api.listUserDataFullInfo(SubgraphBlueprint.basePath)
+    ).filter((f) => f.path.endsWith('.json'))
+    const settled = await Promise.allSettled(res.map(loadBlueprint))
+    settled
+      .filter((i) => 'reason' in i)
+      .forEach(({ reason }) =>
+        console.error('Failed to load subgraph blueprint', reason)
```
christian-byrne: nit: Should we show a toast?

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,313 @@
...
+      python_module: 'blueprint'
+    }
+    const nodeDefImpl = new ComfyNodeDefImpl(nodedefv1)
+    return nodeDefImpl
+  }
+  async function publishSubgraph() {
+    const canvas = canvasStore.getCanvas()
+    const subgraphNode = [...canvas.selectedItems][0]
+    if (
+      canvas.selectedItems.size != 1 ||
```
christian-byrne: Is loose inequality intentional here?

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,313 @@
...
+      !(subgraphNode instanceof SubgraphNode)
+    )
+      throw new TypeError('Must have single SubgraphNode selected to publish')
+
+    const { nodes = [], subgraphs = [] } = canvas._serializeItems([
+      subgraphNode
+    ])
+    if (nodes.length != 1) {
+      throw new TypeError('Must have single SubgraphNode selected to publish')
```
christian-byrne, line 218: nit: Should we surface the errors to the user with toasts then return instead of throwing? Or is it necessary to throw?

src/stores/subgraphStore.ts
```diff
@@ -0,0 +1,313 @@
...
+  const subgraphDefCache = ref<Map<string, ComfyNodeDefImpl>>(new Map())
+  const canvasStore = useCanvasStore()
+  const subgraphBlueprints = computed(() => [
+    ...subgraphDefCache.value.values()
+  ])
+  async function fetchSubgraphs() {
+    async function loadBlueprint(options: {
+      path: string
+      modified: number
+      size: number
```
christian-byrne, line 148: nit: the semantics of the `size` field is a little unclear to me on initial read (might be my fault) and the semantics of `size: 1` are somewhat non-obvious maybe.

src/components/sidebar/tabs/nodeLibrary/NodeTreeLeaf.vue
```diff
@@ -80,6 +110,33 @@ const sidebarLocation = computed<'left' | 'right'>(() =>
 const toggleBookmark = async () => {
   await nodeBookmarkStore.toggleBookmark(nodeDef.value)
 }
+const editBlueprint = async () => {
+  if (!props.node.data)
+    throw new Error(
+      'Failed to edit subgraph blueprint lacking backing node data'
+    )
```
christian-byrne, line 117: nit: Would it be better to show a toast and return without throwing?

christian-byrne (changes requested):
This is looking great.

To help me, is this the correct flow?

1. User clicks "Publish Subgraph" button → publishSubgraph() called
2. Get selected subgraph node → Validates exactly 1 SubgraphNode selected
3. Serialize the node → canvas._serializeItems([subgraphNode])
4. Extract data structures → Gets nodes[] and subgraphs[] arrays
5. Create workflow JSON → Builds a complete workflow structure
6. Save as file → Stores in subgraphs/filename.json
7. Register as node type → Makes it available in node library
8. Cache for editing → Stores loaded workflow for future edits

That is my understanding, want to make sure it's correct.

Also, I like the refactoring of the clipboard logic into distinct methods responsible for clipboard functionality and serialization individually. It makes me want to look into some further refactor to some sort of serialization service like

```typescript
export function useCanvasSerializationService() {
    function serializeItems(items: Iterable<Positionable>): ClipboardItems {
      // All serialization logic here
    }

    function serializeForClipboard(items: Iterable<Positionable>): void {
      const data = serializeItems(items)
      localStorage.setItem('litegrapheditor_clipboard', JSON.stringify(data))
    }

    function serializeForBlueprint(subgraphNode: SubgraphNode): WorkflowData {
      const data = serializeItems([subgraphNode])
      return formatAsWorkflow(data)
    }

    return { serializeItems, serializeForClipboard, serializeForBlueprint }
}
```

Just thinking out loud. Also, feel free to ignore the nits.

src/stores/subgraphStore.ts
AustinMroz: I generally lean towards loose equality, but this is already serialized data. Strict equality is absolutely the correct choice here.

src/stores/subgraphStore.ts
AustinMroz: Agreed. I'm thinking always a single toast that either displays filenames, or number of failed loads if > ~3. Will keep the error message for detailed logging.

src/stores/subgraphStore.ts
AustinMroz: This function is only called from the selectionbar, which should guarantee these as preconditions. Will leave as is.

src/stores/subgraphStore.ts
AustinMroz: Yeah... it tripped me up as well. The following "save" call updates the filesize to the correct value based on server response. Trying to calculate the filesize is a tricky endeavor due to utf-16 shenanigans. Setting this to -1 is a special keyword value that indicates the file is temporary and can not overwrite, but we can use a different placeholder value like -2. Can also add comments for clarity.

src/components/sidebar/tabs/nodeLibrary/NodeTreeLeaf.vue
AustinMroz: ~~Will test error case, but I lean towards probably~~
EDIT: Error should only only occur if edit is somehow called on a non-blueprint node. It does not cover errors in actual loading. Will leave this as is for now.

christian-byrne (approved):
LGTM!

PR #5140, opened 2025-08-21
Fix: Search Box Implementation for keyboard shortcut

### LINKED ISSUE #4122: [Bug]: `Workspce.SearchBox.Toggle` does not respect `Comfy.NodeSearchBoxImpl`
### Custom Node Testing

### Frontend Version

1.21.7

### Expected Behavior

Litegraph search box appears.

### Actual Behavior

Default search box appears.

### Steps to Reproduce

Bind some button to `Toggle Search Box`, set `Node search box implementation` to `litegraph`. Hit the button.

### Debug Logs

```powershell
Nothing interesting.
```

### Browser Logs

Nothing interesting.

### Setting JSON

Basically this (file is on another machine)

```
Comfy.TutorialCompleted: true
Comfy.Keybinding.NewBindings:
  Comfy.SearchBox.Toggle: a
  Comfy.Manager.CustomNodesManager: Shift + N
  Comfy.GroupNode.ConvertSelectedNodesToGroupNode: Shift + G
  Workspace.NextOpenedWorkflow: Alt + Tab
  Workspace.PreviousOpenedWorkflow: Shift + Alt + Tab
Comfy.Keybinding.UnsetBindings:
  Comfy.GroupNode.ConvertSelectedNodesToGroupNode: Alt + G
Comfy.NodeSearchBoxImpl: default
Comfy.DevMode: true
Comfy.NodeSuggestions.number: 10 (does not work either, but that's ok)
```

Google Chrome

### Other Information

_No response_

DrJKL:
## Summary

Moved up the settings check logic for Node search (current vs legacy) so that the keybinding based opening respects the user's preference.

## Changes

- **What**: Litegraph search now opens on hitting the bound key(s)
- **Where**: Near the cursor location
- **Why**: Couldn't tell you. I'm sure there's a reason some folks like the classic search experience.

## Review Focus

Fixes #4122

Check the behavior with each search box setting.

https://github.com/user-attachments/assets/eaf33ce7-0742-4e67-870e-0206c757ea88

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5140-Fix-Search-Box-Implementation-for-keyboard-shortcut-2566d73d3650815aa683cc4e5dcbf72a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -135,7 +135,7 @@ interface ICreateDefaultNodeOptions extends ICreateNodeOptions {
 interface HasShowSearchCallback {
   /** See {@link LGraphCanvas.showSearchBox} */
   showSearchBox: (
-    event: MouseEvent,
+    event: MouseEvent | null,
     options?: IShowSearchOptions
   ) => HTMLDivElement | void
 }
@@ -6870,7 +6870,7 @@ export class LGraphCanvas
   }
 
   showSearchBox(
-    event: MouseEvent,
+    event: MouseEvent | null,
     searchOptions?: IShowSearchOptions
   ): HTMLDivElement {
     // proposed defaults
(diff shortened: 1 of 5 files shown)


src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -7102,14 +7102,25 @@ export class LGraphCanvas
...
+    const safeEvent =
+      event ??
+      new MouseEvent('click', {
+        clientX: rect.left + rect.width * 0.5,
+        clientY: rect.top + rect.height * 0.5,
+        // @ts-expect-error layerY is a nonstandard property
+        layerY: rect.top + rect.height * 0.5
+      })
+
+    const left = safeEvent.clientX - 80
```
arjansingh, line 7119: Could we name these magic numbers for clarity? I assume they are pixel offsets, but I don't know why they are needed just by reading the code.

arjansingh (dismissed):
Only change I can think of is naming the magic numbers in your `safeEvent` math and writing some unit tests for the `searchBoxStore`. I'll send you a suggested spec off thread.

src/lib/litegraph/src/LGraphCanvas.ts
DrJKL: I also don't know where they come from. The existing code had them in there.

Image

AustinMroz: They're probably made up. There's a mess of relevant values in litegraph/public/css/litegraph.css, but they don't even appear to be accurate when taking those into account. From my reading, the offset for the search text x offset should be 10+60+10+5 = 85, and the y should be 2+4+11 = 17

arjansingh: `X_DIALOG_OFFSET` and `Y_DIALOG_OFFSET`?

AustinMroz (approved):

src/lib/litegraph/src/LGraphCanvas.ts
webfiltered: Can almost guarantee they were just made up - so the cursor was automatically over the first item without moving.  The "extra" padding beyond the context menu margin etc was very likely intentional (so it looks/feels like it's just inside the hover highlight hit box).

I've seen this in games & some web uses, but I don't really agree with the practice.  Neither macOS nor Windows do it (anymore?).

DrJKL: 🪄

christian-byrne (approved):
LGTM

PR #5142, opened 2025-08-21
New Workflow Templates Modal

Myestery:
This pull request refactors and simplifies the template workflow card components and related UI in the codebase. The main changes focus on removing unused or redundant components, improving visual and interaction consistency, and enhancing error handling for images. Below are the most important changes grouped by theme:

**Template Workflow Card Refactor and Cleanup**

* Removed the `TemplateWorkflowCard.vue` component and its associated test file `TemplateWorkflowCard.spec.ts`, as well as the `TemplateWorkflowCardSkeleton.vue` and `TemplateWorkflowList.vue` components, indicating a shift away from the previous card-based template workflow UI. [[1]](diffhunk://#diff-49569af0404058e8257f3cc0716b066517ce7397dd58744b02aa0d0c61f2a815L1-L139) [[2]](diffhunk://#diff-9fa6fc1470371f0b520d4deda4129fb313b1bea69888a376556f4bd824f9d751L1-L263) [[3]](diffhunk://#diff-bc35b6f77d1cee6e86b05d0da80b7bd40013c7a6a97a89706d3bc52573e1c574L1-L30) [[4]](diffhunk://#diff-48171f792b22022526fca411d3c3a366d48b675dab77943a20846ae079cbaf3bL1-L68)
* Removed the `TemplateSearchBar.vue` component, suggesting a redesign or replacement of the search/filter UI for templates.

**UI and Interaction Improvements**

* Improved the `CardBottom.vue` component by making its height configurable via a `fullHeight` prop, enhancing layout flexibility.
* Updated the `CardContainer.vue` component to add hover effects (background, border, shadow, and padding) and support a new `none` aspect ratio for more flexible card layouts.

**Image and Input Enhancements**

* Enhanced the `LazyImage.vue` component to display a default placeholder image when an image fails to load, improving error handling and user experience.
* Improved the `SearchBox.vue` component by making the input focusable when clicking anywhere on the wrapper, and added a template ref for better accessibility and usability. [[1]](diffhunk://#diff-08f3b0c51fbfe63171509b9944bf7558228f6c2596a1ef5338e88ab64585791bL2-R5) [[2]](diffhunk://#diff-08f3b0c51fbfe63171509b9944bf7558228f6c2596a1ef5338e88ab64585791bL16-R17) [[3]](diffhunk://#diff-08f3b0c51fbfe63171509b9944bf7558228f6c2596a1ef5338e88ab64585791bR33-R39)

**Minor UI Tweaks**

* Adjusted label styling in `SingleSelect.vue` to remove unnecessary overflow handling, simplifying the visual layout.


src/components/custom/widget/WorkflowTemplateSelector.vue
```diff
@@ -0,0 +1,377 @@
...
+    <template #contentFilter>
+      <div class="relative px-6 pt-2 pb-4 flex gap-2 flex-wrap justify-between">
+        <!-- Filters -->
+        <div class="flex gap-2 flex-wrap">
+          <!-- Model Filter -->
+          <MultiSelect
+            v-model="selectedModelObjects"
+            :label="modelFilterLabel"
+            :options="modelOptions"
+            :has-search-box="true"
```
viva-jinyi:
```
  showSelectedCount?: boolean
  /** Show "Clear all" action in the panel header */
  hasClearButton?: boolean
  /** Placeholder for the search input */
  searchPlaceholder?: string
```
You can inherit these props too.

src/components/input/MultiSelect.vue
```diff
@@ -128,6 +131,36 @@ const selectedItems = defineModel<Option[]>({
 const searchQuery = defineModel<string>('searchQuery')
 const selectedCount = computed(() => selectedItems.value.length)
 
+// Fuse.js configuration for fuzzy search
```
viva-jinyi: This logic should be handled at the container level and passed down.

src/components/widget/nav/NavItem.vue
```diff
@@ -9,21 +9,78 @@
...
 </template>
 
 <script setup lang="ts">
-const {
-  hasFolderIcon = true,
-  active,
-  onClick
-} = defineProps<{
-  hasFolderIcon?: boolean
+import { computed } from 'vue'
```
viva-jinyi: This should also be passed down from the template, and you need to wrap it with a wrapper component. This is a common component.

src/components/widget/panel/LeftSidePanel.vue
```diff
@@ -14,6 +14,7 @@
           <NavItem
             v-for="subItem in item.items"
             :key="subItem.id"
+            :icon="subItem.icon"
```
viva-jinyi: This also shouldn’t be modified because it’s going to be used in other places. Changing it here would affect other parts of the application.

src/services/dialogService.ts
```diff
@@ -124,6 +125,28 @@ export const useDialogService = () => {
     })
   }
 
+  function showWorkflowTemplateSelectorDialog() {
```
viva-jinyi: You don’t need to add a `dialogService`. Control of the dialog is handled within the composable located here: [useModelSelectorDialog.ts](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/composables/useModelSelectorDialog.ts). Please review how it works. Then, please create a new composable— for example, useWorkflowTemplate— following the same pattern and control logic as `useModelSelectorDialog`.

src/types/navTypes.ts
```diff
@@ -1,9 +1,11 @@
 export interface NavItemData {
   id: string
   label: string
+  icon?: string
```
viva-jinyi: This type is only meant to be used within UI components. If you need to use it elsewhere, you should extend it and define it in that context.

viva-jinyi (changes requested):
The common components in the input folder shouldn’t be modified. They are presentation components, not container components, and are shared across different places. If options need to be filtered, that logic should be passed down from the place where MultiSelect is used.

src/types/navTypes.ts
Myestery: Thanks @viva-jinyi 
The PR isnt ready for review yet but I'll follow the steps going foward

browser_tests/tests/templates.spec.ts
```diff
@@ -101,48 +107,76 @@ test.describe('Templates', () => {
     expect(await comfyPage.templates.content.isVisible()).toBe(true)
   })
 
-  test('Uses title field as fallback when the key is not found in locales', async ({
+  test('Uses proper locale files for templates', async ({ comfyPage }) => {
```
arjansingh, line 111: No changes required based on this comment.

Future thought: 

Can we find ways to do these checks with unit or component tests instead of browser tests?

The less we rely on browser tests, the faster our builds can run.

browser_tests/tests/templates.spec.ts
```diff
@@ -101,48 +107,76 @@ test.describe('Templates', () => {
...
-    // Expect the title to be used as fallback for template cards
+    // Verify English titles are shown as fallback
+    // Use a more specific selector to target the title heading
     await expect(
-      comfyPage.templates.content.getByText('FALLBACK TEMPLATE NAME')
-    ).toBeVisible()
+      comfyPage.templates.content.getByRole('heading', {
+        name: 'Image Generation'
+      })
+    ).toBeVisible({ timeout: 5000 })
```
arjansingh: Good job using the waiting functions instead of hard timeouts.

src/components/card/CardContainer.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div :class="containerClasses">
+  <div :class="containerClasses" :style="containerStyle">
```
arjansingh: I was actually about to build something similar. Now I can just use this once it's merged.

src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
```diff
@@ -0,0 +1,686 @@
...
+          <CardContainer
+            v-for="template in displayTemplates"
+            :key="template.name"
+            ref="cardRefs"
+            ratio="none"
+            :max-width="300"
+            :min-width="200"
+            class="cursor-pointer transition-all duration-300 hover:scale-[1.02]"
+            :data-testid="`template-workflow-${template.name}`"
+            @mouseenter="hoveredTemplates[template.name] = true"
```
arjansingh: Okay let's see if my understanding of reactivity is any good! (@DrJKL or @christian-byrne keep me honest if my understanding is off base).

Problem: `hoveredTemplates[template.name] = true` creates an ever-growing object:

```
// This keeps growing forever:
hoveredTemplates.value = {
  "template1": true,
  "template2": true, 
  "template3": true,
  // ... hundreds more, never deleted
}
```

Why it's bad: 
Vue tracks every property for reactivity. Each new template adds another reactive property that Vue must monitor, causing memory leaks and slower reactivity.

Fix:
```js
//something like this
const hoveredTemplate = ref<string | null>(null)

// Only tracks one value:
hoveredTemplate.value = "template1" // previous value is garbage collected
```

Effect: Reduces reactive tracking from O(n) templates to O(1) single value.

arjansingh: Is there a reason we made it an object of records [below](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5142/files#diff-2c860bdc48e907b1b85dbef846599d8376dd02cff90f49e490eebe61371fecedR493)?

src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
```diff
@@ -0,0 +1,686 @@
...
+  set(value: { name: string; value: string }[]) {
+    selectedLicenses.value = value.map((item) => item.value)
+  }
+})
+
+// Loading states
+const isLoading = ref(true)
+const loadingTemplate = ref<string | null>(null)
+const hoveredTemplates = ref<Record<string, boolean>>({})
+const cardRefs = ref<HTMLElement[]>([])
```
arjansingh: At the very least we gotta clean these up on unmount so the refs don't hang around:

```js
onBeforeUnmount(() => {
  hoveredTemplates.value = {}  // Release memory
  cardRefs.value = []          // Release DOM refs
})
```

src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
```diff
@@ -0,0 +1,686 @@
...
+            {{ pageTitle }}
+          </span>
+        </div>
+
+        <!-- Template Cards Grid -->
+        <div
+          class="grid grid-cols-[repeat(auto-fill,minmax(16rem,1fr))] gap-x-4 gap-y-6 px-4 py-4"
+          data-testid="template-workflows-content"
+        >
+          <CardContainer
```
arjansingh, line 171:
```
v-memo="[template.name, hoveredTemplate === template.name]"
```

We could memo this to only re-render this card if name OR hover state actually changed.

src/components/input/SearchBox.vue
```diff
@@ -29,6 +30,13 @@ const {
...
 const searchQuery = defineModel<string>()
 
+const input = ref<{ $el: HTMLElement } | null>()
+const focusInput = () => {
+  if (input.value && input.value.$el) {
+    input.value.$el.focus()
+  }
+}
+
 const wrapperStyle = computed(() => {
```
arjansingh, line 46: Instead of having three reactive dependencies that could trigger a re-render (and need to be followed),  we could combine them into a single reactive object that's only evaluated once per state change:

```js
const styles = computed(() => ({
  wrapper: cn(baseClasses, showBorder ? borderStyles : sizeStyles),
  input: cn('absolute inset-0 w-full h-full pl-11', ...),
  icon: cn(showBorder ? 'text-zinc-300 dark-theme:text-zinc-700' : 'text-neutral')
}))
```

```html
<template>
  <div :class="styles.wrapper">
    <input :class="styles.input">
    <icon :class="styles.icon">
  </div>
</template>
```

src/components/input/SearchBox.vue
```diff
@@ -29,6 +30,13 @@ const {
 // defineModel without arguments uses 'modelValue' as the prop name
 const searchQuery = defineModel<string>()
 
+const input = ref<{ $el: HTMLElement } | null>()
+const focusInput = () => {
+  if (input.value && input.value.$el) {
+    input.value.$el.focus()
+  }
+}
```
arjansingh, line 44: Since you're in here, try this. If it doesn't work forget I said anything:
```ts
const input = ref<InstanceType<typeof InputText> | null>(null)

const focusInput = () => {
  input.value?.$el?.focus() // Now TypeScript knows about $el
  input.value?.someInputMethod?.() // And component methods
}
```

src/components/input/SearchBox.vue
```diff
@@ -1,7 +1,8 @@
 <template>
-  <div :class="wrapperStyle">
+  <div :class="wrapperStyle" @click="focusInput">
     <i-lucide:search :class="iconColorStyle" />
     <InputText
+      ref="input"
```
arjansingh, line 5: a11y roles:

```vue
<div role="searchbox" :aria-label="ariaLabel || 'Search input'">
  <i-lucide:search aria-hidden="true" />
  <InputText 
    :aria-label="ariaLabel || 'Search'"
    // @keydown.enter="handleSearch" - if we have a handler i didn't see one
  />
</div>
```

src/locales/en/main.json
```diff
@@ -915,6 +917,23 @@
...
+    "noResultsHint": "Try adjusting your search or filters",
+    "modelFilter": "Model Filter",
+    "modelsSelected": "{count} Models",
+    "useCasesSelected": "{count} Use Cases",
+    "licensesSelected": "{count} Licenses",
+    "resultsCount": "Showing {count} of {total} templates",
+    "sort": {
+      "recommended": "Recommended",
+      "alphabetical": "A → Z",
+      "newest": "Newest"
```
arjansingh: nit: the search placeholder text could be in here too.

src/composables/useTemplateFiltering.ts
```diff
@@ -1,57 +1,224 @@
...
+    return ['Open Source', 'Closed Source (API Nodes)']
+  })
 
+  const filteredBySearch = computed(() => {
     if (!searchQuery.value.trim()) {
-      return templateData
+      return templatesArray.value
+    }
+
+    const results = fuse.value.search(searchQuery.value)
```
arjansingh: So if i am thinking correctly, every keystroke is triggering a search? Should we add a small debounce (50ms) to lessen the recomputation? @christian-byrne feel free to overrule me there

comfyui-wiki:
## 1. show the model size and VRAM 

Can we also show the model size and VRAM requirement info when hovering the "!" icon, just like this:

image

## 2.  the category should loading from the index.json

image


1. Currently, the category seems fixed. It need loading from the index.[locale].json file. All the categories should loading from index.[locale].json shouldn't be fixed

```
  {
    "moduleName": "default",
    "category": "TOOLS & BUILDING",
    "title": "Area Composition",
    "type": "image",
    "templates": []
    },
     {
    "moduleName": "default",
    "category": "TOOLS & BUILDING",
    "title": "Upscaling",
    "type": "image",
    "templates": []
    },
      {
    "moduleName": "default",
    "category": "TOOLS & BUILDING",
    "title": "Image API",
    "type": "image",
    "templates": []
    }
```
    
Btw, I still need to make updates to the index.[locale].json.
 
2. As for the icon I can add a icon key to the index.json
Such as:

```
      {
    "moduleName": "default",
    "category": "TOOLS & BUILDING",
    "title": "Image API",
    "type": "image",
    "icon": "audio"
    "templates": []
    }
```

But can you tell me how to find the icon class name? So I can update index.json. Also, it would be great to have a default icon when there isn't any icon for a category.

## 3. Custom node templates

Compare with old template modal, each custom node will have it own category

image

But now it's only shows one Extension category, that will be hard for users discover the custom node template

I just checked the design, and I think it should follow the same logic as the old template's modal.

image

## 4.  Getting Started

For the Getting Started category, it's a special category that has the "isEssential" key, just like:

```
  {
    "moduleName": "default",
    "category": "USE CASES",  #
    "title": "Basics",
    "type": "image",
    "isEssential": "true"
    "icon": "edu"
    "templates": [
      {
        "name": "get_start_with_comfyui",
        "title": "1. Get start with ComfyUI",
        "mediaType": "image",
        "mediaSubtype": "webp",
        "description": "xxxx",
        "tutorialUrl": "xxxxx",
        "tags": ["Text to Image", "Image"],
        "models": ["SD1.5"],
        "date": "2025-03-01",
        "size": 1.99
      },
      {
        "name": "how_to_use_subgraph",
        "title": "2. How to use subgraph ",
        "mediaType": "image",
        "mediaSubtype": "webp",
        "description": "xxxx.",
        "tutorialUrl": "xxxxx",
        "tags": ["Text to Image", "Image"],
        "models": ["SD1.5"],
        "date": "2025-03-01",
        "size": 1.99
      },
      ]
  }
```
It only display under the `All Templates`

## 5. default thumbnail

I think we should add `background-size: cover;` for the default thumbnail.

image

src/stores/workflowTemplatesStore.ts
```diff
@@ -169,38 +145,337 @@ export const useWorkflowTemplatesStore = defineStore(
...
+          // Determine performance ("Small Models") based primarily on explicit size prop (<=3GB)
+          // Fallback to heuristic based on model name keywords for backward compatibility.
+          const explicitSize = template.size
+          const heuristicPerformance = template.models?.some(
+            (model) =>
+              model.toLowerCase().includes('turbo') ||
+              model.toLowerCase().includes('fast') ||
+              model.toLowerCase().includes('schnell') ||
+              model.toLowerCase().includes('fp8')
+          )
```
arjansingh: We could optimize this by using a set-based method:

```js
  const PERFORMANCE_KEYWORDS = new Set(['turbo', 'fast', 'schnell', 'fp8'])
  const isPerformance = template.models?.some(model =>
    PERFORMANCE_KEYWORDS.has(model.toLowerCase())```

**Review by arjansingh — CHANGES_REQUESTED** [review]
I tried to focus on potential performance issues.

You should see them inline with potential ways to address them.

@christian-byrne and @DrJKL if i am not understanding view reactivity (or these are not big perf issues) please feel free to tell @Myestery which of these to ignore.

File: browser_tests/tests/templates.spec.ts (hunk 7 above)
Comment (christian-byrne) [reply browser_tests/tests/templates.spec.ts:111]: (replying to other comment) For this check, it makes sense to use e2e since we are testing that we request the file from backend, the backend supplies the locale file with expected format, and the file can be used.

We could add an additional expectation at the end to ensure the language is actually French.

File: src/components/custom/widget/WorkflowTemplateSelectorDialog.vue [hunk 19]
```diff
@@ -0,0 +1,686 @@
...
+    ? navItem.label
+    : navItem.items?.find((i) => i.id === selectedNavItem.value)?.label ||
+        t('templateWorkflows.allTemplates', 'All Templates')
+})
+
+// Initialize
+onMounted(async () => {
+  await loadTemplates()
+  await workflowTemplatesStore.loadWorkflowTemplates()
+  isLoading.value = false
```
christian-byrne: Generally we don't want to block the setup function with awaited calls that may take a long time (in this case, response might take a while).

Could we use something like https://vueuse.org/core/useAsyncState/ and start the fetch without awaiting in `onMounted`? Will allow us to better handle initial loading state -- like by showing the `Skeleton` for each card and sidebar.

src/components/input/MultiSelect.vue
```diff
@@ -152,13 +153,41 @@ const {
...
+const originalOptions = computed(() => (attrs.options as Option[]) || [])
+
+const fuse = computed(
+  () =>
+    new Fuse(originalOptions.value, {
+      keys: ['name', 'value'],
+      threshold: 0.3,
+      includeScore: false
+    })
+)
```
christian-byrne: nit: How about using https://vueuse.org/integrations/useFuse/ and putting template search functionality into a composable?

src/components/templates/TemplateWorkflowCardSkeleton.vue
christian-byrne, line 1: Don't we want to use the Skeleton for loading state still?

src/composables/useTemplateFiltering.ts
christian-byrne: Yeah, that would be the general best practice, agree (referring to the comment about adding debounce).

src/composables/useWorkflowTemplateSelectorDialog.ts
christian-byrne, line 1: This file feels slightly unnecessary. The composable is not maintaining any internal state.

browser_tests/tests/templates.spec.ts
```diff
@@ -79,6 +79,12 @@ test.describe('Templates', () => {
     // Load a template
     await comfyPage.executeCommand('Comfy.BrowseTemplates')
     await expect(comfyPage.templates.content).toBeVisible()
+
+    await comfyPage.page
+      .locator(
+        'nav > div:nth-child(2) > div > span:has-text("Getting Started")'
+      )
```
DrJKL, line 87: Is there not a way to get this through the fixtures?

browser_tests/tests/templates.spec.ts
```diff
@@ -101,48 +107,76 @@ test.describe('Templates', () => {
...
+    // Set locale to French before opening templates
+    await comfyPage.setSetting('Comfy.Locale', 'fr')
+
+    // Capture request for the French index file
+    let frenchIndexRequested = false
+    await comfyPage.page.route('**/templates/index.fr.json', async (route) => {
+      frenchIndexRequested = true
+      // Return the actual French index file
+      await route.continue()
+    })
```
DrJKL: This might be simpler with [a waitForRequest](https://playwright.dev/docs/events#waiting-for-event)

browser_tests/tests/templates.spec.ts
```diff
@@ -152,36 +186,57 @@ test.describe('Templates', () => {
     await comfyPage.executeCommand('Comfy.BrowseTemplates')
     await expect(comfyPage.templates.content).toBeVisible()
 
-    // Wait for at least one template card to appear
-    await expect(comfyPage.page.locator('.template-card').first()).toBeVisible({
-      timeout: 5000
-    })
+    const firstCard = comfyPage.page
+      .locator('[data-testid^="template-workflow-"]')
```
DrJKL, line 187: These feel like they should be in the fixture, to me.

browser_tests/tests/templates.spec.ts
```diff
@@ -152,36 +186,57 @@ test.describe('Templates', () => {
...
+      .first()
+    await expect(firstCard).toBeVisible({ timeout: 5000 })
```
DrJKL, line 189: For all of these, do you need to override the default timeout?

browser_tests/tests/templates.spec.ts
```diff
@@ -152,36 +186,57 @@ test.describe('Templates', () => {
...
+    const mobileGridClass = await templateGrid.getAttribute('class')
+    expect(mobileGridClass).toContain('grid')
 
     // Check cards at tablet size
     await comfyPage.page.setViewportSize({ width: 1024, height: 800 })
     await expect(templateGrid).toBeVisible()
-    await expect(templateGrid).toHaveScreenshot('template-grid-tablet.png')
+    // Grid should still be responsive at tablet size
+    const tabletGridClass = await templateGrid.getAttribute('class')
+    expect(tabletGridClass).toContain('grid')
```
DrJKL, line 222: Is the grid size control not being managed through tailwind's breakpoints?

DrJKL: What would you change in here to allow an `expect(tabletGridClass).not.toContain('grid')` to succeed?

src/components/card/CardBottom.vue
```diff
@@ -1,7 +1,17 @@
...
-<script setup lang="ts"></script>
+<script setup lang="ts">
+import { computed } from 'vue'
+
+const { fullHeight = true } = defineProps<{
+  fullHeight?: boolean
+}>()
+
+const containerClasses = computed(
+  () => `flex-1 w-full${fullHeight ? ' h-full' : ''}`
```
DrJKL: Please use `cn()` for the conditional classes 🙂

src/components/card/CardContainer.vue
```diff
@@ -8,20 +8,36 @@
...
   return `${baseClasses} ${ratioClasses[ratio]}`
 })
+
+const containerStyle = computed(() =>
+  maxWidth || minWidth
+    ? {
+        maxWidth: `${maxWidth}px`,
+        minWidth: `${minWidth}px`
+      }
+    : {}
```
DrJKL: If one of these isn't present, won't you get a `min-width: undefinedpx`?

src/components/input/MultiSelect.vue
```diff
@@ -152,13 +153,41 @@ const {
...
-const searchQuery = defineModel<string>('searchQuery')
+const searchQuery = defineModel<string>('searchQuery', { default: '' })
 const selectedCount = computed(() => selectedItems.value.length)
 
 const popoverStyle = usePopoverSizing({
   minWidth: popoverMinWidth,
   maxWidth: popoverMaxWidth
 })
+const attrs = useAttrs()
+const originalOptions = computed(() => (attrs.options as Option[]) || [])
```
DrJKL, line 172: Why do you need the type assertion here?

src/components/input/SearchBox.vue
DrJKL: I think the search is live-updating and happens through the v-model binding.

DrJKL: But very much yes for the roles!

src/components/templates/thumbnails/AudioThumbnail.vue
```diff
@@ -1,6 +1,9 @@
 <template>
   <BaseThumbnail>
-    <div class="w-full h-full flex items-center justify-center p-4">
+    <div
+      class="w-full h-full flex items-center justify-center p-4"
+      :style="{ backgroundImage: 'url(/assets/images/default-template.png)' }"
```
DrJKL: Could you put this into `style.css` as a `@utility`?

src/composables/useTemplateFiltering.ts
```diff
@@ -1,57 +1,224 @@
+import Fuse from 'fuse.js'
 import { type Ref, computed, ref } from 'vue'
 
 import type { TemplateInfo } from '@/types/workflowTemplateTypes'
 
 // @ts-expect-error unused (To be used later?)
```
DrJKL: 😢

DrJKL: If this isn't used, can we remove it?

src/composables/useTemplateFiltering.ts
```diff
@@ -1,57 +1,224 @@
...
+    | 'model-size-low-to-high'
 }
 
 export function useTemplateFiltering(
   templates: Ref<TemplateInfo[]> | TemplateInfo[]
 ) {
   const searchQuery = ref('')
+  const selectedModels = ref<string[]>([])
+  const selectedUseCases = ref<string[]>([])
+  const selectedLicenses = ref<string[]>([])
```
DrJKL, line 13: Not for this PR, but at some point can we change these from `string` to a string union enum?

DrJKL: or at least a Hinted string?

src/composables/useTemplateFiltering.ts
```diff
@@ -1,57 +1,224 @@
...
+    includeScore: true,
+    includeMatches: true
+  }
+
+  const fuse = computed(() => new Fuse(templatesArray.value, fuseOptions))
+
+  const availableModels = computed(() => {
+    const modelSet = new Set<string>()
+    templatesArray.value.forEach((template) => {
+      if (template.models && Array.isArray(template.models)) {
```
DrJKL:
```suggestion
      if (Array.isArray(template.models)) {
```

DrJKL: I think this function could be simplified with `.flatMap` and `.flat`. It could probably be a one-liner.

DrJKL: Same for the others.

src/composables/useTemplateFiltering.ts
```diff
@@ -1,57 +1,224 @@
...
+    return results.map((result) => result.item)
+  })
+
+  const filteredByModels = computed(() => {
+    if (selectedModels.value.length === 0) {
+      return filteredBySearch.value
+    }
+
+    return filteredBySearch.value.filter((template) => {
+      if (!template.models || !Array.isArray(template.models)) {
```
DrJKL, line 84: For all of these, `Array.isArray(undefined)` returns false, so you shouldn't need to do the additional precondition check.

src/composables/useWorkflowTemplateSelectorDialog.ts
```diff
@@ -0,0 +1,38 @@
...
+      component: WorkflowTemplateSelectorDialog,
+      props: {
+        onClose: hide
+      },
+      dialogComponentProps: {
+        pt: {
+          content: { class: '!px-0 overflow-hidden h-full !py-0' },
+          root: {
+            style:
+              'width: 90vw; height: 85vh; max-width: 1400px; display: flex;'
```
DrJKL, line 27: Could these be tailwind utilities?

src/composables/useWorkflowTemplateSelectorDialog.ts
```diff
@@ -0,0 +1,38 @@
...
+  function show() {
+    dialogService.showLayoutDialog({
+      key: DIALOG_KEY,
+      component: WorkflowTemplateSelectorDialog,
+      props: {
+        onClose: hide
+      },
+      dialogComponentProps: {
+        pt: {
+          content: { class: '!px-0 overflow-hidden h-full !py-0' },
```
DrJKL, line 24: Please remove all of the importants.

src/scripts/api.ts
```diff
@@ -600,11 +600,28 @@ export class ComfyApi extends EventTarget {
...
    */
-  async getCoreWorkflowTemplates(): Promise<WorkflowTemplates[]> {
-    const res = await axios.get(this.fileURL('/templates/index.json'))
-    const contentType = res.headers['content-type']
-    return contentType?.includes('application/json') ? res.data : []
+  async getCoreWorkflowTemplates(
+    locale?: string
+  ): Promise<WorkflowTemplates[]> {
+    const fileName =
+      locale && locale !== 'en' ? `index.${locale}.json` : 'index.json'
```
DrJKL, line 612: Should we just move the english to `index.en.json`?

src/scripts/api.ts
```diff
@@ -600,11 +600,28 @@ export class ComfyApi extends EventTarget {
...
+      const res = await axios.get(this.fileURL(`/templates/${fileName}`))
+      const contentType = res.headers['content-type']
+      return contentType?.includes('application/json') ? res.data : []
+    } catch (error) {
+      // Fallback to default English version if localized version doesn't exist
+      if (locale && locale !== 'en') {
+        console.warn(
+          `Localized templates for '${locale}' not found, falling back to English`
+        )
+        return this.getCoreWorkflowTemplates()
```
DrJKL, line 623: If there's a network issue and we can't get the English locale file, would this end up with a stack overflow and a bunch of failed network requests?

src/services/dialogService.ts
```diff
@@ -235,6 +192,30 @@ export const useDialogService = () => {
...
+      priority: 2,
+      dialogComponentProps: {
+        closable: false,
+        modal: false,
+        position: 'bottom',
+        pt: {
+          root: { class: 'w-[80%] max-w-2xl mx-auto border-none' },
+          content: { class: 'p-0!' },
+          header: { class: 'p-0! border-none' },
+          footer: { class: 'p-0! border-none' }
```
DrJKL, line 213: Please no important.

src/utils/categoryIcons.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    // Extensions and tools
+    extensions: 'icon-[lucide--puzzle]',
+    tools: 'icon-[lucide--wrench]',
+
+    // Fallbacks for common patterns
+    upscaling: 'icon-[lucide--maximize-2]',
+    controlnet: 'icon-[lucide--sliders-horizontal]',
+    'area-composition': 'icon-[lucide--layout-grid]'
+  }
```
DrJKL, line 48: If this never changes, it can be moved to module scope so it's not instantiated per request.

DrJKL (changes requested):

src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
```diff
@@ -0,0 +1,686 @@
...
+            $t(
+              'templateWorkflows.noResultsHint',
+              'Try adjusting your search or filters'
+            )
+          }}
+        </p>
+      </div>
+      <div v-else>
+        <!-- Title -->
+        <div class="px-6 pt-4 pb-2 text-2xl font-semibold text-neutral">
```
viva-jinyi:
```
<!-- Title -->
<div class="px-6 pb-4 text-2xl font-semibold text-neutral">
  <!-- show selected nav -->
  <span>
    {{ pageTitle }}
  </span>
</div>
```
I think this title markup should go right before the #contentFilter template closes. Otherwise, the title disappears once you scroll down.

[screen-capture.webm](https://github.com/user-attachments/assets/9b6fb54a-dac7-42aa-bc9d-74b9e6df37cc)

The video is a series of screen captures arranged in a to-be → as-is sequence.

src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
```diff
@@ -0,0 +1,686 @@
...
+          <!-- show selected nav -->
+          <span>
+            {{ pageTitle }}
+          </span>
+        </div>
+
+        <!-- Template Cards Grid -->
+        <div
+          class="grid grid-cols-[repeat(auto-fill,minmax(16rem,1fr))] gap-x-4 gap-y-6 px-4 py-4"
```
viva-jinyi: We don't need this extra padding. It has already implemented in Layout component.

src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
Myestery: This is valid!

Myestery: redoing now

src/components/input/SearchBox.vue
Myestery: N/A

src/components/templates/TemplateWorkflowCardSkeleton.vue
Myestery: we have it in the dialog now

src/components/input/MultiSelect.vue
Myestery: for now, multi select just filters items within itself when searchbox is used

browser_tests/tests/templates.spec.ts
Myestery: For now, this is the straight foward way to target this group

browser_tests/tests/templates.spec.ts
Myestery: Thanks, will use it now!

browser_tests/tests/templates.spec.ts
Myestery: not really

src/components/input/MultiSelect.vue
Myestery: doesnt reflect without setting it explicitly

src/scripts/api.ts
Myestery: There's no index.en.json

src/scripts/api.ts
Myestery: No, english is the base case

src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
Myestery: Thanks. Nice ctch

src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
Myestery: Its intentional for the hover display

src/components/input/SearchBox.vue
Myestery: N/A

browser_tests/tests/templates.spec.ts
Myestery: that wont change fro input.
means it's redundant, taking it away now

browser_tests/tests/templates.spec.ts
DrJKL: This doesn't feel very straightforward to me

src/scripts/api.ts
DrJKL: Yeah, but if there were, could we remove the special casing?

src/scripts/api.ts
DrJKL: So you'd get the first warning about it falling back to English, then an error log about it failing to load the templates, then the caller would get an empty array?

Myestery: yeah

christian-byrne:
I looked at the 4 failing tests:

- [This testcase](https://090cb8f0.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-f2512260122f99905658) we can potentially delete or change
- [This failure](https://090cb8f0.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-f5a1a01280744a8a5cd7) seems like it might be expected?
- [This one](https://090cb8f0.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-60be6643b4cb745df0a7) seems like potentially an easy fix?
- [This one](https://090cb8f0.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-f5a1a01280744a8a5cd7) seems like we just need to change the locator to match the new system

Myestery:
> I looked at the 4 failing tests:
> 
> * [This testcase](https://090cb8f0.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-f2512260122f99905658) we can potentially delete or change
> * [This failure](https://090cb8f0.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-f5a1a01280744a8a5cd7) seems like it might be expected?
> * [This one](https://090cb8f0.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-60be6643b4cb745df0a7) seems like potentially an easy fix?
> * [This one](https://090cb8f0.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-f5a1a01280744a8a5cd7) seems like we just need to change the locator to match the new system

Yeahh, will resolve them later today

comfyui-wiki:
## Custom node's template tag

The custom node's template seems will always has a `image` tag, need to remove it
image

image

## Tag's background color

Can you make the tag's background color slightly darker? When the cover is white, it will be hard to see the text.

image

## Getting started title needs loading from index.[locale].json as well
So it can supports i18n as well
image

## I18n support is incomplete

image

image

image

image

christian-byrne:
> Can you make the tag's background color slightly darker? When the cover is white, it will be hard to see the text.

We can make a tracking issue and try to solve this one in a future PR

Myestery:
I was able to fix the Getting started submenu
However other keys not translated have a separate issue that will be solved by another PR @comfyui-wiki

src/platform/workflow/templates/repositories/workflowTemplatesStore.ts
```diff
@@ -1,23 +1,28 @@
...
+} from '../types/template'
 
-const SHOULD_SORT_CATEGORIES = new Set([
-  // API Node templates should be strictly sorted by name to avoid any
-  // favoritism or bias towards a particular API. Other categories can
-  // have their ordering specified in index.json freely.
-  'Image API',
-  'Video API'
-])
+// Enhanced template interface for easier filtering
```
christian-byrne, line 17: nit (non-blocking): this could be jsdoc comment so it's available in IDE.

src/platform/workflow/templates/repositories/workflowTemplatesStore.ts
```diff
@@ -1,23 +1,28 @@
...
+interface EnhancedTemplate extends TemplateInfo {
+  sourceModule: string
+  category?: string
+  categoryType?: string
+  categoryGroup?: string // 'GENERATION TYPE' or 'CLOSED SOURCE MODELS'
```
christian-byrne, line 22: nit (non-blocking): same as above

src/scripts/api.ts
christian-byrne, line 1: nit (non-blocking, noting): Generally I think the we should expose errors to the components instead of eating them at the store level. In this case, if an error occurs, we want to throw or expose some reactive error object, then the component is able to decide how to interpret and render that error state in the view layer (e.g., use [NoResultsPlaceholder](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/components/common/NoResultsPlaceholder.vue#L10) and display the error message set by the store).

src/utils/categoryIcons.ts
christian-byrne, line 1: suggestion: this one should be in the templates folder as well, right?

christian-byrne (dismissed):
The code looks good, I am QA testing now.

src/utils/categoryIcons.ts
Myestery: yes

christian-byrne:
Notes from testing:

- I think the "zoom on hover" effect is meant to be applied to the image only (see original implementation from `main`). I don't think the entire card should be increasing its scale when it is hovered, as it creates a somewhat unsettling animation. Generally a scale change on hover shouldn't result in a layout change that has potential to alter the overall layout (e.g., don't make something grow on hover if it's in flex box or grid as it makes it very awkward to try to navigate the items with your mouse).
- The button to show tutorial doesn't seem to be using the design system

christian-byrne:
Found this bug, not sure if it is from this PR or pre-existing:

- Click the filter selector dropdown
- Select some items
- Select the `Chat` item
- Unselect the `Chat` item
- The dropdown disappears unexpectedly

christian-byrne:
Another bug related to the image caching:

1. Go to the Flux sidebar
2. Change the sort order from default to another one like "Newest"
3. The items are re-arranged, but the items that were previously out of view are now not loaded (the images) and stuck in the Skeleton state (hypothesis: they start on load on scroll into view events, but it doesn't account for items being re-arranged).

Related: I also wonder if we should increase the intersection observer buffer a bit?

Myestery:
> Another bug related to the image caching:
> 
> 1. Go to the Flux sidebar
> 2. Change the sort order from default to another one like "Newest"
> 3. The items are re-arranged, but the items that were previously out of view are now not loaded (the images) and stuck in the Skeleton state (hypothesis: they start on load on scroll into view events, but it doesn't account for items being re-arranged).
> 
> Related: I also wonder if we should increase the intersection observer buffer a bit?

For now I'm re-rendering the templates when the sort changes. we have the images force cached for now using the cache service so there's no mush risk of waiting

viva-jinyi:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/5797

Since layout and card style changes are difficult to review purely through code, I went ahead and applied the style adjustments directly in this PR. These changes are to align the UI with the design, so if everything looks good, please go ahead and merge.

**Before**
[screen-capture (1).webm](https://github.com/user-attachments/assets/e23d8625-e12d-4827-ba7b-afb7b923c32c)

**After**
[screen-capture (2).webm](https://github.com/user-attachments/assets/fb21f104-e07e-46b7-818e-10b9d2b5975c)

viva-jinyi:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/5797

Since layout and card style changes are difficult to review purely through code, I went ahead and applied the style adjustments directly in this PR. These changes are to align the UI with the design, so if everything looks good, please go ahead and merge.

**Before**
[before.webm](https://github.com/user-attachments/assets/c870d19f-bc7a-4cdc-8ac4-c26656f197da)

**After**
[after.webm](https://github.com/user-attachments/assets/cc557fb8-5613-4c6b-8d57-e21994a8bae7)

DrJKL:
@Myestery Could you post some updated screenshots of the latest version?

Myestery:
@DrJKL 
https://github.com/user-attachments/assets/48b63edc-ee14-438e-9dd4-184778fd09f7

christian-byrne (approved):
LGTM!!

PR #4685, opened 2025-08-05
[test] Add failing tests to reproduce Firebase network issue #4468

snomiao:
## Summary

- Add failing test cases that reproduce Firebase Auth network errors when offline
- Tests demonstrate current problematic behavior where `getIdToken()` and `getAuthHeader()` throw unhandled errors
- These tests currently pass by expecting the error to be thrown (reproducing the issue)

### `getIdToken()` network error handling
- Tests Firebase Auth `auth/network-request-failed` error when offline
- Currently throws error instead of returning null gracefully

### `getAuthHeader()` network error handling  
- Tests Firebase token refresh failure during auth header retrieval
- Currently propagates error instead of falling back to API key or returning null

## Next Steps

After implementing the fix from issue #4468, these tests should be updated to verify:
- `getIdToken()` returns `null` instead of throwing on network errors
- `getAuthHeader()` falls back gracefully to API key or returns `null`

## Related

Fixes #4468

🤖 Generated with Claude Code

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4685-test-Add-failing-tests-to-reproduce-Firebase-network-issue-4468-2466d73d3650813080a3e7c646a380e9) by [Unito](https://www.unito.io)

PR #5144, opened 2025-08-21
[fix] #4468 gracefully handle Firebase auth failure 

### LINKED ISSUE #4468: [Bug]: Firebase Auth makes network requests when offline without evicting token
## Frontend Version
1.24.1

## Expected Behavior
ComfyUI should function offline without Firebase network requests when not using Firebase features.

## Actual Behavior
Firebase Auth makes network requests that fail when offline or when Firebase is blocked (if there is a token stored from a previous or expired session):

1. **On app initialization**: `onAuthStateChanged` (firebaseAuthStore.ts:82) attempts to validate stored tokens from `firebase:authUser:[PROJECT_ID]:[DEFAULT]`

2. **When queuing prompts**: `getIdToken()` is called (app.ts:1242-1243) which attempts to refresh tokens older than 1 hour

Results in toast error messages and degraded offline experience.

## Steps to Reproduce
1. Sign in to ComfyUI with Firebase Auth
2. Close browser
3. Disconnect from internet (or be behind firewall that blocks Firebase)
4. Wait 1 hour so access token expires (or change timestamp in localstorage)
5. Open ComfyUI
6. Queue a graph
7. Observe network errors in console
## Debug Logs
```
auth/network-request-failed: A network error has occurred.
Failed to refresh token: FirebaseError: Firebase: Error (auth/network-request-failed)
```

## Browser Logs
```
FirebaseError: Firebase: Error (auth/network-request-failed).
    at createErrorInternal (index-6bd8d405.js:474:41)
    at _fail (index-6bd8d405.js:445:11)
    at _performFetchWithErrorHandling (index-6bd8d405.js:998:11)
```

## Setting JSON
Standard settings with Firebase Auth enabled

## What browsers do you use to access the UI?
- Google Chrome
- Mozilla Firefox
- Brave

## Other Information

Firebase Auth behavior:
- ID tokens expire after 1 hour
- `browserLocalPersistence` stores tokens in localStorage/IndexedDB
- Automatic token refresh attempts on expiry
- Exponential backoff for failed requests
- localStorage not cleared on network failures

### Proposed Solutions

**1. Catch Network Errors Gracefully**
```typescript
const getIdToken = async (): Promise<string | null> => {
  if (currentUser.value) {
    try {
      return await currentUser.value.getIdToken()
    } catch (error) {
      if (error.code === 'auth/network-request-failed') {
        return null; // Works offline or behind firewalls
      }
      throw error
    }
  }
  return null
}
```

**2. Lazy Token Validation**
```typescript
onAuthStateChanged(auth, (user) => {
  currentUser.value = user
  isInitialized.value = true
  // Skip immediate token validation
})

// app.ts - only try token if needed
const comfyOrgAuthToken = await useFirebaseAuthStore().getIdToken()
  .catch(() => undefined); // Fail silently
```

**3. Circuit Breaker Pattern**
After multiple failures, stop attempting requests for a cooldown period:
```typescript
class AuthCircuitBreaker {
  private failures = 0;
  private lastFailure = 0;
  private readonly threshold = 3;
  private readonly timeout = 60000; // 1 minute

  async getToken(): Promise<string | null> {
    // If circuit is "open" (too many recent failures), return null immediately
    if (this.failures >= this.threshold && 
        Date.now() - this.lastFailure < this.timeout) {
      return null;
    }
    
    try {
      const token = await getIdToken();
      this.failures = 0; // Reset on success
      return token;
    } catch (error) {
      this.failures++;
      this.lastFailure = Date.now();
      return null;
    }
  }
}
```

**4. Service Worker Intercept**
Intercept and immediately fail Firebase requests to prevent hanging:
```javascript
// In service worker
self.addEventListener('fetch', (event) => {
  const url = new URL(event.request.url);
  
  // Firebase domains that might be blocked
  const firebaseDomains = [
    'googleapis.com',
    'firebaseapp.com', 
    'firebaseio.com'
  ];
  
  if (firebaseDomains.some(domain => url.hostname.includes(domain))) {
    // Try fetch with short timeout
    event.respondWith(
      Promise.race([
        fetch(event.request),
        new Promise((_, reject) => 
          setTimeout(() => reject(new Error('Timeout')), 3000)
        )
      ]).catch(() => 
        new Response(null, { status: 503, statusText: 'Service Unavailable' })
      )
    );
  }
});
```

**5. Offline-First Auth Wrapper**
Wrapper store that tracks Firebase availability and caches state:
```typescript
export const useOfflineAuthStore = () => {
  const authStore = useFirebaseAuthStore();
  const firebaseAvailable = ref(true);
  const lastValidToken = ref<string | null>(null);
  
  const checkFirebaseAvailability = async () => {
    try {
      // Quick health check to Firebase
      await fetch('https://www.googleapis.com/identitytoolkit/v3/relyingparty', {
        method: 'HEAD',
        mode: 'no-cors',
        signal: AbortSignal.timeout(2000)
      });
      firebaseAvailable.value = true;
    } catch {
      firebaseAvailable.value = false;
    }
  };
  
  return {
    ...authStore,
    getIdToken: async () => {
      if (!firebaseAvailable.value) {
        return lastValidToken.value; // Use cached token
      }
      
      try {
        const token = await authStore.getIdToken();
        lastValidToken.value = token; // Cache for offline use
        return token;
      } catch (error) {
        firebaseAvailable.value = false;
        return lastValidToken.value;
      }
    }
  };
}
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4468-Bug-Firebase-Auth-makes-network-requests-when-offline-without-evicting-token-2336d73d365081ab8d50e7499f3239eb) by [Unito](https://www.unito.io)

arjansingh:
## Summary
Fixes #4468 

Firebase network failures are now gracefully handled.

The error is no longer fatal.

The user is still able to use ComfyUI for workflows that don't require authentication.

For workflows that require authentication, they are prompted to log in when the core server rejects their request for improper credentials. (This was an existing fall back workflow).

See attached video for a visual demo.

## Changes

- **What**: 
If Firebase became unavailable during a long lived (> 1 hour) user session, the user would get a fatal error when Firebase tried to refresh their expired token and the network request failed.

We now will gracefully handle this error and pass all others up the chain for handling. See [Firebase Errors](https://github.com/lamodots/error-codes-and-messages-for-firebase-authentication-API) for a comprehensive list.

## Review Focus

`src/stores/firebaseAuthStore.ts` - handles the error triage and propagation logic.
`src/scripts/app.ts` - handles UI concerns by showing an error dialog for any unhandled errors from `getIdToken`
`tests-ui/tests/store/firebaseAuthStore.test.ts` - incorporates the test cases from #4685 that @snomiao generated. I had to modify some of these test cases (and add an additional one) to match the behavior of the actual solution. Sno, i'd love your review here to make sure I didn't accidentally remove an important assertion.

Fixes #4468 

## Screenshots

https://github.com/user-attachments/assets/330a1c6c-a698-4fcd-9a31-00c681aaada0

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5144-fix-4468-gracefully-handle-Firebase-auth-failure-2566d73d3650817a941de7e161df6f36) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -1303,8 +1303,16 @@ export class ComfyApp {
     const executionStore = useExecutionStore()
     executionStore.lastNodeErrors = null
 
-    let comfyOrgAuthToken =
-      (await useFirebaseAuthStore().getIdToken()) ?? undefined
+    let comfyOrgAuthToken = await useFirebaseAuthStore()
+      .getIdToken()
+      .catch((error: unknown) => {
+        useDialogService().showErrorDialog(error, {
+          title: t('errorDialog.defaultTitle'),
+          reportType: 'authenticationError'
+        })
+        console.error(error)
+      })
+
     let comfyOrgApiKey = useApiKeyAuthStore().getApiKey()
 
     try {
@@ -1321,7 +1329,7 @@ export class ComfyApp {
 
           const p = await this.graphToPrompt(this.graph)
           try {
-            api.authToken = comfyOrgAuthToken
+            api.authToken = comfyOrgAuthToken ?? undefined
             api.apiKey = comfyOrgApiKey ?? undefined
             const res = await api.queuePrompt(number, p, {
               partialExecutionTargets: queueNodeIds
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -1303,16 +1303,7 @@ export class ComfyApp {
     const executionStore = useExecutionStore()
     executionStore.lastNodeErrors = null
 
-    let comfyOrgAuthToken = await useFirebaseAuthStore()
-      .getIdToken()
-      .catch((error: unknown) => {
-        useDialogService().showErrorDialog(error, {
-          title: t('errorDialog.defaultTitle'),
-          reportType: 'authenticationError'
-        })
-        console.error(error)
-      })
-
+    let comfyOrgAuthToken = await useFirebaseAuthStore().getIdToken()
     let comfyOrgApiKey = useApiKeyAuthStore().getApiKey()
 
     try {
@@ -1329,7 +1320,7 @@ export class ComfyApp {
 
           const p = await this.graphToPrompt(this.graph)
           try {
-            api.authToken = comfyOrgAuthToken ?? undefined
+            api.authToken = comfyOrgAuthToken
             api.apiKey = comfyOrgApiKey ?? undefined
             const res = await api.queuePrompt(number, p, {
               partialExecutionTargets: queueNodeIds
--- a/src/stores/firebaseAuthStore.ts
+++ b/src/stores/firebaseAuthStore.ts
@@ -1,5 +1,7 @@
+import { FirebaseError } from 'firebase/app'
 import {
   type Auth,
+  AuthErrorCodes,
   GithubAuthProvider,
   GoogleAuthProvider,
   type User,
@@ -20,6 +22,7 @@ import { useFirebaseAuth } from 'vuefire'
 
 import { COMFY_API_BASE_URL } from '@/config/comfyApi'
 import { t } from '@/i18n'
+import { useDialogService } from '@/services/dialogService'
 import { useApiKeyAuthStore } from '@/stores/apiKeyAuthStore'
 import { type AuthHeader } from '@/types/authTypes'
 import { operations } from '@/types/comfyRegistryTypes'
@@ -88,21 +91,28 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
     lastBalanceUpdateTime.value = null
   })
 
-  const getIdToken = async (): Promise<string | null> => {
+  const getIdToken = async (): Promise<string | undefined> => {
     if (currentUser.value) {
       try {
         return await currentUser.value.getIdToken()
-      } catch (error: any) {
-        if (error?.code === 'auth/network-request-failed') {
+      } catch (error: unknown) {
+        if (
+          error instanceof FirebaseError &&
+          error.code === AuthErrorCodes.NETWORK_REQUEST_FAILED
+        ) {
           console.warn(
             'Could not authenticate with Firebase. Features requiring authentication might not work.'
           )
-          return null // handle gracefully
+          return
         }
-        throw error
+
+        useDialogService().showErrorDialog(error, {
+          title: t('errorDialog.defaultTitle'),
+          reportType: 'authenticationError'
+        })
+        console.error(error)
       }
     }
-    return null
   }
 
   /**
--- a/src/stores/firebaseAuthStore.ts
+++ b/src/stores/firebaseAuthStore.ts
@@ -92,26 +92,25 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
   })
 
   const getIdToken = async (): Promise<string | undefined> => {
-    if (currentUser.value) {
-      try {
-        return await currentUser.value.getIdToken()
-      } catch (error: unknown) {
-        if (
-          error instanceof FirebaseError &&
-          error.code === AuthErrorCodes.NETWORK_REQUEST_FAILED
-        ) {
-          console.warn(
-            'Could not authenticate with Firebase. Features requiring authentication might not work.'
-          )
-          return
-        }
-
-        useDialogService().showErrorDialog(error, {
-          title: t('errorDialog.defaultTitle'),
-          reportType: 'authenticationError'
-        })
-        console.error(error)
+    if (!currentUser.value) return
+    try {
+      return await currentUser.value.getIdToken()
+    } catch (error: unknown) {
+      if (
+        error instanceof FirebaseError &&
+        error.code === AuthErrorCodes.NETWORK_REQUEST_FAILED
+      ) {
+        console.warn(
+          'Could not authenticate with Firebase. Features requiring authentication might not work.'
+        )
+        return
       }
+
+      useDialogService().showErrorDialog(error, {
+        title: t('errorDialog.defaultTitle'),
+        reportType: 'authenticationError'
+      })
+      console.error(error)
     }
   }
 
--- a/src/stores/firebaseAuthStore.ts
+++ b/src/stores/firebaseAuthStore.ts
@@ -90,7 +90,17 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
 
   const getIdToken = async (): Promise<string | null> => {
     if (currentUser.value) {
-      return currentUser.value.getIdToken()
+      try {
+        return await currentUser.value.getIdToken()
+      } catch (error: any) {
+        if (error?.code === 'auth/network-request-failed') {
+          console.warn(
+            'Could not authenticate with Firebase. Features requiring authentication might not work.'
+          )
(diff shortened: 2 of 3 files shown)


christian-byrne (commented):
This looks good but I think we can do error handling in the firebaseAuthStore and should be good enough -- i.e., might not need it in app.ts as well. Definitely good call on only catching the network error, we don't want to consume other errors silently.

src/scripts/app.ts
```diff
@@ -1303,8 +1303,16 @@ export class ComfyApp {
...
-      (await useFirebaseAuthStore().getIdToken()) ?? undefined
+    let comfyOrgAuthToken = await useFirebaseAuthStore()
+      .getIdToken()
+      .catch((error: unknown) => {
+        useDialogService().showErrorDialog(error, {
+          title: t('errorDialog.defaultTitle'),
+          reportType: 'authenticationError'
+        })
+        console.error(error)
+      })
```
DrJKL: Could this happen inside the `getIdToken` action?

DrJKL: You can also return `undefined` from the catch if you want to eliminate the nullish coalescing on 1332 (though I think it'd be undefined by default, right?)

src/stores/firebaseAuthStore.ts
```diff
@@ -90,7 +90,17 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
...
     if (currentUser.value) {
-      return currentUser.value.getIdToken()
+      try {
+        return await currentUser.value.getIdToken()
+      } catch (error: any) {
+        if (error?.code === 'auth/network-request-failed') {
+          console.warn(
+            'Could not authenticate with Firebase. Features requiring authentication might not work.'
+          )
+          return null // handle gracefully
```
DrJKL: Ah, sometimes it's `null` and needs to be converted to `undefined`. Could we align the type in the return here and in api.authToken?

src/stores/firebaseAuthStore.ts
```diff
@@ -90,7 +90,17 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
...
+        }
+        throw error
```
DrJKL: Setup Stores can use composables, so you should be able to initiate the dialog in here.

tests-ui/tests/store/firebaseAuthStore.test.ts
```diff
@@ -331,6 +331,70 @@ describe('useFirebaseAuthStore', () => {
...
+
+      // Simulate a user with an expired token that requires network refresh
+      mockUser.getIdToken.mockReset()
+
+      // Mock network failure (auth/network-request-failed error from Firebase)
+      const networkError = new Error(
+        'Firebase: Error (auth/network-request-failed).'
+      )
+      networkError.name = 'FirebaseError'
+      ;(networkError as any).code = 'auth/network-request-failed'
```
DrJKL: Is there _any_ way to avoid `as any`?

src/stores/firebaseAuthStore.ts
```diff
@@ -90,7 +90,17 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
 
   const getIdToken = async (): Promise<string | null> => {
     if (currentUser.value) {
-      return currentUser.value.getIdToken()
+      try {
+        return await currentUser.value.getIdToken()
+      } catch (error: any) {
```
DrJKL: Errors should usually be typed `unknown`.
https://www.typescriptlang.org/play/?ts=4.0.2#example/unknown-in-catch

tests-ui/tests/store/firebaseAuthStore.test.ts
DrJKL: What is the concrete Error type when it actually runs?

arjansingh: this was vibe-coded... guilty as charged. I'll take a look.

src/stores/firebaseAuthStore.ts
arjansingh: i originally had it as an `unknown` then i had to cast it as an `any` to avoid a type error when accessing `error?.code`. I will look at making the casting more concrete.

src/stores/firebaseAuthStore.ts
arjansingh: @christian-byrne aligning on `null` fine?

@DrJKL `api.apiKey` follows this same pattern as well. Thoughts on rabbit holing into standardizing both?

DrJKL: If it's fewer than 10 lines changed, I say go for it.

src/stores/firebaseAuthStore.ts
```diff
@@ -88,11 +91,28 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
     lastBalanceUpdateTime.value = null
   })
 
-  const getIdToken = async (): Promise<string | null> => {
+  const getIdToken = async (): Promise<string | undefined> => {
     if (currentUser.value) {
```
DrJKL: Nit: Want to flip this to a guard clause so you can dedent the rest?

tests-ui/tests/store/firebaseAuthStore.test.ts
```diff
@@ -329,7 +336,73 @@ describe('useFirebaseAuthStore', () => {
...
+    })
+
+    it('should show error dialog when getIdToken fails with non-network error', async () => {
+      // This test verifies that non-network errors trigger the error dialog
+      mockUser.getIdToken.mockReset()
+
+      // Mock a non-network error using actual Firebase Auth error code
+      const authError = new Error('User account is disabled.')
+      authError.name = 'FirebaseError'
+      ;(authError as any).code = 'auth/user-disabled'
```
DrJKL:
```suggestion
      const authError = new FirebaseError(
        /** whichever one is 'auth/user-disabled' */,
        'User account is disabled.'
      )
```

DrJKL (dismissed):

src/stores/firebaseAuthStore.ts
```diff
@@ -88,11 +91,28 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
...
+          )
+          return
+        }
+
+        useDialogService().showErrorDialog(error, {
+          title: t('errorDialog.defaultTitle'),
+          reportType: 'authenticationError'
+        })
+        console.error(error)
+      }
```
DrJKL:
```suggestion
  const getIdToken = async (): Promise<string | undefined> => {
    if (!currentUser.value) return
    try {
      return await currentUser.value.getIdToken()
    } catch (error: unknown) {
      if (
        error instanceof FirebaseError &&
        error.code === AuthErrorCodes.NETWORK_REQUEST_FAILED
      ) {
        console.warn(
          'Could not authenticate with Firebase. Features requiring authentication might not work.'
        )
        return
      }

      useDialogService().showErrorDialog(error, {
        title: t('errorDialog.defaultTitle'),
        reportType: 'authenticationError'
      })
      console.error(error)
```

DrJKL (dismissed):

DrJKL (approved):

snomiao:
LGTM, @arjansingh thank you for your help!

PR #5150, opened 2025-08-21
Fix: Restore api.ts from main branch after incorrect rebase

viva-jinyi:
## Summary
- Fixed api.ts file that was incorrectly rebased from manager/menu-items-migration branch
- Restored the correct version from main branch to fix CI errors
- Removed lodash dependency in favor of direct property access

## Context
The api.ts file was previously incorrectly rebased from the manager/menu-items-migration branch, causing CI failures. This PR brings back the correct version from main branch.

## Changes
- Removed lodash `get` import that was added during incorrect rebase
- Updated `serverSupportsFeature()` and `getServerFeature()` methods to use direct property access instead of lodash
- Aligned code with main branch to ensure CI passes

## Test plan
- [x] TypeScript compilation passes
- [x] Linting passes
- [x] No functional changes, only aligning with main branch

Fixes CI build errors introduced by incorrect rebase

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5150-Fix-Restore-api-ts-from-main-branch-after-incorrect-rebase-2566d73d365081f4bf3ee0b18bb40fb5) by [Unito](https://www.unito.io)


src/locales/zh/main.json
```diff
@@ -825,24 +818,23 @@
...
     "Toggle Logs Bottom Panel": "切换日志底部面板",
-    "Toggle Model Library Sidebar": "切換模型庫側邊欄",
-    "Toggle Node Library Sidebar": "切換節點庫側邊欄",
-    "Toggle Queue Sidebar": "切換佇列側邊欄",
     "Toggle Search Box": "切换搜索框",
     "Toggle Terminal Bottom Panel": "切换终端底部面板",
     "Toggle Theme (Dark/Light)": "切换主题（暗/亮）",
-    "Toggle Workflows Sidebar": "切換工作流程側邊欄",
+    "Toggle shortcuts_essentials Bottom Panel": "切换快捷键基础面板",
+    "Toggle shortcuts_viewControls Bottom Panel": "切换快捷键视图控制面板",
```
DrJKL: These don't look quite right. Should there be lower_snake text in the keys?

DrJKL (commented):
I can't read the languages, so I can't help there...
But it looks like some of the English keys have variables instead of names. Would we be showing "Toggle shortcuts.viewControls Bottom Panel" to users in a tooltip somewhere?

DrJKL (approved):
The api.ts part seems fine though :-)
I don't love unverified type assertions like ` as T`, but I don't know enough about how to clean that up yet.

viva-jinyi:
If this gets merged, it should also resolve the CI issue in [PR #5153](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5153).

src/scripts/api.ts
```diff
@@ -1083,21 +1082,21 @@ export class ComfyApi extends EventTarget {
 
   /**
    * Checks if the server supports a specific feature.
-   * @param featureName The name of the feature to check (supports dot notation for nested values)
+   * @param featureName The name of the feature to check
    * @returns true if the feature is supported, false otherwise
    */
   serverSupportsFeature(featureName: string): boolean {
-    return get(this.serverFeatureFlags, featureName) === true
```
christian-byrne: @viva-jinyi did you see my comment on the other PR? We want to support dotprop access. I think you can just import `get` from es-toolkit compat right?

viva-jinyi: Thanks for catching that. I’ll make the necessary changes.

src/locales/zh/main.json
viva-jinyi: I haven’t touched the locale, so I’m not sure why this got updated 😅

viva-jinyi: Anyway, I modified the locales based on `manager/menu-items-migration`

src/scripts/api.ts
viva-jinyi: I applied it.

christian-byrne (approved):

PR #5154, opened 2025-08-21
[feat] Enhance MultiSelect component and Storybook stories

viva-jinyi:
## Summary
- Enhanced MultiSelect component with improved layout and conditional rendering
- Added comprehensive Storybook stories for better component testing and documentation
- Improved UI/UX for MultiSelect header features with proper spacing

### MultiSelect Component (`src/components/input/MultiSelect.vue`)
- Improved layout with proper spacing between search box and header controls
- Added conditional rendering for header section based on feature flags
- Fixed spacing issues with divider line positioning

### Storybook Stories
- **MultiSelect.stories.ts**: 
  - Added default args for all controls
  - Added 5 new story variants: WithSearchBox, WithSelectedCount, WithClearButton, AllHeaderFeatures, CustomSearchPlaceholder
  - Enhanced controls for hasSearchBox, showSelectedCount, hasClearButton, searchPlaceholder
  
- **SearchBox.stories.ts**: 
  - Added default args with placeHolder and hasBorder
  - Added WithBorder and NoBorder story variants
  - Fixed template syntax and improved story structure

- **MoreButton.stories.ts**: 
  - Updated button type to transparent for better visual consistency
  - Added proper icon sizing (16px)

- **BaseWidget.stories.ts**: 
  - Updated to showcase MultiSelect with all header features enabled

## Test plan
- [x] Run Storybook (`npm run storybook`)
- [x] Test MultiSelect component with all new story variants
- [x] Verify proper spacing and conditional rendering in MultiSelect header
- [x] Test SearchBox with border toggle functionality
- [x] Verify MoreButton displays with transparent styling
- [x] Test BaseWidget with enhanced MultiSelect features
- [x] Run component tests (`npm run test:component`)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5154-feat-Enhance-MultiSelect-component-and-Storybook-stories-2566d73d365081a0b3fddbdc0238065c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/input/MultiSelect.stories.ts
+++ b/src/components/input/MultiSelect.stories.ts
@@ -29,6 +29,13 @@ const meta: Meta<typeof MultiSelect> = {
     searchPlaceholder: {
       control: 'text'
     }
+  },
+  args: {
+    label: 'Select',
+    hasSearchBox: false,
+    showSelectedCount: false,
+    hasClearButton: false,
+    searchPlaceholder: 'Search...'
   }
 }
 
@@ -163,4 +170,42 @@ export const MultipleSelectors: Story = {
       </div>
     `
   })
+}
+
+export const WithSearchBox: Story = {
+  ...Default,
+  args: {
+    hasSearchBox: true
+  }
+}
+
+export const WithSelectedCount: Story = {
+  ...Default,
+  args: {
+    showSelectedCount: true
+  }
+}
+
+export const WithClearButton: Story = {
+  ...Default,
+  args: {
+    hasClearButton: true
+  }
+}
+
+export const AllHeaderFeatures: Story = {
+  ...Default,
+  args: {
+    hasSearchBox: true,
+    showSelectedCount: true,
+    hasClearButton: true
+  }
+}
+
+export const CustomSearchPlaceholder: Story = {
+  ...Default,
+  args: {
+    hasSearchBox: true,
+    searchPlaceholder: 'Filter packages...'
+  }
 }
--- a/src/components/input/MultiSelect.stories.ts
+++ b/src/components/input/MultiSelect.stories.ts
@@ -1,9 +1,23 @@
 import type { Meta, StoryObj } from '@storybook/vue3-vite'
+import type { MultiSelectProps } from 'primevue/multiselect'
 import { ref } from 'vue'
 
 import MultiSelect from './MultiSelect.vue'
 
-const meta: Meta<typeof MultiSelect> = {
+// Combine our component props with PrimeVue MultiSelect props
+// Since we use v-bind="$attrs", all PrimeVue props are available
+interface ExtendedProps extends Partial<MultiSelectProps> {
+  // Our custom props
+  label?: string
+  showSearchBox?: boolean
+  showSelectedCount?: boolean
+  showClearButton?: boolean
+  searchPlaceholder?: string
+  // Override modelValue type to match our Option type
+  modelValue?: Array<{ name: string; value: string }>
+}
+
+const meta: Meta<ExtendedProps> = {
   title: 'Components/Input/MultiSelect',
   component: MultiSelect,
   tags: ['autodocs'],
--- a/src/components/input/MultiSelect.stories.ts
+++ b/src/components/input/MultiSelect.stories.ts
@@ -32,6 +32,12 @@ const meta: Meta<typeof MultiSelect> = {
   },
   args: {
     label: 'Select',
+    options: [
+      { name: 'Vue', value: 'vue' },
+      { name: 'React', value: 'react' },
+      { name: 'Angular', value: 'angular' },
+      { name: 'Svelte', value: 'svelte' }
+    ],
     showSearchBox: false,
     showSelectedCount: false,
     showClearButton: false,
@@ -47,7 +53,7 @@ export const Default: Story = {
     components: { MultiSelect },
     setup() {
       const selected = ref([])
-      const options = [
+      const options = args.options || [
         { name: 'Vue', value: 'vue' },
         { name: 'React', value: 'react' },
         { name: 'Angular', value: 'angular' },
@@ -60,8 +66,11 @@ export const Default: Story = {
         <MultiSelect 
           v-model="selected" 
           :options="options"
-          label="Select Frameworks"
-          v-bind="args"
+          :label="args.label"
+          :showSearchBox="args.showSearchBox"
+          :showSelectedCount="args.showSelectedCount"
+          :showClearButton="args.showClearButton"
+          :searchPlaceholder="args.searchPlaceholder"
         />
         <div class="mt-4 p-3 bg-gray-50 dark-theme:bg-zinc-800 rounded">
           <p class="text-sm">Selected: {{ selected.length > 0 ? selected.map(s => s.name).join(', ') : 'None' }}</p>
@@ -72,10 +81,10 @@ export const Default: Story = {
 }
 
 export const WithPreselectedValues: Story = {
-  render: () => ({
+  render: (args) => ({
     components: { MultiSelect },
     setup() {
-      const options = [
+      const options = args.options || [
         { name: 'JavaScript', value: 'js' },
         { name: 'TypeScript', value: 'ts' },
         { name: 'Python', value: 'python' },
@@ -83,25 +92,43 @@ export const WithPreselectedValues: Story = {
         { name: 'Rust', value: 'rust' }
       ]
       const selected = ref([options[0], options[1]])
-      return { selected, options }
+      return { selected, options, args }
     },
     template: `
       <div>
         <MultiSelect 
           v-model="selected" 
           :options="options"
-          label="Select Languages"
+          :label="args.label"
+          :showSearchBox="args.showSearchBox"
+          :showSelectedCount="args.showSelectedCount"
+          :showClearButton="args.showClearButton"
+          :searchPlaceholder="args.searchPlaceholder"
         />
         <div class="mt-4 p-3 bg-gray-50 dark-theme:bg-zinc-800 rounded">
           <p class="text-sm">Selected: {{ selected.map(s => s.name).join(', ') }}</p>
         </div>
       </div>
     `
-  })
+  }),
+  args: {
+    label: 'Select Languages',
+    options: [
+      { name: 'JavaScript', value: 'js' },
+      { name: 'TypeScript', value: 'ts' },
+      { name: 'Python', value: 'python' },
+      { name: 'Go', value: 'go' },
+      { name: 'Rust', value: 'rust' }
+    ],
+    showSearchBox: false,
+    showSelectedCount: false,
+    showClearButton: false,
+    searchPlaceholder: 'Search...'
+  }
 }
 
 export const MultipleSelectors: Story = {
-  render: () => ({
+  render: (args) => ({
     components: { MultiSelect },
     setup() {
       const frameworkOptions = ref([
@@ -136,7 +163,8 @@ export const MultipleSelectors: Story = {
         tagOptions,
         selectedFrameworks,
         selectedProjects,
-        selectedTags
+        selectedTags,
+        args
       }
     },
     template: `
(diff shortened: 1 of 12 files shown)


Myestery:
I had to make the following mods to multiselect yesterday : https://github.com/Comfy-Org/ComfyUI_frontend/pull/5142/files#diff-2ad6dfa2673924c2a03f905f5e6142f97eaf34067500f9a37447c6d915b0c961.

Pls see if it makes sense to do it here

src/components/input/MultiSelect.stories.ts
```diff
@@ -149,3 +171,41 @@ export const MultipleSelectors: Story = {
     `
   })
 }
+
+export const WithSearchBox: Story = {
+  ...Default,
+  args: {
+    hasSearchBox: true
```
DrJKL: Extreme nit: Can we standardize the verbs across the scenarios/options?
e.g. `showSearchBox` `showSelectedCount`, `showClearButton`, etc.

src/components/input/MultiSelect.stories.ts
```diff
@@ -149,3 +171,41 @@ export const MultipleSelectors: Story = {
...
+  }
+}
+
+export const WithSelectedCount: Story = {
+  ...Default,
+  args: {
+    showSelectedCount: true
+  }
+}
+
```
DrJKL: Could you add some default items to these?
I'm lazy and don't like to have to go edit the controls 😅

DrJKL: Tried adding plain string options, got 

Image

src/components/input/SearchBox.stories.ts
```diff
@@ -25,9 +33,23 @@ export const Default: Story = {
...
-      <div>
-        <SearchBox v-model:="searchQuery" />
+      <div style="max-width: 320px;">
+        <SearchBox v-bind="args" v-model="searchText" />
       </div>
     `
   })
 }
+
+export const WithBorder: Story = {
```
DrJKL, line 43: Does border always imply a magnifying-glass too?

DrJKL (changes requested):

src/components/input/SearchBox.stories.ts
viva-jinyi: No, it’s a prop that controls the border around the entire search bar, not the magnifying-glass icon.

src/components/input/MultiSelect.stories.ts
viva-jinyi: I standardized them with the show prefix.

src/components/input/MultiSelect.stories.ts
viva-jinyi: I’ve added default items!

viva-jinyi:
## 🚀 Additional Improvements Made

Beyond the initial review feedback, I've implemented several enhancements to improve the developer experience:

### Storybook Enhancements
- **Default options in all stories**: Both MultiSelect and SingleSelect stories now have default options in their controls, making it easier to test without manual setup
- **Removed modelValue from controls**: The modelValue is now managed internally as it should be, not exposed in Storybook controls
- **Fixed story reactivity**: All stories (WithPreselectedValues, MultipleSelectors, etc.) now properly respond to control changes
- **Pre-populated sample data**: Stories that demonstrate specific features (like WithSelectedCount, WithClearButton) now come with pre-selected items to better showcase the functionality

### Component Improvements
- **Clear event emission**: Added a `@clear` event that fires when the clear button is clicked, giving parent components full awareness of the clear action while maintaining v-model synchronization
- **Proper Option type usage**: Fixed all stories to use the correct `name` property instead of `label` to match the component's type definition

### SingleSelect Updates
- Applied the same Storybook improvements to SingleSelect for consistency
- Added default options and proper args binding to all SingleSelect stories

These changes make the components more developer-friendly and the Storybook stories more interactive and useful for testing different configurations.

DrJKL:
I think maybe the most recent commit didn't push. I don't see the changes.

christian-byrne (dismissed):

christian-byrne:
I rebased this for you so we can re-run CI with the fixed tests.

DrJKL:
I'm not seeing any default items. The storybook should update on new commits, right?
image

viva-jinyi:
@DrJKL Something went wrong. Let me fix it.

DrJKL:
For the searchbox, can we add +icon -border?
image

viva-jinyi:
@DrJKL  What do you mean `+icon -border`?

DrJKL:
> @DrJKL What do you mean `+icon -border`?

One that has the icon set to visible but without the border. Right now the default and borderless are identical. And it looks like the icon is tied to the border being present.

viva-jinyi:
@DrJKL The version where the icon is visible but the border is removed is actually the default one. I think this might just be an issue with the deployed public Storybook. Anyway, I’ll check into the issue of components breaking when Storybook is deployed.

christian-byrne (dismissed):

viva-jinyi:
Neutralize Storybook's injected styles that override MultiSelect component styling during build.
스크린샷 2025-08-26 오후 2 07 11

src/components/input/MultiSelect.vue
DrJKL, line 1: I know it's not being touched in this PR, but the items in here aren't currently accessible. They're just `<span>`s, so you can't tab to them (or keyboard select) and they don't indicate to the user that they're click interactible.
Could we use the PrimeVue Button instead?

viva-jinyi: Oh, that’s a great point—I should definitely check for keyboard actions when making components. Thanks for pointing it out. I’ll include it in this PR.

christian-byrne (commented):
Should we re-expose the original `props` APIs of the components by using `inheritAttrs: false` + `v-bind="$attrs"`?

viva-jinyi:
@christian-byrne I think it’s better to selectively pass down only the props we actually need. If a custom component is already being used in multiple places, and then that component gets updated, we might not know exactly which props are affected or how they’ll behave. Without that control, there’s a real risk of side effects creeping in.
I also set up Storybook so that the UI changes when you toggle different props. If we were to just expose every single prop, we’d also need to think about which ones should actually be controllable in Storybook. That’s another reason I think it’s better for us to selectively pass them down.

christian-byrne (changes requested):
Vue best practice: if you’re making a wrapper around a library component like PrimeVue’s, you generally shouldn’t put a root wrapper unless you have a very specific reason.

Otherwise, the new plain root element (e.g., `<div>`) essentially privatizes the entire library component. If you do, you should re-bind the props as mentioned previously.

In general, wrappers should be transparent by default. A wrapper exists to add small behavior or convenience, not to restrict the underlying component’s API unless you’re deliberately curating it. Obscuring the API restricts the access and discoverability and can legitimately worsen the output from other devs, designers, and LLMs.

src/components/input/MultiSelect.vue
```diff
@@ -1,93 +1,96 @@
...
+                ? $t('g.itemsSelected', { selectedCount })
+                : $t('g.itemSelected', { selectedCount })
+            }}
+          </span>
+          <TextButton
+            v-if="showClearButton"
+            :label="$t('g.clearAll')"
+            type="transparent"
+            size="fit-content"
+            class="text-sm !text-blue-500 !dark-theme:text-blue-600"
```
DrJKL, line 50: nit: Do these need the important?

src/components/input/MultiSelect.vue
```diff
@@ -1,93 +1,96 @@
...
+    <!-- Custom option row: square checkbox + label (unchanged layout/colors) -->
+    <template #option="slotProps">
+      <div class="flex items-center gap-2">
+        <div
+          class="flex h-4 w-4 p-0.5 flex-shrink-0 items-center justify-center rounded border-[3px] transition-all duration-200"
+          :class="
+            slotProps.selected
+              ? 'border-blue-400 bg-blue-400 dark-theme:border-blue-500 dark-theme:bg-blue-500'
+              : 'border-neutral-300 dark-theme:border-zinc-600 bg-neutral-100 dark-theme:bg-zinc-700'
+          "
```
DrJKL: nit: Can these be extracted and use the cn function?

src/components/input/MultiSelect.vue
```diff
@@ -153,17 +156,25 @@ const pt = computed(() => ({
...
   overlay:
     'mt-2 bg-white dark-theme:bg-zinc-800 text-neutral dark-theme:text-white rounded-lg border border-solid border-zinc-100 dark-theme:border-zinc-700',
   list: {
     class: 'flex flex-col gap-1 p-0 list-none border-none text-xs'
   },
-  // Option row hover tone identical
-  option:
-    'flex gap-1 items-center p-2 hover:bg-neutral-100/50 dark-theme:hover:bg-zinc-700/50',
+  // Option row hover and focus tone
+  option: ({ context }: any) => ({
```
DrJKL: nit: Do we know what the type of the argument is here?

src/components/widget/layout/BaseWidget.stories.ts
```diff
@@ -240,6 +240,9 @@ const createStoryTemplate = (args: StoryArgs) => ({
               v-model="selectedFrameworks"
               label="Select Frameworks"
               :options="frameworkOptions"
+              :has-search-box="true"
+              :show-selected-count="true"
+              :has-clear-button="true"
```
DrJKL, line 245: Should these be `:show-`?

DrJKL (dismissed):

snomiao:
@christian-byrne here，post a new comment starts with “let sno” will spawn a agent helps you perform tasks

src/components/input/MultiSelect.vue
```diff
@@ -1,93 +1,96 @@
...
-            v-if="hasSearchBox"
-            v-model="searchQuery"
-            :has-border="true"
-            :place-holder="searchPlaceholder"
+      <div class="p-2 flex flex-col pb-0">
+        <SearchBox
+          v-if="showSearchBox"
+          v-model="searchQuery"
+          :class="showSelectedCount || showClearButton ? 'mb-2' : ''"
+          :has-border="true"
```
christian-byrne: Should this be show-border?

src/components/input/MultiSelect.vue
```diff
@@ -1,93 +1,96 @@
...
-      :placeholder="label"
-      :max-selected-labels="0"
-      :pt="pt"
+  <MultiSelect
+    v-model="selectedItems"
+    :options="options"
+    option-label="name"
+    unstyled
+    :placeholder="label"
+    :max-selected-labels="0"
```
christian-byrne: Is it necessary to include options, optionLabel, placeholder, and maxSelectedLabels explicitly? They will already be set by the parent. At the very least, we probably don't need to alias them.

christian-byrne (changes requested):

src/components/input/MultiSelect.vue
viva-jinyi: If we don’t add the import in this class, the color override won’t work as intended. I think it would be better to open up a new prop in TextButton in new PR, decide on a name for this blue-styled button type, and then discuss with design how it should be defined and applied.

스크린샷 2025-08-31 오전 9 03 08
스크린샷 2025-08-31 오전 9 03 18
스크린샷 2025-08-31 오전 9 03 27
스크린샷 2025-08-31 오전 9 03 36

src/components/input/MultiSelect.vue
viva-jinyi: Thanks for the review! I'd prefer to keep the Tailwind classes inline in the template as it follows Tailwind's utility-first philosophy - being able to see the styles directly in the template improves developer experience. Since we don't use a cn utility elsewhere in the codebase, introducing it here would break consistency. The PrimeVue PT configuration is already properly organized in the computed property below, which handles that abstraction well.

src/components/widget/layout/BaseWidget.stories.ts
viva-jinyi: I committed to fix it but didn't push it 😅

src/components/input/MultiSelect.vue
viva-jinyi: Good point! Thanks.

src/components/input/MultiSelect.vue
viva-jinyi: Thanks for pointing out!

src/components/input/MultiSelect.vue
viva-jinyi: I’ve removed the unnecessary parts, and for the props needed due to custom styling I added comments above the component to explain them.

viva-jinyi:
## 📝 Changes Summary

### 1. **Improved props handling for MultiSelect/SingleSelect** (commit: 50db4b712)
- MultiSelect: Removed `options` prop, pass all PrimeVue props via `v-bind="$attrs"`
- SingleSelect: Kept `options` prop (required for getLabel function), added `v-bind="$attrs"`
- Fixed Storybook type definitions to resolve TypeScript errors

### 2. **Fixed SearchBox v-model bugs** (commit: 4ed922209)
- Fixed `defineModel` usage (removed empty string parameter)
- Fixed v-model syntax error: `v-model:=` → `v-model`
- Added `v-model:search-query` binding to MultiSelect
(cc. @Myestery)

### 3. **Renamed ModelSelector → SampleModelSelector** (commit: 4ed922209)
- Renamed to clarify it's a sample component
- Updated import path in `useModelSelectorDialog`

**Result:** Search functionality working properly, all TypeScript checks passing

@christian-byrne @DrJKL

christian-byrne (approved):

PR #5158, opened 2025-08-21
Fix/selection toolbox reflow

### LINKED ISSUE #4604: Use transform to position floating selection toolbox

simula-r:
## Summary

After investigating https://github.com/Comfy-Org/ComfyUI_frontend/issues/4604, I've found a solution that eliminates layout thrashing while maintaining all existing functionality.

### The Problem with purposed Transform-based Approach

Initially, I attempted to use CSS transforms for all selection-related DOM components (SelectionOverlay, SelectionToolbox). However, this approach had fundamental issues:

  1. Parent-child scaling conflicts: Using scale on the parent SelectionOverlay meant children
  would inherit the scale, requiring counter-scaling to maintain the toolbox's fixed size
  2. Visual artifacts: CSS borders and other properties scaled incorrectly, creating visual inconsistencies
  3. Complex workarounds: Attempts to use useTransformSettling to snap between transform and absolute positioning created visible jank

  ### The Solution: Hybrid Rendering

  The key insight was recognizing that we don't need the SelectionOverlay as a DOM element at all. Since we're already selecting nodes on the canvas, we can:

  1. Render selection borders directly on canvas via a core extension
    - Eliminates the SelectionOverlay DOM element entirely
    - No transform/scale issues since it's native canvas rendering
    - Perfect sync with node selection since it's in the same coordinate space
    - No need for useTransformSettling composable 
    - No RAFs needed
    - Enables flexible multi-node visual effects in the future

  2. Use transform for toolbox positioning only
    - Apply transform: translate(x, y) for position updates
    - No scaling needed - maintain fixed height/width regardless of zoom level (same UX we currently support)
  
### Results

  - Performance: 0 layout recalculations per second during all interactions (zooming, panning, and dragging)
  - Functionality: Maintains exact same UX - toolbox stays fixed size at all zoom levels
  - Simplicity: Removes complex transform settling logic, counter-scaling calculations, and RAF crutch

## Review Focus

I decided to achieve the selection border via a webgl core/extension, Im not sure if this will cause any problems?

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/4604

https://github.com/user-attachments/assets/26a4c71d-ecb8-4fa5-8ca9-f0a7b77dbb5a

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5158-Fix-selection-toolbox-reflow-2566d73d365081dcb028e11bd2e66bcb) by [Unito](https://www.unito.io)


src/components/graph/GraphCanvas.vue
```diff
@@ -38,9 +38,9 @@
   canvasStore.canvas to be initialized. -->
   <template v-if="comfyAppReady">
     <TitleEditor />
-    <SelectionOverlay v-if="selectionToolboxEnabled">
-      <SelectionToolbox />
-    </SelectionOverlay>
+    <!-- SelectionToolbox is now standalone, no parent overlay needed -->
```
DrJKL:
```suggestion
```

src/components/graph/SelectionToolbox.vue
```diff
@@ -41,23 +43,30 @@ import Load3DViewerButton from '@/components/graph/selectionToolbox/Load3DViewer
...
 const canvasStore = useCanvasStore()
 const extensionService = useExtensionService()
 const canvasInteractions = useCanvasInteractions()
 
-const selectionOverlayState = inject(SelectionOverlayInjectionKey)
-const { shouldAnimate } = useRetriggerableAnimation(
-  selectionOverlayState?.updateCount,
-  { animateOnMount: true }
-)
+// Get position and visibility from the new composable
```
DrJKL:
```suggestion
```

christian-byrne:
```mermaid
flowchart TD
    A[User Action] --> B{Action Type?}
    
    B -->|Select Nodes| C[Selection Event]
    B -->|Drag Nodes| D[Drag Event] 
    B -->|Zoom/Pan Canvas| E[Transform Event]
    B -->|Page Load| F[Init Event]
    
    C --> G[updateSelectionBounds]
    D --> H[Hide Toolbox]
    E --> I[updateTransform]
    F --> J[Initialize DOM]
    
    G --> K[Calculate World Coordinates]
    K --> L[Store in worldPosition ref]
    L --> M[Call updateTransform]
    
    H --> N["style.visibility = 'hidden'<br/>(DOM attribute change)"]
    
    I --> O[Convert World → Screen Coords]
    O --> P["transform: translate(x,y)<br/>(Compositor layer)"]
    
    J --> Q["position: fixed, left: 0, top: 0<br/>(DOM attributes - set once)"]
    
    style N fill:#ffcccc
    style P fill:#ccffcc
    style Q fill:#ccccff
    
    classDef domChange fill:#ffcccc,stroke:#ff0000
    classDef transformChange fill:#ccffcc,stroke:#00aa00  
    classDef initChange fill:#ccccff,stroke:#0000aa
```

## Detailed Flow with Positioning Decisions

```mermaid
sequenceDiagram
    participant User
    participant Canvas
    participant Watcher
    participant Composable
    participant DOM
    participant Browser
    
    Note over DOM: Initial State (useSelectionToolboxPosition.ts:25-30)
    DOM->>DOM: position: fixed, left: 0, top: 0, visibility: hidden
    Note right of DOM: DOM attributes set once - no layout cost
    
    User->>Canvas: Selects nodes
    Canvas->>Canvas: state.selectionChanged = true
    
    Watcher->>Composable: Selection change detected (line 100-114)
    Composable->>Composable: updateSelectionBounds() called
    
    alt No items selected
        Composable->>DOM: visibility: 'hidden' (line 42)
        Note right of DOM: DOM attribute change - minimal layout cost
    else Items selected  
        Composable->>Composable: Calculate world bounds (line 48-58)
        Composable->>Composable: Store in worldPosition ref
        Composable->>Composable: updateTransform() called (line 60)
        
        Composable->>Composable: Convert world→screen coords (line 75-77)
        Composable->>DOM: transform: translate(x,y) (line 87)
        Note right of DOM: Transform only - compositor layer, no layout!
        Composable->>DOM: visibility: 'visible' + willChange: 'transform' (line 88-89)
        
        Browser->>Browser: Composite layer optimization
    end
    
    User->>Canvas: Drags nodes
    Canvas->>Canvas: state.draggingItems = true
    
    Watcher->>Composable: Drag state change (line 119-133)
    Composable->>DOM: visibility: 'hidden' (line 125)
    Note right of DOM: DOM attribute - hides during drag
    
    User->>Canvas: Zoom/Pan canvas
    Canvas->>Composable: useCanvasTransformSync triggers (line 94)
    Composable->>Composable: updateTransform() via continuous sync
    Composable->>DOM: transform: translate(newX,newY)
    Note right of DOM: Only transform changes - smooth performance!
```

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -0,0 +1,141 @@
...
+      // bounds = [x, y, width, height]
+      worldPosition.value = {
+        x: bounds[0] + bounds[2] / 2, // Center X of bounds
+        y: bounds[1], // Top Y of bounds
+        width: bounds[2],
+        height: bounds[3]
+      }
+
+      updateTransform()
+    }
```
DrJKL: Could you flip the bounds existence check to a guard clause / early return?

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -77,16 +63,11 @@ export function useSelectionToolboxPosition() {
     const screenY = (worldPosition.value.y + offset[1]) * scale + canvasRect.top
 
     // Position the toolbox above the selection bounds
-    // The -50% centers it horizontally, and we subtract pixels to position above
-    const toolboxOffset = 45 // Pixels above the selection
+    // The -50% centers it horizontally,
+    const toolboxOffset = 45
```
DrJKL: This is a negative vertical offset, right?

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -37,33 +30,26 @@ export function useSelectionToolboxPosition() {
...
+    if (!bounds) {
+      return
+    }
 
-      updateTransform()
+    worldPosition.value = {
+      x: bounds[0] + bounds[2] / 2,
+      y: bounds[1],
+      width: bounds[2],
+      height: bounds[3]
```
DrJKL: I don't love magic tuples...
Maybe a destructuring assignment to name them?
```ts
const [xBase, y, width, height] = bounds
```

DrJKL (dismissed):

src/composables/canvas/useSelectionToolboxPosition.ts
simula-r: yes and we dont need it, i have a better way.

DrJKL:
The double outlines feels off to me. Can we hide the node's offset outline when we're drawing a group selection one?
Maybe indicate that they're included some other way...
I'll tag the Design folks.

DrJKL:
> The double outlines feels off to me. Can we hide the node's offset outline when we're drawing a group selection one? Maybe indicate that they're included some other way... I'll tag the Design folks.

image

simula-r:
I def agree we could make the design better. But as this is mainly a PR for perf of current stable UX/UI design (which i think above is the same as current stable correct?) can we tackle in a separate PR?

christian-byrne (approved):

DrJKL:
> I def agree we could make the design better. But as this is mainly a PR for perf of current stable UX/UI design (which i think above is the same as current stable correct?) can we tackle in a separate PR?

I thought so, but why did the snapshot screenshots change?

christian-byrne:
The selection outline is slightly changed because the rendering order changed and the style is not exaft 1-to-1 (presumably) which caused visual difference, but it's pretty minor. Since those are valid visual changes, I added the `New Browser Test Expectations` Label which triggers https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/.github/workflows/test-browser-exp.yaml and updates the screnshots.

PR #5159, opened 2025-08-21
Enhanced custom icon support with tailwind

pythongosssss:
## Summary

- Adds support for loading custom icons and using them via tailwind classes
- Updates sidebar tabs to use icon classes instead of components
- Updates "Browse Templates" menu item to use custom icon
- Remove padding from svgs
- Very basic test that just checks an svg variable is set on some icons and that it contains svg markup

## Changes

- **What**:  Adds tailwind build plugin to support custom icons
- **Dependencies**: `@iconify/tailwind`

## Review Focus

I have left in the support for component icons in sidebar tabs for extensions, if we don't think this is worth keeping i'm happy to remove it as the core no longer uses it.

New (left) vs old (right) icon size:
image

Browse templates uses custom icon:
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5159-Enhanced-custom-icon-support-with-tailwind-2566d73d365081dfa54adbc11cacbc0f) by [Unito](https://www.unito.io)


browser_tests/tests/customIcons.spec.ts
```diff
@@ -0,0 +1,57 @@
+import { expect } from '@playwright/test'
+import type { Locator } from '@playwright/test'
+
+import { comfyPageFixture as test } from '../fixtures/ComfyPage'
+
+async function verifyCustomIconSvg(iconElement: Locator) {
```
DrJKL, line 6: What do the errors for this look like in a failure state?

build/customIconCollection.js
DrJKL, line 1: Why not TypeScript?

DrJKL (approved):

christian-byrne (approved):

PR #5161, opened 2025-08-21
Rework theme menu

pythongosssss:
## Summary

Changes the Theme button to a submenu with all available themes. Previously the theme button was overly prominent in the menu for how often it was used and only gave the ability to switch between light & dark.

## Changes

- **What**: Add computed menu item containing all available themes

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5161-Rework-theme-menu-2566d73d36508158af2ac6aab386a155) by [Unito](https://www.unito.io)


browser_tests/tests/menu.spec.ts
```diff
@@ -178,6 +178,75 @@ test.describe('Menu', () => {
       await comfyPage.menu.topbar.triggerTopbarCommand(['ext', 'foo-command'])
       expect(await comfyPage.getVisibleToastCount()).toBe(1)
     })
+
+    test('Can navigate Theme menu and switch between Dark and Light themes', async ({
```
DrJKL, line 182: How do you feel about throwing in some `.screenshot()`, not necessarily as assertions, but for local runs at least.

browser_tests/tests/menu.spec.ts
```diff
@@ -178,6 +178,75 @@ test.describe('Menu', () => {
...
+    }) => {
+      // Open the topbar menu
+      await comfyPage.menu.topbar.openTopbarMenu()
+      const menu = comfyPage.page.locator('.comfy-command-menu')
+      await expect(menu).toBeVisible()
+
+      // Navigate to Theme menu item and open its submenu
+      const themeMenuItem = comfyPage.page.locator(
+        '.p-menubar-item-label:text-is("Theme")'
+      )
```
DrJKL: Could you extract a Fixture for the menus?

DrJKL: (and maybe the submenus, but those would probably be in scope for the Menu fixture itself)

DrJKL: I usually consider explicit `.locator`s in tests to be a minor code smell.

PabloWiedemann (changes requested):
@pythongosssss this is great! love that you can switch between themes without the menu closing. 
only small comment would be. Could we remove the divider between "Edit" and "View" ? 🙏

DrJKL (changes requested):
Yeah, I like the submenu version a lot better 😀

christian-byrne (approved):
LGTM!

PR #5174, opened 2025-08-23
[fix] fix #4438 do not repeatedly try to load 404'ed images

### LINKED ISSUE #4438: [Bug]: Model sidebar makes unnecessary 404 requests for non-existent preview images
Custom Node Testing Not Related

### ComfyUI Frontend Version

main (80e5cf1b)

### Expected Behavior

The model sidebar should efficiently handle model preview images by either:
1. Only requesting preview images that actually exist
2. Gracefully handling missing preview images without generating 404 errors
3. Using a proper error handling mechanism for failed image loads

### Actual Behavior

The model sidebar attempts to load preview images for every single model by constructing URLs to `.webp` files, regardless of whether these preview files exist. This results in numerous 404 errors in the browser console:

```
GET http://localhost:8188/api/experiment/models/preview/checkpoints/0/v1-5-pruned-emaonly-fp16.webp 404 (Not Found)
```

The current implementation in `ModelTreeLeaf.vue` uses CSS `background-image` which:
- Cannot detect or handle 404 errors
- Results in failed requests being part of normal application flow
- Creates unnecessary server load and network traffic

### Steps to Reproduce

1. Open ComfyUI with default models installed (many without preview images)
2. Navigate to the Model sidebar tab
3. Open browser developer console
4. Observe multiple 404 errors for model preview images

### Debug Logs

```
GET http://localhost:8188/api/experiment/models/preview/checkpoints/0/v1-5-pruned-emaonly-fp16.webp 404 (Not Found)
GET http://localhost:8188/api/experiment/models/preview/checkpoints/0/sd_xl_base_1.0.webp 404 (Not Found)
GET http://localhost:8188/api/experiment/models/preview/loras/0/example_lora.webp 404 (Not Found)
[Multiple similar 404 errors for each model without a preview]
```

### Browser Logs

N/A - The issue is visible in the Network tab showing multiple failed requests

### Settings JSON

N/A - Issue occurs with default settings

### Browser

Google Chrome

### Additional Information

**Performance Impact:**
Based on research, these 404 errors have several negative impacts:
- Each failed request still requires a full HTTP request/response cycle
- Multiple 404s increase page load time and consume bandwidth
- Server resources are wasted processing requests for non-existent files
- Log files become cluttered with 404 entries

**Technical Analysis:**
The issue is in `src/components/sidebar/tabs/modelLibrary/ModelTreeLeaf.vue` (lines 46-56) where the `modelPreviewUrl` computed property unconditionally generates preview URLs. The URL is used as a CSS background-image (line 8), which doesn't allow error handling.

**Suggested Solutions:**
1. Switch from CSS `background-image` to `` tags with proper `@error` handlers
2. Use the existing `ComfyImage.vue` component which already has error handling
3. Implement a preview existence check before making requests
4. Add lazy loading to only request previews when models are visible
5. Query the backend for which models have previews available

Related to enhancement requests #739 and #2618 for model preview functionality.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4438-Bug-Model-sidebar-makes-unnecessary-404-requests-for-non-existent-preview-images-22f6d73d365081e3aa38f851babf00bf) by [Unito](https://www.unito.io)

arjansingh:
1. Move rendering of images to TreeExplorer
2. Cache attempts to load so we only do one request for the image src

## Summary

We used to repeatedly make API request to load the image preview of models if the images weren't available (i.e., the server returned a 404). Now we cache errors and successful loads. The 404's do not repeatedly load. See video.

## Changes

- **What**: Cache failures and success of preview images. Do not attempt to load the preview images after one failed attempt to load them.

## Review Focus

1. Anything with idiomatic Vue. I think I avoided horrible stuff.
2. I tried to keep this bug's fix scope tight. I am pretty sure we could refactor this code and LazyImage to use [useImage](https://vueuse.org/core/useImage/), but I am not sure if that's in scope for this PR. Let me know otherwise.

Fixes #4438

### Before

https://github.com/user-attachments/assets/547a3726-f638-4a01-93ab-8b0736179ee7

### After

https://github.com/user-attachments/assets/697d8fe2-6bff-4967-a6b1-d8f66614d93d

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5174-fix-fix-4438-do-not-repeatedly-try-to-load-404-ed-images-2586d73d3650819a8294d05f5e6b6d85) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/common/TreeExplorer.vue
+++ b/src/components/common/TreeExplorer.vue
@@ -36,20 +36,13 @@
       <!-- Always show icon initially, hide it when image loads successfully -->
       <i
         v-show="
-          !imageLoadStatus[node.key] ||
-          imageLoadStatus[node.key] === 'new' ||
-          imageLoadStatus[node.key] === 'failed'
+          !imageLoadStatus[node.key] || imageLoadStatus[node.key] === 'failed'
         "
         :class="node.icon"
       />
       <!-- Load image in background if preview exists and hasn't failed -->
       <img
-        v-if="
-          node.previewImageUrl &&
-          (!imageLoadStatus[node.key] ||
-            imageLoadStatus[node.key] === 'new' ||
-            imageLoadStatus[node.key] === 'loaded')
-        "
+        v-if="node.previewImageUrl && imageLoadStatus[node.key] !== 'failed'"
         v-show="imageLoadStatus[node.key] === 'loaded'"
         :src="node.previewImageUrl"
         class="tree-node-preview-image"
@@ -132,7 +125,7 @@ const getTreeNodeIcon = (node: TreeExplorerNode) => {
   return isExpanded ? 'pi pi-folder-open' : 'pi pi-folder'
 }
 // Track image loading states using a single record with status
-type ImageLoadStatus = 'new' | 'loaded' | 'failed'
+type ImageLoadStatus = 'loaded' | 'failed'
 const imageLoadStatus = ref<Record<string, ImageLoadStatus>>({})
 
 const handleImageLoad = (_e: Event, key: string) => {
--- a/src/components/common/TreeExplorer.vue
+++ b/src/components/common/TreeExplorer.vue
@@ -34,16 +34,28 @@
     </template>
     <template #nodeicon="{ node }">
       <!-- Always show icon initially, hide it when image loads successfully -->
-      <i v-show="!loadedImages[node.key]" :class="node.icon" />
+      <i
+        v-show="
+          !imageLoadStatus[node.key] ||
+          imageLoadStatus[node.key] === 'new' ||
+          imageLoadStatus[node.key] === 'failed'
+        "
+        :class="node.icon"
+      />
       <!-- Load image in background if preview exists and hasn't failed -->
       <img
-        v-if="node.previewImageUrl && !failedImages[node.key]"
-        v-show="loadedImages[node.key]"
+        v-if="
+          node.previewImageUrl &&
+          (!imageLoadStatus[node.key] ||
+            imageLoadStatus[node.key] === 'new' ||
+            imageLoadStatus[node.key] === 'loaded')
+        "
+        v-show="imageLoadStatus[node.key] === 'loaded'"
         :src="node.previewImageUrl"
         class="tree-node-preview-image"
         :alt="node.label"
-        @load="(e) => handleImageLoad(e, node as RenderedTreeExplorerNode)"
-        @error="(e) => handleImageError(e, node as RenderedTreeExplorerNode)"
+        @load="(e) => handleImageLoad(e, node.key)"
+        @error="(e) => handleImageError(e, node.key)"
       />
     </template>
   </Tree>
@@ -119,16 +131,16 @@ const getTreeNodeIcon = (node: TreeExplorerNode) => {
   const isExpanded = expandedKeys.value?.[node.key] ?? false
   return isExpanded ? 'pi pi-folder-open' : 'pi pi-folder'
 }
-// Track image loading states using reactive Records for better performance
-const loadedImages = ref<Record<string, boolean>>({})
-const failedImages = ref<Record<string, boolean>>({})
+// Track image loading states using a single record with status
+type ImageLoadStatus = 'new' | 'loaded' | 'failed'
+const imageLoadStatus = ref<Record<string, ImageLoadStatus>>({})
 
-const handleImageLoad = (_e: Event, node: TreeExplorerNode) => {
-  loadedImages.value[node.key] = true
+const handleImageLoad = (_e: Event, key: string) => {
+  imageLoadStatus.value[key] = 'loaded'
 }
 
-const handleImageError = (_e: Event, node: TreeExplorerNode) => {
-  failedImages.value[node.key] = true
+const handleImageError = (_e: Event, key: string) => {
+  imageLoadStatus.value[key] = 'failed'
 }
 
 const fillNodeInfo = (node: TreeExplorerNode): RenderedTreeExplorerNode => {
--- a/src/components/common/TreeExplorer.vue
+++ b/src/components/common/TreeExplorer.vue
@@ -32,6 +32,20 @@
         <TreeExplorerTreeNode :node="node" />
       </slot>
     </template>
+    <template #nodeicon="{ node }">
+      <!-- Always show icon initially, hide it when image loads successfully -->
+      <i v-show="!loadedImages[node.key]" :class="node.icon" />
+      <!-- Load image in background if preview exists and hasn't failed -->
+      <img
+        v-if="node.previewImageUrl && !failedImages[node.key]"
+        v-show="loadedImages[node.key]"
+        :src="node.previewImageUrl"
+        class="tree-node-preview-image"
+        :alt="node.label"
+        @load="(e) => handleImageLoad(e, node as RenderedTreeExplorerNode)"
+        @error="(e) => handleImageError(e, node as RenderedTreeExplorerNode)"
+      />
+    </template>
   </Tree>
   <ContextMenu ref="menu" :model="menuItems" />
 </template>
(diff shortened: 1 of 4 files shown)


src/components/common/TreeExplorer.vue
```diff
@@ -101,9 +115,22 @@ const getTreeNodeIcon = (node: TreeExplorerNode) => {
   if (node.leaf) {
     return 'pi pi-file'
   }
+
   const isExpanded = expandedKeys.value?.[node.key] ?? false
   return isExpanded ? 'pi pi-folder-open' : 'pi pi-folder'
 }
+// Track image loading states using reactive Records for better performance
+const loadedImages = ref<Record<string, boolean>>({})
+const failedImages = ref<Record<string, boolean>>({})
```
DrJKL: If you used a string enum, could you get away with a single map?
```ts
type ImageLoadStatus = 'new' | 'loaded' | 'failed'
const imageLoadStatus = ref<Record<string, ImageLoadStatus>>({})
```

src/components/common/TreeExplorer.vue
```diff
@@ -270,4 +298,12 @@ defineExpose({
...
   pointer-events: none;
 }
+
+.tree-node-preview-image {
+  width: 1.5rem;
+  height: 1.5rem;
+  object-fit: cover;
+  border-radius: 2px;
+  vertical-align: middle;
+}
```
DrJKL, line 313: Might need to ask @christian-byrne, but do we want to start using tailwind utilities instead of component styles for things like this?

DrJKL (commented):
I would love to see a version with [useImage](https://vueuse.org/core/useImage/)...

But otherwise this looks great. I love how many lines could be removed.

DrJKL (dismissed):

src/components/common/TreeExplorer.vue
```diff
@@ -32,6 +32,20 @@
...
+      <i v-show="!loadedImages[node.key]" :class="node.icon" />
+      <!-- Load image in background if preview exists and hasn't failed -->
+      <img
+        v-if="node.previewImageUrl && !failedImages[node.key]"
+        v-show="loadedImages[node.key]"
+        :src="node.previewImageUrl"
+        class="tree-node-preview-image"
+        :alt="node.label"
+        @load="(e) => handleImageLoad(e, node as RenderedTreeExplorerNode)"
+        @error="(e) => handleImageError(e, node as RenderedTreeExplorerNode)"
```
DrJKL: Oh, while I'm here, why do we need the cast?
These both just need the key, right?

DrJKL: Like this, but then with a different signature below.
```suggestion
        @load="(e) => handleImageLoad(e, node.key)"
        @error="(e) => handleImageError(e, node.key)"
```

src/components/common/TreeExplorer.vue
arjansingh: that is a good suggestion!

src/components/common/TreeExplorer.vue
arjansingh: I'm open to converting it.

src/components/common/TreeExplorer.vue
```diff
@@ -34,16 +34,28 @@
     </template>
     <template #nodeicon="{ node }">
       <!-- Always show icon initially, hide it when image loads successfully -->
-      <i v-show="!loadedImages[node.key]" :class="node.icon" />
+      <i
+        v-show="
+          !imageLoadStatus[node.key] ||
+          imageLoadStatus[node.key] === 'new' ||
+          imageLoadStatus[node.key] === 'failed'
```
DrJKL: Definitely worth extracting a method for this and the one on the ``

DrJKL: Maybe `undefined` is as good as 'new' here

arjansingh: keep it in file? or do you want a more general utility method?

arjansingh: okay forget utility method... i went through and simplified the logic. we end up with:

```diff
diff --git a/src/components/common/TreeExplorer.vue b/src/components/common/TreeExplorer.vue
index afff7cc3..6d650552 100644
--- a/src/components/common/TreeExplorer.vue
+++ b/src/components/common/TreeExplorer.vue
@@ -35,21 +35,12 @@
     <template #nodeicon="{ node }">
       <!-- Always show icon initially, hide it when image loads successfully -->
       <i
-        v-show="
-          !imageLoadStatus[node.key] ||
-          imageLoadStatus[node.key] === 'new' ||
-          imageLoadStatus[node.key] === 'failed'
-        "
+        v-show="!imageLoadStatus[node.key] || imageLoadStatus[node.key] === 'failed'"
         :class="node.icon"
       />
       <!-- Load image in background if preview exists and hasn't failed -->
       <img
-        v-if="
-          node.previewImageUrl &&
-          (!imageLoadStatus[node.key] ||
-            imageLoadStatus[node.key] === 'new' ||
-            imageLoadStatus[node.key] === 'loaded')
-        "
+        v-if="node.previewImageUrl && imageLoadStatus[node.key] !== 'failed'"
         v-show="imageLoadStatus[node.key] === 'loaded'"
         :src="node.previewImageUrl"
         class="tree-node-preview-image"
@@ -132,7 +123,7 @@ const getTreeNodeIcon = (node: TreeExplorerNode) => {
   return isExpanded ? 'pi pi-folder-open' : 'pi pi-folder'
 }
 // Track image loading states using a single record with status
-type ImageLoadStatus = 'new' | 'loaded' | 'failed'
+type ImageLoadStatus = 'loaded' | 'failed' | undefined
 const imageLoadStatus = ref<Record<string, ImageLoadStatus>>({})
 
 const handleImageLoad = (_e: Event, key: string) => {
```

Much more clean but I don't think it's helpful to abstract these conditionals into functions.

Thanks for the tip though, the logic is a  lot easier to understand now.

If you're good let me know and i'll push up these edits.

DrJKL: Yeah, if it's a single `&&` or `||` it's probably fine. Especially if it fits on one line.

DrJKL: And I think you can omit the ` | undefined` for the type.

DrJKL:
Rebased to get the fixed tests.

src/components/common/TreeExplorer.vue
christian-byrne: Yes tailwind is preferred so that the design tokens can be tweaked globally more easily.

src/components/common/TreeExplorer.vue
arjansingh: alright we should be able to go.

src/components/common/TreeExplorer.vue
arjansingh: You got it. I'll convert this in a follow up PR just so we can get this fix out.

DrJKL (approved):

christian-byrne (dismissed):

christian-byrne (changes requested):
This looks good but two main issues:

- We are modifying the common `TreeExplorer` component and adding logic specific to the Model Tree. The `TreeExplorer` is used in other places (e.g., node and workflow sidebars)
- We are maintaining a cache at the component/presentation layer. This type of memory is generally tied to the lifecycle of the component. I.e., when the user closes the sidebar, the objects will be GC'ed.

arjansingh:
> This looks good but two main issues:
> 
> * We are modifying the common `TreeExplorer` component and adding logic specific to the Model Tree. The `TreeExplorer` is used in other places (e.g., node and workflow sidebars)
> * We are maintaining a cache at the component/presentation layer. This type of memory is generally tied to the lifecycle of the component. I.e., when the user closes the sidebar, the objects will be GC'ed.

### Edit
Here's the diff with core for context. I think it's a much simpler solution: https://github.com/comfyanonymous/ComfyUI/compare/master...arjansingh:ComfyUI:fix/add-image-cache-headers?expand=1

### Original Message
@christian-byrne So I had an epiphany talking to @DrJKL this morning. We probably don't need any frontend modifications. We double checked the server and it's not setting any cache headers for images.

I added some code to do that this morning and the results are that we'll load 404's from browser cache for 1 hour. Images will get cached for 1 day. 

These are values are just best guesses and are adjustable. I think it's a simpler solution than trying to manage this stuff in our Vue app. Just give directives to the browser and leverage it's caching.

Let me know if you like this approach. @DrJKL and I prefer it to this PR. I can write some tests and open it up against the core repo.

Loading from cache:
Screenshot 2025-08-25 at 12 15 49 PM

404's expire in 1 hour:
Screenshot 2025-08-25 at 12 20 04 PM

200's expire in 1 day:
Screenshot 2025-08-25 at 12 20 31 PM

arjansingh:
Closing in favor of https://github.com/comfyanonymous/ComfyUI/pull/9560. Handle this server side and let the browser do it's thing normally.

PR #5190, opened 2025-08-24
improve Standard mode

jtydhr88:
## Summary

1. change that make middle scroll to zoom in/out canvas, no ctrl required
2. only switch mode if not already in pan/read_only mode

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5190-remove-isCtrlOrMacMeta-2596d73d365081758a78c7d9573521cb) by [Unito](https://www.unito.io)


jtydhr88:
I deleted some unit tests because they no longer meet the new trigger conditions.

src/lib/litegraph/src/CanvasPointer.ts
```diff
@@ -357,16 +357,47 @@ export class CanvasPointer {
    * Updates the device mode based on event patterns.
    */
   #updateDeviceMode(event: WheelEvent, now: number): void {
-    if (this.#isTrackpadPattern(event)) {
+    console.log('event.deltaX:', event.deltaX)
+    console.log('event.deltaY:', event.deltaY)
```
jtydhr88, line 361: I know it will be noisy, but I prefer to keep these those console.log until we make sure the new detection logic working well on different environments, it allows us easily know the debug info from client

christian-byrne: Maybe we can remove these now since we aren't doing a custom formula that needs heavy testing/debugging? Otherwise we should switch to `log-level` module and set level to debug.

christian-byrne (changes requested):

browser_tests/fixtures/ComfyPage.ts
```diff
@@ -734,10 +734,39 @@ export class ComfyPage {
...
+              deltaY: deltaY,
+              deltaX: 0,
+              deltaMode: 0,
+              clientX: x,
+              clientY: y,
+              bubbles: true,
+              cancelable: true
+            })
+            // Add custom wheelDeltaY property
+            ;(event as any).wheelDeltaY = wheelDeltaY
```
christian-byrne, line 762: nit: Can you switch to an `// ignore` comment instead of type assertion?

browser_tests/tests/interaction.spec.ts
```diff
@@ -923,15 +922,46 @@ test.describe('Canvas Navigation', () => {
...
+          deltaY: -120,
+          deltaX: 0,
+          deltaMode: 0,
+          clientX: 400,
+          clientY: 300,
+          ctrlKey: true,
+          bubbles: true,
+          cancelable: true
+        })
+        ;(event as any).wheelDeltaY = 120
```
christian-byrne, line 939: nit: Same as above

browser_tests/tests/interaction.spec.ts
```diff
@@ -923,15 +922,46 @@ test.describe('Canvas Navigation', () => {
...
+          deltaY: 240,
+          deltaX: 0,
+          deltaMode: 0,
+          clientX: 400,
+          clientY: 300,
+          ctrlKey: true,
+          bubbles: true,
+          cancelable: true
+        })
+        ;(event as any).wheelDeltaY = -240
```
christian-byrne, line 962: nit: same as above

browser_tests/tests/interaction.spec.ts
christian-byrne, line 1: Shouldn't we move the the new wheel logic to the fixture?


```typescript
        await this.page.evaluate(
          ({ deltaY, wheelDeltaY, x, y }) => {
            const canvas = document.getElementById('graph-canvas')
            if (!canvas) return
            const event = new WheelEvent('wheel', {
              deltaY: deltaY,
              deltaX: 0,
              deltaMode: 0,
              clientX: x,
              clientY: y,
              bubbles: true,
              cancelable: true
            })
            // Add custom wheelDeltaY property
            ;(event as any).wheelDeltaY = wheelDeltaY
```

But actually, if we are no longer relying on `wheelDeltaY`, can't we remove all the custom logic for wheel events in these tests?

PR #5217, opened 2025-08-26
When converting single group to subgraph, also convert children

AustinMroz:
A small requested quality of life change.

If convert to subgraph is called on a selection consisting of a single group (the colored background rectangles), the items contained inside are also converted.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5217-When-converting-single-group-to-subgraph-also-convert-children-25b6d73d365081c396f0c2e9105b4311) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraph.ts
+++ b/src/lib/litegraph/src/LGraph.ts
@@ -1465,6 +1465,7 @@ export class LGraph
     const { state, revision, config } = this
     const firstChild = [...items][0]
     if (items.size === 1 && firstChild instanceof LGraphGroup) {
+      items = new Set([firstChild])
       firstChild.recomputeInsideNodes()
       firstChild.children.forEach((n) => items.add(n))
     }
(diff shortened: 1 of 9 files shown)


src/lib/litegraph/src/LGraph.ts
```diff
@@ -1463,6 +1463,11 @@ export class LGraph
     if (items.size === 0)
       throw new Error('Cannot convert to subgraph: nothing to convert')
     const { state, revision, config } = this
+    const firstChild = [...items][0]
+    if (items.size === 1 && firstChild instanceof LGraphGroup) {
+      firstChild.recomputeInsideNodes()
+      firstChild.children.forEach((n) => items.add(n))
```
DrJKL, line 1470: Is there a way to do this without mutating the input?

DrJKL (dismissed):

christian-byrne (dismissed):

src/lib/litegraph/src/LGraph.ts
AustinMroz: I thought I remembered that js would complain if you try and modify a passed argument, but it seems from testing that a `items = new Set()` works just fine.

I'm not too conflicted either way sincce a successful completion results in the contents of items being destoryed.

christian-byrne (approved):

christian-byrne (approved):

PR #5240, opened 2025-08-28
Add safety check for undefined widgets

AustinMroz:
A user on the Banodoco discord server posted a [report for a frontend freeze](https://discord.com/channels/1076117621407223829/1410630269756768338). The error logs indicate that some node contains an undefined widget which crashes the render loop. While I don't have any reproduction on *how* this incorrect state occured (Likely a bugged custom node), this PR adds a minimal safety check to remove the undefined widget entry instead of freezing.

It also marginally improves logging by printing the node which contains the undefined widget. This is probably the best we can do for a followup investigation without resorting to an intrusive proxy on widgets that prevents undefined entries from being inserted.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5240-Add-safety-check-for-undefined-widgets-25d6d73d36508139bad3d44d0a6ad581) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -3918,6 +3918,13 @@ export class LGraphNode
    */
   #arrangeWidgets(widgetStartY: number): void {
     if (!this.widgets || !this.widgets.length) return
+    for (let i = 0; i < this.widgets.length; i++) {
+      if (!this.widgets[i]) {
+        console.error('Removing undefined widget in', this)
+        this.widgets.splice(i, 1)
+        i--
+      }
+    }
```
DrJKL: Want to do this more functional style with a `.filter`?
I'm usually wary of mutating arrays while iterating over them.

AustinMroz: My thanks for the suggestion. This is much nicer.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -3918,12 +3918,10 @@ export class LGraphNode
    */
   #arrangeWidgets(widgetStartY: number): void {
     if (!this.widgets || !this.widgets.length) return
-    for (let i = 0; i < this.widgets.length; i++) {
-      if (!this.widgets[i]) {
-        console.error('Removing undefined widget in', this)
-        this.widgets.splice(i, 1)
-        i--
-      }
+    const realWidgets = this.widgets.filter((w) => w)
```
DrJKL, line 4044: optional nit: I usually like to make it clear that it's filtering out null/undefineds, either like `!!w` or `w != null` (single nequal)

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -3918,12 +3918,10 @@ export class LGraphNode
...
+    if (realWidgets.length != this.widgets.length) {
+      console.error('Removing undefined widgets from', this)
```
DrJKL, line 4046: What does `this` look like in the console?
How do you feel about including the number of widgets removed? I feel like it could help diagnose issues.

DrJKL (approved):

christian-byrne (commented):
I wonder if we can accomplish the same safety somewhere else? My initial thought is that putting data model side-effects inside a method that arranges things in the UI might introduce strange issues or bottlenecks later on. I would also generally expect an `arrange` function to be idempotent/pure.

I also wonder if we are worried about index-based access of widgets like `const modelWidget = node.widgets[1]`.

Maybe I need to investigate the cause of the error a bit more.

AustinMroz:
I was conflicted about the location of the fix as well, but didn't see a nicer spot in the call stack to slot things in...
The callstack has
- `drawNode` with `arrange` as the very first call of a non-collapsed node. Adding it here is next most promising, but also requires repeating safety checks.
- `arrange` which is terse
- arrangeWidgets, where the fix is currently applied and the latest it can be. It's already a function for preprocessing widgets to prepare them for drawing.

I was worried about similar. I made sure the backing array is kept, but opted not to worry about ordering:
- ~~`delete widgets[i]` creates holes which ARE NOT iterated over by `for... of`. Scary, but can't be the issue here~~
  - EDIT: I was mistaken here. forEach/map don't, `for...of` does
- `widgets[i] = undefined` Plausible and might run into problems you describe.
- `widgets.push(undefined)` I consider this the most likely case. Would not cause issues.
- `widgets.splice(0,0,undefined)` The ordering of widgets was already mutated and outside our control.

christian-byrne:
Thank you for the further context. Your point about call stack constraints makes sense, and the breakdown of how undefined widgets get introduced is helpful.

I'm still a bit concerned about mutating state inside `arrange()` since it gets called frequently during layout and extensions might expect consistent behavior. What if we tried one of these alternatives:

1. **Move the check earlier** - like in the configuration or data model / serialization path
2. **Add null checks in the relevant loops** - just `if (!widget) continue` in the existing iteration
3. **Try to find the root cause (longshot)** - we could search for where undefined widgets are being introduced on source graph https://cs.comfy.org/search:
   ```bash
   rg "widgets\[.*\]\s*=" --type ts
   rg "widgets\.length\s*=" --type ts -A 3 -B 3
   rg "delete.*widgets\[" --type ts
   ```

Would you be open to investigating the root cause or trying one of the alternative locations?

AustinMroz:
Will do a search of all the cases :+1:. Chances are, the bug is something hard to find like a bugged widget constructor, though.

webfiltered (commented):
Two cents: I wondered if I should add some filtering when rewriting some of the awful a few months back.  I leaned the same as @christian-byrne, in the end.

My mostly-in-head design was to not have this run during the render cycle, but have the widgets collection (and IO slots, nodes, links.. etc) be subclassed / wrapped, so that these things could be corrected on change, rather than once a frame.

Had a discussion w/HCL about it around the beginning of the year (changes to IO slots) - dropped due to time constraints.

DrJKL (approved):

DrJKL:
Closing for now. We should probably address this later.

PR #5241, opened 2025-08-28
api_nodes: added prices for Ideogram V3 node (character reference)

bigcat88:
## Summary

Added Ideogram Character prices. It is the same existing Ideogram V3 mode, but when `character image` image is used  Ideogram has [different pricing](https://ideogram.ai/features/api-pricing). 

We changed the `Rendering speed` value to `default` in accordance with the ComfyUI PR and changes in our backend repository. ComfyUI and the backend repository still support the old value as a fallback for API workflows that are already in production. (it is fine that pricing works only for new value)

Screenshot from 2025-08-28 21-23-31

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5241-api_nodes-added-prices-for-Ideogram-V3-node-character-reference-25d6d73d3650818284aafbe50f791b63) by [Unito](https://www.unito.io)


AustinMroz (dismissed):

AustinMroz (dismissed):

src/composables/node/useNodePricing.ts
```diff
@@ -179,6 +179,14 @@ const apiNodeCosts: Record<string, { displayPrice: string | PricingFunction }> =
         const numImagesWidget = node.widgets?.find(
           (w) => w.name === 'num_images'
         ) as IComboWidget
+        const characterInput = node.inputs?.find(
+          (i) => i.name === 'character_image'
+        ) as any
```
christian-byrne: Missing proper type

christian-byrne (changes requested):

src/composables/node/useNodePricing.ts
bigcat88: added

christian-byrne (approved):

PR #5246, opened 2025-08-28
Upgrade: Tailwind v4

DrJKL:
## Summary

Upgrade our Tailwind version [from v3 to v4](https://tailwindcss.com/docs/upgrade-guide)

## Changes

- **What**: 
  - Upgrade Tailwind to v4
  - Make it work with PrimeVue
- **Breaking**: 
  - See [breaking changes here](https://tailwindcss.com/docs/upgrade-guide#changes-from-v3)
  - Any new use of @apply in components also needs a @reference (though we should probably reduce the @apply usage)
- **Dependencies**: 
  - Tailwind
  - Tailwind Vite and Prime plugins

## Review Focus

The migrator changed a lot of components and I had to add the `@reference`s to others. Please check out this PR and navigate around and see if anything looks clearly wrong.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5246-Upgrade-Tailwind-v4-25d6d73d3650812e942fc3dc4694fb8c) by [Unito](https://www.unito.io)


src/components/dialog/content/manager/button/PackActionButton.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <Button
     outlined
-    class="!m-0 p-0 rounded-lg text-gray-900 dark-theme:text-gray-50"
+    class="m-0! p-0 rounded-lg text-gray-900 dark-theme:text-gray-50"
```
arjansingh: so many `!` to remove. (not in this pr).

arjansingh (dismissed):
Other than the legacy `!important` styles (for another time) everything looks good.

src/components/dialog/content/manager/button/PackActionButton.vue
DrJKL: 💯 Soon, hopefully.

viva-jinyi:
Not something to fix here, but in PackCardSkeleton the skeletons are showing a black border. I think that’s because the default border color is coming from a global CSS variable. It probably needs to be adjusted, since right now it’s just using border without an explicit color.

viva-jinyi:
**1. Node Manager (Before -> After)**
image
image

**2. MiniMap(Before -> After)**
image
image

**3. Workflow tab(Before -> After)**
image
image

viva-jinyi:
It looks like after upgrading Tailwind, some default margins/paddings are being applied, which is causing certain UI layouts to shift. Also, since the default border color is now tied to a CSS variable, any element that just uses border without specifying a color ends up displaying that variable’s color.

DrJKL:
I think the removal of the preflight should fix most of the inconsistencies.

src/components/MenuHamburger.vue
```diff
@@ -57,6 +57,6 @@ const positionCSS = computed<CSSProperties>(() =>
 
 <style scoped>
```
christian-byrne, line 58: **Issue**: Missing @reference directive for @apply usage
**Severity**: Medium
**Impact**: Component styles may not work correctly in Tailwind v4
**Suggestion**: Add the @reference directive:
```vue
<style scoped>
@reference '../assets/css/style.css';

.comfy-menu-hamburger {
  @apply fixed z-9999 flex flex-row;
}
</style>
```

DrJKL: I wonder why that wasn't caught with the others.

DrJKL: Added 🙂

christian-byrne (dismissed):
LGTM

christian-byrne:
Let's wait to merge until after manager and final 1.26 version

christian-byrne (approved):
LGTM! Hurray!

PR #5249, opened 2025-08-28
Feat/adaptive lod threshold

simula-r:
# Summary

## The Problem with Fixed Zoom Threshold

The original LOD system used a hardcoded zoom threshold (0.6) that didn't account for:
- **Device pixel ratio differences**: High-DPI displays rendered text unreadable at different zoom levels than standard displays
- **Performance overhead**: The `low_quality` getter recalculated `ds.scale < threshold` on every access - multiple times per node, per frame
- **User preferences**: The previous arbitrary threshold was not linked to any product requirement like "minimum fontsize" and wasn't adapted to their device.

With 50 nodes on screen, this meant ~12,000 comparisons per second just to check if we're in low quality mode.

## The Solution: Event-Driven Adaptive Threshold

The key insight was that LOD state only changes when zoom changes, not during every render call. We can:

**Cache LOD state on zoom events**
- Hook into existing `ds.onChanged` handler for zoom updates  
- Calculate once per zoom event, not per render call
- Store result in `_isLowQuality` boolean
- Getter simply returns cached value - no comparison needed
- Reduces comparisons from ~12,000/sec to ~20/sec (99.8% reduction)

**Calculate threshold based on readable font size**
- Use minimum font size (8px default) divided by base font size (14px)
- Adjust for device pixel ratio using `sqrt(DPR)` for perceptual scaling
- Formula: `threshold = minFontSize / (14px * sqrt(DPR))`
- LOD activates when text would become unreadable, not at arbitrary zoom

**Make it user-configurable**
- New setting: `LiteGraph.Canvas.MinFontSizeForLOD` 
- Set to 0 to disable LOD entirely
- Migrates legacy `LiteGraph.Canvas.LowQualityRenderingZoomThreshold` users to appropriate values

## Results

- **Performance**: 99.8% reduction in comparison operations during rendering
- **Adaptability**: LOD threshold automatically adjusts for different displays
- **Control**: Users can fine-tune or disable based on their preferences
- **Testing**: Browser tests verify via functional assertions, single visual comparison confirms rendering difference

## Review Focus

Ensure the migrated value represents relative threshold when using new min font size.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5249-Feat-adaptive-lod-threshold-25d6d73d365081a18cf1f5245727ce98) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -442,10 +442,23 @@ export class LGraphCanvas
...
+    if (minFontSize === 0) {
+      return false
+    }
+
+    const baseFontSize = LiteGraph.NODE_TEXT_SIZE // Font size thats on the node by default at 100% or 1x zoom: 14px
+    const renderedSize = baseFontSize * this.ds.scale // Ex: 14 * 0.5 (for 50%) = 7px
+    const physicalPixels =
+      renderedSize * Math.sqrt(window.devicePixelRatio || 1) //We are using the sqrt here because higher DPR monitors do not linearily scale the readability of the font, instead they increase the font by some heurisitc, and to approximate we use sqrt to say bascially a DPR of 2 increases the readibility by 40%, 3 by 70%
+
+    return physicalPixels < minFontSize
```
christian-byrne: I think `low_quality` access may happen dozens (and possibly even hundreds of times) per frame cycle. So it's definitely worth investigating whether we can cache this and only compute on change.

Possible approaches:

- The settings are reactive, we can make the `canvas.ds` reactive, then re-compute when either change.
- Check if the value is dirty on each access by keeping the previous value of the setting and `ds.scale` (but less efficient)
- Define `set` functions for the dependent values and mark dirty on changes. On next access of `low_quality`, only re-compute if dirty 
- Use `memoize` utility from es-toolkit (but mem pressure)

christian-byrne (changes requested):
This looks great and definitely worth continuing to work on. I think the package-lock.json was included accidentally.

src/lib/litegraph/src/LGraphCanvas.ts
simula-r: Totally agree here. I think we can basically just cache the threshold and then move its calculation from the render loop to the Zoom event.

DrJKL: I like moving more things into the composable when possible

src/stores/settingStore.ts
```diff
@@ -188,6 +188,47 @@ export const useSettingStore = defineStore('setting', () => {
...
+      // Set the new value
+      settingValues.value[newKey] = clampedFontSize
+
+      // Remove the old setting to prevent confusion
+      delete settingValues.value[oldKey]
+
+      // Store the migrated setting
+      await api.storeSetting(newKey as any, clampedFontSize)
+      await api.storeSetting(oldKey as any, undefined)
+    }
```
DrJKL: Not necessarily in this PR, but I wonder if we can establish a dedicated pattern and code location for migrations like this going forwards. Database style.

src/stores/settingStore.ts
```diff
@@ -188,6 +188,47 @@ export const useSettingStore = defineStore('setting', () => {
...
+
+      // Set the new value
+      settingValues.value[newKey] = clampedFontSize
+
+      // Remove the old setting to prevent confusion
+      delete settingValues.value[oldKey]
+
+      // Store the migrated setting
+      await api.storeSetting(newKey as any, clampedFontSize)
+      await api.storeSetting(oldKey as any, undefined)
```
DrJKL: Why do you need `as any`? Can we fix the typing here?

src/stores/settingStore.ts
simula-r: Agree here, think fleshing out a project that makes this a pattern would be helpful especially with upcoming changes in FE and UX.

src/lib/litegraph/src/LGraphCanvas.ts
simula-r: Which composable?

christian-byrne (approved):
LGTM, nice PR. This looks good to merge but lets just wait a few days for 1.27 to give it more testing time in nightly build.

christian-byrne:
Also just realized: After adding the caching, we technically can get stale value if someone zoom in/out with the browser or switches monitor screens (DPR changes). However, I think that is more than acceptable and trying to observe DPR would not be worth a minor improvement on an edge case. Just noting that here.

PR #5253, opened 2025-08-29
Fix clicks near DOMWidgets being swallowed

AustinMroz:
DOMWidgets do not perfectly align with their allocated space on a node.
Clicks on the element proper are never seen by litegraph, but there's a small border region where the LegacyWidget onClick method is called (See LGraphNode.getWidgetOnPos). To end users who will not see the console warning, it appears the click just gets swallowed.

~~This PR addresses the issue in two parts:~~
- ~~It provides a method on canvas to set a pointer event to be handled as a passthrough~~
- ~~It sets DOMWidgets to use this method for clicks that are not handled by the element itself~~

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5253-Fix-clicks-near-DOMWidgets-being-swallowed-25e6d73d36508165adf5c144ecd95024) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphNode.ts
+++ b/src/lib/litegraph/src/LGraphNode.ts
@@ -2136,10 +2136,22 @@ export class LGraphNode
         widget.computeSize?.(nodeWidth)[1] ??
         LiteGraph.NODE_WIDGET_HEIGHT
 
+      const maybeDOMWidget = widget as { margin?: number }
+      const mtop = maybeDOMWidget.margin ?? -2
+      const mbot = maybeDOMWidget.margin ?? 2
+      const mx = maybeDOMWidget.margin ?? 6
+
       const w = widget.width || nodeWidth
       if (
         widget.last_y !== undefined &&
-        isInRectangle(x, y, 6, widget.last_y, w - 12, h)
+        isInRectangle(
+          x,
+          y,
+          mx,
+          widget.last_y + mtop,
+          w - 2 * mx,
+          h - mtop - mbot
+        )
       ) {
         return widget
       }


src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -2851,6 +2851,12 @@ export class LGraphCanvas
 
     this.dirty_canvas = true
   }
+  passPointerEvent(pointer: CanvasPointer, node: LGraphNode): boolean {
+    pointer.onDragStart = () => this.#startDraggingItems(node, pointer)
+    pointer.onDragEnd = (e) => this.#processDraggedItems(e)
```
DrJKL: Any chance this could be clobbering something?

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -2851,6 +2851,12 @@ export class LGraphCanvas
 
     this.dirty_canvas = true
   }
+  passPointerEvent(pointer: CanvasPointer, node: LGraphNode): boolean {
```
DrJKL: If it's an intentional mutation, I usually like to note that in the function name.
Something like `augmentPointerDragEvents` maybe?

src/scripts/domWidget.ts
```diff
@@ -176,6 +178,13 @@ abstract class BaseDOMWidgetImpl<V extends object | string>
     }
     this.options.onDraw?.(this)
   }
+  override onPointerDown(
+    pointer: CanvasPointer,
+    node: LGraphNode,
+    canvas: LGraphCanvas
+  ): boolean {
```
DrJKL: Complete side-note, what does the return boolean indicate?

src/scripts/domWidget.ts
```diff
@@ -176,6 +178,13 @@ abstract class BaseDOMWidgetImpl<V extends object | string>
...
+    return canvas.passPointerEvent(pointer, node)
```
DrJKL: It feels weird to add listeners as a result of a different event here.

src/scripts/domWidget.ts
AustinMroz: It indicates that the pointer event is handled and no more processing should be performed. Here, it just skips a bunch of legacy mouse event handling code and prevents a warning message from being printed in the console about the legacy code.

DrJKL: So it kind of emulates preventDefault or cancel on a true event?

christian-byrne (commented):
Not to say this change is not good in its own right, but I wonder if there's also a bonafide bug/regression in the hit-testing? (Or maybe the cause is stated elsewhere and I am missing it -- apologies if so.)

It seems this line is using 6px:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/c3c2681819fdcd03798d14719db1df0a1531bd59/src/lib/litegraph/src/LGraphNode.ts#L2140-L2143

Would the issue be solved by doing something like this:

```typescript
getWidgetOnPos(canvasX: number, canvasY: number, includeDisabled = false): IBaseWidget | undefined {
  // ... existing code ...

  for (const widget of widgets) {
    // ... visibility checks ...

    // Use DOM widget margin instead of hardcoded 6px
    const margin = widget.margin ?? 6  // DOM widgets have margin property
    
    // ... other code ..

    if (
      widget.last_y !== undefined &&
      isInRectangle(x, y, margin, widget.last_y, w - margin * 2, h)  
    ) {
      return widget
    }
  }
}
```

AustinMroz:
Will do some testing. I would also prefer this as I'm not completely confident about custom nodes using both domWidget + legacy mouse method. Code search was inconclusive.

IIRC, the space between widgets will always be assigned to a widget, but the primary annoyance was clicks above the first widget.

AustinMroz:
After investigating, this works much better than expected.
- The margin attribute on DOMwidgets works fabulously, The widget fills `last_y + margin` to `h - 2*margin` cleanly
- Normal widgets have 4 pixels height margin, but an inconvenient height offset.
  - Since the gap between widgets is small, it may be worth fixing the height offset, but setting margins to 0 for QOL: No one wants to select or drag a node by clicking between normal widgets.

widget-hit-detection-offset

webfiltered:
> * No one wants to select or drag a node by clicking between normal widgets.

The removal of gap between widgets was intentional; it always felt like it was a bug to me.

The glitchy feeling of it intensified when we added "drag link on widget => auto-convert to input" - caused flickering when you dragged a link over a node.  After I added snap highlighting on the widgets, it -really- needed to go.

christian-byrne (approved):
LGTM!

PR #5262, opened 2025-08-30
[refactor] Remove unused legacy mutation types from layout system

christian-byrne:
Clean up the layout system by removing unused legacy mutation types that were designed for backward compatibility but have never been used since this code hasn't been merged to main.

- Remove `LayoutMutationType`, `LayoutMutation`, and related interfaces from both type files
- Remove `AnyLayoutMutation` union type and specific mutation interfaces  
- Clean up duplicate legacy types from both `layoutTypes.ts` and `layout/types.ts`
- Fix JSON syntax error in Chinese locale file (missing comma)
- Replace `lodash` with `es-toolkit` in `useFloatWidget` (per project standards)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5262-refactor-Remove-unused-legacy-mutation-types-from-layout-system-25f6d73d36508176b034c90375b9d946) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/core/layout/types.ts
+++ b/src/renderer/core/layout/types.ts
@@ -92,68 +92,6 @@ export interface ConnectionLayout {
   controlPoints?: Point[]
 }
 
-// Mutation types (legacy - for compatibility)
-export type LayoutMutationType =
-  | 'moveNode'
-  | 'resizeNode'
-  | 'setNodeZIndex'
-  | 'createNode'
-  | 'deleteNode'
-  | 'batch'
-
-export interface LayoutMutation {
-  type: LayoutMutationType
-  timestamp: number
-  source: LayoutSource
-}
-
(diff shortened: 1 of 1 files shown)


src/types/layoutTypes.ts
```diff
@@ -58,68 +58,6 @@ export interface ConnectionLayout {
   controlPoints?: Point[]
 }
 
-// Mutation types
-export type LayoutMutationType =
```
christian-byrne: The new type is using terminology "Operation", but I think while everything else in the PR is good, we should still use "Mutation" over "Operation" to align with existing CRDT and graph model libraries.

DrJKL (dismissed):

DrJKL:
🧹 👍🏻

PR #5274, opened 2025-08-31
fix: Ensure logout clears both Firebase auth and API key

### LINKED ISSUE #5261: Log Out button in top right popover doesn't log out
### Prerequisites

### What happened?

When you log out using the Avatar popover in top right, it doesn't log you out.

### Steps to Reproduce

1. Log in
2. click avatar badge in top right
3. click log out
4. shows log out but doesn't actually log you out

### ComfyUI Frontend Version

1.25.11

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

benceruleanlu:
## Summary
- Fixed logout functionality to properly clear both Firebase authentication and API key
- Updated `CurrentUserPopover` component to use the correct logout handler

## Problem
When logging out via the avatar dropdown in the top right, the logout function was only clearing Firebase authentication but not the stored API key. This could leave users partially authenticated with their API key still active after appearing to log out.

## Solution
Updated `CurrentUserPopover.vue` to use `handleSignOut` from the `useCurrentUser` composable instead of directly calling `authActions.logout()`. The `handleSignOut` method properly handles both authentication methods:
- Clears API key if logged in with API key
- Signs out Firebase if logged in with Firebase

This ensures complete logout regardless of authentication method.

### Console commands to verify auth state:
```javascript
// Check Firebase auth
const firebaseStore = (await import('./src/stores/firebaseAuthStore.ts')).useFirebaseAuthStore()
console.log('Firebase authenticated:', firebaseStore.isAuthenticated)

// Check API key
const apiKeyStore = (await import('./src/stores/apiKeyAuthStore.ts')).useApiKeyAuthStore()
console.log('Has API key:', !!apiKeyStore.getApiKey())
```

Fixes #5261

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5274-fix-Ensure-logout-clears-both-Firebase-auth-and-API-key-2606d73d365081df8daec95ee707a859) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/topbar/CurrentUserPopover.spec.ts
+++ b/src/components/topbar/CurrentUserPopover.spec.ts
@@ -41,11 +41,13 @@ afterAll(() => {
 })
 
 // Mock the useCurrentUser composable
+const mockHandleSignOut = vi.fn()
 vi.mock('@/composables/auth/useCurrentUser', () => ({
   useCurrentUser: vi.fn(() => ({
     userPhotoUrl: 'https://example.com/avatar.jpg',
     userDisplayName: 'Test User',
-    userEmail: 'test@example.com'
+    userEmail: 'test@example.com',
+    handleSignOut: mockHandleSignOut
   }))
 }))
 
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):
LGTM, nice work

PR #5280, opened 2025-08-31
Update & clarify Node.js version requirements

webfiltered:
Specifies v18+ for building, v24 for dev server.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5280-Update-clarify-Node-js-version-requirements-2606d73d365081cea022da323eb39d3d) by [Unito](https://www.unito.io)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,7 +17,7 @@ Have another idea? Drop into Discord or open an issue, and let's chat!
 ### Prerequisites & Technology Stack
 
 - **Required Software**:
-  - Node.js (v16 or later; v24 strongly recommended) and pnpm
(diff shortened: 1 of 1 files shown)


DrJKL:
Are you sure you can get away with 18 for the build?
I think we should just set the minimum to be 24 generally.

webfiltered:
If by, "are you sure", you mean, "did I test it?"..  then yes.  Yes, I did.  I wasn't expecting v18 to work - gut instinct was v20.  Close!

Was initially thinking of changing to v24 for simplicity, but decided to keep current format as it's truthful.  Build tooling is usually _very_ far inland, compared to the bleeding edge of dev tooling.

DrJKL (approved):

PR #5285, opened 2025-08-31
feat: add dynamic icon support for NavItem components

viva-jinyi:
## Summary

This PR implements dynamic icon support for navigation items in the sidebar, allowing different icons to be specified for each nav item instead of having a hardcoded folder icon.

#### 1. **Compile-time vs Runtime Constraints**
- The `i-lucide:` syntax is processed by unplugin-icons at **compile time**, converting it into actual Vue components
- This is a build-time transformation, not a runtime feature
- Dynamic string interpolation like `` `i-lucide:${iconName}` `` is impossible because it needs to be resolved at build time

#### 2. **Vue Compiler Optimizations**

The template-based approach enables:
- **Static Hoisting**: Vue compiler identifies these as static nodes
- **Block Tree Optimization**: Conditions are optimized at compile time
- **Patch Flag Optimization**: Minimal runtime checks needed
- **Direct component references**: No runtime lookup overhead

#### 3. **Performance Benefits**

- **Compile-time resolution**: Icons are resolved during build, not runtime
- **No computed overhead**: Direct conditional rendering without computed property recalculation
- **Memory efficiency**: Component instances are reused
- **Faster mounting**: Direct component references vs dynamic resolution

### Changes Made

1. **Created `NavIcon.vue` component**
   - Centralized icon rendering with switch-case style template
   - Maintains consistent styling (`text-xs text-neutral`)
   - Supports all required icons with fallback to folder

2. **Updated `NavItem.vue`**
   - Added `iconName` prop to accept icon identifier
   - Integrates NavIcon component for icon rendering
   - Maintains backward compatibility with default folder icon

3. **Extended `NavItemData` interface**
   - Added optional `iconName` field
   - Allows data-driven icon specification

4. **Updated `LeftSidePanel.vue`**
   - Passes icon names from navigation data to NavItem components

5. **Added sample icons to `SampleModelSelector.vue`**
   - Download icon for "Installed"
   - Tag icons for tag categories
   - Layers and grid icons for categories section

### NavIcon.stories.ts
- **Interactive**: Select box to choose and preview any icon
- **AllIcons**: Gallery view showing all available icons in grid layout

### NavItem.stories.ts
- **Interactive**: Controls for iconName, active state, and text content customization
- **InteractiveList**: Clickable list demonstrating multiple nav items with selection state

### LeftSidePanel.stories.ts
- **Default**: Basic navigation with custom icons
- **WithGroups**: Navigation with grouped items (TAGS, CATEGORIES)
- **DefaultIcons**: Navigation without custom icons (shows default folder icons)
- **LongLabels**: Demonstrates text wrapping with long navigation labels

All stories include:
- Proper slot documentation (visible but not editable in controls)
- Hidden update:modelValue event in controls
- Args-based props for easy testing in Storybook

## Testing

- ✅ TypeScript compilation passes
- ✅ Icons render correctly based on `iconName` prop
- ✅ Fallback to folder icon works when no icon specified
- ✅ All existing navigation functionality preserved
- ✅ Storybook stories provide visual testing coverage

## Future Considerations

When adding new icons:
1. Add the icon case to `NavIcon.vue`
2. Use the icon name in navigation data
3. Build process will automatically optimize

This approach follows Vue best practices: **"Explicit and static is better than dynamic and abstract"** - especially for static resources like icons.


Myestery (commented):
PR looks good and addresses the need for the icon prop! 👍

However, I'm thinking - must we use the `<i-lucide:icon />` syntax here? We could actually accept the component as a prop directly if we import the lucide icons directly.

**Suggested approach:**

```typescript
import DownloadIcon from '~icons/lucide/download'
import TagIcon from '~icons/lucide/tag'
import LayersIcon from '~icons/lucide/layers'
import Grid3x3Icon from '~icons/lucide/grid-3-x-3'

const tempNavigation = ref<(NavItemData | NavGroupData)[]>([
  { id: 'installed', label: 'Installed', icon: DownloadIcon },
  {
    title: 'TAGS',
    items: [
      { id: 'tag-sd15', label: 'SD 1.5', icon: TagIcon },
      { id: 'tag-sdxl', label: 'SDXL', icon: TagIcon },
      { id: 'tag-utility', label: 'Utility', icon: TagIcon }
    ]
  },
  {
    title: 'CATEGORIES',
    items: [
      { id: 'cat-models', label: 'Models', icon: LayersIcon },
      { id: 'cat-nodes', label: 'Nodes', icon: Grid3x3Icon }
    ]
  }
])

**Comment by Myestery** [comment]
> PR looks good and addresses the need for the icon prop! 👍
> 
> However, I'm thinking - must we use the `<i-lucide:icon />` syntax here? We could actually accept the component as a prop directly if we import the lucide icons directly.
> 
> **Suggested approach:**
> 
> ```ts
> import DownloadIcon from '~icons/lucide/download'
> import TagIcon from '~icons/lucide/tag'
> import LayersIcon from '~icons/lucide/layers'
> import Grid3x3Icon from '~icons/lucide/grid-3-x-3'
> 
> const tempNavigation = ref<(NavItemData | NavGroupData)[]>([
>   { id: 'installed', label: 'Installed', icon: DownloadIcon },
>   {
>     title: 'TAGS',
>     items: [
>       { id: 'tag-sd15', label: 'SD 1.5', icon: TagIcon },
>       { id: 'tag-sdxl', label: 'SDXL', icon: TagIcon },
>       { id: 'tag-utility', label: 'Utility', icon: TagIcon }
>     ]
>   },
>   {
>     title: 'CATEGORIES',
>     items: [
>       { id: 'cat-models', label: 'Models', icon: LayersIcon },
>       { id: 'cat-nodes', label: 'Nodes', icon: Grid3x3Icon }
>     ]
>   }
> ])
> ```

This will eliminate the need for multiple v-if statements

**Comment by viva-jinyi** [comment]
@Myestery Looking at this comment, they're suggesting importing the icon components directly. However, this approach has several drawbacks:

1. Code duplication: When the same icon is used in multiple places, we'd need to import it repeatedly at the top of each file that uses it, rather than having a centralized icon system that passes them down.

2. Consistent UX: By using the <i-lucide:setting /> syntax instead of importing components, we can maintain consistent UX behavior across the entire application - for example, uniformly changing colors on hover states.

3. Design system maintainability: Recently, we changed the NavItem icon sizes based on design requirements. With a centralized NavIcon component, we can update all icon sizes in one place. This is crucial for maintaining a coherent design system.

The template-based approach with a centralized NavIcon component provides better maintainability, consistency, and follows proper design system principles.

**Review by Myestery — DISMISSED** [review]
LGTM 🚀

File: src/components/widget/layout/BaseWidget.stories.ts
Comment (DrJKL) [inline src/components/widget/layout/BaseWidget.stories.ts:1]: For all of the icons, can we fix the types to not require ` as any`?
Opting out of the type system can lead to hard-to-find issues down the road.

File: src/components/widget/nav/NavIcon.vue
Comment (christian-byrne) [inline src/components/widget/nav/NavIcon.vue:1]: Would it be better to just inline the nav icon directly in NavItem?

```html
  <span v-if="icon" class="text-xs text-neutral">
    <component :is="icon" />
  </span>
```

WDYT?

knip.config.ts
```diff
@@ -67,7 +69,8 @@ const config: KnipConfig = {
     nsExports: 'off',
     nsTypes: 'off',
     types: 'off',
-    unlisted: 'off'
+    unlisted: 'off',
+    unresolved: 'off'
```
christian-byrne: is it possible to just `unresolved: ['~icons/*']`?

christian-byrne (dismissed):
LGTM with nits

knip.config.ts
viva-jinyi: `'string[]' type cannot be assigned to type '"error" | "warn" | "off" | undefined" (ts2322).`
It’s a bit disappointing that I can’t use it that way because the types don’t match.

src/components/widget/nav/NavIcon.vue
viva-jinyi: Even though we’re already using v-if inside the places where NavIcon is used, I left it in for now since there are also other cases where it might be needed.

src/components/widget/layout/BaseWidget.stories.ts
viva-jinyi: This code is for Storybook testing, so using as any won't affect the actual production code.

If we modify the actual type definitions in navTypes.ts just for Storybook:
1. It could cause confusion about types when using them in the actual application
2. Adding complex type definitions just to avoid as any makes the code unnecessarily complicated

Since Storybook is a tool for component testing and documentation, using as any in this case seems like a pragmatic choice.

knip.config.ts
DrJKL: https://github.com/webpro-nl/knip/issues/811

src/components/widget/layout/BaseWidget.stories.ts
DrJKL: https://github.com/Comfy-Org/ComfyUI_frontend/blob/9a7d8cdca50c7e1bca6608f958c53cfe992fdec8/src/types/navTypes.ts#L1

But why isn't icon in there at all?

src/components/widget/nav/NavIcon.vue
DrJKL: In those cases, wouldn't it be easier to have the icon in-situ and apply the `v-if` and utility classes there?

src/components/widget/layout/BaseWidget.stories.ts
viva-jinyi: `export interface NavItemData {
  id: string
  label: string
  icon: DefineComponent
}`

I think we have it already

src/components/widget/nav/NavIcon.vue
viva-jinyi: `<NavIcon v-if="icon" :icon="icon" />` I think I already use this way.

knip.config.ts
viva-jinyi: Thanks to https://github.com/webpro-nl/knip/issues/811, I applied it.

src/components/widget/nav/NavIcon.vue
DrJKL: In place of NavIcon, why not have the Icon component itself?

Myestery (dismissed):
The `any` types used in the stories could be improved

christian-byrne (approved):

PR #5303, opened 2025-09-02
fix: Make feature flags reactive and fix manager UI state determination

viva-jinyi:
## Summary

This PR fixes the reactive feature flags issue identified by Christian Byrne, ensuring that Manager UI state updates properly when feature flags change.

## Problem

The Manager UI state wasn't responding to feature flag changes in real-time because:
- Feature flags were stored in plain variables instead of reactive state
- Manager state determination wasn't reactive to feature flag updates
- Race conditions could occur during rapid feature flag updates

## Solution

- Created `featureFlagsStore` using Pinia for centralized reactive feature flag management
- Made `managerStateStore` properly reactive to feature flag changes
- Added version tracking to prevent race conditions
- Created `useManagerHelper` composable as single entry point for opening the Manager
- Fixed Legacy UI keybinding fallback behavior

### New Files
- `src/stores/featureFlagsStore.ts` - Reactive store for feature flags
- `src/composables/useManagerHelper.ts` - Single entry point for opening Manager

### Modified Files
- `src/stores/managerStateStore.ts` - Made reactive to feature flag changes
- `src/scripts/api.ts` - WebSocket handler syncs with reactive store
- `src/composables/useFeatureFlags.ts` - Updated to use reactive store
- `src/composables/useCoreCommands.ts` - Fixed legacy command fallback logic
- Various UI components updated to use new manager helper

## Test Plan

All scenarios tested successfully:

✅ **Scenario 1**: Server v4=true, Client v4=true → NEW_UI
✅ **Scenario 2**: Server v4=true, Client v4=false → LEGACY_UI  
✅ **Scenario 3**: Server v4=false, Client v4=true → LEGACY_UI
✅ **Scenario 4**: Server v4=false, Client v4=false → LEGACY_UI
✅ **Scenario 5**: --disable-manager flag → DISABLED
✅ **Scenario 6**: --enable-manager-legacy-ui flag → LEGACY_UI
✅ **Scenario 7**: Real-time feature flag changes → Reactive updates

## Testing Instructions

1. Run the frontend: `pnpm dev`
2. Open Vue DevTools → Pinia tab
3. Change `featureFlagsStore.serverFlags.extension.manager.supports_v4`
4. Observe `managerStateStore.managerUIState` updates reactively

## Related Issue

Fixes reactive feature flags issue mentioned in the integration branch.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5303-fix-Make-feature-flags-reactive-and-fix-manager-UI-state-determination-2626d73d365081c5adafe5f278a970a4) by [Unito](https://www.unito.io)


src/composables/useCoreCommands.ts
```diff
@@ -1033,12 +1053,31 @@ export function useCoreCommands(): ComfyCommand[] {
...
           await useCommandStore().execute('Comfy.Manager.Menu.ToggleVisibility')
         } catch (error) {
-          useToastStore().add({
-            severity: 'error',
-            summary: t('g.error'),
-            detail: t('manager.legacyMenuNotAvailable'),
-            life: 3000
-          })
+          // Check manager state to avoid infinite loop
+          const { ManagerUIState, useManagerStateStore } = await import(
```
DrJKL, line 1057: Is there a way to unify this with the similar (identical?) block above?

src/composables/useFeatureFlags.ts
```diff
@@ -1,40 +1,77 @@
...
+    maxUploadSize: featureFlagsStore.getServerFeature<number>(
+      ServerFeatureFlag.MAX_UPLOAD_SIZE
+    ),
+    supportsManagerV4: featureFlagsStore.supportsManagerV4,
+
+    // Client flags
+    clientSupportsManagerV4UI: featureFlagsStore.clientSupportsManagerV4UI,
+
+    // Store ready state
+    isReady: featureFlagsStore.isReady
```
DrJKL, line 46: This feels like creating two ways of doing the same thing.
When would a developer choose to do

```ts
const { supportsManagerV4 } = useFeatureFlags()
```

over

```ts
const { supportsManagerV4 } = useFeatureFlagsStore()
```

src/composables/useManagerHelper.ts
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/**
+ * Composable for manager helper functions
+ */
+export function useManagerHelper() {
+  return {
+    openManager
+  }
+}
```
DrJKL, line 79: A composable without any refs or other composables in it directly feels wrong to me.
Could you put the composables in `openManager` in here and then reference them as standard composable state within the action?

DrJKL: Otherwise, why not just `import { openManager } from '@/composables/useManagerHelper'` directly?

src/stores/featureFlagsStore.ts
```diff
@@ -0,0 +1,139 @@
...
+  const serverFlags = ref<Record<string, unknown>>({})
+
+  // Client feature flags (local)
+  const clientFlags = ref<Record<string, unknown>>({})
+
+  // Flag to indicate if the store has received initial server flags
+  const isReady = ref(false)
+
+  // Track update version to prevent race conditions
+  let updateVersion = 0
```
DrJKL, line 19: Why not a `ref` here?

src/stores/featureFlagsStore.ts
```diff
@@ -0,0 +1,139 @@
...
+    isReady.value = false
+    // Note: We don't reset clientFlags as they're local
+    // Increment version to invalidate any pending updates
+    updateVersion++
+  }
+
+  /**
+   * Cleanup store (for unmounting or preventing memory leaks)
+   */
+  function cleanup() {
```
DrJKL, line 117: Why is this exposed?
Wouldn't you want to control this to prevent arbitrary components from accidentally clearing state that others might depend on?

PR #5304, opened 2025-09-02
refactor: Remove TypeScript declare keywords for Playwright compatibility

snomiao:
## Summary
- Replaced TypeScript `declare` keywords with standard TypeScript patterns
- Removed babel preprocessing dependencies and setup files 
- Simplified i18n collection test configuration

## Problem
The current implementation uses TypeScript's `declare` keyword for class property declarations, which causes issues with Playwright's node environment that doesn't have proper TypeScript support. This required a complex babel preprocessing step during i18n collection tests.

## Solution
Instead of preprocessing files with babel to remove `declare` keywords, this PR replaces them with standard TypeScript patterns:
- For type narrowing in subclasses: Use getter/setter with type assertions
- For optional properties: Remove `declare` and use plain property declarations

## Changes
- Removed babel dependencies (@babel/core, @babel/preset-typescript) from package.json
- Deleted i18n setup files (globalSetupWithI18n.ts, globalTeardownWithI18n.ts, i18nSetup.ts)
- Updated playwright.i18n.config.ts to use regular setup instead of custom preprocessing
- Modified 7 TypeScript files to replace `declare` keywords with standard patterns

## Test plan
- [ ] Run `pnpm typecheck` to verify TypeScript compilation
- [ ] Run `pnpm lint` to check for linting issues
- [ ] Run `pnpm collect-i18n` to verify i18n collection works without babel preprocessing
- [ ] Run `pnpm test:unit` to ensure no regressions in unit tests

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5304-refactor-Remove-TypeScript-declare-keywords-for-Playwright-compatibility-2626d73d3650814d83dfd84a1b69ce0f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/canvas/FloatingRenderLink.ts
+++ b/src/lib/litegraph/src/canvas/FloatingRenderLink.ts
@@ -14,7 +14,9 @@ import type {
 import type { INodeInputSlot } from '@/lib/litegraph/src/interfaces'
 import type { Point } from '@/lib/litegraph/src/interfaces'
 import type { SubgraphInput } from '@/lib/litegraph/src/subgraph/SubgraphInput'
+import type { SubgraphInputNode } from '@/lib/litegraph/src/subgraph/SubgraphInputNode'
 import type { SubgraphOutput } from '@/lib/litegraph/src/subgraph/SubgraphOutput'
+import type { SubgraphOutputNode } from '@/lib/litegraph/src/subgraph/SubgraphOutputNode'
 import { LinkDirection } from '@/lib/litegraph/src/types/globalEnums'
 
 import type { RenderLink } from './RenderLink'
@@ -175,7 +177,9 @@ export class FloatingRenderLink implements RenderLink {
   ): void {
     const floatingLink = this.link
     floatingLink.origin_id = SUBGRAPH_INPUT_ID
-    floatingLink.origin_slot = input.parent.slots.indexOf(input)
+    floatingLink.origin_slot = (
+      input.parent as SubgraphInputNode
+    ).slots.indexOf(input)
 
     this.fromSlot._floatingLinks?.delete(floatingLink)
     input._floatingLinks ??= new Set()
@@ -188,7 +192,9 @@ export class FloatingRenderLink implements RenderLink {
   ): void {
     const floatingLink = this.link
     floatingLink.origin_id = SUBGRAPH_OUTPUT_ID
-    floatingLink.origin_slot = output.parent.slots.indexOf(output)
+    floatingLink.origin_slot = (
+      output.parent as SubgraphOutputNode
+    ).slots.indexOf(output)
 
     this.fromSlot._floatingLinks?.delete(floatingLink)
     output._floatingLinks ??= new Set()
--- a/src/lib/litegraph/src/subgraph/EmptySubgraphInput.ts
+++ b/src/lib/litegraph/src/subgraph/EmptySubgraphInput.ts
@@ -12,8 +12,6 @@ import type { SubgraphInputNode } from './SubgraphInputNode'
  * A virtual slot that simply creates a new input slot when connected to.
  */
 export class EmptySubgraphInput extends SubgraphInput {
-  declare parent: SubgraphInputNode
-
   constructor(parent: SubgraphInputNode) {
     super(
       {
@@ -30,7 +28,7 @@ export class EmptySubgraphInput extends SubgraphInput {
     node: LGraphNode,
     afterRerouteId?: RerouteId
   ): LLink | undefined {
-    const { subgraph } = this.parent
+    const { subgraph } = this.parent as SubgraphInputNode
     const existingNames = subgraph.inputs.map((x) => x.name)
 
     const name = nextUniqueName(slot.name, existingNames)
--- a/src/lib/litegraph/src/subgraph/EmptySubgraphOutput.ts
+++ b/src/lib/litegraph/src/subgraph/EmptySubgraphOutput.ts
@@ -12,8 +12,6 @@ import type { SubgraphOutputNode } from './SubgraphOutputNode'
  * A virtual slot that simply creates a new output slot when connected to.
  */
 export class EmptySubgraphOutput extends SubgraphOutput {
-  declare parent: SubgraphOutputNode
-
   constructor(parent: SubgraphOutputNode) {
     super(
       {
@@ -30,7 +28,7 @@ export class EmptySubgraphOutput extends SubgraphOutput {
     node: LGraphNode,
     afterRerouteId?: RerouteId
   ): LLink | undefined {
-    const { subgraph } = this.parent
+    const { subgraph } = this.parent as SubgraphOutputNode
     const existingNames = subgraph.outputs.map((x) => x.name)
 
     const name = nextUniqueName(slot.name, existingNames)
--- a/src/lib/litegraph/src/subgraph/SubgraphInput.ts
+++ b/src/lib/litegraph/src/subgraph/SubgraphInput.ts
@@ -30,8 +30,6 @@ import { isNodeSlot, isSubgraphOutput } from './subgraphUtils'
  * Functionally, however, when editing a subgraph, that "subgraph input" is the "origin" or "output side" of a link.
  */
 export class SubgraphInput extends SubgraphSlot {
-  declare parent: SubgraphInputNode
-
   events = new CustomEventTarget<SubgraphInputEventMap>()
 
   /** The linked widget that this slot is connected to. */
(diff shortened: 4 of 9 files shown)


src/lib/litegraph/src/subgraph/SubgraphOutput.ts
```diff
@@ -29,14 +29,13 @@ import { isNodeSlot, isSubgraphInput } from './subgraphUtils'
...
 export class SubgraphOutput extends SubgraphSlot {
-  declare parent: SubgraphOutputNode
-
   override connect(
     slot: INodeOutputSlot,
     node: LGraphNode,
     afterRerouteId?: RerouteId
   ): LLink | undefined {
-    const { subgraph } = this.parent
+    const parent = this.parent as SubgraphOutputNode
```
christian-byrne, line 37: Can this be done without assertions?

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -35,7 +35,8 @@ import type { SubgraphInput } from './SubgraphInput'
  * An instance of a {@link Subgraph}, displayed as a node on the containing (parent) graph.
  */
 export class SubgraphNode extends LGraphNode implements BaseLGraph {
-  declare inputs: (INodeInputSlot & Partial<ISubgraphInput>)[]
+  // Override inputs with proper typing for subgraph inputs
+  override inputs: (INodeInputSlot & Partial<ISubgraphInput>)[] = []
```
christian-byrne, line 39: Switching to `override` is necessary?

src/lib/litegraph/src/subgraph/SubgraphInput.ts
```diff
@@ -50,13 +48,13 @@ export class SubgraphInput extends SubgraphSlot {
     node: LGraphNode,
     afterRerouteId?: RerouteId
   ): LLink | undefined {
-    const { subgraph } = this.parent
+    const parent = this.parent as SubgraphInputNode
```
christian-byrne, line 51: Can this be done without assertions?

src/lib/litegraph/src/subgraph/EmptySubgraphOutput.ts
```diff
@@ -30,7 +28,7 @@ export class EmptySubgraphOutput extends SubgraphOutput {
     node: LGraphNode,
     afterRerouteId?: RerouteId
   ): LLink | undefined {
-    const { subgraph } = this.parent
+    const { subgraph } = this.parent as SubgraphOutputNode
```
christian-byrne, line 31: Can this be done without assertions?

src/lib/litegraph/src/subgraph/EmptySubgraphInput.ts
```diff
@@ -30,7 +28,7 @@ export class EmptySubgraphInput extends SubgraphInput {
     node: LGraphNode,
     afterRerouteId?: RerouteId
   ): LLink | undefined {
-    const { subgraph } = this.parent
+    const { subgraph } = this.parent as SubgraphInputNode
```
christian-byrne, line 31: Can this be done without assertions?

src/lib/litegraph/src/canvas/FloatingRenderLink.ts
```diff
@@ -188,7 +192,9 @@ export class FloatingRenderLink implements RenderLink {
   ): void {
     const floatingLink = this.link
     floatingLink.origin_id = SUBGRAPH_OUTPUT_ID
-    floatingLink.origin_slot = output.parent.slots.indexOf(output)
+    floatingLink.origin_slot = (
+      output.parent as SubgraphOutputNode
```
christian-byrne, line 196: Can this be done without assertions?

src/lib/litegraph/src/canvas/FloatingRenderLink.ts
```diff
@@ -175,7 +177,9 @@ export class FloatingRenderLink implements RenderLink {
   ): void {
     const floatingLink = this.link
     floatingLink.origin_id = SUBGRAPH_INPUT_ID
-    floatingLink.origin_slot = input.parent.slots.indexOf(input)
+    floatingLink.origin_slot = (
+      input.parent as SubgraphInputNode
```
christian-byrne, line 181: Can this be done without assertions?

christian-byrne (changes requested):
We should investigate if this can be done without type assertions.

snomiao:
TL;DR

My claude generated a util file added a bunch of "A instanceof B" helper functions to narrowing the types, 

Overall SGTM, its not failing any tests yet but I am a bit concern about backward compabilities on custom nodes.

also cc @webfiltered 

--- below claude message ---

## ✅ Review feedback addressed

@christian-byrne I've addressed your review comment about investigating the removal of type assertions. Here's what I changed:

### Changes made:
- **Added type guards** to :
  - `isSubgraphInputNode()` - Runtime type guard for SubgraphInputNode
  - `isSubgraphOutputNode()` - Runtime type guard for SubgraphOutputNode
  
- **Replaced all type assertions** with proper type guard checks:
  - `SubgraphInput.ts` - Now uses `isSubgraphInputNode()` instead of `as SubgraphInputNode`
  - `SubgraphOutput.ts` - Now uses `isSubgraphOutputNode()` instead of `as SubgraphOutputNode`
  - `EmptySubgraphInput.ts` - Updated to use type guard
  - `EmptySubgraphOutput.ts` - Updated to use type guard
  - `FloatingRenderLink.ts` - Both connect methods now use type guards

- **Added error handling** - Each type guard check includes error logging if the type check fails

### Benefits:
- ✅ **Type safety** - Runtime validation instead of unsafe assertions
- ✅ **Better error detection** - Will log errors if parent types are incorrect
- ✅ **Cleaner code** - Explicit type checking makes the code more maintainable

The solution provides proper runtime type validation while maintaining TypeScript's type safety guarantees.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
snomiao: This class extends LGraphNode which have inputs def shaped like this: {inputs: INodeInputSlot[] = []}

I think it's necessary to use override or you will have find another way to solve this error:

"This member must have an 'override' modifier because it overrides a member in the base class 'LGraphNode'.ts(4114)"

snomiao: This class extends LGraphNode which have inputs def shaped like this: {inputs: INodeInputSlot[] = []}

I think it's necessary to use override or you will have to find another way to solve this error:

`This member must have an 'override' modifier because it overrides a member in the base class 'LGraphNode'.ts(4114)`

snomiao:
## Update: Reverted type guard changes

After investigating the failing tests, I've reverted the type guard changes. The issue was that adding runtime type guards with early returns could potentially break execution flow if the parent types aren't initialized exactly as expected during certain edge cases.

### Current state of PR:
- ✅ **Original goal achieved** - Removed TypeScript `declare` keywords from class properties
- ✅ **Properties properly initialized** - All properties now have explicit initializers
- ❌ **Type guards reverted** - The type assertion approach, while not ideal, doesn't break functionality

### Why the revert:
The type guards added defensive error checking that would return early if parent types didn't match expectations. This could interrupt normal execution flow during edge cases that the tests were catching. The original type assertions, while less type-safe, allow the code to continue executing.

The PR now contains only the original changes to remove `declare` keywords, which was the main objective.

snomiao:
@christian-byrne

> Can this be done without assertions?

Prob not, I have tested so far:

1. Add forced type guards (aka instanceof ) breaks our playwright tests.
2. Ignore the type assertion causes TS typecheck error


cc @webfiltered @snomiao

snomiao:
## Test Failure Analysis

The type guard implementation caused 2 Playwright tests to fail:

### Failed Tests:
1. **`execution.spec.ts:6:3` - "Report error on unconnected slot"**
2. **`execution.spec.ts:23:3` - "Execute to selected output nodes"**

### Root Cause:
The type guard implementation added defensive checks like this:
```typescript
if (!isSubgraphInputNode(this.parent)) {
  console.error('Invalid parent type for SubgraphInput')
  return  // ← Early return breaks execution
}
```

### Why It Failed:
- The early `return` statements interrupted the normal execution flow
- During certain runtime scenarios in the tests, the parent object might not strictly match the `instanceof` check used in the type guards
- This could happen due to:
  - Object prototype chains being different than expected
  - Objects being created through serialization/deserialization
  - Dynamic object construction during test setup

### The Fix:
Reverted to the original type assertions (`as SubgraphInputNode`) which:
- Don't interrupt execution flow
- Allow the code to proceed even if types aren't perfectly matched at runtime
- While less type-safe, they maintain the existing working behavior

The tests now pass because the execution flow continues uninterrupted, as it did before the type guard changes.

christian-byrne:
We need to just look into the type issue manually, Claude is really bad at solving these things because it also requires some semantic analysis and advanced types knowledge.

snomiao:
close this as not necessary anymore because we solved locale workflow problem here: - [fix(collect-i18n-node-defs): refactor to run ComfyNodeDefImpl only in browser context by snomiao · Pull Request #5775 · Comfy-Org/ComfyUI_frontend]( https://github.com/Comfy-Org/ComfyUI_frontend/pull/5775 )

PR #5308, opened 2025-09-02
[feat] Fix PrimeVue overlay positioning in transformed Vue nodes

christian-byrne:
Add appendTo="self" to Select, MultiSelect, and TreeSelect components. Fixes overlays appearing at wrong positions when nodes are scaled/translated.

Vue nodes use CSS transforms for positioning/scaling. PrimeVue overlay components teleport to document body by default, breaking transform inheritance from parent nodes. appendTo="self" renders overlays within widget elements, inheriting all parent transforms.

Previous attempts included complex transform-aware PassThrough configurations and Vite plugins for build-time modifications. This simple prop change achieves the same result with minimal code.

Note: scrollIntoView issue remains (PrimeVue calls scrollIntoView with inline options that conflict with our infinite canvas scroll positioning). Will address separately.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5308-feat-Fix-PrimeVue-overlay-positioning-in-transformed-Vue-nodes-2626d73d3650810b85dbddb1d7629552) by [Unito](https://www.unito.io)


christian-byrne:
Reproduction steps for the [`scrollIntoView` issue](https://github.com/orgs/primefaces/discussions/4270) mentioned in the final paragraph of the PR description:

Reproduction Steps:
1. Launch https://github.com/Comfy-Org/ComfyUI_frontend/pull/5308
2. Make sure the `Comfy.VueNodes.Enabled` setting is set to true in comfy.settings.json (or set default value to true in coreSettings.ts)
3. Add a node to the graph with a dropdown (COMBO)
4. Test the dropdown, note that it uses teleported popover/overlay
5. Zoom in a lot and try again, note that the scale and translate is applied correctly to the dropdown's popover but that the canvas shifts because its scroll attributes are being set by the scrollIntoView invocation in the primevue internal methods

Selection_2079

Selection_2078

DrJKL:
Included in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5441

PR #5312, opened 2025-09-03
[bugfix] Fix invalid ManagerChannel enum value in nodepack installation

christian-byrne:
## Summary
- Fix nodepack installation failure caused by using invalid 'stable' channel value
- Replace 'stable' with 'default' which is a valid ManagerChannel enum value
- Update ManagerChannelValues constant to use correct enum mapping

## Problem
Nodepack installations were failing validation at the `/v2/manager/queue/task` endpoint because the frontend was sending `channel: 'stable'` but the backend ManagerChannel enum only accepts: `'default'  |  'recent' | 'legacy' | 'forked' | 'dev' | 'tutorial'`.

After, switching versions works again from the version selection popover

Selection_2081

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5312-bugfix-Fix-invalid-ManagerChannel-enum-value-in-nodepack-installation-2636d73d3650819fbe21e290350cf03f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/manager/PackVersionSelectorPopover.vue
+++ b/src/components/dialog/content/manager/PackVersionSelectorPopover.vue
@@ -110,7 +110,7 @@ const SelectedVersionValues = {
 }
 
 const ManagerChannelValues = {
-  STABLE: 'stable' as ManagerChannel,
+  DEFAULT: 'default' as ManagerChannel,
   DEV: 'dev' as ManagerChannel
 }
 
@@ -242,7 +242,7 @@ const handleSubmit = async () => {
   await managerStore.installPack.call({
     id: nodePack.id,
     repository: nodePack.repository ?? '',
-    channel: ManagerChannelValues.STABLE,
+    channel: ManagerChannelValues.DEFAULT,
     mode: ManagerDatabaseSourceValues.CACHE,
     version: actualVersion,
     selected_version: selectedVersion.value
--- a/src/components/dialog/content/manager/PackVersionSelectorPopover.vue
+++ b/src/components/dialog/content/manager/PackVersionSelectorPopover.vue
@@ -109,15 +109,15 @@ const SelectedVersionValues = {
   NIGHTLY: 'nightly' as SelectedVersion
 }
 
-const ManagerChannelValues = {
-  DEFAULT: 'default' as ManagerChannel,
-  DEV: 'dev' as ManagerChannel
+const ManagerChannelValues: Record<string, ManagerChannel> = {
+  DEFAULT: 'default', // ✅ Valid - will compile
+  DEV: 'dev' // ✅ Valid - will compile
 }
 
-const ManagerDatabaseSourceValues = {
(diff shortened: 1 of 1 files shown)


viva-jinyi (dismissed):
LGTM!

src/components/dialog/content/manager/PackVersionSelectorPopover.vue
```diff
@@ -110,7 +110,7 @@ const SelectedVersionValues = {
 }
 
 const ManagerChannelValues = {
-  STABLE: 'stable' as ManagerChannel,
+  DEFAULT: 'default' as ManagerChannel,
   DEV: 'dev' as ManagerChannel
 }
```
christian-byrne: We switched to API-driven design where there is a single source of truth (the openapi spec) which then generates TS types and Pydantic models and both client and server use those types. We should try to actually enforce the string enum type from those types here instead of defining our own enum inside the component (because it can allow for the exact type of error this PR is trying to resolve).

christian-byrne:
## Additional Improvement: Enhanced Type Safety

After investigating the root cause, I discovered this error occurred because we were using **type assertions** (`as ManagerChannel`) which bypass TypeScript's type checking. This allowed invalid enum values like `'stable'` to slip through compilation.

### What I changed in the second commit:

**Before (unsafe):**
```typescript
const ManagerChannelValues = {
  STABLE: 'stable' as ManagerChannel,  // ❌ Type assertion bypasses validation
  DEV: 'dev' as ManagerChannel
}
```

**After (type-safe):**  
```typescript
const ManagerChannelValues: Record<string, ManagerChannel> = {
  DEFAULT: 'default',  // ✅ Compile-time validation
  DEV: 'dev'           // ✅ TypeScript catches invalid values
}
```

### Verification:
- ✅ **Valid values** (`'default'`, `'dev'`): Compile successfully
- ❌ **Invalid values** (`'stable'`): **Fail at compile time** with clear error:
  `Type '"stable"' is not assignable to type '"default"  < /dev/null |  "legacy" | "recent" | "forked" | "dev" | "tutorial"'`

This ensures that future developers can't accidentally introduce invalid enum values - TypeScript will catch them during development rather than causing runtime API validation failures.

This approach properly leverages our API-driven design where the OpenAPI spec is the single source of truth for types.

src/components/dialog/content/manager/PackVersionSelectorPopover.vue
christian-byrne: I changed to

```typescript
const ManagerChannelValues: Record<string, ManagerChannel> = {
  DEFAULT: 'default',  // ✅ Compile-time validation
  DEV: 'dev'           // ✅ TypeScript catches invalid values
}
```

This ensures that future developers can't accidentally introduce invalid enum values - TypeScript will catch them during development rather than causing runtime API validation failures.

christian-byrne:
Oops forgot to remove comments, will do in followup.

PR #5310, opened 2025-09-03
add litegraph test config

jtydhr88:
run all litegraph tests
`pnpm vitest run --config vitest.litegraph.config.ts`

run single test file
`pnpm vitest run --config vitest.litegraph.config.ts src/lib/litegraph/test/CanvasPointer.deviceDetection.test.ts`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5310-add-litegraph-test-config-2636d73d36508152826aca415a71466d) by [Unito](https://www.unito.io)

PR #5314, opened 2025-09-03
[fix] Update Litegraph test snapshots after PR #5310

christian-byrne:
## Summary
- Fixes 3 failing snapshot tests that were broken after the Litegraph test configuration was added in PR #5310
- Updates outdated snapshots to match current LiteGraph behavior
- 608 out of 615 tests now pass (remaining 5 failures are unrelated existing issues)

## Context
After PR #5310 added the Litegraph test configuration, the test suite revealed that 3 snapshot tests had outdated expectations. This PR updates those snapshots to match the current behavior.

## Test Results
- ✅ 3 snapshot tests now pass
- ✅ Total passing tests: 608/615
- ⚠️ 5 remaining failures are pre-existing code issues unrelated to test configuration

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5314-fix-Update-Litegraph-test-snapshots-after-PR-5310-2636d73d365081eb8edff514d11ae758) by [Unito](https://www.unito.io)


src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -369,8 +369,14 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
       widget: promotedWidget,
       subgraphNode: this
     })
-
-    input.widget = { name: subgraphInput.name }
+    const hasNode = (w: any): w is { node: LGraphNode } =>
```
DrJKL: Might have to add an additional shape check, but I think we can avoid an explicit `any` here.

```suggestion
    const hasNode = (w: unknown): w is { node: LGraphNode } =>
```

DrJKL (dismissed):

src/lib/litegraph/src/subgraph/SubgraphNode.ts
AustinMroz: This is likely a missed checkout after reviewing #5024. Here, unknown is insufficient because an unknown `w` doesn't have a `.node` attribute.

christian-byrne: Oops, was testing Austin's PR forgot to checkout back to main before making this PR.

DrJKL: https://stackoverflow.com/a/68927977

DrJKL (approved):

PR #5316, opened 2025-09-03
refactor: v3 ui slots connection dots

LittleSound:
Reuse code, cut down on repeated parts. by the way, I also adjusted the size of the dots in the design draft.

CleanShot 2025-09-02 at 15 40 44

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5316-refactor-v3-ui-slots-connection-dots-2636d73d365081c3823be566fe583c06) by [Unito](https://www.unito.io)

PR #5317, opened 2025-09-03
fix: feature flags and manager state handling

viva-jinyi:
## Summary

Fix ComfyUI Manager feature flag reactivity issues and improve manager state determination logic.

### 1. Replace reactive feature flags with non-reactive approach (72cb4d268)
- Changed `managerUIState` from computed property to `getManagerUIState()` function
- Ensures fresh computation on each call to avoid Vue reactivity timing issues
- Updates all consumers to use the new function-based approach

### 2. Add HelpCenter manager state handling and API version switching (0cc07bd16)
- Fixed HelpCenter manager extension to check manager state
- Fixed 'Manager' command to respect manager state
- Added dynamic API prefix switching based on manager state (`/api/` for legacy, `/api/v2/` for new)

### 3. Simplify manager state determination and fix API timing issues (45de89eab)
- Remove unnecessary extension check from manager state store
- Use only feature flags (client and server) for state determination
- Default to NEW_UI when server flags not loaded (safer default)
- Fix ImportFailInfoBulk to not send empty requests
- Resolves initial 404 errors on installed API calls

### 4. Correct manager state for non-v4 servers (f00eff204)
- Fix servers without v4 support returning DISABLED instead of LEGACY_UI
- Servers without v4 support should use legacy manager, not disable it
- Clarify condition for server v4 + client non-v4 case

## Testing

- [x] Tested with v4 enabled server
- [x] Tested with v4 disabled server
- [x] Tested manager UI state transitions
- [x] Verified no 404 errors on initial load

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Fixes #5316

### DIFF
--- a/src/components/dialog/content/LoadWorkflowWarning.vue
+++ b/src/components/dialog/content/LoadWorkflowWarning.vue
@@ -107,12 +107,12 @@ const uniqueNodes = computed(() => {
 
 // Show manager buttons unless manager is disabled
 const showManagerButtons = computed(() => {
-  return managerState.shouldShowManagerButtons.value
+  return managerState.shouldShowManagerButtons
 })
 
 // Only show Install All button for NEW_UI (new manager with v4 support)
 const showInstallAllButton = computed(() => {
-  return managerState.shouldShowInstallButton.value
+  return managerState.shouldShowInstallButton
 })
 
 const openManager = async () => {
--- a/src/components/dialog/content/LoadWorkflowWarning.vue
+++ b/src/components/dialog/content/LoadWorkflowWarning.vue
@@ -107,12 +107,12 @@ const uniqueNodes = computed(() => {
 
 // Show manager buttons unless manager is disabled
 const showManagerButtons = computed(() => {
-  return managerState.shouldShowManagerButtons
+  return managerState.shouldShowManagerButtons.value
 })
 
 // Only show Install All button for NEW_UI (new manager with v4 support)
 const showInstallAllButton = computed(() => {
-  return managerState.shouldShowInstallButton
+  return managerState.shouldShowInstallButton.value
 })
 
 const openManager = async () => {
--- a/src/components/dialog/content/LoadWorkflowWarning.vue
+++ b/src/components/dialog/content/LoadWorkflowWarning.vue
@@ -115,12 +115,12 @@ const managerStateStore = useManagerStateStore()
 
 // Show manager buttons unless manager is disabled
 const showManagerButtons = computed(() => {
-  return managerStateStore.getManagerUIState() !== ManagerUIState.DISABLED
+  return managerStateStore.shouldShowManagerButtons()
 })
 
 // Only show Install All button for NEW_UI (new manager with v4 support)
 const showInstallAllButton = computed(() => {
-  return managerStateStore.getManagerUIState() === ManagerUIState.NEW_UI
+  return managerStateStore.shouldShowInstallButton()
 })
 
 const openManager = async () => {
--- a/src/components/dialog/content/LoadWorkflowWarning.vue
+++ b/src/components/dialog/content/LoadWorkflowWarning.vue
@@ -58,8 +58,8 @@ import { computed } from 'vue'
 import NoResultsPlaceholder from '@/components/common/NoResultsPlaceholder.vue'
 import MissingCoreNodesMessage from '@/components/dialog/content/MissingCoreNodesMessage.vue'
 import { useMissingNodes } from '@/composables/nodePack/useMissingNodes'
+import { useManagerState } from '@/composables/useManagerState'
 import { useComfyManagerStore } from '@/stores/comfyManagerStore'
-import { useManagerStateStore } from '@/stores/managerStateStore'
 import type { MissingNodeType } from '@/types/comfy'
 import { ManagerTab } from '@/types/comfyManagerTypes'
 
@@ -74,6 +74,7 @@ const { missingNodePacks, isLoading, error, missingCoreNodes } =
   useMissingNodes()
 
 const comfyManagerStore = useComfyManagerStore()
+const managerState = useManagerState()
 
 // Check if any of the missing packs are currently being installed
 const isInstalling = computed(() => {
@@ -104,20 +105,18 @@ const uniqueNodes = computed(() => {
     })
 })
 
-const managerStateStore = useManagerStateStore()
-
 // Show manager buttons unless manager is disabled
 const showManagerButtons = computed(() => {
-  return managerStateStore.shouldShowManagerButtons()
+  return managerState.shouldShowManagerButtons()
 })
 
 // Only show Install All button for NEW_UI (new manager with v4 support)
 const showInstallAllButton = computed(() => {
-  return managerStateStore.shouldShowInstallButton()
+  return managerState.shouldShowInstallButton()
 })
 
 const openManager = async () => {
-  await managerStateStore.openManager({
+  await managerState.openManager({
     initialTab: ManagerTab.Missing,
     showToastOnLegacyError: true
   })
--- a/src/components/dialog/content/LoadWorkflowWarning.vue
+++ b/src/components/dialog/content/LoadWorkflowWarning.vue
@@ -54,19 +54,12 @@
 import Button from 'primevue/button'
 import ListBox from 'primevue/listbox'
 import { computed } from 'vue'
-import { useI18n } from 'vue-i18n'
 
 import NoResultsPlaceholder from '@/components/common/NoResultsPlaceholder.vue'
 import MissingCoreNodesMessage from '@/components/dialog/content/MissingCoreNodesMessage.vue'
 import { useMissingNodes } from '@/composables/nodePack/useMissingNodes'
-import { useDialogService } from '@/services/dialogService'
 import { useComfyManagerStore } from '@/stores/comfyManagerStore'
-import { useCommandStore } from '@/stores/commandStore'
-import {
-  ManagerUIState,
-  useManagerStateStore
-} from '@/stores/managerStateStore'
-import { useToastStore } from '@/stores/toastStore'
+import { useManagerStateStore } from '@/stores/managerStateStore'
 import type { MissingNodeType } from '@/types/comfy'
 import { ManagerTab } from '@/types/comfyManagerTypes'
 
@@ -124,34 +117,10 @@ const showInstallAllButton = computed(() => {
 })
 
 const openManager = async () => {
-  const state = managerStateStore.getManagerUIState()
-
-  switch (state) {
-    case ManagerUIState.DISABLED:
-      useDialogService().showSettingsDialog('extension')
-      break
-
-    case ManagerUIState.LEGACY_UI:
-      try {
-        await useCommandStore().execute('Comfy.Manager.Menu.ToggleVisibility')
-      } catch {
-        // If legacy command doesn't exist, show toast
-        const { t } = useI18n()
-        useToastStore().add({
-          severity: 'error',
-          summary: t('g.error'),
-          detail: t('manager.legacyMenuNotAvailable'),
-          life: 3000
-        })
-      }
-      break
-
-    case ManagerUIState.NEW_UI:
-      useDialogService().showManagerDialog({
-        initialTab: ManagerTab.Missing
-      })
-      break
-  }
+  await managerStateStore.openManager({
+    initialTab: ManagerTab.Missing,
+    showToastOnLegacyError: true
+  })
 }
 </script>
 
--- a/src/components/dialog/content/LoadWorkflowWarning.vue
+++ b/src/components/dialog/content/LoadWorkflowWarning.vue
@@ -107,12 +107,12 @@ const uniqueNodes = computed(() => {
 
 // Show manager buttons unless manager is disabled
 const showManagerButtons = computed(() => {
-  return managerState.shouldShowManagerButtons()
+  return managerState.shouldShowManagerButtons.value
 })
 
 // Only show Install All button for NEW_UI (new manager with v4 support)
 const showInstallAllButton = computed(() => {
-  return managerState.shouldShowInstallButton()
+  return managerState.shouldShowInstallButton.value
 })
 
 const openManager = async () => {
--- a/src/components/dialog/content/LoadWorkflowWarning.vue
+++ b/src/components/dialog/content/LoadWorkflowWarning.vue
@@ -115,16 +115,16 @@ const managerStateStore = useManagerStateStore()
 
 // Show manager buttons unless manager is disabled
 const showManagerButtons = computed(() => {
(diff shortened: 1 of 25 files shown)


src/services/comfyManagerService.ts
```diff
@@ -34,13 +38,28 @@ enum ManagerRoute {
...
     'Content-Type': 'application/json'
   }
 })
 
+// Add request interceptor to dynamically set baseURL
+managerApiClient.interceptors.request.use((config) => {
+  config.baseURL = getApiBaseURL()
+  return config
+})
+
```
christian-byrne: If the manager state is in LEGACY or DISABLED, the comfyManagerService should never be getting called, so this shouldn't be necessary.

src/components/helpcenter/HelpCenterMenuContent.vue
```diff
@@ -313,8 +317,29 @@ const menuItems = computed<MenuItem[]>(() => {
...
+              )
+            } catch {
+              // If legacy command doesn't exist, fall back to extensions panel
+              dialogService.showSettingsDialog('extension')
+            }
+            break
+
+          case ManagerUIState.NEW_UI:
+            dialogService.showManagerDialog()
+            break
```
christian-byrne: Shuold we move this logic into the `managerStateStore.ts`? I think we repeat it 4 or so times (my fault).

src/stores/managerStateStore.ts
```diff
@@ -13,36 +11,29 @@ export enum ManagerUIState {
...
-    console.log('[Manager State Debug]', {
-      systemStats: systemStats?.system?.argv,
-      clientSupportsV4,
-      serverSupportsV4,
-      hasLegacyManager,
-      extensions: extensionStore.extensions.map((e) => e.name)
-    })
-
-    // Check command line args first
+    // Check command line args first (highest priority)
```
christian-byrne: I realized we should also do a check to ensure `systemStats` is initialized here when doing this. It's possible this will be the first access in the app lifecycle. See:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/0afbf94eb764f3a857690ceb13be4ed82ee6baf1/src/stores/systemStatsStore.ts#L9-L28

src/stores/managerStateStore.ts
```diff
@@ -51,26 +42,26 @@ export const useManagerStateStore = defineStore('managerState', () => {
...
       return ManagerUIState.LEGACY_UI
     }
 
-    // If server feature flags haven't loaded yet, return DISABLED for now
-    // This will update reactively once feature flags load
+    // If server feature flags haven't loaded yet, default to NEW_UI
+    // This is the safest default since v2 API is the current standard
     if (serverSupportsV4 === undefined) {
-      return ManagerUIState.DISABLED
+      return ManagerUIState.NEW_UI
```
christian-byrne: In theory, the feature flags should always be available before other parts of the app are initialized, so this shouldn't be necessary. Can you verify again that it's needed? If so, it might jsut be an issue with way we are accessing.

christian-byrne (commented):
Nice, thank you.

Some added clarification:

During Beta (previous)

* There was a single manager python server
* We needed to support using the old manager _and_ new manager in a single frontend client session
* To do this, we created a new manager server API version

After Release (now)

* There are two manager python server packages. If the user is using `legacy`, the legacy server is initialized. If not, the new manager server is initialized.
* On the frontend
    * If legacy -> the legacy manager JS code is initialized. ComfyUI_frontend can simply disable all manager features
    * if disabled -> no manager JS code is initialized, no manager python server is initialized, ComfyUI_frontend can simply disable all of its manager features
    * else -> the legacy manager JS code is not initialized, ComfyUI_frontend only needs to support the new manager API version, doesn't need to worry about using interceptor or dynamically setting Axios client base URL

viva-jinyi:
@christian-byrne 
### 1) Centralized “Open Manager” logic

- Consolidated the repeated switch statements for Manager state (previously in 8+ files) into a single managerStateStore.openManager() function.
- All Manager entry points now call the centralized function:
    - useCoreCommands.ts: 6 commands
    - CommandMenubar.vue: Manage Extensions menu
    - HelpCenterMenuContent.vue: Manager button
    - LoadWorkflowWarning.vue: Open Manager button

### 2) New openManager() options

```
openManager(options?: {
  initialTab?: ManagerTab;           // Open a specific tab in NEW_UI
  legacyCommand?: string;            // Custom legacy command to run
  showToastOnLegacyError?: boolean;  // Show toast on legacy command failure
  isLegacyOnly?: boolean;            // Mark as legacy-only (error in NEW_UI)
});
```

### 3) Bug fixes

- In LEGACY_UI mode, when a legacy command fails, show an error toast instead of opening the new Manager.
- In NEW_UI mode, fixed an issue where legacy command errors were displayed as a numeric value.

### 🔍 Files changed

- `src/stores/managerStateStore.ts` — added openManager()
- `src/composables/useCoreCommands.ts `— refactored 6 Manager commands
- `src/components/topbar/CommandMenubar.vue` — simplified showManageExtensions()
- `src/components/helpcenter/HelpCenterMenuContent.vue` — simplified Manager action
- `src/components/dialog/content/LoadWorkflowWarning.vue` — simplified openManager() usage

### ✅ Verification

- TypeScript compile ✅
- ESLint pass ✅
- Unit tests ✅
- Manual checks for NEW_UI / LEGACY_UI / DISABLED modes ✅

DrJKL:
I think if we're going to have non-reactive utilities like this, they should probably be separated from the stores. Maybe something in `services/`?

viva-jinyi:
### Refactoring: managerStateStore → useManagerState composable

Thank you for the feedback! @DrJKL is absolutely right about this.

#### Issue
- `managerStateStore` only provides non-reactive utility functions
- Doesn't align with Pinia store's purpose (reactive state management)
- Violates Pinia guideline: "If it's not reactive, it probably shouldn't be in a store"

#### Solution
Converted the Store to a Composable:

**Changes:**
- ✅ Moved `src/stores/managerStateStore.ts` → `src/composables/useManagerState.ts`
- ✅ Updated all import paths (8 files)
- ✅ Relocated and updated test files
- ✅ TypeScript type checking passes
- ✅ Unit tests passing

**Benefits:**
1. **Pattern consistency**: Other utility functions in the repo use composables
2. **Clear separation of concerns**: Stores for reactive state, Composables for business logic
3. **Follows Pinia guidelines**: Non-reactive code excluded from stores

Since Manager UI state doesn't change during runtime and is just a utility checking feature flags and command line args, a composable is the more appropriate pattern.

src/components/graph/GraphCanvas.vue
```diff
@@ -696,6 +697,11 @@ onMounted(async () => {
   const releaseStore = useReleaseStore()
   void releaseStore.initialize()
 
+  // Initialize conflict detection after GraphCanvas is ready
+  // This ensures SystemStats and Manager state are properly loaded
```
christian-byrne: Could also add: "ensures `ComfyApp` is initialized which implies `ComfyApi.init` was called and feature flags were negotiated"

src/components/graph/GraphCanvas.vue
```diff
@@ -696,6 +697,11 @@ onMounted(async () => {
...
+  const conflictDetection = useConflictDetection()
+  void conflictDetection.initializeConflictDetection()
```
christian-byrne: nit: can change to `void useConflictDetection().initializeConflictDetection()`

christian-byrne: Seems good to also check `if (managerState.isManagerEnabled.value)` here, to avoid side-effects of conflict detection store's dependencies like invoking `useManagerStore`

src/composables/useConflictDetection.ts
```diff
@@ -636,6 +637,15 @@ export function useConflictDetection() {
    */
   async function initializeConflictDetection(): Promise<void> {
     try {
+      // Check if manager is disabled before running conflict detection
+      const managerState = useManagerState()
+      if (!managerState.isManagerEnabled.value) {
+        console.log(
+          '[ConflictDetection] Manager is disabled, skipping conflict detection'
+        )
```
christian-byrne: Want to remove this or move it to `console.debug`?

src/composables/useManagerState.ts
```diff
@@ -0,0 +1,192 @@
...
+
+  /**
+   * The current manager UI state.
+   * Computed once and cached until dependencies change (which they don't during runtime).
+   * This follows Vue's conventions and provides better performance through caching.
+   */
+  const managerUIState = readonly(
+    computed((): ManagerUIState => {
+      // Get current values
+      const systemStats = systemStatsStore.systemStats
```
christian-byrne: We should check if systemStats is `null` and if so, await the `fetch` method from that store.

christian-byrne (commented):
This PR is really clean thanks for organizing this.

src/services/comfyManagerService.ts
```diff
@@ -34,6 +35,7 @@ enum ManagerRoute {
   QUEUE_TASK = 'manager/queue/task'
 }
 
+// Always use v2 API for New Manager (no dynamic switching needed)
```
christian-byrne: We can probably remove this right?

viva-jinyi:
@christian-byrne 
### CI/CD Test Failures Detected
In this PR, four Playwright tests failed.
Root cause: In GraphCanvas.vue, calling useConflictDetection() inside setup() introduced side effects that mutated the store.

### Core Issue in GraphCanvas.vue
**Problem:** GraphCanvas.vue directly invokes useConflictDetection().
**Side effects:**
Initializes multiple stores as soon as the composable is created.
Breaks the reactivity chain in the test environment.
Interferes with DOM widget cleanup logic.
Delays UI rendering.

### Initial Mitigation Attempt: Manager Disabled Check
Added conditional in GraphCanvas.vue:
`if (managerState.isManagerEnabled.value) { ... }`
Still problematic: invoking the composable itself triggers store initialization regardless of the condition.

### Attempt to Move Logic into App.vue
Tried relocating conflict detection into App.vue, same as before.
New issue: systemStats wasn’t yet loaded at that point, so managerState could not be correctly resolved.

### Root Cause Identified: systemStatsStore Timing
**Key issue:** systemStatsStore is not handled synchronously.
At store creation time, systemStats is still null.
useManagerState cannot reliably detect --disable-manager from argv.
As a result, the disabled manager state isn’t recognized correctly.

### Implemented Solution
A. Remove Conflict Detection from GraphCanvas.vue

```
// Removed problematic code
const managerState = useManagerState()
if (managerState.isManagerEnabled.value)
  void useConflictDetection().initializeConflictDetection()
```

B. Refactor systemStatsStore with useAsyncState Pattern

```
const { state, isReady, execute } = useAsyncState(
  fetchSystemStatsData,
  null,
  { immediate: true }
)
```

Automatically manages async state.
Provides an isInitialized flag to check readiness.

C. Add Initialization Guard in useManagerState

```
if (!systemStatsStore.isInitialized) {
  return ManagerUIState.DISABLED
}
```

Treats manager as disabled until systemStats has finished loading.

D. Manager Check Inside useConflictDetection

```
if (!managerState.isManagerEnabled.value) {
  console.debug('[ConflictDetection] Manager is disabled')
  return
}
```

### Final Outcome
Removed all side effects from GraphCanvas.vue.
Fixed timing issues in systemStats initialization.
Ensured no conflict detection API calls when the manager is disabled.
Restored stability of the reactive chain in test environments.

src/components/dialog/content/LoadWorkflowWarning.vue
```diff
@@ -111,47 +105,21 @@ const uniqueNodes = computed(() => {
     })
 })
 
-const managerStateStore = useManagerStateStore()
-
 // Show manager buttons unless manager is disabled
 const showManagerButtons = computed(() => {
-  return managerStateStore.managerUIState !== ManagerUIState.DISABLED
+  return managerState.shouldShowManagerButtons.value
```
christian-byrne, line 110: nit: don't need `.value`

src/stores/systemStatsStore.ts
```diff
@@ -62,6 +64,7 @@ export const useSystemStatsStore = defineStore('systemStats', () => {
     systemStats,
     isLoading,
     error,
+    isInitialized,
     fetchSystemStats,
```
christian-byrne: we should remove `fetchSystemStats` (keep private) and only expose `execute` (can alias the name), since `execute` will handle deduplication and setting the `isLoading` state.

src/composables/useConflictDetection.ts
```diff
@@ -78,7 +78,8 @@ export function useConflictDetection() {
     try {
       // Get system stats from store (primary source of system information)
       const systemStatsStore = useSystemStatsStore()
-      if (!systemStatsStore.systemStats) {
+      // Wait for systemStats to be initialized if not already
+      if (!systemStatsStore.isInitialized) {
```
christian-byrne: If needing to wait, we can use

```typescript
import { until } from '@vueuse/core'

await until(systemStatsStore.isInitialized)
```

instead of using the private wrapped execute function

src/composables/useConflictDetection.ts
```diff
@@ -636,7 +640,18 @@ export function useConflictDetection() {
    */
   async function initializeConflictDetection(): Promise<void> {
     try {
-      // Simply perform conflict detection
+      // Check if manager is enabled before proceeding
+      const { useManagerState } = await import('@/composables/useManagerState')
```
christian-byrne, line 643: nit: I think it's okay to import this eagerly in the module scope, the module is always going to be initialized for all users, right?

christian-byrne (changes requested):

src/stores/releaseStore.ts
```diff
@@ -240,7 +240,7 @@ export const useReleaseStore = defineStore('release', () => {
     try {
       // Ensure system stats are loaded
       if (!systemStatsStore.systemStats) {
-        await systemStatsStore.fetchSystemStats()
+        await systemStatsStore.refetchSystemStats()
```
christian-byrne: nit: this can also be `await until(systemStatsStore.isInialized)`

src/stores/versionCompatibilityStore.ts
```diff
@@ -103,7 +103,7 @@ export const useVersionCompatibilityStore = defineStore(
 
     async function checkVersionCompatibility() {
       if (!systemStatsStore.systemStats) {
-        await systemStatsStore.fetchSystemStats()
+        await systemStatsStore.refetchSystemStats()
```
christian-byrne: nit: this can also be `await until(systemStatsStore.isInialized)`

christian-byrne (dismissed):
LGTM with 2 nit

christian-byrne (approved):
LGTM!

PR #5319, opened 2025-09-03
fix: the version of tailwind-merge does not match

LittleSound:
because we're using Tailwind v3, so we need to switch to tailwind-merge v2.6.0

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5319-fix-the-version-of-tailwind-merge-does-not-match-2636d73d36508182a731cad037e0d3b0) by [Unito](https://www.unito.io)

### DIFF
--- a/package.json
+++ b/package.json
@@ -147,7 +147,7 @@
     "primeicons": "^7.0.0",
     "primevue": "^4.2.5",
     "semver": "^7.7.2",
-    "tailwind-merge": "^3.3.1",
+    "tailwind-merge": "^2.6.0",
     "three": "^0.170.0",
     "tiptap-markdown": "^0.8.10",
     "vue": "^3.5.13",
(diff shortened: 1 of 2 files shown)


christian-byrne (commented):
In https://github.com/Comfy-Org/ComfyUI_frontend/pull/5246 (on `main`), we upgraded to Tailwind v4. If we rebase onto `main`, will this change still be necessary?

christian-byrne:
For easier reference, designs are [here](https://www.figma.com/design/31uH3r4x3xbIctuRWYW6NM/V3---Nodes?node-id=5910-2737&p=f&m=dev)

LittleSound:
> In #5246 (on `main`), we upgraded to Tailwind v4. If we rebase onto `main`, will this change still be necessary?

after updating to tailwind 4, this pull request is no longer needed

PR #5320, opened 2025-09-03
feat: widget styles for V3 UI

LittleSound:
Align with the style in the design draft in Figma

CleanShot 2025-09-03 at 18 43 48@2x
CleanShot 2025-09-03 at 18 44 13@2x

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5320-feat-widget-styles-for-V3-UI-2636d73d3650812a925ed19f84cf4ec1) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
@@ -2,10 +2,11 @@
   <WidgetLayoutField :widget="widget">
     <MultiSelect
       v-model="localValue"
-      v-bind="filteredProps"
+      :options="widget.options?.values || []"
       :disabled="readonly"
-      class="w-full text-xs"
+      :class="cn(WidgetInputBaseClass, 'w-full text-xs')"
       size="small"
+      display="chip"
       :pt="{
         option: 'text-xs'
       }"
@@ -16,15 +17,12 @@
 
 <script setup lang="ts">
 import MultiSelect from 'primevue/multiselect'
-import { computed } from 'vue'
 
 import { useWidgetValue } from '@/composables/graph/useWidgetValue'
 import type { SimplifiedWidget } from '@/types/simplifiedWidget'
-import {
-  PANEL_EXCLUDED_PROPS,
-  filterWidgetProps
-} from '@/utils/widgetPropFilter'
+import { cn } from '@/utils/tailwindUtil'
 
+import { WidgetInputBaseClass } from './layout'
 import WidgetLayoutField from './layout/WidgetLayoutField.vue'
 
 const props = defineProps<{
@@ -44,14 +42,4 @@ const { localValue, onChange } = useWidgetValue({
   defaultValue: [],
   emit
 })
-
-// MultiSelect specific excluded props include overlay styles
-const MULTISELECT_EXCLUDED_PROPS = [
-  ...PANEL_EXCLUDED_PROPS,
-  'overlayStyle'
-] as const
-
-const filteredProps = computed(() =>
-  filterWidgetProps(props.widget.options, MULTISELECT_EXCLUDED_PROPS)
-)
 </script>
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
@@ -2,7 +2,7 @@
   <WidgetLayoutField :widget="widget">
     <MultiSelect
       v-model="localValue"
-      :options="widget.options?.values || []"
+      v-bind="filteredProps"
       :disabled="readonly"
       :class="cn(WidgetInputBaseClass, 'w-full text-xs')"
       size="small"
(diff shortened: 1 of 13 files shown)


benceruleanlu (dismissed):
very cool!

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectButton.vue
```diff
@@ -1,62 +1,59 @@
...
+  // Ensure options array is available for SelectButton
+  if (filtered.values && Array.isArray(filtered.values)) {
+    filtered.options = filtered.values
+  }
+
+  return filtered as {
+    options: any[]
+    [key: string]: any
+  }
+})
```
christian-byrne: ~~Was `options` getting filtered? Maybe it's easier if we adjust the filtering logic.~~

src/renderer/extensions/vueNodes/widgets/composables/useWidgetRenderer.ts
```diff
@@ -23,6 +23,12 @@ const TYPE_TO_ENUM_MAP: Record<string, string> = {
   // Selection
   combo: WidgetType.COMBO,
   COMBO: WidgetType.COMBO,
+  selectbutton: WidgetType.SELECTBUTTON,
+  SELECTBUTTON: WidgetType.SELECTBUTTON,
+  multiselect: WidgetType.MULTISELECT,
+  MULTISELECT: WidgetType.MULTISELECT,
+  treeselect: WidgetType.TREESELECT,
+  TREESELECT: WidgetType.TREESELECT,
```
christian-byrne, line 31: Just noting that this is the existing pattern already (my fault), not something introduced in this PR. I can simplify later on so that new widgets don't need uppercase aliases.

src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
```diff
@@ -44,14 +42,4 @@ const { localValue, onChange } = useWidgetValue({
...
-
-// MultiSelect specific excluded props include overlay styles
-const MULTISELECT_EXCLUDED_PROPS = [
-  ...PANEL_EXCLUDED_PROPS,
-  'overlayStyle'
-] as const
-
-const filteredProps = computed(() =>
-  filterWidgetProps(props.widget.options, MULTISELECT_EXCLUDED_PROPS)
-)
```
christian-byrne: Is it possible to keep the prop filtering for this? Was it causing issues? Maybe we can solve it by ensuring `options` is not being filtered (similar to the other comment).

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -3,7 +3,8 @@
     v-model="localValue"
     v-bind="filteredProps"
     :disabled="readonly"
-    class="w-full text-xs"
+    :class="cn(WidgetInputBaseClass, 'w-full text-xs')"
+    :placeholder="widget.name || ''"
```
christian-byrne: I think the consumer will set the `placeholder` prop in a lot of casses. It will be exposed in the API for people making nodes.

For fallback values, what do you think about something like:

```vue
:placeholder="placeholder ?? widget.name || ''"
```

or 

```typescript
const placeholder = computed(() => props.placeholder ?? widget.name || '')
```

christian-byrne:
For easier reference, designs are [here](https://www.figma.com/design/31uH3r4x3xbIctuRWYW6NM/V3---Nodes?node-id=5910-2737&p=f&m=dev)

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectButton.vue
christian-byrne: Looking at PrimeVue SelectButton, it expects:

- options prop for the data array
- optionLabel and optionValue props to specify which fields to use

What about changing FormSelectButton.vue to:

```vue
<script setup lang="ts">
    interface Props {
        modelValue: string | null | undefined
        options: any[]           // Keep this name to match PrimeVue
        optionLabel?: string     // Add these PrimeVue props
        optionValue?: string
        disabled?: boolean
    }
</script>
```

Then in WidgetSelectButton.vue, change:

```vue
<template>
    <FormSelectButton
      v-model="localValue" 
      :options="widget.options?.values || []"  <!-- Direct binding, no transform -->
      :disabled="readonly"
      @update:model-value="onChange"
    />
</template>

<script>
// Can safely remove the values -> options transformation because FormSelectButton now expects the same API as PrimeVue components.
</script>
```

christian-byrne:
Also appreciated:

- Good use of `WidgetInputBaseClass` and `cn()` utility for consistent styling
- Narrowing from `any` to `string` for the value in the SelectButton

Let's continue these patterns!

src/renderer/extensions/vueNodes/widgets/components/form/FormSelectButton.vue
christian-byrne, line 1: Since FormSelectButton is a custom component that needs to handle both primitive arrays (`['a', 'b']`) and object arrays (`[{label: 'A', value: 'a'}]`), what about using generics:
 
```vue
<script setup lang="ts" generic="T extends string | number | { label: string; value: any }">
interface Props {
  options: T[]  // Instead of any[]
}
</script>
```

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectButton.vue
LittleSound: thanks. let me try this way.

LittleSound:
Thanks. I tried to fix these issues

christian-byrne (approved):
LGTM!

PR #5321, opened 2025-09-03
feat: Auto-close LoadWorkflowWarning dialog when all missing nodes are installed

viva-jinyi:
## Summary

Automatically close the LoadWorkflowWarning dialog when all missing nodes have been successfully installed through the Manager.

### Implementation
- Added a computed property `allMissingNodesInstalled` that checks:
  - Loading is complete (`!isLoading`)
  - No missing packs remain (`missingNodePacks.length === 0`) 
  - Not currently installing (`!isInstalling`)
- Added a watcher that triggers when all conditions are met
- Dialog closes automatically after 500ms delay to let users see completion
- Shows success toast notification upon completion

### User Experience Improvements
- Users no longer need to manually close the dialog after successful installation
- Visual feedback through success toast confirms all nodes were installed
- 500ms delay ensures users can see the completion state before dismissal

## Testing

- [x] Tested with workflows containing missing nodes
- [x] Verified dialog closes after successful installation
- [x] Confirmed success toast appears
- [x] Ensured dialog doesn't close prematurely during installation

## Type of Change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Bug fix (non-breaking change which fixes an issue)  
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5321-feat-Auto-close-LoadWorkflowWarning-dialog-when-all-missing-nodes-are-installed-2636d73d3650810ab431e66c12523537) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/LoadWorkflowWarning.vue
+++ b/src/components/dialog/content/LoadWorkflowWarning.vue
@@ -61,12 +61,12 @@ import MissingCoreNodesMessage from '@/components/dialog/content/MissingCoreNode
 import { useMissingNodes } from '@/composables/nodePack/useMissingNodes'
 import { useManagerState } from '@/composables/useManagerState'
 import { useComfyManagerStore } from '@/stores/comfyManagerStore'
+import { useDialogStore } from '@/stores/dialogStore'
+import { useToastStore } from '@/stores/toastStore'
 import type { MissingNodeType } from '@/types/comfy'
 import { ManagerTab } from '@/types/comfyManagerTypes'
 
 import PackInstallButton from './manager/button/PackInstallButton.vue'
-import { useDialogStore } from '@/stores/dialogStore'
-import { useToastStore } from '@/stores/toastStore'
 
 const props = defineProps<{
   missingNodeTypes: MissingNodeType[]
--- a/src/components/dialog/content/LoadWorkflowWarning.vue
+++ b/src/components/dialog/content/LoadWorkflowWarning.vue
@@ -53,7 +53,7 @@
 <script setup lang="ts">
 import Button from 'primevue/button'
 import ListBox from 'primevue/listbox'
-import { computed, watch } from 'vue'
+import { computed, nextTick, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 import NoResultsPlaceholder from '@/components/common/NoResultsPlaceholder.vue'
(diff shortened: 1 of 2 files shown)


src/components/dialog/content/LoadWorkflowWarning.vue
```diff
@@ -153,6 +154,36 @@ const openManager = async () => {
...
+      dialogStore.closeDialog({ key: 'global-load-workflow-warning' })
+
+      // Show success toast
+      useToastStore().add({
+        severity: 'success',
+        summary: t('g.success'),
+        detail: t('manager.allMissingNodesInstalled'),
+        life: 3000
+      })
+    }, 500)
```
christian-byrne: I feel like we don't need the timeout if we are showing a toast anyway, WDYT?

If we are using a timeout, do you think we need to:

1. clear it in `unMounted` or, alternatively, use https://vueuse.org/shared/useTimeout/
2. add deduplication to prevent multiple queued calls

christian-byrne: If the issue is a race condition with `missingNodePacks` changes -> missingNodePacks may not update fully until next tick, then we could just add `nextTick()` for now.

src/components/dialog/content/LoadWorkflowWarning.vue
```diff
@@ -153,6 +154,36 @@ const openManager = async () => {
...
+  return (
+    !isLoading.value &&
+    missingNodePacks.value?.length === 0 &&
+    !isInstalling.value
+  )
+})
+
+// Watch for completion and close dialog
+watch(allMissingNodesInstalled, (allInstalled) => {
+  if (allInstalled && missingNodePacks.value !== null) {
```
DrJKL: Isn't `missingNodePacks.value !== null` implied by allMissingNodesInstalled?

DrJKL (dismissed):
This looks safe enough. Could you add a video showing the behavior?

src/components/dialog/content/LoadWorkflowWarning.vue
viva-jinyi: All missing nodes have been installed" means:
- Loading has finished + No installation in progress + Zero missing packs remaining = true
- If any condition is missing, the dialog may close at the wrong timing.

src/components/dialog/content/LoadWorkflowWarning.vue
viva-jinyi: I will use nextTick to ensure state updates are complete.

viva-jinyi:
[screen-capture (1).webm](https://github.com/user-attachments/assets/14434618-f782-43c2-815a-f1165ded360c)

viva-jinyi:
[screen-capture.webm](https://github.com/user-attachments/assets/f600e713-ea68-4126-98f6-a9b361a119eb)

christian-byrne (approved):
LGTM!

PR #5323, opened 2025-09-03
Add compact menu style for smaller window heights 

### LINKED ISSUE #5289: New menu inaccessible at smaller window sizes
### Prerequisites

### What happened?

Menu inaccessible; placement prevents use of lower buttons.

Image

### Steps to Reproduce

1. Use a small window size (not fullscreen)
1. Open new menu
2. Open help submenu
3. Observe clipped menu

### ComfyUI Frontend Version

HEAD

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-5289-New-menu-inaccessible-at-smaller-window-sizes-2616d73d365081879574f30fa4e87782) by [Unito](https://www.unito.io)

Myestery:
This pull request enhances the responsiveness of the `CommandMenubar` component by introducing a "compact" mode for screens with limited vertical space. When the window height is less than 700px, the menubar adapts its submenu positioning and styling to improve usability on smaller displays. The most important changes are grouped below:

**Responsive Behavior and State Management:**

* Added a reactive `windowHeight` reference and a computed `isCompactHeight` property to detect when the window height is below 700px, enabling "compact" mode for the menubar. Event listeners for window resize are now managed using `onMounted` and `onUnmounted` lifecycle hooks to keep the state updated. [[1]](diffhunk://#diff-10ee0e60e600a168bdd45e0b9f05a148f5b9ee48f54e6c7dee737ebe7b6192e6R314-R326) [[2]](diffhunk://#diff-10ee0e60e600a168bdd45e0b9f05a148f5b9ee48f54e6c7dee737ebe7b6192e6L103-R104)

**Dynamic Styling and Class Application:**

* Updated the root element's class bindings to apply the `comfy-command-menu-compact` class when `isCompactHeight` is true, enabling conditional styling for compact mode.

**Compact Mode Styling:**

* Introduced new CSS rules for `.comfy-command-menu-compact` to force submenus to open to the right and align with the top of the menu container, ensuring better usability on short screens. Adjustments include submenu positioning, alignment for nested menus, and ensuring the submenu container uses the full available height.
* 
/fix #5289

https://github.com/user-attachments/assets/e7908b57-3f07-4fc4-a119-66f2b7e398c5

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5323-Add-compact-menu-style-for-smaller-window-heights-2636d73d3650816cb5aaf4ac76c39739) by [Unito](https://www.unito.io)


src/components/topbar/CommandMenubar.vue
```diff
@@ -310,6 +311,19 @@ const handleItemClick = (item: MenuItem, event: MouseEvent) => {
 const hasActiveStateSiblings = (item: MenuItem): boolean => {
   return menuItemsStore.menuItemHasActiveStateChildren[item.parentPath]
 }
+const windowHeight = ref(window.innerHeight)
```
DrJKL: Can we use this instead? https://vueuse.org/core/useWindowSize/#usewindowsize

src/components/topbar/CommandMenubar.vue
```diff
@@ -338,4 +352,32 @@ const hasActiveStateSiblings = (item: MenuItem): boolean => {
...
   top: -4px !important;
 }
+
+/* Force submenus to open to the right and start from top on compact heights */
+.comfy-command-menu-compact .p-tieredmenu-submenu {
+  /* Position submenu to the right of parent menu */
+  position: absolute !important;
+  left: 100% !important;
+  margin-left: 4px !important;
+  margin-top: 0 !important;
```
DrJKL: I would really like to avoid any additional `!important`s.

src/components/topbar/CommandMenubar.vue
Myestery: Thanks!

src/components/topbar/CommandMenubar.vue
Myestery: done!

DrJKL:
I don't like that the submenu doesn't align at all with the menu trigger here
image

The demo menus [here](https://primevue.org/tieredmenu/) adapt if they're too close to the edge. Are we doing something special that's overriding that behavior?

christian-byrne (dismissed):
LGTM.

I think the approach more aligned with the codebase would be to `useBreakpoints` -- and we can expect that the breakpoints will generally give us a correct idea about the height. Alternatively, we can just handle the edge case by making the menu scrollable and `fill-available` css properties. However, I think this solution is also fine.

christian-byrne:
> I don't like that the submenu doesn't align at all with the menu trigger here

I agree but I think this issue was existing before this PR right?

DrJKL:
> > I don't like that the submenu doesn't align at all with the menu trigger here
> 
> I agree but I think this issue was existing before this PR right?

An earlier version of this PR had a video that had the help menu floating at the very top.

DrJKL:
In my own testing of this, the issue is mitigated by removing the `!important` left and top values in the unscoped style block.
I'd rather find our local overrides that are causing PrimeVue's component to behave differently than in their spec/demo and remove those.
Or we could use a `@media` query to apply the changes via CSS only, if we absolutely have to.

christian-byrne (approved):

christian-byrne:
Tested and it solves the problem.

PR #5279, opened 2025-08-31
[fix] Enable mouse gestures over video previews

christian-byrne:
## Summary
- Fixes ctrl+scroll zoom when cursor is over video/image previews  
- Fixes space+drag panning when cursor is over video/image previews
- Preserves existing drag behavior for node manipulation

## Technical Details
- Extended `useCanvasInteractions` composable with `handlePointer` method
- Added event listeners to video elements in `useNodeVideo` 
- Uses existing canvas space key detection (`read_only` property)
- Follows established canvas store patterns

https://github.com/user-attachments/assets/0c9427ac-78d5-4b3c-ab81-e98981640327

Related:

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/4574
- https://github.com/Comfy-Org/ComfyUI_frontend/issues/2986

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5279-fix-Enable-mouse-gestures-over-video-previews-2606d73d3650812fb378d4c72eedae8b) by [Unito](https://www.unito.io)

PR #5324, opened 2025-09-03
[feat] register UNETLoader, UpscaleModelLoader, StylemModelLoader...

arjansingh:
GLIGENLoader

## Summary

This is in preparation for getting them to work with our new ModelBrowser component.

Also wrote test suite for the store.

## Changes

Nothing should be noticeable to the user.

## Review Focus

1. Make sure the test suite makes sense please.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5324-feat-register-UNETLoader-UpscaleModelLoader-StylemModelLoader-2636d73d3650816bb19acc2fa405504b) by [Unito](https://www.unito.io)

### DIFF
--- a/tests-ui/tests/store/modelToNodeStore.test.ts
+++ b/tests-ui/tests/store/modelToNodeStore.test.ts
@@ -1,0 +1,245 @@
+import { createPinia, setActivePinia } from 'pinia'
+import { beforeEach, describe, expect, it } from 'vitest'
+
+import {
+  ModelNodeProvider,
+  useModelToNodeStore
+} from '@/stores/modelToNodeStore'
+import { type ComfyNodeDefImpl, useNodeDefStore } from '@/stores/nodeDefStore'
+
+describe('useModelToNodeStore', () => {
+  let store: ReturnType<typeof useModelToNodeStore>
+  let nodeDefStore: ReturnType<typeof useNodeDefStore>
+
+  // Create minimal mock for testing - only includes 'name' field since that's
+  // the only property ModelNodeProvider constructor uses and tests verify
+  const createMockNodeDef = (name: string): ComfyNodeDefImpl => {
+    return { name } as ComfyNodeDefImpl
+  }
+
+  beforeEach(() => {
+    setActivePinia(createPinia())
+    store = useModelToNodeStore()
+    nodeDefStore = useNodeDefStore()
+
+    const mockNodeNames = [
+      'CheckpointLoaderSimple',
+      'ImageOnlyCheckpointLoader',
+      'LoraLoader',
+      'LoraLoaderModelOnly',
+      'VAELoader',
+      'ControlNetLoader',
+      'UNETLoader',
+      'UpscaleModelLoader',
+      'StyleModelLoader',
+      'GLIGENLoader'
+    ]
+
+    const mockNodeDefs: Record<string, ComfyNodeDefImpl> = Object.fromEntries(
+      mockNodeNames.map((name) => [name, createMockNodeDef(name)])
+    )
+
+    nodeDefStore.nodeDefsByName = mockNodeDefs
+  })
+
+  describe('modelToNodeMap', () => {
+    it('should initialize as empty', () => {
+      expect(Object.keys(store.modelToNodeMap)).toHaveLength(0)
+    })
+
+    it('should populate after registration', () => {
+      store.registerDefaults()
+      expect(Object.keys(store.modelToNodeMap)).toContain('checkpoints')
+      expect(Object.keys(store.modelToNodeMap)).toContain('unet')
+    })
+  })
+
+  describe('getNodeProvider', () => {
+    it('should return provider for registered model type', () => {
+      store.registerDefaults()
+
+      const provider = store.getNodeProvider('checkpoints')
+      expect(provider).toBeDefined()
+      // Optional chaining used because getNodeProvider() can return undefined for unregistered types
+      expect(provider?.nodeDef.name).toBe('CheckpointLoaderSimple')
+      expect(provider?.key).toBe('ckpt_name')
+    })
+
+    it('should return undefined for unregistered model type', () => {
+      store.registerDefaults()
+      expect(store.getNodeProvider('nonexistent')).toBeUndefined()
+    })
+
+    it('should return first registered provider when multiple providers exist for same model type', () => {
+      store.registerDefaults()
+
+      const provider = store.getNodeProvider('checkpoints')
+      // Using optional chaining for safety since getNodeProvider() can return undefined
+      expect(provider?.nodeDef.name).toBe('CheckpointLoaderSimple')
+    })
+
+    it('should trigger lazy registration when called before registerDefaults', () => {
+      const provider = store.getNodeProvider('checkpoints')
+      expect(provider).toBeDefined()
+    })
+  })
+
+  describe('getAllNodeProviders', () => {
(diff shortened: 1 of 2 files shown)


tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -0,0 +1,245 @@
...
+
+import {
+  ModelNodeProvider,
+  useModelToNodeStore
+} from '@/stores/modelToNodeStore'
+import { type ComfyNodeDefImpl, useNodeDefStore } from '@/stores/nodeDefStore'
+
+describe('useModelToNodeStore', () => {
+  let store: ReturnType<typeof useModelToNodeStore>
+  let nodeDefStore: ReturnType<typeof useNodeDefStore>
```
DrJKL: Not completely necessary, but you might want to check out the comments I left [in here](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5279#discussion_r2317571270).
I generally prefer to minimize how much mutable state there is ever, but especially in test suites.

The initializing of the stores feels like a pretty important part of the Arrange section for any given test, so it'd be nice to have it directly in the `it()` block.

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  // Create minimal mock for testing - only includes 'name' field since that's
+  // the only property ModelNodeProvider constructor uses and tests verify
+  const createMockNodeDef = (name: string): ComfyNodeDefImpl => {
+    return { name } as ComfyNodeDefImpl
```
DrJKL: Does this need a cast?

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -0,0 +1,245 @@
...
+      'ControlNetLoader',
+      'UNETLoader',
+      'UpscaleModelLoader',
+      'StyleModelLoader',
+      'GLIGENLoader'
+    ]
+
+    const mockNodeDefs: Record<string, ComfyNodeDefImpl> = Object.fromEntries(
+      mockNodeNames.map((name) => [name, createMockNodeDef(name)])
+    )
```
DrJKL: If this is going to be the same every time, it could be worth extracting as a CONSTANT.
Or if it needs to be regenerated because it is mutable/mutated, an extracted method.

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  describe('modelToNodeMap', () => {
+    it('should initialize as empty', () => {
+      expect(Object.keys(store.modelToNodeMap)).toHaveLength(0)
+    })
+
+    it('should populate after registration', () => {
+      store.registerDefaults()
+      expect(Object.keys(store.modelToNodeMap)).toContain('checkpoints')
+      expect(Object.keys(store.modelToNodeMap)).toContain('unet')
```
DrJKL: https://vitest.dev/api/expect.html#expect-arraycontaining
Might be clearer than having separate expectations, unless these are each separately important to check.

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -0,0 +1,245 @@
...
+  })
+
+  describe('getNodeProvider', () => {
+    it('should return provider for registered model type', () => {
+      store.registerDefaults()
+
+      const provider = store.getNodeProvider('checkpoints')
+      expect(provider).toBeDefined()
+      // Optional chaining used because getNodeProvider() can return undefined for unregistered types
+      expect(provider?.nodeDef.name).toBe('CheckpointLoaderSimple')
```
DrJKL: If one piece is possibly undefined, the chain should usually go to the terminus.
Otherwise you'll get that `node` isn't present on `undefined`.

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -0,0 +1,245 @@
...
+  // Create minimal mock for testing - only includes 'name' field since that's
+  // the only property ModelNodeProvider constructor uses and tests verify
+  const createMockNodeDef = (name: string): ComfyNodeDefImpl => {
+    return { name } as ComfyNodeDefImpl
+  }
+
+  beforeEach(() => {
+    setActivePinia(createPinia())
+    store = useModelToNodeStore()
+    nodeDefStore = useNodeDefStore()
```
DrJKL: Also, why is one store and the other nodeDefStore? Feels like playing favorites 😛

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -0,0 +1,245 @@
...
+
+      const checkpointProviders = store.getAllNodeProviders('checkpoints')
+      expect(checkpointProviders).toHaveLength(2)
+      expect(checkpointProviders.map((p) => p.nodeDef.name)).toContain(
+        'CheckpointLoaderSimple'
+      )
+      expect(checkpointProviders.map((p) => p.nodeDef.name)).toContain(
+        'ImageOnlyCheckpointLoader'
+      )
+
```
DrJKL: For all of this style assertion: https://vitest.dev/api/expect.html#expect-objectcontaining

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -0,0 +1,245 @@
...
+        'vae',
+        'controlnet',
+        'unet',
+        'upscale_models',
+        'style_models',
+        'gligen'
+      ]
+
+      expectedTypes.forEach((modelType) => {
+        expect(store.getNodeProvider(modelType)).toBeDefined()
```
DrJKL: When you're looping like this, you usually want to [add .soft](https://vitest.dev/api/expect.html#soft) so that you don't have to keep re-running it if there are multiple violations.

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -1,141 +1,184 @@
...
+        output_is_list: [],
+        output_node: false
+      }
+      return new ComfyNodeDefImpl(def)
+    }
+
+    // Mock nodeDefStore dependency - modelToNodeStore relies on this for registration
+    // Most tests expect this to be populated; tests that need empty state can override
+    const nodeDefStore = useNodeDefStore()
+    nodeDefStore.nodeDefsByName = Object.fromEntries(
```
arjansingh: use vitest mocking

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -145,101 +188,106 @@ describe('useModelToNodeStore', () => {
...
-        'style_models',
-        'gligen'
-      ]
-
-      expectedTypes.forEach((modelType) => {
-        expect(store.getNodeProvider(modelType)).toBeDefined()
+      const modelToNodeStore = useModelToNodeStore()
+      modelToNodeStore.registerDefaults()
+
+      EXPECTED_DEFAULT_TYPES.forEach((modelType) => {
```
arjansingh: `for of`

DrJKL (approved):
👍🏻

christian-byrne:
Oh I never got a chance to review. I think for the model selector dialog, we will need a mapping in the other direction as well so we will need to add a method in the store for that.

PR #5327, opened 2025-09-04
fix: remove clipping by removing unnecessary css contain

simula-r:
### Problem:
When zooming, dragging or panning, the vue nodes get clipped/cut off to the size of the initial viewport.

image

This is because the TransformPane parent element's CSS pixel width and height is set only once during the first layout to be the size of the viewport. And using the css property contain: paint clips anything beyond that size.

### Solution:
Remove the contain property because we are already culling/remove nodes outside the viewport and we are relying of the fact that with transform elements skip layout.

image

## Review Focus

Fixes https://www.notion.so/comfy-org/Transform-pane-becomes-misaligned-with-the-canvas-viewport-2626d73d365080bda9a9f2bfe31e3fb5?source=copy_link

### DIFF
--- a/src/components/graph/TransformPane.vue
+++ b/src/components/graph/TransformPane.vue
@@ -117,7 +117,6 @@ useCanvasTransformSync(props.canvas, syncWithCanvas, {
 .transform-pane {
   position: absolute;
   inset: 0;
-  contain: layout style paint;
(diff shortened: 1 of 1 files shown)


christian-byrne (approved):
LGTM!

PR #5332, opened 2025-09-04
[fix] Remove useDefineForClassFields from tsconfig

snomiao:
## Summary
Remove the `useDefineForClassFields` compiler option from tsconfig.json to fix Playwright test configuration compatibility issues.

## Changes
- Removed `useDefineForClassFields: true` from tsconfig.json

## Test plan
- [ ] CI/CD tests pass
- [ ] Playwright tests run successfully
- [ ] Build completes without errors

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5332-fix-Remove-useDefineForClassFields-from-tsconfig-2646d73d365081dea966cc4fa56d5dba) by [Unito](https://www.unito.io)


christian-byrne:
In the spec, it says this is set to `true` if the target is ES2022, which ours is.

snomiao:
## Analysis of i18n Playwright Test Failure

The `Update Locales` workflow is failing with the error:
```
SyntaxError: TypeScript 'declare' fields must first be transformed by @babel/plugin-transform-typescript
```

### Root Cause
The i18n collection scripts (`scripts/collect-i18n-*.ts`) are importing TypeScript source files directly and executing them in Node.js context during Playwright tests. This bypasses the normal Vite build pipeline and instead relies on Playwright's TypeScript compilation, which uses Babel under the hood.

### Why This Happens
1. **Direct imports**: The i18n scripts import TypeScript modules like `../src/constants/coreMenuCommands` directly
2. **Babel transformation**: Playwright uses Babel to transform TypeScript, not the TypeScript compiler (tsc)
3. **Missing plugin configuration**: Babel needs `@babel/plugin-transform-typescript` to handle `declare` fields, but it must run **before** class property plugins

### The Issue with `declare` Fields
When `useDefineForClassFields: false` was removed from tsconfig.json, TypeScript now uses `declare` for class fields that shouldn't be initialized. However, Babel's plugin order matters - the TypeScript plugin must process `declare` fields before class property transformations.

### Possible Solutions
1. **Configure Babel for Playwright** (recommended): Add a babel config specifically for Playwright tests that ensures correct plugin order
2. **Re-add `useDefineForClassFields: true`**: Explicitly set this in tsconfig to avoid `declare` fields
3. **Refactor i18n scripts**: Avoid importing TypeScript source files directly, use compiled outputs instead
4. **Update Playwright's TypeScript handling**: Configure Playwright to use tsc instead of Babel

The TypeScript and Playwright version updates (5.9.2 and 1.55.0) are working correctly for the main test suite - the issue is specific to how the i18n collection scripts compile TypeScript.

snomiao:
close this as the solution is similar to - [refactor: Remove TypeScript declare keywords for Playwright compatibility by snomiao · Pull Request #5304 · Comfy-Org/ComfyUI_frontend]( https://github.com/Comfy-Org/ComfyUI_frontend/pull/5304# )

PR #5337, opened 2025-09-04
[CI] Fix desktop API types update action

webfiltered:
Use --workspace-root flag to install at workspace level.  The GH Action requires this due to pnpm workspace changes.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5337-CI-Fix-desktop-API-types-update-action-2646d73d3650811c9e67daf1be2de36f) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/update-electron-types.yaml
+++ b/.github/workflows/update-electron-types.yaml
@@ -35,7 +35,7 @@ jobs:
             electron-types-tools-cache-${{ runner.os }}-
 
       - name: Update electron types
-        run: pnpm install @comfyorg/comfyui-electron-types@latest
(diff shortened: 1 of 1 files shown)


webfiltered:
@DrJKL Sanity check, please~

... the action, not me.

DrJKL (approved):

DrJKL:
Yeah, we'll have to update it once we move the current app code down into an `/apps`, but for now targeting the root makes sense.

webfiltered:
Follow-up inbound.  I just eyeballed the action; it'll throw if you try to run it.

PR #5344, opened 2025-09-04
Prototype/asset browser

arjansingh:
## Summary

Draft of prototype so we can comment on stuff. DO NOT MERGE.

Most of this is vibe coded hence stuff like handwritten css. Bigger takeaways here will be the integration pattern which i think we can make solid and document for future use.

The Integration guide is claude's version of how it got this to work.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5344-Prototype-asset-browser-2646d73d365081429df7d7fd415a503f) by [Unito](https://www.unito.io)

### DIFF
--- a/docs/WIDGET_INTEGRATION_GUIDE.md
+++ b/docs/WIDGET_INTEGRATION_GUIDE.md
@@ -1,0 +1,929 @@
+# Widget Integration Guide: Vue Components + LiteGraph
+
+## Overview
+This guide documents how to integrate Vue components with the LiteGraph widget system. The primary use case is replacing standard widgets (combo, text, number, etc.) with custom Vue components that provide enhanced UI experiences like modal dialogs, advanced selectors, or interactive controls.
+
+**Common Use Cases:**
+- Asset browsers (checkpoints, LoRAs, VAEs, schedulers)
+- Advanced text editors with syntax highlighting  
+- Interactive parameter controls with sliders and presets
+- File upload interfaces with drag-and-drop
+- Multi-select components with search and filtering
+- Color pickers, date selectors, or other specialized inputs
+
+**Core Pattern:** This guide demonstrates the complete pattern using an asset browser example, but the same architectural principles apply to any widget replacement scenario.
+
+## Architecture Layers
+
+```
+┌─────────────────────────────────────┐
+│           Vue App Layer             │
+│  - YourCustomWidget.vue            │
+│  - YourBrowserDialog.vue           │
+│  - PrimeVue Components             │
+└─────────────────────────────────────┘
+                  ↕
+┌─────────────────────────────────────┐
+│         Integration Layer           │
+│  - useYourCustomWidget()           │
+│  - ComponentWidgetImpl             │
+│  - ComfyWidgets registry           │
+└─────────────────────────────────────┘
+                  ↕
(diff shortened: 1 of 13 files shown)


docs/WIDGET_INTEGRATION_GUIDE.md
```diff
@@ -0,0 +1,929 @@
...
+```vue
+<template>
+  <div class="asset-picker-widget">
+    <!-- Widget display -->
+    <div class="selected-asset-display">
+      <span class="asset-name">{{ displayName }}</span>
+      <Button @click="openAssetBrowser" />
+    </div>
+    
+    <!-- Modal wrapper - CRITICAL: Use PrimeVue Dialog -->
```
arjansingh, line 213: Probably can remove this... that's more useful for building a modal than getting a working vue component

docs/WIDGET_INTEGRATION_GUIDE.md
```diff
@@ -0,0 +1,929 @@
...
+        setValue: (newValue: string) => {
+          // Proper setValue implementation with canvas context
+          const globalApp = globalThis as { app?: { canvas?: LGraphCanvas } }
+          const canvas = globalApp.app?.canvas
+          
+          if (!canvas) {
+            throw new Error('Canvas is required for setValue operation')
+          }
+
+          // Create synthetic CanvasPointerEvent
```
arjansingh, line 339: note: ask it to explain

src/composables/useAssetBrowserDialog.ts
christian-byrne, line 1: Might be cleaner to just make this a method of the dialog service itself.

src/composables/widgets/useAssetComboWidget.ts
christian-byrne, line 1: What about just adding an `onClick` to the LiteGraph widget that called the `useAssetBrowserDialog().show()`?

src/scripts/widgets.ts
```diff
@@ -287,6 +287,6 @@ export const ComfyWidgets: Record<string, ComfyWidgetConstructor> = {
   BOOLEAN: transformWidgetConstructorV2ToV1(useBooleanWidget()),
   STRING: transformWidgetConstructorV2ToV1(useStringWidget()),
   MARKDOWN: transformWidgetConstructorV2ToV1(useMarkdownWidget()),
-  COMBO: transformWidgetConstructorV2ToV1(useComboWidget()),
+  COMBO: transformWidgetConstructorV2ToV1(useAssetComboWidget()),
```
christian-byrne, line 290: Possible alternative:

- Add an extension that pretty much exactly mirrors this structure: https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/extensions/core/uploadImage.ts, except instead of checking for inputs with specific properties, check for specific nodes by name => add some property on the relevant input (e.g., `model`, `checkpoint`)
- Then, in LiteGraph node->widget construction, check if that property exists on the input, if so, add an `onClick` callback (or change in some way to fit desire)

arjansingh: Claude actually originally suggested an extension and i told it not to. I'm going to make sure I understand extensions better. That's a me error not a Claude error. Good call out

PR #5346, opened 2025-09-04
Convert class-based LayoutMutations to composable and define interface where it's used

christian-byrne:
- Convert layout mutations to composable so its methods can be used selectively
- Remove unecessary boilerplate abstraction layer `useLayout`
- Define the interface where it's used

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5346-Convert-class-based-LayoutMutations-to-composable-and-define-interface-where-it-s-used-2646d73d3650818a8de0e8eb7cb624d9) by [Unito](https://www.unito.io)


src/renderer/core/layout/operations/LayoutMutations.ts
```diff
@@ -18,25 +17,70 @@ import {
...
+  // Reroute operations
+  createReroute(
+    rerouteId: string | number,
+    position: Point,
+    parentId?: string | number,
+    linkIds?: (string | number)[]
+  ): void
+  deleteReroute(rerouteId: string | number): void
+  moveReroute(
+    rerouteId: string | number,
```
christian-byrne, line 49: TODO: type reroute and link IDs more explicitly

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -2845,8 +2844,8 @@ export class LGraphNode
     graph._links.set(link.id, link)
 
     // Register link in Layout Store for spatial tracking
-    layoutMutations.setSource(LayoutSource.Canvas)
-    layoutMutations.createLink(
+    graph.layoutMutations.setSource(LayoutSource.Canvas)
+    graph.layoutMutations.createLink(
```
christian-byrne, line 2848: TODO: this should be accessed via `useLayoutMutations` to remove the dependency on `graph` and the dependency on `graph` having a specific type. The long-term goal is for the layout system to *not* be a property of the graph model. Already done correctly in other places other than here.

src/renderer/core/layout/operations/LayoutMutations.ts
christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5352

src/lib/litegraph/src/LGraphNode.ts
christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5356/

PR #5352, opened 2025-09-04
Add explicit types for reroute and link IDs in layout system and don't change type of reroute IDs from LG

christian-byrne:
- Don't convert reroute and link IDs to string (maintain original type `number`)
- Add explicit types for reroute and link IDs
- Use existing NodeId type for Node IDs
- Be more expressive about normalizing node IDs from `string | number` => `string`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5352-Add-explicit-types-for-reroute-and-link-IDs-in-layout-system-2646d73d365081ca96a0cf3242a8d1ed) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/core/layout/operations/LayoutMutations.ts
+++ b/src/renderer/core/layout/operations/LayoutMutations.ts
@@ -6,12 +6,14 @@
  */
 import log from 'loglevel'
 
+import type { NodeId } from '@/lib/litegraph/src/LGraphNode'
 import { layoutStore } from '@/renderer/core/layout/store/LayoutStore'
 import {
   LayoutSource,
-  type NodeId,
+  type LinkId,
   type NodeLayout,
   type Point,
+  type RerouteId,
   type Size
 } from '@/renderer/core/layout/types'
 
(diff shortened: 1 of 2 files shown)


src/renderer/core/layout/operations/LayoutMutations.ts
```diff
@@ -100,13 +103,14 @@ export function useLayoutMutations(): LayoutMutations {
    * Resize a node
    */
   const resizeNode = (nodeId: NodeId, size: Size): void => {
-    const existing = layoutStore.getNodeLayoutRef(nodeId).value
+    const normalizedNodeId = String(nodeId)
```
DrJKL, line 106: So LinkIds are all numbers now but nodeId is sometimes a string?

DrJKL (approved):

src/renderer/core/layout/operations/LayoutMutations.ts
DrJKL: Will the node IDs also become homogenized at some point?

christian-byrne: I think they should become homogenized and converge on `string`, but it's hard to audit everything at this stage.

For this, I switched to importing the NodeId type from Litegraph so that if/when we do change it, it will be easier.

PR #5356, opened 2025-09-04
Removes node's dependency on LGraph for access to layout mutations composable

christian-byrne:
Removes the dependency added which violates desired architectural direction. It's true that the mutations are still happening inside the component but for now we can at least remove this dependency.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5356-Removes-node-s-dependency-on-LGraph-for-access-to-layout-mutations-composable-2646d73d365081919f69e54d040086a6) by [Unito](https://www.unito.io)


DrJKL (approved):
I like this pattern better.

PR #5357, opened 2025-09-04
Fix incorrect id resolution in image preview store

AustinMroz:
Resolves #5354 

The setNodeOutputsByNodeId would call nodeIdToNodeLocatorId without providing the optional subgraph argument. This causes the method to use the activeSubgraph . When a workflow is first loaded (and potentially when outputs are generated during execution) this method is called when the activeSubgraph does not represent the node a preview is being added for.

setNodeOutputsByNodeId lacks the required parameters to be made safe and is only used by the setNodeOutputs function. It has been removed and setNodeOutputs has been reworked to properly supply a node's graph when querying locator ids.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5357-Fix-incorrect-id-resolution-in-image-preview-store-2646d73d36508195b874f2fad3bec600) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/imagePreviewStore.ts
+++ b/src/stores/imagePreviewStore.ts
@@ -1,6 +1,10 @@
 import { defineStore } from 'pinia'
 
-import { LGraphNode, SubgraphNode } from '@/lib/litegraph/src/litegraph'
+import {
+  LGraphNode,
+  Subgraph,
+  SubgraphNode
+} from '@/lib/litegraph/src/litegraph'
 import {
   ExecutedWsMessage,
   ResultItem,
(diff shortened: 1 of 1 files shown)


src/stores/imagePreviewStore.ts
```diff
@@ -136,17 +140,22 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
       )
     } else if (!Array.isArray(filenames)) {
-      setNodeOutputsByNodeId(node.id, filenames)
+      setOutputsByLocatorId(locatorId, filenames)
     } else {
       const resultItems = createOutputs(filenames, folder, isAnimated)
       if (!resultItems?.images?.length) return
-      setNodeOutputsByNodeId(node.id, resultItems)
+      setOutputsByLocatorId(locatorId, resultItems)
     }
```
DrJKL, line 159: Completely optional restructuring proposal: A pure function that takes in `filenames, folder, isAnimated` and returns what would be the second argument to `setOutputsByLocatorId`.
I think it could get us closer to a less procedural and more declarative/reactive store.

DrJKL (approved):

christian-byrne (approved):
This looks good. The code was originally made with naive assumption that the caller would always be inside the _visible graph_. The justification was that the outputs are set in the draw loop in litegraphService, which only affects the active graph. (This should have been documented better.)

- _visible graph_ is root graph => activeSubgraph is falsy => use plain node ID, or
- _visible graph_ is subgraph => activeSubgraph is truthy => prefix with activeSubgraph ID (locator ID)

If it's the case that either

1. outputs are set in other conditions (e.g., get set outside draw loop), or
2. `activeSubgraph` is otherwise incorrect (e.g., not yet initialized)
 
then that system definitely won't work, making this a great change.

---

The only hesitation is that we want to move away from relying on [components](https://en.wikipedia.org/wiki/Entity_component_system) owning state like `node.graph` and instead use IDs to query centralized systems. However, this is fine for now since everything is encapsulated in `imagePreviewStore.ts` and therefore easy to refactor later.

christian-byrne:
Feel free to merge when ready.

src/stores/imagePreviewStore.ts
AustinMroz: I think the direction would be an improvement, but I'd prefer to keep the scope of this change to *just* the bug fix.

PR #5361, opened 2025-09-04
Fix/toolbox node detection

simula-r:
## Summary

Fix pointer events on SelectionToolbox to prevent unintended clicks on parent container by setting pointer-events: none on parent and pointer-events: auto on  child Panel.

## Review Focus

[Fixes #5340](https://github.com/Comfy-Org/ComfyUI_frontend/issues/5340)

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5361-Fix-toolbox-node-detection-2646d73d3650811c9d90d46b40d092fc) by [Unito](https://www.unito.io)


DrJKL:
You can still click the buttons though, right?

DrJKL (dismissed):
Assuming this was tested manually and the Playwright tests cover it otherwise 👍🏻

simula-r:
Looks like playwright is failing due to an item within the toolbox, color picker. Checking out main, it is broken there as well. Not sure how its in main if playwright caught it here. Undoing my changes still shows it broken... I have a fix but not sure if I should roll it into this PR to appease playwright. Edit: patching now to pass tests. Will improve in a follow up.

browser_tests/tests/selectionToolbox.spec.ts-snapshots/selection-toolbox-multiple-selections-border-chromium-linux.png
DrJKL, line 1: These look like the selection logic in the test broke again. Single item selected for multple-selections, etc.

simula-r: The riddle has been solved! This change is a fix for a broken test (either the selection logic in the fixture or the names of the expectations need to be aligned). So, we will need to update the tests in a follow up. Think ok to merge now. @DrJKL concurs.

src/components/graph/SelectionToolbox.vue
```diff
@@ -83,7 +83,6 @@ const extensionToolboxCommands = computed<ComfyCommandImpl[]>(() => {
 <style scoped>
 .selection-toolbox {
   transform: translateX(-50%) translateY(-120%);
-  will-change: transform, opacity;
```
DrJKL, line 86: Not sure why this was removed...

DrJKL (approved):
Went through the test logic with @simula-r , some of the fixture behavior doesn't match the comments' declared intent. Should be looked at later.

christian-byrne (approved):
Context on misleading test snapshot filename:

> It clicks both text encode nodes
> 
> And the test implies that it's going to select the KSampler when in reality the select deselects the KSampler
> 
> Checking for the selection box's appearance and disappearance could catch a real issue
> 

nit: A TODO or comment or change filename would be good

Issue #5328, opened 2025-09-04
Node resizing


wencxxxxxx:
<img width="449" height="718" alt="Image" src="https://github.com/user-attachments/assets/0c3591e5-b1a2-4af2-abae-7fa8e43ae0b6" />

I want to adjust the width of all nodes to be consistent. There are some nodes that cannot be adjusted to the minimum width because their component names are too long.

Tekaiguy:
It's the title that's preventing you. You circled the widget name, but try renaming the node itself and you'll find that it resizes.

wencxxxxxx:
> It's the title that's preventing you. You circled the widget name, but try renaming the node itself and you'll find that it resizes.

I know this can be done, but it only treats the symptoms and not the root cause.

PR #5371, opened 2025-09-05
fix: misaligned slots in Vue Node UI

LittleSound:
Tried to align the slots and connections on the node. I made this work by changing the config file. But I’m not sure if this way is correct. Looking forward to your guidance.

CleanShot 2025-09-05 at 20 22 40

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5371-fix-misaligned-slots-in-Vue-Node-UI-2656d73d36508136814defa33387cd8e) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LiteGraphGlobal.ts
```diff
@@ -63,10 +64,10 @@ export class LiteGraphGlobal {
   NODE_TITLE_TEXT_Y = 20
   NODE_SLOT_HEIGHT = 20
   NODE_WIDGET_HEIGHT = 20
-  NODE_WIDTH = 140
-  NODE_MIN_WIDTH = 50
+  NODE_WIDTH = 445
+  NODE_MIN_WIDTH = 445
   NODE_COLLAPSED_RADIUS = 10
-  NODE_COLLAPSED_WIDTH = 80
+  NODE_COLLAPSED_WIDTH = 445
```
webfiltered, line 70: 445 seems wrong - nodes would be massive.  Does reverting these three change anything?

benceruleanlu (commented):
This leans back on the shared contract system COMFY_VUE_NODE_DIMENSIONS

It can be a good hotfix if the slots layout store's bugs aren't fixed in time for the beta

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -1718,7 +1718,7 @@ export class LGraphNode
       }
     }
 
-    const minWidth = LiteGraph.NODE_WIDTH * (widgets?.length ? 1.5 : 1)
```
DrJKL, line 1721: This might actually address https://github.com/Comfy-Org/ComfyUI_frontend/issues/5328

How does it impact the non-vue nodes?

src/lib/litegraph/src/LiteGraphGlobal.ts
DrJKL: Just an observation: There are a lot of (possibly correlated) magic numbers in here.

LittleSound: 445 related to the position of the output slots. in my side, 445 is the correct value.

christian-byrne:
I think this has been fixed elsewhere, can we close?

Issue #4888, opened 2025-08-10
[Tech Debt]: Gradually enable knip rules for better code health


huchenlei:
### Overview

Knip has been set up to detect unused exports, packages, and files, but currently only unused files checking is enabled. We should gradually enable other knip rules to improve code health and reduce technical debt.

### Current Status

- ✅ Knip is installed and configured
- ✅ CI integration added 
- ✅ Only unused files checking is enabled
- ❌ All other rules are disabled

#### High Priority
- [ ] `devDependencies` - Remove unused dev dependencies to reduce bundle size and maintenance overhead
- [ ] `dependencies` - Remove unused runtime dependencies 
- [ ] `unlisted` - Identify missing dependencies that should be listed in package.json

#### Medium Priority  
- [ ] `exports` - Clean up unused exported functions/constants
- [ ] `duplicates` - Remove duplicate exports

#### Low Priority
- [ ] `types` - Remove unused exported TypeScript types/interfaces
- [ ] `enumMembers` - Remove unused enum members
- [ ] `classMembers` - Remove unused class members
- [ ] `nsExports` - Remove unused namespace exports
- [ ] `nsTypes` - Remove unused namespace types
- [ ] `binaries` - Remove unused binary dependencies

### Implementation Plan

1. **Phase 1**: Enable `devDependencies` and `dependencies` rules
   - Fix any issues found
   - Monitor CI for a few weeks
   
2. **Phase 2**: Enable `unlisted` rule
   - Add missing dependencies to package.json
   - Monitor CI stability

3. **Phase 3**: Enable export-related rules (`exports`, `duplicates`)
   - Clean up unused exports in smaller batches
   - Focus on files with many unused exports first

4. **Phase 4**: Enable remaining rules
   - Enable remaining rules one by one
   - Clean up issues as they're found

### Benefits

- **Reduced bundle size**: Remove unused dependencies
- **Better maintainability**: Remove dead code
- **Cleaner APIs**: Remove unused exports
- **Dependency hygiene**: Ensure all used dependencies are properly listed

### Configuration

Current knip configuration is in `knip.config.ts`. Rules can be enabled by changing them from `'off'` to `'error'` or removing them from the rules object entirely.

### Related

- Knip documentation: https://knip.dev/
- Initial setup: (reference the PR where knip was added)

PR #5374, opened 2025-09-05
Knip: More Pruning

DrJKL:
## Summary

Turning on more rules and cleaning up the existing violations. 
See https://github.com/Comfy-Org/ComfyUI_frontend/issues/4888

## Changes

- **What**: Removing the blanket exceptions and adding targeted ones for things that we know are currently used.

## Review Focus

Anything here that's secretly or externally used and should be retained.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5374-Knip-More-Pruning-2656d73d365081579a9ec98ba8aab42b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/metadata/flac.ts
+++ b/src/scripts/metadata/flac.ts
@@ -1,4 +1,4 @@
-export function getFromFlacBuffer(buffer: ArrayBuffer): Record<string, string> {
+function getFromFlacBuffer(buffer: ArrayBuffer): Record<string, string> {
   const dataView = new DataView(buffer)
 
   // Verify the FLAC signature
--- a/src/scripts/metadata/flac.ts
+++ b/src/scripts/metadata/flac.ts
@@ -1,4 +1,5 @@
-function getFromFlacBuffer(buffer: ArrayBuffer): Record<string, string> {
+/** @knipIgnoreUnusedButUsedByCustomNodes */
+export function getFromFlacBuffer(buffer: ArrayBuffer): Record<string, string> {
   const dataView = new DataView(buffer)
 
   // Verify the FLAC signature
--- a/src/scripts/metadata/png.ts
+++ b/src/scripts/metadata/png.ts
@@ -1,4 +1,5 @@
-function getFromPngBuffer(buffer: ArrayBuffer) {
+/** @knipIgnoreUnusedButUsedByCustomNodes */
+export function getFromPngBuffer(buffer: ArrayBuffer) {
   // Get the PNG data as a Uint8Array
   const pngData = new Uint8Array(buffer)
   const dataView = new DataView(pngData.buffer)
--- a/src/scripts/metadata/png.ts
+++ b/src/scripts/metadata/png.ts
@@ -1,4 +1,4 @@
-export function getFromPngBuffer(buffer: ArrayBuffer) {
+function getFromPngBuffer(buffer: ArrayBuffer) {
   // Get the PNG data as a Uint8Array
   const pngData = new Uint8Array(buffer)
   const dataView = new DataView(pngData.buffer)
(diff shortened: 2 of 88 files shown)


tests-ui/tests/litegraph/subgraph/fixtures/subgraphHelpers.ts
```diff
@@ -424,4 +424,3 @@ export function createEventCapture<T = unknown>(
 }
 
 // Re-export expect from vitest for convenience
```
christian-byrne: Shall we remove this line as well?

src/types/index.ts
christian-byrne, line 1: To an extent, this file serves as the official public API and is the entry point when building the published types package for npm. This should be documented.

src/scripts/ui/components/asyncDialog.ts
christian-byrne, line 1: This one seems to be used by 3 extensions https://cs.comfy.org/search?q=context:global+asyncDialog&patternType=keyword&sm=0

src/scripts/ui/menu/index.ts
```diff
@@ -5,12 +5,6 @@ import { ComfyButtonGroup } from '../components/buttonGroup'
 import './menu.css'
 
 // Export to make sure following components are shimmed and exported by vite
-export { ComfyButton } from '../components/button'
-export { ComfySplitButton } from '../components/splitButton'
-export { ComfyPopup } from '../components/popup'
-export { ComfyAsyncDialog } from '@/scripts/ui/components/asyncDialog'
-export { DraggableList } from '@/scripts/ui/draggableList'
-export { applyTextReplacements, addStylesheet } from '@/scripts/utils'
```
christian-byrne: I think we need this to not break extensions

tests-ui/tests/litegraph/subgraph/fixtures/subgraphHelpers.ts
DrJKL: Yep. I wish there were a `no-nonsense-comment` linter...

src/types/index.ts
DrJKL: Do you think this would be a good target to move to `/packages/`?

src/scripts/ui/components/asyncDialog.ts
DrJKL: I'll add new entry entries for these index files. Might be worth adding the other barrel files if that's what they're for. I think long-term it'd be good to move things that we're not using ourselves to a dedicated package.

DrJKL (commented):
I think all the comments have been addressed 🙂

src/scripts/metadata/png.ts
```diff
@@ -1,4 +1,4 @@
-export function getFromPngBuffer(buffer: ArrayBuffer) {
+function getFromPngBuffer(buffer: ArrayBuffer) {
```
christian-byrne: https://cs.comfy.org/search?q=context:global+getFromPngBuffer&patternType=keyword&sm=0

src/scripts/metadata/flac.ts
```diff
@@ -1,4 +1,4 @@
-export function getFromFlacBuffer(buffer: ArrayBuffer): Record<string, string> {
+function getFromFlacBuffer(buffer: ArrayBuffer): Record<string, string> {
```
christian-byrne: https://cs.comfy.org/search?q=context:global+getFromFlacBuffer&patternType=keyword&sm=0

christian-byrne (changes requested):
I checked almost all code that is public and older than a couple months on CS

src/scripts/metadata/png.ts
DrJKL: Re-exported in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5374/commits/a939664798935690406b7ef1f55047193fecd3a5

src/scripts/metadata/flac.ts
DrJKL: Re-exported in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5374/commits/a939664798935690406b7ef1f55047193fecd3a5

christian-byrne (approved):
LGTM

PR #5378, opened 2025-09-06
fix: correct Vue ref auto-unwrapping in slot components

benceruleanlu:
## Summary

Fixed Vue's auto-unwrapping behavior for exposed template refs in slot components, resolving slot registration failures.

## Changes

- **What**: Corrected type and access pattern for component refs exposed via `defineExpose` in InputSlot.vue and OutputSlot.vue
- **Breaking**: None
- **Dependencies**: None

## Review Focus

The key fix addresses Vue's automatic ref unwrapping when a child component exposes a ref through `defineExpose`. The parent was incorrectly treating the exposed value as a `Ref<HTMLElement>` and calling `.value`, but Vue already unwraps it to `HTMLElement`.

### Exact Issue
- Child component `SlotConnectionDot.vue` exposes a template ref with `defineExpose({ slotElRef })`, where `slotElRef` comes from `useTemplateRef('slot-el')`
- When a child exposes a Ref via `defineExpose`, Vue auto-unwraps it on the parent component instance proxy
- Previously, the parent treated it as `Ref` and read `.value` (double-unwrapping), which produced `undefined` at runtime
- Result: `slotElRef` stayed null and `useDomSlotRegistration` had no element to measure/register, breaking slot hit-testing/linking

### What Changed
**Before:**
- `connectionDotRef: ref<{ slotElRef: Ref<HTMLElement> }>()`
- Used `watch(connectionDotRef, ...)` and accessed `newValue.slotElRef.value`

**After:**
- `connectionDotRef: ref<ComponentPublicInstance<{ slotElRef: HTMLElement | undefined }> | null>()`
- Used `watchEffect(() => { slotElRef.value = connectionDotRef.value?.slotElRef || null })`

### Why It Works
1. **Auto-unwrapping**: Stopped calling `.value` on an already unwrapped HTMLElement
2. **Timing**: `watchEffect` reruns when the child's exposed `slotElRef` appears (after mount), whereas the old `watch` only reacted to component ref assignment

With these fixes, `slotElRef` is reliably populated, `useDomSlotRegistration` can measure and register slot positions, and slot hit-testing/connection logic works as intended.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5378-fix-correct-Vue-ref-auto-unwrapping-in-slot-components-2666d73d3650818c8908f7c88e5a530d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/InputSlot.vue
+++ b/src/renderer/extensions/vueNodes/components/InputSlot.vue
@@ -32,7 +32,14 @@
 </template>
 
 <script setup lang="ts">
-import { type Ref, computed, inject, onErrorCaptured, ref, watch } from 'vue'
+import {
+  type ComponentPublicInstance,
+  computed,
+  inject,
+  onErrorCaptured,
+  ref,
+  watchEffect
+} from 'vue'
 
 import { useErrorHandling } from '@/composables/useErrorHandling'
 import { getSlotColor } from '@/constants/slotColors'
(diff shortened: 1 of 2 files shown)


benceruleanlu:
This is one of like three issues blocking link rendering from depending on the layout tree properly

LittleSound (approved):

christian-byrne (approved):

PR #5385, opened 2025-09-06
[refactor] Centralize type assertions from yjs data

christian-byrne:
Removes type assertions on all the yjs data accesses and centralizes through getters.

This was determined to be best solution based on reading discussion here:

- https://github.com/yjs/yjs/issues/490

Yjs Maps are typed as `Y.Map<unknown>` because CRDTs must handle arbitrary data types and network synchronization where type safety cannot be guaranteed at runtime. Yjs prioritizes correctness of distributed data structures over TypeScript convenience, making type assertions unavoidable when extracting data from shared maps.

Considered three main approaches to handle this limitation:

- zod runtime validation (rejected due to 500-1000x performance overhead)
- generic typed wrapper classes (rejected for adding unnecessary abstraction complexity)
- schema interface-based getter methods (selected).
 
The schema interface approach (also suggested in https://github.com/yjs/yjs/issues/490) seems best because it centralizes type casting in focused methods while maintaining zero runtime overhead and clean call sites. Our implementation follows this pattern by defining schema interfaces for our data structures and creating property-specific getter methods that encapsulate the necessary type assertions, effectively eliminating scattered casting throughout the business logic while acknowledging that some level of type casting is fundamentally required when working with Yjs CRDTs.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5385-refactor-Centralize-type-assertions-from-yjs-data-2666d73d365081f49c36c519ef04178b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/core/layout/store/layoutStore.ts
+++ b/src/renderer/core/layout/store/layoutStore.ts
@@ -99,11 +99,10 @@ class LayoutStoreImpl implements LayoutStore {
     bounds: { x: 0, y: 0, width: 100, height: 50 }
   }
 
-  private static readonly REROUTE_DEFAULTS: Omit<
-    RerouteData,
-    'id' | 'parentId'
-  > = {
+  private static readonly REROUTE_DEFAULTS: RerouteData = {
+    id: 0,
     position: { x: 0, y: 0 },
+    parentId: 0,
     linkIds: []
   }
 
@@ -205,18 +204,11 @@ class LayoutStoreImpl implements LayoutStore {
   private getRerouteField<K extends keyof RerouteData>(
     yreroute: Y.Map<unknown>,
     field: K,
-    defaultValue?: RerouteData[K]
+    defaultValue: RerouteData[K] = LayoutStoreImpl.REROUTE_DEFAULTS[field]
   ): RerouteData[K] {
     const typedReroute = yreroute as TypedYMap<RerouteData>
     const value = typedReroute.get(field)
-    if (value !== undefined) return value
-    if (defaultValue !== undefined) return defaultValue
-    if (field in LayoutStoreImpl.REROUTE_DEFAULTS) {
-      return (LayoutStoreImpl.REROUTE_DEFAULTS as any)[field]
-    }
-    throw new Error(
-      `No default value available for reroute field: ${String(field)}`
-    )
+    return value ?? defaultValue
   }
 
   /**
--- a/src/renderer/core/layout/store/layoutStore.ts
+++ b/src/renderer/core/layout/store/layoutStore.ts
@@ -35,6 +35,7 @@ import {
   type Point,
   type RerouteId,
   type RerouteLayout,
+  type Size,
   type SlotLayout
 } from '@/renderer/core/layout/types'
 import { SpatialIndexManager } from '@/renderer/core/spatial/SpatialIndex'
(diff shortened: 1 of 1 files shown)


src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -49,7 +50,53 @@ const logger = log.getLogger('LayoutStore')
 // Constants
 const REROUTE_RADIUS = 8
 
+interface NodeLayoutData {
+  id: NodeId
+  position: Point
+  size: Size
+  zIndex: number
+  visible: boolean
+  bounds: Bounds
```
DrJKL, line 68: Could bounds be derived from position and size?

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -49,7 +50,53 @@ const logger = log.getLogger('LayoutStore')
...
+}
+
+interface LinkData {
+  id: LinkId
+  sourceNodeId: NodeId
+  targetNodeId: NodeId
+  sourceSlot: number
+  targetSlot: number
```
DrJKL, line 76: Index based slots match what we have now, but longer term it might cause some issues.
Like for nodes that have a variable number of inputs, removing one in the middle would require each subsequent link to have its targetSlot decremented by 1.
Not super relevant for this PR, but noting it here anyway 🙂

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -49,7 +50,53 @@ const logger = log.getLogger('LayoutStore')
...
+}
+
 class LayoutStoreImpl implements LayoutStore {
+  private static readonly NODE_DEFAULTS: NodeLayoutData = {
+    id: 'unknown-node',
+    position: { x: 0, y: 0 },
+    size: { width: 100, height: 50 },
+    zIndex: 0,
+    visible: true,
+    bounds: { x: 0, y: 0, width: 100, height: 50 }
```
DrJKL, line 99: Something like

```ts
get bounds() {
  const { x, y } = this.position
  const { width, height } = this.size
  return { x, y, width, height }
}
```

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -49,7 +50,53 @@ const logger = log.getLogger('LayoutStore')
...
+  }
+
+  private static readonly REROUTE_DEFAULTS: RerouteData = {
+    id: 0 as RerouteId,
+    position: { x: 0, y: 0 },
+    parentId: 0 as LinkId,
```
DrJKL: Do these need the type assertions?

DrJKL: Also, baking in the ids by default feels risky. I can imagine a bug where we forget to properly set it somewhere and have a bunch of Reroutes in the state all linked to 0 building up...
Could we have a `Partial<RerouteData>` or even `Omit<RerouteData, 'id'|'parentId'>` for just the position and linkIds to explicitly require that the other pieces be provided when instantiating?

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -127,6 +174,42 @@ class LayoutStoreImpl implements LayoutStore {
     })
   }
 
+  private getNodeField<K extends keyof NodeLayoutData>(
+    ynode: Y.Map<unknown>,
+    field: K,
+    defaultValue: NodeLayoutData[K] = LayoutStoreImpl.NODE_DEFAULTS[field]
+  ): NodeLayoutData[K] {
+    const typedNode = ynode as TypedYMap<NodeLayoutData>
```
DrJKL, line 191: It's really too bad this can't be enforced at the function interface level ☹️

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -127,6 +174,42 @@ class LayoutStoreImpl implements LayoutStore {
...
+    defaultValue: RerouteData[K] = LayoutStoreImpl.REROUTE_DEFAULTS[field]
+  ): RerouteData[K] {
+    const typedReroute = yreroute as TypedYMap<RerouteData>
+    const value = typedReroute.get(field)
+    return value ?? defaultValue
+  }
+
+  private asRerouteId(id: string | number): RerouteId {
+    return Number(id)
+  }
```
DrJKL: I don't love `RerouteId` and `LinkId` just being aliases for `number` with no additional branding and possibly unsafe type coercion/conversion here.
What happens if the string passed in resolves to `NaN` or one of the infinities in here?

Is there a way we can prevent ever having strings that we'd have to try to numberficate?

DrJKL: (Complete nit otherwise, any method in here that doesn't reverence `this` should probably be a pure function in module scope)

src/renderer/core/layout/store/layoutStore.ts
DrJKL: Or really I'd rather have a
```ts
function getNodeBounds(node: NodeLayoutData): Bounds {...}
```

DrJKL (dismissed):
I love the direction and most of the implementation, but I have a few comments.

src/renderer/core/layout/store/layoutStore.ts
christian-byrne: Done in 326f2c9706e5385a01336aed50cc0beb813af384

christian-byrne:
Need to spend more time thinking about this PR

src/renderer/core/layout/store/layoutStore.ts
christian-byrne: Can it?

src/renderer/core/layout/store/layoutStore.ts
AustinMroz: Many parts of the litegraph codebase check for the truthiness of link or reroute ids as indication that a link exists. Maybe not  the cleanest design, but the code here is safe.

christian-byrne:
I'm going to convert the unresolved comments to Issue Tasks and merge since this PR makes changing layoutStore a lot easier (avoid type assertion inline on every access) and will be very difficult to rebase later.

PR #5399, opened 2025-09-06
[fix] Remove TypeScript declare modifiers from class properties

snomiao:
## Summary
- Replaced TypeScript `declare` modifiers with standard optional property syntax across litegraph files
- Fixed Playwright test to properly await route setup
- Updated test snapshots to reflect property changes

## Details
This PR removes the use of TypeScript's `declare` modifier on class properties and replaces them with standard optional property syntax (`?:`). The `declare` modifier is primarily intended for ambient declarations and can cause type compatibility issues when used on regular class properties.

### Changes made:
- **LGraphNode**: Converted `declare comfyClass?` and `declare isVirtualNode?` to optional properties
- **SubgraphNode**: Added explicit `override` for inputs property with proper typing
- **SubgraphInput/Output**: Removed `declare parent` and use type assertions where needed
- **BaseWidget**: Converted `declare` properties to optional properties
- **FloatingRenderLink**: Added explicit type assertions for parent nodes
- **Playwright test**: Added missing `await` for route setup

## Test plan
- [ ] Type checking passes (`pnpm typecheck`)
- [ ] Unit tests pass (`pnpm test:unit`)
- [ ] Playwright tests pass (`pnpm test:browser`)
- [ ] Linting passes (`pnpm lint`)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5399-fix-Remove-TypeScript-declare-modifiers-from-class-properties-2666d73d36508185969bf96c4eacbc41) by [Unito](https://www.unito.io)


christian-byrne:
It's really confusing to see how removing the `declare` in that location would necessite some of these other changes.

PR #5404, opened 2025-09-06
[feat] Implement AssetService behind settings flag

arjansingh:
## Summary

Asset API integrated via `assetService` and behind a feature switch in settings.

## Changes

1. Added ` Comfy.Assets.UseAssetAPI to CORE_SETTINGS` toggle. If on, frontend uses `assetService` to grab models and load model folders. If off, frontend uses the existing model manager API. Defaults to off.
2. Create `assetService` to manage retrieving model data and model folders. 
3. Configure model store to switch between the two APIs depending on the flag

## Review Focus

1. No dedicated frontend caching logic for now. I think it would better for the backend to send cache-control headers for those endpoints. We can implement an LRU in in an `assetStore` if needed and move the service to read from that.
2. I made some executive decisions on how to handle the folder structure with the Assets API. TL;DR: We're gonna fake the existing structure by hard mocking it. Details in the code.
3. Most of this PR is test code. I'm trying to write unit tests along the way so we can rely less on browser tests eventually.
4. I'm keeping a hardcoded list of tags right now [here](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5404/files#diff-bb7adf4a69a36d101c04c22d0cdbddf5b0651018b73b2fb86c507049a60ea89fR4-R5) that get special treatment for now.

None, everything is behind the scenes.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5404-feat-Implement-AssetService-behind-settings-flag-2666d73d365081c293dfcc93761df52c) by [Unito](https://www.unito.io)


src/services/assetService.ts
```diff
@@ -0,0 +1,151 @@
...
+function isValidAsset(asset: unknown): asset is Asset {
+  return (
+    asset !== null &&
+    typeof asset === 'object' &&
+    'id' in asset &&
+    'name' in asset &&
+    'tags' in asset &&
+    Array.isArray((asset as Asset).tags)
+  )
+}
```
christian-byrne, line 34: nit: great use-case for zod schema and zod validation.

src/services/assetService.ts
christian-byrne, line 1: This does not need to be changed neceessarily because I'm not sure how this service will look after proceeding PRs but if this module / `creatAssetService` have no internal persistent state, we can just put all the functions in the module scope.

If we want to add state e.g. for deduplication, request caching, retry handling, reactivity then we could switch to `useAssertService` and add reactivity and vue.

christian-byrne (approved):
This looks good. I left some comments that could inform follow-up PRs.

Also questions for future:

2. How are we handling the scanning being in a pending state? Theoretically we want to be notified when the scan finishes but could also poll if the database is updated incrementally.
1. Since we are not watching FS on BE and assuming the scanning is done or not started, these routes are all idempotent right? So we can cache the requests e.g. using something like https://github.com/Comfy-Org/ComfyUI_frontend/blob/e2de4b19fc3e2a60aa4072961bc90cc16e79edf8/src/composables/useCachedRequest.ts#L22

src/services/assetService.ts
arjansingh: Good call. Yes I'm deliberately trying to keep stuff simple here and add complexity.

src/constants/coreSettings.ts
```diff
@@ -972,5 +972,14 @@ export const CORE_SETTINGS: SettingParams[] = [
     defaultValue: false,
     experimental: true,
     versionAdded: '1.27.1'
+  },
+  {
+    id: 'Comfy.Assets.UseAssetAPI',
+    name: 'Use Asset API for model library',
+    type: 'boolean',
+    tooltip:
+      'Use new asset API instead of experiment endpoints for model browsing',
```
DrJKL, line 981: I don't think the average user would know what experiment endpoints means.

```suggestion
      'Use new Asset API for model browsing',
```

src/scripts/api.ts
```diff
@@ -682,7 +682,8 @@ export class ComfyApi extends EventTarget {
     }
     const folderBlacklist = ['configs', 'custom_nodes']
     return (await res.json()).filter(
```
DrJKL, line 684: Not necessarily for this PR, but usually when I see `res.json()` I'd expect a Zod parse.

src/services/assetService.ts
DrJKL: I love having guards, but the ones in here are a _little_ hard to reason about.
Like, normally I would only expect a single type guard per type, but in this case there are three for `is Asset`.
Could the filter predicates just return boolean, or maybe indicate a specific subtype of Asset?

src/services/assetService.ts
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * Logic:
+   * 1. Extract directory names directly from asset tags
+   * 2. Filter out blacklisted directories
+   * 3. Return alphabetically sorted directories with assets
+   *
+   * @returns The list of model folder keys
+   */
+  async function getAssetModelFolders(): Promise<
+    { name: string; folders: string[] }[]
```
DrJKL, line 94: This shape is in two places, could it be an `interface`?

src/services/assetService.ts
```diff
@@ -0,0 +1,151 @@
...
+      const directoryTags = data.assets
+        .filter(createAssetFolderFilter())
+        .flatMap((asset) => asset.tags)
+        .filter(
+          (tag) => tag !== MODELS_TAG && !blacklistedDirectories.includes(tag)
+        )
+
+      for (const tag of directoryTags) {
+        discoveredFolders.add(tag)
+      }
```
DrJKL, line 116: You could put the chain into the Set constructor here instead of adding each item.
Or something like `(data?.assets ?? [])`, though that might be a little awkward to read.

src/services/assetService.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    // Return only discovered folders in alphabetical order
+    const sortedFolders = Array.from(discoveredFolders).sort()
```
DrJKL, line 120: https://caniuse.com/mdn-javascript_builtins_array_tosorted

DrJKL: I think this is in the target browsers, and if not our transpiler should be able to handle it for us.

src/services/assetService.ts
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  /**
+   * Gets a list of models in the specified folder from the asset API
+   * @param folder The folder to list models from, such as 'checkpoints'
+   * @returns The list of model filenames within the specified folder
+   */
+  async function getAssetModels(
+    folder: string
+  ): Promise<{ name: string; pathIndex: number }[]> {
```
DrJKL, line 131: This shape also feels like it might deserve a name.

src/services/assetService.ts
```diff
@@ -0,0 +1,151 @@
...
+    const data = await handleAssetRequest(
+      `${ASSETS_ENDPOINT}?include_tags=${MODELS_TAG},${folder}`,
+      `models for ${folder}`
+    )
+
```
DrJKL, line 136: Totally optional style change here.

```suggestion

    return data?.assets?.filter(createFolderAssetFilter(folder))?.map((asset) => ({
          name: asset.name,
          pathIndex: 0
        })) ?? []
```

src/stores/modelStore.ts
```diff
@@ -153,7 +155,12 @@ export class ModelFolder {
   models: Record<string, ComfyModelDef> = {}
   state: ResourceState = ResourceState.Uninitialized
 
-  constructor(public directory: string) {}
+  constructor(
+    public directory: string,
+    private getModelsFunc: (
+      folder: string
+    ) => Promise<{ name: string; pathIndex: number }[]>
```
DrJKL, line 162: Yup, definitely nameable.

src/stores/modelStore.ts
```diff
@@ -197,11 +205,22 @@ export const useModelStore = defineStore('models', () => {
...
+    const useAssetAPI: boolean = settingStore.get('Comfy.Assets.UseAssetAPI')
+
+    const resData = useAssetAPI
+      ? await assetService.getAssetModelFolders()
+      : await api.getModelFolders()
     modelFolderNames.value = resData.map((folder) => folder.name)
     modelFolderByName.value = {}
     for (const folderName of modelFolderNames.value) {
-      modelFolderByName.value[folderName] = new ModelFolder(folderName)
+      const getModelsFunc = useAssetAPI
```
DrJKL, line 216: I'm generally wary when a conditional starts multiplying further down...

tests-ui/tests/services/assetService.test.ts
```diff
@@ -0,0 +1,150 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { api } from '@/scripts/api'
+import { assetService } from '@/services/assetService'
+
+// Test data constants
+const MOCK_ASSETS = {
```
DrJKL, line 7: Just want to say I love the preamble mock setup in here.

tests-ui/tests/services/assetService.test.ts
```diff
@@ -0,0 +1,150 @@
...
+function mockApiError(status: number, statusText = 'Error') {
+  vi.mocked(api.fetchApi).mockResolvedValueOnce(
+    new Response(null, { status, statusText })
+  )
+}
+
+describe('assetService', () => {
+  beforeEach(() => {
+    vi.resetAllMocks()
+    vi.spyOn(api, 'fetchApi')
```
DrJKL, line 49: Do you need to re-spy each time?

tests-ui/tests/services/assetService.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+      expect(api.fetchApi).toHaveBeenCalledWith('/assets?include_tags=models')
+      expect(result).toHaveLength(2)
+
+      const folderNames = result.map((f) => f.name)
+      expect(folderNames).toEqual(['checkpoints', 'vae'])
+      expect(folderNames).not.toContain('configs')
+    })
+
+    it('should handle errors and empty responses', async () => {
```
DrJKL, line 86: Would you mind splitting this into three tests for signal fidelity?

tests-ui/tests/store/modelStore.test.ts
```diff
@@ -13,7 +15,32 @@ vi.mock('@/scripts/api', () => ({
...
+  // Mock settingStore to return the useAssetAPI setting
+  const mockSettingStore = {
+    get: vi.fn().mockImplementation((key: string) => {
+      if (key === 'Comfy.Assets.UseAssetAPI') {
+        return useAssetAPI
+      }
+      return false
+    })
+  }
+  vi.mocked(useSettingStore).mockReturnValue(mockSettingStore as any)
```
DrJKL, line 41: Could you use `{ partial: true }` to avoid the `as any`?

tests-ui/tests/store/modelStore.test.ts
```diff
@@ -95,11 +133,44 @@ describe('useModelStore', () => {
...
+      await store.loadModelFolders()
+      const folderStore = await store.getLoadedModelFolder('checkpoints')
+
+      // Both APIs return objects with .name property, modelStore extracts folder.name in both cases
+      expect(assetService.getAssetModelFolders).toHaveBeenCalledTimes(1)
+      expect(assetService.getAssetModels).toHaveBeenCalledWith('checkpoints')
+      expect(api.getModelFolders).toHaveBeenCalledTimes(0)
+      expect(api.getModels).toHaveBeenCalledTimes(0)
+      expect(folderStore).toBeDefined()
+      expect(Object.keys(folderStore!.models)).toHaveLength(3)
```
DrJKL, line 173: I think this is probably fine for now, but it feels a little change-detectory.

DrJKL (approved):
Nothing critical here, but I left a few possibilities for cleanup/clarity.

arjansingh:
> This looks good. I left some comments that could inform follow-up PRs.
> 
> Also questions for future:
> 
> 2. How are we handling the scanning being in a pending state? Theoretically we want to be notified when the scan finishes but could also poll if the database is updated incrementally.

I originally had the code triggering the scan and watching for it before @bigcat88 added scan on startup. I can add that code back when we handle refreshing. (See below for thoughts there)

> 3. Since we are not watching FS on BE and assuming the scanning is done or not started, these routes are all idempotent right? So we can cache the requests e.g. using something like https://github.com/Comfy-Org/ComfyUI_frontend/blob/e2de4b19fc3e2a60aa4072961bc90cc16e79edf8/src/composables/useCachedRequest.ts#L22

Yup, totally agree these are great candidates for caching.

The tricky bit is we run both locally (users downloading models, triggering scans) and in cloud (we control availability) so caching needs to handle both.

We should do:

Server-side cache headers:
```
  # local: short ttl since users add models
  Cache-Control: max-age=300

  # cloud: longer ttl since we control updates
  Cache-Control: max-age=1800
```

Use `useCachedRequest` for query deduplication:
```typescript
  const folders = ref([])
  const models = ref({})

  const foldersCache = useCachedRequest(
    () => api.fetchApi('/assets?include_tags=models'),
    { maxSize: 10 }
  )

  const loadFolders = async () => {
    folders.value = await foldersCache.call()
  }
```

That way multiple components hitting same endpoint simultaneously get the cached result. No server spamming.

For client-side behavior, like if they click "refresh", claude on first impression came up with this psuedocode. We'd probably want backoff and retry limit logic in addition:
```typescript
  const isScanning = ref(false)
  const scanProgress = ref(0)

  const rescan = async () => {
    // clear caches
    foldersCache.clear()
    modelsCache.clear()

    // trigger scan
    await api.post('/api/assets/scan/schedule', { roots: ['models'] })
    isScanning.value = true

    // poll for status
    while (isScanning.value) {
      const response = await api.get('/api/assets/scan?root=models')
      const scans = response.scans || []
      const modelScan = scans.find(s => s.root === 'models')

      if (modelScan) {
        scanProgress.value = modelScan.progress || 0
        isScanning.value = modelScan.status !== 'completed'
      } else {
        isScanning.value = false
      }

      if (isScanning.value) {
        await new Promise(resolve => setTimeout(resolve, 1000))
      }
    }

    // reactive refetch with fresh data
    await loadFolders()
  }
```

Same frontend code works everywhere, server just sets appropriate cache ttl based on deployment. 

We handle the scan invalidation on frontend when user explicitly rescans. Still want to think through polling. Can we leverage our websocket setup to just tell the client to refresh?

src/services/assetService.ts
arjansingh: to do this: ES2022 --> ES2023 

we to update need this in our tsconfig. shall i go for it?

PR #5079, opened 2025-08-18
[fix] Restore ability to override onMouseDown in node subclasses

christian-byrne:
## Summary

This PR fixes issue #5073 where node subclasses could no longer override the `onMouseDown` method after the title button feature was added.

The regression was introduced in commit 3b4c0b5d where `onMouseDown` was assigned as a property in the constructor, preventing proper method inheritance.

## Changes

- Removed the `onMouseDown` assignment from the `LGraphNode` constructor
- Moved title button click detection to `LGraphCanvas`, where it now checks for button clicks before calling the node's `onMouseDown` method
- This preserves title button functionality while restoring the ability for custom nodes to override `onMouseDown`

## Test Plan

- [x] Title buttons still work correctly (tested manually)
- [x] Custom nodes can now override `onMouseDown` (fixes the reported issue)
- [x] TypeScript compilation passes
- [x] Existing functionality is preserved

Note: Some tests that directly called `node.onMouseDown()` expecting it to handle title buttons may need updates in a follow-up PR, but the user-facing functionality remains unchanged.

Fixes #5073

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5079-fix-Restore-ability-to-override-onMouseDown-in-node-subclasses-2536d73d365081769d36d40c0cb0d9f7) by [Unito](https://www.unito.io)

PR #5405, opened 2025-09-06
Remove accidental onMouseDown handler

AustinMroz:
#5079 removed the onMouseDown handler because it would overwrite prototype onMouseDown handlers added by custom nodes. This handler was accidentally re-added by #4263.

As the replacement mouse handling code added by #5079 still exists, only the removal of the onMouseDown code is required here.

EDIT: Looks like it also reverted test changes.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5405-Remove-accidental-onMouseDown-handler-2666d73d36508126bba7c2e5eaf8a509) by [Unito](https://www.unito.io)


DrJKL (dismissed):
Let's hope this is adequately covered by the tests 🤞🏻

AustinMroz:
It's a travesty that there's now green lines in the PR, but I've added a test that is harder to overwrite than the snapshots.

DrJKL (approved):

christian-byrne (approved):
Great catch

PR #5406, opened 2025-09-06
Fix canvas/screen conversion formulas in useTransformState

benceruleanlu:
## Summary

Fixed the canvas-to-screen and screen-to-canvas conversion formulas to match LiteGraph's actual transform order.

## Changes

- **What**: Corrected the transform formulas in `canvasToScreen` and `screenToCanvas` functions to properly account for the order of operations (translate then scale)
- **Breaking**: None - This fixes alignment issues without changing the API

## Review Focus

The key change is understanding the transform order: LiteGraph applies `scale(scale); translate(offset)` which in CSS terms means "translate first, then scale", resulting in the formula `screen = (canvas + offset) * scale` rather than `screen = canvas * scale + offset`.

This ensures Vue components positioned using these converters will correctly align with their canvas counterparts.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5406-Fix-canvas-screen-conversion-formulas-in-useTransformState-2666d73d365081a0ac0bc8c391e75070) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/core/layout/useTransformState.ts
+++ b/src/renderer/core/layout/useTransformState.ts
@@ -74,6 +74,10 @@ export const useTransformState = () => {
 
   // Computed transform string for CSS
   const transformStyle = computed(() => ({
+    // Match LiteGraph DragAndScale.toCanvasContext():
+    // ctx.scale(scale); ctx.translate(offset)
+    // CSS applies right-to-left, so "scale() translate()" -> translate first, then scale
+    // Effective mapping: screen = (canvas + offset) * scale
     transform: `scale(${camera.z}) translate(${camera.x}px, ${camera.y}px)`,
     transformOrigin: '0 0'
   }))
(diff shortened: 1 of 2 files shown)


benceruleanlu:
although even after this, it isn't the last fix for links to be aligned with slots, we need to offset properly, although after this the links won't go crazy when zooming or freak out when toggle-move-toggle

christian-byrne:
> LiteGraph applies scale(scale); translate(offset)

Do you mean LiteGraph is applying transform via CSS on the canvas HTML element?

christian-byrne:
This is using the Canvas2D transform system rather than CSS/DOM

https://github.com/Comfy-Org/ComfyUI_frontend/blob/2b8d3548627c8185dee33abdac7766efebbdec91/src/lib/litegraph/src/DragAndScale.ts#L123-L140

benceruleanlu:
sorry I should've checked the CC generated PR description, it's rather misleading and nondescriptive

tl;dr: LG does **offset then scale** (effect order). Our TS code effectively did **scale then offset** (effect order), which is out of order. This PR fixes it.

long version:

under the hood, **Canvas2D post-multiplies** each transform into a Current Transform Matrix (CTM): `CTM ← CTM × M`, and points are transformed as `p' = CTM × p`. Because of this, the **rightmost matrix acts first** on the point.

```ts
 toCanvasContext(ctx: CanvasRenderingContext2D): void { 
   ctx.scale(this.scale, this.scale) 
   ctx.translate(this.offset[0], this.offset[1]) 
 }
```

is actually doing this:

* Start: `CTM = I`
* After `scale`: `CTM = S`
* After `translate`: `CTM = S × T`
* Applied to a point: `p' = (S × T) × p = S × (T × p)` → **translate, then scale**, allowed via associative not commutive

PR #5407, opened 2025-09-06
fix: `onNodeRemoved` not called when loading new graph (and tearing down previous)

christian-byrne:
## Summary

Fixes issue where Vue nodes would persist after loading new workflows

The root cause was that LGraph.clear() only called individual node.onRemoved() callbacks but not the graph-level onNodeRemoved() callback that Vue node management relied on. This created a cleanup gap during workflow loading.
## Changes

- Add onNodeRemoved callback to LGraph.clear() to match behavior with individual node removal
- Modify app.clean() to call graph.clear() ensuring proper Vue node cleanup
- Remove redundant graph.clear() calls since app.clean() now handles it
- Also should fix any leaks associated with `onNodeRemoved` reliance (minimap, slot syncing, etc.)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5407-fix-onNodeRemoved-not-called-when-loading-new-graph-and-tearing-down-previous-2666d73d365081fd927cf7830482fbc4) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -1765,6 +1765,12 @@ export class ComfyApp {
     executionStore.lastExecutionError = null
 
     useDomWidgetStore().clear()
+
+    // Subgraph does not properly implement `clear` and the parent class's
+    // (`LGraph`) `clear` breaks the subgraph structure.
+    if (this.graph && !this.canvas.subgraph) {
+      this.graph.clear()
+    }
   }
 
   clientPosToCanvasPos(pos: Vector2): Vector2 {
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -1058,6 +1058,8 @@ export class ComfyApp {
       checkForRerouteMigration = false
     } = {}
   ) {
+    useWorkflowService().beforeLoadNewGraph()
+
     if (clean !== false) {
       this.clean()
     }
@@ -1093,7 +1095,6 @@ export class ComfyApp {
         severity: 'warn'
       })
     }
-    useWorkflowService().beforeLoadNewGraph()
     useSubgraphService().loadSubgraphs(graphData)
 
     const missingNodeTypes: MissingNodeType[] = []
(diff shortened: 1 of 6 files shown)


benceruleanlu (dismissed):

christian-byrne:
Changed `app.clean()` to now encapsulate the call to `graph.clear()`, which removes all nodes from the graph. But the minimap relies on beforeLoadNewGraph() to capture a thumbnail of the canvas we are departing from. 

We can move the thumbnail capture to before the graph is cleared -- i.e., capture the thumbnail before calling app.loadGraphData().

This does call into question the semantics of beforeLoadNewGraph and whether it's widely expected to be able to use this hook to get state of the graph being departed from.

christian-byrne:
Yes, `beforeLoadNewGraph` should definitely be called before clearing/cleaning, that makes given the current uses-cases and the function documentation

https://github.com/Comfy-Org/ComfyUI_frontend/blob/88809c700655e6f3f3d073b2464f40342347d3ff/src/services/workflowService.ts#L280-L291

And actually, it's not used by any custom nodes ([query](https://cs.comfy.org/search?q=context:global+beforeLoadNewGraph&patternType=keyword&sm=0)), so doing some slight refactoring to make the calls more consistent seems allowable.

src/scripts/app.ts
```diff
@@ -1765,6 +1765,12 @@ export class ComfyApp {
     executionStore.lastExecutionError = null
 
     useDomWidgetStore().clear()
+
+    // Subgraph does not properly implement `clear` and the parent class's
+    // (`LGraph`) `clear` breaks the subgraph structure.
```
DrJKL, line 1771: This doesn't sound very safe...

DrJKL (dismissed):

src/scripts/app.ts
christian-byrne: danger is the spice of life

PR #5400, opened 2025-09-06
[refactor] Use getSlotPosition for Vue nodes in link rendering

benceruleanlu:
## Summary

Replace direct node position calls with getSlotPosition utility for Vue nodes in link rendering to ensure consistent slot positioning.

## Changes

- **What**: Updated link rendering in LGraphCanvas to use `getSlotPosition` utility when Vue nodes mode is enabled
- **Breaking**: None
- **Dependencies**: None

## Review Focus

The changes ensure that slot positions are calculated consistently across the canvas rendering system when Vue nodes are enabled. The legacy behavior is preserved when Vue nodes mode is disabled.

https://github.com/Comfy-Org/ComfyUI_frontend/pull/5400 ←https://github.com/Comfy-Org/ComfyUI_frontend/pull/5411

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5400-refactor-Use-getSlotPosition-for-Vue-nodes-in-link-rendering-2666d73d365081609cedd718c95e7622) by [Unito](https://www.unito.io)

PR #5411, opened 2025-09-07
Fix slot position offset in Vue node rendering

benceruleanlu:
## Summary

Fix links being rendered offset from their slots by normalizing slot positions to canvas-relative coordinates before coordinate conversion.

## Problem

When measuring slot positions using `getBoundingClientRect()`, the coordinates were relative to the viewport. This caused misalignment when the canvas container wasn't positioned at (0,0) in the viewport (e.g., when sidebars or panels were present).

## Solution

- Add `canvasRectCache.ts` to track the canvas container's position using VueUse's `useElementBounding`
- Normalize slot coordinates to be relative to the canvas container before converting to canvas space
- Replace window resize listeners with a canvas rect change subscription that handles both resize and layout changes

## Changes

- **Added**: `canvasRectCache.ts` - Tracks canvas container position and notifies subscribers of changes
- **Modified**: `useDomSlotRegistration.ts` - Normalize slot positions relative to canvas container

## Review Focus

- Canvas-relative coordinate calculation in `useDomSlotRegistration.ts`
- Event subscription pattern for handling canvas rect changes

https://github.com/Comfy-Org/ComfyUI_frontend/pull/5400 ←https://github.com/Comfy-Org/ComfyUI_frontend/pull/5411

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5411-Fix-slot-position-offset-in-Vue-node-rendering-2676d73d36508145b7a3d4d9d1dbc60f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -6,6 +6,7 @@ import {
   type LinkRenderContext,
   LitegraphLinkAdapter
 } from '@/renderer/core/canvas/litegraph/litegraphLinkAdapter'
+import { getSlotPosition } from '@/renderer/core/canvas/litegraph/slotCalculations'
 import { layoutStore } from '@/renderer/core/layout/store/layoutStore'
 
 import { CanvasPointer } from './CanvasPointer'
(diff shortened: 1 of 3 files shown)


benceruleanlu:
knip --cache not in lint staged?

christian-byrne:
> knip --cache not in lint staged?



https://github.com/Comfy-Org/ComfyUI_frontend/issues/5160

christian-byrne:
This looks good but I'm wondering if we should use ResizeObserver then tag all slots and vue nodes with data-id keys and query all size/pos in a single sweep after layout is already computed (at most once per full frame cycle -- hopefully skipping when no reflow occurred). This is alreay something that is needed so we can update the layout tree with the correct width/height of nodes (currently only x/y is accurate).

benceruleanlu:
the failing test is interesting, not sure if I should look into that to fix or if we should update expectations

that is

if this pr even goes in

benceruleanlu:
failing test to be fixed by https://github.com/Comfy-Org/ComfyUI_frontend/pull/5447

benceruleanlu:
Replaced by #5460

PR #5421, opened 2025-09-07
Implement selection state management in Vue Nodes

christian-byrne:
## Summary

Lets LGraphCanvas maintain ownership over selection state. Fixes selection state in Vue Nodes to match existing behavior/UX.

## Changes

- **What**: Watch changes to selected items in canvas state, provide a set of selected nodes to children. In LGraphNode.vue, watch the selected items and update selection state of UI.

Key choices:

* `LGraphCanvas` already exposes a reactive state of selected items. It’s one of the few parts of litegraph that transitioned to reactivity before this project, and it’s very ergonomic—showing why we should extend reactivity across the codebase
* Centralized Vue Node event handler needs to update `LGraphCanvas`'s selection state instead of maintaining internally
* `LGraphNode.vue` needs to watch selection state → updates UI (border)
* Instead of having each node component watch and filter the `selected` list by ID, make watcher live in the parent (`GraphCanvas.vue`).
* Use `provide`/`inject` instead of a prop → avoid full node list re-renders
* Store selected IDs in `Set` rather than `Array` → only need O(1) membership check in each node component when selection changes
* Future: no need to store selection state in layout tree now; later can attach actor + and use CRDT but selection/focus/hover is more of an ephemeral client-bound state

https://github.com/user-attachments/assets/1d37a2b1-c829-4962-aa28-4cb78c3e082d

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5421-Implement-selection-state-management-in-Vue-Nodes-2676d73d365081529d04d40c8b75be37) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -44,7 +44,6 @@
       :node-data="nodeData"
       :position="nodePositions.get(nodeData.id)"
       :size="nodeSizes.get(nodeData.id)"
-      :selected="nodeData.selected"
       :readonly="false"
       :executing="executionStore.executingNodeId === nodeData.id"
       :error="
@@ -79,6 +78,7 @@ import {
   computed,
   onMounted,
   onUnmounted,
+  provide,
   ref,
   shallowRef,
   watch,
@@ -112,6 +112,7 @@ import { useWorkflowPersistence } from '@/composables/useWorkflowPersistence'
 import { CORE_SETTINGS } from '@/constants/coreSettings'
 import { i18n, t } from '@/i18n'
 import type { LGraphNode } from '@/lib/litegraph/src/litegraph'
+import { SelectedNodeIdsKey } from '@/renderer/core/canvas/injectionKeys'
 import TransformPane from '@/renderer/core/layout/TransformPane.vue'
 import MiniMap from '@/renderer/extensions/minimap/MiniMap.vue'
 import VueGraphNode from '@/renderer/extensions/vueNodes/components/LGraphNode.vue'
@@ -188,6 +189,21 @@ const handleTransformUpdate = () => {
 const handleNodeSelect = nodeEventHandlers.handleNodeSelect
 const handleNodeCollapse = nodeEventHandlers.handleNodeCollapse
 const handleNodeTitleUpdate = nodeEventHandlers.handleNodeTitleUpdate
+
+// Provide selection state to all Vue nodes
+const selectedNodeIds = ref(new Set<string>())
+provide(SelectedNodeIdsKey, selectedNodeIds)
+watch(
+  () => canvasStore.selectedItems,
+  (newSelectedItems) => {
+    selectedNodeIds.value = new Set(
+      newSelectedItems
+        .filter((item) => item.id !== undefined)
+        .map((item) => String(item.id))
+    )
+  },
+  { immediate: true }
+)
 
 watchEffect(() => {
   nodeDefStore.showDeprecated = settingStore.get('Comfy.Node.ShowDeprecated')
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -191,19 +191,15 @@ const handleNodeCollapse = nodeEventHandlers.handleNodeCollapse
 const handleNodeTitleUpdate = nodeEventHandlers.handleNodeTitleUpdate
 
 // Provide selection state to all Vue nodes
-const selectedNodeIds = ref(new Set<string>())
-provide(SelectedNodeIdsKey, selectedNodeIds)
-watch(
-  () => canvasStore.selectedItems,
-  (newSelectedItems) => {
-    selectedNodeIds.value = new Set(
-      newSelectedItems
+const selectedNodeIds = computed(
+  () =>
+    new Set(
+      canvasStore.selectedItems
         .filter((item) => item.id !== undefined)
         .map((item) => String(item.id))
     )
-  },
-  { immediate: true }
 )
+provide(SelectedNodeIdsKey, selectedNodeIds)
 
 watchEffect(() => {
   nodeDefStore.showDeprecated = settingStore.get('Comfy.Node.ShowDeprecated')
--- a/src/composables/graph/useNodeEventHandlers.ts
+++ b/src/composables/graph/useNodeEventHandlers.ts
@@ -33,12 +33,20 @@ export function useNodeEventHandlers(nodeManager: Ref<NodeManager | null>) {
     const node = nodeManager.value.getNode(nodeData.id)
     if (!node) return
 
-    // Handle multi-select with Ctrl/Cmd key
-    if (!event.ctrlKey && !event.metaKey) {
-      canvasStore.canvas.deselectAllNodes()
-    }
+    const isMultiSelect = event.ctrlKey || event.metaKey
 
-    canvasStore.canvas.selectNode(node)
+    if (isMultiSelect) {
+      // Ctrl/Cmd+click -> toggle selection
+      if (node.selected) {
+        canvasStore.canvas.deselect(node)
+      } else {
+        canvasStore.canvas.select(node)
+      }
+    } else {
+      // Regular click -> single select
+      canvasStore.canvas.deselectAll()
+      canvasStore.canvas.select(node)
+    }
 
     // Bring node to front when clicked (similar to LiteGraph behavior)
     // Skip if node is pinned to avoid unwanted movement
@@ -46,9 +54,6 @@ export function useNodeEventHandlers(nodeManager: Ref<NodeManager | null>) {
       layoutMutations.setSource(LayoutSource.Vue)
       layoutMutations.bringNodeToFront(nodeData.id)
     }
-
-    // Ensure node selection state is set
-    node.selected = true
 
     // Update canvas selection tracking
     canvasStore.updateSelectedItems()
--- a/src/renderer/core/canvas/injectionKeys.ts
+++ b/src/renderer/core/canvas/injectionKeys.ts
@@ -1,0 +1,8 @@
+import type { InjectionKey, Ref } from 'vue'
+
+/**
+ * Injection key for providing selected node IDs to Vue node components.
+ * Contains a reactive Set of selected node IDs (as strings).
+ */
+export const SelectedNodeIdsKey: InjectionKey<Ref<Set<string>>> =
+  Symbol('selectedNodeIds')
--- a/src/renderer/extensions/vueNodes/components/LGraphNode.vue
+++ b/src/renderer/extensions/vueNodes/components/LGraphNode.vue
@@ -154,7 +154,12 @@ const emit = defineEmits<{
 }>()
 
 // Inject selection state from parent
-const selectedNodeIds = inject(SelectedNodeIdsKey, ref(new Set<string>()))
+const selectedNodeIds = inject(SelectedNodeIdsKey)
+if (!selectedNodeIds) {
+  throw new Error(
+    'SelectedNodeIds not provided - LGraphNode must be used within a component that provides selection state'
+  )
+}
 
 // Computed selection state - only this node re-evaluates when its selection changes
 const isSelected = computed(() => {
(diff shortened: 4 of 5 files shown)


src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -10,10 +10,13 @@
...
-        'outline outline-transparent outline-2 hover:outline-black dark-theme:hover:outline-white',
+        'outline outline-transparent outline-2',
         {
-          'border-blue-500 ring-2 ring-blue-300': selected,
-          'border-[#e1ded5] dark-theme:border-[#292A30]': !selected,
+          'outline-black dark-theme:outline-white': isSelected
+        },
+        {
+          'border-blue-500 ring-2 ring-blue-300': isSelected,
+          'border-[#e1ded5] dark-theme:border-[#292A30]': !isSelected,
```
viva-jinyi, line 19: It can't be replaces with tailwind color variable?

tests-ui/tests/composables/graph/useNodeEventHandlers.test.ts
```diff
@@ -0,0 +1,292 @@
...
+  selectedItems: Positionable[]
+  updateSelectedItems: ReturnType<typeof vi.fn>
+}
+
+interface MockLayoutMutations {
+  setSource: ReturnType<typeof vi.fn>
+  bringNodeToFront: ReturnType<typeof vi.fn>
+}
+
+describe('useNodeEventHandlers', () => {
```
viva-jinyi, line 41: I think it should be moved to test code

viva-jinyi (dismissed):

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: Sure let me try

src/components/graph/GraphCanvas.vue
```diff
@@ -189,6 +190,21 @@ const handleNodeSelect = nodeEventHandlers.handleNodeSelect
 const handleNodeCollapse = nodeEventHandlers.handleNodeCollapse
 const handleNodeTitleUpdate = nodeEventHandlers.handleNodeTitleUpdate
 
+// Provide selection state to all Vue nodes
+const selectedNodeIds = ref(new Set<string>())
+provide(SelectedNodeIdsKey, selectedNodeIds)
+watch(
+  () => canvasStore.selectedItems,
+  (newSelectedItems) => {
+    selectedNodeIds.value = new Set(
```
DrJKL: Is this a computed with extra steps?

src/composables/graph/useNodeEventHandlers.ts
```diff
@@ -33,23 +33,28 @@ export function useNodeEventHandlers(nodeManager: Ref<NodeManager | null>) {
...
+      if (node.selected) {
+        canvasStore.canvas.deselect(node)
+      } else {
+        canvasStore.canvas.select(node)
+      }
+    } else {
+      // Regular click -> single select
+      canvasStore.canvas.deselectAll()
+      canvasStore.canvas.select(node)
     }
```
DrJKL, line 49: I don't have a firm suggestion, but this looks like it could be reorganized a bit for clarity.
Like, first maybe deselect X, then decide whether to re-select or add to selection Y.

src/renderer/core/canvas/injectionKeys.ts
```diff
@@ -0,0 +1,8 @@
+import type { InjectionKey, Ref } from 'vue'
+
+/**
+ * Injection key for providing selected node IDs to Vue node components.
+ * Contains a reactive Set of selected node IDs (as strings).
```
DrJKL, line 5: Want to type them as `NodeId` since we have that type alias already anyway?

DrJKL: (Though I still don't like type aliases for things like that unless they include a brand)

src/renderer/extensions/vueNodes/components/LGraphNode.vue
DrJKL: 😁

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -150,6 +153,14 @@ const emit = defineEmits<{
   'update:title': [nodeId: string, newTitle: string]
 }>()
 
+// Inject selection state from parent
+const selectedNodeIds = inject(SelectedNodeIdsKey, ref(new Set<string>()))
```
DrJKL: Do we want a default here? What would it mean to a node to have selection IDs that are scoped to the node alone?

tests-ui/tests/composables/graph/useNodeEventHandlers.test.ts
```diff
@@ -0,0 +1,292 @@
...
+
+vi.mock('@/stores/graphStore')
+vi.mock('@/renderer/core/layout/operations/layoutMutations')
+
+interface MockCanvas {
+  select: ReturnType<typeof vi.fn>
+  deselect: ReturnType<typeof vi.fn>
+  deselectAll: ReturnType<typeof vi.fn>
+  updateSelectedItems: ReturnType<typeof vi.fn>
+}
```
DrJKL: This feels a little icky to me.
Isn't this why we'd normally use `vi.mocked`?

If we need something like this, how would you feel about a utility type that maps function properties from a provided type to `vi.fn()`, that way we know the keys are otherwise aligned.

DrJKL: We could also setup some automocks for these, if we're going to need similar mocks in different composables' test suites.

DrJKL: But normally, if there needs to be an explicit interface, I'd want one that's external to the implementation and mocks that each of them is expected to satisfy.

tests-ui/tests/composables/graph/useNodeEventHandlers.test.ts
```diff
@@ -0,0 +1,292 @@
...
+      bringNodeToFront: vi.fn()
+    }
+
+    // Setup module mocks
+    const { useCanvasStore } = await import('@/stores/graphStore')
+    const { useLayoutMutations } = await import(
+      '@/renderer/core/layout/operations/layoutMutations'
+    )
+
+    // @ts-expect-error - Test mocks only need minimal interface, full Pinia store type too complex
```
DrJKL: Hrmmmmm....

DrJKL: Does `{ partial: true }` satisfy the compiler?

tests-ui/tests/composables/graph/useNodeEventHandlers.test.ts
```diff
@@ -0,0 +1,292 @@
...
+      })
+
+      const nodeData: VueNodeData = {
+        id: 'node-1',
+        title: 'Test Node',
+        type: 'test',
+        mode: 0,
+        selected: false,
+        executing: false
+      }
```
DrJKL: I think this is another thing that we can extract for clarity, maybe.

DrJKL (dismissed):
I'm so excited to see some Dependency Injection 💉

src/renderer/core/canvas/injectionKeys.ts
christian-byrne: The `NodeId` type alias is `string | number`

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: I'll do it in a followup PR so that I can add all the tokens from Figma

src/components/graph/GraphCanvas.vue
christian-byrne: Great catch! We've come full circule from previous discussion.

src/composables/graph/useNodeEventHandlers.ts
christian-byrne: I think I understand what you mean, let me adjust the flow.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: Great point

src/composables/graph/useNodeEventHandlers.ts
christian-byrne: Actually, not really sure how to improve it

tests-ui/tests/composables/graph/useNodeEventHandlers.test.ts
christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5421/commits/7b4afea15ca928073a42aab2b1b981b71ef960b7

src/components/graph/GraphCanvas.vue
christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5421/commits/81dc09033d3881eb93d5285d49958f9b5b3d7cc8

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5421/commits/b43c668a3dc1c14697841b5df048360132a9d24f

tests-ui/tests/composables/graph/useNodeEventHandlers.test.ts
christian-byrne: Fixed indirectly in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5421/commits/7b4afea15ca928073a42aab2b1b981b71ef960b7

tests-ui/tests/composables/graph/useNodeEventHandlers.test.ts
christian-byrne: Switched to `vi.mockObject` https://github.com/Comfy-Org/ComfyUI_frontend/pull/5421/commits/7b4afea15ca928073a42aab2b1b981b71ef960b7

tests-ui/tests/composables/graph/useNodeEventHandlers.test.ts
christian-byrne: Extracted out in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5421/commits/cbc6833c3a31595bb52540f096b6a88cd5f88088

src/renderer/core/canvas/injectionKeys.ts
DrJKL: And we're eventually going to have only string?

tests-ui/tests/composables/graph/useNodeEventHandlers.test.ts
```diff
@@ -0,0 +1,201 @@
...
+
+vi.mock('@/stores/graphStore')
+vi.mock('@/renderer/core/layout/operations/layoutMutations')
+
+describe('useNodeEventHandlers', () => {
+  let mockCanvas: any
+  let mockNode: any
+  let mockNodeManager: any
+  let mockCanvasStore: any
+  let mockLayoutMutations: any
```
DrJKL: 😭

DrJKL: Can they be a mocked or partial of the real classes?

christian-byrne: Importing the actual classes is really hard in the test env. But Claude did something awful in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5421/commits/7b4afea15ca928073a42aab2b1b981b71ef960b7

christian-byrne: I can use type assertion at a higher level though like

```typescript
    vi.mocked(useCanvasStore).mockReturnValue(
      mockCanvasStore as ReturnType<typeof useCanvasStore>
    )
```

christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5421/commits/868fb1a788f78587621dad0ee55d4be6058cde7a. Going forward we should separate folders by domain then add index.ts for all subdomains with interfaces. Then the real contracts can be used in tests very easily without all this effort.

src/renderer/core/canvas/injectionKeys.ts
christian-byrne: For this value, I convert to it to String

PR #5427, opened 2025-09-08
Switch to max 4 threads in eslint script

christian-byrne:
## Summary

Changes eslint to not use more than 4 threads. The current `concurrency=auto` uses half available cores which is causing my 28 core PC to crash or become otherwise unusable.

If this is only happening for me, maybe there's a way to make it configurable per developer?


benceruleanlu:
it works well for me on 16/32

benceruleanlu (approved):

christian-byrne:
Actually mine is breaking even on 4

benceruleanlu (commented):
sounds like a sign!

DrJKL (approved):
4 seems reasonable.
I also opened an issue to try to address speed on the typescript side.

benceruleanlu (changes requested):
MB, didn't pnpm i

benceruleanlu (approved):

benceruleanlu:
something interesting, for whatever reason, this seemingly uses a lot more resources than needed on WSL, and results in a crash actually when ran manually and if not cancelled quickly

result of pnpm lint:fix

image

christian-byrne:
closed in favor of https://github.com/Comfy-Org/ComfyUI_frontend/pull/5448

PR #5431, opened 2025-09-08
Fix bypass nodes slot check only works for precise types

webfiltered:
## Summary

- Follow-up on #4864
- Fixes execution when bypassed nodes are directly connected to slots with non-exact type matches (e.g. list of types)

## Changes

- Resolves logic issue from last-minute refactor in original design (`find` => `findIndex`).

## Review Focus

Requires actual in-app testing with custom nodes before merge.  May or may not do what it says on the tin.

- Fixes #5403

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5431-Fix-bypass-nodes-slot-check-only-works-for-precise-types-2686d73d365081cba0f1d172b2a9b5da) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/subgraph/ExecutableNodeDTO.ts
+++ b/src/lib/litegraph/src/subgraph/ExecutableNodeDTO.ts
@@ -272,7 +272,7 @@ export class ExecutableNodeDTO implements ExecutableLGraphNode {
       const matchingIndex = this.#getBypassSlotIndex(slot, type)
 
       // No input types match
-      if (matchingIndex === undefined) {
+      if (matchingIndex === -1) {
         console.debug(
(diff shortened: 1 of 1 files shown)


AustinMroz:
The fix is good. Here's a minimal, core only reproduction.
image

The indexing performed prior to the reported error uses `Array.at` which supports negative indexes. Consequentially, the error can only occur on a node which has no inputs. Virtual nodes generally don't add inputs corresponding to widgets. Since bypassing a subgraph currently also bypasses all inner nodes, any virtual nodes inside a bypassed subgraph (like primitives or a set node) will trigger this bug.

(Though this also suggests that the nodes inside a bypassed subgraph are being explored. This should not be the case, but is a seperate case I've not found a repro for.)

AustinMroz (approved):

christian-byrne (approved):

PR #5433, opened 2025-09-08
Fix getSlotPosition readonly return value

webfiltered:
Fixes root cause of type issue in #5400, removing need for workaround.  `getSlotPosition` is _creating_ `Point`s; there is no need to constrain the return type to readonly.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5433-Fix-getSlotPosition-readonly-return-value-2686d73d365081ea940ad00ee5a81d6f) by [Unito](https://www.unito.io)


benceruleanlu (approved):
silly me!

PR #5439, opened 2025-09-08
lint: Fix missing defaults for Props with Defaults

DrJKL:
## Summary

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/5434

## Changes

- **What**: Use the `withDefaults` instead of the nullish coalescing for the PrimeVue props.

## Review Focus

Should I do the same for `modelValue`?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5439-lint-Fix-missing-defaults-for-Props-with-Defaults-2686d73d36508123bdc1c67e5c4c1c22) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/form/FormSelectButton.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/form/FormSelectButton.vue
@@ -57,7 +57,9 @@ interface Emits {
 }
 
 const props = withDefaults(defineProps<Props>(), {
-  disabled: false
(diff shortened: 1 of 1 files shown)


webfiltered (approved):
Only difference I can see is if `optionValue` was somehow `null`, it will now be `null`, rather than `'value'`.

But that's fixing a discard of error state, so it's all positive imo.

christian-byrne (approved):

PR #5441, opened 2025-09-08
Fix: Vue node/widget positioning and scroll issue

DrJKL:
## Summary

Building on @christian-byrne's fix for the widgets, also handles the scrollIntoView bug.

## Changes

- **What**: 
  - Attach widget popover elements to the parent instead of the body to scale correctly (initially)
  - Use [overflow: clip](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#clip) to handle the horizontal scrolling issue.

## Review Focus

Any other props that should be consolidated into the composable?
Any other overflow bugs that haven't been addressed yet?

(Just the scroll behavior)

### Before

https://github.com/user-attachments/assets/6bce75bb-54a0-4d4f-ba3d-ce6369181ac1

### After

https://github.com/user-attachments/assets/ddf2ba82-0920-4300-8d3a-653bdf345750

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5441-Fix-Vue-node-widget-positioning-and-scroll-issue-2686d73d36508193aa4ed18c22dba5da) by [Unito](https://www.unito.io)


AustinMroz:
The big evil edge case of canvas scrolling no longer occurs, but there's still a few nit edge cases where behaviour isn't ideal.
- Nodes extending off the left hand side of the screen do not bump the dropdown
- I'm seeing intermittent behaviour where the drop downs move away from the righthand screen edge faster than the node approaches. I have not found a consistent reproduction, but a reload usually fixes and swapping into vue mode is suspect.
gap

christian-byrne:
I think at least one of those behaviors is intended.

DrJKL:
Yeah, the offset on the right side feels correct, since it brings the overlay into a position where a user can interact with it.
The left side one, I'm not sure about, but it's probably worth addressing eventually.

christian-byrne (approved):
LGTM! One-line change, wow.

PR #5442, opened 2025-09-08
Add Centralized Vue Node Size/Pos Tracking

christian-byrne:
## Summary

Implemented centralized resize tracking system with O(1) element registration, single `ResizeObserver` traversal, and extensible ECS architecture to optimize Vue node bounds synchronization.

Here demonstrates that the sizes of the Vue Node elements are updated in the layout tree without indtroducing any new reflows and no thrashing. You can put a `console.count` in the ResizeObserver to confirm no redundant traversals but that is obvious.

https://github.com/user-attachments/assets/d2fd726a-cee4-41fe-b7cb-c59a71238fa7

## Changes

- **What**: Centralized resize registry with single ResizeObserver, optimized math functions for bounds computation, extensible tracking system
- **Breaking**: Moved composable from `src/composables/` to `src/renderer/extensions/vueNodes/composables/` (DDD architecture)
- **Dependencies**: None

## Review Focus

- **Single ResizeObserver**: Replaced per-component observers with centralized registry to minimize browser reflow
- **O(1) Element Access**: Map-based tracking configs eliminate linear searches across element types
- **Single Traversal**: Batch processing groups updates by type in one ResizeObserver callback
- **Optimized Math**: Union bounds computation uses incremental min/max instead of array op
- **ECS Pattern**: Element-Component-System design allows tracking different element types (nodes, widgets, slots) with shared infrastructure
- **Extensible Registry**: `trackingConfigs` Map enables adding new element types without core logic changes
- **Yjs Batch Optimization**: Removed race condition check in `batchUpdateNodeBounds` that could drop updates during concurrent transactions
- **Vue Lifecycle Integration**: Automatic cleanup on component unmount prevents memory leaks
- **Bounds Union**: `computeUnionBounds` uses streaming min/max calculation (O(n)) instead of coordinate array operations (O(n log n))
- **Transform Calculations**: Leverages ResizeObserver's native `contentBoxSize` API for accurate dimensions without layout thrashing


src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -34,17 +36,30 @@ export function useSelectionToolboxPosition(
...
+        if (layout) {
+          allBounds.push([
+            layout.bounds.x,
+            layout.bounds.y,
+            layout.bounds.width,
+            layout.bounds.height
+          ])
+        }
+      }
     }
```
DrJKL: Nit: could you do this with two `.filter()`s?

src/utils/mathUtil.ts
```diff
@@ -1,16 +1,19 @@
...
  * @returns The GCD of the two numbers.
  */
-const gcd = (a: number, b: number): number => {
+export const gcd = (a: number, b: number): number => {
   return b === 0 ? a : gcd(b, a % b)
 }
 
 /**
- * Finds the least common multiple (LCM) for two numbers.
+ * Finds the export least common multiple (LCM) for two numbers.
```
christian-byrne: what is this?

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -1425,6 +1425,33 @@ class LayoutStoreImpl implements LayoutStore {
   getStateAsUpdate(): Uint8Array {
     return Y.encodeStateAsUpdate(this.ydoc)
   }
+
+  /**
+   * Batch update node bounds using Yjs transaction for atomicity.
+   */
+  batchUpdateNodeBounds(
+    updates: Array<{ nodeId: NodeId; bounds: Bounds }>
```
DrJKL: It might be worth naming the type in here.

src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
```diff
@@ -0,0 +1,145 @@
...
+import type { Bounds, NodeId } from '@/renderer/core/layout/types'
+
+/**
+ * Configuration for different types of tracked elements
+ */
+interface ElementTrackingConfig {
+  /** Data attribute name (e.g., 'nodeId') */
+  dataAttribute: string
+  /** Handler for processing bounds updates */
+  updateHandler: (updates: Array<{ id: string; bounds: Bounds }>) => void
```
DrJKL: Why `id` here and `nodeId` above?

DrJKL: Could the types between them be combined?

tests-ui/tests/renderer/extensions/vueNodes/components/LGraphNode.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('should call resize tracking composable with node ID', async () => {
+    const { useVueElementTracking } = vi.mocked(
+      await import(
+        '@/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking'
+      )
+    )
```
DrJKL: The mocks are hoisted, so you should be able to just `import { useVueElementTracking } from '@/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking'` at the top and it'll already be the mocked value here.

src/utils/mathUtil.ts
DrJKL: Zealous autocomplete?

src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
christian-byrne: This should work in any vue component whether node, slot, widget, title, etc.

src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (!elementType || !elementId) continue
+
+    const { inlineSize: width, blockSize: height } = entry.contentBoxSize[0]
+    const rect = element.getBoundingClientRect()
+
+    const bounds: Bounds = {
+      x: rect.left,
+      y: rect.top,
+      width,
+      height
```
AustinMroz:
```suggestion
      height: height-LiteGraph.NODE_TITLE_HEIGHT
```

AustinMroz (changes requested):

AustinMroz (dismissed):
Aside from the missing LiteGraph import I've introduced, the code and its functionality both have my approval.

src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
```diff
@@ -0,0 +1,145 @@
...
+      height: height-LiteGraph.NODE_TITLE_HEIGHT
+    }
+
+    if (!updatesByType.has(elementType)) {
+      updatesByType.set(elementType, [])
+    }
+    const updates = updatesByType.get(elementType)
+    if (updates) {
+      updates.push({ id: elementId, bounds })
+    }
```
benceruleanlu: I’m probably missing something, but should bounds be converted into canvas space before pushing to the store? The store/spatial index operate in canvas coords.

benceruleanlu:
SO SORRY, accidentally hit merge through some git ops

src/utils/mathUtil.ts
christian-byrne: Fixed in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5442/commits/db5a68bc694b64f2cdc73ddb9e6bf2b90fdfd9e6

src/composables/canvas/useSelectionToolboxPosition.ts
christian-byrne: Switched to cleaner functional pattern in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5442/commits/c39cdaf36c2d60da42099e8d60808b40b36efe0d. This is not really a hot path so performance degradation is probably fine.

tests-ui/tests/renderer/extensions/vueNodes/components/LGraphNode.spec.ts
christian-byrne: Fixed in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5442/commits/d39bf36ce045615d0cfdd9092c7a5b7363d4a2b3

src/renderer/core/layout/store/layoutStore.ts
christian-byrne: Add interfaces in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5442/commits/ea93135dbbe420b5fb4924f982d2ca920dd6e041

```typescript
// Generic tracking collects ElementBoundsUpdate[]
const updates: ElementBoundsUpdate[] = [...]

// Converts to specific type for layout store
const nodeUpdates: NodeBoundsUpdate[] = updates.map(({ id, bounds }) => ({
  nodeId: id as NodeId,
  bounds
}))
```

src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
benceruleanlu: I was indeed missing something

PR #5445, opened 2025-09-09
Fix: In standard mode, don't stop when you hit a Vue node.

DrJKL:
## Summary

Forwards the scroll event to litegraph canvas.

Now using the existing `useCanvasInteractions`, thanks @benceruleanlu for pointing me in the right direction.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5445-Fix-In-standard-mode-don-t-stop-when-you-hit-a-Vue-node-2696d73d365081dfaa4dcc24e12e0951) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -36,6 +36,7 @@
     v-if="isVueNodesEnabled && comfyApp.canvas && comfyAppReady"
     :canvas="comfyApp.canvas"
     @transform-update="handleTransformUpdate"
+    @wheel.capture="forwardWheelEvent"
   >
     <!-- Vue nodes rendered based on graph nodes -->
     <VueGraphNode
@@ -453,4 +454,20 @@ onMounted(async () => {
 onUnmounted(() => {
   vueNodeLifecycle.cleanup()
 })
+
+function forwardWheelEvent(e: WheelEvent) {
+  // I don't love this, but it works.
+  const { clientX, clientY, deltaX, deltaY, ctrlKey, metaKey, shiftKey } = e
+  comfyApp.canvasEl.dispatchEvent(
+    new WheelEvent('wheel', {
+      clientX,
+      clientY,
+      deltaX,
+      deltaY,
+      ctrlKey,
+      metaKey,
+      shiftKey
+    })
+  )
+}
 </script>
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -36,7 +36,7 @@
     v-if="isVueNodesEnabled && comfyApp.canvas && comfyAppReady"
     :canvas="comfyApp.canvas"
     @transform-update="handleTransformUpdate"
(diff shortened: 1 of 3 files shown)


src/components/graph/GraphCanvas.vue
```diff
@@ -36,6 +36,7 @@
     v-if="isVueNodesEnabled && comfyApp.canvas && comfyAppReady"
     :canvas="comfyApp.canvas"
     @transform-update="handleTransformUpdate"
+    @wheel.capture="forwardWheelEvent"
```
AustinMroz: Might I humbly suggest
```suggestion
    @wheel.capture="comfyApp.canvas._mousewheel_callback"
```

DrJKL: I was trying to avoid deep coupling. The underscore implies private, right?

DrJKL: I think the long-term solution would be to have an interaction layer above the nodes that would selectively handle things like this and orchestrate behavior like zooming and panning.

AustinMroz: `#` is for private. I've always understood it to mean "internal". 

I see it as: Custom nodes are free to use underscore methods but
- They make have unexpected side effects
- They don't come with documentation
- The frontend team reserves the right to make breaking changes in the future.

I think it's much cleaner here than making a fake event, but I'd be happy to get a second/third opinion.

christian-byrne: Is the effect of forwarding to canvas element vs. the dedicated callback the same in all circumstances? Is there benefit to using the callback for the sake of subgraphs or multi-canvas rotating for example?

AustinMroz: I'm fairly certain the events are equivalent on the core side of things. I don't think multi-canvas ever happened. When swapping between subgraphs, the single canvas source changes the backing graph it is connected to.

Creating a fake event is a little sketchy since actual events bestow special permissions for security purposes (can play audio, can copy/paste, can open new tabs), but I don't foresee any of those being needed for mousewheel events even with custom nodes.

DrJKL: The benefit of using the event system is that the canvas can change what happens on scroll and we wouldn't need to remember to update the private method reference here.

benceruleanlu (dismissed):

benceruleanlu (commented):
actually, what do you think about handleWheel?

in useCanvasInteractions.ts

src/components/graph/GraphCanvas.vue
DrJKL: @AustinMroz What do you think about it now?

AustinMroz: In terms of organization and attention to detail it's great. I'm a little saddened by the need for event construction, but after noticing that the same is already being done in canvas.dispatchEvent, I am no longer concerned that doing so could break things.

It has my approval.

AustinMroz (dismissed):

christian-byrne (approved):

PR #5447, opened 2025-09-09
fix: Stabilize flaky link visibility toggle test

snomiao:
## Summary

This PR fixes a flaky test in the graph canvas menu that was intermittently failing in CI. The test was checking link visibility toggle functionality but wasn't properly waiting for the canvas to update after changing settings.

## Problem

The test `Can toggle link visibility` was flaky because:
1. It clicked the toggle button to change link visibility
2. Only waited for one frame with `nextFrame()`
3. Immediately took a screenshot for comparison

The issue was that changing the `Comfy.LinkRenderMode` setting triggers an async canvas redraw via `setDirty()`, but the actual redraw might not complete within a single frame.

## Solution

Added proper synchronization:
- **Wait for setting to propagate**: Use `waitForFunction` to wait until `canvas.links_render_mode` actually changes
- **Double frame wait**: Call `nextFrame()` twice to ensure the canvas redraw completes
- **Timeout protection**: Add 5-second timeout to prevent test hanging if something goes wrong

## Testing

The fix ensures that:
1. The setting change is confirmed at the canvas level (not just the store level)
2. The canvas has time to complete its redraw cycle
3. Screenshots are taken only after the visual state is stable

This should eliminate the race condition that was causing intermittent failures.

## Related Code

The link visibility toggle is handled by:
- Command: `Comfy.Canvas.ToggleLinkVisibility` in `useCoreCommands.ts`
- Setting watcher: `useLitegraphSettings.ts` watches `Comfy.LinkRenderMode` and updates canvas
- Canvas update: Calls `canvas.setDirty(false, true)` to trigger background redraw

Fixes the flaky test issue reported in CI.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5447-fix-Stabilize-flaky-link-visibility-toggle-test-2696d73d3650815b8d2be6817887bca8) by [Unito](https://www.unito.io)


snomiao:
problem prob solved by waiting for link_hidden status by page.evaluate but this impl is a bit too hacky.

christian-byrne:
This should be fixed on `main` now I think

snomiao:
close this PR as already fixed.

PR #5451, opened 2025-09-09
fix: tailwind v4 utilities layer

LittleSound:
After updating to Tailwind v4, the override order of CSS variables caused issues, leading to problems like input slots misalignment.  
This pull request attempts to fix that.

| Before (Broken) | After (Fixed) |
|-----------------|---------------|
| CleanShot 2025-09-09 at 13 31 55 | CleanShot 2025-09-09 at 13 32 13  |


DrJKL (dismissed):

DrJKL:
That's funny, I would have expected a lot more visual test failures. @christian-byrne was just talking about the need for coverage here. 🔮

christian-byrne:
Why are the tests failing?

LittleSound:
It seems like the snapshots aren't matching. how should I update them?

christian-byrne (approved):

PR #5218, opened 2025-08-26
Floating Selection Toolbox Improvements

Myestery:
This pull request introduces significant improvements to the selection toolbox in the graph UI, focusing on enhanced usability, test coverage, and maintainability. The main changes include a major refactor of the `SelectionToolbox.vue` component to support conditional rendering of buttons, the addition of a comprehensive test suite for the `BypassButton`, and the introduction of a new test file for the "More Options" submenu. Several UI and logic updates were made to individual toolbox buttons to improve their behavior and accessibility.

**Selection Toolbox Refactor and UI Improvements:**

* Refactored `SelectionToolbox.vue` to conditionally render toolbox buttons based on the current selection state, added new components (`InfoButton`, `MoreOptions`, `VerticalDivider`, `FrameNodes`), improved layout and styling, and removed the old `HelpButton` in favor of new options. Also ensures cleanup of "More Options" state on unmount. [[1]](diffhunk://#diff-05d80ee1e28e634dc758394ddf1bfaa8e5ec72a186a6ea2e2b6f5dfba867b264R11-R81) [[2]](diffhunk://#diff-05d80ee1e28e634dc758394ddf1bfaa8e5ec72a186a6ea2e2b6f5dfba867b264R101-R143)
[...]

PR #5454, opened 2025-09-09
Selection state simple buttons

Myestery:
This pull request introduces a new composable, `useSelectionState`, to centralize and simplify selection state logic across the graph selection toolbox and related UI components. It also refactors the selection toolbox buttons (`BypassButton`, `DeleteButton`, `RefreshSelectionButton`) to use this composable, improves their test coverage, and standardizes their appearance and accessibility.

**Centralization of selection state logic:**

- Added the `useSelectionState` composable to encapsulate all computed properties and helper functions related to selection state, such as determining if any nodes are selected, if a node is bypassed, or if the selection includes subgraphs or output nodes. This reduces duplication and makes it easier to maintain consistent selection logic across components.

**Refactoring of selection toolbox buttons:**

- Updated `BypassButton.vue`, `DeleteButton.vue`, and `RefreshSelectionButton.vue` to use the new `useSelectionState` composable for selection state, removing direct dependency on `canvasStore` and improving code clarity. [[1]](diffhunk://#diff-80b781aeba31712968ae157bb70194e4b72bc73430d1cca6a79d718d839daed6L3-R39) [[2]](diffhunk://#diff-d2cc939f5e6b17bd365edb4678e37e61ed46084cc0182ec903d7d8d66b345f5bR22-R30) [[3]](diffhunk://#diff-e6fd6252b370a75218abc1714d3a5d2c977148a9fa65e1c7a6769d6b1f79d5a3L4-R20)
- Standardized button appearance and accessibility by updating severity, adding `data-testid` attributes, and using consistent icon classes. [[1]](diffhunk://#diff-d2cc939f5e6b17bd365edb4678e37e61ed46084cc0182ec903d7d8d66b345f5bL8-R12) [[2]](diffhunk://#diff-e6fd6252b370a75218abc1714d3a5d2c977148a9fa65e1c7a6769d6b1f79d5a3L4-R20)

**Testing improvements:**

- Added a comprehensive unit test suite for `BypassButton.vue` to verify rendering, command execution, styling based on node state, and behavior with multiple selections.

**Localization:**

- Added a new localization key for "Refresh Node" in the English locale to support the updated tooltip in `RefreshSelectionButton.vue`.
part of https://github.com/Comfy-Org/ComfyUI_frontend/pull/5218


src/components/graph/selectionToolbox/RefreshSelectionButton.vue
```diff
@@ -1,17 +1,22 @@
 <template>
   <Button
     v-show="isRefreshable"
-    severity="info"
+    v-tooltip.top="t('g.refreshNode')"
+    severity="secondary"
     text
-    icon="pi pi-refresh"
+    data-testid="refresh-button"
```
arjansingh, line 7: love seeing test ids 💪

src/components/graph/selectionToolbox/DeleteButton.vue
```diff
@@ -17,14 +19,14 @@ import Button from 'primevue/button'
...
-import { useCanvasStore } from '@/stores/graphStore'
 
 const { t } = useI18n()
 const commandStore = useCommandStore()
-const canvasStore = useCanvasStore()
+const { selectedItems } = useSelectionState()
 
 const isDeletable = computed(() =>
-  canvasStore.selectedItems.some((x) => x.removable !== false)
+  selectedItems.value.some((x: any) => x.removable !== false)
```
arjansingh: type this please.

Suggestion for fix:
```typescript
  interface RemovableItem {
    removable?: boolean
  }

  const isRemovableItem = (item: unknown): item is RemovableItem =>
    item != null && typeof item === 'object' &&
    ('removable' in item || true) // All objects are potentially removable

  const isDeletable = computed(() =>
    selectedItems.value
      .filter(isRemovableItem)
      .some((x) => x.removable !== false)
  )
```

@DrJKL can you think of a better way?

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  let commandStore: ReturnType<typeof useCommandStore>
+
+  const i18n = createI18n({
+    legacy: false,
+    locale: 'en',
+    messages: {
+      en: {
+        selectionToolbox: {
+          bypassButton: {
+            tooltip: 'Toggle bypass mode'
```
arjansingh: nit: "Skip Selected Nodes"? Our users know what "bypass" means?

src/components/graph/selectionToolbox/DeleteButton.vue
arjansingh: for future PRs... if we're doing `isDeletable` a lot, we should consider putting it in the composable please.

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,132 @@
+import { computed } from 'vue'
```
arjansingh, line 2: we really should have unit tests here, i had claude suggest some cases, here are the ones i agree with:

```typescript
describe('useSelectionState', () => {

    describe('Selection Detection', () => {
      it('should return false when nothing selected', () => {
        // given: canvasStore.selectedItems = []
        // when: hasAnySelection.value
        // then: expect false
      })

      it('should return true when items selected', () => {
        // given: canvasStore.selectedItems = [node1, node2]
        // when: hasAnySelection.value
        // then: expect true
      })

      it('should distinguish single vs multiple selections', () => {
        // given: single node selected
        // then: hasSingleSelection = true, hasMultipleSelection = false
      })
    })

    describe('Node Type Filtering', () => {
      it('should pick only LGraphNodes from mixed selections', () => {
        // given: [graphNode, comment, connection]
        // when: selectedNodes.value
        // then: only LGraphNode returned
      })

      it('should detect subgraphs in selection', () => {
        // given: SubgraphNode selected
        // when: hasSubgraphs.value
        // then: true
      })
    })

    describe('Node State Computation', () => {
      it('should detect bypassed nodes', () => {
        // given: node with mode = LGraphEventMode.BYPASS
        // when: selectedNodesStates.value.bypassed
        // then: true
      })

      it('should detect pinned/collapsed states', () => {
        // given: mixed pinned/collapsed nodes
        // when: selectedNodesStates.value
        // then: correct state flags
      })

      it('should provide non-reactive state computation', () => {
        // given: performance-sensitive context
        // when: computeSelectionFlags()
        // then: fresh state without watchers
      })
    })

    describe('Data Integrity', () => {
      it('should handle missing removable property', () => {
        // given: node without removable field
        // when: checking deletability
        // then: treats undefined as deletable
      })

      it('should return default states for empty selection', () => {
        // given: no selection
        // when: selectedNodesStates.value
        // then: { collapsed: false, pinned: false, bypassed: false }
      })
    })

    describe('Help Integration', () => {
      it('should show help for single node', () => {
        // given: one node selected
        // when: showNodeHelp()
        // then: opens help sidebar
      })

      it('should ignore help request for multiple nodes', () => {
        // given: multiple nodes selected
        // when: showNodeHelp()
        // then: does nothing
      })
    })

    describe('Button Pattern Consistency', () => {
      it('should provide consistent state for all buttons', () => {
        // given: multiple components using useSelectionState
        // when: selection changes
        // then: all buttons react consistently
      })

      it('should support standardized deletability check', () => {
        // given: selectedItems from composable
        // when: compute isDeletable
        // then: consistent pattern across buttons
      })
    })
  })
```

arjansingh (changes requested):
Overall I like this PR:
1. fix the `any` type
2. get some unit tests for `useSelectionState.ts`

Everything else is a nit.

src/components/graph/selectionToolbox/BypassButton.spec.ts
Myestery: `Bypass/Unbypass Selected Nodes` is what it has been before
I'll update the test tooltip too

src/components/graph/selectionToolbox/DeleteButton.vue
DrJKL: I'd probably type `selectedItems` within `useSelectionState`

DrJKL: Looks like they're already `Positionable`

Myestery: I'm going with the deletable suggestion

DrJKL: Would this work?
```suggestion
  selectedItems.value.some((x) => !x.removable)
```

DrJKL: Separate issue here. This means we show the delete button if any selected node is removable. If a user hits it and nodes remain, wouldn't that be surprising?

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  mode: LGraphEventMode.ALWAYS
+}
+
+vi.mock('@/utils/litegraphUtil', () => ({
+  isLGraphNode: vi.fn(() => true)
+}))
+
+describe('BypassButton', () => {
+  let canvasStore: ReturnType<typeof useCanvasStore>
+  let commandStore: ReturnType<typeof useCommandStore>
```
DrJKL: Could you instantiate these within the tests themselves?

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+        directives: { tooltip: Tooltip },
+        stubs: {
+          'i-lucide:ban': true
+        }
+      }
+    })
+  }
+
+  it('should render bypass button', () => {
+    canvasStore.selectedItems = [mockLGraphNode] as any
```
DrJKL: Can you type the mock node as Positionable? Then you wouldn't need any `any`.

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    const wrapper = mountComponent()
+    const button = wrapper.find('button')
+
+    expect(button.classes()).not.toContain(
+      'dark-theme:[&:not(:active)]:!bg-[#262729]'
+    )
+  })
+
+  it('should show bypassed styling when node is bypassed', async () => {
```
DrJKL: The name hints that it's checking the style, but I don't se any style assertions.

src/components/graph/selectionToolbox/DeleteButton.vue
arjansingh: even better

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    await wrapper.find('button').trigger('click')
+    await wrapper.vm.$nextTick()
+
+    const button = wrapper.find('button')
+    expect(button.exists()).toBe(true)
+  })
+
+  it('should handle multiple selected items', () => {
+    vi.spyOn(commandStore, 'execute').mockResolvedValue()
+    canvasStore.selectedItems = [mockLGraphNode, mockLGraphNode] as any
```
DrJKL: I would expect issues if the selection state contained multiple references to the same node.

src/components/graph/selectionToolbox/BypassButton.vue
```diff
@@ -1,31 +1,40 @@
...
 import { useCommandStore } from '@/stores/commandStore'
-import { useCanvasStore } from '@/stores/graphStore'
 
 const { t } = useI18n()
 const commandStore = useCommandStore()
-const canvasStore = useCanvasStore()
+const { hasAnySelection } = useSelectionState()
+const isBypassed = ref(false)
+
+const byPass = async () => {
```
DrJKL: How do you feel about this instead?
```suggestion
const toggleBypass = async () => {
```

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,132 @@
...
+ * between selection toolbox, context menus, and other UI affordances.
+ */
+export function useSelectionState() {
+  const canvasStore = useCanvasStore()
+  const nodeDefStore = useNodeDefStore()
+  const sidebarTabStore = useSidebarTabStore()
+  const nodeHelpStore = useNodeHelpStore()
+  const { id: nodeLibraryTabId } = useNodeLibrarySidebarTab()
+
+  const selectedItems = computed(() => canvasStore.selectedItems)
```
DrJKL: You could use [storeToRefs](https://pinia.vuejs.org/api/pinia/functions/storeToRefs.html) for this

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  const selectedNodes = computed(() => {
+    return selectedItems.value.filter((i) => isLGraphNode(i)) as LGraphNode[]
```
DrJKL: Do you need the type assertion?

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,132 @@
...
+  )
+
+  const hasSubgraphs = computed(() =>
+    selectedItems.value.some((i) => i instanceof SubgraphNode)
+  )
+
+  const hasImageNode = computed(() => isSingleImageNode.value)
+  const hasOutputNodesSelected = computed(
+    () => filterOutputNodes(selectedNodes.value).length > 0
+  )
```
DrJKL: Are all of these currently being used?

If not, can we limit it to the set of predicates we need right now?

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,132 @@
...
+    nodes: LGraphNode[]
+  ): NodeSelectionState => {
+    if (!nodes.length)
+      return { collapsed: false, pinned: false, bypassed: false }
+    return {
+      collapsed: nodes.some((n) => n.flags?.collapsed),
+      pinned: nodes.some((n) => n.pinned),
+      bypassed: nodes.some((n) => n.mode === LGraphEventMode.BYPASS)
+    }
+  }
```
DrJKL: If this is a pure function, could you move it to module scope?

src/composables/graph/useSelectionState.ts
Myestery: yes they are being used in subsequent PRs.
most proceeding PRs will use some of them

src/components/graph/selectionToolbox/BypassButton.spec.ts
Myestery: this is a good candidate for beforeEach as seen below

src/components/graph/selectionToolbox/DeleteButton.vue
Myestery: This is resolved

src/components/graph/selectionToolbox/BypassButton.spec.ts
DrJKL: If you can avoid mutable global state in tests, it's usually safer. Especially if the stores are mocked.

src/components/graph/selectionToolbox/BypassButton.spec.ts
Myestery: Nice catch
I had to remove the styling from the bypass button cos that state was not reactive

Myestery: taking away the related tests

src/composables/graph/useSelectionState.ts
Myestery: Its only makes sense in this scope for now

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+
+class MockLGraphNode implements Positionable {
+  type = 'TestNode'
+  title = 'Test Node'
+  mode = LGraphEventMode.ALWAYS
+  id = 'node-1'
+  pos = new Float64Array([26, 186])
+  move = vi.fn()
+  snapToGrid = vi.fn()
+  boundingRect: [number, number, number, number] = [0, 0, 100, 100]
```
DrJKL: Do you need all of these properties?
Also, should the boundingRect align with pos?

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+          'Toggle Bypass Selected Nodes'
+      }
+    }
+  })
+
+  beforeEach(() => {
+    setActivePinia(createPinia())
+    // canvasStore = useCanvasStore()
+    // commandStore = useCommandStore()
+    // mockLGraphNode = new MockLGraphNode()
```
DrJKL:
```suggestion
```

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+        directives: { tooltip: Tooltip },
+        stubs: {
+          'i-lucide:ban': true
+        }
+      }
+    })
+  }
+
+  it('should render bypass button', () => {
+    const canvasStore = useCanvasStore()
```
DrJKL: I don't know what all is in there, but normally I'd expect the store modules to be mocked for a component test.

src/components/graph/selectionToolbox/BypassButton.spec.ts
DrJKL, line 1: Could you rename this to `BypassButton.test.ts`?
I usually reserve `.spec.` for Playwright tests and `.test.` for Vitest ones.

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    const mockLGraphNode = new MockLGraphNode()
+
+    canvasStore.selectedItems = [mockLGraphNode]
```
DrJKL: If you mock the store import, you can stub out the return for the standard case to elide it in most of the tests.

src/components/graph/selectionToolbox/BypassButton.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    )
+  })
+
+  it('should handle multiple selected items', () => {
+    const canvasStore = useCanvasStore()
+    const commandStore = useCommandStore()
+    const mockLGraphNode = new MockLGraphNode()
+
+    vi.spyOn(commandStore, 'execute').mockResolvedValue()
+    canvasStore.selectedItems = [mockLGraphNode, new MockLGraphNode()]
```
DrJKL: Wouldn't both of these nodes still have all the same properties and `id`?

src/components/graph/selectionToolbox/BypassButton.vue
```diff
@@ -1,31 +1,40 @@
...
     text
     data-testid="bypass-button"
-    @click="
-      () => commandStore.execute('Comfy.Canvas.ToggleSelectedNodes.Bypass')
-    "
+    :class="{
+      'hover:dark-theme:!bg-[#262729] hover:!bg-[#E7E6E6]': true,
+      'dark-theme:[&:not(:active)]:!bg-[#262729] [&:not(:active)]:!bg-[#E7E6E6]':
+        isBypassed
+    }"
```
DrJKL: Could you use the `cn()` utility here?

src/components/graph/selectionToolbox/BypassButton.vue
```diff
@@ -1,31 +1,40 @@
...
+    @click="toggleBypass"
   >
     <template #icon>
-      <i-game-icons:detour />
+      <i-lucide:ban class="w-4 h-4" />
```
DrJKL:
```suggestion
      <i-lucide:ban class="size-4" />
```

src/components/graph/selectionToolbox/DeleteButton.vue
```diff
@@ -5,26 +5,23 @@
       value: t('commands.Comfy_Canvas_DeleteSelectedItems.label'),
       showDelay: 1000
     }"
-    severity="danger"
+    severity="secondary"
     text
+    icon-class="w-4 h-4"
```
DrJKL:
```suggestion
    icon-class="size-4"
```

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,141 @@
...
+import { filterOutputNodes } from '@/utils/nodeFilterUtil'
+
+interface NodeSelectionState {
+  collapsed: boolean
+  pinned: boolean
+  bypassed: boolean
+}
+interface RemovableItem {
+  removable?: boolean
+}
```
DrJKL: This is already an aspect of Positionable.
I don't think you need the extra interface reference, but if so could you use that directly?

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,141 @@
...
+import { useNodeHelpStore } from '@/stores/workspace/nodeHelpStore'
+import { useSidebarTabStore } from '@/stores/workspace/sidebarTabStore'
+import { isImageNode, isLGraphNode } from '@/utils/litegraphUtil'
+import { filterOutputNodes } from '@/utils/nodeFilterUtil'
+
+interface NodeSelectionState {
+  collapsed: boolean
+  pinned: boolean
+  bypassed: boolean
+}
```
DrJKL: These properties seem orthogonal to whether a given node is selected to me.

src/composables/graph/useSelectionState.ts
DrJKL: Huh, do you know why it didn't work?

src/composables/graph/useSelectionState.ts
DrJKL: It can still be in this module, just not inside the composable function.

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,141 @@
...
+      sidebarTabStore.toggleSidebarTab(nodeLibraryTabId)
+      return
+    }
+
+    if (!isSidebarActive) sidebarTabStore.toggleSidebarTab(nodeLibraryTabId)
+    nodeHelpStore.openHelp(def)
+  }
+
+  const isRemovableItem = (item: unknown): item is RemovableItem =>
+    item != null && typeof item === 'object' && ('removable' in item || true)
```
DrJKL: I think this will be `true` for any `object`, right?

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  const isDeletable = computed(() =>
+    selectedItems.value
+      .filter(isRemovableItem)
+      .some((x) => x.removable !== false)
+  )
```
DrJKL: Same concern about the UX semantics as before. This implies that the selection is deletable in cases where it would only be partially removed.

src/composables/graph/useSelectionState.ts
```diff
@@ -0,0 +1,141 @@
...
+    hasMultipleSelection,
+    isSingleNode,
+    isSingleSubgraph,
+    isSingleImageNode,
+    hasSubgraphs,
+    hasImageNode,
+    hasOutputNodesSelected,
+    selectedNodesStates,
+    computeSelectionFlags,
+    isDeletable
```
DrJKL: Could you cherry-pick the refs/actions here that are used in this PR and re-add the others as they're needed?

tests-ui/tests/composables/graph/useSelectionState.spec.ts
```diff
@@ -0,0 +1,411 @@
...
+import { useSelectionState } from '@/composables/graph/useSelectionState'
+
+// Define constants before mocks
+const LGraphEventMode = {
+  ALWAYS: 0,
+  ON_EVENT: 1,
+  NEVER: 2,
+  ON_TRIGGER: 3,
+  BYPASS: 4
+}
```
DrJKL: Can you reference the pseudo-enum directly?

tests-ui/tests/composables/graph/useSelectionState.spec.ts
DrJKL, line 1: Same naming change here please 🙂

tests-ui/tests/composables/graph/useSelectionState.spec.ts
```diff
@@ -0,0 +1,411 @@
...
+    BYPASS: 4
+  },
+  LGraphNode: class {},
+  SubgraphNode: class {}
+}))
+
+// Mock stores
+const canvasStore = reactive({
+  selectedItems: [] as any[]
+})
```
DrJKL: Could you do a partial mock of the module to get the typing and avoid `any`?

DrJKL: Same below.

DrJKL (changes requested):
Could you add the true types in place of `any` in all the test cases?

I'm also a little unsure about the semantics in useSelectionState, I left comments there.

src/components/graph/selectionToolbox/BypassButton.spec.ts
Myestery: was done in a bid to fulfill the Positionable interface

DrJKL: Generally, if you can, it's better to mock only what you need to be able to test the behavior that you care about.

src/composables/graph/useSelectionState.ts
Myestery: I prefer to not do that for now
Hard keeping track already

src/components/graph/selectionToolbox/BypassButton.vue
Myestery: no real need, vue already handles it well

src/composables/graph/useSelectionState.ts
Myestery: the interface helps in other places to control display of some options

Myestery:
I've refactored most of the tests to rely heavily on mocking and removed some unnecessary tests

PR #5457, opened 2025-09-09
[test] Add Vue Node header component test

christian-byrne:
## Summary

Adds component test cases for the Vue Nodes Header component.

Changes test config to include components put outside of `src/components` folder while excluding yet-migrated litegraph tests.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5457-test-Add-Vue-Node-header-component-test-2696d73d365081ee8a23e800403b7147) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/NodeHeader.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    expect(input.exists()).toBe(false)
+  })
+
+  it('renders correct chevron icon based on collapsed prop', async () => {
+    const wrapper = mountHeader({ collapsed: false })
+    let icon = wrapper.get('i')
+    expect(icon.classes()).toContain('pi-chevron-down')
+
+    await wrapper.setProps({ collapsed: true })
+    icon = wrapper.get('i')
```
DrJKL: I was so close to having 0 comments!
Very minor nit: if this is targeting a separate element, I'd rather have two `const`s instead of a mutable `icon`.

vitest.config.ts
```diff
@@ -20,11 +20,19 @@ export default defineConfig({
...
-    }
+    },
+    exclude: [
+      '**/node_modules/**',
+      '**/dist/**',
+      '**/cypress/**',
+      '**/.{idea,git,cache}/**',
+      '**/{karma,rollup,webpack,vite,vitest,jest,ava,babel,nyc,cypress,tsup,build,eslint,prettier}.config.*',
+      'src/lib/litegraph/test/**'
+    ]
```
DrJKL: I can take a look at this later, but are you sure we need all/each of these?

DrJKL (dismissed):
More tests!

vitest.config.ts
christian-byrne: Oh I just used it from here https://vitest.dev/config/#exclude but it's not needed.

src/renderer/extensions/vueNodes/components/NodeHeader.spec.ts
christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5457/commits/5a282d57c774d1222a079b97263d9fc91bde26f4

vitest.config.ts
christian-byrne: Fixed in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5457/commits/b314c245b8dd8cd9d2a9ca6d4afb1d809569d00a

PR #5463, opened 2025-09-10
Refactor vue slot tracking

benceruleanlu:
## Summary
High-level refactor to standardize canvas-space layout updates for Vue nodes/slots, improve type safety around transform injection, and simplify/optimize ResizeObserver and slot tracking logic.

## Changes
- Typed `TransformState` InjectionKey and switch `provide/inject` sites to use it (no string keys, no ad-hoc assertions).
- Centralized slot element tracking (`useSlotElementTracking`): batched measurements, cached offsets, fewer watchers, and canvas-space persistence.
- Safer ResizeObserver sizing: prefer `contentBoxSize[0]` with `contentRect` fallback; removed `as any` cast.
- Cleaned up `dataset` access to avoid `any`; use typed `element.dataset[...]` for read/write/delete.
- Small readability updates: rename `cfg` → `config`, use get‑or‑init pattern when batching updates.
- Remove legacy modules superseded by the centralized tracker.

## Review Focus
- TransformPane provider + consumers now share a typed `TransformStateKey`.
- RO callback conversion: screen→canvas conversion via per-element context remains; now typed and simpler.
- Slot tracking: single RAF batch per node group; verify cached update path vs DOM read path.

Closes #5411

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5463-Update-slots-to-new-RO-system-26a6d73d365081aaa26cd1d98ccccb89) by [Unito](https://www.unito.io)


src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -72,7 +72,9 @@ export function useSelectionToolboxPosition(
 
     worldPosition.value = {
       x: unionBounds.x + unionBounds.width / 2,
-      y: unionBounds.y
+      // createBounds() applied a default padding of 10px
+      // so adjust Y to maintain visual consistency
+      y: unionBounds.y - 10
```
benceruleanlu: The selectiontoolbox would be too low without this padding, this is to maintain visual parity, createBounds() had this inbuilt default padding

src/renderer/core/layout/slots/useDomSlotRegistration.ts
benceruleanlu, line 1: replaced by useSlotElementTracking.ts

src/renderer/core/layout/injectionKeys.ts
benceruleanlu, line 1: just a type helper

christian-byrne (changes requested):
Can we rebase? 

For the naming with "measure", are we doing any measuring? Or is that leftover from when we used a static formula? Gbcr and such don't do measuring.

benceruleanlu:
fixed conflicts and renamed off of measure

although I still prefer measure

christian-byrne:
> although I still prefer measure

Why's that?

benceruleanlu:
> > although I still prefer measure
> 
> Why's that?

I thought gBCR was all about measuring

src/renderer/core/layout/injectionKeys.ts
```diff
@@ -0,0 +1,18 @@
+import type { InjectionKey } from 'vue'
+
+import type { Point } from '@/renderer/core/layout/types'
+
+interface TransformState {
```
christian-byrne, line 24: Can we add JSdoc comment for this interface?

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -0,0 +1,245 @@
...
+    const slotKey = getSlotKey(nodeId, index, isInput)
+    node.slots.delete(slotKey)
+    layoutStore.deleteSlotLayout(slotKey)
+
+    // If node has no more slots, clean up
+    if (node.slots.size === 0) {
+      if (node.stopWatch) node.stopWatch()
+      nodeRegistry.delete(nodeId)
+    }
+  })
```
christian-byrne: Should we cancel pending layout syncs in the unMounted or does it happen already indirectly?

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
christian-byrne, line 1: In this file we have global state shared across the app and a component hook. Should we separate them into two modules and make the global state use a pinia store?

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -0,0 +1,245 @@
...
+
+    // Remove this slot from registry and layout
+    const slotKey = getSlotKey(nodeId, index, isInput)
+    node.slots.delete(slotKey)
+    layoutStore.deleteSlotLayout(slotKey)
+
+    // If node has no more slots, clean up
+    if (node.slots.size === 0) {
+      if (node.stopWatch) node.stopWatch()
+      nodeRegistry.delete(nodeId)
```
christian-byrne: We should stop the watchers first before other things are done in the callback.

src/renderer/core/layout/dom/canvasRectCache.ts
christian-byrne, line 1: I think we can potentially remove this file and then use the conversion util here

https://github.com/Comfy-Org/ComfyUI_frontend/blob/1845708ddb7975dc26b759a7d9b935c261edd40e/src/composables/element/useCanvasPositionConversion.ts#L11-L31

The canvas should be exposing its offsets already without us needing to query the canvas element and observing it.

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -0,0 +1,245 @@
...
+type SlotEntry = {
+  el: HTMLElement
+  index: number
+  isInput: boolean
+  cachedOffset?: { x: number; y: number }
+}
+
+type NodeEntry = {
+  nodeId: string
+  screenToCanvas?: (p: Point) => Point
```
christian-byrne: Does the `screenToCanvas` need to be coupled to the `NodeEntry`s? Prior it was being provided from transform container right? But repeating across Map entries seems a bit strange and would imply the function is a method particular to that Node.

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -0,0 +1,245 @@
...
+    })
+  }
+
+  if (batch.length) layoutStore.batchUpdateSlotLayouts(batch)
+}
+
+export function useSlotElementTracking(options: {
+  nodeId: string
+  index: number
+  isInput: boolean
```
christian-byrne: What about just using the same type as the `type` field -- 'input' | 'output'

Later we are just doing

```typescript
entry.isInput ? 'input' : 'output'
```

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -0,0 +1,245 @@
...
+}) {
+  const { nodeId, index, isInput, element } = options
+
+  // Get transform utilities from TransformPane
+  const transformState = inject(TransformStateKey)
+
+  onMounted(async () => {
+    if (!nodeId) return
+    await nextTick()
+    const el = element.value
```
christian-byrne: According to [the docs on `onMounted`](https://vuejs.org/api/composition-api-lifecycle#onmounted):

A component is considered mounted after:

> - All of its synchronous child components have been mounted (does not include async components or components inside <Suspense> trees).
> - Its own DOM tree has been created and inserted into the parent container. Note it only guarantees that the component's DOM tree is in-document if the application's root container is also in-document. 


So this check should never be necessary in theory.

Instead of passing the element ref as a param we should follow pattern elsewhere

```typescript
    const element = getCurrentInstance()?.proxy?.$el
    if (!(element instanceof HTMLElement)) return
```

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+
+  if (batch.length) layoutStore.batchUpdateSlotLayouts(batch)
+}
+
+export function useSlotElementTracking(options: {
+  nodeId: string
+  index: number
+  isInput: boolean
+  element: Ref<HTMLElement | null>
```
christian-byrne, line 141: Does this need to be reactive?

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -0,0 +1,245 @@
...
+              newLayout.position.y !== oldLayout.position.y
+            const resized =
+              newLayout.size.width !== oldLayout.size.width ||
+              newLayout.size.height !== oldLayout.size.height
+
+            // Only update from cache on move-only changes.
+            // On resizes (or move+resize), let ResizeObserver resync slots from DOM accurately.
+            if (moved && !resized) {
+              updateNodeSlotsFromCache(nodeId)
+            }
```
christian-byrne: How about short-circuiting? After we compute `moved` or after we compute `resized`, there are scenarios where we can return early.

src/renderer/core/layout/utils/geometry.ts
```diff
@@ -0,0 +1,11 @@
+import type { Bounds, Point } from '@/renderer/core/layout/types'
+
+export function samePoint(a: Point, b: Point): boolean {
+  return a.x === b.x && a.y === b.y
+}
+
+export function sameBounds(a: Bounds, b: Bounds): boolean {
```
christian-byrne: nit: I think more common naming is like "isPointsEqual", "isEqual", "areEqual".

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -0,0 +1,245 @@
...
+      node = {
+        nodeId,
+        screenToCanvas: transformState?.screenToCanvas,
+        slots: new Map()
+      }
+      nodeRegistry.set(nodeId, node)
+      // Subscribe once per node to layout changes for fast cached updates
+      const nodeRef = layoutStore.getNodeLayoutRef(nodeId)
+      const stop = watch(
+        nodeRef,
```
christian-byrne: I think we only need to watch `nodeRef.position` or similar, rather than all properties.

christian-byrne:
> I thought gBCR was all about measuring

It's about getting a value from a layout tree. When the tree is not dirty, it's just a getter.

src/renderer/core/layout/dom/canvasRectCache.ts
benceruleanlu: good catch

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
benceruleanlu: are you sure? that would bring the guard before slots are removed, seemingly never satisfying the node.slots.size === 0 because it wasn't removed before

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
benceruleanlu: if it's gone it won't bother with the expensive checks by returning immediately

if (!node) return
if (!nodeLayout) return
for (const [slotKey, entry] of node.slots) {

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
benceruleanlu: it already short circuits deeper in, I think that's better than short circuiting every time in the call area

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
benceruleanlu: yes, as dynamic slots come in, we can't exactly work with a component root from getCurrentInstance()?.proxy?.$el

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
benceruleanlu: the check is there because despite the child component being mounted, it's not guaranteed that it triggers the watchEffect in time to update its value for onMounted, I think the microtask that is scheduled doesn't happen synchronously compared to onMounted

even so, this seems like a very small thing to worry about in my opinion, the check seems very cheap

as for why I'd rather not use $el:

> as dynamic slots come in, we can't exactly work with a component root from getCurrentInstance()?.proxy?.$el

we would need a lot more (brittle) code to work with a raw component, I don't see the benefit in doing that

src/composables/element/useCanvasPositionConversion.ts
```diff
@@ -1,6 +1,10 @@
 import { useElementBounding } from '@vueuse/core'
 
-import type { LGraphCanvas, Vector2 } from '@/lib/litegraph/src/litegraph'
+import type { LGraphCanvas, Point } from '@/lib/litegraph/src/litegraph'
+import { useCanvasStore } from '@/stores/graphStore'
+
+let sharedConverter: ReturnType<typeof useCanvasPositionConversion> | null =
```
christian-byrne, line 6: Should we just move it to the canvas store?

benceruleanlu: OK

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
benceruleanlu: on second thought, these changes seem to just add complexity for not much gain right now

maybe on a later refactor if this code changes?

benceruleanlu: on third thought, the pinia store sounds OK, I added it

christian-byrne (approved):
LGTM!

PR #5464, opened 2025-09-10
[feat] carve out path to call asset browser in combo widget

arjansingh:
## Summary

Add the forking logic to allow us to start hooking up the Asset Browsing experience.

## Changes

1. Add switching logic to allow the combo widget to have forked behavior for model nodes 
2. Create whitelist of eligible nodes that can get the forked experince.
3. The general logic is `AssetsAPI == enabled && Node in EligibilityList`

## Review Focus

Logic and tests.

With AssetsAPI setting enabled:
Screenshot 2025-09-10 at 7 51 32 PM
Screenshot 2025-09-10 at 7 49 59 PM

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/composables/useComboWidget.ts
+++ b/src/renderer/extensions/vueNodes/widgets/composables/useComboWidget.ts
@@ -1,8 +1,12 @@
 import { ref } from 'vue'
 
 import MultiSelectWidget from '@/components/graph/widgets/MultiSelectWidget.vue'
+import { t } from '@/i18n'
 import type { LGraphNode } from '@/lib/litegraph/src/litegraph'
-import type { IComboWidget } from '@/lib/litegraph/src/types/widgets'
+import type {
+  IBaseWidget,
+  IComboWidget
+} from '@/lib/litegraph/src/types/widgets'
 import { transformInputSpecV2ToV1 } from '@/schemas/nodeDef/migration'
 import {
   ComboInputSpec,
@@ -18,6 +22,8 @@ import {
   type ComfyWidgetConstructorV2,
   addValueControlWidgets
 } from '@/scripts/widgets'
+import { assetService } from '@/services/assetService'
+import { useSettingStore } from '@/stores/settingStore'
 
 import { useRemoteWidget } from './useRemoteWidget'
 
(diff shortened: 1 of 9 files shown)


viva-jinyi:
LGTM!

viva-jinyi (dismissed):

src/renderer/extensions/vueNodes/widgets/composables/useComboWidget.ts
```diff
@@ -48,7 +57,37 @@ const addMultiSelectWidget = (node: LGraphNode, inputSpec: ComboInputSpec) => {
...
 }
 
-const addComboWidget = (node: LGraphNode, inputSpec: ComboInputSpec) => {
+const addComboWidget = (
+  node: LGraphNode,
+  inputSpec: ComboInputSpec
+): IComboWidget => {
+  const settingStore = useSettingStore()
+  const isUsingAssetAPI = settingStore.get('Comfy.Assets.UseAssetAPI')
+  const isAssetBrowserEligible = assetService.isAssetBrowserEligible(
```
christian-byrne: We may want to check based on `node` param as well. We have the `modelToNodeStore` mapping we added.

Otherwise, this may be true for other nodes with these inputs names, including extensions.

src/services/assetService.ts
```diff
@@ -142,9 +142,31 @@ function createAssetService() {
       : []
   }
 
+  /**
+   * Widget spec names that are eligible for asset browser
+   */
+  type AssetBrowserEligibleSpec = 'ckpt_name' | 'lora_name' | 'vae_name'
```
christian-byrne: It feels like it should be a `Set`.

`isAssetBrowserEligible(specName)` is similar like `modelInputNames.has(inputName)`.

src/services/assetService.ts
```diff
@@ -142,9 +142,31 @@ function createAssetService() {
...
+
+  /**
+   * Checks if a widget input spec should use the asset browser
+   *
+   * @param specName - The input spec name (e.g., 'ckpt_name', 'lora_name')
+   * @returns true if this spec should use asset browser
+   */
+  function isAssetBrowserEligible(
+    specName: string
+  ): specName is AssetBrowserEligibleSpec {
```
christian-byrne: I think `inputName` makes more sense than `specName`, since it's the name of the input rather than the name of the spec.

src/renderer/extensions/vueNodes/widgets/composables/useComboWidget.ts
arjansingh: from exploring:
```
node.#nodeKey
'nodeDefs.CheckpointLoaderSimple'
node.type
''
```

is it okay to use `nodeKey`? `type` often comes back empty.

arjansingh: `node.comfyClass` this is what i'm gonna use.

christian-byrne (approved):
LGTM!

PR #5475, opened 2025-09-10
Add Asset Widget

AustinMroz:
Add a quick asset widget implementation based off the string widget.

This removes the buttons and the intermediate select_model popup from the previous COMBO implementation.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5475-Add-Asset-Widget-26a6d73d365081939bf0da3abf76d9f9) by [Unito](https://www.unito.io)


christian-byrne (commented):
This is good

AustinMroz:
There's a little bit of funkiness as this was a PR into @arjansingh's working branch. That branch has since diverged and been merged. 

Happy to rebase the changes here, but I would like to hear back from Arjan first.

I've also updated the original post with pictures for posterity.

arjansingh:
@AustinMroz thanks i'm looking at this right now. I went with @christian-byrne
s initial suggestion to [hard-code](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5464/files#diff-cf6d84525e3edf6994c9bd5270130943fbcca56f522f6f509ba1f51743e812b6R487-R498) the label.

That's probably a main source of the divergence.

arjansingh:
@AustinMroz Okay i looked this approach over. I actually like this a lot. I'll resolve conflicts and get this in a mergeable state then we can someone else from the team review. Thank you for the help here! This will be much better down the line.

arjansingh:
@christian-byrne @DrJKL I think this is in a good place for a review

src/lib/litegraph/src/widgets/AssetWidget.ts
```diff
@@ -0,0 +1,41 @@
...
+  override get _displayValue(): string {
+    return String(this.value) //FIXME: Resolve asset name
+  }
+
+  override drawWidget(
+    ctx: CanvasRenderingContext2D,
+    { width, showText = true }: DrawWidgetOptions
+  ) {
+    // Store original context attributes
+    const { fillStyle, strokeStyle, textAlign } = ctx
```
DrJKL, line 25: Why not use `.save()` and `.restore()`?

DrJKL (approved):
👍🏻 ❤️

christian-byrne (approved):

src/lib/litegraph/src/widgets/AssetWidget.ts
AustinMroz: This code matches the implementation in TextWidget.ts where the decision not to use save/restore tracks all the way back to comfy-org/litegraph.js#484.

I'm not sure why this was done, but I prefer not to invoke Chesterton's fence.

PR #5482, opened 2025-09-10
When toggling selected nodes, align state

AustinMroz:
Previously, when toggling the mode of multiple nodes, each node would have its state individually toggled. Now it sets the target mode if any node does not already match that mode and only disables if all already match.

BypassToggle

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5482-When-toggling-selected-nodes-align-state-26a6d73d3650818cab41c76f81d8505b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/canvas/useSelectedLiteGraphItems.ts
+++ b/src/composables/canvas/useSelectedLiteGraphItems.ts
@@ -123,12 +123,14 @@ export function useSelectedLiteGraphItems() {
     for (const i in selectedNodes) {
       selectedNodeArray.push(selectedNodes[i])
     }
+    const allNodesMatch = !selectedNodeArray.some(
+      (selectedNode) => selectedNode.mode !== mode
+    )
+    const newModeForSelectedNode = allNodesMatch ? LGraphEventMode.ALWAYS : mode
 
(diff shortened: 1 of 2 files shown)


webfiltered:
Hmm, this makes sense from a context menu click (usually because they say "mute" or "unmute" / "bypass").  For me, it doesn't make a huge amount of sense from a keyboard shortcut.  🤔

(p.s. It's entirely possible I am just used to this behaviour - but I use this "feature" pretty regularly, on purpose, the way it is)

DrJKL:
> Hmm, this makes sense from a context menu click (usually because they say "mute" or "unmute" / "bypass"). For me, it doesn't make a huge amount of sense from a keyboard shortcut. 🤔
> 
> (p.s. It's entirely possible I am just used to this behaviour - but I use this "feature" pretty regularly, on purpose, the way it is)

As in you select several heterogeneously enabled things to invert their individual modes?

Agreed about menu vs shortcut.
In the menu we have the power to say things like "mute all", "bypass all", "unmute all", etc.

webfiltered:
Exactly that, yep!

christian-byrne (approved):

PR #5484, opened 2025-09-11
fix(canvas): use vertical-align: top to eliminate baseline gap

benceruleanlu:
## Summary

Remove the small inline baseline gap (≈1–5 px) that caused link/slot misalignment by aligning the canvas to the line box top instead of switching its display mode.

## Changes

- Change `<canvas id="graph-canvas">` to use Tailwind `align-top` (CSS `vertical-align: top`) in `GraphCanvas.vue`.
- Keeps the canvas inline but removes the baseline/descender gap so the canvas and TransformPane share the same visual origin.
- Result: link endpoints land exactly on slot dots across zoom levels; no post-interaction jumps.

## Rationale vs. display:block

- This approach avoids changing layout mode (no switch to block-level), minimizing risk to any code that might rely on inline formatting.
- The baseline gap was the root cause; aligning to the top fixes it without altering width/height or positioning.

## Review Focus

- Verify link endpoints align with slot dots across zoom levels (FULL/REDUCED/MINIMAL LOD).
- Pan/zoom and interact: no post-interaction jumps in link position.
- Confirm no regressions in overlay components (TransformPane, menus, minimap) — they remain absolutely positioned relative to the container and aren’t dependent on inline flow.

## Video

https://github.com/user-attachments/assets/7bf3dbc8-c427-4ba8-bcbb-f7fce68f2354


benceruleanlu:
Miraculously, this also fixes some UI element misalignments across the bottom of the screen.

In fact, this fix works so well that I'm actually suspicious of it, has this been made before and reverted for some reason?

DrJKL (dismissed):

christian-byrne (dismissed):

christian-byrne:
Snapshots all looks correct?

benceruleanlu:
> Snapshots all looks correct?

yes I checked all of them

PR #5493, opened 2025-09-11
Remove requirement for type assertion

webfiltered:
## Summary

Removes the need for type guard / assertion with minor refactor - already extracting concrete type at top of func.

- Ref: #5024


AustinMroz:
It's beautiful.

PR #5510, opened 2025-09-12
Fix/vue nodes viewport culling

simula-r:
## Summary

- Fix viewport culling bug
- Improve performance of culling by keeping all nodes in the DOM and then using display: none instead of removing and adding nodes nodesToRender array.

Note: This approach to culling is a performance trade off (Rendering performance > Memory usage). In the future the high memory usage could be addressed via a client hydration system.

## Changes

- **What**: Viewport culling (fix and improved)
- **Breaking**: N/A
- **Dependencies**: N/A

https://github.com/user-attachments/assets/79c3ab5d-7e8f-46fd-9697-65b1b3b9172f

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5510-Fix-vue-nodes-viewport-culling-26c6d73d365081929475f44e80634983) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -186,9 +186,9 @@ const nodeSizes = vueNodeLifecycle.nodeSizes
 const allNodes = viewportCulling.allNodes
 
 const handleTransformUpdate = () => {
-  viewportCulling.handleTransformUpdate(
-    vueNodeLifecycle.detectChangesInRAF.value
-  )
+  viewportCulling.handleTransformUpdate()
+  // TODO: Fix paste position sync in separate PR
+  vueNodeLifecycle.detectChangesInRAF.value()
 }
 const handleNodeSelect = nodeEventHandlers.handleNodeSelect
 const handleNodeCollapse = nodeEventHandlers.handleNodeCollapse
@@ -458,10 +458,3 @@ onUnmounted(() => {
   vueNodeLifecycle.cleanup()
 })
 </script>
-
-<style>
-/* Simple CSS-based visibility control */
-.node-hidden {
-  display: none !important;
-}
-</style>
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -40,7 +40,7 @@
   >
     <!-- Vue nodes rendered based on graph nodes -->
     <VueGraphNode
-      v-for="nodeData in nodesToRender"
+      v-for="nodeData in allNodes"
       :key="nodeData.id"
       :node-data="nodeData"
       :position="nodePositions.get(nodeData.id)"
@@ -183,7 +183,7 @@ const nodeEventHandlers = useNodeEventHandlers(vueNodeLifecycle.nodeManager)
 
 const nodePositions = vueNodeLifecycle.nodePositions
 const nodeSizes = vueNodeLifecycle.nodeSizes
-const nodesToRender = viewportCulling.nodesToRender
+const allNodes = viewportCulling.allNodes
 
 const handleTransformUpdate = () => {
   viewportCulling.handleTransformUpdate(
@@ -458,3 +458,10 @@ onUnmounted(() => {
   vueNodeLifecycle.cleanup()
 })
 </script>
+
+<style>
+/* Simple CSS-based visibility control */
+.node-hidden {
+  display: none !important;
+}
+</style>
--- a/src/composables/graph/useViewportCulling.ts
+++ b/src/composables/graph/useViewportCulling.ts
@@ -1,15 +1,14 @@
 /**
- * Viewport Culling - Direct DOM manipulation without caching
+ * Vue Nodes Viewport Culling
  *
  * Principles:
  * 1. Query DOM directly using data attributes (no cache to maintain)
- * 2. Use most efficient CSS update method (display: none)
+ * 2. Set display none on element to avoid cascade resolution overheead
  * 3. Only run when transform changes (event driven)
  */
 import { type Ref, computed } from 'vue'
 
 import type { VueNodeData } from '@/composables/graph/useGraphNodeManager'
-import { app as comfyApp } from '@/scripts/app'
 import { useCanvasStore } from '@/stores/graphStore'
 
 interface NodeManager {
@@ -35,7 +34,7 @@ export function useViewportCulling(
    * Queries DOM directly - no cache maintenance needed
    */
   const updateVisibility = () => {
-    if (!nodeManager.value || !canvasStore.canvas || !comfyApp.canvas) return
+    if (!nodeManager.value || !canvasStore.canvas) return
 
     const canvas = canvasStore.canvas
     const manager = nodeManager.value
@@ -63,15 +62,16 @@ export function useViewportCulling(
       const screen_width = node.size[0] * ds.scale
       const screen_height = node.size[1] * ds.scale
 
-      const isVisible = !(
+      const isVisible =
         screen_x + screen_width < -margin ||
         screen_x > viewport_width + margin ||
         screen_y + screen_height < -margin ||
         screen_y > viewport_height + margin
-      )
 
-      // Setting diplay directly SHOULD avoid cascade resolution
-      ;(element as HTMLElement).style.display = isVisible ? '' : 'none'
+      // Setting hidden directly avoid cascade resolution
+      if (element instanceof HTMLElement) {
+        element.style.display = !isVisible ? '' : 'none'
+      }
     }
   }
 
--- a/src/composables/graph/useViewportCulling.ts
+++ b/src/composables/graph/useViewportCulling.ts
@@ -3,7 +3,7 @@
  *
  * Principles:
  * 1. Query DOM directly using data attributes (no cache to maintain)
- * 2. Set display none on element to avoid cascade resolution overheead
+ * 2. Set display none on element to avoid cascade resolution overhead
  * 3. Only run when transform changes (event driven)
  */
 import { type Ref, computed } from 'vue'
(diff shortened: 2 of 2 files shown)


src/composables/graph/useViewportCulling.ts
```diff
@@ -1,16 +1,14 @@
...
  *
- * Handles viewport culling optimization for Vue nodes including:
- * - Transform state synchronization
- * - Visible node calculation with screen space transforms
- * - Adaptive margin computation based on zoom level
- * - Performance optimizations for large graphs
+ * Principles:
+ * 1. Cache DOM elements ONCE when mounted
+ * 2. No timeouts - event driven
+ * 3. No reactivity - pure DOM manipulation
```
simula-r: No reactivity... proceeds to import ref from vue on the next line 0_o
Need to change to "No dom reactivity"

src/components/graph/GraphCanvas.vue
```diff
@@ -458,3 +458,10 @@ onUnmounted(() => {
   vueNodeLifecycle.cleanup()
 })
 </script>
+
+<style>
+/* Simple CSS-based visibility control */
+.node-hidden {
+  display: none !important;
```
DrJKL: Please don't use `!important`. There is always a better way.

Can you add `.hidden` wherever `.node-hidden` is being added?

DrJKL: I'd also like to find a way to get rid of any global styles in our components, but that's a lesser issue, IMO.

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +23,102 @@ export function useViewportCulling(
...
 
-    // Detect node changes during transform updates
-    detectChangesInRAF()
-
-    // Trigger reactivity for nodesToRender
-    void nodesToRender.value.length
+    // Add new nodes to cache
+    for (const [id] of vueNodeData.value) {
+      if (!elementCache.has(id)) {
+        const element = document.querySelector(`[data-node-id="${id}"]`)
```
DrJKL: This seems like it could get pretty heavy

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +23,102 @@ export function useViewportCulling(
...
-   * Calculate if a specific node is visible in viewport
-   * Useful for individual node visibility checks
+   * Update visibility - uses cached elements
    */
-  const isNodeVisible = (nodeData: VueNodeData): boolean => {
-    if (!nodeManager.value || !canvasStore.canvas || !comfyApp.canvas) {
-      return true // Default to visible if culling not available
-    }
+  const updateVisibility = () => {
+    if (!nodeManager.value || !canvasStore.canvas || !comfyApp.canvas) return
```
DrJKL, line 38: I don't know this for sure, but aren't those the same `canvas`?
Do we need to reference it through each of them?

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +23,102 @@ export function useViewportCulling(
...
+      const isVisible = !(
+        screen_x + screen_width < -margin ||
+        screen_x > viewport_width + margin ||
+        screen_y + screen_height < -margin ||
+        screen_y > viewport_height + margin
+      )
+
+      // Toggle visibility class
+      if (isVisible) {
+        element.classList.remove('node-hidden')
```
DrJKL: Would it be possible to have the nodes listen to a store and toggle the class themselves?

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +23,102 @@ export function useViewportCulling(
...
-    return !(
-      screen_x + screen_width < -margin_x ||
-      screen_x > viewport_width + margin_x ||
-      screen_y + screen_height < -margin_y ||
-      screen_y > viewport_height + margin_y
-    )
+    const margin = 200 * ds.scale
+
+    // Update visibility using cached elements
+    elementCache.forEach((element, nodeId) => {
```
DrJKL: @arjansingh @webfiltered

https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/no-array-for-each.md

I don't know if Map's forEach has the same downsides as Array...

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +23,102 @@ export function useViewportCulling(
...
-  const getViewportInfo = () => {
-    if (!canvasStore.canvas || !comfyApp.canvas) {
-      return null
+  const handleTransformUpdate = (detectChangesInRAF?: () => void) => {
+    if (!isVueNodesEnabled.value) return
+
+    // Detect node position/size changes if function provided
+    if (detectChangesInRAF) {
+      detectChangesInRAF()
     }
```
DrJKL:
```suggestion
    detectChangesInRAF?.()
```

DrJKL (changes requested):
A few changes I'd really like to see, but I can't wait for this bug to get fixed.

src/composables/graph/useViewportCulling.ts
christian-byrne: One is a ref

https://github.com/Comfy-Org/ComfyUI_frontend/blob/9aeeab9a9fdf89c3a239a4800fcd8d1070a55a94/src/stores/graphStore.ts#L27

src/components/graph/GraphCanvas.vue
simula-r: Actually i think for this use case it will be better to use direct element property i.e. element.hidden = true Because it SHOULD avoid the specificity resolution that a class would cause.

src/composables/graph/useViewportCulling.ts
simula-r: Truth.

src/composables/graph/useViewportCulling.ts
simula-r: We want to avoid as much reactivity / dependencies to reconcile as possible. I think its best to just surgically update the DOM here.

src/composables/graph/useViewportCulling.ts
DrJKL: So in that case, the ref will always be truthy if we don't dereference it?

src/composables/graph/useViewportCulling.ts
```diff
@@ -88,32 +70,31 @@ export function useViewportCulling(
...
       )
 
-      // Toggle visibility class
-      if (isVisible) {
-        element.classList.remove('node-hidden')
-      } else {
-        element.classList.add('node-hidden')
-      }
+      // Setting diplay directly SHOULD avoid cascade resolution
+      ;(element as HTMLElement).style.display = isVisible ? '' : 'none'
```
DrJKL: Can you use an `instanceof` to narrow the type so you don't have to use a type assertion?

DrJKL (dismissed):

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +22,84 @@ export function useViewportCulling(
...
+
+      const isVisible =
+        screen_x + screen_width < -margin ||
+        screen_x > viewport_width + margin ||
+        screen_y + screen_height < -margin ||
+        screen_y > viewport_height + margin
+
+      // Setting display none directly avoid potential cascade resolution
+      if (element instanceof HTMLElement) {
+        element.style.display = !isVisible ? '' : 'none'
```
christian-byrne: Is this inversed?

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +22,84 @@ export function useViewportCulling(
...
+    // Get all node elements at once
+    const nodeElements = document.querySelectorAll('[data-node-id]')
+
+    // Update each element's visibility
+    for (const element of nodeElements) {
+      const nodeId = element.getAttribute('data-node-id')
+      if (!nodeId) return
+
+      const node = manager.getNode(nodeId)
+      if (!node) return
```
christian-byrne: Should the two `return` be `continue`?

src/composables/graph/useViewportCulling.ts
```diff
@@ -1,17 +1,14 @@
...
+ * Vue Nodes Viewport Culling
  *
- * Handles viewport culling optimization for Vue nodes including:
- * - Transform state synchronization
- * - Visible node calculation with screen space transforms
- * - Adaptive margin computation based on zoom level
- * - Performance optimizations for large graphs
+ * Principles:
+ * 1. Query DOM directly using data attributes (no cache to maintain)
+ * 2. Set display none on element to avoid cascade resolution overheead
```
christian-byrne: nit: typo "overheead"

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +22,84 @@ export function useViewportCulling(
...
+      const node = manager.getNode(nodeId)
+      if (!node) return
+
+      // Calculate if in viewport
+      const screen_x = (node.pos[0] + ds.offset[0]) * ds.scale
+      const screen_y = (node.pos[1] + ds.offset[1]) * ds.scale
+      const screen_width = node.size[0] * ds.scale
+      const screen_height = node.size[1] * ds.scale
+
+      const isVisible =
```
arjansingh: nit: `isInsideViewport` is more descriptive.

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +22,84 @@ export function useViewportCulling(
...
-    )
+    const margin = 500 * ds.scale
+
+    // Get all node elements at once
+    const nodeElements = document.querySelectorAll('[data-node-id]')
+
+    // Update each element's visibility
+    for (const element of nodeElements) {
+      const nodeId = element.getAttribute('data-node-id')
+      if (!nodeId) return
```
arjansingh: This will exit the function `updateVisibility` entirely, did want to `continue` and just skip to next iteration?

For example:

```html
<div data-node-id="node1"></div>
<div data-node-id=""></div><!-- exits -->
<div data-node-id></div><!-- exits -->
<div></div><!-- exits -->
<div data-node-id="node3"></div>
```

Only `node1` gets processed, function would on any of the subsequent nodes if they existed, node3 never gets visibility updated.

src/composables/graph/useViewportCulling.ts
```diff
@@ -25,188 +22,84 @@ export function useViewportCulling(
...
-    return !(
-      screen_x + screen_width < -margin_x ||
-      screen_x > viewport_width + margin_x ||
-      screen_y + screen_height < -margin_y ||
-      screen_y > viewport_height + margin_y
-    )
+    const margin = 500 * ds.scale
+
+    // Get all node elements at once
+    const nodeElements = document.querySelectorAll('[data-node-id]')
```
arjansingh, line 50: nit (definitely can ignore):

```suggestion
    const nodeElements = document.querySelectorAll('[data-node-id]:not([data-node-id=""])')
```

would exclude:
```html
<div data-node-id=""></div> 
```

would not exclude so we still need the `!nodeId` check:
```html
<div data-node-id></div>
```

DrJKL (dismissed):

christian-byrne (approved):
LGTM!

PR #5514, opened 2025-09-12
Remove root-level comment in vue file

christian-byrne:
This is valid syntax according to [this part of vue SFC spec](https://vuejs.org/api/sfc-spec?utm_source=chatgpt.com#comments), but it can cause issues babel.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5514-Remove-root-level-comment-in-vue-file-26c6d73d365081f2b733dd7f5ad90f71) by [Unito](https://www.unito.io)


snomiao:
LGTM

PR #5516, opened 2025-09-12
Component: Vue Widget Slider (new)

DrJKL:
## Summary

Replace PrimeVue Slider with a shadcn/vue / reka-ui component.
Matches the Design styles.
Fixes an issue with the step not applying to the slider.
Step/min/max now apply to both the slider and the paired numeric input.

## Changes

- **What**: Slider works at different zoom levels now.
- **Dependencies**:  tw-animate-css, reka-ui

## Review Focus

Try it out, see how the new slider feels.

## Screenshots

https://github.com/user-attachments/assets/4aa103bf-ee36-4342-b3f1-c7b2df22607a

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5516-Component-Vue-Widget-Slider-new-26c6d73d36508113b877ef6d86e37f4e) by [Unito](https://www.unito.io)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,1 @@
+ignore-workspace-root-check=true
--- a/src/components/ui/slider/Slider.vue
+++ b/src/components/ui/slider/Slider.vue
@@ -41,8 +41,9 @@ const forwarded = useForwardPropsEmits(delegatedProps, emits)
       :class="
         cn(
           'bg-node-stroke relative grow overflow-hidden rounded-full',
+          'cursor-pointer',
           'data-[orientation=horizontal]:h-0.5 data-[orientation=horizontal]:w-full',
-          'data-[orientation=vertical]:h-full data-[orientation=vertical]:w-1.5'
+          'data-[orientation=vertical]:h-full data-[orientation=vertical]:w-0.5'
         )
       "
     >
@@ -56,9 +57,11 @@ const forwarded = useForwardPropsEmits(delegatedProps, emits)
       v-for="(_, key) in modelValue"
       :key="key"
       data-slot="slider-thumb"
+      tabindex="0"
       :class="
         cn(
           'bg-node-component-surface-highlight ring-node-component-surface-selected block size-3.5 shrink-0 rounded-full shadow-sm transition-[color,box-shadow]',
+          'cursor-grab',
           'hover:ring-2 focus-visible:ring-2 focus-visible:outline-hidden disabled:pointer-events-none disabled:opacity-50'
         )
       "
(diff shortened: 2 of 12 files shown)


.npmrc
```diff
@@ -0,0 +1 @@
+ignore-workspace-root-check=true
```
AustinMroz, line 1: Very nice. Thank you.

src/components/ui/slider/Slider.vue
```diff
@@ -0,0 +1,67 @@
+<!-- eslint-disable vue/no-template-shadow vue/multi-word-component-names -->
```
christian-byrne: nit: For disables, should we add rationale in the comment for future readers?

AustinMroz (dismissed):
Nothing blocking or specific code issues I can see, but several functional nits.
- The new slider requires you click on the thumb to start adjusting the value. This isn't blocking and I'm not seeing any design notes, but I think it's annoying.
  - The mouse cursor is in the drag state when hovering the widget even when not over the thumb. This is particularly annoying when trying to grab the thumb from a moderate zoom level.
  - The required y level for clicks is a little unforgiving. Even if exactly aligned on the x axis, about half the y-axis is dead space for clicks.
- Sometimes, the thumb will become locked in the "hovered" state after completing a grab. EDIT: This appears to be caused by a toast being on screen O.O
- (very nit) Changing the widget value in js does not update the position of the slider. Quick testing seems to indicate this is shared with other vue nodes: A primitive set to randomize after generate won't actually display the changed value after a generation.

arjansingh (dismissed):
Appreciate that there's less custom code needed to make this work.

christian-byrne (dismissed):
LGTM!

src/components/ui/slider/Slider.vue
DrJKL: Yes, though I'd also like to fix the issue and get rid of the lint error. Wanted to get all the functional parts working first.
Isn't there a babel bug with top level comments like this somewhere?

PabloWiedemann:
@DrJKL Requested changes:
- Change cursor to be pointer rather than "hand" icon -> Good to have additions: 
     - When dragging then change to grab
     - user can click anywhere on the slider bar to set the handle (hence slider value) to the corresponding value on that sliders position  
- Change target area for changing the cursor to be on the handle, rather than the entire component
- When clearing all the numbers in the number input field and clicking away from it -> return to previous values

src/components/ui/slider/Slider.vue
DrJKL: Turning the rules off for now, will clean up the violations later, hopefully while turning on more rules.

christian-byrne (approved):

PR #5525, opened 2025-09-13
fix: Use generic type in multi-select widget and fix options binding

christian-byrne:
## Summary

Fixed multi-select widget option binding and used generic type. Bring over some improvements from https://github.com/Comfy-Org/ComfyUI_frontend/blob/0e01ca0a980f0a1bf5bbeac90972a53454711e84/src/renderer/extensions/vueNodes/widgets/components/form/FormSelectButton.vue.

## Changes

- **What**: Enhanced [TypeScript generics](https://www.typescriptlang.org/docs/handbook/2/generics.html) and option value binding for FormSelectButton component
- **What**: Added null safety checks for image handling in WidgetGalleria component
- **What**: Improved prop binding with explicit fallbacks in WidgetFileUpload component
- **What**: Enhanced textarea focus handling with cross-environment compatibility in WidgetMarkdown component

## Review Focus

Generic type constraints for option objects and null safety in image array processing. Textarea focus behavior may vary across test environments.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5525-fix-Use-generic-type-in-multi-select-widget-and-fix-options-binding-26d6d73d365081828622e7b4dcfaa203) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
@@ -41,10 +41,10 @@ const emit = defineEmits<{
 }>()
 
 // Use the composable for consistent widget value handling
-const { localValue, onChange } = useWidgetValue({
+const { localValue, onChange } = useWidgetValue<T[]>({
   widget: props.widget,
   modelValue: props.modelValue,
-  defaultValue: [] as T[],
+  defaultValue: [],
   emit
 })
 
@@ -66,8 +66,8 @@ const combinedProps = computed(() => ({
 const multiSelectOptions = computed((): T[] => {
   const options = props.widget.options
 
-  if (options?.values && Array.isArray(options.values)) {
-    return options.values as T[]
+  if (Array.isArray(options?.values)) {
+    return options.values
   }
 
   return []
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
@@ -2,6 +2,7 @@
   <WidgetLayoutField :widget="widget">
     <MultiSelect
       v-model="localValue"
+      :options="multiSelectOptions"
       v-bind="combinedProps"
       :disabled="readonly"
       class="w-full text-xs"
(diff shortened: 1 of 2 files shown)


src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
```diff
@@ -60,4 +61,15 @@ const combinedProps = computed(() => ({
   ...filterWidgetProps(props.widget.options, MULTISELECT_EXCLUDED_PROPS),
   ...transformCompatProps.value
 }))
+
+// Extract multiselect options from widget options
+const multiSelectOptions = computed((): T[] => {
+  const options = props.widget.options
+
+  if (options?.values && Array.isArray(options.values)) {
```
DrJKL: `undefined` isn't an array, so this should be equivalent.
```suggestion
  if (Array.isArray(options?.values)) {
```

src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
```diff
@@ -60,4 +61,15 @@ const combinedProps = computed(() => ({
...
+    return options.values as T[]
```
DrJKL: Any way this can be validated instead of just asserted?

src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
```diff
@@ -30,20 +31,20 @@ import {
...
   readonly?: boolean
 }>()
 
 const emit = defineEmits<{
-  'update:modelValue': [value: any[]]
+  'update:modelValue': [value: T[]]
 }>()
 
 // Use the composable for consistent widget value handling
 const { localValue, onChange } = useWidgetValue({
```
DrJKL:
```suggestion
const { localValue, onChange } = useWidgetValue<T[]>({
 widget: props.widget,
 modelValue: props.modelValue,
  defaultValue: [],
 emit
})
```

DrJKL (dismissed):

src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5525/commits/7acd9b5567508393608a35b449e093497129b60d

src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5525/commits/7acd9b5567508393608a35b449e093497129b60d

src/renderer/extensions/vueNodes/widgets/components/WidgetMultiSelect.vue
christian-byrne: In theory we can expose a validator function but the widget props are set via Python so I don't think it would work that well.

PR #5534, opened 2025-09-13
add sortOrder for setting item

jtydhr88:
## Summary

in some desgin requirement, some settings should display on top of others

## Changes

- **What**: add sortOrder for setting item

https://github.com/user-attachments/assets/6953810c-6a9a-4a50-b966-5e07965d954c

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5534-add-priority-for-setting-item-26d6d73d36508120af77e3c7e1095daa) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/SettingDialogContent.vue
+++ b/src/components/dialog/content/SettingDialogContent.vue
@@ -114,7 +114,6 @@ const sortedGroups = (category: SettingTreeNode): ISettingGroup[] => {
     .map((group) => ({
       label: group.label,
       settings: flattenTree<SettingParams>(group).sort((a, b) => {
-        // Sort by priority first, higher priority comes first
         const sortOrderA = a.sortOrder ?? 0
         const sortOrderB = b.sortOrder ?? 0
 
--- a/src/components/dialog/content/SettingDialogContent.vue
+++ b/src/components/dialog/content/SettingDialogContent.vue
@@ -113,7 +113,18 @@ const sortedGroups = (category: SettingTreeNode): ISettingGroup[] => {
     .sort((a, b) => a.label.localeCompare(b.label))
     .map((group) => ({
       label: group.label,
-      settings: flattenTree<SettingParams>(group)
+      settings: flattenTree<SettingParams>(group).sort((a, b) => {
+        // Sort by priority first, higher priority comes first
+        const priorityA = a.priority ?? 0
+        const priorityB = b.priority ?? 0
+
+        if (priorityA !== priorityB) {
+          return priorityB - priorityA
+        }
+
+        // If priorities are equal, sort alphabetically by name
+        return a.name.localeCompare(b.name)
+      })
     }))
 }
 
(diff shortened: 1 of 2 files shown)


src/components/dialog/content/SettingDialogContent.vue
```diff
@@ -113,7 +113,18 @@ const sortedGroups = (category: SettingTreeNode): ISettingGroup[] => {
...
+        const priorityA = a.priority ?? 0
+        const priorityB = b.priority ?? 0
+
+        if (priorityA !== priorityB) {
+          return priorityB - priorityA
+        }
+
+        // If priorities are equal, sort alphabetically by name
+        return a.name.localeCompare(b.name)
+      })
```
christian-byrne: Keep original ordering for all other settings and make stable:

```typescript
settings: flattenTree<SettingParams>(group).sort((a, b, indexA, indexB) => {
  // Sort by priority first, higher priority comes first
  const priorityA = a.priority ?? 0
  const priorityB = b.priority ?? 0
  if (priorityA !== priorityB) {
    return priorityB - priorityA
  }

  // If priorities are equal, preserve the relative order (stable sort)
  return indexA - indexB // Use original index for stable sorting
})
```

src/types/settingTypes.ts
```diff
@@ -43,6 +43,9 @@ export interface SettingParams<TValue = unknown> extends FormItem {
   versionAdded?: string
   // Version of the setting when it was last modified
   versionModified?: string
+  // Priority for sorting settings within a group. Higher values appear first.
+  // Default is 0 if not specified.
+  priority?: number
```
christian-byrne: Can we change to `sortOrder`?

christian-byrne (changes requested):

src/types/settingTypes.ts
jtydhr88: fixed

src/components/dialog/content/SettingDialogContent.vue
jtydhr88: sort func doesn't have (a, b, indexA, indexB), 
but we can just use
`return priorityB - priorityA` directly 
if both value of priority is 0, it is able to keep the stable sort

DrJKL: https://es-toolkit.dev/reference/array/sortBy.html

jtydhr88:
I saw one failed playwright test but I don't think it is caused by my change

src/components/dialog/content/SettingDialogContent.vue
jtydhr88: I think we can keep the current method - sort and only add 
```
const priorityA = a.priority ?? 0
const priorityB = b.priority ?? 0
return priorityB - priorityA
```
which fulfill my need and keep stable sorting

DrJKL:
I know commenting about naming is the worst, but I would imagine `sortOrder` being more of the final index.
I think `priority` would be okay, I also like `weight`, a la Hugo's pages, especially if bigger goes earlier.

christian-byrne:
But why, seeing a setting type, assume "priority" refers to the ordreing when settings are presented in a list? There's so many things it could mean when referring to settings/configs.

christian-byrne (approved):

PR #5351, opened 2025-09-04
SubgraphNode widgets V2

AustinMroz:
Largely functional, but needs structural improvements.

Configuration of elements is done through a new sidebar tab. Using a tab here is a placeholder.

Known issues:
- Performance has not yet been optimized and isn't great.
- Widgets in subgraphs do not compute themselves as disabled until a draw has occurred.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5351-SubgraphNode-widgets-V2-2646d73d365081ee8427d5448f14e509) by [Unito](https://www.unito.io)

PR #5537, opened 2025-09-13
Subgraph widget promotion - Part 1

AustinMroz:
Small minimally objectionable prerequisite changes required for the widget promotion implementation.
- DOMWidgets ignored their active state and display solely based off if they are calculated as visible
- ComboWidgets used lots of private methods that were particularly thorny to proxy.
- Update the graph swap bookkeeping for DOMWidgets. Since a single DOMWidget element can now be owned by (up to) one widget per graph, slightly more robust logic is required to determine if a widget is active when swapping between subgraphs.
- Add a new Disconnected Widget
  - Not currently used anywhere (which necessitates a temporary knip ignore).
  - To be used as a placeholder if a user deletes a node with a promoted widget inside a subgraph.
  - Will not be added to widget registry or exposed for external use.
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5537-Subgraph-widget-promotion-Part-1-26d6d73d36508105a7dad8a83696dea4) by [Unito](https://www.unito.io)

### DIFF
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -25,7 +25,9 @@ const config: KnipConfig = {
     'src/types/generatedManagerTypes.ts',
     'src/types/comfyRegistryTypes.ts',
     // Used by a custom node (that should move off of this)
-    'src/scripts/ui/components/splitButton.ts'
+    'src/scripts/ui/components/splitButton.ts',
+    // Staged for for use with subgraph widget promotion
+    'src/lib/litegraph/src/widgets/DisconnectedWidget.ts'
   ],
   compilers: {
     // https://github.com/webpro-nl/knip/issues/1008#issuecomment-3207756199
--- a/src/lib/litegraph/src/widgets/ComboWidget.ts
+++ b/src/lib/litegraph/src/widgets/ComboWidget.ts
@@ -45,7 +45,7 @@ export class ComboWidget
     return typeof this.value === 'number' ? String(this.value) : this.value
   }
 
-  getValues(node: LGraphNode): Values {
+  private getValues(node: LGraphNode): Values {
     const { values } = this.options
     if (values == null) throw new Error('[ComboWidget]: values is required')
 
@@ -57,7 +57,7 @@ export class ComboWidget
    * @param increment `true` if checking the use of the increment button, `false` for decrement
    * @returns `true` if the value is at the given index, otherwise `false`.
    */
-  canUseButton(increment: boolean): boolean {
+  private canUseButton(increment: boolean): boolean {
     const { values } = this.options
     // If using legacy duck-typed method, false is the most permissive return value
     if (typeof values === 'function') return false
@@ -93,7 +93,7 @@ export class ComboWidget
     this.tryChangeValue(-1, options)
   }
 
-  tryChangeValue(delta: number, options: WidgetEventOptions): void {
+  private tryChangeValue(delta: number, options: WidgetEventOptions): void {
     const values = this.getValues(options.node)
     const indexedValues = toArray(values)
 
--- a/src/lib/litegraph/src/widgets/ComboWidget.ts
+++ b/src/lib/litegraph/src/widgets/ComboWidget.ts
@@ -45,7 +45,7 @@ export class ComboWidget
     return typeof this.value === 'number' ? String(this.value) : this.value
   }
 
-  #getValues(node: LGraphNode): Values {
+  getValues(node: LGraphNode): Values {
     const { values } = this.options
     if (values == null) throw new Error('[ComboWidget]: values is required')
 
@@ -57,7 +57,7 @@ export class ComboWidget
    * @param increment `true` if checking the use of the increment button, `false` for decrement
    * @returns `true` if the value is at the given index, otherwise `false`.
    */
-  #canUseButton(increment: boolean): boolean {
+  canUseButton(increment: boolean): boolean {
     const { values } = this.options
     // If using legacy duck-typed method, false is the most permissive return value
     if (typeof values === 'function') return false
@@ -78,23 +78,23 @@ export class ComboWidget
    * Handles edge case where the value is both the first and last item in the list.
    */
   override canIncrement(): boolean {
-    return this.#canUseButton(true)
+    return this.canUseButton(true)
   }
 
   override canDecrement(): boolean {
-    return this.#canUseButton(false)
+    return this.canUseButton(false)
   }
 
   override incrementValue(options: WidgetEventOptions): void {
-    this.#tryChangeValue(1, options)
+    this.tryChangeValue(1, options)
   }
 
   override decrementValue(options: WidgetEventOptions): void {
-    this.#tryChangeValue(-1, options)
+    this.tryChangeValue(-1, options)
   }
 
-  #tryChangeValue(delta: number, options: WidgetEventOptions): void {
-    const values = this.#getValues(options.node)
+  tryChangeValue(delta: number, options: WidgetEventOptions): void {
+    const values = this.getValues(options.node)
     const indexedValues = toArray(values)
 
     // avoids double click event
@@ -128,7 +128,7 @@ export class ComboWidget
     if (x > width - 40) return this.incrementValue({ e, node, canvas })
 
     // Otherwise, show dropdown menu
-    const values = this.#getValues(node)
+    const values = this.getValues(node)
     const values_list = toArray(values)
 
     // Handle center click - show dropdown menu
--- a/src/lib/litegraph/src/widgets/DisconnectedWidget.ts
+++ b/src/lib/litegraph/src/widgets/DisconnectedWidget.ts
@@ -1,0 +1,38 @@
+import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
+import type { IButtonWidget } from '@/lib/litegraph/src/types/widgets'
+
+import { BaseWidget, type DrawWidgetOptions } from './BaseWidget'
+
+class DisconnectedWidget extends BaseWidget<IButtonWidget> {
+  constructor(widget: IButtonWidget, node: LGraphNode) {
+    super(widget, node)
+    this.disabled = true
+  }
+
+  override drawWidget(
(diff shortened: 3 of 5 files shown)


src/lib/litegraph/src/widgets/ComboWidget.ts
```diff
@@ -45,7 +45,7 @@ export class ComboWidget
     return typeof this.value === 'number' ? String(this.value) : this.value
   }
 
-  #getValues(node: LGraphNode): Values {
+  getValues(node: LGraphNode): Values {
```
DrJKL: Want to mark them as private, [typescript style](https://www.typescriptlang.org/docs/handbook/2/classes.html#private)?

DrJKL: Or protected if needed by a subclass.

src/lib/litegraph/src/widgets/DisconnectedWidget.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+const conf: IButtonWidget = {
+  type: 'button',
+  value: undefined,
+  name: 'Disconnected',
+  options: {},
+  y: 0,
+  clicked: false
+}
+export const disconnectedWidget = new DisconnectedWidget(conf, {} as LGraphNode)
```
DrJKL: Do you need `as LGraphNode`?

src/lib/litegraph/src/widgets/DisconnectedWidget.ts
```diff
@@ -0,0 +1,38 @@
...
+  constructor(widget: IButtonWidget, node: LGraphNode) {
+    super(widget, node)
+    this.disabled = true
+  }
+
+  override drawWidget(
+    ctx: CanvasRenderingContext2D,
+    { width, showText = true }: DrawWidgetOptions
+  ) {
+    ctx.save()
```
DrJKL, line 16: 🙌🏻

src/scripts/app.ts
```diff
@@ -840,19 +841,23 @@ export class ComfyApp {
...
 
-        // Assertions: UnwrapRef
-        for (const { widget } of widgetStore.activeWidgetStates) {
-          if (!nodeSet.has(widget.node)) {
-            widgetStore.deactivateWidget(widget.id)
-          }
-        }
+        for (const node of newGraph.nodes)
+          for (const w of node.widgets ?? [])
+            if (w instanceof DOMWidgetImpl && w.id) widgetIds[w.id] = w
```
DrJKL: Minor nit, but could you add curly braces or turn this into a functional chain?
Like
```ts
const domWidgets = newGraph.nodes.flatMap(node => node.widgets ?? []).filter(w => w instanceof DOMWidgetImpl)
for (const w of domWidgets) {
  widgetIds[w.id] = w
}
```

Or maybe
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy
or
https://es-toolkit.dev/reference/array/groupBy.html
?

src/scripts/app.ts
```diff
@@ -840,19 +841,23 @@ export class ComfyApp {
...
 
-        for (const { widget } of widgetStore.inactiveWidgetStates) {
-          if (nodeSet.has(widget.node)) {
-            widgetStore.activateWidget(widget.id)
+        // Assertions: UnwrapRef
+        for (const widgetId of widgetStore.widgetStates.keys()) {
```
DrJKL: Does it have an `.entries()`?

src/scripts/app.ts
```diff
@@ -840,19 +841,23 @@ export class ComfyApp {
...
+          const widgetState = widgetStore.widgetStates.get(widgetId)
+          //Unreachable, but required for type safety
+          if (!widgetState) continue
+          if (widgetId in widgetIds) {
```
DrJKL: This makes me think that `widgetIds` might be missing something in its name that helps describe its purpose.

knip.config.ts
```diff
@@ -25,7 +25,9 @@ const config: KnipConfig = {
     'src/types/generatedManagerTypes.ts',
     'src/types/comfyRegistryTypes.ts',
     // Used by a custom node (that should move off of this)
-    'src/scripts/ui/components/splitButton.ts'
+    'src/scripts/ui/components/splitButton.ts',
+    // Staged for for use with subgraph widget promotion
+    'src/lib/litegraph/src/widgets/DisconnectedWidget.ts'
```
DrJKL, line 30: I really don't like code going in before it's used. It's harder to review the design decisions without that context, also YAGNI. 😞

src/lib/litegraph/src/widgets/DisconnectedWidget.ts
arjansingh: At least, `Pick<LGraphNode, /* what we care about */>`?

src/lib/litegraph/src/widgets/ComboWidget.ts
AustinMroz: The feedback I got when inquiring a couple weeks ago was to proceed with blanket removal and, of note, the ComboWidget code is the odd one out compared to all the other litegraph widgets.

Typescript private fields are neat and shouldn't cause issues here. Happy to swap if you think keeping a level of protection desirable here.

src/lib/litegraph/src/widgets/DisconnectedWidget.ts
AustinMroz: Alas, It's sole use is for a super call which requires that exact type, but will never use it.

My personal preference would be to replace the entire class with a minimal object (as is the common style for custom nodes), but that's a far bigger casting crime.

src/scripts/app.ts
AustinMroz: I'm ashamed that it slipped my mind. That is much better

src/scripts/app.ts
AustinMroz: I believe this is just a case of the comment backfiring. `map.has` doesn't function as a type guard. Swapping to `.entries() ` as mentioned above works double here for removing the annoying check.

It's a simple lookup map for DOMWidgets in the current graph by id. The function of the map was refactored while implementing and I agree the name is no longer the most descriptive.

knip.config.ts
AustinMroz: As a reminder, the full functional code exists in #5351 😉

I recall off-hand comments from multiple people advising smaller, isolated PRs for easier review and I consider the subgraph widgets implementation a particularly well suited candidate. Please let me know after the chain if I was mistaken. I put great trust in your opinion.

src/scripts/app.ts
AustinMroz: I'm getting a little whiplash here after the crusade against `forEach`, but that flatMap call is real sleek

How is (with a rename of widgetIds  to the more helpful activeWidgets)
```ts
const activeWidgets: Record<string, BaseDOMWidget<object | string>> =
  Object.fromEntries(newGraph.nodes.flatMap(node => node.widgets ?? [])
    .filter(w => w instanceof DOMWidgetImpl)
    .map(w => [w.id, w]))
```

groupBy looks great. I was specifically digging for a function like this several days back. (If I can remember where) It'll be very useful for cleaning up that code section.

christian-byrne (dismissed):
LGTM!

knip.config.ts
DrJKL: Smaller and isolated, 100%
But my preference is to have small and isolated functional changes. Slicing out a unit of value where each change is of functional benefit to a user, doing one thing.

src/lib/litegraph/src/widgets/ComboWidget.ts
DrJKL: If we don't intend for them to be called externally, I'd mark them `private`. It helps within our code and also for any extension developers who use the published types.

src/lib/litegraph/src/widgets/DisconnectedWidget.ts
DrJKL: I'm surprised nothing fails in the super constructor in that case. 😟 

Just looked at the code, so this will start causing issues when something goes through the widgets and tries to get `.node()` but only gets an empty object?

src/lib/litegraph/src/widgets/DisconnectedWidget.ts
```diff
@@ -0,0 +1,38 @@
+import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
+import type { IButtonWidget } from '@/lib/litegraph/src/types/widgets'
+
+import { BaseWidget, type DrawWidgetOptions } from './BaseWidget'
+
+class DisconnectedWidget extends BaseWidget<IButtonWidget> {
+  constructor(widget: IButtonWidget, node: LGraphNode) {
+    super(widget, node)
+    this.disabled = true
+  }
```
DrJKL: If it really doesn't need the node, you could elide it from this class' constructor and pass in the dangerous default, to make sure that other callers don't try passing in Nodes that they'd expect to connect to the disconnected widget.

DrJKL (dismissed):

src/lib/litegraph/src/widgets/DisconnectedWidget.ts
AustinMroz: Correct. And this is where your other comment on segmenting the PR into pieces gives credence.

The widget will never exist or be owned by a node. It will only ever be displayed by proxy and the proxied instance has it's `.node` wrapped.

Maybe some potential wizardry here in delineating this as a partial widget implementation, but it seems to be in the area of checks typescript struggles with and probably isn't worth the development time.

src/lib/litegraph/src/widgets/DisconnectedWidget.ts
AustinMroz: I initially missed the second comment. I like this direction and, with the benefit of further consideration, have opted to also include a proper placeholder node so there's no cast and types are satisfied.

DrJKL (approved):

christian-byrne (approved):

PR #5540, opened 2025-09-13
fix: Show title, inputs, and outputs on Vue node component even when node fails to initalize (e.g., if missing)

christian-byrne:
## Summary

Enhanced missing node title resolution with improved fallback logic and proper data extraction timing during graph configuration.

## Changes

- **What**: Improved title fallback chain in `useGraphNodeManager` (`title → type → constructor.comfyClass → constructor.title → constructor.name → 'Unknown'`)
- **What**: Added re-extraction of Vue node data after `onAfterGraphConfigured` to capture populated title/slots/widgets
- **What**: Enhanced `NodeHeader` title resolution with proper TypeScript typing and descriptive variable names

## Review Focus

Title resolution during workflow loading vs immediate initialization and fallback behavior for truly missing nodes.

## Screenshots (if applicable)

Selection_2163

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5540-fix-Show-title-inputs-and-outputs-on-Vue-node-component-even-when-node-fails-to-inital-26d6d73d3650819a962bcaf53194a63f) by [Unito](https://www.unito.io)

PR #5541, opened 2025-09-13
Show error state on Vue nodes that caused execution errors

christian-byrne:
## Summary

Similar to https://github.com/Comfy-Org/ComfyUI_frontend/pull/5540, show error state on node when the node caused an execution error (the node's execution on backend triggered an error or the websocket otherwise sent a message indicating as much).

## Changes

- **What**: Created `useExecutionErrorProvider` [composable](https://vuejs.org/guide/reusability/composables.html) for centralized error state and [dependency injection](https://vuejs.org/guide/components/provide-inject.html)
- **What**: Added `ExecutionErrorNodeIdKey` injection key for type-safe error state propagation
- **What**: Enhanced `executionStore` with `lastExecutionErrorNodeLocatorId` for subgraph-aware error tracking
- **What**: Optimized Vue node error styling with computed properties to prevent unnecessary re-renders

## Review Focus

Dependency injection pattern implementation and computed property optimization for error styling performance. Verify error border/outline styling works correctly for nodes in nested subgraphs.

## Screenshots

Selection_2164

Selection_2165

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5541-Show-error-state-on-Vue-nodes-that-caused-execution-errors-26d6d73d3650813ea2b5f2d5cc83a9dc) by [Unito](https://www.unito.io)


src/components/graph/GraphCanvas.vue
```diff
@@ -194,6 +190,8 @@ const handleNodeSelect = nodeEventHandlers.handleNodeSelect
 const handleNodeCollapse = nodeEventHandlers.handleNodeCollapse
 const handleNodeTitleUpdate = nodeEventHandlers.handleNodeTitleUpdate
 
+useExecutionErrorProvider()
```
DrJKL: I usually find it odd to see composables that are called without assignment.
Worth prefixing with `void ` to indicate that it's side-effect only?

DrJKL: Composable? Provide? Why not both?

DrJKL: Why _are_ we using a provide instead of just using the store directly from the component?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -191,6 +191,17 @@ const isSelected = computed(() => {
...
 
+// Execution error via DI (single current node id)
+const execErrorNodeId = inject(ExecutionErrorNodeIdKey, undefined)
+const hasExecutionError = computed(
+  () => execErrorNodeId?.value === props.nodeData.id
+)
+
+// Computed error states for styling
+const hasAnyError = computed(
+  () => hasExecutionError.value || props.nodeData.hasErrors
```
DrJKL: To protect against the one bug where we missed a `.value`
```suggestion
  (): boolean => hasExecutionError.value || props.nodeData.hasErrors
```

DrJKL (commented):
Have some questions about the design

src/components/graph/GraphCanvas.vue
christian-byrne: Yeah this is somewhat strange, typically I guess you would get the value from the composable and `provide` in the component. 

> Why are we using a provide instead of just using the store directly from the component?

Using directly in component requires executing the code for every component.

DrJKL: Wouldn't the store just expose a `ref` that would go into the computed the same way the `inject` value does now?

christian-byrne: You're right

src/stores/executionStore.ts
```diff
@@ -426,6 +439,14 @@ export const useExecutionStore = defineStore('execution', () => {
...
      */
     lastExecutionError,
+    /**
+     * NodeLocatorId for the most recent execution error.
+     */
+    lastExecutionErrorNodeLocatorId,
+    /**
+     * Local node ID for the most recent execution error.
+     */
+    lastExecutionErrorNodeId,
```
DrJKL: The names are pretty descriptive, not sure how much the comments add.

I'm a little annoyed that there are two forms of ID, but I'm sure there's a reason.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -189,6 +190,18 @@ const isSelected = computed(() => {
...
+// Execution error via store
+const executionStore = useExecutionStore()
+const hasExecutionError = computed(
+  () => executionStore.lastExecutionErrorNodeId === props.nodeData.id
+)
+
+// Computed error states for styling
+const hasAnyError = computed(
+  (): boolean =>
+    hasExecutionError.value || !!props.error || !!props.nodeData.hasErrors
```
DrJKL: Do you need the `!!`s?
How would you feel about destructuring the props in here so we can set defaults?

DrJKL (dismissed):

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5541/commits/65d015a0a4e895ab8e78c531245fbdc339b74c72

PR #5535, opened 2025-09-13
Add Inter font used by Figma designs

webfiltered:
## Summary

Adds the Inter fonts used in Figma designs.

## Changes

- Adds and imports fonts from https://fonts.google.com/specimen/Inter
- Being used by desktop immediately, but will be required for other dialogs / interactions.

## Review Focus

- Is this idiomatic Tailwind?

## Screenshots (if applicable)

Inter font, weight 700, 30px.

```
font-inter font-bold text-3xl
```

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5535-Add-ABC-ROM-fonts-used-by-Figma-designs-26d6d73d3650813ca559ce706d43d1fb) by [Unito](https://www.unito.io)

PR #5543, opened 2025-09-13
Add custom dialog handling for Desktop

webfiltered:
## Summary

Adds a basic dialog endpoint for the desktop app.

## Changes

- Adds a route for the desktop app to load individual windows: /desktop-dialog
- Uses GET vars to request the dialog contents

## Caveats

- Uses existing hard-coded strings; i18n can be added separately
- Asserts type of inbound buttons array; schema validation can be added if necessary, however I would argue that should wait until this is separated out from the primary Vue app
- Requires rebase https://github.com/Comfy-Org/ComfyUI_frontend/pull/5535

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5543-Add-custom-dialog-handling-for-Desktop-26d6d73d365081799ae8c84c270dc964) by [Unito](https://www.unito.io)


src/views/DesktopDialogView.vue
```diff
@@ -0,0 +1,119 @@
...
+    return []
+  }
+})
+</script>
+
+<style scoped>
+@reference '../assets/css/style.css';
+
+.dialog-title {
+  font-family: 'ABC ROM';
```
DrJKL, line 66: Could you [register the font](https://tailwindcss.com/docs/font-family#customizing-your-theme) in `style.css` and use the tailwind utility?

src/views/DesktopDialogView.vue
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.p-button {
+  @apply rounded-lg;
```
DrJKL, line 70: Please just add the class to the button, I'm trying to get rid of all of the `@apply`s and prohibit them with a linter soon.

DrJKL: Same for the others <3

src/views/DesktopDialogView.vue
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.p-button-secondary {
+  @apply text-white rounded-lg border-none;
+
+  background: var(--color-button-background, rgba(255, 255, 255, 0.15));
```
DrJKL, line 76: `bg-button-background`

src/views/DesktopDialogView.vue
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.p-button-secondary:hover {
+  background: rgba(255, 255, 255, 0.25);
```
DrJKL, line 80: `bg-white/25`

DrJKL: Similar for the ones below

src/views/DesktopDialogView.vue
```diff
@@ -0,0 +1,119 @@
...
+}
+
+.p-button-secondary:active {
+  background: rgba(255, 255, 255, 0.35);
+}
+
+.p-button-danger {
+  background: rgba(241, 67, 82, 0.5);
```
DrJKL, line 88: Could you name the base color and put it into the color palette?

DrJKL: Same for the others

src/views/DesktopDialogView.vue
```diff
@@ -0,0 +1,119 @@
...
+.p-button-warn {
+  background: rgba(23, 45, 215, 0.66);
+  color: #f0ff41;
+  border: 0;
+}
+
+.p-button-warn:hover {
+  background: rgba(23, 45, 215, 0.88);
+  color: #f0ff41;
+  border: 0;
```
DrJKL, line 109: These should inherit from the original declaration, no need to add them for each modifier state if you're not changing the value.

src/views/DesktopDialogView.vue
```diff
@@ -0,0 +1,119 @@
...
+interface DialogButton {
+  label: string
+  returnValue: string | null
+  severity?: 'primary' | 'secondary' | 'danger' | 'warn'
+}
+
+const route = useRoute()
+
+// Get title and message from query parameters
+const title = computed(() => {
```
DrJKL, line 36: Would [this](https://vueuse.org/router/useRouteQuery/#useroutequery) work?

DrJKL (changes requested):
Mostly CSS comments 😸

src/views/DesktopDialogView.vue
webfiltered: This should probably be draft, sorry~

The font isn't in the repo yet; and this PR shouldn't add it.  But yes, it annoyed me to have to create a class just for this.  😃

src/views/DesktopDialogView.vue
webfiltered: Sorry you had to read that.  The scoped styles are ~80% generated slop - which I pray does not come as a surprise.

src/views/DesktopDialogView.vue
webfiltered: Ah, nice - thank you, and yes, that would be much better.

I am not happy this is using query instead of IPC, but it really does not feel like it is worth the effort as a temporary workaround for a dialog.

webfiltered: Of course, we're not using that, because the kludge for desktop is one of two things using routes.

Do we want another dependency for a temporary workaround?

src/views/DesktopDialogView.vue
DrJKL: 🤖

src/views/DesktopDialogView.vue
DrJKL: 🤔 
@christian-byrne ?

christian-byrne: vueuse is already a dependency

webfiltered: Yeah, but vueuse has multiple packages; this is part of the router addon.

christian-byrne: Oh I see, my mistake.

webfiltered: No problem at all!  But RE: adding the dependency, let's just disregard this.

I'll move this entire feature to IPC at some point in the next 24hrs.  It's more overall code, but just cleaner.

webfiltered:
Replaced by #5605.

PR #5544, opened 2025-09-13
Update test expectations to match core node change

christian-byrne:
This commit https://github.com/comfyanonymous/ComfyUI/commit/80b7c9455bf7afba7a9e95a1eb76b172408ab56c changed LoadVAE default appearance by adding `pixel_space` value. Update test snapshot baseline to reflect new state.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5544-Update-test-expectations-to-match-core-node-change-26d6d73d36508191b2edf482aa157bf8) by [Unito](https://www.unito.io)


browser_tests/tests/widget.spec.ts-snapshots/animated-image-preview-saved-webp-chromium-linux.png
DrJKL, line 1: This one keeps flip-flopping, is there something we're not waiting for in the test?

DrJKL (dismissed):

arjansingh (dismissed):

browser_tests/tests/widget.spec.ts-snapshots/animated-image-preview-saved-webp-chromium-linux.png
christian-byrne: For this, there was a bug where animated webp were being treated as normal webp. So the test asset is an animated webp with 2 frames. The left (image being deleted) shows the second frame -- indicating the node is properly displaying animated webp images. But since it relies on animation of webp, it is just really flaky I guess.

I should just change the test to `fixme`

PR #5548, opened 2025-09-14
Mark flaky animated webp test as fixme

christian-byrne:
## Summary
- Mark the "Can preview saved animated webp image" test as `test.fixme` due to flakiness

## Context
This test keeps flip-flopping because it relies on animated webp timing, which is inherently unreliable in CI environments. The test asset is an animated webp with 2 frames, and the test depends on animation frame timing to verify that animated webp images are properly displayed (as opposed to being treated as static webp).

While the underlying functionality works (animated webp are correctly distinguished from static webp), the test is flaky due to timing dependencies with webp animation frames.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5548-Mark-flaky-animated-webp-test-as-fixme-26e6d73d3650815da2cac17a985d320e) by [Unito](https://www.unito.io)


christian-byrne:
Merging to fix CI for other devs.

PR #5566, opened 2025-09-15
new design for left click and wheel

jtydhr88:
## Summary

new design for left click and wheel to allow user set behavior:
1. Use left click to drag vs select
2. Use scroll wheel to zoom vs Pan

Add a new section at the top of the "Lite Graph" tab called Canvas Navigation
Move there the navigation modes settings -> this one still just has two mode to select from
These settings have settings that:
1. change their values depending on which mode is selected
2. can be overwritten. So if the user selects a combo that is not from the standard nor from the drag mode at the nav mode drop down we will set the text to "custom"
image

## Changes

- **What**: add two new settings, LeftMouseClickBehavior and MouseWheelScroll
- **Breaking**: on LiteGraph, we would not use canvasNavigationMode properties to control behavior

https://github.com/user-attachments/assets/ac673035-b014-4234-94b0-376197b6bcc5

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5566-new-design-for-left-click-and-wheel-26f6d73d365081e480b0f7a52ffa43f7) by [Unito](https://www.unito.io)


CoreyCorza:
Words cannot express how much this pr means to me

jtydhr88:
@CoreyCorza this is something requested by Pablo in slack directly, I am still waiting for his test results, sorry I don’t add too much design details yet, let me convert this into draft for now

christian-byrne (changes requested):
Could you remove the inline imports in the `onChange` callbacks?

jtydhr88:
I checked the current screenshots, likely they are incorrect, I need to update them

christian-byrne:
Yeah, the old ones seemed incorrect as they were zoomed in. The new images show the same zoom level but the canvas is shifted to the left slightly, which seems to match test case description "Shift + mouse wheel should pan canvas horizontally", is that accurate?


image

christian-byrne (approved):

PR #5571, opened 2025-09-15
fix: Set Vue node initial size in layout store instead of CSS

christian-byrne:
## Summary
- Vue nodes now properly set their initial size in the layout store using the `resize()` function from `useNodeLayout` on component mount
- Ensures the layout store is the single source of truth for sizing, preventing conflicts with the ResizeObserver
- Removes CSS-based initial sizing approach that was causing feedback loops

## Problem
Vue nodes were experiencing sizing issues where they would shrink to minimal size after user interaction. This was caused by:

1. Initial size set via CSS custom properties
2. ResizeObserver detecting actual DOM size and updating layout store
3. Layout store overriding the CSS sizing, creating a feedback loop
4. After interaction, nodes would shrink because ResizeObserver measurements took precedence over initial sizing

## Solution
Set the initial size directly in the layout store on component mount using the existing `resize()` function from `useNodeLayout`. This approach:

- Makes the layout store the authoritative source for sizing
- Eliminates conflicts between CSS sizing and ResizeObserver updates  
- Allows ResizeObserver to properly track content-driven size changes
- Maintains consistent sizing behavior after user interactions

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5571-fix-Set-Vue-node-initial-size-in-layout-store-instead-of-CSS-26f6d73d365081979548c0527bda6adf) by [Unito](https://www.unito.io)

### DIFF
--- a/src/assets/css/style.css
+++ b/src/assets/css/style.css
@@ -961,9 +961,7 @@ audio.comfy-audio.empty-audio-widget {
   /* Uses default styling - no overrides needed */
 }
 
-/* Smooth transitions between LOD levels */
 .lg-node {
-  transition: min-height 0.2s ease;
   /* Disable text selection on all nodes */
   user-select: none;
   -webkit-user-select: none;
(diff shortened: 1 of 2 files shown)


christian-byrne:
Merging due to pending PRs changing injected transform state. The type assertion should be removed.

PR #5587, opened 2025-09-15
[WIP] Vue node resizing functionality

christian-byrne:
## Summary

Implemented comprehensive node resizing functionality for Vue-rendered nodes with layout synchronization and performance optimizations.

## Changes

- **What**: Added interactive resize handles to Vue nodes with real-time visual feedback and layout store integration
- **Dependencies**: New composable `useNodeResize.ts` for resize interaction handling

## Review Focus

Coordinate system conversions between screen and canvas space, pointer capture handling, and layout synchronization timing to prevent reactive loops.

```mermaid
graph TD
    A[Resize Handle Interaction] --> B{Drag Started?}
    B -->|Yes| C[Pause Auto Tracking]
    C --> D[Update Local Size State]
    D --> E[Visual Feedback]
    E --> F{Drag Continuing?}
    F -->|Yes| D
    F -->|No| G[Resume Auto Tracking]
    G --> H[Sync to Layout Store]
    H --> I[Force Sync to LiteGraph]

    style I fill:#f9f9f9,stroke:#333,color:#000
```

Key implementation details:
- [Resize composable](https://vuejs.org/guide/reusability/composables.html) with pointer capture and constraint handling
- Tracking pause/resume system to prevent conflicts during manual resize
- Canvas coordinate conversion for proper scaling with zoom levels
- Workflow persistence integration with `forceSyncAll()` method

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5587-WIP-Vue-node-resizing-functionality-26f6d73d365081f0b675d19edcb26fc3) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
christian-byrne, line 1: If I don't pause the RO tracking while resizing, the node "slides" back into its original height/width after you finish resizing. I need to debug it a bit further.

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
christian-byrne, line 1: I want to find a better approach for this module.

src/renderer/core/layout/sync/useLayoutSync.ts
christian-byrne, line 1: This could be moved into a separate PR, as it's not necessary for resizing to work - just ensures the newly sized nodes' dimensions are serialized/saved.

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -1381,6 +1381,24 @@ class LayoutStoreImpl implements LayoutStore {
     // Restore original source
     this.currentSource = originalSource
   }
+
+  /**
+   * Get all node layouts for syncing to external systems
+   */
+  getAllNodeLayouts(): Map<NodeId, NodeLayout | null> {
```
christian-byrne, line 1388: We shouldn't use a function like this as it just creates the potential for de-sync and race conditions.

src/platform/workflow/persistence/composables/useWorkflowPersistence.ts
christian-byrne, line 1: Similar to other file, this can be moved to separate PR

> This could be moved into a separate PR, as it's not necessary for resizing to work - just ensures the newly sized nodes' dimensions are serialized/saved.

AustinMroz (commented):
Functionality and direction both look great for a WIP

The `forceSyncAll` in `persistCurrentWorkflow` is a acceptable, but a little ick. It's a slippery slope to O(n^2) performance where every type of action creates a persistence call which causes every possible action to sync. From my testing,
- Syncing resizes back to litegraph is working fine.
- execution of `persistCurrentWorkflow`  after a resize has strange behaviour currently. A resize seems to set state as dirty. but doesn't initiate a persist itself, so the dirty state persists until the next option that can resolve dirty state, like the end of a canvas pan

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
```diff
@@ -0,0 +1,143 @@
...
+  onEnd?: () => void
+}
+
+export function useNodeResize(
+  resizeCallback: (size: { width: number; height: number }) => void,
+  options: UseNodeResizeOptions = {}
+) {
+  const {
+    minWidth = 200,
+    minHeight = 100,
```
Myestery, line 37: can we have some of these as constants somewhere?

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
arjansingh: `useElementSize`?

https://vueuse.org/core/useElementSize/

```ts
  export function useNodeResize(nodeId) {
    const { batchSync } = useLayoutBatching()
    const nodeElement = useTemplateRef<HTMLElement>('nodeElementRef')

    const { width, height } = useElementSize(nodeElement)

    watch([width, height], ([newWidth, newHeight]: [number, number]) => {
      if (!newWidth || !newHeight) return

      batchSync(nodeId, { width: newWidth, height: newHeight })
    })

    // Separate resize handle logic (not using useDraggable)
    function attachResizeHandles() {
      // Implement custom resize handles with mouse events
      // This directly modifies DOM element size
      // useElementSize will detect the changes and trigger batchSync
    }

    return { width, height, nodeElement }
  }
```

src/platform/workflow/persistence/composables/useWorkflowPersistence.ts
```diff
@@ -12,13 +13,20 @@ import { useSettingStore } from '@/stores/settingStore'
...
   const workflowPersistenceEnabled = computed(() =>
     settingStore.get('Comfy.Workflow.Persist')
   )
 
   const persistCurrentWorkflow = () => {
     if (!workflowPersistenceEnabled.value) return
+
+    // Force sync all layout changes to LiteGraph before serialization
+    if (comfyApp.canvas) {
+      forceSyncAll(comfyApp.canvas)
```
arjansingh, line 27: maybe instead of a forced sync, we can try to just update once a frame?

something like:

```ts
  // useLayoutBatching.ts
  import { reactive, onScopeDispose } from 'vue'

  export function useLayoutBatching() {
    const pendingUpdates = reactive(new Map<NodeId, NodeLayout>())
    let rafId: number | null = null

    function batchSync(nodeId: NodeId, layout: NodeLayout) {
      // Always accumulate updates in reactive map
      pendingUpdates.set(nodeId, layout)

      // Schedule RAF if not already scheduled
      if (rafId) return
      rafId = requestAnimationFrame(processAllUpdates)
    }

    function processAllUpdates() {
      // Create snapshot of updates to process
      const updatesToProcess = new Map(pendingUpdates)
      pendingUpdates.clear()
      flushAllUpdates(updatesToProcess)

      // Single canvas dirty call per frame
      canvas.setDirty()
      rafId = null
    }

    function flushAllUpdates(updates: Map<NodeId, NodeLayout>) {
      updates.forEach((layout, nodeId) => {
        updateNodeLayout(nodeId, layout)
      })
    }

    function updateNodeLayout(nodeId: NodeId, layout: NodeLayout) {
      // Update the Vue reactive store with new node dimensions
      // Should call layoutStore.updateNodeLayout(nodeId, layout) or similar
      // This keeps Vue state in sync with DOM changes
    }

    function cleanup() {
      if (rafId !== null) {
        cancelAnimationFrame(rafId)
        rafId = null
      }
      pendingUpdates.clear()
    }

    // Auto-cleanup when scope is disposed
    onScopeDispose(cleanup)

    return {
      batchSync,
      cleanup // Expose for manual cleanup if needed
    }
  }

```

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -122,6 +124,13 @@
         />
       </div>
     </template>
+
+    <!-- Resize handle -->
+    <div
+      v-if="!readonly"
+      class="absolute bottom-0 right-0 w-3 h-3 cursor-se-resize opacity-0 hover:opacity-20 hover:bg-white transition-opacity duration-200"
+      @pointerdown.stop="startResize"
+    />
```
arjansingh, line 133: Proposed solution would be something like:

```html
  <script setup lang="ts">
  import { useNodeResize } from './composables/useNodeResize'

  interface Props {
    nodeId: string
    title: string
  }

  const props = defineProps<Props>()
  const { nodeElement, width, height } = useNodeResize(props.nodeId)

  function handleNodeClick() {
    // Handle node selection, focus, etc.
    // Should update node state without triggering resize
  }

  function handleContextMenu(event: MouseEvent) {
    // Handle right-click menu for node operations
    // Should prevent default browser context menu
    event.preventDefault()
  }
  </script>

  <template>
    <div
      ref="nodeElementRef"
      :data-node-id="nodeId"
      class="graph-node"
      @click="handleNodeClick"
      @contextmenu="handleContextMenu"
    >
      <!-- Node Content -->
      <!-- Resize Handle -->
      <div class="resize-handle" />
    </div>
  </template>
```

arjansingh (commented):
So this was a fun problem, I rubber ducked with Claude for a bit on it and came up with the following suggestions:

1. I think a batching approach where we just flush updates every RAF could be a good middle ground. That way canvas isn't being set dirty too frequently.
2. I think we can simplify the code with vue-use stuff. I don't think it would be a perf hit but feel free to push back and explain if you think it is.

The sequence in a nutshell:

```mermaid
  sequenceDiagram
      participant User
      participant ResizeHandle
      participant useElementSize
      participant useLayoutBatching
      participant VueStore
      participant Canvas
      participant LiteGraph

      User->>ResizeHandle: Start resize drag
      ResizeHandle->>ResizeHandle: Modify DOM element size directly

      Note over useElementSize: Single source of truth - watches DOM changes
      loop During resize (60fps max)
          useElementSize->>useElementSize: Detect DOM size change
          useElementSize->>useLayoutBatching: batchSync(nodeId, {width, height})
          useLayoutBatching->>useLayoutBatching: pendingUpdates.set(nodeId, layout)

          alt First update this frame
              useLayoutBatching->>useLayoutBatching: Schedule RAF callback
          else Subsequent updates same frame
              useLayoutBatching->>useLayoutBatching: Just update pending map
          end
      end

      Note over useLayoutBatching: RAF executes once per frame maximum
      useLayoutBatching->>useLayoutBatching: processAllUpdates()
      useLayoutBatching->>VueStore: flushAllUpdates() - batch update reactive state
      useLayoutBatching->>Canvas: markCanvasDirty() - ONCE per frame
      Canvas->>LiteGraph: Schedule canvas redraw

      User->>ResizeHandle: End resize
      useElementSize->>useLayoutBatching: Final batchSync()
      useLayoutBatching->>VueStore: Final layout update
      useLayoutBatching->>Canvas: Final setDirty()
```

src/platform/workflow/persistence/composables/useWorkflowPersistence.ts
DrJKL: I think that might be heavier than we need, since litegraph already does batching in its draw.

christian-byrne:
Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/5936

PR #5596, opened 2025-09-15
[fix] add focus state and aria to select components

arjansingh:
## Summary

Focus states weren't working, especially when trying to use the selects with `TAB` indexes and your keyboard.

## Changes

1. Added focus states and cleared them with design. See below.
2. Added storybook stories 
3. Added aria labels since I was in there anyway.

## Screenshots

Demo of working focus states:

https://github.com/user-attachments/assets/a5fd1272-b86c-4f4e-bfbb-13aba3e81713

https://github.com/user-attachments/assets/d2476f53-eda8-4900-819b-110e5f4e1c8a

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5596-fix-add-focus-state-and-aria-to-select-components-26f6d73d365081a38faed2bdb64ee2e2) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/input/MultiSelect.accessibility.stories.ts
+++ b/src/components/input/MultiSelect.accessibility.stories.ts
@@ -1,0 +1,379 @@
+import type { Meta, StoryObj } from '@storybook/vue3-vite'
+import type { MultiSelectProps } from 'primevue/multiselect'
+import { ref } from 'vue'
+
+import MultiSelect from './MultiSelect.vue'
+
+// Combine our component props with PrimeVue MultiSelect props
+interface ExtendedProps extends Partial<MultiSelectProps> {
+  // Our custom props
+  label?: string
+  showSearchBox?: boolean
+  showSelectedCount?: boolean
+  showClearButton?: boolean
+  searchPlaceholder?: string
+  listMaxHeight?: string
+  popoverMinWidth?: string
+  popoverMaxWidth?: string
+  // Override modelValue type to match our Option type
+  modelValue?: Array<{ name: string; value: string }>
+}
+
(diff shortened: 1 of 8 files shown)


src/components/input/MultiSelect.vue
```diff
@@ -170,6 +178,7 @@ const pt = computed(() => ({
       selectedCount.value > 0
         ? 'border-blue-400 dark-theme:border-blue-500'
         : 'border-transparent',
+      'focus-within:border-blue-400 dark-theme:focus-within:border-blue-500',
```
DrJKL, line 182: The border color is being set in two places. It makes sense that focus-within would take precedence.
Want to wrap this in a `cn()` to make sure the options merge in order?

src/components/input/MultiSelect.accessibility.stories.ts
```diff
@@ -0,0 +1,379 @@
...
+    setup() {
+      const selectedFrameworks = ref([])
+      const searchQuery = ref('')
+
+      return {
+        args: {
+          ...args,
+          options: frameworkOptions,
+          modelValue: selectedFrameworks,
+          'onUpdate:modelValue': (value: any) => {
```
DrJKL: But what is it?

DrJKL (dismissed):
❤️

src/components/input/MultiSelect.vue
DrJKL: Actually, it's a little weird that we'd have focus states on both the trigger and the menu item...

DrJKL (approved):

christian-byrne (approved):

PR #5598, opened 2025-09-15
[refactor] create src/platform/assets

arjansingh:
## Summary

Per @christian-byrne's feedback. Just bringing this into the repo sooner to clean up my feature branch.

## Changes

1. creates `/src/platform/assets` directories
2. moves existing service and schema into this directory
3. fix any broken imports

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5598-refactor-create-src-platform-assets-26f6d73d3650811d8ce3f0c9b7992cd3) by [Unito](https://www.unito.io)


src/platform/assets/index.ts
DrJKL, line 1: I'd rather not have barrel files unless they're explicitly for external consumption.

src/platform/assets/services/assetService.ts
```diff
@@ -1,13 +1,14 @@
...
+
 import {
   type AssetResponse,
   type ModelFile,
   type ModelFolder,
   assetResponseSchema
-} from '@/schemas/assetSchema'
-import { api } from '@/scripts/api'
-import { useModelToNodeStore } from '@/stores/modelToNodeStore'
+} from '../schemas/assetSchema'
```
DrJKL: Can we add a rule to require using the tsconfig paths for imports?

src/platform/assets/index.ts
DrJKL: In which case I'd also want to enumerate which pieces we export.

DrJKL (changes requested):

DrJKL:
I love the move otherwise 😄

DrJKL (approved):

src/platform/assets/index.ts
christian-byrne: We shouldn't use them just for DX but defining public API for a domain folder with interfaces is a desirable direction, especially if eventually wanting to create packages, support desktop/cloud/browser in same repo, and generally grow team to a very large size. Definitely should be intentional and concerted. I.e., I agree.

christian-byrne (approved):

PR #5599, opened 2025-09-15
feat: Let mode changes trigger a re-render for Vue nodes

DrJKL:
## Summary

Instrument mode to trigger node data updates.

## Changes

- **What**: Makes bypass/mute work immediately on the Vue nodes

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5599-feat-Let-mode-changes-trigger-a-re-render-for-Vue-nodes-26f6d73d365081dca306dd4a1523dbb3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -806,7 +806,6 @@ export const useGraphNodeManager = (graph: LGraph): GraphNodeManager => {
               })
               break
             case 'mode':
-              console.log('Do the thing')
               vueNodeData.set(nodeId, {
                 ...currentData,
                 mode: typeof event.newValue === 'number' ? event.newValue : 0
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -789,19 +789,28 @@ export const useGraphNodeManager = (graph: LGraph): GraphNodeManager => {
         const currentData = vueNodeData.get(nodeId)
 
         if (currentData) {
-          if (event.property === 'title') {
-            vueNodeData.set(nodeId, {
-              ...currentData,
-              title: String(event.newValue)
-            })
-          } else if (event.property === 'flags.collapsed') {
-            vueNodeData.set(nodeId, {
-              ...currentData,
-              flags: {
-                ...currentData.flags,
-                collapsed: Boolean(event.newValue)
-              }
-            })
+          switch (event.property) {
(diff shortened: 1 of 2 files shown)


src/composables/graph/useGraphNodeManager.ts
```diff
@@ -789,19 +789,28 @@ export const useGraphNodeManager = (graph: LGraph): GraphNodeManager => {
...
+              vueNodeData.set(nodeId, {
+                ...currentData,
+                flags: {
+                  ...currentData.flags,
+                  collapsed: Boolean(event.newValue)
+                }
+              })
+              break
+            case 'mode':
+              console.log('Do the thing')
```
benceruleanlu: this doesn't seem very descriptive or helpful, WDYT about removing this line?

AustinMroz (dismissed):
Another great PR that solves the specific problem advertised, but uncovers a strange and related issue: If a single node is selected after swapping to vue mode, pressing the bypass keybind does not cause the state to toggle. The associated function in useCoreCommands isn't even called.

Excluding the missed log, everything else is functional and beautiful

benceruleanlu (approved):

benceruleanlu (changes requested):
Can this be updated as well?

https://github.com/Comfy-Org/ComfyUI_frontend/blob/2b572917565f3f78757f1a8d136623e20794e47d/src/lib/litegraph/test/LGraphNodeProperties.test.ts#L23-L30

src/composables/graph/useGraphNodeManager.ts
DrJKL: I am deeply ashamed.

benceruleanlu (approved):
I actually don't even know if that test is ran though, I had assumed litegraph tests were enabled some time ago, but if it's not complaining and CI is green then I think it should be fine for now until the tests are enabled later

DrJKL:
> Another great PR that solves the specific problem advertised, but uncovers a strange and related issue: If a single node is selected after swapping to vue mode, pressing the bypass keybind does not cause the state to toggle. The associated function in useCoreCommands isn't even called.
> 
> Excluding the missed log, everything else is functional and beautiful

Mind if I investigate that as a separate issue? I feel like it'll probably be a more involved fix (and maybe cover a few more scenarios)

AustinMroz (approved):
That was my intent, yes. The code here is good.

christian-byrne (approved):

PR #5166, opened 2025-08-22
[feat] Merge ComfyUI_devtools into ComfyUI_frontend

snomiao:
## Summary
Merges ComfyUI devtools components into the ComfyUI frontend monorepo to consolidate development tools.

## Changes
- Added devtools components from ComfyUI repository
- Integrated development nodes and utilities
- Consolidated fake model assets for testing

## Related Issues
Fixes #4683

## Testing
- Devtools components are now available within the frontend monorepo
- Development workflow remains consistent

PR #5600, opened 2025-09-15
ci(tests-ui): unpin ComfyUI_devtools in test-ui

marawan206:
Pins ComfyUI_devtools to $SHA so Playwright runs against the latest devtools test nodes.
Updates all workflow locations that checkout devtools.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5600-ci-tests-ui-pin-ComfyUI_devtools-to-a92e-26f6d73d3650816780e5e5d09ebea116) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/i18n-custom-nodes.yaml
+++ b/.github/workflows/i18n-custom-nodes.yaml
@@ -37,6 +37,7 @@ jobs:
       with:
         repository: Comfy-Org/ComfyUI_devtools
         path: ComfyUI/custom_nodes/ComfyUI_devtools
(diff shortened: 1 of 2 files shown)


DrJKL:
If you don't set ref, shouldn't it always use the latest commit of the default branch?

christian-byrne:
The reason we set the SHA is to allow changing devtools without breaking CI. It also means you can run it on RC branches with different versions of the devtools.

DrJKL:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/5166

PR #5601, opened 2025-09-15
Line Selection toolbox up with Vue Nodes

Myestery:
This pull request improves the selection toolbox behavior during node dragging by ensuring that it correctly responds to both LiteGraph and Vue node drag events. The main changes introduce a reactive drag state for Vue nodes in the layout store and update the selection toolbox composable and Vue node component to use this state.

**Selection toolbox behavior improvements:**

* Added a helper function and separate watchers in `useSelectionToolboxPosition.ts` to hide the selection toolbox when either LiteGraph or Vue nodes are being dragged. This ensures consistent UI feedback regardless of node type. [[1]](diffhunk://#diff-57a51ac5e656e64ae7fd276d71b115058631621755de33b1eb8e8a4731d48713L171-R172) [[2]](diffhunk://#diff-57a51ac5e656e64ae7fd276d71b115058631621755de33b1eb8e8a4731d48713R212-R224)

**Vue node drag state management:**

* Added a reactive `isDraggingVueNodes` property to the `LayoutStoreImpl` class, along with getter and setter methods to manage Vue node drag state. This allows other components to reactively track when Vue nodes are being dragged. [[1]](diffhunk://#diff-80d32fe0fb72730c16cf7259adef8b20732ff214df240b1d39ae516737beaf3bR133-R135) [[2]](diffhunk://#diff-80d32fe0fb72730c16cf7259adef8b20732ff214df240b1d39ae516737beaf3bR354-R367)
* Updated `LGraphNode.vue` to set and clear the Vue node dragging state in the layout store during pointer down and up events, ensuring the selection toolbox is hidden while dragging Vue nodes. [[1]](diffhunk://#diff-a7744614cf842e54416047326db79ad81f7c7ab7bfb66ae2b46f5c73ac7d47f2R357-R360) [[2]](diffhunk://#diff-a7744614cf842e54416047326db79ad81f7c7ab7bfb66ae2b46f5c73ac7d47f2R376-R378)

**Dependency updates:**

* Imported the `layoutStore` in `LGraphNode.vue` to access the new drag state management methods.
* Added missing `ref` import in `layoutStore.ts` to support the new reactive property.

https://github.com/user-attachments/assets/d6e9c15e-63b5-4de2-9688-ebbc6a3be545


src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -168,51 +168,60 @@ export function useSelectionToolboxPosition(
...
+  }
+
+  // Watch for LiteGraph dragging state
+  watch(
+    () => canvasStore.canvas?.state?.draggingItems,
+    (dragging) => handleDragStateChange(dragging || false)
   )
 
+  // Watch for Vue node dragging state when Vue nodes are enabled
+  if (shouldRenderVueNodes.value) {
```
Myestery: Removing this if statement will help this work when switching modes from canvas to vue nodes mode

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -168,51 +168,60 @@ export function useSelectionToolboxPosition(
...
+    const vueNodeDraggingState = layoutStore.getVueNodeDraggingState()
+    watch(vueNodeDraggingState, handleDragStateChange)
+  }
```
arjansingh, line 231: Could we unify to on watcher? Something like:
```
  watch(
    () => ({
      lite: canvasStore.canvas?.state?.draggingItems ?? false,
      vue: layoutStore.getVueNodeDraggingState().value
    }),
    ({ lite, vue }) => handleDragStateChange(lite || vue)
  )
```

I'm worried about race conditions from having multiple watchers.

arjansingh (changes requested):
Can you try unifying the watchers. That's my only concern. Watching LiteGraph and Vue separately.

src/composables/canvas/useSelectionToolboxPosition.ts
Myestery: Done ✅

arjansingh (dismissed):

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -168,50 +168,58 @@ export function useSelectionToolboxPosition(
...
+        }
+        moreOptionsWasOpenBeforeDrag = false
+      })
     }
-  )
+  }
+
+  // Unified dragging state - combines both LiteGraph and Vue node dragging
+  const isDragging = computed(() => {
+    const litegraphDragging = canvasStore.canvas?.state?.draggingItems || false
```
DrJKL:
```suggestion
    const litegraphDragging = canvasStore.canvas?.state?.draggingItems ?? false
```

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -168,50 +168,58 @@ export function useSelectionToolboxPosition(
...
+          moreOptionsRestorePending.value = false
+        }
+        moreOptionsWasOpenBeforeDrag = false
+      })
     }
-  )
+  }
+
+  // Unified dragging state - combines both LiteGraph and Vue node dragging
+  const isDragging = computed(() => {
```
DrJKL: To protect against the missing `.value` thing
```suggestion
  const isDragging = computed((): boolean => {
```

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -168,50 +168,58 @@ export function useSelectionToolboxPosition(
...
-          moreOptionsOpen.value = false
-          moreOptionsRestorePending.value = !!moreOptionsSelectionSignature
-          if (moreOptionsRestorePending.value) {
-            forceCloseMoreOptionsSignal.value++
-          } else {
-            moreOptionsWasOpenBeforeDrag = false
-          }
+  // Helper function to handle drag state changes (shared between LiteGraph and Vue nodes)
+  const handleDragStateChange = (dragging: boolean) => {
+    if (dragging) {
```
DrJKL, line 174: Could you refactor this to flatten it out?
Ideally (but maybe not feasibly) there would be a single point of assignment for each of the refs.
At the very least, guard clauses could help reduce the `else`s.

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -168,50 +168,58 @@ export function useSelectionToolboxPosition(
...
-
-          if (shouldRestore) {
-            restoreMoreOptionsSignal.value++
-          } else {
-            moreOptionsRestorePending.value = false
-          }
-          moreOptionsWasOpenBeforeDrag = false
-        })
+        moreOptionsRestorePending.value = false
+        moreOptionsWasOpenBeforeDrag = false
```
DrJKL: Why isn't this a ref?

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -348,6 +351,20 @@ class LayoutStoreImpl implements LayoutStore {
     return computed(() => this.version)
   }
 
+  /**
+   * Get Vue node dragging state for selection toolbox
+   */
+  getVueNodeDraggingState(): Ref<boolean> {
+    return this.isDraggingVueNodes
+  }
```
DrJKL: Why have a getter and setter instead of exposing the ref?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -368,6 +373,9 @@ const handlePointerUp = (event: PointerEvent) => {
   if (isDragging.value) {
     isDragging.value = false
     void endDrag(event)
+
+    // Clear Vue node dragging state for selection toolbox
+    layoutStore.setVueNodeDraggingState(false)
```
DrJKL: This feels dangerous. What happens if this doesn't get reset?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -353,6 +354,10 @@ const handlePointerDown = (event: PointerEvent) => {
 
   // Start drag using layout system
   isDragging.value = true
+
+  // Set Vue node dragging state for selection toolbox
+  layoutStore.setVueNodeDraggingState(true)
```
DrJKL: What's the difference between isDragging and layoutStore.isDraggingVueNodes?

Myestery: the former is for the node and I started reporting this state to the layoutStore.
Might be a good case for storeToRefs tho

src/composables/canvas/useSelectionToolboxPosition.ts
Myestery: its not part of the exports

src/renderer/extensions/vueNodes/components/LGraphNode.vue
Myestery: for now the only user is the selectiontoolbox
not resetting it will make the selectiontoolbox stay hidden

src/renderer/core/layout/store/layoutStore.ts
Myestery: Layoutstore is not a composable and it only made sense to just use setters here instead of directly mutating the state directly everywhere.

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -168,50 +168,77 @@ export function useSelectionToolboxPosition(
...
+      return
+    }
+
+    // Handle more options cleanup
+    const currentSig = buildSelectionSignature(canvasStore)
+    const selectionChanged = currentSig !== moreOptionsSelectionSignature
+
+    moreOptionsSelectionSignature = selectionChanged
+      ? null
+      : moreOptionsSelectionSignature
```
DrJKL: The boolean logic here and down on 199 is a little hard to follow.

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -168,50 +168,77 @@ export function useSelectionToolboxPosition(
...
+      }
+    })
+  }
+
+  // Unified dragging state - combines both LiteGraph and Vue node dragging
+  const isDragging = computed((): boolean => {
+    const litegraphDragging = canvasStore.canvas?.state?.draggingItems ?? false
+    const vueNodeDragging = shouldRenderVueNodes.value
+      ? layoutStore.getVueNodeDraggingState().value
+      : false
```
DrJKL:
```suggestion
    const vueNodeDragging = shouldRenderVueNodes.value
      && layoutStore.getVueNodeDraggingState().value
```

src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -168,50 +168,77 @@ export function useSelectionToolboxPosition(
...
+      const shouldRestore =
+        moreOptionsWasOpenBeforeDrag &&
+        visible.value &&
+        moreOptionsRestorePending.value &&
+        selectionMatches
+
+      // Single point of assignment for each ref
+      moreOptionsRestorePending.value = shouldRestore
+        ? moreOptionsRestorePending.value
+        : false
```
DrJKL:
```suggestion
      moreOptionsRestorePending.value = shouldRestore
        && moreOptionsRestorePending.value
```

src/renderer/core/layout/store/layoutStore.ts
DrJKL: It's a [store](https://pinia.vuejs.org/core-concepts/#Setup-Stores) though, right?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
DrJKL: And how would a user recover from that?

src/renderer/core/layout/store/layoutStore.ts
DrJKL: If we expose the raw ref in the getter anyway, why not just make isDraggingVueNodes a public readonly property?

Myestery: mutating global state directly is generally not a good idea

src/renderer/extensions/vueNodes/components/LGraphNode.vue
DrJKL: When would they have different values?
Could we eliminate the per-node one and instead of having a boolean ref have the node that's being dragged's ID as both an indicator that dragging is occurring and a source of truth for which node is being dragged?

Myestery: not really
we could be dragging one node and not the other
we dont want the dormant nodes to have a misleading state and using a map complicates this.

src/renderer/core/layout/store/layoutStore.ts
DrJKL: How is it any different than the setter here?

christian-byrne:
@Myestery There is a merge conflict

Myestery:
> @Myestery There is a merge conflict

on it now

christian-byrne (approved):

PR #5533, opened 2025-09-13
feat: enable verbatimModuleSyntax in TypeScript config

snomiao:
## Summary
- Enable `verbatimModuleSyntax` compiler option in TypeScript configuration
- Update all type imports to use explicit `import type` syntax
- This change will Improve tree-shaking and bundler compatibility

## Motivation
The `verbatimModuleSyntax` option ensures that type-only imports are explicitly marked with the `type` keyword. This:
- Makes import/export intentions clearer
- Improves tree-shaking by helping bundlers identify what can be safely removed
- Ensures better compatibility with modern bundlers
- Follows TypeScript best practices for module syntax

## Changes
- Added `"verbatimModuleSyntax": true` to `tsconfig.json`
- Updated another 48+ files to use explicit `import type` syntax for type-only imports
- No functional changes, only import/export syntax improvements

## Test Plan
- [x] TypeScript compilation passes
- [x] Build completes successfully  
- [x] Tests pass
- [ ] No runtime behavior changes

🤖 Generated with [Claude Code](https://claude.ai/code)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5533-feat-enable-verbatimModuleSyntax-in-TypeScript-config-26d6d73d36508190b424ef9b379b5130) by [Unito](https://www.unito.io)

PR #5604, opened 2025-09-16
[refactor] Solve circular dependency in LiteGraph module

snomiao:
## Summary
- Extracted LiteGraph singleton instance into a separate module to break circular dependency chains
- Improves module loading order and prevents initialization issues
- Maintains the same API and functionality

## Context
This PR addresses the **second circular dependency issue** mentioned in #5533:
- The circular import chain: `SubgraphNode → LGraphNode → litegraph.ts → SubgraphNode`
- When `litegraph.ts` exports modules, it triggers loading of `SubgraphNode` which tries to extend `LGraphNode`, but `LGraphNode` imports from `litegraph.ts` causing circular initialization

## Solution
By extracting the LiteGraph singleton into a separate `LiteGraphSingleton.ts` module, we break the circular dependency chain and ensure proper module initialization order.

## Changes
- Created new `LiteGraphSingleton.ts` module to hold the singleton instance
- Updated all imports across the codebase to reference the singleton from the new module
- No functional changes, purely a refactoring to improve module structure

## Test Plan
- [x] All existing tests pass
- [x] No runtime errors from circular dependencies
- [x] Module loading order is now deterministic

### DIFF
--- a/src/lib/litegraph/src/ContextMenu.ts
+++ b/src/lib/litegraph/src/ContextMenu.ts
@@ -1,9 +1,9 @@
+import { LiteGraphInternal } from './LiteGraphInternal'
 import type {
   ContextMenuDivElement,
   IContextMenuOptions,
   IContextMenuValue
 } from './interfaces'
-import { LiteGraph } from './litegraph'
 
 // TODO: Replace this pattern with something more modern.
 export interface ContextMenu<TValue = unknown> {
@@ -182,7 +182,7 @@ export class ContextMenu<TValue = unknown> {
     root.style.left = `${left}px`
     root.style.top = `${top}px`
 
-    if (LiteGraph.context_menu_scaling && options.scale) {
+    if (LiteGraphInternal.context_menu_scaling && options.scale) {
       root.style.transform = `scale(${Math.round(options.scale * 4) * 0.25})`
     }
   }
@@ -355,7 +355,7 @@ export class ContextMenu<TValue = unknown> {
       ) {
         ContextMenu.trigger(
           this.parentMenu.root,
-          `${LiteGraph.pointerevents_method}leave`,
+          `${LiteGraphInternal.pointerevents_method}leave`,
           e
         )
       }
--- a/src/lib/litegraph/src/LGraph.ts
+++ b/src/lib/litegraph/src/LGraph.ts
@@ -14,6 +14,7 @@ import { LGraphCanvas } from './LGraphCanvas'
 import { LGraphGroup } from './LGraphGroup'
 import { LGraphNode, type NodeId } from './LGraphNode'
 import { LLink, type LinkId } from './LLink'
+import { LiteGraphInternal } from './LiteGraphInternal'
 import { MapProxyHandler } from './MapProxyHandler'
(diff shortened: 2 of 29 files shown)


webfiltered (changes requested):
I might be misreading the PR - the current global `LiteGraph` object is _already_ a singleton.  Could you not just move that instead?  It would be a significantly smaller change.

snomiao:
> I might be misreading the PR - the current global `LiteGraph` object is _already_ a singleton. Could you not just move that instead? It would be a significantly smaller change.

Hey @webfiltered, You did not misreading!

I 've renamed it just wanna ensure we are importing direct from the module insteadof from barrel file litegraph.ts. Using a different name helps avoid any confusion between two export files in future.

I think better name is LiteGraphInternal now, just pushed new commit for it : D

webfiltered:
> > I might be misreading the PR - the current global `LiteGraph` object is _already_ a singleton. Could you not just move that instead? It would be a significantly smaller change.
> 
> Hey @webfiltered, You did not misreading!
> 
> I 've renamed it just wanna ensure we are importing direct from the module insteadof from barrel file litegraph.ts. Using a different name helps avoid any confusion between two export files in future.
> 
> I think better name is LiteGraphInternal now, just pushed new commit for it : D

I'm still a bit confused - the imports should always work so long as you don't mess with the barrel file and import from it first.  I balanced this a long while back, and messing with it is complex and frustrating to debug (especially with the ecosystem using the LiteGraph global var).

I just ran `fix-verbatim-module-syntax` npm package on another branch, and had CC convert the script to a vue version then ran it again.  It all seems to be working fine.

PTAL: https://github.com/Comfy-Org/ComfyUI_frontend/compare/verbatim-module-syntax?expand=1

webfiltered:
Argh ran the wrong tests.  Okay, I see the error you saw.  This is actually just an issue with the test files.

webfiltered:
Yeah, can fix this easily by doing type imports properly:

```ts
import { type ThingToImport } from '@/file'
```

This still imports the file when running tests.  We don't notice it in prod because it uses different systems (vite/bundle).

Should be:

```ts
import type { ThingToImport } from '@/file'
```

This type of import is only used for type information, pre-compile - doesn't exist at runtime.

snomiao:
> Argh ran the wrong tests. Okay, I see the error you saw. This is actually just an issue with the test files.

I think we can merge your solution as first step: 

> import { type ThingToImport } from '@/file'

and then gradually refactor in following PRs into:

> import type { ThingToImport } from '@/file'

to make things work smoothy, WDYT? @webfiltered @DrJKL  @christian-byrne

webfiltered:
I think if we're going to do this change and touch all the imports, we may as well just do it all at once.  I would not want to keep having to sanity check this repeatedly.

Pretty sure there's a linter that can auto-fix the imports.  Failing that, we can just stuff the `ast-grep` I used into the pre-commit - it was very fast.

If there are more issues in future, it's just from direct importing.  Runtime imports should be via the barrel, not direct importing from internal LiteGraph files.

snomiao:
Closed this as may not needed anymore since i18n locales workflow fixed by another approach #5775

PR #5605, opened 2025-09-16
Add desktop dialogs framework

webfiltered:
### Summary

Adds desktop dialog framework with data-driven dialog definitions.

### Changes

- Data-driven dialog structure in `desktopDialogs.ts`
- Dynamic dialog view component with i18n support
- Button action types: openUrl, close, cancel
- Button severity levels for styling (primary, secondary, danger, warn)
- Fallback invalid dialog for error handling
- i18n collection script updated for dialog strings

### Review focus

- Colors added to match Figma design specs (coral-red, neutral-550)

### Screenshots

When a corrupt environment is detected during server start:
image

If we make mistakes during dev (should NEVER be seen in prod):
image

### DIFF
--- a/src/assets/css/style.css
+++ b/src/assets/css/style.css
@@ -65,7 +65,6 @@
   --color-charcoal-600: #262729;
   --color-charcoal-700: #202121;
   --color-charcoal-800: #171718;
-  --color-charcoal-900: #61626C;
 
   --color-neutral-550: #636363;
 
--- a/src/assets/css/style.css
+++ b/src/assets/css/style.css
@@ -67,6 +67,8 @@
   --color-charcoal-800: #171718;
   --color-charcoal-900: #61626C;
 
+  --color-neutral-550: #636363;
+
   --color-stone-100: #444444;
   --color-stone-200: #828282;
   --color-stone-300: #bbbbbb;
--- a/src/assets/css/style.css
+++ b/src/assets/css/style.css
@@ -104,9 +104,9 @@
   --color-danger-200: #d62952;
   --color-coral-red: #f14352;
 
-  --color-coral-red-50: rgb(from var(--color-coral-red) r g b / 0.5);
-  --color-coral-red-75: rgb(from var(--color-coral-red) r g b / 0.75);
-  --color-coral-red-88: rgb(from var(--color-coral-red) r g b / 0.88);
+  --color-coral-red-600: #973a40;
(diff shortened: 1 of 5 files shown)


src/views/DesktopDialogView.vue
```diff
@@ -0,0 +1,72 @@
+<template>
+  <div
+    class="w-full h-full flex flex-col rounded-lg p-6 bg-[#2d2d2d] justify-between"
```
DrJKL: Missed a color ;-)

src/views/DesktopDialogView.vue
```diff
@@ -0,0 +1,72 @@
...
+  @apply bg-coral-red-600 border-0;
+}
+
+.p-button-danger:hover {
+  @apply bg-coral-red-500;
+}
+
+.p-button-danger:active {
+  @apply bg-coral-red-400;
+}
```
DrJKL: Nope 🤖

webfiltered: The border-0 needs to go, but how would you like this done?  I do not want to blast over the entire frontend, which is how I would normally go about setting severity colours (globally).

src/views/DesktopDialogView.vue
webfiltered: Danke - fixed in 55d92d0ce01fd14aacb66b2af31f454fb258da4e

DrJKL (dismissed):

src/assets/css/style.css
```diff
@@ -98,6 +100,11 @@
   --color-warning-200: #fcbf64;
   --color-danger-100: #c02323;
   --color-danger-200: #d62952;
+  --color-coral-red: #f14352;
```
DrJKL: Why no number here?

webfiltered: Ah, it evolved from something else. I'll fix the ones I'm using; we can add a full range if it's desirable later.

webfiltered: Fixed in 46f4126711ba829306ca21a701c5c26986b74b94

src/views/DesktopDialogView.vue
webfiltered: Discussed offline:
- This is not the frontend pattern
- It is attempting to emulate a second `style.css`, as we cannot easily use one
- A temporary exception is being made until we either: unify all designs on the same buttons, or this just gets split out into a smaller, desktop-only section.

PR #5608, opened 2025-09-16
Verbatim module syntax

webfiltered:
## Summary

Similar to https://github.com/Comfy-Org/ComfyUI_frontend/pull/5533, but takes another approach, leaving the public objects and their instantiation untouched, but instead fixing the erroneous imports we have throughout the codebase.

All imports _with side effects_ (`import { type Thing }`) for type-only imports have been replaced with `import type { Thing }` syntax.

Not tested - this is just here as an example.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5608-Verbatim-module-syntax-2706d73d3650812d8e47f1e6cf4d2d3f) by [Unito](https://www.unito.io)


christian-byrne:
Is this one still necessary? Sorry didn't see it initially.

webfiltered:
Oh, no, this was not intended for merge sorry - was an example / discussion prop.

PR #5609, opened 2025-09-16
style: Flip the shade direction for charoal to go from light to dark

DrJKL:
## Summary

Should be invisible to users.

If you're curious, you can see the pattern in tailwind's default colors [here](https://tailwindcss.com/docs/colors)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5609-style-Flip-the-shade-direction-for-charoal-to-go-from-light-to-dark-2706d73d365081abb7aacc9b7cb15039) by [Unito](https://www.unito.io)

### DIFF
--- a/src/assets/css/style.css
+++ b/src/assets/css/style.css
@@ -57,14 +57,14 @@
   --font-inter: 'Inter', sans-serif;
 
   /* Palette Colors */
-  --color-charcoal-100: #171718;
-  --color-charcoal-200: #202121;
-  --color-charcoal-300: #262729;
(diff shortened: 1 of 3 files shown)


webfiltered (commented):
Tailwind range for built-in colours appears to be 0-1000.  Should we adopt that for charcoal for consistency?  Make 800 <-> 200?

arjansingh (dismissed):

AustinMroz (approved):

webfiltered (approved):

PR #5612, opened 2025-09-16
Add hook to expose uid

christian-byrne:
## Summary

Add hook to expose anonymized uid

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5612-Add-hook-to-expose-uid-2706d73d36508132aa18e81505bd8577) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/auth/useCurrentUser.ts
+++ b/src/composables/auth/useCurrentUser.ts
@@ -1,4 +1,4 @@
-import { computed } from 'vue'
+import { computed, watch } from 'vue'
 
 import { useFirebaseAuthActions } from '@/composables/auth/useFirebaseAuthActions'
 import { t } from '@/i18n'
@@ -6,6 +6,7 @@ import { useDialogService } from '@/services/dialogService'
 import { useApiKeyAuthStore } from '@/stores/apiKeyAuthStore'
 import { useCommandStore } from '@/stores/commandStore'
 import { useFirebaseAuthStore } from '@/stores/firebaseAuthStore'
+import type { AuthUserInfo } from '@/types/authTypes'
 
 export const useCurrentUser = () => {
   const authStore = useFirebaseAuthStore()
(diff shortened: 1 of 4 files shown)


src/composables/auth/useCurrentUser.ts
```diff
@@ -20,6 +21,32 @@ export const useCurrentUser = () => {
...
+  const onUserResolved = (callback: (user: AuthUserInfo) => void) => {
+    if (resolvedUserInfo.value) {
+      callback(resolvedUserInfo.value)
+    }
+
+    const stop = watch(resolvedUserInfo, (value) => {
+      if (value) {
+        callback(value)
+      }
+    })
```
DrJKL: This could use [whenever](https://vueuse.org/shared/whenever/)

DrJKL (dismissed):
Clean and simple, love it.

src/composables/auth/useCurrentUser.ts
christian-byrne: Good point: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5612/commits/b76a9adfb0d0bbe2480b56d84936b5527f0a4480. 4 lines to 1, clean.

src/composables/auth/useCurrentUser.ts
```diff
@@ -20,6 +22,27 @@ export const useCurrentUser = () => {
...
+    return null
+  })
+
+  const onUserResolved = (callback: (user: AuthUserInfo) => void) => {
+    if (resolvedUserInfo.value) {
+      callback(resolvedUserInfo.value)
+    }
+
+    const stop = whenever(resolvedUserInfo, callback)
+    return () => stop()
```
christian-byrne, line 43: I think this entire block is equivalent to

```typescript
return whenever(resolvedUserInfo, callback, { immediate: true})
```

PR #5614, opened 2025-09-17
Move VueFire persistence configuration to initialization

christian-byrne:
Currently, we set persistence method in the auth store setup. This creates pattern of using the default on init (indexed DB) up until the firebase store is initialized and `setPersistence` is called. For devices that don't support indexed DB or have the connection aggresively terminated or cleared, like [Safari](https://comfy-org.sentry.io/issues/6879071102/?project=4509681221369857&query=is%3Aunresolved&referrer=issue-stream), this can create problems with maintaing auth persistence.

Fix by setting persistence method in the initialization in main.ts

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5614-Move-VueFire-persistence-configuration-to-initialization-2716d73d3650817480e0c8feb1f37b9a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -2,6 +2,11 @@ import { definePreset } from '@primevue/themes'
 import Aura from '@primevue/themes/aura'
 import * as Sentry from '@sentry/vue'
 import { initializeApp } from 'firebase/app'
+import {
+  browserLocalPersistence,
+  browserSessionPersistence,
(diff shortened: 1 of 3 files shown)


DrJKL (dismissed):
🔥

DrJKL:
Change detector tests detected changes.

DrJKL (approved):

PR #5616, opened 2025-09-17
Lint: Turn on rules that should allow for verbatimModuleSyntax

DrJKL:
## Summary

Sorts out type only vs value imports.
- https://typescript-eslint.io/rules/consistent-type-imports
- https://typescript-eslint.io/rules/no-import-type-side-effects

## Changes

- **What**: Rules are now on with the default options, existing violations were all auto-fixable.

## Review Focus

See if I missed anything that `--fix` might have gotten wrong.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5616-Lint-Turn-on-rules-that-should-allow-for-verbatimModuleSyntax-2716d73d3650818d82b6de0cadf44ea1) by [Unito](https://www.unito.io)


christian-byrne:
There seems to be merge conflicts.

christian-byrne (approved):

PR #5617, opened 2025-09-17
Subgraph widget promotion - Part 2

AustinMroz:
Implements proxyWidget support on subgraph nodes.
This registers a special proxyWidgets property on subgraph nodes which is directly mapped to the proxyWidgets displayed on the node ~~with no duplication of state~~. Each proxyWidget directly maps to a real widget inside the subgraph.

Areas of concern
- The proxyWidgets property is serialized as a string ~~on every access~~. Litegraph fully supports leaving this as a list, but this is safer than risking breakage by changing types, is 3-4 orders of magnitude less important than the performance of the proxying code,  and provides a chance to slot in validation
- ~~The code is applied by patching `SubgraphNode._internalConfigureAfterSlots`. SubgraphNodes are a little jank because they sometimes are `configured` twice on initialization and this call deletes all widgets on the node each time. As there is no duplication of state, if the code is applied after the first call, all widgets are cleared from the node on the second. `setTimeout` was previously utilized, but makes for a much uglier workaround.~~
  - Code now uses the standardised onConfigured callback.
- The proxying code itself. ~~Don't be fooled by the `any` cast.~~ Typescript doesn't seem to bother checking types on reflection and the degree of wrapping is rather heavy handed.
- ~~The location where this code is stored. Placing code in scripts isn't great, but it puts it adjacent to the similar DOMWidget code and is consistent with simple act of importing the code applying side effects. Placing the code inside litegraph itself was originally attempted, but non-viable. The code requires access to multiple stores, including the domWidgetstore in order to manage state.~~

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5617-Subgraph-widget-promotion-Part-2-2716d73d3650813d8621fefdce6ae518) by [Unito](https://www.unito.io)


christian-byrne (commented):
## Comprehensive Review: Proxy Widget Implementation

This PR implements a sophisticated proxy widget system that addresses core widget promotion challenges in ComfyUI. Having analyzed the widget ecosystem extensively, I can provide targeted feedback on both implementation and architectural integration.

## Architecture Assessment ✅

### Excellent Design Decisions

1. **Proxy Pattern Choice**: Using JavaScript Proxy for delegation is architecturally sound and aligns with the widget system's need for transparent value synchronization between source and promoted widgets.

2. **DOM Widget Store Integration**: The proper handling of DOM widget activation/deactivation (lines 35-52 in proxyWidget.ts) demonstrates understanding of the dual-layer widget architecture (canvas + DOM).

3. **Nested Subgraph Support**: The colon-separated nodeId resolution (line 99-102) elegantly handles deeply nested subgraphs without complex traversal logic.

4. **Type Safety with Zod**: Using Zod validation for the serialization format provides runtime safety for the string-based storage approach.

## Technical Implementation Strengths

### Property Descriptor Pattern
The custom getter/setter on properties.proxyWidgets is clever:
- Getter automatically maintains serialization sync
- Setter handles widget lifecycle management  
- Avoids manual synchronization issues

### Widget Resolution Logic
The resolveLinkedWidget function correctly handles the graph hierarchy traversal pattern needed for nested subgraphs, which was a key challenge identified in widget promotion analysis.

## Areas for Enhancement

### 1. Type Safety Improvements
Consider strengthening the Proxy handler types instead of using 'any' casts. The current implementation works but could benefit from more specific typing.

### 2. Performance Optimization Consideration
The string serialization on every getter access could be optimized with a dirty flag pattern to cache results when widgets haven't changed.

### 3. Widget Lifecycle Robustness
Consider adding widget position preservation similar to the DOM widget createCopyForNode pattern to maintain proper Y positioning after promotion.

## Integration with Existing Patterns ✅

This implementation aligns well with established widget system patterns:
- Follows the proxy pattern suggested in widget promotion analysis
- Integrates properly with DOM widget store lifecycle
- Maintains LiteGraph widget interface compatibility
- Handles the graph hierarchy navigation correctly

## Testing Coverage Assessment

The test suite covers core functionality well but could benefit from:
- DOM widget specific test cases
- Nested subgraph scenarios
- Widget positioning after promotion/demotion
- Performance testing for the serialization approach

## Overall Assessment

This is a well-architected solution that addresses core widget promotion challenges elegantly. The proxy pattern provides transparent delegation, and integration with existing systems is handled correctly. The string serialization trade-off is reasonable given complexity of alternatives.

**Recommendation: Solid implementation that demonstrates strong understanding of widget system architecture.**

christian-byrne (commented):
## Specific Implementation Feedback

### src/scripts/proxyWidget.ts

**Line 24-30 (Property Getter)**: The automatic serialization in the getter is elegant but consider caching:
```typescript
// Could add dirty flag to avoid re-serializing unchanged widgets
get: () => {
  if (\!this._cachedProxyWidgets || this._proxyWidgetsDirty) {
    this._cachedProxyWidgets = JSON.stringify(/* current logic */)
    this._proxyWidgetsDirty = false
  }
  return this._cachedProxyWidgets
}
```

**Line 109-123 (Proxy Handler)**: The reflection-based approach is solid. Consider documenting the delegation hierarchy:
```typescript
// Documents: overlay properties override, then linkedWidget, then disconnectedWidget
const handler = Object.fromEntries(
  ['get', 'set', 'getPrototypeOf', 'ownKeys', 'has'].map((s) => {
    const func = function (t: object, p: string, ...rest: object[]) {
      if (s == 'get' && p == '_overlay') return overlay
      // Priority: overlay > linkedWidget > disconnectedWidget
```

**Line 95-104 (resolveLinkedWidget)**: Excellent handling of nested subgraphs. The colon-separated ID traversal correctly implements the graph hierarchy pattern identified in widget promotion analysis.

**Line 56 (arrange call)**: Critical for DOM widget positioning\! This demonstrates understanding of the LiteGraph layout system distinction between computeSize() and arrange().

### src/schemas/proxyWidget.ts

**Line 6 (Schema Definition)**: The tuple schema [nodeId, widgetName] is minimal and efficient. Consider if additional metadata might be needed for future features (widget type, positioning, etc.).

### tests-ui/tests/widgets/proxyWidget.test.ts

**Line 83-95 (Value Mirroring Test)**: Excellent bidirectional sync testing. This validates the core proxy delegation mechanism.

**Missing Test Cases**: Consider adding:
- DOM widget activation/deactivation scenarios
- Nested subgraph resolution (nodeId with colons)
- Widget positioning preservation
- Error handling when linked widgets are removed

## Performance Considerations

The current implementation prioritizes correctness over performance, which is appropriate for initial implementation. The string serialization overhead is acceptable given the complexity savings versus maintaining separate state synchronization.

For future optimization, the dirty flag pattern could reduce serialization calls, but measure first to confirm it's actually a bottleneck.

christian-byrne (commented):
## Widget System Integration Analysis

### Alignment with Widget Architecture Patterns

This proxy widget implementation successfully addresses several key challenges identified in the ComfyUI widget system:

#### ✅ Solves Control Widget Promotion Issue
The proxy pattern with dynamic target resolution elegantly handles the control widget promotion problem where control widgets (seed randomization, etc.) need to target promoted widgets instead of their original inner widgets.

#### ✅ DOM Widget System Integration  
The proper handling of DOM widget store activation/deactivation (lines 35-52) shows understanding of the complex DOM widget lifecycle and the need for dual registration (canvas + DOM layers).

#### ✅ Widget Lifecycle Management
By using property descriptors and filtering proxy widgets during reconstruction, this avoids the widget array truncation anti-patterns that cause undefined access issues in extensions.

#### ✅ Graph Hierarchy Navigation
The nodeId resolution with colon separation correctly implements the graph traversal patterns needed for nested subgraphs without requiring complex BFS from root graph.

### Architectural Benefits

1. **No Reference Tracking**: Unlike previous approaches, this doesn't require updating widget references during promotion/demotion - the proxy resolves targets dynamically.

2. **Clean Separation of Concerns**: Proxy widgets remain on SubgraphNodes, source widgets stay on inner nodes, avoiding complex lifecycle coordination.

3. **Extension Compatibility**: Maintains LiteGraph widget interface, ensuring extensions that access widgets by name/type continue working.

4. **Memory Safety**: Avoids circular references and cleanup issues by using weak delegation rather than strong references.

### Future Considerations

This foundation enables several advanced features:
- Widget grouping and conditional visibility
- Enhanced validation patterns  
- Performance optimizations through caching
- Control widget targeting improvements

The implementation provides a solid architectural foundation that aligns well with established widget system patterns while solving core promotion challenges.

src/scripts/proxyWidget.ts
```diff
@@ -0,0 +1,127 @@
...
+  const proxyWidgets = subgraphNode.properties.proxyWidgets
+
+  //Takes no arguements, returns nothing
+  //Sometimes called multiple times on initialization, sometimes once
+  //Clobbers all widgets
+  originalConfigureAfterSlots?.bind(this)?.()
+
+  Object.defineProperty(subgraphNode.properties, 'proxyWidgets', {
+    get: () => {
+      const result = subgraphNode.widgets
```
christian-byrne: **Performance Consideration**: This getter serializes on every access. Consider caching with a dirty flag:
```typescript
let cachedSerialization: string | null = null
let isDirty = true

get: () => {
  if (isDirty || \!cachedSerialization) {
    cachedSerialization = JSON.stringify(result)
    isDirty = false
  }
  return cachedSerialization
}
```

src/scripts/proxyWidget.ts
```diff
@@ -0,0 +1,127 @@
...
+        .filter((w) => isProxyWidget(w))
+        .map((w) => [w._overlay.nodeId, w._overlay.widgetName])
+      return JSON.stringify(result)
```
christian-byrne, line 29: question: How difficult would it be to maintain a dirty flag and then only serialize when the setter marks it dirty? I'm not sure how much properties are accessed but this could be a a lot of serializations potentially?

src/scripts/proxyWidget.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+function resolveLinkedWidget(overlay: Overlay): IBaseWidget | undefined {
+  const { graph, nodeId, widgetName } = overlay
+  let g: LGraph | undefined = graph
+  let n: LGraphNode | SubgraphNode | undefined = undefined
+  for (const id of nodeId.split(':')) {
+    n = g?._nodes_by_id?.[id]
+    g = n?.isSubgraphNode?.() ? n.subgraph : undefined
+  }
+  if (!n) return undefined
```
christian-byrne: Could this use anything in the common graphTraversalUtils.ts or be added to that file if not? It seems it could be good to re-use this and therefore only have to maintain the logic in one place.

src/scripts/proxyWidget.ts
```diff
@@ -0,0 +1,127 @@
...
+        }
+        return (Reflect as any)[s](t, p, ...rest.slice(0, -1), r)
+      }
+      return [s, func]
+    })
+  )
+  const w = new Proxy(disconnectedWidget, handler)
+  subgraphNode.widgets.push(w)
+  return w
+}
```
christian-byrne, line 185: nit: Could we add some jsdoc comment and/or inline comments to this? I think future readers might struggle slightly to understand what is happening.

tests-ui/tests/widgets/proxyWidget.test.ts
christian-byrne, line 1: nit: Potential other test cases:

- nested subgraph resolution
- widget positioning preservation
- error handling when linked widgets are removed

christian-byrne (commented):
Could we move all the files to `src/core/graph/subgraph`

christian-byrne:
Sorry Claude review went haywire.

src/scripts/proxyWidget.ts
AustinMroz: I'll need to think on this one a little longer. My initial reaction was that it would require an onerous level of tracking to keep `node.widgets` as the single source of truth. 

But these proxyWidgets should only ever be managed through this property, and adding this caching layer cleanly solves the double configuration problem.

AustinMroz: Removing the code to calculate displayed widgets and simply using the previously set value is cleaner both in general, and by way of removing the ugly wrap of an internal function. Thanks.

src/scripts/proxyWidget.ts
AustinMroz: After further review. having a single comment wasn't sufficient for describing which parameters are relevant to which handlers, and splitting the handlers out into full individual functions wasn't as bad as I expected it to be.

☝️:nit? It seems like typescript does actual parsing of jsdoc comments to perform type inference, does using a single shared comment for all of them break anything?

src/scripts/proxyWidget.ts
AustinMroz: `getNodeByExecutionId` can be used, but involves passing a non-root graph in order to navigate a partial execution id, which isn't great

I've opted to remove the recursive parsing entirely here. It's not implemented on the display side and helps performance.

tests-ui/tests/widgets/proxyWidget.test.ts
AustinMroz: Added testing for the later two, skipping the first as it is no longer intended functionality.

christian-byrne:
The lint staged should pass if you rebase, since the knip config file was added to tsconfig `include` last week in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5662

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -0,0 +1,177 @@
...
+      this.widgets = this.widgets.filter((w) => !isProxyWidget(w))
+      for (const [nodeId, widgetName] of parsed) {
+        const w = addProxyWidget(this, `${nodeId}`, widgetName)
+        if (w instanceof DOMWidgetImpl) {
+          const widgetState = widgetStates.get(w.id)
+          if (!widgetState) continue
+          widgetState.active = true
+          widgetState.widget = w
+        }
+      }
```
christian-byrne, line 72: suggestion: Move store mutations logic to methods.

1. Currently the domWidget store exposes the `activateWidget` method. It may be beneficial to use that method instead of setting manually so that it will be easier to refactor or adjust the implementation details of what "activating" means later
2. To make point 1 work, we could also create a `setProxyWidget` or just `setWidget` method in domWidget store and then call it here, which would have same benefits as point 1.

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -0,0 +1,177 @@
...
+          widgetState.active = false
+        }
+      }
+      //NOTE: This does not apply to pushed entries, only initial load
+      this.widgets = this.widgets.filter((w) => !isProxyWidget(w))
+      for (const [nodeId, widgetName] of parsed) {
+        const w = addProxyWidget(this, `${nodeId}`, widgetName)
+        if (w instanceof DOMWidgetImpl) {
+          const widgetState = widgetStates.get(w.id)
+          if (!widgetState) continue
```
christian-byrne: question (non-blocking): What would it mean if the condition `if (!widgetState)` is true?

At this point in the code, we have already called `addProxyWidget`, so if it turns out `w` *is* a DOM widget but isn't in the `widgetStates` map, what does that mean for us? Do we need some sort of transaction logic or to move the check for this before we create the proxy?

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -0,0 +1,177 @@
...
+      const parsed = parseProxyWidgets(property)
+      const { widgetStates } = useDomWidgetStore()
+      for (const w of this.widgets.filter((w) => isProxyWidget(w))) {
+        if (w instanceof DOMWidgetImpl && widgetStates.has(w.id)) {
+          const widgetState = widgetStates.get(w.id)
+          if (!widgetState) continue
+          widgetState.active = false
+        }
+      }
+      //NOTE: This does not apply to pushed entries, only initial load
```
christian-byrne: question (non-blocking): Is it important for us to gaurantee it only applies on initial load?

Currently it also applies whenever the property is set right?

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -0,0 +1,177 @@
...
+      }
+      proxyWidgets = property
+      canvasStore.canvas?.setDirty(true, true)
+      this._setConcreteSlots()
+      this.arrange()
+    }
+  })
+  this.properties.proxyWidgets = proxyWidgets
+}
+type Overlay = Partial<IBaseWidget> & {
```
christian-byrne: nit: move types to top of file.

The types describe the shape of data and function signatures. Putting them up front makes it clear what the rest of the file is working with and aligns with current codebase styles.

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -0,0 +1,177 @@
...
+  graph: LGraph
+  nodeId: string
+  widgetName: string
+  isProxyWidget: boolean
+  node?: LGraphNode
+}
+type ProxyWidget = IBaseWidget & { _overlay: Overlay }
```
christian-byrne: suggestion: Add jsdoc comment onto the types.

Adding a jsdoc comment on the type can be beneficial as it allows other developers to hover the types and see the annotation in the future.

src/core/schemas/proxyWidget.ts
```diff
@@ -0,0 +1,23 @@
...
+import type { NodeProperty } from '@/lib/litegraph/src/LGraphNode'
+
+const proxyWidgetsPropertySchema = z.array(z.tuple([z.string(), z.string()]))
+type ProxyWidgetsProperty = z.infer<typeof proxyWidgetsPropertySchema>
+
+export function parseProxyWidgets(
+  property: NodeProperty | undefined
+): ProxyWidgetsProperty {
+  if (typeof property !== 'string') {
+    console.error(`Found non-string value for properties.proxyWidgets`)
```
christian-byrne: suggestion (non-blocking): We could add more context here about what this entails or why it is happening.

src/core/schemas/proxyWidget.ts
```diff
@@ -0,0 +1,23 @@
...
+    return []
+  }
+  const parsed = JSON.parse(property)
```
christian-byrne, line 17: question: Do we want this to throw if the parse fails? Or handle gracefully?

src/core/graph/subgraph/proxyWidget.ts
AustinMroz: While the check was added to satisfy type guards and this *shouldn't* ever come up, it does make a good question. Particularly for the subsequent "set active" code.

A DOMWidget that somehow doesn't exist in the store can't be displayed. Ignoring a widget was fine previously when the getter was calculating from the actual widgets, but now it causes a desync of state. 🤔

Will think on it a bit more, but I'm ever-so-slightly leaning towards throwing an error. If this ever comes up in the future, we'd want to be notified either when testing or through sentry.

src/core/graph/subgraph/proxyWidget.ts
AustinMroz: :+1: Comment should be removed since it's no longer relevant and is now confusing.

When the property was an actual array instead of a string, calling `node.properties.proxyWidgets.push()` would not actually add any proxy widgets because it wouldn't trigger the setter code.

src/core/graph/subgraph/proxyWidget.ts
AustinMroz: Keeping things symmetric was a concern, but point 2 satisfies this.

src/core/schemas/proxyWidget.ts
AustinMroz: I agree, throw would be better here. The current code actually results in the invalid value being set, which is bad.

src/core/schemas/proxyWidget.ts
AustinMroz: I'm not a ware of a non-programmatic way to set a property to a non-string value, but it doesn't hurt.

src/scripts/proxyWidget.ts
christian-byrne: > It seems like typescript does actual parsing of jsdoc comments to perform type inference

What makes you say that?

AustinMroz: [The JSDoc reference page](https://www.typescriptlang.org/docs/handbook/jsdoc-supported-types.html#param-and-returns) completely lacks normal typescript annotations. Under the configuration of this project, not including any annotations results in typecheck errors. This, in tandem with the strongly defined format of JSDoc gave me the types were being inferred from the jsdoc instead and that the example was intentionally leaving of types to indicate this.

I have taken the time to test this thought and am now confident that it is not the case.

src/core/graph/subgraph/proxyWidget.ts
AustinMroz: The other reason I had forgotten was that the earlier changes I made to scripts/app.ts iterate over widgetStates for improved performance.

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -0,0 +1,189 @@
...
+              useNodeOutputStore().getNodeOutputs(linkedNode)?.images ?? []
+            if (images !== linkedNode.images) {
+              linkedNode.images = images
+              useNodeImage(linkedNode).showPreview()
+            }
+            return linkedNode.imgs
+          }
+          return []
+        }
+        return Reflect.get(subgraphNode, p)
```
arjansingh: nit: could we invert the logic an use guard clauses. feels especially appropriate for all this nesting.

arjansingh (commented):
Only got a nit.

src/core/graph/subgraph/proxyWidget.ts
AustinMroz: Yeah. That's good.

christian-byrne (approved):
LGTM

benceruleanlu:
LGTM!

PR #5620, opened 2025-09-17
fix: Add validation and consistent data structure for survey 'Other' options

viva-jinyi:
## Summary
- Added validation to prevent users from proceeding when "Other" is selected but text input is empty
- Changed data structure to send consistent string values to database instead of mixed objects
- Both useCase and industry now send user input directly when "Other" is selected

## Changes
- Added `useCaseOther` field and input to survey form  
- Updated `validStep2` to validate useCaseOther when useCase is 'other'
- Modified submit payload to send string values consistently for both useCase and industry fields

## Test plan
- [x] Select "Other" for use case without filling input → Next button disabled
- [x] Select "Other" for industry without filling input → Next button disabled  
- [x] Fill in "Other" text inputs → Next button enabled
- [x] Submit survey with "Other" selections → Payload sends user input as string values

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5620-fix-Add-validation-and-consistent-data-structure-for-survey-Other-options-2716d73d3650811faa6efc547db14930) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/onboarding/cloud/CloudSurveyView.vue
+++ b/src/platform/onboarding/cloud/CloudSurveyView.vue
@@ -80,6 +80,13 @@
                 >
               </div>
             </div>
+            <div v-if="surveyData.useCase === 'other'" class="mt-4 ml-8">
+              <InputText
+                v-model="surveyData.useCaseOther"
+                class="w-full"
+                placeholder="Please specify"
+              />
+            </div>
           </div>
 
           <div class="flex gap-6 pt-4">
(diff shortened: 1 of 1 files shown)


src/platform/onboarding/cloud/CloudSurveyView.vue
```diff
@@ -80,6 +80,13 @@
                 >
               </div>
             </div>
+            <div v-if="surveyData.useCase === 'other'" class="mt-4 ml-8">
+              <InputText
+                v-model="surveyData.useCaseOther"
+                class="w-full"
+                placeholder="Please specify"
```
DrJKL, line 87: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5625
Once this goes in, the placeholder will be flagged as something to translate.

DrJKL (approved):
Can they only pick one use case?
If it ends up being multiselect this logic will probably need to change.

PR #5567, opened 2025-09-15
[perf] Remove legacy Float32Array usage from LiteGraph

christian-byrne:
## Summary

Removes Float32Array usage throughout LiteGraph codebase, eliminating type conversion overhead for Canvas 2D rendering operations.

- Convert position/size storage from Float32Array to regular arrays
- Replace complex malloc patterns with standard properties  
- Update type system to support both regular arrays and typed arrays
- Maintain full API compatibility with zero breaking changes

## Background

Float32Array usage was originally designed for WebGL integration, but the current Canvas 2D rendering pipeline doesn't use WebGL, making the typed arrays an unnecessary performance overhead. Every Canvas 2D operation that reads coordinates from Float32Array incurs type coercion costs (pad/truncate to/from native 64 bit number format).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5567-perf-Remove-legacy-Float32Array-usage-from-LiteGraph-26f6d73d365081aeb02ccd296f2ca29f) by [Unito](https://www.unito.io)

PR #5628, opened 2025-09-17
Slot functionality for vue nodes

benceruleanlu:
Allows for simple slot functionality in vue nodes mode.

Has:
- Drag new link from slot
- Connect new link from dropping on slot

Now:
- Tests

After:
- Drop on reroute
- Correct link color on connect
- Drop on node
- Hover effects

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5628-Slot-functionality-for-vue-nodes-2716d73d365081c59a3cef7c8a5e539e) by [Unito](https://www.unito.io)


src/renderer/core/canvas/litegraph/litegraphLinkAdapter.ts
```diff
@@ -502,57 +498,33 @@ export class LitegraphLinkAdapter {
     }
   }
 
-  /**
-   * Render a link being dragged from a slot to mouse position
-   * Used during link creation/reconnection
-   */
   renderDraggingLink(
```
benceruleanlu, line 500: renderDraggingLink was actually unused before this, so the removed logic is irrelevant

this change just converts it into a wrapper with the shared defaults/overrides that both litegraph and vue use to render dragging links

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
```diff
@@ -0,0 +1,96 @@
...
+} from '@/lib/litegraph/src/interfaces'
+import { LinkDirection } from '@/lib/litegraph/src/types/globalEnums'
+import { resolveConnectingLinkColor } from '@/lib/litegraph/src/utils/linkColors'
+import type { LinkRenderContext } from '@/renderer/core/canvas/litegraph/litegraphLinkAdapter'
+import {
+  type SlotDragSource,
+  useSlotLinkDragState
+} from '@/renderer/core/linkInteractions/slotLinkDragState'
+
+function buildContext(canvas: LGraphCanvas): LinkRenderContext {
```
benceruleanlu: I do agree that we could potentially have a system to replace these defaults, although that should be its own PR

src/renderer/core/layout/sync/useLinkLayoutSync.ts
```diff
@@ -93,8 +93,8 @@ export function useLinkLayoutSync() {
     const endPos = getSlotPosition(targetNode, link.target_slot, true)
 
     // Get directions
-    const startDir = sourceSlot.dir || LinkDirection.RIGHT
-    const endDir = targetSlot.dir || LinkDirection.LEFT
+    const startDir = sourceSlot.dir ?? LinkDirection.RIGHT
+    const endDir = targetSlot.dir ?? LinkDirection.LEFT
```
benceruleanlu: so the story goes that NONE was improved in the codebase, and started to have usages

the issue is that because LinkDirection is a numerical enum where NONE is 0, || will actually stomp NONE into the default in some parts of the codebase

however, NONE and CENTER have the same link rendering function in the pipeline, save for this ||?? issue

thus, the change from || to ?? is to make this more intentional and safe, making NONE explicitly NONE, and allowing both NONE and CENTER to exist as seemingly duplicates, with existing code to allow for an undefined dir to mean the default

I think this issue never arose because NONE was never used (unconfirmed assumption)? given that NONE and CENTER have the same functionality

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -184,6 +183,8 @@ export function useSlotElementTracking(options: {
 
         // Register slot
         const slotKey = getSlotKey(nodeId, index, type === 'input')
+
+        el.dataset.slotKey = slotKey
```
benceruleanlu, line 187: this is for the hit testing

src/renderer/core/layout/sync/useLinkLayoutSync.ts
benceruleanlu: although this could possibly be updated in the future to make dir required and NONE the default

benceruleanlu: clearly that day is today, so:

undefined: same as NONE
NONE (0): use defaults, literally unset or NONE
...
CENTER (5): no offset, do NOT use defaults

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
```diff
@@ -0,0 +1,83 @@
...
+
+interface CompatibilityResult {
+  allowable: boolean
+  targetNode?: LGraphNode
+  targetSlot?: INodeInputSlot | INodeOutputSlot
+}
+
+function resolveNode(nodeId: string | number) {
+  const canvas = app.canvas
+  const graph = canvas?.graph
```
christian-byrne: nit: don't need to change this now but it may be better to get canvas from canvasStore. Current graph could be retrieved from a dedicated current-graph-getter that will later be replaced per design in https://github.com/Comfy-Org/ComfyUI_frontend/issues/4661

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
```diff
@@ -0,0 +1,83 @@
...
+  if (Number.isNaN(id)) return null
+  return graph.getNodeById(id)
+}
+
+export function evaluateCompatibility(
+  source: SlotDragSource,
+  candidate: SlotDropCandidate
+): CompatibilityResult {
+  if (
+    candidate.layout.nodeId === source.nodeId &&
```
christian-byrne: Shouldn't just this condition alone make it false? That is, you can't connect create a self-loop as we are a directed simple graph.

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
```diff
@@ -0,0 +1,96 @@
...
+    renderBorder: canvas.render_connections_border,
+    lowQuality: canvas.low_quality,
+    highQualityRender: canvas.highquality_render,
+    scale: canvas.ds.scale,
+    linkMarkerShape: canvas.linkMarkerShape,
+    renderConnectionArrows: canvas.render_connection_arrows,
+    highlightedLinks: new Set(Object.keys(canvas.highlighted_links)),
+    defaultLinkColor: canvas.default_link_color,
+    linkTypeColors: (canvas.constructor as typeof LGraphCanvas)
+      .link_type_colors,
```
christian-byrne: nit: is there a better way to access these?

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
```diff
@@ -0,0 +1,96 @@
...
+    disabledPattern: canvas._pattern
+  }
+}
+
+export function attachSlotLinkPreviewRenderer(canvas: LGraphCanvas) {
+  const originalOnRender = canvas.onRender?.bind(canvas)
```
christian-byrne: Is there a solution that doesn't invovle monkey-patching our own functions? We have the freedom to change the litegraph code here.

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
```diff
@@ -0,0 +1,96 @@
...
+  source: SlotDragSource
+): INodeInputSlot | INodeOutputSlot | undefined {
+  const graph = canvas.graph
+  if (!graph) return undefined
+
+  const nodeId = Number(source.nodeId)
+  if (!Number.isFinite(nodeId)) return undefined
+
+  const node = graph.getNodeById(nodeId)
+  if (!node) return undefined
```
christian-byrne: nit: This seems to overlap a lot with the `resolveNode` function

src/renderer/extensions/vueNodes/components/InputOutputSlot.test.ts
christian-byrne, line 1: Thoughts on replacement test?

src/renderer/core/canvas/litegraph/litegraphLinkAdapter.ts
DrJKL: No custom nodes use it?

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
```diff
@@ -0,0 +1,83 @@
...
+} from '@/renderer/core/linkInteractions/slotLinkDragState'
+import { app } from '@/scripts/app'
+
+interface CompatibilityResult {
+  allowable: boolean
+  targetNode?: LGraphNode
+  targetSlot?: INodeInputSlot | INodeOutputSlot
+}
+
+function resolveNode(nodeId: string | number) {
```
DrJKL, line 18: nit: not sure how @christian-byrne feels, but can we use the NodeId type here?

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -0,0 +1,206 @@
...
+      sourceNode.connectSlots(outputSlot, targetNode, inputSlot, undefined)
+      return
+    }
+
+    if (source.type === 'input' && slotLayout.type === 'output') {
+      const inputSlot = sourceSlot as INodeInputSlot | undefined
+      const outputSlot = targetSlot as INodeOutputSlot | undefined
+      if (!inputSlot || !outputSlot) return
+      graph.beforeChange()
+      sourceNode.disconnectInput(source.slotIndex, true)
```
christian-byrne, line 154: Do we need to handle this erroring and not proceed if so?

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -0,0 +1,206 @@
...
+      targetNode.connectSlots(outputSlot, sourceNode, inputSlot, undefined)
```
christian-byrne, line 155: questino: So the validation happens inside of `connectSlots` or it has already happened at this point?

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
```diff
@@ -0,0 +1,83 @@
...
+  const canvas = app.canvas
+  const graph = canvas?.graph
+  if (!graph) return null
+  const id = typeof nodeId === 'string' ? Number(nodeId) : nodeId
+  if (Number.isNaN(id)) return null
```
christian-byrne: Greater than 0 as well?

src/renderer/core/linkInteractions/slotLinkDragState.ts
```diff
@@ -0,0 +1,89 @@
...
+  pointer: defaultPointer,
+  candidate: null
+})
+
+function updatePointerPosition(position: PointerPosition) {
+  state.pointer = shallowReactive({
+    client: position.client,
+    canvas: position.canvas
+  })
+}
```
christian-byrne: We are creating a new reactive object on every pointer move?

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    return candidate
+  }
+
+  const conversion = useSharedCanvasPositionConversion()
+
+  let activePointerId: number | null = null
+
+  const cleanupListeners = () => {
```
christian-byrne: How about using `useEventListener` to handle cleanup?

christian-byrne (changes requested):
How about centralizing all link rendering in something like `src/renderer/core/canvas/links`

src/renderer/core/linkInteractions/slotLinkDragState.ts
DrJKL, line 1: I'm actually extracting a similar composable from LGraphNode in the Notes work.
This is probably worth generalizing.

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (!outputSlot || !inputSlot) {
+      return { allowable: false }
+    }
+
+    const allowable = sourceNode.canConnectTo(targetNode, inputSlot, outputSlot)
+    return { allowable, targetNode, targetSlot: inputSlot }
+  }
+
+  const inputSlot = sourceSlot as INodeInputSlot | undefined
+  const outputSlot = targetSlot as INodeOutputSlot | undefined
```
DrJKL, line 76: Why do you need all the type assertions?

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
DrJKL: Is it static on the class?

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
```diff
@@ -0,0 +1,96 @@
...
+    const start = source.position
+    const sourceSlot = resolveSourceSlot(canvas, source)
+
+    const linkRenderer = canvas.linkRenderer
+    if (!linkRenderer) return
+
+    const context = buildContext(canvas)
+
+    const from: ReadOnlyPoint = [start.x, start.y]
+    const to: ReadOnlyPoint = [pointer.canvas.x, pointer.canvas.y]
```
DrJKL, line 54: Instead of `ReadOnlyPoint`, can we use the `readonly` type modifier on the `Point` interface?

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -2,8 +2,10 @@
   <div v-if="renderError" class="node-error p-1 text-red-500 text-xs">⚠️</div>
   <div
     v-else
-    class="lg-slot lg-slot--input flex items-center cursor-crosshair group rounded-r-lg h-6"
+    class="lg-slot lg-slot--input flex items-center group rounded-r-lg h-6"
     :class="{
+      'cursor-crosshair': !readonly,
+      'cursor-default': readonly,
       'opacity-70': readonly,
```
DrJKL, line 9: You can unify these into a single string.

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -2,8 +2,10 @@
   <div v-if="renderError" class="node-error p-1 text-red-500 text-xs">⚠️</div>
   <div
     v-else
-    class="lg-slot lg-slot--input flex items-center cursor-crosshair group rounded-r-lg h-6"
+    class="lg-slot lg-slot--input flex items-center group rounded-r-lg h-6"
     :class="{
```
DrJKL, line 6: Could you wrap all of the class stuff in `cn()` to get the tailwind-merge resolution?

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -16,6 +18,7 @@
       ref="connectionDotRef"
       :color="slotColor"
       class="-translate-x-1/2"
+      v-on="readonly ? {} : { pointerdown: onPointerDown }"
```
DrJKL, line 21: Could the composable take in readonly as a prop to affect the returned handler?

src/renderer/extensions/vueNodes/components/OutputSlot.vue
DrJKL, line 1: There's a lot of duplication across these two components. Not necessarily for this PR, but it could be worth finding a reasonable abstraction here.

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -203,12 +204,15 @@ export function useSlotElementTracking(options: {
 
     // Remove this slot from registry and layout
     const slotKey = getSlotKey(nodeId, index, type === 'input')
-    node.slots.delete(slotKey)
+    const entry = node.slots.get(slotKey)
+    if (entry) {
+      delete entry.el.dataset.slotKey
+      node.slots.delete(slotKey)
+    }
```
DrJKL, line 211: Should this be in the yjs store? This feels like a CRDT operation kind of thing.

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -0,0 +1,206 @@
...
+        candidate
+      ).allowable
+    }
+
+    return candidate
+  }
+
+  const conversion = useSharedCanvasPositionConversion()
+
+  let activePointerId: number | null = null
```
DrJKL, line 57: Just checking, for any mutable state it might be better to mark it as a ref, unless you know for a fact that changes to it shouldn't ever trigger things.

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -0,0 +1,206 @@
...
+    if (event.pointerId !== activePointerId) return
+    updatePointerState(event)
+    app.canvas?.setDirty(true)
+  }
+
+  const connectSlots = (slotLayout: SlotLayout) => {
+    const canvas = app.canvas
+    const graph = canvas?.graph
+    const source = state.source
+    if (!canvas || !graph || !source) return
```
DrJKL, line 90: Love all the guard clauses, very flat. <3

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -0,0 +1,206 @@
...
+
+    const sourceNode = graph.getNodeById(Number(source.nodeId))
+    const targetNode = graph.getNodeById(Number(slotLayout.nodeId))
+    if (!sourceNode || !targetNode) return
+
+    const sourceSlot =
```
DrJKL, line 96: This smells a little like asking the duck if it needs batteries.

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -0,0 +1,206 @@
...
+    const targetSlot =
+      slotLayout.type === 'input'
+        ? targetNode.inputs?.[slotLayout.index]
+        : targetNode.outputs?.[slotLayout.index]
+
+    if (!sourceSlot || !targetSlot) return
+
+    if (source.type === 'output' && slotLayout.type === 'input') {
+      const outputSlot = sourceSlot as INodeOutputSlot | undefined
+      const inputSlot = targetSlot as INodeInputSlot | undefined
```
DrJKL, line 109: Same question about the type assertions. Is there a place we can add a type so that inference would do the rest?

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -0,0 +1,206 @@
...
+      event.pointerId
+    )
+
+    activePointerId = event.pointerId
+
+    updatePointerState(event)
+
+    window.addEventListener('pointermove', handlePointerMove, true)
+    window.addEventListener('pointerup', handlePointerUp, true)
+    window.addEventListener('pointercancel', handlePointerCancel, true)
```
DrJKL, line 191: Like @christian-byrne said, I think either useEventListener or an abort signal would be good here.

DrJKL (changes requested):

benceruleanlu:
@DrJKL 
>No custom nodes use it?

I wrote it not long ago, and it was non functional as well, there's not much you can do with it, and it was only a simplification for renderLinkDirect anyway

I've been meaning to remove it

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
benceruleanlu: done

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
benceruleanlu: sure

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
benceruleanlu: it will be changed in another PR later, where we allow for it to use the defaults automatically

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
benceruleanlu: onDrawForeground is the official hook, I switched to that

although it really does look like monkey patching, it does preserve previous + patch, which seems to be the intended way to use it

src/renderer/extensions/vueNodes/components/InputOutputSlot.test.ts
benceruleanlu: on its way

src/renderer/core/linkInteractions/slotLinkDragState.ts
benceruleanlu: it does seem suspicious, the idea was to defend from consumers mutating x and y, so this creation + replacement was to trigger reactivity

I changed it to be fully mutable

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: I switched to useEventListener, but I think we still need to manually cleanup when the drag ends

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: github forces extra lines so I'm guessing here you mean sourceNode.disconnectInput(source.slotIndex, true)

if that's the case, I don't think we need to? it actually calls disconnect itself during the connection code, so this is just to be safe

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: validation happened before

    if (state.source) {
      const candidate = candidateFromTarget(event.target)
      if (candidate?.compatible) {
        connectSlots(candidate.layout)
      }
    }

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
benceruleanlu: sure

benceruleanlu: actually, is there a deeper reason for it? I think the isNaN was to prevent throwing, wouldn't nothing change if it just couldn't find the node?

I'll keep it as is for now to prevent defensive coding potentially misleading devs

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
benceruleanlu: done

src/renderer/core/linkInteractions/slotLinkDragState.ts
benceruleanlu: sounds good

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
benceruleanlu: it's typescript being dumb and not knowing this is what sourceSlot/targetSlot actually is from our previous if, so this isn't actually forcing a type

I made it look prettier and now TS infers the right type without needing as

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
benceruleanlu: OK

benceruleanlu: actually, is there a specific reason for this?

isn't ReadOnlyPoint the standard type for this use case?

src/renderer/extensions/vueNodes/components/InputSlot.vue
benceruleanlu: sure

src/renderer/extensions/vueNodes/components/OutputSlot.vue
benceruleanlu: sounds good, can be a followup

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
benceruleanlu: no? this specifically is the cleanup for the drop slot detection

src/renderer/extensions/vueNodes/components/InputSlot.vue
benceruleanlu: OK

src/renderer/extensions/vueNodes/components/InputSlot.vue
benceruleanlu: OK

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
christian-byrne: I'm not 100% positive on the meaning of `-1` id but I thought it was being used in some place - was actually mostly asking you, should have prefixed with "question:"

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: it's not really a type thing, moreso that the code was meant to provide clarity by defining sourceSlot and targetSlot, but typescript isn't able to infer the types that well despite our if guard, so I just changed it to fetch as is, TS can infer that way without needing type assertations

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: can you clarify?

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: changing to ref is not needed right now, I think it would add .value noise in this PR

src/renderer/core/linkInteractions/slotLinkCompatibility.ts
benceruleanlu: -10 and -20 are used by subgraph input and output nodes respectively, but I think it should be fine to allow this for resolveNode since if we guard for it it'd just break on subgraphs

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: useEventListener was used

src/renderer/core/linkInteractions/slotLinkDragState.ts
DrJKL: Could you use [useMouse](https://vueuse.org/core/useMouse/)?

src/renderer/core/linkInteractions/slotLinkPreviewRenderer.ts
DrJKL: Simplicity, mostly.
But it can probably wait until https://github.com/Comfy-Org/ComfyUI_frontend/pull/5567 goes in.

christian-byrne (dismissed):
LGTM!

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
DrJKL: https://codesoapbox.dev/solid-principles-3-the-liskov-substitution-principle/

src/renderer/core/canvas/links/slotLinkCompatibility.ts
```diff
@@ -15,9 +18,10 @@ interface CompatibilityResult {
   targetSlot?: INodeInputSlot | INodeOutputSlot
 }
 
-function resolveNode(nodeId: string | number) {
-  const canvas = app.canvas
-  const graph = canvas?.graph
+function resolveNode(nodeId: NodeId) {
+  const pinia = getActivePinia()
+  const canvasStore = pinia ? useCanvasStore() : null
```
DrJKL, line 23: When would we not have pinia?

src/renderer/core/canvas/links/slotLinkDragState.ts
```diff
@@ -20,9 +20,14 @@ export interface SlotDropCandidate {
   compatible: boolean
 }
 
+interface MutablePoint {
+  x: number
+  y: number
+}
```
DrJKL, line 26: I don't think you need to specify mutable, that's the default in javascript.

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -93,9 +96,10 @@ useSlotElementTracking({
...
 
-const { onPointerDown } = useSlotLinkInteraction({
+const onPointerDown = useSlotLinkInteraction({
   nodeId: props.nodeId ?? '',
   index: props.index,
-  type: 'input'
-})
+  type: 'input',
+  readonly: props.readonly
+})?.onPointerDown
```
DrJKL, line 104:
```suggestion
const { onPointerDown } = useSlotLinkInteraction({
 nodeId: props.nodeId ?? '',
 index: props.index,
  type: 'input',
  readonly: props.readonly
})
```

src/renderer/extensions/vueNodes/components/OutputSlot.vue
```diff
@@ -95,9 +97,10 @@ useSlotElementTracking({
...
 
-const { onPointerDown } = useSlotLinkInteraction({
+const onPointerDown = useSlotLinkInteraction({
   nodeId: props.nodeId ?? '',
   index: props.index,
-  type: 'output'
-})
+  type: 'output',
+  readonly: props.readonly
+})?.onPointerDown
```
DrJKL, line 105:
```suggestion
const { onPointerDown } = useSlotLinkInteraction({
 nodeId: props.nodeId ?? '',
 index: props.index,
  type: 'output',
  readonly: props.readonly
})
```

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -1,32 +1,39 @@
...
   nodeId,
   index,
-  type
-}: SlotInteractionOptions) {
+  type,
+  readonly
+}: SlotInteractionOptions): SlotInteractionHandlers | null {
+  if (readonly) {
+    return null
+  }
```
DrJKL, line 35: I meant gate the behavior within the callback. Composable returns should usually be of a consistent shape.

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -55,26 +62,30 @@ export function useSlotLinkInteraction({
   const conversion = useSharedCanvasPositionConversion()
 
   let activePointerId: number | null = null
+  let stopPointerMove: Fn | null = null
+  let stopPointerUp: Fn | null = null
+  let stopPointerCancel: Fn | null = null
```
DrJKL, line 67: That's a lot of coordinated mutable state :-/

DrJKL (changes requested):

src/renderer/core/canvas/links/slotLinkCompatibility.ts
benceruleanlu: in any call stack that vue/pinia doesn't own, such as when the DOM window event listeners trigger, we fallback to app.canvas?.graph

src/renderer/core/canvas/links/slotLinkDragState.ts
benceruleanlu: OK

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: OK

src/renderer/extensions/vueNodes/components/OutputSlot.vue
benceruleanlu: thanks!

src/renderer/extensions/vueNodes/components/InputSlot.vue
benceruleanlu: thanks!

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: fixed with a pointer session helper

christian-byrne (approved):

DrJKL (approved):

PR #5631, opened 2025-09-18
feat: vue nodes LOD system

simula-r:
## Summary

Replaced reactive (Vue-based) widget LOD with CSS visibility control. Huge JS performance gain off the bat, but doesn't dramatically improve rendering because we are memory bound, but we avoid the mount/unmount overhead during zoom/pan operations. This PR implements the visual component of LOD—complex widgets that need lifecycle management will be addressed separately.

### Problem & Solution
Problem: we want LOD to improve rendering performance and visual feedback but discovered using reactivity in the current setup for it meant mounting/unmounting caused worse perf than it aimed to fix. Switching to render all the nodes all the time but using css visibility proved to be the best solution. However, it doesn't improve rendering performance by much because the GPU texture size is the bottleneck (from TransformPane.vue CSS transforms) and not rasterization.

Solution: Keep all nodes/widgets mounted, use CSS visibility: hidden for LOD. Trade memory for performance stability during zoom/pan/drag operations.

### Technical Decision
We chose Performance > Memory:

- CSS transforms create a single GPU texture whose size depends on node count, not widget complexity
- Mounting/unmounting hundreds of widgets during zoom = noticeable lag from Vue VDOM diffing (since all components are mounted all the time because of viewport culling challenge/trade off see https://github.com/Comfy-Org/ComfyUI_frontend/pull/5510.)
- CSS visibility changes = no reactivity overhead, smooth interactions
- Result: Similar performance, but without interaction stutters

This is the visual layer only. If we want a hook into the LOD state per node / widget that would be the next follow up system to implement.

### Next Steps (maybe)
- Chunked (split up single Transform Pane transform layer) when rendering 1000+ nodes (maybe)
- ~~Selective unmounting API for widgets that register as "expensive"~~
- ~~Client bound hydration system~~

image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5631-feat-vue-nodes-LOD-system-2726d73d365081c6a6c4e14aa634f19c) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/lod/useLOD.ts
```diff
@@ -4,183 +4,37 @@
...
+
+  // Set a default min font size & threshold in case store unavaliable
+  let minFontSize = 12
+  let threshold = 0.4
+
+  watchEffect(() => {
+    const minFontSizeFromStore = settingStore.get(
+      'LiteGraph.Canvas.MinFontSizeForLOD'
+    )
+    if (minFontSizeFromStore) {
```
christian-byrne: Do we allow setting to `0`?

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -94,20 +92,15 @@ const processedWidgets = computed((): ProcessedWidget[] => {
...
-  }
-
   for (const widget of widgets) {
     if (widget.options?.hidden) continue
     if (widget.options?.canvasOnly) continue
     if (!widget.type) continue
     if (!shouldRenderAsVue(widget)) continue
 
-    if (lodLevel === LODLevel.REDUCED && !isEssential(widget.type)) continue
+    if (!isEssential(widget.type)) continue
```
christian-byrne: In this case, we should be removing this line entirely right?

src/renderer/extensions/vueNodes/lod/useLOD.ts
simula-r: Yeah, 0 turns it off. Its not the best discoverability honestly. But in the current settings system we don't have other controls that depend on one another. Ideally, we could have a 2 input setting. one input would be a slider, one input would be a toggle.

tests-ui/tests/renderer/extensions/vueNodes/lod/useLOD.test.ts
```diff
@@ -1,270 +1,132 @@
...
-    })
+describe('useLOD', () => {
+  beforeEach(() => {
+    // Set up default mock behavior
+    vi.mocked(useSettingStore).mockReturnValue({
+      get: (key: string) => {
+        if (key === 'LiteGraph.Canvas.MinFontSizeForLOD') return 12
+        return undefined
+      }
+    } as any)
```
DrJKL: Can you find a way to not use `as any`?

tests-ui/tests/renderer/extensions/vueNodes/lod/useLOD.test.ts
```diff
@@ -1,270 +1,132 @@
...
-      expect(shouldRenderSlotLabels.value).toBe(true)
-      expect(shouldRenderWidgetLabels.value).toBe(true)
-    })
+describe('useLOD', () => {
+  beforeEach(() => {
+    // Set up default mock behavior
+    vi.mocked(useSettingStore).mockReturnValue({
+      get: (key: string) => {
+        if (key === 'LiteGraph.Canvas.MinFontSizeForLOD') return 12
+        return undefined
```
DrJKL: You could probably do this within the mock instantiation, then just `resetAllMocks` here.

tests-ui/tests/renderer/extensions/vueNodes/lod/useLOD.test.ts
```diff
@@ -1,270 +1,132 @@
...
+    const isLOD = ref(false)
 
-      const widgets = [{ type: 'combo' }, { type: 'text' }, { type: 'button' }]
+    useLOD(camera, isLOD)
+    await nextTick()
 
-      expect(getEssentialWidgets(widgets)).toEqual([])
-    })
+    // threshold = 12 / (14 * 1) = 0.857... rounded to 0.86
+    expect(isLOD.value).toBe(false) // z=1 > 0.86, so no LOD
```
DrJKL: That's... a lot of numbers...

tests-ui/tests/renderer/extensions/vueNodes/lod/useLOD.test.ts
```diff
@@ -1,270 +1,132 @@
...
-      expect(lodCssClass.value).toBe('lg-node--lod-minimal')
-
-      zoomRef.value = 0.6
-      expect(lodCssClass.value).toBe('lg-node--lod-reduced')
-
-      zoomRef.value = 1.0
-      expect(lodCssClass.value).toBe('lg-node--lod-full')
-    })
+  afterEach(() => {
+    vi.clearAllMocks()
```
DrJKL: I usually prefer `resetAllMocks` or `restoreAllMocks` in a `beforeEach`

tests-ui/tests/renderer/extensions/vueNodes/lod/useLOD.test.ts
```diff
@@ -1,270 +1,132 @@
...
-      ]
-
-      expect(getEssentialWidgets(widgets)).toEqual(widgets)
+  it('should calculate threshold correctly with minFontSize = 12', async () => {
+    // Mock devicePixelRatio for predictable calculation
+    Object.defineProperty(window, 'devicePixelRatio', {
+      writable: true,
+      configurable: true,
+      value: 1
     })
```
DrJKL: https://vitest.dev/api/vi.html#vi-stubglobal

tests-ui/tests/renderer/extensions/vueNodes/lod/useLOD.test.ts
```diff
@@ -1,270 +1,132 @@
...
-  it('should handle threshold boundary values correctly', () => {
-    expect(supportsFeatureAtZoom(0.8, 'renderWidgets')).toBe(true)
-    expect(supportsFeatureAtZoom(0.8, 'renderContent')).toBe(false)
+  it('should handle minFontSize=0 (keeps default due to falsy check)', async () => {
+    vi.mocked(useSettingStore).mockReturnValueOnce({
+      get: (key: string) => {
+        if (key === 'LiteGraph.Canvas.MinFontSizeForLOD') return 0
+        return undefined
+      }
+    } as any)
```
DrJKL: Same elsewhere, this kind of mock implementation should probably be extracted for clarity and moved outside of the `describe` scope.

src/renderer/extensions/vueNodes/lod/useLOD.ts
```diff
@@ -4,183 +4,38 @@
...
-    return Math.max(0, Math.min(1, zoom))
+export function useLOD(camera: Camera, isLOD: Ref<boolean>) {
+  const settingStore = useSettingStore()
+
+  const baseFontSize = 14
+  const dprAdjustment = Math.sqrt(window.devicePixelRatio || 1)
+
+  // Set a default min font size & threshold in case store unavaliable
+  let minFontSize = 12
+  let threshold = 0.4
```
DrJKL: Why not make these `computed` based on the store value?

src/renderer/extensions/vueNodes/components/SlotConnectionDot.vue
```diff
@@ -24,6 +24,7 @@ defineExpose({
   >
     <div
       ref="slot-el"
+      class="slot-dot"
```
DrJKL, line 27: What's the class for?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -312,25 +288,20 @@ const hasCustomContent = computed(() => {
...
 // Common condition computations to avoid repetition
-const shouldShowWidgets = computed(
-  () => shouldRenderWidgets.value && nodeData.widgets?.length
-)
+const shouldShowWidgets = computed(() => nodeData.widgets?.length)
 
-const shouldShowContent = computed(
-  () => shouldRenderContent.value && hasCustomContent.value
-)
+const shouldShowContent = computed(() => hasCustomContent.value)
```
DrJKL: Is this just an alias?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -312,25 +288,20 @@ const hasCustomContent = computed(() => {
 
 // Computed classes and conditions for better reusability
 const separatorClasses =
-  'bg-sand-100 dark-theme:bg-charcoal-600 h-px mx-0 w-full'
+  'bg-sand-100 dark-theme:bg-charcoal-600 h-px mx-0 w-full lod-toggle'
```
DrJKL, line 254: What's the class for?

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -19,12 +19,15 @@
     />
 
     <!-- Slot Name -->
-    <span
-      v-if="!dotOnly"
-      class="whitespace-nowrap text-sm font-normal dark-theme:text-slate-200 text-stone-200"
-    >
-      {{ slotData.localized_name || slotData.name || `Input ${index}` }}
-    </span>
+    <div class="relative">
```
DrJKL, line 13: Deja vu...

src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -94,17 +94,20 @@
...
+          {{ $t('g.errorLoadingImage') }}
+        </span>
+        <span v-else-if="isLoading" class="text-gray-400">
+          {{ $t('g.loading') }}...
+        </span>
+        <span v-else>
+          {{ actualDimensions || $t('g.calculatingDimensions') }}
+        </span>
+      </div>
+      <LODFallback />
```
DrJKL, line 110: Given how often this pattern occurs, should LODFallback just have a slot and wrap the content?

src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -1,7 +1,10 @@
 <template>
   <div
-    class="transform-pane"
-    :class="{ 'transform-pane--interacting': isInteracting }"
+    class="absolute inset-0 w-full h-full pointer-events-none"
+    :class="[
+      isInteracting ? 'transform-pane--interacting' : 'will-change-auto',
+      isLOD ? 'isLOD' : ''
+    ]"
```
DrJKL: You could use `cn()` here for all the classes.
But also, what's `'isLOD'` used for?

src/assets/css/style.css
```diff
@@ -1007,12 +966,42 @@ audio.comfy-audio.empty-audio-widget {
...
+.isLOD .lod-fallback {
+  display: block;
+}
+
+.isLOD .image-preview img {
+  image-rendering: pixelated;
+}
+
+
+.isLOD .slot-dot {
```
DrJKL, line 1002: Oooooh, I was reading from the bottom up.

Please don't add more CSS here if you can help it. I'd rather have the tailwind utilities directly applied in situ.

src/assets/css/style.css
```diff
@@ -1007,12 +966,42 @@ audio.comfy-audio.empty-audio-widget {
   -webkit-mask-image: none !important;
   mask-image: none !important;
   clip-path: none !important;
+  /* border-radius: 0 !important; */
+  background-image: none !important;
+  text-rendering: optimizeSpeed !important;
+  border-radius: 0 !important;
+  contain: layout content paint;
+   transition: none !important;
```
DrJKL: Can you remove all of the `!important`s?

browser_tests/tests/vueNodes/lod.spec.ts
DrJKL, line 1: Nice and succinct test 😸

browser_tests/tests/vueNodes/lod.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  test.beforeEach(async ({ comfyPage }) => {
+    // Enable Vue nodes rendering
+    await comfyPage.setSetting('Comfy.VueNodes.Enabled', true)
+    await comfyPage.setup()
+    await comfyPage.loadWorkflow('default')
+  })
+
+  test('should toggle LOD based on zoom threshold', async ({ comfyPage }) => {
+    // Wait for Vue nodes to be rendered
+    await comfyPage.vueNodes.waitForNodes()
```
DrJKL, line 13: This feels like it should be baked into the fixture, not for this PR, but in the future.

browser_tests/tests/vueNodes/lod.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    // Verify initial state (not in LOD mode at normal zoom)
+    const initialNodeCount = await comfyPage.vueNodes.getNodeCount()
+    expect(initialNodeCount).toBeGreaterThan(0)
```
DrJKL, line 16: Having nodes indicates not being in LOD mode?

browser_tests/tests/vueNodes/lod.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    await comfyPage.zoom(120, 10) // Zoom out significantly
+    await comfyPage.nextFrame()
+
+    // Take screenshot with LOD active (low detail)
+    await expect(comfyPage.canvas).toHaveScreenshot('vue-nodes-lod-active.png')
+
+    // Verify LOD is active by checking for the isLOD class
+    const lodActiveState = await comfyPage.page.evaluate(() => {
+      // Check if any element has the isLOD class
+      return document.querySelector('.isLOD') !== null
```
DrJKL: Instead of checking for a class, is there a user-facing assertion you could make?

DrJKL (changes requested):

src/renderer/extensions/vueNodes/components/ImagePreview.vue
simula-r: In this case I think because the markup is very specific (each part of the node has unique DOM hierarchy) that would take some work to extract and may change as the nodes design change its best to keep it how it is now.

src/assets/css/style.css
simula-r: See above:
/* START LOD specific styles */
/* LOD styles - Custom CSS avoids 100+ Tailwind selectors that would slow style recalculation when .isLOD toggles */

My understanding is if we moved these selectors to tailwind we would generate a bunch of individual selectors. This is more efficient. I could move this to the TransformPane.vue in the <style> tag and remove the scoped tag. WDYT?

browser_tests/tests/vueNodes/lod.spec.ts
simula-r: Good question. Other than checking zoom, and computing the threshold in the test or screenshotting and checking for simplified pixels on the nodes. Its either check the downstream effect of the previous assertions which is .isLOD in the DOM. And those previous assertions are being tested via the component test. 

Given we have the component test asserting how .isLOD class gets created and assigned, and thats the source of truth for the css being applied on the component do we need to essentially do the same component test in the browser_test as well?

DrJKL: Can you check the visibility or invisibility of elements that would be hidden under different conditions?

DrJKL: https://playwright.dev/docs/best-practices#test-user-visible-behavior

christian-byrne:
Looking good, just some merge conflicts it seems

src/renderer/extensions/vueNodes/lod/useLOD.ts
```diff
@@ -4,183 +4,38 @@
...
-export function useLOD(zoomRef: Ref<number>) {
-  // Continuous LOD score (0-1) for smooth transitions
-  const lodScore = computed(() => {
-    const zoom = zoomRef.value
-    return Math.max(0, Math.min(1, zoom))
+export function useLOD(camera: Camera, isLOD: Ref<boolean>) {
+  const settingStore = useSettingStore()
+
+  const baseFontSize = 14
+  const dprAdjustment = Math.sqrt(window.devicePixelRatio || 1)
```
DrJKL: https://vueuse.org/core/useDevicePixelRatio/#usedevicepixelratio

simula-r:
> Looking good, just some merge conflicts it seems

Yep. Will fix that along with more feedback implementations. 
Lets wait to merge until there in.

christian-byrne (approved):

src/assets/css/style.css
DrJKL: That's not [how tailwind works](https://dev.to/m4xshen/tailwind-css-under-the-hood-m1o).

Tailwind classes are shared by any instances that use them, so it's almost always less code in the bundle.

src/renderer/extensions/vueNodes/components/ImagePreview.vue
DrJKL: https://vuejs.org/guide/components/slots.html

PR #5633, opened 2025-09-18
lint: add tsconfig for browser_tests, fix existing violations

DrJKL:
## Summary

See https://typescript-eslint.io/blog/project-service/ for context.
Creates a browser_tests specific tsconfig so that they can be linted.

Does not add a package.json script to do the linting yet, but `pnpm exec eslint browser_tests` should work for now.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5633-lint-add-tsconfig-for-browser_tests-fix-existing-violations-2726d73d3650819d8ef2c4b0abc31e14) by [Unito](https://www.unito.io)

### DIFF
--- a/tests-ui/tsconfig.json
+++ b/tests-ui/tsconfig.json
@@ -1,0 +1,14 @@
+{
+  // "extends": "../tsconfig.json",
+  "compilerOptions": {
+    /* Test files should not be compiled */
+    "noEmit": true,
+    // "strict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "resolveJsonModule": true
+  },
+  "include": [
+    "**/*.ts",
+  ]
+}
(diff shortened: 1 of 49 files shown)


webfiltered (dismissed):
Should config also be copied into `tests/` and `tests-ui/`?

DrJKL:
> Should config also be copied into `tests/` and `tests-ui/`?

`tests-ui`, yes!

(Where is `tests/`?)

webfiltered:
> (Where is `tests/`?)

In another repo.  \*nods\*

tests-ui/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+  // "extends": "../tsconfig.json",
```
webfiltered: Shouldn't need, or doesn't work?  Curiosity only.

webfiltered (approved):

tests-ui/tsconfig.json
DrJKL: Need for the editor/LSP, but funnily enough works otherwise.

arjansingh (approved):

PR #5637, opened 2025-09-18
Fix cyclic prototype errors with subgraphNodes

AustinMroz:
#5024 added support for connecting primitive nodes to subgraph inputs. To accomplish this, it pulls WidgetLocator information from the node owning the widget.

This `node` property does not exist on all IBaseWidget. `toConcrete` was used to instead have a BaseWidget which is guaranteed to have a node property. The issue that was missed, is that a widget which lacks this information (such as most implemented by custom nodes) sets the node value to the argument which was passed. Here that is the reference to the subgraph node. Sometimes, this `#setWidget` call is made multiple times, and when this occurs, the `input.widget` has itself set as the protoyep, throwing an error.

This is resolved by instead taking an additional input which is unambiguous.

For reference, this is a near minimal workflow using comfy_mtb that replicates the issue
[cyclic.json](https://github.com/user-attachments/files/22412187/cyclic.json)

Special thanks to @melMass for assistance discovering this issue.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5637-Fix-cyclic-prototype-errors-with-subgraphNodes-2726d73d365081fea356f5197e4c2b42) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/subgraph/SubgraphNode.ts
+++ b/src/lib/litegraph/src/subgraph/SubgraphNode.ts
@@ -189,17 +189,14 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
 
     subgraphInput.events.addEventListener(
       'input-connected',
-      () => {
+      (e) => {
         if (input._widget) return
 
-        const linkId = subgraphInput.linkIds[0]
-        const resolvedInput = this.subgraph.links[linkId].resolve(
-          this.subgraph
-        ).input
         const widget = subgraphInput._widget
         if (!widget) return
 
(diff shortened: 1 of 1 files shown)


DrJKL (approved):
🤞🏻

DrJKL:
(I still don't think prototype setting is ever the right move)

christian-byrne (approved):
LGTM!

christian-byrne:
Wait to merge this until 1.28 is released, so we can backport onto 1.27

christian-byrne:
@AustinMroz I can do the cherry pick

PR #5638, opened 2025-09-18
Explicitly add email scope for social auth login.

robinjhuang:
## Summary

Some users were authenticating successfully but their email addresses weren't being extracted from the Firebase token. This happened because we weren't explicitly requesting the email scope during OAuth authentication.
 
While Firebase's default configuration includes basic profile info, it doesn't guarantee email access for all account types - particularly Google Workspace accounts with restrictive policies or users with privacy-conscious settings.

[Github Scopes](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps)

## Changes

Adding email scope for Google + Github social OAuth.

## Review Focus
N/A

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5638-Explicitly-add-email-scope-for-social-auth-login-2726d73d3650817ab356fc9c04f8641b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/firebaseAuthStore.ts
+++ b/src/stores/firebaseAuthStore.ts
@@ -62,10 +62,12 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
 
   // Providers
   const googleProvider = new GoogleAuthProvider()
+  googleProvider.addScope('email')
   googleProvider.setCustomParameters({
     prompt: 'select_account'
   })
   const githubProvider = new GithubAuthProvider()
+  githubProvider.addScope('user:email')
   githubProvider.setCustomParameters({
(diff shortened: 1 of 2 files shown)


DrJKL (dismissed):
I think the mocks in the test need to be updated, but 👍🏻

AustinMroz (commented):
LGTM

DrJKL (approved):

webfiltered (approved):

AustinMroz (approved):

christian-byrne:
I feel like we should just use `uid` for the whitelist, but adding extra consent box on sign up isn't too bad for now.

PR #5643, opened 2025-09-18
Fix SaveAs

### LINKED ISSUE #5592: Bug: File -> Save As doesn’t save current changes
### Prerequisites

### What happened?

You make some changes and want to save the workflow with a new name. File -> Save As and then enter the new workflow name. New tab opens with the new name but none of the changes. The changes remain on the original workflow unsaved. It’s incredibly frustrating because it basically forces you to either save as before you make any changes or to paradoxically save as the old workflow logic rather than the new. Save as should be equivalent to normal save but with a new name, like it used to work

### Steps to Reproduce

See above

### ComfyUI Frontend Version

1.27.3

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-5592-Bug-File-Save-As-doesn-t-save-current-changes-26f6d73d3650817481f8da347f9eb942) by [Unito](https://www.unito.io)

AustinMroz:
Implementing subgraph blueprints (#5139) included changes to saving to ensure that SaveAs generates a new workflow of the correct type. However this code failed to utilize the pre-prepared state when performing the actual save. This produced a couple of problems with both failing to detach the workflow and failing to apply the correct state 

This error is only encountered when using Save As from a non temporary workflow (one loaded from the workflows sidebar tab).

As this state calculation code is only used in this code path, it has been moved into the saveAs function of the workflowStore.

Resolves #5592

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5643-Fix-SaveAs-2726d73d3650818faa7af449d1f13c26) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/workflow/core/services/workflowService.ts
+++ b/src/platform/workflow/core/services/workflowService.ts
@@ -17,7 +17,7 @@ import { downloadBlob } from '@/scripts/utils'
 import { useDialogService } from '@/services/dialogService'
 import { useDomWidgetStore } from '@/stores/domWidgetStore'
 import { useWorkspaceStore } from '@/stores/workspaceStore'
-import { appendJsonExt, generateUUID } from '@/utils/formatUtil'
+import { appendJsonExt } from '@/utils/formatUtil'
 
 export const useWorkflowService = () => {
   const settingStore = useSettingStore()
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):
LGTM!

PR #5662, opened 2025-09-19
[refactor] Migrate manager code to DDD structure

christian-byrne:
## Summary

Reorganized custom nodes manager functionality from scattered technical layers into a cohesive domain-focused module following [domain-driven design](https://en.wikipedia.org/wiki/Domain-driven_design) principles.

## Changes

- **What**: Migrated all manager code from technical layers (`src/components/`, `src/stores/`, etc.) to unified domain structure at `src/workbench/extensions/manager/`
- **Breaking**: Import paths changed for all manager-related modules (40+ files updated)

## Review Focus

Verify all import path updates are correct and no circular dependencies introduced. Check that [Vue 3 composition API](https://vuejs.org/guide/reusability/composables.html) patterns remain consistent across relocated composables.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5662-refactor-Migrate-manager-code-to-DDD-structure-2736d73d3650812c87faf6ed0fffb196) by [Unito](https://www.unito.io)


src/composables/useConflictDetection.ts
```diff
@@ -641,7 +641,9 @@ export function useConflictDetection() {
   async function initializeConflictDetection(): Promise<void> {
     try {
       // Check if manager is new Manager before proceeding
-      const { useManagerState } = await import('@/composables/useManagerState')
+      const { useManagerState } = await import(
+        '@/workbench/extensions/manager/composables/useManagerState'
+      )
```
DrJKL, line 646: I know it's not a part of this PR, but why a dynamic import here?

DrJKL (approved):

src/composables/useConflictDetection.ts
christian-byrne: I don't remember but likely to prevent import side effects in tests or when wanting to disable conflict detection, or to prevent some circular import issue.

viva-jinyi (approved):

PR #5664, opened 2025-09-19
fix: add Safari requestIdleCallback polyfill

christian-byrne:
## Summary

Implemented cross-browser requestIdleCallback polyfill to fix Safari crashes during graph initialization.

## Changes

- **What**: Added [requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) polyfill following [VS Code's pattern](https://github.com/microsoft/vscode/blob/main/src/vs/base/common/async.ts) with setTimeout fallback for Safari
- **Breaking**: None - maintains existing GraphView behavior

## Review Focus

Safari compatibility testing and timeout handling in the 15ms fallback window. Verify that initialization tasks (keybindings, server config, model loading) still execute properly on Safari iOS.

## References

- [VS Code async.ts implementation](https://github.com/microsoft/vscode/blob/main/src/vs/base/common/async.ts) - Source pattern for our polyfill
- [MDN requestIdleCallback](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback) - Browser API documentation  
- [Safari requestIdleCallback support](https://caniuse.com/requestidlecallback) - Browser compatibility table

Fixes CLOUD-FRONTEND-STAGING-N

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5664-fix-add-Safari-requestIdleCallback-polyfill-2736d73d365081cdbcf1fb816fe098d6) by [Unito](https://www.unito.io)

### DIFF
--- a/src/base/common/async.ts
+++ b/src/base/common/async.ts
@@ -1,0 +1,98 @@
+/**
+ * Cross-browser async utilities for scheduling tasks during browser idle time
+ * with proper fallbacks for browsers that don't support requestIdleCallback.
+ *
+ * Implementation based on:
+ * https://github.com/microsoft/vscode/blob/main/src/vs/base/common/async.ts
+ */
+
+interface IdleDeadline {
+  didTimeout: boolean
+  timeRemaining(): number
+}
+
+interface IDisposable {
+  dispose(): void
+}
+
+/**
+ * Internal implementation function that handles the actual scheduling logic.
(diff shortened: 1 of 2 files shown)


christian-byrne:
Need to merge this [because the issue keeps happening](https://comfy-org.sentry.io/issues/6757450175/events/fc0f6a49cd364b69a3af843e7f3791f2/?project=4509681221369857) in prod.

PR #5668, opened 2025-09-19
Add playwright tests for links and slots in vue nodes mode

benceruleanlu:
Tests added
- Should show a link dragging out from a slot when dragging on a slot
- Should create a link when dropping on a compatible slot
- Should not create a link when dropping on an incompatible slot(s)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5668-Add-playwright-tests-for-links-and-slots-in-vue-nodes-mode-2736d73d36508188a47dceee5d1a11e5) by [Unito](https://www.unito.io)


benceruleanlu:
> SyntaxError: /home/runner/work/ComfyUI_frontend/ComfyUI_frontend/ComfyUI_frontend/src/lib/litegraph/src/LGraphNode.ts: TypeScript 'declare' fields must first be transformed by @babel/plugin-transform-typescript.
> If you have already enabled that plugin (or '@babel/preset-typescript'), make sure that it runs before any plugin related to additional class features:
>  - @babel/plugin-transform-class-properties
>  - @babel/plugin-transform-private-methods
>  - @babel/plugin-proposal-decorators
>   405 |   showAdvanced?: boolean
>   406 |
> > 407 |   declare comfyClass?: string
>       |   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   408 |   declare isVirtualNode?: boolean
>   409 |   applyToGraph?(extraLinks?: LLink[]): void
>   410 |

speed-ishowspeed

let me try fixing this

benceruleanlu:
the failing test is the new screenshot needed, feel free to check it out

benceruleanlu:
by the way, I asked design if they had something for links, we can change it at a later time, they haven't made anything just yet, so that's why the links look odd ending abruptly

browser_tests/fixtures/ComfyPage.ts
```diff
@@ -29,6 +29,8 @@ dotenv.config()
 
 type WorkspaceStore = ReturnType<typeof useWorkspaceStore>
 
+type Bounds = readonly [number, number, number, number]
```
DrJKL, line 32: Can we import the type from src?

browser_tests/fixtures/ComfyPage.ts
```diff
@@ -338,6 +340,99 @@ export class ComfyPage {
     return `./browser_tests/assets/${fileName}`
   }
 
+  async fitToView(
+    options: {
+      selectionOnly?: boolean
+      zoom?: number
+      padding?: number
+    } = {}
+  ) {
```
DrJKL, line 349: This method is a lot of logic to put in a fixture.
Normally I like fixtures to be as close to describing what a user would do as possible.

The general rule of browser tests is: unless you expect a user to open the console and type in a series of commands, the test shouldn't do that either.

browser_tests/tests/vueNodes/linkInteraction.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+} from '../../fixtures/ComfyPage'
+
+async function getCenter(locator: Locator): Promise<{ x: number; y: number }> {
+  const box = await locator.boundingBox()
+  if (!box) throw new Error('Slot bounding box not available')
+  return {
+    x: box.x + box.width / 2,
+    y: box.y + box.height / 2
+  }
+}
```
DrJKL, line 16: Can you use [dragTo](https://playwright.dev/docs/input#drag-and-drop)?

DrJKL (changes requested):

browser_tests/fixtures/ComfyPage.ts
benceruleanlu: this has been removed

browser_tests/tests/vueNodes/linkInteraction.spec.ts
benceruleanlu: sure

browser_tests/fixtures/ComfyPage.ts
benceruleanlu: I moved it to a helper

christian-byrne (approved):
LGTM, appreciate adding e2e tests a ton, will give more confidence going forward.

PR #5670, opened 2025-09-19
fix: prevent splitter overlay from loading in focus mode

### LINKED ISSUE #5658: Bug: "Focus Mode(F)" Front-end performance is abnormal
### Prerequisites

### What happened?

In version 1.25, focus mode works correctly when the gallery and command line windows are open. 
In version 1.26, the gallery and command line windows cannot be popped back.

Not sure if it's a problem on my machine.

## Video

https://github.com/user-attachments/assets/5bcf3970-17e7-41f4-b499-1240dc367f06

> version: ComfyUI_frontend v1.25.11 

https://github.com/user-attachments/assets/8cd66cff-f223-4cc0-8cb1-b92654e99b24

> version: ComfyUI_frontend v1.26.11 

### Steps to Reproduce

1. 
2. 
3. 

Crashes ComfyUI completely

### ComfyUI Frontend Version

1.26.11

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

Myestery:
This pull request updates the display logic for the `LiteGraphCanvasSplitterOverlay` component in `GraphCanvas.vue`. The change ensures that the splitter overlay is only shown when the application is ready, the beta menu is enabled, and the workspace is not in focus mode.

Conditional rendering improvement:

* The `LiteGraphCanvasSplitterOverlay` component is now hidden when `workspaceStore.focusMode` is active, preventing it from being displayed in focus mode. (`src/components/graph/GraphCanvas.vue`)
Fixes #5658 

https://github.com/user-attachments/assets/b428246a-853f-4558-b694-337a1f75e483

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5670-fix-prevent-splitter-overlay-from-loading-in-focus-mode-2736d73d3650813fb505fea2f27dc6ca) by [Unito](https://www.unito.io)


DrJKL (dismissed):
So the minimap/controls stuff will be a follow-up?

Myestery:
> * View Report

Let me confirm the expected behaviour internally and revert

> So the minimap/controls stuff will be a follow-up?

src/components/graph/GraphCanvas.vue
```diff
@@ -15,22 +17,16 @@
...
       </div>
-      <GraphCanvasMenu v-if="canvasMenuEnabled" class="pointer-events-auto" />
-
-      <MiniMap
-        v-if="comfyAppReady && minimapEnabled"
-        class="pointer-events-auto"
-      />
     </template>
   </LiteGraphCanvasSplitterOverlay>
-  <GraphCanvasMenu v-if="!betaMenuEnabled && canvasMenuEnabled" />
```
christian-byrne, line 26: I think we accidentally dropped the `!betaMenuEnabled` check here.

src/components/graph/GraphCanvas.vue
```diff
@@ -15,22 +17,16 @@
...
+  <GraphCanvasMenu v-if="canvasMenuEnabled" class="pointer-events-auto" />
+  <MiniMap v-if="comfyAppReady && minimapEnabled" class="pointer-events-auto" />
```
christian-byrne, line 23: Similar to above, the `MiniMap` was previously inside of a parent that only rendered when `!betaMenuEnabled`, so now we may need to move that condition directly into the `v-if` here.

christian-byrne (commented):
Given the massive number of regenerated snapshots, I wonder if it is feasible to apply my suggestions (assuming you agree that they are valid) then open a new PR (given it's only 5 or so lines changes)? WDYT?

src/components/graph/GraphCanvas.vue
Myestery: so I dropped it because we had 2 Graphcanvas components and thebetaMenuEnabled check was logically irrelevant

christian-byrne: 2 GraphCanvasMenu components? Apologies for not understanding

christian-byrne:
There seems to be merge conflicts.

PR #5674, opened 2025-09-20
[style] enforce vue block order

christian-byrne:
## Summary

Standardizes every Vue SFC so their top-level blocks line up with our new linted order, keeping structure predictable for future edits.

## Changes

- **Config** Added [vue/block-order](https://eslint.vuejs.org/rules/block-order.html) with `['docs', 'script', 'template', 'i18n', 'style']` so docs stay first and localization blocks remain ahead of styles.
- **Format** Ran `pnpm lint:fix` to reorder top-level blocks across the existing Vue files; only block placement changed, no logic edits.

## Review Focus

- Double-check the ESLint rule matches our intended ordering and doesn't conflict with other lint settings.
- Spot-check a couple of updated SFCs to confirm the diff is strictly block movement.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5674-style-enforce-vue-block-order-2746d73d36508125b3abe89a76a3b874) by [Unito](https://www.unito.io)

PR #5679, opened 2025-09-20
[style] enforce vue block order in pre-commit

christian-byrne:
## Summary

Adds a pre-commit safeguard that reshuffles staged Vue files into our desired block order before linting, helping us adopt the pattern without touching the whole repo yet. After a while, we can merge the followup:

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/5674

## Changes

- **Hook** Updated `lint-staged.config.js` so staged `.vue` files pass through `eslint --fix` with the `docs → script → template → i18n → style` [block-order rule](https://eslint.vuejs.org/rules/block-order.html).
- **Flow** Kept the existing format + lint + typecheck steps, ensuring the new hook just adds ordering on top of the current pipeline.

## Review Focus

- Validate the lint-staged command syntax (particularly the quoting around the inline rule).
- Confirm the hook only targets staged Vue files and leaves TS/JS handling unchanged.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5679-style-enforce-vue-block-order-in-pre-commit-2746d73d365081b2a9b0dfecf5a31d4d) by [Unito](https://www.unito.io)

### DIFF
--- a/lint-staged.config.js
+++ b/lint-staged.config.js
@@ -1,10 +1,19 @@
+/* eslint-disable @typescript-eslint/no-floating-promises */
+const vueBlockOrderRule =
+  "vue/block-order: ['error', {'order': ['docs', 'script', 'template', 'i18n', 'style']}]"
+
 export default {
   './**/*.js': (stagedFiles) => formatAndEslint(stagedFiles),
 
-  './**/*.{ts,tsx,vue,mts}': (stagedFiles) => [
+  './**/*.{ts,tsx,mts}': (stagedFiles) => [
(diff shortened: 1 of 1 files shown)


christian-byrne:
This is actually not necessary because we have an [action](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/.github/workflows/lint-and-format.yaml) that automatically applies lint fixes to commits - so I think we are good to close this and just merge

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/5674

PR #5695, opened 2025-09-20
Refactor: Composable disentangling

DrJKL:
## Summary

Prerequisite refactor/cleanup to use a global store instead of having nodes throw up events to a parent component that stores a reference to a singleton service that itself bootstraps and synchronizes with a separate service to maintain a partially reactive but not fully reactive set of states that describe some but not all aspects of the nodes on either the litegraph, the vue side, or both.

## Changes

- **What**: Refactoring, the behavior should not change.
- **Dependencies**: A type utility to help with Vue component props

## Review Focus

Is there something about the current structure that this could affect that would not be caught by our tests or using the application?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5695-Refactor-Composable-disentangling-2746d73d365081e6938ce656932f3e36) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/core/canvas/canvasStore.ts
+++ b/src/renderer/core/canvas/canvasStore.ts
@@ -99,6 +99,16 @@ export const useCanvasStore = defineStore('canvas', () => {
   const currentGraph = shallowRef<LGraph | null>(null)
   const isInSubgraph = ref(false)
 
+  // Provide selection state to all Vue nodes
+  const selectedNodeIds = computed(
+    () =>
+      new Set(
+        selectedItems.value
+          .filter((item) => item.id !== undefined)
+          .map((item) => String(item.id))
+      )
+  )
+
   whenever(
     () => canvas.value,
     (newCanvas) => {
@@ -122,6 +132,7 @@ export const useCanvasStore = defineStore('canvas', () => {
   return {
     canvas,
     selectedItems,
+    selectedNodeIds,
     nodeSelected,
     groupSelected,
     rerouteSelected,
--- a/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
+++ b/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
@@ -1,0 +1,93 @@
+import { type MaybeRefOrGetter, computed, ref, toValue } from 'vue'
+
+import type { VueNodeData } from '@/composables/graph/useGraphNodeManager'
+import { useCanvasInteractions } from '@/renderer/core/canvas/useCanvasInteractions'
+import { layoutStore } from '@/renderer/core/layout/store/layoutStore'
+import { useNodeLayout } from '@/renderer/extensions/vueNodes/layout/useNodeLayout'
+
+// Treat tiny pointer jitter as a click, not a drag
+const DRAG_THRESHOLD_PX = 4
+
+export function useNodePointerInteractions(
+  nodeDataMaybe: MaybeRefOrGetter<VueNodeData>,
+  onPointerUp: (
+    event: PointerEvent,
+    nodeData: VueNodeData,
+    wasDragging: boolean
+  ) => void
+) {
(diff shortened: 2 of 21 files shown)


src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -0,0 +1,93 @@
+import { type MaybeRefOrGetter, computed, ref, toValue } from 'vue'
```
arjansingh, line 1: Question (non-blocker):

Can you follow up with these tests? I think it's the most common places where stuff could break or degrade here.

```js
describe('useNodePointerInteractions', () => {
    describe('layout store integration', () => {
      it('should call onNodeDragStart when drag begins', () => {
        // Test that the callback fires when we start dragging
        // Check timing and that it's called exactly once
      })

      it('should call onNodeDrag with correct deltas during drag', () => {
        // Test deltaX/deltaY calculations from last position
        // Make sure coordinates are relative, not absolute
      })

      it('should call onNodeDragEnd when drag completes', () => {
        // Test cleanup callback on pointer up after actual drag
        // Should not fire for clicks (sub-threshold movement)
      })
    })

    describe('canvas interaction coordination', () => {
      it('should forward pointer events when canvas is dragging', () => {
        // When canvas.isDragging is true, events should bubble up
        // Test that node doesn't interfere with canvas panning
      })
    })

    describe('drag threshold logic', () => {
      it('should not set isDragging when movement is below threshold', () => {
        // Movement under 4px should stay as potential click
        // Test with 0px, 1px, 3.9px movements
      })

      it('should set isDragging when movement exceeds threshold', () => {
        // 4px+ movement should trigger drag state
        // Test exactly at 4px and above
      })
    })

    describe('event handling sequence', () => {
      it('should handle rapid successive pointer events', () => {
        // Multiple move events in quick succession
        // Don't break state or cause performance issues
      })

      it('should handle pointer move before pointer down', () => {
        // Graceful handling of weird event orders
        // Browsers sometimes send events out of sequence
      })

      it('should prevent default behavior appropriately', () => {
        // preventDefault() to avoid text selection while dragging
        // Test on pointer down to prevent unwanted browser behavior
      })
    })

    describe('Vue reactivity integration', () => {
      it('should transition isDragging state correctly', () => {
        // false -> true -> false transitions should trigger updates
        // Test that UI responds to state changes
      })

      it('should update dragStyle reactively', () => {
        // "grab" -> "grabbing" -> "grab" cursor changes
        // Test computed property updates with isDragging
      })
    })

    describe('performance characteristics', () => {
      it('should handle high-frequency pointer moves efficiently', () => {
        // Test 60+ events per second don't slow things down
        // Important for smooth dragging experience
      })

      it('should not cause memory leaks with event handlers', () => {
        // Cleanup on unmount, no dangling references
        // Test multiple mount/unmount cycles
      })
    })

    describe('coordinate calculations', () => {
      it('should use Euclidean distance for threshold detection', () => {
        // Test sqrt(deltaX² + deltaY²) calculation
        // Diagonal movements should work correctly
      })
    })
  })
```

src/renderer/core/canvas/canvasStore.ts
```diff
@@ -99,6 +99,16 @@ export const useCanvasStore = defineStore('canvas', () => {
   const currentGraph = shallowRef<LGraph | null>(null)
   const isInSubgraph = ref(false)
 
+  // Provide selection state to all Vue nodes
+  const selectedNodeIds = computed(
```
christian-byrne, line 103: nit: I know you are just moving this and didn't name it yourself, but just realized this will also include groups and reroutes, so the name may be slightly misleading.

christian-byrne (approved):
LGTM!

src/renderer/core/canvas/canvasStore.ts
DrJKL: I'm actually changing that in the next PR!

PR #5708, opened 2025-09-20
Migrate parentIds when converting to subgraph

### LINKED ISSUE #5669: Can't unpack subgraph with a reroute inside
### Prerequisites

### What happened?

TypeError: can't access property "type", app2.graph.links[l] is undefined

### Steps to Reproduce

1. add a reroute
2. convert to subgraph
3. can't unpack subgraph

Image

### ComfyUI Frontend Version

1.28.0

### Browser

Firefox

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-5669-Can-t-unpack-subgraph-with-a-reroute-inside-2736d73d365081a29a02d51a0cd8c1a7) by [Unito](https://www.unito.io)

AustinMroz:
The parentId property on links and reroutes was not handled at all in the "Convert to Subgraph" code.
This needs to be addressed in 4 cases
- A new external input link must have parentId set to the first non-migrated reroute
- A new external output link must have the parentId of it's eldest remaining child set to undefined
- A new internal input link must have the parentId of it's eldest remaining child set to undefined
- A new internal output link must have the parentId set to the first migrated reroute

This is handled in two parts by adding logic where the boundry links is created
- The change involves mutation of inputs (which isn't great) but the function here was already mutating inputs into an invalid state
  - @DrJKL Do you see a quick way to better fix both these cases?

Looks like litegraph tests aren't enabled and cursory glance shows multiple need to be updated to reflect recent changes. I'll still try to add some tests anyways.
EDIT:  Tests are non functional. Seems the subgraph conversion call requires the rest of the frontend is running and has event listeners to register the subgraph node def. More work than anticipated, best revisited later

Resolves #5669

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5708-Migrate-parentIds-when-converting-to-subgraph-2746d73d365081f78acff4454092c74a) by [Unito](https://www.unito.io)


DrJKL:
Often, the resolution to mutating inputs is to introduce an intermediate pure function that returns the new desired state.
So for this case...
1. Take in the link to split
4. Generate the new links
99. Return the set of new links that direct 
    - To the Subgraph externally
    - From the Subgraph Input internally
23. Swap out the original with the replacements

Right?

Another benefit of devolving the structure like this is that the logic can be tested in isolation and then composed.

I feel like there's a way to handle reroutes as links under the hood, but I'd have to look into it further to really check.

AustinMroz:
Appreciate the insight.

> I feel like there's a way to handle reroutes as links under the hood, but I'd have to look into it further to really check.

My understanding is that it was an explicit design goal for the new reroutes to be transparent to the linking logic. It's always safe to pretend they don't exist unless the code you're working with specifically involves reroutes. But that's more a question for @webfiltered, IIRC.

webfiltered:
> My understanding is that it was an explicit design goal for the new reroutes to be transparent to the linking logic. It's always safe to pretend they don't exist unless the code you're working with specifically involves reroutes. But that's more a question for @webfiltered, IIRC. 

Definitely by design.  Reroutes are not links; the original code treating them as links was the cause of extreme misery and woe.

They are virtual waypoints, and the new system treats them as such.

Definitely room to improve the reroute code though; it had legacy constraints.

src/lib/litegraph/src/subgraph/subgraphUtils.ts
```diff
@@ -274,8 +275,19 @@ export function mapSubgraphInputsAndLinks(
...
+      let child: SerialisableLLink | Reroute = linkData
+      let nextReroute = reroutes.get(child.parentId ?? 0)
+      while (child.parentId && nextReroute) {
+        child = nextReroute
+        nextReroute = reroutes.get(child.parentId ?? 0)
+      }
+      //set outside link to first remaining reroute OR undefined
+      link.parentId = child.parentId
+      //ensure end of chain is undefined
+      child.parentId = undefined
```
webfiltered: `undefined` is the idiomatic JS for "not set" / "none" - `0` is actually a valid reroute ID.

The use of `0` as a special value in LiteGraph is not something you see in libraries (for good reason - it allows you to use truthy checks, but breaks **numeric** checks).  Even base JS uses -1 in its always-numeric built-ins.

webfiltered (commented):
Off the top of my head, this looks like it will be corrupt if you convert to subgraph when using "part" of a reroute chain.

Scenarios I wanted to address a while back:

1. A node with an input link that goes outside the subgraph + the "second" reroute only
   The closest reroute is NOT selected
3. A node with an output link that goes outside, taking the first, third, and fifth reroutes
   Skip the second and fourth reroutes
5. Two connected nodes, but only some / none of their reroutes
   - v1: Just bring ALL the reroutes in regardless
7. Pick only a reroute - not the node before or the node after
   - v1: Stupid extreme edge case, just ignore it and pretend you didn't see it

I am aware this is awful.

webfiltered:
> Scenarios I wanted to address a while back:

Ah lost some other words that were meant to come before this.  TL;DR would have been: Reroutes were put in the too-hard basket and this limitation was acknowledged somewhere.  Sorry~

src/lib/litegraph/src/subgraph/subgraphUtils.ts
webfiltered: (Also thanks! Decided to make sure I wasn't losing my mind, and found a bad refactor of mine from Feb.)

AustinMroz: IT IS?  `lastRerouteId` starts at 0, is incremented before use, and there's existing code sections that depend on valid `rerouteId`s always being truthy
```rg
src/LLink.ts
208:    if (!linkSegment.parentId) return []
232:    if (!linkSegment.parentId) return
501:    if (this.parentId) copy.parentId = this.parentId
```

This is one of the cases where js will happily let you index by `undefined` for correct behaviour, and I'm just hoping to satisfy typescript. It can be easily changed  to a ternary if it feels ick to you.

> Off the top of my head, this looks like it will be corrupt if you convert to subgraph when using "part" of a reroute chain.

Ran into a couple of these when trying to write test cases here. Yeah... it gets awful fast. If I'm following along right, 'ghost' reroutes (with no links) are not intended. A reroute that isn't linked through a parentId can be selected, but isn't drawn. There's some actions which can cause them, like a  "Right click slot -> Disconnect Links".

Worthyness of implementation time aside, I'd expect a robust implementation to be a single rule that handles all these cases. Automatic inclusion seems both more UX friendly and seems to have the easier implementation of the 2.
1. Walk the chain of parentIds for every link and reroute, only append chain if it reaches a selected item
2. Assign each reroute a (stable) depth from the reversed parentId list. Sort and rebuild parentId chain from depth for each reroute

Worth editing to the unpacking code to `console.error; continue'`when a chain has an unexpected dead end? It shouldn't actually break anything, but feels like giving up.

webfiltered: > IT IS? `lastRerouteId` starts at 0,

Yeah I forgot I did this - I -think- it was added in.  I found it after I wrote the comment, but decided not to correct myself - there may be workflows from a couple versions that would use the old reroute ID 0 and cause unexpected behaviour.

The TL;DR is that the API design is still that ID 0 is valid - the current impl. detail is that it will not be on workflows created right now.  There should be many more code paths that do nullish checks.  So trying to have 0 also mean "not set" will result in -- what's the technical -- _weird shit happening_.  😃

(It looks like you were typing that response at the same time as I was doing my refactor - and I missed it~)

webfiltered: > 'ghost' reroutes (with no links) are not intended

Yep - that's nasty.

> Automatic inclusion seems both more UX friendly

SGTM!

> Worth editing to the unpacking code to `console.error; continue'`

Pretty much always worth logging unexpected state, imo.  Actionable information beats silently breaking for me.  Discarding error state is a pet hate of mine - it makes debugging impossible.

Spent a lot of time this year making sure LiteGraph absolutely spits the dummy in really obvious ways when things break.  Throw, log, or.. do _something_.  Initial work on subgraph had an insane difficulty spike at the start, due to this - tracing _why_ a change seemed to do nothing just took forever (no errors, no warnings.. just.. didn't work).

> Yeah... it gets awful fast.

Just QFT - sounds like a great tagline / subtitle.

src/lib/litegraph/src/subgraph/subgraphUtils.ts
```diff
@@ -354,9 +372,21 @@ export function mapSubgraphOutputsAndLinks(
...
+        child.parentId === undefined ? undefined : reroutes.get(child.parentId)
+      while (child.parentId !== undefined && nextReroute) {
+        child = nextReroute
+        nextReroute =
+          child.parentId === undefined
+            ? undefined
+            : reroutes.get(child.parentId)
+      }
+      linkData.parentId = child.parentId
+      child.parentId = undefined
```
DrJKL: Deja vu...

src/lib/litegraph/src/LGraph.ts
```diff
@@ -1566,8 +1566,21 @@ export class LGraph
 
     // Inputs, outputs, and links
     const links = internalLinks.map((x) => x.asSerialisable())
-    const inputs = mapSubgraphInputsAndLinks(resolvedInputLinks, links)
-    const outputs = mapSubgraphOutputsAndLinks(resolvedOutputLinks, links)
+
+    const internalReroutes = new Map([...reroutes].map((r) => [r.id, r]))
```
DrJKL, line 1575: Optional: https://es-toolkit.dev/reference/array/groupBy.html#groupby

DrJKL (dismissed):
I'd love to see the duplicate code consolidated, but it's not huge.

DrJKL (approved):

PR #5710, opened 2025-09-21
Increase vue slot/link functionality

benceruleanlu:
## Summary

Increase functionality for slots and links, covered with playwright tests.

## Features

- Allow for reroute anchors to work when dragging from input slot
- Allow for dragging existing links from input slot
- Allow for ctrl/command + alt to create new link from input slot
- Allow shift to drag all connected links on output slot
- Connect links with reroutes (only when dragged from vue slot)

### Playwright

- Dragging input to input drags existing link
- Dropping an input link back on its slot restores the original connection
- Ctrl+alt drag from an input starts a fresh link
- Should reuse the existing origin when dragging an input link
- Shift-dragging an output with multiple links should drag all links
- Rerouted input drag preview remains anchored to reroute
- Rerouted output shift-drag preview remains anchored to reroute

## Notes

The double rendering system for links being dragged, it works right now, maybe they can be coalesced later.

Edit: As in the adapter, can be removed in a followup PR

Also, it's known that more features will arrive in smaller PRs, this PR actually should've been much smaller.

The next ones coming up are drop on canvas support, snap to node, type compatibility highlighting, and working with subgraphs.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5710-Increase-vue-slot-link-functionality-2756d73d3650814f8995f7782244803b) by [Unito](https://www.unito.io)


benceruleanlu:
the two failing tests are screenshot expectations that I added that were bugged from the overlapping dirty workflow dot indicator and the close workflow button

src/lib/litegraph/src/canvas/MovingInputLink.ts
christian-byrne, line 1: question (non-blocking): Can you explain these changes briefly?

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -104,10 +127,161 @@ export function useSlotLinkInteraction({
...
+
+  function isOutputConnectableLink(
+    link: RenderLink
+  ): link is OutputConnectableLink {
+    return (
+      link.toType === 'output' &&
+      typeof (link as { canConnectToOutput?: unknown }).canConnectToOutput ===
+        'function'
+    )
+  }
```
christian-byrne: suggestion (non-blocking): We could consider using more robust type guards and/or discriminated unions for link types.

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -92,10 +103,22 @@ export function useSlotLinkInteraction({
...
+            layout.nodeId,
+            layout.index
+          )
+        } else if (layout.type === 'output') {
+          candidate.compatible = adapter.isOutputValidDrop(
+            layout.nodeId,
+            layout.index
+          )
+        }
+      }
```
christian-byrne, line 121: question: What is the lifecycle of the adapter like? Is it necessary to have explicit cleanup?

src/lib/litegraph/src/canvas/MovingInputLink.ts
```diff
@@ -73,8 +73,30 @@ export class MovingInputLink extends MovingLinkBase {
     return link
   }
 
-  connectToOutput(): never {
-    throw new Error('MovingInputLink cannot connect to an output.')
+  connectToOutput(
+    outputNode: LGraphNode,
+    output: INodeOutputSlot,
+    events: CustomEventTarget<LinkConnectorEventMap>
+  ): LLink | null | undefined {
```
christian-byrne: question: Is it possible to make the meaning of the return types more explicit?

The API contract might be somewhat confusing if `undefined` and `null` have different semantics. If they are the same, would it be possible to only return `null` or only return `undefined`?

browser_tests/tests/vueNodes/interactions/links/linkInteraction.spec.ts
```diff
@@ -132,90 +173,518 @@ test.describe('Vue Node Link Interaction', () => {
...
-      `[data-slot-key="${outputSlotKey}"]`
+    const outputPosition = await samplerOutput.getPosition()
+    const inputPosition = await vaeInput.getPosition()
+    const reroutePoint = getMiddlePoint(outputPosition, inputPosition)
+
+    // Insert a reroute programmatically on the existing link between sampler output[0] and VAE input[0].
+    // This avoids relying on an exact path hit-test position.
+    await comfyPage.page.evaluate(
+      ([targetNodeId, targetSlot, clientPoint]) => {
+        const app = (window as any)['app']
```
christian-byrne, line 423: nit (non-blocking): could we use "expect-error" directive so this is easier to cleanup/refactor/audit later on?

src/renderer/core/canvas/links/slotLinkPreviewRenderer.ts
```diff
@@ -39,57 +36,74 @@ export function attachSlotLinkPreviewRenderer(canvas: LGraphCanvas) {
...
-    const from: ReadOnlyPoint = [start.x, start.y]
-    const to: ReadOnlyPoint = [pointer.canvas.x, pointer.canvas.y]
-
-    const startDir = source.direction ?? LinkDirection.RIGHT
-    const endDir = LinkDirection.CENTER
-
-    const colour = resolveConnectingLinkColor(sourceSlot?.type)
+    const adapter = createLinkConnectorAdapter()
+    const renderLinks = adapter?.renderLinks
+    if (!adapter || !renderLinks || renderLinks.length === 0) return
```
christian-byrne: nit: if `adapter` is falsy here, then so is `renderLinks` by necessity.

christian-byrne (commented):
LGTM with nits

src/lib/litegraph/src/canvas/MovingInputLink.ts
benceruleanlu: this was for testing/iteration, and was supposed to be removed in the 'those who know' commit, but wasn't cleaned up fully

it didn't have any callers, I just removed it now

src/lib/litegraph/src/canvas/MovingInputLink.ts
benceruleanlu: this code was removed after addressing other review comments

src/renderer/core/canvas/links/slotLinkPreviewRenderer.ts
benceruleanlu: fixed

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: true, changed this to intersection typed, and I think a true discriminated union for RenderLink can be done in another PR

browser_tests/tests/vueNodes/interactions/links/linkInteraction.spec.ts
benceruleanlu: wouldn't that get flagged when linting? also, I think the PR that will switch off of as any will have a simple script to *prepend* with expect error

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
benceruleanlu: ignoring the quoted code, the adapter is a weakmap per graph

we need the explicit cleanup because the adapter has its own linkconnector, so the regular linkconnector won't reset for us, and reset is used for things like _dragging flags and anything using listenUntilReset

src/lib/litegraph/src/canvas/MovingInputLink.ts
christian-byrne: Was I supposed to know that?

christian-byrne:
There seems to be some merge conflicts.

src/lib/litegraph/src/canvas/MovingInputLink.ts
benceruleanlu: no, you weren't, and it wasn't obvious

that is a mistake on my end, and I apologize for it

I always aim to do better, so in this case I'm taking away the lesson of having old code that should've been removed, at the detriment of an easy to review pr

benceruleanlu:
the failing tests are the screenshot expectations from the newly added tests, adding the regeneration label now

christian-byrne (approved):
LGTM!

webfiltered (commented):
It appears that the test failures remain after the screenshots were added.  I'm starting to wonder if we should just remove the `[skip ci]` tag from the screenshot add commit.

webfiltered:
> I'm starting to wonder if we should just remove the `[skip ci]` tag from the screenshot add commit.

Just saw the dates on the commits - might not have made a difference, here.

PR #5714, opened 2025-09-21
fix using shift modifier to (de-)select Vue nodes

### LINKED ISSUE #5688: Can't select multiple VUE nodes at the same time
https://github.com/user-attachments/assets/6c434e2b-dc1e-48ea-8f6d-3377c91f021b

christian-byrne:
## Summary

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/5688 by adding shift modifier support for multi-selecting Vue nodes, enabling standard shift+click selection behavior alongside existing ctrl/cmd+click.

## Changes

- **What**: Updated Vue node event handlers to include `event.shiftKey` in multi-select logic
- **Testing**: Added browser tests for both ctrl and shift modifier selection behaviors

## Review Focus

Multi-select behavior consistency across different input modifiers and platform compatibility (Windows/Mac/Linux shift key handling).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5714-fix-using-shift-modifier-to-de-select-Vue-nodes-2756d73d365081bcb5e0fe80eacdb2f0) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/vueNodes/nodeInteractions/selectionState.spec.ts
+++ b/browser_tests/tests/vueNodes/nodeInteractions/selectionState.spec.ts
@@ -1,0 +1,47 @@
+import {
+  comfyExpect as expect,
+  comfyPageFixture as test
+} from '../../../fixtures/ComfyPage'
+
+test.describe('Vue Node Selection', () => {
+  test.beforeEach(async ({ comfyPage }) => {
+    await comfyPage.setSetting('Comfy.VueNodes.Enabled', true)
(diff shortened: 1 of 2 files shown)


browser_tests/tests/vueNodes/nodeInteractions/selectionState.spec.ts
christian-byrne, line 1: These tests could be better because they just test the count of selected nodes matches expected number, but it could easily also test the specific clicked node is the one that is selected.

@christian-byrne

PR #5718, opened 2025-09-22
[refactor] Simplify current user resolved hook implementation

christian-byrne:
## Summary

Refactored `onUserResolved` function in auth composable to use VueUse `whenever` utility instead of manual watch implementation and use `immediate` option instead of invoking manually before creating watcher.

## Changes

- **What**: Replaced manual watch + immediate check pattern with [VueUse whenever](https://vueuse.org/shared/whenever/) utility in `useCurrentUser.ts:37`

## Review Focus

Behavioral equivalence verification - `whenever` with `immediate: true` should maintain identical callback timing and cleanup semantics.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5718-refactor-Simplify-current-user-resolved-hook-implementation-2766d73d365081008b6de156dd78f940) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/auth/useCurrentUser.ts
+++ b/src/composables/auth/useCurrentUser.ts
@@ -1,4 +1,5 @@
-import { computed, watch } from 'vue'
+import { whenever } from '@vueuse/core'
+import { computed } from 'vue'
 
 import { useFirebaseAuthActions } from '@/composables/auth/useFirebaseAuthActions'
 import { t } from '@/i18n'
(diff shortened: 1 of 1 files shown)


christian-byrne:
If you ask, "why not just set the watcher in the consumer?", the answer is that the consumer is probably not using Vue.

DrJKL (approved):

PR #5720, opened 2025-09-22
Allow Vue nodes to have their colors changed from selection toolbox

christian-byrne:
## Summary

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/5680 by allowing Vue nodes to properly synchronize color changes with LiteGraph nodes, implementing header darkening and light theme adjustments.

Screenshot from 2025-09-21 20-00-36

## Changes

- **What**: Implemented color property synchronization between LiteGraph and Vue node rendering systems
- **Core Fix**: Added `nodeData.color` and `nodeData.bgcolor` to [v-memo dependencies](https://vuejs.org/api/built-in-directives.html#v-memo) to trigger re-renders on color changes
- **Color Logic**: Added header darkening using [memoized color adjustments](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/utils/colorUtil.ts) to match LiteGraph's ColorOption system
- **Event System**: Enhanced property change instrumentation in LGraphNode.setColorOption to emit color/bgcolor events

## Review Focus

Vue component reactivity timing - the v-memo fix was critical for immediate color updates. Verify light theme color adjustments match the drawNode monkey patch behavior in app.ts.

## Technical Details

```mermaid
graph TD
    A[User Sets Color] --> B[LGraphNode.setColorOption]
    B --> C[Sets node.color & node.bgcolor]
    C --> D[Triggers property:changed events]
    D --> E[Vue Node Manager Updates]
    E --> F[v-memo Detects Change]
    F --> G[NodeHeader Re-renders]
    G --> H[Header Darkening Applied]

    style H fill:#f9f9f9,stroke:#333,color:#000
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5720-Allow-Vue-nodes-to-have-their-colors-changed-from-selection-toolbox-2766d73d36508123b441d126a74a54b2) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/NodeHeader.vue
```diff
@@ -111,6 +114,30 @@ const tooltipConfig = computed(() => {
...
+  // Apply base header darkening to replicate LiteGraph's ColorOption system
+  // When header and body colors are the same/similar, darken the header
+  if (nodeData.bgcolor && nodeData.color === nodeData.bgcolor) {
+    // Darken header relative to body (opposite of light theme adjustment)
+    headerColor = adjustColor(nodeData.color, { lightness: -0.15 })
+  }
+
+  // Apply light theme lightening on top of base darkening (same as drawNode monkey patch)
+  if (colorPaletteStore.completedActivePalette.light_theme) {
+    headerColor = adjustColor(headerColor, { lightness: 0.5 })
```
christian-byrne: Is there a better way to do this with CSS? `adjustColor` is memoized though.

browser_tests/tests/vueNodes/nodeStates/colors.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+  // TODO: implement loading node colors from workflow in Vue system
+  test.fail('should load node colors from workflow', async ({ comfyPage }) => {
+    await comfyPage.loadWorkflow('nodes/every_node_color')
+    await expect(comfyPage.canvas).toHaveScreenshot(
+      'vue-node-custom-colors-dark-all-colors.png'
+    )
+  })
+
+  // TODO: implement loading node colors from workflow in Vue system
+  test.fail(
```
christian-byrne: Expect these to fail until loading custom colors from workflow is implemented.

src/renderer/extensions/vueNodes/components/NodeHeader.vue
Myestery: color-mix() with white ?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -32,7 +32,9 @@
     :style="[
       {
         transform: `translate(${layoutPosition.x ?? position?.x ?? 0}px, ${(layoutPosition.y ?? position?.y ?? 0) - LiteGraph.NODE_TITLE_HEIGHT}px)`,
-        zIndex: zIndex
+        zIndex: zIndex,
+        // Explicitly set backgroundColor to ensure reset works properly
```
Myestery:
```suggestion
       color: nodeData.color || ''
```
Where are we setting color?

browser_tests/tests/vueNodes/nodeStates/colors.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+    await loadCheckpointNode.getByText('Load Checkpoint').click()
+
+    await comfyPage.page.locator('.selection-toolbox .pi-circle-fill').click()
+    await comfyPage.page
+      .locator('.color-picker-container')
+      .locator('i[data-testid="blue"]')
+      .click()
+
+    await expect(comfyPage.canvas).toHaveScreenshot(
+      'vue-node-custom-color-blue.png'
```
Myestery, line 29: Aside from the screenshot, can we have another test that reads the color or backgroundColor prop from the node?

src/renderer/extensions/vueNodes/components/NodeHeader.vue
DrJKL: Like this? https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_colors/Relative_colors

src/renderer/extensions/vueNodes/components/LGraphNode.vue
DrJKL: The naming in Litegraph is poor.
They're both surface colors, one for the header, one for the body, neither for the text.

DrJKL (dismissed):

browser_tests/tests/vueNodes/nodeStates/colors.spec.ts
DrJKL: What additional signal would that provide?

src/renderer/extensions/vueNodes/components/NodeHeader.vue
christian-byrne: It's safe to assume that would be more performant right? Not sure performance different is measurable though, so maybe other pros/cons should be considered. Thoughts?

DrJKL: If we were able to bind to a CSS variable and derive from it a lighter and a darker shade, we could eliminate all of this logic

christian-byrne: We want to support any arbitrary color set by extensions though.

DrJKL: It would work, the css variable can be dynamic.

christian-byrne: TY, let me try tomorrow

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -332,13 +332,39 @@ export class LGraphNode
...
+        })
+      }
+      if (oldBgcolor !== newBgcolor) {
+        this.graph.trigger('node:property:changed', {
+          nodeId: this.id,
+          property: 'bgcolor',
+          oldValue: oldBgcolor,
+          newValue: newBgcolor
+        })
+      }
```
DrJKL: I think if you add `color` and `bgcolor` [here](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/lib/litegraph/src/LGraphNodeProperties.ts#L9) you can remove this section.

christian-byrne: Thanks, done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5720/commits/0491007c0f73fac91781022a02b7301aa84e0877

browser_tests/tests/vueNodes/nodeStates/colors.spec.ts
christian-byrne: They work now

src/renderer/extensions/vueNodes/components/NodeHeader.vue
christian-byrne: The logic is a bit convulated when trying to do in CSS especially since we need to serialize color in the workflow, decided against it for this PR.

PR #5721, opened 2025-09-22
Refactor: More state management simplification

DrJKL:
## Summary

Remove more procedural synchronization in favor of using reactive references.

> Note: Also includes some fixes for issues caused during HMR.

## Review Focus

In testing it seems to work the same, but let me know if I missed something.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5721-Refactor-More-state-management-simplification-2766d73d3650819b8d7ddc047c460f2b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useViewportCulling.ts
+++ b/src/composables/graph/useViewportCulling.ts
@@ -6,21 +6,18 @@
  * 2. Set display none on element to avoid cascade resolution overhead
  * 3. Only run when transform changes (event driven)
  */
+import { useRafFn } from '@vueuse/core'
 import { computed } from 'vue'
 
 import { useVueNodeLifecycle } from '@/composables/graph/useVueNodeLifecycle'
-import { useVueFeatureFlags } from '@/composables/useVueFeatureFlags'
 import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
 import { app as comfyApp } from '@/scripts/app'
 
 export function useViewportCulling() {
   const canvasStore = useCanvasStore()
-  const { shouldRenderVueNodes } = useVueFeatureFlags()
-  const { vueNodeData, nodeDataTrigger, nodeManager } = useVueNodeLifecycle()
+  const { vueNodeData, nodeManager } = useVueNodeLifecycle()
 
   const allNodes = computed(() => {
-    if (!shouldRenderVueNodes.value) return []
-    void nodeDataTrigger.value // Force re-evaluation when nodeManager initializes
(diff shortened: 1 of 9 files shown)


DrJKL:
The function expression &rarr; declaration change is half stylistic and half precautionary. The bug shows up when the event listener holds on to a reference that's further down in scope and gets wiped.

christian-byrne (dismissed):
Very nice

src/composables/graph/useViewportCulling.ts
```diff
@@ -71,26 +68,9 @@ export function useViewportCulling() {
...
-      cancelAnimationFrame(rafId)
-    }
-
-    // Schedule update in next animation frame
-    rafId = requestAnimationFrame(() => {
-      updateVisibility()
-      rafId = null
-    })
-  }
+  const { resume: handleTransformUpdate } = useRafFn(updateVisibility, {
```
christian-byrne: Shall we expose the `pause()` function? We can pause when the TransformPane emits the "stopped transforming" event (or something similar).

DrJKL: I actually want to get rid of this entirely and just have the nodes react to changes in the transformPane (throttled)

christian-byrne: Isn't that what was happening already?

DrJKL: Nope, it's querying the document with every call.

christian-byrne: Looking at it, I initially thought `handleTransformUpdate` was only called when a transform event is emitted by `TransformPane` which would then trigger an `updateVisibility`.

I thought `resume` from `useRafFn` essentially started calling the rafFn every frame until `pause` is called. So now, we are `resume` when transform is emitted -> starting a call every frame. But before, we just called on every frame that a transform is emitted. Is that wrong?

christian-byrne: Well, the `TransformPane` is emitting update event every frame anyway, because it's using `useCanvasTransformSync` which just simply uses its own rafFn that runs every frame, so I'm incorrect afterall.

christian-byrne: I tested by adding `console.count` call in `handleTransformUpdate`

DrJKL: You were right locally, but that does explain why this code didn't tank performance relative to baseline like it would have otherwise 😁

christian-byrne (approved):

PR #5731, opened 2025-09-23
Make Id types mutually incompatible

AustinMroz:
NodeId, NodeExecutionId, and  NodeLocatorId can all be strings. NodeId, LinkId, and RerouteId can all be numbers. As a result, Typescript will freely allow these values to be cast between one another.

To force these types to be mutually incompatible, they are defined as being extended from "number" and "string" with an additional optional type property. Actually extending primitives is a bad idea. Simply declaring types as extending primitives feels a little ick. Feedback on if such a change should actually be merged, or simply used as a temporary change for testing is appreciated.

Throughout the entire code base, only 1 actual issue was found, which was far fewer than expected: Left clicking on the center dot of a link prior to a reroute will only delete one link instead of all grouped links
reroute-link-deletion

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5731-Make-Id-types-mutually-incompatible-2776d73d365081039b90d44046b7c4b2) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/fixtures/VueNodeHelpers.ts
+++ b/browser_tests/fixtures/VueNodeHelpers.ts
@@ -3,6 +3,8 @@
  */
 import type { Locator, Page } from '@playwright/test'
 
+import type { NodeId } from '@/lib/litegraph/src/LGraphNode'
+
 export class VueNodeHelpers {
   constructor(private page: Page) {}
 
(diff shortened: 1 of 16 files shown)


DrJKL:
I'd rather brand them with unique symbols, if feasible.

AustinMroz:
Agreed. That's a better approach.

I'd be nice if there was a way to also have the "value" of the "brand" property symbol to be a unique symbol itself, but this didn't seem possible from testing.

DrJKL:
> Agreed. That's a better approach.
> 
> I'd be nice if there was a way to also have the "value" of the "brand" property symbol to be a unique symbol itself, but this didn't seem possible from testing.

I would make a symbol adjacent to each type declaration instead of having the utility type.

PR #5732, opened 2025-09-23
Add base url to index.html

deepanjanroy:
cloud.comfy.org/cloud/code/XXXX-XXXX-XXXX was returning a white page. The problem is our assets are using relative path, and so we were requesting assets from cloud.comfy.org/cloud/code/assets/...

We have an SPA fallback handler on the ingest handler which resolves /cloud/ to /, so paths like cloud.comfy.org/cloud/part1 still worked. But it broke down when we have cloud.comfy.org/cloud/part1/part2. 

This PR adds a base url to the html so assets are loaded relative to "/". 

## Tests
Built frontend and copied to ingest server, and checked on local ingest server that it works (it was failing before). 

[Slack thread](https://comfy-organization.slack.com/archives/C08V9NDB3B4/p1758642698302389?thread_ts=1758593690.577669&cid=C08V9NDB3B4)

### DIFF
--- a/index.html
+++ b/index.html
@@ -3,19 +3,21 @@
   <head>
     <meta charset="UTF-8">
     <title>ComfyUI</title>
+    <!-- All assets should be loaded from the root no matter the initial path -->
+    <base href="/">
     <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
     <link rel="stylesheet" type="text/css" href="materialdesignicons.min.css" />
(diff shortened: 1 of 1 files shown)


christian-byrne (approved):

christian-byrne:
Noting: We can also change `base` in vite config if it makes things easier later on. Would be nice if cloud only needed extensions and its own build conifgs

PR #5746, opened 2025-09-24
test(e2e): align test default menu to Top; make legacy specs explicit

benceruleanlu:
## Summary
UseNewMenu has been defaulted to Top in the app for over a year; Playwright’s test default lagged behind. This PR aligns the test default with reality and keeps legacy specs stable.

## Changes
- tests(e2e): default to 'Top' via fixture; specs that previously relied on the old implicit default now explicitly set 'Comfy.UseNewMenu' to 'Disabled'.
- docs(browser-tests): remove outdated README note suggesting tests set 'Top' manually.

## Review Focus
- Intentional uses of 'Top' and 'Bottom' remain unchanged.
- Confirm ComfyPage default remains 'Top' (see browser_tests/fixtures/ComfyPage.ts).

N/A

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5746-test-e2e-align-test-default-menu-to-Top-make-legacy-specs-explicit-2786d73d365081218d06c1346f3ae18e) by [Unito](https://www.unito.io)


DrJKL (dismissed):

benceruleanlu:
the only failing test is just flaky

image

browser_tests/fixtures/ComfyPage.ts
```diff
@@ -1643,7 +1643,7 @@ export const comfyPageFixture = base.extend<{
 
     try {
       await comfyPage.setupSettings({
-        'Comfy.UseNewMenu': 'Disabled',
+        'Comfy.UseNewMenu': 'Top',
```
christian-byrne, line 1646: question (non-blocking): would it work to just remove this line altogether since this is the  default value?

browser_tests/README.md
```diff
@@ -56,14 +56,6 @@ TEST_COMFYUI_DIR=/path/to/your/ComfyUI
...
 ### Common Setup Issues
 
-**Most tests require the new menu system** - Add to your test:
-
-```typescript
-test.beforeEach(async ({ comfyPage }) => {
-  await comfyPage.setSetting('Comfy.UseNewMenu', 'Top')
-})
-```
-
```
christian-byrne, line 66: question (non-blocking): should we just change this to say like "Some tests require the old menu system"? Or maybe not because that should never be necessary anymore?

christian-byrne (approved):
LGTM + 2 questions!

browser_tests/fixtures/ComfyPage.ts
benceruleanlu: yeah I think so, good point, will make this change after merge since we have the setting to dismiss approvals after commits

browser_tests/README.md
benceruleanlu: I think it would be clear enough, since it would say that explicitly by setting it to Disabled, so I don't think this would be needed

PR #5754, opened 2025-09-24
Add Playwright composite action to reduce workflow duplication

snomiao:
This PR introduces a reusable composite action for Playwright setup to reduce duplication across workflows.

## Changes
- Created `.github/actions/setup-playwright/action.yml` composite action that:
  - Detects or uses provided Playwright version
  - Caches Playwright browsers with intelligent cache keys
  - Installs browsers only when cache miss occurs
  - Installs OS dependencies when cache hit occurs
  
## Technical Details
- **Important:** The composite action requires `shell: bash` for all `run` steps as per [GitHub Actions requirements for composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file). This is a mandatory field for composite actions, unlike regular workflow steps.
- Updated workflow paths to account for repository checkout locations (some workflows checkout to subdirectories like `ComfyUI_frontend/`)
- Uses conditional caching to avoid redundant browser installations

## Benefits
- Reduces code duplication across 6 workflow files
- Centralizes Playwright caching logic  
- Consistent browser setup across all workflows
- Easier maintenance and updates
- Faster CI runs through intelligent caching

## Affected Workflows
- `.github/workflows/test-ui.yaml` (2 uses)
- `.github/workflows/i18n-custom-nodes.yaml`
- `.github/workflows/i18n-node-defs.yaml`
- `.github/workflows/i18n.yaml`
- `.github/workflows/test-browser-exp.yaml`


.github/actions/setup-playwright/action.yml
```diff
@@ -0,0 +1,50 @@
...
+outputs:
+  playwright-version:
+    description: 'The detected Playwright version'
+    value: ${{ steps.detect-version.outputs.playwright-version }}
+runs:
+  using: composite
+  steps:
+    - name: Detect Playwright version
+      id: detect-version
+      shell: bash
```
snomiao, line 8: TODO: no need to specify a shell here

snomiao, line 8: - **Important:** The composite action requires `shell: bash` for all `run` steps as per [GitHub Actions requirements for composite actions](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action#creating-an-action-metadata-file). This is a mandatory field for composite actions, unlike regular workflow steps.

.github/workflows/i18n.yaml
```diff
@@ -25,16 +25,11 @@ jobs:
...
       with:
-        path: ~/.cache/ms-playwright
-        key: playwright-browsers-${{ runner.os }}-${{ hashFiles('ComfyUI_frontend/pnpm-lock.yaml') }}
-        restore-keys: |
-          playwright-browsers-${{ runner.os }}-
-    - name: Install Playwright Browsers
-      run: pnpm exec playwright install chromium --with-deps
-      working-directory: ComfyUI_frontend
+        working-directory: ComfyUI_frontend
+        cache-key-suffix: ${{ hashFiles('ComfyUI_frontend/pnpm-lock.yaml') }}
```
DrJKL:
```suggestion
```

.github/workflows/test-browser-exp.yaml
```diff
@@ -11,16 +11,11 @@ jobs:
...
       with:
-        path: ~/.cache/ms-playwright
-        key: playwright-browsers-${{ runner.os }}-${{ hashFiles('ComfyUI_frontend/pnpm-lock.yaml') }}
-        restore-keys: |
-          playwright-browsers-${{ runner.os }}-
-    - name: Install Playwright Browsers
-      run: pnpm exec playwright install chromium --with-deps
-      working-directory: ComfyUI_frontend
+        working-directory: ComfyUI_frontend
+        cache-key-suffix: ${{ hashFiles('ComfyUI_frontend/pnpm-lock.yaml') }}
```
DrJKL:
```suggestion
```

.github/workflows/test-ui.yaml
```diff
@@ -123,22 +123,11 @@ jobs:
...
-      - name: Install Playwright Browsers
-        if: steps.cache-playwright-browsers.outputs.cache-hit != 'true'
-        run: pnpm exec playwright install chromium --with-deps
-        working-directory: ComfyUI_frontend
-
-      - name: Install Playwright Browsers (operating system dependencies)
-        if: steps.cache-playwright-browsers.outputs.cache-hit == 'true'
-        run: pnpm exec playwright install-deps
-        working-directory: ComfyUI_frontend
+          playwright-version: ${{ needs.setup.outputs.playwright-version }}
```
DrJKL: I think I'd rather get rid of this option entirely and let it be handled within the action, unless we really need to specify a version separate from the one in `package.json`

DrJKL: It'd also let us get rid of the step in the setups

.github/actions/setup-playwright/action.yml
```diff
@@ -0,0 +1,50 @@
+name: Setup Playwright
+description: Cache and install Playwright browsers with dependencies
+inputs:
+  playwright-version:
+    description: 'Playwright version for cache key (optional - will be detected if not provided)'
+    required: false
+  working-directory:
+    description: 'Working directory for Playwright commands'
+    required: false
+    default: 'ComfyUI_frontend'
```
DrJKL: Is it ever not set to this? If not, I'd remove the option.

.github/actions/setup-playwright/action.yml
```diff
@@ -0,0 +1,50 @@
...
+  cache-key-suffix:
+    description: 'Additional suffix for cache key (e.g., hash of lockfile)'
+    required: false
```
DrJKL: Why would we want this?

.github/actions/setup-playwright/action.yml
```diff
@@ -0,0 +1,50 @@
...
+outputs:
+  playwright-version:
+    description: 'The detected Playwright version'
+    value: ${{ steps.detect-version.outputs.playwright-version }}
```
DrJKL: Are any of the callers going to consume this right now?

DrJKL (changes requested):
I love it, but I think it could be simplified a bit further.

christian-byrne:
There are some conflicts with main

snomiao:
> There are some conflicts with main

solved! @christian-byrne @DrJKL

.github/workflows/tests-ci.yaml
```diff
@@ -123,22 +116,8 @@ jobs:
...
-        if: steps.cache-playwright-browsers.outputs.cache-hit != 'true'
-        run: pnpm exec playwright install chromium --with-deps
-        working-directory: ComfyUI_frontend
-
-      - name: Install Playwright Browsers (operating system dependencies)
-        if: steps.cache-playwright-browsers.outputs.cache-hit == 'true'
-        run: pnpm exec playwright install-deps
-        working-directory: ComfyUI_frontend
+      - name: Setup Playwright
+        uses: ./ComfyUI_frontend/.github/actions/setup-playwright
```
DrJKL, line 120: This feels weird
Don't need to change it, but it seems odd to reference the composite action in the checked out code instead of relative to this action

DrJKL (approved):

.github/workflows/tests-ci.yaml
snomiao: Agree! maybe we can simplify those things to setup-frontend-and-playwright.yaml in near future so the working directories would look nicer

PR #5765, opened 2025-09-25
[refactor] refactor load3d

jtydhr88:
Summary

 Fully Refactored the Load3D module to improve architecture and maintainability by consolidating functionality into a
  centralized composable pattern and simplifying component structure. and support VueNodes system

  Changes

  - Architecture: Introduced new useLoad3d composable to centralize 3D loading logic and state
  management
  - Component Simplification: Removed redundant components (Load3DAnimation.vue, Load3DAnimationScene.vue,
  PreviewManager.ts) 
- Support VueNodes
- improve config store
- remove lineart output due Animation doesnot support it, may add it back later
- remove Preview screen and keep scene in fixed ratio in load3d (not affect preview3d)
- improve record video feature which will already record video by same ratio as scene
Need BE change https://github.com/comfyanonymous/ComfyUI/pull/10025

https://github.com/user-attachments/assets/9e038729-84a0-45ad-b0f2-11c57d7e0c9a

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5765-refactor-refactor-load3d-2796d73d365081728297cc486e2e9052) by [Unito](https://www.unito.io)


src/components/graph/GraphCanvas.vue
```diff
@@ -175,6 +175,17 @@ const handleVueNodeLifecycleReset = async () => {
   }
 }
 
+const handleTransformPaneWheel = (event: WheelEvent) => {
+  // Check if the event target is inside a Load3D component
```
christian-byrne: Is there possibly a better solution with more specificty?

christian-byrne:
Could you fix the merge conflict then rebase with `main`?

tests-ui/tests/composables/useLoad3d.test.ts
```diff
@@ -0,0 +1,874 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { nextTick, ref } from 'vue'
+
+import { nodeToLoad3dMap, useLoad3d } from '@/composables/useLoad3d'
+import Load3d from '@/extensions/core/load3d/Load3d'
+import Load3dUtils from '@/extensions/core/load3d/Load3dUtils'
+import { useToastStore } from '@/platform/updates/common/toastStore'
+import { api } from '@/scripts/api'
+
+vi.mock('@/extensions/core/load3d/Load3d', () => ({
```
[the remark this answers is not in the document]
jtydhr88: I would like not fix this for now, as it requires additional work testing threejs in playwright.

src/services/load3dService.ts
```diff
@@ -25,84 +20,14 @@ export class Load3dService {
     return Load3dService.instance
   }
 
-  registerLoad3d(
-    node: LGraphNode,
-    container: HTMLElement,
-    inputSpec: CustomInputSpec
-  ) {
+  getLoad3d(node: LGraphNode): Load3d | null {
```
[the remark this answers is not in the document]
jtydhr88: I would like not fix this for now, as it requires redesign this class, may do it later

tests-ui/tests/composables/useLoad3d.test.ts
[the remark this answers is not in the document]
webfiltered: @jtydhr88 imo, disregard this review comment.  This is in the unit tests.

src/composables/useLoad3dViewer.ts
```diff
@@ -171,6 +172,12 @@ export const useLoad3dViewer = (node: LGraphNode) => {
       const cameraConfig = node.properties['Camera Config'] as any
       const lightConfig = node.properties['Light Config'] as any
 
+      const widthWidget = node.widgets?.find((w) => w.name === 'width')
+      const heightWidget = node.widgets?.find((w) => w.name === 'height')
+      if (!(widthWidget && heightWidget)) {
+        isPreview.value = true
+      }
```
christian-byrne: question: what is the logic for this? If there is not a width and a height widget, then it means it's a preview node?

jtydhr88: Yes, and it will be used in, for example, if open 3d viewer from preview3D, drag and drop feature will be disabled.

christian-byrne: What is the heuristic? Why is it the case that, if there is no `width` and no `height`, it means it's a preview widget? It seems a bit hard to understand.

src/components/load3d/controls/CameraControls.vue
```diff
@@ -38,66 +31,31 @@
 import { Tooltip } from 'primevue'
 import Button from 'primevue/button'
 import Slider from 'primevue/slider'
-import { computed, onMounted, onUnmounted, ref, watch } from 'vue'
+import { computed, onMounted, onUnmounted, ref } from 'vue'
 
 import type { CameraType } from '@/extensions/core/load3d/interfaces'
 import { t } from '@/i18n'
 
 const vTooltip = Tooltip
```
christian-byrne: What is this for?

src/components/load3d/controls/LightControls.vue
```diff
@@ -35,24 +34,20 @@
...
 import Button from 'primevue/button'
 import Slider from 'primevue/slider'
-import { onMounted, onUnmounted, ref, watch } from 'vue'
+import { computed, onMounted, onUnmounted, ref } from 'vue'
 
+import type { MaterialMode } from '@/extensions/core/load3d/interfaces'
 import { t } from '@/i18n'
 import { useSettingStore } from '@/platform/settings/settingStore'
 
 const vTooltip = Tooltip
```
christian-byrne: Same question as before for all the `const vTooltip = Tooltip`

src/composables/useLoad3dViewer.ts
jtydhr88: Width and height refer to the output size in load3d. They only exist in load3d, not in preview3d. However, since load3d and preview3d now share this part of the logic, and certain features (such as drag and drop file) only work for load3d, I need to use them to identify whether the source is load3d or preview3d to differentiate between them.

jtydhr88: image

src/components/load3d/Load3DControls.vue
```diff
@@ -20,52 +24,43 @@
...
-        :show-f-o-v-button="showFOVButton"
-        @switch-camera="switchCamera"
-        @update-f-o-v="handleUpdateFOV"
+        v-model:camera-type="cameraConfig.cameraType"
+        v-model:fov="cameraConfig.fov"
       />
 
       <LightControls
-        v-if="activeCategory === 'light'"
+        v-if="activeCategory === 'light' && lightConfig && modelConfig"
```
christian-byrne: for the `v-if` on the Controls, you can use `computed` to prevent re-render on every change of `activeCategory` and config.

src/components/load3d/Load3dViewerContent.vue
```diff
@@ -77,13 +90,24 @@ import ModelControls from '@/components/load3d/controls/viewer/ViewerModelContro
...
 import { useDialogStore } from '@/stores/dialogStore'
 
 const props = defineProps<{
   node: LGraphNode
 }>()
 
+// Drag and drop state
+const isDragging = ref(false)
+const dragMessage = ref('')
+const SUPPORTED_EXTENSIONS = ['.gltf', '.glb', '.obj', '.fbx', '.stl']
```
christian-byrne: question: we are defining this in both `Load3dViewerContent` and `Load3DScene`, are they always going to be the same? If so, can we put it in a constants file that both places import from?

src/components/load3d/Load3dViewerContent.vue
```diff
@@ -92,6 +116,45 @@ const mutationObserver = ref<MutationObserver | null>(null)
...
+
+  const modelFile = files.find(isValidModelFile)
+
+  if (modelFile) {
+    await viewer.handleModelDrop(modelFile)
+  } else {
+    useToastStore().addAlert(t('load3d.unsupportedFileType'))
+  }
+}
+
```
christian-byrne: suggestion: This seems like something that could be moved into a hook and re-used in other components.

src/components/load3d/Load3dViewerContent.vue
```diff
@@ -77,13 +90,24 @@ import ModelControls from '@/components/load3d/controls/viewer/ViewerModelContro
...
 }>()
 
+// Drag and drop state
+const isDragging = ref(false)
+const dragMessage = ref('')
+const SUPPORTED_EXTENSIONS = ['.gltf', '.glb', '.obj', '.fbx', '.stl']
+
+function isValidModelFile(file: File): boolean {
+  const fileName = file.name.toLowerCase()
+  return SUPPORTED_EXTENSIONS.some((ext) => fileName.endsWith(ext))
```
christian-byrne: nit: If this is how `SUPPORTED_EXTENSIONS` is being used, it's more perfomant (O(1)) to make it a `Set`, compared with using an `Array` (O(n)).

src/components/load3d/LoadingOverlay.vue
```diff
@@ -1,43 +1,24 @@
 <template>
   <Transition name="fade">
     <div
-      v-if="modelLoading"
-      class="absolute inset-0 bg-black/50 flex items-center justify-center z-50"
+      v-if="props.loading"
```
christian-byrne: nit (minor, non-blocking): you don't need `props` in the `<template>`.

src/components/load3d/LoadingOverlay.vue
```diff
@@ -1,43 +1,24 @@
...
-const endLoading = async () => {
-  await new Promise((resolve) => setTimeout(resolve, 100))
-  modelLoading.value = false
-}
-
-defineExpose({
-  startLoading,
-  endLoading
-})
+const props = defineProps<{
```
christian-byrne: nit: You can just call `defineProps` without assigning the result.

src/composables/useLoad3d.ts
```diff
@@ -0,0 +1,544 @@
...
+import { useToastStore } from '@/platform/updates/common/toastStore'
+import { api } from '@/scripts/api'
+import { useLoad3dService } from '@/services/load3dService'
+
+type Load3dReadyCallback = (load3d: Load3d) => void
+export const nodeToLoad3dMap = new Map<LGraphNode, Load3d>()
+const pendingCallbacks = new Map<LGraphNode, Load3dReadyCallback[]>()
+
+export const useLoad3d = (nodeOrRef: LGraphNode | Ref<LGraphNode | null>) => {
+  const nodeRef = 'value' in nodeOrRef ? nodeOrRef : ref(nodeOrRef)
```
christian-byrne: There are helpers for this:

```typescript
import { toRef } from '@vueuse/core'
import type { MaybeRef } from '@vueuse/core'

function useNode(nodeOrRef: MaybeRef<LGraphNode>) {
  const nodeRef = toRef(nodeOrRef)
```

src/composables/useLoad3d.ts
```diff
@@ -0,0 +1,544 @@
...
+import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
+import { useToastStore } from '@/platform/updates/common/toastStore'
+import { api } from '@/scripts/api'
+import { useLoad3dService } from '@/services/load3dService'
+
+type Load3dReadyCallback = (load3d: Load3d) => void
+export const nodeToLoad3dMap = new Map<LGraphNode, Load3d>()
+const pendingCallbacks = new Map<LGraphNode, Load3dReadyCallback[]>()
+
+export const useLoad3d = (nodeOrRef: LGraphNode | Ref<LGraphNode | null>) => {
```
christian-byrne: Don't we need to handle scenario in which `nodeOrRef` is of type `Ref<null>`? The composable seems like it would break currently if that were the case.

src/composables/useLoad3d.ts
```diff
@@ -0,0 +1,544 @@
...
+  const isPreview = ref(false)
+
+  const initializeLoad3d = async (containerRef: HTMLElement) => {
+    const node = toRaw(nodeRef.value)
+
+    if (!containerRef || !node) return
+
+    try {
+      load3d = new Load3d(containerRef, {
+        node: node as LGraphNode
```
christian-byrne: question: is this type assertion still necessary if we add null check above? If so, can we use generic `toRaw<LGraphNode>`?

christian-byrne: Same question for all the below type assertions `as LGraphNode` as well.

src/composables/useLoad3d.ts
```diff
@@ -0,0 +1,544 @@
...
+      }
+
+      node.onDrawBackground = function () {
+        if (load3d) {
+          load3d.renderer.domElement.hidden = this.flags.collapsed ?? false
+        }
+      }
+
+      node.onRemoved = function () {
+        useLoad3dService().removeLoad3d(node as LGraphNode)
```
christian-byrne: Don't we also need to clear entries in `nodeToLoad3dMap` and `pendingCallbacks` here?

src/composables/useLoad3d.ts
```diff
@@ -0,0 +1,544 @@
...
+      }
+
+      handleEvents('add')
+    } catch (error) {
+      console.error('Error initializing Load3d:', error)
+      useToastStore().addAlert(t('toastMessages.failedToInitializeLoad3d'))
+    }
+  }
+
+  const restoreConfigurationsFromNode = async (node: LGraphNode) => {
```
christian-byrne, line 128: nit (non-blocking): all of the logic revolving around serialization, config management, config types, and restoring from configs can likely be moved to its own module.

src/composables/useLoad3d.ts
```diff
@@ -0,0 +1,544 @@
...
+          useToastStore().addAlert(t('toastMessages.failedToLoadModel'))
+        } finally {
+          loading.value = false
+          loadingMessage.value = ''
+        }
+      }
+    } else if (cameraStateToRestore) {
+      load3d.setCameraState(cameraStateToRestore)
+    }
+  }
```
christian-byrne, line 178: nit (non-blocking, suggestion for future): We should consider whether this needs to be blocking setup. Currently, we are synchronously restoring the model config, get the url, loading the model, setting the camera state - and everything else must wait on this to happen. I believe we could refactor a bit and have this happen in the background like with `useAsyncState`

src/composables/useLoad3d.ts
```diff
@@ -0,0 +1,544 @@
...
+    }
+
+    if (!pendingCallbacks.has(node as LGraphNode)) {
+      pendingCallbacks.set(node as LGraphNode, [])
+    }
+
+    pendingCallbacks.get(node as LGraphNode)!.push(callback)
+  }
+
+  watch(
```
christian-byrne, line 222: question: it seems that we have these watchers which will dynamically handle updates to the config objects. But we also have the same logic (as far as I can tell) repeated in `restoreConfigurationsFromNode`. Is that correct?

For example, this watcher already handles updates to the CameraConfig:

```typescript
  watch(
    cameraConfig,
    (newValue) => {
      if (load3d && nodeRef.value) {
        nodeRef.value.properties['Camera Config'] = newValue
        load3d.toggleCamera(newValue.cameraType)
        load3d.setFOV(newValue.fov)
      }
    },
    { deep: true }
  )
```

But we are also doing the same thing in the `restoreConfigurationsFromNode`:

```typescript
    const savedCameraConfig = node.properties['Camera Config'] as CameraConfig
    const cameraStateToRestore = savedCameraConfig?.state

    // Apply camera type and FOV immediately
    if (savedCameraConfig) {
      cameraConfig.value = savedCameraConfig
      load3d.toggleCamera(savedCameraConfig.cameraType)
      load3d.setFOV(savedCameraConfig.fov)
    }
```

Shouldn't this line

```typescript
cameraConfig.value = savedCameraConfig
```

actually be triggering that `watch`, making the next two lines redundant?

christian-byrne (changes requested):
I have reviewed 14/42 files, taking a quick break.

src/composables/useLoad3dViewer.ts
christian-byrne: Why not just use the node type? You already have a reference to the node. This current setup seems hard to maintain - what happens if you want to change the name of the `width` widget in 6 months? Will you remember that it's going to break this logic?

jtydhr88: good catch, yes, we can. Just updated

christian-byrne (approved):

PR #5767, opened 2025-09-25
Refactor: Viewport Culling improvements

DrJKL:
## Summary

Reduce the number of DOM changes, but also restructure the logic to create a clean point of separation for when we can make DragAndScale reactive.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5767-Refactor-Viewport-Culling-improvements-2796d73d365081708f02cb8033aac9b1) by [Unito](https://www.unito.io)


christian-byrne:
The [tests results](https://975c3127.comfyui-playwright-chromium.pages.dev/) seem to indicate that the vue nodes are not initially loading in view. The playwright viewport is quite small.

DrJKL:
> The [tests results](https://975c3127.comfyui-playwright-chromium.pages.dev/) seem to indicate that the vue nodes are not initially loading in view. The playwright viewport is quite small.

Yeah, it failed when dragging in a new workflow while in Vue mode. Should be fixed now.

christian-byrne (approved):
LGTM!

src/composables/graph/useViewportCulling.ts
```diff
@@ -6,78 +6,98 @@
...
+type Bounds = [left: number, right: number, top: number, bottom: number]
+
+function getNodeBounds(node: LGraphNode): Bounds {
+  const [nodeLeft, nodeTop] = node.pos
+  const nodeRight = nodeLeft + node.size[0]
+  const nodeBottom = nodeTop + node.size[1]
+  return [nodeLeft, nodeRight, nodeTop, nodeBottom]
+}
+
+function viewportEdges(
```
arjansingh, line 26: Why manual culling and not an [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver)?

We could sett the `root` of the observer to be the canvas and we could just debounce or throttle the processing rate.

DrJKL: @simula-r or @christian-byrne would probably be more equipped to answer that 🙂

simula-r: Good question. My understanding is that IntersectionObserver can't observe elements that have display: none which we rely on to prevent off screen elements from contributing to the compositor layer size. We could try content-visibility instead which seems to work with IntersectionObserver but from my testing it wasnt as performant as display: none.

src/renderer/core/layout/__tests__/TransformPane.spec.ts
```diff
@@ -1,97 +1,96 @@
 import { mount } from '@vue/test-utils'
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
-import { nextTick, ref } from 'vue'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { computed, nextTick } from 'vue'
+
+import type { LGraphCanvas } from '@/lib/litegraph/src/LGraphCanvas'
+import { useTransformState } from '@/renderer/core/layout/transform/useTransformState'
 
 import TransformPane from '../transform/TransformPane.vue'
```
lexxteran666, line 8: Import TransformPane from '../transform/TransformPane.vue'
Helloworld

PR #5773, opened 2025-09-25
Use localized labels for Vue node slots and inputs/outputs (fallback to names)

### LINKED ISSUE #5697: Lots of widget translations are missing in the VUE nodes.
Image

Image

christian-byrne:
## Summary

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/5697 by adding internationalization support for Vue node widget labels and output slot names with fallback to original names.

## Changes

- **What**: Added `label` field to widget data structures and `localized_name` support for output slots
- **Breaking**: None - all changes maintain backward compatibility with fallback to original names

## Review Focus

Translation key resolution for node definitions and proper fallback chain for widget labels and slot names.

Observe that there is parity between Litegraph and Vue nodes w.r.t. localization of labels/names. 

Screenshot from 2025-09-25 13-19-54

Screenshot from 2025-09-25 13-19-38

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5773-Use-localized-labels-for-Vue-node-slots-and-inputs-outputs-fallback-to-names-2796d73d365081e08fb7d38464f4aa90) by [Unito](https://www.unito.io)


christian-byrne:
Test failures are expected (the text next to the widgets changed from the name to the label which changed the apperance of some UI baseline images): https://531ead11.comfyui-playwright-chromium.pages.dev/#?q=s:failed

AustinMroz (dismissed):
nit: (Annoying, but out of scope) changing the language setting while in vue mode does not trigger update of labels. Either toggling vue mode or a reload is required.

src/renderer/extensions/vueNodes/components/NodeHeader.vue
```diff
@@ -119,8 +121,13 @@ const tooltipConfig = computed(() => {
...
   if (title.length > 0) return title
-  const type = (info?.type ?? '').trim()
-  return type.length > 0 ? type : 'Untitled'
+
+  const nodeType = (info?.type ?? '').trim()
+  if (nodeType.length > 0) {
+    const key = `nodeDefs.${normalizeI18nKey(nodeType)}.display_name`
+    return st(key, nodeType)
+  }
+  return 'Untitled'
```
DrJKL: If we had an entry for untitled inputs, could we do this? 
```suggestion
  const nodeType = (info?.type ?? '').trim() || 'Untitled'
  const key = `nodeDefs.${normalizeI18nKey(nodeType)}.display_name`
  return st(key, nodeType)
```

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -135,6 +135,7 @@ const processedWidgets = computed((): ProcessedWidget[] => {
       name: widget.name,
       type: widget.type,
       value: widget.value,
+      label: widget.label,
       options: widget.options,
       callback: widget.callback
```
DrJKL, line 140: This looks like it could just use the spread operator.

DrJKL (dismissed):

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
christian-byrne: I believe it's extracting safe properties, otherwise private fields would be included.

christian-byrne:
> nit: (Annoying, but out of scope) changing the language setting while in vue mode does not trigger update of labels. Either toggling vue mode or a reload is required.

True we have that tracked here:

- https://github.com/Comfy-Org/ComfyUI_frontend/issues/5698

I think we can handle it easily in the next milestone (it's also in notion).

PR #5515, opened 2025-09-12
fix: Add babel configuration for Playwright to handle TypeScript declare fields

snomiao:
## Summary
Fixes babel configuration issues in the Playwright i18n collection script that were causing failures when importing `ComfyNodeDefImpl` after Vue nodes were introduced to the codebase.

## The Problem

The `collect-i18n-node-defs.ts` script was failing because of an import chain that brings Vue components into Node.js context:

### The Import Chain
```
collect-i18n-node-defs.ts
  ↓ imports
ComfyNodeDefImpl (from src/stores/nodeDefStore.ts)
  ↓ imports at line 19
useSubgraphStore (from src/stores/subgraphStore.ts)
  ↓ imports at line 28
ComfyNodeDefImpl (circular, but also imports other stores)
  ↓ which transitively imports
executionStore.ts
  ↓ imports at line 4
ChatHistoryWidget.vue (Vue component!)
```

**Critical Issue**: When babel tries to transform `.vue` files in Node.js context, it crashes because Vue SFC files are not JavaScript - they're templates that need webpack/rollup loaders, not babel transpilation.

### When This Started
This problem began ~3 weeks ago when Vue nodes were introduced in [commit 006e6bd57](https://github.com/Comfy-Org/ComfyUI_frontend/commit/006e6bd57) ([PR #4263](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4263)). Before Vue nodes existed, the import chain didn't include any `.vue` files, so the script worked fine.

## The Solution
This PR implements a robust babel transformation pipeline that:

### 1. Enhanced Playwright Configuration
- Configures a proper babel plugin chain with TypeScript transformation, module resolution, and automatic global injection
- Uses the magic `config['@playwright/test']` configuration to control babel transformations
- Ensures plugins execute in the correct order (module resolver → TypeScript → globals injection)

### 2. Automatic Browser Globals Injection
- Created `babel-plugin-inject-globals.cjs` that automatically injects browser environment setup for matching files
- Eliminates manual import dependencies and order issues
- Configurable via file pattern matching

### 3. Browser Context Evaluation
- Refactored `collect-i18n-node-defs.ts` to evaluate imports within browser context using Playwright
- This avoids Node.js trying to parse Vue components directly
- Vue components are properly available in the browser environment

### 4. Cleanup
- Removed unused `babel-plugin-stub-vue-imports.cjs` (no longer needed with proper babel configuration)
- Improved error handling and ESM support in setup scripts

## The Learning Journey

[...]

PR #5775, opened 2025-09-25
fix(collect-i18n-node-defs): refactor to run ComfyNodeDefImpl only in browser context

snomiao:
## The Problem

The `collect-i18n-node-defs.ts` script started failing ~3 weeks ago when Vue nodes were introduced ([commit 006e6bd57](https://github.com/Comfy-Org/ComfyUI_frontend/commit/006e6bd57), [PR #4263](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4263)). The issue stems from:

1. **Import chain bringing Vue components into Node.js context:**
   ```
   collect-i18n-node-defs.ts
     ↓ imports
   ComfyNodeDefImpl (from nodeDefStore.ts)
     ↓ imports
   useSubgraphStore (from subgraphStore.ts)
     ↓ transitively imports
   executionStore.ts
     ↓ imports
   ChatHistoryWidget.vue (Vue component!)
   ```

2. **TypeScript `declare` fields causing Babel errors:**
   ```
   TypeScript 'declare' fields must first be transformed by @babel/plugin-transform-typescript
   ```

### PR #5515 Approach (Complex)
- Adds custom Babel plugins and configurations
- Implements automatic browser globals injection
- Requires **47,517 additions, 9,469 deletions**
- Modifies the entire Playwright babel transformation pipeline

### This PR's Approach (Simple)
- Uses dynamic imports to defer module loading until runtime
- Avoids Babel compilation of problematic TypeScript/Vue files
- **Only 40 lines changed** in a single file
- No configuration changes needed

## How This Fix Works

```typescript
// Instead of static import that Babel tries to compile:
// import { ComfyNodeDefImpl } from '../src/stores/nodeDefStore'

// We use:
// 1. Type-only import (erased at runtime)
import type { ComfyNodeDefImpl } from '../src/stores/nodeDefStore'

// 2. Dynamic import at runtime (bypasses Babel)
const { ComfyNodeDefImpl: ComfyNodeDefImplClass } = await import(
  '../src/stores/nodeDefStore'
)
```


.github/workflows/i18n.yaml
```diff
@@ -11,7 +11,7 @@ on:
 jobs:
   update-locales:
     # Branch detection: Only run for manual dispatch or version-bump-* branches from main repo
-    if: github.event_name == 'workflow_dispatch' || (github.event.pull_request.head.repo.full_name == github.repository && startsWith(github.head_ref, 'version-bump-'))
+    if: github.event_name == 'workflow_dispatch' || (github.event.pull_request.head.repo.full_name == github.repository && startsWith(github.head_ref, 'version-bump-')) || startsWith(github.head_ref, 'sno-fix-locale')
```
Myestery: now that we've verified it works, can we reset this part?

Myestery (commented):
Thanks a lot. I've anticipated this fix for some time now

scripts/collect-i18n-node-defs.ts
```diff
@@ -26,18 +24,23 @@ test('collect-i18n-node-defs', async ({ comfyPage }) => {
...
+      // @ts-expect-error - app is dynamically added to window
+      const api = window['app'].api
+      const rawNodeDefs = await api.getNodeDefs()
+      const { ComfyNodeDefImpl } = await import('../src/stores/nodeDefStore')
+
+      return (
+        Object.values(rawNodeDefs)
+          // Ignore DevTools nodes (used for internal testing)
+          .filter((def: any) => !def.name.startsWith('DevTools'))
+          .map((def: any) => new ComfyNodeDefImpl(def))
```
DrJKL: Can you add the types?

.github/workflows/i18n.yaml
snomiao: done!

Myestery (dismissed):

DrJKL (changes requested):

scripts/collect-i18n-node-defs.ts
snomiao: done!


        Object.values(rawNodeDefs)
          // Ignore DevTools nodes (used for internal testing)
          .filter((def: ComfyNodeDef) => !def.name.startsWith('DevTools'))
          .map((def: ComfyNodeDef) => new ComfyNodeDefImpl(def))

christian-byrne (approved):
LGTM!

christian-byrne:
I will cherry pick

PR #5777, opened 2025-09-25
Split Tailwind utility functions out to a shared package

webfiltered:
## Summary

Split tailwind utils out to a shared package within the monorepo.

## Changes

- Creates `@comfyorg/tailwind-utils` package
- Does not require export, publishing, etc
- Uses `export` to ensure this change does not impact other PRs (many imports to update)
  - If we _want_ to update all imports, there are two commits ready to be re-applied
  - e.g. `git revert 80960c2a82c0d1ac06eee1bb83ac333216b2b376`

## Review Focus

- Is this pattern desirable?
- Should we just include this in a broader design-system split?  I kind of vote yes, but also it's a good small, first step.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5777-Split-Tailwind-utility-functions-out-to-a-shared-package-2796d73d3650815f976fc73b4fb86ef3) by [Unito](https://www.unito.io)


packages/tailwind-utils/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.json",
+  "compilerOptions": {
+    "rootDir": "src",
+    "outDir": "dist"
+  },
+  "include": ["src/**/*"],
+  "exclude": ["node_modules", "dist"]
```
DrJKL: You shouldn't need to exclude these unless you have a broad scope in include.

DrJKL (dismissed):
I support this, yes 👍🏻 

I would want to do the imports migration and removing `src/utils/tailwindUtil.ts` at some point in the future.

packages/tailwind-utils/tsconfig.json
webfiltered: It needed more stuff.  It's just too skinny for a tsconfig.

... fixed in bec0d2726ae6a65c92867510268de8fe5fab8bdf

DrJKL: The other option would be to keep exclude and let include default to `**/*`, but that doesn't make it any thiccer

webfiltered: frog-bonk

DrJKL (approved):

christian-byrne (approved):

christian-byrne:
Cool

PR #5778, opened 2025-09-25
Vue custom widgets

Myestery:
This pull request introduces a new abstraction for audio widget creation, implements a Vue-specific audio widget factory, and adds a store for managing widget serialization. These changes lay the foundation for a more modular and consistent approach to handling audio widgets in both Vue and LiteGraph environments, and enable proper workflow saving/loading for audio-related widgets.

**Audio Widget Factory Abstraction**

* Added the `AudioWidgetFactory` interface in `AudioWidgetFactory.ts`, defining a consistent API for creating audio widgets and lifecycle hooks, making it easier to support multiple UI frameworks.

**Vue Implementation**

* Implemented `VueAudioWidgetFactory` in `VueAudioWidgetFactory.ts`, providing Vue-compatible audio widget creation, placeholder widgets, and serialization support for audio playback, upload, and recording widgets.
* Added hooks in `VueAudioWidgetFactory` for node definition registration, output updates, and node creation, including integration with an audio service and dynamic widget value updates.

**Widget Serialization Support**

* Introduced the `vueWidgetSerializationStore` in `vueWidgetSerializationStore.ts`, allowing Vue components to register serialization handlers for widgets, enabling proper workflow saving/loading and bridging communication between Vue and LiteGraph widgets.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5778-Vue-custom-widgets-2796d73d36508126b78edb127b2c1126) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/factories/AudioWidgetFactory.ts
+++ b/src/extensions/core/factories/AudioWidgetFactory.ts
@@ -1,0 +1,44 @@
+/**
+ * Factory interface for creating audio widgets
+ * Provides consistent API for both Vue and LiteGraph implementations
+ */
+import type { LGraphNode } from '@/lib/litegraph/src/litegraph'
+import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets'
+
+export interface WidgetCreationResult {
+  widget: IBaseWidget
+  minWidth?: number
+  minHeight?: number
+}
+
+export interface AudioWidgetFactory {
+  /**
+   * Creates an AUDIO_UI widget for audio playback/preview
+   */
+  createAudioUI(node: LGraphNode, inputName: string): WidgetCreationResult
+
+  /**
+   * Creates an AUDIOUPLOAD widget for file uploads
+   */
+  createAudioUpload(node: LGraphNode, inputName: string): WidgetCreationResult
+
+  /**
+   * Creates an AUDIO_RECORD widget for audio recording
+   */
+  createAudioRecord(node: LGraphNode, inputName: string): WidgetCreationResult
+
+  /**
+   * Hook called before registering node definitions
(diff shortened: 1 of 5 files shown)


christian-byrne (commented):
If we added these requirements, would the implementation change?

- Do not need to serialize runtime-generated widgets (AUDIO_UI and upload button)
- When vue nodes are disabled, we can just skip the logic in `uploadAudio.ts` entirely
- We can centralize all logic in the new AudioWidget vue components, so most of the current hooks would not be needed

In general for Vue/Litegraph dual system, I like the approach of having factories that both implement the same interface, but just asking these questions.

Myestery:
> If we added these requirements, would the implementation change?
> 
> * Do not need to serialize runtime-generated widgets (AUDIO_UI and upload button)
> * When vue nodes are disabled, we can just skip the logic in `uploadAudio.ts` entirely
> * We can centralize all logic in the new AudioWidget vue components, so most of the current hooks would not be needed
> 
> In general for Vue/Litegraph dual system, I like the approach of having factories that both implement the same interface, but just asking these questions.

Issue with point 2
Backward compatibility: If we have a workflow with Audio UI, we need to find a way to display it.
Also we need to be able to use the inputs and outputs of that old workflow so this means we have to support the concept of audioUI

christian-byrne:
But Audio UI is just a UI concept. The actual value is just the filename which is in the singular widget we are keeping.

Myestery:
> But Audio UI is just a UI concept. The actual value is just the filename which is in the singular widget we are keeping.

This makes sense since all we need essentially is the serializa function and the rest is UI
We can identify that from the node

PR #5782, opened 2025-09-26
Implement fit-to-view for Vue nodes

christian-byrne:
## Summary

Implemented fit-to-view functionality for Vue nodes with bounds calculation and viewport animation support.

https://github.com/user-attachments/assets/2ec221f1-9194-4564-95f9-ad4da80f190a

## Changes

- **What**: Added Vue nodes support to fit-to-view command with [bounds calculation](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) and LiteGraph integration
- **Dependencies**: Added dependency on `layoutStore` and `selectionBounds` utility

## Review Focus

Bounds calculation accuracy for complex node layouts and animation performance with large node selections. Verify proper fallback to legacy LiteGraph behavior when Vue nodes disabled.

```mermaid
graph TD
    A[Fit to View Command] --> B{Vue Nodes Enabled?}
    B -->|Yes| C[Get Selected Nodes]
    B -->|No| D[Legacy LiteGraph Method]
    C --> E{Nodes Selected?}
    E -->|Yes| F[Calculate Selected Bounds]
    E -->|No| G[Calculate All Nodes Bounds]
    F --> H[Convert to LiteGraph Format]
    G --> H
    H --> I[Animate to Bounds]
    D --> J[Canvas fitViewToSelectionAnimated]
    
    style A fill:#f9f9f9,stroke:#333,color:#333
    style I fill:#f9f9f9,stroke:#333,color:#333
    style J fill:#f9f9f9,stroke:#333,color:#333
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5782-Implement-fit-to-view-for-Vue-nodes-27a6d73d365081cb822cd93f557e77b2) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useCoreCommands.ts
+++ b/src/composables/useCoreCommands.ts
@@ -25,6 +25,8 @@ import {
   useCanvasStore,
   useTitleEditorStore
 } from '@/renderer/core/canvas/canvasStore'
+import { layoutStore } from '@/renderer/core/layout/store/layoutStore'
+import { selectionBounds } from '@/renderer/core/layout/utils/layoutMath'
 import { api } from '@/scripts/api'
 import { app } from '@/scripts/app'
 import { useDialogService } from '@/services/dialogService'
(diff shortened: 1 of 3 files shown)


src/renderer/core/layout/utils/layoutMath.ts
```diff
@@ -19,3 +19,35 @@ export function boundsIntersect(a: Bounds, b: Bounds): boolean {
...
+
+/**
+ * Calculate combined bounds for Vue nodes selection
+ * @param nodes Array of NodeLayout objects to calculate bounds for
+ * @returns Bounds of the nodes or null if no nodes provided
+ */
+export function selectionBounds(nodes: NodeLayout[]): Bounds | null {
+  if (nodes.length === 0) return null
+  return calculateBounds(nodes)
+}
```
christian-byrne, line 53: The reason I made it a separate function is because the semantics of what the bounds of `[]` are can be defined explicitly for the function's purpose, where it otherwise may be unclear.

webfiltered: This feels both logical and proper!  One function is a raw calculator - the second uses the calculator for a specific task.

src/composables/useCoreCommands.ts
```diff
@@ -314,15 +316,55 @@ export function useCoreCommands(): ComfyCommand[] {
...
+          const allNodes = layoutStore.getAllNodes().value
+
+          // Get nodes to fit - selected if any, otherwise all
+          const nodesToFit =
+            selectedNodeIds.size > 0
+              ? Array.from(selectedNodeIds)
+                  .map((id) => allNodes.get(id))
+                  .filter(
+                    (node): node is NonNullable<typeof node> => node != null
+                  )
```
webfiltered:
```suggestion
              ? Array.from(selectedNodeIds)
                  .map((id) => allNodes.get(id))
                  .filter((node) => node != null)
```

Not a hot path, so triple-loop is not a problem here.  Type here is already inferred.

webfiltered (dismissed):
nit

src/composables/useCoreCommands.ts
christian-byrne: Thank you!

webfiltered (approved):

PR #5787, opened 2025-09-26
fix large integer precision handling in Vue int widgets

christian-byrne:
## Summary

Fixed increment/decrement button lockup in number widgets when values exceed JavaScript's safe integer limit (2^53 - 1).

## Changes

- **What**: Added precision-aware button disabling and user feedback to `WidgetInputNumberInput` component using [Number.isSafeInteger()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger)
- We still need to support values greater than 2^53 because they may be in workflows.

## Review Focus

JavaScript floating-point precision behavior at scale - buttons hide when arithmetic operations like `value + 1` would be unreliable due to IEEE 754 limitations. Test coverage includes edge cases (NaN, Infinity) and boundary conditions at MAX_SAFE_INTEGER.

```mermaid
graph TD
    A[User Input] --> B{Value > 2^53?}
    B -->|No| C[Show Buttons]
    B -->|Yes| D[Hide Buttons]
    D --> E[Show Tooltip]
    E --> F[User Can Still Type]
    
    style F fill:#f9f9f9,stroke:#333,color:#000
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5787-fix-large-integer-precision-handling-in-Vue-int-widgets-27a6d73d365081d9ae00e485740cfafb) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
@@ -2,6 +2,7 @@
 import InputNumber from 'primevue/inputnumber'
 import { computed } from 'vue'
 
+import { useNumberWidgetValue } from '@/composables/graph/useWidgetValue'
 import type { SimplifiedWidget } from '@/types/simplifiedWidget'
 import { cn } from '@/utils/tailwindUtil'
 import {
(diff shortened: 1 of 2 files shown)


src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
```diff
@@ -14,10 +15,20 @@ import WidgetLayoutField from './layout/WidgetLayoutField.vue'
...
+  modelValue: number
   readonly?: boolean
 }>()
 
-const modelValue = defineModel<number>({ default: 0 })
+const emit = defineEmits<{
+  'update:modelValue': [value: number]
+}>()
+
+// Use the composable for consistent widget value handling
```
christian-byrne: Remove the LLM comment

arjansingh (dismissed):

DrJKL (approved):

PR #5789, opened 2025-09-26
Cleanup: Litegraph/Vue synchronization work

DrJKL:
## Summary

Cleanup and fixes to the existing syncing logic.

## Review Focus

This is probably enough to review and test now.

Main things that should still work: 
- moving nodes around
- adding new ones
- switching back and forth between Vue and Litegraph

Let me know if you find any bugs that weren't already present there.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5789-WIP-Litegraph-Vue-synchronization-work-27a6d73d3650811682cacacb82367b9e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useVueNodeLifecycle.ts
+++ b/src/composables/graph/useVueNodeLifecycle.ts
@@ -26,7 +26,7 @@ function useVueNodeLifecycleIndividual() {
 
   const initializeNodeManager = () => {
     // Use canvas graph if available (handles subgraph contexts), fallback to app graph
-    const activeGraph = comfyApp.canvas?.graph || comfyApp.graph
+    const activeGraph = comfyApp.graph
     if (!activeGraph || nodeManager.value) return
 
     // Initialize the core node manager
@@ -83,9 +83,7 @@ function useVueNodeLifecycleIndividual() {
 
   // Watch for Vue nodes enabled state changes
   watch(
-    () =>
-      shouldRenderVueNodes.value &&
-      Boolean(comfyApp.canvas?.graph || comfyApp.graph),
+    () => shouldRenderVueNodes.value && Boolean(comfyApp.graph),
     (enabled) => {
       if (enabled) {
         initializeNodeManager()
--- a/src/composables/graph/useVueNodeLifecycle.ts
+++ b/src/composables/graph/useVueNodeLifecycle.ts
@@ -26,7 +26,7 @@ function useVueNodeLifecycleIndividual() {
 
   const initializeNodeManager = () => {
     // Use canvas graph if available (handles subgraph contexts), fallback to app graph
-    const activeGraph = comfyApp.graph
+    const activeGraph = comfyApp.canvas?.graph
     if (!activeGraph || nodeManager.value) return
 
     // Initialize the core node manager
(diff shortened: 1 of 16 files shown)


christian-byrne (commented):
LGTM

arjansingh (dismissed):

src/scripts/app.ts
```diff
@@ -157,10 +157,8 @@ export class ComfyApp {
   // @ts-expect-error fixme ts strict error
   _nodeOutputs: Record<string, any>
   nodePreviewImages: Record<string, string[]>
-  // @ts-expect-error fixme ts strict error
-  #graph: LGraph
   get graph() {
-    return this.#graph
+    return this.canvas?.graph! // TODO: Fix callers who aren't handling undefined graph
```
benceruleanlu: can you confirm that this change from global graph to canvas spaced graph won't cause issues, possibly one of them being that if a user saves a workflow inside of a subgraph, the workflow will only become the subgraph, losing the rest of the workflow?

benceruleanlu (dismissed):

src/scripts/app.ts
benceruleanlu: I could be wrong, but it seems like there are still a few places in the codebase that expect a global graph

I looked at:

```
  const persistCurrentWorkflow = () => {
    const workflow = JSON.stringify(comfyApp.graph.serialize())
    localStorage.setItem('workflow', workflow)
    if (api.clientId) {
      sessionStorage.setItem(`workflow:${api.clientId}`, workflow)
    }
  }
```

where `comfyApp.graph` is supposed to be the global graph, but now is whatever canvas spaced graph is open, which can be the subgraph

https://github.com/user-attachments/assets/6db87b10-24bb-43c3-be19-d61e95e37bbc

src/composables/graph/useVueNodeLifecycle.ts
```diff
@@ -185,24 +147,17 @@ function useVueNodeLifecycleIndividual() {
 
   // Cleanup function for component unmounting
   const cleanup = () => {
+    console.log('cleanup')
```
benceruleanlu: nit: our console isn't very clean right now anyway, but could help

src/scripts/app.ts
DrJKL: Sad, but great catch

src/composables/graph/useVueNodeLifecycle.ts
DrJKL: I'm on my phone, but if you suggest a deletion I'll commit it 😊

DrJKL: Cleaned :-)

src/scripts/app.ts
DrJKL: Okay, restored the places that were grabbing the canvas graph. I'd like to do a rename to something like `rootGraph` and deprecate the old one to give extensions time to clean up.
What do you think?

benceruleanlu: sg

arjansingh (dismissed):

benceruleanlu (dismissed):

PR #5791, opened 2025-09-26
fix Vue nodes capturing wheel event when Ctrl+Shift+drag zooming

christian-byrne:
## Summary

Fixed canvas read-only state during Ctrl+Shift drag-to-zoom to prevent node interaction conflicts. This fix also benefits Litegraph nodes.

## Changes

- **What**: Canvas becomes read-only during [Ctrl+Shift drag zoom](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent) operations and restores original state when finished
- **What**: Added browser test coverage for zoom interaction behavior

## Review Focus

State restoration on zoom interruption and edge cases where drag zoom is cancelled mid-operation.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5791-fix-Vue-nodes-capturing-wheel-event-when-Ctrl-Shift-drag-zooming-27a6d73d3650817d888ee59f8f7fcb01) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -688,8 +688,8 @@ export class LGraphCanvas
 
   /** If true, enable drag zoom. Ctrl+Shift+Drag Up/Down: zoom canvas. */
   dragZoomEnabled: boolean = false
-  /** The start position of the drag zoom. */
-  #dragZoomStart: { pos: Point; scale: number } | null = null
+  /** The start position of the drag zoom and original read-only state. */
+  #dragZoomStart: { pos: Point; scale: number; readOnly: boolean } | null = null
```
DrJKL, line 692: Want to make this typescript private while you're in here?
```suggestion
  private dragZoomStart: { pos: Point; scale: number; readOnly: boolean } | null = null
```

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -3524,6 +3536,10 @@ export class LGraphCanvas
     const { graph, pointer } = this
     if (!graph) return
 
+    if (this.#dragZoomStart) {
```
DrJKL: You have a guard in the method, so you shouldn't need to guard here too.

DrJKL (dismissed):

src/lib/litegraph/src/LGraphCanvas.ts
christian-byrne: Sure!

src/lib/litegraph/src/LGraphCanvas.ts
christian-byrne: Thanks, https://github.com/Comfy-Org/ComfyUI_frontend/pull/5791/commits/667cf0cc35003476d1e187fa5d24b287373d18be

PR #5793, opened 2025-09-26
[fix] Stop widget scroll events from being intercepted by LiteGraph

arjansingh:
## Summary

1. LiteGraph has a handler that intercepts all canvas `wheel` events and preventsDefault on them. See [code](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/lib/litegraph/src/LGraphCanvas.ts#L3617-L3664).
2. To get in front of that I needed to create a composable that sits at the App level to preserve `wheel` events created by our scrollable widgets.

## Changes

1. Create `src/renderer/extensions/vueNodes/composables/usePreserveWidgetScroll.ts`
2. Activate it in App
3. Tests

## Screenshots

Fixed behavior

https://github.com/user-attachments/assets/549b4caa-c3cd-4693-a2e1-d8da5ed32250

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5793-fix-Stop-widget-scroll-events-from-being-intercepted-by-LiteGraph-27a6d73d36508179987bd3fb8e8cd6fb) by [Unito](https://www.unito.io)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -16,6 +16,7 @@ import { computed, onMounted } from 'vue'
 
 import GlobalDialog from '@/components/dialog/GlobalDialog.vue'
 import config from '@/config'
+import { usePreserveWidgetScroll } from '@/renderer/extensions/vueNodes/composables/usePreserveWidgetScroll'
 import { useWorkspaceStore } from '@/stores/workspaceStore'
 import { useConflictDetection } from '@/workbench/extensions/manager/composables/useConflictDetection'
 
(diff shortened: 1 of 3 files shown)


webfiltered (commented):
If I'm reading this right, it looks like a textarea with no scrollbar will still intercept wheel events.

This is how our e.g. CLIP encode nodes used to work; it was quite frustrating to deal with when trying to move around workflows - all of a sudden your scroll zoom stops for no reason.

Just a thought; sometimes it's easier to just add a quick return on the litegraph handler - "if should not handle, return and let frontend do stuff".  (edit: to be clear, "just a thought" was quite literal - I haven't looked to see if that's even remotely viable here!)

PR #5798, opened 2025-09-26
Use input spec to determine asset type in upload widgets

christian-byrne:
Refactored widget rendering to use input specification metadata for determining asset kinds instead of file extension heuristics.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5798-Use-input-spec-to-determine-asset-type-in-upload-widgets-27a6d73d365081f39cfaf45bbc377e49) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -207,10 +207,7 @@ export const useGraphNodeManager = (graph: LGraph): GraphNodeManager => {
         return {
           name: widget.name || 'unknown',
           type: widget.type || 'text',
-          value: undefined, // Already a valid WidgetValue
-          options: undefined,
-          callback: undefined,
-          spec: undefined
+          value: undefined
         }
       }
     })
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -8,7 +8,7 @@ import { useChainCallback } from '@/composables/functional/useChainCallback'
 import { useLayoutMutations } from '@/renderer/core/layout/operations/layoutMutations'
 import { LayoutSource } from '@/renderer/core/layout/types'
 import { type Bounds, QuadTree } from '@/renderer/core/spatial/QuadTree'
-import type { InputSpec as InputSpecV2 } from '@/schemas/nodeDef/nodeDefSchemaV2'
+import type { InputSpec } from '@/schemas/nodeDef/nodeDefSchemaV2'
 import { useNodeDefStore } from '@/stores/nodeDefStore'
 import type { WidgetValue } from '@/types/simplifiedWidget'
 import type { SpatialIndexDebugInfo } from '@/types/spatialIndex'
@@ -46,7 +46,7 @@ export interface SafeWidgetData {
   value: WidgetValue
   options?: Record<string, unknown>
   callback?: ((value: unknown) => void) | undefined
-  spec?: InputSpecV2
+  spec?: InputSpec
 }
 
 export interface VueNodeData {
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -8,6 +8,8 @@ import { useChainCallback } from '@/composables/functional/useChainCallback'
 import { useLayoutMutations } from '@/renderer/core/layout/operations/layoutMutations'
 import { LayoutSource } from '@/renderer/core/layout/types'
 import { type Bounds, QuadTree } from '@/renderer/core/spatial/QuadTree'
+import type { InputSpec as InputSpecV2 } from '@/schemas/nodeDef/nodeDefSchemaV2'
+import { useNodeDefStore } from '@/stores/nodeDefStore'
 import type { WidgetValue } from '@/types/simplifiedWidget'
 import type { SpatialIndexDebugInfo } from '@/types/spatialIndex'
 
@@ -44,6 +46,7 @@ export interface SafeWidgetData {
   value: WidgetValue
   options?: Record<string, unknown>
   callback?: ((value: unknown) => void) | undefined
+  spec?: InputSpecV2
 }
 
 export interface VueNodeData {
@@ -105,6 +108,7 @@ export const useGraphNodeManager = (graph: LGraph): GraphNodeManager => {
   // Get layout mutations composable
   const { moveNode, resizeNode, createNode, deleteNode, setSource } =
     useLayoutMutations()
+  const nodeDefStore = useNodeDefStore()
   // Safe reactive data extracted from LiteGraph nodes
   const vueNodeData = reactive(new Map<string, VueNodeData>())
   const nodeState = reactive(new Map<string, NodeState>())
@@ -189,13 +193,15 @@ export const useGraphNodeManager = (graph: LGraph): GraphNodeManager => {
         ) {
           value = widget.options.values[0]
         }
+        const spec = nodeDefStore.getInputSpecForWidget(node, widget.name)
 
         return {
           name: widget.name,
           type: widget.type,
           value: value,
           options: widget.options ? { ...widget.options } : undefined,
-          callback: widget.callback
+          callback: widget.callback,
+          spec
         }
       } catch (error) {
         return {
@@ -203,7 +209,8 @@ export const useGraphNodeManager = (graph: LGraph): GraphNodeManager => {
           type: widget.type || 'text',
           value: undefined, // Already a valid WidgetValue
           options: undefined,
-          callback: undefined
+          callback: undefined,
+          spec: undefined
         }
       }
     })
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
@@ -2,7 +2,7 @@
   <WidgetSelectDropdown
     v-if="isDropdownUIWidget"
     v-bind="props"
-    :media-kind="mediaKind"
+    :asset-kind="assetKind"
     :allow-upload="allowUpload"
     :upload-folder="uploadFolder"
     @update:model-value="handleUpdateModelValue"
@@ -23,7 +23,7 @@ import {
   isComboInputSpec
 } from '@/schemas/nodeDef/nodeDefSchemaV2'
 import type { SimplifiedWidget } from '@/types/simplifiedWidget'
-import type { MediaKind } from '@/types/widgetTypes'
+import type { AssetKind } from '@/types/widgetTypes'
 
 import WidgetSelectDefault from './WidgetSelectDefault.vue'
 import WidgetSelectDropdown from './WidgetSelectDropdown.vue'
@@ -51,7 +51,7 @@ const specDescriptor = computed(() => {
   const spec = comboSpec.value
   if (!spec) {
     return {
-      kind: 'unknown' as MediaKind,
+      kind: 'unknown' as AssetKind,
       allowUpload: false,
       folder: undefined as ResultItemType | undefined
     }
@@ -68,7 +68,7 @@ const specDescriptor = computed(() => {
     (spec as Partial<Record<'audio_upload', boolean>>).audio_upload
   )
 
-  let kind: MediaKind = 'unknown'
+  let kind: AssetKind = 'unknown'
   if (video_upload) {
     kind = 'video'
   } else if (image_upload || animated_image_upload) {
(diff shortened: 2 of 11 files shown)


src/composables/graph/useGraphNodeManager.ts
```diff
@@ -189,21 +193,24 @@ export const useGraphNodeManager = (graph: LGraph): GraphNodeManager => {
...
+        const spec = nodeDefStore.getInputSpecForWidget(node, widget.name)
 
         return {
           name: widget.name,
           type: widget.type,
           value: value,
           options: widget.options ? { ...widget.options } : undefined,
-          callback: widget.callback
+          callback: widget.callback,
+          spec
```
DrJKL, line 204: optional: This could probably be made a little cleaner with either 
- a splat and override
- destructuring widget and using shorthand for the cloned properties

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -189,21 +193,24 @@ export const useGraphNodeManager = (graph: LGraph): GraphNodeManager => {
...
         }
       } catch (error) {
         return {
           name: widget.name || 'unknown',
           type: widget.type || 'text',
           value: undefined, // Already a valid WidgetValue
           options: undefined,
-          callback: undefined
+          callback: undefined,
+          spec: undefined
```
DrJKL: question: Not sure what the intended shape is here, but does it need explicit `undefined`s?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
```diff
@@ -34,61 +42,63 @@ function handleUpdateModelValue(value: string | number | undefined) {
...
-
-const isVideoWidget = computed(() =>
-  hasFilesWithExtensions(
-    props.widget.options?.values || [],
-    /\.(mp4|webm|ogg|mov)$/i
-  )
-)
+const comboSpec = computed<ComboInputSpec | undefined>(() => {
+  if (!props.widget.spec) return undefined
+  return isComboInputSpec(props.widget.spec) ? props.widget.spec : undefined
```
DrJKL: optional:
```suggestion
  if (props.widget.spec && isComboInputSpec(props.widget.spec)) {
    return props.widget.spec
  }
  return
```

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
```diff
@@ -34,61 +42,63 @@ function handleUpdateModelValue(value: string | number | undefined) {
...
-  )
-)
+const specDescriptor = computed(() => {
+  const spec = comboSpec.value
+  if (!spec) {
+    return {
+      kind: 'unknown' as AssetKind,
+      allowUpload: false,
+      folder: undefined as ResultItemType | undefined
+    }
```
DrJKL, line 63: If you add a return type to the inner function in computed, do these types become inferrable?

DrJKL: Or satisfies to get the shape maybe?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
```diff
@@ -34,61 +42,63 @@ function handleUpdateModelValue(value: string | number | undefined) {
...
-    /\.(safetensors)$/i
+  const {
+    image_upload,
+    animated_image_upload,
+    video_upload,
+    widgetType,
+    image_folder
+  } = spec
+  const audioUpload = Boolean(
+    (spec as Partial<Record<'audio_upload', boolean>>).audio_upload
```
DrJKL: Can we constrain or augment the type of `comboSpec`?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
```diff
@@ -34,61 +42,63 @@ function handleUpdateModelValue(value: string | number | undefined) {
...
+    if (normalized.includes('model')) {
+      kind = 'model'
+    } else if (normalized.includes('audio')) {
+      kind = 'audio'
+    } else if (normalized.includes('image')) {
+      kind = 'image'
+    } else if (normalized.includes('video')) {
+      kind = 'video'
+    }
+  }
```
DrJKL, line 81: This many branches should _definitely_ be extracted to a function.
This could also be a decent case for `switch (true)`, though I know some people don't love that pattern.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
```diff
@@ -34,61 +42,63 @@ function handleUpdateModelValue(value: string | number | undefined) {
...
-function hasFilesWithExtensions(
-  values: any[],
-  extensionPattern: RegExp
-): boolean {
-  if (!Array.isArray(values) || values.length === 0) {
-    return false
+  return {
+    kind,
+    allowUpload: image_upload === true || animated_image_upload === true,
+    folder: image_folder
```
DrJKL, line 91: very nit:
```suggestion
  const allowUpload = image_upload || animated_image_upload
  return {
    kind,
    allowUpload
    folder: image_folder
```

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
```diff
@@ -34,61 +42,63 @@ function handleUpdateModelValue(value: string | number | undefined) {
...
-    if (typeof value !== 'string') return false
-    return extensionPattern.test(value)
-  })
-}
+const assetKind = computed(() => specDescriptor.value.kind)
+const isDropdownUIWidget = computed(() => assetKind.value !== 'unknown')
+const allowUpload = computed(() => specDescriptor.value.allowUpload)
+const uploadFolder = computed<ResultItemType>(() => {
+  return specDescriptor.value.folder ?? 'input'
+})
```
DrJKL, line 100: This looks like how I did a lot of RxJS 😁 
Except with operators

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -24,7 +26,9 @@ const props = defineProps<{
   widget: SimplifiedWidget<string | number | undefined>
   modelValue: string | number | undefined
   readonly?: boolean
-  fileType?: 'image' | 'video' | 'audio' | 'model' | 'unknown'
+  assetKind?: AssetKind
+  allowUpload?: boolean
+  uploadFolder?: ResultItemType
```
DrJKL, line 31: Quick check to make sure these are all independent of one another, please 🙏🏻

DrJKL (approved):
Just a lot of optional code style suggestions.
Nice catches on the i18n stuff.

src/composables/graph/useGraphNodeManager.ts
christian-byrne: What's a splat and override?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
christian-byrne: What do you mean exactly?

christian-byrne:
Removed import alias https://github.com/Comfy-Org/ComfyUI_frontend/pull/5798/commits/6a85dce808cd7fc12088943409b3af4d09892148

src/composables/graph/useGraphNodeManager.ts
christian-byrne: Good point, the optional properties are not needed explicitly in this fallback: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5798/commits/9cb37c1e0330212502e97a893dfab9b206294187

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5798/commits/814b7c38c9dd7128277f16778d54cb4d103f870f

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
christian-byrne: TYSM https://github.com/Comfy-Org/ComfyUI_frontend/pull/5798/commits/cfe2e663a946b01114005ee247691dc5ad3dbbd5

src/composables/graph/useGraphNodeManager.ts
DrJKL, line 205:
```suggestion
return {
...widget,
value,
spec
}
```

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
christian-byrne: Oh wow, that's bad, thanks for catching it: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5798/commits/cfe2e663a946b01114005ee247691dc5ad3dbbd5

src/composables/graph/useGraphNodeManager.ts
DrJKL: Added a suggestion for it. I'll show the other way too 😃

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
christian-byrne: The fallback using `widgetType` is actually bad and was removed -- we don't want to match on nodes who just happen to name a widget "audio" or "image".

https://github.com/Comfy-Org/ComfyUI_frontend/pull/5798/commits/cfe2e663a946b01114005ee247691dc5ad3dbbd5

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5798/commits/d0693f8beb93d2ece226caf1bffcf1d781ff2354

src/composables/graph/useGraphNodeManager.ts
DrJKL, line 205: This would be if you know for sure you only ever want these specific properties to carry forward.

```suggestion
const {name, type, options, callback} = widget
return {
name,
type,
options,
callback,
value,
spec
}
```

christian-byrne: We don't want all the properties of the widget though, there are a lot more and some of them are private (IIRC).

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
```diff
@@ -43,54 +43,51 @@ function handleUpdateModelValue(value: string | number | undefined) {
...
-    } else if (normalized.includes('video')) {
-      kind = 'video'
-    }
   }
+  // TODO: add support for models (checkpoints, VAE, LoRAs, etc.) -- get widgetType from spec
 
+  const allowUpload =
+    image_upload === true ||
+    animated_image_upload === true ||
+    audio_upload === true
```
DrJKL, line 87: optional:
```suggestion
    !!(image_upload ||
    animated_image_upload ||
    audio_upload)
```

src/composables/graph/useGraphNodeManager.ts
christian-byrne: The "safe widget" concept is about extracting the properties we know are not private and that we want to observe (temporary solution)

DrJKL (approved):

src/composables/graph/useGraphNodeManager.ts
DrJKL: Ah, that would be the second option then.

christian-byrne: Is there a specific reason why we should switch to shallow copy in this case? It seems somewhat out of scope of this PR, I'm not sure if we are relying on getting mutations to e.g. widget.options later on

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
christian-byrne: Is this because the API contract was allowing any truthy value before? Or is there a reason to make it less strict?

DrJKL: Code clarity. Doing extra work (explicit checks) usually indicates that there's a type issue that needs to be resolved. Mixing `true` an truthy values in a type that's checked later is a pretty bad code smell.

src/composables/graph/useGraphNodeManager.ts
DrJKL: You could do the same type of separate extraction for options, but in this case if any of _them_ are objects it's only a 1 layer deep copy.
I mostly want to be able to quickly see which pieces are special and which are proxied.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelect.vue
christian-byrne: But this would obscure our world view. At this point, we know the API schema defines the fields as Boolean and that the data has been deserialized from server response. I feel there should be an explicit benefit if wanting to incur that.

PR #4042, opened 2025-06-01
[fix] Automatically fix malformed node def translations

christian-byrne:
Adds cleanup step to i18n workflow to remove null entries that can occur when translation service encounters timing issues with node output changes.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4042-fix-Prevent-null-values-in-translated-node-definition-outputs-2056d73d365081939114e32419822145) by [Unito](https://www.unito.io)

PR #5801, opened 2025-09-26
1.27.6

AustinMroz:
Patch version increment to 1.27.6

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5801-1-27-27a6d73d365081b988b9fc1e2564d193) by [Unito](https://www.unito.io)


christian-byrne:
I'd say it's good to merge, missing translatinos for node defs is fine, as long as we are not corrupting the translation files or deleting translations. There was one bug in the past with lobe creating invalid JSON (see https://github.com/Comfy-Org/ComfyUI_frontend/pull/4042)

PR #5802, opened 2025-09-26
[ci] Update browser tests for new Templates modal

christian-byrne:
## Summary
- Regenerate 3 Playwright screenshot baselines to reflect UI changes from https://github.com/Comfy-Org/ComfyUI_frontend/pull/5142
- Also fixed [this case](https://f01efc75.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-379124415c5b7e9060d2) (test case for responsive sizing) as it was using outdated logic. New logic: ensures that the nav is collapsed on mobile but visible on tablet and desktop screen sizes. It also ensures that, at all the main breakpoints, at least 1 card is visible (covers bug that the case was originally written for wherein the nav was fully covering the cards at narrow screen widths).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5802-ci-Regenerate-Playwright-screenshot-baselines-27a6d73d365081768211da0d24bad2c3) by [Unito](https://www.unito.io)


christian-byrne:
Changes:

- Fixed [this case](https://f01efc75.comfyui-playwright-chromium.pages.dev/#?testId=35f0453d615a452757ca-379124415c5b7e9060d2) (test case for responsive sizing) as it was using outdated logic. New logic: ensures that the nav is collapsed on mobile but visible on tablet and desktop screen sizes. It also ensures that, at all the main breakpoints, at least 1 card is visible (covers bug that the case was originally written for wherein the nav was fully covering the cards at narrow screen widths).

browser_tests/tests/templates.spec.ts
```diff
@@ -181,45 +181,32 @@ test.describe('Templates', () => {
   }) => {
     // Open templates dialog
     await comfyPage.executeCommand('Comfy.BrowseTemplates')
-    await expect(comfyPage.templates.content).toBeVisible()
-
```
Myestery, line 185: These tests will make sense more on the base modal

Myestery (approved):
LGTM!

browser_tests/tests/templates.spec.ts
christian-byrne: The test case was originally about ensuring the grid was visible. The bug from a long time ago was the nav bar fully covering the grid. I think checking the template base modal visibility is covered in other test cases, so it should error rather than fail expectation in this case if not visible.

PR #5813, opened 2025-09-27
[fix] add InputSlot and dot error state

arjansingh:
## Summary

Error states were not getting propagated down to the InputSlots from the API Repsonse

I created a provider and injected error state. It seemed like a way better idea than prop drilling or building a composable that only two nodes (`InputSlot` and `OutputSlot`) would need. 

## Changes

The follow are now error code red when an input node has errors:
1. There's a error round border around the dot.
2. The dot is error colored.
3. The input text is error colored.

This treatment was okay after feedback from design.

## Screenshots - Error State

Screenshot 2025-09-26 at 9 02 58 PM

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5813-fix-add-InputSlot-and-dot-error-state-27b6d73d36508151a955e485f00a2d05) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -205,14 +214,22 @@ const hasExecutionError = computed(
...
+  const nodeValidationErrors = executionStore.lastNodeErrors?.[nodeData.id]
+  if (!nodeValidationErrors?.errors) return false
+
+  return nodeValidationErrors.errors.some(
+    (err) =>
+      err.type === 'required_input_missing' &&
+      err.extra_info?.input_name === inputName
+  )
+}
+
```
christian-byrne: The error state is set on the `NodeInputSlot`, it should be much simpler to get it that way

Image

christian-byrne (changes requested):

src/renderer/extensions/vueNodes/components/LGraphNode.vue
DrJKL: Would this be something we could query a store for, if we set it up?

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -57,7 +60,30 @@ interface InputSlotProps {
...
+  const slotName = props.slotData.name
+  return nodeErrors.errors.some(
+    (error) => error.extra_info?.input_name === slotName
+  )
+})
+
+const errorClassesDot = computed(() => {
+  return hasSlotError.value
+    ? 'ring-2 ring-error dark-theme:ring-error ring-offset-0 rounded-full'
+    : ''
```
DrJKL, line 78: nit: `cn` handles boolean/undefined/string, so it's pretty common to do `condition && 'class list'` which is simpler than a ternary with an empty string branch.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -101,7 +101,11 @@
       >
         <!-- Slots only rendered at full detail -->
         <NodeSlots
-          v-memo="[nodeData.inputs?.length, nodeData.outputs?.length]"
+          v-memo="[
+            nodeData.inputs?.length,
+            nodeData.outputs?.length,
+            executionStore.lastNodeErrors
+          ]"
```
DrJKL, line 108: It's pretty rare that you actually need v-memo. Could you see if it's necessary?

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -81,8 +107,12 @@ onErrorCaptured((error) => {
   return false
 })
 
-// Get slot color based on type
-const slotColor = computed(() => getSlotColor(props.slotData.type))
+const slotColor = computed(() => {
+  if (hasSlotError.value) {
+    return 'var(--color-error)'
```
DrJKL, line 112: I don't think this would do anything.

DrJKL (dismissed):
That is much cleaner, yes.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
arjansingh: i actually removed it during debugging but put it back because i figured i'd get chastised for removing it :Joy:

arjansingh: I think we can remove it, happy to follow on

src/renderer/extensions/vueNodes/components/InputSlot.vue
arjansingh: that's how the dot turns red.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: It prevents the slot from re-rendering on all `nodeData` changes.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -316,26 +316,19 @@ const { latestPreviewUrl, shouldShowPreviewImg } = useNodePreviewState(
...
-  }
-  if (executing.value) {
-    return 'outline-blue-500 dark-theme:outline-blue-500'
-  }
-  return 'outline-black dark-theme:outline-white'
+  return (
+    isSelected.value &&
+    ((hasAnyError.value && 'outline-error dark-theme:outline-error') ||
+      (executing.value && 'outline-blue-500 dark-theme:outline-blue-500') ||
+      'outline-black dark-theme:outline-white')
```
christian-byrne, line 330: nit: I feel just having some `if` blocks makes it a bit more readable.

christian-byrne (approved):
LGTM!

src/renderer/extensions/vueNodes/components/LGraphNode.vue
arjansingh: no remove then.

PR #5815, opened 2025-09-27
feat(auth): Allow SSO login only for whitelisted addresses (localhost)

bigcat88:
## Summary

Hide Google/GitHub SSO login options when the UI is accessed from **non‑local** addresses.
This PR also adds a **static whitelist** (editable in code) so we can allow additional hosts if needed.

Default whitelisted addresses:

1. `localhost` and any subdomain: `*.localhost`
2. IPv4 loopback `127.0.0.0/8` (e.g., `127.x.y.z`)
4. IPv6 loopback `::1` (including equivalent textual forms such as `::0001`)

## Changes

- **What**: 
  * Add `src/utils/hostWhitelist.ts` with `normalizeHost` and `isHostWhitelisted` helpers.
  * Update `SignInContent.vue` to **hide** SSO options when
    `isHostWhitelisted(normalizeHost(window.location.hostname))` returns `false`.
- **Breaking**:
  * Users accessing from Runpod or other previously allowed **non‑local** hosts will **lose** SSO login options.
    If we need to keep SSO there, we should add those hosts to the whitelist in `hostWhitelist.ts`.

## Review Focus

1. Verify that logging in from local addresses (`localhost`, `*.localhost`, `127.0.0.1`, `::1`) **does not change** the current behavior: SSO is visible.
2. Verify that from a **non‑local** address, SSO options are **not** displayed.

UI opened from `192.168.2.109` address:

Screenshot From 2025-09-27 13-22-15

UI opened from default `127.0.0.1` address(nothing changed):

Screenshot From 2025-09-27 13-35-27

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5815-feat-auth-Allow-SSO-login-only-for-whitelisted-addresses-localhost-27b6d73d365081ccbe84c034cf8e416d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/SignInContent.vue
+++ b/src/components/dialog/content/SignInContent.vue
@@ -145,7 +145,7 @@
 import Button from 'primevue/button'
 import Divider from 'primevue/divider'
 import Message from 'primevue/message'
-import { computed, onMounted, onUnmounted, ref } from 'vue'
+import { onMounted, onUnmounted, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 import { useFirebaseAuthActions } from '@/composables/auth/useFirebaseAuthActions'
@@ -167,9 +167,7 @@ const authActions = useFirebaseAuthActions()
 const isSecureContext = window.isSecureContext
 const isSignIn = ref(true)
 const showApiKeyForm = ref(false)
-const ssoAllowed = computed(() =>
-  isHostWhitelisted(normalizeHost(window.location.hostname))
-)
+const ssoAllowed = isHostWhitelisted(normalizeHost(window.location.hostname))
 
 const toggleState = () => {
   isSignIn.value = !isSignIn.value
--- a/src/components/dialog/content/SignInContent.vue
+++ b/src/components/dialog/content/SignInContent.vue
@@ -45,37 +45,39 @@
         <span class="text-muted">{{ t('auth.login.orContinueWith') }}</span>
       </Divider>
 
-      <!-- Social Login Buttons -->
+      <!-- Social Login Buttons (hidden if host not whitelisted) -->
       <div class="flex flex-col gap-6">
-        <Button
-          type="button"
-          class="h-10"
-          severity="secondary"
-          outlined
-          @click="signInWithGoogle"
-        >
-          <i class="pi pi-google mr-2"></i>
-          {{
-            isSignIn
-              ? t('auth.login.loginWithGoogle')
-              : t('auth.signup.signUpWithGoogle')
-          }}
-        </Button>
+        <template v-if="ssoAllowed">
+          <Button
+            type="button"
+            class="h-10"
+            severity="secondary"
+            outlined
+            @click="signInWithGoogle"
+          >
+            <i class="pi pi-google mr-2"></i>
+            {{
+              isSignIn
+                ? t('auth.login.loginWithGoogle')
+                : t('auth.signup.signUpWithGoogle')
+            }}
+          </Button>
 
-        <Button
-          type="button"
-          class="h-10"
-          severity="secondary"
-          outlined
-          @click="signInWithGithub"
-        >
-          <i class="pi pi-github mr-2"></i>
-          {{
-            isSignIn
-              ? t('auth.login.loginWithGithub')
-              : t('auth.signup.signUpWithGithub')
-          }}
-        </Button>
+          <Button
+            type="button"
+            class="h-10"
+            severity="secondary"
+            outlined
+            @click="signInWithGithub"
+          >
+            <i class="pi pi-github mr-2"></i>
+            {{
+              isSignIn
+                ? t('auth.login.loginWithGithub')
+                : t('auth.signup.signUpWithGithub')
+            }}
+          </Button>
+        </template>
 
         <Button
           type="button"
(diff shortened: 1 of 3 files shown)


christian-byrne (commented):
Can you add some unit tests for the utils file?

src/components/dialog/content/SignInContent.vue
```diff
@@ -164,6 +167,9 @@ const authActions = useFirebaseAuthActions()
 const isSecureContext = window.isSecureContext
 const isSignIn = ref(true)
 const showApiKeyForm = ref(false)
+const ssoAllowed = computed(() =>
+  isHostWhitelisted(normalizeHost(window.location.hostname))
+)
```
DrJKL: Does this need to be computed?

src/utils/hostWhitelist.ts
DrJKL, line 1: Would you mind adding just a few unit tests for this? I feel like they'd be a better explanation than the comments.

src/utils/hostWhitelist.ts
```diff
@@ -0,0 +1,99 @@
...
+  if (h.includes('::')) {
+    const [lhs, rhs] = h.split('::')
+    const leftSegs = lhs ? lhs.split(':').filter(Boolean) : []
+    const rightSegs = rhs ? rhs.split(':').filter(Boolean) : []
+    if (![...leftSegs, ...rightSegs].every(validHexSeg)) return false
+    const missing = 8 - (leftSegs.length + rightSegs.length)
+    if (missing < 1) return false // '::' must compress at least one zero group
+    const expanded = [...leftSegs, ...Array(missing).fill('0'), ...rightSegs]
+    return expanded.slice(0, 7).every(zeroSeg) && isOne(expanded[7])
+  }
```
DrJKL: Could this be replaced with a single regex?
https://uibakery.io/regex-library/ip-address

If you had that and some unit tests to double check its behavior I think I'd feel better.
Same for the IPv4 section.

DrJKL (changes requested):

src/components/dialog/content/SignInContent.vue
bigcat88: Good catch - this does not need to be computed because it depends only on `window.location.hostname`, which is not reactive.
Pushed a commit that removes this and makes a plain boolean.

src/utils/hostWhitelist.ts
bigcat88: That's a regex for IPv4 and only for single address. Adding check for `127/8` range will make that regex barely readable, imho.

For IPv6  we need a regex for checking the loopback/local address - it will be larger than that in the link.
It's easier to just check it with code than write a regex that no human will not understand and to make both IPv4 and IPv6.


And I have added tests for everyone to feel better. :rocket:

DrJKL: I understand most regexes better than I do chains of string manipulations, generally.

Also, that site has examples for both IPv4 and IPv6 with capture groups.

bigcat88: If you really need regex, then let it be regex - no problems. Pushed :)

DrJKL (approved):

src/components/dialog/content/SignInContent.vue
```diff
@@ -164,6 +167,7 @@ const authActions = useFirebaseAuthActions()
 const isSecureContext = window.isSecureContext
 const isSignIn = ref(true)
 const showApiKeyForm = ref(false)
+const ssoAllowed = isHostWhitelisted(normalizeHost(window.location.hostname))
```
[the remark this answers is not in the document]
christian-byrne: That still wouldn't be reactive though. We would want https://vueuse.org/core/useBrowserLocation/ right?

src/utils/hostWhitelist.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Fully expanded IPv6 loopback: 0:0:0:0:0:0:0:1 (allow leading zeros up to 4 chars)
+const V6_FULL_LOOPBACK_RE = /^(?:0{1,4}:){7}0{0,3}1$/i
+
+// Compressed IPv6 loopback forms around '::' with only zero groups before the final :1
+// - Left side: zero groups separated by ':' (no trailing colon required)
+// - Right side: zero groups each followed by ':' (so the final ':1' is provided by the pattern)
+// The final group is exactly value 1, with up to 3 leading zeros (e.g., '0001').
+const V6_COMPRESSED_LOOPBACK_RE =
```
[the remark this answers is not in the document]
christian-byrne: This code is executed on the client.

src/utils/hostWhitelist.ts
```diff
@@ -0,0 +1,91 @@
...
+  return h
+}
+
+/** Public check used by the UI. */
+export function isHostWhitelisted(rawHost: string): boolean {
+  const host = normalizeHost(rawHost)
+  if (isLocalhostLabel(host)) return true
+  if (isIPv4Loopback(host)) return true
+  if (isIPv6Loopback(host)) return true
+  const normalizedList = HOST_WHITELIST.map(normalizeHost)
```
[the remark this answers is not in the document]
christian-byrne: Not necessary

christian-byrne (approved):

src/components/dialog/content/SignInContent.vue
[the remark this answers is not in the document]
DrJKL: Yeah, though how often do users change the hostname while staying on the page?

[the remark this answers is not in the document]
christian-byrne: Good point. I guess this is for the specific case where you are using a SPA client-side router that isn't vue router.

PR #5821, opened 2025-09-27
Check if the wheel event is from an element that wants to capture wheel events

### LINKED ISSUE #5691: Try to scroll the Vue node widget options, but it seems to conflict with the canvas zoom in/out.
https://github.com/user-attachments/assets/84752bb0-c278-434e-ae4e-eb8a123cf168

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-5691-Try-to-scroll-the-VUE-node-widget-options-but-it-seems-to-conflict-with-the-canvas-zo-2746d73d365081ec839bef8060ae9d5e) by [Unito](https://www.unito.io)

jtydhr88:
## Summary

  Add generic wheel event capture mechanism for interactive widgets in vueNodes system to prevent event bubbling to canvas.

  ## Changes

  - What: Add event handling logic in LGraphNode.vue and GraphCanvas.vue that checks for data-capture-wheel attribute to determine whether wheel events should be forwarded to the canvas
  - How it works: Components that need to capture wheel events (like Three.js scenes) can add data-capture-wheel="true" attribute to prevent wheel events from bubbling up to the canvas zoom handler

prerequirist for https://github.com/Comfy-Org/ComfyUI_frontend/pull/5765

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5821-Check-if-the-wheel-event-is-from-an-element-that-wants-to-capture-wheel-events-27b6d73d3650812493b5f13849147e6c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/core/canvas/useCanvasInteractions.ts
+++ b/src/renderer/core/canvas/useCanvasInteractions.ts
@@ -30,6 +30,15 @@ export function useCanvasInteractions() {
    * when appropriate (e.g., Ctrl+wheel for zoom in standard mode)
    */
   const handleWheel = (event: WheelEvent) => {
+    // Check if the wheel event is from an element that wants to capture wheel events
+    const target = event.target as HTMLElement
+    const captureElement = target?.closest('[data-capture-wheel="true"]')
+
+    if (captureElement) {
+      // Element wants to capture wheel events, don't forward to canvas
+      return
+    }
+
     // In standard mode, Ctrl+wheel should go to canvas for zoom
     if (isStandardNavMode.value && (event.ctrlKey || event.metaKey)) {
       forwardEventToCanvas(event)
@@ -72,6 +81,15 @@ export function useCanvasInteractions() {
   const forwardEventToCanvas = (
     event: WheelEvent | PointerEvent | MouseEvent
   ) => {
+    // Check if the wheel event is from an element that wants to capture wheel events
+    const target = event.target as HTMLElement
+    const captureElement = target?.closest('[data-capture-wheel="true"]')
+
+    if (captureElement) {
+      // Element wants to capture wheel events, don't forward to canvas
+      return
+    }
+
     const canvasEl = app.canvas?.canvas
     if (!canvasEl) return
     event.preventDefault()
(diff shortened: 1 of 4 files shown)


src/components/graph/GraphCanvas.vue
```diff
@@ -369,6 +369,18 @@ useCanvasDrop(canvasRef)
 useLitegraphSettings()
 useNodeBadge()
 
+const handleTransformPaneWheel = (event: WheelEvent) => {
+  // Check if the wheel event is from an element that wants to capture wheel events
+  const target = event.target as HTMLElement
```
DrJKL: I'd prefer an `instanceof` check for the target, if possible.

src/components/graph/GraphCanvas.vue
```diff
@@ -369,6 +369,18 @@ useCanvasDrop(canvasRef)
...
+  const captureElement = target?.closest('[data-capture-wheel="true"]')
```
DrJKL: Do we have this set on anything currently?

DrJKL (commented):
CC @arjansingh and @christian-byrne who have also been trying to handle Litegraph's aggressive event capturing.

src/components/graph/GraphCanvas.vue
jtydhr88: no, but I will use it in 3d node later if we merge this change.

DrJKL: Could you flip the order so that this is functional at the point that it merges?

arjansingh:
> CC @arjansingh and @christian-byrne who have also been trying to handle Litegraph's aggressive event capturing.

https://github.com/Comfy-Org/ComfyUI_frontend/pull/5793

we're not merging this but this is how proposed dealing with wheel events

jtydhr88:
> > CC @arjansingh and @christian-byrne who have also been trying to handle Litegraph's aggressive event capturing.
> 
> #5793
> 
> we're not merging this but this is how proposed dealing with wheel events

Ok, thanks for the information, I can check this can fix my issue in 3d node

comfyui-wiki:
https://github.com/user-attachments/assets/f8166c38-46bb-478e-9033-85553febb6e8

After testing, I think this PR can solve #5691.

christian-byrne (approved):

PR #5826, opened 2025-09-28
Add UI code for configuring subgraphNode widgets

AustinMroz:
The third PR for managing display of widgets on subgraph nodes. This is the one that actually makes the functionality usable and user visible.

Adds
- A right-side modal for configuring which widgets are promoted, accessed by right click or selection toolbar
  - This menu allows for re-arranging widget order by dragging and dropping.
- Indicators inside the subgraph for which widgets have been promoted.
- Context menu options for promoting or demoting widget inside of a subgraph.
image
image

Known issues
- Some preview widgets are not added to a node until a draw operation occurs. The code does not yet have a way of determining which nodes should have draw operations forced to facilitate initial widget creation.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5826-Add-UI-code-for-configuring-subgraphNode-widgets-27c6d73d36508146accbf395e5bcd36a) by [Unito](https://www.unito.io)


src/components/graph/selectionToolbox/ConfigureSubgraph.vue
```diff
@@ -0,0 +1,24 @@
...
+  </Button>
+</template>
+
+<script setup lang="ts">
+import Button from 'primevue/button'
+import { useI18n } from 'vue-i18n'
+
+import { showSubgraphNodeDialog } from '@/core/graph/subgraph/useSubgraphNodeDialog'
+
+const { t } = useI18n()
```
christian-byrne: nit: in main.ts, the i18n plugin is registered globally:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/a2c7db9dc205df79a0b70cc82cae77175ecaad77/src/main.ts#L66

This plugin registers a `t` globalProperty - allowing the compiler to resolve any `$t` to that function when used in `<template>`. So you can use `$t('translationKey')` without needing to import anything in the `<script>` section.

https://vuejs.org/api/application#app-config-globalproperties

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,288 @@
...
+const activeNode = computed(() => {
+  const node = canvasStore.selectedItems[0]
+  if (node instanceof SubgraphNode) return node
+  useDialogStore().closeDialog()
+  return undefined
+})
+
+const activeWidgets = computed<WidgetItem[]>({
+  get() {
+    if (triggerUpdate.value < 0) console.log('unreachable')
```
christian-byrne: question (non-blocking): What's this?

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,288 @@
...
+    :placeholder="$t('g.search') + '...'"
+  />
+  <div v-if="filteredActive.length" class="widgets-section">
+    <div class="widgets-section-header">
+      <div>{{ t('subgraphStore.shown') }}</div>
+      <a @click.stop="hideAll"> {{ t('subgraphStore.hideAll') }}</a>
+    </div>
+    <div v-if="searchQuery" class="w-full">
+      <div
+        v-for="element in filteredActive"
```
christian-byrne: nit (non-blocking; minor): Is there a name that conveys more meaning than `element` here?

src/core/graph/subgraph/SubgraphNodeWidget.vue
```diff
@@ -0,0 +1,65 @@
...
+    <div class="flex-1">
+      <div class="widget-node">{{ nodeTitle }}</div>
+      <div class="widget-name">{{ widgetName }}</div>
+    </div>
+    <Button
+      size="small"
+      class="shrink-0"
+      text
+      severity="secondary"
+      @click.stop="onClick"
```
christian-byrne: suggestion: here, it seems it might be easier to just `emit` something on click then let the parent handle the visibility logic. Currently we are passing a callback and the params, but the params are not modified in the component as far as I can tell.

src/core/graph/subgraph/SubgraphNodeWidget.vue
```diff
@@ -0,0 +1,65 @@
...
+  font-family: Inter;
+  font-size: 10px;
+}
+.widget-name {
+  color: var(--color-text-primary, #fff);
+
+  /* body-text-small */
+  font-family: Inter;
+  font-size: 12px;
+}
```
christian-byrne: suggestion: this code might benefit from using tailwind utility classes in favor of CSS in `<style>`.

In a lot of cases, we may benefit from using tailwind utility classes as they reference our design tokens which allows for easier refactoring of design system. Also, design token will better handle responsiveness or e.g. browser-level font settings (e.g., changing from `padding: 4px` to `p-1` will use 0.25rem which changes based on the user's browser settings).

See:

- px to tailwind converter: https://nekocalc.com/px-to-tailwind-converter
- more reasons to use Tailwind: https://tailwindcss.com/docs/styling-with-utility-classes

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,288 @@
...
+        :node-id="`${element[0].id}`"
+        :node-title="element[0].title"
+        :widget-name="element[1].name"
+        :toggle-visibility="toggleVisibility"
+      />
+    </div>
+  </div>
+  <div v-if="recommendedWidgets.length" class="justify-center flex py-4">
+    <Button size="small" @click.stop="showRecommended">
+      {{ t('subgraphStore.showRecommended') }}
```
christian-byrne: I think this needs to be added to the `src/locales/en/main.json` file.

src/services/litegraphService.ts
```diff
@@ -824,18 +826,88 @@ export const useLitegraphService = () => {
...
-            this.graph.unpackSubgraph(this)
+        options.unshift(
+          {
+            content: 'Edit Subgraph Widgets',
+            callback: () => {
+              showSubgraphNodeDialog()
+            }
+          },
+          {
+            content: 'Unpack Subgraph',
```
christian-byrne, line 837: nit: can use i18n here.

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,288 @@
...
+  }
+  triggerUpdate.value++
+}
+
+function nodeWidgets(n: LGraphNode): WidgetItem[] {
+  if (!n.widgets) return []
+  return n.widgets.map((w: IBaseWidget) => [n, w])
+}
+
+const candidateWidgets = computed<WidgetItem[]>(() => {
```
christian-byrne, line 105: question: how often does this get re-computed during runtime when using subgraph nodes? If it's a hotpath, there's definitely ways to add some memoization, but it might not be necessary.

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,288 @@
...
+      if (!wNode?.widgets) return []
+      const w = wNode.widgets.find((w) => w.name === name)
+      if (!w) return []
+      return [[wNode, w]]
+    })
+  },
+  set(value: WidgetItem[]) {
+    const node = activeNode.value
+    if (!node)
+      throw new Error('Attempted to toggle widgets with no node selected')
```
christian-byrne: suggestion: throwing in computed setters can have some bad hard-to-debug effects, consider if we can handle this gracefully.

src/core/graph/subgraph/SubgraphNode.vue
christian-byrne, line 1: Two question about this file (SubgraphNode.vue):

1. Can any of the proxy widget logic be refactored to extract some of the common operations and use the proxy widget utils? This might make it easier to add unit tests and ensure fixes are propagated should they occur.
2. Is there anything we should do in the `onUnmounted` lifecycle hook for component? Any proxies to remove, maps to clear, in-flight processes to cancel, etc.?

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,288 @@
...
+import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
+import { useDialogStore } from '@/stores/dialogStore'
+
+type WidgetItem = [LGraphNode, IBaseWidget]
+
+const { t } = useI18n()
+
+const canvasStore = useCanvasStore()
+
+const searchQuery = ref<string>('')
```
christian-byrne, line 39: suggestion (non-blocking): this might be a good use case for https://vueuse.org/shared/refDebounced/

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,288 @@
...
+} from '@/core/schemas/proxyWidget'
+import type { LGraphNode } from '@/lib/litegraph/src/litegraph'
+import { SubgraphNode } from '@/lib/litegraph/src/subgraph/SubgraphNode'
+import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets'
+import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
+import { useDialogStore } from '@/stores/dialogStore'
+
+type WidgetItem = [LGraphNode, IBaseWidget]
+
+const { t } = useI18n()
```
christian-byrne: nit (non-blocking): same comment as [this one](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5826/files#r2386201185) (I think it also applies here).

christian-byrne (commented):
This looks good, I left a lot of very minor comments on the first pass through. Have not tested yet!

src/core/graph/subgraph/SubgraphNode.vue
AustinMroz: A kludge workaround to `@typescript-eslint/no-unused-expressions`. Shame on me for not adding a reminder comment to return to it later. 

I've made a fairly significant rewrite to instead make proxyWidgets access reactive.

src/core/graph/subgraph/SubgraphNode.vue
AustinMroz: Thinking it over, it shouldn't be possible for interior widgets to change without changing selected node. I've split it out into a separate computed so that there's some caching.

src/components/graph/selectionToolbox/ConfigureSubgraph.vue
```diff
@@ -0,0 +1,21 @@
...
+    v-tooltip.top="{
+      value: $t('Edit Subgraph Widgets'),
+      showDelay: 1000
+    }"
+    severity="secondary"
+    text
+    @click="showSubgraphNodeDialog"
+  >
+    <template #icon>
+      <i-lucide:settings2 />
```
DrJKL: not necessary, but I'd like to standardize on the class form of these over the dynamic component ones.

src/components/graph/selectionToolbox/ConfigureSubgraph.vue
```diff
@@ -0,0 +1,21 @@
...
+  <Button
+    v-tooltip.top="{
+      value: $t('Edit Subgraph Widgets'),
+      showDelay: 1000
+    }"
+    severity="secondary"
+    text
+    @click="showSubgraphNodeDialog"
+  >
+    <template #icon>
```
DrJKL: Does this need a `<template>`?

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -43,8 +47,25 @@ function isProxyWidget(w: IBaseWidget): w is ProxyWidget {
   return (w as { _overlay?: Overlay })?._overlay?.isProxyWidget ?? false
 }
 
+export function registerProxyWidgets(canvas: LGraphCanvas) {
+  //NOTE: canvasStore hasn't been initialized yet
```
DrJKL, line 52: This comment raises questions I didn't have before...

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -43,8 +47,25 @@ function isProxyWidget(w: IBaseWidget): w is ProxyWidget {
...
+  canvas.canvas.addEventListener<'subgraph-opened'>('subgraph-opened', (e) => {
+    const { subgraph, fromNode } = e.detail
+    const pw = parseProxyWidgets(fromNode.properties.proxyWidgets)
+    for (const node of subgraph.nodes) {
+      for (const widget of node.widgets ?? []) {
+        widget.promoted = pw.some(([n, w]) => node.id == n && widget.name == w)
```
DrJKL: This feels like logic that should be composed in a store instead of done procedurally in response to events

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -89,16 +113,18 @@ function addProxyWidget(
...
-    computedHeight: undefined,
-    afterQueued: undefined,
+    name,
+    node: subgraphNode,
     onRemove: undefined,
-    node: subgraphNode
+    promoted: undefined,
+    serialize: false,
+    width: undefined,
+    y: 0
```
DrJKL, line 132: nits:
- This is a really big object literal to initialize all at once
- Could you sort it either alphabetically or in order of
  - Base properties
  - New properties
  - Properties that are being explicitly set to undefined (?)
Or something like that

src/core/graph/subgraph/proxyWidget.ts
```diff
@@ -155,7 +178,9 @@ function addProxyFromOverlay(subgraphNode: SubgraphNode, overlay: Overlay) {
       let redirectedReceiver = receiver
       if (property == 'value') redirectedReceiver = backingWidget
       else if (property == 'computedHeight') {
-        //update linkage regularly, but no more than once per frame
+        if (overlay.widgetName.startsWith('$$') && linkedNode)
+          updatePreviews(linkedNode)
+          //update linkage regularly, but no more than once per frame
         ;[linkedNode, linkedWidget] = resolveLinkedWidget(overlay)
```
DrJKL: option:
```suggestion
        if (overlay.widgetName.startsWith('$$') && linkedNode) {
          updatePreviews(linkedNode)
          //update linkage regularly, but no more than once per frame
          }
[linkedNode, linkedWidget] = resolveLinkedWidget(overlay)
```

src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+import type {
+  IContextMenuValue,
+  LGraphNode
+} from '@/lib/litegraph/src/litegraph'
+import type { SubgraphNode } from '@/lib/litegraph/src/subgraph/SubgraphNode'
+import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets.ts'
+import { useSubgraphNavigationStore } from '@/stores/subgraphNavigationStore'
+
+function pushWidgets(node: SubgraphNode, ...widgets: [string, string][]) {
+  const pw = getProxyWidgets(node)
```
DrJKL: nit: I prefer having full names over initials, helps with understanding without having to translate

src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+ * Enables display of a widget on the parent subgraphNode
+ * @param {IBaseWidget} widget - The widget to be promoted
+ * @param {LGraphNode} node - the node which owns the widget
+ */
+function promoteWidget(
+  widget: IBaseWidget,
+  node: LGraphNode,
+  parents: SubgraphNode[]
+) {
+  for (const parent of parents) pushWidgets(parent, [`${node.id}`, widget.name])
```
DrJKL: The path to get to/from here seems a little circuitous.
Is there a way to straighten it out, function chain style?

DrJKL: It's also a lot of mutations

src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+  //Either by including actual parents in the navigation stack,
+  //or by adding a new event for parent listeners to collect from
+  const { navigationStack } = useSubgraphNavigationStore()
+  const subgraph = navigationStack.at(-1)
+  if (!subgraph) throw new Error("Can't promote widget when not in subgraph")
+  const validNodes = []
+  const parentGraph = navigationStack.at(-2) ?? subgraph.rootGraph
+  for (const onode of parentGraph.nodes) {
+    if (onode.type === subgraph.id && onode.isSubgraphNode()) {
+      validNodes.push(onode)
```
DrJKL: nit: could you use `.filter()`?

src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+  node: LGraphNode
+) {
+  const parents = getParentNodes()
+  const promotableParents = parents.filter(
+    (s) =>
+      !getProxyWidgets(s).some(
+        ([id, name]) => node.id == id && widget.name === name
+      )
+  )
+  if (promotableParents.length > 0)
```
DrJKL, line 83: Not for this PR, but I really want to get away from the pattern of mutating arrays/objects that we pass around.

src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+const recommendedNodes = [
+  'CLIPTextEncode',
+  'LoadImage',
+  'SaveImage',
+  'PreviewImage'
+]
+const recommendedWidgetNames = ['seed']
+function nodeWidgets(n: LGraphNode): WidgetItem[] {
+  if (!n.widgets) return []
+  return n.widgets.map((w: IBaseWidget) => [n, w])
```
DrJKL: option:
```suggestion
  return n.widgets?.map((w: IBaseWidget) => [n, w]) ?? []
```

src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -0,0 +1,126 @@
...
+    .filter(
+      ([node, widget]: WidgetItem) =>
+        recommendedNodes.includes(node.type) ||
+        recommendedWidgetNames.includes(widget.name)
+    )
+  const pw: ProxyWidgetsProperty = filteredWidgets.map(([n, w]: WidgetItem) => [
+    `${n.id}`,
+    w.name
+  ])
+  subgraphNode.properties.proxyWidgets = JSON.stringify(pw)
```
DrJKL: Why a string?
That feels wrong to me.

src/core/graph/subgraph/SubgraphNodeWidget.vue
AustinMroz: Easier is debatable, but certainly more idiomatic and cleaner. Thanks

src/services/litegraphService.ts
AustinMroz: Done, but a bit of a rabbit hole. Translation is already being applied to context menu option fields, so adding translation here simply requires adding the field to locales.

As part of this change, I've removed the 🆕 tag from the "Convert to Subgraph" option. It was preventing translation.

src/core/graph/subgraph/SubgraphNode.vue
AustinMroz: 1. This was originally a goal. Took more work than expected, but the result with emit change is clean.
2. Nope. :+1:

src/core/graph/subgraph/SubgraphNode.vue
AustinMroz: There was a little bit of back and forth with design over whether the button should be kept. There's currently a slight lean towards "yes".

In the future, I think I'll commit to marking items which should be translated with a placeholder tag like `t('[PH]Show Recommended Widgets')`. Not wanting to commit to locale changes early is fine. Having mixed, miss-able placeholders is problematic.

AustinMroz:
Current playwritght test failures are from removing the 🆕 character from "Convert to Subgraph" to fix translation and are expected.

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,284 @@
...
+  triggerRef(proxyWidgets)
+}
+function promote([node, widget]: WidgetItem) {
+  const subgraphNode = activeNode.value
+  if (!subgraphNode) return []
+  promoteWidget(node, widget, [subgraphNode])
+  triggerRef(proxyWidgets)
+}
+
+const candidateWidgets = computed<WidgetItem[]>(() => {
```
DrJKL, line 105: nit: Could you order the file so that the refs/computeds/actions/pure functions are together?

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,284 @@
...
+}
+function hideAll() {
+  const node = activeNode.value
+  if (!node) return //Not reachable
+  //Not great from a nesting perspective, but path is cold
+  //and it cleans up potential error states
+  proxyWidgets.value = proxyWidgets.value.filter(
+    ([nodeId, widgetName]) =>
+      !filteredActive.value.some(
+        ([n, w]: WidgetItem) => n.id == nodeId && w.name === widgetName
```
DrJKL: It feels like some of these patterns are repeated enough times to be worth extracting and naming.

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,284 @@
...
+      )
+  )
+}
+const recommendedNodes = [
+  'CLIPTextEncode',
+  'LoadImage',
+  'SaveImage',
+  'PreviewImage'
+]
```
DrJKL: Deja vu...

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,284 @@
...
+    ([n, w]: WidgetItem) =>
+      n.title.toLowerCase().includes(query) ||
+      w.name.toLowerCase().includes(query)
+  )
+})
+</script>
+<template>
+  <SearchBox
+    v-model:model-value="searchQuery"
+    class="model-lib-search-box p-2 2xl:p-4"
```
DrJKL: Where is `.model-lib-search-box` used?

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,284 @@
...
+  <div
+    v-if="filteredActive.length"
+    class="pt-1 pb-4 border-b-1 border-[var(--color-node-divider,#2E3037)]"
+  >
+    <div class="flex py-0 px-4 justify-between">
+      <div class="text-slate-100 text-[9px] font-semibold uppercase">
+        {{ $t('subgraphStore.shown') }}
+      </div>
+      <a
+        class="cursor-pointer text-right text-blue-100 text-[11px] font-normal"
```
DrJKL, line 256: 11px feels very specific.

src/core/graph/subgraph/SubgraphNode.vue
```diff
@@ -0,0 +1,284 @@
...
+    </div>
+    <div v-if="debouncedQuery" class="w-full">
+      <div
+        v-for="widgetItem in filteredActive"
+        :key="toKey(widgetItem)"
+        class="w-full"
+      >
+        <SubgraphNodeWidget
+          :node-title="widgetItem[0].title"
+          :widget-name="widgetItem[1].name"
```
DrJKL: Are you able to destructure the values in `v-for`?

src/core/graph/subgraph/SubgraphNodeWidget.vue
```diff
@@ -0,0 +1,36 @@
...
+  isShown?: boolean
+  isDraggable?: boolean
+}>()
+defineEmits<{
+  (e: 'toggleVisibility'): void
+}>()
+</script>
+<template>
+  <div
+    class="flex pt-1 pr-4 pb-1 pl-0 break-all rounded bg-[var(--p-dialog-background,#202020)] items-center gap-1"
```
DrJKL: Can we use one of our theme colors for the background here?

DrJKL: optional:
```suggestion
    class="flex py-1 pr-4 pl-0 break-all rounded bg-[var(--p-dialog-background,#202020)] items-center gap-1"
```

src/core/graph/subgraph/SubgraphNodeWidget.vue
```diff
@@ -0,0 +1,36 @@
...
+  >
+    <div class="size-4">
+      <i-lucide:grip-vertical v-if="isDraggable" />
+    </div>
```
DrJKL: Do you need the wrapper div in this case?

src/core/graph/subgraph/useSubgraphNodeDialog.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export function showSubgraphNodeDialog() {
+  const dialogStore = useDialogStore()
+  const dialogComponentProps: DialogComponentProps = {
+    headless: true,
+    modal: false,
+    closable: false,
+    position: 'right'
+  }
+  //FIXME: the vuedraggable import has unknown sideffects that break tests.
```
DrJKL: ;_;

src/lib/litegraph/src/infrastructure/LGraphCanvasEventMap.ts
```diff
@@ -14,6 +15,11 @@ export interface LGraphCanvasEventMap {
     /** The old active graph, or `null` if there was no active graph. */
     oldGraph: LGraph | Subgraph | null | undefined
   }
+  'subgraph-opened': {
+    subgraph: Subgraph
+    closingGraph: LGraph
+    fromNode: SubgraphNode
+  }
```
DrJKL, line 22: Not necessarily something to address here, but I've noticed we have a lot of the graph/subgraph state handling split across components, stores, services, and litegraph itself, with a mix of events and mutations.
I'd love to find a way to consolidate the behaviors into a single stream and have a source of truth for the canvas/graph state.

src/lib/litegraph/src/types/widgets.ts
```diff
@@ -306,6 +306,12 @@ export interface IBaseWidget<
 
   hidden?: boolean
   advanced?: boolean
+  /**
+   * Set if the node is displayed on the parent subgraphNode
+   * Promoted widgets have a green border
+   * @readonly [Computed] This property is computed on graph change
+   */
```
DrJKL, line 314: This is a lot of comment...
What happens if we change the theme color for the border here?

src/scripts/domWidget.ts
```diff
@@ -160,16 +160,24 @@ abstract class BaseDOMWidgetImpl<V extends object | string>
...
-    if (this.options.hideOnZoom && lowQuality && this.isVisible()) {
+    if (
+      (this.promoted || (this.options.hideOnZoom && lowQuality)) &&
+      this.isVisible()
+    ) {
       // Draw a placeholder rectangle
       const originalFillStyle = ctx.fillStyle
       ctx.beginPath()
       ctx.fillStyle = LiteGraph.WIDGET_BGCOLOR
+      let margin = this.margin
```
DrJKL: nit: I have a general preference against mutable state like this.

tests-ui/tests/widgets/proxyWidget.test.ts
```diff
@@ -1,21 +1,30 @@
...
+} from '@/lib/litegraph/src/litegraph'
 
 import {
   createTestSubgraph,
   createTestSubgraphNode
 } from '../litegraph/subgraph/fixtures/subgraphHelpers'
 
+registerProxyWidgets({
+  canvas: { addEventListener() {} }
+} as unknown as LGraphCanvas)
```
DrJKL: Normally for something like this I like to explicitly pull out a function that declares an internal `Partial<Foo>` that is then returned as the full, so there's some level of type safety in the properties that are defined.

src/components/graph/selectionToolbox/ConfigureSubgraph.vue
AustinMroz: Pretty sure it wasn't merged when I first started the feature branch. Thanks for the reminder. I prefer it too!

src/core/graph/subgraph/proxyWidget.ts
AustinMroz: In GraphCanvas.vue
```
await comfyApp.setup(canvasRef.value)
attachSlotLinkPreviewRenderer(comfyApp.canvas)
canvasStore.canvas = comfyApp.canvas
```
It's close. Two steps up the callstack and two lines.

src/core/graph/subgraph/proxyWidget.ts
AustinMroz: It puts the Overlay properties first , followed by the IBaseWidget properties in alphabetical order.

I've added a line gap between the two and made the first group alphabetical

src/core/graph/subgraph/proxyWidget.ts
AustinMroz: Hmm... I think this would be the right choice architecturally, but would require some heavy refactoring to the existing logic.

Changes would be prerequisite to properly handling the navigation stack for linked subgraphs, but are best separated into a followup PR.

src/components/graph/selectionToolbox/ConfigureSubgraph.vue
AustinMroz: It does not.

src/core/graph/subgraph/SubgraphNodeWidget.vue
AustinMroz: With the power of class form, it is not.

src/lib/litegraph/src/types/widgets.ts
christian-byrne: I like jsdoc comments especially on a widely used public property `widgets`.

src/scripts/domWidget.ts
christian-byrne: @DrJKL Could you elaborate a bit on what "mutable state" means in this context and in general in these types of scenarios?

DrJKL: Certainly 😁 

I try to avoid ever having to use `let`, if I can instead calculate the closed form and instantiate the variable with its final value so that
- You never need to trace back to see how it started or what operations _might_ have been applied
- Future engineers aren't tempted to further tweak the values in ways that can compound the complexity.

In this case, something like
```ts
ctx.fillStyle = this.promoted ? LiteGraph.WIDGET_PROMOTED_OUTLINE_COLOR : LiteGraph.WIDGET_BGCOLOR
const finalMargin = this.promoted ? this.margin - 1 : this.margin
```

(or `adjustedMargin`, or something that indicates that it's a derived value and not just a local reference to the class field with the same name)

src/core/graph/subgraph/proxyWidgetUtils.ts
AustinMroz: I can point to the exact commit where the change was made (ccc8d6e44), but I can no longer reproduce the type errors I was experiencing prior to the change or find the specific conflicting type definition. Even the presumed but never actually tested edge cases like "Editing the proxyWidgets property through the context menu" work entirely as desired.

Since I too prefer the lesser overhead, I've made the swap back to storing the actual array.

src/lib/litegraph/src/types/widgets.ts
DrJKL: I would too, but in this case
- It's declaring it readonly in jsdoc but not in typescript
- It's specifying a display aspect that isn't related to the behavior of the object (and is subject to change)
- It hints at the fact that it's actually a part of what would otherwise be a subclass / subset of the broader interface (a widget that is specific to subgraphNode) without `@link`ing to the term
- It describes a procedure about how the value _might_ come to be set (I don't trust comments when it comes to provenance, they're often out of date or were lies at the point they were written)

src/core/graph/subgraph/SubgraphNode.vue
AustinMroz: It's from `components/sidebar/tabs/ModelLibrarySidebarTab.vue`, but as it turns out, it's not used there either.

src/lib/litegraph/src/types/widgets.ts
webfiltered: I would _love_ to see the pattern of "`@readonly` only in the JSDocs" only get disappeared.

HCL started adding this as there was no easy way to deal with some of the migration annoyances at the time.  It was fine as a temporary measure, but never got cleaned up.  Then it got mimicked by LLMs repeatedly (I squashed a lot of that).

The idea is that this property is _not_ actually readonly, but should be treated as readonly by downstream consumers.  I learned this through code examination - it was not obvious ... and was not worth the time I spent finding out .. or that we are spending on it now haha.

We can just write "Computed, but should be treated as readonly".  If we need it in generated docs, we should middleware -that-, instead (have an LLM fork vite-dts-plugin or something).

**TL;DR**
If this is a NEW property, ideally it should be a getter pointing at a private (TS/`_convention`) mutable property.

src/core/graph/subgraph/SubgraphNodeWidget.vue
AustinMroz: Sure.

It doesn't inspire confidence that for the closest match (`charcoal-800`) the codebase is split across 4 different values for the now required light theme equivalent (`gray-50, gray-100, white, pure-white`), but I'd rather not be the one to fry that fish.

tests-ui/tests/widgets/proxyWidget.test.ts
AustinMroz: I think the function is a little overkill here, but I quite like the usage of `Partial` over `Unknown`. It lets us get type checking on portion we actually mock.

src/core/graph/subgraph/SubgraphNode.vue
AustinMroz: Requires some repacking, but it's better.

src/core/graph/subgraph/useSubgraphNodeDialog.ts
AustinMroz: Much screaming was had, but the import has been replaced.

src/lib/litegraph/src/infrastructure/LGraphCanvasEventMap.ts
AustinMroz: Agreed. I've been starting to think real hard about "How a user would expect linked subgraphs to work" and "How this could actually be accomplished in code.".

Here, the subgraph navigation stack is insufficient because it just "searches" for a route to the current subgraph every time the graph changes. Things would be much nicer if it properly tracked parent subgraphNodes and could be used as the single source of truth.

src/core/graph/subgraph/SubgraphNode.vue
AustinMroz: I'm trusting the design team here. `text-xs` is slightly too large and I'm not aware of a shorthand for 11px

src/core/graph/subgraph/proxyWidgetUtils.ts
AustinMroz: I keep returning to this one, but I neither see a way it could be cleaned up, or how the number of mutations is anything short of minimal
- The code *really should* account for linked subgraphs even if the length of parent nodes is expected to always be 1
- Each parent has the singular mutation explicitly described by the function name applied.
- forEach is verboten

DrJKL: - You get each parent
- You call pushWidgets
- pushWidgets calls getProxyWidgets
- getProxyWidgets calls parseProxyWidgets
- parseProxyWidgets goes into string land to try to deserialize any existing widgets (using safeParse, but also throwing its own error sometimes)
- Now you have, presumably, an array of _something_ (probably string,string tuples?)
- You dump new values in there (where I'd probably expect a uniqueness constraint, but I don't know if that's valid)
- You then stringify again


`demoteWidgets` has a similar structure but looks different and does things inline instead of abstracting to a function

It's just kind of a lot to have to trace for the logic here.
I still don't understand why we're ever stringifying outside of a serialization context.

DrJKL: The functional form would be more like:
Given an existing set of "proxy widget" tuples and a target, return a new set that either adds or removes the target widget.
Then you could use that pure function across the relevant nodes.
It also makes undo/redo a lot more straightforward if you can stack these sorts of operations.

src/core/graph/subgraph/proxyWidget.ts
AustinMroz: Even with the added braces, Prettier will still add the semicolon back.

I had originally missed this, so I've reopened this conversation for bookkeeping.

christian-byrne (dismissed):

christian-byrne:
I'd like to merge this now and we can continue iterating off `main` as necessary, WDYT?

AustinMroz:
I agree and have no blocking objections. There's a couple of tougher functional issues raised by Pablo I'm still iterating on, but everything here has been addressed.

The new browser test expectations workflow failed, so I've just been manually updating from test failures. This third commit should be sufficient to get tests passing, otherwise we may need to toggle the label to force the workflow to re-run

christian-byrne (approved):

PR #5831, opened 2025-09-28
designate `canvasOnly` on runtime-generated virtual widgets so they are hidden in Vue renderer

christian-byrne:
## Summary

Added `canvasOnly` flag to runtime-generated widgets to prevent Vue renderer from displaying them while keeping canvas functionality intact.

## Changes

- **What**: Added `canvasOnly` widget option to hide upload, webcam, and refresh widgets from Vue renderer

In the Canvas (LiteGraph) system, there was a small set of widgets with strictly defined components. There, if we wanted some unique or relatively complex behavior (like an upload butotn), we needed to create a separate widget that would be coupled to the original widget at runtime (and would not be serialized).

In the Vue renderer system, we can simply add flags to the inputSpec or widget options and conditionally render complex UI additions -- i.e., there is no need for the hard-to-maintain runtime widget associations. Expressing such things entirely in the view layer simplifies business logic related to graph state, as we no longer need to account for preserving the connections between runtime widgets and their special siblings -- we also do not need to worry about the implications for state serialization.

## Related

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/5798

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5831-designate-canvasOnly-on-runtime-generated-virtual-widgets-so-they-are-hidden-in-Vue-ren-27c6d73d365081fb8641feec010190df) by [Unito](https://www.unito.io)


src/extensions/core/widgetInputs.ts
```diff
@@ -217,6 +217,10 @@ export class PrimitiveNode extends LGraphNode {
     widgetName: string,
     recreating: boolean
   ) {
+    console.log('createWidget', inputData, node, widgetName, recreating)
```
DrJKL:
```suggestion
```

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -241,7 +241,7 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
   }
 
   /**
-   * Sets up widget callbacks for a node - now with reduced nesting
```
DrJKL, line 244: And 0 added sugars

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -1851,6 +1851,7 @@ export class LGraphNode
     callback: IBaseWidget['callback'] | string | null,
     options?: IWidgetOptions | string
   ): WidgetTypeMap[Type] | IBaseWidget {
+    console.log('addWidget', type, name, value, callback, options)
```
DrJKL: Or debug if you want to keep them?
```suggestion
```

DrJKL (dismissed):
Not in this PR, but this seems like it would help with the control after generate piece too.

christian-byrne:
> Not in this PR, but this seems like it would help with the control after generate piece too.

I did that in this PR as well

src/scripts/widgets.ts
```diff
@@ -134,7 +134,8 @@ export function addValueControlWidgets(
     function () {},
     {
       values: ['fixed', 'increment', 'decrement', 'randomize'],
-      serialize: false // Don't include this in prompt.
+      serialize: false, // Don't include this in prompt.
+      canvasOnly: true
```
christian-byrne, line 138: This is the `control_after_generate` widget

christian-byrne:
The [9 test failures](https://1080e84c.comfyui-playwright-chromium.pages.dev/#?q=s:failed) are in the Vue Nodes screenshot-based test cases.

In particular, from the `control_after_generate` widgets being removed from the KSamplers and the link bug with runtime widgets being coincidentally solved (because those widgets are removed from this PR).

browser_tests/tests/vueNodes/widgets/load/uploadWidgets.spec.ts-snapshots/vue-nodes-upload-widgets-chromium-linux.png
webfiltered, line 1: @christian-byrne Seems this test is comparing some randomised data - will pretty much always fail.

I don't think any kind of random data is wise in tests.  Pre-generated random array of data is great, but any random generation in a test feels like it's just _asking_ for flakiness.

christian-byrne: Ah, good catch. It's the cache-busting param being added to the `/view` URL construction.  I will fix, thank you for pointing it out.

webfiltered: No worries!  We sending proper HTTP headers for these from core at all?  Or just .. basic HTTP responses, hence random?

christian-byrne: To my recollection, we are not setting the headers, and I imagine aiohttp server has some automatic behavior for images, in addition to browser behavior. So for some reason, it was deemed better to let client choose whether to cache or not via the query param. I don't know whether that was due to convenience (not wanting to wait for PR on core) or for a real technical requirement/reason.

PR #5834, opened 2025-09-28
fix Vue node widgets should be in disabled state if their slots are connected with a link

### LINKED ISSUE #5692: The widget value is still editable or not hidden when converted to input.
https://github.com/user-attachments/assets/95468d24-8f74-4ac1-b557-4a6c84867add

Image

1. Can we also add an input dot style after the widget is automatically converted to an input? 
2. The widget value is still editable

christian-byrne:
## Summary

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/5692 by making widget link connection status trigger on change so Vue widgets with connected links could properly switch to the `disabled` state when they are implicitly converted to inputs. 

## Changes

- **What**: Added `node:slot-links:changed` event tracking and reactive slot data synchronization for Vue widgets

```mermaid
graph TD
    A[Widget Link Change] --> B[NodeInputSlot.link setter]
    B --> C{Is Widget Input?}
    C -->|Yes| D[Trigger slot-links:changed]
    C -->|No| E[End]
    D --> F[Graph Event Handler]
    F --> G[syncNodeSlotData]
    G --> H[Update Vue Reactive Data]
    H --> I[Widget Re-render]
    
    style I fill:#f9f9f9,stroke:#333,color:#000
```

## Review Focus

Widget reactivity performance with frequent link changes and event handler memory management in graph operations.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5834-fix-Vue-node-widgets-should-be-in-disabled-state-if-their-slots-are-connected-with-a-link-27c6d73d365081f6a6c3c1ddc3905c5e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -85,13 +85,28 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
 
     if (!nodeRef || !currentData) return
 
-    const refreshedData = extractVueNodeData(nodeRef)
+    // Only extract slot-related data instead of full node re-extraction
+    const slotMetadata = new Map<string, WidgetSlotMetadata>()
+
+    nodeRef.inputs?.forEach((input, index) => {
+      if (!input?.widget?.name) return
+      slotMetadata.set(input.widget.name, {
+        index,
+        linked: input.link != null
+      })
+    })
+
+    // Update only widgets with new slot metadata, keeping other widget data intact
+    const updatedWidgets = currentData.widgets?.map((widget) => {
+      const slotInfo = slotMetadata.get(widget.name)
+      return slotInfo ? { ...widget, slotMetadata: slotInfo } : widget
+    })
 
     vueNodeData.set(nodeId, {
       ...currentData,
-      widgets: refreshedData.widgets,
-      inputs: refreshedData.inputs,
-      outputs: refreshedData.outputs
+      widgets: updatedWidgets,
+      inputs: nodeRef.inputs ? [...nodeRef.inputs] : undefined,
+      outputs: nodeRef.outputs ? [...nodeRef.outputs] : undefined
     })
   }
 
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -13,7 +13,18 @@ import type { InputSpec } from '@/schemas/nodeDef/nodeDefSchemaV2'
 import { useNodeDefStore } from '@/stores/nodeDefStore'
 import type { WidgetValue } from '@/types/simplifiedWidget'
 
-import type { LGraph, LGraphNode } from '../../lib/litegraph/src/litegraph'
+import type {
+  LGraph,
+  LGraphNode,
+  LGraphTriggerAction,
+  LGraphTriggerParam
+} from '../../lib/litegraph/src/litegraph'
+import { NodeSlotType } from '../../lib/litegraph/src/types/globalEnums'
+
+export interface WidgetSlotMetadata {
+  index: number
+  linked: boolean
+}
 
 export interface SafeWidgetData {
   name: string
@@ -23,6 +34,7 @@ export interface SafeWidgetData {
   options?: Record<string, unknown>
   callback?: ((value: unknown) => void) | undefined
   spec?: InputSpec
+  slotMetadata?: WidgetSlotMetadata
 }
 
 export interface VueNodeData {
@@ -66,6 +78,22 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
   // Non-reactive storage for original LiteGraph nodes
   const nodeRefs = new Map<string, LGraphNode>()
 
+  const refreshNodeSlots = (nodeId: string) => {
+    const nodeRef = nodeRefs.get(nodeId)
+    const currentData = vueNodeData.get(nodeId)
+
+    if (!nodeRef || !currentData) return
+
+    const refreshedData = extractVueNodeData(nodeRef)
+
+    vueNodeData.set(nodeId, {
+      ...currentData,
+      widgets: refreshedData.widgets,
+      inputs: refreshedData.inputs,
+      outputs: refreshedData.outputs
+    })
+  }
+
   // Extract safe data from LiteGraph node for Vue consumption
   const extractVueNodeData = (node: LGraphNode): VueNodeData => {
     // Determine subgraph ID - null for root graph, string for subgraphs
@@ -74,6 +102,16 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
         ? String(node.graph.id)
         : null
     // Extract safe widget data
+    const slotMetadata = new Map<string, WidgetSlotMetadata>()
+
+    node.inputs?.forEach((input, index) => {
+      if (!input?.widget?.name) return
+      slotMetadata.set(input.widget.name, {
+        index,
+        linked: input.link != null
+      })
+    })
+
     const safeWidgets = node.widgets?.map((widget) => {
       try {
         // TODO: Use widget.getReactiveData() once TypeScript types are updated
@@ -90,6 +128,7 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
           value = widget.options.values[0]
         }
         const spec = nodeDefStore.getInputSpecForWidget(node, widget.name)
+        const slotInfo = slotMetadata.get(widget.name)
 
         return {
           name: widget.name,
@@ -98,7 +137,8 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
           label: widget.label,
           options: widget.options ? { ...widget.options } : undefined,
           callback: widget.callback,
-          spec
+          spec,
+          slotMetadata: slotInfo
         }
       } catch (error) {
         return {
@@ -405,29 +445,19 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
       handleNodeRemoved(node, originalOnNodeRemoved)
     }
 
-    // Listen for property change events from instrumented nodes
-    graph.onTrigger = (action: string, param: unknown) => {
-      if (
-        action === 'node:property:changed' &&
-        param &&
-        typeof param === 'object'
-      ) {
-        const event = param as {
-          nodeId: string | number
-          property: string
-          oldValue: unknown
-          newValue: unknown
-        }
-
-        const nodeId = String(event.nodeId)
+    const triggerHandlers: {
+      [K in LGraphTriggerAction]: (event: LGraphTriggerParam<K>) => void
+    } = {
+      'node:property:changed': (propertyEvent) => {
+        const nodeId = String(propertyEvent.nodeId)
         const currentData = vueNodeData.get(nodeId)
 
         if (currentData) {
-          switch (event.property) {
+          switch (propertyEvent.property) {
             case 'title':
               vueNodeData.set(nodeId, {
                 ...currentData,
-                title: String(event.newValue)
+                title: String(propertyEvent.newValue)
               })
               break
             case 'flags.collapsed':
@@ -435,7 +465,7 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
                 ...currentData,
                 flags: {
                   ...currentData.flags,
-                  collapsed: Boolean(event.newValue)
+                  collapsed: Boolean(propertyEvent.newValue)
                 }
               })
               break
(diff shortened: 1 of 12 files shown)


src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -113,7 +113,7 @@
         <!-- Widgets rendered at reduced+ detail -->
         <NodeWidgets
           v-if="nodeData.widgets?.length"
-          v-memo="[nodeData.widgets?.length]"
+          v-memo="[nodeData.widgets?.length, nodeData.inputs]"
```
DrJKL: What would happen if we just removed the [`v-memo`](https://vuejs.org/api/built-in-directives.html#v-memo) and let Vue handle it?
```suggestion
```

src/lib/litegraph/src/node/NodeInputSlot.ts
```diff
@@ -12,10 +12,11 @@ import { type IDrawOptions, NodeSlot } from '@/lib/litegraph/src/node/NodeSlot'
 import type { SubgraphInput } from '@/lib/litegraph/src/subgraph/SubgraphInput'
 import type { SubgraphOutput } from '@/lib/litegraph/src/subgraph/SubgraphOutput'
 import { isSubgraphInput } from '@/lib/litegraph/src/subgraph/subgraphUtils'
+import { NodeSlotType } from '@/lib/litegraph/src/types/globalEnums'
 import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets'
 
 export class NodeInputSlot extends NodeSlot implements INodeInputSlot {
-  link: LinkId | null
+  private _link: LinkId | null
```
DrJKL: Two points of personal preference for this pattern:
I usually like an `Internal` suffix, i.e. `linkInternal` to the underscore prefix. Feels more expressive and helps with IDEs' outlines if you sort fields lexicographically.
I also don't love `null` over `undefined` since it does usually entail adding nullish coalescing to pick the specific correct absent value (when otherwise the absent value is `undefined` in the default case)

src/lib/litegraph/src/node/NodeInputSlot.ts
```diff
@@ -41,11 +42,37 @@ export class NodeInputSlot extends NodeSlot implements INodeInputSlot {
...
+    const index = this.node.inputs.indexOf(this)
+    if (index === -1) return
+
+    graph.trigger('node:slot-links:changed', {
+      nodeId: this.node.id,
+      slotType: NodeSlotType.INPUT,
+      slotIndex: index,
+      connected: this._link != null,
+      linkId: this._link ?? null
+    })
```
DrJKL: This feels like something that we should be doing in a store, if not now, then soon. I don't like complex decisions being made in components.

src/lib/litegraph/src/node/NodeInputSlot.ts
```diff
@@ -41,11 +42,37 @@ export class NodeInputSlot extends NodeSlot implements INodeInputSlot {
...
+    if (!graph) return
+
+    const index = this.node.inputs.indexOf(this)
+    if (index === -1) return
+
+    graph.trigger('node:slot-links:changed', {
+      nodeId: this.node.id,
+      slotType: NodeSlotType.INPUT,
+      slotIndex: index,
+      connected: this._link != null,
```
DrJKL:
```suggestion
      connected: this.isConnected,
```

src/lib/litegraph/src/node/NodeInputSlot.ts
```diff
@@ -41,11 +42,37 @@ export class NodeInputSlot extends NodeSlot implements INodeInputSlot {
...
+      linkId: this._link ?? null
```
DrJKL:
```suggestion
      linkId: this.link
```

src/lib/litegraph/src/node/NodeInputSlot.ts
```diff
@@ -41,11 +42,37 @@ export class NodeInputSlot extends NodeSlot implements INodeInputSlot {
...
+    const graph = this.node.graph
+    if (!graph) return
+
+    const index = this.node.inputs.indexOf(this)
+    if (index === -1) return
+
+    graph.trigger('node:slot-links:changed', {
+      nodeId: this.node.id,
+      slotType: NodeSlotType.INPUT,
+      slotIndex: index,
```
DrJKL: option:
```suggestion
    const slotIndex = this.node.inputs.indexOf(this)
    if (slotIndex === -1) return

    graph.trigger('node:slot-links:changed', {
      nodeId: this.node.id,
      slotType: NodeSlotType.INPUT,
      slotIndex,
```

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -66,6 +66,17 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
   // Non-reactive storage for original LiteGraph nodes
   const nodeRefs = new Map<string, LGraphNode>()
 
+  const syncNodeSlotData = (nodeId: string, node: LGraphNode) => {
+    const currentData = vueNodeData.get(nodeId)
+    if (!currentData) return
+
+    vueNodeData.set(nodeId, {
+      ...currentData,
+      inputs: node.inputs ? [...node.inputs] : undefined,
```
DrJKL: Not for this PR, but what if we defined inputs and outputs as being required, so that a lack of them would just be an empty array? Would that simplify some of the logic?

DrJKL (dismissed):

webfiltered (commented):
I feel like this might be around 40K lines-too-many.  Give or take.  😁

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -66,6 +78,22 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
...
+    if (!nodeRef || !currentData) return
+
+    const refreshedData = extractVueNodeData(nodeRef)
+
+    vueNodeData.set(nodeId, {
+      ...currentData,
+      widgets: refreshedData.widgets,
+      inputs: refreshedData.inputs,
+      outputs: refreshedData.outputs
+    })
```
webfiltered:
```suggestion
    const { widgets, inputs, outputs } = extractVueNodeData(nodeRef)

    vueNodeData.set(nodeId, { ...currentData, widgets, inputs, outputs })
```

This is a sidegrade at best, but... I wanted to see if it would fit.  After I checked, the "work" was already done, so please take / discard as you like.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -444,63 +474,59 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
...
+          refreshNodeSlots(String(slotLinksEvent.nodeId))
         }
       }
+    }
 
-      // Call original trigger handler if it exists
-      if (originalOnTrigger) {
-        originalOnTrigger(action, param)
+    const isTriggerAction = (value: string): value is LGraphTriggerAction =>
+      Object.prototype.hasOwnProperty.call(triggerHandlers, value)
```
webfiltered:
```suggestion
      Object.hasOwn(triggerHandlers, value)
```

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -129,12 +130,21 @@ const processedWidgets = computed((): ProcessedWidget[] => {
 
     const vueComponent = getComponent(widget.type) || WidgetInputText
 
+    const slotMetadata = widget.slotMetadata
```
webfiltered, line 133:
```suggestion
    const { slotMetadata } = widget
```

nit

webfiltered (commented):
Most importantly: LGTM!

I'm still a little confused about why we're using the `LGraph` "trigger" system - the reason I mention that is just because it feels like we have added a few iterations of  complexity and workarounds.

If we need to revisit again, maybe we should go more nuclear on the fix, and remove any possibility of decade-old design decisions haunting us any further?

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
DrJKL: Could pull off options at the same time

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -3009,6 +3019,15 @@ export class LGraphNode
...
         // remove there
         input.link = null
+        if (input.widget) {
+          graph.trigger('node:slot-links:changed', {
+            nodeId: target.id,
+            slotType: NodeSlotType.INPUT,
+            slotIndex: link_info.target_slot,
+            connected: false,
+            linkId: link_info.id
+          })
```
DrJKL, line 3029: If this is the same in all three locations, how would you feel about extracting something like `maybeTriggerLinkChange(...)`?

christian-byrne:
> I'm still a little confused about why we're using the LGraph "trigger" system - the reason I mention that is just because it feels like we have added a few iterations of complexity and workarounds.
> 
> If we need to revisit again, maybe we should go more nuclear on the fix, and remove any possibility of decade-old design decisions haunting us any further?

I think everyone who reads this all wants to go nuclear on the fix, but it doesn't fully align with business goals to dedicate a lot of time to doing it. We will gradually replace the system with signals-based implementation.

christian-byrne:
Planning on making changes:

- Properly type events to not require any extra unneccesary assertions or awkward reflection, using standard patterns
- Add e2e tests for the disabled state on linked widgets

src/lib/litegraph/src/LGraph.ts
```diff
@@ -1180,8 +1191,33 @@ export class LGraph
...
+  trigger(action: string, param: unknown): void
   trigger(action: string, param: unknown) {
-    this.onTrigger?.(action, param)
+    // Convert to discriminated union format for typed handlers
+    if (
+      action === 'node:slot-links:changed' &&
+      param &&
+      typeof param === 'object'
+    ) {
+      this.onTrigger?.({ type: action, ...param } as LGraphTriggerEvent)
```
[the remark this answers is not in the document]
christian-byrne: That's pretty paranoid

src/lib/litegraph/src/LGraphNode.ts
christian-byrne: I think it would only be simplified if it were to be made a class instance property of some sort, which would start to go backwards a bit. If it's moved to a more pure helper, then it becomes even more code, I think. Don't see a way to simplify/refactor cleanly.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -66,6 +79,22 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
   // Non-reactive storage for original LiteGraph nodes
   const nodeRefs = new Map<string, LGraphNode>()
 
+  const refreshNodeSlots = (nodeId: string) => {
+    const nodeRef = nodeRefs.get(nodeId)
+    const currentData = vueNodeData.get(nodeId)
+
+    if (!nodeRef || !currentData) return
+
+    const refreshedData = extractVueNodeData(nodeRef)
```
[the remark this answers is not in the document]
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5834/commits/07de517caff3019ea1418f6085c9ab78ff9b614d

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -2852,7 +2852,17 @@ export class LGraphNode
     output.links ??= []
     output.links.push(link.id)
     // connect in input
-    inputNode.inputs[inputIndex].link = link.id
+    const targetInput = inputNode.inputs[inputIndex]
+    targetInput.link = link.id
+    if (targetInput.widget) {
+      graph.trigger('node:slot-links:changed', {
```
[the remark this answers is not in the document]
christian-byrne: See other comment regarding refactoring this.

src/lib/litegraph/src/types/graphTriggers.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+interface NodeSlotErrorsChangedEvent {
+  type: 'node:slot-errors:changed'
+  nodeId: string | number
+}
+
+interface NodeSlotLinksChangedEvent {
+  type: 'node:slot-links:changed'
+  nodeId: string | number
```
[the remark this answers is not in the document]
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5834/commits/ddbc895eff4b34f944d1e5562b166b8771e04ac6

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -444,63 +475,64 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
...
-              : undefined,
-            outputs: litegraphNode.outputs
-              ? [...litegraphNode.outputs]
-              : undefined
-          })
+      },
+      'node:slot-errors:changed': (slotErrorsEvent) => {
+        refreshNodeSlots(String(slotErrorsEvent.nodeId))
+      },
+      'node:slot-links:changed': (slotLinksEvent) => {
```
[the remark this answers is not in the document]
christian-byrne: No, some things should not be debounced.

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -129,12 +130,21 @@ const processedWidgets = computed((): ProcessedWidget[] => {
...
+
+    let widgetOptions = widget.options
+    if (slotMetadata?.linked) {
```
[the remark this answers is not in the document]
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5834/commits/eff84313dd841127218ae75d80ff18c860729e06

src/lib/litegraph/src/LGraph.ts
```diff
@@ -1180,8 +1191,33 @@ export class LGraph
   }
 
   // ********** GLOBALS *****************
+  trigger<A extends LGraphTriggerAction>(
+    action: A,
+    param: LGraphTriggerParam<A>
+  ): void
+  trigger(action: string, param: unknown): void
   trigger(action: string, param: unknown) {
```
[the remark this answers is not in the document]
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5834/commits/3e542a0d0b4cfca359b6e5868dc4fb124f55f2dd

PR #5842, opened 2025-09-28
fix progress state on Vue nodes in subgraphs

christian-byrne:
## Summary

Fixes two errors with subgraph progress states:

1. Nodes inside subgraphs were not having progress state shown
2. Subgraph nodes (outer representation) themselves did not have a visible progress state

1 is fixed by using locator IDs instead of local node IDs.

2 is fixed by ensuring the subgraph title button does not wrap to a newline and thus block the progress bar under the node header.

## Changes

- **What**: Updated `useNodeExecutionState` composable to use `nodeLocatorId` for tracking execution state across subgraph boundaries
- **What**: Modified NodeHeader layout to fix subgraph enter button positioning with proper flexbox gap

## Review Focus

Execution state tracking accuracy for nested subgraph nodes and NodeHeader layout consistency across different node types.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5842-fix-progress-state-on-Vue-nodes-in-subgraphs-27c6d73d365081cb8335c8bb5dbd74f7) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/execution/useNodeExecutionState.ts
```diff
@@ -9,27 +9,27 @@ import { useExecutionStore } from '@/stores/executionStore'
...
+    return id ? nodeLocationProgressStates.value[id] : undefined
   })
 
+  const executing = computed(() => progressState.value?.state === 'running')
+
   const progress = computed(() => {
-    const state = nodeProgressStates.value[nodeId]
-    return state?.max > 0 ? state.value / state.max : undefined
+    const state = progressState.value
+    return state && state.max > 0 ? state.value / state.max : undefined
```
DrJKL, line 30: Why not use the null-safe operator?
`undefined > 0` should resolve to `false`

Maybe a guard would be clearer than a ternary here?

DrJKL (dismissed):

src/renderer/extensions/vueNodes/execution/useNodeExecutionState.ts
webfiltered: > `undefined > 0` should resolve to `false`

It's not actually valid typescript, though - `>` is a numeric operator.

src/renderer/extensions/vueNodes/components/NodeHeader.vue
```diff
@@ -9,7 +9,7 @@
     :data-testid="`node-header-${nodeData?.id || ''}`"
     @dblclick="handleDoubleClick"
   >
-    <div class="flex items-center justify-between relative">
+    <div class="flex items-center justify-between gap-2 relative">
```
christian-byrne: This is causing tests to fail, as we now have a gap between title and collapse button in node headers. But, in fact, [in the design there is a 10px gap](https://www.figma.com/design/31uH3r4x3xbIctuRWYW6NM/V3---Nodes?node-id=6185-7855&m=dev) (tailwind 2.5)

src/renderer/extensions/vueNodes/execution/useNodeExecutionState.ts
DrJKL: Good point. Yes.

DrJKL (dismissed):

PR #5810, opened 2025-09-27
Extract Zod schemas into shared workspace package

webfiltered:
## Summary

Migrate Zod schemas to a shared `@comfyorg/schemas` package to enable reuse across workspace packages while maintaining backward compatibility through re-exports.

## Changes

- **What**: Created new `@comfyorg/schemas` workspace package and migrated all Zod schemas from `src/schemas/` to the new package
- **Breaking**: None - all existing imports work through re-exports

## Review Focus

The migration uses re-exports to maintain 100% backward compatibility - no import changes needed in consuming code. The git history is preserved through `git mv` operations.

The diff of this PR is massive, as the files were moved then replaced by re-export files (tracking issue will be created to remove these).  The links below show the actual changes in the PR much more cleanly.

- [View package setup and file moves](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5810/files/240255d9e)
- [View re-exports and cleanup](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5810/files/240255d9e..5f94873ae)

PR #5843, opened 2025-09-29
Extract shared utilities into workspace package

webfiltered:
## Summary

Extracts shared formatting and network utilities into dedicated workspace package.

## Changes

- **What**: Created `@comfyorg/shared-frontend-utils` package containing formatUtil and networkUtil
- **Breaking**: None - utilities remain accessible via path aliases in `tsconfig`

Split `createAnnotatedPath` and `electronMirrorCheck` out and left in frontend, due to their tightly-coupled nature.  See [discussion on this PR](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5843#issuecomment-3344724727).

## Review Focus

Package naming - is `@comfyorg/shared-frontend-utils` appropriate for these utilities?  Please think of something better, then tell me about it.

(Edit: No more two-link views required - "discovered" a neat way to squash merge by using different file names and tsconfig aliases.. then just aliasing to the package files direct - just requires a root level tsconfig.  Apparently, this is in the Nx docs.  Spoilsports.)


DrJKL (commented):
Utils is a little generic, yeah.
It feels like it would be hard to define a bright line inclusion rule for it.

Is there a way to hint to git that the files are a move/copy?

What's up with createAnnotatedPath?

webfiltered:
> Utils is a little generic, yeah.

Yeah super not happy with it.  Thought of a bunch of really bad alternatives.  Didn't want to prefix with `frontend-`, but may need to.  How do you feel about `@comfyorg/generic-frontend-pure-functions-and-stuff`?

> Is there a way to hint to git that the files are a move/copy?

So it displays on the "Files changed" tab (without viewing in ranges)?  Not without removing the shims / adding them in a follow-up PR.

I explicitly `commit --no-verify` the pure `git mv` though.  Links to commit ranges are in PR body~

webfiltered:
> What's up with createAnnotatedPath?

It uses the following dependency chain: ResultItem -> [split schemas out to package](https://github.com/Comfy-Org/ComfyUI_frontend/issues/5810) -> i18n -> [nope](https://github.com/Comfy-Org/ComfyUI_frontend/issues/5811).

Path of least resistance was to leave it behind.

webfiltered:
Test failure: confirmed unrelated.

tsconfig.json
```diff
@@ -22,7 +22,10 @@
     "verbatimModuleSyntax": true,
     "baseUrl": ".",
     "paths": {
-      "@/*": ["src/*"]
+      "@/*": ["src/*"],
+      "@/utils/envUtil": ["src/utils/envUtil.shim.ts"],
+      "@/utils/formatUtil": ["src/utils/formatUtil.shim.ts"],
+      "@/utils/networkUtil": ["src/utils/networkUtil.shim.ts"]
```
DrJKL: Does this work?
```suggestion
      "@/*": ["src/*"],
      "@/utils/envUtil": ["packages/shared-frontend-utils/src/envUtil.ts"],
      "@/utils/formatUtil": ["packages/shared-frontend-utils/src/formatUtil.ts"],
      "@/utils/networkUtil": ["packages/shared-frontend-utils/src/networkUtil.ts"]
```

webfiltered: Yeah, I refactored down to this after the last commit.  But the build failure is unrelated, and more painful.

I need to reduce the scope further; will either close PR or force push something else later.

christian-byrne (approved):

DrJKL (approved):

PR #5851, opened 2025-09-29
[chore] make experiment asset api setting hidden

arjansingh:
## Summary

This way no one can accidentally enable it when they are messing around.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5851-chore-make-experiment-asset-api-setting-hidden-27d6d73d36508113a6c1e41a764325f3) by [Unito](https://www.unito.io)


AustinMroz (approved):
I wish I had a better answer for the "how", but the code here is beneficial.

Even the ugly edge case of "how does a user disable a setting that doesn't exist?"  is, at worst, more reports to track from if there *is* something strange happening here.

christian-byrne (approved):

PR #5853, opened 2025-09-30
pass nodeId to widget component

jtydhr88:
## Summary

pass nodeId to widget component, which is a prerequirist for https://github.com/Comfy-Org/ComfyUI_frontend/pull/5765

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5853-pass-nodeData-to-widget-component-27e6d73d3650811e9e48ceb7c3a00545) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/NodeWidgets.vue
+++ b/src/renderer/extensions/vueNodes/components/NodeWidgets.vue
@@ -45,6 +45,7 @@
         :widget="widget.simplified"
         :model-value="widget.value"
         :readonly="readonly"
+        :node-data="nodeData"
         class="flex-1"
         @update:model-value="widget.updateHandler"
       />
--- a/src/renderer/extensions/vueNodes/components/NodeWidgets.vue
+++ b/src/renderer/extensions/vueNodes/components/NodeWidgets.vue
@@ -45,7 +45,7 @@
         :widget="widget.simplified"
         :model-value="widget.value"
         :readonly="readonly"
-        :node-data="nodeData"
+        :node-id="nodeData?.id != null ? String(nodeData.id) : ''"
         class="flex-1"
         @update:model-value="widget.updateHandler"
       />


DrJKL:
I was actually hoping we could move in the other direction. Right now we're passing large objects around through props instead of just the pieces needed by each component. It makes it tricky to draw clear lines of responsibility and might be part of the rendering performance issues.

christian-byrne:
@jtydhr88 Could you give a general explanation of why 3D nodes need the nodedata in order to function properly? Once we all fully understand, we could discuss a good solution.

From my recollection, 3D nodes are storing state on `properties` object/property of LGraphNodes, is that correct? And this is the type of data we want to be serialized so it can persist after reload - so it can't necessarily go straight into the Vue components.

What other details or requirements are there?

jtydhr88:
@christian-byrne you are right, the main reason I need this is to store properties settings on Node level, previously I could get node directly but now I need do in the way https://github.com/Comfy-Org/ComfyUI_frontend/pull/5765/files#diff-f421a89c7c66125361b3049ff1cbe1815b3a6a8c24f5ed4e54795c5148d20f9bL218
Maybe we could have a better solution

jtydhr88:
@DrJKL for my case, in fact I only need nodeId

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -45,6 +45,7 @@
         :widget="widget.simplified"
         :model-value="widget.value"
         :readonly="readonly"
+        :node-id="nodeData?.id != null ? String(nodeData.id) : ''"
```
webfiltered, line 48: This pattern is everywhere and doesn't make sense - `nodeId` is defined as optional (string | undefined), yet we're forcing it to empty string for some reason?  Is that actually valid?

webfiltered (commented):
Comment is almost unrelated to the PR; this PR follows the pattern and is good.

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
jtydhr88: yes, I just copy this line from above code in the same file, and follow the same pattern what I see...

DrJKL: I agree with @webfiltered that we should remove the check and conversion later.

DrJKL (approved):

christian-byrne (approved):

PR #5856, opened 2025-09-30
Fix/vue nodes tooltips

simula-r:
## Summary

Make vue node tooltips support color modes and hide when user drags node or zooms in or out. 

IMPORTANT: this approach is non ideal. But for sake of time its a valid escape hatch. The reason for this approach is because of how primevue exposes tooltips via a directive and the inability to modify its state / conditionally render it directly. Instead we must get the actual tooltip element from the dom and trigger a mouseleave event. 

Follow up PR would be refactoring the entire useNodeTooltip.ts composable and tooltip system from primevue to something like RekaUI which is a much more declarative approach. To start we could just refactor Vue nodes themselves to RekaUI tooltips and then later migrate the other instances.

## Changes

- **What**:  hide tooltips on drag and zoom. Use color themes

## Review Focus

Tooltip hover states after dragging. 
Fixes: 
- https://www.notion.so/comfy-org/Bug-Fix-tooltip-appearing-when-clicking-and-dragging-Vue-node-header-2796d73d365081f692fdde30c51580e1?source=copy_link
- https://www.notion.so/comfy-org/Bug-Fix-chevron-color-customization-in-tooltip-pseudo-element-27b6d73d3650814abc25fc0eefda1648?source=copy_link
- https://www.notion.so/comfy-org/Bug-Bug-Tooltip-positioning-breaks-with-zoom-due-to-missing-appendTo-prop-27b6d73d36508186858bd1cb4336ec03?source=copy_link

https://github.com/user-attachments/assets/06c70f6c-bf64-4b6d-8e6f-739995549940

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5856-Fix-vue-nodes-tooltips-27e6d73d36508177a77bf7f2f27344f4) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
+++ b/src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
@@ -1,23 +1,33 @@
 import type { TooltipDirectivePassThroughOptions } from 'primevue'
-import { type MaybeRef, type Ref, computed, unref, watch } from 'vue'
+import { type MaybeRef, type Ref, computed, ref, unref } from 'vue'
 
 import type { SafeWidgetData } from '@/composables/graph/useGraphNodeManager'
 import { st } from '@/i18n'
 import { useSettingStore } from '@/platform/settings/settingStore'
-import { layoutStore } from '@/renderer/core/layout/store/layoutStore'
 import { useNodeDefStore } from '@/stores/nodeDefStore'
 import { normalizeI18nKey } from '@/utils/formatUtil'
 import { cn } from '@/utils/tailwindUtil'
 
 /**
- * Hide all visible tooltips by finding them via class and dispatching mouseleave
- * Tooltips only exist in DOM when visible, so we can query them directly
+ * Hide all visible tooltips by dispatching mouseleave events
+ *
+ *
+ * IMPORTANT: this escape is needed for many reason due to primevue's directive tooltip system.
+ * We cannot use PT to conditonally render the tooltips because the entire PT object only run
+ * once during the intialization of the directive not every mount/unmount.
+ * Once the directive is constructed its no longer reactive in the traditional sense.
+ * We have to use something non destructive like mouseevents to dismiss the tooltip.
+ *
+ * TODO: use a better tooltip component like RekaUI for vue nodes specifically.
  */
+
+const tooltipsTemporarilyDisabled = ref(false)
+
 const hideTooltipsGlobally = () => {
-  // Get all visible tooltip elements (PrimeVue uses .p-tooltip class)
+  // Get all visible tooltip elements
   const tooltips = document.querySelectorAll('.p-tooltip')
 
-  // Early return if no tooltips are visible - avoid unnecessary work
+  // Early return if no tooltips are visible
   if (tooltips.length === 0) return
 
   tooltips.forEach((tooltipEl) => {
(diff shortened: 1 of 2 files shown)


DrJKL:
https://github.com/primefaces/primevue/issues/6943

src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
```diff
@@ -1,11 +1,77 @@
...
+  if (tooltips.length === 0) return
+
+  tooltips.forEach((tooltipEl) => {
+    const tooltipId = tooltipEl.id
+    if (!tooltipId) return
+
+    // Find the target element that owns this tooltip
+    const targetElements = document.querySelectorAll('[data-pd-tooltip="true"]')
+    for (const targetEl of targetElements) {
+      if ((targetEl as any).$_ptooltipId === tooltipId) {
```
DrJKL, line 40: Can you avoid `any`?

src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
```diff
@@ -1,11 +1,77 @@
...
+}
+
+// Global tooltip hiding system
+const globalTooltipState = { listenersSetup: false }
+
+function setupGlobalTooltipHiding() {
+  if (globalTooltipState.listenersSetup) return
+
+  document.addEventListener('pointerdown', hideTooltipsGlobally)
+  document.addEventListener('pointerup', handlePointerUp)
```
DrJKL, line 67: Could you use [`useEventListener`](https://vueuse.org/core/useEventListener/)?

christian-byrne (approved):

src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
DrJKL: This should also allow you to get rid of the extra module level tracking for the listeners.

src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
```diff
@@ -1,11 +1,77 @@
...
+  window.addEventListener('wheel', hideTooltipsGlobally, {
+    capture: true, //Need this to bypass the event layer from Litegraph
+    passive: true
+  })
```
DrJKL, line 71: This doesn't break anything?

src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
christian-byrne: I think this is supposed to be just a single pair (so never more than 2 globally) of listeners that would persist from the moment any Vue node is made and then for the rest of the app life, so cleanup is not as important. The module level tracking seems mostly just lazy implicit singleton type of initialization pattern, which seems to work for now.

src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
christian-byrne: I don't think it will break anything AFAICT, it just closes tooltips on any wheel. It could potential benefit from being throttled though - e.g., using `throttle` from `es-toolkit` so we don't `querySelectorAll` on each wheel event.

PR #5863, opened 2025-09-30
Select Vue Nodes After Drag

Myestery:
This pull request refactors the node selection logic in the Vue nodes event handler composable to simplify the function signature and improve single vs. multi-selection behavior. The main change is the removal of the `wasDragging` parameter from the `handleNodeSelect` function, with selection logic now determined by the current selection state. Related test code is updated to match the new function signature.

**Node selection logic improvements:**

* Refactored the `handleNodeSelect` function in `useNodeEventHandlersIndividual` to remove the `wasDragging` parameter, making the function signature simpler and relying on selection state to handle single vs. multi-selection.
* Updated the selection logic to check if multiple nodes are already selected using `isLGraphNode`, and only perform single selection if not.

**Code and test updates:**

* Updated all calls to `handleNodeSelect` in the composable to remove the `wasDragging` argument, ensuring consistent usage throughout the codebase. [[1]](diffhunk://#diff-8d3820a1ca9c569bce00671fdd6290af81315ae11b8f3d6f29a5a9d30379d084L125-R123) [[2]](diffhunk://#diff-8d3820a1ca9c569bce00671fdd6290af81315ae11b8f3d6f29a5a9d30379d084L146-R144) [[3]](diffhunk://#diff-8d3820a1ca9c569bce00671fdd6290af81315ae11b8f3d6f29a5a9d30379d084L173-R171)
* Updated all related test cases to use the new `handleNodeSelect` signature without the third parameter. [[1]](diffhunk://#diff-89bfc2a05201c6ff7116578efa45f96097594eb346f18446c70aa7125ab1811aL105-R105) [[2]](diffhunk://#diff-89bfc2a05201c6ff7116578efa45f96097594eb346f18446c70aa7125ab1811aL125-R125) [[3]](diffhunk://#diff-89bfc2a05201c6ff7116578efa45f96097594eb346f18446c70aa7125ab1811aL144-R144) [[4]](diffhunk://#diff-89bfc2a05201c6ff7116578efa45f96097594eb346f18446c70aa7125ab1811aL162-R162) [[5]](diffhunk://#diff-89bfc2a05201c6ff7116578efa45f96097594eb346f18446c70aa7125ab1811aL174-R174) [[6]](diffhunk://#diff-89bfc2a05201c6ff7116578efa45f96097594eb346f18446c70aa7125ab1811aL187-R187)

**Utility import:**

* Added an import for `isLGraphNode` from `@/utils/litegraphUtil` to support the updated selection logic.## Summary

https://github.com/user-attachments/assets/71e856d3-afc2-497d-826e-5b485066e7fe


src/renderer/extensions/vueNodes/composables/useNodeEventHandlers.ts
```diff
@@ -48,12 +45,13 @@ function useNodeEventHandlersIndividual() {
         canvasStore.canvas.select(node)
       }
     } else {
-      // If it wasn't a drag: single-select the node
-      if (!wasDragging) {
+      const selectedMultipleNodes =
+        canvasStore.selectedItems.filter((n) => isLGraphNode(n)).length > 1
```
christian-byrne: nit: we can optimize this by just doing a normal `for` loop and returning  early if we hit 2, then move it to existing utils file or into the module scope.

DrJKL: This feels a little weird to me. Can we change the selection logic in useNodePointerInteractions instead? I think we'll have to change the interface there anyway since that's where the dragging logic lives.

I would expect the selection to be apparent _as_ I click and drag, not after I release the click.

christian-byrne (commented):
It should be fairly comfortable to add a test case here, following the existing patterns:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/35ddb19a6855d333666ea1234ab684d7003fc3cf/browser_tests/tests/vueNodes/interactions/node/select.spec.ts#L23-L38

src/renderer/extensions/vueNodes/composables/useNodeEventHandlers.ts
christian-byrne: I agree - I think we should try to match litegraph behavior exactly (unless it's extremely difficult to do).

DrJKL: Long term if we're doing this in multiple places, it might be a candidate for a computed value in the store. Something like `canvasStore.selectedNodesCount`? (Depending on how expensive the filtering ends up being)

Myestery: Last time I checked, canvas.selectedItems is not reactive

christian-byrne:
The tests might need to be updated

Myestery:
https://github.com/user-attachments/assets/278953a1-ea6f-492e-b425-8dba67885a5d

Myestery:
> Screen.Recording.2025-10-03.at.06.43.01.mov

Added support for dragging with ctrl and selecting nodes at the same time
Ie node1 is selected, hold ctrl and drag node2
everything drags just like litegraph

christian-byrne (changes requested):
For these tests, we should avoid using litegraph test utils/fixtures (since we are manipulating Vue nodes) and also try to accomplish changes by emulating user behaviors rather than calling methods on code objects. For testing code, we can rely moreso on unit tests; for e2e tests we should test the system using more of a black-box approach - which provides a different breadth of coverage and ensures things work from the most important perspective (user), among other reasons. This means:

1. Not using `NodeReference` to get nodes but rather just finding the nodes by using locators that query for visible title text (as a user would do)
2. Not calling `LGraphNode.pin` to pin but rather just selecting the node and pressing the pin hotkey (as a user would do).

To see an example of both, check [here](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/browser_tests/tests/vueNodes/nodeStates/pin.spec.ts).

christian-byrne (dismissed):
LGTM

christian-byrne:
I will fix the conflict

Myestery:
> I will fix the conflict

Already fixed locally

christian-byrne:
I will fix the "Update Playwright Expectations" on `main`, then trigger again afterwards.

christian-byrne (approved):
LGTM!

PR #5869, opened 2025-09-30
Cleanup: YAGNI readonly props, private swap on ComfyApp, Canvas resize events simplification, v-memos on individual instances

DrJKL:
## Summary

Assorted cleanup opportunities found while working through the Vue node rendering logic cleanup.

## Review Focus

Am I wrong that the readonly logic was never actually executing because it was defined as False in GraphCanvas when making each LGraphNode?

Is there an edge case or some other reason that the ResizeObserver wouldn't work as a single signal to resize the canvas?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5869-Cleanup-YAGNI-readonly-props-private-swap-on-ComfyApp-Canvas-resize-events-simplificat-27e6d73d3650811ba1dcf29e8d43091e) by [Unito](https://www.unito.io)


webfiltered (commented):
If Vue nodes were given a "readonly" flag (via LLM?) simply because LiteGraph had one, then yes, _please_ delete that before it goes live.

So much in the handlers could have been simplified if we had just deleted it (I actually did on my SVG branch).  We simply don't use LiteGraph readonly mode.  It _was_ used by an extension, afaik, but I don't recall the feature being popular.

webfiltered (commented):
This PR now does so many different things, the title literally wraps over two lines in full screen mode on a 16" laptop.  >.>

Overall I'm wondering about the same thing as another PR - why are we removing `#private` from ComfyApp?  That object is exposed as `window.app` - it's been a constant source of friction with custom nodes using and overriding functions.

Some large functions were split into small functions, without changing the public API, by making them `#private`.  Marking them `#private` wasn't for our internal code at all.

If this has already been discussed, let me know - I'll stop raising it on all your PRs.  😁

christian-byrne (dismissed):
LGTM

christian-byrne:
We would just like to get away from js private since it interferes so much with Vue in ways that are difficult to forsee - using #private as a tool self-enforces a constraint on our ability to use the frameworks we want. So, I feel we should try to solve the problems it was solving (limiting custom node access) through the larger intiative of better defining the public API, and then at the same time switch to TS private to maintain the internal contracts. Then we can have the best of both worlds.

arjansingh (dismissed):

christian-byrne (dismissed):

christian-byrne (approved):
LGTM

PR #5870, opened 2025-09-30
Fix/vue nodes video

simula-r:
## Summary

Fix the video preview widget and associated dropdown to load and select videos.  

Fixes:
- https://www.notion.so/comfy-org/Video-thumbnails-not-being-used-in-asset-explorer-dialog-27e6d73d365080ec8a3ee7c7ec413657?source=copy_link
- https://www.notion.so/comfy-org/Image-Video-upload-dialog-doesnt-set-mime-type-27e6d73d365080c5bffdf08842855ba0?source=copy_link
- https://www.notion.so/comfy-org/Video-Previews-are-not-displayed-2756d73d365080b2bfb9e0004e9d784d?source=copy_link
- https://www.notion.so/comfy-org/Cannot-load-video-in-Load-Video-node-2756d73d365080009c21d3a67add96c4?source=copy_link

https://github.com/user-attachments/assets/b71dbecb-c9a7-4feb-83a3-c3e044a9c93c

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5870-Fix-vue-nodes-video-27e6d73d36508182b44bef8e90ef4018) by [Unito](https://www.unito.io)


christian-byrne (commented):
In `WidgetSelectDropdown`, we have the prop `widget` which has a `spec` field which will have `image_upload` or `video_upload`.

We can pass the props down the line:

`WidgetSelectDropdown` -> `FormDropdown` -> `FormDropdownMenu` -> `FormDropdownMenuItem`

(or, use `provide`/`inject`).

Then we may no longer need the `getMediaKindFromUrl` logic. WDYT?

src/types/mediaTypes.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Media type constants and utilities
+ * Single source of truth for supported media formats
+ */
+
+type MediaKind = 'image' | 'video' | 'audio' | 'unknown'
+
+const MEDIA_MIME_TYPES = {
+  image: ['image/png', 'image/jpeg', 'image/webp', 'image/gif'] as const,
+  video: ['video/mp4', 'video/webm'] as const,
+  audio: ['audio/mpeg', 'audio/wav', 'audio/ogg', 'audio/flac'] as const
```
christian-byrne: Our backend uses `Pillow`, `av`, `torchaudio` which all are capable of handling a very wide range of formats. If you look at the Python nodes, we simply filter the input folder by the content type ([example with LoadImage](https://github.com/comfyanonymous/ComfyUI/blob/e4f99b479a19730bea890567129f4032b4dd4787/nodes.py#L1630-L1635)). To mirror that behavior, we could just use `accept: "image/*"` and so on.

Of course, there are some things that are not supported (e.g., avif if the user hasn't updated PIL, mkv on Mac, etc.) but I prefer to be permissive and hope that the libraries we use can handle the majority of file types.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -31,6 +32,12 @@ const props = defineProps<{
   uploadFolder?: ResultItemType
 }>()
 
+// Provide assetKind to descendant components
+provide(
+  'assetKind',
```
christian-byrne: nit: You can [type the injection key like this](https://antfu.me/posts/typed-provide-and-inject-in-vue), which is what we usually do.

LittleSound (dismissed):

LittleSound:
LGTM

christian-byrne (approved):

PR #5872, opened 2025-10-01
Chores: Updates for Asset Services

arjansingh:
## Changes

1. Updates schema to match new API
2. Adds additional relevant models to the registry

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5872-Chores-Updates-for-Asset-Services-27f6d73d36508117b89fd473f1a7090d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/modelToNodeStore.ts
+++ b/src/stores/modelToNodeStore.ts
@@ -131,10 +131,24 @@ export const useModelToNodeStore = defineStore('modelToNode', () => {
     quickRegister('loras', 'LoraLoaderModelOnly', 'lora_name')
     quickRegister('vae', 'VAELoader', 'vae_name')
     quickRegister('controlnet', 'ControlNetLoader', 'control_net_name')
-    quickRegister('unet', 'UNETLoader', 'unet_name')
+    quickRegister('diffusion_models', 'UNETLoader', 'unet_name')
     quickRegister('upscale_models', 'UpscaleModelLoader', 'model_name')
-    quickRegister('style_models', 'StyleModelLoader', 'style_model')
+    quickRegister('style_models', 'StyleModelLoader', 'style_model_name')
     quickRegister('gligen', 'GLIGENLoader', 'gligen_name')
+    quickRegister('clip_vision', 'CLIPVisionLoader', 'clip_name')
+    quickRegister('text_encoders', 'CLIPLoader', 'clip_name')
+    quickRegister('audio_encoders', 'AudioEncoderLoader', 'audio_encoder_name')
+    quickRegister('model_patches', 'ModelPatchLoader', 'name')
+    quickRegister(
+      'animatediff_models',
+      'ADE_LoadAnimateDiffModel',
+      'model_name'
+    )
+    quickRegister(
+      'animatediff_motion_lora',
+      'ADE_AnimateDiffLoRALoader',
+      'name'
+    )
   }
 
   return {
--- a/src/stores/modelToNodeStore.ts
+++ b/src/stores/modelToNodeStore.ts
@@ -101,6 +101,7 @@ export const useModelToNodeStore = defineStore('modelToNode', () => {
     nodeProvider: ModelNodeProvider
   ) {
     registerDefaults()
+    if (!nodeProvider.nodeDef) return
     if (!modelToNodeMap.value[modelType]) {
       modelToNodeMap.value[modelType] = []
     }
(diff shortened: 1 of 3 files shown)


DrJKL (dismissed):

DrJKL:
Assuming you'll fix the tests 😊

arjansingh:
> Assuming you'll fix the tests 😊

lol what... none of these changes are visual 😂

christian-byrne:
Please try running the code with the default workflow and/or take a look at the failing test reports from the links in the above comments.

arjansingh:
> Please try running the code with the default workflow and/or take a look at the failing test reports from the links in the above comments.

yes was debugging this locally and found issues. fixes incoming.

src/stores/modelToNodeStore.ts
```diff
@@ -131,10 +134,24 @@ export const useModelToNodeStore = defineStore('modelToNode', () => {
...
+    quickRegister(
+      'animatediff_models',
+      'ADE_LoadAnimateDiffModel',
+      'model_name'
+    )
+    quickRegister(
+      'animatediff_motion_lora',
+      'ADE_AnimateDiffLoRALoader',
+      'name'
+    )
```
christian-byrne, line 155: Won't this create ModelProvider entries with `nodeDef` being undefined which violates the type? If the types are not accurate and there are corrupted entries, then the bug that caused the tests to fail will inevitably resurface

christian-byrne (changes requested):

src/stores/modelToNodeStore.ts
arjansingh: yes so i added safe guard checks and tests to make sure nodes like these get ignored.

guards:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/5872/files#diff-2411edd38323e1b314c82bbf0fa373d34cc8e9fd160488c3196b064a92465ff6R32

https://github.com/Comfy-Org/ComfyUI_frontend/pull/5872/files#diff-2411edd38323e1b314c82bbf0fa373d34cc8e9fd160488c3196b064a92465ff6R42-R43

additional tests:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/5872/files#diff-812de0cbceb659809ccdc7ea1edd8ed0a71027e89ec64a7d081fde0fa08b99c4R267-R279

Or fundamentally, do you want us to rethink how we do our permission check?

Right now it's:

```ts
      // Must be an approved input name
      WHITELISTED_INPUTS.has(inputName) &&
      // Must be a registered node type
      useModelToNodeStore().getRegisteredNodeTypes().has(nodeType)
```

https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/platform/assets/services/assetService.ts#L115-L132

christian-byrne: That seems good. If you filter non-existent nodes, it avoids violating the type and having to filter on every downstream computation/access.

arjansingh: exactly!

christian-byrne: Right now it is creating corrupted state which violates the `ModelNodeProvider` class type.

image

arjansingh: let me double check locally then because that's exactly what my unit test cases (and fixes) should be preventing.

src/stores/modelToNodeStore.ts
```diff
@@ -98,6 +101,7 @@ export const useModelToNodeStore = defineStore('modelToNode', () => {
     nodeProvider: ModelNodeProvider
   ) {
     registerDefaults()
+    if (!nodeProvider.nodeDef) return
```
arjansingh, line 104: Image

src/stores/modelToNodeStore.ts
arjansingh: @christian-byrne okay i added on more check to make sure keys don't even get created:

https://github.com/Comfy-Org/ComfyUI_frontend/pull/5872/files#r2400274212

src/stores/modelToNodeStore.ts
arjansingh: Expanded
Image

tests-ui/tests/store/modelToNodeStore.test.ts
```diff
@@ -173,6 +186,22 @@ describe('useModelToNodeStore', () => {
   })
 
   describe('registerNodeProvider', () => {
+    it('should not register provider when nodeDef is undefined', () => {
```
arjansingh, line 189: Test case to cover the scenario of the user not having a node for the model on their system.

christian-byrne (approved):
LGTM!

PR #5877, opened 2025-10-01
Cleanup/Perf: Float32Array/Float64Array removal

DrJKL:
## Summary

Redoing https://github.com/Comfy-Org/ComfyUI_frontend/pull/5567, without the link rendering changes.

## Changes

- **What**: Standardizing the Point/Size/Rect logic around numeric tuples instead of typed arrays.

## Review Focus

Cutting here and going to continue in a second PR.

Do the simpler types make sense?
Do we want to keep the behavior of Rectangle as it is now?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5877-WIP-Float32Array-Float64Array-removal-27f6d73d36508169a39eff1e4a87a61c) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -413,7 +412,7 @@ export class LGraphNode
   }
 
   /** @inheritdoc {@link renderArea} */
-  #renderArea: Float32Array = new Float32Array(4)
+  #renderArea: Rectangle = new Rectangle()
```
webfiltered:
```suggestion
  #renderArea = new Rectangle()
```

nit - Consistency

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -5204,8 +5206,8 @@ export class LGraphCanvas
 
     // clip if required (mask)
     const shape = node._shape || RenderShape.BOX
-    const size = LGraphCanvas.#temp_vec2
-    size.set(node.renderingSize)
+    const size = LGraphCanvas.temp_vec2
+    size.splice(0, 2, ...node.renderingSize)
```
webfiltered: This feels like an incredibly complex way to force two number assignments fit on one line.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -236,11 +238,11 @@ export class LGraphCanvas
...
-  static #temp = new Float32Array(4)
-  static #temp_vec2 = new Float32Array(2)
-  static #tmp_area = new Float32Array(4)
-  static #margin_area = new Float32Array(4)
-  static #link_bounding = new Float32Array(4)
+  private static temp = new Rectangle()
+  private static temp_vec2: Point = [0, 0]
+  private static tmp_area = new Rectangle()
+  private static margin_area = new Rectangle()
+  private static link_bounding = new Rectangle()
```
webfiltered: This is now full-circle from vanilla litegraph.js, it seems.

If you really can't stand #private fields, please just move them to module scope.  These never have been, and never should never be accessible at runtime; they are internal memory buffers to handle the 2015-era structure of this monsterclass.

src/lib/litegraph/src/draw.ts
```diff
@@ -67,7 +67,7 @@ interface IDrawTextInAreaOptions {
  */
 export function strokeShape(
   ctx: CanvasRenderingContext2D,
-  area: Rect,
+  area: Rectangle,
```
webfiltered, line 70: Hmm, I might be missing something - are we getting something by constraining this further?  It should still work on all `Rect`, afaik?

src/lib/litegraph/src/infrastructure/Rectangle.ts
```diff
@@ -12,17 +10,17 @@ import { isInRectangle } from '@/lib/litegraph/src/measure'
...
  *
- * Sub-array properties ({@link Float64Array.subarray}):
+ * Sub-array properties ({@link Float32Array.subarray}):
  * - {@link pos}: The position of the top-left corner of the rectangle.
  * - {@link size}: The size of the rectangle.
  */
-export class Rectangle extends Float64Array {
-  #pos: Point | undefined
-  #size: Size | undefined
+export class Rectangle extends Float32Array {
```
webfiltered: Is this so it fits in with the places you added it to?  Shouldn't we be doing the _opposite_?

src/lib/litegraph/src/infrastructure/Rectangle.ts
```diff
@@ -81,10 +79,10 @@ export class Rectangle extends Float64Array {
    */
   get pos(): Point {
     this.#pos ??= this.subarray(0, 2)
-    return this.#pos!
+    return this.#pos! as unknown as Point
```
webfiltered, line 82: \<gasp\> You _didn't_!

src/lib/litegraph/src/interfaces.ts
```diff
@@ -226,52 +226,25 @@ export interface IFoundSlot extends IInputOrOutput {
...
-export type ReadOnlyPoint =
-  | readonly [x: number, y: number]
-  | ReadOnlyTypedArray<Float32Array>
-  | ReadOnlyTypedArray<Float64Array>
-
-/** A size represented as `[width, height]` that will not be modified */
-export type ReadOnlySize =
-  | readonly [width: number, height: number]
-  | ReadOnlyTypedArray<Float32Array>
-  | ReadOnlyTypedArray<Float64Array>
```
webfiltered, line 250: Really happy to see these are no longer necessary - thank you!

src/lib/litegraph/src/LGraphCanvas.ts
webfiltered:
```suggestion
    size[0] = node.renderingSize[0]
    size[1] = node.renderingSize[1]
```

src/lib/litegraph/src/LGraphCanvas.ts
DrJKL: It's more that Proxies don't like #private fields, but I'm happy to pull these out of the class.
(I don't like them either, but...)

src/lib/litegraph/src/draw.ts
DrJKL: Long term, the goal is to have fewer options, but better support. A loose contract is a maintenance burden.

src/lib/litegraph/src/infrastructure/Rectangle.ts
DrJKL: How fine-grained are we going with these rectangles?

src/lib/litegraph/src/infrastructure/Rectangle.ts
DrJKL: I know, I hate it...

src/lib/litegraph/src/interfaces.ts
DrJKL: 🫡

src/lib/litegraph/src/infrastructure/Rectangle.ts
webfiltered: Using Float32Array over Float64Array was a choice made by people over 10 years ago.  These days, it's a memory optimisation at the cost of CPU, complexity, and lossy conversions every time it is read, calculated, and written.

There's just no good reason to use it unless you have a LOT of data (we do not come close).

src/lib/litegraph/src/LGraphCanvas.ts
webfiltered: As discussed - LGraphCanvas roundabout has been done to death.  Making it reactive is a no-go zone.  It breaks several things, and not just the private stuff - can't actually recall.

If there's seriously benefit to attempting this, we should check out whether it's even possible first, before refactoring to support it.

src/lib/litegraph/src/infrastructure/Rectangle.ts
DrJKL: Yeah, I plan on getting rid of it entirely in a subsequent PR, but I wanted to get this one out there for review before focusing on the Rectangle shaped pieces that have more complexity to them (synchronized views on the buffer).

webfiltered: I'm still a bit confused on why pushing everything to Float32Array is necessary.  It impacts all the things that were already using `Float64Array`-based `Rectangle`.

Honestly, not the end of the world - just felt like it might be an unnecessary step backwards.

DrJKL: I think this was the only `Float64Array`, but there were a few places that had their own non-`Rectangle` Float32Arrays that functioned in the same way.
First step: Unify them all and isolate the direct invocation of `Float\d\dArray` 
Second step: Fix the Rectangle contract to not need the TypedArray inheritance
Third step: ???

webfiltered: Right - this was the only place the code had literally `Float64Array`, but Rectangle is already _used_ in a few places.

Option 1: Moving Float32Array to Float64Array may result in _some_ off-by-`Number.MIN_VALUE` errors - this is the only reason I didn't do it a year ago.

Option 2: Changing from 64bit -> 32bit though, will *immediately* impact workflows - save them, and most/all 64 bit values will have changed.

Unless I'm seriously missing something, there's just no upside to using Float32Array.

DrJKL: 🧞 As you wish: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5877/commits/2f4a398fac9c6374832c07e235848d7b9420095d

webfiltered (commented):
Thank you, it looks splendid~

webfiltered (approved):

christian-byrne (approved):

PR #5878, opened 2025-10-01
Add MediaAssetCard presentation components

viva-jinyi:
## Summary

Implements a comprehensive media asset card component system for the Asset Manager sidebar, enabling display and interaction with various media types (images, videos, audio, and 3D models).

### New Components
- **MediaAssetCard**: Main card component for displaying media assets
- **Media type-specific components**: Specialized display logic for each media type
  - MediaImageTop/Bottom
  - MediaVideoTop/Bottom  
  - MediaAudioTop/Bottom
  - Media3DTop/Bottom
- **MediaAssetActions**: Top-left action buttons (delete, download, more options)
- **MediaAssetMoreMenu**: Dropdown menu for additional actions
- **SquareChip**: Chip component for displaying duration and file format with dark/light variants
- **MediaAssetButtonDivider**: Visual separator for button groups

### Features
- **Video playback**: Autoplay with native video controls
  - Dynamic duration chip positioning based on control visibility
  - Hides overlays when video is playing
- **Audio playback**: Audio icon with HTML5 audio element
  - Duration chip with consistent positioning
- **3D model support**: Icon display for 3D assets
- **Selection state**: Proper hover and selected state handling with CSS priority fixes

### Architecture Improvements
- **Domain-Driven Design structure**: Organized under `src/platform/mediaAsset/` following DDD principles
- **Provide/Inject pattern**: Eliminates props drilling with MediaAssetKey InjectionKey
- **Composable pattern**: `useMediaAssetActions` manages all action handlers
- **Type safety**: Comprehensive TypeScript types for media assets and actions

### UI/UX Enhancements
- **CardTop component**: Added custom class props for slot positioning
- **SquareChip component**: Backdrop blur effects with variant system
- **Lazy loading**: Image optimization with LazyImage component
- **Responsive states**: Loading, selected, and hover states

### Utilities
- **formatDuration**: Converts milliseconds to human-readable format (45s, 1m 23s, 1h 2m)

## Testing
- Comprehensive Storybook stories for all media types
- Grid layout examples
- Loading and selected state demonstrations

## File Structure
```
src/platform/assets/
├── components/
│   ├── MediaAssetCard.vue
│   ├── MediaAssetCard.stories.ts
│   ├── MediaAssetActions.vue
│   ├── MediaAssetMoreMenu.vue
│   ├── MediaAssetButtonDivider.vue
│   ├── MediaImageTop.vue
│   ├── MediaImageBottom.vue
│   ├── MediaVideoTop.vue
│   ├── MediaVideoBottom.vue
│   ├── MediaAudioTop.vue
│   ├── MediaAudioBottom.vue
│   ├── Media3DTop.vue
│   └── Media3DBottom.vue
├── composables/
│   └── useMediaAssetActions.ts
└── schemas/
    └── mediaAssetSchema.ts
```

## Screenshots
[media_asset_record.webm](https://github.com/user-attachments/assets/d13b5cc0-a262-4850-bb81-ca1daa0dd969)


src/platform/mediaAsset/composables/useMediaAssetActions.ts
```diff
@@ -0,0 +1,40 @@
...
+  onCopy: (assetId: string) => void
+  onCopyJobId: (assetId: string) => void
+  onMore: (assetId: string) => void
+  onAddToWorkflow: (assetId: string) => void
+  onOpenWorkflow: (assetId: string) => void
+  onExportWorkflow: (assetId: string) => void
+  onOutputCountClick: (assetId: string) => void
+}
+
+export function useMediaAssetActions(emit: any): MediaAssetActions {
```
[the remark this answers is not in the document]
viva-jinyi: good point

src/platform/mediaAsset/components/cards/MediaVideoTop.vue
```diff
@@ -0,0 +1,76 @@
...
+    <!-- Show video player if playing -->
+    <div
+      v-if="showVideoPlayer"
+      class="h-full w-full bg-black"
+      @mouseenter="showControls = true"
+      @mouseleave="showControls = false"
+    >
+      <video
+        :controls="showControls"
+        autoplay
```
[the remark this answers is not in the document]
viva-jinyi: good point

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -0,0 +1,216 @@
+<template>
+  <CardContainer
```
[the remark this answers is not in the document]
viva-jinyi: good point

src/components/card/CardContainer.vue
christian-byrne, line 1: note: I like the way you destructured the `class`, removed automatic binding, then remerged with `cn` - that's a nice pattern.

src/components/button/MoreButton.vue
```diff
@@ -25,12 +28,24 @@
...
 import { cn } from '@/utils/tailwindUtil'
 
 import IconButton from './IconButton.vue'
 
 const popover = ref<InstanceType<typeof Popover>>()
 
+const { size, isVertical = false } = defineProps<{
+  size?: ButtonSize
+  isVertical?: boolean
+  containerClass?: string
```
christian-byrne: Is this unused?

src/components/button/MoreButton.vue
```diff
@@ -25,12 +28,24 @@
...
 
+import type { ButtonSize } from '@/types/buttonTypes'
 import { cn } from '@/utils/tailwindUtil'
 
 import IconButton from './IconButton.vue'
 
 const popover = ref<InstanceType<typeof Popover>>()
 
+const { size, isVertical = false } = defineProps<{
+  size?: ButtonSize
```
christian-byrne: Looking at how this is used:

- Binded directly to `IconButton`
- `IconButton` binds it directly to `Button` (PrimeVue button)

In `IconButton`, we have a default of "md" https://github.com/Comfy-Org/ComfyUI_frontend/blob/0851f2ff80641c8c7eb0c30c1d8c9f7ce5fa2923/src/components/button/IconButton.vue#L35

But in the PrimeVue props interface, the valid values are "small" and "medium" https://primevue.org/button/#api.button.props.size, so it seems to be incorrect, is that accurate?

src/components/card/CardContainer.vue
```diff
@@ -8,26 +8,77 @@
...
+
+const {
+  size = 'regular',
+  variant = 'default',
+  rounded = 'md',
+  customAspectRatio,
+  noBorder = false,
+  noBackground = false,
+  noShadow = false,
+  noCursor = false,
```
christian-byrne: nit (non-blocking): I think `hasBorder`, `hasShadow` is easier to understand compared with double negatives.

src/components/common/LazyImage.vue
```diff
@@ -45,12 +46,14 @@ import { useMediaCache } from '@/services/mediaCacheService'
...
   src,
   alt = '',
+  containerClass = '',
   imageClass = '',
   imageStyle,
   rootMargin = '300px'
 } = defineProps<{
   src: string
   alt?: string
+  containerClass?: string | string[] | Record<string, boolean>
```
christian-byrne: suggestion: How about using the official class?

```typescript
import type { ClassValue } from 'vue'
```

src/platform/mediaAsset/components/cards/Media3DBottom.vue
```diff
@@ -0,0 +1,30 @@
...
+<script setup lang="ts">
+import { computed } from 'vue'
+
+import { formatSize, getFilenameDetails } from '@/utils/formatUtil'
+
+import type { AssetContext, AssetMeta } from '../../types'
+
+const { asset } = defineProps<{
+  asset: AssetMeta
+  context: AssetContext
```
christian-byrne: nit: Where is this used?

src/platform/assets/components/MediaAssetMoreMenu.vue
christian-byrne, line 1: Are the labels automatically translated or do we need i18n?

src/platform/assets/components/MediaAudioBottom.vue
```diff
@@ -0,0 +1,30 @@
...
+<script setup lang="ts">
+import { computed } from 'vue'
+
+import { formatSize, getFilenameDetails } from '@/utils/formatUtil'
+
+import type { AssetContext, AssetMeta } from '../../types'
+
+const { asset } = defineProps<{
+  asset: AssetMeta
+  context: AssetContext
```
christian-byrne, line 24: nit: Same question as before. Maybe we can also re-use this component? (low priority).

src/platform/mediaAsset/components/cards/MediaAudioTop.vue
```diff
@@ -0,0 +1,29 @@
...
+    />
+  </div>
+</template>
+
+<script setup lang="ts">
+import type { AssetContext, AssetMeta } from '../../types'
+
+const { asset } = defineProps<{
+  asset: AssetMeta
+  context: AssetContext
```
christian-byrne: question: Same as with other 2

src/platform/assets/components/MediaAudioTop.vue
```diff
@@ -0,0 +1,29 @@
...
+      :src="asset.src"
+      @click.stop
+    />
+  </div>
+</template>
+
+<script setup lang="ts">
+import type { AssetContext, AssetMeta } from '../../types'
+
+const { asset } = defineProps<{
```
christian-byrne, line 25: nit: No need to destructure this if it's not used in the script.

src/platform/assets/components/MediaImageBottom.vue
```diff
@@ -0,0 +1,30 @@
...
+<script setup lang="ts">
+import { computed } from 'vue'
+
+import { getFilenameDetails } from '@/utils/formatUtil'
+
+import type { AssetContext, AssetMeta } from '../../types'
+
+const { asset } = defineProps<{
+  asset: AssetMeta
+  context: AssetContext
```
christian-byrne, line 24: nit: Same as others

src/platform/mediaAsset/components/cards/MediaImageTop.vue
```diff
@@ -0,0 +1,28 @@
...
+</template>
+
+<script setup lang="ts">
+import LazyImage from '@/components/common/LazyImage.vue'
+
+import type { AssetContext, AssetMeta } from '../../types'
+
+const { asset } = defineProps<{
+  asset: AssetMeta
+  context: AssetContext
```
christian-byrne: nit: same as others.

packages/design-system/src/icons/play.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M4 2L13.3333 8L4 14V2Z" fill="currentColor" stroke="currentColor" stroke-width="1.3" stroke-linecap="round" stroke-linejoin="round"/>
```
arjansingh: It's worth running this through https://svgomg.net/ or https://github.com/svg/svgo

We can probably reduce the size by ~10%. This matters a lot more with larger SVGs though.

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -451,3 +451,26 @@ export function stringToLocale(locale: string): SupportedLocale {
     ? (locale as SupportedLocale)
     : 'en'
 }
+
+export function formatDuration(milliseconds: number): string {
```
arjansingh, line 455: It's worth writing a few tests for this to document the the expected formats.

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -0,0 +1,229 @@
...
+    image: defineAsyncComponent(() => import('./cards/MediaImageTop.vue')),
+    '3D': defineAsyncComponent(() => import('./cards/Media3DTop.vue'))
+  },
+  bottom: {
+    video: defineAsyncComponent(() => import('./cards/MediaVideoBottom.vue')),
+    audio: defineAsyncComponent(() => import('./cards/MediaAudioBottom.vue')),
+    image: defineAsyncComponent(() => import('./cards/MediaImageBottom.vue')),
+    '3D': defineAsyncComponent(() => import('./cards/Media3DBottom.vue'))
+  }
+}
```
christian-byrne, line 140: cool design!

christian-byrne (commented):
Looking good!

src/components/button/MoreButton.vue
arjansingh: question: Where's `containerClass` used?

src/components/card/CardContainer.vue
```diff
@@ -8,26 +8,77 @@
...
 const containerClasses = computed(() => {
-  const baseClasses =
-    'cursor-pointer flex flex-col bg-white dark-theme:bg-zinc-800 rounded-lg shadow-sm border border-zinc-200 dark-theme:border-zinc-700 overflow-hidden'
+  // Base structure classes
+  const structureClasses = 'flex flex-col overflow-hidden'
 
-  if (type === 'workflow-template-card') {
-    return `cursor-pointer p-2 flex flex-col hover:bg-white dark-theme:hover:bg-zinc-800 rounded-lg transition-background duration-200 ease-in-out`
+  // Rounded corners
+  const roundedClasses = {
```
arjansingh: nit: this could be extracted outside of the computed as a `const`. it never changes.

src/components/card/CardTop.vue
```diff
@@ -4,28 +4,42 @@
...
-      class="absolute bottom-2 left-2 flex flex-wrap justify-start gap-2"
+      :class="getSlotClasses('bottom-left', bottomLeftClass)"
     >
       <slot name="bottom-left"></slot>
     </div>
 
     <div
       v-if="slots['bottom-right']"
-      class="absolute right-2 bottom-2 flex flex-wrap justify-end gap-2"
+      :class="getSlotClasses('bottom-right', bottomRightClass)"
```
arjansingh: question: Could these be computeds to avoid computation on re-renders?

src/components/chip/SquareChip.vue
```diff
@@ -1,13 +1,28 @@
...
+const { label, variant = 'dark' } = defineProps<{
   label: string
+  variant?: 'dark' | 'light'
 }>()
+
+const chipClasses = computed(() => {
+  const baseClasses =
+    'inline-flex shrink-0 items-center justify-center gap-1 rounded px-2 py-1 text-xs font-bold'
+
+  const variantStyles = {
```
arjansingh: nit: these styles could also be defined outside of the computed.

src/components/widget/SampleModelSelector.vue
```diff
@@ -88,7 +88,7 @@
     <template #content>
       <!-- Card Examples -->
       <div :style="gridStyle">
-        <CardContainer v-for="i in 100" :key="i" ratio="square">
+        <CardContainer v-for="i in 100" :key="i" size="regular">
```
arjansingh, line 91: I think we can delete this file and the related composable. The Asset Browser Modal takes over for it completely.

src/platform/mediaAsset/types/index.ts
```diff
@@ -0,0 +1,51 @@
+/**
+ * Media Asset domain types
```
arjansingh: Change Request:

Definitely double check what you have here with [assetSchema.ts](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/platform/assets/schemas/assetSchema.ts).

This schema is what the API actually returns currently. We will need to fallback gracefully if we don't have the data we need right now.


For example we could do this for `AssetMeta`:

```ts
  export const mediaKindSchema = z.enum(['video', 'audio', 'image', '3D'])
  export type MediaKind = z.infer<typeof mediaKindSchema>

  // Extend the base asset schema with media-specific fields
  export const mediaAssetDisplayItemSchema = assetItemSchema.extend({
    // New required fields
    kind: mediaKindSchema,
    src: z.string().url(),

    // New optional fields
    duration: z.number().nonnegative().optional(),
    dimensions: dimensionsSchema.optional(),
    jobId: z.string().optional(),
    isMulti: z.boolean().optional()
  })
```

src/platform/mediaAsset/composables/useMediaAssetActions.ts
```diff
@@ -0,0 +1,40 @@
...
+  return {
+    // Actions that emit to parent
+    onSelect: (asset: AssetMeta) => emit('select', asset),
+    onView: (assetId: string) => emit('view', assetId),
+    onDownload: (assetId: string) => emit('download', assetId),
+    onDelete: (assetId: string) => emit('delete', assetId),
+    onPlay: (assetId: string) => emit('play', assetId),
+    // Internal actions (can be extended later)
+    onCopy: (assetId: string) => console.log('Copy:', assetId),
```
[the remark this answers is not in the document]
arjansingh: I agree with claude here, these should be `noops`

src/platform/mediaAsset/components/cards/MediaAssetActions.vue
```diff
@@ -0,0 +1,48 @@
...
+import MoreButton from '@/components/button/MoreButton.vue'
+
+import { MediaAssetKey } from '../../types'
+import MediaAssetMoreMenu from './MediaAssetMoreMenu.vue'
+
+const emit = defineEmits<{
+  menuStateChanged: [isOpen: boolean]
+}>()
+
+const { asset, actions } = inject(MediaAssetKey)!
```
arjansingh: Question: Do we need to inject here? Can we leverage a composable or another method for providing this info?

src/platform/mediaAsset/composables/useMediaAssetActions.ts
```diff
@@ -0,0 +1,45 @@
...
+    event: K,
+    ...args: MediaAssetEmits[K]
+  ) => void
+): MediaAssetActions {
+  return {
+    // Actions that emit to parent
+    onSelect: (asset: AssetMeta) => emit('select', asset),
+    onView: (assetId: string) => emit('view', assetId),
+    onDownload: (assetId: string) => emit('download', assetId),
+    onDelete: (assetId: string) => emit('delete', assetId),
```
arjansingh: Question: I'm not familiar with this pattern, how are we going to use it?

could we use a more direct pattern like:

```js
  export function useMediaAssetActions() {
    const assetService = useAssetService()
    const toast = useToast()
    const confirmDialog = useConfirmDialog()

    const deleteAsset = async (assetId: string) => {
      const confirmed = await confirmDialog.show({
        title: 'Delete Asset',
        message: 'This action cannot be undone.'
      })
      if (!confirmed) return false

      await assetService.deleteAsset(assetId)
      toast.success('Asset deleted')
      return true
    }

    // ... other real implementations

    return { deleteAsset, downloadAsset, copyAssetUrl }
  }
```

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -0,0 +1,229 @@
...
+
+const isHovered = useElementHover(cardContainerRef)
+
+const showHoverActions = computed(() => {
+  return !loading && !!asset && (isHovered.value || isMenuOpen.value)
+})
+
+const actions = useMediaAssetActions(emit)
+
+provide(MediaAssetKey, {
```
arjansingh, line 167: Question: Do we need to provide/inject here? Could the Parent get the information from the `assetService` and just pass as props?

arjansingh (changes requested):
For the most part this is good.

I have questions around the use of `emit`.

Agree with @christian-byrne this should be in `/platform/assets/`.

Once you do that we can also leverage `assetSchema.ts` to help a lot with the types and validation. See my comments.

src/platform/mediaAsset/components/cards/Media3DBottom.vue
viva-jinyi: I deleted context props. Thx.

src/components/card/CardContainer.vue
viva-jinyi: Good Point. Thx.

src/components/card/CardContainer.vue
viva-jinyi: It should be changed. Thx.

src/components/button/MoreButton.vue
viva-jinyi: Oh, It's unused. Thx.

viva-jinyi: I should remove it. Thx.

src/components/button/MoreButton.vue
viva-jinyi: Good catch! But actually there's no issue here. Since we declare size in defineProps, Vue 3 automatically excludes it from $attrs. So even with v-bind="$attrs", the size prop never reaches PrimeVue Button. We use unstyled Button and apply our own size classes through buttonStyle. The 'sm'/'md' values are for our custom styling system, completely independent from PrimeVue's sizing.

viva-jinyi: To clarify - we're not using PrimeVue's size prop:
**If we were using PrimeVue's size (we're not):**
스크린샷 2025-10-11 오후 6 10 24
**Our actual implementation:**
스크린샷 2025-10-11 오후 6 10 05

For type safety, we use BaseButtonProps interface which defines our custom button types:
  - size?: 'fit-content' | 'sm' | 'md' - our custom sizes
  - type?: 'primary' | 'secondary' | 'transparent'
  - Other shared button properties

 This ensures consistency across all our button components while maintaining complete control over the styling.

src/platform/assets/components/MediaAudioTop.vue
viva-jinyi: The asset is used, so I think it is necessary

src/components/common/LazyImage.vue
viva-jinyi: Actually, ClassValue is from @comfyorg/tailwind-utils package. I've added the correct import.

Good suggestion though! We should consider using this type consistently across other common components as well for better type safety and flexibility when passing class props.

src/platform/assets/components/MediaAssetMoreMenu.vue
viva-jinyi: I will add the translation key when I connect the features.

packages/design-system/src/icons/play.svg
viva-jinyi: Thank you! I will update it.

packages/shared-frontend-utils/src/formatUtil.ts
viva-jinyi: Good point about documenting the expected formats. However, our current test setup doesn't run tests inside the packages/shared-frontend-utils/ directory - the vitest configuration only includes tests-ui/ and src/ directories. We'd need to set up a proper test environment for the shared packages first before adding tests there. I'll keep this in mind for future improvements when we establish package-level testing infrastructure.

src/components/card/CardTop.vue
viva-jinyi: Good catch! You're absolutely right. I've refactored it to use a computed property that caches all slot classes. This avoids unnecessary recalculation on every render. Thanks for the suggestion!

src/components/chip/SquareChip.vue
viva-jinyi: Yes, You're right.

src/components/widget/SampleModelSelector.vue
viva-jinyi: The Asset Browser Modal doesn't completely replace this component. It currently copies the card component styles instead of using the actual card component. I'll refactor the Asset Browser to use the proper card components in a separate PR, and then we can delete this file. For now, these components serve as the foundation for the consistent card UI across the application.

src/platform/mediaAsset/composables/useMediaAssetActions.ts
viva-jinyi: Thanks for good example!

src/platform/assets/components/MediaAssetCard.vue
viva-jinyi: I think we don't need provide/inject here. It can simply call useMediaAssetActions() directly where needed.

src/platform/mediaAsset/components/cards/MediaAssetActions.vue
viva-jinyi: I will use a hybrid approach - composables for actions and provide/inject for data (avoids prop drilling through dynamically rendered components).

src/platform/mediaAsset/types/index.ts
viva-jinyi: I've implemented your suggestion by creating `mediaAssetSchema.ts` and extending `assetItemSchema`:

1. **Created new schema file**: `/platform/assets/schemas/mediaAssetSchema.ts`
2. **Extended existing schema**: Using `assetItemSchema.extend()` to match API response
3. **Zod schema-based types**: Ensuring runtime validation and type safety

viva-jinyi:
@christian-byrne @arjansingh I’ve revised almost everything based on the review.
For the comments I marked as “resolved” without adding a reply — those were issues I already fixed elsewhere while addressing similar feedback in other parts of the code.
For the areas that didn’t require changes, I left explanatory comments instead.

christian-byrne (commented):
LGTM

src/platform/assets/schemas/mediaAssetSchema.ts
```diff
@@ -0,0 +1,46 @@
...
+  // New optional fields
+  duration: z.number().nonnegative().optional(),
+  dimensions: zDimensionsSchema.optional(),
+  jobId: z.string().optional(),
+  isMulti: z.boolean().optional()
+})
+
+// Asset context schema
+const zAssetContextSchema = z.object({
+  type: z.enum(['input', 'output']),
```
arjansingh, line 29: comment: FYI for future work, the types are going to come as asset tags.
The openapi schema for assets is here.

as you continue to develop, you probably should standardize your types around the records the API is returning.

src/platform/assets/schemas/mediaAssetSchema.ts
```diff
@@ -0,0 +1,46 @@
...
+const zMediaAssetDisplayItemSchema = assetItemSchema.extend({
+  // New required fields
+  kind: zMediaKindSchema,
+  src: z.string().url(),
+
+  // New optional fields
+  duration: z.number().nonnegative().optional(),
+  dimensions: zDimensionsSchema.optional(),
+  jobId: z.string().optional(),
+  isMulti: z.boolean().optional()
```
arjansingh, line 24: question: where is this data going to be coming from?

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -0,0 +1,233 @@
...
+  return mediaComponents.top[kind] || mediaComponents.top.image
+}
+
+function getBottomComponent(kind: MediaKind) {
+  return mediaComponents.bottom[kind] || mediaComponents.bottom.image
+}
+
+const { context, asset, loading, selected } = defineProps<{
+  context: AssetContext
+  asset?: AssetMeta
```
arjansingh, line 152: Follow Up:

To avoid confusion this should be renamed `assetMeta`.

`asset` conventionally refers to full record from the API. See [here](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/platform/assets/schemas/assetSchema.ts#L49).

src/platform/assets/components/MediaAssetCard.stories.ts
```diff
@@ -0,0 +1,318 @@
...
+          id: 'grid-1',
+          name: 'image-file.jpg',
+          kind: 'image',
+          size: 2097152,
+          duration: 4500,
+          created_at: Date.now().toString(),
+          src: SAMPLE_MEDIA.image1,
+          dimensions: { width: 1920, height: 1080 },
+          tags: []
+        },
```
arjansingh, line 237: Where is this data coming from?

packages/shared-frontend-utils/src/formatUtil.ts
arjansingh: @christian-byrne FYI ☝️. I will add to tech improvements.

src/platform/assets/components/MediaAssetActions.vue
```diff
@@ -0,0 +1,50 @@
...
+  menuStateChanged: [isOpen: boolean]
+}>()
+
+const { asset } = inject(MediaAssetKey)!
+const actions = useMediaAssetActions()
+
+const handleDelete = () => {
+  if (asset.value) {
+    actions.deleteAsset(asset.value.id)
+  }
```
arjansingh, line 42: nit:

```suggestion
  if (!asset.value) return
  actions.deleteAsset(asset.value.id)
```

src/platform/assets/components/MediaAssetActions.vue
```diff
@@ -0,0 +1,50 @@
...
+}
+
+const handleDownload = () => {
+  if (asset.value) {
+    actions.downloadAsset(asset.value.id)
+  }
```
arjansingh, line 48: nit:

```suggestion
  if (!asset.value) return
  actions.downloadAsset(asset.value.id)
```

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -0,0 +1,233 @@
...
+})
+
+const showOutputCount = computed(() => {
+  return !loading && context?.outputCount && !isVideoPlaying.value
+})
+
+const handleCardClick = () => {
+  if (asset) {
+    actions.selectAsset(asset)
+  }
```
arjansingh, line 231: nit:
```suggestion
  if (!asset) return
  actions.selectAsset(asset)
```

arjansingh (approved):
I have some questions and follow up requests (see comments) but we can merge this. Approved!

PR #5887, opened 2025-10-02
ci: fork-safe checkout for lint workflow

benceruleanlu:
Use conditional ref to support forks and avoid checkout failures.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5887-ci-fork-safe-checkout-for-lint-workflow-2806d73d36508139b9effa6d18e1cadb) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/lint-and-format.yaml
+++ b/.github/workflows/lint-and-format.yaml
@@ -15,7 +15,7 @@ jobs:
       - name: Checkout PR
         uses: actions/checkout@v5
         with:
-          ref: ${{ github.event.pull_request.head.repo.full_name == github.repository && github.head_ref || github.ref }}
+          ref: ${{ !github.event.pull_request.head.repo.fork && github.head_ref || github.ref }}
 
       - name: Install pnpm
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-and-format.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Checkout PR
         uses: actions/checkout@v5
         with:
-          ref: ${{ github.head_ref }}
+          ref: ${{ !github.event.pull_request.head.repo.fork && github.head_ref || github.ref }}
```
benceruleanlu, line 18: github.event.pull_request.head.repo.fork exists per docs:

response example: https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository
response spec, you'll need to dig deep here: https://docs.github.com/en/webhooks/webhook-events-and-payloads#pull_request

DrJKL: GitHub makes me sad.

DrJKL (approved):

PR #5888, opened 2025-10-02
Tests: Vitest configuration cleanup

DrJKL:
## Summary

Simplify default scripts. Filtering is still available to users, we can revisit tagging or grouping later. 
This fixes the issue where we had tests that were in the codebase but never run because they weren't under `/src/components`

Also deletes the duplicate litegraph tests and their associated vitest config file.

## Changes

- **What**: Test cleanup

## Review Focus

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5888-Tests-Vitest-configuration-cleanup-2806d73d36508197b800f68f0b028279) by [Unito](https://www.unito.io)


AustinMroz (commented):
LGTM

arjansingh (commented):
LGTM.

I'm jealous, i'll never get a commit to take  line line count that negative. You win.

christian-byrne (commented):
I think the only tests that are being removed entirely are 

- CanvasPointer.deviceDetection.test.ts
- LGraphNodeProperties.test.ts

I think the first is intentional, but what about the second?

DrJKL:
> I think the only tests that are being removed entirely are
> 
>     * CanvasPointer.deviceDetection.test.ts
> 
>     * LGraphNodeProperties.test.ts
> 
> 
> I think the first is intentional, but what about the second?

I'll bring them both back, though I do think the first one's been obviated by more recent design decisions, yeah.

christian-byrne (approved):
LGTM! High score!

PR #5892, opened 2025-10-02
feat(api-nodes): add pricing for new Kling 2.5 Turbo model

bigcat88:
## Summary

Added prices for new model to the `KlingTextToVideoNode` and `KlingImage2VideoNode` nodes

Screenshot From 2025-10-02 10-51-04

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5892-feat-api-nodes-add-pricing-for-new-Kling-2-5-Turbo-model-2806d73d3650810eb33cf794dca23266) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/node/useNodePricing.ts
+++ b/src/composables/node/useNodePricing.ts
@@ -355,7 +355,12 @@ const apiNodeCosts: Record<string, { displayPrice: string | PricingFunction }> =
         const modelValue = String(modelWidget.value)
 
         // Same pricing matrix as KlingTextToVideoNode
-        if (
+        if (modelValue.includes('v2-5-turbo')) {
+          if (durationValue.includes('10')) {
+            return '$0.70/Run'
+          }
+          return '$0.35/Run' // 5s default
+        } else if (
           modelValue.includes('v2-1-master') ||
           modelValue.includes('v2-master')
         ) {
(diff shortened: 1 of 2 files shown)


src/composables/node/useNodePricing.ts
```diff
@@ -355,7 +355,12 @@ const apiNodeCosts: Record<string, { displayPrice: string | PricingFunction }> =
         const modelValue = String(modelWidget.value)
 
         // Same pricing matrix as KlingTextToVideoNode
-        if (
+        if (modelValue.includes('v2-5-turbo')) {
+          if (durationValue.includes('10')) {
+            return '$0.7/Run'
+          }
+          return '$0.35/Run' // 5s default
```
DrJKL, line 362: Users can't see this comment. Do we hint at length per run anywhere else?

src/composables/node/useNodePricing.ts
```diff
@@ -505,7 +510,12 @@ const apiNodeCosts: Record<string, { displayPrice: string | PricingFunction }> =
         const modeValue = String(modeWidget.value)
 
         // Pricing matrix from CSV data based on mode string content
-        if (modeValue.includes('v2-1-master')) {
+        if (modeValue.includes('v2-5-turbo')) {
+          if (modeValue.includes('10')) {
+            return '$0.7/Run'
```
DrJKL: So this would be 70 cents?
How would you feel about `'$0.70/Run'`?

DrJKL (dismissed):

src/composables/node/useNodePricing.ts
bigcat88: yes, we do that in every node that support length

src/composables/node/useNodePricing.ts
bigcat88: that makes sense, changed to `$0.70/Run`

christian-byrne (approved):

PR #5909, opened 2025-10-04
fix: node deletion handling in vue nodes when switching from litegrap…

simula-r:
## Summary

Fix node deletion when switching from litegraph -> vue node mode. 

## Background:
 
There is a race condition where we have [2 watch() functions that are both tracking shouldRenderVueNodes.value](https://github.com/Comfy-Org/ComfyUI_frontend/blob/f58365b20b2eb768f0db21eafa29d32291e064fb/src/composables/graph/useVueNodeLifecycle.ts#L85):
  - For vue node lifecycle
  - For slot/link sync lifecycle

Each watch() separately sets graph.onNodeRemoved to its own cleanup handler. But without flush: 'sync', the slot sync watch runs AFTER vue nodes and overwrites the callback. So when deletion happens, graph.onNodeRemoved points to the slot handler (or undefined) instead of the vue cleanup handler, and the vue node never gets removed from the DOM.

## Review Focus
We are making use of the watch() option "fush: sync" in the link and slot watch() so maybe there is side effects from firing this when dependencies change synchronously. 

https://github.com/user-attachments/assets/6951fa50-61d5-4c63-88e9-e113f603ff77

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5909-fix-node-deletion-handling-in-vue-nodes-when-switching-from-litegrap-2826d73d365081bdba35c3d8728d6251) by [Unito](https://www.unito.io)


DrJKL (dismissed):

christian-byrne:
This seems unexpected: https://f4221684.comfyui-playwright-chromium.pages.dev/#?testId=2bf9a4bcfdc4cfb07393-89182de5dbf3d9e4b95f

christian-byrne:
On second look, the new screenshots are actually correct as they reflect the [real test workflow](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/browser_tests/assets/nodes/every_node_color.json) state. The prior screenshots were wrong and the undrelying reason for that baseline existing is evidently fixed by this change.

browser_tests/tests/vueNodes/nodeStates/colors.spec.ts-snapshots/vue-node-custom-colors-dark-all-colors-chromium-linux.png
DrJKL, line 1: Where'd all the new loaders come from?

DrJKL (approved):

browser_tests/tests/vueNodes/nodeStates/colors.spec.ts-snapshots/vue-node-custom-colors-dark-all-colors-chromium-linux.png
christian-byrne: Those are actually in the test workflow

PR #5911, opened 2025-10-04
Subgraph widget promotion fixes

AustinMroz:
- Fixes automatic promotion of image previews by ~~more correctly handling a usage of `requestAnimationFrame` and~~ introducing a means to perform actions upon successful load of preview.
- When workflows contain an old proxyWidget property in string form, parse it instead of throwing an error.
- Do not overlay the `label` of promoted widgets. Further consideration is needed, but this resolves the primary annoyance and prevents untranslated widget names
See #5914

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5911-Subgraph-widget-promotion-fixes-2826d73d365081838ffeeea4b8d7068c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/node/useNodeImage.ts
+++ b/src/composables/node/useNodeImage.ts
@@ -98,7 +98,7 @@ const useNodePreview = <T extends MediaElement>(
 /**
  * Attaches a preview image to a node.
  */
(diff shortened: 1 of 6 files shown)


christian-byrne (approved):
LGTM!

PR #5920, opened 2025-10-04
use PrimeVue tooltip for "enter subgraph" button on Vue node headers

christian-byrne:
## Summary

Replaced HTML `title` attribute with PrimeVue tooltip component for the "Enter Subgraph" button in Vue node headers.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5920-use-PrimeVue-tooltip-for-enter-subgraph-button-on-Vue-node-headers-2826d73d365081bdb0b9dcc7b4509d39) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/NodeHeader.vue
+++ b/src/renderer/extensions/vueNodes/components/NodeHeader.vue
@@ -45,11 +45,11 @@
       <div class="flex items-center lod-toggle shrink-0">
         <IconButton
           v-if="isSubgraphNode"
+          v-tooltip.top="enterSubgraphTooltipConfig"
           size="sm"
           type="transparent"
           class="text-stone-200 dark-theme:text-slate-300"
           data-testid="subgraph-enter-button"
-          title="Enter Subgraph"
           @click.stop="handleEnterSubgraph"
           @dblclick.stop
         >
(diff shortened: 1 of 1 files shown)


src/renderer/extensions/vueNodes/components/NodeHeader.vue
```diff
@@ -123,6 +123,10 @@ const tooltipConfig = computed(() => {
   return createTooltipConfig(description)
 })
 
+const enterSubgraphTooltipConfig = computed(() => {
+  return createTooltipConfig(st('enterSubgraph', 'Enter Subgraph'))
+})
```
DrJKL, line 128: If there's nothing reactive here, I don't think this needs to be `computed`

DrJKL (approved):

src/renderer/extensions/vueNodes/components/NodeHeader.vue
christian-byrne: The `computed()` is actually necessary here because `st()` is reactive - it uses Vue i18n's `t()` and `te()` functions which update when the locale changes. Without `computed()`, the tooltip wouldn't update when users switch languages. To verify, add a console.log in the computed callback and change the locale in the settings.

PR #5921, opened 2025-10-04
fix mmb navigation when click starts on Vue nodes

### LINKED ISSUE #5860: Vue nodes feedback
Was just checking out the beautiful awesomely cool new vue nodes
Just one critique:

Using middle mouse button to pan the nodegraph is great

Using middle mouse button to pan the nodegraph but also move nodes depending on if you're hovering over a node or in a blank space on the graph. Not great.
- No reason to introduce this, left mouse press already moves nodes reliably
- Extremely annoying trying to find a space inbetween nodes to middle mouse press - to pan the nodegraph
- Accidentally moving nodes instead of panning nodegraph
- Goes against the muscle memory people have already developed
- Totally not cool. At all.

Anyway, vue nodes looking amazing so far! Great work! ❤️

christian-byrne:
## Summary

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/5860
 by updating Vue node pointer interactions to forward middle mouse button events to canvas instead of handling them locally.

## Review Focus

Middle mouse button event detection logic using both `button` property and `buttons` bitmask for cross-browser compatibility. Test coverage for pointer event forwarding behavior.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5921-fix-mmb-navigation-when-click-starts-on-Vue-nodes-2826d73d3650819688eec4600666755d) by [Unito](https://www.unito.io)


src/renderer/core/canvas/useCanvasInteractions.ts
```diff
@@ -54,11 +54,28 @@ export function useCanvasInteractions() {
     // Otherwise, let the component handle it normally
   }
 
+  const isMiddlePointerInput = (event: PointerEvent | MouseEvent): boolean => {
+    if ('button' in event && event.button === 1) {
+      return true
+    }
+
+    if ('buttons' in event && typeof event.buttons === 'number') {
+      return (event.buttons & 4) !== 0
```
DrJKL: bitwise &, huh? Fancy.

question: Do you know if there are any situations where we'd have `buttons` but not `button`?

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts
```diff
@@ -4,10 +4,12 @@ import { nextTick, ref } from 'vue'
...
+const forwardEventToCanvasMock = vi.fn()
+
 // Mock the dependencies
 vi.mock('@/renderer/core/canvas/useCanvasInteractions', () => ({
   useCanvasInteractions: () => ({
-    forwardEventToCanvas: vi.fn(),
+    forwardEventToCanvas: forwardEventToCanvasMock,
     shouldHandleNodePointerEvents: ref(true)
   })
 }))
```
DrJKL, line 15: If you do it this way, then you can grab the mock in the test like you would in a component.

```suggestion
// Mock the dependencies
vi.mock('@/renderer/core/canvas/useCanvasInteractions', () => {
  const forwardEventToCanvas = vi.fn()
  return {
    useCanvasInteractions: () => ({
        forwardEventToCanvas,
        shouldHandleNodePointerEvents: ref(true)
    })
  }
})
```

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -34,6 +34,16 @@ export function useNodePointerInteractions(
   const { forwardEventToCanvas, shouldHandleNodePointerEvents } =
     useCanvasInteractions()
 
+  const isMiddlePointer = (event: PointerEvent): boolean => {
+    return event.button === 1 || (event.buttons & 4) !== 0
+  }
```
DrJKL: Deja vu...

DrJKL: This makes me think that useCanvasInteractions and useNodePointerInteractions should probably be unified.

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -165,8 +181,11 @@ export function useNodePointerInteractions(
   })
 
   const pointerHandlers = {
+    onPointerdownCapture: forwardMiddlePointerIfNeeded,
     onPointerdown: handlePointerDown,
+    onPointermoveCapture: forwardMiddlePointerIfNeeded,
     onPointermove: handlePointerMove,
+    onPointerupCapture: forwardMiddlePointerIfNeeded,
```
DrJKL: I would rather not double all of these until we absolutely need them. Right now they're only used in the test, right?

src/renderer/core/canvas/useCanvasInteractions.ts
christian-byrne: This is just bad, I should make it readable and expressive.

christian-byrne: You're right\! The bitwise logic was unnecessary complexity. Simplified to direct comparison in 6d4d497. Regarding your question: the  property is available during drag operations (mousemove/pointermove) when  might not be, but for middle mouse detection, the direct comparison is cleaner and sufficient.

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts
christian-byrne: Good catch\! The current implementation with forwardEventToCanvasMock declared outside the mock is actually the correct approach for accessing the mock in tests. Your suggested syntax has some issues, but the intent is right - and we're already doing it correctly.

src/base/pointerUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function isMiddlePointerInput(
+  event: PointerEvent | MouseEvent
+): boolean {
+  if ('button' in event && event.button === 1) {
+    return true
+  }
+
+  if ('buttons' in event && typeof event.buttons === 'number') {
+    return event.buttons === 4
```
DrJKL, line 18: The bitwise & was the right way to do the comparison for `buttons`, I was just wondering if we'll ever reach this statement.

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
christian-byrne: Fixed in cbf19a7

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
christian-byrne: Fixed in 53df229

PR #5922, opened 2025-10-04
fix dragging video/image components on Vue nodes triggers node drag

christian-byrne:
## Summary

Fixed pointer capture behavior on video and image preview components to prevent unintended node dragging. Below video shows behavior after fix:

https://github.com/user-attachments/assets/95563a2d-8958-47e1-a19c-977fb539d162

## Changes

- **What**: Added `data-capture-node="true"` attribute to `VideoPreview.vue` and `ImagePreview.vue` components
- **What**: Enhanced `useNodePointerInteractions.ts` composable to detect and handle pointer events on elements with capture attribute

## Review Focus

Pointer event delegation logic and interaction behavior between preview components and canvas drag operations.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5922-fix-dragging-video-image-components-on-Vue-nodes-triggers-node-drag-2826d73d365081ce83e7fd61510167e2) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
+++ b/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
@@ -52,6 +52,16 @@ export function useNodePointerInteractions(
       return
     }
 
+    const stopNodeDragTarget = (event.target as HTMLElement | null)?.closest(
+      '[data-capture-node="true"]'
+    )
+    if (stopNodeDragTarget) {
+      if (!shouldHandleNodePointerEvents.value) {
+        forwardEventToCanvas(event)
+      }
+      return
+    }
+
     // Only start drag on left-click (button 0)
     if (event.button !== 0) {
       return
--- a/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
+++ b/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
@@ -52,9 +52,10 @@ export function useNodePointerInteractions(
       return
     }
 
-    const stopNodeDragTarget = (event.target as HTMLElement | null)?.closest(
-      '[data-capture-node="true"]'
-    )
+    const stopNodeDragTarget =
+      event.target instanceof HTMLElement
+        ? event.target.closest('[data-capture-node="true"]')
+        : null
     if (stopNodeDragTarget) {
       if (!shouldHandleNodePointerEvents.value) {
         forwardEventToCanvas(event)
--- a/src/renderer/extensions/vueNodes/VideoPreview.vue
+++ b/src/renderer/extensions/vueNodes/VideoPreview.vue
@@ -5,6 +5,7 @@
     tabindex="0"
     role="region"
     :aria-label="$t('g.videoPreview')"
+    data-capture-node="true"
     @mouseenter="handleMouseEnter"
     @mouseleave="handleMouseLeave"
     @keydown="handleKeyDown"
--- a/src/renderer/extensions/vueNodes/components/ImagePreview.vue
+++ b/src/renderer/extensions/vueNodes/components/ImagePreview.vue
@@ -2,6 +2,7 @@
   <div
     v-if="imageUrls.length > 0"
     class="image-preview relative group flex flex-col items-center"
+    data-capture-node="true"
     tabindex="0"
     role="region"
     :aria-label="$t('g.imagePreview')"


src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -52,6 +52,16 @@ export function useNodePointerInteractions(
       return
     }
 
+    const stopNodeDragTarget = (event.target as HTMLElement | null)?.closest(
+      '[data-capture-node="true"]'
```
DrJKL: Could you add an `instanceof` guard instead of a type assertion here?

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -52,6 +52,16 @@ export function useNodePointerInteractions(
...
+    )
+    if (stopNodeDragTarget) {
+      if (!shouldHandleNodePointerEvents.value) {
+        forwardEventToCanvas(event)
+      }
```
DrJKL, line 62: Just commenting for signal: I still really don't like the pattern of decorating components and shuffling events and synthetic events around.

DrJKL (dismissed):

DrJKL:
Side-note: I would ask a few users what they would expect here. That's a pretty big chunk of a node to not be part of the drag target, especially when zoomed out.
I'd be a little surprised/annoyed if I were trying to rearrange my workflow and accidentally opened a copy of it by dragging the output onto the canvas.

christian-byrne:
> Side-note: I would ask a few users what they would expect here. That's a pretty big chunk of a node to not be part of the drag target, especially when zoomed out.
> 
> I'd be a little surprised/annoyed if I were trying to rearrange my workflow and accidentally opened a copy of it by dragging the output onto the canvas.


That's true, this was a feature we all decided we wanted and even began implementing but never finished (dragging images). In particular, so that we could add the ability to drop onto canvas and insert the worfklow (rather than open in a new tab).

We should definitely observe how this feels and keep in mind whether it's something we should remove later and maintain current behavior for.

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
christian-byrne: Changed to use instanceof guard in 7909204

silveroxides:
@christian-byrne Let me get this confirmed so I understand what you are doing in the video.
1. You are placing a Load Image and a Save Image node on canvas.
2. You run the workflow, thus saving the loaded image.
3. You delete the Load Image node and place three other nodes: Save Image, CLIP Text Encode (prompt) and Empty Latent Image.
4. You then demonstrate that dragging the preview image from the Save Image node will open the workflow in that image.
5. You finally demonstrate that dragging the image preview from the Load Image node converts it to a subgraph?

So now instead of users dealing with not being able to move multiple selected nodes when accidentally trying to drag them by dragging from the image preview, they are now going to have to deal with frontend opening new workflows or creating subgraph while they are trying to organise their nodes? This is especially annoying since everytime you open a new workflow on frontend currently, the RAM consumption increases without any memory getting freed. Maybe focus on essential stuff than nonsensical implementations which has high likelihood of causing annoyance when trying to use ComfyUI normally.

silveroxides:
Solution. Solve the multi select dragging instead and add buttons which appear with delay when image is hovered over with cursor that adds option to "Load embedded workflow from preview image" and "add Load Image node from the preview image". The delay is to ensure that the risk of clicking them is minimal when trying to arrange nodes by dragging multiple with cursor being on the preview image area.

without-ordinary:
I cannot see a reason why I would want to do DND load of an image's embedded workflow from a preview node. I also cannot think of a normal use case where a preview node's image would have an embedded wf that differs usefully from the currently loaded wf (ie. the wf in the image is likely to be at least a version of the wf currently loaded as it was likely what created the embedded wf in it).

This feels like a weird edge use case undo function. 

When watching the demo video, I thought this new functionality was a bug being addressed by the change, not the intended functionality change. 👀

If attempting to move image nodes is likely to lead to unintended wf opening, that's going to be painful. Keep in mind that there is an option to disable to multiple open wf support but turning off the wf tabs. There is the possibility that this isn't just a minor inconvenience of opening an unwanted addition wf when attempting to move a node, but actually losing the currently open wf.

christian-byrne:
Thanks for the feedback. We will test next week and re calibrate.

PR #5923, opened 2025-10-04
fix stone-100 and stone-200 color variable values swapped

christian-byrne:
## Summary

The `stone-100` and `stone-200` were flipped. This PR swaps them to align with [color variables](https://www.figma.com/design/vALUV83vIdBzEsTJAhQgXq/Comfy-Design-System?node-id=0-1&p=f&vars=1&var-id=538-1913&m=dev)

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5923-fix-stone-100-and-stone-200-color-variable-values-swapped-2826d73d3650815c8c4ae0627aa045d9) by [Unito](https://www.unito.io)


webfiltered (commented):
Test failure: title moves, and text colour changes.

Is `stone` being used for text colour on nodes?  Need to flip usage, if so.  Otherwise, disregard.

(_Deciphering the test madness is getting taxing_.)

DrJKL:
I'm really surprised more screenshots didn't trigger.

christian-byrne:
> Is `stone` being used for text colour on nodes? Need to flip usage, if so. Otherwise, disregard.

Good point. We will be using semantic tokens soon though, so we could wait for that.

PR #5930, opened 2025-10-05
Fix untyped subgraph node outputs in vue mode

AustinMroz:
Under some infrequent circumstances, the node outputs of subgraphNodes lack the boundingRect property. As this is a required property of `INodeSlot`s, it's presence is required to satisfy isSlotObject and the type information is discarded to instead display a placeholder output.

This boundingRect property is not used by vue nodes and the type of nodeData.outputs already satisfies INodeSlot, so this entire typeguard and computed can be removed.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5930-Fix-untyped-subgraph-node-outputs-in-vue-mode-2836d73d3650814f993fc71590eca79b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/NodeSlots.vue
+++ b/src/renderer/extensions/vueNodes/components/NodeSlots.vue
@@ -14,9 +14,9 @@
       />
     </div>
 
-    <div v-if="filteredOutputs.length" class="flex flex-col gap-1 ml-auto">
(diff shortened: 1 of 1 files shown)


DrJKL (approved):
Test failure looks unrelated.

Want to add before/after screenshots to the description?

AustinMroz:
Yep. The same test failing on other PRs I have up and I can repro the change on main

christian-byrne (approved):
LGTM

PR #5935, opened 2025-10-06
fix: unignore stylelint

DrJKL:
## Summary

Minimal fix to let Knip succeed but also use its built-in [stylelint plugin](https://knip.dev/reference/plugins/stylelint).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5935-fix-unignore-stylelint-2846d73d3650816d9949faf405012416) by [Unito](https://www.unito.io)


christian-byrne (approved):
Thanks

PR #5936, opened 2025-10-06
make Vue nodes resizable

christian-byrne:
## Summary

Implemented node resizing functionality for Vue nodes.

https://github.com/user-attachments/assets/a7536045-1fa5-401b-8d18-7c26b4dfbfc3

Resolves https://github.com/Comfy-Org/ComfyUI_frontend/issues/5675.

## Review Focus

ResizeObserver as single source of truth pattern eliminates feedback loops between manual resize and reactive layout updates. Intrinsic content sizing calculation temporarily resets DOM styles to measure natural content dimensions.

```mermaid
graph TD
    A[User Drags Handle] --> B[Direct DOM Style Update]
    B --> C[ResizeObserver Detects Change]
    C --> D[Layout Store Update]
    D --> E[Slot Position Sync]
    
    style E fill:#f9f9f9,stroke:#333,color:#000
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5936-make-Vue-nodes-resizable-2846d73d36508160b3b9db49ad8b273e) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -113,6 +139,13 @@
         </div>
       </div>
     </template>
+
+    <!-- Resize handle -->
+    <div
+      v-if="!readonly"
```
DrJKL:
```suggestion
```

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -173,15 +212,19 @@ const { selectedNodeIds } = storeToRefs(useCanvasStore())
 
 // Inject transform state for coordinate conversion
 const transformState = inject(TransformStateKey)
+if (!transformState) {
+  throw new Error(
+    'TransformState must be provided for node resize functionality'
+  )
+}
```
DrJKL, line 188: This might be worth keeping as a `console.error` to start, then letting `useNodeResize` degrade gracefully if it's not provided.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -264,13 +307,26 @@ onMounted(() => {
   }
 })
 
+const { startResize } = useNodeResize(
+  (newSize, element) => {
+    // Apply size directly to DOM element - ResizeObserver will pick this up
+    if (!isCollapsed.value) {
+      element.style.width = `${newSize.width}px`
+      element.style.height = `${newSize.height}px`
+    }
```
DrJKL: very nit: I'd flip this to "ignore if collapsed" instead of "do things if not collapsed", but it's very much a personal aversion to nested code.

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
```diff
@@ -0,0 +1,131 @@
+import { useEventListener } from '@vueuse/core'
+import { ref } from 'vue'
+
+import type { TransformState } from '@/renderer/core/layout/injectionKeys'
+
+interface UseNodeResizeOptions {
+  /** Transform state for coordinate conversion */
+  transformState: TransformState
+  /** Callback when resize ends */
+  onEnd?: () => void
```
DrJKL: Is this being used?

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
```diff
@@ -0,0 +1,131 @@
...
+    element: HTMLElement
+  ) => void,
+  options: UseNodeResizeOptions
+) {
+  const { transformState } = options
+
+  const isResizing = ref(false)
+  const resizeStartPos = ref<{ x: number; y: number } | null>(null)
+  const resizeStartSize = ref<{ width: number; height: number } | null>(null)
+  const intrinsicMinSize = ref<{ width: number; height: number } | null>(null)
```
DrJKL: I think these shapes are used enough times to be worth _naming_

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
```diff
@@ -0,0 +1,131 @@
...
+
+        target.releasePointerCapture(upEvent.pointerId)
+        stopMoveListen()
+        stopUpListen()
+        options.onEnd?.()
+      }
+    }
+
+    const stopMoveListen = useEventListener('pointermove', handlePointerMove)
+    const stopUpListen = useEventListener('pointerup', handlePointerUp)
```
DrJKL, line 128: I think adding the listeners within function calls might be part of how we're leaking so many.
The handlers are already ignoring events if we're not resizing, so would it be safe to just pull them out and invert some of the start/stop logic?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: Why? This should always exist.

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
christian-byrne: > adding the listeners within function calls might be part of how we're leaking so many

What does this mean exactly?

> The handlers are already ignoring events if we're not resizing, so would it be safe to just pull them out and invert some of the start/stop logic

So have the listeners active across the node lifecycle instead of just during resize? Isn't that worse?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5936/commits/d4d66597adb9e79d8d448f9e5ce65575ef2d95da

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
christian-byrne: Good catch, was leftover from before. Removed in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5936/commits/f91345127458017c5bf8f8375998205d93cb634e

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
christian-byrne: Good point, done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5936/commits/f6504f0eb333924cc487ea68e6ae69c8cf2e152b

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/5936/commits/94457260b11dbec9a6dabac12b9ed48bd39f67be

christian-byrne:
I'd like to merge this because I have two other features done that rely on it:

1. Serialize Vue node sizes to workflow
2. Mid-execution preview component which has a resize button

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: This appears to be an empty suggestion. Could you clarify what you'd like me to change here?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: There was a conflict between your console.error suggestion and @christian-byrne's view that TransformState should always exist. Since this is an internal component with guaranteed injection, I kept the current error throwing approach for fail-fast debugging.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: Implemented early return pattern in f332a8b53

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
christian-byrne: Removed unused onEnd callback option in 8b0d56bb5

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
christian-byrne: Added named Size and Position types in a1f272b2f

src/renderer/extensions/vueNodes/composables/useNodeResize.ts
christian-byrne: Analyzed this approach but the current implementation is actually optimal - listeners only exist during resize operations (seconds), avoiding unnecessary event processing. useEventListener handles cleanup properly, so no leaks occur.

simula-r (dismissed):
LGTM

PR #5939, opened 2025-10-06
fix: emit layout change for batch node bounds

christian-byrne:
## Summary

Fixes issue where node size changes are not serialized by routing DOM-driven node bounds updates through a single CRDT operation so Vue node geometry stays synchronized with LiteGraph.

## Changes

- **What**: Added `BatchUpdateBoundsOperation` to the layout store, applied it via the existing Yjs pipeline, notified link sync to recompute touched nodes, and covered the path with a regression test

## Review Focus

Correctness of the new batch operation when multiple nodes update simultaneously, especially remote replay/undo scenarios and link geometry recomputation.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5939-fix-emit-layout-change-for-batch-node-bounds-2846d73d365081db8f8cca5bf7b85308) by [Unito](https://www.unito.io)


src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -1092,6 +1099,32 @@ class LayoutStoreImpl implements LayoutStore {
     change.nodeIds.push(operation.nodeId)
   }
 
+  private handleBatchUpdateBounds(
+    operation: BatchUpdateBoundsOperation,
+    change: LayoutChange
+  ): void {
+    for (const nodeId of operation.nodeIds) {
```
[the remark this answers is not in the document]
christian-byrne: After investigating, this check is not needed. The `nodeIds` array is:
1. Always initialized as an empty array (line 1408)
2. Only assigned to the operation after validation (line 1423 checks `!nodeIds.length`)
3. Never exposed to external deserialization

The existing TypeScript type safety and single creation point provide sufficient protection. Adding the check would be inconsistent with other operation handlers which validate entity existence, not operation structure.

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -1092,6 +1099,32 @@ class LayoutStoreImpl implements LayoutStore {
...
+      const data = operation.bounds[nodeId]
+      const ynode = this.ynodes.get(nodeId)
+      if (!ynode || !data) continue
+
+      this.spatialIndex.update(nodeId, data.bounds)
```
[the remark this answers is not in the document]
christian-byrne: Implemented in 25d14525a. Added `batchUpdate()` method to SpatialIndexManager that performs all updates before invalidating the cache once, eliminating N-1 redundant cache clears for batch operations.

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -1092,6 +1099,32 @@ class LayoutStoreImpl implements LayoutStore {
...
+      ynode.set('size', {
+        width: data.bounds.width,
+        height: data.bounds.height
+      })
+      ynode.set('bounds', data.bounds)
+
+      change.nodeIds.push(nodeId)
+    }
+
+    if (change.nodeIds.length) {
```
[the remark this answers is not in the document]
christian-byrne: After review, the premise is incorrect. `change.type` is initialized to `'update'` at line 825 before the handler runs, so it's never undefined. The conditional at line 1123 is redundant (always true when nodeIds.length > 0), but this matches the pattern of other handlers where empty changes are still notified to observers.

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -1372,19 +1405,38 @@ class LayoutStoreImpl implements LayoutStore {
...
-        ynode.set('position', { x: bounds.x, y: bounds.y })
-        ynode.set('size', { width: bounds.width, height: bounds.height })
+    for (const { nodeId, bounds } of updates) {
+      const ynode = this.ynodes.get(nodeId)
+      if (!ynode) continue
+      const currentLayout = yNodeToLayout(ynode)
+
+      boundsRecord[nodeId] = {
+        bounds,
+        previousBounds: currentLayout.bounds
```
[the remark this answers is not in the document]
christian-byrne: Skipping this change to maintain architectural consistency. All 7 operation types store `previous*` values (previousPosition, previousSize, previousZIndex, previousLayout, etc.) for future undo/redo support, even though none are currently read. Removing just `previousBounds` would break this pattern. If memory is a concern, it should be addressed holistically across all operations in a separate refactoring.

src/renderer/core/layout/types.ts
```diff
@@ -184,10 +184,11 @@ interface SetNodeVisibilityOperation extends NodeOpBase {
 /**
  * Batch update operation for atomic multi-property changes
  */
-interface BatchUpdateOperation extends NodeOpBase {
-  type: 'batchUpdate'
-  updates: Partial<NodeLayout>
-  previousValues: Partial<NodeLayout>
+export interface BatchUpdateBoundsOperation extends OperationMeta {
```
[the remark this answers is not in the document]
christian-byrne: No duplicate fields exist. The `type` field narrows OperationMeta's `type: OperationType` union to the literal `'batchUpdateBounds'`—this is the standard TypeScript discriminated union pattern. BatchUpdateBoundsOperation cannot extend NodeOpBase because NodeOpBase adds `nodeId: NodeId` (singular), whereas batch operations require `nodeIds: NodeId[]` (plural). The current implementation is architecturally correct.

tests-ui/tests/renderer/core/layout/layoutStore.test.ts
```diff
@@ -176,6 +176,45 @@ describe('layoutStore CRDT operations', () => {
...
+
+    const changes: LayoutChange[] = []
+    const unsubscribe = layoutStore.onChange((change) => {
+      changes.push(change)
+    })
+
+    const newBounds = { x: 40, y: 60, width: 220, height: 120 }
+    layoutStore.batchUpdateNodeBounds([{ nodeId, bounds: newBounds }])
+
+    await new Promise((resolve) => setTimeout(resolve, 50))
```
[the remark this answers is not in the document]
christian-byrne: Implemented in 81b67b2e7. Replaced arbitrary `setTimeout(50)` with vitest's `vi.waitFor()` which polls for the expected condition. This eliminates flaky failures on slower systems while maintaining fast execution when conditions are met quickly.

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -1091,6 +1098,38 @@ class LayoutStoreImpl implements LayoutStore {
...
+      const data = operation.bounds[nodeId]
+      const ynode = this.ynodes.get(nodeId)
+      if (!ynode || !data) continue
+
+      ynode.set('position', { x: data.bounds.x, y: data.bounds.y })
+      ynode.set('size', {
+        width: data.bounds.width,
+        height: data.bounds.height
+      })
+      ynode.set('bounds', data.bounds)
```
DrJKL, line 1125: I still hate that these are separate properties

src/renderer/core/spatial/SpatialIndex.ts
```diff
@@ -55,6 +55,17 @@ export class SpatialIndexManager {
     this.invalidateCache()
   }
 
+  /**
+   * Batch update multiple nodes' bounds in the spatial index
+   * More efficient than calling update() multiple times as it only invalidates cache once
+   */
+  batchUpdate(updates: Array<{ nodeId: NodeId; bounds: Bounds }>): void {
+    for (const { nodeId, bounds } of updates) {
```
DrJKL, line 63: Does it really count as a batch if you do it for each item?

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -267,13 +268,19 @@ const handleContextMenu = (event: MouseEvent) => {
...
 onMounted(() => {
-  if (size.value && transformState?.camera) {
-    const scale = transformState.camera.z
-    const screenSize = {
-      width: size.value.width * scale,
-      height: size.value.height * scale
-    }
-    resize(screenSize)
+  // Set initial DOM size from layout store, but respect intrinsic content minimum
+  if (size.value && nodeContainerRef.value && transformState) {
```
DrJKL, line 274: nit: guard clause?

DrJKL: But also, what happens if this boat gets missed?

src/renderer/extensions/vueNodes/utils/calculateIntrinsicSize.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  // Restore original size
+  element.style.width = originalWidth
+  element.style.height = originalHeight
+
+  // Convert from screen coordinates to canvas coordinates
+  return {
+    width: intrinsicRect.width / scale,
+    height: intrinsicRect.height / scale
+  }
```
DrJKL, line 33: There's got to be a better way...

DrJKL (dismissed):
I know you said there was an issue you're still looking into, but I want this.

src/renderer/core/spatial/SpatialIndex.ts
christian-byrne: Batching because we don't rebuild the tree until after.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: What do you mean exactly?

src/renderer/extensions/vueNodes/utils/calculateIntrinsicSize.ts
christian-byrne: Well, this is used in multiple places and this PR centralizes so this is the shared location - so if we find a better way it will be very easy to swap implementations later.

PR #5949, opened 2025-10-06
refactor: Reorganize GitHub Actions for better reusability

snomiao:
## Summary

This PR refactors the GitHub Actions workflow structure to improve reusability, maintainability, and CI performance.

### New Actions
- **setup-comfyui-server**: New composite action that handles ComfyUI server setup and launch
  - Checks out ComfyUI repository
  - Installs ComfyUI_devtools custom node
  - Sets up Python environment and dependencies
  - Optionally launches the server with configurable parameters

### Refactored Actions
- **setup-frontend**: Simplified to focus only on frontend-specific tasks
  - Installs pnpm and Node.js
  - Installs dependencies
  - Optionally builds the frontend (can be skipped when using cached builds)
  - No longer handles server setup or checkout

#### tests-ci.yaml
- Introduced a setup job that builds once and caches the entire workspace
- Test jobs now restore the cached workspace instead of rebuilding
- Eliminated redundant setup steps in each test shard
- Better separation between setup and test execution phases
- Significant performance improvement through workspace caching

#### Locale Update Workflows
- Updated `update-locales.yaml` to use the new action structure
- Updated `update-locales-for-given-custom-node-repository.yaml` with proper custom node installation
- Updated `update-node-definitions-locales.yaml` to use new actions
- Removed `working-directory` references where appropriate

#### Other Workflows
- Updated `update-playwright-expectations.yaml` to use new action structure
- Consistent action usage across all workflows

## Benefits

1. **Better Performance**: Workspace caching eliminates redundant builds in CI, significantly reducing test execution time
2. **Improved Maintainability**: Clear separation of concerns makes actions easier to understand and modify
3. **Enhanced Reusability**: Actions can be composed in different ways for different workflows
4. **DRY Principle**: Eliminated code duplication across workflows
5. **Easier Debugging**: Smaller, focused actions make it easier to identify and fix issues

## Testing

- [ ] Verify tests-ci workflow runs successfully
- [ ] Verify locale update workflows function correctly
- [ ] Verify playwright expectations update workflow works
- [ ] Confirm cache/restore mechanism works as expected

## Related Issues

This refactoring addresses workflow complexity and reduces CI runtime by leveraging GitHub Actions caching more effectively.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5949-refactor-Reorganize-GitHub-Actions-for-better-reusability-2846d73d365081ae8e16f151423b5a88) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/actions/setup-comfyui-server/action.yml
+++ b/.github/actions/setup-comfyui-server/action.yml
@@ -8,7 +8,7 @@ inputs:
   launch_server:
     description: 'Whether to launch the server after setup'
     required: false
-    default: 'true'
+    default: 'false'
 runs:
   using: 'composite'
   steps:
--- a/.github/actions/setup-comfyui-server/action.yml
+++ b/.github/actions/setup-comfyui-server/action.yml
@@ -1,0 +1,50 @@
+name: Setup ComfyUI Server
+description: 'Setup ComfyUI server for continuous integration (with ComfyUI_devtools node installed)'
+inputs:
+  extra_server_params:
+    description: 'Additional parameters to pass to ComfyUI server'
+    required: false
+    default: ''
+  launch_server:
+    description: 'Whether to launch the server after setup'
+    required: false
+    default: 'true'
+runs:
+  using: 'composite'
+  steps:
+    # Note: this workflow assume frontend repo is checked out and is built in ../dist
+
+    # Checkout ComfyUI repo, install the dev_tools node and start server
+    - name: Checkout ComfyUI
+      uses: actions/checkout@v5
+      with:
+        repository: 'comfyanonymous/ComfyUI'
+        path: 'ComfyUI'
+
+    - name: Install ComfyUI_devtools from frontend repo
+      shell: bash
+      run: |
+        mkdir -p ComfyUI/custom_nodes/ComfyUI_devtools
+        cp -r ./tools/devtools/* ComfyUI/custom_nodes/ComfyUI_devtools/
+
+    - name: Setup Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+
+    - name: Install Python requirements
+      shell: bash
+      working-directory: ComfyUI
+      run: |
+        python -m pip install --upgrade pip
+        pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
+        pip install -r requirements.txt
+        pip install wait-for-it
+
+    - name: Start ComfyUI server
+      if: ${{ inputs.launch_server == 'true' }}
+      shell: bash
+      working-directory: ComfyUI
+      run: |
+        python main.py --cpu --multi-user --front-end-root ../dist ${{ inputs.extra_server_params }} &
+        wait-for-it --service 127.0.0.1:8188 -t 600
--- a/.github/actions/setup-comfyui-server/action.yml
+++ b/.github/actions/setup-comfyui-server/action.yml
@@ -25,7 +25,12 @@ runs:
       shell: bash
       run: |
         mkdir -p ComfyUI/custom_nodes/ComfyUI_devtools
-        cp -r ./tools/devtools/* ComfyUI/custom_nodes/ComfyUI_devtools/
+        if ! cp -r ./tools/devtools/* ComfyUI/custom_nodes/ComfyUI_devtools/; then
+          echo "::error::Failed to copy ComfyUI_devtools from ./tools/devtools/"
+          echo "::error::This action assumes the ComfyUI_frontend repository is checked out in the current working directory."
+          echo "::error::Please ensure you have run 'actions/checkout@v5' before calling this action."
+          exit 1
+        fi
 
     - name: Setup Python
       uses: actions/setup-python@v4
--- a/.github/actions/setup-playwright/action.yml
+++ b/.github/actions/setup-playwright/action.yml
@@ -26,15 +26,3 @@ runs:
       if: steps.cache-playwright-browsers.outputs.cache-hit == 'true'
       shell: bash
       run: pnpm exec playwright install-deps
-
-    - name: Cache tool outputs
-      uses: actions/cache@v4
-      with:
-        path: |
-          ./.cache
-          ./tsconfig.tsbuildinfo
-        key: playwright-setup-cache-${{ runner.os }}-${{ hashFiles('./pnpm-lock.yaml') }}-${{ hashFiles('./src/**/*.{ts,vue,js}', './*.config.*') }}
-        restore-keys: |
-          playwright-setup-cache-${{ runner.os }}-${{ hashFiles('./pnpm-lock.yaml') }}-
-          playwright-setup-cache-${{ runner.os }}-
-          playwright-tools-cache-${{ runner.os }}-
--- a/.github/actions/setup-playwright/action.yml
+++ b/.github/actions/setup-playwright/action.yml
@@ -6,7 +6,6 @@ runs:
     - name: Detect Playwright version
       id: detect-version
       shell: bash
-      working-directory: ComfyUI_frontend
       run: |
         PLAYWRIGHT_VERSION=$(pnpm ls @playwright/test --json | jq --raw-output '.[0].devDependencies["@playwright/test"].version')
         echo "playwright-version=$PLAYWRIGHT_VERSION" >> $GITHUB_OUTPUT
@@ -22,10 +21,20 @@ runs:
       if: steps.cache-playwright-browsers.outputs.cache-hit != 'true'
       shell: bash
       run: pnpm exec playwright install chromium --with-deps
-      working-directory: ComfyUI_frontend
 
     - name: Install Playwright Browsers (operating system dependencies)
       if: steps.cache-playwright-browsers.outputs.cache-hit == 'true'
       shell: bash
       run: pnpm exec playwright install-deps
-      working-directory: ComfyUI_frontend
+
+    - name: Cache tool outputs
+      uses: actions/cache@v4
+      with:
+        path: |
+          ./.cache
+          ./tsconfig.tsbuildinfo
+        key: playwright-setup-cache-${{ runner.os }}-${{ hashFiles('./pnpm-lock.yaml') }}-${{ hashFiles('./src/**/*.{ts,vue,js}', './*.config.*') }}
+        restore-keys: |
+          playwright-setup-cache-${{ runner.os }}-${{ hashFiles('./pnpm-lock.yaml') }}-
+          playwright-setup-cache-${{ runner.os }}-
+          playwright-tools-cache-${{ runner.os }}-
--- a/.github/workflows/tests-ci.yaml
+++ b/.github/workflows/tests-ci.yaml
@@ -152,7 +152,7 @@ jobs:
         uses: ./.github/actions/setup-playwright
 
       - name: Download blob reports
-        uses: actions/download-artifact@v4w
+        uses: actions/download-artifact@v4
         with:
           path: ./all-blob-reports
           pattern: blob-report-chromium-*
--- a/.github/workflows/tests-ci.yaml
+++ b/.github/workflows/tests-ci.yaml
@@ -13,63 +13,27 @@ jobs:
     outputs:
       cache-key: ${{ steps.cache-key.outputs.key }}
     steps:
-      - name: Checkout ComfyUI
+      - name: Checkout repository
         uses: actions/checkout@v5
+
+      # Setup Test Environment, build frontend but do not start server yet
+      - name: Setup ComfyUI server
+        uses: ./.github/actions/setup-comfyui-server
         with:
-          repository: 'comfyanonymous/ComfyUI'
-          path: 'ComfyUI'
-          ref: master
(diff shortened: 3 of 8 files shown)


.github/actions/setup-comfyui-server/action.yml
```diff
@@ -0,0 +1,50 @@
...
+    required: false
+    default: ''
+  launch_server:
+    description: 'Whether to launch the server after setup'
+    required: false
+    default: 'true'
+runs:
+  using: 'composite'
+  steps:
+    # Note: this workflow assume frontend repo is checked out and is built in ../dist
```
DrJKL, line 15: Want to add a step to validate this and fail with an actionable message otherwise?

.github/actions/setup-playwright/action.yml
```diff
@@ -22,10 +21,20 @@ runs:
...
-      working-directory: ComfyUI_frontend
\ No newline at end of file
+
+    - name: Cache tool outputs
+      uses: actions/cache@v4
+      with:
+        path: |
+          ./.cache
+          ./tsconfig.tsbuildinfo
+        key: playwright-setup-cache-${{ runner.os }}-${{ hashFiles('./pnpm-lock.yaml') }}-${{ hashFiles('./src/**/*.{ts,vue,js}', './*.config.*') }}
```
DrJKL: This would invalidate practically every run, right?

snomiao:
## Fix Applied

Fixed the failing  job. The issue was with the blob report output path.

### Problem
When removing the `working-directory: ComfyUI_frontend` from test jobs, the `PLAYWRIGHT_BLOB_OUTPUT_DIR` was still set to `../blob-report`, which would try to write outside the repo root. This caused the blob report artifacts to be empty, leading to the merge-reports job failure.

### Solution
Changed `PLAYWRIGHT_BLOB_OUTPUT_DIR` from `../blob-report` to `./blob-report` to output the reports in the repo root directory, where the artifact upload step expects them.

The workflow is now running again with this fix applied.

.github/actions/setup-comfyui-server/action.yml
snomiao: Sure! I think we can show this message when fail to cp -r from ./tools/devtools/*

snomiao: actionable message added below

.github/actions/setup-playwright/action.yml
snomiao: Indeed, I have to check what's included in .cache now

snomiao:
## Response to cache invalidation concern

I've refactored the `.cache` directory caching strategy to address this concern. Here's what changed:

### Changes Made
1. **Moved cache to `setup-frontend` action** - Now restored BEFORE build operations (not after)
2. **Removed from `setup-playwright`** - Eliminated the duplicate cache step
3. **Better cache key** - Now includes both `pnpm-lock.yaml` AND source file hashes:
   ```yaml
   key: tool-cache-${{ runner.os }}-${{ hashFiles('./pnpm-lock.yaml') }}-${{ hashFiles('./src/**/*.{ts,vue,js,mts}', './*.config.*') }}
   ```

### Why This Works Better

**Cache invalidation is now appropriate:**
- Invalidates when dependencies change (`pnpm-lock.yaml`)
- Invalidates when source files or configs change
- But NOT on every run - only when actual code changes

**What's cached:**
The `.cache` directory contains incremental outputs from:
- ESLint (`.eslintcache`)
- Prettier (`.prettierrcache`) 
- Stylelint (`.stylelintcache`)
- Knip cache
- TypeScript (`tsconfig.tsbuildinfo`)

**In `tests-ci.yaml`:**
The entire workspace (including `.cache`) is saved after the setup job completes, so test shards can restore everything at once. This means:
- Setup job: Restores tool cache, builds, then saves entire workspace
- Test jobs: Restore entire workspace (includes the already-populated `.cache`)
- No redundant individual cache saves needed

This approach is more efficient and follows GitHub Actions best practices for cache lifecycle management.

.github/actions/setup-playwright/action.yml
snomiao: Migrated to frontend build action

.github/actions/setup-comfyui-server/action.yml
```diff
@@ -0,0 +1,55 @@
...
+      uses: actions/checkout@v5
+      with:
+        repository: 'comfyanonymous/ComfyUI'
+        path: 'ComfyUI'
+
+    - name: Install ComfyUI_devtools from frontend repo
+      shell: bash
+      run: |
+        mkdir -p ComfyUI/custom_nodes/ComfyUI_devtools
+        if ! cp -r ./tools/devtools/* ComfyUI/custom_nodes/ComfyUI_devtools/; then
```
DrJKL, line 28: So this means the frontend isn't under `ComfyUI_frontend/ComfyUI_frontend/` anymore? 😁 
This makes me happy.

.github/workflows/tests-ci.yaml
```diff
@@ -151,45 +95,27 @@ jobs:
...
-
-      - name: Install requirements
-        run: |
-          python -m pip install --upgrade pip
-          pip install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
-          pip install -r requirements.txt
-          pip install wait-for-it
-        working-directory: ComfyUI
-
+          build_frontend: 'false'
```
DrJKL: Just a quick check: When we invoke this, is `build_frontend` set to false or true more often?

If it's turned off explicitly more often than left on by default, can we flip the default to make building opt-in instead of opt-out?

I usually prefer being able to find the heavier uses by searching for when options are turned on.

DrJKL:
Keep in mind, we've started moving some components of the codebase into packages, so not everything that impacts the build is under `/src` anymore
At that point, how much are we saving with the caching? How would you feel about removing that piece and seeing how it performs, then if it's a problem we can address it but with a baseline for comparison?

.github/actions/setup-comfyui-server/action.yml
snomiao: Yes!

christian-byrne:
@snomiao I think you accidentally removed the change from this PR when resolving conflicts:

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/5960

snomiao:
~Set every heavy frontend build site to pass build_frontend: 'true' so only reuse paths opt-out. Should make it easier to find build-heavy workflow invocations while keeping existing behaviour.~

.github/workflows/tests-ci.yaml
snomiao: We have 8 of `uses: ./.github/actions/setup-frontend`, 

with 3 of `build_frontend: 'false',`

so its 5 true vs 3 false

snomiao:
Renamed the composite action input from `not_build` to `skip_build` for better clarity. The parameter still defaults to 'false' so building stays the default, and all reuse paths pass `skip_build: 'true'`. Heavy workflows explicitly set `skip_build: 'false'` to keep them searchable.

snomiao:
> @snomiao I think you accidentally removed the change from this PR when resolving conflicts:
> 
> * [Devex: Add additional trigger for Playwright updates #5960](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5960)

solved!

.github/workflows/tests-ci.yaml
snomiao: renamed it into "skip_build" now. defaults to false

snomiao:
> Keep in mind, we've started moving some components of the codebase into packages, so not everything that impacts the build is under `/src` anymore At that point, how much are we saving with the caching? How would you feel about removing that piece and seeing how it performs, then if it's a problem we can address it but with a baseline for comparison?

Good Idea! I prefer make another PR to do this, keep this PR as only refactor commit :D @DrJKL 

Issue created:

- [CI: Remove ./.cache caching in github action and then watch how it performs. · Issue #5988 · Comfy-Org/ComfyUI_frontend]( https://github.com/Comfy-Org/ComfyUI_frontend/issues/5988 )

DrJKL:
> `skip_build`

I was thinking the inverse, a positive option that _does_ build and if you don't include it the caller just gets the cloning/file moving.
Side-note: It's a little weird to have the action with Build in the name but an option to not build.

snomiao:
> > `skip_build`
> 
> I was thinking the inverse, a positive option that _does_ build and if you don't include it the caller just gets the cloning/file moving. Side-note: It's a little weird to have the action with Build in the name but an option to not build.

Indeed! Just renamed it into `Setup ComfyUI frontend` so we can expect that this is a workflow prepareing everything for frontend. (and including a skip-able building process)

@DrJKL @christian-byrne WDYT

DrJKL:
> > > `skip_build`
> > 
> > 
> > I was thinking the inverse, a positive option that _does_ build and if you don't include it the caller just gets the cloning/file moving. Side-note: It's a little weird to have the action with Build in the name but an option to not build.
> 
> Indeed! Just renamed it into `Setup ComfyUI frontend` so we can expect that this is a workflow prepareing everything for frontend. (and including a skip-able building process)
> 
> @DrJKL @christian-byrne WDYT

I'd still rather it be a positive option like `include_build_step` that defaults to `false`

snomiao:
> > > > `skip_build`
> > > 
> > > 
> > > I was thinking the inverse, a positive option that _does_ build and if you don't include it the caller just gets the cloning/file moving. Side-note: It's a little weird to have the action with Build in the name but an option to not build.
> > 
> > 
> > Indeed! Just renamed it into `Setup ComfyUI frontend` so we can expect that this is a workflow prepareing everything for frontend. (and including a skip-able building process)
> > @DrJKL @christian-byrne WDYT
> 
> I'd still rather it be a positive option like `include_build_step` that defaults to `false`

done! @DrJKL

.github/actions/setup-comfyui-server/action.yml
DrJKL: That should work, but I was thinking of checking that `/dist` existed and was non-empty.

.github/actions/setup-comfyui-server/action.yml
```diff
@@ -0,0 +1,55 @@
...
+description: 'Setup ComfyUI server for continuous integration (with ComfyUI_devtools node installed)'
+inputs:
+  extra_server_params:
+    description: 'Additional parameters to pass to ComfyUI server'
+    required: false
+    default: ''
+  launch_server:
+    description: 'Whether to launch the server after setup'
+    required: false
+    default: 'true'
```
DrJKL: Can this also default to false?
The general principle is: Doing things is opt-in, not opt-out.

snomiao: sure!

.github/workflows/tests-ci.yaml
```diff
@@ -150,45 +95,25 @@ jobs:
       matrix:
         browser: [chromium-2x, chromium-0.5x, mobile-chrome]
     steps:
+      # download built frontend repo from setup job
       - name: Wait for cache propagation
         run: sleep 10
-
       - name: Restore cached setup
-        uses: actions/cache/restore@v4
+        uses: actions/cache/restore@5a3ec84eff668545956fd18022155c47e93e2684
```
DrJKL, line 102:
```suggestion
        uses: actions/cache/restore@v4
```

.github/workflows/update-node-definitions-locales.yaml
```diff
@@ -44,4 +50,4 @@ jobs:
         branch: update-locales-node-defs-${{ github.event.inputs.trigger_type }}-${{ github.run_id }}
         base: main
         labels: dependencies
-        path: ComfyUI_frontend
\ No newline at end of file
+        path: ComfyUI_frontend
```
DrJKL: Is this right now?

DrJKL (dismissed):

.github/workflows/update-node-definitions-locales.yaml
snomiao:
```suggestion
        path: .
```

snomiao: Keen eyes!

fixed!

christian-byrne:
I fixed the merge conflict

DrJKL:
Manually resolved the changes in the expectations workflow (Hope I didn't break anything...)

.github/workflows/update-node-definitions-locales.yaml
DrJKL:
```suggestion
```

DrJKL (dismissed):

DrJKL:
I'm okay merging as is and following up with any tweaks later 🙂

DrJKL (approved):

snomiao:
✅ Done! I've updated the setup-frontend action as requested:

- Changed the parameter name from `skip_build` to `include_build_step`
- Set the default to `false` (building is now opt-in)
- Updated all workflow usages:
  - Workflows that need a built frontend now explicitly set `include_build_step: 'true'`
  - Workflows that don't need building no longer pass any parameter (using the default `false`)

This makes the action behavior more intuitive as @DrJKL suggested - by default it just sets up the environment without building, and you explicitly opt into building when needed.

PR #5951, opened 2025-10-06
CI: Use main version number instead of pinning

DrJKL:
## Summary

Minimal change to use the versioned cache action instead of pinning to a specific commit.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5951-CI-Use-main-version-number-instead-of-pinning-2846d73d36508130821ffb659cb52464) by [Unito](https://www.unito.io)


snomiao (approved):
LGTM

PR #5953, opened 2025-10-07
fix mask editor bug under vueNodes

### LINKED ISSUE #5816: Vue node: MaskEditor is not loading the right image, and after editing the mask, the mask isn't saved correctly either.
https://github.com/user-attachments/assets/9098bfcd-6262-4d4c-b446-5d2a8e092883

Can we also change the mask icon? For non-English speakers, this icon would be very confusing.

Image

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-5816-Vue-node-MaskEditor-is-not-loading-the-right-image-and-after-editing-the-mask-the-m-27b6d73d365081aa9788e0b1797deece) by [Unito](https://www.unito.io)

jtydhr88:
## Summary

fix mask editor issues on vueNodes

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5953-fix-mask-editor-bug-under-vueNodes-2856d73d3650810aa8a2e1a94c4d97a6) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/ImagePreview.vue
+++ b/src/renderer/extensions/vueNodes/components/ImagePreview.vue
@@ -32,21 +32,12 @@
       <!-- Main Image -->
       <img
         v-else
+        ref="currentImageEl"
         :src="currentImageUrl"
         :alt="imageAltText"
         class="block size-full object-contain"
         @load="handleImageLoad"
         @error="handleImageError"
-      />
-
-      <!-- Hidden preloaded images for mask editor -->
-      <img
-        v-for="(url, i) in imageUrls"
-        v-show="false"
-        :key="i"
-        ref="imageEls"
-        :src="url"
-        alt=""
       />
 
       <!-- Floating Action Buttons (appear on hover) -->
@@ -154,7 +145,7 @@ const actualDimensions = ref<string | null>(null)
 const imageError = ref(false)
 const isLoading = ref(false)
 
-const imageEls = ref<HTMLImageElement[]>([])
+const currentImageEl = ref<HTMLImageElement>()
 
 // Computed values
 const currentImageUrl = computed(() => props.imageUrls[currentIndex.value])
@@ -197,11 +188,11 @@ const handleImageError = () => {
 
 // In vueNodes mode, we need to set them manually before opening the mask editor.
 const setupNodeForMaskEditor = () => {
-  if (!props.nodeId) return
+  if (!props.nodeId || !currentImageEl.value) return
   const node = app.rootGraph?.getNodeById(props.nodeId)
   if (!node) return
   node.imageIndex = currentIndex.value
-  node.imgs = imageEls.value
+  node.imgs = [currentImageEl.value]
   app.canvas?.select(node)
 }
 
--- a/src/renderer/extensions/vueNodes/components/ImagePreview.vue
+++ b/src/renderer/extensions/vueNodes/components/ImagePreview.vue
@@ -39,6 +39,16 @@
         @error="handleImageError"
       />
 
+      <!-- Hidden preloaded images for mask editor -->
+      <img
+        v-for="(url, i) in imageUrls"
+        v-show="false"
+        :key="i"
+        ref="imageEls"
+        :src="url"
+        alt=""
+      />
+
       <!-- Floating Action Buttons (appear on hover) -->
       <div v-if="isHovered" class="actions absolute top-2 right-2 flex gap-1">
         <!-- Mask/Edit Button -->
@@ -118,6 +128,7 @@ import { computed, ref, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 import { downloadFile } from '@/base/common/downloadUtil'
+import { app } from '@/scripts/app'
 import { useCommandStore } from '@/stores/commandStore'
 import { useNodeOutputStore } from '@/stores/imagePreviewStore'
 
@@ -142,6 +153,8 @@ const isHovered = ref(false)
 const actualDimensions = ref<string | null>(null)
 const imageError = ref(false)
 const isLoading = ref(false)
+
+const imageEls = ref<HTMLImageElement[]>([])
 
 // Computed values
 const currentImageUrl = computed(() => props.imageUrls[currentIndex.value])
@@ -182,7 +195,18 @@ const handleImageError = () => {
   actualDimensions.value = null
 }
 
+// In vueNodes mode, we need to set them manually before opening the mask editor.
+const setupNodeForMaskEditor = () => {
+  if (!props.nodeId) return
+  const node = app.rootGraph?.getNodeById(props.nodeId)
+  if (!node) return
+  node.imageIndex = currentIndex.value
+  node.imgs = imageEls.value
+  app.canvas?.select(node)
+}
+
 const handleEditMask = () => {
+  setupNodeForMaskEditor()
   void commandStore.execute('Comfy.MaskEditor.OpenMaskEditor')
 }
 
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -380,11 +380,15 @@ export class ComfyApp {
     const paintedIndex = selectedIndex + 1
     const combinedIndex = selectedIndex + 2
 
+    // for vueNodes mode
+    const images =
+      node.images ?? useNodeOutputStore().getNodeOutputs(node)?.images
+
     ComfyApp.clipspace = {
       widgets: widgets,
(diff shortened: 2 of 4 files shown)


comfyui-wiki:
Yes this PR fixed the https://github.com/Comfy-Org/ComfyUI_frontend/issues/5816

https://github.com/user-attachments/assets/7c3a210f-86cd-4ea1-93cb-077747725305

src/scripts/app.ts
```diff
@@ -379,11 +379,19 @@ export class ComfyApp {
     const paintedIndex = selectedIndex + 1
     const combinedIndex = selectedIndex + 2
 
+    // for vueNodes mode
+    let images = node.images
```
DrJKL: I don't like that we have `imgs` and `images` in the same scope.

DrJKL: Side-nit: Could you use a nullish coalescing operator to not need the conditional or a mutable reference here?

src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -186,8 +187,29 @@ const handleImageError = () => {
...
+      node.imageIndex = currentIndex.value
+
+      node.imgs = await Promise.all(
+        props.imageUrls.map((url) => {
+          return new Promise<HTMLImageElement>((resolve) => {
+            const img = new Image()
+
+            img.onload = () => resolve(img)
+
+            img.src = url
```
DrJKL: @christian-byrne Is this going to run into the same CORS issue you were working on fixing for Cloud?

src/scripts/app.ts
jtydhr88: fixed.

src/renderer/extensions/vueNodes/components/ImagePreview.vue
christian-byrne: Maybe, we could add just in case. See this for details:

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/5957

jtydhr88: added

AustinMroz:
Accessing nodeOutputs by nodeId instead of nodeLocatorId will cause the code to break when run inside of subgraphs.

But from looking into it, it's a decent amount of work to cleanup the surrounding code to use locatorIds instead of assuming the node resides in the active graph which is kinda out of scope. I can handle it in a separate PR after this one is merged.

christian-byrne (changes requested):
So the fundamental issue is that, when in Litegraph, both `LoadImage` and `SaveImage` have their `imgs` property assigned when their preview/output is udpated. For `SaveImage` it is set via this code https://github.com/Comfy-Org/ComfyUI_frontend/blob/3f84bb0db213d06f881c503dc3f9677d8781c114/src/composables/node/useNodeImage.ts#L101-L116

And `useNodeImage` is invoked in the draw loop here https://github.com/Comfy-Org/ComfyUI_frontend/blob/3f84bb0db213d06f881c503dc3f9677d8781c114/src/services/litegraphService.ts#L872-L896

But for `LoadImage`, I guess it is set in the `useNodeImageUpload` widget logic (which Vue nodes don't have). So, in Vue mode, only the `SaveImage` is getting `node.imgs` set. Then, the solution in this PR involves getting the image, initializing the `Image` object, then setting it manually. 

A few questions:

1. Should this logic go into its own function? Being a side-effect of `handleEditMask` is fairly confusing and I don't think future maintainers will understand the logic and what is happening
2. Do we need to create a new `Image` object? Aren't we doubling the work? Can we not just do this:

  ```vue
  <template>
    <img
      v-for="(url, i) in imageUrls"
      :key="i"
      :src="url"
      ref="imageEls"
      class="..."
    />
  </template>
  
  <script setup lang="ts">
  const imageEls = ref<HTMLImageElement[]>([])
  
  const handleEditMask = async () => {
    if (!props.nodeId) return
    const node = app.rootGraph?.getNodeById(props.nodeId)
    if (!node) return
  
    node.imageIndex = currentIndex.value
    node.imgs = imageEls.value          // ← already HTMLImageElements
    app.canvas?.selectNode(node)
    await commandStore.execute('Comfy.MaskEditor.OpenMaskEditor')
  }
  </script>
  ```

src/scripts/app.ts
```diff
@@ -410,7 +414,8 @@ export class ComfyApp {
           ComfyApp.clipspace.imgs[ComfyApp.clipspace.combinedIndex].src
       }
       if (ComfyApp.clipspace.imgs && node.imgs) {
-        if (node.images && ComfyApp.clipspace.images) {
+        // Update node.images even if it's initially undefined (vueNodes mode)
+        if (ComfyApp.clipspace.images) {
```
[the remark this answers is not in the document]
jtydhr88: we need to remove this confition as under VueNodes, node.images will be undefined and need to be updated

src/scripts/app.ts
```diff
@@ -512,6 +517,13 @@ export class ComfyApp {
       }
 
       app.graph.setDirtyCanvas(true)
+
+      if (node.images && app.nodeOutputs[node.id]) {
+        useNodeOutputStore().nodeOutputs[node.id] = {
```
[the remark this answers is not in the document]
jtydhr88: fixed

src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -43,6 +43,16 @@
...
 
+      <!-- Hidden preloaded images for mask editor -->
+      <img
+        v-for="(url, i) in imageUrls"
+        v-show="false"
+        :key="i"
+        ref="imageEls"
+        :src="url"
+        alt=""
+      />
```
christian-byrne: This feels like it might be slightly suboptimal since we are creating extra `img`

```vue
      <!-- Main Image -->
      <img
        v-else
        :src="currentImageUrl"
        :alt="imageAltText"
        class="block h-[352px] w-full object-contain"
        @load="handleImageLoad"
        @error="handleImageError"
      />
```

```vue
      <!-- Hidden preloaded images for mask editor -->
      <img
        v-for="(url, i) in imageUrls"
        v-show="false"
        :key="i"
        ref="imageEls"
        :src="url"
        alt=""
      />
```

Is the problem just that we are not loading the other images? I just tested running default workflow with batch size of 4, and when I open a mask editor from the node, it only shows the "active" image anyway, so I think we can just use the active image and not create a second `img` component. WDYT?

christian-byrne (changes requested):

src/renderer/extensions/vueNodes/components/ImagePreview.vue
jtydhr88: fixed

src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -182,7 +186,18 @@ const handleImageError = () => {
...
 }
 
+// In vueNodes mode, we need to set them manually before opening the mask editor.
+const setupNodeForMaskEditor = () => {
+  if (!props.nodeId || !currentImageEl.value) return
+  const node = app.rootGraph?.getNodeById(props.nodeId)
+  if (!node) return
+  node.imageIndex = currentIndex.value
+  node.imgs = [currentImageEl.value]
+  app.canvas?.select(node)
```
christian-byrne, line 196: This is necessary? How can the node be unselected if this is being called?

jtydhr88: image
Yes, this is necessary, please check the screenshot. In this case, even if no node is selected, the user may still see the icon to open the mask editor when hovering the mouse over the image. However, because no node is currently selected, the mask editor cannot be opened in this situation, since the mask editor requires accessing the selected node.

christian-byrne (approved):
LGTM!

christian-byrne:
In follow up (when we have time), we can do:

Add something to keep in sync with `LGraphNode.images` for compatibility generally, following some pattern similar to

```typescript
watch(
  () => props.imageUrls,
  (newUrls) => {
    const node = app.graph.getNodeById(props.nodeId)
    if (!node) return

    node.images = [currentImageEl] as string[]
  },
  { deep: true, flush: 'post' }
)
```

Then we can just remove this, which might be a little bit confusingly named currently: https://github.com/Comfy-Org/ComfyUI_frontend/blob/bee09d5310409ae15d550995296224ec4a887c29/src/stores/imagePreviewStore.ts#L334-L352

PR #5954, opened 2025-10-07
Only run failed tests for snapshot update

benceruleanlu:
Note for reviewers: Contains a lot of bash scripting

## Summary

Before, our snapshot update would run all playwright tests, even the ones that did not fail.

With this PR, we now only run the failed tests for snapshot updates.

## Changes

1. Generates a custom text manifest containing failed test_file:line, which playwright supports as a way to run that test in particular
2. Update workflow finds this latest text manifest from the GH API, using it to update snapshots

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5954-Only-run-failed-tests-for-snapshot-update-2856d73d36508138a6bbd1f22beda154) by [Unito](https://www.unito.io)


benceruleanlu:
@christian-byrne I manually wrote this PR description to be as helpful as possible, LMKWYT if you have time

DrJKL (commented):
Discussed offline, but for the record:
I can't block this, but I would strongly oppose adding such complexity before even trying simpler solutions like filtering the update runs using `--last-failed` with the results we're already storing.
I'm also worried about the mix in here of having: an inlined github-script, an inlined bash script, and a separate typescript script that collectively depend on one another by loose contract.

benceruleanlu:
> Discussed offline, but for the record: I can't block this, but I would strongly oppose adding such complexity before even trying simpler solutions like filtering the update runs using `--last-failed` with the results we're already storing. I'm also worried about the mix in here of having: an inlined github-script, an inlined bash script, and a separate typescript script that collectively depend on one another by loose contract.

  - We considered --last-failed, but it relies on Playwright’s local state from the
    previous run. Our update job runs in a fresh workflow and is sharded, so we’d have to
    persist and merge Playwright’s internal “last failed” state, which isn’t exposed via
    the reporters we already store. That ends up being more brittle and doesn’t filter to
    screenshot-only failures.
  - The current approach uses artifacts we already produce (merged JSON) and targets only
    tests that have screenshot attachments, which reduces unnecessary re-runs.
  - I agree on keeping things simple. If helpful, I can collapse the bash loop and
    manifest handoff into a single Node step that reads the merged JSON and invokes
    Playwright per project directly. That removes the extra artifact/loop while keeping
    the same precision.

.github/workflows/tests-ci.yaml
```diff
@@ -168,6 +168,26 @@ jobs:
...
+      - name: Build failed screenshot manifest
+        if: ${{ needs.playwright-tests-chromium-sharded.result == 'failure' }}
+        run: |
+          set -euo pipefail
+          if ! pnpm tsx scripts/cicd/build-failed-screenshot-manifest.ts; then
+            echo "ERROR: Failed to generate screenshot manifest"
+            echo "This may indicate an issue with the Playwright JSON report or the manifest script"
+            exit 1
+          fi
+        working-directory: ComfyUI_frontend
```
snomiao, line 180: working-directory: ComfyUI_frontend

is not necessary anymore

.github/workflows/tests-ci.yaml
```diff
@@ -168,6 +168,26 @@ jobs:
...
+
+      - name: Upload failed screenshot manifest
+        if: ${{ needs.playwright-tests-chromium-sharded.result == 'failure' }}
+        uses: actions/upload-artifact@v4
+        with:
+          name: failed-screenshot-tests
+          path: ComfyUI_frontend/ci-rerun/*.txt
```
snomiao, line 187: working-directory: ComfyUI_frontend

is not necessary anymore

scripts/cicd/build-failed-screenshot-manifest.ts
```diff
@@ -0,0 +1,87 @@
...
+import fs from 'node:fs'
+import fsp from 'node:fs/promises'
+import path from 'node:path'
+
+const argv = process.argv.slice(2)
+const getArg = (flag: string, fallback: string) => {
+  const i = argv.indexOf(flag)
+  if (i >= 0 && i + 1 < argv.length) return argv[i + 1]
+  return fallback
+}
```
snomiao, line 16: nice to have:

parse argv by minimist, and use it like `argv[flag-name] || fallback`

- [minimist - npm]( https://www.npmjs.com/package/minimist )
> var argv = require('minimist')(process.argv.slice(2));

snomiao (commented):
LGTM!

snomiao:
related: - [Shard Update Test Expectations PR by Myestery · Pull Request #6100 · Comfy-Org/ComfyUI_frontend]( https://github.com/Comfy-Org/ComfyUI_frontend/pull/6100 )

christian-byrne:
Just need to resolve conflicts and set "ready to review"

benceruleanlu:
Replaced

PR #5955, opened 2025-10-07
Lint: Add eslint import plugin

DrJKL:
## Summary

Adds the linter, turns on the recommended and a few extra rules, fixes existing violations.

Doesn't prohibit `../../...` imports yet, that'll be it's own PR.

## Changes

- **What**: Consistent and fixable imports
- **Dependencies**: The plugin and parser

## Review Focus

How do you feel about the recommended rules?
What about the extra ones?
[Any more](https://github.com/un-ts/eslint-plugin-import-x?tab=readme-ov-file#rules) you'd want to turn on?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5955-Lint-Add-eslint-import-plugin-2856d73d3650819985c0fb9ca3fa94b0) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM

To review, I simply checked important types files like litegraph augmentation, barrel exports, the index.ts used to build types package, etc. and then skimmed through everything else. I also looked at the config changes and new syntax.

PR #5961, opened 2025-10-07
fix: node header DOM hierarchy

simula-r:
## Summary

Fix: node header DOM hierarchy to position tooltip over text and constrain LOD fallback to text instead of full width. Keep node header full width to accommodate for colored background.

image
image
image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5961-fix-node-header-DOM-hierarchy-2856d73d3650814eae04ef96fae062fe) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM

PR #5973, opened 2025-10-08
Fix/vue nodes snap to grid

simula-r:
## Summary

Enable node snap to grid in vue nodes mirroring the same behavior as litegraph.

- Show node snap preview (semi transparent white box target behind node)
- Resize snap to grid
- Shift + drag / Auto snap 
- Multi select + group snap

## Changes

- **What**: useNodeSnap.ts useShifyKeySync.ts setups the core hooks into both the vue node positioning/resizing system and the event forwarding technique for communicating to litegraph.

## Review Focus

Both new composables and specifically the useNodeLayout modifications to batch the mutations when snapping.
A key tradeoff/note is why we are using the useShifyKeySync.ts which dispatches a new shift event to the canvas layer. This approach is the cleaner / more declaritive method mimicking how other vue node -> litegraph realtime events are passed.

https://github.com/user-attachments/assets/5a04dd0a-8cc2-45db-aa19-eb72e3cc416a

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5973-Fix-vue-nodes-snap-to-grid-2866d73d365081c1a058d223c8c52576) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/composables/useNodeSnap.ts
+++ b/src/renderer/extensions/vueNodes/composables/useNodeSnap.ts
@@ -1,0 +1,75 @@
+import { computed } from 'vue'
+
+import { snapPoint } from '@/lib/litegraph/src/measure'
+import { useSettingStore } from '@/platform/settings/settingStore'
+
+/**
+ * Composable for node snap-to-grid functionality
+ *
+ * Provides reactive access to snap settings and utilities for applying
+ * snap-to-grid behavior to Vue nodes during drag and resize operations.
+ */
+export function useNodeSnap() {
+  const settingStore = useSettingStore()
+
+  // Reactive snap settings
+  const gridSize = computed(() => settingStore.get('Comfy.SnapToGrid.GridSize'))
+  const alwaysSnap = computed(() => settingStore.get('pysssss.SnapToGrid'))
+
+  /**
+   * Determines if snap-to-grid should be applied based on shift key and settings
+   * @param event - The pointer event to check for shift key
+   * @returns true if snapping should be applied
+   */
+  function shouldSnap(event: PointerEvent): boolean {
+    return event.shiftKey || alwaysSnap.value
+  }
+
+  /**
+   * Applies snap-to-grid to a position
+   * @param position - Position object with x, y coordinates
+   * @returns The snapped position (mutates input and returns it)
+   */
+  function applySnapToPosition(position: { x: number; y: number }): {
+    x: number
+    y: number
+  } {
+    const size = gridSize.value
+    if (!size) return position
+
+    const posArray: [number, number] = [position.x, position.y]
+    if (snapPoint(posArray, size)) {
+      position.x = posArray[0]
+      position.y = posArray[1]
+    }
+    return position
+  }
+
+  /**
+   * Applies snap-to-grid to a size (width/height)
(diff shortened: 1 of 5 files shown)


src/renderer/extensions/vueNodes/layout/useNodeLayout.ts
```diff
@@ -144,6 +147,29 @@ export function useNodeLayout(nodeIdMaybe: MaybeRefOrGetter<string>) {
...
+          const nodeLayout = layoutStore.getNodeLayoutRef(otherNodeId).value
+          if (nodeLayout) {
+            const snappedOtherPos = applySnapToPosition({
+              ...nodeLayout.position
+            })
+            mutations.moveNode(otherNodeId, snappedOtherPos)
+          }
+        }
+      }
+    }
```
christian-byrne, line 231: Would it be possible to use this somehow - or batch some way?

https://github.com/Comfy-Org/ComfyUI_frontend/blob/874ef3ba0c85b04811258a8aafa885c8e8e9d2fc/src/renderer/core/layout/store/layoutStore.ts#L1364-L1389

src/renderer/extensions/vueNodes/composables/useNodeSnap.ts
christian-byrne, line 1: nit (can be done later): We could move these types to shared types file in layout then use in both `useNodeResize` and `useNodeSnap` https://github.com/Comfy-Org/ComfyUI_frontend/blob/8a9b3f6ea20c4244329b97e51ed282cb21815fb5/src/renderer/extensions/vueNodes/composables/useNodeResize.ts#L7-L15

src/renderer/extensions/vueNodes/composables/useShiftKeySync.ts
```diff
@@ -0,0 +1,107 @@
...
+    if (isShiftPressed === shiftKeyState.value) return
+
+    // Lazy-initialize canvas reference on first use
+    if (!canvasEl) {
+      canvasEl = app.canvas?.canvas ?? null
+      if (!canvasEl) return // Canvas not ready yet
+    }
+
+    shiftKeyState.value = isShiftPressed
+    canvasEl.dispatchEvent(
```
simula-r, line 47: This approach mimics how other events get forwarded to litegraph such as mouseWheel. I think its better scoped and declarative as opposed to registering litegraph’s event listener on a parent element and filtering out view node input/textarea targets.

src/renderer/extensions/vueNodes/composables/useShiftKeySync.ts
```diff
@@ -0,0 +1,107 @@
...
+   *   }
+   * }
+   * ```
+   */
+  function trackShiftKey(initialEvent: PointerEvent): () => void {
+    // Sync initial shift state
+    syncShiftState(initialEvent.shiftKey)
+
+    // Listen for shift key press/release during the operation
+    const handleKeyEvent = (e: KeyboardEvent) => {
```
simula-r, line 81: This lets us listen for the shift key event only when a vue node is being dragged. Also by managing its own listener we only need to use trackShiftKey() once in useNodeLayout.ts and useNodeResizer.ts and it will be auto cleaned up.

christian-byrne (approved):
LGTM

PR #5976, opened 2025-10-08
perf: avoid calling getBoundingClientRect too often in useSelectionToolboxPosition

LittleSound:
┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5976-perf-avoid-calling-getBoundingClientRect-too-often-in-useSelectionToolboxPosition-2866d73d365081639755d8087e80377c) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM

In the future we might be able to centralize these, although the performance benefit would be fairly minor.

https://github.com/Comfy-Org/ComfyUI_frontend/blob/874ef3ba0c85b04811258a8aafa885c8e8e9d2fc/src/components/graph/widgets/DomWidget.vue#L99

https://github.com/Comfy-Org/ComfyUI_frontend/blob/874ef3ba0c85b04811258a8aafa885c8e8e9d2fc/src/composables/element/useCanvasPositionConversion.ts#L16

PR #5980, opened 2025-10-08
Floating Menus - UI rework

pythongosssss:
## Summary

Enhancing and further modernizing the UI, giving users more usable area whilst keeping farmiliar positioning and feel of elements.

## Changes

- **What**: Significant restructure of the UI elements, changing elements from large blocks to floating elements, updating:
- Side toolbar menu (floating style, supports small/normal mode, combines to scroll on height overflow)
- Bottom tabs panel (floating style, tabs redesigned)
- Action bar (support for docking/undocking menu)
    - Added login/user menu button to top right
- Restyled breadcrumbs (still collapse when overflows)
- Add litegraph support for fps info position (so it isn't covered by the sidebar)

- **Breaking**: 
- Removed various elements and added new ones, I have tested custom sidebars, custom actions, etc but if scripts are inserting elements into "other" elements they may have been (re)moved.
- Remove support for bottom menu
- Remove support for 2nd-row tabs

## Screenshots 
ui

https://github.com/user-attachments/assets/571dbda5-01ec-47e8-b235-ee1b88c93dd0

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5980-Floating-Menus-UI-rework-2866d73d3650810aac60cc1afe979b60) by [Unito](https://www.unito.io)


JorgeR81:
This will be an option, right ?

JorgeR81:
I'm starting to think this is the only layout you have planned ...

I understand that this is now a popular layout and it may help to increase Comfy UI adoption.

But keep in mind that these floating layouts are normally used in much simpler applications.
The floating panels look very similar to nodes, and it may be confusing for some users.  
  
You can leave this as default, but you should offer an alternative.

Please consider an option for something like this: 

- Move the Top and Side panels to the edges to optimize space. 
- Unify the Top and Side panels into single bars

You can keep the internal panel layout, and dimensions.
And I think it still looks pretty good. 

img1

PabloWiedemann:
@JorgeR81 thanks for sharing your thoughts, really appreciate it!

You’re right, edge-aligned layouts can feel more structured. The main reason we went with floating panels was to give the canvas more room and make the UI feel lighter. In node-based workflows, the canvas is where most of the focus goes, so reducing frame weight helps a lot. We’ve also seen similar patterns in other complex and more modern creative tools.

That said, we know some users prefer a more anchored layout, so there will be an option to dock the left panel for that.

The top bar will stay floating since it mainly holds breadcrumbs, run controls, and user settings. Keeping it separate saves space and makes those frequently used actions more discoverable.

Appreciate the feedback! 🙏

JorgeR81:
> The top bar will stay floating since it mainly holds breadcrumbs, run controls, and user settings. Keeping it separate saves space and makes those frequently used actions more discoverable.

@PabloWiedemann, In that case, can you add options to hide the top floating panels ( Breadcrumbs and Queue menu ) ?

We can replace that functionally with keybinds. 

And we already have a setting to hide the other floating panel ( canvas menu ).

JorgeR81:
@PabloWiedemann, in the long term, there may be another solution.

As long as you have Top floating panels, and a Topbar above them, you'll have users wanting to drag and dock them to the Topbar.
You probably don't want to allow this, since that functionality was just removed in the v3 UI.
But, I bet this is the first thing most new users will try to do, and it will create frustration and a bad first impression, when they realise they can't do it.

So, why not integrate the functionality of the Top floating panels, in the Sidebar:

- Add a new single button to "Run" the workflow ( below the main Comfy logo ).
  * You can have another button to "Cancel" below that.
  * Alternatively, the "Run" button could change shape and function, to become the Cancel button, while the workflow is running ( this would save space in the Sidebar ).
  * The other Queue menu controls ( change "Run mode"; Batch count ; Clear pending tasks ) can be added to the Queue / gallery tab ( in the "mini header" ). 

- Add a new Sidebar tab for breadcrumbs: 
  * The subgraph levels can be shown vertically. 
  * You can remove the breadcrumb's dropdown menu, because the buttons are mostly duplicated, but you can also have some in the Sidebar's "mini header". 
  *  You can also add a small number to the new sidebar's breadcrumbs button / icon ( like in the Queue sidebar's icon ), but here it would indicate the current subgraph level. 

- Finally, add the single button to "Log in", ( maybe next to the other buttons at the bottom of the Sidebar. )

pythongosssss:
There is one failing test, this looks like a legit issue with bypassing nodes not correctly updating on the minimap but not related to these changes

PabloWiedemann:
> @PabloWiedemann, in the long term, there may be another solution.
> 
> As long as you have Top floating panels, and a Topbar above them, you'll have users wanting to drag and dock them to the Topbar. You probably don't want to allow this, since that functionality was just removed in the v3 UI. But, I bet this is the first thing most new users will try to do, and it will create frustration and a bad first impression, when they realise they can't do it.
> 
> So, why not integrate the functionality of the Top floating panels, in the Sidebar:
> 
> * Add a new single button to "Run" the workflow ( below the main Comfy logo ).
>   
>   * You can have another button to "Cancel" below that.
>   * Alternatively, the "Run" button could change shape and function, to become the Cancel button, while the workflow is running ( this would save space in the Sidebar ).
>   * The other Queue menu controls ( change "Run mode"; Batch count ; Clear pending tasks ) can be added to the Queue / gallery tab ( in the "mini header" ).
> * Add a new Sidebar tab for breadcrumbs:
>   
>   * The subgraph levels can be shown vertically.
>   * You can remove the breadcrumb's dropdown menu, because the buttons are mostly duplicated, but you can also have some in the Sidebar's "mini header".
>   * You can also add a small number to the new sidebar's breadcrumbs button / icon ( like in the Queue sidebar's icon ), but here it would indicate the current subgraph level.
> * Finally, add the single button to "Log in", ( maybe next to the other buttons at the bottom of the Sidebar. )

@JorgeR81 we actually explored quite a few options, including integrating some of these menus into the left sidebar. In the end, we decided to go with the current design, but we’ll definitely keep your suggestions in mind for future iterations. This certainly won’t be the last UI update we do 🙂

Really appreciate how much thought you’re putting into this, it’s awesome to see people care so much about Comfy and its UX!

benceruleanlu:
@codex review

christian-byrne (approved):

brucew4yn3rp:
I agree with @JorgeR81. The floating elements definitely add a modern feel to ComfyUI but the non-floating topbar seems to clash with the floating menus. Especially with the active workflow tab having the same exact text displayed as the floating button to manage the workflow, it seems redundant. (I know the floating text expands to show subgraphs when expanded)

image

I wonder if we could make the topbar disappear until the mouse hovers over that area - not dissimilar to the "Hide taskbar" setting on Windows. That way, we avoid the redundancy and extra screen space taken for 95% of the use and when a user needs to open a new workflow tab, the topbar will still be there.

Also, separate issue, but I think the login button belongs in the settings. For users that don't want to use the API nodes, it seems a bit obtrusive to have a large persistent "Log In" button that just stays at the top right of the screen.  

@PabloWiedemann @pythongosssss - thoughts?

JorgeR81:
> I wonder if we could make the topbar disappear until the mouse hovers over that area - not dissimilar to the "Hide taskbar" setting on Windows. That way, we avoid the redundancy and extra screen space taken for 95% of the use and when a user needs to open a new workflow tab, the topbar will still be there.

Yeah, this would be nice, in the near term. 

But in the long term, I think the only way to please all users ( new and advanced alike ) would be to give them the ability to "build" their own GUI, ( like we do with nodes ! ).

Basically, allowing to combine floating menus, and to dock all menus in any edge. 

I go in detail in here: https://github.com/Comfy-Org/ComfyUI_frontend/issues/6104#issuecomment-3418382999

lawlietdream:
The progress bar in comfyui-crystools is broken, now there are only unaligned numbers and percentages

brucew4yn3rp:
@JorgeR81 I threw together a [custom node](https://github.com/brucew4yn3rp/ComfyUI_FrontEndCleanup) to modify the frontend in a way that is more visually effective in my opinion. 

## Before

image


## After

image

## Optional Settings Toggles to Hide Subgraph Breadcrumbs or Progress Menu

image

PR #5983, opened 2025-10-08
[feat] Auto-remove claude-review label after CI review completes

christian-byrne:
## Summary
Automatically removes the `claude-review` label after the Claude PR review workflow completes, regardless of success or failure.

## Changes
- Added a cleanup step to `.github/workflows/claude-pr-review.yml` that removes the label using `gh pr edit --remove-label`
- Uses `if: always()` to ensure the label is removed even if the review fails
- This prevents label accumulation and allows the label to be re-applied for additional reviews

## Benefits
- Cleaner PR label management
- Labels can be re-applied to trigger additional reviews without manual cleanup
- Reduces noise in the PR interface

## Test Plan
- Apply the `claude-review` label to this PR to verify the workflow removes it automatically after completion

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5983-feat-Auto-remove-claude-review-label-after-CI-review-completes-2866d73d365081da929cd393996010e1) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/claude-pr-review.yml
+++ b/.github/workflows/claude-pr-review.yml
@@ -73,10 +73,10 @@ jobs:
         with:
           label_trigger: "claude-review"
           prompt: |
-            Read the file .claude/commands/comprehensive-pr-review.md and follow ALL the instructions exactly. 
-            
-            CRITICAL: You must post individual inline comments using the gh api commands shown in the file. 
-            DO NOT create a summary comment. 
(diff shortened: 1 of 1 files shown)


christian-byrne:
The claude-code-action validates that the workflow file in the PR exactly matches the one on the main branch. This prevents malicious PRs from modifying the workflow to do bad things.

Need to merge in order to test this.

PR #6003, opened 2025-10-09
feat: add default keybindings for changing the mask brush size

davvid-comfy:
## Summary

Make it easier to change the mask brush size. These keybindings are familiar to users of other image editors.

## Changes

- **What**: Added `[` to decrease and `]` to increase the brush size.

## Review Focus

Is this generically useful enough to include the default keybindings?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6003-feat-add-default-keybindings-for-changing-the-mask-brush-size-2876d73d365081de8832fcd4d94729f6) by [Unito](https://www.unito.io)

## References

- [GIMP brush size shortcuts](https://docs.gimp.org/3.0/en/gimp-using-variable-size-brush.html)
- [Photoshop brush size shortcuts](https://helpx.adobe.com/photoshop/using/default-keyboard-shortcuts.html)
- [Krita brush size shortcuts](https://imagy.app/shortcuts-for-krita-on-windows/#Shortcuts_for_Painting)

### DIFF
--- a/src/constants/coreKeybindings.ts
+++ b/src/constants/coreKeybindings.ts
@@ -82,6 +82,18 @@ export const CORE_KEYBINDINGS: Keybinding[] = [
     },
     commandId: 'Comfy.ShowSettingsDialog'
   },
+  {
+    combo: {
+      key: '['
+    },
+    commandId: 'Comfy.MaskEditor.BrushSize.Decrease'
+  },
+  {
+    combo: {
+      key: ']'
+    },
+    commandId: 'Comfy.MaskEditor.BrushSize.Increase'
+  },
   // For '=' both holding shift and not holding shift
   {
     combo: {


jtydhr88 (dismissed):
Tested on my local, worked as expected

christian-byrne (commented):
The mask editor has an internal Keybinding manager because we don't have a keybinding context system yet and adding a keybinding globally will trigger whether or not the mask editor is initialized or opened.

davvid-comfy:
I see what you mean about the local keybinding manager. I added a 2nd commit that handles this behavior using the scoped keyDown callback instead of with a global keybinding. If you think the end result looks better I'll squash the two commits together.

brucew4yn3rp:
Hi @davvid-comfy this feature already exists as a keybinding setting.

image

christian-byrne (dismissed):
LGTM!

davvid-comfy:
I just squashed the commits together.

@brucew4yn3rp the commit isn't here anymore because I squashed them together, but this PR was originally just an addition to the core keybindings to make this behavior preset. At least in my usage, that keybinding wasn't present by default.

What's here now is scoped behavior to just the editor and it's unconditional which is kinda nice but it does beg the question as to what happens w/ this keyDown approach if someone were to have the same hotkey set in their settings.

Maybe the original corekeybindings approach was better in that it avoids that issue, despite the concerns raised earlier? Here's a link for comparison:

https://github.com/davvid-comfy/ComfyUI_frontend/commit/200fbb7ed32cdc2d68d8b9b63f8d5efd81e7f10c

I can always swap that back in here instead if we think that's a cleaner implementation. That's the `maskeditor-keyboard-shortcuts-v0` branch in my fork.

brucew4yn3rp:
@davvid-comfy we had initially incorporated default keybindings but got [this feedback](https://github.com/Comfy-Org/ComfyUI_frontend/pull/4361#discussion_r2203060717) from the team, so I’ll defer to @webfiltered and others if we now want to incorporate default keys for brush re-sizing. 

In my opinion, the existence of the setting is more versatile and serves the purpose well.

davvid-comfy:
Thanks for the link to that feedback. (heads-up @webfiltered)

The non-us keyboard concern makes sense, but I think a better solution for that would be to allow multiple keyboard shortcuts to bind to the same action. That way a non-US shortcut could be provided as well, but.. honestly.. is that really a problem? That didn't seem to prevent this shortcut from being adopted elsewhere.

Providing it by default is really the better approach in general. All of the popular image editing programs use this same shortcut and it can be argued that it's on the same level as `ctrl-z` for undo -- it really is that ubiquitous and it's a very common operation.

Consider for a moment that it would take an average user 30 seconds to configure this, and that's assuming they already know that it exists and is configurable. Multiply that by even 10k users and we're already talking about 5k wasted minutes collectively.

Those are user-side arguments in favor of it, at least. It's not a hill I'm going to die on, though, so if you don't want it let's close this PR.

brucew4yn3rp:
I appreciate your preference for the `[` and `]` keybindings. As mentioned, I’d defer to @webfiltered since we removed the default assignment based on earlier feedback. 

That said, perhaps the right approach isn’t to hardcode a message broker command, but instead to have the keybinding settings preset to `[` and `]` for brush size decrease and increase (respectively). This would mitigate the concern of every user having to set a somewhat standard keybinding but would still allow the versatility for users who want to set non-standard bindings.

davvid-comfy:
> That said, perhaps the right approach isn’t to hardcode a message broker command, but instead to have the keybinding settings preset to [ and ] for brush size decrease and increase (respectively). This would mitigate the concern of every user having to set a somewhat standard keybinding but would still allow the versatility for users who want to set non-standard bindings.

Makes sense. I agree. That's effectively what I had in the original version of this PR (where it just edited the core keybindings and nothing else) so I've restored those changes.

Users are able to override these presets so this PR works the way you describe now. cheers

christian-byrne:
Closing this PR — thanks for the contribution @davvid-comfy, and for being patient while we discussed this!

**Why we're closing:**
- The team previously decided in #4361 to not include default brush size keybindings, based on concerns about non-US keyboard layouts where `[` and `]` are hard to reach (German, French, etc.)
- The keybindings would be globally scoped (no `targetElementId` limiting them to the mask editor), which could interfere with other interactions
- The consensus leaned toward keeping these as user-configurable only — and the commands (`Comfy.MaskEditor.BrushSize.Increase/Decrease`) already work perfectly via the keybinding settings UI
- The target file has since been relocated (`src/constants/coreKeybindings.ts` → `src/platform/keybindings/defaults.ts`)

Users who want `[`/`]` for brush size can easily add them in keybinding settings. Thanks again! 🙏

PR #6006, opened 2025-10-10
Implement DOMWidget for vue

AustinMroz:
vue-dom-widget

Did testing on about a dozen custom nodes. Most just work.
- Some custom nodes have copy/pasted the `addDOMWidget` call with types like `customtext` and get converted to textareas -> Not feasible to fix here. Can open PRs into custom nodes if complaints arise.
- Only the KJNodes spline editor had mouse issues -> Can investigate/open PR into KJNodes later.
- Many nodes don't resize gracefully. Probably best handled in a future PR.
  - Some expect to be handled like textareas. These currently have minsize and don't scale.
  - Others, like VHS previews, scale self properly, but don't update height inside a drag operation -> node height can be set to less than fit.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6006-Implement-DOMWidget-for-vue-2886d73d3650817ca497c15d87d70f4f) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM! Great start, clean implementation.

PR #6018, opened 2025-10-11
fix Vue node border styles in different states (executing, error, selected)

christian-byrne:
- Use exact tokens from Figma
- Fix issue in which node is stuck in `executing` state after it errors

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6018-fix-Vue-node-border-styles-in-different-states-executing-error-selected-2896d73d365081f39000fc3e42811f0d) by [Unito](https://www.unito.io)


christian-byrne:
Going to add a test for Vue nodes that error

christian-byrne:
Merging as future PRs will depend on new tokens.

PR #6019, opened 2025-10-11
Make nodeData.widgets reactive

AustinMroz:
Makes the litegraph `node.widgets` array `shallowReactive` and makes the `nodeData.widgets` a `reactiveComputed` derived from the litegraph widget data.
reactive-widgets

Making changes to the structure of litegraph items is somewhat dangerous, but code search verifies that there are no custom nodes using `defineProperty` on `node.widgets`

This fixes display of promoted widgets on subgraph node and any custom nodes that dynamically add or remove widgets.

TODO:
- Investigate occasional dropped widgets.
  - Some of this was confusion with `canvasOnly` widgets and widgets not implemented in vue. Will keep investigating, but I'm not terribly concerned with actual test cases and it being an objective improvement.
  
Known Issue:
- Node does not grow/shrink to fit changed widgets

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6019-Make-nodeData-widgets-reactive-2896d73d3650815691b6ee370a86a22c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -2,13 +2,15 @@
  * Vue node lifecycle management for LiteGraph integration
  * Provides event-driven reactivity with performance optimizations
  */
-import { reactive } from 'vue'
+import { reactiveComputed } from '@vueuse/core'
+import { reactive, shallowReactive } from 'vue'
 
 import { useChainCallback } from '@/composables/functional/useChainCallback'
 import type {
   INodeInputSlot,
   INodeOutputSlot
 } from '@/lib/litegraph/src/interfaces'
+import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets'
 import { useLayoutMutations } from '@/renderer/core/layout/operations/layoutMutations'
 import { LayoutSource } from '@/renderer/core/layout/types'
 import type { InputSpec } from '@/schemas/nodeDef/nodeDefSchemaV2'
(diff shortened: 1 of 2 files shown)


src/composables/graph/useGraphNodeManager.ts
```diff
@@ -132,44 +134,57 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
...
-          type: widget.type || 'text',
-          value: undefined
-        }
+    const reactiveWidgets = shallowReactive<IBaseWidget[]>(node.widgets ?? [])
+    Object.defineProperty(node, 'widgets', {
+      get() {
+        return reactiveWidgets
+      },
+      set(v) {
+        reactiveWidgets.splice(0, reactiveWidgets.length, ...v)
```
christian-byrne, line 143: nit: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/Proxy/set#return_value

christian-byrne (approved):
Nice, LGTM!

src/composables/graph/useGraphNodeManager.ts
AustinMroz: `defineProperty` uses different handlers than proxy. Their call signature is different (`t,p,v,r` vs `v`)  and the proxy handler will throw very visible errors if an incorrect return value is given. See the [example on MDN for defining a set handler](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty#creating_a_property)

christian-byrne: Oh I see, good to know - thank you!

PR #6022, opened 2025-10-11
fix Vue node opacity conditions (user node opacity, bypass state, muted state)

christian-byrne:
## Summary

Fixed Vue node opacity calculation to properly combine global opacity setting with muted/bypassed state opacity.

**Root Cause**: When global opacity setting was added as inline style (481aa8252), it began overriding CSS `opacity-50` classes due to higher specificity.

**Solution**: Modified `nodeOpacity` computed property to calculate effective opacity as `globalOpacity * 0.5` for muted/bypassed states, removing conflicting CSS classes.

## Changes

- **What**: Fixed [CSS specificity conflict](https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity) where inline `opacity` style overrode `opacity-50` classes for muted/bypassed nodes
- **Breaking**: None - restores intended opacity behavior

## Review Focus

Multiplicative opacity calculation ensuring muted/bypassed nodes apply 0.5 opacity on top of global opacity setting rather than being overridden by it.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6022-fix-Vue-node-opacity-conditions-user-node-opacity-bypass-state-muted-state-2896d73d365081c290f1da37c195c2f5) by [Unito](https://www.unito.io)


christian-byrne:
Need to update the "muted" playwright tests to not rely on checking classnames. Can check style or just use a screenshot, I think the latter is better as it also ensures the css is working as expected.

PR #6030, opened 2025-10-12
[fix] Improve cloud onboarding UX with email verification polling and signup flow

viva-jinyi:
## Summary
Improve cloud onboarding flow by adding email verification polling and fixing signup auto-logout issue.

## Changes
- **What**: 
  - Add polling mechanism to automatically check email verification status every 5 seconds on the verify email page
  - Fix auto-logout issue after signup by redirecting to root instead of login page  
  - Remove automatic logout on login page mount to preserve user session after signup
- **Breaking**: None
- **Dependencies**: None

## Review Focus
- Email verification polling implementation uses Firebase Auth's `reload()` method to refresh user state
- Polling stops after 5 minutes to prevent indefinite resource usage
- Proper cleanup of intervals/timeouts in `onUnmounted` hook to prevent memory leaks
- Signup flow now maintains user session instead of forcing re-login

## User Experience Improvements
1. **Before**: Users had to manually refresh the page after clicking email verification link
   **After**: Page automatically detects verification and proceeds to next step

2. **Before**: After signup, users were redirected to login page and automatically logged out, requiring them to sign in again
   **After**: Users stay logged in after signup and are redirected to root, maintaining their session

## Testing
- Tested email verification polling with both verified and unverified states
- Verified that invite codes are preserved throughout the flow
- Confirmed no memory leaks from polling intervals
- Tested signup flow with email/password authentication

## Related Issues
- Resolves user complaints about having to sign in twice during signup flow
- Addresses email verification page not auto-advancing after verification

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6030-fix-Improve-cloud-onboarding-UX-with-email-verification-polling-and-signup-flow-28a6d73d365081be8020caee6337c3e7) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/onboarding/cloud/CloudVerifyEmailView.vue
+++ b/src/platform/onboarding/cloud/CloudVerifyEmailView.vue
@@ -35,18 +35,53 @@
 </template>
 
 <script setup lang="ts">
-import { onMounted } from 'vue'
+import { onMounted, onUnmounted, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { useRoute, useRouter } from 'vue-router'
+import { useFirebaseAuth } from 'vuefire'
 
 import { useToastStore } from '@/platform/updates/common/toastStore'
 import { useFirebaseAuthStore } from '@/stores/firebaseAuthStore'
 
 const authStore = useFirebaseAuthStore()
+const auth = useFirebaseAuth()!
 
 const router = useRouter()
 const route = useRoute()
 const { t } = useI18n()
+
+let intervalId: number | null = null
+let timeoutId: number | null = null
+const redirectInProgress = ref(false)
+
+function clearPolling(): void {
+  if (intervalId !== null) {
+    clearInterval(intervalId)
+    intervalId = null
+  }
+  if (timeoutId !== null) {
+    clearTimeout(timeoutId)
+    timeoutId = null
+  }
+}
+
+async function redirectToNextStep(): Promise<void> {
+  if (redirectInProgress.value) return
+
+  redirectInProgress.value = true
+  clearPolling()
+
+  const inviteCode = route.query.inviteCode as string | undefined
+
+  if (inviteCode) {
+    await router.push({
+      name: 'cloud-invite-check',
+      query: { inviteCode }
+    })
+  } else {
+    await router.push({ name: 'cloud-user-check' })
+  }
+}
 
 const goBack = async () => {
   const inviteCode = route.query.inviteCode as string | undefined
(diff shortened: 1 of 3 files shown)


viva-jinyi:
https://github.com/Comfy-Org/website/blob/3583ca8f17cb57ed2000899246b5aabd8148d939/pages/verify-email.vue#L20-L33

src/platform/onboarding/cloud/CloudVerifyEmailView.vue
```diff
@@ -86,23 +121,35 @@ async function onSend() {
...
-  } else {
-    await onSend()
+    return redirectToNextStep()
   }
+
+  // Send initial verification email
+  await onSend()
+
+  // Start polling to check email verification status
+  intervalId = window.setInterval(async () => {
```
arjansingh, line 134: We can probably do this with `useTimeoutPoll`: https://vueuse.org/core/useTimeoutPoll/

There's vue-use functions for handling intervals and timeouts directly if `useTimeoutPoll` doesn't work.

src/platform/onboarding/cloud/CloudVerifyEmailView.vue
```diff
@@ -86,23 +121,35 @@ async function onSend() {
...
+    if (auth.currentUser && !redirectInProgress.value) {
```
arjansingh, line 135: Non-blocker: `useCurrentUser`, could simplify a little further, but not necessary. https://vuefire.vuejs.org/api/modules/vuefire.html#useCurrentUser

```ts
  import { useCurrentUser } from 'vuefire'
  import { useTimeoutPoll } from '@vueuse/core'

  const user = useCurrentUser()

  const { pause } = useTimeoutPoll(
    async () => {
      if (user.value && !redirectInProgress.value) {
        await user.value.reload()
        if (user.value.emailVerified) {
          pause()
          await redirectToNextStep()
        }
      }
    },
    5000,
    { timeout: 5 * 60 * 1000 }
  )
```

src/platform/onboarding/cloud/CloudVerifyEmailView.vue
```diff
@@ -35,19 +35,54 @@
...
+function clearPolling(): void {
+  if (intervalId !== null) {
+    clearInterval(intervalId)
+    intervalId = null
+  }
+  if (timeoutId !== null) {
+    clearTimeout(timeoutId)
+    timeoutId = null
+  }
+}
```
arjansingh, line 66: You don't need the `null` checks.

`setTimeout` and `setInterval` return `number | undefined`.

You could do something like:

```ts
let intervalId: number | undefined
let timeoutId: number | undefined

function clearPolling(): void {
  clearInterval(intervalId)
  clearTimeout(timeoutId)
  intervalId = undefined
  timeoutId = undefined
}
```

But more important: See my comments about the vue-use methods we should be using instead.

src/platform/onboarding/cloud/CloudVerifyEmailView.vue
```diff
@@ -35,19 +35,54 @@
...
+
+async function redirectToNextStep(): Promise<void> {
+  if (redirectInProgress.value) return
+
+  redirectInProgress.value = true
+  clearPolling()
+
+  const inviteCode = route.query.inviteCode as string | undefined
+
+  if (inviteCode) {
```
arjansingh, line 76: I don't think that type assertion protects us at runtime.

Eventually, the types resolve to:
```ts
route.query: Record<string, string | null | (string | null)[]>
route.query.inviteCode: string | null | (string | null)[] | undefined
```

So `route.query` could still return an array at runtime and we'd have a runtime error.

This code would protect, and then you could remove the assertion.

```ts
const inviteCode = route.query.inviteCode
if (inviteCode && !(Array.isArray(inviteCode) && inviteCode.length === 0)) {
  // continue
}
```

src/platform/onboarding/cloud/CloudVerifyEmailView.vue
arjansingh:
```ts
  import { useTimeoutPoll } from '@vueuse/core'

  const { pause } = useTimeoutPoll(
    async () => {
      if (auth.currentUser && !redirectInProgress.value) {
        await auth.currentUser.reload()
        if (auth.currentUser?.emailVerified) {
          pause() // Stop polling
          await redirectToNextStep()
        }
      }
    },
    5000, // interval: 5 seconds
    { timeout: 5 * 60 * 1000 } // max duration: 5 minutes
  )
```

something like that could work.

arjansingh (changes requested):
We could do this a lot more simply using: https://vueuse.org/core/useTimeoutPoll/

arjansingh (approved):
Talked it over with @christian-byrne consider these follow-ons. Not gonna block merge on this.

PR #6034, opened 2025-10-12
Add "control after generate" widget variant in Vue nodes system

christian-byrne:
## Summary

Adds revamped system for widgets that have control after/before generate variant on them.

https://github.com/user-attachments/assets/3d1ea6e3-b728-49c2-9f46-dff783ee7b8d

## TODO

- [ ] fix step value being 10 (legacy system)
- [ ] ensure it works with COMBO
- [ ] ensure it works with FLOAT
- [ ] ensure it works with primitive nodes
- [ ] make "random option" selected by default
- [ ] either implement or remove the config button functionality - think it should open settings?
- [ ] align popover UI with design exactly

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6034-Add-control-after-generate-widget-variant-in-Vue-nodes-system-28a6d73d365081878d4cc83e980a9578) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -17,8 +17,8 @@ import type { InputSpec } from '@/schemas/nodeDef/nodeDefSchemaV2'
 import { isDOMWidget } from '@/scripts/domWidget'
 import { useNodeDefStore } from '@/stores/nodeDefStore'
 import type { WidgetValue } from '@/types/simplifiedWidget'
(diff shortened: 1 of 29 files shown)


hartmark:
Looking good, what's the status of this PR?

Is it that if none of the options are checked the seed is a fixed value?

christian-byrne:
Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/6985

PR #6036, opened 2025-10-13
fix FLOAT widget incrementing broken & disabled state styles on widget number input (Vue)

christian-byrne:
## Summary

Align Vue node number widgets with Figma by centralising button styling and surfacing disabled-state tokens in the design system.

## Changes

- **What**: Added shared [`useNumberWidgetButtonPt`](src/renderer/extensions/vueNodes/widgets/composables/useNumberWidgetButtonPt.ts) helper so both PrimeVue `InputNumber` widgets reuse the same Tailwind token classes, and added the `color/tokens/alpha` values in [`packages/design-system/src/css/style.css`](packages/design-system/src/css/style.css#L89-L212) so semantic aliases remain token-driven ([PrimeVue passthrough docs](https://www.primefaces.org/primevue/passthrough) w [`color-mix`](https://www.w3.org/TR/css-color-5/#color-mix))

## Review Focus

Confirm hover/active/disabled colours match the design tokens in both light and dark themes and that float precision still respects the safe-range guard.

Screenshot from 2025-10-12 17-53-23

Screenshot from 2025-10-12 17-53-32

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6036-fix-disabled-state-styles-on-Vue-widget-number-input-INT-and-FLOAT-widgets-28b6d73d365081f8aef7fa860b641f7d) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
```diff
@@ -65,17 +66,24 @@ const useGrouping = computed(() => {
 
 // Check if increment/decrement buttons should be disabled due to precision limits
 const buttonsDisabled = computed(() => {
-  const currentValue = localValue.value || 0
-  return !Number.isSafeInteger(currentValue)
```
christian-byrne, line 69: This was bugged, as this component is also used for FLOAT widgets.

christian-byrne:
Verified that all changes are expected snapshot drift.

PR #6055, opened 2025-10-14
allow copy and paste when using minimap or canvas menu (bottom right)

christian-byrne:
## Summary

Fixed intermittent copy/paste failures when clicking minimap or canvas control buttons by expanding target whitelist in document-level clipboard handlers.

## Changes

- **What**: Expanded `isTargetInGraph` check in `useCopy.ts` and `usePaste.ts` to include minimap and canvas controls
- **Added IDs**:
  - `comfy-minimap` on minimap container (`src/renderer/extensions/minimap/MiniMap.vue`)
  - `graph-canvas-controls` on canvas controls ButtonGroup (`src/components/graph/GraphCanvasMenu.vue`)
- **Test Infrastructure**: Created `Minimap` test fixture following existing pattern (Topbar, SidebarTab)

## Technical Details

Document-level clipboard handlers (added Feb 2025) replace LiteGraph's canvas-only handlers but require event target to be within graph workspace. Previous implementation only checked:

```typescript
e.target.id === 'graph-canvas' ||
e.target.classList.contains('litegraph') ||
e.target.classList.contains('graph-canvas-container')
```

Clicking UI elements (minimap buttons, zoom controls) caused focus loss, making target check fail silently.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6055-allow-copy-and-paste-when-using-minimap-or-canvas-menu-bottom-right-28c6d73d365081529049e2b1c993cd53) by [Unito](https://www.unito.io)


src/composables/usePaste.ts
```diff
@@ -38,11 +38,17 @@ export const usePaste = () => {
   }
 
   useEventListener(document, 'paste', async (e) => {
+    // Check if target is graph canvas or within graph UI (minimap, controls, etc.)
     const isTargetInGraph =
       e.target instanceof Element &&
-      (e.target.classList.contains('litegraph') ||
+      (e.target.id === 'graph-canvas' ||
+        e.target.id === 'comfy-minimap' ||
```
Myestery: suggestion: we might be able to reuse the code from the useCopy. The conditions are many so its a good idea to reuse

christian-byrne: Extracted to shared helper in src/workbench/eventHelpers.ts in a059806

AustinMroz:
I've been exploring similar options over in #6087.

Thoughts on broadening the target restrictions to `e.target instanceof HTMLTextAreaElement || e.target instanceof HTMLInputElement`? IIRC the only reason we don't grab *every* clipboard event was that supporting undo/redo for text inputs was non-feasible. Are there any cases other than text inputs where we don't want our copy/paste handler to apply?

Edit to clarify: The changes here are an improvement and LGTM, even if further changes will be made down the line.

benceruleanlu:
@codex review

src/composables/useCopy.ts
```diff
@@ -20,14 +21,14 @@ export const useCopy = () => {
...
+    if (!isEventTargetInGraph(e.target)) {
+      return
+    }
 
     // copy nodes and clear clipboard
     const canvas = canvasStore.canvas
-    if (isTargetInGraph && canvas?.selectedItems) {
+    if (canvas?.selectedItems) {
       canvas.copyToClipboard()
       // clearData doesn't remove images from clipboard
```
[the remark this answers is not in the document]
benceruleanlu: technically true, but not impactful yet because I think primevue for input number uses type=text under the hood

benceruleanlu (dismissed):
LGTM, any ideas on a more robust solution that won't need id upkeep to prevent drifting?

christian-byrne:
> LGTM, any ideas on a more robust solution that won't need id upkeep to prevent drifting?

Nothing comes to mind. Is ID drifting common? Would be pretty bad to change IDs and not audit all usage first.

Thankfully since I added tests, we will be able to spot that case immediately when the test fails.

christian-byrne:
Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/6087

PR #5292, opened 2025-09-01
Rework desktop install / startup UX

webfiltered:
### Summary

Complete redesign of Desktop app installer, implementing the new app startup progress system and error reporting.

### Review Focus

I would like to put some limitations on the review scope, if that is acceptable.  This PR explicitly _does not_ attempt to refactor / fix existing bad logic that is simply being moved or added to.  The project identifies it as explicitly out of scope.

Some code has been moved between files (e.g. mirror config), looking like it is new, but really is just refactored enough to continue functioning exactly as it did before, within the new design.

Adding nits to every apparent change would take an excessive amount of time, and produce a lot of noise that could be addressed outside of this PR.

#### 3: Online Behaviour

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5292-Rework-desktop-install-startup-UX-2616d73d36508132b0e2c1f3e154459c) by [Unito](https://www.unito.io)

PR #6056, opened 2025-10-14
fix terminal style

christian-byrne:
## Summary

After #5292, at certain browser zoom levels, the xterm terminal did not fill horizontal space properly, showing a gap with mismatched background colors (`#171717` viewport vs `black` terminal content).

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/6049

## Problem

The hardcoded `#171717` terminal theme background only affected the xterm viewport, while terminal content remained black, causing a visible color mismatch at low zoom levels where the gap was more apparent.

Fixed by keeping original theme when not on desktop distribution.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6056-fix-terminal-style-28c6d73d36508132b521e5767f41d540) by [Unito](https://www.unito.io)


AustinMroz (commented):
LGTM

PR #6061, opened 2025-10-14
[style] prevent Vue node selection outline being obscured by image output

christian-byrne:
## Summary

Adds `px-2` on image to prevent this issue (below) - I think there's a better solution but I'm not really sure what it is. We use outline for selection state and it's somewhat complex how our ring/border/outline with many different node states and interactions works right now. It will take some CSS skill to allow the images to be totally flush.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6061-style-prevent-Vue-node-selection-outline-being-obscured-by-image-output-28c6d73d365081d59b34d8f91252de92) by [Unito](https://www.unito.io)


AustinMroz (dismissed):
I've been toying with similar code sections to better handle the sizing of DOMWidgets.

Thoughts on swapping calculateIntrinsicSize to `min-content`? From quick testing, it allows for significant cleanup to the image preview styling here.

christian-byrne:
> Thoughts on swapping calculateIntrinsicSize to min-content? From quick testing, it allows for significant cleanup to the image preview styling here.

We can do it, but it has the downside of then allowing you to resize the node too much (hard to explain). That being said, I think there might be better solutions for that - since this `calculateIntrinsicSize` and hardcoding the node height/width is incuring a large cost, what do you think?

DrJKL:
I think this is a fine short-term improvement until we are able to work out the sizing-on-resizing. Want to add an `After` to the description?

DrJKL (dismissed):

DrJKL:
Want to do the same for the Video Preview?

DrJKL (dismissed):
Video next?

christian-byrne:
> Video next?

Will do, good call.

PR #6063, opened 2025-10-15
fix Cloudbadge

Myestery:
This pull request refactors how top bar badges are handled in the application, making badge rendering extensible and moving cloud badge logic into the extension system. The main changes include replacing the old `CloudBetaBadge` component with a new, generic badge system, introducing a Pinia store for badge management, and updating the extension API to support top bar badges.

**Badge System Refactor and Extensibility**

* Replaced the hardcoded `CloudBetaBadge` with a new `TopbarBadges` component, which dynamically renders badges from the store instead of relying on the `isCloud` flag in `TopMenubar.vue`. [[1]](diffhunk://#diff-b7d7bf1028f09fb907c09edf27631214d005c93b80eaff7cf15cfd53671b1e8aL9-R9) [[2]](diffhunk://#diff-b7d7bf1028f09fb907c09edf27631214d005c93b80eaff7cf15cfd53671b1e8aL43-R48)

* Renamed and refactored `CloudBetaBadge.vue` to `TopbarBadge.vue`, making it accept a generic `badge` prop and removing i18n logic from the component.

* Added a new `TopbarBadges.vue` component to render all badges from the `topbarBadgeStore`.

**Badge Data Management**

* Introduced a new Pinia store `topbarBadgeStore` that aggregates top bar badges from all extensions, enabling dynamic badge management.

**Extension System Integration**

* Updated the extension API (`ComfyExtension` interface) to support a new `topbarBadges` property, allowing extensions to contribute badges to the top bar.

* Added a core extension (`cloudBadge.ts`) that registers a "Comfy Cloud" beta badge when running in a cloud environment, using the new badge system. [[1]](diffhunk://#diff-b7818ca9daae2411d56695777160b8132507f2a3ff4f700d2510453c8833ca75R1-R16) [[2]](diffhunk://#diff-236993d9e4213efe96d267c75c3292d32b93aa4dd6c3318d26a397e0ae56bc87R2)

**Type Definitions**

* Added a new `TopbarBadge` type to `comfy.ts` to define the structure for top bar badges, supporting optional labels.

PR #6064, opened 2025-10-15
subscription page

jtydhr88:
Summary

  Implements cloud subscription management UI and flow for ComfyUI Cloud users.

  Core Features:
  - Subscription Status Tracking: Global reactive state management for subscription status across all components
  using shared subscriptionStatus ref
  - Subscribe to Run Button: Replaces the Run button in the actionbar with a "Subscribe to Run" button for users
  without active subscriptions
  - Subscription Required Dialog: Modal dialog with subscription benefits, pricing, and checkout flow with video
  background
  - Subscription Settings Panel: New settings panel showing subscription status, renewal date, and quick access to
  billing management
  - Auto-detection & Polling: Automatically polls subscription status after checkout completion and syncs state
  across the application

https://github.com/user-attachments/assets/f41b8e6a-5845-48a7-8169-3a6fc0d2e5c8

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6064-subscription-page-28d6d73d36508135a2a0fe7c94b40852) by [Unito](https://www.unito.io)


src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
```diff
@@ -0,0 +1,125 @@
...
+}
+
+const stopPolling = () => {
+  if (pollInterval) {
+    clearInterval(pollInterval)
+    pollInterval = null
+  }
+  isPolling.value = false
+  isLoading.value = false
+}
```
arjansingh: ~I think [useTimeoutPoll](https://vueuse.org/core/useTimeoutPoll/#usetimeoutpoll) could handle most of this. Less code for us to write.~

I thought about this more... I don't actually think useTimeoutPoll is clearer.

src/stores/firebaseAuthStore.ts
```diff
@@ -367,6 +386,75 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
     return response.json()
   }
 
+  /**
+   * Fetch the current cloud subscription status for the authenticated user
+   * @returns Subscription status or null if no subscription exists
+   */
+  const fetchSubscriptionStatus =
```
arjansingh: The subscription logic might be better of in it's own dedicated store.

src/components/dialog/content/setting/SubscriptionPanel.vue
```diff
@@ -0,0 +1,285 @@
...
+  const date = new Date(authStore.subscriptionStatus.renewal_date)
+  return date.toLocaleDateString('en-US', {
+    month: 'short',
+    day: 'numeric',
+    year: 'numeric'
+  })
+})
+
+const latestEvents = ref<AuditLog[]>([])
+const isLoadingEvents = ref(false)
```
arjansingh: suggestion: if we moved some of this logic into a composable we could probably break this up into several smaller components.

src/App.vue
```diff
@@ -41,6 +48,32 @@ const showContextMenu = (event: MouseEvent) => {
   }
 }
 
+const checkSubscriptionStatus = async () => {
+  if (!isCloud) return
+
+  if (!isLoggedIn.value) return
+
+  try {
+    const status = await authActions.fetchSubscriptionStatus()
```
arjansingh: Another reason to break out the subscription work, we can do what @Myestery in https://github.com/Comfy-Org/ComfyUI_frontend/pull/6063 an dynamically load the subscription code and components only when we need them.

src/components/actionbar/ComfyActionbar.vue
```diff
@@ -33,15 +34,28 @@ import Panel from 'primevue/panel'
 import type { Ref } from 'vue'
 import { computed, inject, nextTick, onMounted, ref, watch } from 'vue'
 
+import { isCloud } from '@/platform/distribution/types'
 import { useSettingStore } from '@/platform/settings/settingStore'
+import { useFirebaseAuthStore } from '@/stores/firebaseAuthStore'
 import { cn } from '@/utils/tailwindUtil'
 
 import ComfyQueueButton from './ComfyQueueButton.vue'
+import SubscribeToRun from './SubscribeToRun.vue'
```
arjansingh: this is another good example of where we could dynamically import the component.

arjansingh (commented):
Overall this is a very good draft. 

I made a couple of comments on how and where we can be dynamically importing off of `isCloud`.

That and moving the subscription stuff out of firebase and into it's own store are the big comments.

src/components/dialog/content/setting/SubscriptionPanel.vue
```diff
@@ -0,0 +1,285 @@
...
+        </div>
+      </div>
+
+      <div v-if="hasActiveSubscription" class="grow overflow-auto">
+        <div class="rounded-lg border border-charcoal-400 p-4">
+          <div>
+            <div class="flex items-center justify-between">
+              <div>
+                <div class="flex items-baseline gap-1">
+                  <span class="text-2xl font-bold">$20.00</span>
```
arjansingh: Can we make this number a `const` so it's easy to update if we need to?

src/stores/firebaseAuthStore.ts
arjansingh: @jtydhr88 we can't put this in a subscription store that we can dynamically import?

src/services/dialogService.ts
```diff
@@ -485,6 +486,22 @@ export const useDialogService = () => {
...
+      dialogComponentProps: {
+        closable: true,
+        style: 'width: 720px; height: 434px;',
+        pt: {
+          content: {
+            class: 'overflow-hidden p-0!'
+          }
+        }
+      }
+    })
```
arjansingh, line 518:
```suggestion
  if(isCloud) {
    dialogStore.showDialog({
      key: 'subscription-required',
      component: SubscriptionRequiredDialogContent,
      dialogComponentProps: {
        closable: true,
        style: 'width: 720px; height: 434px;',
        pt: {
          content: {
            class: 'overflow-hidden p-0!'
          }
        }
      }
    })
  }
```

@christian-byrne would we handle it like this so it compiles to a noop? Or should we actually create a cloud dialog service?

src/services/dialogService.ts
```diff
@@ -8,6 +8,7 @@ import LoadWorkflowWarning from '@/components/dialog/content/LoadWorkflowWarning
 import MissingModelsWarning from '@/components/dialog/content/MissingModelsWarning.vue'
 import PromptDialogContent from '@/components/dialog/content/PromptDialogContent.vue'
 import SignInContent from '@/components/dialog/content/SignInContent.vue'
+import SubscriptionRequiredDialogContent from '@/components/dialog/content/SubscriptionRequiredDialogContent.vue'
```
arjansingh: This should be a dynamic import.

src/platform/settings/composables/useSettingUI.ts
```diff
@@ -78,6 +80,17 @@ export function useSettingUI(
     )
   }
 
+  const subscriptionPanel: SettingPanelItem = {
```
arjansingh:
```suggestion
  const subscriptionPanel: SettingPanelItem | null = !isCloud ? null : {
```

@christian-byrne ternaries should get compiled out right? Basing code on line 145-146 below.

src/extensions/core/cloudSubscription.ts
```diff
@@ -0,0 +1,25 @@
...
+    const { requireActiveSubscription } = useSubscription()
+
+    const checkSubscriptionStatus = () => {
+      if (!isLoggedIn.value) return
+
+      void requireActiveSubscription()
+    }
+
+    watch(() => isLoggedIn.value, checkSubscriptionStatus, { immediate: true })
+  }
```
arjansingh: if you can we want to try to copy this ternary setup pattern from here:

https://github.com/Comfy-Org/ComfyUI_frontend/pull/6063/files#diff-b7818ca9daae2411d56695777160b8132507f2a3ff4f700d2510453c8833ca75R8-R15

src/components/actionbar/ComfyActionbar.vue
```diff
@@ -31,13 +32,29 @@ import {
...
 import { cn } from '@/utils/tailwindUtil'
 
 import ComfyQueueButton from './ComfyQueueButton.vue'
 
+const { isActiveSubscription } = useSubscription()
+
+const SubscribeToRun = isCloud
+  ? defineAsyncComponent(() => import('./SubscribeToRun.vue'))
+  : null
+
```
arjansingh: `useSubscription` should be dynamically imported and all of that code should be wrapped up in `isCloud`.

something like:
```ts
// Top-level - safe to import

// Inside component setup
let subscriptionState = ref(false)
let SubscribeToRun: Component | null = null

if (isCloud) {
  const { useSubscription } = await import('@/composables/auth/useSubscription')
  const { isActiveSubscription } = useSubscription()
  subscriptionState = isActiveSubscription

  SubscribeToRun = defineAsyncComponent(
    () => import('./SubscribeToRun.vue')
  )
}
```

src/composables/auth/useFirebaseAuthActions.ts
```diff
@@ -145,12 +145,32 @@ export const useFirebaseAuthActions = () => {
...
+      throw new Error(
+        t('toastMessages.failedToInitiateSubscription', {
+          error: 'No checkout URL returned'
+        })
+      )
+    }
+
+    window.open(response.checkout_url, '_blank')
+  }, reportError)
+
```
arjansingh: We should pull these out into `useSubscription`. It's already importing the `authStore`.

src/composables/auth/useSubscription.ts
```diff
@@ -0,0 +1,91 @@
+import { computed } from 'vue'
```
arjansingh: This composable is 🔥. I like it.

public/assets/images/cloud-subscription.webm
arjansingh, line 1: @christian-byrne do we want assets here or should this just get uploaded to GCS and we serve up a publicly signed link?

arjansingh (commented):
@jtydhr88 left a few more comments. The overall feature code looks solid. It's mainly around making sure we properly set up build time flags so we can remove this code where it's not needed.

@christian-byrne I think we should move the subscription-specific files into `@src//platform/cloud/subscription/`. It would definitely centralize everything. Thoughts?

christian-byrne:
There is a merge conflict it seems.

jtydhr88:
@christian-byrne fixed

christian-byrne:
Are these failing tests expected? Can you check the visual results: https://dbd3e99a.comfyui-playwright-chromium.pages.dev/#?q=s:failed

src/components/actionbar/ComfyActionbar.vue
```diff
@@ -34,7 +34,8 @@
             )
           "
         />
-        <ComfyQueueButton />
+        <SubscribeToRun v-if="!activeSubscription" />
+        <ComfyQueueButton v-else />
```
christian-byrne: Won't this hide the queue button on OSS build? We can use something like

```vue
<component :is=runButtonImplementation />
```

jtydhr88: fixed

src/components/actionbar/ComfyActionbar.vue
christian-byrne, line 1: Can you encapsulate the subscription logic into a single component like

```vue
// ExampleWrapper.vue
<template>
  <ComfyRunButton v-if=subscribed  />
  <SubscribeToRunButton v-else />
</template>
<script>
// all cloud-specific modules go here (like useSubscription.ts)
</script>
```

Then this becomes simpler and easier to work with:

```typescript
const runButtonImplementation = computed(() => {
  if (!isCloud) {
    return defineAsyncComponent(
      () => import('@/components/actionbar/ComfyQueueButton.vue')
    )
  } else {
    return defineAsyncComponent(
      () => import('@/components/actionbar/ExampleWrapper.vue')
    )
  }
})
```

christian-byrne: Actually, if we do the above, then we can simplify even further by following pattern outlined here: https://github.com/Comfy-Org/ComfyUI_frontend/pull/6112#pullrequestreview-3352394794

with 

```
ComfyRunButton/
    ExampleWrapper.vue
    ComfyQueueButton.vue
    index.ts
```

src/extensions/core/cloudSubscription.ts
```diff
@@ -0,0 +1,27 @@
...
+
+import { useCurrentUser } from '@/composables/auth/useCurrentUser'
+import { useSubscription } from '@/platform/cloud/subscription/composables/useSubscription'
+import { isCloud } from '@/platform/distribution/types'
+import { useExtensionService } from '@/services/extensionService'
+
+useExtensionService().registerExtension({
+  name: 'Comfy.CloudSubscription',
+
+  setup: isCloud
```
christian-byrne: You don't need to check `isCloud` here, because the extension is only imported on cloud anyway

src/platform/cloud/subscription/components/SubscriptionPanel.vue
```diff
@@ -0,0 +1,289 @@
+<template>
+  <TabPanel value="Plan & Credits" class="subscription-container h-full">
```
christian-byrne: Will this be translated?

src/platform/cloud/subscription/components/SubscriptionPanel.vue
```diff
@@ -0,0 +1,289 @@
...
+                          {{
+                            event.createdAt
+                              ? customerEventService.formatDate(event.createdAt)
+                              : ''
+                          }}
+                        </span>
+                      </div>
+                      <div
+                        v-if="event.params?.amount !== undefined"
+                        :class="['font-bold']"
```
christian-byrne:
```suggestion
                        class="font-bold"
```

src/platform/cloud/subscription/components/SubscriptionPanel.vue
```diff
@@ -0,0 +1,289 @@
...
+  }
+}
+
+onMounted(async () => {
+  await Promise.all([
+    authActions.fetchBalance(),
+    fetchStatus(),
+    fetchLatestEvents()
+  ])
+})
```
christian-byrne: Does this need to block the setup? Can we start these functions without `await` and display a loading state? Or, use something like https://vueuse.org/core/useAsyncState/

src/platform/cloud/subscription/components/SubscriptionPanel.vue
```diff
@@ -0,0 +1,289 @@
...
+
+const handleRefreshBalance = async () => {
+  await Promise.all([
+    authActions.fetchBalance(),
+    fetchStatus(),
+    fetchLatestEvents()
+  ])
```
christian-byrne: This seems like it could be refactored into a single function since it is done in 3 places.

src/platform/cloud/subscription/composables/useSubscription.ts
```diff
@@ -0,0 +1,209 @@
...
+import { COMFY_API_BASE_URL } from '@/config/comfyApi'
+import { t } from '@/i18n'
+import { isCloud } from '@/platform/distribution/types'
+import { useDialogService } from '@/services/dialogService'
+import {
+  FirebaseAuthStoreError,
+  useFirebaseAuthStore
+} from '@/stores/firebaseAuthStore'
+
+const MONTHLY_SUBSCRIPTION_PRICE = 20
```
christian-byrne: Can we move it to a file like in `src/config/subscriptionPricesConfig.ts` so that backend/ops teams can change it more easily in the future (or we can make it a build-time var)?

src/platform/cloud/subscription/composables/useSubscription.ts
```diff
@@ -0,0 +1,209 @@
...
+      async (loggedIn) => {
+        if (loggedIn) {
+          await fetchSubscriptionStatus()
+        } else {
+          subscriptionStatus.value = null
+        }
+      },
+      { immediate: true }
+    )
+  }
```
christian-byrne, line 158: Is the purpose of this to react to whenver the user logs in or changes account? Can it use the `onUserResolved` API from useCurrentUser composable?

src/services/dialogService.ts
christian-byrne: (replying to comment about whether this needs a no op) This should be tree-shaken due to the early return above I think - so I think it's fine how it is. (no change necessary).

src/platform/cloud/subscription/components/SubscriptionPanel.vue
jtydhr88: here would not be used in i18n, translated label is using in useSettingUI.ts https://github.com/Comfy-Org/ComfyUI_frontend/blob/c6b528b8be5f99cc6677250bbe92639a073ec779/src/platform/settings/composables/useSettingUI.ts#L73

src/platform/cloud/subscription/composables/useSubscription.ts
jtydhr88: I tried but onUserResovlved could not detect when if user logout, it could only get when user login, and I need to know when if user log out then set subscriptionStatus.value = null
image

src/platform/cloud/subscription/components/SubscriptionPanel.vue
jtydhr88: but this TabPanel's value is also matched to the label in useSettingUI.ts

src/platform/cloud/subscription/composables/useSubscription.ts
christian-byrne: Got it. In the future we could add a similar hook `onAuthUserLoggedOut`

src/extensions/core/cloudSubscription.ts
jtydhr88: fixed

src/platform/cloud/subscription/components/SubscriptionPanel.vue
jtydhr88: fixed

src/platform/cloud/subscription/components/SubscriptionPanel.vue
jtydhr88: fixed

src/platform/cloud/subscription/components/SubscriptionPanel.vue
jtydhr88: fixed

src/platform/cloud/subscription/components/SubscriptionPanel.vue
jtydhr88: fixed

src/platform/cloud/subscription/composables/useSubscription.ts
jtydhr88: fixed

christian-byrne (approved):

PR #6079, opened 2025-10-15
Prevent partial copy of custom widgets when performing linked promotion on subgraphs

AustinMroz:
The `toConcrete` call creates a restricted view of a widget that extends from `BaseWidget`. A copy of the widget created by `createCopyForNode` will also inherit this restricted view. This creates two problems
- Some widget properties (like `displayValue`) have been judged unsafe and are explicitly blacklisted from being copied 
- The widget now extends from `BaseWidget`. This results in the widget being processed differently in some logic, such as `#processWidgetClick`
  - Because `LegacyWidget` provides an implementation for `onClick`, the presence of click handlers can not be used to determine which should be used.

As a proposed, minimal workaround. Widgets which do not already extend from BaseWidget are no longer cloned through `createCopyForNode`.

Because this PR involves side-stepping properties which have been explicitly blacklisted. I'd recommend waiting to merge/backport until after the release of 1.28.7

Resolves Kosinkadink/ComfyUI-VideoHelperSuite#569

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6079-Prevent-partial-copy-of-custom-widgets-when-performing-linked-promotion-on-subgraphs-28d6d73d36508198950efd401186ddef) by [Unito](https://www.unito.io)


christian-byrne:
This PR is only 9 LOC, but there's a lot of history, depth, and complexity to it, haha.

AustinMroz:
Yeah. Wound up merging the proposed workaround in VHS so there's no rush here. Probably best I do some more digging and add test cases for each of the edge cases that can cause things to break here.

AustinMroz:
Some more note taking.
- Blacklisting was added by comfy-org/litegraph.js#1023
  - The blacklisting is just a list of BaseWidget properties which only have getters
    - Thus, custom node props wasn't a factor at time of creation... Shouldn't be an issue now.
  - Blacklisting `displayValue` instead of` _displayValue` is a bug?
    - `displayValue` was changed from `_displayValue` due to a conflict with VHS, but the value in VHS is a safe, well-behaved, function
- The existing implementation `createCopyForNode` is only a shallow copy
  - (Making a change to `widget.options.min` also changes `promotedWidget.options.min`)
- Spread skips properties inherited from prototypes
  - This makes it inherently incomplete. Should be avoided?
- `{__proto__: widget}`? Still breaks private props/getters. Probably safer, but Alex has declared disain.
  - Perhaps combining? `{...widget, __proto__: widget}` is less shallow, but still struggles with getters/private props

DrJKL:
@coderabbitai review

DrJKL (approved):

PR #6087, opened 2025-10-16
Support cross domain/application copy/paste

AustinMroz:
AnimateDiff_00001

Browsers place very heavy restrictions on what can be copied and pasted. See:
- https://alexharri.com/blog/clipboard
- https://www.w3.org/TR/clipboard-apis/#mandatory-data-types-x

Further thoughts and TODO
- Should the local storage of copied nodes be removed entirely?
  - It's unlikely a person would want to paste nodes from several days back, but having multiple items in the clipboard may be desirable. Should "copy node -> copy text -> paste text -> paste node" be possible?
- Need to add extra safety to the checks instead of a try/catch block

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6087-Experimental-cross-domain-application-copy-paste-28e6d73d36508154a0a8deeb392f43a4) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useCopy.ts
+++ b/src/composables/useCopy.ts
@@ -2,6 +2,11 @@ import { useEventListener } from '@vueuse/core'
 
 import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
 
+const clipboardHTMLWrapper = [
+  '<meta charset="utf-8"><div><span data-metadata="',
+  '"></span></div><span style="white-space:pre-wrap;">Text</span>'
+]
+
 /**
  * Adds a handler on copy that serializes selected nodes to JSON
  */
@@ -23,9 +28,7 @@ export const useCopy = () => {
       // clearData doesn't remove images from clipboard
       e.clipboardData?.setData(
         'text/html',
-        '<meta charset="utf-8"><div><span data-metadata="' +
-          btoa(serializedData) +
-          '"></span></div><span style="white-space:pre-wrap;">Text</span>'
+        clipboardHTMLWrapper.join(btoa(serializedData))
       )
       e.preventDefault()
       e.stopImmediatePropagation()
--- a/src/composables/useCopy.ts
+++ b/src/composables/useCopy.ts
@@ -1,6 +1,7 @@
 import { useEventListener } from '@vueuse/core'
 
 import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
+import { shouldIgnoreCopyPaste } from '@/workbench/eventHelpers'
 
 const clipboardHTMLWrapper = [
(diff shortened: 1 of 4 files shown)


src/composables/useCopy.ts
```diff
@@ -9,28 +9,24 @@ export const useCopy = () => {
   const canvasStore = useCanvasStore()
 
   useEventListener(document, 'copy', (e) => {
-    if (!(e.target instanceof Element)) {
-      return
-    }
     if (
-      (e.target instanceof HTMLTextAreaElement &&
-        e.target.type === 'textarea') ||
-      (e.target instanceof HTMLInputElement && e.target.type === 'text')
```
christian-byrne, line 18: Did we want to return early for all `HTMLInputElement` including things like radio, checkboxes, button?

AustinMroz: Very good point. From a quick pass of all input element types, looks like we'd want.  
- Handled by canvas: button, checkbox, file, hidden, image, radio, range, reset, search, submit
- Handled by element: color, date, datetime-local, email, month, number, password, search, tel, text, time, url, week, datetime

Probably best implemented as a whitelist of input element types we allow the canvas to handle for future-proofing?

src/composables/usePaste.ts
```diff
@@ -8,6 +8,25 @@ import { app } from '@/scripts/app'
 import { useWorkspaceStore } from '@/stores/workspaceStore'
 import { isAudioNode, isImageNode, isVideoNode } from '@/utils/litegraphUtil'
 
+function pasteClipboardItems(data: DataTransfer): boolean {
+  try {
+    const tempElement = document.createElement('div')
+    tempElement.innerHTML = data.getData('text/html')
```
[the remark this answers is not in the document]
AustinMroz: Was having difficulty testing if script tags would actually run before being added to the document, but the actual data being base64 encoded makes it trivial to swap to regex here.

src/composables/useCopy.ts
AustinMroz: And also probably best to revert changes here to be *just* clipboard support and have any targeting changes be handled over in #6055

AustinMroz: Added whitelist here as suggested. For an easily testable places this matters: Try pasting after toggling the value of a boolean in vue mode.

christian-byrne (approved):
LGTM

geroldmeisinger:
@AustinMroz thanks for your PR and contribution. however it appears this change broke my copy+paste functionality in ComfyUI, see #6660. I have to revert back to 1.30.1.

system:
Firefox
Cinnamon Desktop Manager
Debian Linux

PR #6090, opened 2025-10-16
feat: deprecated API alert

LittleSound:
When an extension imports deprecated APIs, this log will appear:
```bash
[ComfyUI Deprecated] Importing from <file path> is deprecated. <guidance information>. This will be removed in <version>. See: <url>
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6090-feat-deprecated-API-alert-28e6d73d365081898c5ddafcc025d6f8) by [Unito](https://www.unito.io)


christian-byrne (commented):
How about something like:

- For files in `src/scripts/ui` we use the current message and set 1.34 as the target version - but we don't really have a replacement. These are the files we are definitely going to remove
- For all other files, we just say something like

  > [ComfyUI Notice] "${shimFileName}" is an internal module, not part of the public API. Future updates may break this import.
  
Then, we can also do something like this to prevent excessive warnings?

```typescript
const warnedInternals = new Set()
if (!warnedInternals.has(shimFileName)) {
  console.warn(`[ComfyUI Warning] Importing from "${shimFileName}" is not part of the public API. Internal paths may change or be removed at any time.`)
  warnedInternals.add(shimFileName)
}
```

(or something similar)?

christian-byrne:
Other files I thought of (will be updating this):

- src/extensions/core/maskEditorOld.ts

LittleSound:
> Then, we can also do something like this to prevent excessive warnings?

It will only display once because it is at the root of the file.

LittleSound:
> Other files I thought of (will be updating this):
> 
> * src/extensions/core/maskEditorOld.ts

How should this file be handled?

christian-byrne:
> How should this file be handled?

This is another one that we want to remove in 1.34. 

Same with:

- public/materialdesignicons.min.css
- public/fonts/materialdesignicons-webfont.woff2
- src/extensions/core/maskEditorOld.js
- src/scripts/ui/*
- src/scripts/ui.js
- src/extensions/core/groupNode.js
- maybe some utils and metadata parsing files that are likely to be moved/refactored

LittleSound:
I added these `.js` files, but other files cannot use this method to warn.

LittleSound:
related: https://github.com/comfyanonymous/ComfyUI/issues/10423

> I think It would be nice If we had a link to open a migration document.

christian-byrne (approved):

PR #6110, opened 2025-10-17
fix minimap not re-rendering when Vue nodes change mode (bypass, mute)

christian-byrne:
## Summary

Fixes minimap not updating when node bypass/mute state changes in vue nodes mode by adding `onTrigger` tracking of relevant instrumented properties.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6110-fix-minimap-not-re-rendering-when-Vue-nodes-change-mode-bypass-mute-28f6d73d36508112bab1ce031705c6a9) by [Unito](https://www.unito.io)


benceruleanlu (dismissed):

christian-byrne:
This fixes it when using the feature, but not in tests.

christian-byrne:
Wow, this is depressing.

christian-byrne:
Hm, clicking the canvas works. I'll add a timeout to see whether I just need to wait generally or I need to explicitly trigger litegraph re-render

Issue #1502, opened 2024-11-11
[Feature Request]: Filter Queue Images for each `Preview Image` node 


JorgeR81:
### Is there an existing issue for this?

- [X] I have searched the existing issues and checked the recent builds/commits

### What would your feature do ?

This feature would allow filtering Queue images, per image node.

This way we could see all the images displayed on a specific `Preview Image`, during the current session. 

Currently we only have 2 options to see either, the first image or all images from a specific generation. 

But if the workflow has multiple extra stages ( e.g. Refiner, Face Detailer, Upscaler ), it may hard to know which is which, in the full image sequence.
And the first image of the workflow, it's usually the least important. 

Also, some workflows may produce images that we don't need to see in the Queue ( e.g. preprocessor images ).

### Proposed workflow

- We would have search field Queue sidebar ( like on the other sidebars )
- We could type the title name we gave to a `Preview Image` node or its ID badge number ( e.g.  `#10`)
- The Queue sidebar will only display images from that node   

![q0](https://github.com/user-attachments/assets/f9b46d33-06c3-4ed8-81ee-7f1ce7e3c1fb)

### Additional information

This could replace a similar feature request I've made.

- https://github.com/Comfy-Org/ComfyUI_frontend/issues/1422

JorgeR81:
This would be a better solution:  https://github.com/Comfy-Org/ComfyUI_frontend/issues/1502

PR #6112, opened 2025-10-17
Media Assets Management Sidebar Tab Implementation

### LINKED ISSUE #6267: The queue windows are covering the generated images with information, that's not really needed.

The queue explorer doesn't need the "Save image" and timestamps. 
Image

When a prompt has been completed. . 

Multiple times per day

Minor annoyance

### Current workarounds

_No response_

The information isn't really needed, if you really want to know the generation times, you can look in the cli. If a image exists then there's no reason to know the state. 

### Additional context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6267-The-queue-windows-are-covering-the-generated-images-with-information-that-s-not-reall-2976d73d3650817daadadbf4e61cfa2c) by [Unito](https://www.unito.io)

viva-jinyi:
## 📋 Overview
Implemented a new Media Assets sidebar tab in ComfyUI for managing user-uploaded input files and generated output files. This feature supports both local and cloud environments and is currently enabled only in development mode.

### 1. Media Assets Sidebar Tab
- **Imported** / **Generated** files separated by tabs
- Visual display with file preview cards
- Gallery view support (navigable with arrow keys)

### 2. Environment-Specific Implementation
- **`useInternalMediaAssets`**: For local environment
  - Fetches file list via `/files` API
  - Retrieves generation task execution time via `/history` API
  - Processes history data using the same logic as QueueSidebarTab
- **`useCloudMediaAssets`**: For cloud environment
  - File retrieval through assetService
  - History data processing using TaskItemImpl
  - Auto-truncation of long filenames over 20 characters (e.g., `very_long_filename_here.png` → `very_long_...here.png`)

### 3. Execution Time Display
- Shows task execution time on generated image cards (e.g., "2.3s")
- Calculated from History API's `execution_start` and `execution_success` messages
- Displayed at MediaAssetCard's duration chip location

### 4. Gallery Feature
- Full-screen gallery mode on image click
- Navigate between images with keyboard arrows
- Exit gallery with ESC key
- Reuses ResultGallery component from QueueSidebarTab

### 5. Development Mode Only
- Excluded from production builds using `import.meta.env.DEV` condition
- Feature in development, scheduled for official release after stabilization

### New Files Added
- `src/components/sidebar/tabs/AssetsSidebarTab.vue` - Main sidebar tab component
- `src/composables/sidebarTabs/useAssetsSidebarTab.ts` - Sidebar tab definition
- `src/composables/useInternalMediaAssets.ts` - Local environment implementation
- `src/composables/useCloudMediaAssets.ts` - Cloud environment implementation
- `packages/design-system/src/icons/image-ai-edit.svg` - Icon addition

### Modified Files
- `src/stores/workspace/sidebarTabStore.ts` - Added dev mode only tab display logic
- `src/platform/assets/components/MediaAssetCard.vue` - Added execution time display, zoom event
- `src/platform/assets/components/MediaImageTop.vue` - Added image dimension detection
- `packages/shared-frontend-utils/src/formatUtil.ts` - Added media type determination utility functions
- `src/locales/en/main.json` - Added translation keys

[media_asset_OSS_cloud.webm](https://github.com/user-attachments/assets/a6ee3b49-19ed-4735-baad-c2ac2da868ef)


packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -474,3 +474,51 @@ export function formatDuration(milliseconds: number): string {
...
+ */
+export function getMediaTypeFromFilename(filename: string): string {
+  if (!filename) return 'images'
+  const ext = filename.split('.').pop()?.toLowerCase()
+  if (!ext) return 'images'
+
+  const imageExts = ['png', 'jpg', 'jpeg', 'gif', 'webp', 'bmp']
+  const videoExts = ['mp4', 'webm', 'mov', 'avi']
+  const audioExts = ['mp3', 'wav', 'ogg', 'flac']
+  const threeDExts = ['obj', 'fbx', 'gltf', 'glb']
```
christian-byrne: nit: We can move these to the module scope (avoid re-initialization on every call)

packages/shared-frontend-utils/src/formatUtil.ts
christian-byrne, line 1: Can we combine the concept of media kind and media type? It seems to be exactly the same but just differing between plural vs. singular. I think singular is better to align with MIME https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types

src/composables/useCloudMediaAssets.ts
```diff
@@ -0,0 +1,133 @@
...
+                    : ''
+
+                // If name without extension is still long, truncate it
+                if (nameWithoutExt.length > 10) {
+                  displayName =
+                    nameWithoutExt.substring(0, 10) +
+                    '...' +
+                    nameWithoutExt.substring(nameWithoutExt.length - 10) +
+                    extension
+                }
```
christian-byrne: Suggestion: extract this to a helper function then use es-toolkit truncate helper.

src/composables/useCloudMediaAssets.ts
```diff
@@ -0,0 +1,133 @@
...
+                  nodeId: output.nodeId,
+                  subfolder: output.subfolder,
+                  ...(executionTimeInSeconds && {
+                    executionTimeInSeconds
+                  }),
+                  ...(output.format && {
+                    format: output.format
+                  }),
+                  ...(taskItem.workflow && {
+                    workflow: taskItem.workflow
```
christian-byrne: Suggestion: extract this logic to a dedicate mapping function so that:

- it is more explicit and expressive
- we don't need the inline `&&` notation
- we can declare the output type explicitly to prevent future mistakes

src/composables/useInternalMediaAssets.ts
```diff
@@ -0,0 +1,129 @@
...
+                  subfolder: output.subfolder,
+                  ...(executionTimeInSeconds && {
+                    executionTimeInSeconds
+                  }),
+                  ...(output.format && {
+                    format: output.format
+                  }),
+                  ...(taskItem.workflow && {
+                    workflow: taskItem.workflow
+                  })
```
christian-byrne: Same comment here as with https://github.com/Comfy-Org/ComfyUI_frontend/pull/6112/files#r2441443691

christian-byrne (commented):
Can we put the composables in the `platform/assets` folder and then re-arrange so it's like:

```
useMediaAssets/
    useAssetsApi.ts
    useInternalFilesApi.ts
    index.ts
```

then inside the index.ts file you can put:

```typescript

if (isCloud) { export { useAssetsApi as useMediaAssets } from './useAssetsApi' } 
else { export { useInternalFilesApi as useMediaAssets } from './useInternalFilesApi' }
```

Then we should also define an interface or abstract class

```typescript
interface IAssetsProvider {
   loading: ref<boolean>
   error: ref<null | string>
   fetchMediaList: () => UninifedType
}
```

This way, no other parts of the codebase need to understand the difference in implementations for assets - everything is abstracted behind the `UnifiedType` and `AssetsProvider` concepts. This makes refactoring, reasoning about the code, and especially dealing with types considerably easier and more maintainable.

The additional benefit is that for OSS build, the entire `useAssetsApi.ts` file actually gets tree-shaken.

packages/shared-frontend-utils/src/formatUtil.ts
viva-jinyi: Thx. I modified it.

packages/shared-frontend-utils/src/formatUtil.ts
viva-jinyi: Thx. I modified it.

src/composables/useCloudMediaAssets.ts
viva-jinyi: Thx. I modified it.

src/composables/useInternalMediaAssets.ts
viva-jinyi: Thx. I modified it.

src/composables/useCloudMediaAssets.ts
viva-jinyi: Thx. I modified it.

src/platform/assets/composables/useMediaAssets/useInternalFilesApi.ts
```diff
@@ -0,0 +1,84 @@
...
+          )
+        })
+
+      // Sort by creation date (newest first)
+      return assetItems.sort(
+        (a, b) =>
+          new Date(b.created_at).getTime() - new Date(a.created_at).getTime()
+      )
+    } catch (err) {
+      const errorMessage = err instanceof Error ? err.message : 'Unknown error'
```
DrJKL: Thank you so much, I love it 😁

src/platform/assets/composables/useMediaAssets/useInternalFilesApi.ts
```diff
@@ -0,0 +1,84 @@
...
+  mapTaskOutputToAssetItem
+} from './assetMappers'
+
+/**
+ * Composable for fetching media assets from local environment
+ * Uses the same logic as QueueSidebarTab for history processing
+ */
+export function useInternalFilesApi() {
+  const loading = ref(false)
+  const error = ref<string | null>(null)
```
DrJKL: I like this a lot, overall.
But for composables like this, I would usually rather have the state that correlates with the requested assets be tied together.
So in this instance, the composable would accept the `directory` as an argument, bake it in, then you could return a function to kick off the request.

Something like
```ts
const { media, loading, error, fetchMediaList } = useInternalFilesApi('input')
```

Then you could guard against possibly calling `fetchMediaList` multiple times which would make the loading and error values ambiguous.

src/platform/assets/composables/useMediaAssets/useInternalFilesApi.ts
```diff
@@ -0,0 +1,84 @@
...
+        })
+        if (!response.ok) {
+          throw new Error(`Failed to fetch ${directory} files`)
+        }
+        const filenames: string[] = await response.json()
+
+        return filenames.map((name, index) =>
+          mapInputFileToAssetItem(name, index, directory)
+        )
+      }
```
DrJKL: I think this big of a branch probably deserves to be broken out.

src/platform/assets/composables/useMediaAssets/useAssetsApi.ts
```diff
@@ -0,0 +1,77 @@
...
+import { mapTaskOutputToAssetItem } from './assetMappers'
+
+/**
+ * Composable for fetching media assets from cloud environment
+ * Includes execution time from history API
+ */
+export function useAssetsApi() {
+  const loading = ref(false)
+  const error = ref<string | null>(null)
+
```
DrJKL: ... Deja vu...

DrJKL: Is there a reasonably small abstraction that could unify these two?

src/platform/assets/composables/media/IAssetsProvider.ts
```diff
@@ -0,0 +1,22 @@
...
+  /** Error state, null when no error */
+  error: Ref<string | null>
+
+  /**
+   * Fetch list of media assets from the specified directory
+   * @param directory - 'input' or 'output'
+   * @returns Promise resolving to array of AssetItem
+   */
+  fetchMediaList: (directory: 'input' | 'output') => Promise<AssetItem[]>
+}
```
DrJKL, line 29: If you could find a way to unify the composables, you wouldn't need an interface 😉

src/platform/assets/composables/useMediaAssets/assetMappers.ts
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Maps a TaskItemImpl output to an AssetItem format
+ * @param taskItem The task item containing execution data
+ * @param output The output from the task
+ * @param useDisplayName Whether to truncate the filename for display
+ * @returns AssetItem formatted object
+ */
+export function mapTaskOutputToAssetItem(
+  taskItem: TaskItemImpl,
+  output: any,
```
DrJKL: Can you find a better type? Maybe it needs to be generic with some constraints?

src/platform/assets/components/MediaImageTop.vue
```diff
@@ -17,11 +18,35 @@
...
+
+// Emit dimensions when image is loaded
+watch(isReady, (ready) => {
+  if (ready && state.value) {
+    const width = state.value.naturalWidth
+    const height = state.value.naturalHeight
+    if (width && height) {
+      emit('image-loaded', { width, height })
+    }
+  }
```
DrJKL: Very nit, but could you flip the logic here to use guard clauses?

Also, I think you could use `whenever` to remove the top level conditional.

src/platform/assets/components/MediaImageTop.vue
```diff
@@ -17,11 +18,35 @@
...
+  'image-loaded': [dimensions: { width: number; height: number }]
+}>()
+
+// Use same image loading logic as AssetCard
+const { state, error, isReady } = useImage({
+  src: asset.src ?? '',
+  alt: asset.name
+})
+
+const shouldShowImage = computed(() => asset.src && !error.value)
```
DrJKL: I think you should be able to just gate on `!error` in the template. What happens if you give `useImage` an empty string for the resource path?

src/platform/assets/components/MediaImageTop.vue
```diff
@@ -1,11 +1,12 @@
...
+  <div
+    class="relative h-full w-full overflow-hidden rounded bg-zinc-200 dark-theme:bg-zinc-700/50"
+  >
+    <img
+      v-if="shouldShowImage"
       :src="asset.src"
       :alt="asset.name"
-      :container-class="'aspect-square'"
-      :image-class="'w-full h-full object-cover'"
+      class="h-full w-full object-contain"
```
DrJKL: Very nit:
```suggestion
      class="size-full object-contain"
```

src/platform/assets/components/MediaImageTop.vue
```diff
@@ -1,11 +1,12 @@
 <template>
-  <div class="relative h-full w-full overflow-hidden rounded">
-    <LazyImage
-      v-if="asset.src"
+  <div
+    class="relative h-full w-full overflow-hidden rounded bg-zinc-200 dark-theme:bg-zinc-700/50"
```
DrJKL: If we don't have a design token for the background here, could we make one?

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -272,7 +323,11 @@ const handleOverlayMouseLeave = () => {
 
 const handleZoomClick = () => {
   if (asset) {
-    galleryStore.openSingle(asset)
+    emit('zoom', asset)
   }
 }
+
+const handleImageLoaded = (dimensions: { width: number; height: number }) => {
```
DrJKL: Not sure how many times it's used, but it could be worth naming `{ width: number; height: number }`

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -229,36 +273,43 @@ const showHoverActions = computed(
...
-  () =>
-    showHoverActions.value &&
-    (!isVideoPlaying.value || isCardOrOverlayHovered.value)
-)
+const showActionsOverlay = false
+// const showActionsOverlay = computed(
+//   () =>
+//     showHoverActions.value &&
+//     (!isVideoPlaying.value || isCardOrOverlayHovered.value)
+// )
```
DrJKL:
```suggestion
```

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -165,28 +164,65 @@ function getBottomComponent(kind: MediaKind) {
...
+    tags: asset.tags || [],
+    created_at: asset.created_at,
+    duration: asset.user_metadata?.duration
+      ? Number(asset.user_metadata.duration)
+      : undefined,
+    dimensions:
+      imageDimensions.value ||
+      (asset.user_metadata?.dimensions as
+        | { width: number; height: number }
+        | undefined)
```
DrJKL: Why do you need the type assertion?

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -165,28 +164,65 @@ function getBottomComponent(kind: MediaKind) {
...
+const imageDimensions = ref<{ width: number; height: number } | undefined>()
+
 const isHovered = useElementHover(cardContainerRef)
 
 const actions = useMediaAssetActions()
-const galleryStore = useMediaAssetGalleryStore()
+
+// Get asset type from tags[0]
+const assetType = computed(() => {
+  return (asset?.tags?.[0] as 'input' | 'output') || 'input'
```
DrJKL: Same question, could you find a way to wire the types through consistently?

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -79,7 +82,7 @@
           <IconTextButton
             type="secondary"
             size="sm"
-            :label="context?.outputCount?.toString() ?? '0'"
+            :label="'0'"
```
DrJKL: 0?

DrJKL:
```suggestion
            label="0"
```

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -2,9 +2,7 @@
   <CardContainer
     ref="cardContainerRef"
     role="button"
-    :aria-label="
-      asset ? `${asset.name} - ${asset.kind} asset` : 'Loading asset'
-    "
+    :aria-label="asset ? `${asset.name} - ${fileKind} asset` : 'Loading asset'"
```
DrJKL: I don't know that most people think of things in terms of "assets". It might be worth thinking of more common/casual ways to describe things here.

DrJKL: Also, should this be going through `t()`?

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -0,0 +1,145 @@
...
+  }
+}
+
+watch(activeTab, () => {
+  void refreshAssets()
+})
+
+onMounted(() => {
+  void refreshAssets()
+})
```
DrJKL: If you made the `watch` immediate, could you eliminate the `onMounted`?

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -0,0 +1,145 @@
...
+            :asset="item"
+            :selected="selectedAsset?.id === item.id"
+            @click="handleAssetSelect(item)"
+            @zoom="handleZoomClick(item)"
+          />
+        </template>
+      </VirtualGrid>
+      <div v-else-if="loading">
+        <ProgressSpinner
+          style="width: 50px; left: 50%; transform: translateX(-50%)"
```
DrJKL: Instead of style, could you use tailwind classes?

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -474,3 +474,93 @@ export function formatDuration(milliseconds: number): string {
...
+ */
+export function getMediaTypeFromFilenamePlural(filename: string): string {
+  const type = getMediaTypeFromFilename(filename)
+  switch (type) {
+    case 'image':
+      return 'images'
+    case 'video':
+      return 'videos'
+    case 'audio':
+      return 'audios'
```
DrJKL: I don't usually hear `audios`

DrJKL (commented):
Reviewed from the bottom up 🙃

src/platform/assets/components/MediaAssetCard.vue
christian-byrne: "assets" is the terminology we have already converged on.

src/platform/assets/composables/media/IAssetsProvider.ts
christian-byrne: This is the design I asked for.

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -474,3 +474,93 @@ export function formatDuration(milliseconds: number): string {
 
   return parts.join(' ')
 }
+
+// Module scope constants to avoid re-initialization on every call
```
christian-byrne:
```suggestion
```

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -474,3 +474,93 @@ export function formatDuration(milliseconds: number): string {
...
+/**
+ * @deprecated Use getMediaTypeFromFilename instead - kept for backward compatibility
+ * @param filename The filename to analyze
+ * @returns The media kind: 'image', 'video', 'audio', or '3D'
+ */
+export function getMediaKindFromFilename(
+  filename: string
+): 'image' | 'video' | 'audio' | '3D' {
+  return getMediaTypeFromFilename(filename)
+}
```
christian-byrne: Could we just replace the instances where the old one is used?

packages/shared-frontend-utils/src/formatUtil.ts
viva-jinyi: Just noticed that `ResultItemImpl` uses:
- `mediaType === 'images'` (plural)
- `mediaType === 'audio'` (singular)  
- `mediaType === 'video'` (singular)

So `getMediaTypeFromFilename()` should return `'images'` (plural) for images to match existing code:

```typescript
if (imageExts.includes(ext)) return 'images'  // not 'image'
if (videoExts.includes(ext)) return 'video'
if (audioExts.includes(ext)) return 'audio'
```

viva-jinyi: ✅

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: ✅

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: Thank you!

viva-jinyi: ✅

src/platform/assets/components/MediaAssetCard.vue
viva-jinyi: Updated the `MediaAssetCard` component to use translation keys for accessibility labels as per review feedback.

viva-jinyi: ✅

src/platform/assets/components/MediaAssetCard.vue
viva-jinyi: ✅

viva-jinyi: This is only temporary code.

viva-jinyi: For more context, in this PR I created a composable to view the assets list. I also made another PR for the job detail view, which opens when you click the output count. #6164

src/platform/assets/composables/useMediaAssets/useAssetsApi.ts
viva-jinyi: Refactored `useInternalFilesApi` and `useAssetsApi` composables to use VueUse's `useAsyncState` pattern based on review feedback.

viva-jinyi: ✅

src/platform/assets/composables/useMediaAssets/useInternalFilesApi.ts
viva-jinyi: ✅

src/platform/assets/composables/useMediaAssets/useInternalFilesApi.ts
viva-jinyi: ✅

src/platform/assets/composables/useMediaAssets/assetMappers.ts
viva-jinyi: Thank you for pointing out.

viva-jinyi: ✅

src/platform/assets/components/MediaImageTop.vue
viva-jinyi: I'll use `whenever` pattern instead of current pattern

viva-jinyi: ✅

src/platform/assets/components/MediaImageTop.vue
viva-jinyi: ✅

src/platform/assets/components/MediaImageTop.vue
viva-jinyi: ✅

src/platform/assets/components/MediaAssetCard.vue
viva-jinyi: You're right

viva-jinyi: ✅

src/platform/assets/components/MediaAssetCard.vue
viva-jinyi: ✅

src/platform/assets/components/MediaAssetCard.vue
viva-jinyi: ✅

src/platform/assets/components/MediaAssetCard.vue
viva-jinyi: ✅

src/platform/assets/components/MediaImageTop.vue
viva-jinyi: I'll address this in a separate PR.

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -0,0 +1,145 @@
...
+const galleryActiveIndex = ref(-1)
+const galleryItems = computed(() => {
+  // Convert AssetItems to ResultItemImpl format for gallery
+  return mediaAssets.value.map((asset) => {
+    const resultItem = new ResultItemImpl({
+      filename: asset.name,
+      subfolder: '',
+      type: 'output',
+      nodeId: '0',
+      mediaType: getMediaTypeFromFilenamePlural(asset.name)
```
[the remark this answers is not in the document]
viva-jinyi: ✅

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -0,0 +1,145 @@
...
+    ...asset,
+    key: asset.id
+  }))
+})
+
+const refreshAssets = async () => {
+  const files = await fetchMediaList(activeTab.value)
+  mediaAssets.value = files
+  if (error.value) {
+    console.error('Failed to refresh assets:', error.value)
```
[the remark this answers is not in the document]
viva-jinyi: ✅

src/platform/assets/composables/useMediaAssets/assetMappers.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  // Store original filename if using display name
+  if (useDisplayName) {
+    metadata.originalFilename = output.filename
+  }
+
+  return {
+    id: `${taskItem.promptId}-${output.nodeId}-${output.filename}`,
+    name: useDisplayName
+      ? truncateFilename(output.filename, 20)
```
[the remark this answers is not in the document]
viva-jinyi: ✅

src/platform/assets/composables/useMediaAssets/assetMappers.ts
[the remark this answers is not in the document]
viva-jinyi: ✅

src/platform/assets/composables/useMediaAssets/useAssetsApi.ts
```diff
@@ -0,0 +1,77 @@
...
+  ): Promise<AssetItem[]> => {
+    loading.value = true
+    error.value = null
+
+    try {
+      // For input directory, just return assets without history
+      if (directory === 'input') {
+        const assets = await assetService.getAssetsByTag(directory, false)
+        // Process assets to truncate long filenames for display
+        return assets.map((asset) => ({
```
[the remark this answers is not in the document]
viva-jinyi: ✅

src/platform/assets/components/MediaAssetCard.vue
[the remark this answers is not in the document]
viva-jinyi: ✅

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -474,3 +474,93 @@ export function formatDuration(milliseconds: number): string {
...
+
+  return 'image'
+}
+
+/**
+ * @deprecated Use getMediaTypeFromFilename instead - returns plural form for legacy compatibility
+ * @param filename The filename to analyze
+ * @returns The media type in plural form: 'images', 'videos', 'audios', '3D'
+ */
+export function getMediaTypeFromFilenamePlural(filename: string): string {
```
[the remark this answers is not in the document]
viva-jinyi: ✅

src/platform/assets/composables/media/assetMappers.ts
```diff
@@ -0,0 +1,81 @@
...
+  if (taskItem.workflow) {
+    metadata.workflow = taskItem.workflow
+  }
+
+  // Store original filename if using display name
+  if (useDisplayName) {
+    metadata.originalFilename = output.filename
+  }
+
+  return {
```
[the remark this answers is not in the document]
viva-jinyi: ✅

src/stores/workspace/sidebarTabStore.ts
```diff
@@ -102,6 +104,11 @@ export const useSidebarTabStore = defineStore('sidebarTab', () => {
    * Register the core sidebar tabs.
    */
   const registerCoreSidebarTabs = () => {
+    // Only show AssetsSidebarTab in development mode
+    if (import.meta.env.DEV) {
```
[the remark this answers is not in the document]
viva-jinyi: It's okay

tests-ui/tests/utils/formatUtil.test.ts
```diff
@@ -0,0 +1,147 @@
+import { describe, expect, it } from 'vitest'
+
+import {
+  getMediaTypeFromFilename,
+  getMediaTypeFromFilenamePlural,
+  truncateFilename
+} from '@/utils/formatUtil'
+
+describe('formatUtil', () => {
+  describe('truncateFilename', () => {
```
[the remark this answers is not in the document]
viva-jinyi: ✅

viva-jinyi:
I refactored the code based on all the feedback I received in this PR, and I marked each addressed review comment with a ✅ emoji.

1. **Composable Architecture Enhancement**
   - Refactored `useInternalFilesApi` and `useAssetsApi` using factory pattern
   - Created independent instances for each directory ('input'/'output')
   - Simplified state management using VueUse's `useAsyncState`

2. **Type Safety Improvements**
   - Removed `any` types and used proper TypeScript interfaces
   - Eliminated type casting with `getAssetType` utility function
   - Explicitly typed `ResultItemImpl` parameter

3. **Accessibility Improvements**
   - Added i18n support for all aria-labels
   - Translatable to Korean and other languages

4. **Code Quality Enhancements**
   - Removed unnecessary computed properties
   - Simplified watch logic using VueUse's `whenever` pattern
   - Extracted large branches into separate functions (`fetchInputFiles`, `fetchOutputFiles`)

5. **Enhanced Testing**
   - Added edge case tests for `getMediaTypeFromFilename` function
   - Handles null/undefined, special characters, long filenames

6. **Dead Code Removal**
   - Removed `getMediaTypeFromFilenamePlural` function
   - Cleaned up unnecessary comments and code

### Technical Details
- Improved async state management with `useAsyncState`
- Independent composable instances per directory
- Improved image load event emit signature

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -0,0 +1,285 @@
+<template>
+  <AssetsSidebarTemplate>
+    <template #top>
+      <span v-if="!isInFolderView" class="font-bold">
+        {{ $t('sideToolbar.mediaAssets') }}
+      </span>
+      <div v-else class="flex w-full items-center justify-between gap-2">
+        <div class="flex items-center gap-2">
+          <span class="font-bold">{{ $t('Job ID') }}:</span>
```
christian-byrne, line 9: Is the key supposed to be 'Job ID'?

christian-byrne:
Can you verify to make sure we still account for the bugs fixed in this PR:

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/2125

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -0,0 +1,285 @@
...
+          type="secondary"
+          @click="exitFolderView"
+        >
+          <template #icon>
+            <i class="icon-[lucide--arrow-left] size-4" />
+          </template>
+        </IconTextButton>
+      </div>
+      <!-- Normal Tab View -->
+      <TabList v-else v-model="activeTab" class="pt-4 pb-1">
```
christian-byrne, line 38: In the future, we should find a way to persist this so the user's previous active tab is restored when re-opening the assets sidebar.

christian-byrne (commented):
This looks good. So the current system sort of works like:

1. Open assets tab -> use assets API to get outputs/inputs
2. The inputs/outputs are not reactive (so generating more outputs while the tab is visible won't update the items)

In the queue sidebar, the way it works is we listen for status updates (indicating something has changed about the jobs on the server). On status messages, we call `queueStore.update` ([code](https://github.com/Comfy-Org/ComfyUI_frontend/blob/b48636ad0e1373b540b59efb9361561a59f86cc4/src/views/GraphView.vue#L189-L192)).

queueStore's update is [here](https://github.com/Comfy-Org/ComfyUI_frontend/blob/b48636ad0e1373b540b59efb9361561a59f86cc4/src/stores/queueStore.ts#L454-L460):

```typescript
const update = async () => {
    isLoading.value = true
    try {
      const [queue, history] = await Promise.all([
        api.getQueue(),
        api.getHistory(maxHistoryItems.value)
      ])
```

Then it reconciles the new state of the queue/running/history and updates its reactive `tasks`. The queue sidebar tab derives computed values that react to those `tasks`.

---

Two possible quick solutions to make inputs/outputs reactive across the app:

*solution 1*

* use the `useAssetsApi` gateway in the queueStore to get the appropriate fetchers (dependent on whether the new assets API is enabled in the environment)
* add `fetchAssets` to the queue store's `update` function like

```typescript
const update = async () => {
    isLoading.value = true
    try {
      const [queue, history, assets] = await Promise.all([
        api.getQueue(),
        api.getHistory(maxHistoryItems.value),
        fetchAssets(),
      ])
```

*solution 2*

* create an assetsStore which mirrors the queueStore
* add a listener in GraphView that mirrors the queueStore 'status' listener
* on status update, re-fetch the the outputs and have the store expose a reactive `outputs` (mirroring the reactive `tasks`
* can then even add triggers that would re-fetch inputs (e.g., the `/upload` method in `api.ts` was called)
* make the assets sidebar interact with assetsStore the same way queue sidebar interacts with queueStore

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -474,3 +474,72 @@ export function formatDuration(milliseconds: number): string {
 
   return parts.join(' ')
 }
+
+// Module scope constants to avoid re-initialization on every call
+const IMAGE_EXTENSIONS = ['png', 'jpg', 'jpeg', 'gif', 'webp', 'bmp']
+const VIDEO_EXTENSIONS = ['mp4', 'webm', 'mov', 'avi']
+const AUDIO_EXTENSIONS = ['mp3', 'wav', 'ogg', 'flac']
+const THREE_D_EXTENSIONS = ['obj', 'fbx', 'gltf', 'glb']
```
DrJKL:
```suggestion
const IMAGE_EXTENSIONS = ['png', 'jpg', 'jpeg', 'gif', 'webp', 'bmp'] as const
const VIDEO_EXTENSIONS = ['mp4', 'webm', 'mov', 'avi'] as const
const AUDIO_EXTENSIONS = ['mp3', 'wav', 'ogg', 'flac'] as const
const THREE_D_EXTENSIONS = ['obj', 'fbx', 'gltf', 'glb'] as const
```

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -474,3 +474,72 @@ export function formatDuration(milliseconds: number): string {
...
+}
+
+/**
+ * Determines the media type from a filename's extension (singular form)
+ * @param filename The filename to analyze
+ * @returns The media type: 'image', 'video', 'audio', or '3D'
+ */
+export function getMediaTypeFromFilename(
+  filename: string
+): 'image' | 'video' | 'audio' | '3D' {
```
DrJKL: You could follow the same pattern for this as the extensions
```ts
const MEDIA_TYPES = [ 'image' , 'video' , 'audio' , '3D' ] as const
type MediaType = typeof MEDIA_TYPES[number]
```

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -474,3 +474,72 @@ export function formatDuration(milliseconds: number): string {
...
+  if (IMAGE_EXTENSIONS.includes(ext)) return 'image'
+  if (VIDEO_EXTENSIONS.includes(ext)) return 'video'
+  if (AUDIO_EXTENSIONS.includes(ext)) return 'audio'
+  if (THREE_D_EXTENSIONS.includes(ext)) return '3D'
+
+  return 'image'
+}
+
+/**
+ * @deprecated Use getMediaTypeFromFilename instead - kept for backward compatibility
```
DrJKL: Optional
```suggestion
 * @deprecated Use {@link getMediaTypeFromFilename} instead - kept for backward compatibility
```

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: Yep, I think it doesn't have to be translated.

packages/shared-frontend-utils/src/formatUtil.ts
viva-jinyi: ✅

packages/shared-frontend-utils/src/formatUtil.ts
viva-jinyi: ✅

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: Thanks for pointing out. I will remember it.

packages/shared-frontend-utils/src/formatUtil.ts
viva-jinyi: ✅

packages/shared-frontend-utils/src/formatUtil.ts
viva-jinyi: ✅

packages/shared-frontend-utils/src/formatUtil.ts
viva-jinyi: ✅

viva-jinyi:
[screen-capture (1).webm](https://github.com/user-attachments/assets/f25adaab-1a03-4e66-8363-8612921e9906)

I created a new store called AssetStore, and it now reactively fetches data whenever a file is added to the input.
Also, the bugs that occurred in #2125 no longer happen.

src/components/tab/TabList.vue
```diff
@@ -0,0 +1,21 @@
...
+    <slot />
+  </div>
+</template>
+
+<script setup lang="ts">
+import { provide, toRef } from 'vue'
+
+const props = defineProps<{
+  modelValue: string
+}>()
```
DrJKL: Could this use [defineModel](https://vuejs.org/api/sfc-script-setup.html#definemodel)?

src/composables/sidebarTabs/useAssetsSidebarTab.ts
DrJKL, line 1: I don't usually like composables that aren't using refs or other composables.
Could this be a constant?

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -272,7 +326,15 @@ const handleOverlayMouseLeave = () => {
...
+    emit('zoom', asset)
   }
 }
+
+const handleImageLoaded = (width: number, height: number) => {
+  imageDimensions.value = { width, height }
+}
+
+const handleOutputCountClick = () => {
+  emit('output-count-click')
```
DrJKL, line 340: Is this not being consumed yet?

src/platform/assets/components/MediaImageBottom.vue
```diff
@@ -4,18 +4,20 @@
       class="m-0 line-clamp-1 text-sm font-bold text-zinc-900 dark-theme:text-white"
       :title="asset.name"
     >
-      {{ fileName }}
+      {{ truncateFilename(fileName) }}
```
DrJKL: It feels like there's a lot of duplication across these, could they be joined into a single component?

src/platform/assets/components/MediaImageTop.vue
```diff
@@ -1,27 +1,44 @@
...
+import { useImage, whenever } from '@vueuse/core'
 
 import type { AssetMeta } from '../schemas/mediaAssetSchema'
 
 const { asset } = defineProps<{
   asset: AssetMeta
 }>()
+
+const emit = defineEmits<{
+  'image-loaded': [width: number, height: number]
```
DrJKL, line 30: If it weren't in the litegraph section, I'd say you should use the `Size` type 🫤

src/platform/assets/composables/media/useMediaAssets.ts
DrJKL, line 1: I generally don't like `index.ts` files, especially if we're not using them in most cases.
What do you think about renaming the directory to be `media` and this to be `useMediaAssets.ts`?

src/stores/assetsStore.ts
```diff
@@ -0,0 +1,142 @@
...
+    inputLoading,
+    historyLoading,
+    inputError,
+    historyError,
+    isLoading,
+
+    // Actions
+    updateInputs,
+    updateHistory,
+    update
```
DrJKL: I like this store, just to check, all of these are used within this PR?

src/utils/uuidUtil.ts
```diff
@@ -0,0 +1,45 @@
...
+}
+
+/**
+ * Check if a string is a valid UUID (any version)
+ * @param str - The string to check
+ * @returns true if the string is a valid UUID
+ */
+export function isValidUuid(str: string): boolean {
+  return uuidValidate(str)
+}
```
DrJKL: Could we just use the `uuid` method in place wherever this is being called?

src/utils/uuidUtil.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+const UUID_REGEX =
+  /^([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})/i
+
+/**
+ * Extract UUID from the beginning of a string
+ * @param str - The string to extract UUID from
+ * @returns The extracted UUID or null if not found
+ */
+export function extractUuidFromString(str: string): string | null {
```
DrJKL: What strings do we have that start with a UUID but contain other text?

tests-ui/tests/utils/formatUtil.test.ts
```diff
@@ -0,0 +1,138 @@
...
+      const filename = 'my.file.with.multiple.dots.txt'
+      const truncated = truncateFilename(filename, 15)
+      expect(truncated.endsWith('.txt')).toBe(true)
+      expect(truncated).toContain('...')
+    })
+  })
+
+  describe('getMediaTypeFromFilename', () => {
+    describe('image files', () => {
+      it('should identify image extensions correctly', () => {
```
DrJKL: Did I already recommend [it.for](https://vitest.dev/api/#test-for)?

DrJKL (dismissed):

src/platform/assets/composables/media/useMediaAssets.ts
christian-byrne: This is the design I asked for to enable static adatper type of pattern per distribution.

christian-byrne:
> Also, the bugs that occurred in https://github.com/Comfy-Org/ComfyUI_frontend/pull/2125 no longer happen.

I can reproduce it on this branch. Steps:

1. Generate some images
2. Open Gallery with an image in the middle of the list
3. Use `Ctrl`+`Enter` to queue a job while the gallery is still open
4. the gallery changes automatically (it shoudln't)

src/composables/sidebarTabs/useAssetsSidebarTab.ts
christian-byrne: > I don't usually like composables that aren't using refs or other composables.

Can you explain reasoning to help others learn?

JorgeR81:
This looks nice, but it would be more useful if it had a search field. 
If you have more than 1000 images, it's not practical to use just by scrolling images.

You could also add ( to each image ), other information useful for the search:

- the name of the workflow in which image was generated
- The node's ID number in which image was generated


This last parameter is useful because complex workflows create lots of images ( e.g. Control Nets, multiple samplers, upscalers ), but in the gallery, we may only want to look at the final image.

I already asked for something similar for Queue sidebar, a while algo: 
- https://github.com/Comfy-Org/ComfyUI_frontend/issues/1502

viva-jinyi:
@JorgeR81 Thank you for your valuable feedback! We're planning to implement search and sort functionality in the next iteration. This will include:
- A search field to filter images by file type and workflow name
- Sorting options by date, and execution time
These features will make it much more practical to manage large collections of images.

tests-ui/tests/utils/formatUtil.test.ts
viva-jinyi: ✅

src/utils/uuidUtil.ts
viva-jinyi: ✅

src/platform/assets/components/MediaImageBottom.vue
viva-jinyi: ✅

src/components/tab/TabList.vue
viva-jinyi: ✅

src/platform/assets/components/MediaAssetCard.vue
viva-jinyi: It is being consumed - Used in AssetsSidebarTab line 62 to call enterFolderView(item) when users click the output count badge to see all outputs from a batch job.

src/stores/assetsStore.ts
viva-jinyi: I've removed the 2 unused exports (isLoading and update). All remaining exports are actively used across the codebase.

src/utils/uuidUtil.ts
viva-jinyi: I deleted this file!

viva-jinyi: ✅

DrJKL (approved):

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -0,0 +1,297 @@
...
+    const mediaType = getMediaTypeFromFilename(asset.name)
+    const resultItem = new ResultItemImpl({
+      filename: asset.name,
+      subfolder: '',
+      type: 'output',
+      nodeId: '0',
+      mediaType: mediaType === 'image' ? 'images' : mediaType
+    })
+
+    Object.defineProperty(resultItem, 'url', {
```
arjansingh, line 187: Non-blocker: You probably should extend `ResultItemImpl` to have optional `url` field. That's the better way to do this. Could also do a subclass.

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -0,0 +1,297 @@
...
+watch(displayAssets, (newAssets) => {
+  if (currentGalleryAssetId.value && galleryActiveIndex.value !== -1) {
+    const newIndex = newAssets.findIndex(
+      (asset) => asset.id === currentGalleryAssetId.value
+    )
+    if (newIndex !== -1) {
+      galleryActiveIndex.value = newIndex
+    }
+  }
+})
```
arjansingh, line 168: non-blocker: in general, looking for ways to avoid nesting logic will make the code more readable.

Double check the suggestion logic, but something like this:

```suggestion
watch(displayAssets, (newAssets) => {
  if (!currentGalleryAssetId.value || galleryActiveIndex.value == -1) return
  
  const newIndex = newAssets.findIndex((asset) => asset.id === currentGalleryAssetId.value)
    
  if (newIndex === -1) return
  
  galleryActiveIndex.value = newIndex
})
```

src/stores/assetsStore.ts
```diff
@@ -0,0 +1,133 @@
+import { useAsyncState } from '@vueuse/core'
```
arjansingh, line 1: follow-up: This should go in /platform/assets

arjansingh (commented):
Generally, I think if you reduced the number of components you could use inject/emit a little less and simplify the code.

Only follow up I have is the assetStore, we should definitely bring that into the platform directory for assets.

arjansingh (approved):

PR #6114, opened 2025-10-17
Add subgraph management keybinds

AustinMroz:
Adds a new keybind to attempt to toggle the promotion state of the currently hovered widget

Adds a keybind to open the subgraph configuration panel

Neither of these options are given a default binding.

Cleans up some incorrect version and icon information that was forgotten about.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6114-Add-subgraph-management-keybinds-28f6d73d365081e6b6f3c78f99afc094) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useCoreCommands.ts
+++ b/src/composables/useCoreCommands.ts
@@ -920,7 +920,7 @@ export function useCoreCommands(): ComfyCommand[] {
     },
     {
       id: 'Comfy.Graph.UnpackSubgraph',
-      icon: 'icon-[lucide--grow]',
+      icon: 'icon-[lucide--expand]',
       label: 'Unpack the selected Subgraph',
       versionAdded: '1.26.3',
       function: () => {
--- a/src/composables/useCoreCommands.ts
+++ b/src/composables/useCoreCommands.ts
@@ -5,7 +5,11 @@ import {
   DEFAULT_DARK_COLOR_PALETTE,
   DEFAULT_LIGHT_COLOR_PALETTE
 } from '@/constants/coreColorPalettes'
-import { promoteRecommendedWidgets } from '@/core/graph/subgraph/proxyWidgetUtils'
+import {
+  promoteRecommendedWidgets,
+  tryToggleWidgetPromotion
+} from '@/core/graph/subgraph/proxyWidgetUtils'
+import { showSubgraphNodeDialog } from '@/core/graph/subgraph/useSubgraphNodeDialog'
 import { t } from '@/i18n'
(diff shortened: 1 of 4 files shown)


src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -96,6 +97,21 @@ export function addWidgetPromotionOptions(
     })
   }
 }
+export function tryToggleWidgetPromotion() {
+  const canvas = useCanvasStore().getCanvas()
+  const [x, y] = canvas.canvas_mouse
```
benceruleanlu: question: canvas_mouse seems to be deprecated, should we replace with graph_mouse?

src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -96,6 +97,21 @@ export function addWidgetPromotionOptions(
...
+  const node = canvas.graph?.getNodeOnPos(x, y, canvas.visible_nodes)
+  if (!node) return
+  const widget = node.getWidgetOnPos(x, y, true)
+  const parents = getParentNodes()
```
benceruleanlu, line 116: nit: I think calling getParentNodes() will throw on root graph, which I guess isn't terribly critical? since it's a noop to the user, only in console but a change to warning or guarding could be nice

AustinMroz: Correct. This creates a nice error toast in the top-right, but I might revisit to add translation and limit the lifetime of the toast.

EDIT: Wording is a little unclear. Throwing an error inside a core command creates a toast. The prior state was not invisible to the user, but it's properly handled now.

src/composables/useCoreCommands.ts
```diff
@@ -931,6 +934,20 @@ export function useCoreCommands(): ComfyCommand[] {
         graph.unpackSubgraph(subgraphNode)
       }
     },
+    {
+      id: 'Comfy.Graph.EditSubgraphWidgets',
+      label: 'Toggle promotion of hovered widget',
```
benceruleanlu:
```suggestion
      label: 'Edit Subgraph Widgets',
```

benceruleanlu: nit: the label is the same as Comfy.Graph.ToggleWidgetPromotion

benceruleanlu (dismissed):

christian-byrne (approved):
LGTM

PR #6116, opened 2025-10-18
[ci] enable minification in build workflows

christian-byrne:
## Summary

Continuation of https://github.com/Comfy-Org/ComfyUI_frontend/pull/6068: set `SHOULD_MINIFY` in build workflows to enable minification.

Our build plugin dictates what files are available for extensions to import at runtime: https://github.com/Comfy-Org/ComfyUI_frontend/blob/bfe083dcbad61965b0ef04cf91d06a9cddaca361/build/plugins/comfyAPIPlugin.ts#L52

The shimming marks these files and their named exports and named memebrs as side-effectful, so names should be preserved from the perspective of 3rd party extensions, making this safe in theory.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6116-ci-enable-minification-in-build-workflows-2906d73d36508141be57f63b619f9f16) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/publish-desktop-ui.yaml
+++ b/.github/workflows/publish-desktop-ui.yaml
@@ -44,6 +44,7 @@ jobs:
       contents: read
     env:
       PLAYWRIGHT_SKIP_BROWSER_DOWNLOAD: '1'
+      ENABLE_MINIFY: 'true'
     steps:
       - name: Validate inputs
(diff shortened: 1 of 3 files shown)


christian-byrne:
Going to merge and test on a nightly release.

PR #6122, opened 2025-10-18
[refactor] change `isElectron` callsites to use isDesktop distribution type

christian-byrne:
## Summary

Simplify this so it can be converted into a build-time variable later, improving performance, testability, and unifying with cloud distribution pattern.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6122-refactor-change-isElectron-callsites-to-use-isDesktop-distribution-type-2906d73d365081cc9405e84683d0c965) by [Unito](https://www.unito.io)


christian-byrne:
Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/8710

PR #6135, opened 2025-10-19
Fix link resolution of virtual nodes

AustinMroz:
Some virtual nodes (like get/set nodes) perform link redirection at prompt resolution. The prior implementation incorrectly tried to return the source of the virtual link after resolution, but this causes things to break when the source of the virtual link is a subgraph IO.

Instead, this PR changes the code section to restart resolution from the destination of the virtual link so that the existing subgraph boundary resolution code is applied.

Edit: Also fix a bug with reconnection of complex/any types on conversion to subgraph.
- This second change gives me pause. The level of complexity suggests that there was an intentional reason to handle things in the prior (broken) way, but the adjacent subgraphInput code does already uses this new indexing methodology.
  - Traced down original commit. I'm now confident fix is correct.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6135-Fix-link-resolution-of-virtual-nodes-2916d73d36508183b891ef6eb39bad4c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraph.ts
+++ b/src/lib/litegraph/src/LGraph.ts
@@ -1665,7 +1665,7 @@ export class LGraph
         continue
       }
 
-      const input = subgraphNode.findInputSlotByType(link.type, true, true)
+      const input = subgraphNode.inputs[i - 1]
       outputNode.connectSlots(output, subgraphNode, input, link.parentId)
     }
 
--- a/src/lib/litegraph/src/subgraph/ExecutableNodeDTO.ts
+++ b/src/lib/litegraph/src/subgraph/ExecutableNodeDTO.ts
@@ -294,25 +294,21 @@ export class ExecutableNodeDTO implements ExecutableLGraphNode {
       // Fallback check for nodes performing link redirection
       const virtualLink = this.node.getInputLink(slot)
       if (virtualLink) {
-        const outputNode = this.graph.getNodeById(virtualLink.origin_id)
-        if (!outputNode)
+        const { inputNode } = virtualLink.resolve(this.graph)
+        if (!inputNode)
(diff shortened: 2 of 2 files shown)


src/lib/litegraph/src/LGraph.ts
```diff
@@ -1665,7 +1665,7 @@ export class LGraph
         continue
       }
 
-      const input = subgraphNode.findInputSlotByType(link.type, true, true)
+      const input = subgraphNode.inputs[i - 1]
```
DrJKL, line 1668: Does this need an undefined check?

src/lib/litegraph/src/subgraph/ExecutableNodeDTO.ts
```diff
@@ -294,25 +294,21 @@ export class ExecutableNodeDTO implements ExecutableLGraphNode {
       // Fallback check for nodes performing link redirection
       const virtualLink = this.node.getInputLink(slot)
       if (virtualLink) {
-        const outputNode = this.graph.getNodeById(virtualLink.origin_id)
-        if (!outputNode)
+        const { inputNode } = virtualLink.resolve(this.graph)
```
DrJKL, line 297: This makes more sense to me. I visualize Comfy stuff as pull based, usually.

DrJKL (approved):

christian-byrne (commented):
LGTM

src/lib/litegraph/src/LGraph.ts
AustinMroz: Probably not. None of the adjacent code  indexing with `i` includes checks and the subsequent `connectSlots` safely handles undefined inputs.

PR #6137, opened 2025-10-19
[perf] manually chunk vendored code

christian-byrne:
## Summary

Added a `manualChunks` strategy in `vite.config.mts` that splits primevue, tiptap, chart.js, three/@xterm, core Vue/Pinia, and the remaining dependencies into dedicated vendor bundles. This reduces the main application chunk size and allows browsers to cache heavy third-party code across releases, improving load times when those libraries stay unchanged.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6137-perf-manually-chunk-vendored-code-2916d73d36508140a44ec0de228ef9cc) by [Unito](https://www.unito.io)

### DIFF
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -188,7 +188,36 @@ export default defineConfig({
     target: 'es2022',
     sourcemap: GENERATE_SOURCEMAP,
     rollupOptions: {
-      treeshake: true
+      treeshake: true,
+      output: {
+        manualChunks: (id) => {
+          if (!id.includes('node_modules')) {
+            return undefined
+          }
+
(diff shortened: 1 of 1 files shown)


christian-byrne:
> App Entry Points 🟢 -5.37 MB (6.32 MB) · Vendor & Third-Party 🔴 +5.36 MB (5.36 MB) 

Moving to other chunks which don't need to be re-fetched everytime a new frontend version is released, improving page load for users. Downside is just a few additional requests one time.

DrJKL (approved):

PR #6144, opened 2025-10-19
make "require subscription" toggleable in build

christian-byrne:
## Summary

Adds build time feature flags system starting with a flag that indicates whether subscription is required to use the app. This is only used on cloud.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6144-make-require-subscription-toggleable-in-build-2916d73d3650813bb140c5e96bcce1ce) by [Unito](https://www.unito.io)


jtydhr88 (approved):
LGTM

PR #6149, opened 2025-10-19
[style] unify Vue widget/slot label colors

christian-byrne:
## Summary

Change labels on all widgets and slots to the same value which matches design spec.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6149-style-unify-Vue-widget-slot-label-colors-2916d73d3650810a98f3ee75e0b22da0) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/form/FormSelectButton.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/form/FormSelectButton.vue
@@ -23,7 +23,7 @@
           },
           {
             'text-neutral-900': isSelected(option) && !disabled,
-            'text-zinc-500': !isSelected(option) || disabled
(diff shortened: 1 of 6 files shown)


src/renderer/extensions/vueNodes/widgets/components/form/FormSelectButton.vue
```diff
@@ -23,7 +23,7 @@
           },
           {
             'text-neutral-900': isSelected(option) && !disabled,
-            'text-zinc-500': !isSelected(option) || disabled
+            'text-secondary': !isSelected(option) || disabled
```
DrJKL: Want to use a ternary instead?

DrJKL (dismissed):

DrJKL (approved):

PR #6157, opened 2025-10-20
[bugfix] Fix enable pack functionality to use proper API endpoint

viva-jinyi:
## Summary
- Fixed issue where enabling a disabled pack incorrectly triggered installation instead of using the dedicated enable endpoint
- Added proper `enablePack` method in the manager service layer
- Updated store and component to use the correct API call

## Problem
When users toggled a disabled pack to enable it via `PackEnableToggle` component, the system was incorrectly calling the install endpoint with full installation parameters instead of the simpler enable endpoint that only requires the pack ID.

## Solution
1. **Added dedicated `enablePack` method in `comfyManagerService.ts`**:
   - Uses the `'enable'` task kind with `EnablePackParams`
   - Only requires `cnr_id` parameter (simpler than install)

2. **Updated `comfyManagerStore.ts`**:
   - Created proper `enablePack` function that queues an enable task
   - Removed the incorrect aliasing where `enablePack` was pointing to `installPack`

3. **Simplified `PackEnableToggle.vue`**:
   - Now calls `enablePack` with only required parameters (id, version)
   - Removed unnecessary installation-specific parameters

## Test plan
- [x] Enable a disabled pack and verify it uses the enable endpoint (not install)
- [x] Confirm the pack is properly enabled after the operation
- [x] Check that the task queue shows "Enabling" message (not "Installing")
- [x] Verify existing install/uninstall functionality still works

Fixes #6154

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6157-bugfix-Fix-enable-pack-functionality-to-use-proper-API-endpoint-2926d73d3650819fa4caf1b848f99735) by [Unito](https://www.unito.io)

### DIFF
--- a/src/workbench/extensions/manager/components/manager/button/PackEnableToggle.test.ts
+++ b/src/workbench/extensions/manager/components/manager/button/PackEnableToggle.test.ts
@@ -31,7 +31,7 @@ const mockNodePack = {
 }
 
 const mockIsPackEnabled = vi.fn()
-const mockEnablePack = { call: vi.fn().mockResolvedValue(undefined) }
+const mockEnablePack = vi.fn().mockResolvedValue(undefined)
 const mockDisablePack = vi.fn().mockResolvedValue(undefined)
 vi.mock('@/workbench/extensions/manager/stores/comfyManagerStore', () => ({
   useComfyManagerStore: vi.fn(() => ({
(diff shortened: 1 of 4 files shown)


christian-byrne (approved):
LGTM

PR #6167, opened 2025-10-20
[fix] prevent duplicate verification emails on page refresh

viva-jinyi:
## Summary
- Fixed duplicate verification email issue where emails were sent every time users returned to the root page
- Emails are now only sent automatically when coming from signup/login flow
- Added proper toast notifications to cloud onboarding pages

## Changes
- **Conditional email sending**: Only send verification email when `fromAuth=true` query parameter is present (from signup/login flow)
- **Auto-cleanup**: Remove `fromAuth` parameter after sending email to prevent re-sending on page refresh
- **Toast system fix**: 
  - Added `GlobalToast` component to `CloudLayoutView` for proper toast display in onboarding pages
  - Migrated from PrimeVue `useToast()` to ComfyUI's `useToastStore()`
- **UI improvements**:
  - Better spacing and layout for email verification page
  - Added multiline support for tips and instructions
  - Improved toast messages with clearer titles and summaries

## Problem it solves
Previously, when users signed up and received a verification email, every time they navigated back to the root page (`/`), the router guard would redirect them to the email verification page which would automatically send another email. This caused multiple emails to be sent, often ending up in spam folders.

## Test plan
- [x] Sign up for a new account → Should receive ONE verification email
- [x] Navigate away and back to root → Should NOT receive another email
- [x] Click "Resend email" button → Should receive a new email
- [x] Refresh the verification page → Should NOT receive another email
- [x] Toast notifications appear correctly in all auth flows

[screen-capture (1).webm](https://github.com/user-attachments/assets/25ffad94-d129-4051-b29e-5bdec696cd11)


benceruleanlu (approved):
👍

christian-byrne (approved):

PR #6172, opened 2025-10-20
custom_node provided blueprints

AustinMroz:
Frontend implementation for a system to allow custom_nodes to provide a set of subgraph blueprints.

Requires comfyanonymous/ComfyUI#10438, but shouldn't break anything if merged first

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6172-Early-POC-custom_node-provided-blueprints-2926d73d3650814982ecd43f12abd873) by [Unito](https://www.unito.io)


AustinMroz:
Updated blueprints to load asynchronously and verified correctness by adding artificial delay to the backend.
There's no indicator that blueprints are still loading, but this ever mattering would prove the need. I don't suspect this is currently an actual issue, but saving even fractions of a second on load times helps.

christian-byrne (dismissed):
LGTM

christian-byrne:
Fixed merge conflict with other PR that added function to api.ts in similar location.

Issue #5705, opened 2025-09-20
When the link inside the subgraph is misaligned with the input.


comfyui-wiki:
<img width="2060" height="1322" alt="Image" src="https://github.com/user-attachments/assets/acd5a9d8-363e-49e1-b8f4-9c674c31df69" />

christian-byrne:
IS this fixed?

PR #6181, opened 2025-10-21
Fix links to wrong slots in vue mode

AustinMroz:
Error is divided into 2 parts
- A widget lacking slotMetadata would create a slot overwriting the index 0 slot
- Slot data wasn't reactive. Any dynamic widgets would not have slotMetadata
image

See #5705
- Describes incorrect links internal to subgraphs. Likely a different, already solved issue

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6181-Fix-links-to-wrong-slots-in-vue-mode-2936d73d36508136ad43d8c818bf9fba) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -126,14 +126,6 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
     // Extract safe widget data
     const slotMetadata = new Map<string, WidgetSlotMetadata>()
 
-    node.inputs?.forEach((input, index) => {
-      if (!input?.widget?.name) return
-      slotMetadata.set(input.widget.name, {
-        index,
-        linked: input.link != null
-      })
-    })
-
     const reactiveWidgets = shallowReactive<IBaseWidget[]>(node.widgets ?? [])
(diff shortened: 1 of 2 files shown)


src/composables/graph/useGraphNodeManager.ts
```diff
@@ -144,8 +136,15 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
       }
     })
 
-    const safeWidgets = reactiveComputed<SafeWidgetData[]>(
-      () =>
+    const safeWidgets = reactiveComputed<SafeWidgetData[]>(() => {
```
christian-byrne, line 139: By chance do you know how often this is recomputed during regular usage?

christian-byrne (approved):
LGTM

src/composables/graph/useGraphNodeManager.ts
AustinMroz: From testing, only when there's a change in widget order when in vue mode! It feels like a textbook example for the benefits of computed properties as a system.

christian-byrne: Nice, that's good.

Issue #6151, opened 2025-10-19
[perf] Performance in new Vue Nodes system is bottlenecked by interactions/events


christian-byrne:
This is a good thing, because it's one of the easiest things to fix.
 
If I load a workflow with a huge number of nodes and images, the performance seems to degrade during navigation.

However, if I navigate with the minimap, I can get the performance again from the early days (60 fps on giant graphs).

To me, this suggests there's some huge bottle necks in the interaction/event handling on transform pane and its children. As using the minimap to navigate involves all the same systems as it relates to rendering - the only difference is the interaction flow.

This aligns with the previous profiling done showing event listener bottlenecks.

It's good that the performance can still be quite good after we clean this up. I.e., we are just a few optimizations away from restoring the original performance which generally out-competes all DOM based competitor at least for navigation.

https://github.com/user-attachments/assets/4e7eecc9-0b6e-4cf0-bed5-15dbbb5034eb

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6151-perf-Performance-in-new-Vue-Nodes-system-is-bottlenecked-by-interactions-events-2916d73d365081ecbcb7ce8969c7df7c) by [Unito](https://www.unito.io)

PR #6186, opened 2025-10-21
disable transform settling reflow when panning the graph

christian-byrne:
## Summary

- disable pan tracking in `useTransformSettling` so we stop wiring high-frequency pointer listeners during canvas drags
- the post-navigation-interaction forced reflow is only necessary when zooming since it is for fixing pixel stretch that results from `scale` (which doesn't happen during panning/`translate`)
- extend settle delay to 512ms to reduce unnecessary reflow while preserving post-zoom pixel fix

After this PR, there should be 0 reflows when panning the graph.

First PR in series to address:

- https://github.com/Comfy-Org/ComfyUI_frontend/issues/6151

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6186-disable-transform-settling-reflow-when-panning-the-graph-2936d73d365081c2b357e3c72d711439) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/core/layout/transform/TransformPane.vue
+++ b/src/renderer/core/layout/transform/TransformPane.vue
@@ -45,8 +45,8 @@ const { isLOD } = useLOD(camera)
 
 const canvasElement = computed(() => props.canvas?.canvas)
 const { isTransforming: isInteracting } = useTransformSettling(canvasElement, {
-  settleDelay: 200,
-  trackPan: true
+  settleDelay: 512,
+  trackPan: false // panning doesn't require a reflow because there's no pixel-stretching
 })
 
 provide(TransformStateKey, {
--- a/src/renderer/core/layout/transform/TransformPane.vue
+++ b/src/renderer/core/layout/transform/TransformPane.vue
@@ -45,8 +45,7 @@ const { isLOD } = useLOD(camera)
 
 const canvasElement = computed(() => props.canvas?.canvas)
 const { isTransforming: isInteracting } = useTransformSettling(canvasElement, {
-  settleDelay: 512,
-  trackPan: false // panning doesn't require a reflow because there's no pixel-stretching
+  settleDelay: 512
 })
 
 provide(TransformStateKey, {
--- a/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts
+++ b/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts
@@ -1,3 +1,4 @@
+import { createPinia, setActivePinia } from 'pinia'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { nextTick, ref } from 'vue'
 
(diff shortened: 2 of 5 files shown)


src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -45,8 +45,8 @@ const { isLOD } = useLOD(camera)
 
 const canvasElement = computed(() => props.canvas?.canvas)
 const { isTransforming: isInteracting } = useTransformSettling(canvasElement, {
-  settleDelay: 200,
-  trackPan: true
+  settleDelay: 512,
+  trackPan: false // panning doesn't require a reflow because there's no pixel-stretching
```
DrJKL: This defaults to false and it looks like this is the only place it's set. Can we just get rid of the option?

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts
```diff
@@ -71,11 +72,17 @@ const createMouseEvent = (
...
     vi.clearAllMocks()
+    setActivePinia(createPinia())
     // Reset layout store state between tests
     const { layoutStore } = await import(
       '@/renderer/core/layout/store/layoutStore'
     )
     layoutStore.isDraggingVueNodes.value = false
+    const { useCanvasStore } = await import(
+      '@/renderer/core/canvas/canvasStore'
+    )
```
DrJKL: Should we be mocking this?

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts
```diff
@@ -71,11 +72,17 @@ const createMouseEvent = (
...
+    const canvasStore = useCanvasStore()
+    canvasStore.setVueNodePointerInteractionsDisabled(false)
```
DrJKL: Missing commit?

src/renderer/core/layout/transform/TransformPane.vue
christian-byrne: True, that's better.

christian-byrne: Removed in f9cd45e

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts
christian-byrne: Removed the redundant canvas store setup in e44a4f8. The canvas interaction behavior is already properly mocked via the useCanvasInteractions mock.

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts
christian-byrne: Fixed in e44a4f8. Removed the call to non-existent method since canvas interactions are already mocked.

PR #6189, opened 2025-10-22
load assets browser before fetch completes and show loading state

christian-byrne:
## Summary

Moves the fetch and post-fetch logic associated with the asset browser into the component and shows a loading state while fetching.

To test, use this branch: https://github.com/comfyanonymous/ComfyUI/pull/10045

https://github.com/user-attachments/assets/718974d5-efc7-46a0-bcd6-e82596d4c389

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6189-load-assets-browser-before-fetch-completes-and-show-loading-state-2946d73d365081879d1bd05d86e8c036) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/assets/components/AssetBrowserModal.vue
+++ b/src/platform/assets/components/AssetBrowserModal.vue
@@ -136,7 +136,7 @@ const primaryCategoryTag = computed(() => {
 
   if (props.assetType) return props.assetType
 
-  return t('assetBrowser.allModels')
+  return 'models'
 })
 
 const activeCategoryTag = computed(() => {
--- a/src/platform/assets/components/AssetBrowserModal.vue
+++ b/src/platform/assets/components/AssetBrowserModal.vue
@@ -40,6 +40,7 @@
     <template #content>
       <AssetGrid
         :assets="filteredAssets"
+        :loading="isLoading"
         @asset-select="handleAssetSelectAndEmit"
       />
     </template>
@@ -47,7 +48,9 @@
 </template>
 
 <script setup lang="ts">
-import { computed, provide } from 'vue'
+import { useAsyncState } from '@vueuse/core'
+import { computed, provide, watch } from 'vue'
+import { useI18n } from 'vue-i18n'
 
 import SearchBox from '@/components/input/SearchBox.vue'
 import BaseModalLayout from '@/components/widget/layout/BaseModalLayout.vue'
@@ -57,6 +60,9 @@ import AssetGrid from '@/platform/assets/components/AssetGrid.vue'
 import type { AssetDisplayItem } from '@/platform/assets/composables/useAssetBrowser'
 import { useAssetBrowser } from '@/platform/assets/composables/useAssetBrowser'
 import type { AssetItem } from '@/platform/assets/schemas/assetSchema'
+import { assetService } from '@/platform/assets/services/assetService'
+import { formatCategoryLabel } from '@/platform/assets/utils/categoryLabel'
+import { useModelToNodeStore } from '@/stores/modelToNodeStore'
 import { OnCloseKey } from '@/types/widgetTypes'
 
 const props = defineProps<{
@@ -65,9 +71,11 @@ const props = defineProps<{
   onSelect?: (asset: AssetItem) => void
   onClose?: () => void
   showLeftPanel?: boolean
-  assets?: AssetItem[]
   title?: string
+  assetType?: string
 }>()
+
+const { t } = useI18n()
 
 const emit = defineEmits<{
   'asset-select': [asset: AssetDisplayItem]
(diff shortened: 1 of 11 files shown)


src/platform/assets/components/AssetBrowserModal.vue
```diff
@@ -65,29 +71,86 @@ const props = defineProps<{
...
+  execute
+} = useAsyncState<AssetItem[]>(fetchAssets, [], { immediate: false })
+
+watch(
+  () => [props.nodeType, props.assetType],
+  async () => {
+    await execute()
+  },
+  { immediate: true }
+)
```
arjansingh, line 111: nit: instead of `{ immediate: true }` I think could you do something like this to make the lifecycle more explicit:

```ts
onMount(() => execute())

watch(() => [props.nodeType, props.assetType], () => execute())
```

arjansingh (dismissed):
One nit (not blocking) otherwise we're good.

src/platform/assets/components/AssetBrowserModal.vue
arjansingh: I see you @DrJKL!

DrJKL: 😜

christian-byrne: Thanks for the review! After merging with main, this code has been superseded. The component now accepts `assets` as a prop instead of fetching internally, so the `useAsyncState` + `watch({immediate: true})` pattern is no longer present.

For context on the original implementation: `watch({immediate: true})` runs during component setup (before DOM mount), while `onMounted()` runs after mount. For data fetching that doesn't need DOM access, starting earlier provides better perceived performance.

src/platform/assets/schemas/assetSchema.ts
```diff
@@ -4,9 +4,9 @@ import { z } from 'zod'
 const zAsset = z.object({
   id: z.string(),
   name: z.string(),
-  asset_hash: z.string().optional(),
+  asset_hash: z.string().nullish(),
```
DrJKL, line 7: Changing something from optional to required but nullable can be risky.

tests-ui/platform/assets/components/AssetBrowserModal.test.ts
```diff
@@ -1,32 +1,40 @@
...
+import { flushPromises, mount } from '@vue/test-utils'
 import { createPinia, setActivePinia } from 'pinia'
-import { describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import AssetBrowserModal from '@/platform/assets/components/AssetBrowserModal.vue'
 import type { AssetItem } from '@/platform/assets/schemas/assetSchema'
 
-// Mock @/i18n for useAssetBrowser and AssetFilterBar
+const mockAssetService = vi.hoisted(() => ({
```
DrJKL, line 8: Ooooh, very nice.

src/platform/assets/composables/useAssetBrowser.ts
```diff
@@ -67,7 +68,10 @@ export interface AssetDisplayItem extends AssetItem {
  * Asset Browser composable
  * Manages search, filtering, asset transformation and selection logic
  */
-export function useAssetBrowser(assets: AssetItem[] = []) {
+export function useAssetBrowser(
+  assetsSource: Ref<AssetItem[] | undefined> = ref<AssetItem[] | undefined>([])
```
DrJKL, line 70: Bit of a nit, but I like to make composables more flexible by using `MaybeRefOrGetter` and `toValue`.
It'd also have saved you from having to update all the tests that were passing in pure values instead of refs.

DrJKL (dismissed):

src/platform/assets/components/AssetBrowserModal.vue
arjansingh: Thank you for the explanation.

src/platform/assets/composables/useAssetBrowser.ts
christian-byrne: > make composables more flexible by using MaybeRefOrGetter and toValue

YAGNI

src/platform/assets/schemas/assetSchema.ts
christian-byrne: On the OSS, it returns null, on the cloud, it's undefined. Thanks for flagging this out - cc @arjansingh

src/platform/assets/components/AssetBrowserModal.vue
christian-byrne: Claude got slightly confused on this one.

PR #6194, opened 2025-10-22
Introduce grow-parent class for widgets

AustinMroz:
Only some widgets actually want to grow. Flex makes this difficult. This PR sets up a `widget-expands` class that widgets can use to indicate that they want to dynamically grow and applies it to the textarea widget
image

There's a potential alternative avenue with using `flex-shrink` instead of `flex-grow`, and using `min-content` for `calculateIntrinsicSize`, but I've been poking around with that for a while with no success.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6194-Introduce-grow-parent-class-for-widgets-2946d73d3650812f9d03c305ab04e212) by [Unito](https://www.unito.io)


DrJKL (commented):
nit: How do you feel about `.widget-expands` instead? That way the class naming describes the thing it labels, then the parent _chooses_ to respond based on the class inside.

DrJKL:
Would the image/video previews also be candidates for this behavior?

AustinMroz:
> nit: How do you feel about `.widget-expands` instead? That way the class naming describes the thing it labels, then the parent _chooses_ to respond based on the class inside.

I like it. It makes it more obvious that the class is something specific to ComfyUI and not tailwind or primevue.

> Would the image/video previews also be candidates for this behavior?

It was a consideration, yes. As well as Legacy DOM/canvas widgets. Probably best left for a future PR, though.

DrJKL (dismissed):

DrJKL (approved):

AustinMroz:
Yeah... There's something strange going on with Load Image (as Mask). I'll look into it after cleaning up other in-flight PRs.

image

PR #6206, opened 2025-10-22
style: remove pulsing animation on executing Vue nodes

christian-byrne:
## Summary

Removes pulsing animation which wasn't part of the original design and has performance overhead.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6206-style-remove-pulsing-animation-on-executing-Vue-nodes-2946d73d3650816ab877da8120ab4085) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/LGraphNode.vue
+++ b/src/renderer/extensions/vueNodes/components/LGraphNode.vue
@@ -18,7 +18,6 @@
         borderClass,
         outlineClass,
         {
-          'animate-pulse': executing,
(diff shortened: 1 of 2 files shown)


DrJKL (approved):
Do we have another way to indicate from a distance which node is "executing"?

christian-byrne:
> Do we have another way to indicate from a distance which node is "executing"?

Yeah there's a progress bar. For low LOD, there's a discussion about it on Figma in the LOD section of Node v3 design file.

PR #6209, opened 2025-10-22
Speed: Remove some optimizations that weren't optimizing

DrJKL:
## Summary

Simplify the TransformPane.

## Changes

- **What**: Remove the settling and culling composables. Gets rid of a frequent event emission and some event handling addition/removals.

## Review Focus

In testing with a huge workflow in Vue mode, it was a lot faster without these than with.
Can you check to see if you experience the same benefits?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6209-Speed-Remove-some-optimizations-that-weren-t-optimizing-2946d73d3650815197a4df3c58a61575) by [Unito](https://www.unito.io)


src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -3,8 +3,7 @@
     data-testid="transform-pane"
     :class="
       cn(
-        'absolute inset-0 w-full h-full pointer-events-none',
-        isInteracting ? 'transform-pane--interacting' : 'will-change-auto',
+        'absolute inset-0 w-full h-full pointer-events-none will-change-auto',
```
simula-r: IIRC transform-pane--interacting, will-change: transform which served 2 purposes:
1. A browser hint to promote the layer to the compositor. Without will-change: transform what are we relying on to ensure promotion? Is the css transform declaration on the transform pane itself enough? https://webperf.tips/tip/layers-and-compositing/#when-are-layers-created
2. The flip back to will-change: auto forced a reflow and subsequent paint which ensured at close zoom pixel stretch wouldn't be noticeable. @christian-byrne I remember earlier conversations where I was unsure if it was needed but found a few examples where without it, the node's text and borders were significantly stretched. Can we confirm this is not needed now?

src/components/graph/GraphCanvas.vue
```diff
@@ -196,7 +194,6 @@ const { shouldRenderVueNodes } = useVueFeatureFlags()
 
 // Vue node system
 const vueNodeLifecycle = useVueNodeLifecycle()
-const { handleTransformUpdate } = useViewportCulling()
```
simula-r, line 205: Viewport Culling was one of if not the most needle moving performance improvement to large workflows (removing nodes from the layout tree via display: none in order to shrink the layer bounds) from the performance dev tool flame graph traces I ran.

simula-r: I compared this branch with main (that has viewport culling) and in 300+ node large workflows when zoomed in helps the responsiveness by a considerable amount. What tests did you use to determine it was unnecessary?

DrJKL: I tested with a workflow with 896 nodes and it was a lot smoother to both zoom and pan.

christian-byrne: This could definitely vary across OS/Browser/Browser settings.

src/renderer/core/layout/transform/TransformPane.vue
DrJKL: Can you confirm that you checked and couldn't find an issue here?

src/components/graph/GraphCanvas.vue
DrJKL: I've tested this with several people on different systems, there's near universally better subjective performance when panning and zooming without the culling.

src/renderer/core/layout/transform/TransformPane.vue
simula-r: Yes. Confirmed. Cheers.

src/components/graph/GraphCanvas.vue
simula-r: Interesting - I still see a 10-15fps gain when zoomed in close on large workflows, but no real difference when zoomed out. Happy to cut it / simplify if it's not a big win either way.

simula-r (approved):

PR #4849, opened 2025-08-08
[fix] Add bounds checking for clipspace indices to prevent paste errors

christian-byrne:
Fixes #4838. The clipspace paste operation was failing when copying from regular Load Image nodes because combinedIndex was accessing out-of-bounds array indices. Added proper bounds checking in both app.ts and maskeditor.ts to handle cases where the imgs array has fewer elements than expected.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-4849-fix-Add-bounds-checking-for-clipspace-indices-to-prevent-paste-errors-2496d73d3650811f94e6cf2c400ff186) by [Unito](https://www.unito.io)

PR #6210, opened 2025-10-23
Fix copying from multi-outputs with clipspace

### LINKED ISSUE #6202: Clipspace Copy + Paste from a "Preview Image" node with multiple images to a "Load Image" node will paste the correct image but *show* the wrong image
### Prerequisites

### What happened?

After generating multiple images and showing them in a "Preview Image" node (or any similar node), performing in a clipspace copy and paste operation from that node to a "Load Image" node will, most of the time, cause the "Load Image" node to display the wrong image.

The more images are in the "Preview Image" node, the more likely the issue is to occur. At 9 images, it almost always happens.

*Note that the "Load Image" node will still **use** the correct image, it just **displays** the wrong one.*

Switching to a different workflow and coming back fixes the issue (the "Load Image" node now displays the correct image).

This issue is relatively recent, having started a few versions ago. Unfortunately, I couldn't tell you when it started exactly.

### Steps to Reproduce

1. Generate multiple images by setting `batch_size` to a number over 4 in the "Empty Latent Image" node. (9 is a good choice).
2. Select one of the images in the "Preview Image" node.
3. Right click the node and select "Copy (Clipspace)".
4. Right click on the "Load Image" node and select "Paste (Clipspace)".
    *The wrong image may be displayed. Try a different one if the issue does not occur.*
5. Navigate to a different workflow and then navigate back.
    *The correct image is now displayed.*

[Example Workflow](https://github.com/user-attachments/files/23051114/Example.Workflow.json)
[Example Video](https://github.com/user-attachments/assets/b1d86d74-2c6a-48c8-b51b-431b45ce7357)

Causes confusion, even the "Load Image" node works as intended otherwise.

### ComfyUI Frontend Version

1.28.7

### Browser

Chrome/Chromium

### Console Errors

```javascript
*No errors in console.*
```

### Logs

```shell
*No errors in terminal.*
```

### Additional Context

_No response_

AustinMroz:
`copyToClipspace` sets the `paintedIndex` and `combinedIndex` regardless of if any editing was actually performed on the image. Previously, they  were set to be simple increments above the currently selected image. 

When attempting to copy from a node previewing multiple images, these indexes may refer to images that already exist.
This is resolved by instead setting the indexes to be used for paintedIndex and combinedIndex to always be beyond the range of current images.
image

Resolves #6202

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6210-Fix-copying-from-multi-outputs-with-clipspace-2956d73d365081c5a07be4ddb0b352a3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -379,8 +379,8 @@ export class ComfyApp {
       selectedIndex = node.imageIndex
     }
 
-    const paintedIndex = selectedIndex + 1
-    const combinedIndex = selectedIndex + 2
+    const paintedIndex = imgs ? imgs.length + 1 : 1
+    const combinedIndex = imgs ? imgs.length + 2 : 2
 
     // for vueNodes mode
     const images =


christian-byrne:
Related to

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/4849

christian-byrne (approved):

PR #6212, opened 2025-10-23
Feat/vue nodes arrange alg

simula-r:
### Problem: 
The Vue nodes renderer/feature introduces new designs for each node i.e. the equivalent Litegraph node design is smaller and the vue node design is non uniformly larger.

### Example: 

Litegraph Ksampler node: 200w x 220h

image

Vue Node Ksampler node: 445w x 430h 

image

This means if users load a workflow in Litegraph and then switches to Vue nodes renderer the nodes are using the same Litegraph positions which would cause a visual overlap and overall look broken.

### Example:

image
image

### Solution:

Scale the positions of the nodes in lite graph radially from the center of the bounds of all nodes. And then simply move the Vue nodes to those new positions.

1. Get the `center of the bounds of all LG nodes`.
2. Get the `xy of each LG node`.
3. Get the vector from `center of the bounds of all LG nodes` `-` `xy of each LG node`.
4. Scale it by a factor (e.g. 1.75x which is the average Vue node size increase plus some visual padding.)
5. Move each Vue node to the scaled `xy of each LG node`. 

Result: The nodes are spaced apart removing overlaps while keeping the spatial layout intact.

image

### Further concerns.

This vector scaling algorithm needs to run once per workflow when in vue nodes. This means when in Litegraph and switching to Vue nodes, it needs to run before the nodes render. And then now that the entire app is in vue nodes, we need to run it each time we load a workflow. However, once its run, we do not need to run it again. Therefore we must persist a flag that it has run somewhere. This PR also adds that feature by leveraging the `extra` field in the workflow schema.

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -30,7 +30,6 @@ import {
   type NodeId,
   isSubgraphDefinition
 } from '@/platform/workflow/validation/schemas/workflowSchema'
-import { useFixVueNodeOverlap } from '@/renderer/extensions/vueNodes/composables/useFixVueNodeOverlap'
 import type {
   ExecutionErrorWsMessage,
   NodeError,
@@ -100,6 +99,7 @@ import { $el, ComfyUI } from './ui'
 import { ComfyAppMenu } from './ui/menu/index'
 import { clone } from './utils'
 import { type ComfyWidgetConstructor } from './widgets'
+import { scaleLayoutForVueNodes } from '@/renderer/extensions/vueNodes/layout/scaleLayoutForVueNodes'
 
 export const ANIM_PREVIEW_WIDGET = '$$comfy_animation_preview'
 
@@ -1191,7 +1191,7 @@ export class ComfyApp {
       }
 
       if (vueMode && !this.graph.extra.vueNodesScaled) {
-        useFixVueNodeOverlap()
+        scaleLayoutForVueNodes()
         this.graph.extra.vueNodesScaled = true
       }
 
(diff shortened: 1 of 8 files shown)


src/scripts/app.ts
```diff
@@ -1180,6 +1182,20 @@ export class ComfyApp {
...
+      if (!this.graph.extra) {
+        this.graph.extra = {}
+      }
+
+      // Run overlap fix only if in Vue mode and not already scaled
+      if (vueMode && !this.graph.extra.vueNodesScaled) {
+        useFixVueNodeOverlap()
+        this.graph.extra.vueNodesScaled = true
+      }
+
```
christian-byrne, line 1197: So in theory we could check `!vueMode && this.graph.extra.vueNodesScaled` (someone loading a vue nodes workflow in litegraph mode), then only after some change(s) are made -> set `vueNodesScaled = false`.

But it's not trivial to implement and can be done later.

christian-byrne (dismissed):
LGTM

src/scripts/app.ts
simula-r: Yep exactly. Also there is some hashing stuff I want to try. But warrants a larger discussion around workflow file version and future considerations.

christian-byrne:
For the tests, I wonder if we should have a flag or setting to disable the arrange. It seems to change all the baselines haha

christian-byrne (approved):

DrJKL:
😢

simula-r:
Re: @DrJKL 😢
I have a new potential enhancement will make in a follow up PR :)

PR #6231, opened 2025-10-23
Show default node list after clearing search input

### LINKED ISSUE #4887: [Bug]: Searchbox doesn't show default node list after clearing search input
### Frontend Version

main

### Expected Behavior

When the searchbox input is cleared (after typing and then deleting all text), it should display the default list of nodes, consistent with the behavior when the searchbox initially opens with an empty input.

### Actual Behavior

- When searchbox initially opens: Shows a default list of available nodes
- After typing something and then deleting all text to get back to empty state: No search options are displayed

This creates an inconsistent user experience where the same empty input state produces different results.

### Steps to Reproduce

1. Open the searchbox (double-click on canvas or drag a link and release)
2. Observe that a default list of nodes is displayed
3. Type any text in the search input (e.g., "image")
4. Delete all the text using backspace/delete to return to an empty input
5. Observe that no nodes are displayed, unlike the initial empty state

### Debug Logs

_No response_

### Browser Logs

_No response_

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4887-Bug-Searchbox-doesn-t-show-default-node-list-after-clearing-search-input-24b6d73d36508181b975cd94237ad360) by [Unito](https://www.unito.io)

AustinMroz:
searchbox-clear
Resolves #4887 

<details>
<summary>The dirty details</summary>
There's really not a cleaner workaround here. Prime vue is hardcoded to hide results when a query is received with length 0. With our search box, we never want completions not to be shown and the sanest, if not the only viable solution, is to simply block the hiding completely.

Future TODO:
- Completely remove the reFocusInput jank. If we can make the search box a frame more responsive, we should.

```ts
onInput(event) {
    if (this.typeahead) {
        if (this.searchTimeout) {
            clearTimeout(this.searchTimeout);
        }

        let query = event.target.value;

        if (!this.multiple) {
            this.updateModel(event, query);
        }

        if (query.length === 0) {
            this.hide();
            this.$emit('clear');
        } else {
            if (query.length >= this.minLength) {
                this.focusedOptionIndex = -1;

                this.searchTimeout = setTimeout(() => {
                    this.search(event, query, 'input');
                }, this.delay);
            } else {
                this.hide();
            }
        }
    }
}
hide(isFocus) {
    const _hide = () => {
        this.$emit('before-hide');
        this.dirty = isFocus;
        this.overlayVisible = false;
        this.clicked = false;
        this.focusedOptionIndex = -1;

        isFocus && focus(this.multiple ? this.$refs.focusInput : this.$refs.focusInput?.$el);
    };

    setTimeout(() => {
        _hide();
    }, 0); // For ScreenReaders
}
```
</details>

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6231-Show-default-node-list-after-clearing-search-input-2956d73d36508134960df537c7409646) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/searchbox/NodeSearchBox.vue
+++ b/src/components/searchbox/NodeSearchBox.vue
@@ -35,6 +35,7 @@
     </Dialog>
 
     <AutoCompletePlus
+      ref="acpref"
       :model-value="filters"
       class="comfy-vue-node-search-box z-10 grow"
       scroll-height="40vh"
@@ -42,7 +43,6 @@
       :input-id="inputId"
       append-to="self"
       :suggestions="suggestions"
-      :min-length="0"
       :delay="100"
       :loading="!nodeFrequencyStore.isLoaded"
       complete-on-focus
@@ -50,6 +50,7 @@
       force-selection
       multiple
       :option-label="'display_name'"
+      @clear="search('')"
       @complete="search($event.query)"
       @option-select="emit('addNode', $event.value)"
       @focused-option-changed="setHoverSuggestion($event)"
@@ -105,6 +106,7 @@ const { filters, searchLimit = 64 } = defineProps<{
   filters: FuseFilterWithValue<ComfyNodeDefImpl, string>[]
   searchLimit?: number
 }>()
+const acpref = ref()
 
 const nodeSearchFilterVisible = ref(false)
 const inputId = `comfy-vue-node-search-box-input-${Math.random()}`
@@ -140,7 +142,11 @@ const reFocusInput = async () => {
   }
 }
 
-onMounted(reFocusInput)
+onMounted(() => {
+  acpref.value.hide = () => {}
+  search('')
+  acpref.value.show()
+})
 const onAddFilter = (
   filterAndValue: FuseFilterWithValue<ComfyNodeDefImpl, string>
 ) => {
--- a/src/components/searchbox/NodeSearchBox.vue
+++ b/src/components/searchbox/NodeSearchBox.vue
@@ -35,7 +35,7 @@
     </Dialog>
 
     <AutoCompletePlus
-      ref="acpref"
+      ref="autoCompletePlus"
       :model-value="filters"
       class="comfy-vue-node-search-box z-10 grow"
       scroll-height="40vh"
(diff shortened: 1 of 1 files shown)


src/components/searchbox/NodeSearchBox.vue
```diff
@@ -140,7 +142,11 @@ const reFocusInput = async () => {
   }
 }
 
-onMounted(reFocusInput)
+onMounted(() => {
+  acpref.value.hide = () => {}
```
DrJKL: Why not set this in the template?

AustinMroz:
Current test failures seem unrelated. Is "unhovered widget inputs are incorrectly displayed in vue mode" a problem other PRs are seeing?

I'm not having luck reproducing locally.

src/components/searchbox/NodeSearchBox.vue
```diff
@@ -140,7 +141,13 @@ const reFocusInput = async () => {
   }
 }
 
-onMounted(reFocusInput)
+onMounted(() => {
+  inputElement ??= document.getElementById(inputId) as HTMLInputElement
+  if (inputElement) inputElement.focus()
+  autoCompletePlus.value.hide = () => search('')
+  search('')
```
christian-byrne, line 148: How did you discover this solution? I also tried to fix this in the past but wasn't able.

AustinMroz: A fairly lengthy journey of frustration. I started with an `@clear="search('')` and was excited to see that it would cause the computed list of options to update. I came real close to opening that as a one line fix only to see that, of course, the problem wasn't that easy. 

From there, I started doing debugging on the actual surrounding code. The relevant fragments are included in the top post, but one of `hide` or `onInput` needed to be changed and hide was simpler.

Swapping `hide` to call `search`('') instead of a no-op was a later commit to solve edge cases with options showing, but not being added to the graph on click.

christian-byrne: Dang, that's good.

christian-byrne (approved):

PR #6251, opened 2025-10-24
DO NOT MERGE - History v2 conflicts

arjansingh:
## Summary

Using this as a working reference for integrating this feature into `main`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6251-DO-NOT-MERGE-History-v2-conflicts-2966d73d365081c996d2f54bd73a0aea) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/fixtures/utils/taskHistory.ts
+++ b/browser_tests/fixtures/utils/taskHistory.ts
@@ -34,17 +34,23 @@ const getContentType = (filename: string, fileType: OutputFileType) => {
 }
 
 const setQueueIndex = (task: TaskItem) => {
-  task.prompt[0] = TaskHistory.queueIndex++
+  task.prompt.priority = TaskHistory.queueIndex++
 }
 
 const setPromptId = (task: TaskItem) => {
-  task.prompt[1] = uuidv4()
+  if (!task.prompt.prompt_id || task.prompt.prompt_id === 'prompt-id') {
+    task.prompt.prompt_id = uuidv4()
+  }
 }
 
 export default class TaskHistory {
   static queueIndex = 0
   static readonly defaultTask: Readonly<HistoryTaskItem> = {
-    prompt: [0, 'prompt-id', {}, { client_id: uuidv4() }, []],
+    prompt: {
+      priority: 0,
+      prompt_id: 'prompt-id',
+      extra_data: { client_id: uuidv4() }
+    },
     outputs: {},
     status: {
       status_str: 'success',
@@ -66,10 +72,37 @@ export default class TaskHistory {
   )
 
   private async handleGetHistory(route: Route) {
+    const url = route.request().url()
+
+    // Handle history_v2/:prompt_id endpoint
+    const promptIdMatch = url.match(/history_v2\/([^?]+)/)
+    if (promptIdMatch) {
+      const promptId = promptIdMatch[1]
+      const task = this.tasks.find((t) => t.prompt.prompt_id === promptId)
+      const response: Record<string, any> = {}
+      if (task) {
+        response[promptId] = task
+      }
+      return route.fulfill({
+        status: 200,
+        contentType: 'application/json',
+        body: JSON.stringify(response)
+      })
+    }
+
+    // Handle history_v2 list endpoint
+    // Convert HistoryTaskItem to RawHistoryItem format expected by API
+    const rawHistoryItems = this.tasks.map((task) => ({
+      prompt_id: task.prompt.prompt_id,
+      prompt: task.prompt,
+      status: task.status,
+      outputs: task.outputs,
+      ...(task.meta && { meta: task.meta })
+    }))
     return route.fulfill({
       status: 200,
       contentType: 'application/json',
-      body: JSON.stringify(this.tasks)
+      body: JSON.stringify({ history: rawHistoryItems })
     })
   }
 
@@ -93,7 +126,7 @@ export default class TaskHistory {
 
   async setupRoutes() {
     return this.comfyPage.page.route(
-      /.*\/api\/(view|history)(\?.*)?$/,
+      /.*\/api\/(view|history_v2)(\/[^?]*)?(\?.*)?$/,
       async (route) => {
         const request = route.request()
         const method = request.method()
--- a/browser_tests/tests/historyApi.spec.ts
+++ b/browser_tests/tests/historyApi.spec.ts
@@ -1,0 +1,131 @@
+import {
+  comfyExpect as expect,
+  comfyPageFixture as test
+} from '../fixtures/ComfyPage'
+
+test.describe('History API v2', () => {
+  const TEST_PROMPT_ID = 'test-prompt-id'
+  const TEST_CLIENT_ID = 'test-client'
+
+  test('Can fetch history with new v2 format', async ({ comfyPage }) => {
+    // Set up mocked history with tasks
+    await comfyPage.setupHistory().withTask(['example.webp']).setupRoutes()
+
+    // Verify history_v2 API response format
+    const result = await comfyPage.page.evaluate(async () => {
+      try {
+        const response = await window['app'].api.getHistory()
+        return { success: true, data: response }
+      } catch (error) {
+        console.error('Failed to fetch history:', error)
+        return { success: false, error: error.message }
+      }
+    })
+
+    expect(result.success).toBe(true)
+    expect(result.data).toHaveProperty('History')
+    expect(Array.isArray(result.data.History)).toBe(true)
+    expect(result.data.History.length).toBeGreaterThan(0)
+
+    const historyItem = result.data.History[0]
+
+    // Verify the new prompt structure (object instead of array)
+    expect(historyItem.prompt).toHaveProperty('priority')
+    expect(historyItem.prompt).toHaveProperty('prompt_id')
+    expect(historyItem.prompt).toHaveProperty('extra_data')
+    expect(typeof historyItem.prompt.priority).toBe('number')
+    expect(typeof historyItem.prompt.prompt_id).toBe('string')
+    expect(historyItem.prompt.extra_data).toHaveProperty('client_id')
+  })
+
+  test('Can load workflow from history using history_v2 endpoint', async ({
+    comfyPage
+  }) => {
+    // Simple mock workflow for testing
+    const mockWorkflow = {
+      version: 0.4,
+      nodes: [{ id: 1, type: 'TestNode', pos: [100, 100], size: [200, 100] }],
+      links: [],
+      groups: [],
+      config: {},
+      extra: {}
+    }
+
+    // Set up history with workflow data
+    await comfyPage
+      .setupHistory()
+      .withTask(['example.webp'], 'images', {
+        prompt: {
+          priority: 0,
+          prompt_id: TEST_PROMPT_ID,
+          extra_data: {
+            client_id: TEST_CLIENT_ID,
+            extra_pnginfo: { workflow: mockWorkflow }
+          }
+        }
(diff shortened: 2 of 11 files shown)


browser_tests/fixtures/utils/taskHistory.ts
```diff
@@ -34,17 +34,23 @@ const getContentType = (filename: string, fileType: OutputFileType) => {
 }
```
arjansingh, line 34: Do not bring in.

browser_tests/tests/sidebar/workflows.spec.ts
```diff
@@ -187,6 +187,7 @@ test.describe('Workflows sidebar', () => {
 
   test('Can save workflow as with same name', async ({ comfyPage }) => {
     await comfyPage.menu.topbar.saveWorkflow('workflow5.json')
+    await comfyPage.nextFrame()
```
arjansingh, line 190: Do not bring in.

browser_tests/tests/historyApi.spec.ts
```diff
@@ -0,0 +1,131 @@
+import {
```
arjansingh, line 1: Do not bring in

src/scripts/ui.ts
```diff
@@ -264,15 +264,16 @@ class ComfyList {
                   ? item.remove
                   : {
                       name: 'Delete',
```
arjansingh, line 266: Do not bring in?

christian-byrne: No need - legacy code

src/scripts/api.ts
```diff
@@ -869,15 +900,23 @@ export class ComfyApi extends EventTarget {
...
         // Running action uses a different endpoint for cancelling
-        Running: data.queue_running.map((prompt: Record<number, any>) => ({
+        Running: data.queue_running.map((prompt: any) => ({
           taskType: 'Running',
+<<<<<<< HEAD
           prompt,
           // prompt[1] is the prompt id
           remove: { name: 'Cancel', cb: () => api.interrupt(prompt[1]) }
+=======
+          prompt: this.parseQueuePrompt(prompt),
```
arjansingh, line 910: parseQueuePrompt is already a pseudo mapping function?

christian-byrne: What do you mean?

src/schemas/apiSchema.ts
```diff
@@ -216,13 +208,11 @@ const zStatus = z.object({
...
 })
 
-const zTaskPrompt = z.tuple([
-  zQueueIndex,
-  zPromptId,
-  zPromptInputs,
-  zExtraData,
-  zOutputsToExecute
-])
+const zTaskPrompt = z.object({
```
arjansingh, line 211: Map new `zTaskPrompt` to old `zTaskPrompt`.

src/scripts/api.ts
```diff
@@ -856,6 +865,28 @@ export class ComfyApi extends EventTarget {
     return this.getHistory()
   }
 
+  /**
+   * Parses queue prompt data from array or object format
+   * @param rawPrompt The raw prompt data from the API
+   * @returns Normalized TaskPrompt object
+   */
+  private parseQueuePrompt(rawPrompt: any): TaskPrompt {
+    if (Array.isArray(rawPrompt)) {
```
arjansingh, line 874: translation: if OSS prompt

would this ever happen in Cloud?

src/scripts/api.ts
```diff
@@ -856,6 +865,28 @@ export class ComfyApi extends EventTarget {
...
+      // Queue format: [priority, prompt_id, workflow, outputs]
```
arjansingh, line 875: Comment has wrong format? Investigate. See: https://github.com/Comfy-Org/ComfyUI_frontend/pull/6251/files#diff-b769532e74f826ca909951c0c34331b9246efb3f6901ff95a856ecf01ad826beL219-L225

src/scripts/api.ts
```diff
@@ -869,15 +900,23 @@ export class ComfyApi extends EventTarget {
       const data = await res.json()
       return {
         // Running action uses a different endpoint for cancelling
-        Running: data.queue_running.map((prompt: Record<number, any>) => ({
+        Running: data.queue_running.map((prompt: any) => ({
```
arjansingh, line 903: figure out actual type

src/scripts/ui.ts
```diff
@@ -264,15 +264,16 @@ class ComfyList {
...
-                      cb: () => api.deleteItem(this.#type, item.prompt[1])
+                      cb: () =>
```
arjansingh, line 267: legacy UI with menu turned off.

src/scripts/api.ts
```diff
@@ -856,6 +865,28 @@ export class ComfyApi extends EventTarget {
...
+  /**
+   * Parses queue prompt data from array or object format
+   * @param rawPrompt The raw prompt data from the API
+   * @returns Normalized TaskPrompt object
+   */
+  private parseQueuePrompt(rawPrompt: any): TaskPrompt {
+    if (Array.isArray(rawPrompt)) {
+      // Queue format: [priority, prompt_id, workflow, outputs]
+      const [priority, prompt_id, workflow] = rawPrompt
+      return {
```
arjansingh, line 877: the prompt is a lossy object.

src/scripts/api.ts
arjansingh: normalize back to the array and see if we can get that working.

then we can look at semantic refactoring.

arjansingh: > What do you mean?

@DrJKL and I are talking through this and I am just taking live notes.

We think (and are pretty sure now) that this is doing a mapping where it converts array to object and drops the inputs and outputs for compatability with the cloud api. 

We wouldn't necessarily being this function in verbatim, but we might use it as a basis for building any adapter/translation layer.

src/components/sidebar/tabs/QueueSidebarTab.vue
```diff
@@ -205,8 +206,16 @@ const menuItems = computed<MenuItem[]>(() => {
     {
       label: t('g.loadWorkflow'),
       icon: 'pi pi-file-export',
-      command: () => menuTargetTask.value?.loadWorkflow(app),
-      disabled: !menuTargetTask.value?.workflow
+      command: () => {
```
arjansingh, line 209: right click on history to load workflow

src/platform/workflow/core/services/workflowService.ts
```diff
@@ -144,6 +158,32 @@ export const useWorkflowService = () => {
     await app.loadGraphData(blankGraph)
   }
 
+  /**
+   * Load a workflow from a task item (queue/history)
+   * For history items, fetches workflow data from /history_v2/{prompt_id}
+   * @param task The task item to load the workflow from
+   */
+  const loadTaskWorkflow = async (task: TaskItemImpl) => {
```
arjansingh, line 166: used here: https://github.com/Comfy-Org/ComfyUI_frontend/pull/6251/files#diff-86e2bace7f5b0fa4713944d03b32dc6472f31dc10d095b813515acf5e5b4ba5bR209

src/scripts/api.ts
```diff
@@ -894,20 +933,51 @@ export class ComfyApi extends EventTarget {
...
       return { History: [] }
     }
   }
 
+  /**
+   * Gets workflow data for a specific prompt from history
+   * @param prompt_id The prompt ID to fetch workflow for
+   * @returns Workflow data for the specific prompt
+   */
+  async getWorkflowFromHistory(
```
arjansingh, line 959: right click menu uses this

src/stores/queueStore.ts
```diff
@@ -477,19 +483,38 @@ export const useQueueStore = defineStore('queue', () => {
...
       )
       const existingHistoryItems = historyTasks.value.filter((item) =>
         allIndex.has(item.queueIndex)
       )
-      historyTasks.value = [...newHistoryItems, ...existingHistoryItems]
+      const sortedTasks = [...newHistoryItems, ...existingHistoryItems]
         .slice(0, maxHistoryItems.value)
-        .sort((a, b) => b.queueIndex - a.queueIndex)
+        .sort((a, b) => {
+          const aTime = a.executionStartTimestamp ?? 0
```
arjansingh, line 505: we should be able to just sort by execution start time because both OSS and Cloud support it.

christian-byrne: We don't want to sort by execution time because it's not reliable. We can do it on cloud by letting the adapter map execution time to queue index.

src/scripts/api.ts
arjansingh, line 874: maybe basic shallow validation? 

also think about an integration test to test server api?

src/scripts/api.ts
christian-byrne: Shouldn't be necessary if we use the adapter to map back to original format, I believe -- the concepts associated with differeing backends and data shapes should all be abstracted and contained in the static adatper.

src/platform/workflow/core/services/workflowService.ts
christian-byrne: Same as https://github.com/Comfy-Org/ComfyUI_frontend/pull/6251#discussion_r2461882618

src/scripts/ui.ts
christian-byrne: can skip

src/scripts/api.ts
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/6251#discussion_r2461882618

src/scripts/api.ts
christian-byrne: https://github.com/Comfy-Org/ComfyUI_frontend/pull/6251#discussion_r2461882618

src/scripts/api.ts
christian-byrne: this is the type of adapting we would want

src/stores/queueStore.ts
arjansingh: Understood, I will head down that route.

src/scripts/api.ts
arjansingh: Understood... this comment was just context for what code was doing.

src/scripts/api.ts
christian-byrne: shallow validation is not in scope for this, we will known the shape based on distribution.

src/scripts/api.ts
arjansingh: TL;DR: `history_v2` doesn't return `extra_data`. This method/endpoint gets it upon user requests, and is probably worth keeping. Otherwise we have to fetch workflows for every history item to achieve a full v1 mapping.

@christian-byrne so something to think about here is that `history_v2` doesn't return `extra_data` by default.

Compare `history_v2`:

Image

With `history`:

Image


I think having this function to just grab the workflow data when the user needs, is probably the most efficient option in cloud.

Otherwise if we want to do that with a map at [line 936](https://github.com/Comfy-Org/ComfyUI_frontend/pull/6251/files#diff-bee7eddf5267c50e3abc9d0b96326137993f615be5d05bf62478430e1659c8afR936), we'd have to call `/history_v2/${prompt_id}` for each (new) history item so we can have `extra_data` present.

It's probably just better to grab the workflow data when user actually needs it.

The rest of the mapping works fine (absent the service worker issue in development). I'm getting a PR up for that.

christian-byrne:
Can close now?

arjansingh:
> Can close now?

Sorry didnt' see this. yes i am closing.

PR #6264, opened 2025-10-25
Style: Remove currently non-functional collapsed slots

DrJKL:
## Summary

Address request to remove confusing hover behavior in the collapsed state

## Review Focus

Confirmed with Alex that it's okay to remove these now and add them back with the implementation instead of leaving them as decorative elements.

Also fixes a bug where the slots _all_ revealed when a node was hovered.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6264-Style-Remove-currently-non-functional-collapsed-slots-2976d73d36508191914def8889491e8e) by [Unito](https://www.unito.io)


christian-byrne (dismissed):

christian-byrne (approved):
LGTM!

> Also fixes a bug where the slots all revealed when a node was hovered.


I guess expectations updates are from this part

PR #6294, opened 2025-10-26
remove all auth service work related code

christian-byrne:
## Summary

Removes all service worker auth code, as it is being replaced by a more robust standard solution for authenticating view and viewvideo requests in https://github.com/Comfy-Org/ComfyUI_frontend/pull/6295.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6294-remove-all-auth-service-work-related-code-2986d73d36508170a24bf1c42cad401e) by [Unito](https://www.unito.io)


christian-byrne:
and good riddance!

PR #6301, opened 2025-10-26
set Sentry config based on distribution

christian-byrne:
Set the config based on compile-time DISTRIBUTION env var.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6301-set-Sentry-config-based-on-distribution-2986d73d3650815f9b7ef821bbdd745f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -49,11 +49,18 @@ Sentry.init({
   dsn: __SENTRY_DSN__,
   enabled: __SENTRY_ENABLED__,
   release: __COMFYUI_FRONTEND_VERSION__,
-  integrations: [],
-  autoSessionTracking: false,
-  defaultIntegrations: false,
   normalizeDepth: 8,
-  tracesSampleRate: 0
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -49,11 +49,18 @@ Sentry.init({
...
   enabled: __SENTRY_ENABLED__,
   release: __COMFYUI_FRONTEND_VERSION__,
-  integrations: [],
-  autoSessionTracking: false,
-  defaultIntegrations: false,
   normalizeDepth: 8,
-  tracesSampleRate: 0
+  tracesSampleRate: isCloud ? 1.0 : 0,
+  replaysSessionSampleRate: 0,
+  replaysOnErrorSampleRate: 0,
```
christian-byrne, line 55: Set these to fix https://comfy-org.sentry.io/issues/6929291501/?project=4509681221369857&query=is%3Aunresolved&referrer=issue-stream

christian-byrne:
Oh this doesn't need to be backported.

PR #6307, opened 2025-10-26
feat(api-nodes): add pricing for new LTXV-2 models

bigcat88:
## Summary

For the upcoming LTXV API nodes.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6307-feat-api-nodes-add-pricing-for-new-LTXV-2-models-2986d73d365081db9994deffc219c6c4) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/node/useNodePricing.ts
+++ b/src/composables/node/useNodePricing.ts
@@ -169,6 +169,46 @@ const byteDanceVideoPricingCalculator = (node: LGraphNode): string => {
     : `$${minCost.toFixed(2)}-$${maxCost.toFixed(2)}/Run`
 }
 
+const ltxvPricingCalculator = (node: LGraphNode): string => {
+  const modelWidget = node.widgets?.find(
+    (w) => w.name === 'model'
+  ) as IComboWidget
+  const durationWidget = node.widgets?.find(
+    (w) => w.name === 'duration'
+  ) as IComboWidget
+  const resolutionWidget = node.widgets?.find(
+    (w) => w.name === 'resolution'
+  ) as IComboWidget
+
+  const fallback = '$0.04-0.24/second'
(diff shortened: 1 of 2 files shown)


src/composables/node/useNodePricing.ts
```diff
@@ -169,6 +169,46 @@ const byteDanceVideoPricingCalculator = (node: LGraphNode): string => {
...
+      '1920x1080': 0.06,
+      '2560x1440': 0.12,
+      '3840x2160': 0.24
+    },
+    'ltx-2 (fast)': {
+      '1920x1080': 0.04,
+      '2560x1440': 0.08,
+      '3840x2160': 0.16
+    }
+  }
```
DrJKL, line 200: Not sure if it's worthwhile yet, but using `as const` here and checking the keys could add some safety, make it less likely we ever hit the fallback.
```suggestion
  const priceByModel = {
    'ltx-2 (pro)': {
      '1920x1080': 0.06,
      '2560x1440': 0.12,
      '3840x2160': 0.24
    },
    'ltx-2 (fast)': {
      '1920x1080': 0.04,
      '2560x1440': 0.08,
      '3840x2160': 0.16
    }
  } as const
```

DrJKL (approved):

src/composables/node/useNodePricing.ts
bigcat88: I hope we can soon formalize API node pricing into a declarative schema and move the calculation logic to the Python backend.

Any idea why backport failed?

bigcat88:
@DrJKL as I understand it, we need to add a tag with the version for which we want backport, and then rerun the action, right?
Which version will be released on Monday?

christian-byrne:
Ignore the comments

PR #6330, opened 2025-10-27
Fix: Vue-Litegraph conversion bug with Reroutes

DrJKL:
## Summary

The private fields triggered an error in intitializing the linkLayoutSync. Turns out that wasn't necessary anymore.

> [!NOTE]
> Edit: Doing some more investigation, it looks like the slot sync can also be removed?

## Changes

- **What**: Converts JS private fields to typescript private, adds some readonly declarations
- **What**: Removes the useLinkLayoutSync usage in useVueNodeLifecycle
- **What**: Removes the useSlotLayoutSync usage in useVueNodeLifecycle

## Review Focus

Was the sync doing something that wouldn't be caught in normal usage/testing?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6330-Fix-Vue-Litegraph-conversion-bug-with-Reroutes-2996d73d3650819ebb24e4aa2fc51c65) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LGraph.ts
```diff
@@ -223,15 +223,15 @@ export class LGraph
   /** Internal only.  Not required for serialisation; calculated on deserialise. */
   #lastFloatingLinkId: number = 0
 
-  #floatingLinks: Map<LinkId, LLink> = new Map()
+  private readonly floatingLinksInternal: Map<LinkId, LLink> = new Map()
```
arjansingh, line 226: Okay if I understand correctly (mdn)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy#no_private_field_forwarding], this works by making these available to proxy, whereas JS privates can't be accessed by proxies.

arjansingh (approved):

Myestery (approved):

PR #6341, opened 2025-10-28
add TOS and privacy policy text on cloud login view

christian-byrne:
## Summary

Adds these missing sentences that might have been dropped when bringing this component over from main or during some refactoring.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6341-add-TOS-and-privacy-policy-text-on-cloud-login-view-29a6d73d36508124b9dad689184a4ae3) by [Unito](https://www.unito.io)


christian-byrne:
I guess they were just moved to bottom of page actually

PR #6357, opened 2025-10-29
Feat/vue noes arrange alg improved

simula-r:
## Summary

Improve the previous [vue node arrange alg](https://github.com/Comfy-Org/ComfyUI_frontend/blob/5a1284660c41f2705eb8c633054a7babbac92442/src/renderer/extensions/vueNodes/layout/scaleLayoutForVueNodes.ts) to match Litegraph much closer visually.

- In addition to the scaling of the LG node's x,y and then setting the vue nodes position to that, we can also scale the width and height of the LG node and set that for the Vue node wxh. 
- Change from scaling from the center to the top left
- Change the zoom offset to account for the scale for seamless transition

image

image (1)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6357-Feat-vue-noes-arrange-alg-improved-29b6d73d365081e7813bd6f19ce1202a) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
```diff
@@ -19,68 +19,76 @@ export function scaleLayoutForVueNodes() {
...
     group.size = [group.size[0] * SCALE_FACTOR, group.size[1] * SCALE_FACTOR]
   })
 
-  if (scaledLgBounds) {
-    canvas.ds.fitToBounds(scaledLgBounds, {
-      zoom: 0.5 //Makes it so the fit to view is slightly zoomed out and not edge to edge.
-    })
-  }
+  const originScreen = canvas.ds.convertOffsetToCanvas([originX, originY])
+  canvas.ds.changeScale(canvas.ds.scale / SCALE_FACTOR, originScreen)
```
DrJKL, line 90: Could we use the fit nodes in view command here?

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
```diff
@@ -19,68 +19,76 @@ export function scaleLayoutForVueNodes() {
...
+
+  if (!graph.extra) graph.extra = {}
```
DrJKL, line 92: nit, optional:
```suggestion
  graph.extra ??= {}
```

DrJKL (approved):
I love it, thank you <3

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
simula-r: that was in previously. Think it feels better during switching per workflow to not have it pull the view around, this makes it seamless.

Will add a fit to view later with banner.

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
simula-r: can do in follow up

PR #6366, opened 2025-10-29
remove email verification temporarily

christian-byrne:
## Summary

- Temporarily remove email verification. After susbcription gating was enabled, this is less important
- Will re-add the logic back at a later time, defering requirement until time to subscribe
- For time being, typo emails can be resolved through custom service (https://support.comfy.org/hc/en-us/requests/new?tf_42243568391700=ccloud&tf_123456=X)
- Keep the route and redirect for deprecation. Not really needed since the server falls back to root route anyway but generally good practice and is more resistant to future changes + avoids a single extra routing step in that scenario.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6366-remove-email-verification-temporarily-29b6d73d3650810095a4e7c4591b3327) by [Unito](https://www.unito.io)


src/platform/assets/components/AssetCard.vue
```diff
@@ -31,6 +31,7 @@
               'dark-theme:text-white'
             )
           "
+          :title="asset.name"
```
viva-jinyi, line 34: I think it should be deleted ?

viva-jinyi (approved):
LGTM

src/platform/assets/components/AssetCard.vue
christian-byrne: Just wrong base branch but this change was requested

PR #6254, opened 2025-10-24
feat: Add pagination support for history assets in Assets sidebar

viva-jinyi:
## Overview

Added pagination support with infinite scroll to the Assets sidebar for efficient loading of large history collections.

### 1. API Enhancement
- Added `offset` parameter to `api.getHistory()` method
- Safely handle query parameters using URLSearchParams

### 2. AssetsStore Refactoring
- Implemented pagination logic following ManagerContent pattern
- Load history items in pages (50 items per page)
- Stable data merging with duplicate removal
- Preserve server-provided order (removed frontend sorting)

### 3. UI/UX Improvements
- Infinite scroll using VirtualGrid's `approach-end` event
- Loading spinner during additional data fetch
- Pagination only active in Output tab

## Technical Details

- **Page size**: 50 items
- **Loading trigger**: When VirtualGrid scroll approaches end
- **State management**: Manual state management with refs (instead of useAsyncState pattern)

## Testing

- [x] Initial 50 items load correctly
- [x] Additional items load on scroll
- [x] Loading indicator displays properly
- [x] Input tab unaffected
- [x] No duplicate items

## Notes

- Server API provides data in oldest-first order, sorting options needed in future
- Currently displays in server-provided order

[screen-capture (2).webm](https://github.com/user-attachments/assets/9df41548-2256-4590-bd88-b30fcc96208d)

PR #6373, opened 2025-10-29
feat: Add pagination support for media assets history

viva-jinyi:
## Summary
- Implement pagination for media assets history to handle large datasets efficiently
- Add infinite scroll support with approach-end event handler  
- Support offset parameter in history API for both V1 and V2 endpoints

## Changes
- Add offset parameter support to `api.getHistory()` method
- Update history fetchers (V1/V2) to include offset in API requests
- Implement `loadMoreHistory()` in assetsStore with pagination state management
- Add `loadMore`, `hasMore`, and `isLoadingMore` to IAssetsProvider interface
- Add approach-end handler in AssetsSidebarTab for infinite scroll
- Set BATCH_SIZE to 200 for efficient loading

## Implementation Improvements
Simplified offset-based pagination by removing unnecessary reconciliation logic:
- Remove `reconcileHistory`, `taskItemsMap`, `lastKnownQueueIndex` (offset is sufficient)
- Replace `assetItemsByPromptId` Map → `loadedIds` Set (store IDs only)
- Replace `findInsertionIndex` binary search → push + sort (faster for batch operations)
- Replace `loadingPromise` → `isLoadingMore` boolean (simpler state management)
- Fix memory leak by cleaning up Set together with array slice

## Test Plan
- [x] TypeScript compilation passes
- [x] ESLint and Prettier formatting applied
- [x] Test infinite scroll in media assets tab
- [x] Verify network requests include correct offset parameter
- [x] Confirm no duplicate items when loading more

### DIFF
--- a/src/components/sidebar/tabs/AssetsSidebarTab.vue
+++ b/src/components/sidebar/tabs/AssetsSidebarTab.vue
@@ -41,9 +41,27 @@
       </TabList>
     </template>
     <template #body>
-      <div v-if="displayAssets.length" class="relative size-full">
+      <!-- Loading state -->
+      <div v-if="loading">
+        <ProgressSpinner class="absolute left-1/2 w-[50px] -translate-x-1/2" />
+      </div>
+      <!-- Empty state -->
+      <div v-else-if="!displayAssets.length">
+        <NoResultsPlaceholder
+          icon="pi pi-info-circle"
+          :title="
+            $t(
+              activeTab === 'input'
+                ? 'sideToolbar.noImportedFiles'
+                : 'sideToolbar.noGeneratedFiles'
+            )
+          "
+          :message="$t('sideToolbar.noFilesFoundMessage')"
+        />
+      </div>
+      <!-- Content -->
+      <div v-else class="relative size-full">
         <VirtualGrid
-          v-if="!loading && displayAssets.length"
           :items="mediaAssetsWithKey"
           :grid-style="{
             display: 'grid',
@@ -67,24 +85,6 @@
             />
           </template>
         </VirtualGrid>
-        <div v-else>
-          <ProgressSpinner
-            class="absolute left-1/2 w-[50px] -translate-x-1/2"
-          />
-        </div>
-      </div>
-      <div v-else-if="!loading">
-        <NoResultsPlaceholder
-          icon="pi pi-info-circle"
-          :title="
-            $t(
-              activeTab === 'input'
-                ? 'sideToolbar.noImportedFiles'
-                : 'sideToolbar.noGeneratedFiles'
-            )
-          "
-          :message="$t('sideToolbar.noFilesFoundMessage')"
-        />
       </div>
     </template>
     <template #footer>
--- a/src/components/sidebar/tabs/AssetsSidebarTab.vue
+++ b/src/components/sidebar/tabs/AssetsSidebarTab.vue
@@ -43,7 +43,7 @@
     <template #body>
       <div v-if="displayAssets.length" class="relative size-full">
         <VirtualGrid
-          v-if="displayAssets.length"
+          v-if="!loading"
           :items="mediaAssetsWithKey"
           :grid-style="{
             display: 'grid',
@@ -51,6 +51,7 @@
             padding: '0.5rem',
             gap: '0.5rem'
           }"
+          @approach-end="handleApproachEnd"
         >
           <template #item="{ item }">
             <MediaAssetCard
@@ -66,24 +67,24 @@
             />
           </template>
         </VirtualGrid>
-        <div v-else-if="loading">
+        <div v-else>
           <ProgressSpinner
             class="absolute left-1/2 w-[50px] -translate-x-1/2"
           />
         </div>
-        <div v-else>
-          <NoResultsPlaceholder
-            icon="pi pi-info-circle"
-            :title="
-              $t(
-                activeTab === 'input'
-                  ? 'sideToolbar.noImportedFiles'
-                  : 'sideToolbar.noGeneratedFiles'
-              )
-            "
-            :message="$t('sideToolbar.noFilesFoundMessage')"
-          />
-        </div>
+      </div>
+      <div v-else>
+        <NoResultsPlaceholder
+          icon="pi pi-info-circle"
+          :title="
+            $t(
+              activeTab === 'input'
+                ? 'sideToolbar.noImportedFiles'
+                : 'sideToolbar.noGeneratedFiles'
+            )
+          "
+          :message="$t('sideToolbar.noFilesFoundMessage')"
+        />
       </div>
     </template>
     <template #footer>
@@ -291,6 +292,7 @@ watch(
   activeTab,
   () => {
     clearSelection()
+    // Reset pagination state when tab changes
     void refreshAssets()
   },
   { immediate: true }
@@ -394,5 +396,17 @@ const handleDeleteSelected = async () => {
   const selectedAssets = getSelectedAssets(displayAssets.value)
   await deleteMultipleAssets(selectedAssets)
   clearSelection()
+}
+
+const handleApproachEnd = async () => {
+  if (
+    activeTab.value === 'output' &&
+    !isInFolderView.value &&
+    outputAssets.loadMore &&
+    outputAssets.hasMore?.value &&
+    !outputAssets.isLoadingMore?.value
+  ) {
+    await outputAssets.loadMore()
+  }
 }
 </script>
--- a/src/components/sidebar/tabs/AssetsSidebarTab.vue
+++ b/src/components/sidebar/tabs/AssetsSidebarTab.vue
@@ -148,6 +148,7 @@
 </template>
 
 <script setup lang="ts">
+import { useDebounceFn } from '@vueuse/core'
 import ProgressSpinner from 'primevue/progressspinner'
 import { useToast } from 'primevue/usetoast'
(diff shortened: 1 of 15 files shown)


viva-jinyi:
#6254

src/platform/assets/composables/media/IAssetsProvider.ts
```diff
@@ -26,4 +26,19 @@ export interface IAssetsProvider {
    * Refresh the media list (alias for fetchMediaList)
    */
   refresh: () => Promise<AssetItem[]>
+
+  /**
+   * Load more items (for pagination)
+   */
+  loadMore?: () => Promise<void>
```
christian-byrne: question: can this be required?

viva-jinyi: For input assets, pagination is not used, so making `loadMore` optional is the correct approach. Input files are typically fewer in number and loaded all at once, unlike history assets which can accumulate indefinitely and require infinite scroll pagination.

src/stores/assetsStore.ts
```diff
@@ -85,16 +89,22 @@ function mapHistoryToAssets(historyItems: HistoryTaskItem[]): AssetItem[] {
...
-  const getFetchInputFiles = () => {
-    if (isCloud) {
-      return fetchInputFilesFromCloud
-    }
-    return fetchInputFilesFromAPI
-  }
-  const fetchInputFiles = getFetchInputFiles()
+  const allHistoryItems = ref<AssetItem[]>([])
+
+  const loadedIds = new Set<string>()
```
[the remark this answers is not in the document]
viva-jinyi: Changed `loadedIds` from a module-level Set to `shallowReactive(new Set<string>())` inside the store definition.

src/stores/assetsStore.ts
```diff
@@ -109,23 +119,113 @@ export const useAssetsStore = defineStore('assets', () => {
...
+          return false // Already loaded
+        }
+        loadedIds.add(asset.id)
+        return true
+      })
+
+      allHistoryItems.value.push(...uniqueAssets)
+
+      // Sort by date (newest first)
+      allHistoryItems.value.sort(
```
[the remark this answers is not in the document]
viva-jinyi: Replaced the full array sort on every batch with insertion at the correct sorted position

src/stores/assetsStore.ts
```diff
@@ -109,23 +119,113 @@ export const useAssetsStore = defineStore('assets', () => {
...
+    } finally {
+      historyLoading.value = false
     }
-  })
+  }
+
+  /**
+   * Load more history items (infinite scroll)
+   */
+  const loadMoreHistory = async () => {
```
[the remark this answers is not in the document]
viva-jinyi: Combined the two separate guards into a single atomic check:
```typescript
// Before: potential race condition between checks
if (!hasMoreHistory.value) return
if (isLoadingMore.value) return
isLoadingMore.value = true

// After: atomic check eliminates race window
if (!hasMoreHistory.value || isLoadingMore.value) return
isLoadingMore.value = true
```

src/stores/assetsStore.ts
```diff
@@ -109,23 +119,113 @@ export const useAssetsStore = defineStore('assets', () => {
...
+    // Guard: prevent concurrent loads
+    if (isLoadingMore.value) return
+
+    isLoadingMore.value = true
+    historyError.value = null
+
+    try {
+      await fetchHistoryAssets(true)
+      historyAssets.value = allHistoryItems.value
+    } catch (err) {
```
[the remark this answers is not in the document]
viva-jinyi: Added data preservation logic to `loadMoreHistory` to match `updateHistory`:
```typescript
} catch (err) {
  console.error('Error loading more history:', err)
  historyError.value = err
  // Keep existing data when error occurs (consistent with updateHistory)
  if (!historyAssets.value.length) {
    historyAssets.value = []
  }
}
```

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -395,4 +397,16 @@ const handleDeleteSelected = async () => {
   await deleteMultipleAssets(selectedAssets)
   clearSelection()
 }
+
+const handleApproachEnd = async () => {
```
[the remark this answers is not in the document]
viva-jinyi: Wrapped `handleApproachEnd` with `useDebounceFn`:
```typescript
import { useDebounceFn } from '@vueuse/core'

const handleApproachEnd = useDebounceFn(async () => {
  if (
    activeTab.value === 'output' &&
    !isInFolderView.value &&
    outputAssets.hasMore.value &&
    !outputAssets.isLoadingMore.value
  ) {
    await outputAssets.loadMore()
  }
}, 300)
```

src/platform/assets/composables/media/IAssetsProvider.ts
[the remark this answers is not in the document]
viva-jinyi: Changed optional properties to required

src/platform/remote/comfyui/history/fetchers/fetchHistoryV1.ts
```diff
@@ -19,9 +19,14 @@ import type {
  */
 export async function fetchHistoryV1(
   fetchApi: (url: string) => Promise<Response>,
-  maxItems: number = 200
+  maxItems: number = 200,
+  offset?: number
 ): Promise<HistoryV1Response> {
-  const res = await fetchApi(`/history?max_items=${maxItems}`)
+  let url = `/history?max_items=${maxItems}`
```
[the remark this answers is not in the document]
viva-jinyi: Implemented input validation before URL construction in both V1 and V2 fetchers:
```typescript
// Validate offset parameter
if (offset !== undefined && (offset < 0 || !Number.isInteger(offset))) {
  throw new Error(
    `Invalid offset parameter: ${offset}. Must be a non-negative integer.`
  )
}
```

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -43,14 +43,15 @@
     <template #body>
       <div v-if="displayAssets.length" class="relative size-full">
         <VirtualGrid
-          v-if="displayAssets.length"
+          v-if="!loading && displayAssets.length"
           :items="mediaAssetsWithKey"
```
DrJKL, line 65: nit: The different conditionals in here are a little hard to track.
Normally I'd expect a top level loader (or error state), then a fallthrough to the contentful item.

Also, I think the second `displayAssets.length` is redundant since if we get to the VirtualGrid, the outer div already guarantees that we have some items.

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -66,24 +67,24 @@
...
-                activeTab === 'input'
-                  ? 'sideToolbar.noImportedFiles'
-                  : 'sideToolbar.noGeneratedFiles'
-              )
-            "
-            :message="$t('sideToolbar.noFilesFoundMessage')"
-          />
-        </div>
+      </div>
+      <div v-else-if="!loading">
```
DrJKL: For the loading stuff, if you pulled it out you'd be able to reduce the places you have to check for it.

src/platform/assets/components/MediaVideoTop.vue
```diff
@@ -7,7 +7,11 @@
     <video
       ref="videoRef"
       :controls="shouldShowControls"
-      preload="none"
+      preload="metadata"
+      autoplay
```
DrJKL, line 11: question: Are you sure we always want autoplay here?

src/platform/remote/comfyui/history/fetchers/fetchHistoryV1.ts
```diff
@@ -15,13 +15,27 @@ import type {
...
+  // Validate offset parameter
+  if (offset !== undefined && (offset < 0 || !Number.isInteger(offset))) {
+    throw new Error(
+      `Invalid offset parameter: ${offset}. Must be a non-negative integer.`
+    )
+  }
+
+  let url = `/history?max_items=${maxItems}`
+  if (offset !== undefined) {
+    url += `&offset=${offset}`
```
DrJKL: This should probably use [UrlSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

DrJKL (dismissed):

src/platform/assets/components/MediaVideoTop.vue
viva-jinyi: Yes, we’ve decided to go with the same UX as the current Queue tab, where videos auto-play.

src/platform/remote/comfyui/history/fetchers/fetchHistoryV1.ts
viva-jinyi: ✅

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: ✅

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: ✅

DrJKL: Beautiful, thanks!

DrJKL (approved):

PR #6385, opened 2025-10-29
Fix: Allow click and drag selection when editing text (e.g. Node name)

DrJKL:
## Summary

Previously, the node would be dragged when trying to select text.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6385-Fix-Allow-click-and-drag-selection-when-editing-text-e-g-Node-name-29b6d73d36508197b73fc8f965bb34ef) by [Unito](https://www.unito.io)


AustinMroz (approved):
Seems chrome and firefox were both bugged in different ways, but the fix works for both.

PR #6401, opened 2025-10-30
Fix selectionToolbox not being on node

AustinMroz:
`getBounding()` doesn't work if a draw hasn't occurred.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6401-Fix-selectionToolbox-not-being-on-node-29c6d73d3650812cbfbff116a9b8159d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/canvas/useSelectionToolboxPosition.ts
+++ b/src/composables/canvas/useSelectionToolboxPosition.ts
@@ -5,7 +5,11 @@ import type { Ref } from 'vue'
 import { useSelectedLiteGraphItems } from '@/composables/canvas/useSelectedLiteGraphItems'
 import { useVueFeatureFlags } from '@/composables/useVueFeatureFlags'
 import type { ReadOnlyRect } from '@/lib/litegraph/src/interfaces'
-import { LGraphGroup, LGraphNode } from '@/lib/litegraph/src/litegraph'
+import {
+  LGraphGroup,
+  LGraphNode,
+  LiteGraph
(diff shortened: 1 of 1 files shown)


src/composables/canvas/useSelectionToolboxPosition.ts
```diff
@@ -95,8 +99,13 @@ export function useSelectionToolboxPosition(
...
         // Fallback to LiteGraph bounds for regular nodes or non-string IDs
         if (item instanceof LGraphNode || item instanceof LGraphGroup) {
-          const bounds = item.getBounding()
-          allBounds.push([bounds[0], bounds[1], bounds[2], bounds[3]] as const)
+          const bounds = [
+            item.pos[0],
+            item.pos[1] - LiteGraph.NODE_TITLE_HEIGHT,
+            item.size[0],
+            item.size[1] + LiteGraph.NODE_TITLE_HEIGHT
+          ]
```
DrJKL:
```suggestion
          ] as const
```

DrJKL (dismissed):

src/composables/canvas/useSelectionToolboxPosition.ts
AustinMroz: Was in a rush to get the PR out before a scheduled power outage. Alas.

Chose to resolve by removing the temp variable so there isn't a cast.

simula-r (approved):
LGTM

PR #6487, opened 2025-10-31
Fix partial execution inside subgraphs

### LINKED ISSUE #6480: An output node in a subgraph can't be used to trigger a "run this output"
### Prerequisites

### What happened?

When selected, an output node has a 'play' arrow on it, which can be used to run this output only.

If you click that arrow on an output node which is in a subgraph, the error message "Prompt has no outputs" is displayed.

### Steps to Reproduce

1. put an output node in a subgraph
2. click it's play button
3. 

Minor inconvenience

### ComfyUI Frontend Version

1.28.8

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6480-An-output-node-in-a-subgraph-can-t-be-used-to-trigger-a-run-this-output-29d6d73d3650812c8211e705b5b3c341) by [Unito](https://www.unito.io)

AustinMroz:
`getExecutionIdsForSelectedNodes`  is only used for partial execution. The prior implementation solved the wrong problem. Given a list of nodes, it would explore into subgraphs and return a list of partial ExecutionIds for all contained nodes. Because this does not resolve the partial execution path to the current subgraph, this is incorrect when the current graph is not the root graph. Woefully, this incorrect functionality is never useful because the recursive exploration only applies to subgraph nodes which never satisfy the outputNode filter applied by the parent function.

An extra function is used to correctly append the parent execution path, but the existing, probably never useful code for recursively collecting children is otherwise left in place.

Resolves #6480

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6487-Fix-partial-execution-inside-subgraphs-29d6d73d36508197924bfb3a0fb6699e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/utils/graphTraversalUtil.ts
+++ b/src/utils/graphTraversalUtil.ts
@@ -541,8 +541,16 @@ export function collectFromNodes<T = LGraphNode, C = void>(
  * @returns Array of execution IDs for selected nodes and all nodes within selected subgraphs
  */
 export function getExecutionIdsForSelectedNodes(
-  selectedNodes: LGraphNode[]
+  selectedNodes: LGraphNode[],
+  startGraph = selectedNodes[0].graph
 ): NodeExecutionId[] {
+  if (!startGraph) return []
+  const rootGraph = startGraph.rootGraph
+  const parentPath = startGraph?.isRootGraph
+    ? ''
+    : findPartialExecutionPathToGraph(startGraph, rootGraph)
+  if (parentPath === undefined) return []
+
   return collectFromNodes<NodeExecutionId, string>(selectedNodes, {
     collector: (node, parentExecutionId) => {
       const nodeId = String(node.id)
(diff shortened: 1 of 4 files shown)


src/utils/graphTraversalUtil.ts
```diff
@@ -541,8 +541,16 @@ export function collectFromNodes<T = LGraphNode, C = void>(
  * @returns Array of execution IDs for selected nodes and all nodes within selected subgraphs
  */
 export function getExecutionIdsForSelectedNodes(
-  selectedNodes: LGraphNode[]
+  selectedNodes: LGraphNode[],
+  startGraph = selectedNodes[0]?.graph
 ): NodeExecutionId[] {
+  if (!startGraph) return []
+  const rootGraph = startGraph.rootGraph
+  const parentPath = startGraph?.isRootGraph
```
DrJKL: nit: Already did a presence check
```suggestion
  const parentPath = startGraph.isRootGraph
```

src/utils/graphTraversalUtil.ts
```diff
@@ -552,7 +560,20 @@ export function getExecutionIdsForSelectedNodes(
...
+function findPartialExecutionPathToGraph(
+  target: LGraph,
+  root: LGraph
+): string | undefined {
+  for (const node of root.nodes)
+    if (node.isSubgraphNode()) {
+      if (node.subgraph === target) return `${node.id}`
+      const subpath = findPartialExecutionPathToGraph(target, node.subgraph)
+      if (subpath !== undefined) return node.id + ':' + subpath
+    }
```
DrJKL: nit/personal style preference:
```suggestion
  for (const node of root.nodes) {
    if (!node.isSubgraphNode()) continue

    if (node.subgraph === target) return `${node.id}`
    
    const subpath = findPartialExecutionPathToGraph(target, node.subgraph)
    if (subpath !== undefined) return node.id + ':' + subpath
    }
  }
```

DrJKL (dismissed):

DrJKL (approved):

Issue #6484, opened 2025-10-31
[ Vue nude ] Add a compact versions of the new nodes


JorgeR81:
### Is there an existing issue for this?

- [x] I have searched the existing issues and checked the recent builds/commits

### What problem are you experiencing?

The new Vude nodes look good, but they are very large. 

This makes it difficult to create and navigate large workflows, especially if we have node sizes and positions very optimized. 

Below are the smallest sizes we can get for the KSampler node, in Vue and Litegraph ( both at 100% Zoom ).

![Image](https://github.com/user-attachments/assets/ef8199ee-d57f-4508-9a63-aa297685a7f7)

### When does this problem occur?

Using Vue nodes in large workflows.

### How often do you encounter this problem?

Daily

### How much does this problem affect your workflow?

Causes moderate inconvenience

### Current workarounds

We can reduce zoom level, but this does solve do problem because

- The nodes are still much wider than the old design. 
So the workflow layout also needs to be changed, if it's very optimized already. 

- The font it's harder to read at lower zoom levels ( https://github.com/Comfy-Org/ComfyUI_frontend/issues/6481#issuecomment-3472335861 )
I dont think a larger font would be a good fix, because lot's of parameter names would be partially hidden.

### Ideas for solutions (Optional)

We could have a compact version of the Vue node.

- Parameter name inside the widget.
- Smaller minimum width. 

<img width="358" height="461" alt="Image" src="https://github.com/user-attachments/assets/0b9f66c3-cd45-4371-8fed-5b6fea3cca8f" />

--- 

I would also add an "ultra compact" version

- shorter header
- no space between headers 

<img width="350" height="372" alt="Image" src="https://github.com/user-attachments/assets/e29fe556-d0aa-4b6d-bb6c-e8d215238652" />

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6484-Vue-nude-Add-a-compact-versions-of-the-new-nodes-29d6d73d365081d98619e78337105a3f) by [Unito](https://www.unito.io)

jovan2009:
I agree, there even seems to be a minimum horizontal size for a node which is especially annoying, making nodes that otherwise have nothing much to show as interface occupy unreasonable amount of space.

<img width="780" height="404" alt="Image" src="https://github.com/user-attachments/assets/338f27fd-6a3b-422b-b4a8-85d36d6b833d" />

JorgeR81:
[...]

PR #6516, opened 2025-11-01
Fix badge overflow

AustinMroz:
Prevents node badges from overflowing out of the node header in vue mode
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6516-Fix-badge-overflow-29e6d73d365081328250cbeb3ebbb670) by [Unito](https://www.unito.io)


JorgeR81:
Vue nodes are already too wide ( https://github.com/Comfy-Org/ComfyUI_frontend/issues/6484 )

In a future update, consider putting the extra badge in a 2nd row.
You can do that without increasing the header's height, as you can see in the mockup.   

b1

AustinMroz:
I understand your frustration and greatly appreciate your continued involvement and feedback. This is a bugfix PR and I would like to keep it limited in scope.

(I'd even argue that it's decreasing the minimum width by enabling title truncation.)

DrJKL (approved):

DrJKL:
Merging since it's better than overflowing for now. I would like to figure out a better way to handle the initial sizing.

PR #6524, opened 2025-11-01
Remove Object URL usage for image previews

AustinMroz:
If an [Object URL](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/blob) is not revoked after being created, it causes a memory leak. [Data URLs](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data) don't have this issue, but they can clutter the DOM, require more memory due to the base64 encoding, and may result in duplicate copies being kept in memory.

This is a WIP testing branch to determine
- If this overhead is significant
- If this affects the reports of memory leaks in the frontend

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6524-Remove-Object-URL-usage-for-image-previews-29e6d73d3650818cbe3cd1ba7b3e0f22) by [Unito](https://www.unito.io)


DrJKL:
Doesn't help. Experiment was a success in that we learned something, so that's cool.

PR #6525, opened 2025-11-02
Fix node header height for subgraphs

AustinMroz:
Node headers are slightly too tall on subgraph nodes due to the height of the "enter subgraph" button. This is fixed by explicitly setting the height of the enter subgraph button.
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6525-Fix-node-header-height-for-subgraphs-29f6d73d3650810eaf95e7a6d107c2bd) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/NodeHeader.vue
+++ b/src/renderer/extensions/vueNodes/components/NodeHeader.vue
@@ -71,6 +71,7 @@
           size="sm"
           type="transparent"
           data-testid="subgraph-enter-button"
+          class="h-5"
           @click.stop="handleEnterSubgraph"
           @dblclick.stop
         >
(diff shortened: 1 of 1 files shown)


src/renderer/extensions/vueNodes/components/NodeHeader.vue
```diff
@@ -71,6 +71,7 @@
           size="sm"
           type="transparent"
           data-testid="subgraph-enter-button"
+          class="h-5"
```
DrJKL: nit: You _might_ want to do an explicit `size-5` and maybe remove the `size` attribute, not sure what all else is handled within `IconButton`

DrJKL (dismissed):

src/renderer/extensions/vueNodes/components/NodeHeader.vue
AustinMroz: I was concerned about creating side effects in the size of the icon. After testing things, I prefer your suggestion.

christian-byrne (approved):

PR #6545, opened 2025-11-02
[bugfix] Fix Vue nodes combo widgets not displaying deserialized values

christian-byrne:
Fixes an issue where combo widgets in Vue nodes would not display values from deserialized workflows if those values were not in the current options list (e.g., deleted model files, removed checkpoints). This brings Vue nodes behavior in line with legacy canvas rendering, which always displays the current value regardless of whether it exists in the options list.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6545-bugfix-Fix-Vue-nodes-combo-widgets-not-displaying-deserialized-values-29f6d73d3650810ca8cdd7976ffdeb30) by [Unito](https://www.unito.io)


christian-byrne:
All [7 Playwright test failures](https://da2d3a70.comfyui-playwright-chromium.pages.dev/#?testId=1cce8372b6f641eb4635-fc7d25b0fe1c2c8e2cc7&q=s:failed) are due to the widget now showing the value from the de-serialized data (how litegraph works).

christian-byrne:
## Refactored Approach

I've simplified the implementation based on the suggestion. Instead of modifying the options array to include deserialized values, we now display them as the placeholder text when they're not in the current options list.

**Changes:**
- Removed utility function and tests (simpler approach doesn't need them)
- No array manipulation - options remain unchanged
- Deserialized values (e.g., deleted models) are shown as placeholder text
- Addresses all performance concerns from the code review

This maintains the legacy behavior while being cleaner and more performant.

christian-byrne:
Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/6555

PR #6551, opened 2025-11-03
Fixing race condition

maludwig:
## Summary

Original PR and discussion: https://github.com/Comfy-Org/litegraph.js/pull/1082

Original Example HTML to visually describe the issue: https://github.com/Comfy-Org/litegraph.js/blob/8e5c4452cd67ccd91146ba79514bb001c470900a/public/mousemoving_example_with_blur.html

The ordering of mouseup and mousemove events is not defined. So there may be a mouseup event before the next mousemove event where buttons=0, but there may also be a mouseup immediately after.

The original code was designed to handle the case where a user click and drags, then tabs out. It turns out, you still get mousemove events in that case, so we can reset the CanvasPointer the moment it happens, as opposed to attempting to detect if it did happen in the past. The original code was designed under the assumption that if there was no mouseup, and then there was a mousemove with buttons=0, then we must have lost the focus at some point. Unfortunately, on some computers (notably my Macbook M2) this has the side effect of making it so that if your mouse pointer was even slightly moving when releasing the mouse button, it would reset and not connect the graph nodes. Extremely frustrating.

## Changes

- **What**: Window blur reset handling.
- **Breaking**: Nothing.
- **Dependencies**: Nothing.

## Review Focus

To replicate this, use a high pixel density display and a sensitive mouse device, such as a Retina Display and a macbook touchpad. This also happens in low resolution displays with less sensitive mouse devices, but the race condition is much harder to create, because you need to move at least 1 pixel during the release of the mouse button. If you want to replicate the bug, and you're on a non-Retina display or you have a basic USB mouse, you'll need to move your mouse rather vigorously. If you're on a macbook M2, just use the computer normally.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6551-Fixing-race-condition-2a06d73d365081abac2feb92ff49be63) by [Unito](https://www.unito.io)


maludwig:
Do I need to do a thing to get this merged or reviewed? It's annoying that whenever I update Comfy, I have to also patch the frontend with this fix to make the nodes work on my Macbook. Presumably it's an issue for anyone with a high DPI display and fast input device.

jtydhr88:
as discussed and testing internally we could not reproduce this issue. please re-open with the clarification if it's still a problem.
close this PR for now.

maludwig:
Hey, I'm still getting this issue. Every time I update ComfyUI I pull in this branch, rebuild it, and update the files. Find a device with a high pixel density, and a high refresh rate mouse. Any Apple device should do. ComfyUI itself can run on another computer in the cloud or a local PC or whatever. I'm more than happy to show you on a Zoom call or Discord screen share or whatever. I can reproduce it on my Windows PC, but only on my 4K monitor, with a specific gaming mouse. You need the "mousemove" event to fire at the same time as the "mouseup". So if you slowly go to a node, linger over the input for a second, and let go, it'll work. But if you quickly move to a node and let go while your cursor is moving, you'll be able to reproduce it.

I also had (in the original PR) a bare bones HTML page that showed it.

I'm turiyag on Discord. webfiltered and Robin have talked with me there previously.

maludwig:
Also, @jtydhr88 , mgrimes is able to reproduce the bug. Talk to them! Please unclose this!

PR #6555, opened 2025-11-03
Fix empty combo widget for missing models in vue

AustinMroz:
When a loaded workflow uses a model which does not exist, the combo widget would previously display as empty. This PR
- Sets the placeholder value to the localValue, so that an invalid selection still displays
- Sets the selection as invalid when the value is not in valid options so that it displays in red.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6555-Austin-vue-combo-missing-2a06d73d36508181a96ddd5c604efba4) by [Unito](https://www.unito.io)


AustinMroz:
New test expectations are required for the now displayed value, but I want to draw attention to `vue-nodes-default.png` which had fairly dramatic styling changes that didn't trip the test detection.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
```diff
@@ -68,4 +70,5 @@ const selectOptions = computed(() => {
 
   return []
 })
+const isValid = computed(() => selectOptions.value.includes(localValue.value))
```
DrJKL: optional: Using `invalid`, you can shorthand the prop instead of having to negate it
```suggestion
      :placeholder="invalid ? `${localValue ?? ''} : undefined`"
      :invalid
:options="selectOptions"
v-bind="combinedProps"
:class="cn(WidgetInputBaseClass, 'w-full text-xs')"
@@ -68,4 +70,5 @@ const selectOptions = computed(() => {

 return []
})
const invalid = computed(() => selectOptions.value.includes(localValue.value))
```

DrJKL (dismissed):

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
AustinMroz: I started with this, but it caused the computed definition to barely overflow onto 3 lines. :cry:

AustinMroz: After checking, using `invalid` as suggested instead of `isInvalid` does fit on the single line. This is much better!

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
```diff
@@ -2,6 +2,8 @@
   <WidgetLayoutField :widget>
     <Select
       v-model="localValue"
+      :placeholder="invalid ? `${localValue ?? ''}` : undefined"
```
christian-byrne: We can fall back to `combinedProps.placeholder` then to `undefined` to allow the placeholder to be set from the Python side still.

AustinMroz: Nice catch! Since `v-bind` was applied after `placeholder`,  `combinedProps.placeholder` would take priority over the logic here and still display with red coloration.

christian-byrne (approved):

PR #6562, opened 2025-11-03
fix(vite): use dynamic base URL based on cloud distribution

christian-byrne:
## Summary
- Replace hardcoded `<base href="/">` in index.html with dynamic vite base config
- Set `base: DISTRIBUTION === 'cloud' ? '/' : ''` in vite.config.mts
- Ensures proper asset loading across different deployment contexts

## Test plan
- [ ] Verify cloud distribution builds work correctly
- [ ] Verify localhost/desktop distributions work correctly
- [ ] Test asset loading in both contexts

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6562-fix-vite-use-dynamic-base-URL-based-on-cloud-distribution-2a06d73d365081c8b5d2e58870ebd14d) by [Unito](https://www.unito.io)

### DIFF
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -122,7 +122,7 @@ const gcsRedirectProxyConfig: ProxyOptions = {
 }
 
 export default defineConfig({
-  base: '',
+  base: DISTRIBUTION === 'cloud' ? '/' : '',
   server: {
     host: VITE_REMOTE_DEV ? '0.0.0.0' : undefined,
     watch: {


christian-byrne:
Going to see if I can change go server instead first, as I don't want a different base url across distributions - makes things inconsistent especially for FE 3p extension devs.

christian-byrne:
~~Replaced by https://github.com/Comfy-Org/cloud/pull/1320. No need to bring this from cloud branch.~~

Edit: still needed for nested paths.

christian-byrne:
The reason we don't want to set it for OSS is that some apps serve the application on a nested path IIRC and it may or may not break some extensions. The fact that extensions might behave differently on cloud vs. OSS is something we need to fix eventually for this - but server-side fix was a bit too involved.

PR #6563, opened 2025-11-04
Fix session cookie creation race: skip initial token refresh, wrap extension auth hooks

benceruleanlu:
Fixes a race causing “No auth header available for session creation” during sign‑in, by skipping the initial token refresh event, and wrapping extension auth hooks with async error handling.

Sentry: https://comfy-org.sentry.io/issues/6990347926/?alert_rule_id=1614600&project=4509681221369857

Context
- Error surfaced as an unhandled rejection when session creation was triggered without a valid auth header.
- Triggers: both onAuthUserResolved and onAuthTokenRefreshed fired during initial login.
- Pre‑fix, onIdTokenChanged treated the very first token emission as a “refresh” as well, so two concurrent createSession() calls ran back‑to‑back.
- One of those calls could land before a Firebase ID token existed, so getAuthHeader() returned null → createSession threw “No auth header available for session creation”.

Exact pre‑fix failure path
- src/extensions/core/cloudSessionCookie.ts
  - onAuthUserResolved → useSessionCookie().createSession()
  - onAuthTokenRefreshed → useSessionCookie().createSession()
- src/stores/firebaseAuthStore.ts
  - onIdTokenChanged increments tokenRefreshTrigger even for the initial token (treated as a refresh)
  - getAuthHeader() → getIdToken() may be undefined briefly during initialization
- src/platform/auth/session/useSessionCookie.ts
  - createSession(): calls authStore.getAuthHeader(); if falsy, throws Error('No auth header available for session creation')

What this PR changes
1) Skip initial token “refresh”
   - Track lastTokenUserId and ignore the first onIdTokenChanged for a user; only subsequent token changes count as refresh events.
   - File: src/stores/firebaseAuthStore.ts
2) Wrap extension auth hooks with async error handling
   - Use wrapWithErrorHandlingAsync for onAuthUserResolved/onAuthTokenRefreshed/onAuthUserLogout callbacks to avoid unhandled rejections.
   - File: src/services/extensionService.ts

Result
- Eliminates the timing window where createSession() runs before getIdToken() returns a token.
- Ensures any remaining errors are caught and reported instead of surfacing as unhandled promise rejections.

Notes
- Lint and typecheck run clean (pnpm lint:fix && pnpm typecheck).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6563-Fix-session-cookie-creation-race-dedupe-calls-skip-initial-token-refresh-wrap-extensio-2a16d73d365081ef8c22c5ac8cb948aa) by [Unito](https://www.unito.io)


benceruleanlu:
Only the 'Skip initial token “refresh”' is needed, rest are additional for additional hardening

src/platform/auth/session/useSessionCookie.ts
```diff
@@ -14,27 +23,40 @@ export const useSessionCookie = () => {
...
+      return
     }
 
-    const response = await fetch(api.apiURL('/auth/session'), {
-      method: 'POST',
-      credentials: 'include',
-      headers: {
-        ...authHeader,
-        'Content-Type': 'application/json'
+    createSessionInFlight = (async () => {
```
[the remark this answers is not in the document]
benceruleanlu: Resolved by removing this since it wasn't needed for the fix

benceruleanlu:
@codex review

src/services/extensionService.ts
```diff
@@ -77,22 +79,31 @@ export const useExtensionService = () => {
 
     if (extension.onAuthUserResolved) {
       const { onUserResolved } = useCurrentUser()
+      const handleUserResolved = wrapWithErrorHandlingAsync(
+        (user: AuthUserInfo) => extension.onAuthUserResolved?.(user, app)
+      )
       onUserResolved((user) => {
-        void extension.onAuthUserResolved?.(user, app)
+        void handleUserResolved(user)
```
benceruleanlu, line 97: Addressed: the async wrapper now adds structured error context for auth hooks — we log the extension name and the hook (e.g., `onAuthUserResolved`) and still route to the shared toast. This makes triage clearer without changing UX.

src/services/extensionService.ts
```diff
@@ -77,22 +79,31 @@ export const useExtensionService = () => {
...
     }
 
     if (extension.onAuthTokenRefreshed) {
       const { onTokenRefreshed } = useCurrentUser()
+      const handleTokenRefreshed = wrapWithErrorHandlingAsync(() =>
+        extension.onAuthTokenRefreshed?.()
+      )
       onTokenRefreshed(() => {
-        void extension.onAuthTokenRefreshed?.()
+        void handleTokenRefreshed()
```
benceruleanlu, line 115: Clarification: the wrapper’s return is intentionally discarded (`void handleUserResolved(user)`). Extension hooks are `void | Promise<void>`, so a potential `undefined` return from the wrapper doesn’t affect behavior.

src/stores/firebaseAuthStore.ts
```diff
@@ -102,6 +110,11 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
   // Listen for token refresh events
   onIdTokenChanged(auth, (user) => {
```
benceruleanlu, line 111: Clarification: the Pinia store is a singleton for the app’s lifetime, so we register one `onIdTokenChanged` listener and it doesn’t accumulate in normal runtime. If preferred for tests, we can add an explicit `dispose()` to call the unsubscribe, but it isn’t necessary for the app.

src/stores/firebaseAuthStore.ts
```diff
@@ -102,6 +110,11 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
...
     if (user && isCloud) {
+      // Skip initial token change
+      if (lastTokenUserId.value !== user.uid) {
```
benceruleanlu, line 114: Clarification: `onIdTokenChanged` events are delivered sequentially. We treat the first token for any `uid` as initialization by setting `lastTokenUserId` and returning; subsequent events for the same `uid` increment the trigger. We reset `lastTokenUserId` on sign‑out in `onAuthStateChanged`. This avoids spurious increments without introducing races; tests cover this behaviour in `tests-ui/tests/store/firebaseAuthStore.test.ts`.

benceruleanlu (commented):
Addressed review feedback:
- Removed in-flight dedupe in session creation to avoid suppressing legitimate calls after user switches.
- Added structured error context for extension auth hooks (extension + hook name) while preserving toast UX.
- Clarified async wrapper return is intentionally discarded for hooks typed as void/Promise<void>.
- Explained token listener lifecycle: single registration tied to store lifetime; no accumulation in runtime.
- Clarified race considerations for rapid user switches; logic treats first token per uid as initialization, subsequent tokens increment; sign-out resets.

Please let me know if you want a test-only dispose hook for the auth store’s token listener.

christian-byrne (approved):
LGTM. We can verify the fix in the data after the next deploy

PR #6565, opened 2025-11-04
More forgiving connections in vue

### LINKED ISSUE #6488: [ Vue nodes ] When I try to connect to a widget, it only connects to the first connectable widget.
### Prerequisites

### What happened?

When I try to connect to widget, it only connects to the first connectable widget. 

### Steps to Reproduce

Try to connect an INT node to the `batch_size` widget, in the Empty Latent Image node.
It only connects to the first widget ( `width` ).

Image

To make the connection, we need to connect to all widgets above.

Image

And then remove the connections above. 

Image

--- 

I also tried the KSampler node, with INT and FLOAT nodes, with the same results.

Here, I'm trying to connect to `denoise`, but it only wants to connect to `cfg`. 

Image

Only after `cfg` is connected, I can connect to `denoise`.

Image

### ComfyUI Frontend Version

1.32.0

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6488-Vue-nodes-When-I-try-to-connect-to-a-widget-it-only-connects-to-the-first-connect-29d6d73d3650811fa1b8df2bfde5131a) by [Unito](https://www.unito.io)

AustinMroz:
The previous link connection code uses [closest](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest) to find a slot. Closest only checks parents, not siblings. Since the sought element has no children, this meant connection to a slot required the mouse be directly over the slot.

This is changed by finding the closest (parent) widget or slot, and then querying for the slot. For simplicity, this means introducing an `lg-node-widget` class. As a result, connections can be made by hovering anywhere over a valid widget.

vue-connections_00001

Resolves #6488

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6565-More-forgiving-connections-in-vue-2a16d73d365081e1bf46f5d54ec382d6) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/core/canvas/links/linkDropOrchestrator.ts
+++ b/src/renderer/core/canvas/links/linkDropOrchestrator.ts
@@ -20,7 +20,9 @@ export const resolveSlotTargetCandidate = (
   const { state: dragState, setCompatibleForKey } = useSlotLinkDragUIState()
   if (!(target instanceof HTMLElement)) return null
 
-  const elWithKey = target.closest<HTMLElement>('[data-slot-key]')
+  const elWithKey = target
+    .closest('.lg-slot, .lg-node-widget')
+    ?.querySelector<HTMLElement>('[data-slot-key]')
   const key = elWithKey?.dataset['slotKey']
   if (!key) return null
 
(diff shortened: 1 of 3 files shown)


DrJKL (approved):

DrJKL:
As we discussed, one of the two places might be vestigial, but we can leave that investigation for later 🙂

PR #6566, opened 2025-11-04
Add .pnpm-store to gitignore

benceruleanlu:
Adds .pnpm-store to gitignore. It's a cache folder that sometimes appears in certain scenarios that would never need to be checked into git.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6566-Add-pnpm-store-to-gitignore-2a16d73d365081319220f90516cfad3c) by [Unito](https://www.unito.io)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ yarn.lock
 .stylelintcache
 
 node_modules
+.pnpm-store
 dist
 dist-ssr
 *.local


christian-byrne (approved):
pnpm creates these? I thought it used a system wide store

benceruleanlu:
> pnpm creates these? I thought it used a system wide store

whenever I try to mimic a CI environment, it ends up at project root, not too sure how

benceruleanlu:
image

just happened like 5 seconds ago from codex trying pnpm i

I think it might be a codex cli thing, from the sandboxing, and layering wsl2 maybe?

christian-byrne:
> whenever I try to mimic a CI environment, it ends up at project root, not too sure how

It would make sense then if you are using docker since the project root is the system root in that case right?

PR #6568, opened 2025-11-04
Hide browser tab star when autosave is enabled

benceruleanlu:
- Hide "*" indicator in the browser tab title when autosave is enabled (Comfy.Workflow.AutoSave === 'after delay').
- Refactor: extract readable computed values (`shouldShowUnsavedIndicator`, `isActiveWorkflowModified`, `isActiveWorkflowPersisted`).
- Aligns with workflow tab behavior; also hides while Shift is held (matches in-app tab logic).

Files touched:
- src/composables/useBrowserTabTitle.ts

Validation:
- Ran `pnpm lint:fix` and `pnpm typecheck` — both passed.

Manual test suggestions:
- With autosave set to 'after delay': modify a workflow → browser tab should not show `*`.
- With autosave 'off': modify or open non-persisted workflow → browser tab shows `*`.
- Hold Shift: indicator hidden while held (consistent with workflow tab).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6568-Hide-browser-tab-star-when-autosave-is-enabled-refactor-title-logic-2a16d73d365081549906e9d1fed07a42) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useBrowserTabTitle.ts
+++ b/src/composables/useBrowserTabTitle.ts
@@ -5,6 +5,7 @@ import { t } from '@/i18n'
 import { useSettingStore } from '@/platform/settings/settingStore'
 import { useWorkflowStore } from '@/platform/workflow/management/stores/workflowStore'
 import { useExecutionStore } from '@/stores/executionStore'
+import { useWorkspaceStore } from '@/stores/workspaceStore'
 
 const DEFAULT_TITLE = 'ComfyUI'
 const TITLE_SUFFIX = ' - ComfyUI'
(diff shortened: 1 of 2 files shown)


src/composables/useBrowserTabTitle.ts
```diff
@@ -24,11 +26,27 @@ export const useBrowserTabTitle = () => {
     () => settingStore.get('Comfy.UseNewMenu') !== 'Disabled'
   )
 
+  const isAutoSaveEnabled = computed(
+    () => settingStore.get('Comfy.Workflow.AutoSave') === 'after delay'
```
christian-byrne, line 30: nit: we could make the 'after delay' a shared enum or we could try typing `settingStore.get<settings['Comfy.Workflow.AutoSave']>()`. (so that if this changes in the future, a type error will occur.)

christian-byrne (approved):

src/composables/useBrowserTabTitle.ts
benceruleanlu: SGTM, my thinking at the time was that the tests would've failed on that change

maybe someone will refactor this reliance on standard strings for all settings since this seems to be a pattern in the codebase

PR #6580, opened 2025-11-04
fix minimap navigation on touch devices

christian-byrne:
Fixes minimap navigation (dragging the viewport box on the minimap) on touch devices.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6580-fix-minimap-navigation-on-touch-devices-2a16d73d36508195b070da2b8e4b908a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/minimap/composables/useMinimapInteraction.ts
+++ b/src/renderer/extensions/minimap/composables/useMinimapInteraction.ts
@@ -58,6 +58,7 @@ export function useMinimapInteraction(
   }
 
   const releasePointer = (e?: PointerEvent) => {
+    isDragging.value = false
     if (!e) return
 
     const target = e.currentTarget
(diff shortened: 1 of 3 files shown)


src/renderer/extensions/minimap/composables/useMinimapInteraction.ts
```diff
@@ -53,7 +57,25 @@ export function useMinimapInteraction(
...
+
+  const handlePointerUp = (e?: PointerEvent) => {
+    releasePointer(e)
+    isDragging.value = false
+  }
+
+  const handlePointerCancel = (e: PointerEvent) => {
+    releasePointer(e)
     isDragging.value = false
   }
```
DrJKL: Very nit, also optional:
To make it clear that they're doing the same thing (and should unless someone intentionally causes the behavior to diverge)

```suggestion
  const releasePointer = (e?: PointerEvent) => {
    isDragging.value = false
    if (!e) return

    const target = e.currentTarget
    if (
      target instanceof HTMLElement &&
      target.hasPointerCapture(e.pointerId)
    ) {
      target.releasePointerCapture(e.pointerId)
    }
  }

  const handlePointerUp = releasePointer

  const handlePointerCancel = releasePointer
```

DrJKL (dismissed):

PR #6582, opened 2025-11-04
Add front end support for type matching

AustinMroz:
This PR implements front end logic to handle MatchType inputs and outputs.
See  comfyanonymous/ComfyUI#10644

This allows for the implementation of nodes such as a "switch node"  where input types change based on the connections made.
switch-node

As part of this implementation, significant cleanup is being performed in the reroute code. Extra testing will be required to make sure these changes don't introduce regressions.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6582-Add-front-end-support-for-type-matching-2a16d73d36508189b042cd23f82a332e) by [Unito](https://www.unito.io)


AustinMroz:
Test failure is unexpected, ~~but I am not sure of the cause.~~

~~I am unable to run the test locally and manually performing the steps seems to function without issue.~~

EDIT: Figured out both issue with test environment and test, ~~working on fix~~

src/extensions/core/matchType.ts
```diff
@@ -0,0 +1,151 @@
...
+  if (!isStrings(types)) return undefined
+  const filteredTypes = types.filter((t) => t !== '*')
+  if (!filteredTypes.length) return '*'
+  const combinedSet = filteredTypes.reduce(
+    (combined, partial) => combined.intersection(new Set(partial.split(','))),
+    new Set(filteredTypes[0].split(','))
+  )
+  if (combinedSet.size == 0) return undefined
+  return [...combinedSet].join(',')
+}
```
christian-byrne, line 155: nit: I wonder if we can make this code more expressive so people can read it quickly and understand everything that it's doing.

Example:

```typescript
import { intersection, without, split } from 'es-toolkit';

/** comment explain the semantics of what `undefined` return value represents, etc. **/
export function combineTypes(...types: ISlotType[]): ISlotType | undefined {
  if (!isStrings(types)) return undefined

  const withoutWildcards = without(types, '*')

  const allWereWildcards = withoutWildcards.length === 0
  if (allWereWildcards) return '*'

  const allTypes = withoutWildcards.map(t => split(t, ','))
  const commonTypes = intersection(...allTypes) // order follows the first array

  const hasCommonTypes = !!commonTypes.length
  return hasCommonTypes ? commonTypes.join(',') : undefined
}
```

AustinMroz: Agreed.

Was real excited to see es-toolkit intersection, but it seems to have the same 'limit 2 inputs' which requires the ugly `reduce`.

I've opted to write my own intersection implementation which does handle variable number of inputs. I think it covers the same need of segmenting out the complex logic so it's clear how types are combined.

christian-byrne (approved):
I think we can merge this.

For future, the things I would consider:

- Explaining choice between changing litegraph directly vs. using monkey-patches
- Running the code through an LLM and asking how it could be made more expressive/maintainable (e.g., "how can I refactor or document this code such that a new hire could read it and fix a bug related to it?")
- Trying to see if there's any ways to make minor adjustments that would make the code better match the style of codebase a bit more (minor)

From a technical perspective, I think it is great code and I definitely want to see the feature get in the users' hands as soon as possible.

PR #6590, opened 2025-11-05
set config via feature flags

christian-byrne:
In cloud environment, allow all the config values to be set by the feature flag endpoint and be updated dynamically (on 30s poll). Retain origianl behavior for OSS. On cloud, config changes shouldn't be changed via redeploy and the promoted image should match the staging image.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6590-set-config-via-feature-flags-2a26d73d3650819f8084eb2695c51f22) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/signin/ApiKeyForm.vue
+++ b/src/components/dialog/content/signin/ApiKeyForm.vue
@@ -89,7 +89,10 @@ import { computed } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 import { getComfyPlatformBaseUrl } from '@/config/comfyApi'
-import { remoteConfig } from '@/platform/remoteConfig/remoteConfig'
+import {
+  configValueOrDefault,
+  remoteConfig
+} from '@/platform/remoteConfig/remoteConfig'
 import { apiKeySchema } from '@/schemas/signInSchema'
 import { useApiKeyAuthStore } from '@/stores/apiKeyAuthStore'
 import { useFirebaseAuthStore } from '@/stores/firebaseAuthStore'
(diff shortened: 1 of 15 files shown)


src/components/dialog/content/signin/ApiKeyForm.vue
```diff
@@ -88,14 +88,23 @@ import Message from 'primevue/message'
...
 import { useFirebaseAuthStore } from '@/stores/firebaseAuthStore'
 
 const authStore = useFirebaseAuthStore()
 const apiKeyStore = useApiKeyAuthStore()
 const loading = computed(() => authStore.loading)
+const comfyPlatformBaseUrl = computed(() => {
+  const runtimeUrl = remoteConfig.value.comfy_platform_base_url
+  if (runtimeUrl && runtimeUrl.length > 0) {
+    return runtimeUrl
+  }
```
DrJKL: nit (v1):
Empty strings are already falsy
```suggestion
  if (runtimeUrl) {
    return runtimeUrl
  }
```

src/components/dialog/content/signin/ApiKeyForm.vue
```diff
@@ -88,14 +88,23 @@ import Message from 'primevue/message'
...
+
+  return getComfyPlatformBaseUrl()
```
DrJKL: nit (v2):
Since it's falsy, we can just use an `||` with a fallback. In this case it would need to be `||` to handle empty strings as well as non-values.
```suggestion
  return runtimeUrl || getComfyPlatformBaseUrl()
```

src/config/comfyApi.ts
```diff
@@ -1,7 +1,38 @@
...
+export function getComfyApiBaseUrl(): string {
+  if (!isCloud) {
+    return BUILD_TIME_API_BASE_URL
+  }
+
+  const runtimeUrl = remoteConfig.value.comfy_api_base_url
+  return typeof runtimeUrl === 'string' && runtimeUrl.length > 0
+    ? runtimeUrl
+    : BUILD_TIME_API_BASE_URL
+}
```
DrJKL: Since this happens in several places, can we put the logic into remoteConfig?

DrJKL (dismissed):

DrJKL (dismissed):

DrJKL (approved):

AustinMroz (approved):

PR #6593, opened 2025-11-05
fix: improve template URL loading UX and prevent re-triggering

christian-byrne:
Fixes the janky UX when loading templates via URL query parameters by moving the loading logic earlier in the app lifecycle (from GraphView.onGraphReady to useWorkflowPersistence.restorePreviousWorkflow). The saved workflow now loads first as a background tab, then the template loads as the active tab, eliminating the visual flash where the saved workflow briefly appears before being replaced. After loading, the template and source query parameters are removed from the URL using router.replace to prevent the template from re-loading on page refresh. This preserves user work by keeping both workflows open in separate tabs and matches the existing behavior when clicking templates from the dialog. All 15 tests pass including 3 new tests for URL cleanup.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6593-fix-improve-template-URL-loading-UX-and-prevent-re-triggering-2a26d73d36508137a0cae6cf92c842fc) by [Unito](https://www.unito.io)


src/platform/workflow/persistence/composables/useWorkflowPersistence.ts
christian-byrne, line 1: It feels wrong because the logic doesn't seem to belong in a function named `restorePreviousWorkflow`

src/platform/workflow/templates/composables/useTemplateUrlLoader.ts
christian-byrne, line 1: Maybe we can just remove this or majorly refactor it.

src/platform/workflow/persistence/composables/useWorkflowPersistence.ts
Myestery: how about loadWorkflowFromSource?

christian-byrne: Good idea. I'll refactor into a workflowSources array and change the loader/initializer to be more generic

PR #6594, opened 2025-11-05
[bugfix] Extend missing combo value fix to dropdown widget

christian-byrne:
Extends the fix from PR #6555 to WidgetSelectDropdown (used for image/video/audio upload combos). When a loaded workflow uses a model or file which does not exist, the dropdown widget now displays the missing value as a placeholder, consistent with WidgetSelectDefault behavior.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6594-bugfix-Extend-missing-combo-value-fix-to-dropdown-widget-2a26d73d365081479d04d887207b84c8) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
@@ -127,7 +127,18 @@ const dropdownItems = computed<DropdownItem[]>(() => {
   }
 })
 
+const invalid = computed(() => {
+  if (!localValue.value) return false
+  const values = props.widget.options?.values || []
+  return !values.includes(localValue.value)
+})
+
 const mediaPlaceholder = computed(() => {
+  // Show invalid value as placeholder (similar to WidgetSelectDefault)
+  if (invalid.value) {
+    return `${localValue.value}`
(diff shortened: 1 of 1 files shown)


DrJKL (dismissed):

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -127,7 +127,18 @@ const dropdownItems = computed<DropdownItem[]>(() => {
   }
 })
 
+const invalid = computed(() => {
+  if (!localValue.value) return false
+  const values = props.widget.options?.values || []
+  return !values.includes(localValue.value)
+})
+
```
DrJKL, line 135:
```suggestion
const invalid = computed(() => {
  if (!localValue.value) return false
  const { placeholder, values } = props.widgets.options
  if (localValue.value === placeholder) return false
  const safeValues = values ?? []
  return !safeValues.includes(localValue.value)
})

```

DrJKL:
Looking into this, but I think the validity needs to be pushed down into the `FormDropdown`

DrJKL:
Closing this for now

PR #6606, opened 2025-11-06
Fix subgraph conversion of primitives

AustinMroz:
AnimateDiff_00001

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6606-Fix-subgraph-conversion-of-primitives-2a36d73d3650818e9e74dd383a7f9007) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/widgetInputs.ts
+++ b/src/extensions/core/widgetInputs.ts
@@ -168,16 +168,14 @@ export class PrimitiveNode extends LGraphNode {
 
   #onFirstConnection(recreating?: boolean) {
     // First connection can fire before the graph is ready on initial load so random things can be missing
-    if (!this.outputs[0].links) {
+    if (!this.outputs[0].links || !this.graph) {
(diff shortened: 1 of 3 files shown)


AustinMroz:
Connections after conversion were failing because the process of configuring a node would result in the config reference of `input.widget` being dropped.

I'm quite conflicted on the location of this change. The current fix for this is functional, but not great

AustinMroz:
Some further digging:
- Widgets are normally setup through a `graph.onConfigure`.  Adding a `graph.onConfigure?.apply(subgraph, data)` does result in proper setup of widget config, and seems to better handle some cases (conversion of a primitive along with it's dependent), but this isn't valid from a types perspective.

EDIT: Resolved by directly applying the node callbacks, which don't take args.

DrJKL (approved):

christian-byrne (approved):

PR #6607, opened 2025-11-06
fix: use WidgetSelectDropdown for models

arjansingh:
## Summary

As the commit says, the model loaders were broken in cloud if you enabled Vue Nodes (not a thing I think user does yet). 

This fixes it by configuring the `WidgetSelectDropdown` to load so the user load models like they would load a input or output asset.

## Review Focus

Probably `useAssetWidgetData` to make sure it's idomatic.

This part of [assetsStore](https://github.com/Comfy-Org/ComfyUI_frontend/pull/6607/files#diff-18a5914c9f12c16d9c9c3a9f6d0e203a9c00598414d3d1c8637da9ca77339d83R158-R234) as well.

## Screenshots

Screenshot 2025-11-05 at 5 34 22 PM

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6607-fix-use-WidgetSelectDropdown-for-models-2a36d73d36508143b185d06d736e4af9) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
@@ -1,10 +1,21 @@
 <script setup lang="ts">
-import { computed, provide, ref, watch } from 'vue'
+import { capitalize } from 'es-toolkit'
+import { computed, provide, ref, toRef, watch } from 'vue'
 
 import { useWidgetValue } from '@/composables/graph/useWidgetValue'
 import { useTransformCompatOverlayProps } from '@/composables/useTransformCompatOverlayProps'
 import { t } from '@/i18n'
 import { useToastStore } from '@/platform/updates/common/toastStore'
+import FormDropdown from '@/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue'
+import { AssetKindKey } from '@/renderer/extensions/vueNodes/widgets/components/form/dropdown/types'
+import type {
+  DropdownItem,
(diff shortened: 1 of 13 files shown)


src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -132,9 +151,16 @@ const outputItems = computed<DropdownItem[]>(() => {
 })
 
 const allItems = computed<DropdownItem[]>(() => {
+  if (props.isAssetMode && assetData) {
```
DrJKL, line 154: Having the mode check in so many different places is a little smelly.

src/renderer/extensions/vueNodes/widgets/composables/useAssetWidgetData.ts
```diff
@@ -0,0 +1,99 @@
...
+ * Cloud-only composable - returns empty data when not in cloud environment.
+ *
+ * @param nodeType - ComfyUI node type (ref, getter, or plain value). Can be undefined.
+ *   Accepts: ref('CheckpointLoaderSimple'), () => 'CheckpointLoaderSimple', or 'CheckpointLoaderSimple'
+ * @returns Reactive data including category, assets, dropdown items, loading state, and errors
+ */
+export function useAssetWidgetData(
+  nodeType: MaybeRefOrGetter<string | undefined>
+) {
+  if (isCloud) {
```
DrJKL, line 24: Could this condition be flipped?

DrJKL (approved):

PR #6623, opened 2025-11-06
sync devtools nodes with test ComfyUI dir in test global setup and add dev nodes for all new widget types 

christian-byrne:
Current issues:

- the widget.options and widget.spec need to be unified and the apiSchema needs to be updated to reflect the props API. Then this PR can be greatly simplified (or some other way to wire through the widget spec into the component props and have those types to the apiSchema)
- relying on type guards for all widget types instead of discriminated union

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6623-sync-devtools-nodes-with-test-ComfyUI-dir-in-test-global-setup-and-add-dev-nodes-for-all--2a36d73d365081fb8a16ee85b8ac92b5) by [Unito](https://www.unito.io)


jtydhr88:
as it keeps draft for months, and has conflicts, please rebase and resend if we still need it.
close it for now

PR #6624, opened 2025-11-06
feat: navigate to previously active tab when closing current tab

christian-byrne:
When closing a tab, the UI now returns to the most recently active tab instead of always going to the first/next tab. This matches standard browser tab behavior and prevents accidental edits in the wrong workflow. Implementation uses a lazy-cleanup history array (max 32 entries) that tracks tab activations and skips closed tabs when finding the previous tab to switch to. Fixes #6599

https://github.com/user-attachments/assets/0bb87969-fd01-4e6b-96e8-c0f741f23ff8

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6624-feat-navigate-to-previously-active-tab-when-closing-current-tab-2a36d73d365081f5be95db51ff7a03f6) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/workflow/management/stores/workflowStore.ts
+++ b/src/platform/workflow/management/stores/workflowStore.ts
@@ -158,6 +158,7 @@ interface WorkflowStore {
   isActive: (workflow: ComfyWorkflow) => boolean
   openWorkflows: ComfyWorkflow[]
   openedWorkflowIndexShift: (shift: number) => ComfyWorkflow | null
+  getMostRecentWorkflow: () => ComfyWorkflow | null
   openWorkflow: (workflow: ComfyWorkflow) => Promise<LoadedComfyWorkflow>
   openWorkflowsInBackground: (paths: {
     left?: string[]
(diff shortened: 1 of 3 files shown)


src/platform/workflow/management/stores/workflowStore.ts
```diff
@@ -405,6 +426,39 @@ export const useWorkflowStore = defineStore('workflow', () => {
...
+   * Get the most recently active workflow from history (excluding current).
+   * Lazily cleans invalid paths from history.
+   * @returns The most recent valid workflow or null if none found.
+   */
+  const getMostRecentWorkflow = (): ComfyWorkflow | null => {
+    const currentPath = activeWorkflow.value?.path
+    const validPaths: string[] = []
+
+    // Scan backwards through history
+    for (let i = tabActivationHistory.value.length - 1; i >= 0; i--) {
```
DrJKL, line 439: If we're always going backwards, can we just flip to unshift/pop above so that we can traverse forward?

DrJKL (approved):

src/platform/workflow/management/stores/workflowStore.ts
christian-byrne: Do you mean unshift/shift? And keep it LIFO? I think it could be better but downside would be other developers not expecting a history stack to be using the start of the array rather than end.

PR #6642, opened 2025-11-09
WIP: Linear Mode graph view

christian-byrne:
Add core infrastructure for Linear Mode - a simplified form-based UI for ComfyUI workflows:

- Add optional onQueued callback to app.queuePrompt() for tracking which jobs were queued in linear mode (to display a filtered output list)
- Create Linear Mode type definitions (widgets, templates, output images)
- Add template configuration with 10 promoted widgets (placeholder)
- Implement Pinia store with client-side history filtering via prompt ID tracking
- Add service layer for template loading and widget value manipulation
- Add composable for queue operations with automatic prompt tracking
- TODO: components

This is a very minimal implementation with the following limitations:

- Not bi-directional: changes made in the linear mode are propogated when converting to graph but cannot be propagated back up (when opening a workflow in linear mode, always open a new copy of one of the given hardcoded options)
- Hardcoded list of available workflows: logic does not exist to promote a given set of widgets/inputs (via user interaction or some domain-specific heuristic) or to convert arbitrary workflows into a linear view. Instead, a small number of workflow templates will be added and those templates will have logic hardcoded for them in particular in order to manipulate their widget values and create some other hand-tailored interactions.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6642-WIP-Linear-Mode-graph-view-2a66d73d3650818eaf23c5ca8fb55a8b) by [Unito](https://www.unito.io)

### DIFF
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -41,7 +41,9 @@ const config: KnipConfig = {
     'src/workbench/extensions/manager/types/generatedManagerTypes.ts',
     'packages/registry-types/src/comfyRegistryTypes.ts',
     // Used by a custom node (that should move off of this)
-    'src/scripts/ui/components/splitButton.ts'
+    'src/scripts/ui/components/splitButton.ts',
+    // Linear Mode infrastructure - exports will be used by UI components in next PR
+    'src/renderer/extensions/linearMode/**/*.ts'
   ],
   compilers: {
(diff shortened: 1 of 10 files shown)


christian-byrne:
Replaced by real implementation https://github.com/Comfy-Org/ComfyUI_frontend/pull/6670 that works on all graphs.

PR #6643, opened 2025-11-09
cleanup: Don't offset the size calculation by the litegraph title height.

DrJKL:
## Summary

Addresses nodes shrinking by just a little bit each time the resize handler starts.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6643-cleanup-Don-t-offset-the-size-calculation-by-the-litegraph-title-height-2a66d73d365081ab8473cd630b4d69c5) by [Unito](https://www.unito.io)


AustinMroz (commented):
Causes some  regressions for positioning and sizing.
- The selection indicator now has a empty gap above nodes in vue mode.
image

- Nodes are offset downwards when swapping into vue mode.
image

- Node surface area is no longer closely aligned with litegraph
image

PR #6651, opened 2025-11-11
feat: add streamer extension

rickstaa:
This commit adds a simple streamer exdtension to get streaming to work inside comfyui.


rickstaa:
Sorry for that this was should have been created on my own fork as it was for testing purposes.

PR #6664, opened 2025-11-12
Fix/vue nodes auto scale

simula-r:
## Summary

**Problem:** ensureCorrectLayoutScale scales up LG -> Vue. But doesn't scale down from Vue -> LG. 

**Solution:** Bi directional scaling.

**Bonus:** fix edge cases such as subgraphs, groups, and reroutes. Also, set auto scale: true now that we 'preserve' LG scale.

**IMPORTANT:** useVueNodeResizeTracking.ts sets vue node height - Litegraph.NODE_TITLE_HEIGHT on workflow load using a resize observer. Reloading the page (loading a workflow) in Vue mode, will subtract height each time. This can look like a problem caused by ensureCorrectLayoutScale. It is not. Need to fix. Here was an attempt by [removing the Litegraph.NODE_TITLE_HEIGHT entirely](https://github.com/Comfy-Org/ComfyUI_frontend/pull/6643).

## Review Focus

Full lifecycle of loading workflows and switching between vue and lg. Race conditions could be present. For example switching the mode using keybind very fast.

https://github.com/user-attachments/assets/5576b760-13a8-45b9-b8f7-64e1caf443c1

https://github.com/user-attachments/assets/46d6f870-df76-4084-968a-53cb629fc123


simula-r:
Need to update browser snapshots due to changing vue nodes to use borderBox instead of contentBox sizing.

src/lib/litegraph/src/LGraph.ts
```diff
@@ -79,6 +79,8 @@ export type {
   LGraphTriggerParam
 } from './types/graphTriggers'
 
+export type rendererType = 'LG' | 'Vue'
```
DrJKL, line 82: nit: RendererType, types should be UpperCamelCase

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
```diff
@@ -18,77 +26,178 @@ export function ensureCorrectLayoutScale() {
...
 
   if (!graph || !graph.nodes) return
 
-  if (graph.extra?.vueNodesScaled === true) {
-    return
+  // Use renderer from graph, default to 'LG' for the check (but don't modify graph yet)
+  if (!renderer) {
+    // Always assume legacy LG format when unknown (pre-dates this feature)
+    renderer = 'LG'
   }
```
DrJKL, line 40: Want to set this as a default in the args?

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
```diff
@@ -18,77 +26,178 @@ export function ensureCorrectLayoutScale() {
...
+
+  if (needsUpscale) {
+    graph.extra.workflowRendererVersion = 'Vue'
+  }
 
-  if (!graph.extra) graph.extra = {}
-  graph.extra.vueNodesScaled = true
+  if (needsDownscale) {
+    graph.extra.workflowRendererVersion = 'LG'
+  }
```
DrJKL, line 202: Couldn't you set this unconditionally at the top, before the guard at line 46?

src/scripts/app.ts
```diff
@@ -825,6 +825,23 @@ export class ComfyApp {
...
+        // Only scale when switching between graphs (not during initial setup)
+        // oldGraph is null/undefined during initial setup, so skip scaling then
+        if (oldGraph) {
+          ensureCorrectLayoutScale(
+            newGraph.extra.workflowRendererVersion,
+            newGraph
+          )
+        }
+      }
+    )
```
DrJKL, line 843: Any chance this could have an ordering issue with the other event listener above?

DrJKL (dismissed):

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
simula-r: Can take a look as part of a follow up.

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
simula-r: It's somewhat non trivial so want room to comment. Can take a look as part of a follow up.

src/scripts/app.ts
simula-r: Not as far as I know. :)

DrJKL (approved):

christian-byrne (approved):

christian-byrne:
No need to backport -- forgot Vue Nodes beta started on current minor version already (1.32)

PR #6670, opened 2025-11-13
Add linear mode

AustinMroz:
linear-mode

See also: #6642

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6670-Add-linear-mode-2aa6d73d365081d08887e4a6db3a8fa0) by [Unito](https://www.unito.io)


src/platform/workflow/management/stores/workflowStore.ts
```diff
@@ -329,6 +330,7 @@ export const useWorkflowStore = defineStore('workflow', () => {
       tabActivationHistory.value.shift()
     }
 
+    useCanvasStore().linearMode = !!loadedWorkflow.activeState.extra?.linearMode
```
christian-byrne, line 333: big nit: Generally I think of the dependency layers as going from:

1. libraries
2. base
3. platform
4. workbench
5. renderer

If we think of the workflow management happening at platform layer, we would generally not want it to import and use the canvas store, which operates more at the renderer layer. Instead, there could be hooks, events, or signals used to allow lower levels to react to a workflow changing. Then, the linear mode logic could be confined to its layer. A good rule of thumb is that you should be able to remove something in the lower level and not break something above.

However, this standard/architecture is not really enforced effectively at all and definitely shouldn't block a PR at this stage.

src/views/GraphView.vue
christian-byrne, line 1: note: This sort of highlights a problem we have where all the app state is bootstrapped in (and tied to the lifecycle of) the GraphView. Because in a perfect world, the LinearView and GraphView would be side-by-side (rather than parent-child) and the router could be used to switch between them. But I am guessing that was not really possible to do easily in this case?

src/views/LinearView.vue
```diff
@@ -0,0 +1,189 @@
...
+    })
+  }
+  await commandStore.execute(commandId, {
+    metadata: {
+      subscribe_to_run: false,
+      trigger_source: 'button'
+    }
+  })
+}
+function openFeedback() {
```
christian-byrne, line 98: nit: We should make a command in command store for this since it's used in multiple places now.

christian-byrne (approved):
LGTM

src/views/GraphView.vue
AustinMroz: Yeah. Having a router based implementation would be preferred, but we'd need quite a bit of refactoring to get there. GraphCanvas also owns a lot of state and I'm not terribly happy about needing to put a `v-show` on it instead of a `v-if/v-else`

src/platform/workflow/management/stores/workflowStore.ts
AustinMroz: Was getting a similar smell. I'll make a note to follow up in a future PR.

PR #6690, opened 2025-11-14
filter out templates that have custom nodes when not on cloud (temporary)

christian-byrne:
Templates are being added that have custom nodes. As a temporary measure, filter these out on local. In a followup, add a UX to allow local users to do something like opt-in to use these or to show these on the condition that the user has these custom nodes installed.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6690-filter-out-templates-that-have-custom-nodes-when-not-on-cloud-temporary-2ab6d73d3650819981c0e9b26d34acff) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/workflow/templates/repositories/workflowTemplatesStore.ts
+++ b/src/platform/workflow/templates/repositories/workflowTemplatesStore.ts
@@ -244,8 +244,7 @@ export const useWorkflowTemplatesStore = defineStore(
       const filteredTemplates = isCloud
         ? allTemplates
         : allTemplates.filter(
-            (template) =>
-              !template.requiresCustomNodes?.length
+            (template) => !template.requiresCustomNodes?.length
           )
 
       return filteredTemplates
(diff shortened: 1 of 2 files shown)


src/platform/workflow/templates/repositories/workflowTemplatesStore.ts
```diff
@@ -237,7 +237,19 @@ export const useWorkflowTemplatesStore = defineStore(
...
+      // TODO: Temporary filtering of custom node templates on local installations
+      // Future: Add UX that allows local users to opt-in to templates with custom nodes,
+      // potentially conditional on whether they have those specific custom nodes installed.
+      // This would provide better template discovery while respecting local user workflows.
+      const filteredTemplates = isCloud
+        ? allTemplates
+        : allTemplates.filter(
+            (template) =>
+              !template.requiresCustomNodes ||
+              template.requiresCustomNodes.length === 0
```
DrJKL: Optional:
```suggestion
              !template.requiresCustomNodes?.length
```

DrJKL (dismissed):

DrJKL (approved):

PR #6693, opened 2025-11-14
refactor: change telemetry service to use hook registry instead of importing directly from modules in lower layers 

christian-byrne:
Refactors the telemetry service such that it does not depend on other lower-layer modules (only library or other platform-level modules). This way, it can be imported early in the app bootstrap (e.g., in onboarding views).

Alsos extends the service interface to better support composition with multiple providers (Clickhouse, Mixpanel, DD, Sentry, etc.).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6693-refactor-change-telemetry-service-to-use-hook-registry-instead-of-importing-directly-fro-2ab6d73d365081a7bcfae20550770ef5) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/api.ts
+++ b/src/scripts/api.ts
@@ -1251,6 +1251,28 @@ export class ComfyApi extends EventTarget {
   getServerFeatures(): Record<string, unknown> {
     return { ...this.serverFeatureFlags }
   }
+
+  /**
+   * Posts analytics event to cloud analytics service
+   * @param eventName The name of the analytics event
+   * @param eventData The event data (any JSON-serializable object)
+   * @returns Promise resolving to the response
+   */
+  async postCloudAnalytics(
+    eventName: string,
+    eventData: any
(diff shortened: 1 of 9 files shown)


src/scripts/api.ts
```diff
@@ -1257,6 +1251,28 @@ export class ComfyApi extends EventTarget {
...
     return { ...this.serverFeatureFlags }
   }
+
+  /**
+   * Posts analytics event to cloud analytics service
+   * @param eventName The name of the analytics event
+   * @param eventData The event data (any JSON-serializable object)
+   * @returns Promise resolving to the response
+   */
+  async postCloudAnalytics(
```
christian-byrne, line 1261: TODO: make this tree-shaken on local

christian-byrne:
This was done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/8354, closing this.

PR #6723, opened 2025-11-17
Remove queue cancel controls

benceruleanlu:
## Summary
- remove the interrupt/clear controls from the run button cluster
- drop the queue pending count dependencies that only fed those controls

This is at the request of product design, because this functionality is being relocated to the queue progress overlay in bl-execution

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6723-Remove-queue-cancel-controls-2ae6d73d365081ddaa63ea9e3447ad7f) by [Unito](https://www.unito.io)


benceruleanlu:
Not too sure how, but browser_tests/tests/loadWorkflowInMedia.spec.ts-snapshots/dropped-workflow-url-hidream-dev-example-png-chromium-linux.png seems to be incorrectly generated, checking

benceruleanlu:
I just reset it, the two buttons that changed aren't even visible on that test.

image

christian-byrne (approved):

PR #6687, opened 2025-11-13
[Style] Compact Modern Nodes

DrJKL:
## Summary

Simple and clean is the way that we're making the nodes tonight.

## Changes

- **What**: Smaller minimum widths for nodes and labels
- **What**: Smaller font for the labels
- **What**: Removed outlines for widgets
- **What**: Fixes a text/background issue with buttons on widgets
- **What**: Smaller header
- **What**: Less padding within the node itself

## Review Focus

Check out the new styles and how they align with the Designs.

## Screenshots

| Before | After |
| --- | --- |
| image | image |

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6687-Style-Compact-Modern-Nodes-2aa6d73d365081c48db3c5491c556dc9) by [Unito](https://www.unito.io)

PR #6731, opened 2025-11-18
fix: arbitrary styles, min size <= content, ensure layout calc, trunc…

simula-r:
### Problem:
After [vue node compacting PR](https://github.com/Comfy-Org/ComfyUI_frontend/pull/6687) the white space within the node has been greatly reduced, lowering the min intrinsic size, thus allowing us to reduce the amount we need to scale up via ensureCorrectLayoutScale(), therefore increasing readability of nodes. Great! 

However, a side effect of reducing the scale factor means nodes with larger min content will not be scaled up enough causing nodes to be too large in many cases.

For example, if the min intrinsic width is very long due to input length:
image

### Solution:
Allow for nodes to be resized less than their intrinsic min width. And truncate widget inputs like many other node UIs do.

IMPORTANT: when a node is added via search or other, it will still get a min size based on its intrinsic content it just wont be the min width! So best of both worlds. 

image

this means we choose a default min width and clamp resize to it. This also means we have to remove the arbitrary min width values that were sprinkled around the vue node widgets. They are not needed because instead of min width, they can take up full width and inherit the sizing from the node min width! This makes nodes like little browser windows and widgets are just responsive elements with in. Much more natural imo.

### Bonus
- Set ensureCorrectLayouScale() to scale factor of 1.2 which means vue nodes are now only being set 20% bigger than LG. That covers for the height difference we cant change! 
- Fix ensureCorrectLayouScale() to offset y position for groups / better alignment
- Get rid of arbitrary inflexible min width like min-[417px] which shouldnt have been used the first place
- Make Select and Input overlay portals width set to their content

## Changes

**What**: 
- Node resizing behavior
- Node widget min width
- Widget input and slot truncation
- Misc arbitrary styling that should have been fluid

https://github.com/user-attachments/assets/3ea4b8fe-565a-47f7-b3ab-6cef56cecde5

https://github.com/user-attachments/assets/2fe1e1a0-a9dc-4000-b865-ce2d8c7f3606

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6731-fix-arbitrary-styles-min-size-content-ensure-layout-calc-trunc-2af6d73d365081eab507c2f1638a4194) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM

Issue #5827, opened 2025-09-28
Vue node: The tab for filtering images in the input/output folder seems not working.


comfyui-wiki:
https://github.com/user-attachments/assets/75e51bcd-7a3e-478f-9041-a1cffa5c9b76

<img width="764" height="915" alt="Image" src="https://github.com/user-attachments/assets/5d040b89-9151-4ca6-900b-40b7c64e2c9c" />

LittleSound:
I don’t know how to make this filter feature. hope to get some guidance 🙏

comfyui-wiki:
@LittleSound  I don't know if this helps. There is another node "Load image (from output)", and it uses a different route.

<img width="1017" height="443" alt="Image" src="https://github.com/user-attachments/assets/9896e8c5-8892-4828-adc7-915b770a7ad0" />

https://github.com/comfyanonymous/ComfyUI/blob/master/nodes.py#L1756

christian-byrne:
Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/5894

comfyui-wiki:
I tried it again today, but still found that only very few output images can be loaded 

In the load image (from output) node, you can see that there is a long image list
<img width="1852" height="1900" alt="Image" src="https://github.com/user-attachments/assets/d5c23315-45ed-4101-b0cd-d4cf31531843" />

But after switching to the Vue node, only one image shows up.

<img width="2882" height="1710" alt="Image" src="https://github.com/user-attachments/assets/763bab71-2802-4c20-84cf-6dd33dafa536" />


https://github.com/user-attachments/assets/0fb4870f-dc7a-4a33-a816-43236853700e

PR #6734, opened 2025-11-18
fix: dropdown widget fetching output files

LittleSound:
related https://github.com/Comfy-Org/ComfyUI_frontend/issues/5827#issuecomment-3539629932

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6734-fix-dropdown-widget-fetching-output-files-2af6d73d365081988340f74694b7cee7) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
@@ -7,6 +7,7 @@ import { useTransformCompatOverlayProps } from '@/composables/useTransformCompat
 import { appendCloudResParam } from '@/platform/distribution/cloudPreviewUtil'
 import { SUPPORTED_EXTENSIONS_ACCEPT } from '@/extensions/core/load3d/constants'
 import { useAssetFilterOptions } from '@/platform/assets/composables/useAssetFilterOptions'
+import { useMediaAssets } from '@/platform/assets/composables/media/useMediaAssets'
 import {
   filterItemByBaseModels,
   filterItemByOwnership
@@ -33,9 +34,9 @@ import { useAssetWidgetData } from '@/renderer/extensions/vueNodes/widgets/compo
 import type { ResultItemType } from '@/schemas/apiSchema'
 import { api } from '@/scripts/api'
 import { useAssetsStore } from '@/stores/assetsStore'
-import { useQueueStore } from '@/stores/queueStore'
 import type { SimplifiedWidget } from '@/types/simplifiedWidget'
 import type { AssetKind } from '@/types/widgetTypes'
+import { getMediaTypeFromFilename } from '@/utils/formatUtil'
 import {
   PANEL_EXCLUDED_PROPS,
   filterWidgetProps
@@ -68,7 +69,8 @@ const modelValue = defineModel<string | undefined>({
 
 const { t } = useI18n()
 const toastStore = useToastStore()
-const queueStore = useQueueStore()
+
+const outputMediaAssets = useMediaAssets('output')
 
 const transformCompatProps = useTransformCompatOverlayProps()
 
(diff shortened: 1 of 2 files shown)


src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdownMenu.vue
```diff
@@ -67,15 +67,15 @@ const searchQuery = defineModel<string>('searchQuery')
...
-            class="icon-[lucide--circle-off] size-30 text-zinc-500/20"
-          />
-        </div>
+        <template v-if="items.length === 0">
+          <div class="absolute inset-0 flex items-center justify-center">
+            <i
+              title="No items"
+              class="icon-[lucide--circle-off] size-30 text-zinc-500/20"
+            />
+          </div>
```
christian-byrne: note: We also have this https://github.com/Comfy-Org/ComfyUI_frontend/blob/e42715086ef2175eb02ee3f9a9f60efd353a5211/src/components/common/NoResultsPlaceholder.vue#L4

(not blocking this PR).

src/stores/outputsStore.ts
```diff
@@ -0,0 +1,72 @@
...
+      const ext = filename.toLowerCase().split('.').pop() || ''
+      return ['mp4', 'webm', 'mov', 'avi', 'mkv'].includes(ext)
+    })
+  })
+  const outputAudios = computed(() => {
+    return outputs.value.filter((filename) => {
+      const ext = filename.toLowerCase().split('.').pop() || ''
+      return ['mp3', 'ogg', 'wav', 'flac'].includes(ext)
+    })
+  })
```
christian-byrne: Do you think it could be optimized to only need a single traversal?

christian-byrne (dismissed):

src/stores/outputsStore.ts
```diff
@@ -0,0 +1,72 @@
+import { api } from '@/scripts/api'
+import { defineStore } from 'pinia'
+import { computed, ref } from 'vue'
+
+export const useOutputsStore = defineStore('outputs', () => {
```
viva-jinyi: https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/platform/assets/composables/media/useMediaAssets.ts
[const outputAssets = useMediaAssets('output')](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/components/sidebar/tabs/AssetsSidebarTab.vue#L229)
It looks like most of the code in `src/stores/outputsStore.ts` could be replaced with something like `const outputAssets = useMediaAssets('output')`. I think it would be better to use the existing store. What do you think?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -121,33 +122,30 @@ const inputItems = computed<DropdownItem[]>(() => {
...
-      if (output.type === 'output' && isTargetType) {
-        const path = output.subfolder
-          ? `${output.subfolder}/${output.filename}`
-          : output.filename
-        // Add [output] annotation so the preview component knows the type
-        const annotatedPath = `${path} [output]`
-        outputs.add(annotatedPath)
-      }
-    })
+  const outputFiles = ((): string[] => {
```
viva-jinyi: https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/platform/assets/composables/useMediaAssetFiltering.ts#L62
And you can reuse this format utility!

viva-jinyi (changes requested):
It doesn’t seem to cover both Cloud and OSS. Since `internal/files` can only fetch output files in the OSS environment, I think it would be better to use `useMediaAssets.ts`. It’s already implemented to work for both OSS and Cloud, and we can update it later when the new API becomes available.

LittleSound:
I tried this API, but it doesn't work. I don't know why. you can see this image, so I use the fetch to get the list

However, I still did it according to your request. You can see it here https://github.com/Comfy-Org/ComfyUI_frontend/pull/6809, and you can actually run the code to compare their differences. 

CleanShot 2025-11-22 at 00 12 31@2x

viva-jinyi:
@LittleSound Can you resolve conflicts?

LittleSound:
Okay, I will solve this.

viva-jinyi (dismissed):
LGTM

christian-byrne:
Rebased onto current main (squashed into a single clean commit) and updated the implementation:

- Uses `useMediaAssets('output')` as viva-jinyi requested (works for both Cloud and OSS)
- Uses the existing `getMediaTypeFromFilename` utility instead of duplicating extension lists
- Converts `FormDropdown` `isOpen` from `ref` to `defineModel` so the parent can react to open/close
- Refreshes output assets on dropdown open
- Adds audio support to output file filtering
- Typecheck passes, ESLint clean

Ready for re-review.

christian-byrne (approved):
LGTM

PR #6741, opened 2025-11-18
Fix: Simplify the widget state logic

DrJKL:
## Summary

Fixes the case where a value is updated in the graph but the result doesn't reflect on the widget representation on the relevant node.

## Changes

- **What**: Uses vanilla Vue utilities instead of a special utility
- **What**: Fewer places where state could be desynced.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6741-Fix-WIP-Simplify-the-widget-state-logic-2af6d73d36508160b729db50608a2ea9) by [Unito](https://www.unito.io)


AustinMroz (dismissed):
This also solves a syncronization bug with linear mode I was procrastinating. Very nice!

christian-byrne (dismissed):

PR #6092, opened 2025-10-16
Implement workflow progress panel

benceruleanlu:
Adds a workflow progress panel component underneath the `actionbar-container`.

I suggest starting a review at the extraneous changes that were needed. Including but not limited to:

- `get createTime()` in queueStore
- `promptIdToWorkflowId`, `initializingPromptIds`, and `nodeProgressStatesByPrompt` in executionStore
- `create_time` handling in v2ToV1Adapter
- `pointer-events-auto` on ComfyActionbar.vue

The rest of the changes should be contained under `QueueProgressOverlay.vue`, and has less of a blast radius in case something goes wrong.

PR #6752, opened 2025-11-19
Support renaming widgets

AustinMroz:
widget-rename_00002

Widget labels are saved by serializing the value on inputs. This requires minor changes to ensure widgets inputs are serialized when required.

Currently only exposed by right clicking on widgets directly. Should probably be added to the subgraph config panel in the future.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6752-Support-renaming-widgets-2b06d73d36508196bff2e511c6e7b89b) by [Unito](https://www.unito.io)


src/services/litegraphService.ts
```diff
@@ -822,14 +823,31 @@ export const useLitegraphService = () => {
           }
         )
       }
-      if (this.graph && !this.graph.isRootGraph) {
-        const [x, y] = canvas.graph_mouse
-        const overWidget = this.getWidgetOnPos(x, y, true)
-        if (overWidget) {
+      const [x, y] = canvas.graph_mouse
+      const overWidget = this.getWidgetOnPos(x, y, true)
+      if (overWidget) {
```
DrJKL, line 828: Nit: Want to flip this and `if (input)` to be guard clauses?

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -835,6 +835,9 @@ export class LGraphNode
       for (const w of this.widgets) {
         if (!w) continue
 
+        const input = this.inputs.find((i) => i.widget?.name === w.name)
+        if (input?.label) w.label = input.label
```
DrJKL, line 839: Minor concern about mutation here.
Is there a way to restore the original name?

DrJKL (dismissed):

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: No. :thinking:  Might need to kick this question to design.

The input displays the original name as placeholder and renaming to an empty string removes the label?

DrJKL (approved):

huchenlei (commented):
See https://github.com/Comfy-Org/ComfyUI_frontend/pull/6092#discussion_r2542953864

I think these playwright expectation update are unrelated to this PR. Must be regressed by a previous change.

AustinMroz:
The styling changes not tripping tests is a known issue, yes.

The new context menu item requires new expectations and the stale styling changes will come along with it.

DrJKL:
@huchenlei I think the issue is Playwright's [default tolerance for color changes](https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-1-option-threshold) is very forgiving when it comes to shades of gray.

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: https://github.com/user-attachments/assets/af344318-dac2-4611-b080-910cdfa1e87d

Quick demonstration. Correctly wiring the placeholder value would be a little invovled, but we can differentiate between canceled and empty

christian-byrne (approved):

PR #6769, opened 2025-11-20
Fix linear mode with vue

AustinMroz:
Previously, entering linear mode from vue mode would cause all nodes to be set to position 0.

This is fixed by ignoring resize updates with a contentRect of all 0s

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6769-Fix-linear-mode-with-vue-2b16d73d36508188964bcfb8b465dcb1) by [Unito](https://www.unito.io)


DrJKL (dismissed):

AustinMroz:
Test failures are not expected. Should we swap to the more restrictive linearMode check?

DrJKL:
> Test failures are not expected. Should we swap to the more restrictive linearMode check?

As a quick fix, yeah.

christian-byrne (approved):

PR #6801, opened 2025-11-21
[bugfix] Fix double-click required after pasting URL in upload model dialog

luke-mino-altherr:
## Summary
Fixed an issue where users had to click twice to continue after pasting a URL in the upload model dialog - once to blur the input, then again to click the button.

## Changes
- **What**: Replaced `UrlInput` with plain `InputText` in `UploadModelUrlInput` to emit value immediately on input instead of only on blur
- **Cleanup**: Moved URL cleaning/normalization to the `fetchMetadata` handler, removed unused `disableValidation` prop from `UrlInput` component

## Review Focus
- URL normalization logic in `useUploadModelWizard.ts`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6801-bugfix-Fix-double-click-required-after-pasting-URL-in-upload-model-dialog-2b26d73d3650811881aed0cc064efcc7) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/assets/composables/useUploadModelWizard.ts
+++ b/src/platform/assets/composables/useUploadModelWizard.ts
@@ -63,11 +63,11 @@ export function useUploadModelWizard(modelTypes: Ref<ModelTypeOption[]>) {
     if (!canFetchMetadata.value) return
 
     // Clean and normalize URL
-    let cleanedUrl = wizardData.value.url.replace(/\s+/g, '')
+    let cleanedUrl = wizardData.value.url.trim()
     try {
-      cleanedUrl = new URL(cleanedUrl).toString()
+      cleanedUrl = new URL(encodeURI(cleanedUrl)).toString()
     } catch {
-      // If URL parsing fails, just use the cleaned input
+      // If URL parsing fails, just use the trimmed input
     }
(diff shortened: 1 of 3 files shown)


christian-byrne (dismissed):

src/platform/assets/composables/useUploadModelWizard.ts
```diff
@@ -62,6 +62,15 @@ export function useUploadModelWizard(modelTypes: Ref<ModelTypeOption[]>) {
   async function fetchMetadata() {
     if (!canFetchMetadata.value) return
 
+    // Clean and normalize URL
+    let cleanedUrl = wizardData.value.url.replace(/\s+/g, '')
```
DrJKL: Some URLs have spaces in them that would normally be encoded as `%20` but this would just remove them.
I think you might want to do a `.trim()` and then [`encodeURI`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI).

DrJKL (changes requested):
Forgot to hit request changes.

DrJKL (approved):

PR #6811, opened 2025-11-21
Allow graph navigation by browser forward/backward

AustinMroz:
- On graph change, set the `graph.id` as location hash
- On hash change, navigate to the target `graph.id` either in the current, or any other loaded workflow.

`canvasStore.currentGraph` does not trigger when `app.loadGraphData` is called. A trigger could be forced here, but I'm concerned about side effects. Instead `updateHash` is manually called.

Code search shows that there are no current custom nodes using `onhashchange`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6811-Allow-graph-navigation-by-browser-forward-backward-2b26d73d365081bb8414fdf7c3686124) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/fixtures/ws.ts
+++ b/browser_tests/fixtures/ws.ts
@@ -29,7 +29,7 @@ export const webSocketFixture = base.extend<{
             function ([data, url]) {
               if (!url) {
                 // If no URL specified, use page URL
-                const u = new URL(window.location.toString())
+                const u = new URL(window.location.toString().split('#')[0])
                 u.protocol = 'ws:'
                 u.pathname = '/'
                 url = u.toString() + 'ws'
--- a/browser_tests/fixtures/ws.ts
+++ b/browser_tests/fixtures/ws.ts
@@ -29,7 +29,8 @@ export const webSocketFixture = base.extend<{
             function ([data, url]) {
               if (!url) {
                 // If no URL specified, use page URL
-                const u = new URL(window.location.toString().split('#')[0])
+                const u = new URL(window.location.href)
+                u.hash = ''
                 u.protocol = 'ws:'
                 u.pathname = '/'
                 url = u.toString() + 'ws'
--- a/browser_tests/tests/subgraph.spec.ts
+++ b/browser_tests/tests/subgraph.spec.ts
@@ -767,11 +767,9 @@ test.describe('Subgraph Operations', { tag: ['@slow', '@subgraph'] }, () => {
       })
 
       // Click breadcrumb to navigate back to parent graph
-      const homeBreadcrumb = comfyPage.page.getByRole('link', {
-        // In the subgraph navigation breadcrumbs, the home/top level
-        // breadcrumb is just the workflow name without the folder path
-        name: 'subgraph-with-promoted-text-widget'
-      })
+      const homeBreadcrumb = comfyPage.page.locator(
+        '.p-breadcrumb-list > :first-child'
+      )
       await homeBreadcrumb.waitFor({ state: 'visible' })
       await homeBreadcrumb.click()
       await comfyPage.nextFrame()
--- a/src/components/breadcrumb/SubgraphBreadcrumbItem.vue
+++ b/src/components/breadcrumb/SubgraphBreadcrumbItem.vue
@@ -1,5 +1,5 @@
 <template>
-  <a
+  <div
     ref="wrapperRef"
     v-tooltip.bottom="{
       value: tooltipText,
@@ -22,7 +22,7 @@
     <span class="p-breadcrumb-item-label px-2">{{ item.label }}</span>
     <Tag v-if="item.isBlueprint" value="Blueprint" severity="primary" />
     <i v-if="isActive" class="pi pi-angle-down text-[10px]"></i>
-  </a>
+  </div>
   <Menu
     v-if="isActive || isRoot"
     ref="menu"
--- a/src/components/breadcrumb/SubgraphBreadcrumbItem.vue
+++ b/src/components/breadcrumb/SubgraphBreadcrumbItem.vue
@@ -6,7 +6,6 @@
       showDelay: 512
     }"
     draggable="false"
-    href="#"
     class="p-breadcrumb-item-link h-8 cursor-pointer px-2"
     :class="{
       'flex items-center gap-1': isActive,
--- a/src/stores/subgraphNavigationStore.ts
+++ b/src/stores/subgraphNavigationStore.ts
@@ -1,12 +1,14 @@
 import QuickLRU from '@alloc/quick-lru'
 import { defineStore } from 'pinia'
 import { computed, ref, shallowRef, watch } from 'vue'
+import { useRouteHash } from '@vueuse/router'
 
 import type { DragAndScaleState } from '@/lib/litegraph/src/DragAndScale'
 import type { Subgraph } from '@/lib/litegraph/src/litegraph'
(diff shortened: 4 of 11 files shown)


src/stores/subgraphNavigationStore.ts
```diff
@@ -157,6 +156,60 @@ export const useSubgraphNavigationStore = defineStore(
         onNavigated(newValue, oldValue)
       }
     )
+    //Allow navigation with forward/back buttons
+    //TODO: Extend for dialogues?
+    //TODO: force update widget.promoted
+    function onHashChange() {
```
christian-byrne: How about on page reload? I.e., initial load

christian-byrne:
Would consider using https://vueuse.org/router/useRouteHash/

AustinMroz:
> Would consider using https://vueuse.org/router/useRouteHash/

Already got pinged by Alex for this and made the change. Will push once I fix your suggestion of making navigation work on initial load.

browser_tests/fixtures/ws.ts
```diff
@@ -29,7 +29,7 @@ export const webSocketFixture = base.extend<{
             function ([data, url]) {
               if (!url) {
                 // If no URL specified, use page URL
-                const u = new URL(window.location.toString())
+                const u = new URL(window.location.toString().split('#')[0])
```
DrJKL: Could you use this? https://developer.mozilla.org/en-US/docs/Web/API/Location/hash

src/components/breadcrumb/SubgraphBreadcrumbItem.vue
DrJKL: Should this be a [RouterLink](https://router.vuejs.org/api/variables/RouterLink.html#RouterLink)?

src/stores/subgraphNavigationStore.ts
```diff
@@ -157,6 +158,72 @@ export const useSubgraphNavigationStore = defineStore(
...
       }
     )
+    const routeHash = useRouteHash()
+    let blockHashUpdate = false
+    let initialLoad = true
+
+    //Allow navigation with forward/back buttons
+    //TODO: Extend for dialogues?
+    //TODO: force update widget.promoted
+    async function navigateToHash(newHash: string | undefined | null) {
```
DrJKL: I think you're doing a lot imperatively here that could be done through Route definitions and use of the Router, but I don't know for sure how easy it would be to transition to that.

src/components/breadcrumb/SubgraphBreadcrumbItem.vue
AustinMroz: From quick testing, it wouldn't be a quick substitution. It's causing errors on click about missing state.

Perhaps we move in the opposite direction and replace the `<a>` with a `<div>` since it's not actually performing a navigation?

AustinMroz: From quick testing, it wouldn't be a quick substitution. It's causing errors on click about missing state.

Perhaps we move in the opposite direction and replace the <a> with a <div> since it's not actually performing a navigation?

src/stores/subgraphNavigationStore.ts
AustinMroz: My understanding is that it would be particularly hard. The navigation to a new graph happens before any event listeners or watched values change. The correct router based approach would mean initiating a navigation to the new address which causes the new graph or subgraph to load.

src/stores/subgraphNavigationStore.ts
```diff
@@ -158,13 +158,87 @@ export const useSubgraphNavigationStore = defineStore(
       }
     )
 
+    //Allow navigation with forward/back buttons
+    const routeHash = ref(window.location.hash)
+    const originalOnHashChange = window.onhashchange
+    window.onhashchange = (...args) => {
```
Myestery: we could use `addEventListener` here to avoid overwriting existing handlers

AustinMroz: That's much nicer. :+1:

browser_tests/fixtures/ws.ts
DrJKL: Optional:
```suggestion
                const u = new URL(window.location.href)
                u.hash = ''
```

browser_tests/tests/subgraph.spec.ts
```diff
@@ -671,11 +671,9 @@ test.describe('Subgraph Operations', () => {
...
 
       // Click breadcrumb to navigate back to parent graph
-      const homeBreadcrumb = comfyPage.page.getByRole('link', {
-        // In the subgraph navigation breadcrumbs, the home/top level
-        // breadcrumb is just the workflow name without the folder path
-        name: 'subgraph-with-promoted-text-widget'
-      })
+      const homeBreadcrumb = comfyPage.page.locator(
+        '.p-breadcrumb-list > :first-child'
+      )
```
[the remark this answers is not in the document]
DrJKL: @AustinMroz Should we add a testid for these? Hit this in my PR too.

src/components/breadcrumb/SubgraphBreadcrumbItem.vue
```diff
@@ -1,12 +1,11 @@
 <template>
-  <a
+  <div
```
DrJKL, line 2: Is it technically a Button?

browser_tests/tests/subgraph.spec.ts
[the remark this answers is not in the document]
AustinMroz: That'd be best, yeah.

In one of my rebases now lost to the seas of time, I was using a locator on `.back-button`, but that's slightly different since it's "navigate to parent" instead of "navigate to root"

src/stores/subgraphNavigationStore.ts
```diff
@@ -158,13 +161,80 @@ export const useSubgraphNavigationStore = defineStore(
       }
     )
 
+    //Allow navigation with forward/back buttons
+    let blockHashUpdate = false
+    let initialLoad = true
+
+    async function navigateToHash(newHash: string) {
```
DrJKL, line 168: Can definitely do as a follow-up, but this seems like it's worth extracting for more direct testability (and maybe further decomposition, maybe even reuse if we want to support Navigation API stuff for domains other than subgraphs)

src/stores/subgraphNavigationStore.ts
```diff
@@ -158,13 +161,80 @@ export const useSubgraphNavigationStore = defineStore(
...
+      if (!routeHash.value) await router.replace('#' + app.rootGraph.id)
+      const currentId = routeHash.value?.slice(1)
+      if (!newId || newId === currentId) return
+
+      await router.push('#' + newId)
+    }
+    //update navigation hash
+    //NOTE: Doesn't apply on workflow load
+    watch(() => canvasStore.currentGraph, updateHash)
+    watch(routeHash, () => navigateToHash(String(routeHash.value)))
```
DrJKL, line 228: This feels like it could lead to an infinite loop preeeetty easily.

DrJKL (dismissed):
I'd love an integration test or two, but let's try it 👍🏻

DrJKL (approved):

src/stores/subgraphNavigationStore.ts
AustinMroz: Agreed.

Part of the reason for the renewed push was an interest in allowing forward/back to enter/exit app mode. Assuming this PR doesn't cause problems, I'm planning a follow-up PR in the near future.

PR #6820, opened 2025-11-22
make vue node settings appear higher in the settings dialog

christian-byrne:
makes setting groups/categories be sorted by highest internal setting field `sortOrder` and adds high `sortOrder` values to the Vue Nodes (Nodes 2.0) settings.

Selection_2371


AustinMroz (approved):
LGTM

PR #6874, opened 2025-11-23
fix load image from outputs node combo value displays on cloud

christian-byrne:
show filenames instead of hashes when on cloud

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6874-fix-load-image-from-outputs-node-combo-value-displays-on-cloud-2b46d73d365081948b48ef9ff912dc20) by [Unito](https://www.unito.io)


christian-byrne:
This doesn't work. Only inputs have mappings, so this basically breaks the node.

Need to fix once we are on assets API entirely.

This node is basically obsolete anyway.

PR #6881, opened 2025-11-24
Remove unsupported workflow description fields

benceruleanlu:
## Summary
Remove top-level `description` keys from workflows because they are not valid in GitHub Actions workflow syntax (see https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax?utm_source=chatgpt.com).

## Changes
- **What**: delete the unsupported `description:` field from all workflow YAMLs under `.github/workflows/`.

## Review Focus
- Confirm workflows still show intended names and triggers without the invalid `description` metadata.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6881-Remove-unsupported-workflow-description-fields-2b56d73d365081ed9f20eb7f57956bc6) by [Unito](https://www.unito.io)


DrJKL (approved):
Nit: I think we could probably get rid of `Description: `, if it's the only comment at the top of each file, but LGTM otherwise.

christian-byrne (approved):

PR #6893, opened 2025-11-24
[feat] Improve queue job item UX based on design feedback

benceruleanlu:
## Summary
- Running jobs now show cancel button at all times (always visible, not just on hover)
- Cancel/delete buttons use destructive red styling by default with hover state
- Changed pending job icon from clock to loader-circle with spin animation
- Fixed icon buttons to be square (size-6) instead of rectangular
- Added TODO comment for future declarative button config system
- Pending hint ("Job added to queue") now shows only once per entry and no longer resets when other jobs update
- Spinner animation now applies only to the pending loader icon; completed/check icons no longer spin
- Queue overlay hover/active state also triggers when hovering the top menu bar so controls stay visible

## Design Spec
https://www.notion.so/comfy-org/Design-Queue-Dialog-Job-Ordering-and-Cancel-Button-Visibility-2b46d73d365081748a43d5cc9fbe2639

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6893-feat-Improve-queue-job-item-UX-based-on-design-feedback-2b56d73d365081a2bc7ef6f6fea1c739) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/TopMenuSection.vue
+++ b/src/components/TopMenuSection.vue
@@ -1,5 +1,10 @@
 <template>
-  <div v-if="!workspaceStore.focusMode" class="ml-1 flex gap-x-0.5 pt-1">
+  <div
+    v-if="!workspaceStore.focusMode"
+    class="ml-1 flex gap-x-0.5 pt-1"
+    @mouseenter="isTopMenuHovered = true"
+    @mouseleave="isTopMenuHovered = false"
+  >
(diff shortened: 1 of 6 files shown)


christian-byrne (approved):
LGTM. Code rabbit nits are worth doing but don't need to block these improvements.

PR #6894, opened 2025-11-24
[feat] Add rectangular hover area tracking for queue overlay

benceruleanlu:
## Summary
Adds hover detection over the entire rectangular bounding box of the queue area (actionbar + QueueProgressOverlay combined).

## Implementation
Uses `useMouse` + `useElementBounding` from VueUse to compute whether the mouse position is within the element's bounds:
```typescript
const isQueueAreaHovered = computed(
  () =>
    mouseX.value >= left.value &&
    mouseX.value <= right.value &&
    mouseY.value >= top.value &&
    mouseY.value <= bottom.value
)
```

This approach is needed because `QueueProgressOverlay` has `pointer-events-none` on its wrapper, which would create "holes" in hover detection with standard `mouseenter`/`mouseleave` events.

## Status
⏳ **Waiting for design confirmation** on whether we need the strict rectangular hover area tracking, or if standard hover events would suffice.

## Depends on
- #6893

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6894-feat-Add-rectangular-hover-area-tracking-for-queue-overlay-2b56d73d365081809591d100eaab1b5d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/TopMenuSection.vue
+++ b/src/components/TopMenuSection.vue
@@ -4,7 +4,7 @@
       <SubgraphBreadcrumb />
     </div>
 
-    <div class="mx-1 flex flex-col items-end gap-1">
+    <div ref="queueAreaRef" class="mx-1 flex flex-col items-end gap-1">
       <div
         class="actionbar-container pointer-events-auto flex h-12 items-center rounded-lg border border-interface-stroke px-2 shadow-interface"
       >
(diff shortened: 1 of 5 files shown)


benceruleanlu:
Even if design says we need it, we can probably replace this with dom/css gymnastics

benceruleanlu:
this is already implemented (?)

PR #6895, opened 2025-11-24
[feat] Show "Finished in" duration for completed jobs in cloud

benceruleanlu:
## Summary
In cloud distribution, completed jobs now show "Finished in Xh Ym Zs" as the primary text instead of the filename.

- Uses `formatDuration` to display time as `1h 30m 45s`, `30m 45s`, or `45s`
- Gated with `isCloud` - non-cloud continues to show filename
- Added i18n key `queue.completedIn` for localization

Filename is not fetchable right now in cloud. This is what design wanted as the alternative.

image

### DIFF
--- a/src/composables/queue/useJobList.ts
+++ b/src/composables/queue/useJobList.ts
@@ -3,6 +3,7 @@ import { useI18n } from 'vue-i18n'
 
 import { useQueueProgress } from '@/composables/queue/useQueueProgress'
 import { st } from '@/i18n'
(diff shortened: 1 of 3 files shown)


benceruleanlu:
@codex review

christian-byrne (approved):
LGTM but does the `isCloud` need to be part of the job context? I guess if we eventually allow switching between desktop and cloud? Otherwise it might be simpler to just use as a global variable.

PR #6896, opened 2025-11-24
Add cloud backport tagging workflow

benceruleanlu:
## Summary
- add workflow to tag merged backport-labeled PRs into cloud/* with cloud/v<package.json version>
- validate branch/version alignment and skip when tag already exists
- use .nvmrc (Node 24) for setup-node

## Question
- This workflow expects the version bump to already be in the PR when you merge, as it fails if the tag already exists to keep immutability. Should we autobump in this case?

### DIFF
--- a/.github/workflows/cloud-backport-tag.yaml
+++ b/.github/workflows/cloud-backport-tag.yaml
@@ -1,0 +1,69 @@
+---
+name: Cloud Backport Tag
+
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


christian-byrne:
~~Hold off on this for a moment. I want to see if we can avoid needing two different RC branches. After deciding, we can pick this back up.~~ okay let's proceed with this

christian-byrne (approved):

PR #6897, opened 2025-11-24
[Draft] Investigate queue grouping anomaly (no code changes)

benceruleanlu:
Reverted prior attempts to change queue grouping/order so we can align with existing queueIndex semantics and investigate the cloud-specific ordering issue without altering behavior.

Status: branch now matches main for queue overlay logic; no functional changes active. Pending further investigation into why "Today" groups land mid-stack in cloud despite queueIndex alignment.

Tests (prior to revert):
- pnpm typecheck
- pnpm lint:fix

Next steps: inspect cloud queue/history ordering and timestamps to reproduce the observed grouping misplacement.

### DIFF
--- a/src/composables/queue/useJobList.ts
+++ b/src/composables/queue/useJobList.ts
@@ -87,6 +87,30 @@ type TaskWithState = {
   state: JobState
 }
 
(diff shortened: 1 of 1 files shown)


benceruleanlu:
This will probably be dropped

PR #6898, opened 2025-11-25
Cleanup: Consistent use of Nodes 2.0 in user facing strings.

DrJKL:
https://github.com/Comfy-Org/ComfyUI_frontend/issues/6888

## Summary

Did a quick pass to find user facing strings. Not sure if this gets them all.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6898-Cleanup-Consistent-use-of-Nodes-2-0-in-user-facing-strings-2b66d73d36508124aee2f5f8373a4b60) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/topbar/TryVueNodeBanner.vue
+++ b/src/components/topbar/TryVueNodeBanner.vue
@@ -63,7 +63,7 @@ const handleTryItOut = async (): Promise<void> => {
   try {
     await settingStore.set('Comfy.VueNodes.Enabled', true)
   } catch (error) {
-    console.error('Failed to enable Vue nodes:', error)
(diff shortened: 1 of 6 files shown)


AustinMroz (dismissed):
LGTM. Should we also hit the tooltip for the AutoScaleLayout setting?
Currently `Automatically scale node positions when switching to Vue rendering to prevent overlap`

DrJKL:
> LGTM. Should we also hit the tooltip for the AutoScaleLayout setting? Currently `Automatically scale node positions when switching to Vue rendering to prevent overlap`

Yes. Yes we should. 💪🏻

PR #6904, opened 2025-11-25
Feat: Babel plugin for Vite define replacements in Playwright

### LINKED ISSUE #10981: fix: fresh SubgraphNode instances leak promoted widget values through shared inner-widget state
## Problem

Multiple SubgraphNode instances sharing the same subgraph blueprint leak promoted widget values to each other when the value is set through the promoted view on one instance but never explicitly set on siblings.

**Root cause**: The `PromotedWidgetView` value setter writes to both `_instanceWidgetValues` (per-instance) AND shared state (widget state store, `resolveAtHost().widget.value`). The getter checks `_instanceWidgetValues` first, but for fresh instances (never explicitly set), it falls through to shared state — which was polluted by a sibling's setter.

This affects both display (`value` getter) and execution (`serializeValue` falls through to `this.value`).

## Reproduction

1. Create two SubgraphNode instances from the same blueprint
2. Edit a promoted widget on instance A
3. Read the same widget on instance B → returns A's value instead of the default

## Investigation (from PR #10849)

Three approaches were tried:

1. **Seed `_instanceWidgetValues` at creation** — breaks nested subgraph propagation (direct inner widget changes like `comboWidget.value = 'b'` are shadowed by the seed)
2. **Remove shared state writes from setter** — breaks 12 tests (linked widget cascading, three-level nested propagation, rendering sync)
3. **Skip widget state store in getter** — setter also writes to `resolveAtHost().widget.value`, so the fallback still picks up polluted values

## Proposed solutions

- **Inner widget observer**: Add a change listener on the concrete inner widget so `_instanceWidgetValues` stays in sync with direct mutations. This lets the getter use only the per-instance map.
- **Sole source of truth**: Make `_instanceWidgetValues` the exclusive read path for the getter. Move all shared state syncing (for rendering, linked widgets) to a separate concern that runs when an instance becomes the active/focused one.

Both require architectural changes to the value propagation flow.

## Related

- PR #10849 — per-instance widget values (configure/serialize path)
- Issue #10146 — original shared-value collision bug

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-10981-fix-fresh-SubgraphNode-instances-leak-promoted-widget-values-through-shared-inner-wi-33d6d73d365081b0af4ae40d68c5c228) by [Unito](https://www.unito.io)

snomiao:
## Summary

Implements Babel plugins to handle Vite define replacements during Playwright's compilation for i18n collection tests. This resolves ReferenceErrors caused by unprocessed compile-time constants like `__DISTRIBUTION__`.

Related to Notion task: https://www.notion.so/comfy-org/Implement-Babel-plugin-for-Vite-define-replacements-in-Playwright-2b56d73d365081d5bb63e02712912b17

## Problem

Playwright's Babel compilation does not run Vite's `define` replacements, causing reference errors during i18n collection. The `__DISTRIBUTION__` reference from `coreMenuCommands.ts` (via `src/platform/distribution/types.ts`) remains raw and causes failures.

See failing CI: https://github.com/Comfy-Org/ComfyUI_frontend/actions/runs/19624786760/job/56191480523

## Solution

Following the approach from PR #5515, this PR adds:

### 1. **babel-plugin-vite-define.cjs**
- Replaces Vite define constants (`__DISTRIBUTION__`, `__COMFYUI_FRONTEND_VERSION__`, etc.) during Babel transformation
- Uses default values suitable for Playwright test environment
- Handles various value types (strings, booleans, numbers)

### 2. **babel-plugin-inject-globals.cjs**
- Automatically injects browser globals setup import into i18n collection test files
- Pattern-matches files starting with `collect-i18n-`

### 3. **setup-i18n-globals.mjs**
- Provides JSDOM-based browser environment for Node.js execution
- Sets up `window`, `document`, `navigator`, and other browser globals
- Required because the codebase references browser globals that don't exist in Node.js

### 4. **Updated playwright.i18n.config.ts**
- Configures Babel plugins for Playwright test compilation
- Includes module resolver for `@` alias support
- TypeScript transformation with declare fields support

## Changes

- ✨ Add `babel-plugin-vite-define.cjs` to replace Vite define constants
- ✨ Add `babel-plugin-inject-globals.cjs` to inject browser globals setup  
- ✨ Add `setup-i18n-globals.mjs` for JSDOM-based browser environment
- 🔧 Update `playwright.i18n.config.ts` with Babel plugin configuration
- 🔧 Update `knip.config.ts` to ignore babel-plugin-module-resolver
- 📦 Install `babel-plugin-module-resolver` for @ alias support

## Test Plan

- [ ] Verify i18n collection tests run without ReferenceErrors
- [ ] Confirm `__DISTRIBUTION__` and other defines are properly replaced
- [ ] Check that browser globals are available during test execution
- [ ] Ensure CI passes for i18n collection workflow

## Notes

- Implementation follows PR #5515's approach but adapted for current codebase
- Babel plugins run during Playwright's test compilation phase
- Default define values are set for test environment (e.g., `__DISTRIBUTION__ = "localhost"`)

Fixes #10981

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6904-Feat-Babel-plugin-for-Vite-define-replacements-in-Playwright-2b66d73d365081a39e4dd46273f64499) by [Unito](https://www.unito.io)


snomiao:
this PR also solves https://github.com/Comfy-Org/ComfyUI_frontend/pull/6943#issuecomment-3587447241

DrJKL:
Converting to draft as a backup.

snomiao:
closed, can open later if needed :D @DrJKL

PR #6906, opened 2025-11-25
[bugfix] Add vite-define shim for Playwright i18n collection

snomiao:
## Summary

Fixes the `ReferenceError: __DISTRIBUTION__ is not defined` error when running i18n collection tests.

## Problem

PR #6879 added conditional menu commands based on distribution (hiding memory unload commands in cloud). This introduced a dependency on `isCloud` which uses the `__DISTRIBUTION__` Vite define variable in `coreMenuCommands.ts`.

When Playwright's test runner imports this file during i18n collection, it fails because Vite define variables are only replaced during Vite's build/dev process, not during Playwright's TypeScript compilation.

## Solution

Created a simple shim (`scripts/vite-define-shim.ts`) that:
1. Defines all Vite define variables as global constants with default values
2. Provides a minimal `window` shim for Node environment
3. Is imported at the top of `collect-i18n-general.ts` before any code that uses these variables

This approach is simpler than:
- Creating a custom Babel plugin (attempted in this PR, see commit history)
- Using `ctViteConfig` (only works for component testing, not regular Playwright tests)
- Post-build regex replacement (fragile and error-prone)

## Test Plan

Run `pnpm collect-i18n` and verify:
- ✅ No more `ReferenceError: __DISTRIBUTION__ is not defined`
- ✅ No more `ReferenceError: window is not defined`
- ⏱️ Tests may timeout if dev server is not running on port 5173, but that's a separate issue

## Related

- Fixes issue introduced by PR #6879
- Related to Notion task: https://www.notion.so/comfy-org/Implement-Babel-plugin-for-Vite-define-replacements-in-Playwright-2b56d73d365081d5bb63e02712912b17

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6906-bugfix-Add-vite-define-shim-for-Playwright-i18n-collection-2b66d73d36508182b4d6d69b88ae2771) by [Unito](https://www.unito.io)


snomiao:
this PR also solves https://github.com/Comfy-Org/ComfyUI_frontend/pull/6943#issuecomment-3587447241

DrJKL (approved):

PR #6925, opened 2025-11-25
Reduce width of run button dock zone

AustinMroz:
Tiny PR to make the docking area line up with the new size of the actionbar after #6723
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6925-Reduce-width-of-run-button-dock-zone-2b66d73d36508137aa0cc18178172264) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/actionbar/ComfyActionbar.vue
+++ b/src/components/actionbar/ComfyActionbar.vue
@@ -257,7 +257,7 @@ watch(isDragging, (dragging) => {
 })
 const actionbarClass = computed(() =>
   cn(
-    'w-[265px] border-dashed border-blue-500 opacity-80',
+    'w-[200px] border-dashed border-blue-500 opacity-80',
     'm-1.5 flex items-center justify-center self-stretch',
     'rounded-md before:w-50 before:-ml-50 before:h-full',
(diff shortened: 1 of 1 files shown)


src/components/actionbar/ComfyActionbar.vue
```diff
@@ -257,7 +257,7 @@ watch(isDragging, (dragging) => {
 })
 const actionbarClass = computed(() =>
   cn(
-    'w-[265px] border-dashed border-blue-500 opacity-80',
+    'w-[200px] border-dashed border-blue-500 opacity-80',
```
DrJKL, line 260: Any way we can tie them together?

DrJKL (approved):

src/components/actionbar/ComfyActionbar.vue
DrJKL: Same with the height, the docking area looks a little short.

AustinMroz: I very much would like to, but my initial tests have not been promising. Probably best to revisit later.

christian-byrne (approved):

PR #6930, opened 2025-11-25
Fix job details popover sticking after cancel/delete

### LINKED ISSUE #6907: Job Details popup gets stuck open
The Job Details popup gets stuck open and permenantly displays information that the user didn't request.
Untill it's annoyingly hovered over a second time.

https://github.com/user-attachments/assets/942b442b-649a-4ffe-a0c2-d56b0f28c9d6

### ComfyUI Frontend Version

v1.33.8

### Browser

Chrome/Chromium

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6907-Job-Details-popup-gets-stuck-open-2b66d73d365081489157cdc422c82ee3) by [Unito](https://www.unito.io)

benceruleanlu:
## Summary
- close the job details popover when its job disappears or timers fire after list changes, and clear hover timers on unmount
- emit an explicit details-leave on cancel/delete clicks so the popover closes even if hover-out never fires

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/6907

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6930-Fix-job-details-popover-sticking-after-cancel-delete-2b66d73d365081dc990ae87d01455bad) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/queue/job/JobGroupsList.vue
+++ b/src/components/queue/job/JobGroupsList.vue
@@ -36,13 +36,12 @@
 </template>
 
 <script setup lang="ts">
-import { computed, onBeforeUnmount, ref, watch } from 'vue'
+import { onBeforeUnmount, ref, watch } from 'vue'
 
 import QueueJobItem from '@/components/queue/job/QueueJobItem.vue'
 import type { JobGroup, JobListItem } from '@/composables/queue/useJobList'
 
 const props = defineProps<{ displayedJobGroups: JobGroup[] }>()
-const displayedJobGroups = computed(() => props.displayedJobGroups)
 
 const emit = defineEmits<{
   (e: 'cancelItem', item: JobListItem): void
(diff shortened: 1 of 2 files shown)


benceruleanlu:
[chromium] › browser_tests/tests/litegraphEvent.spec.ts:30:3 › Canvas Event › Emit litegraph:canvas empty-double-click 
Is timing out, pretty sure it's unrelated.

src/components/queue/job/JobGroupsList.vue
```diff
@@ -89,4 +90,23 @@ const onDetailsLeave = (jobId: string) => {
...
+    clearHideTimer()
+    clearShowTimer()
+    activeDetailsId.value = null
+  }
+})
+
+onBeforeUnmount(() => {
+  clearHideTimer()
+  clearShowTimer()
+})
```
christian-byrne: How expensive is this with a long history?

src/components/queue/job/JobGroupsList.vue
```diff
@@ -36,12 +36,13 @@
...
 <script setup lang="ts">
-import { ref } from 'vue'
+import { computed, onBeforeUnmount, ref, watch } from 'vue'
 
 import QueueJobItem from '@/components/queue/job/QueueJobItem.vue'
 import type { JobGroup, JobListItem } from '@/composables/queue/useJobList'
 
-defineProps<{ displayedJobGroups: JobGroup[] }>()
+const props = defineProps<{ displayedJobGroups: JobGroup[] }>()
+const displayedJobGroups = computed(() => props.displayedJobGroups)
```
DrJKL: Props should be [reactive by default](https://vuejs.org/guide/components/props.html#reactive-props-destructure).

src/components/queue/job/JobGroupsList.vue
benceruleanlu: We have a maxhistory of 64, if that gets removed, then it might be bad after a few thousand? Technically speaking this can be removed and each case handled on its own.

christian-byrne (dismissed):

src/components/queue/job/JobGroupsList.vue
christian-byrne: We do plan to allow scrolling through history and not removing the tail and the maxhistory is configurable right? This isn't really blocking though

benceruleanlu:
I see the replies and will get to the fixes soon

it's technically not blocking, yes, but like, it's probably going to get lost after someone merges

but of course now that I say that explicitly it'll never get lost and haunt me until I fix it

src/components/queue/job/JobGroupsList.vue
benceruleanlu: fixed

src/components/queue/job/JobGroupsList.vue
benceruleanlu: made a touch more performant, we can change later if its an issue, this isn't per-frame

benceruleanlu:
@codex review

christian-byrne (approved):

PR #6952, opened 2025-11-26
feat: right side panel

LittleSound:
CleanShot 2025-11-26 at 16 01 15

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6952-feat-right-side-panel-2b76d73d36508112b121c283a479f42a) by [Unito](https://www.unito.io)


src/components/rightSidePanel/layout/SidePanelSearch.vue
```diff
@@ -0,0 +1,79 @@
...
+  }>(),
+  {
+    searcher: async () => {}
+  }
+)
+
+const searchQuery = defineModel<string>({ default: '' })
+
+const isQuerying = ref(false)
+const debouncedSearchQuery = refDebounced(searchQuery, 700, {
```
AustinMroz, line 23:
```suggestion
const debouncedSearchQuery = refDebounced(searchQuery, 100, {
```
700 ms feels super unresponsive here. Can we lower it?

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -0,0 +1,69 @@
...
+      <slot name="label">
+        {{ label ?? $t('rightSidePanel.inputs') }}
+      </slot>
+    </template>
+
+    <div class="space-y-4 rounded-lg bg-interface-surface px-4">
+      <div
+        v-for="({ widget, node }, index) in widgets"
+        :key="`widget-${index}-${widget.name}`"
+        class="widget-item flex flex-col gap-1.5"
```
AustinMroz:
```suggestion
    <div class="space-y-4 rounded-lg bg-interface-surface px-4
      grid grid-cols-[minmax(80px,max-content)_minmax(125px,auto)]">
      <div
        v-for="({ widget, node }, index) in widgets"
        :key="`widget-${index}-${widget.name}`"
        class="widget-item gap-1.5 group col-span-full grid grid-cols-subgrid items-stretch"
```

Looks like this PR got shoved a bit by the grid widgets change. I don't quite match Alex's wizardy, but this is enough to make it usable

AustinMroz (commented):
Nice!

silveroxides:
What about the right side positioning of the sidebar? This will obviously cause massive inconveniences for anyone that prefer to have the sidebar there, me included. There is already issues with devs not considering the right side positioned sidebar as is so this would only increase the complexity of making it usable.
Also a side menu like this one does not really fit the concept of a node graph editor which is that ComfyUI frontend is supposed to be.
It is better relegated to a separate overlay style UI similar to SwarmUI if that is the kind of simplicity and screen space wasting feature users want to have.

christian-byrne (approved):
LGTM!

PR #6963, opened 2025-11-26
feat: propagate errors up subgraphs and show slot errors in subgraphs

christian-byrne:
https://github.com/user-attachments/assets/6531879d-a8a2-420a-aaca-ee329386dd1a

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6963-feat-propagate-errors-up-subgraphs-and-show-slot-errors-in-subgraphs-2b76d73d3650813e8391fac0a5e6dc9b) by [Unito](https://www.unito.io)


src/stores/executionStore.ts
```diff
@@ -552,21 +568,17 @@ export const useExecutionStore = defineStore('execution', () => {
...
     uniqueExecutingNodeIdStrings,
     // Raw executing progress data for backward compatibility in ComfyApp.
     _executingNodeProgress,
     // NodeLocatorId conversion helpers
     executionIdToNodeLocatorId,
     nodeLocatorIdToExecutionId,
-    promptIdToWorkflowId
+    // Node error lookup helpers
+    getNodeErrors,
+    slotHasError
```
DrJKL, line 665: Are these just exposed for the test?
If so, is there a way to test the behavior through the public API?

DrJKL: https://testing.googleblog.com/2015/01/testing-on-toilet-prefer-testing-public.html

https://shoulditestprivatemethods.com/

DrJKL (dismissed):
Testing internal logic piece can (and should) be fixed later.

PR #6965, opened 2025-11-26
fix: don't use registry when only checking for presence of missing nodes

christian-byrne:
Changes the client-side validation of whether a graph has missing nodes to use a simpler check (is the node name in the node defs from `/object_info`), rather than using the Comfy Node Registry API (api.comfy.org). The Registry API is still needed to get full metadata/info about missing nodes, but that can be deferred until the user actually needs that info.

This also fixes an issue where a node you coded yourself locally which is neither a core node nor a node in the registry would be considered "missing."

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6965-fix-don-t-use-registry-when-only-checking-for-presence-of-missing-nodes-2b76d73d365081e7b8fbcb3f2a1c4502) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/breadcrumb/SubgraphBreadcrumbItem.vue
+++ b/src/components/breadcrumb/SubgraphBreadcrumbItem.vue
@@ -64,11 +64,13 @@ import {
   ComfyWorkflow,
   useWorkflowStore
 } from '@/platform/workflow/management/stores/workflowStore'
+import { app } from '@/scripts/app'
 import { useDialogService } from '@/services/dialogService'
 import { useCommandStore } from '@/stores/commandStore'
+import { useNodeDefStore } from '@/stores/nodeDefStore'
 import { useSubgraphNavigationStore } from '@/stores/subgraphNavigationStore'
 import { appendJsonExt } from '@/utils/formatUtil'
-import { useMissingNodes } from '@/workbench/extensions/manager/composables/nodePack/useMissingNodes'
+import { graphHasMissingNodes } from '@/workbench/extensions/manager/utils/graphHasMissingNodes'
 
 interface Props {
   item: MenuItem
@@ -79,7 +81,10 @@ const props = withDefaults(defineProps<Props>(), {
   isActive: false
 })
 
-const { hasMissingNodes } = useMissingNodes()
+const nodeDefStore = useNodeDefStore()
+const hasMissingNodes = computed(() =>
+  graphHasMissingNodes(app.graph, nodeDefStore.nodeDefsByName)
+)
 
 const { t } = useI18n()
 const menu = ref<InstanceType<typeof Menu> & MenuState>()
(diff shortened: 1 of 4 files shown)


AustinMroz (commented):
LGTM

src/components/breadcrumb/SubgraphBreadcrumbItem.vue
```diff
@@ -79,7 +81,10 @@ const props = withDefaults(defineProps<Props>(), {
   isActive: false
 })
 
-const { hasMissingNodes } = useMissingNodes()
+const nodeDefStore = useNodeDefStore()
+const hasMissingNodes = computed(() =>
+  graphHasMissingNodes(app.graph, nodeDefStore.nodeDefsByName)
```
DrJKL, line 86: Seeing this in multiple places makes me think it could be readonly state within NodeDefStore.

src/workbench/extensions/manager/utils/graphHasMissingNodes.ts
```diff
@@ -0,0 +1,36 @@
...
+
+type NodeDefLookup = Record<string, unknown>
+
+const isNodeMissingDefinition = (
+  node: LGraphNode,
+  nodeDefsByName: NodeDefLookup
+) => {
+  const nodeName = node?.type
+  if (!nodeName) return false
+  return !nodeDefsByName[nodeName]
```
DrJKL, line 20: Very nit: (also double check my logic here)
```suggestion
  return !(nodeName && nodeDefsByName[nodeName])
```

src/components/breadcrumb/SubgraphBreadcrumbItem.vue
christian-byrne: Could you elaborate a bit?

tests-ui/tests/workbench/extensions/manager/utils/graphHasMissingNodes.test.ts
```diff
@@ -0,0 +1,83 @@
...
+} from '@/lib/litegraph/src/litegraph'
+import {
+  collectMissingNodes,
+  graphHasMissingNodes
+} from '@/workbench/extensions/manager/utils/graphHasMissingNodes'
+
+type NodeDefs = Record<string, unknown>
+
+const createGraph = (nodes: LGraphNode[] = []): LGraph => {
+  return { nodes } as unknown as LGraph
```
DrJKL: Nit: Doing a partial to full adds a bit of extra validation when the real type changes, especially for removed fields.
```suggestion
  return { nodes } as Partial<LGraph> as LGraph
```

tests-ui/tests/workbench/extensions/manager/utils/graphHasMissingNodes.test.ts
```diff
@@ -0,0 +1,83 @@
...
+  LGraph,
+  LGraphNode,
+  Subgraph
+} from '@/lib/litegraph/src/litegraph'
+import {
+  collectMissingNodes,
+  graphHasMissingNodes
+} from '@/workbench/extensions/manager/utils/graphHasMissingNodes'
+
+type NodeDefs = Record<string, unknown>
```
DrJKL: Do we have a real type somewhere to reference here?

DrJKL (dismissed):

PR #6985, opened 2025-11-27
Support "control after generate" in vue

AustinMroz:
Continuation of #6034 with
- Updated synchronization for seed
- Properly truncates the displayed widget value for the button
- Synchronizes control after generate state with litegraph and allows for serialization

Several issues from original PR have not (yet) been addressed, but are likely better moved to future PR
- fix step value being 10 (legacy system)
- ensure it works with COMBO (Fixed in #7095)
- ensure it works with FLOAT (Fixed in #7095)
- either implement or remove the config button functionality - think it should open settings?

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6985-Support-control-after-generate-in-vue-2b86d73d365081d8b01ce489d887ff00) by [Unito](https://www.unito.io)


lostdisc:
>either implement or remove the config button functionality - think it should open settings?

Personally as a random SDXL user, I do think the "control _before_ generate" setting should be made a lot more obvious (or better yet, the default), since it's more intuitive for users wanting to quickly retry a seed.  In before-mode, when I get a promising-but-not-perfect result from a random seed, I can just switch to "fixed" mode, tweak my prompt, and rerun.

In after-mode, by contrast, the seed has already changed, and I have to retrieve the old value by digging into the image's metadata in a separate program and manually copy-pasting it back into the seed field.  I can't even copy seeds from the job history, only a useless* job ID.

*: I'm sure it's useful for something, just like control-after-generate must be useful for something, but I can't fathom why after-mode is the default.  So I'm glad there's a before/after setting, but I wish its existence and lifesaving practical effect were made more obvious.

christian-byrne (dismissed):

christian-byrne:
@AustinMroz I can fix these merge conflicts right now, let me know if you are currently also doing this and I'll avoid pushing. If not, please ignore.

christian-byrne:
> > either implement or remove the config button functionality - think it should open settings?
> 
> Personally as a random SDXL user, I do think the "control _before_ generate" setting should be made a lot more obvious (or better yet, the default), since it's more intuitive for users wanting to quickly retry a seed. In before-mode, when I get a promising-but-not-perfect result from a random seed, I can just switch to "fixed" mode, tweak my prompt, and rerun.
> 
> In after-mode, by contrast, the seed has already changed, and I have to retrieve the old value by digging into the image's metadata in a separate program and manually copy-pasting it back into the seed field. I can't even copy seeds from the job history, only a useless* job ID.
> 
> *: I'm sure it's useful for something, just like control-after-generate must be useful for something, but I can't fathom why after-mode is the default. So I'm glad there's a before/after setting, but I wish its existence and lifesaving practical effect were made more obvious.

This is great feedback, what would you say would be most intuitive or obvious to you?

lostdisc:
I think "control before generate" would be the more intuitive default.  With "control after generate", it's confusing for the seed to change at the end of a gen, since it no longer corresponds with the output on display, and there's no way within ComfyUI to retrieve the old seed for a retry with different prompts/settings, which is a common use case.

The one downside I can think of for "control before generate" is that increment/decrement would also change the starting seed for the first gen, which would be unexpected.  Maybe it can skip changing the seed when the seed value has just been manually changed by the user.  (This exception would also make sense for randomize mode.)

src/scripts/utils.ts
```diff
@@ -53,6 +51,21 @@ export async function addStylesheet(
 
 export { downloadBlob } from '@/base/common/downloadUtil'
 
+if (typeof window !== 'undefined') {
```
christian-byrne: Cannot remember exactly why this was necessary

christian-byrne:
Cleaning a few things up

src/scripts/utils.ts
christian-byrne: Removed, we will see

DrJKL:
> I think "control before generate" would be the more intuitive default. With "control after generate", it's confusing for the seed to change at the end of a gen, since it no longer corresponds with the output on display, and there's no way within ComfyUI to retrieve the old seed for a retry with different prompts/settings, which is a common use case.
> 
> The one downside I can think of for "control before generate" is that increment/decrement would also change the starting seed for the first gen, which would be unexpected. Maybe it can skip changing the seed when the seed value has just been manually changed by the user. (This exception would also make sense for randomize mode.)

There is an option in the settings to change it to control before generate.
image

lostdisc:
> There is an option in the settings to change it to control before generate.

I know; that's how I can articulate the concept now.  But until recently, I was just suffering through the default mode with no idea it could be changed, until someone else pointed out this setting existed and did something useful.  I bet many users follow the same path, or never find out.

I think putting a toggle right in the KSampler node would help a lot with discoverability, but since it's a shared setting, maybe it's hard to keep the UI state synced across multiple nodes?

Barring that, maybe the simplest improvement would be changing the terminology from "control before/after generate" to "**update seed** before/after generate", which makes the practical effect much clearer.  Similarly, the setting could be relabeled from "widget control mode: before/after" to "update seed fields: before generate / after generate".

(Or is the current wording a vague "control" because it also affects things other than seed RNGs?)

DrJKL:
@coderabbitai full review

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -82,6 +87,29 @@ export interface GraphNodeManager {
   cleanup(): void
 }
 
+function validateControlWidgetValue(val: unknown): ControlWidgetOptions {
+  //TODO: Is there a way to do this without repeating?
```
DrJKL: I'd do something like I did with the expandables.
```ts
const CONTROL_OPTIONS = [ 'fixed', 'increment', 'decrement', 'randomize' ] as const
export function validateControlOption(maybeOption: string): ControlOption {
 if (CONTROL_OPTIONS.includes(maybeOption as ControlOption)) return maybeOption
 return 'randomize'
}
```
You could also use a type guard in there.

src/locales/en/main.json
```diff
@@ -1910,6 +1910,22 @@
       "placeholderVideo": "Select video...",
       "placeholderModel": "Select model...",
       "placeholderUnknown": "Select media..."
+    },
+    "numberControl": {
+      "controlHeaderBefore": "Automatically update the value",
+      "controlHeaderAfter": "AFTER",
+      "controlHeaderBefore2": "BEFORE", 
+      "controlHeaderEnd": "running the workflow:",
```
DrJKL: Optional: You could nest these
```suggestion
      "controlHeader": {
        "Before": "Automatically update the value",
        "After": "AFTER",
        "Before2": "BEFORE", 
        "End": "running the workflow:",
      }
```

src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberSlider.vue
```diff
@@ -75,25 +76,7 @@ const precision = computed(() => {
...
-
-  if (precision.value === 0) {
-    return 1
-  }
-
-  // For precision > 0, step = 1 / (10^precision)
-  // precision 1 → 0.1, precision 2 → 0.01, etc.
-  return 1 / Math.pow(10, precision.value)
-})
+const stepValue = useNumberStepCalculation(widget.options, precision, true)
```
DrJKL, line 75: I think this could be a utility function instead of a composable for now.

src/renderer/extensions/vueNodes/widgets/composables/useNumberStepCalculation.ts
```diff
@@ -0,0 +1,36 @@
...
+  precision?: number
+}
+
+/**
+ * Shared composable for calculating step values in number input widgets
+ * Handles both explicit step2 values and precision-derived steps
+ */
+export function useNumberStepCalculation(
+  options: NumberWidgetOptions | undefined,
+  precision: Ref<number | undefined>,
```
DrJKL: https://vuejs.org/api/reactivity-utilities.html#tovalue

This is a good place for `MaybeRefOrGetter`

src/renderer/extensions/vueNodes/widgets/composables/useStepperControl.ts
```diff
@@ -0,0 +1,117 @@
...
+
+import { numberControlRegistry } from '../services/NumberControlRegistry'
+
+export enum NumberControlMode {
+  FIXED = 'fixed',
+  INCREMENT = 'increment',
+  DECREMENT = 'decrement',
+  RANDOMIZE = 'randomize',
+  LINK_TO_GLOBAL = 'linkToGlobal'
+}
```
DrJKL, line 14:
```suggestion
export const NumberControlMode = {
  FIXED: 'fixed',
  INCREMENT: 'increment',
  DECREMENT: 'decrement',
  RANDOMIZE: 'randomize',
  LINK_TO_GLOBAL: 'linkToGlobal'
} as const
```

That thing we talked about

src/renderer/extensions/vueNodes/widgets/components/NumberControlPopover.vue
```diff
@@ -0,0 +1,171 @@
...
+
+import { NumberControlMode } from '../composables/useStepperControl'
+
+type ControlOption = {
+  description: string
+  icon?: string
+  mode: NumberControlMode
+  text?: string
+  title: string
+}
```
DrJKL, line 18: Alphabetical is great, but required before optional please <3

src/renderer/extensions/vueNodes/widgets/composables/useStepperControl.ts
```diff
@@ -0,0 +1,117 @@
...
+  options: StepperControlOptions,
+  defaultValue?: ControlOptions
+) {
+  const controlMode = ref<NumberControlMode>(convertToEnum(defaultValue))
+  const controlId = Symbol('numberControl')
+  const globalSeedStore = useGlobalSeedStore()
+
+  const applyControl = () => {
+    const { min = 0, max = 1000000, step2, step = 1, onChange } = options
+    const safeMax = Math.min(1125899906842624, max)
```
DrJKL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER maybe?

src/renderer/extensions/vueNodes/widgets/composables/useStepperControl.ts
```diff
@@ -0,0 +1,117 @@
...
+    if (onChange) {
+      onChange(newValue)
+    } else {
+      modelValue.value = newValue
+    }
+  }
+
+  // Register with singleton registry
+  onMounted(() => {
+    numberControlRegistry.register(controlId, applyControl)
```
DrJKL, line 97: Double check to see if you need to be DOM dependent here.

src/types/simplifiedWidget.ts
```diff
@@ -15,6 +15,25 @@ export type WidgetValue =
...
   | File[]
 
+const CONTROL_OPTIONS = [
+  'fixed',
+  'increment',
+  'decrement',
+  'randomize'
+] as const
+export type ControlOptions = (typeof CONTROL_OPTIONS)[number]
+
```
DrJKL, line 25:
```suggestion
function isValidControlOption(optionMaybe: string): optionMaybe is ControlOption {
  return CONTROL_OPTIONS.includes(optionMaybe as ControlOptions)
}
```

src/types/simplifiedWidget.ts
```diff
@@ -15,6 +15,25 @@ export type WidgetValue =
...
+export function validateControlOption(val: unknown): ControlOptions {
```
DrJKL: Also, this is technically normalizing, right?

tests-ui/tests/renderer/extensions/vueNodes/widgets/composables/useStepperControl.test.ts
```diff
@@ -0,0 +1,272 @@
...
+      getControlCount: vi.fn(() => 0),
+      clear: vi.fn()
+    },
+    executeNumberControls: vi.fn()
+  })
+)
+
+describe('useStepperControl', () => {
+  beforeEach(() => {
+    setActivePinia(createPinia())
```
DrJKL: createTestingPinia please

src/stores/globalSeedStore.ts
DrJKL, line 1: I think this piece could be pruned for now, it's a deserving feature and should be treated with such deference.

src/renderer/extensions/vueNodes/widgets/composables/useStepperControl.ts
AustinMroz: Followup: sorta. It's being used as a convenient way to ensure it gets unregistered at some point. Since I'm proposing removal in a future PR, it's probably not worth rewriting the implementation here.

christian-byrne (dismissed):

christian-byrne:
Confirmed all test failures are from the seed control, re-generating baseline snapshots now!

christian-byrne (approved):

PR #6997, opened 2025-11-27
[feat] Add ZIP download support for multiple assets

viva-jinyi:
## Summary
When downloading multiple selected assets, they are now bundled into a single ZIP file instead of downloading each file individually.

## Changes
- Use `client-zip` library to generate ZIP files in the browser
- Use `Promise.allSettled` to include remaining files in ZIP even if some fail
- Automatically add numbers to duplicate filenames
- Notify users with toast messages based on download status
- Auto-generate filename with timestamp to prevent filename collisions

### Modified Files
- `src/platform/assets/composables/useMediaAssetActions.ts`: Implement ZIP download logic
- `src/locales/en/main.json`: Add new i18n strings
- `package.json`: Add `client-zip` dependency

[screen-capture.webm](https://github.com/user-attachments/assets/a57ddfc2-1525-466e-b913-d1d3a38bfdaa)

Co-Authored-By: Claude <noreply@anthropic.com>

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6997-feat-Add-ZIP-download-support-for-multiple-assets-2b86d73d36508101af82f5d446cc066a) by [Unito](https://www.unito.io)


christian-byrne (commented):
This is a good idea but what are pros/cons of doing this on the client vs. doing this on the server?

viva-jinyi:
@christian-byrne I think for large files, backend processing would be better. But for downloading multiple images like this, client-side processing seems more appropriate. In my opinion, it would be good to implement it on the frontend for now, and then switch to backend processing later when the backend provides API support for it.

christian-byrne (commented):
The tentative current plan we have discussed in various other threads is something like:

1. For downloading batches, we can temporarily just use the multiple download feature of the browser, which downloads sequentially one by one - rather than trying to parallelize and zip on the client
2. Sometime in next 2-4 weeks, the backend will switch to supporting the zip endpoint, which we will implement on cloud first
3. After item 2 (above) is done, we will mirror the route in the backend (for downloading zipped assets)
4. After item 3 (above) is done, we will make the bulk download operate the same on cloud and local
5. Somewhat separately, we want to make sure that all the items inside the stacked outputs are downloaded, rather than just the top level visible output (for jobs that created mutliple outputs)

I think items 3 and 4 are not really necessary in most cases since the user is presumed to be using a local backend anyway; however, I think we should do it for simplicity and also because some people do remote into their computer to use ComfyUI.

PR #7006, opened 2025-11-27
[fix] Re-encode cloud-subscription video to VP9 for Safari compatibility

Myestery:
The original video was encoded with AV1 codec which Safari does not
support, causing "Plug-in handled load" errors. Re-encoded to VP9
which is supported across all major browsers including Safari 14.1+.

 Summary

  - Re-encode cloud-subscription.webm from AV1 to VP9 codec for Safari browser compatibility

  Problem

  Safari does not support AV1 codec in webm containers, causing the cloud subscription video to
  fail with "Plug-in handled load" error.

  Solution

  Re-encoded the video using VP9 codec (libvpx-vp9) which is supported by all major browsers
  including Safari 14.1+.

  Test plan

  - Verify video plays correctly in Safari
  - Verify video plays correctly in Chrome/Firefox/Edge

[cloud-subscription.webm](https://github.com/user-attachments/assets/ab2dcaa1-f3b3-47c9-84d0-9fc8ea8c6f17)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7006-fix-Re-encode-cloud-subscription-video-to-VP9-for-Safari-compatibility-2b86d73d365081ad9262f97fc1ebb5ee) by [Unito](https://www.unito.io)


christian-byrne (approved):
Nice, LGTM

PR #7021, opened 2025-11-28
Handle HTML fallbacks for node help and skip blueprint docs

benceruleanlu:
explicitly prevents subgraphs from making an api call since they don't have docs, this was previously reliant on a non-ok resolution

also doesn't try returning anything that has contenttype of text/html to prevent the markdown renderer from crashing

## Summary
- short-circuit blueprint/subgraph nodes in help: skip doc fetch and return the node description, avoiding SPA fallback responses
- guard node help fetch against HTML/SPA fallbacks using content-type checks; treat them as missing and trigger the existing description fallback
- keep base URL logic unchanged for non-blueprint nodes

## Testing
- pnpm typecheck
- pnpm lint:fix
- pnpm test:unit

### DIFF
--- a/src/services/nodeHelpService.ts
+++ b/src/services/nodeHelpService.ts
@@ -23,6 +23,7 @@ class NodeHelpService {
     locale: string
   ): Promise<string> {
     const customNodeName = extractCustomNodeName(node.python_module)
+    let lastError: string | undefined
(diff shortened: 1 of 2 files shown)


benceruleanlu:
@codex review

christian-byrne (approved):

PR #6514, opened 2025-11-01
Drag vuenodes input

Myestery:
This pull request introduces several improvements to Vue reactivity and user experience in the graph node and widget system. The main focus is on ensuring that changes to node and widget data reliably trigger updates in Vue components, improving drag-and-drop support for nodes, and enhancing widget value handling for better compatibility and reactivity.

**Vue Reactivity Improvements:**

* In `useGraphNodeManager.ts`, node data updates now create a completely new object and add a timestamp (`_updateTs`) to force Vue's reactivity system to detect changes. Additionally, node data is re-set on the next tick to guarantee component updates. [[1]](diffhunk://#diff-f980db6f42cef913c3fe92669783b255d617e40b9ccef9a1ab9cc8e326ff1790L272-R280) [[2]](diffhunk://#diff-f980db6f42cef913c3fe92669783b255d617e40b9ccef9a1ab9cc8e326ff1790R326-R335)
* Widget value composables (`useWidgetValue` and related helpers) now accept either a direct value or a getter function for `modelValue`, and always normalize it to a getter. Watches are updated to use this getter for more reliable reactivity. [[1]](diffhunk://#diff-92dc3c8b09ab57105e400e115196aae645214f305685044f62edc3338afa0911L13-R14) [[2]](diffhunk://#diff-92dc3c8b09ab57105e400e115196aae645214f305685044f62edc3338afa0911R49-R57) [[3]](diffhunk://#diff-92dc3c8b09ab57105e400e115196aae645214f305685044f62edc3338afa0911L82-R91) [[4]](diffhunk://#diff-92dc3c8b09ab57105e400e115196aae645214f305685044f62edc3338afa0911L100-R104) [[5]](diffhunk://#diff-92dc3c8b09ab57105e400e115196aae645214f305685044f62edc3338afa0911L117-R121) [[6]](diffhunk://#diff-92dc3c8b09ab57105e400e115196aae645214f305685044f62edc3338afa0911L140-R144) [[7]](diffhunk://#diff-0c43cefa9fb524ae86541c7ca851e97a22b3fd01f95795c83273c977be77468fL47-R47)
* In `useImageUploadWidget.ts`, widget value updates now use a new array/object to ensure Vue detects the change, especially for batch uploads.

**Drag-and-Drop Support for Nodes:**

[...]

PR #7027, opened 2025-11-28
Fix: image upload widget value saved incorrectly

doommm:
## Summary

image upload widget value was saved incorrectly(didn't trigger `changeTracker`)

## Changes

- call `workflowStore.activeWorkflow?.changeTracker?.checkState()` after value setted and callback has been called.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7027-Fix-image-upload-widget-value-saved-incorrectly-2b96d73d36508104aa9ffe5d27360f98) by [Unito](https://www.unito.io)


christian-byrne:
This seems reasonable and we can merge. Could you provide reproduction steps to reproduce the original observed bug?

doommm:
> This seems reasonable and we can merge. Could you provide reproduction steps to reproduce the original observed bug?

this seems unnecessary after this PR merged 🤣(https://github.com/Comfy-Org/ComfyUI_frontend/pull/6514)

DrJKL:
> > This seems reasonable and we can merge. Could you provide reproduction steps to reproduce the original observed bug?
> 
> this seems unnecessary after this PR merged 🤣(#6514)

If you still see issues with it, definitely let us know!

PR #7053, opened 2025-11-30
fix resizing from top

christian-byrne:
## Changes

- **What**: 
- **Breaking**: 
- **Dependencies**: 

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7053-fix-resizing-from-top-2bb6d73d3650812cab29e2198da6045a) by [Unito](https://www.unito.io)


christian-byrne:
Decided against this approach. Planning to disable resizing from top on 1.32 and 1.33 then try an alternative solution from 1.34 onwards.

PR #7054, opened 2025-11-30
fix wrong 3d node def after refactor

jtydhr88:
## Summary

fix wrong node def for 3d load after refactor, to match BE change in https://github.com/comfyanonymous/ComfyUI/pull/10025

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7054-fix-wrong-3d-node-def-after-refactor-2bb6d73d365081da956cdee057f1e8b3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/locales/en/nodeDefs.json
+++ b/src/locales/en/nodeDefs.json
@@ -4799,7 +4799,7 @@
     }
   },
   "Load3D": {
-    "display_name": "Load 3D",
+    "display_name": "Load 3D & Animation",
     "inputs": {
       "model_file": {
         "name": "model_file"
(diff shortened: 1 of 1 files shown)


christian-byrne (commented):
This usually happens automatically, but the locale generation workflow is currently broken.

jtydhr88:
@christian-byrne on currect version of core, it is still missing these changes

DrJKL (approved):

DrJKL:
When do you think we'll have the locale workflow working again?

christian-byrne:
We shouldn't manually edit these files really, it may disrupt the CI/CD. Only if ths current translations are breaking the node.

jtydhr88:
Ok, @christian-byrne let me close this. But as you mentioned, current GitHub workflow is down, just let me know when it is available, I will double check this issue to fix later

PR #7061, opened 2025-11-30
add shared comfy credit conversion helpers

christian-byrne:
Introduces cents<->usd<->credit converters plus basic formatters and adds test. Lays groundwork to start converting UI components into displaying comfy credits.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7061-add-shared-comfy-credit-conversion-helpers-2bb6d73d3650810bb34fdf9bb3fc115b) by [Unito](https://www.unito.io)


tests-ui/tests/base/credits/comfyCredits.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    expect(usdToCredits(2.1)).toBeCloseTo(1)
+    expect(creditsToUsd(3.5)).toBeCloseTo(7.35)
+  })
+
+  test('formats credits and USD values using en-US locale', () => {
+    const locale = 'en-US'
+    expect(formatCredits(1234.567, undefined, locale)).toBe('1,234.57')
+    expect(formatCreditsFromCents(210, undefined, locale)).toBe('1.00')
+    expect(formatCreditsFromUsd(4.2, undefined, locale)).toBe('2.00')
+    expect(formatUsd(4.2, undefined, locale)).toBe('4.20')
```
DrJKL: Having to put undefined in here makes me think we should just have an object param, wdyt?

DrJKL: Or, we could flip the order which would also serve to match the order in `Intl.NumberFormat`

src/base/credits/comfyCredits.ts
```diff
@@ -0,0 +1,115 @@
...
+    merged.minimumFractionDigits = merged.maximumFractionDigits
+  }
+
+  return new Intl.NumberFormat(locale, merged).format(value)
+}
+
+export const COMFY_CREDIT_RATE_CENTS = 210
+export const COMFY_CREDIT_RATE_USD = COMFY_CREDIT_RATE_CENTS / 100
+
+export const usdToCents = (usd: number): number => Math.round(usd * 100)
```
AustinMroz, line 34: A little concerning that rounding is only applied to cents, but I don't think it'll be an issue.

I'm also questioning if `COMFY_CREDIT_RATE_CENTS` is backwards. A dollar is worth < 1 credit? And a balance of 1 cent should round down to 0.00 credits?

AustinMroz (dismissed):

src/base/credits/comfyCredits.ts
christian-byrne: Thanks, good catch.

christian-byrne:
Due to the deadline, I really need to merge this. Future PRs will be fixing the style to make it match the designs exactly. Please leave reviews still and I will implement changes tomorrow.

christian-byrne:
Will implement designs properly in followup PR.

PR #7085, opened 2025-12-02
Feat: Add model import button to Vue model list popover.

DrJKL:
## Summary

Adds a button to trigger the model upload on the Asset dropdown.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7085-Feat-Add-model-import-button-to-Vue-model-list-popover-2bd6d73d365081958c8ef23de55a341e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/assets/composables/useModelUpload.ts
+++ b/src/platform/assets/composables/useModelUpload.ts
@@ -1,0 +1,35 @@
+import { useFeatureFlags } from '@/composables/useFeatureFlags'
+import UploadModelDialog from '@/platform/assets/components/UploadModelDialog.vue'
+import UploadModelDialogHeader from '@/platform/assets/components/UploadModelDialogHeader.vue'
+import type { AssetItem } from '@/platform/assets/schemas/assetSchema'
+import { useDialogStore } from '@/stores/dialogStore'
+import type { UseAsyncStateReturn } from '@vueuse/core'
+import { computed } from 'vue'
+
+export function useModelUpload(
+  execute?: UseAsyncStateReturn<AssetItem[], [], true>['execute']
+) {
+  const dialogStore = useDialogStore()
+  const { flags } = useFeatureFlags()
+  const isUploadButtonEnabled = computed(() => flags.modelUploadButtonEnabled)
+
+  function showUploadDialog() {
+    dialogStore.showDialog({
+      key: 'upload-model',
+      headerComponent: UploadModelDialogHeader,
+      component: UploadModelDialog,
+      props: {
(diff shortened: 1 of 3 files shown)


src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdownMenuFilter.vue
```diff
@@ -8,10 +10,12 @@ defineProps<{
 }>()
 
 const filterSelected = defineModel<OptionId>('filterSelected')
+
+const { isUploadButtonEnabled, showUploadDialog } = useModelUpload()
 </script>
```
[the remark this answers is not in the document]
DrJKL: It's in there. Line 37

src/platform/assets/composables/useModelUpload.ts
```diff
@@ -0,0 +1,35 @@
...
+import type { AssetItem } from '@/platform/assets/schemas/assetSchema'
+import { useDialogStore } from '@/stores/dialogStore'
+import type { UseAsyncStateReturn } from '@vueuse/core'
+import { computed } from 'vue'
+
+export function useModelUpload(
+  execute?: UseAsyncStateReturn<AssetItem[], [], true>['execute']
+) {
+  const dialogStore = useDialogStore()
+  const { flags } = useFeatureFlags()
```
christian-byrne, line 13: question: Is it better for feature flags to be accessed same way everywhere or passed through via the state-threading/DI/prop-drilling type of pattern?

christian-byrne (approved):
LGTM

src/platform/assets/composables/useModelUpload.ts
DrJKL: They're universal, so they should be accessed directly. DI/Prop drilling is for things that are contextual and where the context can be derived from the DOM/component hierarchy.

christian-byrne: So in theory, this compostable shouldn't re expose a computer version of the flags?

DrJKL: In this case, the importance isn't that it's a flag, it's whether the behavior here should be exposed.
Putting the flag check into the composable means that each consumer only needs to know the value, not how to get it.

DrJKL: In this case, the importance isn't that it's a flag, it's whether the behavior here should be exposed.
Putting the flag check into the composable means that each consumer only needs to know the value, not how to get it.

PR #7102, opened 2025-12-02
Feat: Add preview as plaintext toggle for Preview As Text

DrJKL:
## Summary

Adds a toggle to conditionally render the text as Markdown.

## Also...

Fixes some type issues across our myriad Widget types. We should probably clean those up.

## Example

https://github.com/user-attachments/assets/24fed943-1e79-4ea4-a962-826b06d68761

This could be a good minimal testcase for dynamic widgets @AustinMroz

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7102-WIP-Feat-Add-preview-as-plaintext-toggle-for-Preview-As-Text-2bd6d73d3650810c8b25c84866c8875c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/previewAny.ts
+++ b/src/extensions/core/previewAny.ts
@@ -24,12 +24,41 @@ useExtensionService().registerExtension({
           app
         ).widget as DOMWidget<HTMLTextAreaElement, string>
 
-        showValueWidget.options.read_only = true
+        const showValueWidgetPlain = ComfyWidgets['STRING'](
+          this,
+          'preview',
+          ['STRING', { multiline: true }],
+          app
+        ).widget as DOMWidget<HTMLTextAreaElement, string>
 
+        const showAsPlaintextWidget = ComfyWidgets['BOOLEAN'](
+          this,
+          'previewMode',
+          [
+            'BOOLEAN',
+            { label_on: 'Markdown', label_off: 'Plaintext', default: false }
+          ],
+          app
+        )
+
+        showAsPlaintextWidget.widget.callback = (value) => {
+          showValueWidget.hidden = !value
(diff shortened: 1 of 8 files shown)


DrJKL:
I know it should be Heading not Header. Don't @ me.

src/extensions/core/previewAny.ts
```diff
@@ -39,9 +70,12 @@ useExtensionService().registerExtension({
           ? void 0
           : onExecuted.apply(this, [message])
 
-        const previewWidget = this.widgets?.find((w) => w.name === 'preview')
+        const previewWidgets =
+          this.widgets?.filter((w) => w.name === 'preview') ?? []
+
+        if (!previewWidgets) return
```
DrJKL:
```suggestion
```

AustinMroz (dismissed):
Functionality works well.

Some passing concern on the change to how `hidden` functions. There's a fairly lengthy tail of recommending custom node authors add a `widget.computeSize` to set the height to `-4 ` when hidden. (This wouldn't quite work here due to the sizing on textAreas). However, I'm reasonably confident this change is safe. I can't think of a time you'd want a widget to be hidden, but still take up space.

DrJKL:
> Functionality works well.
> 
> Some passing concern on the change to how `hidden` functions. There's a fairly lengthy tail of recommending custom node authors add a `widget.computeSize` to set the height to `-4 ` when hidden. (This wouldn't quite work here due to the sizing on textAreas). However, I'm reasonably confident this change is safe. I can't think of a time you'd want a widget to be hidden, but still take up space.

Very worth calling out. I'll keep an eye out.

AustinMroz (approved):

Issue #6993, opened 2025-11-27
Copy and paste still broken


CoreyCorza:
Barely any improvement since September and when it first broke https://github.com/Comfy-Org/ComfyUI_frontend/issues/5449
Still broken

https://github.com/user-attachments/assets/900a01ee-bba2-4139-af3f-cc1a2ff8b776

### ComfyUI Frontend Version

1.34.0

### Browser

Chrome/Chromium

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6993-Copy-and-paste-still-broken-2b86d73d36508194a7ddfe8227262b88) by [Unito](https://www.unito.io)

CoreyCorza:
Barely any improvement in the sense that it still behaves the same way as when it was first introduced. So its still the same bug I guess.
It doesn't do it _as much_ on newer versions, but still inevitable and happens numerous times per session

geroldmeisinger:
all issues related to copy+paste #6660 #6993 #5449 #6822 #6485 #5683
it's especially annoying as the new comfyui release nudges me to use the new frontend

jtydhr88:
close it for now as it should fix in https://github.com/Comfy-Org/ComfyUI_frontend/pull/7103 and https://github.com/Comfy-Org/ComfyUI_frontend/pull/7166.
Please feel free to reopen if it still exists after these changes

diGitTil:
Updating to the latest build reintroduces the copy paste bug.

Keyboard shortcut doesn’t copy at all. The context menu only succeeds if selecting one single node (copy success notification appears). Tried this in Edge, Chromium, Firefox with all extensions disabled. Windows, 32GB system RAM.

The result is consistent. I noticed a substantial increase in RAM usage when loading comfy (without running any Workflows) after updating to the most recent version of the UI. 

Example: fresh boot, Firefox without any open tabs, loading only ComfyUI, not loading any workflows… 85% RAM usage!

My theory is some type of memory hog/leak, perhaps from nodes 2.0?

Gonna need to downgrade/revert to earlier version I suppose.

diGitTil:
UPDATE after additional experimentation:

Updated all video drivers as well, and completely reinstalled comfyUI without any custom nodes whatsoever. The cut command will fail if the total number of nodes exceeds somewhere between 4 and 8, but it’s arbitrary. If a grouping node is part of the selection, it appears to be a higher consistency of failing the copy.

[...]

Issue #5449, opened 2025-09-09
v1.27.2 Copy+paste nodes broken


CoreyCorza:
### Prerequisites

- [x] I am running the latest version of ComfyUI
- [x] I have searched existing issues to make sure this isn't a duplicate
- [x] I have tested with all custom nodes disabled ([see how](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled))

### What happened?

With v1.27.2 copying and pasting nodes has become unreliable and broken.
Somehow now 'zoom dependant'.

Sometimes works zoomed in, sometimes doesn't.
Somteimes **fails to copy node**, pastes old nodes on clipboard.
Sometimes copies, **fails to paste nodes**.
Also affects copying and pasting nodes across different tabs/workflows.

---> Zooming out all the way, seems to mitagate the issue.
--> Being at close/mid zoomed-in levels, issue appears.

Its not immediately apparent in the gif but I'm rapidly pressing ctrl+c ctrl+v to copy and paste nodes and its only working about 8% of the time.
Unless I zoom all the way out.
[-- Failed to upload "brave_B7er30Tgq5.gif" ]

### Steps to Reproduce

asad

### How is this affecting you?

Feature doesn't work as expected

### ComfyUI Frontend Version

 v1.27.2

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

CoreyCorza:
Possibly related to this new implementation?
<img width="752" height="66" alt="Image" src="https://github.com/user-attachments/assets/48accf58-84a4-487b-8740-85d7b1f7ad97" />

CoreyCorza:
@dosu
Further investigation: The issue is persistant no matter what the LOD slider is set to

CoreyCorza:
@dosu
Further investigation: Happens on native nodes aswell that dont appear to have hidden widgets or slots. 
Issue persists with --disable-all-custom-nodes

CoreyCorza:
@dosu
Further investigation: Taking selection into account as you've suggested - I dont see any replicable patterns.
It still acts very unreliably.
Sometimes being zoomed all the way in will copy and paste the nodes, experimenting with all zoom levels when selecting the nodes.
Sometimes not.
The only consistent reproducable factor (that I can find) is that if you zoom all the way out the issue is mitigated.

CoreyCorza:
[...]

PR #7103, opened 2025-12-02
fix: handle Unicode characters in clipboard copy/paste and add Paste menu option

jtydhr88:
## Summary
Use TextEncoder/TextDecoder for UTF-8 safe base64 encoding/decoding

When copying nodes containing non-Latin1 characters (e.g., Chinese characters in localized_name field), btoa() throws an error:

InvalidCharacterError: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.

The copy operation saved data to localStorage successfully, but failed to write to the browser clipboard. On paste, the browser clipboard still contained old data, causing the wrong node to be pasted.

https://github.com/Comfy-Org/ComfyUI_frontend/issues/6993
https://github.com/Comfy-Org/ComfyUI_frontend/issues/5449
https://github.com/comfyanonymous/ComfyUI/issues/8481

before
https://github.com/user-attachments/assets/8abd9049-91bb-4200-8853-e26753376007

after
https://github.com/user-attachments/assets/7d969f32-bb0f-4c7a-baa2-65d576a4eba2

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7103-fix-handle-Unicode-characters-in-clipboard-copy-paste-and-add-Paste-menu-option-2bd6d73d365081f39c40e7e7f832b97c) by [Unito](https://www.unito.io)


DrJKL (dismissed):
I'd love two unit tests for this, if feasible.

DrJKL (dismissed):

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -8031,7 +8031,13 @@ export class LGraphCanvas
           has_submenu: true,
           callback: LGraphCanvas.onMenuAdd
         },
-        { content: 'Add Group', callback: LGraphCanvas.onGroupAdd }
+        { content: 'Add Group', callback: LGraphCanvas.onGroupAdd },
+        {
+          content: 'Paste',
```
DrJKL, line 8036: Are these internationalized by default?

jtydhr88: Yes, it should be, but right now our translation workflow is down

DrJKL (approved):

PR #7104, opened 2025-12-02
[feat] Allow opening workflows with same name as new tabs

Myestery:
When importing a workflow file that has the same name as an existing open workflow, create a new tab with a unique suffix (2), (3), etc. instead of silently failing or replacing the existing workflow.

- Add forceNew parameter to createTemporary() in workflowStore
- Always create new temporary workflow when importing via file picker
- Remove logic that looked up persisted workflows by name during import

https://github.com/user-attachments/assets/9c9aebd3-37c2-464f-9fb4-9ef871ec2673

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7104-feat-Allow-opening-workflows-with-same-name-as-new-tabs-2bd6d73d365081cd9182eaa35bf37c29) by [Unito](https://www.unito.io)


src/platform/workflow/management/stores/workflowStore.ts
```diff
@@ -365,21 +365,29 @@ export const useWorkflowStore = defineStore('workflow', () => {
...
       ComfyWorkflow.basePath + (path ?? 'Unsaved Workflow.json')
     )
-    const existingWorkflow = workflows.value.find((w) => w.fullFilename == path)
-    if (
-      path &&
-      workflowData &&
-      existingWorkflow?.changeTracker &&
-      !existingWorkflow.directory.startsWith(
-        ComfyWorkflow.basePath.slice(0, -1)
+    if (!forceNew) {
```
DrJKL: Nit: I would favor not negating the condition here.
Also, could this be merged with the internal condition?

src/platform/workflow/core/services/workflowService.ts
```diff
@@ -310,31 +310,37 @@ export const useWorkflowService = () => {
...
+          loadedWorkflow.changeTracker.restore()
+          return
+        }
+      }
+
       const tempWorkflow = workflowStore.createTemporary(
         path ? appendJsonExt(path) : undefined,
-        workflowData
+        workflowData,
+        true
```
DrJKL: Nit: Bare boolean arguments are a code smell, could you convert this to have named parameters / options arg?

DrJKL: Or splitting this into two separate methods, since it's really two different possible tasks.

Myestery: good idea

christian-byrne (approved):
LGTM. I thought of some other edge cases, but after testing manually, it seems they all still work properly.

PR #7127, opened 2025-12-03
refactor: rename `isSubscribed` for better expressiveness and avoid initializing cloud modules on local

christian-byrne:
Renames the shared subscription state export to keep the descriptive name `isSubscribedOrIsNotCloud`. Previously it was just `isSubscribed` which led to some misunderstandings and could lie

In some call sites, short-circuit on `isCloud` earlier, to avoid initializing a cloud-specific module (which can have non-breaking but inadvertant side-effects like network requests) on local. We want that all to be tree-shaken on local

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7127-refactor-rename-isSubscribed-for-better-expressiveness-and-avoid-initializing-cloud-mo-2be6d73d3650816caf69d20b07dc835e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/actionbar/ComfyRunButton/CloudRunButtonWrapper.vue
+++ b/src/components/actionbar/ComfyRunButton/CloudRunButtonWrapper.vue
@@ -1,7 +1,7 @@
 <template>
   <component
     :is="currentButton"
-    :key="isSubscribedOrIsNotCloud ? 'queue' : 'subscribe'"
+    :key="isSubscriptionRequirementMet ? 'queue' : 'subscribe'"
   />
 </template>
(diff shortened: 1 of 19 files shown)


DrJKL (changes requested):
I don't love compound logic in names like this.

Is there something more _essential_ about the gating that we can describe without exposing what specifically goes into it right now?

Maybe `subscriptionIneligible`?

christian-byrne:
> I don't love compound logic in names like this.

Why's that?

christian-byrne:
This is a mess. Going to re-open and address this later once these areas of the code are not being actively worked on.

PR #7132, opened 2025-12-03
fix: should allow autoCols=true when server doesn't provide size

### LINKED ISSUE #7117: Logs display is limited to less than half the width of the browser, with complete black area on the right.
### Prerequisites

### What happened?

Image

Does anyone else has this problem with the log display or it's just me? 

### Steps to Reproduce

1. It started a few weeks ago after an update. It happens on 2 different installations on different machines. 
2.
3.

Crashes ComfyUI completely

### ComfyUI Frontend Version

1.32.10

### Browser

Firefox

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

Running on Linux debian 12 and 13 as well.

This happens on chrome too.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-7117-Logs-display-is-limited-to-less-than-half-the-width-of-the-browser-with-complete-blac-2be6d73d3650814e8c65d93687a89cc2) by [Unito](https://www.unito.io)

jtydhr88:
## Summary

set autoCols = true by default for Logs Terminal

## Changes
Previously, the logs terminal had autoCols: false to preserve server-side progress bar rendering, However, this caused large black empty areas when the server terminal was narrower than the UI panel, Many server environments don't provide terminal size at all, making this tradeoff not worthwhile.

The original implementation set autoCols: false because the server renders progress bars based on its own terminal width. If the frontend used a different column count, progress bars would display incorrectly (e.g., wrapped or truncated).

However, this created a poor user experience:
  1. The terminal only filled a portion of the panel width (often 80 columns)
  2. The remaining space appeared as a large black empty area
  3. Many backend environments don't provide size data at all, making the fixed-width approach pointless

fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/7117

before
https://github.com/user-attachments/assets/e38af410-9cf1-4175-8acc-39d11a5b73ce

after
https://github.com/user-attachments/assets/3d180318-609f-44c5-aac5-230f9e4ef596

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7132-fix-should-allow-autoCols-true-when-server-doesn-t-provide-size-2be6d73d365081e7bf3bf5988bdba39a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/bottomPanel/tabs/terminal/LogsTerminal.vue
+++ b/src/components/bottomPanel/tabs/terminal/LogsTerminal.vue
@@ -19,7 +19,7 @@ import type { Ref } from 'vue'
 import { onMounted, onUnmounted, ref } from 'vue'
 
 import type { useTerminal } from '@/composables/bottomPanelTabs/useTerminal'
-import type { LogEntry, LogsWsMessage, TerminalSize } from '@/schemas/apiSchema'
+import type { LogEntry, LogsWsMessage } from '@/schemas/apiSchema'
 import { api } from '@/scripts/api'
 import { useExecutionStore } from '@/stores/executionStore'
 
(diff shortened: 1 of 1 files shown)


src/components/bottomPanel/tabs/terminal/LogsTerminal.vue
```diff
@@ -32,17 +32,35 @@ const terminalCreated = (
...
+    autoCols: true,
+    minCols: 80
+  })
 
   const update = (entries: Array<LogEntry>, size?: TerminalSize) => {
-    if (size) {
+    // Determine sizing mode once on first response
+    if (serverProvidesSize === null) {
+      serverProvidesSize = size?.cols != null && size.cols > 0
+    }
```
DrJKL: Optional:
1. Nullish coalescing assign
2. `!= null` isn't more expressive than a truthiness check, right?
```suggestion
    serverProvidesSize ??= size?.cols && size.cols > 0
```

Question: Would the server ever end a negative number?

src/components/bottomPanel/tabs/terminal/LogsTerminal.vue
```diff
@@ -32,17 +32,35 @@ const terminalCreated = (
...
+
+    // Use server size if provided, otherwise auto sizing handles it
+    if (serverProvidesSize && size?.cols) {
```
DrJKL: So if the server starts sending the columns after the first response, we ignore it?

DrJKL (dismissed):
I have a question about the logic around serverProvidesSize, but this is probably better than the baseline and if the response order sizing becomes an issue we can address it then.

Also a small style based suggestion.

DrJKL:
Oh, check out the rabbit comments too.

jtydhr88:
After investigating the backend code, the original design doesn't make practical sense:

1. Backend terminal size is irrelevant to frontend display: The server's terminal_service.py tries to get the terminal size via os.get_terminal_size(), but this only works when ComfyUI is launched from a real terminal. In most environments (Windows GUI, Docker, Colab, Jupyter), it falls back to a hardcoded (80, 24).
2. The progress bar concern is already handled: Progress bars render correctly in the frontend terminal. The minCols: 80 ensures a minimum width, which was the original fix for the Colab issue
3. The cost outweighs the benefit: The original autoCols: false was meant to handle a narrow edge case (Colab), but it sacrificed the entire auto-sizing functionality. Users end up with a fixed 80-column terminal that doesn't utilize the available panel width.

The correct approach is simply autoCols: true with minCols: 80 - let the frontend determine its own width based on the container, while maintaining the minimum width for edge cases. The server-provided size serves no practical purpose in the web frontend context.

DrJKL (approved):

PR #7140, opened 2025-12-03
Enable shift-drop context menu test

benceruleanlu:
## Summary
- turn the shift-drop context menu release action test back on
- keep the drag distance shorter to reduce flake risk while preserving behavior

## Testing
- pnpm typecheck
- pnpm lint:fix

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7140-Enable-shift-drop-context-menu-test-2be6d73d36508104a913d55279d4d3e5) by [Unito](https://www.unito.io)


benceruleanlu:
the test failed because it dragged too far, which was canvas before, but after the QPO's button, it dragged onto the run button, which ate the action, and started to make it fail

christian-byrne (approved):
Thank you!

PR #7095, opened 2025-12-02
Reactive widget.value and node.inputs in vue

AustinMroz:
Known Issues:
- Breaks undo

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7095-Reactive-widget-value-and-node-inputs-in-vue-2bd6d73d3650814099c2c4fbb71326f4) by [Unito](https://www.unito.io)

PR #7142, opened 2025-12-04
Implement a CustomCombo node

AustinMroz:
Adds a frontend support for a "Custom Combo" node which contains a Combo Widget, and a growing list of string inputs that determine the options available to the combo.

Custom Combo

By promoting the combo widget on this node, a list of options determined by workflow builders can be exposed to end users.

CC: @Kosinkadink

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7142-Implement-a-CustomCombo-node-2bf6d73d36508161951df01009a7262f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/customCombo.ts
+++ b/src/extensions/core/customCombo.ts
@@ -1,50 +1,115 @@
+import { useChainCallback } from '@/composables/functional/useChainCallback'
+import type { CanvasPointerEvent } from '@/lib/litegraph/src/types/events'
 import { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
-import { LiteGraph } from '@/lib/litegraph/src/litegraph'
-import { useLitegraphService } from '@/services/litegraphService'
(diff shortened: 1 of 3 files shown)


DrJKL:
I think this doesn't work when using Nodes 2.0

AustinMroz:
- Selecting entries works fine as is.
- Post #7095, adding and removing entries is functional.
- I'll push a fix to make combo options reactive shortly.

christian-byrne (approved):

christian-byrne:
From offline discussion:

> Of mention, it's no longer being implemented as a frontend only node. After speaking with teammate, we've added a skeleton node on the backend to address several of the edge cases (i.e. Some nodes can't be connected to).
> 
> More generally, I think the required maintenance will be near non-existant. The PR consists entirely of green lines, and exclusively uses old as dirt well defined litegraph APIs. I don't think we could break it without also creating far bigger problems with other custom ndoes.

PR #7154, opened 2025-12-04
Fix workflow loading from PNG images with both workflow and parameter…

### LINKED ISSUE #6633: Loading workflows from image fails
### Prerequisites

### What happened?

Loading complex workflows saved with custom_nodes are not being correctly interpreted anymore since Frontend version 1.32.2 (according to #6561 also since 1.32.1). <br>
I know that there's already an open issue for this, but the suggestes solution does not actually work and the suggested Merge-Request that is supposed to fix this issue is from June since it was a problem that already popped up in the past once. <br>
When trying to load this image (workflow attached) it instead loads a basic comfyui workflow with just the prompts and a sampler instead of the advanced workflow.
Image

When loading it in the last stable frontend release (1.28.8) it shows the full workflow as seen here:
Image

While in the new 1.32.2 it shows a basic workflow with just the prompts and sampler settings filled out:
Image

Dosubot suggested that the problem stems from values like is_changed being NaN instead of null or none.
I checked older images and there was a node with a NaN value and was about to open a PR in the image-saver node's repo to sanitize NaN's before writing the data to the image - but upon deleting the affected node and saving an image again (the image attached at the top) i noticed that the issue still persists even without any NaN values in the metadata of the image.

### Steps to Reproduce

1. Download attached image(workflow)
2. Open in frontend version 1.28.8 and the workflow loads correctly
3. Open in frontend version 1.32.2 (or 1.32.1) and the workflow fails to load correctly

### ComfyUI Frontend Version

1.32.2

### Browser

Firefox

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

I have not added any logs since there are no errors being thrown or anything different being shown other than the frontend version difference. <br>
I have tried this in 2 different sterila environments - once with all custom nodes required for the workflow installed and once without any nodes installed and just a clean comfyui installation. Both on v.1.28.8 and v.1.32.2

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6633-Loading-workflows-from-image-fails-2a46d73d365081b399c1e05c5a35a70e) by [Unito](https://www.unito.io)

r-vage:
…s metadata

- Reorder handleFile() to check workflow before parameters
- Add validation to prevent JSON parse errors from crashing imports
- Fix loadGraphData() to use explicit type validation instead of falsy check
- Ensures ComfyUI-generated PNGs with both metadata types load the workflow, not parameters

Fixes issue where large workflows (e.g., 634 nodes) were replaced with basic A1111 format when importing PNG files.

## Summary

Fixed workflow loading from PNG images to prioritize workflow metadata over parameters, preventing large workflows from being replaced with basic A1111 format.

## Changes

- **What**: Reordered `handleFile()` to check workflow before parameters, added JSON parse error handling and validation, fixed `loadGraphData()` to use explicit type checking instead of falsy check
- **Dependencies**: None

## Review Focus

The key issue was in `handleFile()` where parameters were checked before workflow, causing ComfyUI-generated PNGs (which contain both workflow and parameters metadata) to incorrectly import as A1111 format. The fix ensures:
1. Workflow is always checked first and validated properly
2. Parameters are only used as a fallback when no workflow exists
3. Invalid/malformed workflow data doesn't crash the import process

Additionally, `loadGraphData()` was using a falsy check (`if (!graphData)`) which could incorrectly replace valid but falsy values. Now uses explicit type validation.

Tested with real-world PNG containing 634-node workflow (780KB) + parameters (1KB) - now correctly loads the workflow instead of discarding it.

N/A - Backend logic fix, no UI changes

Fixes #6633

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7154-Fix-workflow-loading-from-PNG-images-with-both-workflow-and-parameter-2bf6d73d365081ecb7a6c4bf6b6ccd51) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -1441,31 +1441,30 @@ export class ComfyApp {
     // Check workflow first - it should take priority over parameters
     // when both are present (e.g., in ComfyUI-generated PNGs)
     if (workflow) {
-      let workflowObj: ComfyWorkflowJSON
+      let workflowObj: ComfyWorkflowJSON | undefined = undefined
       try {
         workflowObj =
           typeof workflow === 'string' ? JSON.parse(workflow) : workflow
+
+        // Only load workflow if parsing succeeded AND validation passed
+        if (
+          workflowObj &&
+          typeof workflowObj === 'object' &&
+          !Array.isArray(workflowObj)
+        ) {
+          await this.loadGraphData(workflowObj, true, true, fileName, {
+            openSource
+          })
+          return
+        } else {
+          console.error('Invalid workflow structure, trying parameters fallback')
+          this.showErrorOnFileLoad(file)
+        }
       } catch (err) {
         console.error('Failed to parse workflow:', err)
         this.showErrorOnFileLoad(file)
-        return
+        // Fall through to check parameters as fallback
       }
-
-      // Validate workflow is a proper object
-      if (
-        !workflowObj ||
-        typeof workflowObj !== 'object' ||
-        Array.isArray(workflowObj)
-      ) {
-        console.error('Invalid workflow structure')
-        this.showErrorOnFileLoad(file)
-        return
-      }
-
-      await this.loadGraphData(workflowObj, true, true, fileName, {
-        openSource
-      })
-      return
     }
 
     // Use parameters as fallback when no workflow exists
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -1441,29 +1441,29 @@ export class ComfyApp {
     // Check workflow first - it should take priority over parameters
     // when both are present (e.g., in ComfyUI-generated PNGs)
     if (workflow) {
-      let workflowObj: ComfyWorkflowJSON | undefined
+      let workflowObj: ComfyWorkflowJSON | undefined = undefined
       try {
         workflowObj =
           typeof workflow === 'string' ? JSON.parse(workflow) : workflow
+
+        // Only load workflow if parsing succeeded AND validation passed
+        if (
+          workflowObj &&
+          typeof workflowObj === 'object' &&
+          !Array.isArray(workflowObj)
+        ) {
+          await this.loadGraphData(workflowObj, true, true, fileName, {
+            openSource
+          })
+          return
+        } else {
+          console.error('Invalid workflow structure, trying parameters fallback')
+          this.showErrorOnFileLoad(file)
+        }
       } catch (err) {
         console.error('Failed to parse workflow:', err)
         this.showErrorOnFileLoad(file)
         // Fall through to check parameters as fallback
-      }
-
-      // Only load workflow if parsing succeeded AND validation passed
-      if (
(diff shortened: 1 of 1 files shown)


src/scripts/app.ts
```diff
@@ -1432,6 +1438,37 @@ export class ComfyApp {
...
+    // when both are present (e.g., in ComfyUI-generated PNGs)
+    if (workflow) {
+      let workflowObj: ComfyWorkflowJSON
+      try {
+        workflowObj =
+          typeof workflow === 'string' ? JSON.parse(workflow) : workflow
+      } catch (err) {
+        console.error('Failed to parse workflow:', err)
+        this.showErrorOnFileLoad(file)
+        return
```
DrJKL: Instead of returning here, why not continue on to try `parameters` or `prompt`?

src/scripts/app.ts
```diff
@@ -1432,6 +1438,37 @@ export class ComfyApp {
...
+        console.error('Invalid workflow structure')
+        this.showErrorOnFileLoad(file)
+        return
+      }
+
+      await this.loadGraphData(workflowObj, true, true, fileName, {
+        openSource
+      })
+      return
+    }
```
DrJKL, line 1468: I'd probably extract a method that returns `ComfyWorkflowJSON|undefined`, but how about this?

```suggestion
    if (workflow) {
      let workflowObj: ComfyWorkflowJSON
      try {
        workflowObj =
          typeof workflow === 'string' ? JSON.parse(workflow) : workflow
      } catch (err) {
        console.error('Failed to parse workflow:', err)
        this.showErrorOnFileLoad(file)
      }

      // Validate workflow is a proper object
      if (
        typeof workflowObj !== 'object' ||
        Array.isArray(workflowObj)
      ) {
        console.error('Invalid workflow structure')
      } else {
        await this.loadGraphData(workflowObj, true, true, fileName, {
          openSource
        })
        return
      }
    }
```

r-vage: if (workflow) {
  let workflowObj: ComfyWorkflowJSON | undefined
  try {
    workflowObj =
      typeof workflow === 'string' ? JSON.parse(workflow) : workflow
  } catch (err) {
    console.error('Failed to parse workflow:', err)
    // Fall through to check parameters
  }
  
  // Only load workflow if parsing succeeded AND validation passed
  if (
    workflowObj &&
    typeof workflowObj === 'object' &&
    !Array.isArray(workflowObj)
  ) {
    await this.loadGraphData(workflowObj, true, true, fileName, {
      openSource
    })
    return
  } else if (workflowObj !== undefined) {
    console.error('Invalid workflow structure, trying parameters fallback')
  }
}

DrJKL (dismissed):

DrJKL (approved):

mirabarukaso:
# The parameters section should under prompt

The API workflow uses prompt to stroge workflow because we don't care about node position.

Otherwise, it won't fully fix for API generated.

PR #7161, opened 2025-12-04
On adding output matchType, initialize type

AustinMroz:
The output type of a matchType output is initialized to COMFY_MATCHTYPE_V3, but is updated soon after to the value calculated from input types. Under some difficult to reproduce circumstances, this output type may be incorrectly evaluated in connections to other switch nodes. Since the initial type is never a valid connection, this can produce errors.

Instead, the output type of a matchtype node is initialized to allow connections to anything to ensure that the subsequent restriction of output types is guaranteed to be directional

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7161-On-adding-output-matchType-initialize-type-2bf6d73d3650819ab169ffe9a4ecfeb4) by [Unito](https://www.unito.io)

### DIFF
--- a/src/services/litegraphService.ts
+++ b/src/services/litegraphService.ts
@@ -219,7 +219,8 @@ export const useLitegraphService = () => {
    */
   function addOutputs(node: LGraphNode, outputs: OutputSpec[]) {
     for (const output of outputs) {
-      const { name, type, is_list } = output
+      const { name, is_list } = output
+      const type = output.type === 'COMFY_MATCHTYPE_V3' ? '*' : output.type
       const shapeOptions = is_list ? { shape: LiteGraph.GRID_SHAPE } : {}
       const nameKey = `${nodeKey(node)}.outputs.${output.index}.name`
       const typeKey = `dataTypes.${normalizeI18nKey(type)}`
--- a/src/services/litegraphService.ts
+++ b/src/services/litegraphService.ts
@@ -220,6 +220,7 @@ export const useLitegraphService = () => {
   function addOutputs(node: LGraphNode, outputs: OutputSpec[]) {
     for (const output of outputs) {
       const { name, is_list } = output
+      // TODO: Fix the typing at the node spec level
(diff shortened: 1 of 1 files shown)


src/services/litegraphService.ts
```diff
@@ -219,7 +219,8 @@ export const useLitegraphService = () => {
    */
   function addOutputs(node: LGraphNode, outputs: OutputSpec[]) {
     for (const output of outputs) {
-      const { name, type, is_list } = output
+      const { name, is_list } = output
+      const type = output.type === 'COMFY_MATCHTYPE_V3' ? '*' : output.type
```
DrJKL, line 224:
```suggestion
      // TODO: Fix the typing at the node spec level
      const type = output.type === 'COMFY_MATCHTYPE_V3' ? '*' : output.type
```

DrJKL (dismissed):

src/services/litegraphService.ts
AustinMroz: I agree there should be future improvement, but I'm specifically conflicted on if the backend spec should be changed.

If more dynamic output types are introduced in the future, then it is more correct to have the logic be performed as a lookup on the output type (as is done on the input side) instead of blindly checking for a dozen possible output properties.

How does YAGNI feel about adding todo's that may not be accurate in the future?

DrJKL: Would you have to continue adding types to upcast to `*` here otherwise?

DrJKL: By spec level, I mean when pulling in the node spec on the frontend, not at the point that we're handling outputs.

AustinMroz: The thought was that new output types would require more advanced logic than just overriding type (like setting up listeners, or adding multiple outputs). It's definitely a case where I don't have a firm grasp on what may be required, but knowing the suggestion is reasonable knowing it won't necessitate synchronization between frontend and backend versions.

DrJKL (approved):

PR #7169, opened 2025-12-04
feat: Add Jobs API infrastructure (PR 1 of 3)

ric-yu:
## Summary

Add Jobs API infrastructure in preparation for migrating from legacy `/history`, `/history_v2`, and `/queue` endpoints to the unified `/jobs` API.

**This is PR 1 of 3** - Additive changes only, no breaking changes.

## Changes

- **What**:
  - Add Zod schemas for runtime validation of Jobs API responses (`JobListItem`, `JobDetail`)
  - Add `fetchQueue`, `fetchHistory`, `fetchJobDetail` fetchers for `/jobs` endpoint
  - Add `extractWorkflow` utility for extracting workflow from nested job detail response
  - Add synthetic priority assignment for queue ordering (pending > running > history)
  - Add comprehensive tests for all new fetchers

- **Non-breaking**: All changes are additive - existing code continues to work

## Review Focus

1. **Zod schema flexibility**: Using `.passthrough()` to allow extra API fields - ensures forward compatibility but less strict validation
2. **Priority computation**: Synthetic priority ensures display order: pending (queued) → running → completed (history)
3. **Test coverage**: Verify tests adequately cover edge cases

## Files Added

- `src/platform/remote/comfyui/jobs/` - New Jobs API module
  - `types/jobTypes.ts` - Zod schemas and TypeScript types
  - `fetchers/fetchJobs.ts` - API fetchers with validation
  - `index.ts` - Barrel exports
- `tests-ui/tests/platform/remote/comfyui/jobs/fetchers/fetchJobs.test.ts` - Tests

## Next PRs

- **PR 2**: Migrate `getQueue()` and `getHistory()` to use Jobs API
- **PR 3**: Remove legacy history code and unused types

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7169-feat-Add-Jobs-API-infrastructure-PR-1-of-3-2bf6d73d3650812eae4ac0555a86969c) by [Unito](https://www.unito.io)

PR #7170, opened 2025-12-04
feat: Migrate to Jobs API (PR 2 of 3)

ric-yu:
## Summary

Migrate frontend from legacy `/history`, `/history_v2`, and `/queue` endpoints to the unified `/jobs` API with memory optimization and lazy loading.

**This is PR 2 of 3** - Core migration, depends on PR 1.

## Changes

- **What**:
  - Replace `api.getQueue()` and `api.getHistory()` implementations to use Jobs API fetchers
  - Implement lazy loading for workflow and full outputs via `/jobs/{id}` endpoint in `useJobMenu`
  - Add `TaskItemImpl` class wrapping `JobListItem` for queue store compatibility
  - Rename `reconcileHistory` to `reconcileJobs` for clarity
  - Use `execution_start_time` and `execution_end_time` from API for execution timing
  - Use `workflowId` from job instead of nested `workflow.id`
  - Update `useJobMenu` to fetch job details on demand (`openJobWorkflow`, `exportJobWorkflow`)

- **Breaking**: Requires backend Jobs API support (ComfyUI with `/jobs` endpoint)

## Review Focus

1. **Lazy loading in `useJobMenu`**: `openJobWorkflow` and `exportJobWorkflow` now fetch from API on demand instead of accessing `taskRef.workflow`
2. **`TaskItemImpl` wrapper**: Adapts `JobListItem` to existing queue store interface
3. **Error reporting**: Uses `execution_error` field from API for rich error dialogs
4. **Memory optimization**: Only fetches full job details when needed

## Files Changed

- `src/scripts/api.ts` - Updated `getQueue()` and `getHistory()` to use Jobs API
- `src/stores/queueStore.ts` - Added `TaskItemImpl`, updated to use `JobListItem`
- `src/composables/useJobMenu.ts` - Lazy loading for workflow access
- `src/composables/useJobList.ts` - Updated types
- Various test files updated

## Dependencies

- **Depends on**: PR 1 (Jobs API Infrastructure) - #7169

## Next PR

- **PR 3**: Remove legacy history code and unused types

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7170-feat-Migrate-to-Jobs-API-PR-2-of-3-2bf6d73d3650811b94f4fbe69944bba6) by [Unito](https://www.unito.io)


src/composables/queue/useJobMenu.ts
```diff
@@ -52,10 +53,20 @@ export function useJobMenu(
...
   const mediaAssetActions = useMediaAssetActions()
 
+  /**
+   * Fetches workflow data for a job, lazy loading from API if needed.
+   */
+  const getJobWorkflow = async (
+    jobId: string
+  ): Promise<ComfyWorkflowJSON | undefined> => {
+    const jobDetail = await fetchJobDetail((url) => api.fetchApi(url), jobId)
+    return extractWorkflow(jobDetail) as ComfyWorkflowJSON | undefined
```
DrJKL: Do you need the assertion here?
Can you ensure the type and annotate the return of `extractWorkflow`?

src/components/queue/job/useJobErrorReporting.ts
```diff
@@ -63,11 +37,16 @@ export const useJobErrorReporting = ({
...
   const reportJobError = () => {
-    const error = extractExecutionError(taskForJob.value)
-    if (error?.detail) {
-      dialog.showExecutionErrorDialog(error.detail)
+    const task = taskForJob.value
+
+    // Use execution_error from list response if available (includes prompt_id, timestamp)
+    const executionError = task?.executionError
+    if (executionError) {
+      dialog.showExecutionErrorDialog(executionError as ExecutionErrorWsMessage)
```
christian-byrne: I saw immediately that we are doing a type assertion from ExecutionError to ExecutionErrorWSMessage which isn't correct right? Those are not the same type. I also can't tell why the error in the job is used to trigger execution error dialog. Is this part of the plan to move execution errors to be controlled/owned/triggered by /jobs rather than WS message? If so, that's really good, but we need to update the schema rather than using type assertion.

christian-byrne (commented):
Reviewed 4/39 files. Taking break

src/components/queue/job/useJobErrorReporting.ts
ric-yu: Good catch! The ExecutionError is built from the WS message, so ideally, they should be the same shape/type. The OpenAPI schema is missing a few fields within ExecutionErrorWsMessage. For now, the function will take a subtype that both `ExecutionError` and `ExecutionErrorWSMessage` will have the fields for. Would that make sense?

benceruleanlu:
Reviewing this now

src/platform/remote/comfyui/jobs/fetchJobs.ts
```diff
@@ -139,8 +140,12 @@ export async function fetchJobDetail(
...
  */
-export function extractWorkflow(job: JobDetail | undefined): unknown {
+export function extractWorkflow(
+  job: JobDetail | undefined
+): ComfyWorkflowJSON | undefined {
   const parsed = zWorkflowContainer.safeParse(job?.workflow)
   if (!parsed.success) return undefined
-  return parsed.data.extra_data?.extra_pnginfo?.workflow
+  return parsed.data.extra_data?.extra_pnginfo?.workflow as
+    | ComfyWorkflowJSON
```
benceruleanlu: this is out of scope, but we should probably take a look at this tech debt of z.unknown

benceruleanlu: after this project*

src/platform/remote/comfyui/history/reconciliation.ts
```diff
@@ -1,122 +1,40 @@
...
-  )
+  // Items from server not yet in client cache
+  const newItems = serverJobs.filter((item) => !clientIds.has(item.id))
 
-  const clientItemsStillOnServer = clientHistory.filter((item) =>
-    serverPromptIds.has(item.prompt[1])
+  // Retain client items that still exist on server
+  const clientItemsStillOnServer = clientJobs.filter((item) =>
+    serverIds.has(item.id)
   )
```
benceruleanlu: question/nit: isn't the server the source of truth here? it seems like we're prefering whatever the client has, stale data risk?

src/platform/remote/comfyui/jobs/fetchJobs.ts
benceruleanlu: ['completed'] in fetchHistory, ['in_progress', 'pending'] in fetchQueue

```
const zJobStatus = z.enum([
  'pending',
  'in_progress',
  'completed',
  'failed',
  'cancelled'
])
```

cancelled failed missing, noone using fetchJobsRaw outside this file, we are missing cancelled and failed state jobs, I queued a job and cancelled it, the job is missing, presumably the same for failed jobs as well, I think that's a regression

benceruleanlu (changes requested):
things generally look good, I haven't taken a look at the tests yet, there are some minor edge case/state possible refinements that could be improved but is nitty for now, biggest concern is the failed/cancelled state missing, can you PTAL?

src/composables/queue/useJobMenu.ts
```diff
@@ -81,23 +92,29 @@ export function useJobMenu(
...
+      return
+    }
+
+    // Fall back to simple error dialog
+    const message = item.taskRef?.errorMessage
+    if (message) {
+      useDialogService().showErrorDialog(new Error(message), {
+        reportType: 'queueJobError'
+      })
+    }
```
christian-byrne, line 106: After failed jobs are included, we should check we need this code at all. The error dialog is already shown via this code https://github.com/Comfy-Org/ComfyUI_frontend/blob/ea96c718182842bf5e4b8ae131a2cad5b2895896/src/scripts/app.ts#L693-L695

src/composables/queue/useResultGallery.ts
```diff
@@ -1,27 +1,85 @@
...
+type FetchApi = (url: string) => Promise<Response>
+
+const getPreviewableOutputs = (outputs?: readonly ResultItemImpl[]) =>
+  outputs?.filter((o) => o.supportsPreview) ?? []
+
+const findActiveIndex = (items: ResultItemImpl[], url?: string): number => {
+  if (!url) return 0
+  const idx = items.findIndex((o) => o.url === url)
+  return idx >= 0 ? idx : 0
+}
```
christian-byrne: nit: I think it might be better to put these as class methods on ResultItemImpl or TaskItemImpl just to continue the existing pattern - regardless of it's the better pattern overall.

src/composables/queue/useResultGallery.ts
```diff
@@ -1,27 +1,85 @@
 import { ref, shallowRef } from 'vue'
 
 import type { JobListItem } from '@/composables/queue/useJobList'
-import type { ResultItemImpl } from '@/stores/queueStore'
+import type { ResultItemImpl, TaskItemImpl } from '@/stores/queueStore'
+
+type FetchApi = (url: string) => Promise<Response>
+
+const getPreviewableOutputs = (outputs?: readonly ResultItemImpl[]) =>
```
christian-byrne: In case it's helpful, I just checked and if I remove the `?`, it still compiles without error.

src/stores/queueStore.ts
```diff
@@ -429,28 +402,53 @@ export class TaskItemImpl {
       : undefined
   }
 
-  public async loadWorkflow(app: ComfyApp) {
-    let workflowData = this.workflow
+  /**
+   * Loads full outputs for tasks that only have preview data
+   * Returns a new TaskItemImpl with full outputs and execution status
+   */
+  public async loadFullOutputs(
```
christian-byrne: nit: I do find it unintuitive that making a request to `/jobs` with prompt_id param gives a different (richer) data structure than what you get when requesting `/jobs` normally (getting pages). Not sure what the best pattern would be.

src/composables/queue/useResultGallery.ts
```diff
@@ -1,27 +1,85 @@
...
+    const cacheKey = String(task.promptId)
+    const cached = loadedTasksCache.get(cacheKey)
+    if (cached) {
+      return getPreviewableOutputs(cached.flatOutputs)
+    }
+
+    const loadedTask = await task.loadFullOutputs(fetchApi)
+    loadedTasksCache.set(cacheKey, loadedTask)
+    return getPreviewableOutputs(loadedTask.flatOutputs)
+  }
```
christian-byrne: We should extract this out to its own location (like its own store for fetching and caching lazily loaded job outputs), rather than being in a component hook here

src/stores/queueStore.ts
```diff
@@ -215,29 +212,24 @@ export class ResultItemImpl {
...
-    this.taskType = taskType
-    this.prompt = prompt
-    this.status = status
+    this.job = job
+    // If no outputs provided but job has preview_output, create synthetic outputs
+    const effectiveOutputs =
+      outputs ??
+      (job.preview_output
+        ? { preview_node: { images: [job.preview_output] } }
+        : {})
```
christian-byrne, line 243: This seems like we are fighting a bit against our own types and server responses - is there a way for the server to return something more consistent since its own our API?

src/composables/queue/useResultGallery.ts
```diff
@@ -1,27 +1,85 @@
...
+
+    const targetTask = item.taskRef
+    let targetOutputs: ResultItemImpl[] = []
+
+    if (targetTask) {
+      targetOutputs = await getOutputsForTask(targetTask)
+    }
+
+    // Abort if a newer request was made while loading
+    if (requestId !== currentRequestId) return
```
christian-byrne: If applying above and moving this logic to a store, the view-layer logic doesn't need to manually handle request deduping.

src/composables/queue/useResultGallery.ts
```diff
@@ -1,27 +1,85 @@
...
+
+    const activeUrl = item.taskRef?.previewOutput?.url
+
+    if (targetOutputs.length > 0) {
+      galleryItems.value = targetOutputs
+      galleryActiveIndex.value = findActiveIndex(targetOutputs, activeUrl)
+    } else {
+      const items = tasks.flatMap((t) => {
+        const preview = t.previewOutput
+        return preview?.supportsPreview ? [preview] : []
```
christian-byrne: If we move this to class property and adjust it slightly, we don't need to check `supportPreview` at each access site and can just have the implementation return an empty array in those cases - so all consumers deal with simpler contract.

src/platform/remote/comfyui/history/reconciliation.ts
christian-byrne: We might want to just keep it how it is for now since we know it works (that is, if requesting window past what we have, keep optimistic data on client from previous requests).

src/platform/remote/comfyui/jobs/fetchJobs.ts
```diff
@@ -77,7 +77,7 @@ export async function fetchHistory(
 ): Promise<JobListItem[]> {
   const { jobs, total } = await fetchJobsRaw(
     fetchApi,
-    ['completed'],
+    ['completed', 'failed', 'cancelled'],
```
christian-byrne, line 81: Can get rid of cancelled.

src/scripts/ui.ts
```diff
@@ -259,7 +259,7 @@ class ComfyList {
         $el('div.comfy-list-items', [
           // @ts-expect-error fixme ts strict error
           ...(this.#reverse ? items[section].reverse() : items[section]).map(
-            (item: TaskItem) => {
+            (item: any) => {
```
DrJKL: `unknown` and a guard to check that it's an `object`?
Anything but `any` 🥺

src/stores/queueStore.ts
```diff
@@ -30,6 +24,9 @@ import { getMediaTypeFromFilename } from '@/utils/formatUtil'
 // Task type used in the API.
 type APITaskType = 'queue' | 'history'
 
+// Internal task type derived from job status
+type TaskType = 'Running' | 'Pending' | 'History'
```
christian-byrne: Would be good for this to be in the schema itself

src/stores/queueStore.ts
```diff
@@ -285,88 +289,73 @@ export class TaskItemImpl {
   }
 
   get queueIndex() {
-    return this.prompt[0]
+    return this.job.priority
   }
```
christian-byrne, line 310: Can we just rename things to have a singular name? Instead of aliasing job.priority to queueIndex, we can just rename everything to use job.priority.

src/stores/queueStore.ts
```diff
@@ -285,88 +289,73 @@ export class TaskItemImpl {
...
-    return this.prompt[1]
-  }
-
-  get promptInputs() {
-    return this.prompt[2]
-  }
-
-  get extraData() {
-    return this.prompt[3]
+    return this.job.id
```
christian-byrne, line 313: Similar to above, we should begin to start just using job.id everywhere.

src/stores/queueStore.ts
```diff
@@ -285,88 +289,73 @@ export class TaskItemImpl {
...
   }
 
-  get workflow(): ComfyWorkflowJSON | undefined {
-    return this.extraPngInfo?.workflow
+  /**
+   * Execution error details if job failed with traceback.
+   * Returns the error object for detailed error dialogs.
+   */
+  get executionError() {
+    return this.job.execution_error ?? undefined
```
christian-byrne: Maybe we don't need these comments since the code is self-explantory and now the comments need to be maintained.

src/stores/queueStore.ts
```diff
@@ -285,88 +289,73 @@ export class TaskItemImpl {
...
    */
-  get errorMessage(): string | undefined {
-    const messages = this.status?.messages
-    if (!Array.isArray(messages) || !messages.length) return undefined
-    const record = messages.find(
-      (entry: unknown) => Array.isArray(entry) && entry[0] === 'execution_error'
-    ) as [string, { exception_message?: string }?] | undefined
-    return record?.[1]?.exception_message
+  get workflow(): undefined {
+    return undefined
```
christian-byrne: Let's try to see how we can remove this entirely.

src/stores/queueStore.ts
```diff
@@ -285,88 +289,73 @@ export class TaskItemImpl {
...
    */
-  get executionError(): ExecutionErrorWsMessage | undefined {
-    const messages = this.status?.messages
-    if (!Array.isArray(messages) || !messages.length) return undefined
-    const record = messages.find(
-      (entry: unknown) => Array.isArray(entry) && entry[0] === 'execution_error'
-    ) as [string, ExecutionErrorWsMessage?] | undefined
-    return record?.[1]
+  get messages(): Array<[string, unknown]> {
+    return []
```
christian-byrne: If the execution error is derived in another way now, we should try to remove this as well.

src/stores/queueStore.ts
```diff
@@ -285,88 +289,73 @@ export class TaskItemImpl {
...
-    ) as [string, { exception_message?: string }?] | undefined
-    return record?.[1]?.exception_message
+  get workflow(): undefined {
+    return undefined
   }
 
   /**
-   * Extracts the full execution error from status messages.
-   * Returns the ExecutionErrorWsMessage for detailed error dialogs.
+   * Execution messages - not available in Jobs API
```
DrJKL:
```suggestion
   * Execution messages - not available in Jobs API
   * @deprecated DON'T USE THIS PLZKTHX
   
```

src/stores/queueStore.ts
```diff
@@ -285,88 +289,73 @@ export class TaskItemImpl {
...
-  get interrupted() {
-    return _.some(
-      this.messages,
-      (message) => message[0] === 'execution_interrupted'
-    )
+  /**
+   * Whether the job was interrupted/cancelled
+   */
+  get interrupted(): boolean {
+    return this.job.status === 'cancelled'
```
christian-byrne: Seems like 'cancelled' is not a valid job status, so we would use a slight modification of the original code rather than this?

src/stores/queueStore.ts
```diff
@@ -429,28 +402,53 @@ export class TaskItemImpl {
...
-        (url) => app.api.fetchApi(url),
-        this.promptId
-      )
+    if (!jobDetail?.outputs) {
+      return this
+    }
+
+    // Create new TaskItemImpl with full outputs
+    return new TaskItemImpl(this.job, jobDetail.outputs)
+  }
```
christian-byrne: Seems we are using the same logic elsewhere but with caching. maybe they can all use same store that this code also uses and then it will get caching as well.

src/stores/queueStore.ts
christian-byrne, line 1: nit: not related to this PR necessarilly, but would be good to move TaskItemImpl to its own file.

src/stores/queueStore.ts
```diff
@@ -429,28 +402,53 @@ export class TaskItemImpl {
...
   }
 
-  public async loadWorkflow(app: ComfyApp) {
-    let workflowData = this.workflow
+  /**
+   * Loads full outputs for tasks that only have preview data
+   * Returns a new TaskItemImpl with full outputs and execution status
+   */
+  public async loadFullOutputs(
+    fetchApi: (url: string) => Promise<Response>
```
DrJKL: Can this use the global fetchApi now?

src/stores/queueStore.ts
```diff
@@ -490,31 +483,14 @@ export class TaskItemImpl {
...
-      taskType: 'History',
-      prompt: this.prompt,
-      status: this.status!,
-      outputs: this.outputs
-    }
-    return item
+  /**
+   * Returns the underlying job data
+   */
+  public toJob(): JobListItem {
```
christian-byrne: nit: `toJob` seems like a bad namer for a getter.

christian-byrne: Actually we probably don't even need a function for this, just access the property anyway. The property is readonly so no need for it to be computed

tests-ui/tests/composables/useJobMenu.test.ts
```diff
@@ -58,12 +58,21 @@ vi.mock('@/platform/workflow/management/stores/workflowStore', () => ({
   useWorkflowStore: () => workflowStoreMock
 }))
 
+const fetchJobDetailMock = vi.fn()
+const extractWorkflowMock = vi.fn()
+vi.mock('@/platform/remote/comfyui/jobs/fetchJobs', () => ({
+  fetchJobDetail: (...args: any[]) => fetchJobDetailMock(...args),
+  extractWorkflow: (...args: any[]) => extractWorkflowMock(...args)
+}))
+
```
DrJKL:
```suggestion
vi.mock('@/platform/remote/comfyui/jobs/fetchJobs', () => ({
  fetchJobDetail: vi.fn(),
  extractWorkflow: vi.fn()
}))

```

christian-byrne (commented):
forgot to submit this comment originally

src/stores/queueStore.ts
ric-yu: This is a standard REST pattern. List endpoints return lightweight summaries while detail endpoints return full data.

src/stores/queueStore.ts
ric-yu: Tackling this in a follow-up PR

src/composables/queue/useJobMenu.ts
ric-yu: The execution_error event shows the dialog automatically during execution, but this code lets users re-view errors later when browsing failed jobs in history.

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -398,24 +400,66 @@ const handleZoomClick = (asset: AssetItem) => {
...
+  if (!promptId) {
     console.warn('Missing required folder view data')
     return
   }
 
   folderPromptId.value = promptId
   folderExecutionTime.value = executionTimeInSeconds
 
-  folderAssets.value = allOutputs.map((output) => ({
+  // Determine which outputs to display
```
christian-byrne, line 421:
```suggestion
```

src/stores/queueStore.ts
christian-byrne: When you say "standard REST pattern", do you mean in the sense of common industry practice rather than something prescribed by REST itself? I’m not aware of a REST constraint that requires list endpoints to return summaries and filtered variants to return richer representations. In practice, it also seems more common to expose the full representation via something like `GET /jobs/{jobId}`, with `/jobs` (including filtered variants) returning a consistent, lighter-weight shape.

This is not blocking the PR by any means.

christian-byrne (commented):
19/45, continuing later tonight.

src/stores/jobOutputStore.ts
```diff
@@ -0,0 +1,160 @@
...
+import { extractWorkflow } from '@/platform/remote/comfyui/jobs/fetchJobs'
+import type { ComfyWorkflowJSON } from '@/platform/workflow/validation/schemas/workflowSchema'
+import { api } from '@/scripts/api'
+import { ResultItemImpl } from '@/stores/queueStore'
+import type { TaskItemImpl } from '@/stores/queueStore'
+
+const MAX_TASK_CACHE_SIZE = 50
+const MAX_JOB_DETAIL_CACHE_SIZE = 50
+
+function createLRUCache<T>(maxSize: number) {
```
christian-byrne: Can we use the existing LRU library in the repo (Quick LRU)?

src/stores/jobOutputStore.ts
christian-byrne, line 1: Should we simplify this file by using https://github.com/Comfy-Org/ComfyUI_frontend/blob/dbb0bd961fe65bd42d06da7764385abe1dea4e22/src/composables/useCachedRequest.ts? It has LRU caching with our library and request deduping with abort signal and exposes an API that makes it easier for components to decide when to evict or cancel.

src/stores/queueStore.ts
christian-byrne: @coderabbitai make an issue to track this please

src/stores/queueStore.ts
christian-byrne: @coderabbitai make an issue to track this please

christian-byrne (approved):
LGTM generally

src/stores/jobOutputStore.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    // Job details & workflows
+    getJobDetail,
+    getJobWorkflow,
+
+    // Cache management
+    clearTaskCache,
+    clearJobDetailCache,
+    clearAllCaches
+  }
```
DrJKL: Are all of these used (not including tests)?

ric-yu: Updated teh tests to not need cache management. the rest are used by the assets management or results gallery.

PR #7179, opened 2025-12-05
feat(sidebar): autofocus search input in Workflows, Model, and Node tabs

Myestery:
## Summary
This PR improves the user experience by automatically focusing the search input field when opening the Workflows, Model Library, or Node Library sidebar tabs.

## Changes
- **SearchBox.vue**: Exposed a `focus()` method to allow parent components to programmatically set focus on the input.
- **WorkflowsSidebarTab.vue**: Added a watcher to focus the search box when the tab is activated.
- **ModelLibrarySidebarTab.vue**: Added autofocus on component mount.
- **NodeLibrarySidebarTab.vue**: Added autofocus on component mount.

## Motivation
Users often switch to these tabs specifically to search for an item. Automatically focusing the search box reduces friction and saves a click.

https://github.com/user-attachments/assets/8438e71c-a5e4-4b6c-8665-04d535d3ad8e

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7179-feat-sidebar-autofocus-search-input-in-Workflows-Model-and-Node-tabs-2c06d73d36508199b8c0e83d19f1cd84) by [Unito](https://www.unito.io)


DrJKL (changes requested):
Are you able to just use [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus)?

Myestery:
> Are you able to just use [autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus)?

So I tried using it but got errors because the tab switch button is becoming the active element 

> Autofocus processing was blocked because a document already has a focused element.

PR #7181, opened 2025-12-05
Fix desktop menu docs links regression

benceruleanlu:
## Summary
- make `useExternalLink` rely on the global i18n locale so it can be used safely outside setup
- restore `electronAdapter` to use the shared `useExternalLink` helper for docs URLs and static links

## Motivation
Desktop menu items disappeared because a top-level call to `useExternalLink` in `electronAdapter` triggered `useI18n` at module-eval time, throwing and blocking extension registration. By making the composable global-locale-only and using it in `electronAdapter`, the module can load without setup context while preserving link behavior.

## Testing
- pnpm typecheck
- pnpm lint:fix

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7181-Fix-desktop-menu-docs-links-regression-2c06d73d36508157ae48cff078b9173e) by [Unito](https://www.unito.io)

### DIFF
--- a/tests-ui/tests/composables/useExternalLink.test.ts
+++ b/tests-ui/tests/composables/useExternalLink.test.ts
@@ -1,5 +1,4 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import { createI18n } from 'vue-i18n'
 
 // Mock the environment utilities
 vi.mock('@/utils/envUtil', () => ({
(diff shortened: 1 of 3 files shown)


tests-ui/tests/composables/useExternalLink.test.ts
```diff
@@ -9,12 +9,14 @@ vi.mock('@/utils/envUtil', () => ({
...
 }))
 
-// Mock vue-i18n
+// Mock global i18n locale ref
 const mockLocale = ref('en')
-vi.mock('vue-i18n', () => ({
-  useI18n: vi.fn(() => ({
-    locale: mockLocale
-  }))
+vi.mock('@/i18n', () => ({
```
DrJKL, line 17: You should be able to just `createI18n` as a plugin for the test

benceruleanlu: ok fixed

benceruleanlu: actually this sounds somewhat excessive, we only need i18n.global.locale.value, can you recheck? idm either way eod though

simula-r (approved):
LGTM

tests-ui/tests/composables/useExternalLink.test.ts
DrJKL: https://github.com/search?q=repo%3AComfy-Org%2FComfyUI_frontend+path%3A.test.ts+createI18n({&type=code

christian-byrne (approved):

PR #7190, opened 2025-12-05
Design: Model management

DrJKL:
## Summary

Assorted updates to the components involved in uploading personal models.

## Changes

- Standardize Import buttons
- Let the images fill the space on the card
- Order by recent by default
- Nicer display on the model select popover

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7190-Design-Model-management-2c06d73d365081e7b9fed7a83b730c0f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/assets/components/AssetFilterBar.vue
+++ b/src/platform/assets/components/AssetFilterBar.vue
@@ -1,6 +1,12 @@
 <template>
-  <div :class="containerClasses" data-component-id="asset-filter-bar">
-    <div :class="leftSideClasses" data-component-id="asset-filter-bar-left">
+  <div
+    class="flex gap-4 items-center justify-between px-6 pt-2 pb-6"
+    data-component-id="asset-filter-bar"
+  >
+    <div
+      class="flex gap-4 items-center"
+      data-component-id="asset-filter-bar-left"
+    >
       <MultiSelect
         v-if="availableFileFormats.length > 0"
         v-model="fileFormats"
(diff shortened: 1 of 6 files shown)


luke-mino-altherr (approved):

src/platform/assets/components/AssetFilterBar.vue
```diff
@@ -16,18 +22,18 @@
...
       />
     </div>
 
-    <div :class="rightSideClasses" data-component-id="asset-filter-bar-right">
+    <div class="flex items-center" data-component-id="asset-filter-bar-right">
       <SingleSelect
         v-model="sortBy"
         :label="$t('assetBrowser.sortBy')"
         :options="sortOptions"
-        :class="selectClasses"
```
christian-byrne, line 30: Maybe we should enable that linting rule. Does one exist?

src/platform/assets/components/AssetFilterBar.vue
```diff
@@ -48,43 +54,38 @@ import type { SelectOption } from '@/components/input/types'
 import { t } from '@/i18n'
```
christian-byrne, line 54: Not related to this PR, but it's definitely a mistake that we are using `@/i18n` here instead of the `useI18n`. Basically forgoing reactivity, right?

Another thing we could add a linter rule for ("prefer reactive translations in component setup functions").

christian-byrne (approved):

src/platform/assets/components/AssetFilterBar.vue
christian-byrne: Oh that was actually fine before, didn't realize it was a variable.

DrJKL: https://github.com/francoismassart/eslint-plugin-tailwindcss/blob/master/docs/rules/no-custom-classname.md

PR #7211, opened 2025-12-06
Color links as common type

### LINKED ISSUE #7196: Reroute Nodes and Wires Should Inherit Upstream Output Color
### Feature Idea

It would be really helpful if reroute nodes and their connecting wires could automatically inherit the color of the output they’re receiving. It would make large graphs much easier to read and keep the visual logic consistent throughout the workflow. 

Image

### Existing Solutions

_No response_

### Other

_No response_        

---

*This issue is transferred from: https://github.com/comfyanonymous/ComfyUI/issues/11122*

Original issue was created by @M8bits at 2025-12-05T14:16:11.000Z

**Original Comments:**

@dynatron832: <pre>i would call that more like a "bug" ...the feature you want (and me too) is already in comfy since years. and if you connect nodes directly they seem to stick to the correct colors. but as soon as the classic reroutes are in the line the colors are bugged. the last updates have also killed rgthree´s reroute nodes which are super flexible with the direction option.

it looks like comfy now has this round type reroute nodes. seems they stick to the line colors ...but they are creepy.</pre>

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-7196-Reroute-Nodes-and-Wires-Should-Inherit-Upstream-Output-Color-2c06d73d36508162bc16eb00301b7a66) by [Unito](https://www.unito.io)

AustinMroz:
Previously the color of a link would simply use the type of the target slot and fallback to the type of the origin slot. When a connection is made to a node that accepts the any type ('*'), the link has the green color of an unknown type.

Instead, when a connection is made, the type of a link is now calculated as the greatest common type of the source and destination. This means that connections to reroutes are correctly colored.

| Before | After |
| ------ | ----- |
| before| after|

The code for calculating common types already exists, it has simply been moved into litegraph and given a more descriptive name.

Resolves #7196

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7211-Color-links-as-common-type-2c16d73d365081188460f6b5973db962) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/utils/type.ts
+++ b/src/lib/litegraph/src/utils/type.ts
@@ -49,7 +49,7 @@ function intersection(...sets: string[][]): string[] {
     for (const item of new Set(set))
       itemCounts[item] = (itemCounts[item] ?? 0) + 1
   return Object.entries(itemCounts)
-    .filter(([, count]) => count == sets.length)
+    .filter(([, count]) => count === sets.length)
     .map(([key]) => key)
 }
 
--- a/src/lib/litegraph/src/utils/type.ts
+++ b/src/lib/litegraph/src/utils/type.ts
@@ -1,4 +1,6 @@
-import type { IColorable } from '@/lib/litegraph/src/interfaces'
+import { without } from 'es-toolkit'
+
+import type { IColorable, ISlotType } from '@/lib/litegraph/src/interfaces'
 
 /**
  * Converts a plain object to a class instance if it is not already an instance of the class.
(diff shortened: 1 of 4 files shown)


src/lib/litegraph/src/utils/type.ts
```diff
@@ -26,3 +28,31 @@ export function isColorable(obj: unknown): obj is IColorable {
...
+  const itemCounts: Record<string, number> = {}
+  for (const set of sets)
+    for (const item of new Set(set))
+      itemCounts[item] = (itemCounts[item] ?? 0) + 1
+  return Object.entries(itemCounts)
+    .filter(([, count]) => count == sets.length)
+    .map(([key]) => key)
+}
+
+function isStrings(types: unknown[]): types is string[] {
```
AustinMroz, line 56: A little conflicted about moving this here, but I think it's good enough.
\+ commonType should be in litegraph and should see greater use.
\+ Having litegraph reference outside code is bad.
\- There's some conflation of litegraph connection types and programming types.
\- The function is no longer exported. It's less clear that it's intended as utility.

DrJKL: Intent is less important than how it's currently being used, so I support this.
If we had something that was serving as a utility and could be put into the frontend utilities package for use here, that would also be good, but YAGNI.

DrJKL (approved):

PR #7231, opened 2025-12-08
Tests: Playwright test timeouts

DrJKL:
## Summary

See where we can use proper DOM waiting instead of waitForTimeout.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7231-WIP-Playwright-test-timeouts-2c36d73d3650812b966ac3d9c338dfd4) by [Unito](https://www.unito.io)


Myestery (approved):
Its Interesting  how they pass without the timeouts

PR #7233, opened 2025-12-08
fix: Vue Nodes 2.0 slot link drag not working on mobile/touch devices

Myestery:
## Summary
- Fix slot link drag and snap/attraction not working on mobile browsers in Vue Nodes 2.0 mode
- Use `document.elementFromPoint()` to get the actual element under the pointer instead of relying on `event.target`

## Root Cause
On touch/mobile devices, pointer events have "implicit pointer capture" - when you touch an element, all subsequent pointer events (`pointermove`, `pointerup`) for that touch are sent to the same element where the touch started, regardless of where the pointer moves.

The code was using `event.target` to find slots under the pointer for snap/attraction. On touch devices, this always returned the original slot element (where the drag started), not the element currently under the touch point. This caused:
- No snap/attraction when dragging links over other slots
- Connections failing when dropping on target slots

## Before
https://github.com/user-attachments/assets/55b56d5c-9744-4d6c-abfd-3a2136ab25bc

## After
https://github.com/user-attachments/assets/5bdf2a22-0025-4ae1-9358-35f0100b67d4

## Test plan
- [ ] Enable Vue Nodes 2.0 mode in settings
- [ ] Test on mobile browser or Chrome DevTools mobile simulation
- [ ] Drag a link from one node's output slot to another node's input slot
- [ ] Verify the link snaps/attracts to compatible slots during drag
- [ ] Verify the connection is made successfully on drop

Fixes #7224


src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -501,9 +505,14 @@ export function useSlotLinkInteraction({
...
-    const hasConnected = connectByPriority(canvasEvent.target, snappedCandidate)
+    // Use elementFromPoint to get the actual element under the pointer.
+    // On touch devices, event.target stays as the element where touch started.
+    const dropTarget =
+      document.elementFromPoint(event.clientX, event.clientY) ??
+      canvasEvent.target
+    const hasConnected = connectByPriority(dropTarget, snappedCandidate)
 
-    if (!hasConnected && event.target === app.canvas?.canvas) {
+    if (!hasConnected && dropTarget === app.canvas?.canvas) {
```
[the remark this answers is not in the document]
DrJKL: @coderabbitai what would that test look like and where would it go?

src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.test.ts
```diff
@@ -0,0 +1,121 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { resolvePointerTarget } from '@/renderer/extensions/vueNodes/composables/useSlotLinkInteraction'
+
+describe('resolvePointerTarget', () => {
+  let originalElementFromPoint: typeof document.elementFromPoint
+
+  beforeEach(() => {
+    originalElementFromPoint = document.elementFromPoint
```
DrJKL: Could you use vitest's mocking for this?

DrJKL: https://vitest.dev/api/vi.html#vi-spyon

PR #7251, opened 2025-12-08
fix: Vue Nodes dropdown doesn't show selected model from Asset Browser

luke-mino-altherr:
## Summary
Fixes Vue Nodes 2.0 dropdowns not displaying the selected model when created from the Asset Browser.

## Root Cause
Widget values were set AFTER the node was added to the graph, causing Vue's reactivity system to capture stale initial values.

## Solution
Set widget value BEFORE adding node to graph in `createModelNodeFromAsset.ts`.

## Changes
- **1 file changed**: `createModelNodeFromAsset.ts`
- **4 lines added, 1 removed**: Move widget value assignment before graph.add()

This ensures Vue's reactivity system captures the correct initial widget value when the node is created.

### DIFF
--- a/src/platform/assets/utils/createModelNodeFromAsset.ts
+++ b/src/platform/assets/utils/createModelNodeFromAsset.ts
@@ -199,8 +199,11 @@ export function createModelNodeFromAsset(
     }
   }
 
-  targetGraph.add(node)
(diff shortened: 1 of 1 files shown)


DrJKL (approved):

AustinMroz (approved):
Change is reasonable and minimal.  No objections even with the impending reactivity PR.

PR #7268, opened 2025-12-09
fix: loading state to show loader only if it takes more than 250ms

simula-r:
## Summary

To prevent the flash of "loading..." and "calculating dimensions" when loading cached images only set loading set if longer than 250ms

## Changes

- **What**: ImagePreview.vue
- **Breaking**: 
- **Dependencies**: 

The retrigger loading is because i have throttled 4g slow in the demo. So cache takes time. Normally this doesn't happen.

https://github.com/user-attachments/assets/335ca7e4-4ce1-43dd-b7d0-9ee88e187069

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7268-fix-loading-state-to-show-loader-only-if-it-takes-more-than-150ms-2c46d73d365081a6b311f78ba3e1cffd) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -150,9 +147,26 @@ const isHovered = ref(false)
...
+}
+
+const clearLoaderTimeout = () => {
+  if (loaderTimeoutId) {
+    clearTimeout(loaderTimeoutId)
+    loaderTimeoutId = null
+  }
+  showLoader.value = false
+}
+
```
christian-byrne: This could be a good use-case for 

```typescript
import { useTimeoutFn } from '@vueuse/core'

const { start, stop, isPending } = useTimeoutFn(() => {
  showLoader.value = true
}, 150)
```

christian-byrne (dismissed):

christian-byrne (approved):

PR #7277, opened 2025-12-09
feat: Enable system notifications on cloud

christian-byrne:
Re-enables the system notification popup for cloud distribution, allowing cloud devs to notify cloud users about new features and updates without requiring a new release.

Cloud now fetches release notes from the "cloud" project (instead of "comfyui") and uses the `cloud_version` field for version comparison. Since cloud versions are git hashes rather than semver, a helper handles both formats gracefully.

The "What's New" popup is enabled for cloud, while the update toast and red dot indicator remain desktop-only since cloud auto-updates and doesn't require user action.

You can test this by doing `pnpm dev:cloud` and you will see a notification I added (for testing):

image

Content is controlled by non-devs at cms.comfy.org.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7277-feat-Enable-system-notifications-on-cloud-2c46d73d365081bcb33cd79ec18faefe) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/updates/common/releaseStore.ts
+++ b/src/platform/updates/common/releaseStore.ts
@@ -1,6 +1,6 @@
 import { until } from '@vueuse/core'
 import { defineStore } from 'pinia'
-import { compare } from 'semver'
+import { compare, valid } from 'semver'
 import { computed, ref } from 'vue'
 
 import { isCloud } from '@/platform/distribution/types'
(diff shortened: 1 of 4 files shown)


src/platform/updates/common/releaseStore.ts
```diff
@@ -55,22 +57,33 @@ export const useReleaseStore = defineStore('release', () => {
...
   const THREE_DAYS_MS = 3 * 24 * 60 * 60 * 1000 // 3 days
 
+  const compareVersions = (
+    releaseVersion: string,
+    currentVer: string
+  ): number => {
+    if (valid(releaseVersion) && valid(currentVer)) {
+      return compare(releaseVersion, currentVer)
+    }
+    // Non-semver (e.g. git hash): assume different = newer
```
DrJKL, line 67: That's an assumption alright

tests-ui/tests/store/releaseStore.test.ts
DrJKL, line 1: These tests seem pretty granular. Are you getting real signals from them?

DrJKL (approved):

src/platform/updates/common/releaseStore.ts
christian-byrne: Yes, I think this should work though.

tests-ui/tests/store/releaseStore.test.ts
christian-byrne: I suppose it can be helpful when testing changes given the natural number of logical combinations that semver comparison yields.

PR #7286, opened 2025-12-09
[fix] Fall back to current minor when next minor branch doesn't exist

christian-byrne:
## Summary
- When the next minor branch (e.g., `core/1.35`) doesn't exist yet, fall back to current minor (`core/1.34`) for patch releases instead of failing
- Fixes the weekly release workflow failure when ComfyUI is on version 1.33.x and `core/1.34` doesn't exist yet

## Test plan
- [x] Tested locally with version where next minor exists (1.33.5 → finds core/1.34)
- [x] Tested fallback when next minor doesn't exist (1.34.7 → falls back to core/1.34)
- [x] Tested error case when neither branch exists

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7286-fix-Fall-back-to-current-minor-when-next-minor-branch-doesn-t-exist-2c46d73d365081009762c732954e148d) by [Unito](https://www.unito.io)

### DIFF
--- a/scripts/cicd/resolve-comfyui-release.ts
+++ b/scripts/cicd/resolve-comfyui-release.ts
@@ -134,22 +134,38 @@ function resolveRelease(
 
   const [major, currentMinor, patch] = currentVersion.split('.').map(Number)
 
-  // Calculate target minor version (next minor)
-  const targetMinor = currentMinor + 1
-  const targetBranch = `core/1.${targetMinor}`
-
-  // Check if target branch exists in frontend repo
+  // Fetch all branches
   exec('git fetch origin', frontendRepoPath)
-  const branchExists = exec(
+
+  // Try next minor first, fall back to current minor if not available
+  let targetMinor = currentMinor + 1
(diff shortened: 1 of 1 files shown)


scripts/cicd/resolve-comfyui-release.ts
```diff
@@ -134,22 +134,38 @@ function resolveRelease(
...
-  const targetBranch = `core/1.${targetMinor}`
-
-  // Check if target branch exists in frontend repo
+  // Fetch all branches
   exec('git fetch origin', frontendRepoPath)
-  const branchExists = exec(
+
+  // Try next minor first, fall back to current minor if not available
+  let targetMinor = currentMinor + 1
+  let targetBranch = `core/1.${targetMinor}`
```
DrJKL, line 142: I know you could do this without having mutable variables.

DrJKL (approved):

scripts/cicd/resolve-comfyui-release.ts
christian-byrne: True

PR #7295, opened 2025-12-09
🤖 Testing section update

DrJKL:
## Summary

Standing on the shoulders of giants.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7295-Testing-section-update-2c46d73d3650818f935bcb2ac65d9830) by [Unito](https://www.unito.io)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -161,15 +161,6 @@ The project uses **Nx** for build orchestration and task management
   - Unit/Component: `**/*.test.ts`
   - E2E: `browser_tests/**/*.spec.ts`
   - Litegraph Specific: `src/lib/litegraph/test/`
-- Coverage: text/json/html reporters enabled
-  - aim to cover critical logic and new features
-- Playwright:
-  - optional tags like `@mobile`, `@2x` are respected by config
-- Tests to avoid
-  - Change detector tests
-    - e.g. a test that just asserts that the defaults are certain values
-  - Tests that are dependent on non-behavioral features like utility classes or styles
-  - Redundant tests
 
 ### General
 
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -161,7 +143,7 @@ The project uses **Nx** for build orchestration and task management
 14. Write code that is expressive and self-documenting to the furthest degree possible. This reduces the need for code comments which can get out of sync with the code itself. Try to avoid comments unless absolutely necessary
 15. Do not add or retain redundant comments, clean as you go
 16. Whenever a new piece of code is written, the author should ask themselves 'is there a simpler way to introduce the same functionality?'. If the answer is yes, the simpler course should be chosen
-17. Refactoring should be used to make complex code simpler
+17. [Refactoring](https://refactoring.com/catalog/) should be used to make complex code simpler
```
christian-byrne, line 146: > https://refactoring.com/catalog/

Wow, he generalized and refactored refactoring

DrJKL: I also like https://refactoring.guru/ but it feels like it might be less LLM friendly.

AGENTS.md
```diff
@@ -170,6 +152,42 @@ The project uses **Nx** for build orchestration and task management
...
+4. [Don’t Mock What You Don’t Own](https://hynek.me/articles/what-to-mock-in-5-mins/)
+
+### Vitest / Unit Tests
+
+1. Do not write tests that just test the mocks  
+   Ensure that the tests fail when the code itself would behave in a way that was not expected or desired
+2. For mocking, leverage [Vitest's utilities](https://vitest.dev/guide/mocking.html) where possible
+3. Keep your module mocks contained  
+   Do not use global mutable state within the test file  
+   Use `vi.hoisted()` if necessary to allow for per-test Arrange phase manipulation of deeper mock state
```
christian-byrne, line 181: > Arrange phase manipulation

What does that mean?

christian-byrne (approved):

AGENTS.md
DrJKL: [For example](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/renderer/core/layout/__tests__/TransformPane.test.ts#L67)

When Arranging the state for the test, we need to change something that would otherwise be a deep / computed property.
We don't want to have mutable module-scoped variables that we have to then wire through, we want to let the code under test access values more naturally.
So we hoist an object to contain the more flexible mock data.
Then we reference that within the module mock.
And in a given test we can tweak the value as needed.

PR #7306, opened 2025-12-10
fix: move selected groups when dragging nodes in vueNodes mode

jtydhr88:
## Summary
Captures selected groups at drag start and moves them using frame delta to match LiteGraph's behavior.

Litegraph doesn't have this issue.

### Before

https://github.com/user-attachments/assets/0e4ff907-376e-438b-aa89-106c146a8ac1

### After

https://github.com/user-attachments/assets/d954da99-3468-4bd8-9e1a-835e1a90a3bd

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7306-fix-move-selected-groups-when-dragging-nodes-in-vueNodes-mode-2c56d73d3650816a83efd11bbc36262e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
+++ b/src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
@@ -1,6 +1,7 @@
 import { storeToRefs } from 'pinia'
 import { toValue } from 'vue'
 
+import type { LGraphGroup } from '@/lib/litegraph/src/LGraphGroup'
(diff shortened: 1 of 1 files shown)


DrJKL (commented):
Would this be a good time to handle Reroutes too?

DrJKL (approved):
Approving because this is better, but I feel like there should be a way to handle all the Positionables together.

christian-byrne (approved):

PR #7322, opened 2025-12-10
Implement widget borders in vue

AustinMroz:
Adds support for displaying the "promoted" and "advanced" border indicators when in vue mode.

Requires some (hopefully minor and generally beneficial) styling changes to make sure that the widgets are contained within their border.

Note that the 'advanced' functionality sees minimal use and is likely to be revamped in the future.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7322-Implement-widget-borders-in-vue-2c56d73d36508187b881f97e373d433b) by [Unito](https://www.unito.io)


src/composables/graph/useGraphNodeManager.ts
```diff
@@ -105,17 +105,22 @@ export function safeWidgetMapper(
       }
       const spec = nodeDefStore.getInputSpecForWidget(node, widget.name)
       const slotInfo = slotMetadata.get(widget.name)
+      const borderColor =
+        (widget.promoted && 'promoted') ||
+        (widget.advanced && 'advanced') ||
+        undefined
```
DrJKL: question: Could this happen lower down?
Maybe in NodeWidgets?

DrJKL: Nit: I prefer ternary chains to short-circuit logic.
It's usually safer and feels clearer in intent.
```suggestion
      const borderColor =
        widget.promoted ? 'promoted' :
        widget.advanced ? 'advanced' :
        undefined
```

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -57,7 +57,14 @@
           :model-value="widget.value"
           :node-id="nodeData?.id != null ? String(nodeData.id) : ''"
           :node-type="nodeType"
-          class="flex-1 col-span-2"
+          :class="
+            cn(
+              'flex-1 col-span-2',
+              widget.borderColor && 'border rounded-lg',
+              widget.borderColor === 'promoted' && 'border-fuchsia-600/80',
+              widget.borderColor === 'advanced' && 'border-azure-400/80'
```
DrJKL: These should probably be theme colors.

Also, could the whole composite set of classes be joined when the widget's status is polled and just put unconditionally?

src/renderer/extensions/vueNodes/widgets/components/layout/WidgetLayoutField.vue
```diff
@@ -13,7 +13,7 @@ const hideLayoutField = inject<boolean>('hideLayoutField', false)
 
 <template>
   <div
-    class="grid grid-cols-subgrid h-7.5 min-w-0 items-center justify-between gap-1"
+    class="grid grid-cols-subgrid grid-rows-[30px] min-w-0 items-center justify-between gap-1 -ml-1 pl-1"
```
DrJKL: When would a widget comprise multiple rows?

AustinMroz: They don't. Widgets were offset because the row height wasn't being properly inherited.
image
image

src/composables/graph/useGraphNodeManager.ts
AustinMroz: I'm specifically trying to limit the runaway growth of our SimplifiedWidgets object.

I'd prefer to move in the other direction and convert things into an actual color here, but tailwind doesn't allow that.

DrJKL (dismissed):

DrJKL (approved):

PR #7324, opened 2025-12-10
Improve subscription dialog width for laptop screens

christian-byrne:
## Summary
- Increase Stripe subscription dialog width for better experience on laptop screens

When too narrow, it forces pricing options grid to go into single column layout which looks bad and should only happen when absolutely necessary (e.g,. mobile viewport).


DrJKL (approved):

DrJKL:
Pretty sure something got dropped, or the description is a test.

christian-byrne:
> Pretty sure something got dropped, or the description is a test.


Good catch.

PR #7326, opened 2025-12-10
Fix loading of subgraph blueprints on cloud

AustinMroz:
Cloud doesn't like the trailing slash when querying  directories.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7326-Fix-loading-of-subgraph-blueprints-on-cloud-2c56d73d36508136a6eae50668b15742) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/subgraphStore.ts
+++ b/src/stores/subgraphStore.ts
@@ -181,7 +181,7 @@ export const useSubgraphStore = defineStore('subgraph', () => {
     }
 
     const userSubs = (
-      await api.listUserDataFullInfo(SubgraphBlueprint.basePath)
+      await api.listUserDataFullInfo(SubgraphBlueprint.basePath.slice(0, -1))
     ).filter((f) => f.path.endsWith('.json'))
     const settled = await Promise.allSettled([
       ...userSubs.map(loadBlueprint),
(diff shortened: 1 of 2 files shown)


src/stores/subgraphStore.ts
```diff
@@ -181,7 +181,7 @@ export const useSubgraphStore = defineStore('subgraph', () => {
     }
 
     const userSubs = (
-      await api.listUserDataFullInfo(SubgraphBlueprint.basePath)
+      await api.listUserDataFullInfo(SubgraphBlueprint.basePath.slice(0, -1))
```
luke-mino-altherr: can we more explicitly remove the trailing slash, in case the SubgraphBlueprint.basePath changes in the future?

luke-mino-altherr: Can you also verify -- is this an issue for workflows as well?

AustinMroz: Probably. Slash being included in the basePath is consistent with the workflow implementation. It's just the workflow code is bad and hardcodes the `workflows` directory without the trailing slash when querying for userData.

Would updating the `listUserDataFullInfo` method in `scripts/api.ts` to strip trailing slashes be better? Would there be a time the functionality is desirable?

luke-mino-altherr: > Would updating the listUserDataFullInfo method in scripts/api.ts to strip trailing slashes be better?

This makes sense to me

luke-mino-altherr (approved):

DrJKL (approved):

AustinMroz:
Test failure is both unrelated and seems to have an incorrect expected.

PR #7332, opened 2025-12-11
add warning when using legacy mask editor (indicating it will be removed in next version)

christian-byrne:
## Summary

Telemetry data shows zero users using the legacy mask editor. It has been considerable time since we switched to the new one, and there really is no reason to use the legacy version given how lacking it is.

Will add this warning for 1 minor version just for maximum safety then remove all legacy mask editor code.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7332-add-warning-when-using-legacy-mask-editor-indicating-it-will-be-removed-in-next-version-2c66d73d365081a0bad7d63ba4d414af) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/maskeditor.ts
+++ b/src/extensions/core/maskeditor.ts
@@ -27,6 +27,9 @@ function openMaskEditor(node: LGraphNode): void {
   if (useNewEditor) {
     useMaskEditor().openMaskEditor(node)
   } else {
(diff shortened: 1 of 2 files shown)


DrJKL (dismissed):
Maybe a toast too?

trsommer (dismissed):

PR #7352, opened 2025-12-11
fix: flaky e2e test for dropping assets on nodes

christian-byrne:
Fix flaky "Can drag-and-drop animated webp image" test that was reading the widget value before the upload completed, causing intermittent failures where filenames appeared truncated. Added `waitForUpload` option to `dragAndDropFile` helper that waits for the `/upload/` response before returning. This is opt-in since not all drag-and-drop operations trigger uploads (e.g., loading workflows from media files).

### DIFF
--- a/browser_tests/fixtures/ComfyPage.ts
+++ b/browser_tests/fixtures/ComfyPage.ts
@@ -624,6 +624,14 @@ export class ComfyPage {
     // Dropping a URL (e.g., dropping image across browser tabs in Firefox)
     if (url) evaluateParams.url = url
 
+    // Set up response waiter for file uploads before triggering the drop
+    const uploadResponsePromise = fileName
+      ? this.page.waitForResponse(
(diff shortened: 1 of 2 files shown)


DrJKL (approved):

browser_tests/fixtures/ComfyPage.ts
```diff
@@ -624,6 +630,14 @@ export class ComfyPage {
     // Dropping a URL (e.g., dropping image across browser tabs in Firefox)
     if (url) evaluateParams.url = url
 
+    // Set up response waiter for file uploads before triggering the drop
+    const uploadResponsePromise = waitForUpload
+      ? this.page.waitForResponse(
+          (resp) => resp.url().includes('/upload/') && resp.status() === 200,
+          { timeout: 10000 }
```
DrJKL, line 637: I'd probably have this as a standalone method on the fixture, but I can follow up with that later.

PR #7370, opened 2025-12-11
cleanup: remove the legacy mask editor and all related code

christian-byrne:
## Summary

Removes the legacy mask editor. May also want to remove the "Beta" tags on all the current mask editor components/mentions in the app now as well.

## Context

Telemetry data shows zero users using the legacy mask editor. It has been considerable time since we switched to the new one, and there really is no reason to use the legacy version given how lacking it is.

In https://github.com/Comfy-Org/ComfyUI_frontend/pull/7332 (v1.35.2 - Dec 11, 2025), we added a final warning that the legacy mask editor is being removed.

On 1.36, this PR can be merged, as more than enough warning will have been given.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7370-cleanup-remove-the-legacy-mask-editor-and-all-related-code-2c66d73d365081d58fbed0f3c84bcb0d) by [Unito](https://www.unito.io)


christian-byrne:
We added the warning for users a few releases ago, and ensured from telemetry data on cloud that no one opts into using the legacy version.

For devs, we added the deprecation warning a couple months ago and as of today there is [0 usage from custom nodes per source graph](https://cs.comfy.org/search?q=context:global+maskEditorOld&patternType=keyword&sm=0).

This should be good to merge now (on 1.36, as promised).

PR #7374, opened 2025-12-11
refactor: replace runtime `isElectron` with buildtime `isDesktop`

christian-byrne:
## Summary

Dependent on https://github.com/Comfy-Org/desktop/pull/1479

Whether the environment is desktop or not is known at build time and will not be dynamic at runtime, so we can use this pattern instead. The benefits are:

- Allowing DCR and tree-shaking
- Allow better minifying
- More performant
- Less error-prone as it doesn't rely on potentially changeable logic of `'electronAPI' in window && window.electronAPI !== undefined`
- Matches pattern used for cloud, ensuring there's only a single way to do this type of thing going forward

Evidence of improved DCR and tree-shaking:

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7374-refactor-replace-runtime-isElectron-with-buildtime-isDesktop-2c66d73d365081edbbcafdce5fe76eba) by [Unito](https://www.unito.io)


christian-byrne:
Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/8710

PR #7399, opened 2025-12-12
Cleanup app.graph usage

AustinMroz:
Prior to the release of subgraphs, there was a single graph accessed through `app.graph`. Now that there's multiple graphs, there's a lot of code that needs to be reviewed and potentially updated depending on if it cares about nearby nodes, all nodes, or something else requiring specific attention.

This was done by simply changing the type of `app.graph` to unknown so the typechecker will complain about every place it's currently used. References were then updated to `app.rootGraph` if the previous usage was correct, or actually rewritten.

By not getting rid of `app.graph`, this change already ensures that there's no loss of functionality for custom nodes, but the prior typing of `app.graph` can always be restored if future dissuasion of `app.graph` usage creates issues.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7399-Cleanup-app-graph-usage-2c76d73d365081178743dfdcf07f44d0) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetRecordAudio.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetRecordAudio.vue
@@ -157,8 +157,8 @@ const isWaveformActive = computed(() => isRecording.value || isPlaying.value)
 const modelValue = defineModel<string>({ default: '' })
 
 const litegraphNode = computed(() => {
-  if (!props.nodeId || !app.rootGraph) return null
-  return app.rootGraph.getNodeById(props.nodeId) as LGraphNode | null
+  if (!props.nodeId || !app.canvas.graph) return null
+  return app.canvas.graph.getNodeById(props.nodeId) as LGraphNode | null
 })
 
 async function handleRecordingComplete(blob: Blob) {
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -1,4 +1,3 @@
-import type { LGraph } from '@/lib/litegraph/src/litegraph'
 import type {
   DeviceStats,
   EmbeddingsResponse,
@@ -71,6 +70,6 @@ declare global {
     app?: ComfyApp
 
     /** For use by extensions and in the browser console. Where possible, import `app` and access via `app.graph` instead. */
-    graph?: LGraph
+    graph?: unknown
   }
 }
--- a/tests-ui/tests/composables/useMissingNodes.test.ts
+++ b/tests-ui/tests/composables/useMissingNodes.test.ts
@@ -468,7 +468,7 @@ describe('useMissingNodes', () => {
     it('calls collectAllNodes with the app graph and filter function', () => {
       const mockGraph = { nodes: [], subgraphs: new Map() }
       // @ts-expect-error - Mocking app.graph for testing
-      app.graph = mockGraph
+      app.rootGraph = mockGraph
 
       const { missingCoreNodes } = useMissingNodes()
       // Access the computed to trigger the function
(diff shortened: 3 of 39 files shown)


tests-ui/tests/composables/useMissingNodes.test.ts
```diff
@@ -586,7 +586,7 @@ describe('useMissingNodes', () => {
       }
 
       // @ts-expect-error - Mocking app.graph for testing
```
DrJKL: What is the type error here?

tests-ui/tests/composables/useMissingNodes.test.ts
```diff
@@ -468,7 +468,7 @@ describe('useMissingNodes', () => {
     it('calls collectAllNodes with the app graph and filter function', () => {
       const mockGraph = { nodes: [], subgraphs: new Map() }
```
DrJKL, line 459: Does this work?
```suggestion
      const mockGraph: LGraph = { nodes: [], subgraphs: new Map() } as Partial<LGraph>
```

tests-ui/tests/renderer/extensions/vueNodes/components/NodeHeader.subgraph.test.ts
```diff
@@ -58,9 +58,9 @@ describe('NodeHeader - Subgraph Functionality', () => {
     const { app } = await import('@/scripts/app')
 
     if (hasGraph) {
-      ;(app as any).graph = { rootGraph: {} }
+      ;(app as any).rootGraph = { rootGraph: {} }
     } else {
-      ;(app as any).graph = null
+      ;(app as any).rootGraph = null
```
DrJKL: ...
😢

DrJKL (dismissed):
Beautiful.

src/types/index.ts
```diff
@@ -71,6 +70,6 @@ declare global {
     app?: ComfyApp
 
     /** For use by extensions and in the browser console. Where possible, import `app` and access via `app.graph` instead. */
-    graph?: LGraph
+    graph?: unknown
```
DrJKL, line 73: Want to mark it as `@deprecated`?

tests-ui/tests/composables/useMissingNodes.test.ts
AustinMroz: Both app.graph and app.rootGraph are typed to be only getters, so setting would be illegal if it wasn't a mock instance?

tests-ui/tests/composables/useMissingNodes.test.ts
AustinMroz: Not quite, but I took a different approach that's much less `as`

(Regretfully, LGraphNode trickier: You can't do `{} as Partial<LGraphNode>` and creating actual LGraphNode instances creates import errors here)

tests-ui/tests/renderer/extensions/vueNodes/components/NodeHeader.subgraph.test.ts
AustinMroz: Save those tears for the autogrow PR.

src/renderer/extensions/vueNodes/widgets/components/WidgetRecordAudio.vue
```diff
@@ -157,8 +157,8 @@ const isWaveformActive = computed(() => isRecording.value || isPlaying.value)
 const modelValue = defineModel<string>({ default: '' })
 
 const litegraphNode = computed(() => {
-  if (!props.nodeId || !app.rootGraph) return null
-  return app.rootGraph.getNodeById(props.nodeId) as LGraphNode | null
+  if (!props.nodeId || !app.canvas.graph) return null
+  return app.canvas.graph.getNodeById(props.nodeId) as LGraphNode | null
 })
```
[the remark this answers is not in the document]
christian-byrne: If that was true, wouldn't this trigger one of our linting rules for unneccessary type assertion? Not sure tbh.

christian-byrne (approved):
Nice, good progress

PR #7403, opened 2025-12-12
fix: correct grammar in pricing table description

christian-byrne:
## Summary

Change "amount" to "number," as "video" is a countable noun.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7403-fix-correct-grammar-in-pricing-table-description-2c76d73d36508144828fd060b046c1a5) by [Unito](https://www.unito.io)


DrJKL (approved):
Are we saving that many characters with Approx. vs Approximate?

christian-byrne:
> Are we saving that many characters with Approx. vs Approximate?

Valid point

PR #7410, opened 2025-12-12
ci: make nightly release happen automatically every night

christian-byrne:
## Summary

Makes the existing "Release: Version Bump" workflow run at 00:00 UTC every day.

## Details

- concurrency keeps only one run active while manual dispatch remains available for ad-hoc bumps.
- inputs are normalized inside the workflow so scheduled runs (which lack `workflow_dispatch` inputs) safely fall back to `patch`/`main`, and the version bump + PR formatting steps only use the optional `pre_release` flag when it is provided
- each nightly invocation closes any lingering bot-authored `version-bump-*` PRs/branches before creating a new patch PR, preventing stale locale bumps from conflicting
- checkout now disables credential persistence and `pnpm/action-setup` is pinned to a commit for supply-chain safety.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7410-ci-make-nightly-release-happen-automatically-every-night-2c76d73d3650813a9e02ee8878370929) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/release-version-bump.yaml
+++ b/.github/workflows/release-version-bump.yaml
@@ -20,6 +20,12 @@ on:
         required: true
         default: 'main'
         type: string
+  schedule:
+    - cron: '0 0 * * *'
+
+concurrency:
+  group: release-version-bump
+  cancel-in-progress: true
 
 jobs:
   bump-version:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-version-bump.yaml
```diff
@@ -29,15 +36,99 @@ jobs:
...
+                continue
+              }
+
+              if (pr.user?.login !== 'github-actions[bot]') {
+                continue
+              }
+
+              // Only clean up stale nightly PRs targeting main.
+              // Adjust here if other target branches should be cleaned.
+              if (pr.base?.ref !== 'main') {
```
DrJKL, line 92: Optional:
I think you could filter the original query with this.
https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests

DrJKL (approved):

Issue #6561, opened 2025-11-03
Fail to load workflow from png saved with some custom nodes.


49RpK5dY:
Dragging or directly opening images saved with some custom nodes fails to load embedded workflow on 1.32.1. 

<img width="4920" height="1900" alt="Image" src="https://github.com/user-attachments/assets/663ec0be-8c4a-4b46-9ba5-52662ea96d1e" />

[test.saver.json](https://github.com/user-attachments/files/23318503/test.saver.json)

Tested https://github.com/wallish77/wlsh_nodes and https://github.com/alexopus/ComfyUI-Image-Saver
Trying to open first saved image results in new workflow but it is just a copy of the one that was already open. The second one builds a new workflow from scratch based on metadata instead of opening embedded workflow. Third one works as intended.

Dragging images saved on 1.32.1 in 1.32.0 work just fine.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6561-Fail-to-load-workflow-from-png-saved-with-some-custom-nodes-2a06d73d365081728d27e107d8c4a0fd) by [Unito](https://www.unito.io)

49RpK5dY:
[@dosu](https://go.dosu.dev/dosubot) It seem whatever was reverted did not solve the issue. 1.32.2 still fails to load workflow from images saved with custom nodes.

49RpK5dY:
Since #6464 has been merged into stable does this mean this is now intended behavior and there are no plans to make loading workflows from images backwards compatible?

Sarashana:
It's not just a backwards compatibility issue. The newest frontend fails to load workflows with custom nodes from images in the described manner, even when the images were created using frontend version 1.34.0.

49RpK5dY:
Yeah. What I meant was there is no easy way to load workflow from dozens of GB of old images saved over the years. But there is still a default `Save Image` node for newly created ones if this will not get fixed.

PR #7411, opened 2025-12-12
fix: loading api-format workflow that contains "parameters" string

### LINKED ISSUE #6988: Adjust the logic in handleFile function of app.ts
### Prerequisites

### What happened?

In the current logic, the `parameters` are positioned too early. This causes images saved using the `Image Saver` node with A1111 compatibility to fail to load the workflow correctly. Since the code first detects the presence of the `parameters`, it attempts to read them in A1111 compatibility mode instead of using the `prompt` correctly to load the workflow.

It is recommended that the priority of `parameters` is lowered to the very end so that compatibility mode is only used for evaluation if all preceding checks fail.

I traced back through the versions and discovered that, although this issue has probably been present for long time, bug it appeared after today I updated to latest ComfyUI.

https://github.com/Comfy-Org/ComfyUI_frontend/blob/c57ceaf8265fd9b7f6b180967652cdbef8722c27/src/scripts/app.ts#L1421

```
  async handleFile(file: File, openSource?: WorkflowOpenSource) {
    const fileName = file.name.replace(/\.\w+$/, '') // Strip file extension
    const workflowData = await getWorkflowDataFromFile(file)
    if (!workflowData) {
      return
    }

    const { workflow, prompt, parameters, templates } = workflowData

    if (templates) {
      this.loadTemplateData({ templates })
    }

    if (workflow) {
      const workflowObj =
        typeof workflow === 'string' ? JSON.parse(workflow) : workflow
      await this.loadGraphData(workflowObj, true, true, fileName, {
        openSource
      })
      return
    }

    if (prompt) {
      const promptObj = typeof prompt === 'string' ? JSON.parse(prompt) : prompt
      this.loadApiJson(promptObj, fileName)
      return
    }
	
	if (parameters) {
      // Note: Not putting this in `importA1111` as it is mostly not used
      // by external callers, and `importA1111` has no access to `app`.
      useWorkflowService().beforeLoadNewGraph()
      importA1111(this.graph, parameters)
      useWorkflowService().afterLoadNewGraph(
        fileName,
        this.graph.serialize() as unknown as ComfyWorkflowJSON
      )
      return
    }

  }
```

### Steps to Reproduce

1. Update to latest start comfyui
2. Load `2025-11-27-150211_3065179543.png` in to comfyui
3. Shows the A1111 workflow
4. Load `2025-11-27-150211_3065179543 - Copy.png` which has simply change `parameters` in `tExt` to `marameters`
5. Shows the correct workflow

Image

Image

# Incorrect Workflow
2025-11-27-150211_3065179543.png

Image

# Correct Workflow
2025-11-27-150211_3065179543 - Copy.png

Image

### ComfyUI Frontend Version

1.34.0

### Browser

Edge

### Console Errors

```javascript

```

### Logs

```shell
System Info

Checkpoint files will always be loaded safely.
Total VRAM 24576 MB, total RAM 130756 MB
pytorch version: 2.9.0+cu130
Enabled fp16 accumulation.
Set vram state to: NORMAL_VRAM
Device: cuda:0 NVIDIA TITAN RTX : cudaMallocAsync
Enabled pinned memory 58840.0
working around nvidia conv3d memory bug.
Using sage attention
Python version: 3.13.9 (tags/v3.13.9:8183fa5, Oct 14 2025, 14:09:13) [MSC v.1944 64 bit (AMD64)]
ComfyUI version: 0.3.75
ComfyUI frontend version: 1.34.0
[Prompt Server] web root: C:\Users\oOkami\AppData\Roaming\Python\Python313\site-packages\comfyui_frontend_package\static
Total VRAM 24576 MB, total RAM 130756 MB
pytorch version: 2.9.0+cu130
Enabled fp16 accumulation.
Set vram state to: NORMAL_VRAM
Device: cuda:0 NVIDIA TITAN RTX : cudaMallocAsync
Enabled pinned memory 58840.0
```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6988-Adjust-the-logic-in-handleFile-function-of-app-ts-2b86d73d365081c1966eebaf04bc359b) by [Unito](https://www.unito.io)

### LINKED ISSUE #7096: After updating ComfyUI, loading old workflows by dragging in an image no longer works. Even worse, the two images produce completely different behaviors
### Prerequisites

### What happened?

Normal Version: Core 0.3.75, Backend 1.30.6
Problem Version: Core 0.3.76, Backend 1.33.10

Issue Description:
After updating ComfyUI, loading old workflows by dragging in an image no longer works.
Even worse, the two images produce completely different behaviors:

510584052853891_1-底图00001 → loads as a completely blank workflow (nothing appears)

02-59-06-1056516826837278_底图00001 → loads, but the workflow becomes extremely simplified compared to the original

510584052853891_1-BaseImage_00001_.json — the original workflow file

Image
Image

[510584052853891_1-底图_00001_.json](https://github.com/user-attachments/files/23880526/510584052853891_1-._00001_.json)

### Steps to Reproduce

1.
2.
3.

Visual/UI issue only

### ComfyUI Frontend Version

1.33.10

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-7096-After-updating-ComfyUI-loading-old-workflows-by-dragging-in-an-image-no-longer-works--2bd6d73d365081a9aa44eba251f66736) by [Unito](https://www.unito.io)

christian-byrne:
## Summary

This change extends https://github.com/Comfy-Org/ComfyUI_frontend/pull/7154 by making sure the `prompt` metadata tag is parsed before the legacy A1111 fallback when files are dropped onto the canvas.

ComfyUI embeds two structured payloads into every first-class export format we support (PNG, WEBP, WEBM, MP4/MOV/M4V, GLB, SVG, MP3, OGG/FLAC, etc.): `workflow`, which is the full editor JSON with layout state, and `prompt`, which is the API graph sent to `/prompt`.

During import we try format-specific decoders first and only as a last resort look for an A1111 file by scanning text chunks for a `parameters` entry. That compatibility path was always meant to be a best-effort option, but when we refactored the loader it accidentally enforced the order `workflow → parameters → prompt`. As soon as a dropped asset contained a `parameters` chunk—something Image Saver’s “A1111 compatibility” mode always adds—the A1111 converter activated and blocked the subsequent `prompt` loading logic.

PR #7154 already lifted `workflow` ahead of the fallback, yet any file lacking the `workflow` chunk but holding both `prompt` and `parameters` still regressed. Reordering to `workflow → prompt → parameters` preserves the compatibility shim for genuine A1111 exports while guaranteeing native Comfy metadata always wins, eliminating the entire class of failures triggered merely by the presence of the word `parameters` in an unrelated metadata chunk.

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/7096, fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/6988

## Related 

(fixed by https://github.com/Comfy-Org/ComfyUI_frontend/pull/7154)

- https://github.com/Comfy-Org/ComfyUI_frontend/issues/6633
- https://github.com/Comfy-Org/ComfyUI_frontend/issues/6561


browser_tests/assets/workflowInMedia/workflow_prompt_parameters.png
DrJKL, line 1: What is this, an image for ants?

src/scripts/app.ts
```diff
@@ -1468,7 +1468,21 @@ export class ComfyApp {
       }
     }
 
-    // Use parameters as fallback when no workflow exists
+    if (prompt) {
+      try {
+        const promptObj =
+          typeof prompt === 'string' ? JSON.parse(prompt) : prompt
+        if (this.isApiJson(promptObj)) {
+          this.loadApiJson(promptObj as ComfyApiWorkflow, fileName)
```
DrJKL: Should `isApiJson` `assert` that `promptObj is ComfyApiWorkflow`?

DrJKL (dismissed):

src/scripts/app.ts
christian-byrne: Great point, yes

PR #7416, opened 2025-12-12
fix: "Copy Image", "Open Image", "Save Image", "Open in Mask Editor" missing from context menu on SaveImage vue node

christian-byrne:
## Summary

| Before                                                                                                                                        | After                                                                                                                                         |
| --------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
| Screenshot from 2025-12-12 05-22-16 | Screenshot from 2025-12-12 05-21-10 |

Fixes the "Copy Image" context menu option not working in Vue Nodes (Nodes 2.0). The issue was that `node.imgs` wasn't being populated in Vue nodes like it is in legacy nodes causing it to not include the image context menu options properly.

Added `syncNodeImgs()` to `ImagePreview.vue` that syncs the image element to `node.imgs` when images load, enabling backwards compatibility with legacy systems that rely on this property. Also simplified `setupNodeForMaskEditor()` since the sync now happens automatically. Added unit test for the sync behavior and browser test for mask editor button functionality.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7416-fix-Copy-Image-Open-Image-Save-Image-Open-in-Mask-Editor-missing-from-context-2c76d73d365081b8936fcaade7d2b15a) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -197,6 +197,17 @@ watch(
...
+/**
+ * Sync node.imgs for backwards compatibility with legacy systems (e.g., Copy Image).
+ */
+const syncNodeImgs = () => {
+  if (!props.nodeId || !currentImageEl.value) return
+  const node = app.rootGraph?.getNodeById(props.nodeId)
+  if (!node) return
+  node.imgs = [currentImageEl.value]
+  node.imageIndex = currentIndex.value
+}
```
christian-byrne: Realizing now that this is not ideal. Managing the state of the nodes should not be responsibility of a component in the view layer - should be controlled at higher level in the business logic of the renderer.

Should be easy to move this into the nodeOutputsStore.

DrJKL: @coderabbitai What would that look like?

tests-ui/tests/renderer/extensions/vueNodes/components/ImagePreview.test.ts
```diff
@@ -308,4 +313,27 @@ describe('ImagePreview', () => {
     expect(imgElement.exists()).toBe(true)
     expect(imgElement.attributes('alt')).toBe('Node output 2')
   })
+
+  it('syncs node.imgs on image load for legacy compatibility', async () => {
+    // Reset mock node state
+    delete mockNode.imgs
+    delete mockNode.imageIndex
```
DrJKL: Just noting for later, this can be an issue if a second test tried to depend on the `mockNode`. They'd become order dependent immediately.

browser_tests/tests/vueNodes/maskEditor.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    // Close mask editor
+    await comfyPage.page.keyboard.press('Escape')
+    await expect(maskEditorDialog).not.toBeVisible()
+
+    // Test 2: Open from image overlay button
+    const imageWrapper = comfyPage.page.locator('.image-preview [role="img"]')
+    await imageWrapper.hover()
+
+    const overlayMaskButton = comfyPage.page.locator(
+      '.image-preview [aria-label="Edit or mask image"]'
```
DrJKL: We could use [getByLabel](https://playwright.dev/docs/locators#locate-by-label) here

DrJKL (dismissed):

browser_tests/tests/nodeSearchBox.spec.ts-snapshots/added-node-no-connection-chromium-linux.png
DrJKL, line 1: Seems unrelated, not sure what's up here.

src/stores/imagePreviewStore.ts
```diff
@@ -2,6 +2,7 @@ import { useTimeoutFn } from '@vueuse/core'
 import { defineStore } from 'pinia'
 import { ref } from 'vue'
 
+import { LiteGraph } from '@/lib/litegraph/src/litegraph'
 import type { LGraphNode, SubgraphNode } from '@/lib/litegraph/src/litegraph'
 import { useWorkflowStore } from '@/platform/workflow/management/stores/workflowStore'
```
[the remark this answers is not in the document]
christian-byrne: This is actually a linting rule we have that enforces this.

christian-byrne:
There's issues here with redundancy and the test not working, still working on it.

christian-byrne:
Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/8143

PR #7434, opened 2025-12-13
fix: pricing table links to wrong page in docs (p2)

christian-byrne:
## Summary

Continuation of updating pricing table link (p1 was https://github.com/Comfy-Org/ComfyUI_frontend/pull/7402). Create a constant since it's now used in multiple locations.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7434-fix-pricing-table-links-to-wrong-page-in-docs-p2-2c86d73d36508199b320d49faa6a0c73) by [Unito](https://www.unito.io)


DrJKL (approved):
Agree about the `as const`, not as sure about the extra `type`

AustinMroz (approved):

PR #7435, opened 2025-12-13
feat: when restored position has no nodes in viewport, automatically fit to view

christian-byrne:
## Summary

Sometimes the saved position is super far away from any of the nodes, which causes general confusion. This PR changes the `loadGraphData` logic to fit-to-view in those scenarios.

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/7425

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7435-feat-when-restored-position-has-no-nodes-in-viewport-automatically-fit-to-view-2c86d73d36508119bf2ed9d361ec868f) by [Unito](https://www.unito.io)


browser_tests/tests/viewport.spec.ts-snapshots/viewport-fits-when-saved-offscreen-chromium-linux.png
christian-byrne, line 1: This workflow's viewport data is way down to the right with no nodes in view. This snapshot verified the fit-to-view happened automatically.

AustinMroz (approved):
I'm assuming canvas.computeVisibleNodes isn't used because `node.boundingRect` isn't initialized yet?

I suppose that creates an edge case where a single node title barely displaying off the bottom of the screen will still trigger fit, but that's not worth solving.

LGTM

PR #7447, opened 2025-12-13
fix: inner groups being moved double when moving outer group (in vue mode)

christian-byrne:
## Summary

Fixes issue when dragging a group that had inner groups when in vue mode.

When dragging the outer group in Vue mode:

1. getAllNestedItems(selected) returns ALL items: outer group + inner groups + nodes
2. moveChildNodesInGroupVueMode loops through all items
3. For outer group G1: calls G1.move(delta, true) then moveGroupChildren(G1, ...)
4. moveGroupChildren calls G2.move(delta) (no skipChildren) - this moves G2 AND G2's children!
5. Then the loop reaches G2: calls G2.move(delta, true) - moves G2 again
6. Plus moveGroupChildren(G2, ...) processes G2's children again

This PR fixes it by adding `skipChildren=true` to the `move` call.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7447-fix-inner-groups-being-moved-double-when-moving-outer-group-in-vue-mode-2c86d73d365081ce97abec682f2a8518) by [Unito](https://www.unito.io)


browser_tests/tests/vueNodes/groups/groups.spec.ts
```diff
@@ -32,4 +32,60 @@ test.describe('Vue Node Groups', () => {
...
+          outerGroup.pos[0] + 50,
+          outerGroup.pos[1] + 15
+        ])
+
+        return {
+          outer: { x: outerGroup.pos[0], y: outerGroup.pos[1] },
+          inner: { x: innerGroup.pos[0], y: innerGroup.pos[1] },
+          screenPos: { x: screenPos[0], y: screenPos[1] }
+        }
+      })
```
DrJKL: Nit, maybe for later:
This kind of logic is good to put into the Fixture/Page Object and expose via a more usery method.
Like:
```ts
expect(comfyPage.getGroupBounds('Outer Group')).toBe(...)

await comfyPage.dragGroup({name: 'Outer Group', deltaX: 100, deltaY: 80})

expect(comfyPage.getGroupBounds('Outer Group')).toBe(...)
```

DrJKL: That way the test reads like how you'd describe the behavior to a user (or they'd describe the expectations/bug to you).

DrJKL (dismissed):

PR #7448, opened 2025-12-13
WIP: add graph state store and use to propagate execution and missing nodes errors up subgraphs

christian-byrne:
┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7448-WIP-add-graph-state-store-and-use-to-propagate-execution-and-missing-nodes-errors-up-sub-2c86d73d365081888618e19d2096c5b5) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/graph/state/graphStateStore.ts
+++ b/src/core/graph/state/graphStateStore.ts
@@ -3,23 +3,23 @@ import { customRef } from 'vue'
 
 import type { NodeLocatorId } from '@/types/nodeIdentification'
 
-export interface NodeState {
+interface NodeState {
   hasError: boolean
 }
 
-export interface SetNodeErrorCommand {
+interface SetNodeErrorCommand {
   type: 'SetNodeError'
   version: 1
   nodeId: NodeLocatorId
   hasError: boolean
 }
 
-export interface ClearAllErrorsCommand {
+interface ClearAllErrorsCommand {
   type: 'ClearAllErrors'
   version: 1
 }
 
-export type GraphStateCommand = SetNodeErrorCommand | ClearAllErrorsCommand
+type GraphStateCommand = SetNodeErrorCommand | ClearAllErrorsCommand
 
 export const useGraphStateStore = defineStore('graphState', () => {
   const nodes = new Map<NodeLocatorId, NodeState>()
--- a/src/core/graph/state/graphStateStore.ts
+++ b/src/core/graph/state/graphStateStore.ts
@@ -1,0 +1,79 @@
+import { defineStore } from 'pinia'
+import { customRef } from 'vue'
+
+import type { NodeLocatorId } from '@/types/nodeIdentification'
+
+export interface NodeState {
+  hasError: boolean
+}
+
+export interface SetNodeErrorCommand {
+  type: 'SetNodeError'
+  version: 1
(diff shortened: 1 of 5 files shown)


christian-byrne:
@corerabbitai review

christian-byrne:
@coderabbitai review

src/core/graph/state/graphStateStore.ts
```diff
@@ -0,0 +1,79 @@
...
+  const stateRef = customRef<number>((track, trigger) => ({
+    get() {
+      track()
+      return revision
+    },
+    set() {
+      revision++
+      trigger()
+    }
+  }))
```
DrJKL, line 37: Question: Could you check to see if this works the same? 
The setter is currently ignoring the value passed in on line 57, right?
```suggestion
  const stateRef = ref(0)
```

src/core/graph/state/graphStateStore.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  const execute = (command: GraphStateCommand): void => {
+    switch (command.type) {
+      case 'SetNodeError': {
+        const existing = nodes.get(command.nodeId)
```
DrJKL, line 42: nit: This could technically use `getNodeState` (though it'd have to be declared earlier since it isn't using a function declaration 😉) 

Though what are the odds we change how we access a specific node state?

src/core/graph/state/graphStateStore.ts
```diff
@@ -0,0 +1,79 @@
...
+  const getNodeState = (nodeId: NodeLocatorId): NodeState | undefined => {
+    return nodes.get(nodeId)
+  }
+
+  const getNodesWithErrors = (): NodeLocatorId[] => {
+    const result: NodeLocatorId[] = []
+    for (const [nodeId, state] of nodes) {
+      if (state.hasError) result.push(nodeId)
+    }
+    return result
```
DrJKL, line 69: Very nit: Could be a single line with a `reduce`

src/core/graph/state/useGraphErrorState.ts
```diff
@@ -0,0 +1,45 @@
...
+    subgraphNode.has_errors = true
+    propagateErrorToParents(subgraphNode)
+  })
+}
+
+export const useGraphErrorState = () => {
+  const store = useGraphStateStore()
+
+  watch(
+    () => store.stateRef,
```
DrJKL, line 27: Optional: You _could_ use `storeToRefs` for this.

src/stores/executionStore.ts
```diff
@@ -574,56 +575,54 @@ export const useExecutionStore = defineStore('execution', () => {
...
-          if (!slotName) continue
-
-          const slot = node.inputs.find((s) => s.name === slotName)
-          if (slot) {
-            slot.hasErrors = true
-          }
-        }
+      const locatorId = executionIdToNodeLocatorId(executionId)
+      if (locatorId) {
+        graphStateStore.execute({
```
DrJKL, line 605: This would be a good place to have a facade.
Something like `graphStateStore.setNodeError(locatorId)` so that callers don't need to track internal details.

christian-byrne:
Closing, although we still need a graph state store.

PR #7457, opened 2025-12-13
Support display of multitype slots

AustinMroz:
Example with forcibly modified types for testing
image

Vue mode doesn't currently seem to display optional inputs, but the SVGs here include support for being made hollow with `--shape: url(#hollow)`
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7457-Support-display-of-multitype-slots-2c86d73d3650818594afd988e73827e3) by [Unito](https://www.unito.io)


src/lib/litegraph/src/node/NodeSlot.ts
```diff
@@ -33,6 +33,7 @@ export interface IDrawOptions {
 /** Shared base class for {@link LGraphNode} input and output slots. */
 export abstract class NodeSlot extends SlotBase implements INodeSlot {
   pos?: Point
+  rotation = -Math.PI / 2
```
DrJKL: If this is a constant, can it be either a `readonly static` or a `CONSTANT` in the module scope?

src/lib/litegraph/src/node/NodeSlot.ts
```diff
@@ -163,23 +165,60 @@ export abstract class NodeSlot extends SlotBase implements INodeSlot {
...
+          ctx.strokeStyle = 'black'
+          ctx.lineWidth = 0.5
+          types.forEach((_, idx) => {
+            ctx.moveTo(pos[0], pos[1])
+            const xOffset = Math.cos(arcLen * idx + this.rotation) * radius
+            const yOffset = Math.sin(arcLen * idx + this.rotation) * radius
+            ctx.lineTo(pos[0] + xOffset, pos[1] + yOffset)
+          })
+          ctx.stroke()
+          ctx.restore()
```
DrJKL, line 213: Nit: Instead of a comment, how about extracting a named function?
(for this, and also the ones above)

src/renderer/extensions/vueNodes/components/SlotConnectionDot.vue
```diff
@@ -1,17 +1,40 @@
...
+  const types = [...typesSet].slice(0, 3)
+  if (types.length === 1) return { backgroundColor: types[0] }
+  const angle = 360 / types.length
+  const slices = types.map(
+    (type, idx) => `${type} ${angle * idx}deg ${angle * (idx + 1)}deg`
+  )
+  return {
+    background: `conic-gradient(${slices.join(',')})`,
+    backgroundOrigin: 'border-box'
+  }
```
DrJKL: Nit: Same as above. This is fine as is, but extracting a utility could help make the intent and scope clearer (and also make an obvious target for future related changes).

src/renderer/extensions/vueNodes/components/SlotConnectionDot.vue
```diff
@@ -1,17 +1,40 @@
...
+}
+const style = getStyle()
```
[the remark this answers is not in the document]
DrJKL: @coderabbitai wouldn't updating the template binding as suggested be redundant given the shorthand syntax in Vue 3.5?

AustinMroz:
Going to try and cleanup the vue implementation on this one by moving to SVGs. For posterity, my very rough attempt at dividing lines with conic gradient
image

src/lib/litegraph/src/node/NodeSlot.ts
AustinMroz: Tragic to see the door to spinning closed off, but YAGNI is an impartial arbiter

DrJKL: Closed off _for now_ ;-)

src/renderer/extensions/vueNodes/components/SlotConnectionDot.vue
```diff
@@ -27,19 +52,26 @@ defineExpose({
...
-          'bg-slate-300 rounded-full',
-          'transition-all duration-150',
-          'border border-solid border-node-component-slot-dot-outline',
-          !multi &&
-            'cursor-crosshair group-hover/slot:[--node-component-slot-dot-outline-opacity-mult:5] group-hover/slot:scale-125',
-          multi ? 'w-3 h-6' : 'size-3'
-        )
-      "
+      :style="{ backgroundColor: types[0] }"
+      :class="slotClass"
```
DrJKL, line 58: Why was this pulled out of the template?

AustinMroz: It's used twice. I wanted to avoid the extra lines and the chance that the two copies become desynced.

I'm not 100% happy with how the border outline or scaling of optional multi-type inputs work and would like to eventually separate out/re-inline the class, but the amount of extra work this would require is not something I can justify right now.

DrJKL (approved):

PR #7459, opened 2025-12-13
Add functionality to quickly disconnect moved input links

AustinMroz:
Disconnections are frequently performed by dragging a link from an input slot and dropping it on the canvas, but needing to wait for the searchbox to pop up, and then needing to manually close out of this can make it feel slow. Sometimes, this will even result in users disabling the link release action for more responsive graph building.

Instead, this PR introduces new functionality where a link which is moved only a short distance from a node input and dropped will be immediately disconnected instead of performing the default link release action.
fast-disco_00001

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7459-Add-functionality-to-quickly-disconnect-moved-input-links-2c86d73d365081919052f3856db8e672) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -5015,6 +5015,7 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
           }
         )
       }
+      if (renderLink instanceof MovingInputLink) this.setDirty(false, true)
 
       ctx.fillStyle = colour
       ctx.beginPath()
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -2,7 +2,7 @@ import { toString } from 'es-toolkit/compat'
 import { toValue } from 'vue'
 
 import { PREFIX, SEPARATOR } from '@/constants/groupNodeConstants'
-import { MovingLinkBase } from '@/lib/litegraph/src/canvas/MovingLinkBase'
+import { MovingInputLink } from '@/lib/litegraph/src/canvas/MovingInputLink'
 import { LitegraphLinkAdapter } from '@/renderer/core/canvas/litegraph/litegraphLinkAdapter'
 import type { LinkRenderContext } from '@/renderer/core/canvas/litegraph/litegraphLinkAdapter'
 import { getSlotPosition } from '@/renderer/core/canvas/litegraph/slotCalculations'
@@ -5015,8 +5015,8 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
           }
         )
       }
-      if (renderLink instanceof MovingLinkBase)
-        renderLink.drawConnectionCircle(ctx)
+      if (renderLink instanceof MovingInputLink)
+        renderLink.drawConnectionCircle(ctx, this.graph_mouse)
 
       ctx.fillStyle = colour
       ctx.beginPath()
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -5016,7 +5016,7 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
         )
       }
       if (renderLink instanceof MovingInputLink)
-        renderLink.drawConnectionCircle(ctx, this.graph_mouse)
+        renderLink.drawConnectionCircle(ctx, highlightPos)
 
       ctx.fillStyle = colour
       ctx.beginPath()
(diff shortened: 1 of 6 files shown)


src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -4769,6 +4770,8 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
               }
             )
           }
+          if (renderLink instanceof MovingInputLink)
+            renderLink.drawConnectionCircle(ctx, this.graph_mouse)
```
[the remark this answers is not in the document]
DrJKL: Is the context state already saved and restored inside the method?

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -1,6 +1,7 @@
 import { toString } from 'es-toolkit/compat'
 
 import { PREFIX, SEPARATOR } from '@/constants/groupNodeConstants'
+import { MovingInputLink } from '@/lib/litegraph/src/canvas/MovingInputLink'
```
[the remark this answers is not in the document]
DrJKL: Lies.

src/renderer/core/canvas/links/linkConnectorAdapter.ts
```diff
@@ -72,11 +73,20 @@ export class LinkConnectorAdapter {
   beginFromInput(
     nodeId: NodeId,
     inputIndex: number,
-    opts?: { moveExisting?: boolean; fromRerouteId?: RerouteId }
+    opts?: {
+      moveExisting?: boolean
+      fromRerouteId?: RerouteId
+      layout: SlotLayout
+    }
```
DrJKL, line 81: Nit: Required before Optional
```suggestion
    opts?: {
      layout: SlotLayout
      moveExisting?: boolean
      fromRerouteId?: RerouteId
    }
```

without-ordinary:
Stupid question, doesn't ctrl+alt+click doing this already?

AustinMroz:
Yes. The functionality is similar and ctrl+alt+click is not being removed.

This functionality is
- More discoverable and doesn't require remembering specific modifiers
- Doesn't require a keyboard
- 0 cost. Unless you're disconnecting a node and creating a new one directly on top, functionality is unchanged.

But I do value your opinon. Is there a problem I've missed?

without-ordinary:
> * More discoverable and doesn't require remembering specific modifiers

Unless the user doesn't read patch notes and is expecting the default setting of `Action on link release (No modifier) ` to be followed.

If this is an opt-in behavior that will be added under the existing `Link Release` settings or will have a clear visual cue that a different action will be preformed based on which side of a invisible radius the drop occurs on, then I retract the above.

The me that knows short link release will disconnect wants this. The me that didn't know that was an intentional change would be filing a bug report for link release being inconsistently broken. :v

src/lib/litegraph/src/canvas/LinkConnector.ts
```diff
@@ -860,6 +862,11 @@ export class LinkConnector {
   }
 
   dropOnNothing(event: CanvasPointerEvent): void {
+    remove(
+      this.renderLinks,
+      (link) => link instanceof MovingInputLink && link.disconnectOnDrop
+    ).forEach((link) => (link as MovingLinkBase).disconnect())
+    if (this.renderLinks.length === 0) return
     // For external event only.
```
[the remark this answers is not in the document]
DrJKL: Still would rather have a real type guard 🙂

src/lib/litegraph/src/canvas/MovingInputLink.ts
```diff
@@ -130,4 +136,23 @@ export class MovingInputLink extends MovingLinkBase {
...
+
+  drawConnectionCircle(ctx: CanvasRenderingContext2D, to: Readonly<Point>) {
+    if (!this.disconnectOnDrop) return
+
+    const [originX, originY] = this.disconnectOrigin
+    const radius = 35
+    const distSquared = (originX - to[0]) ** 2 + (originY - to[1]) ** 2
+
+    ctx.save()
+    ctx.strokeStyle = LiteGraph.WIDGET_OUTLINE_COLOR
```
DrJKL, line 148: Should we be doing a save/restore for the context?

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -5908,6 +5910,12 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
       // Never draw slots when the pointer is down
       if (!this.pointer.isDown) reroute.drawSlots(ctx)
     }
+
+    const highlightPos = this._getHighlightPosition()
+    this.linkConnector.renderLinks
+      .filter((rl) => rl instanceof MovingInputLink)
```
DrJKL, line 5916: Like this!

DrJKL (approved):

src/lib/litegraph/src/canvas/MovingInputLink.ts
DrJKL: (I'm blind)

PR #7470, opened 2025-12-14
fix: "convert to subgraph" not shown in context menu if subgraph inside the selection context

### LINKED ISSUE #7453: Cant convert nodes to subgraph
Because this selection of nodes contains a subgraph - there is no option to convert these nodes to a subgraph
Image

But with the legacy menu -this selection of nodes can be converted to a subgraph
Image

So should this selection of nodes be allowed to be converted to a subgraph, or not?
They can be when using the legacy menu but not when using the new menu?
And because nested subgraphs are supported technically, this would indicate a bug with the new menu?

### ComfyUI Frontend Version

1.35.5

### Browser

Chrome/Chromium

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-7453-Cant-convert-nodes-to-subgraph-2c86d73d365081459582c070950630e3) by [Unito](https://www.unito.io)

christian-byrne:
Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/7453.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7470-fix-convert-to-subgraph-not-shown-in-context-menu-if-subgraph-inside-the-selection-con-2c96d73d36508146a475e8d39c64183c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useMoreOptionsMenu.ts
+++ b/src/composables/graph/useMoreOptionsMenu.ts
@@ -173,7 +173,12 @@ export function useMoreOptionsMenu() {
     }
 
     // Section 5: Subgraph operations
-    options.push(...getSubgraphOptions(hasSubgraphsSelected))
+    options.push(
+      ...getSubgraphOptions({
+        hasSubgraphs: hasSubgraphsSelected,
+        hasMultipleSelection: hasMultipleNodes.value
+      })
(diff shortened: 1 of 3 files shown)


Myestery (approved):
LGTM

PR #7473, opened 2025-12-14
fix: prevent unrelated groups from moving when dragging nodes in vueNodes mode

jtydhr88:
## Summary

Previously, when dragging a node that was not part of the selection, any selected groups would still move along with it. This fix ensures groups only move when the dragged node is actually part of the selection.

before

https://github.com/user-attachments/assets/ff9a18c2-59b2-4bbd-81b4-7a6ecb35e659

after

https://github.com/user-attachments/assets/019a6cc6-b1e2-41d1-bfec-d6af7ae84091

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7473-fix-prevent-unrelated-groups-from-moving-when-dragging-nodes-in-vueNodes-mode-2c96d73d365081a194a6fef57f9c1108) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
+++ b/src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
@@ -57,7 +57,10 @@ function useNodeDragIndividual() {
     const selectedNodes = toValue(selectedNodeIds)
 
     // capture the starting positions of all other selected nodes
-    if (selectedNodes?.has(nodeId) && selectedNodes.size > 1) {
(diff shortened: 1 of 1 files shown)


christian-byrne (approved):
Can you add a test in the future? I just added a test in [this PR](https://github.com/Comfy-Org/ComfyUI_frontend/pull/7447) for dragging groups and you should be able to use the same utilities

jtydhr88:
sure, I will add it soon

PR #7488, opened 2025-12-15
feat: implement virtual list with sliding window for TreeExplorer

Yourz:
## Summary

[Tech design](https://www.notion.so/comfy-org/TreeExplorer-VirtualList-TechDesign-2f06d73d36508054a207d0c30c368fc5?source=copy_link)

Implements virtual list (windowing) for the `TreeExplorer` component to enable efficient rendering of large tree structures with thousands of nodes, reducing DOM nodes from O(n) to O(visible) and significantly improving scroll performance.

### What
- Added virtual list implementation to `TreeExplorer` component using window-based rendering
- Extracted reusable virtual list utilities to `virtualListUtils.ts` for modularity and testability
- Implemented position and height caching to eliminate O(n²) behavior during scroll events
- Added comprehensive unit tests (32 tests) and E2E tests for virtual list functionality
- Optimized tree traversal to compute positions and heights in a single O(n) pass

### Breaking
- **Behavior change**: Tree nodes are now dynamically added/removed from DOM based on scroll position. Extensions or customizations that rely on all nodes being present in the DOM simultaneously may need updates.
- **Scroll behavior**: The visible nodes are now limited by a sliding window (60 nodes max). Users scrolling through very large trees will see nodes being recycled.

### Key Functions Extracted to `virtualListUtils.ts`:
- `mergeWindowRange`: Intelligently merges window ranges to minimize updates
- `calculateWindowRangeByHeights`: Generic window range calculation based on item positions
- `calculateSpacerHeightsVariable`: Calculates spacer heights for variable-height items
- `calculateTreePositionsAndHeights`: Generic tree traversal for position/height calculation
- `calculateChildrenListBounds`: Calculates list boundaries for parent node children
- `applyWindow`: Recursively applies windowing to tree nodes
- `createInitialWindowRange`: Creates default window range

## Review Focus

- **Virtual List Algorithm**: Window range calculation and merging logic in `virtualListUtils.ts`
- **Edge Cases**: Empty trees, deeply nested structures, rapid scrolling, simultaneous expand/scroll
- **CSS Spacers**: Verify correct scroll height maintenance using `--top-spacer` and `--bottom-spacer` CSS variables
- **Test Coverage**: Review unit tests for `calculateTreePositionsAndHeights` and `calculateChildrenListBounds`

## Screenshots

Before(it takes more than 5 seconds to load the list):

Kapture 2025-12-15 at 18 07 53

After:
Kapture 2025-12-15 at 18 10 24

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7488-feat-implement-virtual-list-with-sliding-window-for-TreeExplorer-2ca6d73d36508195aceef2041db107c2) by [Unito](https://www.unito.io)


src/components/common/TreeExplorer.vue
```diff
@@ -84,12 +102,240 @@ const {
...
+const shiftWindowsBackward = () => shiftWindows('backward')
+
 const renderedRoot = computed<RenderedTreeExplorerNode>(() => {
   const renderedRoot = fillNodeInfo(props.root)
   return newFolderNode.value
     ? combineTrees(renderedRoot, newFolderNode.value)
     : renderedRoot
 })
+
+// Build a lookup map for O(1) node access instead of O(n) tree traversal
```
Myestery, line 210: 👍

Myestery:
I ran into this error while testing. If you scroll really fast, the list kinda disappears

https://github.com/user-attachments/assets/c0b257e4-f449-43f3-b110-f8432e5a3eb8

Yourz:
> I ran into this error while testing. If you scroll really fast, the list kinda disappears
> 
>  Screen.Recording.2025-12-15.at.21.03.17.mov

maybe shorten the throttle latency would solve it? I'll try it again

Yourz:
i tried it and got the repro path:

not only u need to scroll really fast(it seems that really fast scrolling will lead to a blank status, but finally the nodes will be displayed at the end), but also u need to move ur cursor out of the node library sidebar when blank sidebar showed up after a really fast scrolling.

Does this repro path makes sense to you? @Myestery

Myestery:
> i tried it and got the repro path:
> 
> not only u need to scroll really fast(it seems that really fast scrolling will lead to a blank status, but finally the nodes will be displayed at the end), but also u need to move ur cursor out of the node library sidebar when blank sidebar showed up after a really fast scrolling.
> 
> Does this repro path makes sense to you? @Myestery

Actually, no.
Its only related to fast scrolling.
You may want to try calculating the target window position directly from scroll position
feel free to use this link to reproduce https://pr-1716.testenvs.comfy.org

Yourz:
@Myestery @christian-byrne Just fix the blank side bar

src/components/common/TreeExplorer.vue
```diff
@@ -53,7 +63,16 @@ import type {
...
+import { combineTrees } from '@/utils/treeUtil'
+import type { WindowRange } from '@/utils/virtualListUtils'
+import {
+  applyWindow as applyWindowUtil,
+  calculateScrollPercentage,
+  calculateSpacerHeights,
+  createInitialWindowRange,
+  shiftWindowBackward as shiftWindowBackwardUtil,
+  shiftWindowForward as shiftWindowForwardUtil
+} from '@/utils/virtualListUtils'
```
christian-byrne, line 74: Should we get rid of the import aliases?

src/components/common/TreeExplorer.vue
```diff
@@ -84,28 +103,282 @@ const {
   }
 )
 
+// Sliding window configuration
+const WINDOW_SIZE = 60
+const BUFFER_SIZE = 20
+const NODE_HEIGHT = 28 // Approximate height per tree node in pixels
+const SCROLL_FORWARD_THRESHOLD = 0.7 // Shift window forward when scrolled past 70%
+const SCROLL_BACKWARD_THRESHOLD = 0.3 // Shift window backward when scrolled below 30%
+const SCROLL_THROTTLE_MS = 100 // Throttle scroll handler to every 100ms
```
christian-byrne: For these comments, could we express the same information in the variable names and then not need the comments at all?

src/components/common/TreeExplorer.vue
```diff
@@ -84,28 +103,282 @@ const {
...
+    for (const key in oldKeys) {
+      if (oldKeys[key] && !newKeys[key]) {
+        delete parentWindowRanges.value[key]
+      }
+    }
+  },
+  { deep: true }
+)
+
+// Ref to access the tree container for scroll detection
```
christian-byrne: Should we cut down on some of these comments? Not sure how much they are helping

src/components/common/TreeExplorer.vue
christian-byrne, line 1: Do you think this could be simplified at all by pulling some of the observed state out into computed values? In essence, we have a list of items and a window range. We need to observe the list of items, determine our visible window, then calculation the spacer heights. See [example](https://github.com/Comfy-Org/ComfyUI_frontend/blob/08a3c767ac154038bee27cedbcfa1f73a1800943/src/components/common/VirtualGrid.vue#L4). Perhaps there are some unique challenges here however. What do you think?

Yourz: makes sense. will do it.

src/components/common/TreeExplorer.vue
Yourz: done

src/components/common/TreeExplorer.vue
Yourz: done

src/components/common/TreeExplorer.vue
Yourz: done

PR #7498, opened 2025-12-15
Revert "feat: add showScrollbar prop to VirtualGrid"

christian-byrne:
Reverts Comfy-Org/ComfyUI_frontend#7227

Reason: seems to mess up the scrollbars. I see three different ones. The scrollbar it is showing is hidden behind the scrollpanel scrollbar, which also should be hidden.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7498-Revert-feat-add-showScrollbar-prop-to-VirtualGrid-2ca6d73d365081329889ddddb45f4c17) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/common/VirtualGrid.vue
+++ b/src/components/common/VirtualGrid.vue
@@ -117,7 +117,16 @@ onBeforeUnmount(() => {
 .scroll-container {
   height: 100%;
   overflow-y: auto;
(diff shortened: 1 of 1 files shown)


christian-byrne:
Seems this was not actually the issue

PR #7507, opened 2025-12-15
fix: prevent 'no results' flash during assets tab switching

christian-byrne:
## Summary

This PR fixes an issue in the media assets sidebar where switching between Generated and Imported tabs would show "no results found" placeholder even when results existed. The issue occurred because the loading state would become false before the display assets finished updating, creating a moment where the template condition `!loading && !displayAssets.length` was incorrectly met. The fix adds an `isTransitioning` state that remains true during tab switches until asset fetching completes, ensuring the spinner displays throughout the transition and prevents the erroneous placeholder from appearing.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7507-fix-prevent-no-results-flash-during-assets-tab-switching-2ca6d73d36508174be2cee202bb11085) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/sidebar/tabs/AssetsSidebarTab.vue
+++ b/src/components/sidebar/tabs/AssetsSidebarTab.vue
@@ -59,10 +59,10 @@
     </template>
     <template #body>
       <Divider type="dashed" class="m-2" />
-      <div v-if="loading && !displayAssets.length">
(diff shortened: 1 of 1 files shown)


christian-byrne:
I can no longer reproduce this bug, so closing the PR.

PR #7538, opened 2025-12-16
feat: add ImageCompare node

jtydhr88:
## Summary

add ImageCompare node, which is high demand among custom nodes, such as rgthree, we should support as core node

Need BE change https://github.com/comfyanonymous/ComfyUI/pull/11343

https://github.com/user-attachments/assets/a37bdcd0-de59-4bdd-bfc7-1adbe92f5298

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7538-feat-add-ImageCompare-node-2cb6d73d36508163a7d5f4807aece01a) by [Unito](https://www.unito.io)


src/extensions/core/imageCompare.ts
```diff
@@ -0,0 +1,53 @@
...
+  async nodeCreated(node) {
+    if (node.constructor.comfyClass !== 'ImageCompare') return
+
+    const [oldWidth, oldHeight] = node.size
+    node.setSize([Math.max(oldWidth, 400), Math.max(oldHeight, 350)])
+
+    const onExecuted = node.onExecuted
+
+    node.onExecuted = function (message: any) {
+      onExecuted?.apply(this, arguments as any)
```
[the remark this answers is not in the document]
jtydhr88: no, not make sense. 
Becasue we already use "any" for all node.onExecuted, for example, saveMesh, uploadAudio, load3D.

src/extensions/core/imageCompare.ts
```diff
@@ -0,0 +1,53 @@
...
+
+useExtensionService().registerExtension({
+  name: 'Comfy.ImageCompare',
+
+  async beforeRegisterNodeDef(_nodeType, nodeData) {
+    if (nodeData.name !== 'ImageCompare') return
+
+    // Inject the IMAGECOMPARE widget type into the node
+    // @ts-expect-error InputSpec is not typed correctly
+    nodeData.input.required.compare_view = ['IMAGECOMPARE']
```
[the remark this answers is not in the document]
jtydhr88: no, it is necessary.
and we also use the same pattern in other components, for example, saveMesh, uploadAudio, load3D.

src/extensions/core/imageCompare.ts
```diff
@@ -0,0 +1,50 @@
...
+import { useExtensionService } from '@/services/extensionService'
+
+useExtensionService().registerExtension({
+  name: 'Comfy.ImageCompare',
+
+  async beforeRegisterNodeDef(_nodeType, nodeData) {
+    if (nodeData.name !== 'ImageCompare') return
+
+    // @ts-expect-error InputSpec is not typed correctly
+    nodeData.input.required.compare_view = ['IMAGECOMPARE']
```
christian-byrne: We would like to move away from this pattern as it is quite hard to maintain. We have the ability to add first-class input types, how would this look like if we just defined IMAGECOMPARE in the node defs in core?

jtydhr88: added https://github.com/comfyanonymous/ComfyUI/pull/11343/files#diff-08b0c3241eee64407fff1a77cb22585958c9431f2012987d564cedd0a5716e2dR1103 here, let me know it is correct that what you want

src/extensions/core/imageCompare.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  async nodeCreated(node) {
+    if (node.constructor.comfyClass !== 'ImageCompare') return
+
+    const [oldWidth, oldHeight] = node.size
+    node.setSize([Math.max(oldWidth, 400), Math.max(oldHeight, 350)])
+
+    const onExecuted = node.onExecuted
+
+    node.onExecuted = function (message: any) {
```
christian-byrne: Can we use the type `onExecuted?(output: NodeExecutionOutput): void`?

jtydhr88: fixed

christian-byrne (approved):

PR #7539, opened 2025-12-16
Fix widget reactivity

AustinMroz:
Closes #7095

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7539-Fix-widget-reactivity-2cb6d73d3650816d8977ebda35ab88b9) by [Unito](https://www.unito.io)

PR #7550, opened 2025-12-16
Fix doubled control application

AustinMroz:
With reactivity fixed, control widgets would apply twice. This is fixed by using the litegraph implementation.

Also adds control widget support for combos

Followup to #7539.

Known Issue:
- Primitive node do not have litegraph callbacks properly setup. As a result, they will display an updated value when modified by control widgets. Fixing this will requires a larger, separate PR

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7550-Fix-doubled-control-application-2cb6d73d365081739a2fc40fdfb3630e) by [Unito](https://www.unito.io)


src/types/simplifiedWidget.ts
```diff
@@ -72,3 +72,8 @@ export interface SimplifiedWidget<
 
   controlWidget?: SafeControlWidget
 }
+
+export type SimplifiedControlWidget<
+  T extends WidgetValue = WidgetValue,
+  O = Record<string, any>
+> = SimplifiedWidget<T, O> & { controlWidget: SafeControlWidget }
```
DrJKL: Nit: [Extending an interface has some benefits over intersections](https://github.com/microsoft/Typescript/wiki/Performance#preferring-interfaces-over-intersections)
```suggestion
export interface SimplifiedControlWidget<
  T extends WidgetValue = WidgetValue,
  O = Record<string, any>
> extends SimplifiedWidget<T, O> {
 controlWidget: SafeControlWidget
}
```

src/renderer/extensions/vueNodes/widgets/components/WidgetWithControl.vue
```diff
@@ -0,0 +1,61 @@
...
+  WidgetValue
+} from '@/types/simplifiedWidget'
+
+const NumberControlPopover = defineAsyncComponent(
+  () => import('./NumberControlPopover.vue')
+)
+
+const props = defineProps<{
+  widget: SimplifiedWidget<T> & { controlWidget: SafeControlWidget }
+  comp: unknown
```
DrJKL: Nit: `component` for clarity. Or `as` maybe?

src/renderer/extensions/vueNodes/widgets/components/WidgetWithControl.vue
```diff
@@ -0,0 +1,61 @@
...
+  SimplifiedWidget,
+  WidgetValue
+} from '@/types/simplifiedWidget'
+
+const NumberControlPopover = defineAsyncComponent(
+  () => import('./NumberControlPopover.vue')
+)
+
+const props = defineProps<{
+  widget: SimplifiedWidget<T> & { controlWidget: SafeControlWidget }
```
DrJKL: Could this use `SimplifiedControlWidget`?

src/renderer/extensions/vueNodes/widgets/components/WidgetWithControl.vue
```diff
@@ -0,0 +1,61 @@
...
+import type {
+  ControlOptions,
+  SafeControlWidget,
+  SimplifiedWidget,
+  WidgetValue
+} from '@/types/simplifiedWidget'
+
+const NumberControlPopover = defineAsyncComponent(
+  () => import('./NumberControlPopover.vue')
+)
```
DrJKL: Can the `Number` here go away now too?

DrJKL (commented):
So much less code 😋 

I have a few suggestions, none blocking.

Waiting to approve for the tests to be fixed.

src/renderer/extensions/vueNodes/widgets/components/WidgetWithControl.vue
AustinMroz: It should. I'll need to give the actual component a read-over to make sure the descriptions aren't specific to numbers.

DrJKL (approved):
I think it's still missing increment-wrap, but that can be its own PR.

PR #7551, opened 2025-12-16
Remove queue overlay header more menu

benceruleanlu:
Removes the triple-dot (more options) menu from `QueueOverlayHeader` and cleans up related wiring.

This is being deleted instead of fixed because it deletes all user assets on cloud, will be touched by /jobs, replaced by the second iteration of design changes, along with changes to asset deletion happening soon. It would be too risky to have a proper fix for it in the cloud deploy tomorrow.

Slack context thread here: https://comfy-organization.slack.com/archives/C09FY39CC3V/p1765860644099299


benceruleanlu:
oops

PR #7553, opened 2025-12-16
feat: handling subscription tier button link parameter

Yourz:
## Summary

Discussion here: https://comfy-organization.slack.com/archives/C0A0XANFJRE/p1764899027465379

Implement: Subscription tier query parameter for direct checkout flow

Example button link: `/cloud/subscribe?tier=standard`

`tier` could be `standard`, `creator` or `pro`
`cycle` could be `monthly` or `yearly`. it is optional, and `monthly` by default.

## Changes

- **What**: 
  - Add a landing page called `CloudSubscriptionRedirectView.vue` to handling the subscription tier button link parameter
  - Extract subscription handling logic from `PriceTable.vue`
 
- **Breaking**: 
  - Code change touched  `PriceTable.vue`
- **Dependencies**: 

## Review Focus
- link will redirect to login url, when cloud app not login
- after login, the cloud app will redirect to CloudSubscriptionRedirect page
- wait for several seconds, the cloud app will be redirected to checkout page

Kapture 2025-12-16 at 18 43 28

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7553-feat-handling-subscription-tier-button-link-parameter-2cb6d73d365081ee9580e89090248300) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/cloud/onboarding/CloudSubscriptionRedirectView.vue
+++ b/src/platform/cloud/onboarding/CloudSubscriptionRedirectView.vue
@@ -11,6 +11,7 @@ import { useErrorHandling } from '@/composables/useErrorHandling'
 import { useSubscription } from '@/platform/cloud/subscription/composables/useSubscription'
 import type { TierKey } from '@/platform/cloud/subscription/constants/tierPricing'
 import { performSubscriptionCheckout } from '@/platform/cloud/subscription/utils/subscriptionCheckoutUtil'
+
 import type { BillingCycle } from '../subscription/utils/subscriptionTierRank'
 
 const { t } = useI18n()
@@ -39,7 +40,6 @@ const runRedirect = wrapWithErrorHandlingAsync(async () => {
   const rawCycle = route.query.cycle
   let tierKeyParam: string | null = null
   let cycleParam = 'monthly'
-
 
   if (typeof rawType === 'string') {
     tierKeyParam = rawType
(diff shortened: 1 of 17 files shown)


src/platform/cloud/onboarding/CloudSubscriptionRedirectView.vue
```diff
@@ -0,0 +1,88 @@
...
+
+  const validTierKeys: TierKey[] = ['standard', 'creator', 'pro']
+  if (!validTierKeys.includes(tierKeyParam as TierKey)) {
+    await router.push('/')
+    return
+  }
+
+  selectedTierKey.value = tierKeyParam as TierKey
+
+  if (isActiveSubscription.value) {
```
christian-byrne, line 80: I think there might be a race condition where this code (below) is still executing https://github.com/Comfy-Org/ComfyUI_frontend/blob/5d1bf6dfb3626d876c8d363fb44f934cd1e8b752/src/platform/cloud/subscription/composables/useSubscription.ts#L192-L203

When we access it here (when we do `if (isActiveSubscription.value)`).

One possible solution might be to add a `isInitialized` or `isFetching` ref to `useSubscription`. Then we can await it here like

```typescript
import { until } from '@vueuse/core'

await until(isInitialized).toBe(true)
if (isActiveSubscription.value) {
    // ... rest of code
```

What do you think?

Yourz: makes sense. updated.

christian-byrne (approved):

christian-byrne (approved):
LGTM

PR #7558, opened 2025-12-16
feat: improve vue node video preview loading and a11y

christian-byrne:
## Summary

Applies appropriate logic from below PRs (which affected image outputs/previews) to video previews

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/7268: don't port the 250ms logic for videos, as it is not as relevant - videos typically take longer than 250ms and most browsers have a built-in loading state that will be displayed between component mount and video onloaded that we don't want to flash. Use the native video loaded event instead
- https://github.com/Comfy-Org/ComfyUI_frontend/pull/7252: apply to videos 1-for-1

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7558-feat-improve-vue-node-video-preview-loading-and-a11y-2cb6d73d365081eab4dcfeb1e62c553b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/VideoPreview.vue
+++ b/src/renderer/extensions/vueNodes/VideoPreview.vue
@@ -2,16 +2,20 @@
   <div
     v-if="imageUrls.length > 0"
     class="video-preview group relative flex size-full min-h-16 min-w-16 flex-col px-2"
-    tabindex="0"
-    role="region"
-    :aria-label="$t('g.videoPreview')"
-    @mouseenter="handleMouseEnter"
-    @mouseleave="handleMouseLeave"
     @keydown="handleKeyDown"
   >
     <!-- Video Wrapper -->
     <div
+      ref="videoWrapperEl"
       class="relative h-full w-full grow overflow-hidden rounded-[5px] bg-node-component-surface"
+      tabindex="0"
+      role="region"
+      :aria-label="$t('g.videoPreview')"
+      :aria-busy="showLoader"
+      @mouseenter="handleMouseEnter"
+      @mouseleave="handleMouseLeave"
+      @focusin="handleFocusIn"
+      @focusout="handleFocusOut"
     >
       <!-- Error State -->
       <div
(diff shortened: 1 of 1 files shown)


src/renderer/extensions/vueNodes/VideoPreview.vue
```diff
@@ -27,18 +31,18 @@
...
       <!-- Loading State -->
       <Skeleton
-        v-if="isLoading && !videoError"
+        v-if="showLoader && !videoError"
         class="absolute inset-0 size-full"
         border-radius="5px"
-        width="16rem"
-        height="16rem"
+        width="100%"
+        height="100%"
```
DrJKL, line 38: Not for this PR
I still find it weird that we define the sizing in 3 ways
- width/height
- absolute + inset-0
- size-full

src/renderer/extensions/vueNodes/VideoPreview.vue
```diff
@@ -126,12 +133,18 @@ const props = defineProps<VideoPreviewProps>()
 const { t } = useI18n()
 const nodeOutputStore = useNodeOutputStore()
 
+const actionButtonClass =
```
DrJKL, line 136: I'll keep this in mind for the button variants.

src/renderer/extensions/vueNodes/VideoPreview.vue
```diff
@@ -207,6 +220,16 @@ const handleMouseLeave = () => {
   isHovered.value = false
 }
 
+const handleFocusIn = () => {
+  isFocused.value = true
```
DrJKL, line 224: Nit: It really feels like we're using JavaScript to do what the DOM does well otherwise (hovered/focused state management)

DrJKL: Could we do a group/focus-within to achieve the same goals?

DrJKL (approved):

PR #7576, opened 2025-12-17
Fix promoted assets not being assets in vue

AustinMroz:
- Fixes asset widgets which have been promoted failing to display as asset widgets and having red names in vue mode.
- Fixes promoted subgraph widgets failing to resolve inputSpec for use in vue mode.

| Before | After |
| ------ | ----- |
| before | after|

I'm not excited that this creates further bloat of SimplifiedWidget.

Known issue
- Similar to #7550, subgraph widgets will have an incorrect callback and will fail to update value on a fresh reload. This can be "fixed" (made worse) by entering and exiting the subgraph. Since this creates a 'leaked' widget callback which will then be called.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7576-Fix-promoted-assets-not-being-assets-in-vue-2cc6d73d3650814b8734f69b225b0228) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -6,6 +6,7 @@ import { reactiveComputed } from '@vueuse/core'
 import { reactive, shallowReactive } from 'vue'
 
 import { useChainCallback } from '@/composables/functional/useChainCallback'
+import { isProxyWidget } from '@/core/graph/subgraph/proxyWidget'
 import type {
   INodeInputSlot,
   INodeOutputSlot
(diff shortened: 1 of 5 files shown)


DrJKL (commented):
(Want to add a test?)

DrJKL (dismissed):
More twists to the Gordian knot...
We'll untie it eventually.

AustinMroz:
I would like to add a test, but didn't see any existing precedence for tests targeting cloud. I may have to take a rain check here.

DrJKL (approved):

christian-byrne (approved):

PR #7579, opened 2025-12-17
Ensure widgets always get a single callback

AustinMroz:
The other side of reactivity. Ensure that vue mode always registers a callback on litegraph nodes and never registers more than one.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7579-Ensure-widgets-always-get-a-single-callback-2cc6d73d365081e2a488c38ae394efc0) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -105,7 +105,6 @@ function widgetWithValueRef(
     if (valueRef.value !== widget.value)
       valueRef.value = normalizeWidgetValue(widget.value) ?? undefined
   })
-  //@ts-expect-error duck violence
   widget.valueRef = valueRef
 }
 export function useReactiveWidgetValue(widget: IBaseWidget) {
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -3,8 +3,7 @@
  * Provides event-driven reactivity with performance optimizations
  */
 import { reactiveComputed } from '@vueuse/core'
-import { reactive, shallowReactive, ref, watch } from 'vue'
-import type { Ref } from 'vue'
+import { customRef, reactive, shallowReactive } from 'vue'
 
 import { useChainCallback } from '@/composables/functional/useChainCallback'
 import { isProxyWidget } from '@/core/graph/subgraph/proxyWidget'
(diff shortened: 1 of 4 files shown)


src/composables/graph/useGraphNodeManager.ts
```diff
@@ -86,6 +87,25 @@ export interface GraphNodeManager {
   cleanup(): void
 }
 
+const valueRefMap = new WeakMap<IBaseWidget, Ref<WidgetValue>>()
+
+function getValueRef(widget: IBaseWidget): Ref<WidgetValue> {
```
DrJKL: Is this meant to be a composable?

I don't usually expect to see watchers and refs being created inside functions.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -90,6 +90,23 @@ export interface GraphNodeManager {
...
 
+function widgetWithVueTrack(
+  widget: IBaseWidget
+): asserts widget is IBaseWidget & { vueTrack: () => void } {
+  if (widget.vueTrack) return
+
+  customRef((track, trigger) => {
+    widget.callback = useChainCallback(widget.callback, trigger)
+    widget.vueTrack = track
+    return { get() {}, set() {} }
```
Myestery, line 101: is it intentional that this value is not used outside the ref?
I mean, the custom ref is never returned

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -106,40 +123,63 @@ function getNodeType(node: LGraphNode, widget: IBaseWidget) {
...
+  if (
+    typeof value === 'string' ||
+    typeof value === 'number' ||
+    typeof value === 'boolean'
+  ) {
+    return value
+  }
+  if (typeof value === 'object') {
+    // Check if it's a File array
+    if (
```
Myestery, line 142: is this check needed if we still return the value regardless?

src/composables/graph/useGraphNodeManager.ts
Myestery: never mind

src/composables/graph/useGraphNodeManager.ts
AustinMroz: IMO, no. There's been some long term discussion about wanting to phase away from simplifiedWidget and `WidgetValue` being a different type than `TValue` (in `IBaseWidget`) feels like a smell.

Since this function is just being moved without changes. I lean towards keeping this PR focused on the reactivity issue.

src/composables/graph/useGraphNodeManager.ts
AustinMroz: Correct.

Myestery (approved):

PR #7591, opened 2025-12-17
Always wait for next tick before layout init

AustinMroz:
A frequent pattern is to add a node to the graph, and then update the nodes position afterwards.

Some of these cases (like subgraph unpacking) can set the node position in advance, but others, (like importA1111) require information on nodes in order to perform arranging.

Alternatives, like allowing code to either modify `app.configuringGraph` or otherwise set a temporary state were considered, but create the same problem of requiring fixes in many places.

As a proposed alternative, when a node is created, an extra tick of delay is always added before initializing layout.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7591-Always-wait-for-next-tick-before-layout-init-2cc6d73d365081f4ababc38020645670) by [Unito](https://www.unito.io)


DrJKL (dismissed):

DrJKL (dismissed):

DrJKL (dismissed):

AustinMroz:
A bit of a wild postmortem on the test failure:
- The failing test creates a KSampler node and converts it to a subgraph
  - Prior to this PR, the KSampler is shrunk to have incorrect width
- Because the Ksampler has smaller width, the fitview instantcall results in the graph being scaled to a larger size
  - Note that the subgraph outputs is arranged to a different position here, but subgraphIO are not considered for fitview call. The only relevant change causing test failure is the canvas state post fitview
- The target mouse position is very far to the right of where it should be
- This happens because the fixture code is incorrectly estimating the position of the slot from the node properties
  - The slot isn't able to be found because not-yet-created slots are held under the `.emptySlot` attr, not `.slots`
  - The gap estimate was just a magic constant guess of 10 which my PR got unlucky on :(
- With the correct slot being grabbed, there's no longer a need for the gapHeight, or any of the fallbacks

| Before PR| Before Text Fix| After Test Fix|
| ------ | ----- | ----- |
| before | after| fix|

browser_tests/fixtures/utils/litegraphUtils.ts
```diff
@@ -79,48 +79,15 @@ export class SubgraphSlotReference {
...
-          slotY = node.pos[1] + node.size[1] / 2
-        }
-
         // Convert from offset to canvas coordinates
         const canvasPos = window['app'].canvas.ds.convertOffsetToCanvas([
-          slotX,
-          slotY
+          node.emptySlot.pos[0],
+          node.emptySlot.pos[1]
         ])
```
DrJKL, line 91: Very nit...
Could you just pass `node.emptySlot.pos`?

DrJKL (approved):

browser_tests/fixtures/utils/litegraphUtils.ts
AustinMroz: Yeah, that would have been better. Automerge took the reigns this time, but I'll make sure to fix if I'm ever in the code section again.

DrJKL:
@AustinMroz plz fix

PR #7600, opened 2025-12-18
refactor: type the checkout request and response

simula-r:
## Summary

Refactor: use openapi.yml types for the strip checkout api.

Currently the types do not include the yearly subscription tier. But will once https://github.com/Comfy-Org/comfy-api/pull/757 is merged. So will want to tie the tier types to openapi after.

## Changes

- **What**: PricingTable.vue
- **Breaking**: 
- **Dependencies**: 

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7600-refactor-type-the-checkout-request-and-response-2cd6d73d3650813ca536f6184ec988f3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/cloud/subscription/components/PricingTable.vue
+++ b/src/platform/cloud/subscription/components/PricingTable.vue
@@ -258,11 +258,13 @@ import {
   FirebaseAuthStoreError,
   useFirebaseAuthStore
 } from '@/stores/firebaseAuthStore'
-import type { components } from '@/types/comfyRegistryTypes'
+import type { components, operations } from '@/types/comfyRegistryTypes'
 
 type SubscriptionTier = components['schemas']['SubscriptionTier']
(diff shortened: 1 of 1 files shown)


christian-byrne (approved):

christian-byrne (approved):

src/platform/cloud/subscription/components/PricingTable.vue
```diff
@@ -429,7 +433,7 @@ const initiateCheckout = async (tierKey: TierKey) => {
     )
   }
 
-  return await response.json()
+  return (await response.json()) as CheckoutResponse
```
christian-byrne, line 436: note for other reviewers: it would actually be somewhat bad to do runtime validation here if we ever believe this code is going to run on local - because it would prevent us from changing the API in a functionally backwards-compatible way (but not strictly same type) in the future.

PR #7601, opened 2025-12-18
fix: nodes2.0's combo widget does not respect "scale menus" setting

LittleSound:
Users report that text scaling in Nodes 2.0 does not respect the "scale menus" setting. This is affecting the visual consistency and user experience when users adjust their menu scaling preferences.

The "scale menus" setting is intended to control the scaling of menu elements within the interface. However, text scaling is not respecting this setting, meaning text does not adjust proportionally when users change their menu scaling preferences.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7601-fix-nodes2-0-s-combo-widget-does-not-respect-scale-menus-setting-2cd6d73d365081d396ede2c64c45dd1d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useTransformCompatOverlayProps.ts
+++ b/src/composables/useTransformCompatOverlayProps.ts
@@ -1,6 +1,8 @@
 import type { HintedString } from '@primevue/core'
 import { computed } from 'vue'
 
+import { useSettingStore } from '@/platform/settings/settingStore'
+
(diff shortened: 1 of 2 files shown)


christian-byrne:
I think this setting we do not need to port over. If we get some user requests, we can re-open this PR.

PR #7619, opened 2025-12-18
improve logic of 3d scene size

jtydhr88:
## Summary

improve 3d scene size logic, for preview3d, it should not have target size

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7619-improve-logic-of-3d-scene-size-2cd6d73d36508101b9f5dcef9bbe1314) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useLoad3dViewer.ts
+++ b/src/composables/useLoad3dViewer.ts
@@ -187,17 +187,18 @@ export const useLoad3dViewer = (node?: LGraphNode) => {
       const width = node.widgets?.find((w) => w.name === 'width')
       const height = node.widgets?.find((w) => w.name === 'height')
 
+      const hasTargetDimensions = !!(width && height)
+
       load3d = new Load3d(containerRef, {
         width: width ? (toRaw(width).value as number) : undefined,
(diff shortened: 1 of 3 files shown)


src/composables/useLoad3dViewer.ts
```diff
@@ -187,17 +187,18 @@ export const useLoad3dViewer = (node?: LGraphNode) => {
...
-            ? () => ({
-                width: width.value as number,
-                height: height.value as number
-              })
-            : undefined,
-        isViewerMode: true
+        getDimensions: hasTargetDimensions
+          ? () => ({
+              width: width.value as number,
+              height: height.value as number
```
DrJKL, line 198: Question: Why do we need so many type assertions around these?

jtydhr88: width and height need number type, but width.value could be others

DrJKL (approved):

src/composables/useLoad3dViewer.ts
DrJKL: Can we constrain the type upstream?

PR #7621, opened 2025-12-18
fix: 3d resize vertically

kaili-yang:
## Summary

Fix an issue #7620  where 3D nodes could be resized vertically only in the expanding direction but could not be shrunk.

## Changes
- **What**:  
  - Allow the 3D node content container to shrink by removing `h-full` and adding `min-h-0 flex-1`.
  - Prevent the internal canvas from reasserting its previous height during parent resize by avoiding `h-full` and constraining it with `max-h-full`.
- **Dependencies**:  
  - None

## Review Focus

- Verify that vertical resize works in both directions for 3D nodes.
- Ensure the canvas still resizes correctly with the node and that three.js rendering is unaffected.
- Confirm no regression in pointer events, overlays, or resize handles.

## Screenshots 
before

https://github.com/user-attachments/assets/6b9e3bb0-98eb-4b06-adb3-b3aab60a7f86

after

https://github.com/user-attachments/assets/76d1a962-6cdc-4ca4-941e-7a3874629b29

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7621-fix-3d-resize-vertically-2cd6d73d365081efb281e8a40c87bf9e) by [Unito](https://www.unito.io)


jtydhr88 (dismissed):
Great! super thanks for this!

src/components/load3d/Load3DScene.vue
```diff
@@ -1,7 +1,7 @@
 <template>
   <div
     ref="container"
-    class="relative h-full w-full"
+    class="relative h-full w-full min-h-0 overflow-hidden"
```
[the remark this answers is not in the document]
jtydhr88: no, not make sence, this issue caused by threejs canvas which not be able set css class directly, we should keep this workaround for this special case

src/components/load3d/Load3DScene.vue
```diff
@@ -65,3 +65,15 @@ onUnmounted(() => {
   props.cleanup()
 })
 </script>
+
+<style scoped>
```
DrJKL: I'd like to avoid new `<style>` or `<style scoped>` tags.
Is there a way you can target these to the canvas directly?

Also, same question about any `!important`s. There's _always_ ™️ a better way.

DrJKL: https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/extensions/core/load3d/Load3d.ts#L75
Looks like here maybe?

kaili-yang: I appreciate the guidance and the good example. 👍 I’ll revise the code and align with the approach going forward.

src/extensions/core/load3d/Load3d.ts
```diff
@@ -72,7 +72,15 @@ class Load3d {
     this.renderer.setClearColor(0x282828)
     this.renderer.autoClear = false
     this.renderer.outputColorSpace = THREE.SRGBColorSpace
-    this.renderer.domElement.classList.add('flex', '!h-full', '!w-full')
+    this.renderer.domElement.classList.add(
+      'absolute',
+      'top-0',
+      'left-0',
+      'h-full',
+      'w-full',
```
DrJKL: Minor nit: You could replace these with `inset-0`
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/inset

src/extensions/core/load3d/Load3d.ts
```diff
@@ -72,7 +72,15 @@ class Load3d {
...
+      'block',
```
DrJKL: Is it not a block level element to begin with?

DrJKL (dismissed):

src/extensions/core/load3d/Load3d.ts
kaili-yang: Good code review! Thanks!

src/extensions/core/load3d/Load3d.ts
kaili-yang: Yes, it is. The code is updated.

PR #7627, opened 2025-12-19
Fix buttons displayed behind images in litegraph

AustinMroz:
#7394 caused a regression with preview image buttons showing behind the image in litegraph. This is fixed by also deferring button draws

| Before | After |
| ------ | ----- |
| before | after|


DrJKL (dismissed):
Agree that there could be some issues with hardcoding coordinates, and it would be nice to have a method on the page to encapsulate the node image logic.

christian-byrne (dismissed):

christian-byrne (approved):

PR #7630, opened 2025-12-19
Disable desktop release notifications

benceruleanlu:
Disable desktop release toast and red-dot notifications.

## What changed
- Stop showing release toast and update red-dot on Electron builds; keep them for non-desktop platforms.
- Update releaseStore unit tests to reflect the desktop/web behavior split.

## Why
- Desktop updates are handled by todesktop, so showing core version update notifications is misleading.
- This keeps the existing release UX for web while avoiding duplicate or confusing prompts on desktop.
- Tradeoff: desktop no longer surfaces core version release indicators; that remains the responsibility of todesktop notifications.

## Evidence
- None

## References
- None

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7630-Disable-desktop-release-notifications-2ce6d73d365081648fd4c3256ac912bb) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/updates/common/releaseStore.ts
+++ b/src/platform/updates/common/releaseStore.ts
@@ -94,8 +94,9 @@ export const useReleaseStore = defineStore('release', () => {
 
   // Show toast if needed
   const shouldShowToast = computed(() => {
-    // Only show on desktop version
-    if (!isElectron() || isCloud) {
+    // Disable on desktop; updates are handled by todesktop
+    // Disable on cloud; users do not update cloud
+    if (isElectron() || isCloud) {
       return false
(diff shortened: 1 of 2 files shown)


benceruleanlu:
I think instead of disabling, it needs to be refactored.

An electron update store based on todesktop, and we have the red dot be based on that. The toast is duplicating the popup, we can replace the popup with the toast, and prevent the todesktop default update UI from showing since we have an inapp version.

jtydhr88:
as it keeps draft for months and has conflicts now, please rebase and resend if we still need it.
close it for now

PR #7632, opened 2025-12-19
Fix animated webp test and remove screenshot

AustinMroz:
Remove screenshot expectation and restore test.

We care that it's displaying an image not what frame of the image is being displayed.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7632-Fix-animated-webp-test-and-remove-screenshot-2ce6d73d365081a59d38dc9d82fa1a51) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/widget.spec.ts
+++ b/browser_tests/tests/widget.spec.ts
@@ -262,13 +262,7 @@ test.describe('Animated image widget', () => {
     expect(filename).toContain('animated_webp.webp')
   })
 
-  // FIXME: This test keeps flip-flopping because it relies on animated webp timing,
-  // which is inherently unreliable in CI environments. The test asset is an animated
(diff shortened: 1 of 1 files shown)


DrJKL (approved):

AustinMroz:
Test failure is unrelated.

christian-byrne:
> We care that it's displaying an image not what frame of the image is being displayed.

The bug that motivated this test case was one in which animated WEBP would not animate in the LoadImage (it would just sit on the first frame forever), so this may be slightly inaccurate. There's definitely a better way to test, "did the frame change at some point across some period of time" but no idea how to do that without it being inherently flaky.

AustinMroz:
:thinking:  I'll make a note to return in the future. Perhaps 3 screenshots taken at half the animation interval and then testing that not all 3 frames are equal?

PR #7643, opened 2025-12-19
Test: Fix templates Spec, scroll card into view

DrJKL:
## Summary

- Scroll the target card into view before clicking
- Hopefully stabilize the locale check by enqueuing the check earlier in the process

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7643-Test-Fix-templates-Spec-scroll-card-into-view-2ce6d73d36508175a009f81502f0fe16) by [Unito](https://www.unito.io)


AustinMroz (approved):
Test is passing and fix seems reasonable. :+1:

PR #7646, opened 2025-12-19
fix: properly type extractWorkflow return

ric-yu:
## Summary
Type `extractWorkflow` return as `ComfyWorkflowJSON | undefined` instead of `unknown` for better type safety downstream.

## Context
This is a small preparatory PR to slim down the Jobs API migration PR (PR 2 of 3).

## Test plan
- [x] TypeCheck passes
- [x] Lint passes

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7646-fix-properly-type-extractWorkflow-return-2ce6d73d365081c794eac3669632271f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/remote/comfyui/jobs/fetchJobs.ts
+++ b/src/platform/remote/comfyui/jobs/fetchJobs.ts
@@ -6,6 +6,7 @@
  * All distributions use the /jobs endpoint.
  */
 
+import type { ComfyWorkflowJSON } from '@/platform/workflow/validation/schemas/workflowSchema'
 import type { PromptId } from '@/schemas/apiSchema'
 
 import type {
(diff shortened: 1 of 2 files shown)


src/platform/remote/comfyui/jobs/fetchJobs.ts
```diff
@@ -139,8 +140,12 @@ export async function fetchJobDetail(
...
-export function extractWorkflow(job: JobDetail | undefined): unknown {
+export function extractWorkflow(
+  job: JobDetail | undefined
+): ComfyWorkflowJSON | undefined {
   const parsed = zWorkflowContainer.safeParse(job?.workflow)
   if (!parsed.success) return undefined
-  return parsed.data.extra_data?.extra_pnginfo?.workflow
+  return parsed.data.extra_data?.extra_pnginfo?.workflow as
+    | ComfyWorkflowJSON
+    | undefined
```
DrJKL: Why doesn't the Zod type give you the expected type here?

ric-yu: The workflow is opaque at the API layer (we don't type the workflows in cloud :(), so the Zod schema uses z.unknown(). Full workflow validation happens downstream via validateComfyWorkflow - this just validates the container structure exists.

christian-byrne: In that case wouldn't we not want to assert the type before it's been validated?

christian-byrne: @ric-yu any way we could look at how the type inference happens in the lifecycle of this object and see if there's a better way?

christian-byrne (approved):

PR #7647, opened 2025-12-19
feat: position properties panel opposite to sidebar

csongorczezar:
## Problem

When sidebar is positioned on the right, the properties panel also appears on the right, causing both panels to compete for space and creating a poor layout.

## Solution

Properties panel now dynamically positions itself opposite to the sidebar:
- Sidebar left → Properties panel right (default)
- Sidebar right → Properties panel left

## Changes

- Modified `LiteGraphCanvasSplitterOverlay.vue` to conditionally render properties panel based on sidebar location
- Updated splitter refresh key to recalculate layout when sidebar position changes
- Added dynamic close button icon in `RightSidePanel.vue` that points in the correct direction

## Testing

- Created E2E tests to verify positioning behavior
- Manually verified visual behavior in browser

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7647-feat-position-properties-panel-opposite-to-sidebar-2ce6d73d365081049683e74c8d03dbdd) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/LiteGraphCanvasSplitterOverlay.vue
+++ b/src/components/LiteGraphCanvasSplitterOverlay.vue
@@ -22,6 +22,18 @@
         state-storage="local"
         @resizestart="onResizestart"
       >
+        <!-- Properties Panel on left when sidebar is on right -->
+        <SplitterPanel
+          v-if="
+            rightSidePanelVisible && !focusMode && sidebarLocation === 'right'
+          "
+          class="bg-comfy-menu-bg pointer-events-auto"
+          :min-size="15"
+          :size="20"
+        >
+          <slot name="right-side-panel" />
+        </SplitterPanel>
+
         <SplitterPanel
           v-if="sidebarLocation === 'left' && !focusMode"
           :class="
@@ -96,9 +108,11 @@
           />
         </SplitterPanel>
 
-        <!-- Right Side Panel - independent of sidebar -->
+        <!-- Properties Panel on right when sidebar is on left (default) -->
         <SplitterPanel
-          v-if="rightSidePanelVisible && !focusMode"
+          v-if="
+            rightSidePanelVisible && !focusMode && sidebarLocation === 'left'
+          "
           class="bg-comfy-menu-bg pointer-events-auto"
           :min-size="15"
           :size="20"
@@ -159,12 +173,16 @@ function onResizestart({ originalEvent: event }: SplitterResizeStartEvent) {
 }
 
 /*
- * Force refresh the splitter when right panel visibility changes to recalculate the width
+ * Force refresh the splitter when right panel visibility or sidebar location changes
+ * to recalculate the width and panel order
  */
 const splitterRefreshKey = computed(() => {
-  return rightSidePanelVisible.value
-    ? 'main-splitter-with-right-panel'
-    : 'main-splitter'
+  const parts = ['main-splitter']
+  if (rightSidePanelVisible.value) {
+    parts.push('with-right-panel')
+  }
+  parts.push(sidebarLocation.value)
+  return parts.join('-')
 })
 </script>
 
--- a/src/components/LiteGraphCanvasSplitterOverlay.vue
+++ b/src/components/LiteGraphCanvasSplitterOverlay.vue
@@ -22,43 +22,42 @@
         state-storage="local"
         @resizestart="onResizestart"
       >
-        <!-- Properties Panel on left when sidebar is on right -->
+        <!-- First panel: sidebar when left, properties when right -->
         <SplitterPanel
           v-if="
-            rightSidePanelVisible && !focusMode && sidebarLocation === 'right'
+            !focusMode && (sidebarLocation === 'left' || rightSidePanelVisible)
           "
-          class="bg-comfy-menu-bg pointer-events-auto"
-          :min-size="15"
-          :size="20"
-        >
-          <slot name="right-side-panel" />
-        </SplitterPanel>
-
-        <SplitterPanel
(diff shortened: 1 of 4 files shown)


src/components/LiteGraphCanvasSplitterOverlay.vue
```diff
@@ -22,6 +22,18 @@
...
+        <!-- Properties Panel on left when sidebar is on right -->
+        <SplitterPanel
+          v-if="
+            rightSidePanelVisible && !focusMode && sidebarLocation === 'right'
+          "
+          class="bg-comfy-menu-bg pointer-events-auto"
+          :min-size="15"
+          :size="20"
+        >
+          <slot name="right-side-panel" />
```
DrJKL: Would it be possible to just swap the slots between the two panels instead?

csongorczezar: Hmmm. I thought about that too, but the issue is that the Splitter component renders its children in the order they appear in the template. So to get the properties panel on the left vs right, it needs to be the first child vs the last child in the DOM.

We could technically do a computed array of components and dynamically reorder them, but that felt more convoluted than just having two conditional renders. 
The actual duplication is minimal: it's just the SplitterPanel wrapper, the content itself ([<slot name="right-side-panel" />]) is shared.
If you have a cleaner pattern in mind though, I'm all ears :)

src/components/LiteGraphCanvasSplitterOverlay.vue
```diff
@@ -73,38 +86,37 @@
...
+          :min-size="sidebarLocation === 'right' ? 10 : 15"
           :size="20"
-          :style="{
-            display:
-              sidebarPanelVisible && sidebarLocation === 'right'
-                ? 'flex'
-                : 'none'
-          }"
+          :style="
+            sidebarLocation === 'right'
```
Myestery: we could use a computed prop here

Myestery (dismissed):
LGTM

src/components/LiteGraphCanvasSplitterOverlay.vue
csongorczezar: Done

src/components/LiteGraphCanvasSplitterOverlay.vue
csongorczezar: Done.

Myestery (dismissed):

src/components/LiteGraphCanvasSplitterOverlay.vue
```diff
@@ -159,12 +164,30 @@ function onResizestart({ originalEvent: event }: SplitterResizeStartEvent) {
...
 const splitterRefreshKey = computed(() => {
-  return rightSidePanelVisible.value
-    ? 'main-splitter-with-right-panel'
-    : 'main-splitter'
+  const parts = ['main-splitter']
+  if (rightSidePanelVisible.value) {
+    parts.push('with-right-panel')
+  }
+  parts.push(sidebarLocation.value)
+  return parts.join('-')
```
DrJKL: Nit: I'd probably just do this with a single template string

src/components/LiteGraphCanvasSplitterOverlay.vue
```diff
@@ -159,12 +164,30 @@ function onResizestart({ originalEvent: event }: SplitterResizeStartEvent) {
...
+})
+
+const firstPanelStyle = computed(() => {
+  if (sidebarLocation.value === 'left') {
+    return { display: sidebarPanelVisible.value ? 'flex' : 'none' }
+  }
+  return undefined
```
DrJKL, line 179: Question: Could this be done with tailwind utilities instead of through the style attribute?

src/components/LiteGraphCanvasSplitterOverlay.vue
```diff
@@ -118,6 +122,7 @@ import type { SplitterResizeStartEvent } from 'primevue/splitter'
 import SplitterPanel from 'primevue/splitterpanel'
 import { computed } from 'vue'
 
+import { t } from '@/i18n'
```
DrJKL: The global `t` function is deprecated. Could you use the `useI18n` composable instead?

src/components/LiteGraphCanvasSplitterOverlay.vue
csongorczezar: Done

src/components/LiteGraphCanvasSplitterOverlay.vue
csongorczezar: Great observation. With this we are removing the computed styles.
Done!

src/components/LiteGraphCanvasSplitterOverlay.vue
csongorczezar: Done!

src/components/LiteGraphCanvasSplitterOverlay.vue
csongorczezar: Reverted back to the original way, because Tailwind was interacting differently with PrimeVue's Splitter component, causing subtle layout shifts that affect the canvas rendering.

comfydesigner (approved):
Can't comment on the code per say, but checked out the branch and the properties panel seems to be functioning as we discussed. Looks good to me

src/components/LiteGraphCanvasSplitterOverlay.vue
DrJKL: 😢

DrJKL (approved):

LittleSound (commented):
LGTM

PR #7650, opened 2025-12-19
refactor: encapsulate error extraction in TaskItemImpl getters

ric-yu:
## Summary
- Add `errorMessage` and `executionError` getters to `TaskItemImpl` that extract error info from status messages
- Update `useJobErrorReporting` composable to use these getters instead of standalone function
- Remove the standalone `extractExecutionError` function

This encapsulates error extraction within `TaskItemImpl`, preparing for the Jobs API migration where the underlying data format will change but the getter interface will remain stable.

## Test plan
- [x] All existing tests pass
- [x] New tests added for `TaskItemImpl.errorMessage` and `TaskItemImpl.executionError` getters
- [x] TypeScript, lint, and knip checks pass

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7650-refactor-encapsulate-error-extraction-in-TaskItemImpl-getters-2ce6d73d365081caae33dcc7e1e07720) by [Unito](https://www.unito.io)


src/stores/queueStore.ts
```diff
@@ -319,6 +320,32 @@ export class TaskItemImpl {
...
+  /**
+   * Extracts the execution error message from status messages.
+   * Used by error reporting UI components.
+   */
+  get errorMessage(): string | undefined {
+    const messages = this.status?.messages
+    if (!Array.isArray(messages) || !messages.length) return undefined
+    const record = messages.find(
+      (entry: unknown) => Array.isArray(entry) && entry[0] === 'execution_error'
+    ) as [string, { exception_message?: string }?] | undefined
```
DrJKL: Why isn't the type here inferrable?

src/stores/queueStore.ts
```diff
@@ -323,6 +323,10 @@ export class TaskItemImpl {
   /**
    * Extracts the execution error message from status messages.
    * Used by error reporting UI components.
+   *
+   * Note: The type cast is required because `messages` is typed as
+   * `Array<[string, unknown]>` - TypeScript cannot narrow the second tuple
```
DrJKL: Can you update the expected type of `messages`?

src/stores/queueStore.ts
ric-yu: So, the zod isn't available until https://github.com/Comfy-Org/ComfyUI_frontend/pull/7170, where we clean up the type definitions. These PR's were kinda hard to separate with all the dependencies, so this is kind of a placeholder.

src/stores/queueStore.ts
ric-yu: see: https://github.com/Comfy-Org/ComfyUI_frontend/pull/7650#discussion_r2684624568

christian-byrne (approved):

PR #7658, opened 2025-12-20
feat: support effective_balance_micros for user balance display

huntcsg:
## Summary

Add support for the new `effective_balance_micros` field to show users their effective balance accounting for pending charges.

## Changes

- **What**: Update balance display components to use `effective_balance_micros` (with fallback to `amount_micros` for backwards compatibility)
- **Types**: Add `pending_charges_micros` and `effective_balance_micros` to `GetCustomerBalance` response type in registry-types

## Review Focus

- The fallback pattern ensures backwards compatibility if the API doesn't return the new field
- The `effective_balance_micros` can be negative when pending charges exceed the available balance

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7658-feat-support-effective_balance_micros-for-user-balance-display-2cf6d73d36508193a5a7e999f3185078) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/common/UserCredit.test.ts
+++ b/src/components/common/UserCredit.test.ts
@@ -1,0 +1,134 @@
+import { mount } from '@vue/test-utils'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { createI18n } from 'vue-i18n'
+
+import enMessages from '@/locales/en/main.json' with { type: 'json' }
+
+import UserCredit from './UserCredit.vue'
+
+vi.mock('firebase/app', () => ({
+  initializeApp: vi.fn(),
+  getApp: vi.fn()
+}))
+
+vi.mock('firebase/auth', () => ({
+  getAuth: vi.fn(),
+  setPersistence: vi.fn(),
+  browserLocalPersistence: {},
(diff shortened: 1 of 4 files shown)


packages/registry-types/src/comfyRegistryTypes.ts
christian-byrne, line 1: These types are generated by manually running [this workflow](https://github.com/Comfy-Org/ComfyUI_frontend/actions/workflows/api-update-registry-api-types.yaml) (it will create a PR). Are these new fields in the deployed prod https://api.comfy.org/openapi?

Manually editing them is also fine though, as long as it matches the type that this workflow will set next time it runs.

huntcsg: I just ran the generate command locally and it brought in a bunch of other changes too!
I think it would be best to just run the workflow and merge, and I will then rebase on main once it's merged, just to be sure it's exactly matching.

huntcsg: Synced in #7659 
Will rebase onto that once it's merged.

christian-byrne (approved):

christian-byrne (approved):

PR #7688, opened 2025-12-20
fix: QPO progress bar now shows node name in subgraphs

csongorczezar:
## Summary

Resolve the queue progress node label from queued prompt metadata so subgraph execution IDs show the correct node name without depending on the live canvas.

## Changes

- **What**: Store a prompt-scoped `executionId -> { title, type }` lookup from `p.output` when queueing a job, and use that lookup for the active job's executing node label.
- **What**: Reuse the same job-scoped node info for the browser tab title so it stays aligned with the queue overlay.
- **What**: Add unit coverage for root and subgraph execution IDs, and merge the branch forward to current `main`.

## Review Focus

This keeps the fix scoped to the existing singular `activeJobId` path. It fixes subgraph labels and avoids the workflow-switching regression from resolving against `app.rootGraph`, but it does not redesign concurrent multi-job selection yet.

Longer term, the cleaner solution is still prompt-scoped execution metadata from the backend rather than frontend reconstruction.

N/A


tests-ui/tests/store/executionStore.executingNode.test.ts
```diff
@@ -0,0 +1,157 @@
...
+import type { LGraphNode } from '@/lib/litegraph/src/litegraph'
+import { useExecutionStore } from '@/stores/executionStore'
+
+// Create mock functions
+const mockNodeExecutionIdToNodeLocatorId = vi.fn()
+const mockNodeIdToNodeLocatorId = vi.fn()
+const mockNodeLocatorIdToNodeExecutionId = vi.fn()
+
+// Create a mocked graph that we can manipulate
+let mockRootGraph: any
```
DrJKL: Could you do a full mocked object here as a const?
Or maybe a mock factory?

In general, I prefer avoiding global state for mocks in tests.

DrJKL: Oh, also a `Partial` type instead of `any` please.

tests-ui/tests/store/executionStore.executingNode.test.ts
```diff
@@ -0,0 +1,157 @@
...
+      type: 'SubgraphNode',
+      isSubgraphNode: () => true,
+      subgraph: {
+        id: 'sub-uuid',
+        getNodeById: vi.fn((id) => {
+          return id === '789' ? mockNodeInSubgraph : null
+        }),
+        _nodes: []
+      }
+    } as unknown as LGraphNode
```
DrJKL: Why does it need to be cast through unknown?
Could this be pulled up into a factory to avoid having to use a type assertion?

tests-ui/tests/store/executionStore.executingNode.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    revokePreviews: vi.fn(),
+    nodePreviewImages: {}
+  }
+}))
+
+// Mock the workflowStore
+vi.mock('@/platform/workflow/management/stores/workflowStore', async () => {
+  const { ComfyWorkflow } = await vi.importActual<
+    typeof import('@/platform/workflow/management/stores/workflowStore')
+  >('@/platform/workflow/management/stores/workflowStore')
```
DrJKL: Could you use `importOriginal` here?

DrJKL: https://vitest.dev/guide/mocking/modules.html#mocking-a-module

tests-ui/tests/store/executionStore.executingNode.test.ts
AustinMroz: LGraphNode is particularly thorny to mock because it has a typed constructor that makes an empty object not satisfy `Partial<LGraphNode>`

AustinMroz (commented):
Not going through activePrompt causes a regression where swapping (or modifying) workflows mid execution will result in either failure to find, or an incorrect node being returned.

But given we also have the same issue with (the more frequently seen) node previews, it's probably not worth worrying too much about.

tests-ui/tests/store/executionStore.executingNode.test.ts
```diff
@@ -0,0 +1,181 @@
...
+    return {
+      ComfyWorkflow,
+      useWorkflowStore: vi.fn(() => ({
+        nodeExecutionIdToNodeLocatorId: mockNodeExecutionIdToNodeLocatorId,
+        nodeIdToNodeLocatorId: mockNodeIdToNodeLocatorId,
+        nodeLocatorIdToNodeExecutionId: mockNodeLocatorIdToNodeExecutionId
+      }))
+    }
+  }
+)
```
DrJKL: Optional: Just moving the mock functions out of the global scope and into the mock.
Do we need the original import for this?
I like the testing here, but I think it could be simplified to target just the correspondence between the executing ID and the identity of the node.

```suggestion

// Keep track of the current mock root graph
const mockAppState = {
  rootGraph: createMockRootGraph()
}

// Mock the app import and its rootGraph property
vi.mock('@/scripts/app', () => ({
  app: {
    get rootGraph() {
      return mockAppState.rootGraph
    },
    set rootGraph(value) {
      mockAppState.rootGraph = value
    },
    revokePreviews: vi.fn(),
    nodePreviewImages: {}
  }
}))

// Mock the workflowStore
vi.mock(
  '@/platform/workflow/management/stores/workflowStore',
  async (importOriginal) => {
    const { ComfyWorkflow } =
      await importOriginal<
        typeof import('@/platform/workflow/management/stores/workflowStore')
      >()
    const mockNodeExecutionIdToNodeLocatorId = vi.fn()
    const mockNodeIdToNodeLocatorId = vi.fn()
    const mockNodeLocatorIdToNodeExecutionId = vi.fn()
      
    return {
      ComfyWorkflow,
      useWorkflowStore: vi.fn(() => ({
        nodeExecutionIdToNodeLocatorId: mockNodeExecutionIdToNodeLocatorId,
        nodeIdToNodeLocatorId: mockNodeIdToNodeLocatorId,
        nodeLocatorIdToNodeExecutionId: mockNodeLocatorIdToNodeExecutionId
      }))
    }
  }
)
```

tests-ui/tests/store/executionStore.executingNode.test.ts
DrJKL, line 1: I appreciate separating things out for clarity, but I think I'd rather keep the unit tests for each store in a single file if possible, especially for shared mocking consistency.

And if that gets too big, it might be a sign that we have a store that's doing too many things.

DrJKL (dismissed):
Only comments are about the tests 🙂

benceruleanlu:
Thanks for reviewing this @DrJKL and @AustinMroz, I couldn't pick this up before the holidays.

tests-ui/tests/store/executionStore.executingNode.test.ts
csongorczezar: Thank you.
Your suggestion has been applied.

tests-ui/tests/store/executionStore.executingNode.test.ts
csongorczezar: Good idea.
Merged it to the main executionStore test.

tests-ui/tests/store/executionStore.test.ts
```diff
@@ -297,3 +303,103 @@ describe('useExecutionStore - Node Error Lookups', () => {
...
+      title: 'Nested Node',
+      type: 'NestedNode'
+    } as LGraphNode
+
+    // Create a real subgraph with the test helper
+    const testSubgraph = createTestSubgraph({
+      id: 'sub-uuid'
+    })
+    // Add the mock node to the subgraph
+    testSubgraph._nodes.push(mockNodeInSubgraph)
```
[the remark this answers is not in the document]
csongorczezar: Done!

benceruleanlu:
Will bring to finish

AustinMroz (dismissed):
I'll approve, but I have the same complaint as before: Not going through `activeprompt`(/`activeJob`?) causes a regression where swapping (or modifying) workflows mid execution will result in either failure to find, or an incorrect node being returned.

benceruleanlu:
> I'll approve, but I have the same complaint as before: Not going through `activeprompt`(/`activeJob`?) causes a regression where swapping (or modifying) workflows mid execution will result in either failure to find, or an incorrect node being returned.

Will fix this before merge then

benceruleanlu:
> I'll approve, but I have the same complaint as before: Not going through `activeprompt`(/`activeJob`?) causes a regression where swapping (or modifying) workflows mid execution will result in either failure to find, or an incorrect node being returned.

When a prompt is queued, it's stored in execution store along with promptOutput, and this is used as the lookup now, which fixes the two concerns of

1. Wrong node named when switching between workflows, because it's keyed with promptid
2. Wrong node named when editing the graph

Also fixes another caller in the tab title.

A note is that this is not multi execution proof yet.

AustinMroz (dismissed):

benceruleanlu:
I'm going to merge based on the last approval. I just forgot to merge this one.

PR #7692, opened 2025-12-21
feat: pass target tier to billing portal for subscription updates

huntcsg:
## Summary

Pass target tier to billing portal API for deep linking to Stripe's subscription update confirmation screen when user has an active subscription.

## Changes

- **What**: When a user with an active subscription clicks a tier in PricingTable, pass the target tier (including billing cycle) to `accessBillingPortal` which sends it as `target_tier` in the request body. This enables the backend to create a Stripe billing portal deep link directly to the subscription update confirmation screen.
- **Dependencies**: Requires comfy-api PR for `POST /customers/billing` `target_tier` support

## Review Focus

- PricingTable now differentiates between new subscriptions (checkout flow) and existing subscriptions (billing portal with deep link)
- Type derivation uses `Parameters<typeof authStore.accessBillingPortal>[0]` to avoid duplicating the tier union (matches codebase pattern)
- Registry types manually updated to include `target_tier` field (will be regenerated when API is deployed)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7692-feat-pass-target-tier-to-billing-portal-for-subscription-updates-2d06d73d365081b38fe4c81e95dce58c) by [Unito](https://www.unito.io)


src/composables/auth/useFirebaseAuthActions.ts
```diff
@@ -102,17 +102,22 @@ export const useFirebaseAuthActions = () => {
...
-          error: 'No billing portal URL returned'
-        })
-      )
-    }
-    window.open(response.billing_portal_url, '_blank')
-  }, reportError)
+  const accessBillingPortal = wrapWithErrorHandlingAsync(
+    async (
+      targetTier?: Parameters<typeof authStore.accessBillingPortal>[0]
+    ) => {
```
christian-byrne: nit: Since [`wrapWithErrorHandlingAsync`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/9514e5d36c42af26a553f6ad3f883943a7389da8/src/composables/useErrorHandling.ts#L79-L87
) is generic, I believe the more idiomatic pattern would be to provide the tuple and return generics at the call site and let those flow through to the paramter.

I also wouldn't use `Parameters<typeof authStore.accessBillingPortal>[0]` because it hides which argument the wrapper expects because readers have to mentally index into the tuple. Spelling out `[BillingPortalTargetTier | undefined]` keeps the intent obvious, and we still fail fast the moment `accessBillingPortal` changes.

Suggestion:

```typescript
const accessBillingPortal = wrapWithErrorHandlingAsync<
  [BillingPortalTargetTier | undefined],
  void
>(async (targetTier) => {
  const response = await authStore.accessBillingPortal(targetTier)
  // …
}, reportError)
```

christian-byrne (commented):
LGTM, should be good after backend is deployed.

christian-byrne (dismissed):

christian-byrne:
Oops, made a mistake while fixing conflict in GitHub UI, fixing now.

christian-byrne (approved):

PR #7701, opened 2025-12-22
Fix(cloud)/pricing annual misc

simula-r:
## Summary

Fix: PricingTable showed "Current Plan" on the wrong billing cycle (e.g., showing it on Yearly when subscribed to Monthly) because we weren't checking subscription_duration. Now we check for ANNUAL | MONTHLY match.

Fix: Subscribed users were being sent to billing portal instead of checkout. Now routes to checkout.

Improved: Types now use openapi.yml as source of truth. Tier names in user popover and subscription panels now reflect the billing cycle (YEARLY/MONTHLY).

Recommended to merge this before https://github.com/Comfy-Org/ComfyUI_frontend/pull/7692

### DIFF
--- a/src/platform/cloud/subscription/components/PricingTable.vue
+++ b/src/platform/cloud/subscription/components/PricingTable.vue
@@ -331,8 +331,9 @@ const tiers: PricingTierConfig[] = [
 
 const { n } = useI18n()
 const { getAuthHeader } = useFirebaseAuthStore()
-const { isActiveSubscription, subscriptionTier } = useSubscription()
-const { accessBillingPortal, reportError } = useFirebaseAuthActions()
+const { isActiveSubscription, subscriptionTier, isYearlySubscription } =
+  useSubscription()
+const { reportError } = useFirebaseAuthActions()
 const { wrapWithErrorHandlingAsync } = useErrorHandling()
 
 const isLoading = ref(false)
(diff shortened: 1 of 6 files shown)


src/platform/cloud/subscription/components/PricingTable.vue
```diff
@@ -428,13 +441,9 @@ const handleSubscribe = wrapWithErrorHandlingAsync(
     loadingTier.value = tierKey
 
     try {
-      if (isActiveSubscription.value) {
-        await accessBillingPortal()
-      } else {
-        const response = await initiateCheckout(tierKey)
-        if (response.checkout_url) {
-          window.open(response.checkout_url, '_blank')
-        }
```
christian-byrne, line 437: Does the checkout URL use the portal link to change plan when already subscribed now? Previously, we did `accessBillingPortal` when already subscribed otherwise it would take you to checkout session and then if you wanted to change subscription from there, you would need to re-auth on Stripe side.

christian-byrne: Oh I see, we have https://github.com/Comfy-Org/ComfyUI_frontend/pull/7692 after which will make the billing portal deep link to the update subscription flow.

christian-byrne:
I'll apply CodeRabbit's nit while waiting for CI.

christian-byrne:
Rebasing for https://github.com/Comfy-Org/ComfyUI_frontend/pull/7702.

christian-byrne (approved):

PR #7736, opened 2025-12-23
Add infra for parallel execution support

benceruleanlu:
WIP

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7736-Add-infra-for-parallel-execution-support-2d26d73d365081748ed9c647d132109d) by [Unito](https://www.unito.io)


benceruleanlu:
Closing this as stale and superseded as an implementation vehicle.

The underlying browser parallel-execution problem is still legitimate, but this PR is a partial Phase 1 store rewrite from December 2025. It leaves output/preview isolation, per-job UI, interrupt routing, browser-tab policy, and behavioral coverage unimplemented.

Since then, `main` has moved substantially and all three source files changed here now conflict. The execution domain has moved from prompt-centric to job-centric naming, and per-job previews, per-workflow output restoration, progress-text job identity, and workflow status indicators were implemented through later architecture.

A more complete concurrent-execution design was explored in #9786/#9789 around explicit active-vs-focused job semantics, feature gating, UI controls, and behavioral tests. Review of that work also exposed cross-job event races that this older patch does not comprehensively address. The backend rollout has likewise evolved to tier- and auth-method-gated concurrency, with production browser parallelism still deliberately disabled.

If browser parallel execution is resumed, it should be implemented against current `main` and the later reviewed design rather than rebasing this patch.

PR #7737, opened 2025-12-23
[QPOv2] Add list view to assets sidepanel

benceruleanlu:
This adds the list view to the media assets sidepanel, while also adding the active jobs to be displayed right now.

The design for this is actually changing, which is why it is in draft right now. There are technical limitations of the virtual grid that doesn't make it easy for both the active jobs and generated assets to exist on the same container. Currently WIP right now.

Part of the QPO v2 iteration, figma design can be found [here](https://www.figma.com/design/LVilZgHGk5RwWOkVN6yCEK/Queue-Progress-Modal?node-id=3330-37286&m=dev). This will be implemented in a series of stacked PRs that can be reviewed and merged individually.

main <-- #7737, #7743, #7745

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7737-QPOv2-Add-list-view-to-assets-sidepanel-2d26d73d365081858e22c48902bd56e2) by [Unito](https://www.unito.io)


src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -337,6 +347,20 @@ const displayAssets = computed(() => {
...
+    (!isListView.value || activeJobsCount.value === 0)
+)
+
+const showEmptyState = computed(
+  () =>
+    !loading.value &&
+    displayAssets.value.length === 0 &&
+    (!isListView.value || activeJobsCount.value === 0)
+)
+
```
benceruleanlu, line 375: This change was made because now the view can show even when there are no generated assets, because running jobs were added to this view

src/config/uiFeatureFlags.ts
```diff
@@ -1 +1 @@
-export const isQPOV2Enabled = false
+export const isQPOV2Enabled = true
```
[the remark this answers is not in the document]
benceruleanlu: fixed, sorry

benceruleanlu:
This PR can be split into three if needed, one for the card, one for the actions, and another for the list itself.

DrJKL:
@coderabbitai full review

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -96,7 +96,15 @@
         />
       </div>
       <div v-else class="relative size-full" @click="handleEmptySpaceClick">
+        <AssetsSidebarListView
+          v-if="isListView"
+          :assets="displayAssets"
+          :is-selected="isSelected"
+          @select-asset="handleAssetSelect"
+          @approach-end="handleApproachEnd"
+        />
```
christian-byrne, line 105: We should ensure this doesn't exceed screen height on mobile otherwise you won't be able to physically access the generated assets section.

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -0,0 +1,210 @@
+<template>
+  <div class="flex h-full flex-col">
+    <div v-if="activeJobItems.length" class="flex flex-col gap-2 px-2">
+      <AssetsListCard
```
christian-byrne: nit: Could we not name this one `Card`? It seems more like an entry or item.

src/platform/assets/components/AssetsListItem.vue
christian-byrne, line 1: nit (definitely non-blocking): Maybe for future: Can we potentially make this a more generalized `Item` component and then have the `AssetListItem` that implements the more jobs-specific stuff?

src/platform/assets/components/AssetsListItem.vue
```diff
@@ -0,0 +1,102 @@
...
+        />
+      </div>
+    </div>
+
+    <div class="relative z-1 flex min-w-0 flex-1 flex-col gap-1">
+      <div
+        v-if="$slots.primary || primaryText"
+        class="text-xs leading-none text-text-primary"
+      >
+        <slot name="primary">{{ primaryText }}</slot>
```
christian-byrne, line 64: Can we make the text non-selectable if shift or control is down? To prevent text selection when doing Ctrl/Shift selection. Can use https://vueuse.org/core/useMagicKeys/

src/platform/assets/components/AssetsListItem.vue
```diff
@@ -0,0 +1,102 @@
...
+      v-if="
+        progressTotalPercent !== undefined ||
+        progressCurrentPercent !== undefined
+      "
+      class="absolute inset-0"
+    >
+      <div
+        v-if="progressTotalPercent !== undefined"
+        class="pointer-events-none absolute inset-y-0 left-0 h-full bg-interface-panel-job-progress-primary transition-[width]"
+        :style="{ width: `${clampPercent(progressTotalPercent)}%` }"
```
christian-byrne: Can we generalize this to some composable or directive so any component can simply get the progress bars background styles?

src/platform/assets/components/AssetsListItem.vue
```diff
@@ -0,0 +1,102 @@
...
+          aria-hidden="true"
+          :class="
+            cn(
+              iconName ?? 'icon-[lucide--image]',
+              'size-4 text-text-secondary',
+              iconClass
+            )
+          "
+        />
+      </div>
```
christian-byrne: Can we make this a slot so the icons props don't have to be set from the outside in (whoever has asset card has to define 20 properties to add a custom icon).

src/platform/assets/components/AssetsListItem.vue
christian-byrne: Check how https://github.com/unovue/shadcn-vue/tree/dev/apps/v4/registry/new-york-v4/ui/item is organized for reference.

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -0,0 +1,210 @@
...
+        :preview-url="job.iconImageUrl"
+        :preview-alt="job.title"
+        :icon-name="job.iconName"
+        :icon-class="getJobIconClass(job)"
+        :primary-text="job.title"
+        :secondary-text="job.meta"
+        :progress-total-percent="job.progressTotalPercent"
+        :progress-current-percent="job.progressCurrentPercent"
+        @mouseenter="onJobEnter(job.id)"
+        @mouseleave="onJobLeave(job.id)"
```
christian-byrne, line 25: Can we use https://vueuse.org/core/useElementHover/?

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -0,0 +1,210 @@
...
+    listCardBaseClass,
+    'cursor-pointer',
+    selected &&
+      'bg-secondary-background-hover ring-1 ring-inset ring-modal-card-border-highlighted'
+  )
+}
+
+function getJobCardClass(): string {
+  return cn(listCardBaseClass, 'cursor-default')
+}
```
christian-byrne: We can just inline this right?

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -0,0 +1,210 @@
...
+          :preview-alt="item.asset.name"
+          :icon-name="getAssetIconName(item.asset)"
+          :primary-text="getAssetPrimaryText(item.asset)"
+          :secondary-text="getAssetSecondaryText(item.asset)"
+          @click.stop="emit('select-asset', item.asset)"
+        />
+      </template>
+    </VirtualGrid>
+  </div>
+</template>
```
christian-byrne, line 82: Seems like we can just have 2 `VirtualGrid` and clamp the first one's height to half the visible space (to prevent edge-case issue where hundreds of queued jobs make it uncomfortable to go to generated assets section).

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -0,0 +1,210 @@
...
+)
+
+const listGridStyle = {
+  display: 'grid',
+  gridTemplateColumns: 'minmax(0, 1fr)',
+  padding: '0 0.5rem',
+  gap: '0.5rem'
+}
+
+const listCardBaseClass =
```
christian-byrne: Could be good to just inline this

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -0,0 +1,210 @@
...
+  return ''
+}
+
+function getAssetIconName(asset: AssetItem): string {
+  const mediaType = getMediaTypeFromFilename(asset.name)
+  if (mediaType === 'video') return 'icon-[lucide--video]'
+  if (mediaType === 'audio') return 'icon-[lucide--music]'
+  if (mediaType === '3D') return 'icon-[lucide--box]'
+  return 'icon-[lucide--image]'
+}
```
christian-byrne: Would be good to extract this since we already have done so in https://github.com/Comfy-Org/ComfyUI_frontend/commits/b13aca47cdf6d216a6a2f7ae5f78701c4e14e69c/src/utils/queueDisplay.ts

src/composables/queue/useJobActions.ts
christian-byrne, line 1: Would be good to put in domain specific folders rather than `composables`

src/composables/queue/useJobActions.ts
```diff
@@ -0,0 +1,56 @@
...
+    job.showClear === false ? [] : (jobActionSets.value[job.state] ?? [])
+
+  const runJobAction = wrapWithErrorHandlingAsync(
+    async (action: JobAction, job: JobListItem) => {
+      currentJob.value = job
+
+      if (action.key === 'cancel') {
+        await cancelJob()
+      }
+    }
```
christian-byrne: Seems overly refactored if only used for the cancel action. YAGNI

src/composables/queue/useJobActions.ts
```diff
@@ -0,0 +1,56 @@
...
+type JobActionKey = 'cancel'
+
+export type JobAction = {
+  key: JobActionKey
+  icon: string
+  label: string
+  variant: 'destructive' | 'secondary' | 'textonly'
+}
+
+export function useJobActions() {
```
christian-byrne: Seems like it would make more sense if the composable wrapped the job (and took it as an argument).

src/composables/queue/useJobActions.ts
```diff
@@ -0,0 +1,56 @@
...
+  const { t } = useI18n()
+  const { wrapWithErrorHandlingAsync } = useErrorHandling()
+  const currentJob = ref<JobListItem | null>(null)
+  const { cancelJob } = useJobMenu(() => currentJob.value)
```
christian-byrne: Seems like we are doing a lot of round trips for this:

1. runJobAction => sets currentJob
2. The cancelJob relies on getter that returns currentJob
3. Which we then wait on

Is there a cleaner way to organize this?

src/composables/queue/useJobActions.ts
```diff
@@ -0,0 +1,56 @@
...
+      label: t('sideToolbar.queueProgressOverlay.cancelJobTooltip'),
+      variant: 'destructive'
+    }
+
+    return {
+      pending: [cancelAction],
+      initialization: [cancelAction],
+      running: [cancelAction]
+    }
+  })
```
christian-byrne: This is slightly over refactored as well since it just returns a constant object and there's no reactive values internally (besides i18n). YAGNI

src/components/sidebar/tabs/AssetsSidebarListView.vue
DrJKL: https://www.shadcn-vue.com/docs/components/item

src/components/sidebar/tabs/AssetsSidebarTab.vue
benceruleanlu: can this be non blocking and we fix afterwards? most of the vertical height comes from portions of the sidebar tab template and I think we'd also need to change the virtual grid to account for this

src/platform/assets/components/AssetsListItem.vue
benceruleanlu: sure, although I don't think it wouldn't have any consumers because the QPO itself is planned for deletion after this change

benceruleanlu: @coderabbitai can you write an issue tracking this?

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: renamed to ...item

src/platform/assets/components/AssetsListItem.vue
benceruleanlu: I just prevented text selection in general, which matches the grid implementation and is simpler, let me know if we want to gate it to just shift being held

src/platform/assets/components/AssetsListItem.vue
benceruleanlu: done

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: done

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: we can, but we'd have to add a lot more code to support using useElementHover because useElementHover only gives a boolean for whether or not the item is hovered when we need the job id, mouseenter and mouseleave seems to be the better fit here.

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: it has two usages, but I inlined it anyway

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: done

src/composables/queue/useJobActions.ts
benceruleanlu: done

src/composables/queue/useJobActions.ts
benceruleanlu: done

src/composables/queue/useJobActions.ts
benceruleanlu: done

src/composables/queue/useJobActions.ts
benceruleanlu: fixed by two other review comments

src/platform/assets/components/AssetsListItem.vue
benceruleanlu: sure I made it into a slot but existing callers on props

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: image


done

AustinMroz (commented):
Changes look good to me.

packages/shared-frontend-utils/src/formatUtil.ts
christian-byrne, line 1: Does this belong in `formatUtil`?

src/components/sidebar/tabs/AssetsSidebarListView.vue
christian-byrne: @coderabbitai can we add custom args to useElementHover for this use-case?

src/components/sidebar/tabs/AssetsSidebarListView.vue
christian-byrne: That's good, is there no scrollbar in the active jobs when it overflows?

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -0,0 +1,201 @@
...
+          <Button
+            v-for="action in getJobActions(job)"
+            :key="action.key"
+            :variant="action.variant"
+            size="icon"
+            :aria-label="action.label"
+            @click.stop="runCancelJob(job)"
+          >
+            <i :class="action.icon" class="size-4" />
+          </Button>
```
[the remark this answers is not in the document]
christian-byrne: @benceruleanlu This is true right? It's always `runCancelJob(job)` as the click action? Is that intended?

src/components/sidebar/tabs/AssetsSidebarTab.vue
christian-byrne: @coderabbitai make a tracking issue for this and assign to Ben please

src/composables/queue/useJobActions.ts
christian-byrne: The `useJobActions` still doesn't take the job as an argument, I might be misunderstanding.

src/composables/useProgressBarBackground.ts
christian-byrne, line 1: Is there any way to move the state of the progress bar into the composable?

src/components/sidebar/tabs/AssetsSidebarListView.vue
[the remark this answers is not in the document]
benceruleanlu: yeah, the original code was extensible and supported multiple actions, but a review comment said "yagni" because only the cancel button was in the design and the ... menu is another slice of code, so this is the "yagni" implementation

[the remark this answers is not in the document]
benceruleanlu: Cleaned this up so that the action.key and other remnants of the previous system is removed so it's more clear that it's just a cancel button

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: there is, it fades away if not being used, its scrollable

packages/shared-frontend-utils/src/formatUtil.ts
benceruleanlu: probably not, moved out

src/composables/queue/useJobActions.ts
benceruleanlu: I misunderstood, now made the change for useJobActions to take a job ref and wrap it reactively

src/composables/queue/useJobActions.ts
benceruleanlu: Queue is already domain‑specific but nested under a type bucket. Do you mean a domain root like src/queue/*?

christian-byrne: Yes

benceruleanlu: K, [did that](https://github.com/Comfy-Org/ComfyUI_frontend/pull/7924) at the top of the stack to prevent git headaches and keep this one unblocked

draft because I'm going to see how closer I can make it match DDD, it was a first general pass

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -0,0 +1,203 @@
...
+  gridTemplateColumns: 'minmax(0, 1fr)',
+  padding: '0 0.5rem',
+  gap: '0.5rem'
+}
+
+function isActiveJobState(state: JobState): boolean {
+  return (
+    state === 'pending' || state === 'initialization' || state === 'running'
+  )
+}
```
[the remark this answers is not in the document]
christian-byrne: @coderabbitai make a tracking issue for this in the future

src/composables/queue/useJobActions.ts
```diff
@@ -0,0 +1,59 @@
...
+  const { wrapWithErrorHandlingAsync } = useErrorHandling()
+  const { cancelJob } = useJobMenu()
+
+  const cancelAction: JobAction = {
+    icon: 'icon-[lucide--x]',
+    label: t('sideToolbar.queueProgressOverlay.cancelJobTooltip'),
+    variant: 'destructive'
+  }
+
+  const cancellableStates: JobState[] = ['pending', 'initialization', 'running']
```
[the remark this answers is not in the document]
christian-byrne: @coderabbitai make a tracking issue for this in the future

christian-byrne (approved):
LGTM!

PR #7748, opened 2025-12-24
Fix slot renaming in vue

AustinMroz:
Adds an additional check for `slotData.label` so that renamed slots properly display in vue mode

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7748-Fix-slot-renaming-in-vue-2d36d73d365081dc8247f2d9e9e06a7a) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM!

PR #7749, opened 2025-12-24
feat: Vue widget registration for custom extensions

Myestery:
## Summary

- Add `getCustomVueWidgets()` extension hook for registering Vue SFC widgets
- Expose Vue globally as `window.Vue` for extension developers
- Support `display` field in Python input definitions for custom widget type hints
- Add widget registry with alias support

## Test plan

- [x] Unit tests for widget registry (`widgetRegistry.test.ts`)
- [ ] Manual test with `test_vue_widget_node` extension

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7749-feat-Vue-widget-registration-for-custom-extensions-2d36d73d365081d5bf04c708153f224b) by [Unito](https://www.unito.io)


src/utils/vueExtensionApi.ts
```diff
@@ -0,0 +1,19 @@
...
+import * as Vue from 'vue'
+
+declare global {
+  interface Window {
+    Vue: typeof Vue
+  }
+}
+
+export function exposeVueApi(): void {
+  window.Vue = Vue
```
jtydhr88, line 18: I am not sure this solution could work properly, but as my https://github.com/jtydhr88/ComfyUI_frontend_vue_basic is official example for extension developer, we need to make sure this sure this solution can work with example.

Or I will suggest we skip this change in this PR and work on it in another PR.

jtydhr88 (changes requested):

PR #7258, opened 2025-12-09
performance fix: prevent gcd infinite loop with floating-point step values

jtydhr88:
## Summary

report and fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/3919

- Convert recursive gcd to iterative to avoid stack overflow
- Add epsilon tolerance (1e-10) for floating-point precision issues

This fixes workflow loading hangs when node trying merge values like 0.01 and 0.001, which caused the original recursive gcd to run indefinitely due to floating-point modulo never reaching exactly zero.

please notice, we need both iterative and epsilon together to fix this gcd issue

Call Chain

PrimitiveNode.onAfterGraphConfigured
  → #mergeWidgetConfig
    → #isValidConnection
      → mergeIfValid
        → mergeInputSpec
          → mergeNumericInputSpec
            → lcm(step1, step2)
              → gcd(a, b)  ← Problem here

Why It Happened
When some nodes connect to multiple nodes, it may merge values using LCM, which internally calls GCD.

Original recursive implementation:
```
export const gcd = (a: number, b: number): number => {
   return b === 0 ? a : gcd(b, a % b)
}
```

Issues:
1. Stack Overflow: Recursive calls with many nodes exhausted the call stack.
2. Floating-Point Precision: For values like gcd(0.01, 0.001):
 ` 0.01 % 0.001 = 0.0009999999999999994  // Not exactly 0!`
3. Due to Ifloating-point representation, the modulo never reaches exactly zero, causing hundreds or thousands of iterations.

### before

https://github.com/user-attachments/assets/cca4342c-a882-4590-a8d4-1e0bea19e5b7

### fix with only iterative, without epsilon

https://github.com/user-attachments/assets/1dc52aa4-a86a-40b5-8bac-904094c4c36b

### final fix with iterative and epsilon
https://github.com/user-attachments/assets/7b868b50-c3c9-4be4-8594-27cecbc08a26

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7258-performance-fix-prevent-gcd-infinite-loop-with-floating-point-step-values-2c46d73d3650818cbe8cf455c934a114) by [Unito](https://www.unito.io)

PR #7033, opened 2025-11-28
[fix] Add fraction digits props to number input widget

Myestery:
Add min-fraction-digits and max-fraction-digits props to InputNumber
component to allow typing decimal values. Without these props, users
could not manually type values like "0.8" even though the increment
buttons worked correctly.

This aligns the input-only variant with the slider variant which
already had these props configured.

fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/7016
## Screenshots (if applicable)

https://github.com/user-attachments/assets/fa205023-13c4-45ac-874b-b241808cf56d

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7033-fix-Add-fraction-digits-props-to-number-input-widget-2b96d73d36508124ac65e466802bbb0c) by [Unito](https://www.unito.io)

PR #7765, opened 2025-12-26
Improve float widget precision handling

vivekchavan14:
fixes issue https://github.com/Comfy-Org/ComfyUI_frontend/issues/7697
Improves how float widgets determine display precision for v2 nodes so that values are rounded consistently based on step/round configuration, with an option for user override via settings.

Details
- Introduced `computePrecisionFromResolution` to derive precision from a resolution value (e.g. `round` or `step`), with a default when none is provided.
- Prefer the `round` option as the primary resolution when present, falling back to explicit `step` where available.
- Computed an automatic precision from the chosen resolution (e.g. `0.01` → 2 decimal places, integer steps → 0).
- Respect `FloatRoundingPrecision` from settings when set to a value greater than 0, overriding the auto-calculated precision.
- Updated `useFloatWidget` tests to cover:
  - Default precision when no resolution is specified.
  - Precision derived from step/round.
  - Settings-based override behavior.

Testing
- Intended: `pnpm test:unit -- tests-ui/tests/renderer/extensions/vueNodes/widgets/composables/useFloatWidget.test.ts`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7765-Improve-float-widget-precision-handling-2d56d73d3650818da9efdc2ea7b7637d) by [Unito](https://www.unito.io)


tests-ui/tests/renderer/extensions/vueNodes/widgets/composables/useFloatWidget.test.ts
```diff
@@ -77,4 +93,80 @@ describe('useFloatWidget', () => {
       expect(widget.value).toBeCloseTo(3, 10)
     })
   })
+
+  describe('widget constructor precision', () => {
+    const buildNode = () => {
+      const widgets: any[] = []
```
Myestery: we could use  IBaseWidget[] here, there are many optional fields

Myestery (commented):
A few nits
- It will be nice to have no new `any` added
- we could add the precedence tests like coderabbit mentioned

tests-ui/tests/renderer/extensions/vueNodes/widgets/composables/useFloatWidget.test.ts
vivekchavan14: Okay, that makes sense.

vivekchavan14: Done, I replaced any type with IBaseWidget[]

christian-byrne:
Closing this PR — thank you for the contribution @vivekchavan14!

**Why we're closing:**
- Several related PRs have since landed that address float widget precision from different angles: #7768 (custom number nodes), #8992 (gradient-slider widget for FLOAT inputs — directly modified `useFloatWidget.ts`), #7033 (fraction digits props), and #7258 (GCD infinite loop fix for floating-point steps)
- The target file (`useFloatWidget.ts`) has diverged significantly since this PR was opened, resulting in merge conflicts
- Review feedback from @Myestery (remove new `any` types, add precedence tests) was not addressed

If you believe there's still a precision gap not covered by the merged work above, feel free to open a fresh PR against the current `main` branch. Thanks again! 🙏

PR #7768, opened 2025-12-27
Frontend code for custom number nodes

AustinMroz:
Allows creation of Int and Float widgets with configurable, min, max, step, and precision.  
This PR has been fairly heavily reworked.  Options are no longer exposed as widgets, but set as properties on the node.

Since the changes no longer modify the sizing or serialization of the node, backend changes are no longer required and the extended functionality has been added directly onto the existing PrimitiveFloat and PrimitiveInt nodes.

There's intent to expose these configuration parameters on the new properties panel, but this PR can be merged as is.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7768-Frontend-code-for-custom-number-nodes-2d66d73d365081879541cbda7db3c24f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/customWidgets.ts
+++ b/src/extensions/core/customWidgets.ts
@@ -189,7 +189,7 @@ function onCustomFloatCreated(this: LGraphNode) {
 }
 
 app.registerExtension({
-  name: 'Comfy.CustomCombo',
+  name: 'Comfy.CustomWidgets',
   beforeRegisterNodeDef(nodeType, nodeData) {
     if (nodeData?.name === 'CustomCombo')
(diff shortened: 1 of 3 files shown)


src/extensions/core/customWidgets.ts
pythongosssss, line 1: I think we should rename the extension to something clearer now it doesn't just control custom combos - something like CustomWidgetConfig

AustinMroz: Agreed

pythongosssss (approved):

Tekaiguy:
Eygvi1mWQAAPQsV

PR #7775, opened 2025-12-28
fix: viewport overflow in manager

csongorczezar:
### **Summary**
Fixes viewport overflow in version selector dropdown when Manager dialog is positioned near edges

**Changes**
Removed popover arrow for visual consistency across all positions
Implemented dialog boundary detection to constrain popover within Manager viewport

**Testing**
All existing unit tests pass (17/17)
Visually tested across different screen positions

after-fix-viewport-all-positions
before-fix-viewport-low
before-fix-viewport-high

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7775-fix-viewport-overflow-in-manager-2d76d73d365081a88c1df2a103c5925e) by [Unito](https://www.unito.io)


src/workbench/extensions/manager/components/manager/PackVersionBadge.vue
```diff
@@ -104,3 +162,11 @@ watch(
   }
 )
 </script>
+
+<style>
+.no-arrow-popover::before,
+.no-arrow-popover::after,
+.no-arrow-popover .p-popover-arrow {
+  display: none !important;
+}
```
viva-jinyi: Could you use tailwind class instead of using style?

viva-jinyi (changes requested):

src/workbench/extensions/manager/components/manager/PackVersionBadge.vue
csongorczezar: Unfortunately, we can't use Tailwind here because:

PrimeVue's `pt` prop cannot target pseudo-elements (`::before, ::after`)
The popover arrow is rendered using these pseudo-elements
Since the popover uses `append-to="body"`, scoped styles won't work
The `!important` is needed to override PrimeVue's default arrow styles. 
This pattern is also used in WorkflowTabPopover.vue for similar reason.

csongorczezar: Found a solution after all, fix has been applied!

viva-jinyi (approved):
LGTM

PR #7783, opened 2025-12-28
Fixed Brush Settings Post Refactor and Added Numeric Control

brucew4yn3rp:
## Summary
Refactored BrushSettingsPanel layout to stack labels and number inputs above sliders, and fixed brush size keybinding limits to match the updated 1-250 range.

## Changes
- **What**: 
    - Reorganized BrushSettingsPanel UI to display labels and number inputs in a row above each slider (instead of side-by-side), creating a cleaner vertical layout with better visual hierarchy. 
    - Updated brush size increase/decrease keybindings to clamp between 1-250 (previously 1-100) to match the refactored slider limits.
    - Added setting for color picker keybinding
- **Breaking**: None

## Review Focus
- Verify the stacked layout (label + number input above slider) works well across different panel widths
- Confirm all slider controls properly sync with their corresponding number inputs
- Test brush size keybindings (increase/decrease) respect the new 1-250 limits

## Screenshot
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7783-Fixed-Brush-Settings-Port-Refactor-and-Added-Numeric-Control-2d76d73d365081bda7a8e12d3c649085) by [Unito](https://www.unito.io)


src/extensions/core/maskeditor.ts
```diff
@@ -82,13 +82,27 @@ app.registerExtension({
...
+    },
+    {
+      id: 'Comfy.MaskEditor.ColorPicker',
+      icon: 'pi pi-palette',
+      label: 'Open Color Picker in MaskEditor',
+      function: () => {
+        // Trigger the color picker's click event
+        const colorInput = document.querySelector(
+          'input[type="color"]'
+        ) as HTMLInputElement
```
DrJKL: Nit: We're trying to avoid type assertions where possible
```suggestion
        const colorInput = document.querySelector<HTMLInputElement>(
          'input[type="color"]'
        )
```

DrJKL: But also, for commands like this, I think we need the same kind of status gating that's in `changeBrushSize` to make sure that any keybind that would trigger it doesn't work unless the mask editor is open.
It should be possible to bind a ref for the element in the useMaskEditor composable too so that you know precisely which input needs a synthetic click.

src/components/maskeditor/BrushSettingsPanel.vue
```diff
@@ -10,119 +10,208 @@
...
+  onInputChange: (e: Event) => {
+    const value = clamp(Number((e.target as HTMLInputElement).value), min, max)
+    setter(value)
+  }
+})
+
+const thickness = createHandlers((v) => store.setBrushSize(v), 1, 250)
+const opacity = createHandlers((v) => store.setBrushOpacity(v), 0, 1)
+const hardness = createHandlers((v) => store.setBrushHardness(v), 0, 1)
+const stepSize = createHandlers((v) => store.setBrushStepSize(v), 1, 100)
```
DrJKL: Would it be possible to change this to bind the same `v-model` for the slider/numeric pair instead of wiring it through events?
The value in the store should already be `reactive`

src/components/maskeditor/BrushSettingsPanel.vue
```diff
@@ -10,119 +10,208 @@
...
+
+    <!-- Hardness -->
+    <div class="flex flex-col gap-2">
+      <div class="flex items-center justify-between">
+        <span class="text-left text-xs font-sans text-[var(--descrip-text)]">
+          {{ t('maskEditor.hardness') }}
+        </span>
+        <input
+          type="number"
+          class="w-16 px-2 py-1 text-sm text-center border rounded-md bg-[var(--comfy-menu-bg)] border-[var(--p-form-field-border-color)] text-[var(--input-text)]"
```
DrJKL: For all of the utilities, please avoid using arbitrary values if the [utility itself](https://tailwindcss.com/docs/color#basic-example) would work
```suggestion
          class="w-16 px-2 py-1 text-sm text-center border rounded-md bg-comfy-menu-bg border-p-form-field-border-color text-input-text"
```

DrJKL (changes requested):
Overall I like the direction, but I have some changes to request.

src/components/maskeditor/BrushSettingsPanel.vue
brucew4yn3rp: It seems that we can't have v-model used on the numeric inputs if we want to clamp values before they reach the store - meaning without manual @input handlers, users will be able to input numbers outside the min and max bounds. 

Wou can use v-model, but then we'd need to add watch() to clamp values after they hit the store (which allows invalid data momentarily)

DrJKL (approved):
Awesome, thanks 👍🏻 
Approving for my part, @jtydhr88 wdyt?

jtydhr88 (approved):
LGTM

brucew4yn3rp:
Thank you, both.

Hi @trsommer - would you mind reviewing?

brucew4yn3rp:
Hi @christian-byrne @webfiltered, if all other reviews are complete, are we okay to merge in the PR?

jtydhr88:
I think we are good to merge this, the failure test is not about this change, I will merge it with skip tests completed

PR #7798, opened 2025-12-30
fix: keep links hidden under bypassed nodes

### LINKED ISSUE #7755: what can i say.

Can't the connecting lines be placed at the bottom layer?
https://github.com/user-attachments/assets/988eac2f-00e7-4e36-86f5-068a917e8382

.

Multiple times per day

Blocks me from completing tasks

### Current workarounds

_No response_

_No response_

### Additional context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-7755-what-can-i-say-2d36d73d3650811a8325ecbc63ed639c) by [Unito](https://www.unito.io)

jtydhr88:
## Summary
(Not sure we want it, but try to fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/7755 which makes sense to me)

Remove opacity reduction for bypassed nodes so that links remain hidden underneath instead of showing through. The bypass visual is now indicated only by the color overlay.

## Screenshots  - litegraph
before
image

after
image

## Screenshots  - vueNodes

before
image

after 
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7798-fix-keep-links-hidden-under-bypassed-nodes-2d96d73d365081cf9523ed26b65952cf) by [Unito](https://www.unito.io)


PabloWiedemann:
What problem is this solving?

jtydhr88:
@PabloWiedemann Normally, the line is under the node. 
However, when the node is bypassed, the line becomes visible because we added transparency to the bypass. check my screenshot.
This can be problematic for certain users, such as in the scenario described in the issue I mentioned.
This PR removes opacity of node to make sure the line could be always invisible and under the nodes

PR #7802, opened 2025-12-30
Fix reactivity washing in refreshNodeSlots

AustinMroz:
Creating a copy with spread resulted in a copy which was not reactive.

Solves a bug where all widgets on a node in vue mode would cease to be reactive after any connection is made to the node.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7802-Fix-reactivity-washing-in-refreshNodeSlots-2d96d73d3650819e842ff46030bebfa1) by [Unito](https://www.unito.io)


tests-ui/tests/composables/graph/useGraphNodeManager.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    node.addWidget('number', 'testnum', 2, () => undefined, {})
+    graph.add(node)
+
+    const { vueNodeData } = useGraphNodeManager(graph)
+    const onReactivityUpdate = vi.fn()
+    graph.trigger('node:slot-links:changed', {
+      nodeId: '1',
+      slotType: NodeSlotType.INPUT
+    })
+    watch(vueNodeData, onReactivityUpdate)
```
DrJKL: Nit: I think if the common Arrange pieces were pulled out, it might read a bit cleaner here.

Also, want to add the same negative assertion from lines 22-23 above in this one for consistency?

DrJKL (dismissed):
👍🏻

tests-ui/tests/composables/graph/useGraphNodeManager.test.ts
AustinMroz: I have some mixed feelings about the negative assertion even in the first test (I'm not sure it would ever catch anything since there's not an `await nextTick()` to let reactivity setttle) but symmetry is good and establishing that the update should not be caused by the `graph.trigger` call is important.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -245,17 +245,10 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
...
-      return slotInfo ? { ...widget, slotMetadata: slotInfo } : widget
-    })
-
-    vueNodeData.set(nodeId, {
-      ...currentData,
-      widgets: updatedWidgets,
-      inputs: nodeRef.inputs ? [...nodeRef.inputs] : undefined,
-      outputs: nodeRef.outputs ? [...nodeRef.outputs] : undefined
-    })
+      if (slotInfo) widget.slotMetadata = slotInfo
```
DrJKL, line 250: Just noting for the future, but this is the kind of thing that could be done PRN in the context of a WidgetDataStore.

DrJKL (approved):

PR #7807, opened 2025-12-30
fix: spacebar panning in vueNodes mode

### LINKED ISSUE #7806: Spacebar panning in Nodes 2.0

Not sure if I'm missing something, but I used spacebar to pan while dragging a node connection in the litegraph UI, and that doesn't appear to work in Nodes 2.0. I'm not sure what you're supposed to do instead? I've had to instead start zooming out while pointed at one edge of the screen then zoom in pointed at the opposite end just to navigate around.

I noticed the "Navigation Mode" setting which has a new "Standard" mode, and the standard mode maps the scroll wheel to pan... I sincerely hope this is not the intended permanent solution, for a node based UI, where you're panning horizontally just as much as you are vertically. And on top of that the scroll wheel panning stops working if your cursor just happens to hover over a field element.

If spacebar is not an option for whatever reason, I wouldn't even mind using middle click to pan while dragging a connection, as awkward as that is, but that doesn't seem to work either.

While dragging a connection between nodes and attempting to use spacebar or middle click to pan.

Multiple times per day

Significantly slows down my work

### Current workarounds

Zooming in and out or using the mousewheel scroll mode.

_No response_

### Additional context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-7806-Spacebar-panning-in-Nodes-2-0-2d96d73d365081dc9ca8edaec4d01a12) by [Unito](https://www.unito.io)

Myestery:
## Summary
- Forward keydown events to litegraph's `processKey` when Vue nodes have focus (litegraph only binds keydown to canvas)
- Sync pointer state with litegraph canvas during link dragging for spacebar panning while dragging connections
- Add tests for keydown forwarding behavior

## Test plan
- [ ] Enable vueNodes mode and "Drag Navigation" (legacy) mode
- [ ] Select a Vue node, hold spacebar and drag - canvas should pan
- [ ] Drag a link connection, hold spacebar and drag - canvas should pan while link stays attached

Fixes #7806

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7807-fix-spacebar-panning-in-vueNodes-mode-2d96d73d3650813987a1e1f17a6d5bdd) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/composables/useSlotLinkInteraction.ts
```diff
@@ -703,6 +712,10 @@ export function useSlotLinkInteraction({
     )
 
     pointerSession.begin(event.pointerId)
+    if (canvas.pointer) {
+      canvas.pointer.isDown = true
+    }
```
DrJKL: This feels like something that we should not be storing statefully. Can we derive the button state from the MouseEvent `buttons` instead?

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -6,12 +6,34 @@ import { useVueNodeLifecycle } from '@/composables/graph/useVueNodeLifecycle'
...
+    'keydown',
+    (e) => {
+      const canvas = app.canvas
+      if (!canvas) return
+      if (e.target === canvas.canvas) return
+      canvas.processKey(e)
+    },
+    true
+  )
+}
```
DrJKL: I don't want to add more arbitrary event listeners in various composables if we can help it.
Could you use [this](https://vueuse.org/core/useMagicKeys/#usemagickeys)?

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -10,29 +12,51 @@ import { useNodeDrag } from '@/renderer/extensions/vueNodes/layout/useNodeDrag'
...
+// Forward spacebar key events to litegraph for panning when Vue nodes have focus
+let spacebarForwardingInitialized = false
+
+function isEditableElement(el: Element | null): boolean {
+  if (!el) return false
+  const tag = el.tagName.toUpperCase()
+  if (tag === 'INPUT' || tag === 'TEXTAREA') return true
+  if (el instanceof HTMLElement && el.isContentEditable) return true
+  return false
+}
```
DrJKL: Could we consolidate the logic here with the logic in `keybindingService`?

DrJKL: Or just use keybindingService somehow?

Myestery: let me see

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -10,29 +12,51 @@ import { useNodeDrag } from '@/renderer/extensions/vueNodes/layout/useNodeDrag'
...
+
+function initSpacebarForwarding() {
+  if (spacebarForwardingInitialized) return
+  spacebarForwardingInitialized = true
+
+  const { space } = useMagicKeys()
+  const activeElement = useActiveElement()
```
DrJKL: I don't think we need to (or should) have these instantiated conditionally. We can just include them in the composable directly.

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
Myestery: so I just checked and its a bit complicated since it will require keybindingService to handle keyup events and we now need to take away the magickeys and alter graphview too

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -1,13 +1,40 @@
...
+  const canvas = app.canvas
+  if (!canvas) return
+
+  // Skip if canvas has focus (litegraph handles it) or if in editable element
+  if (activeElement.value === canvas.canvas) return
+  if (isEditableElement(activeElement.value || null)) return
+
+  // pointer events will bubble to litegraph
+  canvas.read_only = isPressed
+})
```
DrJKL: Are you sure we need these changes?

Myestery: yeah
We want to filter some events

DrJKL:
Still seeing the "Jump to initial position" issue.

PR #7811, opened 2025-12-31
chore: migrate tests from tests-ui/ to colocate with source files

DrJKL:
## Summary

Migrates all unit tests from `tests-ui/` to colocate with their source files in `src/`, improving discoverability and maintainability.

## Changes

- **What**: Relocated all unit tests to be adjacent to the code they test, following the `<source>.test.ts` naming convention
- **Config**: Updated `vitest.config.ts` to remove `tests-ui` include pattern and `@tests-ui` alias
- **Docs**: Moved testing documentation to `docs/testing/` with updated paths and patterns

## Review Focus

- Migration patterns documented in `temp/plans/migrate-tests-ui-to-src.md`
- Tests use `@/` path aliases instead of relative imports
- Shared fixtures placed in `__fixtures__/` directories

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7811-chore-migrate-tests-from-tests-ui-to-colocate-with-source-files-2da6d73d36508147a4cce85365dee614) by [Unito](https://www.unito.io)


DrJKL:
@coderabbitai review

DrJKL:
@coderabbitai full review

DrJKL:
Awwww :'(

packages/shared-frontend-utils/src/formatUtil.test.ts
```diff
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest'
 
-import { getMediaTypeFromFilename, truncateFilename } from '@/utils/formatUtil'
+import { getMediaTypeFromFilename, truncateFilename } from './formatUtil'
```
AustinMroz, line 3: Weren't we trying to move away from relative import paths in the codebase?

AustinMroz (approved):
That's a lot of linting fixes. :+1:

packages/shared-frontend-utils/src/formatUtil.test.ts
DrJKL: Yes, very much so, except...
https://github.com/Comfy-Org/ComfyUI_frontend/pull/7811/changes#diff-e844d7a368da4f4a1557249eb137591abecbeb8bd86d1e372f0c31ed99817811R5

The relative imports that cause problems are often ones like this, where it's going up, over, and down somewhere else.

Within unit tests, I think that there is some value in clearly marking the modules that are being tested as being adjacent to the test file. But I could also entertain an argument for a universal import format.

AustinMroz (approved):

christian-byrne (approved):

christian-byrne:
Seems like it doesn't actually need to be backported -- all the other backports still worked.

PR #7812, opened 2025-12-31
feat: right side panel favorites, no selection state, and more...

LittleSound:
Most of the features in this pull request are completed and can be reviewed and merged.

## TODO

- [x] no selection panel
- [x] group selected panel
- [x] tabs 
  - [x] favorites tab
  - [x] global settings tab
  - [x] nodes tab
- [x] widget actions menu 
  - [x] [Bug]: style bugs
- [x] button zoom to the node on canvas.
- [x] rename widgets on widget actions
  - [ ] [Bug]: the canvas has not been updated after renaming. 
- [x] global settings
  - [ ] setting item: "show advanced parameters"
    - blocked by other things. skip for now.
  - [x] setting item: show toolbox on selection 
  - [x] setting item: nodes 2.0
  - [ ] setting item: "background color"
    - blocked by other things. skip for now.
  - [x] setting item: grid spacing
  - [x] setting item: snap nodes to grid
  - [x] setting item: link shape
  - [x] setting item: show connected links
  - [x] form style reuses the form style of node widgets
- [x] group node cases
  - [x] group node settings
  - [x] show all nodes in group
  - [x] show frame name on nodes when multiple selections are made
  - [x] group multiple selections
  - [x] [Bug]: nodes without widgets cannot display the location and their group
  - [x] [Bug]: labels layout
- [x] favorites
  - [x] the indicator on widgets
  - [x] favorite and unfavorite buttons on widgets
  - [x] [Bug]: show node name in favorite widgets + improve labels layout
  - [ ] [Bug]: After canceling the like, the like list will not be updated immediately.
  - [x] [Bug]: The favorite function does not work for the project on Subgraph.
- [x] subgraph
  - [x] add the node name from where this parameter comes from when node is subgraph
  - [x] show and hide directly on Inputs
    - [x] some bugs need to be fixed.
- [x] advanced widgets 
  - [x] button: show advanced inputs
    - Clicking button expands the "Advanced Inputs" section on the right side panel, regardless of whether the panel is open or not
    - [x] [Bug]: style bugs
  - [x] advanced inputs section when node is subgraph
- [x] inputs tab rearranging
  - [x] favorited inputs rearranging
  - [x] subgraph inputs rearranging
- [ ] review and reconstruction to improve complexity and architecture

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7812-feat-right-side-panel-favorites-no-selection-state-and-more-2da6d73d36508134b503d676f9b3d248) by [Unito](https://www.unito.io)


src/components/rightSidePanel/parameters/TabNormalInputs.vue
```diff
@@ -0,0 +1,132 @@
...
+    v-if="isSingleSubgraphSelected && advancedInputsWidgets.length > 0"
+    v-model:collapse="advancedInputsCollapsed"
+    :label="t('rightSidePanel.advancedInputs')"
+    :parents="parents"
+    :widgets="advancedInputsWidgets"
+    :default-collapse="advancedInputsCollapsed"
+    class="border-b border-interface-stroke"
+    data-section="advanced-inputs"
+    show-node-name
+  />
```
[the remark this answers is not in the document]
LittleSound: In some cases, this parameter is still necessary.

src/components/rightSidePanel/parameters/WidgetItem.vue
```diff
@@ -0,0 +1,152 @@
...
+    >
+      <i
+        class="absolute -right-1 -top-1 pi pi-star-fill text-xs text-muted-foreground pointer-events-none"
+      />
+    </div>
+    <!-- widget content -->
+    <component
+      :is="widgetComponent"
+      :widget="widget"
+      :model-value="widget.value"
```
AustinMroz:
```suggestion
      :model-value="useReactiveWidgetValue(widget)"
```
A tiny recommendation to make changes on the actual widget update the value displayed in the properties panel as well.

LittleSound: Fixed. Thanks!

src/components/rightSidePanel/RightSidePanel.vue
```diff
@@ -40,29 +56,120 @@ const panelIcon = computed(() =>
...
+
+    const candidateArea = groupRect[2] * groupRect[3]
+    const parentArea = parentRect[2] * parentRect[3]
 
-const selectedNode = computed(() => {
+    if (candidateArea < parentArea) parent = group as LGraphGroup
+  }
+
+  return parent
+}
```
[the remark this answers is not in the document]
LittleSound: Actually, adding this code is related to the issue where group sub-items were not updating correctly after a page refresh. It can be removed after the bug is fixed.

src/components/rightSidePanel/RightSidePanel.vue
```diff
@@ -132,29 +262,36 @@ function handleTitleCancel() {
...
+              @edit="handleTitleEdit"
+              @cancel="handleTitleCancel"
+              @click="isEditing = true"
+            />
+            <i
+              v-if="!isEditing"
+              class="icon-[lucide--pencil] size-4 text-muted-foreground ml-2 content-center relative top-[2px] hover:text-base-foreground cursor-pointer shrink-0"
+              @click="isEditing = true"
+            />
+          </template>
```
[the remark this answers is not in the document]
LittleSound: It should actually be considered to make EditableText accessible via keyboard, rather than a pencil icon.

src/components/rightSidePanel/layout/PropertiesAccordionItem.vue
```diff
@@ -1,23 +1,42 @@
...
 
 const isCollapse = defineModel<boolean>('collapse', { default: false })
+
+if (props.defaultCollapse) {
+  isCollapse.value = true
+}
+watch(
+  () => props.defaultCollapse,
+  (value) => (isCollapse.value = value)
+)
```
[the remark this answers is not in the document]
LittleSound: This doesn't constitute a problem.

src/components/rightSidePanel/parameters/TabGlobalParameters.vue
```diff
@@ -0,0 +1,143 @@
...
+)
+
+const searchedFavoritedWidgets = shallowRef<ValidFavoritedWidget[]>(
+  favoritedWidgets.value
+)
+
+async function searcher(query: string) {
+  isSearching.value = query.trim().length > 0
+  searchedFavoritedWidgets.value = searchWidgets(favoritedWidgets.value, query)
+}
```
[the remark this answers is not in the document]
LittleSound: There is no such question.

src/components/rightSidePanel/parameters/TabNormalInputs.vue
```diff
@@ -0,0 +1,91 @@
...
+  widgetsSectionDataList.value
+)
+const isSearching = ref(false)
+
+async function searcher(query: string) {
+  const list = widgetsSectionDataList.value
+  const target = searchedWidgetsSectionDataList
+  isSearching.value = query.trim() !== ''
+  target.value = searchWidgetsAndNodes(list, query)
+}
```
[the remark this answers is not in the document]
LittleSound: The function must adhere to the interface of the caller.

src/components/rightSidePanel/parameters/WidgetItem.vue
```diff
@@ -0,0 +1,187 @@
...
+
+<template>
+  <div
+    :class="
+      cn(
+        'widget-item col-span-full grid grid-cols-subgrid rounded-lg group',
+        isDraggable &&
+          'draggable-item !will-change-auto drag-handle cursor-grab bg-comfy-menu-bg [&.is-draggable]:cursor-grabbing outline-comfy-menu-bg [&.is-draggable]:outline-4 [&.is-draggable]:outline-offset-0 [&.is-draggable]:opacity-70'
+      )
+    "
```
LittleSound, line 113: This is very important, removing it will cause the selector dropdown to be misaligned and overlap.

src/stores/workspace/rightSidePanelStore.ts
```diff
@@ -27,12 +35,42 @@ export const useRightSidePanelStore = defineStore('rightSidePanel', () => {
...
+    // Scroll to the section after a short delay to ensure the panel is open
+    setTimeout(() => {
+      const sectionElement = document.querySelector(
+        `[data-section="${section}"]`
+      )
+      if (sectionElement) {
+        sectionElement.scrollIntoView({ behavior: 'smooth', block: 'start' })
+      }
+    }, 300)
+  }
```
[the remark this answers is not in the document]
LittleSound: This won't work, we need a delayed operation because expanding the animation requires extra time.

src/stores/workspace/rightSidePanelStore.ts
```diff
@@ -27,12 +43,42 @@ export const useRightSidePanelStore = defineStore('rightSidePanel', () => {
...
```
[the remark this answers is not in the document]
LittleSound: This may not be a problem.

LittleSound:
Fixed a lot of suggestions from coderabbitai. It would take too much time to go through the remaining nitpicks one by one, and blindly adopting them might introduce new bugs.

src/components/button/MoreButton.vue
```diff
@@ -1,6 +1,11 @@
 <template>
   <div class="relative inline-flex items-center">
-    <Button size="icon" variant="secondary" @click="popover?.toggle">
+    <Button
+      size="icon"
+      variant="secondary"
+      :class="buttonClass"
```
christian-byrne: nit: I wonder if just using `v-bind` is better for cases like this on common `MoreButton` type of components. 110% Non-blocking

src/components/rightSidePanel/layout/PropertiesAccordionItem.vue
christian-byrne: nit: this could just use `immediate` right? Or potentially better: shouldn't this just be a `computed`?

src/components/rightSidePanel/layout/PropertiesAccordionItem.vue
```diff
@@ -1,23 +1,42 @@
...
+const props = defineProps<{
   isEmpty?: boolean
+  defaultCollapse?: boolean
+  label?: string
+  enableEmptyState?: boolean
+  noTooltip?: boolean
+  tooltip?: string
 }>()
 
 const isCollapse = defineModel<boolean>('collapse', { default: false })
```
christian-byrne: nit: do we need a `defaultCollapse`? For all cases it's being used, could we just set `collapse=...` on the component?

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -1,33 +1,96 @@
...
+  widgets: widgetsProp,
+  showLocateButton = false,
+  isDraggable = false,
+  hiddenFavoriteIndicator = false,
+  showNodeName = false,
+  defaultCollapse = false,
+  parents = [],
+  enableEmptyState = false,
+  noTooltip = false,
+  tooltip
```
christian-byrne: question (non-blocking): could this be simplified?

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -1,33 +1,96 @@
...
+  noTooltip?: boolean
+  tooltip?: string
 }>()
 
+const collapse = defineModel<boolean>('collapse')
+
+const widgetsContainer = ref<HTMLElement>()
+
+const widgets = shallowRef(widgetsProp)
+watchEffect(() => (widgets.value = widgetsProp))
```
christian-byrne, line 53: question: is there a way to simplify this or make more idiomatic (e.g., using a single "need re-render" version counter ref)?

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -39,49 +102,106 @@ function onWidgetValueChange(
...
+
+const targetNode = computed<LGraphNode | null>(() => {
+  if (node) return node
+  if (isEmpty.value) return null
+
+  const firstNodeId = widgets.value[0].node.id
+  const allSameNode = widgets.value.every(({ node }) => node.id === firstNodeId)
+
+  return allSameNode ? widgets.value[0].node : null
+})
```
christian-byrne, line 99: See above comment about this fallback. If `node` needs to be `undefined`, it seems that we lose a lot of certainty which makes this code hard to think about - and the uncertainty seeps down to all other reliant code.

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -1,33 +1,96 @@
...
+  showNodeName = false,
+  defaultCollapse = false,
+  parents = [],
+  enableEmptyState = false,
+  noTooltip = false,
+  tooltip
+} = defineProps<{
   label?: string
+  parents?: SubgraphNode[]
+  node?: LGraphNode
```
christian-byrne, line 34: In what case would this (`node`) be undefined? Seems like the code would be a lot simpler if we forced it to be defined.

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -39,49 +102,106 @@ function onWidgetValueChange(
...
+            :title="parentGroup.title"
+          >
+            {{ parentGroup.title }}
+          </span>
+        </span>
+        <Button
+          v-if="canShowLocateButton"
+          variant="textonly"
+          size="icon-sm"
+          class="subbutton shrink-0 mr-3 size-8 cursor-pointer text-muted-foreground hover:text-base-foreground"
```
christian-byrne: Is `subbutton` a class or is this code leftover from previous versions of the PR?

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -119,12 +119,59 @@ function getControlWidget(widget: IBaseWidget): SafeControlWidget | undefined {
...
+    borderStyle: widget.promoted
+      ? 'ring ring-component-node-widget-promoted'
+      : widget.advanced
+        ? 'ring ring-component-node-widget-advanced'
+        : undefined,
+    label: widget.label,
+    options: widget.options
+  }
+}
+
```
christian-byrne, line 175: Instead of creating a new `getSharedWidgetEnhancements`, could we just get the graph node manager singleton and get the widget enhancements that way (in `WidgetItem.vue`)? It seems we might be adding some unnecessary complexity with this, but maybe there was another reason this is needed?

src/components/rightSidePanel/RightSidePanel.vue
```diff
@@ -75,53 +185,76 @@ type RightSidePanelTabList = Array<{
...
-  const node = toValue(selectedNode)
+  const node = selectedGroups.value[0] || selectedNodes.value[0]
   if (!node) return
 
   if (trimmedTitle === node.title) return
 
   node.title = trimmedTitle
-  canvasStore.canvas?.setDirty(true, false)
+  canvasStore.canvas?.setDirty(true, true)
+  triggerItems()
```
christian-byrne: It seems strange we have this `triggerItems` for full refreshing but it's only used in one place and it's for title editing. Is there other ways we could approach this?

src/components/rightSidePanel/RightSidePanel.vue
```diff
@@ -40,29 +56,123 @@ const panelIcon = computed(() =>
     : 'icon-[lucide--panel-right]'
 )
 
-const hasSelection = computed(() => selectedItems.value.length > 0)
+// Use shallowRef to avoid deep reactivity overhead
+const flattedItems = shallowRef<MixedSelectionItem[]>([])
+const selectedNodes = shallowRef<LGraphNode[]>([])
+const selectedGroups = shallowRef<LGraphGroup[]>([])
+const nodeToParentGroup = shallowRef<Map<LGraphNode, LGraphGroup>>(new Map())
```
christian-byrne: question: how much could the code be simplified if we dropped `shallowRef` and just relied on reactivity in a more common/simpler way?

Curious what type of perf bottlenecks you were observing when testing this -- I could definitely be wrong!

src/components/rightSidePanel/RightSidePanel.vue
```diff
@@ -40,29 +56,123 @@ const panelIcon = computed(() =>
...
+provide(GetNodeParentGroupKey, (node: LGraphNode) => {
+  if (!shouldShowGroupNames.value) return null
+  return nodeToParentGroup.value.get(node) ?? findParentGroupInGraph(node)
 })
 
-const isSingleNodeSelected = computed(() => selectedNodes.value.length === 1)
+/**
+ * TODO: This traverses the entire graph and could be very slow; needs optimization.
+ */
+function findParentGroupInGraph(node: LGraphNode): LGraphGroup | null {
```
christian-byrne: Do you think this type of function best fits in the view layer (in a component)? Would it be good to move this to the selection service/store/composables?

src/components/rightSidePanel/RightSidePanel.vue
```diff
@@ -40,29 +56,123 @@ const panelIcon = computed(() =>
...
+  return selectedSingleNode.value instanceof SubgraphNode
+})
+
+const selectionCount = computed(() => flattedItems.value.length)
+
+const rootLevelNodes = computed((): LGraphNode[] => {
+  // Depend on activeWorkflow to trigger recomputation when workflow changes
+  void workflowStore.activeWorkflow?.path
+  return (canvasStore.canvas?.graph?.nodes ?? []) as LGraphNode[]
+})
```
christian-byrne: Similar question to [above](https://github.com/Comfy-Org/ComfyUI_frontend/pull/7812/changes#r2677717284) comment

src/stores/workspace/favoritedWidgetsStore.ts
```diff
@@ -0,0 +1,358 @@
...
+    try {
+      const favorites: FavoritedWidgetId[] = favoritedIds.value
+        .map(parseFavoriteKey)
+        .filter((id): id is FavoritedWidgetId => id !== null)
+
+      const data: FavoritedWidgetStorage = { favorites }
+
+      // Ensure extra object exists
+      graph.extra ??= {}
+      graph.extra.favoritedWidgets = data
```
christian-byrne, line 175: note: for future, might be slightly better to just have `graph.extra.favoritedWidgets` instead of an extra layer of indirection and duplication with `favoritedIds` ref.

Would need to make it reactive though and solve some potential issues with serialization, private properties, centralized graph state.

Just noting for future, not blocking on this PR!

src/stores/workspace/favoritedWidgetsStore.ts
christian-byrne, line 1: Nice job on this!

src/stores/workspace/rightSidePanelStore.ts
```diff
@@ -27,12 +44,43 @@ export const useRightSidePanelStore = defineStore('rightSidePanel', () => {
...
+
+    // Scroll to the section after a short delay to ensure the panel is open
+    setTimeout(() => {
+      const sectionElement = document.querySelector(
+        `[data-section="${section}"]`
+      )
+      if (sectionElement) {
+        sectionElement.scrollIntoView({ behavior: 'smooth', block: 'start' })
+      }
+    }, 300)
```
christian-byrne: Can this be done in the component rather than the store? Or would there be a way to use a more Vue-idiomatic approach rather than `querySelector`? Or when we are passing the `section`, can we get the `ref`?

src/components/rightSidePanel/parameters/SectionWidgets.vue
LittleSound: src/components/rightSidePanel/RightSidePanel.vue
```src/components/rightSidePanel/RightSidePanel.vue
group-has-[.subbutton:hover]:text-muted-foreground
```

It is used to prevent child buttons from triggering the interaction effect of the parent button.

src/components/rightSidePanel/parameters/SectionWidgets.vue
LittleSound: Because of the existence of features like favorited lists and subgraphs, it's necessary to consider that each widget might have different nodes. Therefore, a unified node is optional. It didn't even exist in previous versions. Adding it was primarily to provide node title display and locate node functionality.

src/components/button/MoreButton.vue
LittleSound: Sorry, I didn't quite understand. What do you mean?

src/components/rightSidePanel/layout/PropertiesAccordionItem.vue
LittleSound: You are right!

src/components/rightSidePanel/RightSidePanel.vue
LittleSound: I don't think we need to extract this into a composable/store right now. `flatAndCategorizeSelectedItems` is already in `shared.ts` for reusability. we should wait until there's a second consumer before creating `useFlattedSelection`?

src/composables/graph/useGraphNodeManager.ts
LittleSound: Thanks. however, `nodeManager` from `useVueNodeLifecycle()` is only available when Nodes 2.0 is enabled. Since the Right Side Panel needs to work regardless of whether Nodes2.0 is enabled or not, we need to keep `getSharedWidgetEnhancements` as a standalone utility that can extract widget metadata directly from `IBaseWidget` without depending on the `vueNodeData` map.

src/components/button/MoreButton.vue
LittleSound: Oh. I understand. solved.

src/components/rightSidePanel/RightSidePanel.vue
LittleSound: I have simplified this part of the code

src/components/rightSidePanel/parameters/TabNodes.vue
```diff
@@ -0,0 +1,82 @@
...
+import SectionWidgets from './SectionWidgets.vue'
+
+const canvasStore = useCanvasStore()
+const workflowStore = useWorkflowStore()
+
+const nodes = computed((): LGraphNode[] => {
+  // Depend on activeWorkflow to trigger recomputation when workflow changes
+  void workflowStore.activeWorkflow?.path
+  return (canvasStore.canvas?.graph?.nodes ?? []) as LGraphNode[]
+})
```
christian-byrne, line 22: This type of code we can update later when we have better solutions for centralized reactive graph state

src/components/rightSidePanel/settings/SetNodeColor.vue
christian-byrne, line 1: We should explore whether this can use the same functionalities and modules as the other node color picker systems we have. @coderabbitai make a tracking Issue for this

src/components/rightSidePanel/settings/SetNodeState.vue
christian-byrne, line 1: I think we can use the layoutStore for this instead of changing the Litegraph objects directly. @coderabbitai make an Issue to track this

src/components/rightSidePanel/RightSidePanel.vue
christian-byrne, line 1: nit: The selection state logic seems like it might be worth refactoring into a separate module

src/components/rightSidePanel/shared.ts
christian-byrne, line 1: This might be worth using our Fuse library for

src/components/rightSidePanel/shared.ts
```diff
@@ -0,0 +1,271 @@
...
+ * Renames a widget and its corresponding input.
+ * Handles both regular widgets and proxy widgets in subgraphs.
+ *
+ * @param widget The widget to rename
+ * @param node The node containing the widget
+ * @param newLabel The new label for the widget (empty string or undefined to clear)
+ * @param parents Optional array of parent SubgraphNodes (for proxy widgets)
+ * @returns true if the rename was successful, false otherwise
+ */
+export function renameWidget(
```
christian-byrne, line 219: This might be worth moving to a higher level module so all parts of the app can use it -- i.e., just moving outside of the `rightSidePanel` folder.

christian-byrne (dismissed):
Let's merge this for now so followup work is easier to manage. The PR is too large now, and we just went to 1.38 so there is plenty of time to iterate.

src/components/rightSidePanel/settings/SetNodeColor.vue
christian-byrne: @coderabbitai you didn't make a new issue. 8023 already exists and is something else.

LittleSound:
It appears that the failed E2E tests are related to the right-click menu. It might be because the right-click menu has added the Widgets favorite feature.

christian-byrne (dismissed):

PR #7815, opened 2025-12-31
fix 3d-min-resize

kaili-yang:
## Summary

Set up a minimum height on the viewer container to prevent layout collapse caused by the absolutely positioned canvas.

## Changes

What: Updated handleResize apply a minimum height style to the parent element

## Screenshots 
before

https://github.com/user-attachments/assets/0ddb2681-3083-4dfe-b59f-77724072002d

after

https://github.com/user-attachments/assets/672570ad-8792-4e09-8050-6dfcb921cd36

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7815-fix-3d-min-resize-2da6d73d3650811ca66fd82ad13258b9) by [Unito](https://www.unito.io)


src/extensions/core/load3d/Load3d.ts
```diff
@@ -629,9 +629,11 @@ class Load3d {
       console.warn('Parent element not found')
       return
     }
+    const MIN_HEIGHT = 200
+    parentElement.style.minHeight = `${MIN_HEIGHT}px`
 
     const containerWidth = parentElement.clientWidth
-    const containerHeight = parentElement.clientHeight
+    const containerHeight = Math.max(parentElement.clientHeight, MIN_HEIGHT)
```
jtydhr88: can we do it in css in its vue componemnt? for example https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/components/load3d/Load3DScene.vue or somewhere?

jtydhr88 (changes requested):

src/extensions/core/load3d/Load3d.ts
kaili-yang: Thanks for the suggestion. I’ll implement this in CSS within the Vue component.

jtydhr88 (approved):
LGTM

PR #7825, opened 2026-01-02
feat: Add visual crop preview widget for ImageCrop node - widget ImageCrop

jtydhr88:
## Summary

Another implementation for image crop node, alternative for https://github.com/Comfy-Org/ComfyUI_frontend/pull/7014
As discussed with @christian-byrne and @DrJKL we could have single widget - IMAGECROP with 4 ints and UI preview.

However, this solution requires changing the definition of image crop node in BE (sent [here](https://github.com/comfyanonymous/ComfyUI/pull/11594)), which will break the exsiting workflow, also it would not allow connect separate int node as input, I am not sure it is a good idea.

So I keep two PRs openned for references

## Screenshots

https://github.com/user-attachments/assets/fde6938c-4395-48f6-ac05-6282c5eb8157

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7825-feat-Add-visual-crop-preview-widget-for-ImageCrop-node-widget-ImageCrop-2dc6d73d3650812bb8a2cdff4615032b) by [Unito](https://www.unito.io)


Tekaiguy:
If possible, I would request adding a **ratio** input which accepts values like "1.00" or "16:9" to lock the aspect ratio while resizing/dragging. I do a lot of crop work and it is a huge time saver.

The current best custom implementation of visual crop is [here](https://github.com/o-l-l-i/ComfyUI-Olm-DragCrop), if you want to take a look. Yours is less visually cluttered though, which I like!

RandomGuyWithIssues:
Lets say i detect a face in some image and get bbox location, i want to crop that location from the image, do some sampling and return that crop back to the original image. How can i do it with this node if only input and output are image and no other data?

src/composables/useImageCrop.ts
```diff
@@ -0,0 +1,477 @@
...
+
+  const syncFromModelValue = () => {
+    const value = modelValue.value
+    cropX.value = value.x
+    cropY.value = value.y
+    cropWidth.value = value.width || 512
+    cropHeight.value = value.height || 512
+  }
+
+  const syncToModelValue = () => {
```
christian-byrne: It seems strange to have a sort of flush/sync function on a model value. Can we simplify by making the underlying values model or using a more idiomatic approach with computed values?

src/composables/useImageCrop.ts
christian-byrne, line 1: Wondering: are there any libraries or component libraries that have this type of component that we could use? (non-blocking)

src/lib/litegraph/src/widgets/ImageCropWidget.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    ctx.strokeStyle = this.outline_color
+    ctx.strokeRect(15, y, width - 30, height)
+
+    ctx.fillStyle = this.text_color
+    ctx.font = '11px monospace'
+    ctx.textAlign = 'center'
+    ctx.textBaseline = 'middle'
+
+    const text = 'ImageCrop: Vue-only'
```
christian-byrne: Can we change to "Nodes 2.0 only"? And also add a note about how to enable Nodes 2.0?

src/composables/useImageCrop.ts
jtydhr88: fixed

src/components/imagecrop/WidgetImageCrop.vue
```diff
@@ -0,0 +1,142 @@
...
+
+import { useImageCrop } from '@/composables/useImageCrop'
+import type { BoundingBoxValue } from '@/lib/litegraph/src/types/widgets'
+import type { NodeId } from '@/platform/workflow/validation/schemas/workflowSchema'
+
+const props = defineProps<{
+  nodeId: NodeId
+}>()
+
+const modelValue = defineModel<BoundingBoxValue>({
```
DrJKL: Can you share the code with the widget above?

src/lib/litegraph/src/widgets/ImageCropWidget.ts
```diff
@@ -0,0 +1,49 @@
...
+      fillStyle,
+      strokeStyle,
+      textAlign,
+      textBaseline,
+      font
+    })
+  }
+
+  onClick(_options: WidgetEventOptions): void {
+    // This is a widget that only has a Vue widgets implementation
```
DrJKL: This would be a fun place for an Easter egg.

src/locales/en/main.json
```diff
@@ -1710,6 +1710,21 @@
     "unsupportedFileType": "Unsupported file type (supports .gltf, .glb, .obj, .fbx, .stl)",
     "uploadingModel": "Uploading 3D model..."
   },
+  "imageCrop": {
+    "loading": "Loading...",
+    "noInputImage": "No input image connected",
+    "cropPreviewAlt": "Crop preview",
+    "x": "x",
```
DrJKL: Can we remove these from here and delegate to the boundingBox keys?

src/renderer/extensions/vueNodes/widgets/composables/useBoundingBoxWidget.ts
```diff
@@ -0,0 +1,81 @@
...
+          if (subWidget) {
+            subWidget.value = (widget.value as BoundingBoxValue)[field]
+          }
+        }
+      },
+      {
+        serialize: true,
+        canvasOnly: false
+      }
+    ) as IBoundingBoxWidget | IImageCropWidget
```
DrJKL: I think there should be a way to get the correct types without assertions

src/schemas/nodeDef/nodeDefSchemaV2.ts
```diff
@@ -64,6 +64,21 @@ const zImageCompareInputSpec = zBaseInputOptions.extend({
   options: z.record(z.unknown()).optional()
 })
 
+const zBoundingBoxInputSpec = zBaseInputOptions.extend({
+  type: z.literal('BOUNDINGBOX'),
+  name: z.string(),
+  isOptional: z.boolean().optional(),
+  component: z.enum(['ImageCrop']).optional(),
```
DrJKL, line 71: I like it. I'll think of how we can make this piece extensible for custom node devs to hook into.

src/components/boundingbox/WidgetBoundingBox.vue
```diff
@@ -0,0 +1,78 @@
...
+  <div class="grid grid-cols-[auto_1fr] gap-x-2 gap-y-1">
+    <label class="content-center text-xs text-node-component-slot-text">
+      {{ $t('boundingBox.x') }}
+    </label>
+    <input
+      v-model.number="x"
+      type="number"
+      :min="0"
+      class="h-7 rounded-lg border-none bg-component-node-widget-background px-2 text-xs text-component-node-foreground focus:outline-0"
+    />
```
DrJKL, line 12: It's probably worth adding `step="1"` to these.
Might be premature to extract a component, but I wouldn't object.

src/components/boundingbox/WidgetBoundingBox.vue
```diff
@@ -0,0 +1,78 @@
...
+  set: (val) => {
+    modelValue.value = { ...modelValue.value, width: val }
+  }
+})
+
+const height = computed({
+  get: () => modelValue.value.height,
+  set: (val) => {
+    modelValue.value = { ...modelValue.value, height: val }
+  }
```
DrJKL: Very nit, but if you use the key name you can also use shorthand for the object construction.
```suggestion
  set: (height) => {
    modelValue.value = { ...modelValue.value, height }
  }
```

src/composables/useImageCrop.ts
DrJKL: Same question. I'd love to be able to reduce the code we have to maintain.

src/lib/litegraph/src/types/widgets.ts
```diff
@@ -259,6 +261,32 @@ export interface IAssetWidget extends IBaseWidget<
   value: string
 }
 
+/** Bounding box value with x, y, width, height */
+export interface BoundingBoxValue {
+  x: number
+  y: number
+  width: number
+  height: number
+}
```
DrJKL: Just a quick check, is there already an interface with this signature that we could reference?

src/lib/litegraph/src/widgets/BoundingBoxWidget.ts
```diff
@@ -0,0 +1,45 @@
...
+  extends BaseWidget<IBoundingBoxWidget>
+  implements IBoundingBoxWidget
+{
+  override type = 'boundingbox' as const
+
+  drawWidget(ctx: CanvasRenderingContext2D, options: DrawWidgetOptions): void {
+    const { width } = options
+    const { y, height } = this
+
+    const { fillStyle, strokeStyle, textAlign, textBaseline, font } = ctx
```
DrJKL: Could you use save/restore instead?

src/lib/litegraph/src/types/widgets.ts
jtydhr88: good catch, we have two same signature but in different places:
1. [types.ts](https://github.com/Comfy-Org/ComfyUI_frontend/blob/d3bd85db7f9e0ec202dfb00a80b2a3d8a67f84f0/src/renderer/core/layout/types.ts#L28)
2. [QuadTree.ts](https://github.com/Comfy-Org/ComfyUI_frontend/blob/d3bd85db7f9e0ec202dfb00a80b2a3d8a67f84f0/src/renderer/core/spatial/QuadTree.ts#L10)

src/lib/litegraph/src/widgets/BoundingBoxWidget.ts
jtydhr88: fixed

src/lib/litegraph/src/types/widgets.ts
jtydhr88: use first one instead

src/components/boundingbox/WidgetBoundingBox.vue
jtydhr88: fixed

src/components/boundingbox/WidgetBoundingBox.vue
jtydhr88: fixed

src/lib/litegraph/src/widgets/ImageCropWidget.ts
```diff
@@ -0,0 +1,49 @@
...
+  extends BaseWidget<IImageCropWidget>
+  implements IImageCropWidget
+{
+  override type = 'imagecrop' as const
+
+  drawWidget(ctx: CanvasRenderingContext2D, options: DrawWidgetOptions): void {
+    const { width } = options
+    const { y, height } = this
+
+    const { fillStyle, strokeStyle, textAlign, textBaseline, font } = ctx
```
DrJKL: Same here.
This seems common enough now to extract as a utility, or even to consolidate as something like `VueOnlyWidgetWarning`

jtydhr88: fixed

DrJKL (approved):

src/composables/useImageCrop.ts
```diff
@@ -0,0 +1,473 @@
...
+  const handleImageError = () => {
+    isLoading.value = false
+    imageUrl.value = null
+  }
+
+  const capturePointer = (e: PointerEvent) =>
+    (e.target as HTMLElement).setPointerCapture(e.pointerId)
+
+  const releasePointer = (e: PointerEvent) =>
+    (e.target as HTMLElement).releasePointerCapture(e.pointerId)
```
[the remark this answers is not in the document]
DrJKL: @coderabbitai Open an issue to fix all similar casts in the code.

PR #7841, opened 2026-01-04
Added MaskEditor Rotate and Mirror Functions

brucew4yn3rp:
# Canvas Rotation and Mirroring

## Overview
Adds rotation (90° left/right) and mirroring (horizontal/vertical) capabilities to the mask editor canvas. All three layers (image, mask, RGB) transform together. Redo and Undo respect transformations as new states. Keyboard shortcuts also added for all four functions in Keybinding settings.

Additionally, fixed the issue of ctrl+z and ctrl+y keyboard commands not restricting to the mask editor canvas while opened. 

https://github.com/user-attachments/assets/fb8d5347-b357-4a3a-840a-721cdf8a6125

### New Files
- **`src/composables/maskeditor/useCanvasTransform.ts`**
  - Core transformation logic for rotation and mirroring
  - GPU texture recreation after transformations

#### **`src/composables/useCoreCommands.ts`**
- Added check to see if Mask Editor is opened for undo and redo commands

#### **`src/stores/maskEditorStore.ts`**
- Added GPU texture recreation signals

#### **`src/composables/maskeditor/useBrushDrawing.ts`**
- Added watcher for `gpuTexturesNeedRecreation` signal
- Handles GPU texture recreation when canvas dimensions change
- Recreates textures with new dimensions after rotation
- Updates preview canvas and readback buffers accordingly
- Ensures proper ArrayBuffer backing for WebGPU compatibility

#### **`src/components/maskeditor/TopBarHeader.vue`**
- Added 4 new transform buttons with icons:
  - Rotate Left (counter-clockwise)
  - Rotate Right (clockwise)
  - Mirror Horizontal
  - Mirror Vertical
- Added visual separators between button groups

#### **`src/extensions/core/maskEditor.ts`**
- Added keyboard shortcut settings for rotate and mirror

#### **Translation Files** (e.g., `src/locales/en.json`)
- Added i18n keys:

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7841-Added-MaskEditor-Rotate-and-Mirror-Functions-2de6d73d365081bc9b84ea4919a3c6a1) by [Unito](https://www.unito.io)


src/components/maskeditor/MaskEditorContent.vue
```diff
@@ -4,6 +4,7 @@
     class="maskEditor-dialog-root flex h-full w-full flex-col"
     @contextmenu.prevent
     @dragstart="handleDragStart"
+    @keydown.stop
```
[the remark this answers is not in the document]
brucew4yn3rp: This is incorrect. The keydowns are prevented from reaching outside Mask Editor when opened, but all keyboard shortcuts within Mask Editor work as expected.

trsommer (dismissed):
works as described. Only thing I can see is that rotation and undo/redo buttons are very similar in design. That could lead to some confusion

brucew4yn3rp:
Thank you @trsommer. Just updated the rotate buttons with a different SVG for ease of interpretation. 

image

src/components/maskeditor/dialog/TopBarHeader.vue
```diff
@@ -35,6 +35,74 @@
           </svg>
         </button>
 
+        <div class="h-5 w-px bg-[var(--p-form-field-border-color)]" />
```
DrJKL: I need to add a linter or something to check for these.
Can you use the direct utilities instead of the `[var(...)]` ones?

src/composables/maskeditor/useCanvasHistory.ts
```diff
@@ -1,12 +1,17 @@
 import { ref, computed } from 'vue'
 import { useMaskEditorStore } from '@/stores/maskEditorStore'
 
+// Define the state interface for better readability
+interface CanvasState {
+  mask: ImageData | ImageBitmap
+  rgb: ImageData | ImageBitmap
+  img: ImageData | ImageBitmap
```
DrJKL, line 8: Any chance we can consolidate on a single form for these?

DrJKL: Not necessarily in this PR, more just asking because you'd probably know if it's feasible.

src/composables/maskeditor/useCanvasHistory.ts
```diff
@@ -87,71 +115,86 @@ export function useCanvasHistory(maxStates = 20) {
...
 
-    if (state.rgb instanceof ImageBitmap) {
-      rgbCtx.clearRect(0, 0, state.rgb.width, state.rgb.height)
-      rgbCtx.drawImage(state.rgb, 0, 0)
-    } else {
-      rgbCtx.putImageData(state.rgb, 0, 0)
+    // Update canvas dimensions to match state (handles rotation undo/redo)
+    const refData = state.mask
+    const newWidth =
+      refData instanceof ImageBitmap ? refData.width : refData.width
```
DrJKL: Why a ternary with identical branches?

src/composables/maskeditor/useCanvasTransform.ts
```diff
@@ -0,0 +1,270 @@
...
+export function useCanvasTransform() {
+  const store = useMaskEditorStore()
+
+  /**
+   * Rotates a canvas 90 degrees clockwise or counter-clockwise
+   */
+  const rotateCanvas = (
+    ctx: CanvasRenderingContext2D,
+    canvas: HTMLCanvasElement,
+    clockwise: boolean
```
DrJKL, line 15: I have a general aversion to positional boolean arguments (including the unary ones down below)...

Since these aren't exposed, it's probably okay for now, though I wouldn't mind using an options object with a default or a string enum (also with a default).

For the `AllLayers` ones though, given the way they're separately bound and not dependent on some other boolean that's passed through, what do you think about just splitting them into `rotateCanvasClockwise`, `rotateCanvasCounterclockwise` for the consumers of the composable?

That way they don't need to worry about the fact that there are multiple layers to transform and also don't have to dive in to see which state the boolean's truthiness represents.

src/composables/maskeditor/useCanvasTransform.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    // Copy data
+    maskData.set(maskImageData.data)
+    rgbData.set(rgbImageData.data)
+    /*
+    if (maskData.buffer instanceof SharedArrayBuffer || rgbData.buffer instanceof SharedArrayBuffer) {
+      console.error('[useCanvasTransform] SharedArrayBuffer detected, WebGPU writeTexture will fail')
+      return
+    }
+    */
```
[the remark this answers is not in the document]
DrJKL: I agree with removing commented out code.

src/composables/maskeditor/useCanvasTransform.ts
```diff
@@ -0,0 +1,270 @@
...
+    // Premultiply alpha for GPU
+    premultiplyData(maskData)
+    premultiplyData(rgbData)
+
+    // Store the premultiplied data for useBrushDrawing to pick up
+    // Cast to ensure ArrayBuffer backing (not SharedArrayBuffer)
+    store.pendingGPUMaskData = maskData as Uint8ClampedArray & {
```
DrJKL: The casting just tricks the type checker. Is there a way to guarantee the type further up the chain?

DrJKL:
I left a few questions and an optional plea for a change in the public signature of the composable, but it looks good. Works for me locally.
Waiting to approve until the tests are fixed 👍🏻

src/composables/maskeditor/useCanvasHistory.ts
brucew4yn3rp: The type is ImageData | ImageBitmap because after rotations/transforms, the code stores the result which could be either type depending on the operation so keeping separate is cleanest imo

src/composables/maskeditor/useCanvasTransform.ts
brucew4yn3rp: Great point. Public API has been split into rotateClockwise(), rotateCounterclockwise(), mirrorHorizontal(), and mirrorVertical(). The internal helper functions (rotateCanvas, mirrorCanvas) still use booleans since they're not exposed

brucew4yn3rp:
Thank you for catching some of the improvements and the unnecessary ternary, @DrJKL. Made the changes you mentioned but unfortunately don't think we can consolidate the mask, rgb, and img ImageData | ImageBitmap.

src/composables/maskeditor/useCanvasHistory.ts
DrJKL: Oh well 👍🏻 
Thanks for looking into it.

DrJKL:
Looks like the unit tests need to be updated, otherwise should be good.

brucew4yn3rp:
@DrJKL Updated `useCanvasHistory.test.ts` to track all three layers and added a `useCanvasTransform.test.ts`

src/composables/maskeditor/useCanvasTransform.test.ts
```diff
@@ -0,0 +1,656 @@
...
+  rgbCtx: IMockContext | null
+  imgCtx: IMockContext | null
+  tgpuRoot: unknown
+  canvasHistory: IMockCanvasHistory
+  gpuTexturesNeedRecreation: boolean
+  gpuTextureWidth: number
+  gpuTextureHeight: number
+  pendingGPUMaskData: Uint8ClampedArray | null
+  pendingGPURgbData: Uint8ClampedArray | null
+}
```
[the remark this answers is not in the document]
brucew4yn3rp: The as Partial<MaskEditorStore> as MaskEditorStore approach was causing TypeScript errors because canvasHistory had incompatible types between the mock and the full store type. Ignoring this suggestion.

brucew4yn3rp:
@DrJKL I think the tests are fine as they are now and they all seem to pass. Mind reviewing the real changes and approving if satisfactory?

DrJKL (approved):

DrJKL:
Looks great, thanks!

DrJKL:
@brucew4yn3rp How do you feel about coderabbit so far?

brucew4yn3rp:
@DrJKL I think it's a useful tool to provide a first layer of review and help identify inefficiencies or potential nascent issues, but at the same time it almost tries too hard and implementing some of its suggestions would either not be practical or not work outright.

DrJKL:
@brucew4yn3rp If you have any specific awkward suggestions to call out, we can try to edit the AGENTS files to better guide it.

PR #7846, opened 2026-01-05
feat: use Web Worker for OBJ loading to prevent UI blocking

### LINKED ISSUE #7843: Loading large 3D models should not block the entire UI
When attempting to load large 3D models (e.g., larger than 50MB), although the Load3D node displays a loading indicator, the entire ComfyUI UI freezes during this process, which significantly impacts user experience. This should be implemented as asynchronous loading. 

Additionally, through testing, files larger than approximately 200MB cannot be loaded at all, and there is currently no error message or notification to inform users of this limitation.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-7843-Loading-large-3D-models-should-not-block-the-entire-UI-2de6d73d365081538cbde869714057a2) by [Unito](https://www.unito.io)

jtydhr88:
## Summary

- Replace OBJLoader with OBJLoader2Parallel from wwobjloader2
- OBJ parsing now runs in a Web Worker, keeping UI responsive
- Add 100MB file size limit with user-friendly error message

reduce loading time for 97M obj from 9s to 3s

fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/7843

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7846-feat-use-Web-Worker-for-OBJ-loading-to-prevent-UI-blocking-2df6d73d36508140bea3c87e83d11278) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/load3d/Load3dUtils.ts
+++ b/src/extensions/core/load3d/Load3dUtils.ts
@@ -34,8 +34,25 @@ class Load3dUtils {
     return await resp.json()
   }
 
+  static readonly MAX_UPLOAD_SIZE_MB = 100
+
   static async uploadFile(file: File, subfolder: string) {
     let uploadPath
+
+    const fileSizeMB = file.size / 1024 / 1024
+    if (fileSizeMB > this.MAX_UPLOAD_SIZE_MB) {
+      const message = t('toastMessages.fileTooLarge', {
+        size: fileSizeMB.toFixed(1),
+        maxSize: this.MAX_UPLOAD_SIZE_MB
+      })
+      console.warn(
+        '[Load3D] uploadFile: file too large',
+        fileSizeMB.toFixed(2),
+        'MB'
+      )
+      useToastStore().addAlert(message)
+      return undefined
+    }
 
     try {
       const body = new FormData()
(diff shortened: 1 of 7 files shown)


src/extensions/core/load3d/Load3dUtils.ts
```diff
@@ -34,9 +34,26 @@ class Load3dUtils {
...
   }
 
+  static readonly MAX_UPLOAD_SIZE_MB = 100
+
   static async uploadFile(file: File, subfolder: string) {
     let uploadPath
 
+    const fileSizeMB = file.size / 1024 / 1024
+    if (fileSizeMB > this.MAX_UPLOAD_SIZE_MB) {
+      const message = t('toastMessages.fileTooLarge', {
```
benceruleanlu, line 44: nit: we moved to usei18n for t here, so the import might want changing

src/extensions/core/load3d/Load3dUtils.ts
```diff
@@ -34,9 +34,26 @@ class Load3dUtils {
...
+        size: fileSizeMB.toFixed(1),
+        maxSize: this.MAX_UPLOAD_SIZE_MB
+      })
+      console.warn(
+        '[Load3D] uploadFile: file too large',
+        fileSizeMB.toFixed(2),
+        'MB'
+      )
+      useToastStore().addAlert(message)
```
benceruleanlu, line 53: nit: all but one downstream caller handles undefined as well by showing their own toast, which causes a double toast ux annoyance on this case

benceruleanlu (approved):
lgtm, small nit is just be careful of uploadFile's return type being any

PR #7852, opened 2026-01-05
Subgraph/workflow breadcrumbs menu updates

pythongosssss:
## Summary
For users who don't use subgraphs, the workflow name in the top left can be unnecessarily obstructive so this updated collapses it to a simple icon until a subgraph is entered.

## Changes

- Add menu button to WorkflowTab for quick workflow actions
- Add menu and back button to SubgraphBreadcrumb
- Extract shared menu items to useBreadcrumbMenu composable
- Add Comfy.RenameWorkflow command for renaming persisted workflows
- Menu always shows root workflow menu, even when in subgraph

image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7852-Subgraph-workflow-breadcrumbs-menu-updates-2df6d73d3650815b8490ca0a9a92d540) by [Unito](https://www.unito.io)


comfydesigner:
Is it possible to also put "Export, "Export (API)", and "Bookmark" into this menu as well?

Screenshot 2026-01-08 at 11 17 06 AM

Into

image

comfydesigner:
Thoughts on also showing this menu on right clicking the tab too? We should utilize right click more (long press on mobile)
Screenshot 2026-01-08 at 11 26 08 AM

To keep the functionality I suppose we should throw the other tab closing options in here as well, but I'm not sure how often they're used. They're more dangerous than just closing the current workflow, so placing them at the bottom away from the other actions is preferred. Maybe these extra tab options should only show up when right clicking? Thoughts?
Screenshot 2026-01-08 at 11 26 23 AM

comfydesigner:
Speedy right clicks show a context menu that would be hidden under the tooltip or workflow preview. No issues when manually clicking on the "menu" button of an active tab though

Screenshot 2026-01-10 at 6 54 57 PM

comfydesigner:
image

Perhaps when right clicking on a tab, it should close the older menu that was triggered from the button?
Thanks for adding better distinguishing icons for the "Close tab" variants btw.

------------------
(Also a future iteration for the close tab variants could be to hide them behind a submenu? No need to take action on this yet)

I could see the list as:

| Clear workflow                |
| Close specific tabs    >  |
| Close tab                         |
Screenshot 2026-01-10 at 7 00 31 PM

christian-byrne (dismissed):

christian-byrne:
@pythongosssss Sorry there is a conflict after merging https://github.com/Comfy-Org/ComfyUI_frontend/pull/7853 (seemingly)

christian-byrne (approved):
LGTM

PR #7853, opened 2026-01-05
Integrated tab bar UI elements

pythongosssss:
## Summary

The current help / feedback is often overlooked by users, this adds a setting that makes it more visible moving it up into the tab bar and moves the user login/profile button out of the "action bar" into the tab bar.

## Changes
- Add 'Comfy.UI.TabBarLayout' setting with Default/Integrated options
- Move Help & User controls to tab bar when Integrated mode is enabled
- Extract help center logic into shared useHelpCenter composable

image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7853-Integrated-tab-bar-UI-elements-2df6d73d365081b1beb8f7c641c2fa43) by [Unito](https://www.unito.io)


comfydesigner:
Thanks for putting this behind a settings option!
Unrelated to this PR, but in the future should we organize all these into a collection? Perhaps these can be flagged as "beta" and that could be the collection to sort by to see/test out the latest changes. 

As for the PR, the user account button is a bit cramped on my side vs what I see in your screenshot. Is this just an issue on my end? Using Chrome
Screenshot 2026-01-08 at 11 10 15 AM

christian-byrne (approved):

PR #7860, opened 2026-01-06
fix RightSidePanel title not updating when litegraph node title changes

pythongosssss:
## Summary

When editing a node title via the LiteGraph node with the right panel open, the title is not reactively updated in the panel UI.

## Changes

- Used `useVueNodeLifecycle` to allow reactive access to the title

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7860-fix-RightSidePanel-title-not-updating-when-litegraph-node-title-changes-2e06d73d365081ae84a2f08305c854f8) by [Unito](https://www.unito.io)


AustinMroz (commented):
- The description seems a little misleading since the reactivity is only fixed in vue mode (where litegraph isn't used)
- There's some pretty significant work being done on the right side panel over in #7812

I'd want a second opinion here, but actual content of the PR is an improvement and LGTM

PR #7862, opened 2026-01-06
fix: prevent image preview resize issues when switching to vueNodes mode

jtydhr88:
## Summary
- Fix infinite resize growth issue when dragging resize handle in vueNodes mode by adding isResizingVueNodes guard to ResizeObserver callback
- Fix duplicate rendering issue for image preview nodes when switching from litegraph to vueNodes mode by setting canvasOnly: true on ImagePreviewWidget

## Problem

 When switching from litegraph to vueNodes mode, image preview nodes (LoadImage, PreviewImage) had two issues:

  1. Node becoming longer: The ImagePreviewWidget was being rendered twice - once as a WidgetLegacy canvas (with stale computedHeight from litegraph mode) and once as Vue's ImagePreview component
  2. Resize feedback loop: During active resize, ResizeObserver would continuously fire and update the layout store, causing potential infinite growth

## Solution

  1. Added isResizingVueNodes check in ResizeObserver to skip processing during active resize operations
  2. Set canvasOnly: true for ImagePreviewWidget so it won't render as WidgetLegacy in Vue mode (Vue's ImagePreview.vue already handles image display)

before

https://github.com/user-attachments/assets/925c4fb4-bc9a-4da5-b8ae-3557c2d3836b

after

https://github.com/user-attachments/assets/5faa6878-c56d-44dd-86f5-728bff9ad58a

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7862-fix-prevent-image-preview-resize-issues-when-switching-to-vueNodes-mode-2e06d73d36508122ae86d134d03dffe3) by [Unito](https://www.unito.io)


browser_tests/tests/widget.spec.ts-snapshots/image-preview-changed-by-combo-value-chromium-linux.png
DrJKL, line 1: Is it rendering a different image than is selected now?

PR #7879, opened 2026-01-07
fix: use pre-bundled wwobjloader2 worker for production builds

jtydhr88:
## Summary

The unbundled worker from 'wwobjloader2/worker' has ES module imports that fail in production builds because Vite's ?url suffix doesn't bundle dependencies. 
Switch to 'wwobjloader2/bundle/worker/module' which is self-contained with all dependencies included.

Fixes OBJ loading failing in production with Worker error events.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7879-fix-use-pre-bundled-wwobjloader2-worker-for-production-builds-2e16d73d365081f4a485c993852be1d3) by [Unito](https://www.unito.io)


DrJKL (approved):

christian-byrne:
Seems like it doesn't actually need to be backported -- fixing something that is only on 1.37 anyway.

PR #7899, opened 2026-01-08
Feat(cloud)/new top up dialog

simula-r:
## Summary

- Implement the new add credits (top up) dialog. 
- Refactor the subscription dialog to make different credit types easier to understand

## Changes

- **What**: TopUpCreditsDialogContent.vue, SubscriptionPanel.vue, /en/main.json
- **Breaking**: 
- **Dependencies**: 

## Review Focus

https://github.com/user-attachments/assets/a6454651-e195-4430-bfcc-0f2a8c1dc80b

Relevant notion links:
https://www.notion.so/comfy-org/Implement-New-Top-Up-Dialog-with-Custom-Amount-Input-2df6d73d36508142b901fc0edb0d1fc1?source=copy_link
https://www.notion.so/comfy-org/Implement-Update-confusing-credits-remaining-this-month-message-2df6d73d36508168b7e5ed46754cec60?source=copy_link


src/components/dialog/content/TopUpCreditsDialogContent.vue
christian-byrne, line 1: Would it worthwhile to move the input and its logic to a separate component? Also would any of the number inputs from the component libraries be worth using here?

src/platform/cloud/subscription/components/SubscriptionPanel.test.ts
```diff
@@ -33,6 +33,9 @@ const mockSubscriptionData = {
     const baseName = TIER_TO_NAME[mockSubscriptionTier.value]
     return mockIsYearlySubscription.value ? `${baseName} Yearly` : baseName
   }),
+  subscriptionStatus: computed(() => ({
+    renewal_date: '2024-12-31T00:00:00Z'
```
christian-byrne, line 37: question: is it ever worth making this value dynamic with real date/time?

simula-r: hmmm in this case we are just testing conditional logic to show "expiry" vs "renews" so a real date (new Date()) would just make assertions harder to write and introduce potential timezone ci flakiness without testing anything additional.

src/components/dialog/content/TopUpCreditsDialogContent.vue
simula-r: Good point, refactored into its own component for reuse. In terms of other number inputs id say this is sufficiently bespoke to wrap native.

christian-byrne (approved):
LGTM!

PR #7907, opened 2026-01-08
Fix "Markdown" and "Text Preview" widgets mouse events

kaalibro:
## Summary

Fixes canvas interaction issues in "Markdown" and "Text Preview" widgets by properly delegating wheel and pointer events to the canvas interaction system, preventing interaction blocking.

## Changes

- **What**: Added event handlers to delegate wheel and pointer events from widgets to the canvas interaction system
  - `useMarkdownWidget.ts`: Implemented comprehensive wheel handling with trackpad detection, pinch zoom prevention, and proper vertical scroll delegation
  - `TextPreviewWidget.vue`: Added wheel and pointerdown handlers using `useCanvasInteractions` composable

### Before:

https://github.com/user-attachments/assets/ca874217-8c7c-4229-922a-2b6f10a8ac9e

### After:

https://github.com/user-attachments/assets/4fd07590-2fb7-4657-a572-2cc37921d64f

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7907-Fix-Markdown-and-Text-Preview-widgets-mouse-events-2e26d73d3650812bb512ea2e2b6f8237) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/composables/useMarkdownWidget.ts
+++ b/src/renderer/extensions/vueNodes/widgets/composables/useMarkdownWidget.ts
@@ -8,9 +8,12 @@ import TiptapStarterKit from '@tiptap/starter-kit'
 import { Markdown as TiptapMarkdown } from 'tiptap-markdown'
 
 import type { LGraphNode } from '@/lib/litegraph/src/litegraph'
+import { useSettingStore } from '@/platform/settings/settingStore'
 import type { InputSpec } from '@/schemas/nodeDef/nodeDefSchemaV2'
 import { app } from '@/scripts/app'
 import type { ComfyWidgetConstructorV2 } from '@/scripts/widgets'
+
+const TRACKPAD_DETECTION_THRESHOLD = 50
 
 function addMarkdownWidget(
   node: LGraphNode,
@@ -40,6 +43,7 @@ function addMarkdownWidget(
   inputEl.classList.add('comfy-markdown')
   const textarea = document.createElement('textarea')
   inputEl.append(textarea)
+  const editorDom: HTMLElement = editor.view.dom
 
   const widget = node.addDOMWidget(name, 'MARKDOWN', inputEl, {
     getValue(): string {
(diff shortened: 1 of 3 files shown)


kaalibro:
I'm as far from development as a butcher is from surgery :) So I couldn't think of anything better than just copying the code for `useMarkdownWidget.ts` from `useStringWidget.ts` as is. It might not be the most elegant solution, but at least it works.

src/renderer/extensions/vueNodes/widgets/composables/useMarkdownWidget.ts
```diff
@@ -91,6 +95,59 @@ function addMarkdownWidget(
     }
   })
 
+  inputEl.addEventListener('wheel', (event: WheelEvent) => {
+    const gesturesEnabled = useSettingStore().get(
+      'LiteGraph.Pointer.TrackpadGestures'
+    )
```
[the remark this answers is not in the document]
Myestery: I will go with code rabbit here, lets cache this value once cos its a bit expensive to call

Myestery:
Let's cache the settings call and this will be good to go.
The new behaviour looks great

kaalibro:
@Myestery Since the 'wheel' event handler code is identical for `useMarkdownWidget.ts` and [`useStringWidget.ts`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/43c162a8627c63e19364c3143975e35ff1d29ffb/src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts#L57-L59), should I also implement caching in `useStringWidget.ts` the same way in this PR?

DrJKL:
> @Myestery Since the 'wheel' event handler code is identical for `useMarkdownWidget.ts` and [`useStringWidget.ts`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/43c162a8627c63e19364c3143975e35ff1d29ffb/src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts#L57-L59), should I also implement caching in `useStringWidget.ts` the same way in this PR?

Or we can open an issue to follow-up.

kaalibro:
> Or we can open an issue to follow-up.

I've got the commit ready, so I'll push it here to keep things consistent and save us the extra hassle.

src/renderer/extensions/vueNodes/widgets/composables/useMarkdownWidget.ts
```diff
@@ -91,6 +98,60 @@ function addMarkdownWidget(
...
+      event.stopPropagation()
+      app.canvas.processMouseWheel(event)
+      return
+    }
+
+    // Detect if this is likely a trackpad gesture vs mouse wheel
+    // Trackpads usually have deltaX or smaller deltaY values (< TRACKPAD_DETECTION_THRESHOLD)
+    // Mouse wheels typically have larger discrete deltaY values (>= TRACKPAD_DETECTION_THRESHOLD)
+    const isLikelyTrackpad =
+      Math.abs(deltaX) > 0 || Math.abs(deltaY) < TRACKPAD_DETECTION_THRESHOLD
```
jtydhr88: I don’t want to introduce any logic of detection for trackpad/mouse.
Last year, I spent months to deal with this problem, but we don’t have any good solutions for this problem, thershold approach would not work properly in the real world
So please remove all code related to detection
For this issue, we could only focus on mouse wheel

jtydhr88 (changes requested):

src/renderer/extensions/vueNodes/widgets/composables/useMarkdownWidget.ts
kaalibro: Done for both `useMarkdownWidget.ts` and `useStringWidget.ts`

jtydhr88 (approved):

PR #7912, opened 2026-01-08
Fix run badge anchoring

christian-byrne:
## Summary
Restore the shared button's positioning context so the run-queue badge anchors to the correct spot.

## Changes
- **What**: add `position: relative` back to `button.variants.ts` so badge overlays stay attached to their buttons

## Review Focus
- Make sure no buttons rely on being `position: static` (should be unaffected) and that the run badge now sits beside the Run button instead of the window edge.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7912-Fix-run-badge-anchoring-2e26d73d365081aa8fefe5381f37cfa4) by [Unito](https://www.unito.io)


christian-byrne:
>   12 pixels (ratio 0.01 of all image pixels) are different.

I cannot even see the error in the diff

DrJKL (dismissed):

PR #7934, opened 2026-01-09
fix: respect node resizable property in vueNodes mode

jtydhr88:
## Summary
Custom nodes like ComfyUI-KJNodes set `this.resizable = false` to disable resizing. This worked in litegraph but was ignored in vueNodes mode.

Extract the resizable property from LGraphNode to VueNodeData and use it to conditionally render the resize handle and block resize interactions.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7934-fix-respect-node-resizable-property-in-vueNodes-mode-2e36d73d365081a0a92ade8b23ee3ce8) by [Unito](https://www.unito.io)


Myestery (approved):
Looks Good

PR #7963, opened 2026-01-12
fix protovis coordinate calculation for DOM widgets

jtydhr88:
## Summary

Fixes drag interaction in protovis-based widgets (like KJNodes Spline Editor and Points Editor) when running in vueNodes mode.

Problem: Protovis calculates mouse position using pageX/pageY minus the offsetParent chain, but CSS transform on TransformPane changes visual position without affecting offsetLeft/offsetTop.

Solution:
- Add useProtovisCoordinate composable that patches pv.Mark.prototype.mouse
- Use getBoundingClientRect() which accounts for CSS transforms
- Only apply patch when widget contains protovis content
- Stop pointer events (not mouse events) to prevent node drag while
  allowing protovis mouse events to propagate

Need KJNodes PR https://github.com/kijai/ComfyUI-KJNodes/pull/497 merged

Before

https://github.com/user-attachments/assets/8494ef2f-5185-4d47-8fb6-7ffe5d9a93d1

After

https://github.com/user-attachments/assets/daff7445-a4b2-4e6a-8470-b1fc28bb8b26

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7963-fix-protovis-coordinate-calculation-for-DOM-widgets-2e66d73d3650810db7ebe91c279def5b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetDOM.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetDOM.vue
@@ -2,6 +2,7 @@
 import { onMounted, ref } from 'vue'
 
 import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
+import { useProtovisCoordinate } from '@/renderer/extensions/vueNodes/widgets/composables/useProtovisCoordinate'
 import { isDOMWidget } from '@/scripts/domWidget'
 import type { SimplifiedWidget } from '@/types/simplifiedWidget'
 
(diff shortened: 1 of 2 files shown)


christian-byrne (commented):
This is a case where it might be better to just expose a generalized API for the pre-transformed coordinates, instead of fixing specifically for protovis, WDYT?

AustinMroz:
I looked into this problem a while ago and gave up. This PR corroborates with my understanding: rather than using the local event coordinates, Protovis is manually walking the DOM tree to try and calculate transforms itself.

While I'm sure whatever reason Protovis has for doing this introspection is well intentioned, it is very much an edge case. Pointer events already provide multiple properties like `event.clientX` and `event.offsetX` for DOM elements to use depending on their need for pre or post-transform coordinates.

DrJKL:
This sounds like the same issue we had with the PrimeVue slider.

jtydhr88:
use https://github.com/Comfy-Org/ComfyUI_frontend/pull/8162/files instead, fix the issue in kjnodes instead https://github.com/kijai/ComfyUI-KJNodes/pull/514

PR #8009, opened 2026-01-13
feat: add overflow tooltip to NavItem

viva-jinyi:
## Summary
- Add text truncation with ellipsis to NavItem when text overflows
- Show tooltip on hover only when text is truncated
- Use on-demand overflow check (mouseenter) instead of ResizeObserver for better performance

## Test plan
- [x] Verify NavItem text truncates with ellipsis when it overflows
- [x] Verify tooltip appears on hover only when text is truncated
- [x] Verify tooltip does not appear when text fits without truncation

스크린샷 2026-01-13 오후 2 20 20

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8009-feat-add-overflow-tooltip-to-NavItem-2e76d73d3650815a9d7cefb3f0bf2daa) by [Unito](https://www.unito.io)


src/components/widget/nav/NavItem.vue
```diff
@@ -29,4 +37,15 @@ const { icon, active, onClick } = defineProps<{
   active?: boolean
   onClick: () => void
 }>()
+
+const textRef = ref<HTMLElement | null>(null)
+const isOverflowing = ref(false)
+
+const checkOverflow = () => {
```
DrJKL, line 44: Could thjs be a computed?

viva-jinyi: I’d like it to compute only once when the user hovers. If we use a computed, it runs even when there’s no overflow, so we end up doing unnecessary work.

christian-byrne (approved):

DrJKL (approved):

PR #8028, opened 2026-01-13
Faster search

AustinMroz:
Removes primevue from the searchbox, reduces animation duration, and utilizes virtualList. 

Currently causes some variation in design, but changes can be easily made since it moves away from primevue.

| Before | After |
| ------ | ----- |
| before | after|
| before | after|

Performance is difficult to depict visually. I encourage checking out and trying the branch.

Known issues:
- Refining search does not reset scroll state, allowing for scroll past the end of the list

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8028-Faster-search-2e76d73d3650815a8962d39f2ffde4b3) by [Unito](https://www.unito.io)


DrJKL (approved):

christian-byrne:
Closing as we have a totally new search box now.

PR #8036, opened 2026-01-14
fix: version mismatch warning appearing in Playwright tests despite DisableWarnings setting

christian-byrne:
PR #7004 added a setting to disable version warnings in e2e tests, but it wasn't working on release branches. The issue was a race condition (hypothesis): the version check ran before settings finished loading from the backend, so the DisableWarnings setting read its default value (false) instead of the configured value (true).

Fixed by making the warningsDisabled check reactive so it updates when settings load and adding `nextTick` (settings are loaded, but ref updates flush in a microtask. The immediate `whenever` runs before that flush, so computeds may see stale/default values -- `nextTick` waits for reactive microtasks to flush, so computeds will be correct. It's fine).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8036-fix-version-mismatch-warning-appearing-in-Playwright-tests-despite-DisableWarnings-setti-2e86d73d36508132b4d1fd73ade76e63) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/updates/common/useFrontendVersionMismatchWarning.ts
+++ b/src/platform/updates/common/useFrontendVersionMismatchWarning.ts
@@ -1,5 +1,5 @@
 import { whenever } from '@vueuse/core'
-import { computed, onMounted } from 'vue'
+import { computed, nextTick, onMounted } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 import { useToastStore } from './toastStore'
@@ -65,9 +65,12 @@ export function useFrontendVersionMismatchWarning(
     versionCompatibilityStore.dismissWarning()
   }
 
-  onMounted(() => {
+  onMounted(async () => {
     // Only set up the watcher if immediate is true
     if (immediate) {
+      // Wait for next tick to ensure reactive updates from settings load have propagated
+      await nextTick()
+
       whenever(
         () => versionCompatibilityStore.shouldShowWarning,
         () => {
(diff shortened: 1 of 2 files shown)


src/platform/settings/settingStore.ts
christian-byrne, line 1: The changes here do not make sense.

We can be confident in the reactivity change and at worst add nextTick. The lifecycle and ordering of events is pretty easy to discern here -- so a good solution should be easy to find.

christian-byrne:
Bootstrap sequence:

1. main.ts: app created, pinia + router installed, app.mount('#vue-app')
2. Router loads GraphView.vue
3. GraphView setup() runs (sync):
   - useFrontendVersionMismatchWarning({ immediate: true }) registers an onMounted callback
   - versionCompatibilityStore.initialize() scheduled via nextTick
4. GraphView template renders, including <GraphCanvas>
5. GraphCanvas setup() runs
6. GraphCanvas onMounted() (async):
   - await settingStore.loadSettingValues() loads settings
   - CORE_SETTINGS registered
7. GraphView onMounted():
   - useFrontendVersionMismatchWarning runs
   - whenever(immediate: true) evaluates shouldShowWarning
   - warningsDisabled reads from settingStore.get()
   - settings exist, but reactivity may not be flushed yet

The problem: Settings are loaded, but ref updates flush in a microtask. The immediate whenever runs before that flush, so computeds may see stale/default values.

`nextTick` waits for reactive microtasks to flush, so computeds will be correct. It is acceptable here.

DrJKL (approved):

PR #8043, opened 2026-01-14
feat: implement fuzzy search for widgets and nodes using Fuse in Right Side Panel

LittleSound:
related https://github.com/Comfy-Org/ComfyUI_frontend/pull/7812#discussion_r2685117810

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8043-feat-implement-fuzzy-search-for-widgets-and-nodes-using-Fuse-in-Right-Side-Panel-2e86d73d365081d7869cfa1956dfc0ad) by [Unito](https://www.unito.io)


src/components/rightSidePanel/shared.ts
```diff
@@ -31,27 +40,50 @@ export function searchWidgets<T extends { widget: IBaseWidget }[]>(
...
-    return words.every(
-      (word) =>
-        name.includes(word) ||
-        label?.includes(word) ||
-        type?.includes(word) ||
-        value?.includes(word)
-    )
-  }) as T
+
+  const searchableToOriginal = new Map<WidgetSearchItem<T[number]>, T[number]>()
```
christian-byrne: What does `searchableToOriginal` mean exactly? (The variable naming.)

src/components/rightSidePanel/shared.ts
christian-byrne, line 1: Should we extract the search logic to a service or composable? To make it re-usable, testable, deduplicate, and potentially implement lifecycle hooks if they become necessary?

src/components/rightSidePanel/shared.ts
LittleSound: I have simplified this part of the code by removing this variable.

src/components/rightSidePanel/shared.ts
LittleSound: This search logic was reused in the entire right side panel. They use the same data structure and search widgets. I don't know the situation in other places.

christian-byrne (approved):

PR #8052, opened 2026-01-14
refactor: rename CLAUDE.md to AGENTS.md

DrJKL:
## Summary

Pure rename of CLAUDE.md files to AGENTS.md (no content changes).

## Changes

| Old Path | New Path |
|----------|----------|
| `.github/CLAUDE.md` | `.github/AGENTS.md` |
| `.storybook/CLAUDE.md` | `.storybook/AGENTS.md` |
| `browser_tests/CLAUDE.md` | `browser_tests/AGENTS.md` |
| `src/CLAUDE.md` | `src/AGENTS.md` |
| `src/components/CLAUDE.md` | `src/components/AGENTS.md` |
| `src/lib/litegraph/CLAUDE.md` | `src/lib/litegraph/AGENTS.md` |

Root `CLAUDE.md` deleted (content will be merged into `AGENTS.md` in follow-up PR).

## Follow-up

A second PR will add glob-based guidance files and consolidate redundancies.

PR #8053, opened 2026-01-14
refactor: consolidate guidance and add glob-based docs

DrJKL:
## Summary

Consolidates AGENTS.md files and adds glob-based guidance per [Amp Manual patterns](https://ampcode.com/manual#granular-guidance).

**Depends on:** #8052 (rename CLAUDE.md to AGENTS.md)

### New Glob-Based Guidance (`docs/guidance/`)

| File | Globs | Purpose |
|------|-------|---------|
| `vue-components.md` | `**/*.vue` | PrimeVue migrations, VueUse patterns |
| `typescript.md` | `**/*.ts`, `**/*.tsx` | Type safety, API utilities |
| `vitest.md` | `**/*.test.ts` | Unit test conventions |
| `playwright.md` | `**/*.spec.ts` | E2E test conventions |
| `storybook.md` | `**/*.stories.ts` | Story structure |

### Root AGENTS.md Updates

- Added `@docs/guidance/*.md` reference
- Added Development Workflow section (from CLAUDE.md)
- Added Git Conventions section (from CLAUDE.md)
- Added `pnpm storybook` command

### Redundancies Eliminated

- PrimeVue migrations (was in 3 files → vue-components.md)
- API utilities (was in 2 files → src/AGENTS.md only)
- Vue 3 patterns (was in 3 files → root AGENTS.md only)
- vue-i18n guidance (was in 3 files → root AGENTS.md only)
- Build commands (was in 3 files → root AGENTS.md only)

## Results

| Metric | Before | After |
|--------|--------|-------|
| Redundant topics | 10 | 0 |
| Glob-based files | 0 | 5 |
| Lines changed | - | -53 net |

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8053-refactor-consolidate-guidance-and-add-glob-based-docs-2e86d73d3650813f9547e2434eb7815e) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/AGENTS.md
+++ b/.github/AGENTS.md
@@ -1,32 +1,10 @@
-# ComfyUI Frontend - Claude Review Context
+# PR Review Context
 
-This file provides additional context for the automated PR review system.
-
-## Quick Reference
-
-### PrimeVue Component Migrations
-
-When reviewing, flag these deprecated components:
-- `Dropdown` → Use `Select` from 'primevue/select'
-- `OverlayPanel` → Use `Popover` from 'primevue/popover'
-- `Calendar` → Use `DatePicker` from 'primevue/datepicker'
-- `InputSwitch` → Use `ToggleSwitch` from 'primevue/toggleswitch'
-- `Sidebar` → Use `Drawer` from 'primevue/drawer'
-- `Chips` → Use `AutoComplete` with multiple enabled and typeahead disabled
-- `TabMenu` → Use `Tabs` without panels
-- `Steps` → Use `Stepper` without panels
-- `InlineMessage` → Use `Message` component
-
-### API Utilities Reference
-
-- `api.apiURL()` - Backend API calls (/prompt, /queue, /view, etc.)
-- `api.fileURL()` - Static file access (templates, extensions)
-- `$t()` / `i18n.global.t()` - Internationalization
-- `DOMPurify.sanitize()` - HTML sanitization
+Context for automated PR review system.
 
 ## Review Scope
 
-This automated review performs comprehensive analysis including:
+This automated review performs comprehensive analysis:
 - Architecture and design patterns
 - Security vulnerabilities
 - Performance implications
--- a/docs/guidance/storybook.md
+++ b/docs/guidance/storybook.md
@@ -1,0 +1,55 @@
+---
+globs:
+  - '**/*.stories.ts'
+---
+
+# Storybook Conventions
+
+## File Placement
+
+Place `*.stories.ts` files alongside their components:
+```
+src/components/MyComponent/
+├── MyComponent.vue
+└── MyComponent.stories.ts
+```
+
+## Story Structure
+
+```typescript
+import type { Meta, StoryObj } from '@storybook/vue3'
+import ComponentName from './ComponentName.vue'
+
+const meta: Meta<typeof ComponentName> = {
+  title: 'Category/ComponentName',
+  component: ComponentName,
+  parameters: { layout: 'centered' }
+}
+
+export default meta
+type Story = StoryObj<typeof meta>
+
+export const Default: Story = {
+  args: { /* props */ }
+}
+```
+
+## Required Story Variants
+
+Include when applicable:
+- **Default** - Minimal props
+- **WithData** - Realistic data
+- **Loading** - Loading state
+- **Error** - Error state
+- **Empty** - No data
+
+## Mock Data
+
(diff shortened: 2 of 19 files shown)


.github/AGENTS.md
```diff
@@ -1,36 +1,14 @@
...
-- `Steps` → Use `Stepper` without panels
-- `InlineMessage` → Use `Message` component
-
-### API Utilities Reference
-
-- `api.apiURL()` - Backend API calls (/prompt, /queue, /view, etc.)
-- `api.fileURL()` - Static file access (templates, extensions)
-- `$t()` / `i18n.global.t()` - Internationalization
-- `DOMPurify.sanitize()` - HTML sanitization
+Context for automated PR review system.
```
christian-byrne, line 3: This file is also used when doing development inside the repo, should we adjust the framing away from "This is a doc explaining how to do PR review" to something more generalizable like "this is a file for best practices" (or whatever the standard way of framing these files is)?

docs/guidance/storybook.md
christian-byrne, line 1: You don't like the other stuff that we had before like the guide on mock data? Just worry we would miss something that otherwise takes the agent some trial and error

DrJKL: @viva-jinyi You've written a few stories. What advice do you think we should keep in here?

viva-jinyi: For shared / common components, I think what matters more than mock data is how the component renders based on the props it receives or inherits.

Mock data tends to break whenever data types change, which means it often requires constant updates. From a maintenance perspective, that makes heavy mock-data–driven stories less useful over time.

Instead, I think Storybook is most valuable when it focuses on:

- how the component looks when certain props are present or missing
- how new props or options affect rendering
- visual variants based on different prop combinations

In other words, Storybook works best as a place to clearly show
“given these props, this is how the component behaves and looks.”

That kind of prop-driven variant documentation feels much more aligned with the purpose of shared components and much more useful in the long run.

.github/AGENTS.md
DrJKL: I'm unsure when this would actually be triggered or what humans would look for it.

viva-jinyi (approved):
LGTM

docs/guidance/storybook.md
DrJKL: I also think that it might be good to curate good examples to set as references.

.github/AGENTS.md
christian-byrne: Claude will read this when making changes to .github/* files. 90% confident. Can test when back at computer. This isn't blocking though as it already existed

christian-byrne: Example:

Image

DrJKL: That piece of the workflow shouldn't need to know about PrimeVue Components or API utilities, right?

Added compatibility shims for Claude.

docs/guidance/vue-components.md
```diff
@@ -0,0 +1,64 @@
...
+|------------|-------------|-------------|
+| `Dropdown` | `Select` | `primevue/select` |
+| `OverlayPanel` | `Popover` | `primevue/popover` |
+| `Calendar` | `DatePicker` | `primevue/datepicker` |
+| `InputSwitch` | `ToggleSwitch` | `primevue/toggleswitch` |
+| `Sidebar` | `Drawer` | `primevue/drawer` |
+| `Chips` | `AutoComplete` | `primevue/autocomplete` (with `multiple` enabled, `typeahead` disabled) |
+| `TabMenu` | `Tabs` | `primevue/tabs` (without panels) |
+| `Steps` | `Stepper` | `primevue/stepper` (without panels) |
+| `InlineMessage` | `Message` | `primevue/message` |
```
christian-byrne: Since we have a linter rule for this, seems like an easy win for cutting down on agent context right?

christian-byrne (commented):
It's a bit hard to review when files are moved _and_ changed in same PR. I can't be 100% sure I know what was just moved vs. what content was actually changed as well.

DrJKL:
> It's a bit hard to review when files are moved _and_ changed in same PR. I can't be 100% sure I know what was just moved vs. what content was actually changed as well.

That's why the move happened in the parent PR. The other way it looks like the content from the old files was just deleted and the new files created without history.

docs/guidance/vue-components.md
DrJKL: Removed. There are probably other pieces we can consolidate. Especially if we can get linting fast enough.

christian-byrne (approved):

PR #8054, opened 2026-01-14
Linear mode bug fixes

AustinMroz:
┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8054-Linear-mode-bug-fixes-2e86d73d365081ed8d75d6e2af679f6c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useCoreCommands.ts
+++ b/src/composables/useCoreCommands.ts
@@ -1234,7 +1234,7 @@ export function useCoreCommands(): ComfyCommand[] {
     {
       id: 'Comfy.ToggleLinear',
       icon: 'pi pi-database',
-      label: 'toggle linear mode',
+      label: 'Toggle Simple Mode',
       function: () => {
         const newMode = !canvasStore.linearMode
         app.rootGraph.extra.linearMode = newMode
--- a/src/renderer/extensions/linearMode/LinearControls.vue
+++ b/src/renderer/extensions/linearMode/LinearControls.vue
@@ -72,13 +72,18 @@ function nodeToNodeData(node: LGraphNode) {
   }
 }
 const partitionedNodes = computed(() => {
-  return partition(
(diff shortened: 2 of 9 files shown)


src/composables/useCoreCommands.ts
```diff
@@ -1234,7 +1234,7 @@ export function useCoreCommands(): ComfyCommand[] {
     {
       id: 'Comfy.ToggleLinear',
       icon: 'pi pi-database',
-      label: 'toggle linear mode',
+      label: 'Toggle Simple Mode',
```
DrJKL, line 1237: Nit: Does this go through i18n?

src/renderer/extensions/linearMode/LinearControls.vue
```diff
@@ -55,27 +56,35 @@ function nodeToNodeData(node: LGraphNode) {
...
-          label: t('linearMode.dragAndDropImage')
+          label: t('linearMode.dragAndDropImage'),
+          onClick: () => node.widgets?.[1]?.callback?.(undefined)
         }
   const nodeData = extractVueNodeData(node)
   for (const widget of nodeData.widgets ?? []) widget.slotMetadata = undefined
 
   return {
     ...nodeData,
+    //note lastNodeErrors uses exeuctionid, node.id is execution for root
```
DrJKL, line 67:
```suggestion
    // Note: lastNodeErrors uses ExecutionId, node.id is execution for root
```

DrJKL (approved):

src/composables/useCoreCommands.ts
AustinMroz: As best I can tell the label is only used in automation scripts to generate entries in locales/en/commands.json (which is keyed by id) and are never used afterwards.

PR #8056, opened 2026-01-14
fix: use staging platform URL for usage history link

christian-byrne:
## Summary

Uses `getComfyPlatformBaseUrl()` for the usage history link so it correctly points to `stagingplatform.comfy.org` on staging builds instead of being hardcoded to `platform.comfy.org`.

## Changes

- Added `usageHistoryUrl` computed that uses `getComfyPlatformBaseUrl()`
- Updated template to use dynamic `:href` binding

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8056-fix-use-staging-platform-URL-for-usage-history-link-2e86d73d36508186a845ffd84e5caaf2) by [Unito](https://www.unito.io)


DrJKL (dismissed):
Do we want to add the locale to this one?

DrJKL (approved):

christian-byrne:
> Do we want to add the locale to this one?

That's only for docs.comfy.org I believe

PR #8070, opened 2026-01-15
fix: prevent Record Audio waveform from overflowing node bounds

jtydhr88:
## Summary

Add min-w-0 to flex containers to allow shrinking, reduce gaps and padding, and use shrink-0 on fixed-size elements to ensure the waveform area clips properly when the node is at minimum width.

before
https://github.com/user-attachments/assets/215455aa-555b-4ade-9b36-9e89ac7c14aa

after
https://github.com/user-attachments/assets/bf56028b-ae02-4388-be83-460c7b1f14e1

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8070-fix-prevent-Record-Audio-waveform-from-overflowing-node-bounds-2e96d73d3650816b9660e5532ffa0bc3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetRecordAudio.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetRecordAudio.vue
@@ -12,11 +12,11 @@
     </div>
     <div
       v-if="isRecording || isPlaying || recordedURL"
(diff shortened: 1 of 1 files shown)


Myestery:
It will be nice to have the waveform fill the whole width when the node is resized

DrJKL (approved):

PR #8083, opened 2026-01-15
Road to No Explicit Any Part 6: Composables and Extensions

Myestery:
## Summary
- Type `onExecuted` callbacks with `NodeExecutionOutput` in saveMesh.ts and uploadAudio.ts
- Type composable parameters and return values properly (useLoad3dViewer, useImageMenuOptions, useJobMenu, useResultGallery, useContextMenuTranslation)
- Type `taskRef` as `TaskItemImpl` with updated test mocks
- Fix error catch and index signature patterns without `any`
- Add `NodeOutputWith<T>` generic helper for typed access to passthrough properties on `NodeExecutionOutput`

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] Unit tests pass for affected files
- [x] Sourcegraph checks confirm no external usage of modified types

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8083-Road-to-No-Explicit-Any-Part-6-Composables-and-Extensions-2e96d73d3650810fb033d745bf88a22b) by [Unito](https://www.unito.io)

PR #7171, opened 2025-12-04
chore: Remove legacy history code (PR 3 of 3)

ric-yu:
## Summary

Remove legacy history code that is no longer needed after migrating to the Jobs API.

**This is PR 3 of 3** - Cleanup only, depends on PR 2.

## Changes

- **What**:
  - Delete legacy `/history` and `/history_v2` fetchers and adapters
  - Delete legacy history types (`HistoryTaskItem`, `RunningTaskItem`, `PendingTaskItem`, etc.)
  - Delete `getWorkflowFromHistory` wrapper (replaced by direct `fetchJobDetail` + `extractWorkflow`)
  - Delete browser test fixtures for legacy task history
  - Remove unused Zod schemas from `apiSchema.ts`

- **Non-breaking**: Only removes dead code after migration

### Legacy History Module
- `src/platform/remote/comfyui/history/` (entire directory)
  - `adapters/v2ToV1Adapter.ts`
  - `fetchers/fetchHistoryV1.ts`
  - `fetchers/fetchHistoryV2.ts`
  - `types/historyV1Types.ts`
  - `types/historyV2Types.ts`
  - `index.ts`

### Legacy Workflow Helper
- `src/platform/workflow/cloud/getWorkflowFromHistory.ts`
- `src/platform/workflow/cloud/index.ts`

### Test Fixtures
- `tests-ui/fixtures/historyFixtures.ts`
- `tests-ui/fixtures/historySortingFixtures.ts`
- `browser_tests/fixtures/utils/taskHistory.ts`

### Legacy Tests
- `tests-ui/tests/platform/remote/comfyui/history/` (all tests)

## Files Modified

- `src/schemas/apiSchema.ts` - Removed unused legacy types:
  - `zQueueIndex`, `zExtraData`
  - `RunningTaskItem`, `PendingTaskItem`, `HistoryTaskItem`, `TaskItem`

## Dependencies

- **Depends on**: PR 2 (Jobs API Migration) - #7170

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7171-chore-Remove-legacy-history-code-PR-3-of-3-2bf6d73d365081bb8fa4e57498044a14) by [Unito](https://www.unito.io)

PR #8092, opened 2026-01-16
Road to No Explicit Any Part 7: Scripts and Dialog Cleanup

Myestery:
## Summary

Continues the TypeScript strict typing improvements by removing `any` types from core scripts and dialog components.

### Changes

**api.ts (6 instances)**
- Define `V1RawPrompt` and `CloudRawPrompt` tuple types for queue prompt formats
- Export `QueueIndex`, `PromptInputs`, `ExtraData`, `OutputsToExecute` from apiSchema
- Type `#postItem` body, `storeUserData` data, and `getCustomNodesI18n` return

**groupNodeManage.ts (all @ts-expect-error removed)**
- Add `GroupNodeConfigEntry` interface to LGraph.ts
- Extend `GroupNodeWorkflowData` with `title`, `widgets_values`, and typed `config`
- Type all class properties with definite assignment assertions
- Type all method parameters and event handlers
- Fix save button callback with proper generic types for node ordering

**changeTracker.ts (4 instances)**
- Type `nodeOutputs` as `Record<string, ExecutedWsMessage['output']>`
- Type prompt callback with `CanvasPointerEvent` and proper value types

**asyncDialog.ts and dialog.ts**
- Make `ComfyAsyncDialog` generic with `DialogAction<T>` type
- Type `ComfyDialog` constructor and show method parameters
- Update `ManageGroupDialog.show` signature to match base class

## Test plan
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] Sourcegraph checks for external usage

---
Related: Continues from #8083

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8092-Road-to-No-Explicit-Any-Part-7-Scripts-and-Dialog-Cleanup-2ea6d73d365081fbb890e73646a6ad16) by [Unito](https://www.unito.io)


src/schemas/apiSchema.ts
christian-byrne, line 1: This might conflict with other in-flight PR https://github.com/Comfy-Org/ComfyUI_frontend/pull/7171

christian-byrne (commented):
Can we pause on this one until https://github.com/Comfy-Org/ComfyUI_frontend/pull/7171 is in? Major refactor to related code I believe

src/scripts/api.ts
```diff
@@ -1251,7 +1251,7 @@ export class ComfyApi extends EventTarget {
    *
    * @returns The custom nodes i18n data
    */
-  async getCustomNodesI18n(): Promise<Record<string, any>> {
+  async getCustomNodesI18n(): Promise<Record<string, unknown>> {
```
christian-byrne: We should have a schema for this, or define one otherwise

src/scripts/api.ts
```diff
@@ -1074,7 +1074,7 @@ export class ComfyApi extends EventTarget {
    */
   async storeUserData(
     file: string,
-    data: any,
+    data: BodyInit | Record<string, unknown> | null,
```
christian-byrne: May as well use `unknown` here?

src/extensions/core/groupNodeManage.ts
```diff
@@ -15,15 +17,17 @@ import './groupNodeManage.css'
 
 const ORDER: symbol = Symbol()
 
-// @ts-expect-error fixme ts strict error
-function merge(target, source) {
+function merge(
```
AustinMroz, line 20:
```suggestion
function merge(
  target: Record<string, unknown>,
  source: Record<string, unknown>
): Record<string, unknown> {
  for (const key in source) {
    const sv = source[key]
    if (typeof sv === 'object') {
      let tv = target[key] as Record<string, unknown> | undefined
      merge(tv ?? {}, sv as Record<string, unknown>)
    } else {
      target[key] = sv
    }
  }

  return target
}
```
We can further reduce the number of casts here.

AustinMroz (dismissed):
This one was pretty straightforward :+1:

src/extensions/core/groupNodeManage.ts
Myestery: There's a little issue with the suggestion line 28 around merging with an unassigned tv but I have simplified it as requested.

AustinMroz (approved):

src/extensions/core/groupNode.ts
```diff
@@ -386,7 +386,8 @@ export class GroupNodeConfig {
       let type: string | number | null = linksFrom[0]?.[0]?.[5] ?? null
       if (type === 'COMBO') {
         // Use the array items
-        const source = node.outputs?.[0]?.widget?.name
+        const output = node.outputs?.[0] as GroupNodeOutput | undefined
```
DrJKL, line 389: Using type assertions isn't much safer than `any`

PR #8094, opened 2026-01-16
Fix copypasted primitives inside subgraphs

AustinMroz:
Copying a subgraph which contains primitive nodes would cause the primitives to fail to initialize. This was caused because they did not have their `onGraphConfigured` and `onAfterGraphConfigured` callbacks applied.

There's already a copy of `forEachNode` in `@/utils/graphTraversalUtil.ts`, but the method is small and I want to avoid litegraph referencing outside code.

See also #6606, where a similar fix was needed

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8094-Fix-copypasted-primitives-inside-subgraphs-2ea6d73d365081f189f1ea4c9248f5ed) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/subgraph/subgraphUtils.ts
+++ b/src/lib/litegraph/src/subgraph/subgraphUtils.ts
@@ -499,15 +499,3 @@ export function isNodeSlot(
     ('link' in slot || 'links' in slot)
   )
 }
-export function forEachNode(
-  func: (node: LGraphNode) => void,
-  nodes: LGraphNode[],
-  visited = new Set()
-) {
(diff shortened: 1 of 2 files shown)


src/lib/litegraph/src/subgraph/subgraphUtils.ts
christian-byrne, line 1: Could we move this to graphTraversalUtils?

AustinMroz: ok

christian-byrne (approved):

PR #8112, opened 2026-01-16
Disable control widgets on link to parent

AustinMroz:
When a link is made to a widget with control (like seed) , the control widget can no longer be used to update it's state. To better communicate this, the control widgets are now given the disabled property when their parent widget is linked.

| Before | After |
| ------ | ----- |
| before | after|

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8112-Disable-control-widgets-on-link-to-parent-2ea6d73d365081afad77db6c5f56e085) by [Unito](https://www.unito.io)


christian-byrne (dismissed):

browser_tests/tests/mobileBaseline.spec.ts-snapshots/mobile-settings-dialog-mobile-chrome-linux.png
AustinMroz, line 1: Well this is tragic to see. Is this just a *slight* difference of length between `1` and `0` ?

christian-byrne (approved):

PR #8120, opened 2026-01-17
Fix crosshair cursor in vue mode

AustinMroz:
When the mouse cursor is at the very edge of a a node in vue mode, a crosshair cursor will sometimes display. This happens because the mouse is over the canvas, and `LGraphCanvas.processMouseMove` determines the cursor is still above the node.

~~This fix feels kinda ick. I don't believe that we can trust the `getNodeOnPos` call to be accurate when in vue mode, but this change could likely have side effects. Perhaps the check should be performed in `processMouseMove` instead?~~

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8120-Fix-crosshair-cursor-in-vue-mode-2eb6d73d36508116a3cfdd407c5e1e9c) by [Unito](https://www.unito.io)


AustinMroz:
PR causes a regression with requiring a draggedlink to be directly over a vue node to connect instead of just being near to one.

christian-byrne (approved):

PR #8122, opened 2026-01-17
Add a slider indicator for number widgets in vue mode.

AustinMroz:
Sometimes it's difficult to gauge the valid range of values for a widget. Litegraph includes a "slider" widget which displays the distance from the min and max values as a colored bar. However, this implementation is rather strongly disliked because it prevents entering an exact number. Vue mode makes it simple to add just the indicator onto our existing widget.

In addition to requiring both min and max be set, not every widget would want this functionality. It's not useful information for seed, but also has potential to cause confusion on widgets like CFG, that allow inputting numbers up to 100 even though values beyond ~15 are rarely desirable.

As a proposed heuristic, the ratio of "step" to distance between min and max is currently used, but this could fairly easily be changed to an opt-in only system.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8122-Add-a-slider-indicator-for-number-widgets-in-vue-mode-2eb6d73d365081218fc8e86f37001958) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
@@ -187,7 +187,7 @@ const sliderWidth = computed(() => {
     return 0
   const value = dragValue.value ?? modelValue.value
   const ratio = (value - min) / (max - min)
-  return (ratio * 100) | 0
+  return (ratio * 100).toFixed(0)
 })
 </script>
 
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
@@ -177,8 +177,14 @@ const buttonTooltip = computed(() => {
 })
 
 const sliderWidth = computed(() => {
-  const { min, max } = filteredProps.value
-  if (min === undefined || max === undefined) return 0
+  const { max, min, step } = filteredProps.value
+  if (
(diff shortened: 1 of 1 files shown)


src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
```diff
@@ -175,6 +175,20 @@ const buttonTooltip = computed(() => {
...
+  if (
+    min === undefined ||
+    max === undefined ||
+    step === undefined ||
+    (max - min) / step >= 100
+  )
+    return 0
+  const value = dragValue.value ?? modelValue.value
+  const ratio = (value - min) / (max - min)
+  return (ratio * 100) | 0
```
christian-byrne: What's the main motivations for bitwise OR here?

christian-byrne (dismissed):
LGTM

src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberInput.vue
AustinMroz: It's to truncate the value to a round percent. I was a little concerned that irrational numbers would add unwanted clutter to the dom or cause excessive rerenders, but it's super minor.

DrJKL: If you just need the string value clamped, You can probably use `toPrecision` or `toFixed`.

christian-byrne (approved):

PR #8129, opened 2026-01-17
add thumbnail for 3d generation

jtydhr88:
## Summary

add thrumbnail for 3d genations, feature requested by @PabloWiedemann 

## Screenshots

https://github.com/user-attachments/assets/4fb9b88b-dd7b-4a69-a70c-e850472d3498

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8129-add-thumbnail-for-3d-generation-2eb6d73d365081f2a30bc698a4fde6e0) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useLoad3d.ts
+++ b/src/composables/useLoad3d.ts
@@ -511,6 +511,22 @@ export const useLoad3d = (nodeOrRef: MaybeRef<LGraphNode | null>) => {
       hasSkeleton.value = load3d?.hasSkeleton() ?? false
       // Reset skeleton visibility when loading new model
       modelConfig.value.showSkeleton = false
+
+      if (load3d) {
+        const node = nodeRef.value
+
+        const modelWidget = node?.widgets?.find(
+          (w) => w.name === 'model_file' || w.name === 'image'
+        )
+        const value = modelWidget?.value
+        if (typeof value === 'string') {
+          void Load3dUtils.generateThumbnailIfNeeded(
+            load3d,
+            value,
+            isPreview.value ? 'output' : 'input'
+          )
+        }
+      }
     },
     skeletonVisibilityChange: (value: boolean) => {
       modelConfig.value.showSkeleton = value
--- a/src/extensions/core/load3d/Load3d.ts
+++ b/src/extensions/core/load3d/Load3d.ts
@@ -754,6 +754,60 @@ class Load3d {
     this.forceRender()
   }
 
+  public async captureThumbnail(
+    width: number = 256,
+    height: number = 256
+  ): Promise<string> {
+    if (!this.modelManager.currentModel) {
+      throw new Error('No model loaded for thumbnail capture')
+    }
+
+    const savedState = this.cameraManager.getCameraState()
+    const savedCameraType = this.cameraManager.getCurrentCameraType()
+    const savedGridVisible = this.sceneManager.gridHelper.visible
+
+    try {
+      this.sceneManager.gridHelper.visible = false
+
+      if (savedCameraType !== 'perspective') {
+        this.cameraManager.toggleCamera('perspective')
+      }
+
+      const box = new THREE.Box3().setFromObject(this.modelManager.currentModel)
+      const size = box.getSize(new THREE.Vector3())
+      const center = box.getCenter(new THREE.Vector3())
+
+      const maxDim = Math.max(size.x, size.y, size.z)
+      const distance = maxDim * 1.5
+
+      const cameraPosition = new THREE.Vector3(
+        center.x - distance * 0.8,
+        center.y + distance * 0.4,
+        center.z + distance * 0.3
+      )
+
+      this.cameraManager.perspectiveCamera.position.copy(cameraPosition)
+      this.cameraManager.perspectiveCamera.lookAt(center)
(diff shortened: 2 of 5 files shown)


src/composables/useLoad3d.ts
```diff
@@ -511,6 +539,22 @@ export const useLoad3d = (nodeOrRef: MaybeRef<LGraphNode | null>) => {
       hasSkeleton.value = load3d?.hasSkeleton() ?? false
       // Reset skeleton visibility when loading new model
       modelConfig.value.showSkeleton = false
+
+      if (load3d) {
+        const node = toRaw(nodeRef.value) as LGraphNode | null
+
+        const modelWidget = node?.widgets?.find(
+          (w) => w.name === 'model_file' || w.name === 'image'
+        )
```
christian-byrne: Do you need to use `toRaw`? What happens if you just access the underlying node using `.value`? There was an issue with types?

src/composables/useLoad3d.ts
```diff
@@ -511,6 +539,22 @@ export const useLoad3d = (nodeOrRef: MaybeRef<LGraphNode | null>) => {
...
+        const modelPath = modelWidget?.value as string | undefined
+        if (modelPath) {
+          void generateThumbnailIfNeeded(
+            load3d,
+            modelPath,
```
christian-byrne: nit: Guard might be better than assertion here?

```suggestion
        const value = modelWidget?.value
        if (typeof value === 'string') {
          void generateThumbnailIfNeeded(
            load3d,
            value,
```

src/extensions/core/load3d/Load3d.ts
```diff
@@ -754,6 +754,60 @@ class Load3d {
     this.forceRender()
   }
 
+  public async captureThumbnail(
+    width: number = 256,
+    height: number = 256
+  ): Promise<string> {
+    if (!this.modelManager.currentModel) {
+      throw new Error('No model loaded for thumbnail capture')
```
christian-byrne, line 762: nit: Is it better for this to fail gracefully rather than `throw`? Maybe it's already wrapped in `try` block at all call-sites?

src/composables/useLoad3d.ts
```diff
@@ -463,6 +463,34 @@ export const useLoad3d = (nodeOrRef: MaybeRef<LGraphNode | null>) => {
     }
   }
 
+  async function generateThumbnailIfNeeded(
```
christian-byrne: This function and `generateThumbnailForSaveMesh` seem to be almost the same. Can we create a single function that they both share?

src/extensions/core/load3d/Load3dUtils.ts
```diff
@@ -122,6 +122,51 @@ class Load3dUtils {
...
+    try {
+      const url = api.apiURL(this.getResourceURL(subfolder, filename, type))
+      const response = await fetch(url, { method: 'HEAD' })
+      return response.ok
+    } catch {
+      return false
+    }
+  }
+
+  static async uploadThumbnail(
```
christian-byrne, line 169: Should we re-use the helpers that already exist -- e.g., the one that is 100 lines above this https://github.com/Comfy-Org/ComfyUI_frontend/blob/d3bd85db7f9e0ec202dfb00a80b2a3d8a67f84f0/src/extensions/core/load3d/Load3dUtils.ts#L39

Or this https://github.com/Comfy-Org/ComfyUI_frontend/blob/d3bd85db7f9e0ec202dfb00a80b2a3d8a67f84f0/src/extensions/core/load3d/Load3dUtils.ts#L7

---
                                                                                                    
Not really for this PR, but after looking, we seem to have a lot of duplication across the codebase for the upload endpoint usage:                                                                               
                                                                                                                
  - src/composables/node/useNodeImageUpload.ts:20-49 (most generic implementation)                              
  - src/composables/maskeditor/useMaskEditorSaver.ts:266-299                                                    
  - src/components/common/BackgroundImageUpload.vue:55-74                                                       
  - src/extensions/core/webcamCapture.ts                                                                        
  - src/extensions/core/uploadAudio.ts:28-73                                                                    
  - src/services/audioService.ts                                                                                
  - src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue

src/platform/assets/components/Media3DTop.vue
```diff
@@ -1,12 +1,35 @@
...
+import type { AssetMeta } from '../schemas/mediaAssetSchema'
+
+const { asset } = defineProps<{ asset: AssetMeta }>()
+
+const thumbnailError = ref(false)
+
+const thumbnailSrc = computed(() => {
+  if (!asset?.src) return ''
+  return asset.src.replace(/([?&]filename=)([^&]*)/, '$1$2.png')
+})
```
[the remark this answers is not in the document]
christian-byrne: Seems valid (replying to coderabbit comment)

src/extensions/core/load3d/Load3dUtils.ts
christian-byrne: @coderabbitai make an Issue to centralize all the upload helpers listed above into a singlar shared util or service that these modules then import from. Allows easier refactoring later whenever we adjust the upload logic or parameters.

christian-byrne (approved):

PR #8143, opened 2026-01-18
fix: sync node.imgs for legacy context menu in Vue Nodes mode

christian-byrne:
## Summary

Fixes missing "Copy Image", "Open Image", "Save Image", and "Open in Mask Editor" context menu options on SaveImage nodes when Vue Nodes mode is enabled.

## Changes

- Add `syncLegacyNodeImgs` store method to sync loaded image elements to `node.imgs`
- Call sync on image load in ImagePreview component
- Simplify mask editor handling to call composable directly

## Technical Details

- Only runs when `vueNodesMode` is enabled (no impact on legacy mode)
- Reuses already-loaded `` element from Vue (no duplicate network requests)
- Store owns the sync logic, component just hands off the element

Supersedes #7416

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8143-fix-sync-node-imgs-for-legacy-context-menu-in-Vue-Nodes-mode-2ec6d73d365081c59d42cd1722779b61) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -201,6 +198,10 @@ const handleImageLoad = (event: Event) => {
   if (img.naturalWidth && img.naturalHeight) {
     actualDimensions.value = `${img.naturalWidth} x ${img.naturalHeight}`
   }
+
+  if (props.nodeId) {
+    nodeOutputStore.syncLegacyNodeImgs(props.nodeId, img, currentIndex.value)
```
DrJKL, line 212: Seeing `sync` always worries me.

src/stores/imagePreviewStore.ts
```diff
@@ -364,6 +365,32 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
+   * @param activeIndex - The current image index (for multi-image outputs)
+   */
+  function syncLegacyNodeImgs(
+    nodeId: string | number,
+    element: HTMLImageElement,
+    activeIndex: number = 0
+  ) {
+    if (!LiteGraph.vueNodesMode) return
+
+    const node = app.rootGraph?.getNodeById(Number(nodeId))
```
DrJKL, line 417: We really need to standardize our NodeId type.

src/renderer/extensions/vueNodes/components/ImagePreview.vue
christian-byrne: It's just doing what happens in litegraph

DrJKL: Why does that not comfort me?

DrJKL (dismissed):
I still think this is dirty, but we can clean it up later.

PR #8149, opened 2026-01-18
feat: add isNightly build flag for nightly-only features

christian-byrne:
## Summary

Adds a compile-time `__IS_NIGHTLY__` constant that detects whether the build is from the main branch (nightly) or a core/* branch (RC/stable). The detection logic in vite.config.mts auto-detects based on `GITHUB_REF_NAME === 'main'` in CI, with explicit override support via `IS_NIGHTLY` environment variable. Exports `isNightly` from `src/platform/distribution/types.ts` for use throughout the codebase. Includes unit tests for the detection logic.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8149-feat-add-isNightly-build-flag-for-nightly-only-features-2ec6d73d365081c09930edec1c6644f5) by [Unito](https://www.unito.io)


src/platform/distribution/types.test.ts
christian-byrne, line 1: Change detector test is not necessary!

DrJKL: This could either be 3 tests or 1 test, but instead...

DrJKL: Optionally: You could use `it.for` if you want to get fancy.

DrJKL (dismissed):

PR #8159, opened 2026-01-19
feat: optimize empty search to use cached /nodes endpoint

Myestery:
## Summary

Optimizes the Manager dialog to use the cached `GET /nodes` endpoint instead of `GET /nodes/search` for empty search queries (when the dialog first opens). This significantly reduces Algolia usage since empty searches account for the majority of search requests.

## Changes

- **registrySearchProvider.ts**: Modified `searchPacks()` to detect empty queries and route them to `listAllPacks()` instead of `search()`
- **registrySearchProvider.test.ts**: Added 5 new test cases covering empty query behavior
- Cache clearing now clears both `search` and `listAllPacks` caches

## Technical Details

**Empty Query Flow (NEW):**
- Query: `""` or whitespace
- Endpoint: `GET /nodes?limit=X&page=Y`
- Cache: Server-side cached (via omitting `latest` parameter)
- Result: Fast, cached node pack list

**Non-Empty Query Flow (UNCHANGED):**
- Query: Any non-empty string
- Endpoint: `GET /nodes/search?search=X` or `comfy_node_search=X`
- Result: Search results as before

## Testing

```bash
pnpm test:unit -- src/services/providers/registrySearchProvider.test.ts
pnpm typecheck
```

### DIFF
--- a/src/services/providers/registrySearchProvider.ts
+++ b/src/services/providers/registrySearchProvider.ts
@@ -25,14 +25,37 @@ export const useComfyRegistrySearchProvider = (): NodePackSearchProvider => {
   ): Promise<SearchPacksResult> => {
     const { pageSize, pageNumber, restrictSearchableAttributes } = params
 
-    // Determine search mode based on searchable attributes
+    // For empty queries, use the cached listAllPacks endpoint instead of search
+    if (!query || query.trim() === '') {
+      const listParams = {
+        limit: pageSize,
+        page: pageNumber + 1 // Registry API uses 1-based pagination
+        // Note: omitting 'latest' parameter defaults to cached result
+      }
+
+      const listResult = await registryStore.listAllPacks.call(listParams)
+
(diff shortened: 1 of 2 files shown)


src/services/providers/registrySearchProvider.ts
```diff
@@ -25,14 +25,37 @@ export const useComfyRegistrySearchProvider = (): NodePackSearchProvider => {
...
+        return {
+          nodePacks: [],
+          querySuggestions: []
+        }
+      }
+
+      return {
+        nodePacks: listResult.nodes,
+        querySuggestions: []
+      }
```
DrJKL: Optional:
```suggestion
      const nodePacks = listResult?.nodes ?? []

      return {
        nodePacks,
        querySuggestions: []
      }
```

christian-byrne (approved):
LGTM. Tested and it doesn't make repeated requests when clearing search bar (after caching the `/nodes` response once).

PR #8187, opened 2026-01-20
fix: garbage collect subgraph definitions when SubgraphNode is removed

### LINKED ISSUE #8145: node.onRemoved isn't called for nodes inside of subgraphs and node missing graphs sub-subgraphs
### Prerequisites

### What happened?

Sorry, I really don't want to make a million tickets, I've just am trying to fix up my node and am finding all these issues.  So I joined this one.

When switching workflow tabs, all the nodes in the main workflow have their `onRemoved` method called. But not ones in subgraphs.  So event listeners can pile up.

Also, inside an event listener inside a subgraph, I can access the graph via node.graph.  But if I'm in a subgraph in a subgraph, then node.graph doesn't exist.

### Steps to Reproduce

add logs to onRemoved, put that node in a subgraph, watch as it's not called

### ComfyUI Frontend Version

1.36.14

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-8145-node-onRemoved-isn-t-called-for-nodes-inside-of-subgraphs-and-node-missing-graphs-sub--2ec6d73d3650812eb5c3fac15a9f285a) by [Unito](https://www.unito.io)

christian-byrne:
## Summary

When removing a SubgraphNode via `LGraph.remove()`:
- Fire `onRemoved` for all nodes inside the subgraph
- Fire `onNodeRemoved` callback on the subgraph for each inner node
- Remove subgraph definition from `rootGraph.subgraphs` when no other nodes reference it (checks both root graph nodes and nodes inside other subgraphs)

This ensures proper cleanup of subgraph definitions and lifecycle callbacks for nested nodes when subgraph nodes are deleted.

### LGraph.ts
Added SubgraphNode-specific cleanup in `remove()` method that:
1. Iterates inner nodes and fires their `onRemoved` callbacks
2. Fires `onNodeRemoved` on the subgraph for downstream listeners (e.g., minimap)
3. Garbage collects the subgraph definition when no other nodes reference it

### SubgraphNode.ts
Fixed `graph` property to match `LGraphNode` lifecycle contract. Previously it was declared as `override readonly graph` via constructor parameter promotion, which prevented `LGraph.remove()` from setting `node.graph = null`. Changed to a regular mutable property with null guard in `rootGraph` getter.

### LGraph.test.ts
Added 4 tests:
- `removing SubgraphNode fires onRemoved for inner nodes`
- `removing SubgraphNode fires onNodeRemoved callback`
- `subgraph definition is removed when last referencing node is removed`
- `subgraph definition is retained when other nodes still reference it`

## Related

- Fixes #8145
- Part of the subgraph lifecycle cleanup plan (Slice 2: Definition garbage collection)

### DIFF
--- a/src/lib/litegraph/src/LGraph.test.ts
+++ b/src/lib/litegraph/src/LGraph.test.ts
@@ -1,6 +1,10 @@
 import { describe, expect, it } from 'vitest'
 
 import { LGraph, LGraphNode, LiteGraph } from '@/lib/litegraph/src/litegraph'
+import {
+  createTestSubgraphData,
+  createTestSubgraphNode
+} from './subgraph/__fixtures__/subgraphHelpers'
 
 import { test } from './__fixtures__/testExtensions'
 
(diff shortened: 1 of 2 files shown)


christian-byrne:
## Additional Fix: SubgraphNode.graph Type Correction

This PR also includes a fix for a TypeScript error that was blocking the build.

### Problem

`LGraph.remove()` sets `node.graph = null` after removal, but `SubgraphNode` overrode the `graph` property as `readonly` via constructor parameter promotion:

```typescript
constructor(
  override readonly graph: GraphOrSubgraph,  // ❌ readonly + non-nullable
  ...
)
```

This violated the `LGraphNode` lifecycle contract where `graph` is mutable and becomes `null` after removal.

### Solution

Changed `SubgraphNode` to properly conform to the base class contract:

1. **Declared `graph` as a regular mutable property:**
   ```typescript
   override graph: GraphOrSubgraph | null
   ```

2. **Changed constructor to assign manually:**
   ```typescript
   constructor(graph: GraphOrSubgraph, ...) {
     super(...)
     this.graph = graph
   }
   ```

3. **Added `NullGraphError` guard in `rootGraph` getter** (consistent with how other methods handle null graph)

4. **Added null checks** in `ExecutableNodeDTO` and `imagePreviewStore` where `subgraphNode.graph` is accessed

5. **Added non-null assertions in test files** where tests access `graph` immediately after node creation (before any removal could occur)

### Why Not Use a Type Assertion?

The initial workaround `(node as LGraphNode).graph = null` was a code smell - it suppressed a real contract mismatch. The proper fix ensures `SubgraphNode.graph` follows the same lifecycle semantics as `LGraphNode.graph`.

AustinMroz (approved):

christian-byrne:
Merging now so we can test on 1.39

PR #8175, opened 2026-01-20
feat: add feature usage tracker for nightly surveys

christian-byrne:
Introduces `useFeatureUsageTracker` composable that tracks how many times a user has used a specific feature, along with first and last usage timestamps. Data persists to localStorage using `@vueuse/core`'s `useStorage`. This composable provides the foundation for triggering surveys after a configurable number of feature uses. Includes comprehensive unit tests.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8175-feat-add-feature-usage-tracker-for-nightly-surveys-2ee6d73d36508118859ece6fcf17561d) by [Unito](https://www.unito.io)

PR #8189, opened 2026-01-20
feat: add composable to determine if user is eligible for nightly survey(s)

christian-byrne:
## Summary

Adds `useSurveyEligibility` composable that determines whether a user should see a survey based on multiple criteria: nightly localhost build only (`isNightly && !isCloud && !isDesktop`), configurable usage threshold (default 3), 14-day global cooldown between any surveys, once-per-feature-ever display, optional percentage-based sampling, and user opt-out support. All state persists to localStorage. Includes extensive unit tests covering all eligibility conditions.

See:

- https://github.com/Comfy-Org/ComfyUI_frontend/pull/8149
- https://github.com/Comfy-Org/ComfyUI_frontend/pull/8175

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8189-feat-add-composable-to-determine-if-user-is-eligible-for-nightly-survey-s-2ee6d73d365081f088f2fd76032cc60a) by [Unito](https://www.unito.io)


src/platform/surveys/useSurveyEligibility.ts
```diff
@@ -0,0 +1,109 @@
+import { useStorage } from '@vueuse/core'
+import type { MaybeRefOrGetter } from 'vue'
+import { computed, toValue } from 'vue'
+
+import { isCloud, isDesktop, isNightly } from '@/platform/distribution/types'
+
+import { useFeatureUsageTracker } from './useFeatureUsageTracker'
+
+/** @public */
```
christian-byrne: Remove this `/** @public */` annotation.

src/platform/surveys/useSurveyEligibility.ts
```diff
@@ -0,0 +1,109 @@
...
+}
+
+interface SurveyState {
+  seenSurveys: Record<string, number>
+  lastSurveyShown: number | null
+  optedOut: boolean
+}
+
+const STORAGE_KEY = 'Comfy.SurveyState'
+const GLOBAL_COOLDOWN_MS = 14 * 24 * 60 * 60 * 1000 // 14 days
```
christian-byrne: Change global cooldown to 4 days. The release cycle is short and sample size too low to be 14 days.

src/platform/surveys/useSurveyEligibility.ts
```diff
@@ -0,0 +1,109 @@
...
+const DEFAULT_DELAY_MS = 5000
+
+function getStorageState() {
+  return useStorage<SurveyState>(STORAGE_KEY, {
+    seenSurveys: {},
+    lastSurveyShown: null,
+    optedOut: false
+  })
+}
+
```
christian-byrne: Is there a better way to make it testable without requiring this clunky initialization function? Would prefer to just inline `useStorageState`

src/platform/surveys/useSurveyEligibility.ts
```diff
@@ -0,0 +1,109 @@
...
+  return {
+    isEligible,
+    hasReachedThreshold,
+    hasSeenSurvey,
+    isInGlobalCooldown,
+    hasOptedOut,
+    delayMs,
+    markSurveyShown,
+    optOut,
+    resetState
```
DrJKL, line 95: A quick check: Do we need to expose all/each of these, or can we get away with just the aggregate `isEligble` and the actions?

src/platform/surveys/useSurveyEligibility.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  const { useCount } = useFeatureUsageTracker(resolvedConfig.value.featureId)
+
+  const threshold = computed(
+    () => resolvedConfig.value.triggerThreshold ?? DEFAULT_THRESHOLD
+  )
+  const delayMs = computed(
+    () => resolvedConfig.value.delayMs ?? DEFAULT_DELAY_MS
+  )
+  const enabled = computed(() => resolvedConfig.value.enabled ?? true)
```
DrJKL: `enabled` is a little generic. What's enabled?

src/platform/surveys/useSurveyEligibility.ts
DrJKL: Usually through module mocking.

DrJKL: https://github.com/HatcherDX/dx-engine/blob/b6e06f7334de2c12c004c65e2fdf43b4c3800c66/apps/web/src/composables/useTheme.spec.ts#L22

src/platform/surveys/useSurveyEligibility.ts
```diff
@@ -0,0 +1,109 @@
...
+  featureId: string
+  typeformId: string
+  triggerThreshold?: number
+  delayMs?: number
+  enabled?: boolean
+}
+
+interface SurveyState {
+  seenSurveys: Record<string, number>
+  lastSurveyShown: number | null
```
DrJKL: Nit: Can this be optional instead of nullable?
(I wouldn't mind sorting the fields either)

src/platform/surveys/useSurveyEligibility.ts
```diff
@@ -0,0 +1,109 @@
...
+    if (isInGlobalCooldown.value) return false
+    if (hasOptedOut.value) return false
+
+    return true
+  })
+
+  function markSurveyShown() {
+    const now = Date.now()
+    state.value.seenSurveys[resolvedConfig.value.featureId] = now
+    state.value.lastSurveyShown = now
```
DrJKL: Very nit: Maintaining this as a separate value instead of deriving it from the collection can lead to skew.

src/platform/surveys/useSurveyEligibility.test.ts
```diff
@@ -0,0 +1,303 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const SURVEY_STATE_KEY = 'Comfy.SurveyState'
+const FEATURE_USAGE_KEY = 'Comfy.FeatureUsage'
+
+const mockIsNightly = vi.hoisted(() => ({ value: true }))
+const mockIsCloud = vi.hoisted(() => ({ value: false }))
+const mockIsDesktop = vi.hoisted(() => ({ value: false }))
```
DrJKL: Optional, but then you could just return this object instead of having to assign and read from `.value` for each of them.

```suggestion
const mockData = vi.hoisted(() => ({
 isNightly: true,
 isCloud: false,
 isDesktop: false
}))
```

src/platform/surveys/useSurveyEligibility.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    }
+    localStorage.setItem(FEATURE_USAGE_KEY, JSON.stringify(existing))
+  }
+
+  describe('eligibility checks', () => {
+    it('is not eligible when not nightly', async () => {
+      mockIsNightly.value = false
+      setFeatureUsage('test-feature', 5)
+
+      const { useSurveyEligibility } = await import('./useSurveyEligibility')
```
DrJKL: Could you import this at the top instead of doing a dynamic import?

DrJKL (dismissed):

DrJKL (dismissed):

PR #8073, opened 2026-01-15
feat: add per-tab workspace authentication infrastructure

christian-byrne:
## Summary
Add workspace authentication composables and types for per-tab workspace isolation. This infrastructure enables users to work in different workspaces in different browser tabs.

## Changes
- **useWorkspaceAuth composable** - workspace token management
  - Exchange Firebase token for workspace-scoped JWT via `POST /api/auth/token`
  - Auto-refresh tokens 5 minutes before expiry
  - Per-tab sessionStorage caching
- **useWorkspaceSwitch composable** - workspace switching with unsaved changes confirmation
- **WorkspaceWithRole/WorkspaceTokenResponse types** - aligned with backend API
- **firebaseAuthStore.getAuthHeader()** - prioritizes workspace tokens over Firebase tokens
- **useSessionCookie** - uses Firebase token directly (getIdToken()) since getAuthHeader() now returns workspace token

## Backend Dependency
- `POST /api/auth/token` - exchange Firebase token for workspace token
- `GET /api/workspaces` - list user's workspaces

## Related
- https://github.com/Comfy-Org/ComfyUI_frontend/pull/6295

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8073-feat-add-per-tab-workspace-authentication-infrastructure-2e96d73d3650816c8cf9dae9c330aebb) by [Unito](https://www.unito.io)

PR #8194, opened 2026-01-21
feat: add workspace session, auth, and store infrastructure

simula-r:
## Summary
- Add `teamWorkspaceStore` Pinia store for workspace state management (workspaces, members, invites, current workspace)
- Add `workspaceApi` client for workspace CRUD, member management, and invite operations
- Update `useWorkspaceSwitch` composable for workspace switching logic
- Update `useSessionCookie` for workspace-aware sessions
- Update `firebaseAuthStore` for workspace aware auth
- Use `workspaceAuthStore` for workspace auth flow

## Test plan
- [x] 59 unit tests passing (50 store tests + 9 switch tests)
- [x] Typecheck passing
- [x] Lint passing
- [x] Knip passing

Note: This PR depends on the `team_workspaces_enabled` feature flag being available (already in main).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8194-feat-add-workspace-session-auth-and-store-infrastructure-2ef6d73d3650814984afe8ee7ba0a209) by [Unito](https://www.unito.io)


Myestery:
Implementation looks clean at first glance

src/platform/workspace/api/workspaceApi.ts
```diff
@@ -0,0 +1,334 @@
+import type { AxiosResponse } from 'axios'
+import axios from 'axios'
+
+import { t } from '@/i18n'
+import { api } from '@/scripts/api'
+import { useFirebaseAuthStore } from '@/stores/firebaseAuthStore'
+import type { AuthHeader } from '@/types/authTypes'
+
+// Types aligned with backend API
```
christian-byrne:
```suggestion
```

src/platform/workspace/api/workspaceApi.ts
```diff
@@ -0,0 +1,334 @@
...
+  id: string
+  name: string
+  type: WorkspaceType
+}
+
+export interface WorkspaceWithRole extends Workspace {
+  role: WorkspaceRole
+}
+
+// Member type from API
```
christian-byrne:
```suggestion
```

src/platform/workspace/api/workspaceApi.ts
```diff
@@ -0,0 +1,334 @@
...
+  members: Member[]
+  pagination: PaginationInfo
+}
+
+export interface ListMembersParams {
+  offset?: number
+  limit?: number
+}
+
+// Pending invite type from API
```
christian-byrne:
```suggestion
```

src/platform/workspace/api/workspaceApi.ts
```diff
@@ -0,0 +1,334 @@
...
+
+interface CreateWorkspacePayload {
+  name: string
+}
+
+interface UpdateWorkspacePayload {
+  name: string
+}
+
+// API responses
```
christian-byrne:
```suggestion
```

src/platform/workspace/api/workspaceApi.ts
```diff
@@ -0,0 +1,334 @@
...
+    }
+    throw err
+  }
+}
+
+/**
+ * Wrapper that uses Firebase ID token directly (not workspace token).
+ * Used for token exchange where we need the Firebase token to get a workspace token.
+ */
+async function withFirebaseAuth<T>(
```
christian-byrne: The semantics of deciding which token to use feels like logic that should be confined to upper-layer auth domain rather than be specific to the workspace-related business logic. I.e., would it make sense to move the majority of this file to https://github.com/Comfy-Org/ComfyUI_frontend/tree/main/src/platform/auth?

src/platform/workspace/services/sessionManager.ts
```diff
@@ -0,0 +1,148 @@
...
+   * Clear the workspace token from sessionStorage
+   */
+  clearWorkspaceToken(): void {
+    try {
+      sessionStorage.removeItem(WORKSPACE_STORAGE_KEYS.TOKEN)
+      sessionStorage.removeItem(WORKSPACE_STORAGE_KEYS.EXPIRES_AT)
+    } catch {
+      console.warn('Failed to clear workspace token from sessionStorage')
+    }
+  },
```
christian-byrne: This might be overlapping a bit with https://github.com/Comfy-Org/ComfyUI_frontend/tree/main/src/platform/auth/workspace

src/stores/firebaseAuthStore.ts
```diff
@@ -201,10 +203,19 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
     return useApiKeyAuthStore().getAuthHeader()
   }
 
+  /**
+   * Returns Firebase auth header for user-scoped endpoints (e.g., /customers/*).
+   * Use this for endpoints that need user identity, not workspace context.
+   */
```
christian-byrne, line 209: Might be worthwhile to make this module (`firebaseAuthStore`) just be something like `authStore`, then all the logic of determining which token to use can be encapsulated here and not be responsibility of callers. WDYT?

Then we can also organizationally reconcile it better with this existing function whose existence might otherwise be confusing after we add a colocated public `getFirebaseAuthHeader` function https://github.com/Comfy-Org/ComfyUI_frontend/blob/9669100c14174038d9754d8a9bffc66edf45431e/src/stores/firebaseAuthStore.ts#L175

christian-byrne (commented):
Looks good - just posing some questions about organization and some suggestions to drop some comments.

src/stores/firebaseAuthStore.ts
simula-r: I agree with that. Calling it firebaseAuthStore and then mixing firebase auth token and workspace auth token is confusing. And callingfirebaseAuthStore.getAuthHeader()  returns workspace auth. Not ideal. I think for now we shouldnt rename to authStore in this PR because its used 144 places. But can do after.

src/platform/workspace/services/sessionManager.ts
simula-r: I am going back to using workspaceAuthStore you made in your PR https://github.com/Comfy-Org/ComfyUI_frontend/pull/8073

src/stores/firebaseAuthStore.ts
christian-byrne: @coderabbitai can you make an issue for the above comments?

christian-byrne (approved):

PR #8206, opened 2026-01-21
Fix doubled player on VHS LoadAudio in vue

AustinMroz:
In vue mode, the VHS Load Audio (Upload) node had 2 audio previews. This occurred because the native AudioPreview widget was being applied to any combo widget with the name `audio`. This native preview does not support the advanced preview functions VHS provides like seeking to specific start time, trimming to a target duration, or converting from formats the browser may not support.

This is fixed through a fairly involved cleanup to instead display the litegraph AudioUI widget as an AudioPreview widget when in vue mode.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8206-Fix-doubled-player-on-VHS-LoadAudio-in-vue-2ef6d73d365081ce8907dca2706214a1) by [Unito](https://www.unito.io)


src/extensions/core/uploadAudio.ts
```diff
@@ -118,26 +128,18 @@ app.registerExtension({
...
-          () => {
-            if (!audioUIWidget.element) return
-            audioUIWidget.element.pause()
-            audioUIWidget.element.src = ''
-            audioUIWidget.element.remove()
-          }
-        )
+        let value = ''
+        audioUIWidget.options.getValue = () => value
+        audioUIWidget.options.setValue = (v) => (value = v)
```
DrJKL, line 142: This seems sus...

DrJKL (approved):

src/extensions/core/uploadAudio.ts
AustinMroz: Agreed. I'm real unhappy with the setup, but the existing DOMwidget code has already defined `widget.value`. It's impossible to set `widget.value` on a DOMWidget unless you set up a getter and setter like this.

PR #8250, opened 2026-01-22
Linear: progressbar, tooltips, and output fixes

AustinMroz:
- Fixes only the first output being displayed in linear mode after the jobs migration
- Fixes selected output no longer scrolling into view in history
- Adds a progress bar indicator on running job
  image
- Moves linear toggle button to v-tooltip
- Fixes placeholder sometimes continuing to display after a new output.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8250-Linear-progressbar-tooltips-and-output-fixes-2f06d73d365081ca9fa3ebf0e2516487) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/linearMode/OutputHistory.vue
+++ b/src/renderer/extensions/linearMode/OutputHistory.vue
@@ -81,7 +81,9 @@ watch(selectedIndex, () => {
   const [index, key] = selectedIndex.value
   if (!outputsRef.value) return
 
-  const outputElement = outputsRef.value?.children?.[index]?.children?.[key]
+  const outputElement = outputsRef.value?.querySelectorAll(
+    `[data-output-index="${index}"]`
+  )?.[key]
   if (!outputElement) return
 
   //container: 'nearest' is nice, but bleeding edge and chrome only
@@ -318,6 +320,7 @@ useEventListener(document.body, 'keydown', (e: KeyboardEvent) => {
                   'border-2'
               )
             "
+            :data-output-index="index"
             :src="output.url"
             @click="selectedIndex = [index, key]"
           />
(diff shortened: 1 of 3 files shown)


src/renderer/extensions/linearMode/OutputHistory.vue
```diff
@@ -72,36 +88,76 @@ watch(selectedIndex, () => {
...
+
+  const outputRef = useAsyncState(
+    getJobDetail(user_metadata.promptId).then((jobDetail) => {
+      if (!jobDetail?.outputs) return []
+      return Object.entries(jobDetail.outputs).flatMap(flattenNodeOutput)
+    }),
+    []
+  ).state
+  outputsCache[item!.id] = outputRef
+  return outputRef
```
christian-byrne, line 145: This code seems like it might fit better in a centralized location or composable rather than in a component, WDYT?

AustinMroz: All the different history implementations are in need of a lot of polish/consolidation. The other implementations are close, but insufficient to fit the needs I have here. I think the correct move is to spend more time polishing those in order to be usable, but I couldn't justify the time that would be required to do so right now

That the implementation here isn't in a centralized location is for the best. I want to avoid my implementation becoming another competing standard.

The `@/stores/queueStore` implementation is closest, it just
- Doesn't support loading more assets for use with `useInfiniteScroll`
- Incorrectly casts outputs to `ResultItem[]`
  - Which means text outputs aren't handled
- Is a little too needy in required inputs for construction of `TaskItemImpl`

christian-byrne (approved):

PR #8251, opened 2026-01-22
Further number widget fixes

AustinMroz:
- The slider indicator is now only rounded at the ends and doesn't display outside the widget at small values
- Prevents a bug where scrubbing would result in a 1/10 chance of causing text selection after a completed scrub.

| Before | After |
| ------ | ----- |
| before | after|

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8251-Further-number-widget-fixes-2f06d73d3650811f9548ded527ca16ae) by [Unito](https://www.unito.io)


DrJKL (approved):
Screenshots would be nice, but 👍🏻

PR #8253, opened 2026-01-23
refactor: improve TypeScript patterns in test files (Group 1/8)

Myestery:
## Summary
Improves type safety in test files by replacing unsafe type patterns with proper TypeScript idioms.

## Changes
- Define typed `TestWindow` interface extending `Window` for Playwright tests with custom properties
- Use `Partial<HTMLElement>` with single type assertion for DOM element mocks
- Remove redundant type imports
- Fix `console.log` → `console.warn` in test fixture

## Files Changed
16 test files across browser_tests, packages, and src/components

## Test Plan
- ✅ `pnpm typecheck` passes
- ✅ No new `any` types introduced
- ✅ All pre-commit hooks pass

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8253-refactor-improve-TypeScript-patterns-in-test-files-Group-1-8-2f16d73d365081548f9ece7bcf0525ee) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/featureFlags.spec.ts
+++ b/browser_tests/tests/featureFlags.spec.ts
@@ -2,22 +2,6 @@ import { expect } from '@playwright/test'
 
 import { comfyPageFixture as test } from '../fixtures/ComfyPage'
 
-interface CapturedMessages {
-  clientFeatureFlags: { type: string; data: Record<string, unknown> } | null
-  serverFeatureFlags: Record<string, unknown> | null
-}
-
-interface AppReadiness {
-  featureFlagsReceived: boolean
-  apiInitialized: boolean
-  appInitialized: boolean
-}
-
-interface TestWindow extends Window {
-  __capturedMessages: CapturedMessages
-  __appReadiness: AppReadiness
-}
-
 test.beforeEach(async ({ comfyPage }) => {
   await comfyPage.setSetting('Comfy.UseNewMenu', 'Disabled')
 })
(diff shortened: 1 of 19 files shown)


browser_tests/tests/featureFlags.spec.ts
```diff
@@ -274,7 +290,8 @@ test.describe('Feature Flags', () => {
     // Set up monitoring before navigation
     await newPage.addInitScript(() => {
       // Track when various app components are ready
-      ;(window as any).__appReadiness = {
+      const testWindow = window as TestWindow
```
AustinMroz:
```suggestion
      const testWindow: Partial<TestWindow> = window
```

browser_tests/tests/nodeHelp.spec.ts
```diff
@@ -345,7 +347,9 @@ This is documentation for a custom node.
 
       // Find and select a custom/group node
       const nodeRefs = await comfyPage.page.evaluate(() => {
-        return window['app'].graph.nodes.map((n: any) => n.id)
+        return window['app'].graph.nodes.map(
```
AustinMroz: Do you think there's a way we could globally type window in tests like we have in `@/types/index.ts`?

src/components/common/UserAvatar.test.ts
```diff
@@ -1,3 +1,5 @@
+import type { ComponentProps } from 'vue-component-type-helpers'
```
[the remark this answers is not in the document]
Myestery: we have no case where the props are being used with that type forcing the attributes in it to be non optional so this is okay

AustinMroz (dismissed):

browser_tests/tests/nodeHelp.spec.ts
Myestery: let me explore that

Myestery: so the global typing we have in @/types/index.ts already applies to tests

Myestery: I have added the WS fields to the global window

AustinMroz (approved):

PR #8259, opened 2026-01-23
fix: prevent middle-click paste duplicating workflow on Linux

### LINKED ISSUE #4464: [Bug]: Middle Mouse Pan Duplicates Entire Workflow
### Custom Node Testing

### Frontend Version

ComfyUI 0.3.44
ComfyUI_frontend v1.23.4
ComfyUI-Manager V3.34

### Expected Behavior

Middle drag should only pan the canvas
No paste operation should occur
Original workflow should remain unchanged

### Actual Behavior

Canvas pans correctly during middle drag
On release, entire workflow is duplicated as new nodes
Duplicate nodes are often offset/offscreen and can be difficult to notice immediately
Leads to cluttered workflows and confusion

### Steps to Reproduce

Use ComfyUI on Linux system
Create a workflow with multiple nodes
Select any text elsewhere on the system (to populate PRIMARY clipboard)
Middle drag on the graph canvas to pan
Release middle button

### Debug Logs

```powershell
none
```

### Browser Logs

none

### Setting JSON

{
    "Comfy.TutorialCompleted": true,
    "Comfy.Queue.ImageFit": "contain"
}

Mozilla Firefox

## Related Issues
- **Similar but different from**: https://github.com/Comfy-Org/ComfyUI_frontend/issues/2409
  
### Key Differences:
- **Scope**: This issue affects the entire `#graph-canvas`, not just the mask editor
- **Trigger**: Occurs with any text in PRIMARY clipboard (normal text selection), not just copied nodes
- **Frequency**: Happens on every middle drag anywhere on canvas, not limited to specific tools

The existing issue #2409 appears to be a subset of this broader middle-click paste problem. While that issue focuses on the mask editor with node data in clipboard, this bug report covers the more general case affecting all canvas interactions with any clipboard content.

### Clipboard Analysis
- Inspected PRIMARY clipboard content - contained last selected text, not node data
- Confirmed middle-click paste works correctly in other applications (text editors, terminals)
- Verified this is specific to ComfyUI web interface, not system-wide issue

### Technical Analysis
Created test web page to reproduce the issue:
- Confirmed middle-click paste can be easily prevented with `preventDefault()`
- Issue occurs because ComfyUI doesn't handle middle-click events properly
- The `auxclick` event (fired on middle drag release) needs to be prevented

### Root Cause
ComfyUI's `#graph-canvas` element doesn't prevent the default middle-click paste behavior while still allowing the drag/pan functionality.

## Proposed Solution
Add event listener to `#graph-canvas` to prevent middle-click paste:

```javascript
canvas.addEventListener('auxclick', function(e) {
    if (e.button === 1) {  // Middle mouse button
        e.preventDefault();
    }
}, true);
```

This solution:
- Preserves middle drag panning functionality
- Prevents unwanted paste operations
- Doesn't affect other mouse interactions
- Uses capture phase to intercept before other handlers

## Notes
- I know LMB can pan!! But muscle memory from other CAD software (FreeCAD, OpenSCAD, KiCAD) is particularly hard to shake
- Issue manifests even with simple text selections (emails, URLs, etc.) in PRIMARY clipboard maybe even empty
- Affects core canvas navigation, not specialized editing tools

This suggests the fix should be implemented at the canvas level rather than in specific components, which would resolve both this issue and #2409 simultaneously.

## Temporary Workaround
Users can inject this code in browser console as a temporary fix:

```javascript
document.querySelector('#graph-canvas').addEventListener('auxclick', function(e) {
    if (e.button === 1) {
        e.preventDefault();
    }
}, true);
```

## Environment
- **OS**: Artix Linux.  KDE. Wayland
- **Browser**: Tested on Firefox only

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-4464-Bug-Middle-Click-Paste-Duplicates-Entire-Workflow-During-Middle-Drag-on-Graph-Canva-2326d73d365081f390d1c4a08b9a389d) by [Unito](https://www.unito.io)

bimakw:
## Summary

Adds `auxclick` event listener to prevent the browser's default middle-click paste behavior on Linux systems.

**Problem:** On Linux, middle-clicking anywhere triggers a paste from the PRIMARY clipboard. When middle-dragging to pan the canvas, this causes the entire workflow to be duplicated as new nodes on mouse release.

**Solution:** Add `auxclick` event listener with `preventDefault()` to the graph canvas, blocking the paste while preserving pan functionality.

## Changes

- Add `auxclick` event listener in `bindEvents()` 
- Add corresponding `removeEventListener` in `unbindEvents()`

## Test Plan

- [ ] On Linux: Middle-drag to pan canvas - should pan without duplicating nodes
- [ ] On Linux: Verify left/right click behaviors unchanged
- [ ] On Windows/macOS: Verify no regression (auxclick should have no effect)

Fixes #4464

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8259-fix-prevent-middle-click-paste-duplicating-workflow-on-Linux-2f16d73d3650812b98f9cada699f5508) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -1965,6 +1965,8 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
     canvas.addEventListener('pointercancel', this._mousecancel_callback, true)
 
     canvas.addEventListener('contextmenu', this._doNothing)
+    // Prevent middle-click paste on Linux (PRIMARY clipboard) - fixes #4464
+    canvas.addEventListener('auxclick', this._doNothing)
 
     // Keyboard
     this._key_callback = this.processKey.bind(this)
@@ -2003,6 +2005,7 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
     canvas.removeEventListener('keydown', this._key_callback!)
     document.removeEventListener('keyup', this._key_callback!)
     canvas.removeEventListener('contextmenu', this._doNothing)
+    canvas.removeEventListener('auxclick', this._doNothing)
     canvas.removeEventListener('dragenter', this._doReturnTrue)
 
     this._mousedown_callback = undefined
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -172,7 +172,7 @@ interface IDialogExtensions extends ICloseable {
   is_modified: boolean
 }
 
-interface IDialog extends HTMLDivElement, IDialogExtensions {}
+interface IDialog extends HTMLDivElement, IDialogExtensions { }
 type PromptDialog = Omit<IDialog, 'modified'>
 
 interface IDialogOptions {
@@ -839,15 +839,15 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
           const linkReleaseContext =
             this.linkConnector.state.connectingTo === 'input'
               ? {
-                  node_from: firstLink.node as LGraphNode,
-                  slot_from: firstLink.fromSlot as INodeOutputSlot,
-                  type_filter_in: firstLink.fromSlot.type
-                }
+                node_from: firstLink.node as LGraphNode,
+                slot_from: firstLink.fromSlot as INodeOutputSlot,
+                type_filter_in: firstLink.fromSlot.type
+              }
               : {
-                  node_to: firstLink.node as LGraphNode,
-                  slot_to: firstLink.fromSlot as INodeInputSlot,
-                  type_filter_out: firstLink.fromSlot.type
-                }
+                node_to: firstLink.node as LGraphNode,
+                slot_to: firstLink.fromSlot as INodeInputSlot,
+                type_filter_out: firstLink.fromSlot.type
+              }
 
           const afterRerouteId = firstLink.fromReroute?.id
 
@@ -1221,8 +1221,8 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
     }
   }
 
-  static onMenuCollapseAll() {}
-  static onMenuNodeEdit() {}
+  static onMenuCollapseAll() { }
+  static onMenuNodeEdit() { }
 
   /** @param _options Parameter is never used */
   static showMenuNodeOptionalOutputs(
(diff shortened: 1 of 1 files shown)


christian-byrne (changes requested):
This doesn't happen to me on Ubuntu 24. 

Also wouldn't we want to restrict the behavior to the target system?

bimakw:
Updated — now only prevents middle click (`button === 1`) instead of blocking all non-primary buttons. See latest commit.

Regarding OS detection: this is a PRIMARY clipboard paste issue specific to X11/XWayland setups, which varies by distro, display server, and browser combo. The reporter was on Artix Linux + KDE + Wayland + Firefox — it might not repro on GNOME-based Ubuntu 24 since clipboard handling differs.

I chose not to add OS detection because:
- `navigator.platform` is deprecated and unreliable for this
- Preventing middle-click auxclick default on a canvas element is harmless on all platforms — there's no useful default behavior being lost
- The existing `contextmenu` handler follows the same pattern (no OS check)

So it's effectively a no-op on systems where the issue doesn't occur, and a fix where it does.

Pauan:
> @bimakw Regarding OS detection: this is a PRIMARY clipboard paste issue specific to X11/XWayland setups, which varies by distro, display server, and browser combo.

The PRIMARY clipboard (and middle click to paste) is hardcoded into X11, and affects all distros that use X11. Because it's hardcoded in the source code, it's impossible to disable middle click paste in X11.

GNOME actually patched the X11 source code to make it possible to disable middle mouse click, but this patch only affects GTK apps.

There are [several hacky workarounds](https://askubuntu.com/questions/4507/how-do-i-disable-middle-mouse-button-click-paste) which (partially) disable middle click paste, but they all have various problems. The current best solution seems to be [XMousePasteBlock](https://github.com/milaq/XMousePasteBlock).

Wayland uses a different system, where the PRIMARY clipboard and middle click to paste is emulated in userland, and can be disabled by the user.

This whole situation is *very* messy, so I agree with you that it doesn't make sense to add in detection code. It should just always disable middle click.

christian-byrne (dismissed):
Approving — the rebuttal to the OS-detection concern is sound:

- `navigator.platform` is deprecated and unreliable
- The behavior varies by X11/XWayland/display server/browser combo, making OS detection infeasible
- The existing `contextmenu` handler follows the same pattern (no OS check) — it uses `_doNothing` unconditionally on all platforms
- Blocking auxclick (button === 1) on canvas is harmless — the existing middle-click behavior uses `pointerdown`, not `auxclick`, so there's no conflict
- The `button === 1` guard correctly scopes this to middle-click only

Note: Would be good to add a brief code comment explaining why auxclick is blocked (Linux X11 clipboard paste prevention) for future maintainers.

christian-byrne:
Apologies for such a delayed response, appreciate the help on this to everyone who contrbuted.

christian-byrne:
I pushed some formatting changes manually since the automatic formatter is failing on forks currently. Will merge after CI checks.

Tekaiguy:
Thank you

PR #8266, opened 2026-01-23
fix: use authenticated API for remote config polling

christian-byrne:
## Summary
- Fixes remote config polling to use authenticated API
- Consolidates `loadRemoteConfig` into `refreshRemoteConfig` with auth control
- Adds unit tests for both auth modes

## Problem
The cloud extension's polling interval was using unauthenticated `fetch`, causing it to receive only default feature flags instead of user-specific configurations.

**Root cause:**
1. Bootstrap called `loadRemoteConfig()` (raw `fetch`, no auth) - correct, auth not initialized yet
2. Extension watch called `refreshRemoteConfig()` (`api.fetchApi`, with auth) - correct
3. Extension interval called `loadRemoteConfig()` (raw `fetch`, no auth) - **bug**

## Solution
- Consolidate into single `refreshRemoteConfig()` with optional `useAuth` parameter (defaults to `true`)
- Bootstrap: `refreshRemoteConfig({ useAuth: false })`
- Polling: `refreshRemoteConfig()` (authenticated by default)

## Test Plan
- Unit tests verify both auth modes
- `pnpm typecheck`, `pnpm lint`, `pnpm test:unit` all pass

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8266-fix-use-authenticated-API-for-remote-config-polling-2f16d73d3650817ea7b0e3a7e3ccf12a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/remoteConfig/refreshRemoteConfig.test.ts
+++ b/src/platform/remoteConfig/refreshRemoteConfig.test.ts
@@ -1,0 +1,109 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import { api } from '@/scripts/api'
+
+import { refreshRemoteConfig } from './refreshRemoteConfig'
+import { remoteConfig } from './remoteConfig'
+
+vi.mock('@/scripts/api', () => ({
+  api: {
+    fetchApi: vi.fn()
+  }
+}))
+
+global.fetch = vi.fn()
+
+describe('refreshRemoteConfig', () => {
+  const mockConfig = { feature1: true, feature2: 'value' }
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+    remoteConfig.value = {}
+    window.__CONFIG__ = {}
+  })
+
+  describe('with auth (default)', () => {
+    it('uses api.fetchApi when useAuth is true', async () => {
+      vi.mocked(api.fetchApi).mockResolvedValue({
+        ok: true,
+        json: async () => mockConfig
+      } as Response)
+
+      await refreshRemoteConfig({ useAuth: true })
+
+      expect(api.fetchApi).toHaveBeenCalledWith('/features', {
+        cache: 'no-store'
+      })
+      expect(global.fetch).not.toHaveBeenCalled()
+      expect(remoteConfig.value).toEqual(mockConfig)
+      expect(window.__CONFIG__).toEqual(mockConfig)
+    })
+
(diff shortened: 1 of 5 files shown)


src/platform/remoteConfig/refreshRemoteConfig.ts
DrJKL, line 1: Nit: I know it's not coming from this PR, but this feels like it should be a part of the remoteConfig module, not in a separate file.

src/platform/remoteConfig/refreshRemoteConfig.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    remoteConfig.value = {}
+    window.__CONFIG__ = {}
+  })
+
+  describe('with auth (default)', () => {
+    it('uses api.fetchApi when useAuth is true', async () => {
+      vi.mocked(api.fetchApi).mockResolvedValue({
+        ok: true,
+        json: async () => mockConfig
+      } as Response)
```
DrJKL, line 30: Does this need the type assertion?
Could you get away with `satisfies Partial<Response>`?

src/platform/remoteConfig/refreshRemoteConfig.test.ts
```diff
@@ -0,0 +1,109 @@
...
+import { refreshRemoteConfig } from './refreshRemoteConfig'
+import { remoteConfig } from './remoteConfig'
+
+vi.mock('@/scripts/api', () => ({
+  api: {
+    fetchApi: vi.fn()
+  }
+}))
+
+global.fetch = vi.fn()
```
DrJKL, line 14: Very nit: I usually like [stubGlobal](https://vitest.dev/api/vi.html#vi-stubglobal) for consistency and control

src/platform/remoteConfig/refreshRemoteConfig.test.ts
```diff
@@ -0,0 +1,109 @@
...
+      expect(global.fetch).not.toHaveBeenCalled()
+    })
+  })
+
+  describe('without auth', () => {
+    it('uses raw fetch when useAuth is false', async () => {
+      vi.mocked(global.fetch).mockResolvedValue({
+        ok: true,
+        json: async () => mockConfig
+      } as Response)
```
DrJKL, line 60: Nit: This pattern seems repeated enough times in here to be worth extracting.

DrJKL (approved):

PR #8272, opened 2026-01-23
fix: image selection modal Inputs Outputs filtering is not working

LittleSound:
filteredItems only updates during search, and does not update when the filter changes. Trigger a re-search when items change through the updateKey prop chain.

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
@@ -78,6 +78,8 @@ const maxSelectable = computed(() => {
   return 1
 })
 
+const itemsKey = computed(() => props.items.map((item) => item.id).join('|'))
+
 const filteredItems = ref<DropdownItem[]>([])
 
 const defaultSorter = computed<SortOption['sorter']>(() => {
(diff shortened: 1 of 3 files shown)


LittleSound:
> The watcher in `FormSearchInput.vue` is shallow (no `{ deep: true }` flag) and watches `updateKeyRef` via `watch([debouncedSearchQuery, updateKeyRef], ..., { immediate: true })`. When bound directly to the `items` prop (line 212), reference changes are detected, but in-place mutations are not. If items are modified without creating a new array, the searcher won't re-run, leaving the search stale.

fixed

christian-byrne (commented):
How about hoisting the search effect to `FormDropdown` since it owns both `items` and `filteredItems`? Then would we be able to:

- Remove `updateKey` prop from all 4 components
- Remove `itemsKey` computed from `FormDropdown`
- `FormSearchInput` becomes a simple input (keeps debounce UI/spinner, emits query changes)

This follows general idiomatic approach where the inputs emit user intent (query changes) and containers run effects (search/filtering) based on reactive dependencies they own.

Will it work?

LittleSound:
Thanks for the suggestion! I considered this approach, but I'd prefer to keep the current design for a few reasons:

FormSearchInput is designed to be reusable across multiple contexts, not just FormDropdown. It encapsulates both debounce logic and loading state management (the spinner animation). The isQuerying state needs to know when the search completes, not just when the debounced query changes:

```js
.finally(() => {
  if (!isCleanup) isQuerying.value = false
})
```

If we hoist the search effect to FormDropdown, FormSearchInput would lose visibility into search completion. We'd need to pass an isLoading prop back down, which adds complexity without much gain.

The updateKey prop is a lightweight way to let the container signal "re-run your search" while keeping the component self-contained. The trade-off feels acceptable for maintaining reusability.

christian-byrne (approved):

PR #8275, opened 2026-01-23
fix: `recomputeInsideNodes` does not support nested group processing

LittleSound:
related https://github.com/Comfy-Org/ComfyUI_frontend/pull/8274

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8275-fix-recomputeInsideNodes-does-not-support-nested-group-processing-2f16d73d3650815e9b44dbd182f8dd74) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphGroup.test.ts
+++ b/src/lib/litegraph/src/LGraphGroup.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect } from 'vitest'
 
-import { LGraphGroup } from '@/lib/litegraph/src/litegraph'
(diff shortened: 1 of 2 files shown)


AustinMroz (commented):
Having `recomputeInsideNodes` take a Set of visited group ids would be preferable over maxDepth

In the following example, the current code has the innermost group recompute it's contents 4 tiems when the outermost is selected.
image

LittleSound:
> Having recomputeInsideNodes take a Set of visited group ids would be preferable over maxDepth

You are right. I added a Set.

AustinMroz (approved):

PR #8276, opened 2026-01-23
fix: default image input for the template is displayed as empty on dropdown selection

LittleSound:
The image input for nodes loaded from templates appears empty in the Properties Panel.

When the widget's current value (saved in the template) is not in the available file list returned by the server, the selectedSet is empty, causing a placeholder to be displayed instead of the actual value.

Added a missingValueItem computed property in WidgetSelectDropdown.vue. When the current value is not in inputItems or outputItems, it creates a fallback item and adds it to dropdownItems.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8276-fix-default-image-input-for-the-template-is-displayed-as-empty-on-dropdown-selection-2f16d73d3650817eaad5e4e33637fb74) by [Unito](https://www.unito.io)


christian-byrne (commented):
If you run with `pnpm dev:cloud` and open a template, does the same issue occur? If not, can we investigate how we might use the same solution that the nodes use here instead of using fallback?

christian-byrne:
If we do use a placeholder, it might be better to just use a tooltip or something explanatory rather than "undefined". See the visual diff in this test result: https://e867908e.comfyui-playwright-chromium.pages.dev/#?testId=0a6cb6e601f2f6d19e73-f6f8443e643451dee3c7

LittleSound:
I cannot access the cloud environment. However, I tried to fix it.

AustinMroz (dismissed):
Functionality looks fine. Obligatory style nit that we're pushing for `undefined` instead of `null`

PR also fixes display of nodes on cloud.

| Before | After |
| ------ | ----- |
| before | after|

LittleSound:
> Functionality looks fine. Obligatory style nit that we're pushing for undefined instead of null

fixed.

christian-byrne:
I am looking into why the snapshot upload does not work.

PR #8278, opened 2026-01-23
fix: add tooltips for various settings in the right side panel

LittleSound:
CleanShot 2026-01-23 at 21 27 45@2x

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8278-fix-add-tooltips-for-various-settings-in-the-right-side-panel-2f16d73d365081e2816cfdf24fa7687f) by [Unito](https://www.unito.io)


src/locales/en/main.json
christian-byrne, line 1: Can we re-use the existing tooltips/descriptions for these features so the terminology/phrrasing is standardized across the UI?

LittleSound: fixed

christian-byrne (approved):

PR #8279, opened 2026-01-23
fix: hide label of textarea in right side panel + align switch to the left

LittleSound:
image
 | 
CleanShot 2026-01-23 at 21 25 04@2x

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8279-fix-hide-label-of-textarea-in-right-side-panel-align-switch-to-the-left-2f16d73d365081e9b932fb2be873b660) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetToggleSwitch.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetToggleSwitch.vue
@@ -29,10 +29,11 @@
 
 <script setup lang="ts">
 import ToggleSwitch from 'primevue/toggleswitch'
-import { computed, inject } from 'vue'
+import { computed } from 'vue'
 
 import type { IWidgetOptions } from '@/lib/litegraph/src/types/widgets'
 import type { SimplifiedWidget } from '@/types/simplifiedWidget'
+import { useHideLayoutField } from '@/types/widgetTypes'
 import { cn } from '@/utils/tailwindUtil'
 import {
   STANDARD_EXCLUDED_PROPS,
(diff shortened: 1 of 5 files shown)


christian-byrne (dismissed):
In future can we use typed injection keys?

src/renderer/extensions/vueNodes/widgets/components/WidgetToggleSwitch.vue
```diff
@@ -1,6 +1,10 @@
 <template>
   <WidgetLayoutField :widget>
-    <div class="ml-auto flex w-fit items-center gap-2">
+    <div
+      :class="
+        cn('flex w-fit items-center gap-2', { 'ml-auto': !hideLayoutField })
```
christian-byrne: The semantics seem strange here -- the widget is left-alligned if the widget label is hidden? Shouldn't the alignment be set by its own prop?

LittleSound: What do you mean by "own prop"?

LittleSound:
> In future can we use typed injection keys?

fixed

src/renderer/extensions/vueNodes/widgets/components/WidgetToggleSwitch.vue
LittleSound: I don't think there should be any problem with this. If there is Field Layout, this style is used for layout in Field Layout. So we remove it when there is no Field Layout.

DrJKL: Nit: Using booleans like this is more common.
```suggestion
        cn('flex w-fit items-center gap-2', !hideLayoutField && 'ml-auto')
```

DrJKL (dismissed):

PR #8283, opened 2026-01-23
Add expandable output stacks to assets list view

benceruleanlu:
Add expandable output stacks to the assets list view.

Monolith ver. of https://github.com/Comfy-Org/ComfyUI_frontend/pull/8298 and its children

List view currently collapses multi-output jobs into a single row, which makes sibling outputs easy to miss and causes selection/zoom behavior to drift once items are expanded elsewhere. This change adds a stack toggle to list rows, expands child outputs derived from job data, and keeps list-view selection and gallery navigation aligned with the expanded list. Output mapping and “load full outputs” checks are centralized so folder view and stacks share the same helper, and job-detail parsing now yields previewable outputs for the list view. Asset actions now prefer metadata prompt IDs to support the composite IDs used by stacked outputs.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8283-Add-expandable-output-stacks-to-assets-list-view-2f16d73d365081a99fc6f1519ac2e57c) by [Unito](https://www.unito.io)


src/platform/assets/composables/useMediaAssetActions.ts
```diff
@@ -203,9 +203,10 @@ export function useMediaAssetActions() {
     const targetAsset = asset ?? mediaContext?.asset.value
     if (!targetAsset) return
 
-    // Try asset.id first (OSS), then fall back to metadata (Cloud)
     const metadata = getOutputAssetMetadata(targetAsset.user_metadata)
-    const promptId = targetAsset.id || metadata?.promptId
+    const promptId =
+      metadata?.promptId ||
+      (getAssetType(targetAsset) === 'output' ? targetAsset.id : undefined)
```
benceruleanlu, line 144: metadata.promptId is set by the frontend mappers here, so the comment was outdated. In cloud, asset.id is the asset UUID (not the job/prompt ID), so preferring metadata.promptId is the correct behavior.

src/platform/assets/utils/outputAssetUtil.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  return mapOutputsToAssetItems({
+    promptId: metadata.promptId,
+    outputs: outputsToDisplay,
+    createdAt,
+    executionTimeInSeconds: metadata.executionTimeInSeconds,
+    workflow: metadata.workflow,
+    excludeFilename: excludeOutputKey
+  })
+}
```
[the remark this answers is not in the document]
benceruleanlu: It was added after opening

src/platform/assets/composables/useOutputStacks.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  return {
+    assetItems,
+    selectableAssets,
+    isStackExpanded,
+    toggleStack
+  }
+}
```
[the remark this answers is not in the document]
benceruleanlu: It was added after opening

src/platform/assets/utils/outputAssetUtil.ts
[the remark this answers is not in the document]
benceruleanlu: resolving to keep PR readable scroll-wise

src/services/jobOutputCache.ts
```diff
@@ -75,6 +76,40 @@ export async function getOutputsForTask(
...
+            })
+        )
+      })
+  )
+
+  return ResultItemImpl.filterPreviewable(resultItems)
+}
+
+function isResultItem(item: unknown): item is ResultItem {
+  return typeof item === 'object' && item !== null
```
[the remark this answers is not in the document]
benceruleanlu: fetchJobDetail parses zJobDetail and zTaskOutput/zResultItem, so bad shapes should be rejected before getPreviewableOutputs runs, I think this is OK for now, for this PR.

src/platform/assets/components/AssetsListItem.vue
```diff
@@ -74,13 +74,45 @@
...
+          :class="
+            cn(
+              stackExpanded
+                ? 'icon-[lucide--chevron-down]'
+                : 'icon-[lucide--chevron-right]',
+              'size-3'
+            )
+          "
+        />
+      </Button>
```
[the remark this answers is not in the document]
benceruleanlu: resolved

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -404,7 +406,10 @@ const showEmptyState = computed(
     activeJobsCount.value === 0
 )
 
-watch(displayAssets, (newAssets) => {
+watch(visibleAssets, (newAssets) => {
+  // Alternative: keep hidden selections and surface them in UI; for now prune
+  // so selection stays consistent with what this view can act on.
+  reconcileSelection(newAssets)
   if (currentGalleryAssetId.value && galleryActiveIndex.value !== -1) {
```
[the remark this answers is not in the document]
benceruleanlu: true, made the expensive refactor to bring useStack up to shared parent

src/platform/assets/composables/useAssetSelection.ts
```diff
@@ -88,6 +88,38 @@ export function useAssetSelection() {
...
+
+    for (const id of selectionStore.selectedAssetIds) {
+      if (visibleIds.has(id)) {
+        nextSelectedIds.push(id)
+      }
+    }
+
+    if (nextSelectedIds.length === selectionStore.selectedAssetIds.size) {
+      return
+    }
```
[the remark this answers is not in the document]
benceruleanlu: switched to also rely on asset IDs

src/services/jobOutputCache.ts
```diff
@@ -75,6 +76,40 @@ export async function getOutputsForTask(
...
+}
+
+export function getPreviewableOutputsFromJobDetail(
+  jobDetail?: JobDetail
+): ResultItemImpl[] {
+  return getPreviewableOutputs(jobDetail?.outputs)
+}
```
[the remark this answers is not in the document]
benceruleanlu: Added already

src/services/jobOutputCache.ts
Myestery:
```suggestion
  return item instanceOf ResultItem
```

src/platform/assets/utils/outputAssetUtil.ts
```diff
@@ -0,0 +1,84 @@
...
+  excludeFilename
+}: OutputAssetMapOptions): AssetItem[] {
+  const createdAtValue = createdAt ?? new Date().toISOString()
+
+  return outputs
+    .filter((output) => output.filename && output.filename !== excludeFilename)
+    .map((output) => ({
+      id: `${promptId}-${output.nodeId}-${output.filename}`,
+      name: output.filename,
+      size: 0,
```
Myestery, line 71: is this intentional?

src/components/sidebar/tabs/AssetsSidebarListView.stories.ts
```diff
@@ -138,16 +140,33 @@ function renderAssetsSidebarListView(args: StoryArgs) {
     setup() {
       setMockJobItems(args.jobs)
       setMockJobActions(args.actionsByJobId ?? {})
+      const { assetItems, selectableAssets, isStackExpanded, toggleStack } =
+        useOutputStacks({
+          assets: computed(() => args.assets)
```
Myestery: is there real need for it to be computed here?

src/platform/assets/components/AssetsListItem.vue
```diff
@@ -74,13 +74,46 @@
...
+
+    <div
+      v-if="typeof stackCount === 'number' && stackCount > 1"
+      class="relative z-1 flex shrink-0 items-center"
+    >
+      <Button
+        variant="secondary"
+        size="md"
+        class="gap-1 font-bold"
+        :aria-label="stackIndicatorLabel || undefined"
```
Myestery:
```suggestion
        :aria-label="stackIndicatorLabel"
```
stackIndicatorLabel could be undefined already

src/platform/assets/composables/useAssetSelection.ts
```diff
@@ -21,6 +21,25 @@ export function useAssetSelection() {
   const metaKey = computed(() => isActive.value && metaKeyRaw.value)
   const cmdOrCtrlKey = computed(() => ctrlKey.value || metaKey.value)
 
+  function setAnchor(index: number, assetId: string | null) {
```
Myestery, line 24: How much of these can be computed properties so we dont have to keep reconciling in every mutation?

src/platform/assets/composables/useAssetSelection.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    const assets: AssetItem[] = [
+      { id: 'a', name: 'a.png', tags: [] },
+      { id: 'b', name: 'b.png', tags: [] }
+    ]
+
+    store.setSelection(['a', 'b'])
+    store.setLastSelectedIndex(0)
+    store.setLastSelectedAssetId('a')
+
+    selection.reconcileSelection([assets[1]])
```
Myestery: If this logic were exposed via a computed property, the test would be clearer and more representative of real usage. It would also avoid explicitly invoking the internal update function, keeping the test focused on observable behavior rather than implementation details.

src/platform/assets/composables/useOutputStacks.test.ts
```diff
@@ -0,0 +1,200 @@
...
+  let resolve!: (value: T) => void
+  let reject!: (reason?: unknown) => void
+  const promise = new Promise<T>((resolveFn, rejectFn) => {
+    resolve = resolveFn
+    reject = rejectFn
+  })
+  return { promise, resolve, reject }
+}
+
+function createAsset(overrides: Partial<AssetItem> = {}): AssetItem {
```
Myestery, line 36: Great mock pattern ✅

src/services/jobOutputCache.ts
```diff
@@ -75,6 +76,40 @@ export async function getOutputsForTask(
   }
 }
 
+function getPreviewableOutputs(outputs?: TaskOutput): ResultItemImpl[] {
+  if (!outputs) return []
+  const resultItems = Object.entries(outputs).flatMap(([nodeId, nodeOutputs]) =>
+    Object.entries(nodeOutputs)
+      .filter(([mediaType, items]) => mediaType !== 'animated' && items)
```
Myestery: we need a better condition to filter items or we leave it out of the filter and replace with

```suggestion
      .filter(([mediaType, _]) => mediaType !== 'animated')
```

Myestery:
@coderabbitai full review

src/services/jobOutputCache.ts
benceruleanlu: ResultItem is a type alias, it has no prototype for instanceof to check

benceruleanlu: it's like a very funny problem, because the schema itself is very forgiving so we can't really zod parse without modifying which we really can't, and it was also very poorly named given the original implementation, so I think I'll just tighten with a manual guard

benceruleanlu: I also renamed it to isResultItemLike for clarity

src/platform/assets/utils/outputAssetUtil.ts
benceruleanlu: the ID, previously yes it's to prevent collisions, but this is still susceptible to collisions across subfolders and nodes, so I just brought back the commit from the stacked PRs to this branch to fix that

src/components/sidebar/tabs/AssetsSidebarListView.stories.ts
benceruleanlu: for the storybook args in controls to be reactive yes, but I switched it to toref

src/platform/assets/components/AssetsListItem.vue
benceruleanlu: got it

src/platform/assets/composables/useAssetSelection.ts
benceruleanlu: These properties are already computed, did you mean to select another line?

src/platform/assets/composables/useAssetSelection.test.ts
benceruleanlu: I'm not too sure I understand, maybe the line selection was off? reconcileSelection is an imperative action that mutates the store, it can't really be a computed

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -79,18 +79,27 @@
...
             :secondary-text="getAssetSecondaryText(item.asset)"
+            :stack-count="getStackCount(item.asset)"
+            :stack-indicator-label="t('mediaAsset.actions.seeMoreOutputs')"
+            :stack-expanded="isStackExpanded(item.asset)"
             @mouseenter="onAssetEnter(item.asset.id)"
             @mouseleave="onAssetLeave(item.asset.id)"
             @contextmenu.prevent.stop="emit('context-menu', $event, item.asset)"
-            @click.stop="emit('select-asset', item.asset)"
+            @click.stop="emit('select-asset', item.asset, selectableAssets)"
+            @stack-toggle="toggleStack(item.asset)"
```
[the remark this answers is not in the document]
benceruleanlu: Done!

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -470,9 +486,10 @@ watch(
   { immediate: true }
 )
 
-const handleAssetSelect = (asset: AssetItem) => {
-  const index = displayAssets.value.findIndex((a) => a.id === asset.id)
-  handleAssetClick(asset, index, displayAssets.value)
+const handleAssetSelect = (asset: AssetItem, assets?: AssetItem[]) => {
+  const assetList = assets ?? visibleAssets.value
+  const index = assetList.findIndex((a) => a.id === asset.id)
+  handleAssetClick(asset, index, assetList)
```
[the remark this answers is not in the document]
benceruleanlu: Done!

src/services/jobOutputCache.ts
benceruleanlu: OK, removed the items check

src/platform/assets/utils/outputAssetUtil.ts
```diff
@@ -0,0 +1,109 @@
...
+export function getOutputKey({
+  nodeId,
+  subfolder,
+  filename
+}: OutputKeyParts): string | null {
+  if (!nodeId || !subfolder || !filename) {
+    return null
+  }
+
+  return `${nodeId}-${subfolder}-${filename}`
```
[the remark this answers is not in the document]
benceruleanlu: Done!

Myestery:
👀

src/platform/assets/utils/outputAssetUtil.ts
```diff
@@ -0,0 +1,109 @@
...
+  excludeOutputKey?: string
+}
+
+type ResolveOutputAssetItemsOptions = {
+  createdAt?: string
+  excludeOutputKey?: string
+}
+
+type OutputKeyParts = {
+  nodeId?: string | number | null
```
Myestery, line 24: we have a NodeId type we could use here

src/platform/assets/utils/outputAssetUtil.ts
```diff
@@ -0,0 +1,109 @@
...
+  return outputs.reduce<AssetItem[]>((items, output) => {
+    const outputKey = getOutputKey(output)
+    if (!output.filename || !outputKey || outputKey === excludeOutputKey) {
+      return items
+    }
+
+    items.push({
+      id: `${promptId}-${outputKey}`,
+      name: output.filename,
+      size: 0,
```
Myestery, line 71: I'm not sure why the size is hardcoded to 0 here

Myestery (approved):

src/platform/assets/utils/outputAssetUtil.ts
benceruleanlu: Good point, I'll add that in a followup PR

src/platform/assets/utils/outputAssetUtil.ts
benceruleanlu: It’s hardcoded because the output objects we map (ResultItemImpl) simply don’t carry file size, and the history/job APIs don’t provide it yet. AssetItem.size is optional with a “TBD…provided by history API in the future” note, so we’re effectively using 0 as “unknown” and to keep the shape consistent. Also, the UI only shows size if it’s truthy, so 0 suppresses display.

src/platform/assets/utils/outputAssetUtil.ts
benceruleanlu: https://github.com/Comfy-Org/ComfyUI_frontend/pull/8547

PR #8286, opened 2026-01-24
Templates: Search speed

DrJKL:
## Summary

...

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8286-Templates-Search-speed-2f26d73d365081dab3d1cccd7878a1de) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useTemplateFiltering.ts
+++ b/src/composables/useTemplateFiltering.ts
@@ -1,4 +1,4 @@
-import { refDebounced, watchDebounced } from '@vueuse/core'
+import { refThrottled, watchDebounced } from '@vueuse/core'
 import Fuse from 'fuse.js'
 import type { IFuseOptions } from 'fuse.js'
 import { computed, ref, watch } from 'vue'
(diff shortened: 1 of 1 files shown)


src/composables/useTemplateFiltering.ts
```diff
@@ -84,7 +84,7 @@ export function useTemplateFiltering(
     return ['ComfyUI', 'External or Remote API']
   })
 
-  const debouncedSearchQuery = refDebounced(searchQuery, 50)
+  const debouncedSearchQuery = refThrottled(searchQuery, 50)
```
DrJKL, line 87: Should I rename this? Yes.
Will I?
Only if asked.

AustinMroz: I'd want it left as is specifically to serve as a reminder that this is wrong and awful

AustinMroz (approved):
It's a massive improvement and I'm hitting approve just because it's a massive improvement.

In terms of "how search should work", I can confidently say it should not be like this. This is entirely at odds with my understanding of how performance should be done. That it *does* substantially improve performance leaves me perplexed.

---
Almost guaranteed unrelated: My computer OOM'd and killed the frontend browser tab while I was writing this review, seemingly suggesting 14GB in use by the tab. Will try to repro.

src/composables/useTemplateFiltering.ts
DrJKL: As you wish!

PR #8287, opened 2026-01-24
Style: Dark mode body on load alternative

DrJKL:
## Summary

Alternative to https://github.com/Comfy-Org/ComfyUI_frontend/pull/8077

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8287-Style-Dark-mode-body-on-load-alternative-2f26d73d365081fb8231f167e75beb83) by [Unito](https://www.unito.io)

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,7 +10,13 @@
     <meta name="mobile-web-app-capable" content="yes">
     <!-- Status bar style (eg. black or transparent) -->
     <meta name="apple-mobile-web-app-status-bar-style" content="black">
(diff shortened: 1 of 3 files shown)


christian-byrne (approved):

light-and-ray:
I see, maybe removing this color at all is a good idea. But do you have other patches, or is it everything?

I no longer see the background image in comfy UI after applying the changes on my local setup, because you have removed it too @DrJKL

PR #8302, opened 2026-01-25
fix: fallback to asset metadata/name when filename missing

DrJKL:
## Summary

Fix model node creation failing when `user_metadata.filename` is missing by falling back to `asset.metadata.filename` or `asset.name`.

## Changes

- Add fallback chain for filename: `userMetadata.filename || validAsset.metadata?.filename || validAsset.name`

## Testing

Manual testing with assets that have filename in different metadata locations.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8302-fix-fallback-to-asset-metadata-name-when-filename-missing-2f36d73d365081478299e2f2c1abde81) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/assets/utils/createModelNodeFromAsset.ts
+++ b/src/platform/assets/utils/createModelNodeFromAsset.ts
@@ -82,7 +82,8 @@ export function createModelNodeFromAsset(
     }
   }
 
-  const filename = userMetadata.filename
+  const filename =
+    userMetadata.filename || validAsset.metadata?.filename || validAsset.name
   if (typeof filename !== 'string' || filename.length === 0) {
     console.error(
       `Asset ${validAsset.id} has invalid user_metadata.filename (expected non-empty string, got ${typeof filename})`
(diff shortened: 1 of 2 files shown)


src/platform/assets/utils/createModelNodeFromAsset.ts
```diff
@@ -82,7 +82,8 @@ export function createModelNodeFromAsset(
     }
   }
 
-  const filename = userMetadata.filename
+  const filename =
+    userMetadata.filename || validAsset.metadata?.filename || validAsset.name
```
luke-mino-altherr, line 75: I'm not sure if userMetadata is required in lines above. To be safe can we do 
```
const userMetadata = validAsset.user_metadata ?? {}
```

luke-mino-altherr: not sure what else assume we will have user_metadata. That is only populated on the backend now after the user manually edits the model asset

luke-mino-altherr (approved):

christian-byrne (approved):

PR #8307, opened 2026-01-25
refactor: parallelize bootstrap and simplify lifecycle with VueUse

DrJKL:
## Summary

Refactors bootstrap and lifecycle management to parallelize initialization, use Vue best practices, and fix a logout state bug.

### Bootstrap Store (`bootstrapStore.ts`)
- Extract early bootstrap logic into a dedicated store using `useAsyncState`
- Parallelize settings, i18n, and workflow sync loading (previously sequential)
- Handle multi-user login scenarios by deferring settings/workflows until authenticated

### GraphCanvas Refactoring
- Move non-DOM composables (`useGlobalLitegraph`, `useCopy`, `usePaste`, etc.) to script setup level for earlier initialization
- Move `watch` and `whenever` declarations outside `onMounted` (Vue best practice)
- Use `until()` from VueUse to await bootstrap store readiness instead of direct async calls

### GraphView Simplification
- Replace manual `addEventListener`/`removeEventListener` with `useEventListener`
- Replace `setInterval` with `useIntervalFn` for automatic cleanup
- Move core command registration to script setup level

### Bug Fix
Using `router.push()` for logout caused `isSettingsReady` to persist as `true`, making new users inherit the previous user's cached settings. Reverted to `window.location.reload()` with TODOs for future store reset implementation.

## Testing

- Verified login/logout cycle clears settings correctly
- Verified bootstrap sequence completes without errors


christian-byrne (commented):
No multi layered DI containers and bootstrap.ts per distribution? 😭

src/stores/workspaceStore.ts
```diff
@@ -18,7 +19,7 @@ import { useSidebarTabStore } from './workspace/sidebarTabStore'
 
 export const useWorkspaceStore = defineStore('workspace', () => {
   const spinner = ref(false)
-  const shiftDown = ref(false)
+  const { shift: shiftDown } = useMagicKeys()
```
AustinMroz, line 22: My Hero!

This resolves #5162

src/components/graph/GraphCanvas.vue
```diff
@@ -404,35 +401,69 @@ useCanvasDrop(canvasRef)
...
+    if (!oldLocale) return
+    await until(() => isSettingsReady.value || !!settingsError.value).toBe(true)
+    await Promise.all([
+      until(() => isI18nReady.value || !!i18nError.value).toBe(true),
+      newUserService().initializeIfNewUser(settingStore)
+    ])
+    await useCommandStore().execute('Comfy.RefreshNodeDefinitions')
+    await useWorkflowService().reloadCurrentWorkflow()
+  }
+)
```
AustinMroz, line 429: Worth adding an `onCleanup` here?

There's some race condition potential here, but I think the chances of it occurring are near nonexistant.

src/composables/node/useNodeBadge.ts
```diff
@@ -71,6 +71,9 @@ export const useNodeBadge = () => {
   }
 
   onMounted(() => {
+    if (extensionStore.isExtensionInstalled('Comfy.NodeBadge')) return
```
AustinMroz, line 74: :+1:

AustinMroz (dismissed):
Lots of long standing issues cleaned up as a side effect of this PR.

src/platform/settings/settingStore.ts
```diff
@@ -170,7 +170,11 @@ export const useSettingStore = defineStore('setting', () => {
       throw new Error('Settings must have an ID')
     }
     if (setting.id in settingsById.value) {
-      throw new Error(`Setting ${setting.id} must have a unique ID.`)
+      // Setting already registered - skip to allow component remounting
+      // TODO: Add store reset methods to bootstrapStore and settingStore, then
+      // replace window.location.reload() with router.push() in SidebarLogoutIcon.vue
+      console.warn(`Setting already registered: ${setting.id}`)
+      return
```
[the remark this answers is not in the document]
DrJKL: Nope.

src/stores/bootstrapStore.ts
```diff
@@ -0,0 +1,112 @@
...
+    },
+    undefined,
+    { immediate: false }
+  )
+
+  function syncWorkflows() {
+    if (!isWorkflowsReady.value && !isWorkflowsLoading.value) {
+      void executeSyncWorkflows()
+    }
+  }
```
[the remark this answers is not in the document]
DrJKL: Don't suggest exposing internals that aren't used. Only useful properties should be exposed (as demonstrated by their active use).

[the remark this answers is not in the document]
DrJKL: @coderabbitai I was talking to you... which says something about me, I'm sure...

@coderabbitai Don't suggest exposing internals that aren't used. Only useful properties should be exposed (as demonstrated by their active use).

src/components/graph/GraphCanvas.vue
DrJKL: Possibly, but I'm not sure what I'd be cleaning up yet.

AustinMroz (dismissed):

src/stores/bootstrapStore.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  }))
+}))
+
+vi.mock('@/stores/workspaceStore', () => ({
+  useWorkspaceStore: vi.fn(() => ({
+    workflow: {
+      syncWorkflows: vi.fn().mockResolvedValue(undefined)
+    }
+  }))
+}))
```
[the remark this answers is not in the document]
DrJKL: Maybe later.

src/components/graph/GraphCanvas.vue
```diff
@@ -404,35 +402,72 @@ useCanvasDrop(canvasRef)
...
   // ChangeTracker needs to be initialized before setup, as it will overwrite
   // some listeners of litegraph canvas.
   ChangeTracker.init()
-  await loadCustomNodesI18n()
-  try {
-    await settingStore.loadSettingValues()
-  } catch (error) {
-    if (error instanceof UnauthorizedError) {
+
+  await until(() => isSettingsReady.value || !!settingsError.value).toBe(true)
```
christian-byrne, line 445: Should this logic/nuance be abstracted and exported from the store? So it's used like `await untilSettingsReady()`? To be fair it's already quite expressive.

src/components/graph/GraphCanvas.vue
```diff
@@ -404,35 +402,72 @@ useCanvasDrop(canvasRef)
...
+  await Promise.all([
+    until(() => isI18nReady.value || !!i18nError.value).toBe(true),
+    newUserService().initializeIfNewUser(settingStore)
+  ])
+  if (i18nError.value) {
+    console.warn(
+      '[GraphCanvas] Failed to load custom nodes i18n:',
+      i18nError.value
+    )
+  }
```
christian-byrne, line 471: Does this make more sense to happen in bootstrap store?

src/platform/settings/settingStore.ts
```diff
@@ -170,7 +170,11 @@ export const useSettingStore = defineStore('setting', () => {
       throw new Error('Settings must have an ID')
     }
     if (setting.id in settingsById.value) {
-      throw new Error(`Setting ${setting.id} must have a unique ID.`)
+      // Setting already registered - skip to allow component remounting
+      // TODO: Add store reset methods to bootstrapStore and settingStore, then
+      // replace window.location.reload() with router.push() in SidebarLogoutIcon.vue
```
christian-byrne, line 210: `router.push()` on logout won't invalidate all the app state which would cause issues when changing accounts.

src/stores/bootstrapStore.ts
```diff
@@ -0,0 +1,112 @@
...
+      await useSettingStore().loadSettingValues()
+    },
+    undefined,
+    { immediate: false }
+  )
+
+  function loadSettings() {
+    // TODO: This check makes the store "sticky" across logouts. Add a reset
+    // method to clear isSettingsReady, then replace window.location.reload()
+    // with router.push() in SidebarLogoutIcon.vue
```
christian-byrne: Same comment as above.

src/views/GraphView.vue
```diff
@@ -198,15 +194,12 @@ watchEffect(() => {
...
-  useSidebarTabStore().registerCoreSidebarTabs()
-  useBottomPanelStore().registerCoreBottomPanelTabs()
-  app.extensionManager = useWorkspaceStore()
-}
+const coreCommands = useCoreCommands()
+useCommandStore().registerCommands(coreCommands)
+useMenuItemStore().registerCoreMenuCommands()
+useKeybindingService().registerCoreKeybindings()
+useSidebarTabStore().registerCoreSidebarTabs()
+useBottomPanelStore().registerCoreBottomPanelTabs()
```
christian-byrne, line 202: question: Can this happen in the bootstrap domain and not in view file? Or that would require a lot more changes?

src/App.vue
```diff
@@ -21,15 +20,13 @@ import { useWorkspaceStore } from '@/stores/workspaceStore'
 import { useConflictDetection } from '@/workbench/extensions/manager/composables/useConflictDetection'
 
 import { electronAPI, isElectron } from './utils/envUtil'
+import { app } from '@/scripts/app'
 
 const workspaceStore = useWorkspaceStore()
+app.extensionManager = useWorkspaceStore()
```
christian-byrne, line 26: note: I suppose the definite type assertion on this property will now be actually correct, or closer to it.

https://github.com/Comfy-Org/ComfyUI_frontend/blob/b551064a6a8ef9795144f25c1dab2bd9c72c72c1/src/scripts/app.ts#L157

src/views/GraphView.vue
```diff
@@ -291,27 +284,7 @@ onMounted(() => {
...
-
-  // Clean up tab count tracking
-  if (tabCountInterval) {
-    window.clearInterval(tabCountInterval)
-    tabCountInterval = null
-  }
-  if (tabCountChannel) {
-    tabCountChannel.close()
-    tabCountChannel = null
-  }
```
christian-byrne, line 314: nit: We can honestly remove all the heartbeat and tab count tracking code and the types in the telemetry modules, as the events are disabled and too expensive to track anyway.

src/views/GraphView.vue
```diff
@@ -337,18 +310,17 @@ const onGraphReady = () => {
...
     // Set up page visibility tracking (cloud only)
-    if (isCloud && telemetry && !visibilityListener) {
-      visibilityListener = () => {
+    if (isCloud && telemetry) {
+      useEventListener(document, 'visibilitychange', () => {
         telemetry.trackPageVisibilityChanged({
           visibility_state: document.visibilityState as 'visible' | 'hidden'
         })
-      }
-      document.addEventListener('visibilitychange', visibilityListener)
```
christian-byrne, line 346: Seems this won't track page focus events anymore (besides the first)?

src/components/graph/GraphCanvas.vue
DrJKL: Yes. I considered doing that for each of the lazily initialized stores instead of keeping the logic in bootstrap. I'll try it out.

src/components/graph/GraphCanvas.vue
DrJKL: It should, but first we have to fix the login/logout to not be dependent on Views with distributed persistent state.
I had a version that used the router instead of reload but that had issues with our Stores, thus the user based initialization happening here.

src/platform/settings/settingStore.ts
DrJKL: Oh, yeah, this. (My comment above)

src/stores/bootstrapStore.ts
DrJKL: Getting the user specific logic and state isolated is going to be its own whole project.

src/views/GraphView.vue
DrJKL: It would require more changes, but that is where I'd like things to go.

src/App.vue
DrJKL: Closer...

src/views/GraphView.vue
DrJKL: It might be the opposite, actually...
I'll check

src/components/graph/GraphCanvas.vue
DrJKL: Just lacking an analogous store for i18n pieces.

src/platform/settings/settingStore.ts
```diff
@@ -242,8 +260,11 @@ export const useSettingStore = defineStore('setting', () => {
   return {
     settingValues,
     settingsById,
+    isReady,
+    isLoading,
+    error,
+    load,
```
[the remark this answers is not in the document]
DrJKL: Lies.

AustinMroz (dismissed):

src/platform/settings/settingStore.ts
christian-byrne: Should we add this aspect to the comment/todo? Or make an Issue?

christian-byrne (approved):
LGTM

src/platform/settings/settingStore.ts
DrJKL: @coderabbitai Open a tracking issue to fix the state such that we can use the router to login/logout

PR #8312, opened 2026-01-25
Allow augmenting input spec by property

AustinMroz:
In order for GLSL nodes to be configurable by subgraph, it's desirable that there be a way to provide a dynamicCombo adjacent experience which is defined in workflow, similar to the the custom numbers or custom combo.

This PR adds support for a special `dynamicSpec` property on nodes which, if present and safely parsable to an inputSpecV2, will add a corresponding input to the node.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8312-Allow-augmenting-input-spec-by-property-2f36d73d36508143bc32fc57f5190cfe) by [Unito](https://www.unito.io)

### DIFF
--- a/src/schemas/nodeDef/nodeDefSchemaV2.ts
+++ b/src/schemas/nodeDef/nodeDefSchemaV2.ts
@@ -132,7 +132,7 @@ const zCustomInputSpec = zBaseInputOptions.extend({
   isOptional: z.boolean().optional()
 })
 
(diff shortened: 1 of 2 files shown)


AustinMroz:
Functionality here has a much larger surface area than requested and is likely to cause problems down the line. Closing this one as the proposed functionality isn't desired.

PR #8318, opened 2026-01-26
fix: resolve no-misused-spread lint warnings in test files

Myestery:
## Summary

Replace spread operators with Object.assign() to fix 3 no-misused-spread lint warnings in test utilities and test files.

## Changes

- **What**: Replaced spread operators with Object.assign() in createMockFileList and mock node creation functions to avoid spreading arrays into objects and class instances that could lose prototypes. Simplified BypassButton.test.ts by removing redundant type annotations.
- **Breaking**: None

## Review Focus

Type safety is preserved without using weak TypeScript patterns (no `any`, `as unknown as`, or unnecessary casts).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8318-fix-resolve-no-misused-spread-lint-warnings-in-test-files-2f46d73d365081aca3f6cf208556d492) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphHierarchy.test.ts
+++ b/src/composables/graph/useGraphHierarchy.test.ts
@@ -14,11 +14,13 @@ import { useGraphHierarchy } from './useGraphHierarchy'
 vi.mock('@/renderer/core/canvas/canvasStore')
 
 function createMockNode(overrides: Partial<LGraphNode> = {}): LGraphNode {
-  return {
-    ...createMockLGraphNode(),
-    boundingRect: new Rectangle(100, 100, 50, 50),
-    ...overrides
-  } as LGraphNode
+  return Object.assign(
+    createMockLGraphNode(),
+    {
+      boundingRect: new Rectangle(100, 100, 50, 50)
+    },
+    overrides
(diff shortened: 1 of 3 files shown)


src/composables/graph/useGraphHierarchy.test.ts
```diff
@@ -14,11 +14,13 @@ import { useGraphHierarchy } from './useGraphHierarchy'
...
-    boundingRect: new Rectangle(100, 100, 50, 50),
-    ...overrides
-  } as LGraphNode
+  return Object.assign(
+    createMockLGraphNode(),
+    {
+      boundingRect: new Rectangle(100, 100, 50, 50)
+    },
+    overrides
+  ) as LGraphNode
```
DrJKL: Would this let you remove the type assertion?
```suggestion
  )
```

DrJKL (dismissed):

src/composables/graph/useGraphHierarchy.test.ts
Myestery: Good catch

PR #8335, opened 2026-01-27
Fix: Implement scope-aware filtering for template modal

Myestery:
## Summary
- Fixes template filtering to be scope-aware, preventing empty results when switching categories
- Filters now only show as selected when they exist in the current category scope
- User selections are still persisted in localStorage for convenience

## Problem
When users selected filters (e.g., Image-specific filters) and switched to a different category (e.g., Video), the persisted filters would cause empty results because those filter values didn't exist in the new scope.

## Solution
- Modified `useTemplateFiltering` composable to track which filters are "active" (applicable to current scope)
- Only active filters are shown as selected in the UI
- Only active filters are applied to the filtering logic
- All user selections are still persisted in localStorage
- When returning to a category, previously selected filters automatically reactivate if applicable

## Test Plan
Tested manually in browser:
1. Navigate to Image category
2. Select filters like "Image" and "Image Edit" from Tasks dropdown
3. Switch to Video category - filters show as "0 selected" (Image/Image Edit don't exist in Video)
4. Return to Image category - filters are automatically reselected (2 selected)
5. No empty results when switching between categories

## Changes
- `useTemplateFiltering.ts`: Added scope-aware filtering logic with active/inactive filter distinction
- `WorkflowTemplateSelectorDialog.vue`: Pass navigation context and use active filters for UI
- `useTemplateFiltering.test.ts`: Added comprehensive tests for scope-aware behavior

https://github.com/user-attachments/assets/706ddabf-abad-4ff9-a378-6603d275d15a

### DIFF
--- a/src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
+++ b/src/components/custom/widget/WorkflowTemplateSelectorDialog.vue
@@ -547,13 +547,15 @@ const navigationFilteredTemplates = computed(() => {
   return workflowTemplatesStore.filterTemplatesByCategory(selectedNavItem.value)
 })
 
-// Template filtering
+// Template filtering with scope awareness
 const {
   searchQuery,
   selectedModels,
   selectedUseCases,
   selectedRunsOn,
   sortBy,
+  activeModels,
+  activeUseCases,
   filteredTemplates,
   availableModels,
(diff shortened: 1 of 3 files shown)


viva-jinyi (approved):
LGTM

PR #8342, opened 2026-01-27
Prevent configuring a node to a placeholder nodeId

AustinMroz:
Litegraph uses `-1 ` as a placeholder node id to indicate that a node should be assigned a new node id when added to the graph. Under some unknown circumstances it's possible for a node to have this placeholder id saved in a workflow file. 

When this occurs, the node is loaded and assigned a new id, but then configured back to the placeholder id. This PR makes it so the newly assigned id is kept instead.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8342-Prevent-configuring-a-node-to-a-placehodler-nodeId-2f56d73d365081ed8217e989187b15c8) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphNode.ts
+++ b/src/lib/litegraph/src/LGraphNode.ts
@@ -785,7 +785,7 @@ export class LGraphNode
     if (this.graph) {
       this.graph._version++
     }
-    if (info.id == -1) info.id = this.id
+    if (info.id === -1) info.id = this.id
     for (const j in info) {
       if (j == 'properties') {
         // i don't want to clone properties, I want to reuse the old container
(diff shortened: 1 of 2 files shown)


src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -785,6 +785,7 @@ export class LGraphNode
     if (this.graph) {
       this.graph._version++
     }
+    if (info.id == -1) info.id = this.id
```
DrJKL: Nit:
```suggestion
    if (info.id === -1) info.id = this.id
```

DrJKL (dismissed):

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: Loose equality was intentional. NodeId can be a number or a string and an id of `'-1'` would be equally problematic.

Would appreciate a second opinion though. `LGraph.add()` does use strict equality to `-1` when a node is added to the graph :thinking:

DrJKL (approved):

Issue #8134, opened 2026-01-17
Template search input has performance issue


fappaz:
### Prerequisites

- [x] I am running the latest version of ComfyUI
- [x] I have searched existing issues to make sure this isn't a duplicate
- [x] I have tested with all custom nodes disabled ([see how](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled))

### What happened?

When searching for templates, the search box is extremely slow, freezing while you type.

In the video below, I typed "z-image turbo" but it took a few seconds for the input box to receive all the characters and display the filtered results. Later on, I started deleting the characters, but it also took a while to process the keyboard events.

https://github.com/user-attachments/assets/cb801523-5b34-49ec-b158-8e71f0b07b52

### Steps to Reproduce

1. Open the Templates dialog
2. Start typing in the search box
3. Notice how it freezes for a while before it shows the typed filter

### How is this affecting you?

Visual/UI issue only

### ComfyUI Frontend Version

1.38.3

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

```
ComfyUI version: 0.9.2
ComfyUI frontend version: 1.38.3
OS: win32
Python Version: 3.11.6 (tags/v3.11.6:8b6ee5b, Oct 2 2023, 14:57:12) [MSC v.1935 64 bit (AMD64)]
Embedded Python: true
Pytorch Version: 2.9.0+cu128
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-8134-Template-search-input-has-performance-issue-2eb6d73d36508191b469ef00a77248f9) by [Unito](https://www.unito.io)

fappaz:
Just a note, I updated the frontend to version 1.38.4 released just a few hours ago, but the problem persists.

christian-byrne:
Fixed by PR #8343 (merged Jan 30). The search input now uses 300ms debouncing instead of 50ms throttling.

If you still experience performance issues on the latest version, please reopen.

PR #8343, opened 2026-01-27
Improve template search input performance issue 

kaili-yang:
## Summary

Improve Template search input performance issue #8134
This was caused by the search logic running too frequently (throttled at 50ms), causing the main thread to block on every few keystrokes.

## Changes
Use debouncing that wait until you stop typing for a specific time (300ms) before running.
It makes the searching function more smoothly.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8343-Improve-template-search-input-performance-issue-2f56d73d36508144bdf9fa5e0cd76818) by [Unito](https://www.unito.io)


DrJKL:
@kaili-yang That was the opposite of my experience 🙃 

When I tested it the search resolved a lot faster using throttled. @AustinMroz ?

AustinMroz:
Conceptually, I think the template search should use be a debounce and have a higher delay (100-150).

But 50ms throttled feels much better than 50ms debounced and appears to be better than the proposed 300ms debounced of this PR, too.

I went ahead and did some digging myself since the circumstances were frustrating to me. `WorkflowTemplateSelectorDialog` is watching and resetting pagination on many things it should not be, including the raw searchQuery.  Since pagination is computed from `filteredTemplates`, I'm unsure if the watch or `resetPagination` call are needed at all.

kaili-yang:
The problem might be`watch searchQuery` in the `src/components/custom/widget/WorkflowTemplateSelectorDialog.vue`.

Every time the user presses a key, immediately **hard-reset** the list (templateListKey++) and reset pagination.
This invalidates the entire purpose of Debouncing/Throttling. Even if your search logic was delayed, your UI re-render logic was instant.

I suggest to try the following approach. Logically, this ensures the expensive re-render happens only once (after the 300ms delay), rather than 10 times while you type "hello world".

image

kaili-yang:
The performance issue was NOT the search algo, but the UI Watcher on `searchQuery`. It forced the entire list to re-render on every single keystroke, causing the freeze.
With that bug present, throttle at least updated the results during the freeze, creating an illusion of responsiveness. debounce just froze with old data.
Now that I've removed `searchQuery` from the watcher.`Debounce` is definitely better than 'throttled'. It ensures zero re-renders while typing, giving the smoothest possible search experience.

AustinMroz (approved):

PR #8359, opened 2026-01-28
feat: support dev-only nodes

guill:
## Summary

Support `dev_only` property to node definitions that hides nodes from search and menus unless dev mode is enabled. Dev-only nodes display a "DEV" badge when visible.

This functionality is primarily intended to support unit-testing nodes on Comfy Cloud, but also has other uses.

## Changes

- **What**: Nodes flagged as dev_only in the node schema will only appear in search and menus if Dev Mode is on.

With Dev Mode off:
image

With Dev Mode on:
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8359-feat-support-dev-only-nodes-2f66d73d36508102839ee7cd66a26129) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -261,20 +261,6 @@ watchEffect(() => {
 })
 
 watchEffect(() => {
-  const devModeEnabled = settingStore.get('Comfy.DevMode')
-  nodeDefStore.showDevOnly = devModeEnabled
-
-  // Update skip_list on all registered node types for dev-only nodes
-  // This ensures LiteGraph's getNodeTypesCategories/getNodeTypesInCategory
-  // correctly filter dev-only nodes from the right-click context menu
-  for (const nodeType of Object.values(LiteGraph.registered_node_types)) {
-    if (nodeType.nodeData?.dev_only) {
-      nodeType.skip_list = !devModeEnabled
-    }
-  }
-})
-
-watchEffect(() => {
   const spellcheckEnabled = settingStore.get('Comfy.TextareaWidget.Spellcheck')
   const textareas = document.querySelectorAll<HTMLTextAreaElement>(
     'textarea.comfy-multiline-input'
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -261,7 +261,17 @@ watchEffect(() => {
 })
 
 watchEffect(() => {
-  nodeDefStore.showDevOnly = settingStore.get('Comfy.DevMode')
+  const devModeEnabled = settingStore.get('Comfy.DevMode')
+  nodeDefStore.showDevOnly = devModeEnabled
+
+  // Update skip_list on all registered node types for dev-only nodes
+  // This ensures LiteGraph's getNodeTypesCategories/getNodeTypesInCategory
+  // correctly filter dev-only nodes from the right-click context menu
+  for (const nodeType of Object.values(LiteGraph.registered_node_types)) {
+    if (nodeType.nodeData?.dev_only) {
+      nodeType.skip_list = !devModeEnabled
+    }
+  }
 })
 
 watchEffect(() => {
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -261,6 +261,10 @@ watchEffect(() => {
 })
 
 watchEffect(() => {
+  nodeDefStore.showDevOnly = settingStore.get('Comfy.DevMode')
+})
+
+watchEffect(() => {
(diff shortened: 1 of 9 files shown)


src/components/graph/GraphCanvas.vue
```diff
@@ -260,6 +260,20 @@ watchEffect(() => {
   )
 })
 
+watchEffect(() => {
+  const devModeEnabled = settingStore.get('Comfy.DevMode')
+  nodeDefStore.showDevOnly = devModeEnabled
```
DrJKL: Nit: I would favor putting this directly into the nodeDefStore as a computedRef. The pattern of reading from one store and mutating another within components is one I'm working on eliminating.

src/components/graph/GraphCanvas.vue
```diff
@@ -260,6 +260,20 @@ watchEffect(() => {
...
+
+  // Update skip_list on all registered node types for dev-only nodes
+  // This ensures LiteGraph's getNodeTypesCategories/getNodeTypesInCategory
+  // correctly filter dev-only nodes from the right-click context menu
+  for (const nodeType of Object.values(LiteGraph.registered_node_types)) {
+    if (nodeType.nodeData?.dev_only) {
+      nodeType.skip_list = !devModeEnabled
+    }
+  }
+})
```
DrJKL: Nit: Same for this, if possible I would rather have this as a computed property somewhere (probably also the NodeDefStore) so we can avoid mutating the definitions.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -233,6 +233,12 @@ export class LGraphNode
   static description?: string
   static filter?: string
   static skip_list?: boolean
+  static nodeData?: {
+    dev_only?: boolean
+    deprecated?: boolean
+    experimental?: boolean
+    [key: string]: unknown
```
DrJKL: Do we need the index type? If not I'd rather limit the properties to known current keys.
```suggestion
```

src/components/graph/GraphCanvas.vue
DrJKL: Could `skip_list` be a getter?

src/stores/nodeDefStore.ts
```diff
@@ -174,6 +176,7 @@ export class ComfyNodeDefImpl
   get nodeLifeCycleBadgeText(): string {
     if (this.deprecated) return '[DEPR]'
     if (this.experimental) return '[BETA]'
+    if (this.dev_only) return '[DEV]'
     return ''
```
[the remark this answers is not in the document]
guill: This badge is consistent with all other badges.

src/stores/nodeDefStore.ts
```diff
@@ -422,6 +426,14 @@ export const useNodeDefStore = defineStore('nodeDef', () => {
       predicate: (nodeDef) => showExperimental.value || !nodeDef.experimental
     })
 
+    // Dev-only nodes filter
+    registerNodeDefFilter({
+      id: 'core.dev_only',
+      name: 'Hide Dev-Only Nodes',
+      description: 'Hides nodes marked as dev-only unless dev mode is enabled',
+      predicate: (nodeDef) => showDevOnly.value || !nodeDef.dev_only
+    })
```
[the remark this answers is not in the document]
guill: The unlocalized name and description are consistent with other filters.

src/components/graph/GraphCanvas.vue
guill: So it looks like actually turning this into a computed property or a getter would be non-trivial. It would require either updating Litegraph to import and depend on Vue code (which could easily result in circular dependencies) or an injection of a callback function.
I did at least move this watcher to the NodeDefStore where it makes more sense. Do you think it's worth actually doing the effort to move it into the Litegraph code itself?

src/lib/litegraph/src/LGraphNode.ts
guill: I removed this, but had to update one pre-existing test to avoid a TypeScript error. Please let me know if it looks reasonable!

src/stores/nodeDefStore.ts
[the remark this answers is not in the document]
DrJKL: @coderabbitai Open an issue for this, do not assign anyone.

src/components/graph/GraphCanvas.vue
DrJKL: Probably not :-)

src/lib/litegraph/src/LGraphNode.ts
DrJKL: The new enumeration looks good

DrJKL (approved):

PR #8363, opened 2026-01-28
Experiment: E2E after i18n (minimal)

DrJKL:
## Summary

See if all we need is the trigger.
Also cleans up the ignore list and adds the ability to manually trigger the checks.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8363-Experiment-E2E-after-i18n-minimal-2f66d73d365081f0b3abc10b15669f1f) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/ci-tests-e2e.yaml
+++ b/.github/workflows/ci-tests-e2e.yaml
@@ -5,7 +5,7 @@ on:
   push:
     branches: [main, master, core/*, desktop/*]
   pull_request:
-    branches-ignore: [wip/*, draft/*, temp/*]
+    branches-ignore: [wip/*, draft/*, temp/*, version-bump-*]
   workflow_dispatch:
   workflow_run:
     workflows: ['i18n: Update Core']
--- a/.github/workflows/ci-tests-e2e.yaml
+++ b/.github/workflows/ci-tests-e2e.yaml
@@ -5,8 +5,11 @@ on:
   push:
     branches: [main, master, core/*, desktop/*]
   pull_request:
-    branches-ignore:
-      [wip/*, draft/*, temp/*, vue-nodes-migration, sno-playwright-*]
+    branches-ignore: [wip/*, draft/*, temp/*]
+  workflow_dispatch:
+  workflow_run:
+    workflows: ['i18n: Update Core']
+    types: [completed]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}


.github/workflows/ci-tests-e2e.yaml
```diff
@@ -5,8 +5,11 @@ on:
   push:
     branches: [main, master, core/*, desktop/*]
   pull_request:
-    branches-ignore:
-      [wip/*, draft/*, temp/*, vue-nodes-migration, sno-playwright-*]
+    branches-ignore: [wip/*, draft/*, temp/*]
```
snomiao: maybe also ignores version-bump* ?


*Currently we have only "version-bump*" branches runs "i18n: Update Core"

.github/workflows/ci-tests-e2e.yaml
```diff
@@ -5,8 +5,11 @@ on:
...
+  workflow_dispatch:
+  workflow_run:
```
snomiao, line 10: and when its running by workflow_run event, allow only version-bump* to prevent duplicate runs

snomiao (changes requested):

.github/workflows/ci-tests-e2e.yaml
DrJKL: That's what made me think we could omit it here.

.github/workflows/ci-tests-e2e.yaml
DrJKL: Chatted about this offline 🙂

.github/workflows/ci-tests-e2e.yaml
DrJKL: Added it to the ignore list 🫡

.github/workflows/ci-tests-e2e.yaml
```diff
@@ -5,8 +5,11 @@ on:
   push:
     branches: [main, master, core/*, desktop/*]
   pull_request:
-    branches-ignore:
-      [wip/*, draft/*, temp/*, vue-nodes-migration, sno-playwright-*]
+    branches-ignore: [wip/*, draft/*, temp/*, version-bump-*]
+  workflow_dispatch:
+  workflow_run:
+    workflows: ['i18n: Update Core']
+    types: [completed]
```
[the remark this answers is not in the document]
snomiao: sg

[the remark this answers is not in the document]
DrJKL: We'd want the Playwright signal regardless, right?

AustinMroz (approved):

PR #8364, opened 2026-01-28
[feat] Add node replacement store and types

viva-jinyi:
## Summary
Add infrastructure for automatic node replacement feature that allows missing/deprecated nodes to be replaced with their newer equivalents.

## Changes
- **Types**: `NodeReplacement`, `NodeReplacementResponse` types matching backend API spec (PR #12014)
- **Service**: `fetchNodeReplacements()` API wrapper
- **Store**: `useNodeReplacementStore` with `getReplacementFor()`, `hasReplacement()`, `isEnabled()`
- **Setting**: `Comfy.NodeReplacement.Enabled` toggle (experimental)
- **Tests**: 11 unit tests covering store functionality

## Related
- Backend PR: Comfy-Org/ComfyUI#12014

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8364-feat-Add-node-replacement-store-and-types-2f66d73d3650816bb771c9cc6a8e1774) by [Unito](https://www.unito.io)

## Summary by CodeRabbit

* **New Features**
  * Automatic node replacement is now available, allowing missing nodes to be replaced with their newer equivalents when replacement mappings exist.
  * Added "Enable automatic node replacement" experimental setting in Workflow preferences (enabled by default).
  * Replacement data is loaded during app initialization.


christian-byrne (changes requested):
The current structure scatters files across generic folders:
```
src/services/nodeReplacementService.ts
src/stores/nodeReplacementStore.ts
src/types/nodeReplacementTypes.ts
```

Recommended: Colocate under `src/platform/nodeReplacement/` (following the pattern of existing small domains like `settings/` and `telemetry/`):

```
src/platform/nodeReplacement/
├── nodeReplacementService.ts       # API fetching
├── nodeReplacementStore.ts         # Pinia store
├── nodeReplacementStore.test.ts    # Tests colocated
└── types.ts                        # Domain types
```

Reasons:

1. **Matches existing patterns** - `src/platform/settings/` and `src/platform/telemetry/` use flat structures for small domains
2. **Bounded context** - All node replacement concerns grouped together
3. **Discoverability** - Easier to find related code
4. **No over-engineering** - Flat files at domain root; only nest (`components/`, `composables/`) when complexity warrants it

src/platform/nodeReplacement/nodeReplacementStore.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  function getReplacementFor(nodeType: string): NodeReplacement | null {
+    if (!isEnabled()) return null
+    return replacements.value[nodeType]?.[0] ?? null
+  }
+
+  function hasReplacement(nodeType: string): boolean {
+    if (!isEnabled()) return false
+    return !!replacements.value[nodeType]?.length
+  }
```
christian-byrne: You can probably make a computed value for `isEnabled` (which will update whenever the backing setting changes). Then the function calls below just check the computed `.value`. This makes it push-based and avoids us having to instantiate the settings store everytime we call the functions

christian-byrne (dismissed):

src/platform/nodeReplacement/nodeReplacementStore.test.ts
```diff
@@ -0,0 +1,240 @@
...
+      }
+
+      expect(store.hasReplacement('OldNode')).toBe(true)
+    })
+
+    it('should return false when node type does not exist', () => {
+      store.replacements = {}
+
+      expect(store.hasReplacement('NonExistentNode')).toBe(false)
+    })
```
christian-byrne, line 144: This is not really testing the case where the node type does not exist -- but rather if the node type doesn't exist in the replacement list right?

I think to test the described behavior ('should return false when node type does not exist'), the replacement store would have to be interacting the the nodeDefStore and we would have to be mocking that to some extent maybe.

christian-byrne (approved):
LGTM with nit on a test.

src/platform/nodeReplacement/nodeReplacementStore.test.ts
viva-jinyi: I will refactor it on other branch! thx

PR #8365, opened 2026-01-28
feat: add provider logo overlays to workflow template thumbnails

christian-byrne:
## Summary

Add support for overlaying provider logos on workflow template thumbnails at runtime.

## Changes

- **What**: 
  - Add `LogoInfo` interface and `logos` field to `TemplateInfo` type
  - Create `LogoOverlay.vue` component for rendering positioned logos
  - Fetch logo index from `templates/index_logo.json` in store
  - Add `getLogoUrl` helper to `useTemplateWorkflows` composable
  - Integrate `LogoOverlay` into `WorkflowTemplateSelectorDialog`

## Review Focus

- Logo positioning uses Tailwind classes (e.g. `absolute bottom-2 right-2`)
- Supports multiple logos per template with configurable size/opacity
- Gracefully handles missing logos (returns empty string, renders nothing)
- Templates must explicitly declare logos - no magic inference from models

## Dependencies

Requires separate PR in workflow_templates repo to:
1. Update `index.schema.json` with logos definition
2. Add `logos` field to templates in `index.json`

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8365-feat-add-provider-logo-overlays-to-workflow-template-thumbnails-2f66d73d365081309236c6b991cb6f7b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/workflow/templates/composables/useTemplateWorkflows.ts
+++ b/src/platform/workflow/templates/composables/useTemplateWorkflows.ts
@@ -158,6 +158,7 @@ export function useTemplateWorkflows() {
    */
   const fetchTemplateJson = async (id: string, sourceModule: string) => {
     if (sourceModule === 'default') {
+      // Default templates provided by frontend are served on this separate endpoint
       return fetch(api.fileURL(`/templates/${id}.json`)).then((r) => r.json())
     } else {
       return fetch(
--- a/src/platform/workflow/templates/composables/useTemplateWorkflows.ts
+++ b/src/platform/workflow/templates/composables/useTemplateWorkflows.ts
@@ -182,7 +182,6 @@ export function useTemplateWorkflows() {
     getTemplateThumbnailUrl,
     getTemplateTitle,
     getTemplateDescription,
-    loadWorkflowTemplate,
-    getLogoUrl: workflowTemplatesStore.getLogoUrl
+    loadWorkflowTemplate
   }
 }
--- a/src/platform/workflow/templates/composables/useTemplateWorkflows.ts
+++ b/src/platform/workflow/templates/composables/useTemplateWorkflows.ts
@@ -158,13 +158,19 @@ export function useTemplateWorkflows() {
    */
   const fetchTemplateJson = async (id: string, sourceModule: string) => {
     if (sourceModule === 'default') {
-      // Default templates provided by frontend are served on this separate endpoint
       return fetch(api.fileURL(`/templates/${id}.json`)).then((r) => r.json())
     } else {
       return fetch(
         api.apiURL(`/workflow_templates/${sourceModule}/${id}.json`)
       ).then((r) => r.json())
     }
+  }
+
+  /**
+   * Gets logo URL for a provider name
+   */
+  const getLogoUrl = (provider: string): string => {
+    return workflowTemplatesStore.getLogoUrl(provider)
   }
 
   return {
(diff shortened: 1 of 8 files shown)


src/platform/workflow/templates/composables/useTemplateWorkflows.ts
```diff
@@ -167,6 +166,13 @@ export function useTemplateWorkflows() {
     }
   }
 
+  /**
+   * Gets logo URL for a provider name
+   */
+  const getLogoUrl = (provider: string): string => {
+    return workflowTemplatesStore.getLogoUrl(provider)
+  }
```
christian-byrne: Is it necessary to just alias this method here?

src/platform/workflow/templates/repositories/workflowTemplatesStore.ts
```diff
@@ -498,6 +503,22 @@ export const useWorkflowTemplatesStore = defineStore(
...
 
+    async function fetchLogoIndex(): Promise<LogoIndex> {
+      try {
+        const response = await fetch(api.fileURL('/templates/index_logo.json'))
+        if (!response.ok) return {}
+        return await response.json()
+      } catch {
+        return {}
+      }
+    }
```
christian-byrne, line 518: We should make this better align with other ways that fetching is done (in even this module alone) -- putting the actual call in api.ts such that headers can be applied

comfyui-wiki:
Seems the size currently isn't working, but I think we might not need size, since it's always fixed.

image

## logo size

image

Style updates:
```diff
<div class="pointer-events-none absolute z-10 top-2 right-2">
<div 
- class="flex items-center gap-1.5 rounded-full bg-black/20 px-2 py-1"
+ class="flex items-center gap-1.5 rounded-full bg-black/20 pr-2"
"
 style="opacity: 0.85;">
<img src="/templates/logo/stability.png" alt="Stability" 
-  class="h-5 w-5  rounded-[50%]" 
+ class="h-6 w-6  rounded-[50%]" 
 rounded-[50%]" 
 draggable="false"><span class="text-sm font-medium text-white">Stability</span></div></div>
```


## Multiple logos support

The status of multiple providers does not seem very good.
image

So I'm wondering if it's possible to change it to like this

```
logos: 
{
"label": "WaveSpeed & Hunyuan",
"gap": "5px" # optional, should support  negative numbers as well
"position": "" # Optional, maybe we should remove it to unify all logo overlays` position
"provider": ["wavespeed", "hunyuan"]
}
```

image

- Each logo has a unified 2px border width
- Use the `gap` to control the space

Not sure if we need to support the `border` setting. When there are multiple logos, a border makes it easier to distinguish them. Need advice here
image

Myestery (dismissed):
Valid suggestions from rabbit around the size but I guess there's no need for it at the moment

src/platform/workflow/templates/repositories/workflowTemplatesStore.ts
```diff
@@ -498,6 +505,37 @@ export const useWorkflowTemplatesStore = defineStore(
...
+          api.fileURL('/templates/index_logo.json')
+        )
+        const contentType = response.headers['content-type']
+        if (!contentType?.includes('application/json')) return {}
+        const result = zLogoIndex.safeParse(response.data)
+        return result.success ? result.data : {}
+      } catch {
+        return {}
+      }
+    }
```
christian-byrne, line 518: all api calls need to be centralized and not scattered about in places like pinia stores -- so we can change auth and headers and cookies and so on systematically at any point in the future

PR #8366, opened 2026-01-28
fix: prevent image/video preview reset on dynamic widget addition

christian-byrne:
## Summary

Fixes image/video previews getting stuck in loading state when widgets are added dynamically to a node.

## Problem

When dynamic widgets are added to a node (e.g., by extensions), Vue reactivity triggers the watch on `imageUrls` prop even when the URL content is identical—the array has a new reference but the same values. This caused:
1. `startDelayedLoader()` to reset loading state to pending
2. If the cached image doesn't trigger `@load` before the 250ms timeout, the loader shows and stays stuck

## Solution

Compare URL arrays by content, not reference. Only reset loading state when URLs actually change:
- Check array length and element-by-element equality
- Return early if URLs are identical (just a new array reference)
- Remove `deep: true` since we compare manually

## Screenshots

image

Screenshot from 2026-01-28 15-24-18

Screenshot from 2026-01-28 15-24-05

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8366-fix-prevent-image-video-preview-reset-on-dynamic-widget-addition-2f66d73d3650819483b2d5cbfb78187f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/ImagePreview.test.ts
+++ b/src/renderer/extensions/vueNodes/components/ImagePreview.test.ts
@@ -308,4 +308,80 @@ describe('ImagePreview', () => {
     expect(imgElement.exists()).toBe(true)
     expect(imgElement.attributes('alt')).toBe('Node output 2')
   })
+
+  describe('URL change detection', () => {
+    it('should NOT reset loading state when imageUrls prop is reassigned with identical URLs', async () => {
+      vi.useFakeTimers()
+      try {
+        const urls = ['/api/view?filename=test.png&type=output']
+        const wrapper = mountImagePreview({ imageUrls: urls })
+
+        // Simulate image load completing
+        const img = wrapper.find('img')
+        await img.trigger('load')
+        await nextTick()
+
+        // Verify loader is hidden after load
+        expect(wrapper.find('[aria-busy="true"]').exists()).toBe(false)
+
+        // Reassign with new array reference but same content
+        await wrapper.setProps({ imageUrls: [...urls] })
+        await nextTick()
+
+        // Advance past the 250ms delayed loader timeout
+        await vi.advanceTimersByTimeAsync(300)
+        await nextTick()
+
+        // Loading state should NOT have been reset - aria-busy should still be false
+        // because the URLs are identical (just a new array reference)
+        expect(wrapper.find('[aria-busy="true"]').exists()).toBe(false)
+      } finally {
+        vi.useRealTimers()
+      }
+    })
+
+    it('should reset loading state when imageUrls prop changes to different URLs', async () => {
+      const urls = ['/api/view?filename=test.png&type=output']
+      const wrapper = mountImagePreview({ imageUrls: urls })
+
+      // Simulate image load completing
+      const img = wrapper.find('img')
+      await img.trigger('load')
+      await nextTick()
+
+      // Verify loader is hidden
+      expect(wrapper.find('[aria-busy="true"]').exists()).toBe(false)
+
+      // Change to different URL
+      await wrapper.setProps({
+        imageUrls: ['/api/view?filename=different.png&type=output']
+      })
+      await nextTick()
+
+      // After 250ms timeout, loading state should be reset (aria-busy="true")
+      // We can check the internal state via the Skeleton appearing
+      // or wait for the timeout
+      await new Promise((resolve) => setTimeout(resolve, 300))
+      await nextTick()
+
+      expect(wrapper.find('[aria-busy="true"]').exists()).toBe(true)
+    })
+
+    it('should handle empty to non-empty URL transitions correctly', async () => {
+      const wrapper = mountImagePreview({ imageUrls: [] })
+
+      // No preview initially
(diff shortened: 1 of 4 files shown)


src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -188,7 +196,7 @@ watch(
     imageError.value = false
     if (newUrls.length > 0) startDelayedLoader()
   },
-  { deep: true, immediate: true }
```
christian-byrne, line 191: Note for reviewers: `deep` is redundant on string arrays -- adding, removing, reassigning all trigger reactivity. Proof: [Vue Playground Link](https://play.vuejs.org/#__SSR__eNqNVEtu2zAQvcqUKCAZUOQW7cqwjaZBFsnCKeyimzBAFGlsM5ZIgaRkG4K2PUCP2JN0SEW20TRpVhrOmzd886Eadl6WcV0hG7GxSbUoLRi0VTnlUhSl0hYa0LiMYJvYdA0tLLUqICBGwCWXqZLGQqVzAxMXF94GSRBB8BDcDXo0V6sedE4uh0OYKcgQyxFYXeE7Ln32MBzAZOqzxXWSVxhBKHH7I8kjUHlGX483XIJL2sXEZWXW4b3VYrVCjdkI3jfXi5tZbMglV2K5D5+4Lfz++es52t0waO9JW+v1LSuZWqEkJFl2ZbEgWf7OozCq5zaO41OlQRrcUYITtsZC1fhiguMhNrlIMfwQwdlHJ+Ikh0kKvFDSorQz3M6ph//X0skYD7tx0iDpQCLKPLFIJ4BxJmpvkFlOHXEETdNNsW3HQzd8Dz5U1pKILynp20w4e+oHZ9PzLANB5njYxbzEOPaASHN/eBPvWd1EX5APaKWcMwIam9upvxOtdW9RkVCfLZWmdCFR9D4CMQAh3e5wBqMN7gkSlJmK9wG++r45vTUeHprHImYNKViKVfxolKRH44fBWaqKUuSob0o3NsMZdbTTQV3Lc7W99j637VHvT9eYbv7hfzQ75+Psm0aDukbODphN9AptB18uZrgj+wAWKqtyin4FnKNReeU0dmFfK5mR7JM4r/bKP316IN/N5Y76bfqinFAX2fp4zug/cPFK6Ue5n+LPnke7ydo/nYuAmg==)

src/renderer/extensions/vueNodes/components/ImagePreview.test.ts
```diff
@@ -308,4 +308,80 @@ describe('ImagePreview', () => {
...
+
+        // Verify loader is hidden after load
+        expect(wrapper.find('[aria-busy="true"]').exists()).toBe(false)
+
+        // Reassign with new array reference but same content
+        await wrapper.setProps({ imageUrls: [...urls] })
+        await nextTick()
+
+        // Advance past the 250ms delayed loader timeout
+        await vi.advanceTimersByTimeAsync(300)
```
AustinMroz, line 332: Didn't know this was a thing. Super cool

AustinMroz (commented):
Change to watches look correct, but when manually testing, it seems that Videos are still reloading anytime the dynamicCombo is changed because the `rand` parameter in the url is changed.

Will keep investigating.

AustinMroz:
In `LGraphNode.vue` `nodeMedia` is recomputing because the `hasVideoInput` iterates over node inputs (which have changed).

If we move `hasVideoInput` into a separate computed, then `nodeMedia` will only refire when the actual outputs have changed.

christian-byrne:
> Change to watches look correct, but when manually testing, it seems that Videos are still reloading anytime the dynamicCombo is changed because the rand parameter in the url is changed.

This was just to fix the previews breaking entirely when dynamic widget changes. The `rand` and aggresive cache-busting on each `/view` call (on local) I feel is a separate thing. But curious: do you think we should remove that at least for videos? I can see people deleting image outputs then creating new ones with the same name, but videos might be less likely (and the perf hit is worse). WDYT?

AustinMroz:
>This was just to fix the previews breaking entirely when dynamic widget changes.

I'm pretty sure that's the part I'm having issues with. I'm running this branch with a single added `console.log(urlsChanged)` to the watch along with Comfy-Org/ComfyUI#11940. Every time I update the `codec` option, the video reloads and I get a log that the url has changed.

> do you think we should remove that

I've got a slight leaning towards "not handing video differently than images". The ideal case is removing the cache busting entirely and trusting `cache-control` headers. That way we minimize the code changes between local and cloud.

christian-byrne:
> If we move hasVideoInput into a separate computed, then nodeMedia will only refire when the actual outputs have changed.

Done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/8366/changes/9013914a5b90a551872fffd8ab065de2ec0c5738

christian-byrne:
Tested and works better now (doesn't re-render when widgets are added).

AustinMroz (approved):

PR #8367, opened 2026-01-28
fix: suppress link rendering during slot sync after graph reconfigure

christian-byrne:
## Description

Fixes link alignment issues after undo/redo operations in Vue Nodes 2.0. When multiple connections exist from the same output to different nodes, performing an undo would cause the connections to become misaligned with their inputs.

## Root Cause

When undo triggers `loadGraphData`, the graph is reconfigured and Vue node components are destroyed and recreated. The new slot elements mount and schedule RAF-batched position syncs via `scheduleSlotLayoutSync`. However, links are drawn **before** the RAF batch completes, causing `getSlotPosition()` to return stale/missing positions.

## Solution

- Export a new `flushPendingSlotLayoutSyncs()` function from `useSlotElementTracking.ts`
- Create a `useGraphConfigureSlotSync` composable that flushes pending syncs after graph configuration
- Integrate the flush into `addAfterConfigureHandler` in `app.ts`, called after `onAfterGraphConfigured`
- Force canvas redraw after flushing to render links with correct positions

## Testing

- Added unit tests for `flushPendingSlotLayoutSyncs`
- Added unit tests for `useGraphConfigureSlotSync` composable
- Manual verification: connections now align correctly after undo/redo operations

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8367-fix-flush-pending-slot-layout-syncs-after-graph-configure-2f66d73d365081a3a564fac96c44a048) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
+++ b/src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
@@ -36,9 +36,9 @@ function scheduleSlotLayoutSync(nodeId: string) {
 
 export function flushScheduledSlotLayoutSync() {
   if (pendingNodes.size === 0) {
-    // No pending nodes - don't clear the flag here as late mounts may still
-    // call scheduleSlotLayoutSync. The RAF callback will clear it after
-    // processing actual nodes.
+    // No pending nodes and no RAF scheduled - clear the flag to avoid
+    // permanently suppressing link rendering (e.g., workflows without Vue nodes)
+    layoutStore.setPendingSlotSync(false)
     return
   }
   const conv = useSharedCanvasPositionConversion()
--- a/src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
+++ b/src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
@@ -28,13 +28,17 @@ const raf = createRafBatch(() => {
 
 function scheduleSlotLayoutSync(nodeId: string) {
   pendingNodes.add(nodeId)
+  // Re-assert pending flag for late mounts (Vue components mounting after
+  // flushScheduledSlotLayoutSync was called synchronously in onConfigure)
+  layoutStore.setPendingSlotSync(true)
   raf.schedule()
 }
 
 export function flushScheduledSlotLayoutSync() {
   if (pendingNodes.size === 0) {
-    // Even if no pending nodes, clear the flag (e.g., graph with no nodes)
-    layoutStore.setPendingSlotSync(false)
+    // No pending nodes - don't clear the flag here as late mounts may still
+    // call scheduleSlotLayoutSync. The RAF callback will clear it after
+    // processing actual nodes.
     return
   }
   const conv = useSharedCanvasPositionConversion()
--- a/src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
+++ b/src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
@@ -37,8 +37,15 @@ function scheduleSlotLayoutSync(nodeId: string) {
 
 export function flushScheduledSlotLayoutSync() {
   if (pendingNodes.size === 0) {
-    // No pending nodes and no RAF scheduled - clear the flag to avoid
-    // permanently suppressing link rendering (e.g., workflows without Vue nodes)
+    // No pending nodes - check if we should wait for Vue components to mount
+    const graph = app.canvas?.graph
+    const hasNodes = graph && graph._nodes && graph._nodes.length > 0
+    if (hasNodes) {
+      // Graph has nodes but Vue hasn't mounted them yet - keep flag set
+      // so late mounts can re-assert it via scheduleSlotLayoutSync()
+      return
+    }
+    // No nodes in graph - safe to clear the flag (no Vue components will mount)
     layoutStore.setPendingSlotSync(false)
     app.canvas?.setDirty(true, true)
     return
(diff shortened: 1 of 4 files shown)


src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -31,13 +31,19 @@ function scheduleSlotLayoutSync(nodeId: string) {
   raf.schedule()
 }
 
-function flushScheduledSlotLayoutSync() {
-  if (pendingNodes.size === 0) return
+export function flushScheduledSlotLayoutSync() {
+  if (pendingNodes.size === 0) {
+    // Even if no pending nodes, clear the flag (e.g., graph with no nodes)
```
benceruleanlu: Possible race condition: `flushScheduledSlotLayoutSync()` is called synchronously right after `onConfigure`. If Vue slot elements haven’t mounted yet, `pendingNodes.size === 0` here, so this block clears `pendingSlotSync` and returns. That re‑enables link rendering before any DOM-based slot positions exist, which can reintroduce the misaligned-links bug under load. `pendingNodes` only becomes non‑empty when `useSlotElementTracking` runs in `onMounted` and calls `scheduleSlotLayoutSync()`. Consider only clearing the flag after a real flush (non-empty), or setting `pendingSlotSync` to true in `scheduleSlotLayoutSync()` so late mounts re‑assert it.

src/scripts/app.ts
```diff
@@ -730,6 +732,11 @@ export class ComfyApp {
   private addAfterConfigureHandler(graph: LGraph) {
     const { onConfigure } = graph
     graph.onConfigure = function (...args) {
+      // Set pending sync flag to suppress link rendering until slots are synced
```
benceruleanlu, line 735: Potential stuck state: if `onConfigure` (or callbacks it triggers) throws, `pendingSlotSync` is never cleared because the flush/clear logic is after the call. That can permanently suppress link rendering. Suggest wrapping the handler body in `try/finally` and clearing the flag in `finally` (or at least in a catch) so links don’t stay hidden on error paths.

benceruleanlu (dismissed):
LGTM after taking a look at issues mentioned

Also, did the agent forget to commit or push the unit tests mentioned? having those as a regression test would be nice

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
christian-byrne: Addressed in commit 2e1845bb - `scheduleSlotLayoutSync()` now re-asserts the `pendingSlotSync` flag for late-mounting Vue components, and we only clear the flag when there are actual nodes to process.

src/scripts/app.ts
christian-byrne: Addressed in commit c2d548b - wrapped the `onConfigure` handler body in `try/finally` to ensure the flush runs even if an error is thrown.

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -28,16 +28,26 @@ const raf = createRafBatch(() => {
...
+    return
+  }
   const conv = useSharedCanvasPositionConversion()
   for (const nodeId of Array.from(pendingNodes)) {
     pendingNodes.delete(nodeId)
     syncNodeSlotLayoutsFromDOM(nodeId, conv)
   }
+  // Clear the pending sync flag - slots are now synced
+  layoutStore.setPendingSlotSync(false)
 }
```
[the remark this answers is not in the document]
christian-byrne: Addressed in commit 14d5bbe - the function now checks if the graph has nodes before deciding whether to clear the flag. If the graph has nodes but `pendingNodes` is empty, we keep the flag set to allow Vue components to mount. If the graph is empty (no nodes), we clear the flag since no Vue components will mount.

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -12,6 +12,7 @@ import { useSharedCanvasPositionConversion } from '@/composables/element/useCanv
 import { LiteGraph } from '@/lib/litegraph/src/litegraph'
 import { getSlotKey } from '@/renderer/core/layout/slots/slotIdentifier'
 import { layoutStore } from '@/renderer/core/layout/store/layoutStore'
+import { app } from '@/scripts/app'
```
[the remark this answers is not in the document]
christian-byrne: This is a false positive - the `@/scripts/app` import resolves correctly and lint passes locally with no errors. The project's alias resolver includes this path.

christian-byrne:
## Review Comments Addressed

Pushed 3 commits addressing the review feedback:

### 1. CodeRabbit - Clear reroute cache (LGraphCanvas.ts)
Commit `bf21a055`: Now clears `this.#visibleReroutes` before the early return to prevent stale reroute entries from affecting hit-testing while `pendingSlotSync` is true.

### 2. @benceruleanlu - Race condition with late mounts (useSlotElementTracking.ts)
Commit `2e1845bb`: `scheduleSlotLayoutSync()` now re-asserts the `pendingSlotSync` flag so Vue components mounting after the synchronous `flushScheduledSlotLayoutSync()` call will keep link rendering suppressed until their slots are synced. The flag is only cleared when there are actual nodes to process.

### 3. @benceruleanlu - Error handling stuck state (app.ts)
Commit `c2d548bf`: Wrapped the `onConfigure` handler body in `try/finally` to ensure `pendingSlotSync` is cleared even if an error is thrown during graph configuration, preventing links from being permanently hidden.

PR #8372, opened 2026-01-28
fix: remove redundant forceRender call and add ResizeObserver guard

jtydhr88:
## Summary
- Remove duplicate forceRender() in ResizeObserver callback since handleResize() already calls it
- Add guard for environments without ResizeObserver support
- Disconnect existing observer before reassigning to prevent leaks

requested by @DrJKL in https://github.com/Comfy-Org/ComfyUI_frontend/pull/8351

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8372-fix-remove-redundant-forceRender-call-and-add-ResizeObserver-guard-2f66d73d3650811bb3a6de5c59b3c1fb) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/load3d/Load3d.ts
+++ b/src/extensions/core/load3d/Load3d.ts
@@ -157,9 +157,11 @@ class Load3d {
   }
 
   private initResizeObserver(container: Element | HTMLElement): void {
+    if (typeof ResizeObserver === 'undefined') return
+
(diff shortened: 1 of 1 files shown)


DrJKL (dismissed):
Want to remove the other forceRender calls while you're in here?

DrJKL (approved):

PR #8378, opened 2026-01-28
feat: add category support for blueprints and protect global blueprints

christian-byrne:
## Summary

This PR adds two related features for subgraph blueprints:

### 1. Protect Global Blueprints from Deletion
- Added `isGlobalBlueprint()` helper that distinguishes blueprints by `python_module` field
- User blueprints have `python_module: 'blueprint'`
- Global (installed) blueprints have `python_module` set to the node pack name
- Guarded `deleteBlueprint()` to show a warning toast for global blueprints
- Hidden delete menu in node library for global blueprints

### 2. Category Support for Blueprints
- User blueprints now use category `Subgraph Blueprints/User`
- Global blueprints use `Subgraph Blueprints/{category}` if category is provided, otherwise `Subgraph Blueprints`
- Extended `GlobalSubgraphData.info` type with optional `category` field
- Added `category` to `zSubgraphDefinition` schema

## Files Changed
- `src/stores/subgraphStore.ts` - Core logic for both features
- `src/stores/subgraphStore.test.ts` - Tests for `isGlobalBlueprint`
- `src/components/sidebar/tabs/nodeLibrary/NodeTreeLeaf.vue` - Hide delete menu for global blueprints
- `src/scripts/api.ts` - Extended `GlobalSubgraphData` type
- `src/platform/workflow/validation/schemas/workflowSchema.ts` - Added category to schema
- `src/locales/en/main.json` - Added translation key

## Testing
- ✅ `pnpm typecheck` passed
- ✅ `pnpm lint` passed

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8378-feat-add-category-support-for-blueprints-and-protect-global-blueprints-2f66d73d36508137aa67c2d88c358b69) by [Unito](https://www.unito.io)

PR #8379, opened 2026-01-28
Fix assigning category of user blueprints on save

AustinMroz:
PR into #8378 

From testing, when a user blueprint is saved, or published, it wouldn't have the `User` category assigned until the next reload.

Also fixes a doubled registration.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8379-Fix-assigning-category-of-user-blueprints-on-save-2f66d73d365081f5a882fe88ad3e2931) by [Unito](https://www.unito.io)


christian-byrne (approved):
Thank you

PR #8396, opened 2026-01-29
[bugfix] Fix shift+click deselection in asset panel

viva-jinyi:
## Summary
Fix shift+click range selection not properly deselecting assets when selecting a smaller range, and improve selection performance.

## Changes
- **Bug Fix**: Shift+click now replaces selection with the new range instead of combining with existing selection
- **Performance**: Remove unnecessary `.every()` check in `setSelection` (O(n) → O(1))
- **Tests**: Add 23 unit tests for asset selection logic

## Test Plan
- [x] Click 1st asset → only 1st selected
- [x] Shift+click 3rd asset → items 1-3 selected
- [x] Shift+click 1st asset → only 1st selected (was broken, now fixed)
- [x] All 23 new unit tests pass

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8396-bugfix-Fix-shift-click-deselection-in-asset-panel-2f76d73d3650814ca060d1e6a40cf6d4) by [Unito](https://www.unito.io)


src/platform/assets/composables/useAssetSelectionStore.ts
```diff
@@ -21,14 +21,7 @@ export const useAssetSelectionStore = defineStore('assetSelection', () => {
   }
 
   function setSelection(assetIds: string[]) {
```
benceruleanlu, line 23: nit: This now always replaces the Set, even if the IDs are identical. Probably fine, but if any watchers/effects do expensive work on selection change, we could trigger extra recomputation. Worth confirming no heavy watchers depend on `selectedAssetIds`, or consider a cheap equality guard.

benceruleanlu (approved):
Looks good overall. Main behavior fix matches expected shift+click range selection. Left a small perf consideration about always replacing the Set in setSelection.

christian-byrne (commented):
LGTM (read briefly)

PR #8409, opened 2026-01-29
[bugfix] Fix manager missing node tab with shared composable

viva-jinyi:
## Summary
Fix the manager missing nodes tab by converting `useWorkflowPacks` to a shared singleton composable using `createSharedComposable` from VueUse.

## Changes
- **What**: Wrap `useWorkflowPacks` with `createSharedComposable` to ensure all components share the same instance
- **Why**: Multiple instances were causing the missing nodes tab to not work correctly

## Technical Details
- Removed unused `UseNodePacksOptions` parameter since the shared composable pattern doesn't support per-call options
- All components now share the same workflow packs state and data fetching

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8409-bugfix-Fix-manager-missing-node-tab-with-shared-composable-2f76d73d365081e682acfa3a7450cdd6) by [Unito](https://www.unito.io)

### DIFF
--- a/src/workbench/extensions/manager/composables/nodePack/useWorkflowPacks.ts
+++ b/src/workbench/extensions/manager/composables/nodePack/useWorkflowPacks.ts
@@ -1,3 +1,4 @@
+import { createSharedComposable } from '@vueuse/core'
 import { computed, onUnmounted, ref } from 'vue'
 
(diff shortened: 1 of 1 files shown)


christian-byrne (approved):
LGTM!

PR #8418, opened 2026-01-29
feat: add Chatterbox model support for Cloud asset browser

christian-byrne:
## Summary

Adds support for creating Chatterbox TTS nodes when clicking Chatterbox models in the Cloud asset browser.

### modelToNodeStore.ts
- Add `findProvidersWithFallback()` helper for hierarchical model type lookups (e.g., `parent/child` falls back to `parent`)
- Register 4 Chatterbox model directories with empty widget keys:
  - `chatterbox/chatterbox` → `FL_ChatterboxTTS`
  - `chatterbox/chatterbox_turbo` → `FL_ChatterboxTurboTTS`
  - `chatterbox/chatterbox_multilingual` → `FL_ChatterboxMultilingualTTS`
  - `chatterbox/chatterbox_vc` → `FL_ChatterboxVC`

### createModelNodeFromAsset.ts
- Skip widget assignment when `provider.key` is empty (for nodes that auto-load models without a widget selector)

### Tests
- Add tests for hierarchical fallback behavior
- Add tests for empty widget key (auto-load nodes)
- Add Chatterbox node types to mock data

## Notes

- Empty `key` convention: Chatterbox nodes auto-load their models and don't have a model selector widget, so we register them with `key: ''` and skip the widget assignment step
- Hierarchical fallback only goes one level deep (`a/b/c` → `a`, not `a/b`)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8418-feat-add-Chatterbox-model-support-for-Cloud-asset-browser-2f76d73d365081be822bc369b155f099) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/modelToNodeStore.ts
+++ b/src/stores/modelToNodeStore.ts
@@ -81,6 +81,10 @@ export const useModelToNodeStore = defineStore('modelToNode', () => {
   function findProvidersWithFallback(
     modelType: string
   ): ModelNodeProvider[] | undefined {
+    if (!modelType || typeof modelType !== 'string') {
+      return undefined
+    }
+
     const exactMatch = modelToNodeMap.value[modelType]
     if (exactMatch && exactMatch.length > 0) return exactMatch
 
--- a/src/stores/modelToNodeStore.ts
+++ b/src/stores/modelToNodeStore.ts
@@ -90,7 +90,7 @@ export const useModelToNodeStore = defineStore('modelToNode', () => {
     const fallback = modelToNodeMap.value[topLevel]
 
     if (fallback && fallback.length > 0) return fallback
-    
+
     return undefined
   }
 
--- a/src/stores/modelToNodeStore.ts
+++ b/src/stores/modelToNodeStore.ts
@@ -9,7 +9,7 @@ export class ModelNodeProvider {
   /** The node definition to use for this model. */
   public nodeDef: ComfyNodeDefImpl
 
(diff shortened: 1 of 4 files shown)


christian-byrne:
This code is somewhat temporary, as it will all get replaced when either asset-model strings become first-class types or we start specifying asset/model metadata on the node class.

src/stores/modelToNodeStore.ts
```diff
@@ -73,23 +73,45 @@ export const useModelToNodeStore = defineStore('modelToNode', () => {
...
+  ): ModelNodeProvider[] | undefined {
+    const exactMatch = modelToNodeMap.value[modelType]
+    if (exactMatch && exactMatch.length > 0) return exactMatch
+
+    const topLevel = modelType.split('/')[0]
+    if (topLevel !== modelType) {
+      const fallback = modelToNodeMap.value[topLevel]
+      if (fallback && fallback.length > 0) return fallback
+    }
+    return undefined
```
DrJKL: Nit: Technically adds an additional return, but it flattens overall structure so all of the non-fallthrough returns are guard clausy
(Thought about trying to make the last one a ternary but I'm not sure)
```suggestion
    if (topLevel === modelType) return undefined

    const fallback = modelToNodeMap.value[topLevel]

    if (fallback && fallback.length > 0) return fallback
    
    return undefined
```

DrJKL (dismissed):

DrJKL (approved):

PR #8377, opened 2026-01-28
Chore: Actions updates and cleanup

DrJKL:
## Summary

...

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8377-WIP-Chore-Actions-updates-and-cleanup-2f66d73d3650818483a8dffa32a6f245) by [Unito](https://www.unito.io)

PR #8432, opened 2026-01-29
fix: use merge-multiple for snapshot artifact download

christian-byrne:
## Summary

Fixes the snapshot merge failure introduced by PR #8377 (actions/download-artifact v4→v7 upgrade).

## Root Cause

The v5+ release of `download-artifact` changed behavior: when a `pattern` matches only a **single artifact**, files are extracted directly to `path/` without the artifact name subdirectory. When only one shard had changes, the merge loop couldn't find the expected `snapshots-shard-*/` directories.

## Fix

Use `merge-multiple: true` — the documented pattern for combining sharded artifacts. This merges all matched artifacts directly into the target path, eliminating directory structure assumptions.

## Testing

This fix can be validated by re-running the workflow on [PR #8276](https://github.com/Comfy-Org/ComfyUI_frontend/pull/8276) after merge.

---
- Fixes snapshot update workflow regression from #8377

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8432-fix-use-merge-multiple-for-snapshot-artifact-download-2f76d73d3650810b97fdfe28cd3c7694) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/pr-update-playwright-expectations.yaml
+++ b/.github/workflows/pr-update-playwright-expectations.yaml
@@ -180,7 +180,7 @@ jobs:
         with:
           pattern: snapshots-shard-*
           path: ./downloaded-snapshots
-          merge-multiple: false
+          merge-multiple: true
 
       - name: List downloaded files
         run: |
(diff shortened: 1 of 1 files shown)


DrJKL (dismissed):
Want to force this to run on this PR?

browser_tests/tests/loadWorkflowInMedia.spec.ts-snapshots/dropped-workflow-url-hidream-dev-example-png-chromium-linux.png
DrJKL, line 1: Noooooooo

christian-byrne:
Fixing

DrJKL (approved):

PR #8435, opened 2026-01-29
Fix invalid keybind flash

AustinMroz:
Previously the save keybind action would
- apply the new keybind
- wait for a network request to persist the change
- close the dialogue regardless of the results of the above changes

During this network request, the dialog would show a warning that the keybind is invalid because the dialogue "contains a modified keybind which conflicts with an existing keybind"
image

This PR changes the order these actions are applied in.
- The dialogue is immediately closed
- The keybinding is updated if valid
- The keybinding is persisted.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8435-Fix-invalid-keybind-flash-2f76d73d3650815c9657f35e77d331fe) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/setting/KeybindingPanel.vue
+++ b/src/components/dialog/content/setting/KeybindingPanel.vue
@@ -265,18 +265,15 @@ function cancelEdit() {
 }
 
 async function saveKeybinding() {
-  if (currentEditingCommand.value && newBindingKeyCombo.value) {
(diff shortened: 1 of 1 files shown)


christian-byrne (approved):
Nice, was also attempting to fix it last night ([branch](https://github.com/Comfy-Org/ComfyUI_frontend/compare/main...fix/keybind-save-flash)) -- you did better than me. This solution makes sense.

PR #8440, opened 2026-01-29
refactor: migrate ES private fields to TypeScript private for Vue Proxy compatibility

DrJKL:
## Summary

Migrates ECMAScript private fields (`#`) to TypeScript private (`private`) across LiteGraph to fix Vue Proxy reactivity incompatibility.

## Problem

ES private fields (`#field`) are incompatible with Vue's Proxy-based reactivity system - accessing `#field` through a Proxy throws `TypeError: Cannot read private member from an object whose class did not declare it`.

## Solution

- Converted all `#field` to `private _field` across 10 phases
- Added `toJSON()` methods to `LGraph`, `NodeSlot`, `NodeInputSlot`, and `NodeOutputSlot` to prevent circular reference errors during serialization (TypeScript private fields are visible to `JSON.stringify` unlike true ES private fields)
- Made `DragAndScale.element.data` non-enumerable to break canvas circular reference chain

## Testing

- All 4027 unit tests pass
- Added 9 new serialization tests to catch future circular reference issues
- Browser tests (undo/redo, save workflows) verified working

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8440-refactor-migrate-ES-private-fields-to-TypeScript-private-for-Vue-Proxy-compatibility-2f76d73d365081a3bd82d429a3e0fcb7) by [Unito](https://www.unito.io)


src/lib/litegraph/src/node/NodeInputSlot.ts
```diff
@@ -79,4 +79,12 @@ export class NodeInputSlot extends NodeSlot implements INodeInputSlot {
...
     ctx.textAlign = textAlign
   }
+
+  override toJSON(): INodeInputSlot {
+    return {
+      ...super.toJSON(),
+      link: this.link,
+      widget: this.widget
+    }
+  }
```
christian-byrne, line 89: Were there other things you considered for this?

christian-byrne (approved):
Only concern previously was with shadowing properties added by 3P code. Using underscore prefix should solve that. See no other issues. LGTM

src/lib/litegraph/src/node/NodeInputSlot.ts
AustinMroz: I was pondering the same. Would it break things if we made all the `toJSON` calls just return `serialize`?

DrJKL: Yes, a few things.
The best solution would be to remove the circular references, but that would be a much larger undertaking and likely™️ cause issues with existing extensions.
Another option is to replace our internal calls to JSON.stringify with a serialize that does much the same thing here, but that has the same issue of not working for existing external callers who would then hit the circular dereference bug at runtime.

DrJKL: @AustinMroz We'd need to implement `serialize` on each of these still, then call it from `toJSON`, right?

AustinMroz: :thinking: Probably best we don't implement new serialize methods. The smaller places you added it (like NodeSlot) won't have `toJSON` called if the parent classes are using their (already safe) `serialize` methods. And your newly written `toJSON` methods are still an improvement when debugging with `console.log`

AustinMroz (approved):

PR #8464, opened 2026-01-30
fix: use GitHub App token to trigger e2e tests after lint/i18n commits

DrJKL:
Use GitHub App token instead of GITHUB_TOKEN so that commits pushed by lint and i18n workflows trigger downstream e2e tests.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8464-fix-use-GitHub-App-token-to-trigger-e2e-tests-after-lint-i18n-commits-2f86d73d36508156b4b9e06afd70470f) by [Unito](https://www.unito.io)


snomiao:
Amazing idea! (this should work, and currently CI have scopes problem need to fix..)

## Ref:

Token differences:

When to Use Which


Use the GITHUB_TOKEN by default when your workflow only needs to interact with resources within the same repository (e.g., creating an issue, commenting on a pull request, or pushing code).


Use a GitHub App token when you need your workflow to access resources in different repositories, perform actions that the default token is not allowed to do (like triggering another workflow), or build a long-lived, non-user-dependent integration. You can create a GitHub App on the [GitHub Developer settings page](https://github.com/settings/tokens) and use an action like the [tibdex/github-app-token action](https://github.com/marketplace/actions/github-app-token) to generate the token within your workflow.

PR #8340, opened 2026-01-27
fix: await needsLogin

DrJKL:
## Summary

Add additional protection for bootstrap order issues.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8340-fix-await-needsLogin-2f56d73d365081c9b3c6d5a091c1eb8d) by [Unito](https://www.unito.io)

PR #8483, opened 2026-01-30
feat: classify missing nodes by replacement availability

viva-jinyi:
## Summary
- Extend `MissingNodeType` with `isReplaceable` and `replacement` fields
- Classify missing nodes by checking `nodeReplacementStore.getReplacementFor()` during graph load
- Wrap hardcoded node patches (T2IAdapterLoader, ConditioningAverage, etc.) in `if (!isEnabled)` guard so they only run when node replacement setting is disabled
- Change `useNodeReplacementStore().load()` from fire-and-forget (`void`) to `await` so replacement data is available before missing node detection
- Fix guard condition order in `nodeReplacementStore.load()`: check `isEnabled` before `isLoaded`
- Align `InputMap` types with actual API response (flat `old_id`/`set_value` fields instead of nested `assign` wrapper)

## Test plan
- [x] Load workflow with deprecated nodes (T2IAdapterLoader, Load3DAnimation, SDV_img2vid_Conditioning)
- [x] Verify missing nodes are classified with `isReplaceable: true` and `replacement` object
- [x] Verify hardcoded patches only run when node replacement setting is OFF
- [x] Verify `nodeReplacementStore.load()` is not called when setting is disabled
- [x] Unit tests pass (`nodeReplacementStore.test.ts` - 16 tests)
- [x] Typecheck passes


christian-byrne (commented):
```
bootstrapStore.ts — loadAuthenticatedStores():
  void settingStore.load()
  void workflowStore.loadWorkflows()
+ void nodeReplacementStore.load()    // starts in parallel

app.ts — collectMissingNodesAndModels():
+ const { isLoaded } = storeToRefs(useNodeReplacementStore())
+ await until(isLoaded).toBe(true)    // reactive wait at consumption
  // ...then use getReplacementFor() safely
````

This gives parallel loading (no startup latency hit) + safe consumption (reactive wait) + centralized bootstrap (single orchestration point), fully aligned with #8340's architecture.

christian-byrne (dismissed):

viva-jinyi:
I updated the code based on the review above, but it caused an issue where workflows no longer load. I'm investigating more.

viva-jinyi:
@christian-byrne 
The pattern of calling `void nodeReplacementStore.load()` in `bootstrapStore.loadAuthenticatedStores()` and awaiting `until(isLoaded).toBe(true)` in `loadGraphData()` does not work due to a settings loading order dependency.

In `loadAuthenticatedStores()`, `settingStore.load()` is also called with void (fire-and-forget). This means when `nodeReplacementStore.load()` executes, settings haven’t been loaded yet. As a result, isEnabled reads its default value (false), and load() returns immediately without setting isLoaded to true. Later in loadGraphData(), after settings are loaded, isEnabled becomes true and passes the guard, but isLoaded remains false forever — causing `await until(isLoaded).toBe(true)` to hang indefinitely and blocking workflow loading entirely.

The existing await `useNodeReplacementStore().load()` in setup() works correctly because settings are already loaded at that point, so isEnabled has the correct value.

christian-byrne (approved):

PR #8521, opened 2026-02-01
feat: add Painter Node

jtydhr88:
## Summary                                                                                                                             
Add PainterNode widget for freehand mask drawing directly on the canvas, with brush/eraser tools, opacity, hardness, and background color controls.

need BE changes https://github.com/Comfy-Org/ComfyUI/pull/12294

https://github.com/user-attachments/assets/7222063a-0e40-40bb-b72e-b42c8984beb9

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8521-feat-add-Painter-Node-2fa6d73d36508124ab2ede449a0cc67a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/painter/WidgetPainter.vue
+++ b/src/components/painter/WidgetPainter.vue
@@ -1,0 +1,359 @@
+<template>
+  <div
+    class="widget-expands flex h-full w-full flex-col gap-1"
+    @pointerdown.stop
+    @pointermove.stop
+    @pointerup.stop
+  >
+    <div
+      class="flex min-h-0 flex-1 items-center justify-center overflow-hidden rounded-lg bg-node-component-surface"
+    >
+      <div class="relative max-h-full w-full" :style="canvasContainerStyle">
+        <img
+          v-if="inputImageUrl"
+          :src="inputImageUrl"
+          class="absolute inset-0 size-full"
+          draggable="false"
+          @load="handleInputImageLoad"
+          @dragstart.prevent
+        />
+        <canvas
+          ref="canvasEl"
+          class="absolute inset-0 size-full cursor-none touch-none"
+          @pointerdown="handlePointerDown"
+          @pointermove="handlePointerMove"
+          @pointerup="handlePointerUp"
+          @pointerenter="handlePointerEnter"
+          @pointerleave="handlePointerLeave"
+        />
+        <div
+          v-show="cursorVisible"
+          class="pointer-events-none absolute left-0 top-0 rounded-full border border-black/60 shadow-[0_0_0_1px_rgba(255,255,255,0.8)]"
+          :style="cursorStyle"
+        />
+      </div>
+    </div>
+
+    <div
+      v-if="isImageInputConnected"
+      class="text-center text-xs text-muted-foreground"
+    >
+      {{ canvasWidth }} x {{ canvasHeight }}
+    </div>
+
+    <div
+      ref="controlsEl"
+      :class="
+        cn(
+          'grid shrink-0 gap-x-1 gap-y-1',
+          compact ? 'grid-cols-1' : 'grid-cols-[auto_1fr]'
+        )
+      "
+    >
+      <div
+        v-if="!compact"
+        class="flex w-28 items-center truncate text-sm text-muted-foreground"
+      >
+        {{ $t('painter.tool') }}
+      </div>
+      <div
+        class="flex h-8 items-center gap-1 rounded-sm bg-component-node-widget-background p-1"
+      >
+        <Button
+          variant="textonly"
+          size="unset"
+          :class="
+            cn(
+              'flex-1 self-stretch px-2 text-xs transition-colors',
+              tool === PAINTER_TOOLS.BRUSH
+                ? 'rounded-sm bg-component-node-widget-background-selected text-base-foreground'
+                : 'text-node-text-muted hover:text-node-text'
+            )
+          "
+          @click="tool = PAINTER_TOOLS.BRUSH"
+        >
+          {{ $t('painter.brush') }}
+        </Button>
+        <Button
+          variant="textonly"
+          size="unset"
+          :class="
+            cn(
+              'flex-1 self-stretch px-2 text-xs transition-colors',
+              tool === PAINTER_TOOLS.ERASER
+                ? 'rounded-sm bg-component-node-widget-background-selected text-base-foreground'
+                : 'text-node-text-muted hover:text-node-text'
+            )
+          "
+          @click="tool = PAINTER_TOOLS.ERASER"
+        >
+          {{ $t('painter.eraser') }}
+        </Button>
+      </div>
+
+      <div
+        v-if="!compact"
+        class="flex w-28 items-center truncate text-sm text-muted-foreground"
+      >
+        {{ $t('painter.size') }}
+      </div>
+      <div
+        class="flex h-8 items-center gap-2 rounded-lg bg-component-node-widget-background pr-2 pl-3"
+      >
+        <Slider
+          :model-value="[brushSize]"
+          :min="1"
+          :max="200"
+          :step="1"
+          class="flex-1"
+          @update:model-value="(v) => v?.length && (brushSize = v[0])"
+        />
+        <span class="w-8 text-center text-xs text-node-text-muted">{{
+          brushSize
+        }}</span>
+      </div>
+
+      <template v-if="tool === PAINTER_TOOLS.BRUSH">
+        <div
+          class="flex w-28 items-center truncate text-sm text-muted-foreground"
+        >
+          {{ $t('painter.color') }}
+        </div>
+        <div
+          class="flex h-8 w-full items-center gap-2 rounded-lg bg-component-node-widget-background px-4"
+        >
+          <input
+            type="color"
+            :value="brushColorDisplay"
+            class="h-4 w-8 cursor-pointer appearance-none overflow-hidden rounded-full border-none bg-transparent [&::-webkit-color-swatch-wrapper]:p-0 [&::-webkit-color-swatch]:border-none [&::-webkit-color-swatch]:rounded-full [&::-moz-color-swatch]:border-none [&::-moz-color-swatch]:rounded-full"
+            @input="
+              (e) => (brushColorDisplay = (e.target as HTMLInputElement).value)
+            "
+          />
+          <span class="min-w-[4ch] truncate text-xs">{{
+            brushColorDisplay
+          }}</span>
+          <span class="ml-auto flex items-center text-xs text-node-text-muted">
+            <input
+              type="number"
+              :value="brushOpacityPercent"
+              min="0"
+              max="100"
+              step="1"
+              class="w-7 appearance-none border-0 bg-transparent text-right text-xs text-node-text-muted outline-none [&::-webkit-inner-spin-button]:appearance-none [&::-webkit-outer-spin-button]:appearance-none [-moz-appearance:textfield]"
+              @click.prevent
+              @change="
+                (e) => {
+                  const val = Math.min(
+                    100,
+                    Math.max(0, Number((e.target as HTMLInputElement).value))
+                  )
+                  brushOpacityPercent = val
+                  ;(e.target as HTMLInputElement).value = String(val)
+                }
+              "
+            />%</span
+          >
+        </div>
+
+        <div
+          class="flex w-28 items-center truncate text-sm text-muted-foreground"
+        >
+          {{ $t('painter.hardness') }}
+        </div>
+        <div
+          class="flex h-8 items-center gap-2 rounded-lg bg-component-node-widget-background pr-2 pl-3"
+        >
+          <Slider
+            :model-value="[brushHardnessPercent]"
+            :min="0"
+            :max="100"
+            :step="1"
+            class="flex-1"
+            @update:model-value="
+              (v) => v?.length && (brushHardnessPercent = v[0])
+            "
+          />
+          <span class="w-8 text-center text-xs text-node-text-muted"
+            >{{ brushHardnessPercent }}%</span
+          >
+        </div>
+      </template>
+
+      <template v-if="!isImageInputConnected">
+        <div
+          class="flex w-28 items-center truncate text-sm text-muted-foreground"
+        >
+          {{ $t('painter.width') }}
+        </div>
+        <div
+          class="flex h-8 items-center gap-2 rounded-lg bg-component-node-widget-background pr-2 pl-3"
+        >
+          <Slider
+            :model-value="[canvasWidth]"
+            :min="64"
+            :max="4096"
+            :step="64"
+            class="flex-1"
+            @update:model-value="(v) => v?.length && (canvasWidth = v[0])"
+          />
+          <span class="w-10 text-center text-xs text-node-text-muted">{{
+            canvasWidth
+          }}</span>
+        </div>
+
+        <div
+          class="flex w-28 items-center truncate text-sm text-muted-foreground"
+        >
+          {{ $t('painter.height') }}
+        </div>
+        <div
+          class="flex h-8 items-center gap-2 rounded-lg bg-component-node-widget-background pr-2 pl-3"
+        >
+          <Slider
+            :model-value="[canvasHeight]"
+            :min="64"
+            :max="4096"
+            :step="64"
+            class="flex-1"
+            @update:model-value="(v) => v?.length && (canvasHeight = v[0])"
+          />
+          <span class="w-10 text-center text-xs text-node-text-muted">{{
+            canvasHeight
+          }}</span>
+        </div>
+
+        <div
+          class="flex w-28 items-center truncate text-sm text-muted-foreground"
+        >
+          {{ $t('painter.background') }}
+        </div>
+        <div
+          class="flex h-8 w-full items-center gap-2 rounded-lg bg-component-node-widget-background px-4"
+        >
+          <input
+            type="color"
+            :value="backgroundColorDisplay"
+            class="h-4 w-8 cursor-pointer appearance-none overflow-hidden rounded-full border-none bg-transparent [&::-webkit-color-swatch-wrapper]:p-0 [&::-webkit-color-swatch]:border-none [&::-webkit-color-swatch]:rounded-full [&::-moz-color-swatch]:border-none [&::-moz-color-swatch]:rounded-full"
+            @input="
+              (e) =>
+                (backgroundColorDisplay = (e.target as HTMLInputElement).value)
+            "
+          />
+          <span class="min-w-[4ch] truncate text-xs">{{
+            backgroundColorDisplay
+          }}</span>
+        </div>
+      </template>
+
+      <Button
+        variant="secondary"
+        size="md"
+        :class="
+          cn(
+            'gap-2 rounded-lg border border-component-node-border bg-component-node-background text-xs text-muted-foreground hover:text-base-foreground',
+            !compact && 'col-span-2'
+          )
+        "
+        @click="handleClear"
+      >
+        <i class="icon-[lucide--undo-2]" />
+        {{ $t('painter.clear') }}
+      </Button>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { useElementSize } from '@vueuse/core'
+import { computed, useTemplateRef } from 'vue'
+
+import Button from '@/components/ui/button/Button.vue'
+import Slider from '@/components/ui/slider/Slider.vue'
+import { PAINTER_TOOLS, usePainter } from '@/composables/painter/usePainter'
+import { toHexFromFormat } from '@/utils/colorUtil'
+import { cn } from '@/utils/tailwindUtil'
+
+const { nodeId } = defineProps<{
+  nodeId: string
+}>()
+
+const modelValue = defineModel<string>({ default: '' })
+
+const canvasEl = useTemplateRef<HTMLCanvasElement>('canvasEl')
+const controlsEl = useTemplateRef<HTMLDivElement>('controlsEl')
+const { width: controlsWidth } = useElementSize(controlsEl)
+const compact = computed(
+  () => controlsWidth.value > 0 && controlsWidth.value < 350
+)
+
+const {
+  tool,
+  brushSize,
+  brushColor,
+  brushOpacity,
+  brushHardness,
+  backgroundColor,
+  canvasWidth,
+  canvasHeight,
+  cursorX,
+  cursorY,
+  cursorVisible,
+  displayBrushSize,
+  inputImageUrl,
+  isImageInputConnected,
+  handlePointerDown,
+  handlePointerMove,
+  handlePointerUp,
+  handlePointerEnter,
+  handlePointerLeave,
+  handleInputImageLoad,
+  handleClear
+} = usePainter(nodeId, { canvasEl, modelValue })
+
+const canvasContainerStyle = computed(() => ({
+  aspectRatio: `${canvasWidth.value} / ${canvasHeight.value}`,
+  backgroundColor: isImageInputConnected.value
+    ? undefined
+    : backgroundColor.value
+}))
+
+const cursorStyle = computed(() => {
+  const size = displayBrushSize.value
+  const x = cursorX.value - size / 2
+  const y = cursorY.value - size / 2
+  return {
+    width: `${size}px`,
+    height: `${size}px`,
+    transform: `translate(${x}px, ${y}px)`
+  }
+})
+
+const brushOpacityPercent = computed({
+  get: () => Math.round(brushOpacity.value * 100),
+  set: (val: number) => {
+    brushOpacity.value = val / 100
+  }
+})
+
+const brushHardnessPercent = computed({
+  get: () => Math.round(brushHardness.value * 100),
+  set: (val: number) => {
+    brushHardness.value = val / 100
+  }
+})
+
+const brushColorDisplay = computed({
+  get: () => toHexFromFormat(brushColor.value, 'hex'),
+  set: (val: unknown) => {
+    brushColor.value = toHexFromFormat(val, 'hex')
+  }
+})
+
+const backgroundColorDisplay = computed({
+  get: () => toHexFromFormat(backgroundColor.value, 'hex'),
+  set: (val: unknown) => {
+    backgroundColor.value = toHexFromFormat(val, 'hex')
+  }
+})
+</script>
--- a/src/composables/painter/usePainter.ts
+++ b/src/composables/painter/usePainter.ts
@@ -1,0 +1,780 @@
+import type { Ref } from 'vue'
+import { computed, onMounted, onUnmounted, ref, watch } from 'vue'
+import { useElementSize } from '@vueuse/core'
+import { useI18n } from 'vue-i18n'
+
+import {
+  getEffectiveBrushSize,
+  getEffectiveHardness
+} from '@/composables/maskeditor/brushUtils'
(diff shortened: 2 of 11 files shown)


asagi4:
Would it be difficult to add a "mask only" brush mode (could be a toggle in the color picker to keep the interface uncluttered)? I can see it being useful to add bits to the mask output without actually necessarily painting over the image.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,670 @@
...
+  function getCtx() {
+    return (
+      canvasEl.value?.getContext('2d', { willReadFrequently: true }) ?? null
+    )
+  }
+
+  function getCanvasPoint(e: PointerEvent): Point | null {
+    const el = canvasEl.value
+    if (!el) return null
+    const rect = el.getBoundingClientRect()
```
christian-byrne: We need to avoid querying computed layout tree on every pixel

jtydhr88: fixed

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+    const points = strokeProcessor.addPoint(point)
+    if (points.length === 0 && lastPoint) {
+      points.push(point)
+    }
+
+    if (points.length === 0) return
+
+    strokeCtx.save()
+    applyBrushStyle(strokeCtx)
+
```
christian-byrne, line 448: This is the biggest perf opportunity in the whole PR. Right now every single `pointermove` event goes straight through to `continueStroke()`, which does coordinate transforms, stroke processing, segment drawing, gradient creation, and a full canvas composite. All synchronously, all per event.

The thing is, `pointermove` fires at the input device's polling rate, not the display refresh rate. A typical mouse is 125 Hz, a gaming mouse is 1000 Hz, and pen tablets are usually 200+. But the screen only refreshes at 60-144 Hz. So most of that work produces pixels that never get shown.

The standard fix is to just store the latest event and batch the actual work into a `requestAnimationFrame`:

```ts
let pendingMoveEvent: PointerEvent | null = null
let rafId: number | null = null

function handlePointerMove(e: PointerEvent) {
  updateCursorPos(e)
  if (!isDrawing) return

  pendingMoveEvent = e
  if (!rafId) {
    rafId = requestAnimationFrame(() => {
      if (pendingMoveEvent) continueStroke(pendingMoveEvent)
      rafId = null
      pendingMoveEvent = null
    })
  }
}
```

But the caveat is that you'll lose intermediate points between frames, which can cause angular strokes at low frame rates. If that matters (not sure), you can accumulate events in an array between frames, or use `e.getCoalescedEvents()` in the RAF callback. https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/getCoalescedEvents

But honestly for a node widget painter, the simple version above is probably fine.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+  function syncCanvasSizeFromWidgets() {
+    const w = getWidgetByName('width')
+    const h = getWidgetByName('height')
+    canvasWidth.value = (w?.value as number) ?? 512
+    canvasHeight.value = (h?.value as number) ?? 512
+  }
+
+  function activeHardness(): number {
+    return tool.value === PAINTER_TOOLS.ERASER ? 1 : brushHardness.value
+  }
```
christian-byrne, line 182: Two things here:

**1)** This gets called a lot — every `compositeStrokeToMain`, every `resizeCanvas`, `handleClear`, `isCanvasEmpty`, etc. `getContext('2d', ...)` technically returns the same context each time (browsers cache it per canvas), but the option negotiation still has overhead. Would be cleaner to just grab it once and hold it:

```ts
let mainCtx: CanvasRenderingContext2D | null = null

function getCtx() {
  if (!mainCtx && canvasEl.value) {
    mainCtx = canvasEl.value.getContext('2d') ?? null
  }
  return mainCtx
}
```

(Reset `mainCtx = null` in `resizeCanvas` since the canvas element's dimensions change.)

**2)** More importantly — `willReadFrequently: true` tells the browser to keep this canvas in CPU memory instead of GPU-accelerated. That makes reads (`getImageData`) faster but makes **all drawing operations slower**. This is a drawing-heavy canvas — you're writing to it on every pointermove. The only reads are in `resizeCanvas()` and `isCanvasEmpty()`, which are cold paths.

I'd drop `willReadFrequently` here. If you really need fast reads later, do them on a separate offscreen canvas.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+          drawCircle(ctx, { x, y })
+        }
+      }
+    } else {
+      ctx.beginPath()
+      ctx.moveTo(from.x, from.y)
+      ctx.lineTo(to.x, to.y)
+      ctx.stroke()
+      drawCircle(ctx, to)
+    }
```
christian-byrne, line 308: So this runs on every `continueStroke()` call (which is every pointermove during a stroke):

```ts
if (hasBaseSnapshot && baseCanvas) {
  ctx.clearRect(0, 0, el.width, el.height)   // clear entire canvas
  ctx.drawImage(baseCanvas, 0, 0)             // redraw entire base
}
ctx.drawImage(strokeCanvas, 0, 0)             // draw entire stroke layer
```

That's 2-3 full-canvas blits per pointermove. For a 2048×2048 canvas that's ~50MB of pixel data being shuffled around per event.

The reason you need this is because you're drawing the stroke onto a separate canvas with full opacity, then compositing with `brushOpacity` — so you can't just incrementally add to the main canvas without the opacity stacking.

A couple options:

- **Dirty rect approach:** Track the bounding box of what changed this frame (you'd need to extend `drawSegment` to return the affected region), then only `clearRect` + `drawImage` that region. Canvas `drawImage` supports source/dest rects.

- **Double-buffer swap:** Instead of clear+redraw+composite every frame, keep two full canvases and swap. But this is more complex.

The dirty rect approach is probably the sweet spot — you'd only be blitting a small region around the latest brush stroke instead of the entire canvas.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+    const radius = brushSize.value / 2
+    const effectiveRadius = getEffectiveBrushSize(radius, activeHardness())
+    const effectiveDiameter = effectiveRadius * 2
+    return effectiveDiameter * (el.clientWidth / canvasWidth.value)
+  })
+
+  function getCtx() {
+    return (
+      canvasEl.value?.getContext('2d', { willReadFrequently: true }) ?? null
+    )
```
christian-byrne: Everything inside `drawCircle` is **constant for the entire stroke** — brush size, hardness, color, effective radius — none of these can change while the pointer is down. But this function runs per-dab (and soft brushes can generate dozens of dabs per segment), so we're recomputing all of this every time:

```ts
const radius = brushSize.value / 2                              // same every call
const hardness = activeHardness()                                // same every call
const effectiveRadius = getEffectiveBrushSize(radius, hardness)  // same every call
const effectiveHardness = getEffectiveHardness(...)              // same every call
const { r, g, b } = hexToRgb(brushColor.value)                  // string parsing, same every call
```

Would be a lot cleaner to snapshot these once in `startStroke()`:

```ts
let strokeBrush: {
  radius: number
  effectiveRadius: number
  effectiveHardness: number
  r: number; g: number; b: number
  hardness: number
} | null = null

function startStroke(e: PointerEvent) {
  const radius = brushSize.value / 2
  const hardness = activeHardness()
  const effectiveRadius = getEffectiveBrushSize(radius, hardness)
  strokeBrush = {
    radius,
    effectiveRadius,
    effectiveHardness: getEffectiveHardness(radius, hardness, effectiveRadius),
    ...hexToRgb(brushColor.value),
    hardness
  }
  // ... rest of startStroke
}
```

Then `drawCircle` just reads from `strokeBrush` — no recomputation, no string parsing, no reactive `.value` access in the hot loop.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+
+  function drawCircle(ctx: CanvasRenderingContext2D, point: Point) {
+    const radius = brushSize.value / 2
+    const hardness = activeHardness()
+    const effectiveRadius = getEffectiveBrushSize(radius, hardness)
+    const effectiveHardness = getEffectiveHardness(
+      radius,
+      hardness,
+      effectiveRadius
+    )
```
christian-byrne: The soft brush path here can generate a *lot* of circles:

```ts
const step = Math.max(1, effectiveRadius / 4)
const steps = Math.ceil(dist / step)
for (let i = 1; i <= steps; i++) {
  drawCircle(ctx, { x, y })  // 👈 each one creates a radial gradient
}
```

For a fast drag with a large brush, `dist` can be hundreds of pixels and `step` can be small, so you're creating and filling dozens of radial gradients per pointermove. Each gradient also allocates `{ x, y }` on line 210.

Two quick wins:

1. Bump the step to `effectiveRadius / 2` — you'll barely notice the visual difference but halve the dab count
2. Reuse a point object instead of allocating `{ x, y }` per dab:

```ts
const dabPoint: Point = { x: 0, y: 0 }
for (let i = 1; i <= steps; i++) {
  const t = i / steps
  dabPoint.x = from.x + dx * t
  dabPoint.y = from.y + dy * t
  drawCircle(ctx, dabPoint)
}
```

Also `Math.hypot(dx, dy)` does a square root — if you're only comparing distances or computing step counts, you don't always need the actual distance. But here you do use `dist` for division, so it's fine. Just noting it.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+    ctx.save()
+    applyBrushStyle(ctx)
+    drawCircle(ctx, point)
+    ctx.restore()
+
+    compositeStrokeToMain(true)
+  }
+
+  function continueStroke(e: PointerEvent) {
+    if (!isDrawing || !strokeProcessor || !strokeCtx) return
```
christian-byrne, line 434:
```ts
cursorPos.value = { x: e.offsetX, y: e.offsetY }
```

This allocates a new object on every pointermove, and since `cursorPos` is a `ref`, Vue's reactivity will trigger all downstream computeds/watchers — including `brushLeft` and `brushTop` style bindings on the cursor ring element in the template.

Two options depending on how far you want to go:

**Simple:** Just use two separate refs. Avoids the object allocation and Vue can skip the update if only one axis changed:
```ts
const cursorX = ref(0)
const cursorY = ref(0)

function updateCursorPos(e: PointerEvent) {
  cursorX.value = e.offsetX
  cursorY.value = e.offsetY
}
```

**Better (pairs with RAF throttle):** If you do the RAF batching from the other comment, the cursor update gets batched for free — just update the cursor pos inside the RAF callback alongside the stroke drawing. Then you get exactly one reactive update per frame for both the stroke and the cursor.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+  function startStroke(e: PointerEvent) {
+    const point = getCanvasPoint(e)
+    if (!point) return
+
+    const el = canvasEl.value
+    if (!el) return
+
+    const bCtx = ensureBaseCanvas()
+    if (bCtx) {
+      bCtx.clearRect(0, 0, el.width, el.height)
```
christian-byrne, line 410:
```ts
cachedRect = null
```

You cache the rect on `pointerdown` (good!), but then null it on every stroke end. If the user draws multiple strokes without moving the canvas, you're re-reading `getBoundingClientRect()` at the start of each one for no reason — the element didn't move.

I'd just leave the cache alive and invalidate it on actual layout changes. You could add a `ResizeObserver` or just re-cache on the next `pointerdown` without nulling here. The `pointerdown` in `handlePointerDown` already calls `cacheCanvasRect()`, so removing this line should be safe.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+      bCtx.drawImage(el, 0, 0)
+      hasBaseSnapshot = true
+    }
+
```
christian-byrne, line 414: This does a full `getImageData` → resize → `putImageData` round-trip:

```ts
const prevData = el.width > 0 && el.height > 0
  ? getCtx()?.getImageData(0, 0, el.width, el.height)
  : null
el.width = canvasWidth.value
el.height = canvasHeight.value
if (prevData) {
  getCtx()?.putImageData(prevData, 0, 0)
}
```

`getImageData`/`putImageData` is expensive — it forces a GPU→CPU sync if the canvas is hardware-accelerated (and with `willReadFrequently` removed per my other comment, it would be). The standard approach is to use `drawImage` with an offscreen canvas:

```ts
function resizeCanvas() {
  const el = canvasEl.value
  if (!el) return
  const ctx = getCtx()
  if (!ctx) return

  // Copy current content to a temp canvas
  const tmp = document.createElement('canvas')
  tmp.width = el.width
  tmp.height = el.height
  tmp.getContext('2d')!.drawImage(el, 0, 0)

  // Resize (clears the canvas)
  el.width = canvasWidth.value
  el.height = canvasHeight.value
  mainCtx = null  // force re-acquisition

  // Copy back
  getCtx()?.drawImage(tmp, 0, 0)

  // ... rest of cleanup
}
```

`drawImage` stays on the GPU path and avoids the pixel data round-trip.

christian-byrne (commented):
Hey, overall this is a solid first pass — the architecture is clean, the stroke processor reuse from the mask editor is smart, and the base/stroke canvas separation is the right approach.

I left a bunch of comments, mostly around the pointermove hot path. This thing is going to get hammered at 100-200+ Hz during strokes and right now we're doing a lot of work per event that we don't need to. None of it is blocking but some of these are genuinely big wins, especially the RAF throttling and the gradient recomputation.

Also flagged a couple of the deep watches and the `willReadFrequently` thing — those are more subtle but worth addressing before this lands.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+      applyBrushStyle(strokeCtx)
+      let prev = lastPoint ?? points[0]
+      for (const p of points) {
+        drawSegment(strokeCtx, prev, p)
+        prev = p
+      }
+      strokeCtx.restore()
+    }
+
+    compositeStrokeToMain()
```
christian-byrne, line 476: Nice — you already have the `hasStrokes` early-exit, so the full-canvas scan only runs if there have been actual strokes. That's a good optimization.

But `isCanvasEmpty` is called from `serializeValue`, which runs on widget serialization. If the graph serializes frequently (auto-save, workflow export, etc.), this could be called more often than you'd expect. The full alpha-channel scan on a 2048×2048 canvas is ~4M iterations.

Since you're already tracking `hasStrokes`, could you just trust it?

```ts
function isCanvasEmpty(): boolean {
  return !hasStrokes
}
```

The only case where `hasStrokes` would be wrong is if someone draws then erases everything back to transparent — but in that case, uploading a fully-transparent image is probably fine too. If you want to be more careful, you could set `hasStrokes = false` in `handleClear` (which you already do) and just trust the flag.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+    const blob = await new Promise<Blob | null>((resolve) =>
+      el.toBlob(resolve, 'image/png')
+    )
+    if (!blob) return modelValue.value
+
+    const name = `painter-${nodeId}-${Date.now()}.png`
+    const file = new File([blob], name)
+    const body = new FormData()
+    body.append('image', file)
+    if (!isCloud) body.append('subfolder', 'painter')
```
christian-byrne, line 633: These deep watches are going to be a problem:

```ts
watch(() => nodeOutputStore.nodeOutputs, updateInputImageUrl, { deep: true })
watch(() => nodeOutputStore.nodePreviewImages, updateInputImageUrl, { deep: true })
```

`nodeOutputs` and `nodePreviewImages` are store-wide maps that get mutated whenever **any** node produces output — not just the node connected to this painter. With `deep: true`, Vue traverses the entire object tree on every mutation to check for changes. If there are 50 nodes producing preview images, this watcher fires and traverses for all of them.

Much better to watch only what you actually care about:

```ts
watch(
  () => {
    const node = litegraphNode.value
    if (!node) return null
    const inputNode = node.getInputNode(0)
    if (!inputNode) return null
    return nodeOutputStore.getNodeImageUrls(inputNode)?.[0] ?? null
  },
  () => updateInputImageUrl()
)
```

This only reacts when the specific upstream node's first image URL changes, which is what you actually need. Removes the deep traversal entirely.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+    }
+
+    ctx.fill()
+  }
+
+  function drawSegment(ctx: CanvasRenderingContext2D, from: Point, to: Point) {
+    const hardness = activeHardness()
+
+    if (hardness < 1) {
+      const dx = to.x - from.x
```
christian-byrne, line 286: Same `willReadFrequently` note as the main canvas — this stroke canvas is write-heavy (you're drawing circles and segments onto it every pointermove) and never read via `getImageData`. The only read is via `drawImage` in `compositeStrokeToMain`, which doesn't need `willReadFrequently`.

Dropping this should let the browser keep the stroke canvas GPU-accelerated.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+    const el = canvasEl.value
+    const ctx = getCtx()
+    if (!el || !ctx) return
+    ctx.clearRect(0, 0, el.width, el.height)
+    isDirty.value = true
+    hasStrokes = false
+  }
+
+  function updateCursorPos(e: PointerEvent) {
+    cursorPos.value = { x: e.offsetX, y: e.offsetY }
```
christian-byrne: Tiny thing — `FormData.append` already accepts `Blob` with a filename:

```ts
// Current:
const file = new File([blob], name)
body.append('image', file)

// Simpler:
body.append('image', blob, name)
```

Skips the `File` constructor allocation. The server sees the same multipart payload either way. Not a hot path at all, just cleaner.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+  function handlePointerEnter() {
+    cursorVisible.value = true
+  }
+
+  function handleInputImageLoad(e: Event) {
+    const img = e.target as HTMLImageElement
+    const widthWidget = getWidgetByName('width')
+    const heightWidget = getWidgetByName('height')
+    if (widthWidget) {
+      widthWidget.value = img.naturalWidth
```
christian-byrne, line 582: You're calling `useToastStore()` three times across the catch blocks here (lines 477, 488, 499). Pinia composable calls are cheap but not free — `useToastStore()` looks up the store in the pinia instance map each time.

Just grab it once at the top of the composable alongside `useNodeOutputStore()`:

```ts
const toastStore = useToastStore()  // 👈 once
```

Then use `toastStore.addAlert(...)` in the catch blocks.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+    if (!inputNode) {
+      inputImageUrl.value = null
+      return
+    }
+
+    const urls = nodeOutputStore.getNodeImageUrls(inputNode)
+    inputImageUrl.value = urls?.length ? urls[0] : null
+  }
+
+  function syncCanvasSizeFromWidgets() {
```
christian-byrne, line 173: nit: `displayBrushSize` reads `el.clientWidth` inside a computed:

```ts
return effectiveDiameter * (el.clientWidth / canvasWidth.value)
```

`clientWidth` triggers a synchronous layout read. This computed re-evaluates whenever `brushSize`, `brushHardness`, or `canvasWidth` change — which is fine, those don't change during strokes. But if someone is actively dragging the brush size slider, you'd get layout reads on every slider tick.

Not a big deal since it's not on the pointermove path, but if you ever wanted to tighten it up: you could track the canvas element's display width with `useElementSize` (which uses `ResizeObserver` under the hood — you're already importing it for `controlsEl` in the template).

src/components/painter/WidgetPainter.vue
```diff
@@ -0,0 +1,351 @@
...
+          @pointerup="handlePointerUp"
+          @pointerenter="handlePointerEnter"
+          @pointerleave="handlePointerLeave"
+        />
+        <div
+          v-show="cursorVisible"
+          class="pointer-events-none absolute rounded-full border border-white/80"
+          :style="{
+            width: `${displayBrushSize}px`,
+            height: `${displayBrushSize}px`,
```
christian-byrne: This cursor div's style binding re-creates the object on every pointermove (since `cursorPos` is a dependency):

```vue
:style="{
  width: `${displayBrushSize}px`,
  height: `${displayBrushSize}px`,
  left: `${cursorPos.x - displayBrushSize / 2}px`,
  top: `${cursorPos.y - displayBrushSize / 2}px`
}"
```

Two thoughts:

1. Moving the style to a named `computed` is a minor win — avoids the template-level object construction and makes it easier for Vue to diff.

2. If you really want smooth cursor tracking, consider using `transform: translate(x, y)` instead of `left`/`top`. Transforms are compositor-friendly — the browser can update them without triggering layout. `left`/`top` changes cause layout recalculation.

```vue
:style="cursorStyle"
```
```ts
const cursorStyle = computed(() => ({
  width: `${displayBrushSize.value}px`,
  height: `${displayBrushSize.value}px`,
  transform: `translate(${cursorPos.value.x - displayBrushSize.value / 2}px, ${cursorPos.value.y - displayBrushSize.value / 2}px)`
}))
```

This is what most drawing apps do for cursor rendering — it avoids the layout→paint→composite cycle on every move.

src/components/painter/WidgetPainter.vue
```diff
@@ -0,0 +1,351 @@
...
+    @pointermove.stop
+    @pointerup.stop
+  >
+    <div
+      class="flex min-h-0 flex-1 items-center justify-center overflow-hidden rounded-lg bg-node-component-surface"
+    >
+      <div
+        class="relative max-h-full w-full"
+        :style="{
+          aspectRatio: `${canvasWidth} / ${canvasHeight}`,
```
christian-byrne: Same idea as the cursor — this inline style object is rebuilt whenever any of its dependencies change:

```vue
:style="{
  aspectRatio: `${canvasWidth} / ${canvasHeight}`,
  backgroundColor: isImageInputConnected ? undefined : backgroundColor
}"
```

This one is less critical since `canvasWidth`/`canvasHeight`/`backgroundColor` only change on slider interaction (not every pointermove), but it'd be cleaner as a named computed just to keep the template simple.

src/composables/painter/usePainter.ts
```diff
@@ -0,0 +1,679 @@
...
+    ctx.lineWidth = effectiveDiameter
+    ctx.lineCap = 'round'
+    ctx.lineJoin = 'round'
+  }
+
+  function ensureStrokeCanvas() {
+    const el = canvasEl.value
+    if (!el) return null
+
+    if (
```
christian-byrne, line 328:
```ts
strokeProcessor = new StrokeProcessor(Math.max(1, brushSize.value / 4))
```

Minor — you're allocating a new `StrokeProcessor` per stroke. That's totally fine, but if you ever see allocation pressure during rapid short strokes (like stippling), you could reuse the instance and just reset it:

```ts
if (!strokeProcessor || strokeProcessor.stepSize !== step) {
  strokeProcessor = new StrokeProcessor(step)
} else {
  strokeProcessor.reset()
}
```

Not worth doing unless profiling shows it matters. Just noting it.

src/composables/painter/usePainter.ts
jtydhr88: this particular optimization would break certain node chains. For example, with Load Image → Image Crop → Painter, getInputNode(0) returns the Image Crop node, but the reactive dependency tracking in the watch getter may not correctly capture when Image Crop's output updates in the store — resulting in the painter showing Load Image's output instead of Image Crop's cropped result.

src/components/painter/WidgetPainter.vue
jtydhr88: fixed

src/components/painter/WidgetPainter.vue
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
jtydhr88: fixed

src/composables/painter/usePainter.ts
christian-byrne: fair enough, that makes sense - getInputNode(0) wouldn't track intermediate node outputs correctly in chains like Load Image → Image Crop → Painter. the deep watch is the safe approach here.

src/components/painter/WidgetPainter.vue
```diff
@@ -0,0 +1,351 @@
...
+          compact ? 'grid-cols-1' : 'grid-cols-[auto_1fr]'
+        )
+      "
+    >
+      <div
+        v-if="!compact"
+        class="flex w-28 items-center truncate text-sm text-muted-foreground"
+      >
+        {{ $t('painter.tool') }}
+      </div>
```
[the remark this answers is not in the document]
christian-byrne: @coderabbitai Make an Issue to do this in a followup PR on a future date

src/components/painter/WidgetPainter.vue
```diff
@@ -0,0 +1,351 @@
...
+                  const val = Math.min(
+                    100,
+                    Math.max(0, Number((e.target as HTMLInputElement).value))
+                  )
+                  brushOpacityPercent = val
+                  ;(e.target as HTMLInputElement).value = String(val)
+                }
+              "
+            />%</span
+          >
```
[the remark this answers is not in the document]
christian-byrne: @coderabbitai Make an Issue to do this in a followup PR on a future date

christian-byrne (approved):
LGTM. I created followup issues for all things not addressed in review.

PR #8523, opened 2026-02-01
fix: dedupe queueStore.update() to prevent race conditions

christian-byrne:
## Summary

Prevents stale API responses from overwriting fresh queue state during rapid consecutive `update()` calls.

## Problem

When `queueStore.update()` is called multiple times in quick succession (e.g., during websocket reconnection), responses could resolve out-of-order. A stale response resolving after a fresh one would overwrite the current state with outdated data.

## Solution

Added request ID tracking that:
1. Increments a counter on each `update()` call
2. Guards all state mutations with a staleness check
3. Only sets `isLoading=false` for the latest request

This pattern matches the existing approach in `jobOutputCache.ts`.

## Changes

- `src/stores/queueStore.ts`: Added `updateRequestId` counter and staleness guards
- `src/stores/queueStore.test.ts`: Added tests for deduplication behavior

## Testing

- All 49 existing tests pass
- Added 3 new tests for race condition handling:
  - Stale responses are discarded
  - isLoading state is correctly managed
  - Stale request failures don't affect latest state

## Fixes COM-12784

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8523-fix-dedupe-queueStore-update-to-prevent-race-conditions-2fa6d73d3650810daa0dc63af359e1ea) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/queueStore.test.ts
+++ b/src/stores/queueStore.test.ts
@@ -49,6 +49,9 @@ const createTaskOutput = (
   }
 })
 
+type QueueResponse = { Running: JobListItem[]; Pending: JobListItem[] }
+type QueueResolver = (value: QueueResponse) => void
+
 // Mock API
 vi.mock('@/scripts/api', () => ({
   api: {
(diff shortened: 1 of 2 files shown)


src/stores/queueStore.ts
```diff
@@ -545,7 +551,9 @@ export const useQueueStore = defineStore('queue', () => {
         return existing
       })
     } finally {
-      isLoading.value = false
+      if (requestId === updateRequestId) {
+        isLoading.value = false
```
DrJKL, line 558: This seems like it could get us stuck in a perpetual loading state.

DrJKL: Not super likely, but all it would take is a slight tweak above, like tightening the scope of the try block or splitting it.

src/stores/queueStore.test.ts
```diff
@@ -802,4 +802,136 @@ describe('useQueueStore', () => {
...
+        Pending: JobListItem[]
+      }>((resolve) => {
+        resolveFirst = resolve
+      })
+      const secondQueuePromise = new Promise<{
+        Running: JobListItem[]
+        Pending: JobListItem[]
+      }>((resolve) => {
+        resolveSecond = resolve
+      })
```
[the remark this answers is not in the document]
DrJKL: Agree with cleaning up the types in here. They deserve a name.

DrJKL (dismissed):

PR #8524, opened 2026-02-01
fix: add Frame Nodes to core menu items for multi-selection context menu

christian-byrne:
## Summary

Fix "Frame Nodes" option being incorrectly categorized as an extension item instead of appearing in the core menu section when multiple nodes are selected.

## Changes

- **What**: Added "Frame Nodes" to `CORE_MENU_ITEMS` set and `MENU_ORDER` array in contextMenuConverter.ts, and added frame equivalents to duplicate detection to prevent duplicate menu items

## Review Focus

The fix ensures the Vue-based "Frame Nodes" menu option (i18n key `g.frameNodes`) is recognized as a core menu item alongside the legacy LiteGraph "Frame selection" label. Both are now treated as equivalent for deduplication.

Fixes COM-13922

N/A - menu item positioning fix

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8524-fix-add-Frame-Nodes-to-core-menu-items-for-multi-selection-context-menu-2fa6d73d365081989799f7bc74c65868) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/contextMenuConverter.test.ts
+++ b/src/composables/graph/contextMenuConverter.test.ts
@@ -134,6 +134,27 @@ describe('contextMenuConverter', () => {
       // Node Info (section 4) should come before or with Color (section 4)
       expect(getIndex('Node Info')).toBeLessThanOrEqual(getIndex('Color'))
     })
+
+    it('should recognize Frame Nodes as a core menu item', () => {
+      const options: MenuOption[] = [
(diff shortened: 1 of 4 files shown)


christian-byrne:
Currently it is only shown on the context menu from the toolbox.

Myestery (approved):

PR #8530, opened 2026-02-01
fix: properly parse PNG iTXt chunks per specification

### LINKED ISSUE #8150: PNG iTXt parsing is not compliant with specification
### Prerequisites

### What happened?

From the website (https://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.iTXt):

> This chunk is semantically equivalent to the tEXt and zTXt chunks, but the textual data is in the UTF-8 encoding of the Unicode character set instead of Latin-1. This chunk contains:
> ```
> Keyword:             1-79 bytes (character string)
> Null separator:      1 byte
> Compression flag:    1 byte
> Compression method:  1 byte
> Language tag:        0 or more bytes (character string)
> Null separator:      1 byte
> Translated keyword:  0 or more bytes
> Null separator:      1 byte
> Text:                0 or more bytes
> ```

Currently, the iTXt parser ignores the compression flag, method, and 2 null terminators:

https://github.com/Comfy-Org/ComfyUI_frontend/blob/30907f99f12034842b88b779f8c5f13fdc230cba/src/scripts/metadata/png.ts#L31-L35

Example PNG that follows the specification:

Image

Viewed on a hex editor, there are more NULL bytes after the `prompt\x00` keyword:

Image

When attempting to load, it produces an error due to the 4 leading NULL bytes:

Image

Viewing on a debugger confirms the leading 4 null bytes:

Image

### Steps to Reproduce

1. Load the example PNG
2. Workflow fails to load

### ComfyUI Frontend Version

1.36.14

### Browser

Firefox

### Console Errors

```javascript
Failed to parse prompt: SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
    handleFile app.ts:1488
    addDropHandler app.ts:548
```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-8150-PNG-iTXt-parsing-is-not-compliant-with-specification-2ec6d73d365081238507f47cee673730) by [Unito](https://www.unito.io)

christian-byrne:
## Summary
Fixes PNG iTXt chunk parsing to comply with the PNG specification.

## Problem
The current iTXt parser incorrectly reads the text content immediately after the keyword null terminator, but iTXt chunks have additional fields:
- Compression flag (1 byte)
- Compression method (1 byte)
- Language tag (null-terminated)
- Translated keyword (null-terminated)
- Text content

This caused PNGs that correctly follow the spec to fail loading with JSON parse errors due to leading null bytes.

## Solution
Skip the compression flag, method, language tag, and translated keyword fields before reading the text content.

Fixes #8150

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8530-fix-properly-parse-PNG-iTXt-chunks-per-specification-2fa6d73d36508189bef4cc5fa3899096) by [Unito](https://www.unito.io)


christian-byrne (commented):
This needs a test. Thankfully we have tests for normal PNGs to ensure this doesn't introduce a regression.

AustinMroz (dismissed):

christian-byrne:
Manually confirmed that:

- when loading the test asset added in this PR to current `master` ComfyUI, it gives "failed to parse workflow" error toast canvas doesn't change
- When loading the test asset added in this PR to this branch on dev server, it loads the workflow

The issue is the test asset's workflow looks the same as the default workflow, which might be what's already on the graph, so if there's a divergence, we might not know. Going to update the test asset to have a more unique workflow embedded. 

---

_edit_: Done!

PR #8533, opened 2026-02-01
feat: add aspect ratio lock for ImageCrop widget

jtydhr88:
## Summary

Add ratio preset dropdown (1:1, 3:4, 4:3, 16:9, 9:16, custom) and lock toggle to the ImageCrop widget. When locked, only corner handles are shown and resizing maintains the selected aspect ratio using diagonal projection for smooth, jump-free interaction. 
Locking with custom selected captures the current crop's ratio.

https://github.com/user-attachments/assets/a7b5c0a0-c18c-4785-940f-59793702e892

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8533-feat-add-aspect-ratio-lock-for-ImageCrop-widget-2fa6d73d365081a98546e43ed0d7d4fe) by [Unito](https://www.unito.io)

## Summary by CodeRabbit

* **New Features**
  * Image crop widget adds a ratio selector with preset and custom options, plus a lock/unlock toggle to preserve aspect while cropping.
  * Constrained corner-resize behavior enforces locked aspect ratios during resizing.

* **Documentation**
  * Added UI text entries for ratio controls (labels for ratio, lock/unlock, custom).

### DIFF
--- a/src/composables/useImageCrop.ts
+++ b/src/composables/useImageCrop.ts
@@ -22,14 +22,14 @@ const CORNER_SIZE = 10
 const MIN_CROP_SIZE = 16
 const CROP_BOX_BORDER = 2
 
-export const ASPECT_RATIOS: Record<string, number | null> = {
+export const ASPECT_RATIOS = {
   '1:1': 1,
   '3:4': 3 / 4,
   '4:3': 4 / 3,
   '16:9': 16 / 9,
   '9:16': 9 / 16,
   custom: null
-}
+} as const
 
 interface UseImageCropOptions {
   imageEl: Ref<HTMLImageElement | null>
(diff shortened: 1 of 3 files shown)


src/composables/useImageCrop.ts
```diff
@@ -22,6 +22,15 @@ const CORNER_SIZE = 10
...
 const CROP_BOX_BORDER = 2
 
+export const ASPECT_RATIOS: Record<string, number | null> = {
+  '1:1': 1,
+  '3:4': 3 / 4,
+  '4:3': 4 / 3,
+  '16:9': 16 / 9,
+  '9:16': 9 / 16,
+  custom: null
+}
```
DrJKL: Nit: I think this'll provide better typechecking/autocomplete.

```suggestion
export const ASPECT_RATIOS = {
  '1:1': 1,
  '3:4': 3 / 4,
  '4:3': 4 / 3,
  '16:9': 16 / 9,
  '9:16': 9 / 16,
  custom: null
} as const
```

DrJKL: Also, what do you think about using `0` instead of null for the custom case?

src/composables/useImageCrop.ts
```diff
@@ -88,6 +97,47 @@ export function useImageCrop(nodeId: NodeId, options: UseImageCropOptions) {
   const resizeStartCropWidth = ref(0)
   const resizeStartCropHeight = ref(0)
 
+  const selectedRatio = ref<string>('custom')
+  const isLockEnabled = ref(false)
+  const lockedRatio = ref<number | null>(null)
+
```
DrJKL: I feel like there's probably a way to unify these into a single ref and some computeds.
That way you wouldn't need the watchers.

src/composables/useImageCrop.ts
jtydhr88: I'd prefer to keep null — it clearly expresses "no ratio constraint," and 0 as an aspect ratio is mathematically meaningless (width/height = 0). The null checks throughout the code (== null) read naturally as "if no ratio is set." Using 0 would lose that clarity and TypeScript's ability to enforce handling the unconstrained case.

src/composables/useImageCrop.ts
jtydhr88: fixed

src/composables/useImageCrop.ts
```diff
@@ -88,6 +97,55 @@ export function useImageCrop(nodeId: NodeId, options: UseImageCropOptions) {
...
 
+  const lockedRatio = ref<number | null>(null)
+
+  const selectedRatio = computed({
+    get: () => {
+      if (lockedRatio.value == null) return 'custom'
+      const entry = Object.entries(ASPECT_RATIOS).find(
+        ([, v]) => v === lockedRatio.value
+      )
+      return entry ? entry[0] : 'custom'
```
DrJKL, line 108: Completely optional, but:
```suggestion
      return entry?.[0] ?? 'custom'
```

src/composables/useImageCrop.ts
```diff
@@ -88,6 +97,55 @@ export function useImageCrop(nodeId: NodeId, options: UseImageCropOptions) {
   const resizeStartCropWidth = ref(0)
   const resizeStartCropHeight = ref(0)
 
+  const lockedRatio = ref<number | null>(null)
+
+  const selectedRatio = computed({
+    get: () => {
+      if (lockedRatio.value == null) return 'custom'
+      const entry = Object.entries(ASPECT_RATIOS).find(
```
DrJKL, line 105: Having to loop through entries in a map like this makes me think the map is backwards.

src/composables/useImageCrop.ts
```diff
@@ -200,7 +258,9 @@ export function useImageCrop(nodeId: NodeId, options: UseImageCropOptions) {
     }
   }
 
-  const resizeHandles = computed<ResizeHandle[]>(() => {
+  const CORNER_DIRECTIONS = new Set<ResizeDirection>(['nw', 'ne', 'sw', 'se'])
```
DrJKL, line 261: You could do a `const` array, then a derived `ReadonlySet` from the values.

DrJKL (approved):

PR #8537, opened 2026-02-02
feat: show legacy manager search tip in no-results empty state

christian-byrne:
## Summary

Show a non-intrusive tip in the no-results empty state when users search legacy manager-related terms in the new manager UI.

## Changes

- **What**: Add `useLegacySearchTip` composable that detects when search query matches legacy manager keywords ("manager", "comfyui-manager", etc.) and appends a tip to the empty state message suggesting the `--enable-manager-legacy-ui` flag
- **Integration**: Tip appears as secondary muted text below "Try a different search" message when no results found
- **Tests**: 9 test cases covering detection logic and edge cases

## Review Focus

- Keyword list covers common legacy manager search terms
- Non-intrusive approach: tip only shows in no-results state, no dismiss button needed

Fixes COM-12509


viva-jinyi (approved):
LGTM

PR #8542, opened 2026-02-02
feat: code splitting optimization - reduce initial bundle by 36%

DrJKL:
## Summary

Reduces initial module preload from 12.94 MB to 8.24 MB (-4.7 MB, -36%).

## Changes

- Split vendor chunks for better cache isolation (firebase, sentry, i18n, zod, etc.)
- Exclude heavy optional chunks from initial preload: THREE.js, xterm, tiptap, chart.js, yjs
- Lazy load 16 dialog components in dialogService
- Add \endor-yjs\ chunk for CRDT library

## Metrics

| Metric | Before | After |
|--------|--------|-------|
| Initial preload | 12.94 MB | 8.24 MB |
| vendor-other | 4,006 KB | 2,156 KB |
| dialogService | 1,860 KB | 1,304 KB |

All excluded chunks still load on-demand when their features are used.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8542-feat-code-splitting-optimization-reduce-initial-bundle-by-36-2fb6d73d36508146aaf7fdaed3274033) by [Unito](https://www.unito.io)


christian-byrne:
How did the overall bundle decrease 2MB?

src/services/load3dService.ts
```diff
@@ -1,11 +1,76 @@
...
+  [key: string]: unknown
+}
+
+// Type for SkeletonUtils module
+type SkeletonUtilsModule = { clone: (source: Object3D) => Object3D }
+
+// Cache for lazy-loaded modules
+let cachedNodeToLoad3dMap: Map<LGraphNode, Load3d> | null = null
+let cachedUseLoad3dViewer: UseLoad3dViewerFn | null = null
+let cachedSkeletonUtils: SkeletonUtilsModule | null = null
```
christian-byrne, line 44: question: Do other big codebases have a better pattern for this? Perhaps something like a dependency container that you can inject things in to, which has a global state and makes the typing and access patterns a lot cleaner and avoids mutable state?

vite.config.mts
```diff
@@ -419,6 +419,31 @@ export default defineConfig({
     minify: SHOULD_MINIFY,
     target: 'es2022',
     sourcemap: GENERATE_SOURCEMAP,
+    // Exclude heavy optional vendor chunks from initial module preload
+    // These chunks are only needed when their features are used (3D, terminal, etc.)
```
christian-byrne, line 423: note (no action item): I think deferring imports of a lot of chunks could exacerbate the stale chunk issue a lot (client with stale version's index.html is making requests for stale chunks across a longer timespan rather than requesting upfront). But then again, xterm/firebase/sentry/threejs we can assume are unchanging across release. Hopefully this can go out around the same time we finish the new cloud frontend deployment impl. (which will involve holding old chunks for some period of time).

vite.config.mts
```diff
@@ -442,52 +467,101 @@ export default defineConfig({
...
       output: {
         keepNames: true,
         codeSplitting: {
           groups: [
+            // Framework core - highest priority, very stable
+            {
+              name: 'vendor-vue-core',
+              test: /[\\/]node_modules[\\/](vue|@vue|pinia|vue-router)[\\/]/,
+              priority: 20
+            },
```
christian-byrne, line 479: nit/question: I sort of agreed with amp on this that it might not make sense to have a separate chunk for vue when it's <20kB, WDYT?

vite.config.mts
```diff
@@ -442,52 +467,101 @@ export default defineConfig({
...
+
+            // Cloud services - large, stable per version
```
christian-byrne: nit: Firebase also ships on local, since we have partner nodes that need auth. Sentry also ships on local (previously was used for feedback buttons, now still used for desktop distro). I believe we could drop Sentry from local dist for further optimization.

christian-byrne (dismissed):

christian-byrne:
I also saw in your prompt that you (or amp by themself) was aware of the issue with all chunks being invalidated across every release due to something like a library adding a timestamp somewhere. Did you get to that in this PR? I don't see anything related. So we should make a separate task for it?

vite.config.mts
DrJKL: Removed the comment. The actual pruning is DCE based.

src/services/load3dService.ts
DrJKL: "There's got to be a better way!"

vite.config.mts
DrJKL: No strong opinion there. I'd want to see if it affected the page load/initialization in real conditions.

browser_tests/tests/mobileBaseline.spec.ts-snapshots/mobile-settings-dialog-mobile-chrome-linux.png
DrJKL, line 1: Order change because of the lazy loading.

vite.config.mts
christian-byrne: Well I guess this is better, since we will cache this for a really long time, so it's just 1 extra request every few months at the benefit of 20kB less load.

christian-byrne (approved):
LGTM!

PR #8570, opened 2026-02-03
refactor: move cancellable states to module-level constant

christian-byrne:
## Summary

Moves static data outside the `useJobActions` composable to module-level.

## Changes

- **What**: Extracted `cancellableStates` array to a module-level `CANCELLABLE_STATES` constant. The `cancelAction` object remains inside the composable because it depends on `t()` for i18n reactivity.

## Review Focus

This is a pure refactor with no behavior change - the constant is now allocated once per module instead of per-composable-call.

Fixes #7946

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8570-refactor-move-cancellable-states-to-module-level-constant-2fc6d73d3650814fa1fefa172f7ace2f) by [Unito](https://www.unito.io)

## Summary by CodeRabbit

* **Chores**
  * Internal code refactoring to improve maintainability of job cancellation logic.

### DIFF
--- a/src/composables/queue/useJobActions.ts
+++ b/src/composables/queue/useJobActions.ts
@@ -13,6 +13,8 @@ export type JobAction = {
   variant: 'destructive' | 'secondary' | 'textonly'
 }
 
+const CANCELLABLE_STATES: JobState[] = ['pending', 'initialization', 'running']
+
 export function useJobActions(
   job: MaybeRefOrGetter<JobListItem | null | undefined>
 ) {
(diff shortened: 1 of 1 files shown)


src/composables/queue/useJobActions.ts
```diff
@@ -13,6 +13,8 @@ export type JobAction = {
   variant: 'destructive' | 'secondary' | 'textonly'
 }
 
+const CANCELLABLE_STATES: JobState[] = ['pending', 'initialization', 'running']
```
DrJKL, line 16: This would give better type hinting.
```suggestion
const CANCELLABLE_STATES = ['pending', 'initialization', 'running'] as const satisfies JobState[]
```

DrJKL (approved):

src/composables/queue/useJobActions.ts
christian-byrne: Whoops I didn't see it in time.

PR #8592, opened 2026-02-03
feat: add workflow execution state indicators to tabs

christian-byrne:
## Summary

Add visual indicators to workflow tabs showing execution state (running, completed, error).

## Changes

- **executionStore**: Added `WorkflowExecutionResult` type and `lastExecutionResultByWorkflowId` state to track execution results per workflow
- **useWorkflowExecutionState**: New composable that derives execution state (idle/running/completed/error) from store data
- **WorkflowExecutionIndicator**: New component displaying 16x16 icons with animations and semantic colors
- **WorkflowTab**: Integrated indicator with 5-second auto-clear for completed state
- **WorkflowsSidebarTab**: Added execution badge to sidebar workflow list via WorkflowExecutionBadge wrapper

## Design

- 🔄 Running: `loader-circle` icon with `animate-spin`, `text-muted-foreground`
- ✅ Completed: `circle-check` icon, `text-jade-600`, auto-hides after 5 seconds
- ❌ Error: `circle-alert` icon, `text-coral-600`, persists until next execution

## Review Focus

- State derivation logic in `useWorkflowExecutionState` - correctly maps prompt IDs to workflow IDs
- Cleanup of timeout in `WorkflowTab.vue` on unmount
- Test coverage for both the composable and indicator component

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8592-feat-add-workflow-execution-state-indicators-to-tabs-2fc6d73d365081bc855bf58899cc6067) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/topbar/WorkflowExecutionIndicator.vue
+++ b/src/components/topbar/WorkflowExecutionIndicator.vue
@@ -1,0 +1,25 @@
+<template>
+  <i
+    v-if="state === 'running'"
+    class="icon-[lucide--loader-circle] size-4 shrink-0 animate-spin text-muted-foreground"
+    :aria-label="$t('workflowExecution.running')"
+  />
+  <i
+    v-else-if="state === 'completed'"
+    class="icon-[lucide--circle-check] size-4 shrink-0 text-jade-600"
+    :aria-label="$t('workflowExecution.completed')"
+  />
+  <i
+    v-else-if="state === 'error'"
+    class="icon-[lucide--circle-alert] size-4 shrink-0 text-coral-600"
+    :aria-label="$t('workflowExecution.error')"
+  />
+</template>
+
+<script setup lang="ts">
+import type { WorkflowExecutionState } from '@/stores/executionStore'
+
+const { state } = defineProps<{
+  state: WorkflowExecutionState
+}>()
+</script>
--- a/src/components/topbar/WorkflowTab.vue
+++ b/src/components/topbar/WorkflowTab.vue
@@ -20,6 +20,10 @@
         >
           {{ workflowOption.workflow.filename }}
         </span>
+        <WorkflowExecutionIndicator
+          v-if="showExecutionIndicator"
+          :state="executionState"
+        />
         <div class="relative">
           <span
             v-if="shouldShowStatusIndicator"
@@ -68,11 +72,12 @@ import {
   ContextMenuSeparator,
   ContextMenuTrigger
 } from 'reka-ui'
-import { computed, onUnmounted, ref } from 'vue'
+import { computed, onUnmounted, ref, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 import WorkflowActionsList from '@/components/common/WorkflowActionsList.vue'
 import Button from '@/components/ui/button/Button.vue'
+import { useWorkflowExecutionState } from '@/composables/useWorkflowExecutionState'
 import {
   usePragmaticDraggable,
   usePragmaticDroppable
@@ -87,6 +92,7 @@ import { useCommandStore } from '@/stores/commandStore'
 import { useWorkspaceStore } from '@/stores/workspaceStore'
 import type { WorkflowMenuItem } from '@/types/workflowMenuItem'
 
+import WorkflowExecutionIndicator from './WorkflowExecutionIndicator.vue'
 import WorkflowTabPopover from './WorkflowTabPopover.vue'
 
 interface WorkflowOption {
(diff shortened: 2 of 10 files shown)


src/components/topbar/WorkflowExecutionIndicator.vue
```diff
@@ -0,0 +1,29 @@
...
+<script setup lang="ts">
+import { useI18n } from 'vue-i18n'
+
+import type { WorkflowExecutionState } from '@/composables/useWorkflowExecutionState'
+
+defineProps<{
+  state: WorkflowExecutionState
+}>()
+
+const { t } = useI18n()
```
christian-byrne: If only using `t` in the `template`, you can just use `$t`, as it is injected into the compiler. So importing something is not needed either. @coderabbitai can you add this to our learnings?

src/components/topbar/WorkflowTab.vue
```diff
@@ -76,6 +82,33 @@ const workflowTabRef = ref<HTMLElement | null>(null)
...
+
+  if (newState === 'completed') {
+    clearTimeoutId.value = setTimeout(() => {
+      clearResult()
+      clearTimeoutId.value = null
+    }, 5000)
+  }
+})
+
+const showExecutionIndicator = computed(() => executionState.value !== 'idle')
```
christian-byrne, line 150: Why would we need a timeout? We have other WS messages to indicate a job is completed -- that same system (and shared centralized services/stores) should be used in the same way they are already used in the repo to determine the state of a job. There's no need to repeat work. Check the queueStore and entire Assets / Jobs / QueueProgress / Assets Side Panel domains for more details.

src/composables/useWorkflowExecutionState.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  const state = computed<WorkflowExecutionState>(() => {
+    const wid = toValue(workflowId)
+    if (!wid) return 'idle'
+
+    for (const promptId of executionStore.runningPromptIds) {
+      if (executionStore.promptIdToWorkflowId.get(promptId) === wid) {
+        return 'running'
+      }
+    }
```
christian-byrne: This should belong to the execution store -- knowing which workflow is running is generally useful and will be accessed across the app. The state should not be managed in a composable here which can have multiple instances.

src/stores/executionStore.ts
```diff
@@ -387,17 +417,28 @@ export const useExecutionStore = defineStore('execution', () => {
...
+    const wid = promptIdToWorkflowId.value.get(pid)
+    if (wid) {
+      const next = new Map(lastExecutionResultByWorkflowId.value)
+      next.set(wid, {
+        state: 'error',
+        timestamp: Date.now(),
+        promptId: pid
+      })
+      lastExecutionResultByWorkflowId.value = next
+    }
```
christian-byrne: This inline logic is repeated twice.

christian-byrne:
Addressed all review comments:

1. **WorkflowExecutionIndicator.vue** - Now uses `$t` directly in template instead of importing `useI18n`
2. **executionStore.ts** - Moved state derivation logic into `getWorkflowExecutionState()` function; extracted duplicate result-setting code into `setWorkflowExecutionResult()` helper
3. **useWorkflowExecutionState.ts** - Now delegates to store's `getWorkflowExecutionState()` instead of containing its own logic
4. **WorkflowTab.vue** - Removed manual 5-second timeout; state is now managed entirely via WS messages

The composable remains for convenience in components but the core logic is centralized in the store as suggested.

christian-byrne:
Note: this is currently non-functional until https://github.com/Comfy-Org/cloud/pull/2364 is merged on cloud and the equivalent on ComfyUI. It can still be reviewed though.

christian-byrne:
@coderabbitai review

benceruleanlu:
Reviewing now

src/stores/queueStore.ts
```diff
@@ -556,6 +557,31 @@ export const useQueueStore = defineStore('queue', () => {
...
+        const state =
+          job.status === 'failed'
+            ? 'error'
+            : job.status === 'completed'
+              ? 'completed'
+              : undefined
+        if (!state) continue
+        const existing =
+          executionStore.lastExecutionResultByWorkflowId.get(workflowId)
+        const jobTimestamp = job.create_time * 1000
```
benceruleanlu, line 575: nit: I think the rest of the codebase treats job.create_time like milliseconds but we multiply by 1000 here while comparing with existing.timestamp that is also in milliseconds (Date.now)

benceruleanlu (approved):
small concern LGTM otherwise

PR #8594, opened 2026-02-03
feat: add WidgetValueStore for centralized widget value management

DrJKL:
## Summary

Implements Phase 1 of the **Vue-owns-truth** pattern for widget values. Widget values are now canonical in a Pinia store; `widget.value` delegates to the store while preserving full backward compatibility.

## Changes

- **New store**: `src/stores/widgetValueStore.ts` - centralized widget value storage with `get/set/remove/removeNode` API
- **BaseWidget integration**: `widget.value` getter/setter now delegates to store when widget is associated with a node
- **LGraphNode wiring**: `addCustomWidget()` automatically calls `widget.setNodeId(this.id)` to wire widgets to their nodes
- **Test fixes**: Added Pinia setup to test files that use widgets

## Why

This foundation enables:
- Vue components to reactively bind to widget values via `computed(() => store.get(...))`
- Future Yjs/CRDT backing for real-time collaboration
- Cleaner separation between Vue state and LiteGraph rendering

## Backward Compatibility

| Extension Pattern | Status |
|-------------------|--------|
| `widget.value = x` | ✅ Works unchanged |
| `node.widgets[i].value` | ✅ Works unchanged |
| `widget.callback` | ✅ Still fires |
| `node.onWidgetChanged` | ✅ Still fires |

## Testing

- ✅ 4252 unit tests pass
- ✅ Build succeeds

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8594-feat-add-WidgetValueStore-for-centralized-widget-value-management-2fc6d73d36508160886fcb9f3ebd941e) by [Unito](https://www.unito.io)


src/lib/litegraph/src/widgets/BaseWidget.ts
```diff
@@ -66,17 +72,49 @@ export abstract class BaseWidget<
...
   last_y?: number
   width?: number
-  disabled?: boolean
   computedDisabled?: boolean
-  hidden?: boolean
-  advanced?: boolean
-  promoted?: boolean
   tooltip?: string
+
+  private _state: WidgetState
```
DrJKL: There are likely more properties that we'll want to move into the store.

src/stores/widgetValueStore.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Strips graph/subgraph prefixes from a scoped node ID to get the bare node ID.
+ * e.g., "graph1:subgraph2:42" → "42"
+ */
+export function stripGraphPrefix(scopedId: NodeId | string): NodeId {
+  return String(scopedId).replace(/^(.*:)+/, '') as NodeId
+}
+
+export interface WidgetState<
```
DrJKL, line 25: I'm hoping to be able to remove this and instead compose types soon.
This should also let us eliminate some redundant interfaces floating around in various modules.

src/lib/litegraph/src/LGraph.ts
```diff
@@ -914,6 +914,16 @@ export class LGraph
       }
     }
 
+    // Register all widgets with the WidgetValueStore now that node has a valid ID.
+    // Widgets added before the node was in the graph deferred their setNodeId call.
+    if (node.widgets) {
```
AustinMroz, line 967: Node is configured and has `node.id` reassigned after being added to the graph.

Worth keeping an eye on, but the code here is correct. If any issues arise, it's a bug that needs to be handled elsewhere.

AustinMroz (commented):
Code looks great and everything seems to work from my quick testing, but I'll want to thoroughly red team this before giving formal approval.

src/stores/widgetValueStore.ts
```diff
@@ -0,0 +1,79 @@
...
+      .filter(([key]) => key.startsWith(prefix))
+      .map(([, state]) => state)
+  }
+
+  function getWidget(
+    nodeId: NodeId,
+    widgetName: string
+  ): WidgetState | undefined {
+    return widgetStates.value.get(makeKey(nodeId, widgetName))
+  }
```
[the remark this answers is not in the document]
DrJKL: Decent suggestion, but not yet.

src/lib/litegraph/src/widgets/BaseWidget.ts
```diff
@@ -97,13 +136,22 @@ export abstract class BaseWidget<
...
+  /**
+   * Associates this widget with a node ID and registers it in the WidgetValueStore.
+   * Once set, value reads/writes will be delegated to the store.
+   */
+  setNodeId(nodeId: NodeId): void {
+    this._state = useWidgetValueStore().registerWidget({
+      ...this._state,
+      nodeId
+    })
   }
```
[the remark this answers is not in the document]
DrJKL: Bad suggestion, removing widgets is too overloaded. It's called on subgraph conversion.

browser_tests/tests/subgraph-duplicate-ids.spec.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    const result = await comfyPage.page.evaluate(() => {
+      const graph = window.app!.canvas.graph!
+      // TODO: Extract allGraphs accessor (root + subgraphs) into LGraph
+      // TODO: Extract allNodeIds accessor into LGraph
+      const allGraphs = [graph, ...graph.subgraphs.values()]
+      const allIds = allGraphs
+        .flatMap((g) => g._nodes)
+        .map((n) => n.id)
+        .filter((id): id is number => typeof id === 'number')
```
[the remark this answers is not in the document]
DrJKL: Do more investigation.

src/extensions/core/previewAny.ts
```diff
@@ -71,7 +73,7 @@ useExtensionService().registerExtension({
           : onExecuted.apply(this, [message])
 
         const previewWidgets =
-          this.widgets?.filter((w) => w.name === 'preview') ?? []
+          this.widgets?.filter((w) => w.name.startsWith('preview_')) ?? []
```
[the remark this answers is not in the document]
DrJKL: No.

browser_tests/tests/subgraph-duplicate-ids.spec.ts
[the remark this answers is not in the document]
DrJKL: It's kind of impressive how many things are wrong in here.

src/stores/widgetValueStore.ts
```diff
@@ -0,0 +1,79 @@
...
+    widgetStates.value.set(key, state)
+    return widgetStates.value.get(key) as WidgetState<TValue>
+  }
+
+  function getNodeWidgets(nodeId: NodeId): WidgetState[] {
+    const prefix = `${nodeId}:`
+    return [...widgetStates.value]
+      .filter(([key]) => key.startsWith(prefix))
+      .map(([, state]) => state)
+  }
```
christian-byrne, line 63: Are we afraid of derived maps or secondary views that would allow for O(widgets on node) queries rather than O(total widgets in graph). I know this is pretty fundamental to the entire ECS idea.

christian-byrne: In theory we can leverage reactivity to determine if a sub-map is dirty or not.

src/stores/widgetValueStore.ts
christian-byrne, line 1: nit/suggestion: Consider adding logging to Sentry if node:widget-name is not unique so we can verify that invariant.

christian-byrne, line 1: Do we need to add `unregisterWidget(nodeId, name)` and `unregisterNode(nodeId)`, calls in `LGraph.remove()` or the node manager's `handleNodeRemoved`, or wherever appropriate?

src/lib/litegraph/src/widgets/BaseWidget.ts
```diff
@@ -66,17 +72,50 @@ export abstract class BaseWidget<
...
   width?: number
-  disabled?: boolean
   computedDisabled?: boolean
-  hidden?: boolean
-  advanced?: boolean
-  promoted?: boolean
   tooltip?: string
+
+  private _state: Omit<WidgetState, 'nodeId'> &
+    Partial<Pick<WidgetState, 'nodeId'>>
```
christian-byrne, line 84: nit: Is this actually a mirrored state or just like a pending state? I am curious how to name/document this better? Also should we document that the _state reference swpas on `setNodeId`?

src/lib/litegraph/src/widgets/BaseWidget.ts
```diff
@@ -66,17 +72,50 @@ export abstract class BaseWidget<
...
+  set hidden(value: boolean | undefined) {
+    this._state.hidden = value ?? false
+  }
+
+  get disabled(): boolean | undefined {
+    return this._state.disabled
+  }
+  set disabled(value: boolean | undefined) {
+    this._state.disabled = value ?? false
+  }
```
christian-byrne: question: should we add additional coercion that wasn't there before (i.e., should we be coercing to false? Worried about downstream effects)?

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -1957,6 +1957,18 @@ export class LGraphNode
...
     this.widgets.push(widget)
+
+    // Only register with store if node has a valid ID (is already in a graph).
+    // If the node isn't in a graph yet (id === -1), registration happens
+    // when the node is added via LGraph.add() -> node.onAdded.
+    if (
+      this.id !== -1 &&
+      'setNodeId' in widget &&
+      typeof widget.setNodeId === 'function'
+    ) {
```
christian-byrne: nit: this duck type guard is repeated a few times and might be worth extracting

src/stores/widgetValueStore.ts
DrJKL: Not afraid, no. I would want to see that there was a real performance issue before introducing intermediate state.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -228,20 +205,40 @@ function safeWidgetMapper(
         node.widgets?.forEach((w) => w.triggerDraw?.())
       }
 
+      // Extract only render-critical options (canvasOnly, advanced)
+      const options = widget.options
+        ? {
+            canvasOnly: widget.options.canvasOnly,
+            advanced: widget.advanced
+          }
+        : undefined
```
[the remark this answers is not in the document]
DrJKL: Does IWidgetOptions have an `advanced` property?

christian-byrne (commented):
~Not done reviewing yet, posting current comments before stepping away.~ Finished first review.

src/lib/litegraph/src/widgets/BaseWidget.ts
christian-byrne: I.e., should we be coercing to false? Worried about downstream effects.

src/lib/litegraph/src/LGraph.ts
```diff
@@ -952,6 +960,16 @@ export class LGraph
       node.flags.ghost = true
     }
 
+    // Register all widgets with the WidgetValueStore now that node has a valid ID.
+    // Widgets added before the node was in the graph deferred their setNodeId call.
+    if (node.widgets) {
+      for (const widget of node.widgets) {
+        if ('setNodeId' in widget && typeof widget.setNodeId === 'function') {
+          widget.setNodeId(node.id)
```
christian-byrne: This is happening in `add`, why no cleanup in `remove`?

src/components/rightSidePanel/shared.ts
```diff
@@ -265,8 +265,8 @@ export function computedSectionDataList(nodes: MaybeRefOrGetter<LGraphNode[]>) {
           (w) =>
             !(
               w.options?.canvasOnly ||
-              w.options?.hidden ||
-              (w.options?.advanced && !includesAdvanced.value)
+              w.hidden ||
+              (w.advanced && !includesAdvanced.value)
```
christian-byrne: Won't this break if the input spec is not wired to the widget properties? This PR: https://github.com/Comfy-Org/ComfyUI/pull/12197 relies on the fact that the inputSpec is just passed to widget.options, and I think this would break it unless we create a system to wire inputSpec options onto widget properties directly (which we should consider doing).

christian-byrne: note: I think this also will break for old/legacy widgets that are not `instanceof BaseWidget`

src/extensions/core/previewAny.ts
```diff
@@ -19,14 +19,14 @@ useExtensionService().registerExtension({
...
-          'preview',
+          'preview_markdown',
           ['MARKDOWN', {}],
           app
         ).widget as DOMWidget<HTMLTextAreaElement, string>
 
         const showValueWidgetPlain = ComfyWidgets['STRING'](
           this,
-          'preview',
+          'preview_text',
```
christian-byrne, line 29: nit (non-blocking): Technically a breaking change for a few extensions: https://cs.comfy.org/search?q=context:global+.name+%3D%3D+preview&patternType=keyword&sm=0

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -228,20 +205,40 @@ function safeWidgetMapper(
         node.widgets?.forEach((w) => w.triggerDraw?.())
       }
 
+      // Extract only render-critical options (canvasOnly, advanced, read_only)
```
christian-byrne, line 213: nit/question: Ah, didn't see this when I made previous comment. Wonder if it's the best solution for wiring inputSpec onto the widgets.

christian-byrne (commented):
What I might recommend for this PR, but totally optional:

- Move the LGraph.ensureGloablIdUniqueness, subgraph nested tests/assets/fixtures, subgraph collision ID fix to a separate PR
- Possible move the LGraph.state get/set and subgraph sharing to above PR or in a stack
- Move parallelization of node defs to a separate PR and considering merging on next minor version given risk vs. reward

src/stores/widgetValueStore.ts
christian-byrne: Would you use "intermediate state" to characterize derived/computed views on the same backing data? If computed's are intermediate state, then a lot of what we do in Vue, a lot of caching/memoization, etc. are all bad. I typically think of "intermediate state" as something that has risk of becoming desynced and/or requires cleanup and manually syncing. If not, then it can't have an inherent negative connotation really.

src/lib/litegraph/src/LGraph.ts
DrJKL: Because the lifecycle for nodes is a little silly, especially when packing/unpacking/traversing subgraphs.

src/lib/litegraph/src/widgets/BaseWidget.ts
DrJKL: It's the pending state until the widget is registered in the store, then it's swapped for the reactive proxy.

src/lib/litegraph/src/widgets/BaseWidget.ts
DrJKL: We could leave it as `undefined`. I would hope we're not differentiating between `undefined` and `false` anywhere, but I wouldn't be shocked to find out that we were.

src/stores/widgetValueStore.ts
DrJKL: Once we sort out the Nodes' state similarly to this and have a real sense of their removal from the overall workflow as opposed to the currently rendered graph, yes. But right now the node's underlying info is tied to whether we intend to display it which is the cause of several of the subgraph bugs (cleanup hooks being tied to removal that get called when entering or exiting a subgraph)

src/lib/litegraph/src/LGraphNode.ts
DrJKL: Looks like it was twice, but I love a type guard 😁

src/lib/litegraph/src/LGraph.ts
DrJKL: See earlier response 🙂

src/extensions/core/previewAny.ts
DrJKL: They depend on the internals of the `Preview as Text` node?

src/composables/graph/useGraphNodeManager.ts
DrJKL: Almost certainly not. we'd probably want to compose this with the `nodeDef` store. Part of that will also be eliminating ProxyWidgets so that we can avoid having to follow the subgraph chain through to the concrete widget.

AustinMroz:
Thoroughly discussed offline. PR has my approval, but will be merged after 1.40.0

src/components/rightSidePanel/shared.ts
DrJKL: Removed hidden/advanced, keeping those half in options for now.

DrJKL:
> What I might recommend for this PR, but totally optional:
> 
>     * Move the LGraph.ensureGloablIdUniqueness, subgraph nested tests/assets/fixtures, subgraph collision ID fix to a separate PR
> 
>     * Possible move the LGraph.state get/set and subgraph sharing to above PR or in a stack
> 
>     * Move parallelization of node defs to a separate PR and considering merging on next minor version given risk vs. reward

Option exercised 🫡

AustinMroz (approved):

PR #8602, opened 2026-02-04
fix: prevent duplicate context menu items by using content-based comparison

### LINKED ISSUE #9019: 在节点上右键弹出的菜单，菜单内容重复
在节点上右键菜单，菜单内内容重复，看截图:
Comfyui ：v0.13.0，其它版本也有这个问题

Image

Myestery:
## Summary
- Switches from reference-based to content-based duplicate detection for context menu items
- Fixes cases where extensions create duplicate menu items (different objects with same content)
- Improves removal detection accuracy by comparing content strings instead of object references

## Details
The previous implementation compared menu items by object reference, which would miss duplicates when extensions added new objects with the same content. This change:
- Creates Sets of content strings from menu items for efficient duplicate detection
- Filters additions by checking if their content already exists
- Provides accurate count of removed items through content comparison

## Test plan
- [x] Unit tests pass (`pnpm test:unit src/lib/litegraph/src/contextMenuCompat.test.ts`)
- [x] TypeScript compilation succeeds (`pnpm typecheck`)
- [x] Linting passes (`pnpm lint`)
- [x] Pre-commit hooks pass

## Before
Screenshot 2026-02-04 045422

## After
Screenshot 2026-02-04 045918

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved context menu compatibility: more accurate detection of added/removed menu items and clearer warnings when items are removed.
* **Refactor**
  * Updated numeric input formatting to use locale-aware formatting logic, preserving grouping and precision behavior.
* **Tests**
  * Added a test ensuring legacy menu extraction handles items with undefined content correctly.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8602-fix-prevent-duplicate-context-menu-items-by-using-content-based-comparison-2fd6d73d36508197aa74c3409c7425fa) by [Unito](https://www.unito.io)


src/lib/litegraph/src/contextMenuCompat.ts
```diff
@@ -152,19 +152,51 @@ class LegacyMenuCompat {
...
+      const createItemKey = (item: IContextMenuValue): string => {
+        const parts = [
+          item.content ?? '',
+          item.title ?? '',
+          item.className ?? '',
+          item.property ?? '',
+          item.type ?? ''
+        ]
+        return parts.join('|')
+      }
```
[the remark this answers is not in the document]
Myestery: the current approach is enough for this use case

Myestery:
WidgetInputNumber isnt supposed to be part of this PR

src/lib/litegraph/src/contextMenuCompat.ts
```diff
@@ -152,19 +152,51 @@ class LegacyMenuCompat {
...
+        originalItems
+          .filter(
+            (item): item is IContextMenuValue =>
+              item !== null && typeof item === 'object' && 'content' in item
+          )
+          .map(createItemKey)
+      )
+      const addedItems = patchedItems.filter((item) => {
+        if (item === null) return false
+        if (typeof item !== 'object' || !('content' in item)) return false
```
benceruleanlu, line 181: nit: this filter drops `null` items entirely, so legacy patches that add separators (`null`) won’t be surfaced in `addedItems` anymore. That’s a behavior change vs the old reference-based diff.

src/lib/litegraph/src/contextMenuCompat.ts
```diff
@@ -152,19 +152,51 @@ class LegacyMenuCompat {
...
+      if (!patchedItems) {
+        return []
+      }
+      // Use content-based diff to detect additions (not reference-based)
+      // Create composite keys from multiple properties to handle undefined content
+      const createItemKey = (item: IContextMenuValue): string => {
+        const parts = [
+          item.content ?? '',
+          item.title ?? '',
+          item.className ?? '',
```
benceruleanlu, line 164: keying only on `content/title/className/property/type` means two items with the same label but different callbacks/submenus/values are treated as duplicates. That’s a behavior change vs reference-based diffing.

src/lib/litegraph/src/contextMenuCompat.ts
```diff
@@ -152,19 +152,51 @@ class LegacyMenuCompat {
...
+          item.property ?? '',
+          item.type ?? ''
+        ]
+        return parts.join('|')
```
benceruleanlu, line 168: nit: `parts.join('|')` can collide if any field contains `|` (e.g., `content: 'A|B'`), which could cause false de-duplication.

src/lib/litegraph/src/contextMenuCompat.ts
```diff
@@ -152,19 +152,51 @@ class LegacyMenuCompat {
...
+      const patchedKeys = new Set(
+        patchedItems
+          .filter(
+            (item): item is IContextMenuValue =>
+              item !== null && typeof item === 'object' && 'content' in item
+          )
+          .map(createItemKey)
+      )
+      const removedCount = [...originalKeys].filter(
+        (key) => !patchedKeys.has(key)
```
benceruleanlu, line 195: nit: removal count uses a Set of keys, so duplicates collapse. If original had multiple items with the same key, removing one may yield `removedCount = 0`, and removing several only counts as 1. Warning becomes approximate.

note: the review comment here is about the removedcount and similar drifting, not the duplication collapse itself, which is intended

benceruleanlu (commented):
WDYT about the key issue? Rest are nits and pretty easy to fix, but I don't see a good way to solve the key issue because it kinda goes against content comparison

src/lib/litegraph/src/contextMenuCompat.ts
Myestery: This won't be an issue cos the resulting key will be A|B||

src/lib/litegraph/src/contextMenuCompat.ts
Myestery: adding multiple items with same key is wrong by occurence so yes I'd filter them out

src/lib/litegraph/src/contextMenuCompat.ts
Myestery: separators are already handled properly.
Remember the goal is to remove duplicate items. separators cannot be treated as duplicate for that case

src/lib/litegraph/src/contextMenuCompat.ts
Myestery: I'd say we should do this at the rare cost of having 2 menu items with same name but different callbacks and pick only one.
Also, theres no way to check if 2 callbacks are the same unless they are references to the same function

benceruleanlu: Fine for now then

src/lib/litegraph/src/contextMenuCompat.ts
benceruleanlu: OK

src/lib/litegraph/src/contextMenuCompat.ts
benceruleanlu: if an item property is empty it's '' here, so:

     Two different items can produce the same key if any field contains |.
     Example:
      - Item A: content = 'A|B', title = 'C', others empty → key A|B|C||
      - Item B: content = 'A', title = 'B|C', others empty → key A|B|C||

but it's an edge case and a nit

src/lib/litegraph/src/contextMenuCompat.ts
benceruleanlu: This comment is about warning accuracy, not whether duplicates should be filtered.

     If duplicates exist (even if “wrong”), the Set
     collapses them and removedCount can undercount. Saying “duplicates
     shouldn’t happen” doesn’t change that the warning becomes approximate
     when they do occur.

benceruleanlu (approved):
OK given the key issue isn't that blocking

PR #8603, opened 2026-02-04
New bottom button and badges

AustinMroz:
- "Enter Subgraph" "Show advanced inputs" and a new "show node Errors"  button now use a combined button design at the bottom of the node.
- A new "Errors" tab is added to the right side panel
- After a failed queue, the label of an invalid widget is now red.
- Badges other than price are now displayed on the bottom of the node.
- Price badge will now truncate from the first space, prioritizing the sizing of the node title
- An indicator for the node resize handle is now displayed while mousing over the node.

image

resize


src/composables/node/usePriceBadge.ts
```diff
@@ -36,7 +36,9 @@ export const usePriceBadge = () => {
     return badges
   }
 
-  function isCreditsBadge(badge: LGraphBadge | (() => LGraphBadge)): boolean {
+  function isCreditsBadge(
+    badge: Partial<LGraphBadge> | (() => LGraphBadge)
```
DrJKL: Should we setup our own utility like MaybeRefOrGetter but without Ref?
```suggestion
    badge: Partial<LGraphBadge> | (() => Partial<LGraphBadge>)
```

src/renderer/extensions/vueNodes/components/LGraphNode.subgraph.test.ts
```diff
@@ -7,11 +7,11 @@ import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import type {
   LGraph,
-  LGraphNode,
+  LGraphNode as LGLGraphNode,
```
DrJKL, line 10: ...

...

?

src/renderer/extensions/vueNodes/components/LGraphNode.subgraph.test.ts
```diff
@@ -7,11 +7,11 @@ import { beforeEach, describe, expect, it, vi } from 'vitest'
...
   SubgraphNode
 } from '@/lib/litegraph/src/litegraph'
 import type { VueNodeData } from '@/composables/graph/useGraphNodeManager'
-import NodeHeader from '@/renderer/extensions/vueNodes/components/NodeHeader.vue'
+import LGraphNode from '@/renderer/extensions/vueNodes/components/LGraphNode.vue'
```
DrJKL, line 14: We should probably rename this in a separate PR.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -28,19 +28,18 @@
...
       )
     "
     :style="[
       {
         transform: `translate(${position.x ?? 0}px, ${(position.y ?? 0) - LiteGraph.NODE_TITLE_HEIGHT}px)`,
         zIndex: zIndex,
         opacity: nodeOpacity,
-        '--component-node-background': applyLightThemeColor(nodeData.bgcolor)
+        '--component-node-background': applyLightThemeColor(nodeData.bgcolor),
+        backgroundColor: applyLightThemeColor(nodeData?.color)
```
DrJKL, line 42: This seems sus. color and bgcolor both being applied to something that looks like it'd be the background.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -120,42 +119,76 @@
...
+      "
+      as-child
+    >
+      <button
+        v-if="hasAnyError"
+        @click.stop="useRightSidePanelStore().openPanel('error')"
+      >
+        <span>{{ t('g.error') }}</span>
+        <i class="icon-[lucide--info] size-4" />
+      </button>
```
DrJKL, line 141: I see what you're going for here, but the structure here is a little tricky to trace.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -120,42 +119,76 @@
...
+        <span>{{ t('rightSidePanel.hideAdvancedInputsButton') }}</span>
+        <i class="icon-[lucide--chevron-up] size-4" />
+      </button>
+      <button
+        v-else-if="showAdvancedInputsButton"
+        @click.stop="showAdvancedState = !showAdvancedState"
+      >
+        <span>{{ t('rightSidePanel.showAdvancedInputsButton') }} </span>
+        <i class="icon-[lucide--settings-2] size-4" />
+      </button>
```
DrJKL, line 162: These two, for instance, seem eerily similar to one another.

src/renderer/extensions/vueNodes/components/NodeBadges.vue
```diff
@@ -0,0 +1,49 @@
...
+          class="h-6 first:pl-2 last:pr-2"
+        />
+      </template>
+    </div>
+    <div
+      v-if="extension.length"
+      class="rounded-full bg-component-node-widget-background h-6 flex justify-center items-center overflow-clip"
+    >
+      <template v-for="(badge, index) of extension" :key="badge.text">
+        <div v-if="index">|</div>
```
DrJKL: Clever. I hate it.

Could you do the separators using CSS?

src/renderer/extensions/vueNodes/components/NodeHeader.vue
```diff
@@ -63,56 +55,46 @@
...
+              'flex h-5 bg-component-node-widget-background p-1 items-center text-xs shrink-0',
+              badge.rest ? 'rounded-l-full pr-1' : 'rounded-full'
+            )
+          "
         >
-          <span>{{ $t('g.edit') }}</span>
-          <i class="icon-[lucide--scaling] size-5" />
-        </Button>
-      </div>
+          <i class="h-full icon-[lucide--component] bg-[#FABC25]" />
```
DrJKL: What's the arbitrary color?

src/renderer/extensions/vueNodes/components/NodeHeader.vue
```diff
@@ -63,56 +55,46 @@
...
+        </span>
+        <span
+          v-if="badge.rest"
+          class="truncate -ml-2.5 grow-1 basis-0 bg-component-node-widget-background rounded-r-full max-w-max min-w-0"
+        >
+          <span class="pr-2" v-text="badge.rest" />
+        </span>
+      </template>
+      <NodeBadge v-if="statusBadge" v-bind="statusBadge" />
+      <i-comfy:pin
```
DrJKL: The component form of badges has had issues, that's why we standardized around `<i class="icon-[...]" />`

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -146,21 +154,23 @@ const { getWidgetTooltip, createTooltipConfig } = useNodeTooltips(
 const widgetValueStore = useWidgetValueStore()
 
 interface ProcessedWidget {
+  advanced: boolean
+  hasLayoutSize: boolean
+  hasError: boolean
+  hidden: boolean
```
DrJKL, line 160: Not for this PR, but this is the kind of thing that I want to restructure so that we can compose interfaces and not duplicate the fields across different interfaces.

src/renderer/extensions/vueNodes/composables/usePartitionedBadges.ts
```diff
@@ -0,0 +1,135 @@
...
+        if (relevantNames.length > 0 && nodeData?.id != null) {
+          for (const name of relevantNames) {
+            // Access value from store to create reactive dependency
+            void widgetStore.getWidget(nodeData.id, name)?.value
+          }
+        }
+        // Access input connections for regular inputs
+        const inputNames = relevantInputNames.value
+        if (inputNames.length > 0) {
+          nodeData?.inputs?.forEach((inp) => {
```
DrJKL, line 66: inp

src/renderer/extensions/vueNodes/composables/usePartitionedBadges.ts
```diff
@@ -0,0 +1,135 @@
...
+            ) {
+              void inp.link // Access link to create reactive dependency
+            }
+          })
+        }
+      }
+    }
+    return [...(nodeData?.badges ?? [])].map(toValue)
+  })
+  const nodeDef = useNodeDefStore().nodeDefsByName[nodeData.type]
```
DrJKL, line 87: This might need to be `computed`

src/renderer/extensions/vueNodes/composables/usePartitionedBadges.ts
```diff
@@ -0,0 +1,135 @@
...
+  return computed(() => {
+    const displaySource = settingStore.get(
+      'Comfy.NodeBadge.NodeSourceBadgeMode'
+    )
+    const isCoreNode =
+      nodeDef?.isCoreNode && displaySource === NodeBadgeMode.ShowAll
+    const core: NodeBadgeProps[] = []
+    const extension: NodeBadgeProps[] = []
+    const pricing: { required: string; rest?: string }[] = []
```
DrJKL, line 96: This is very procedural

DrJKL (dismissed):

src/composables/node/usePriceBadge.ts
AustinMroz: Will keep an eye on it. It probably makes for better code if we just accept `MaybeRefOrGetter` any place we'd want this and use the standard `toValue` for normalization.

src/renderer/extensions/vueNodes/components/LGraphNode.subgraph.test.ts
AustinMroz: Agreed

src/renderer/extensions/vueNodes/components/LGraphNode.vue
AustinMroz: Agreed, it looks sketch. Since the NodeHeader color is now shared between the the Header and bottom button, the PR swaps things around so `nodeData.color` is the layer of color beneath the primary node color defined by `nodeData.bgcolor`

Within the scope of this PR, I can point to the exchange being an extraction and screenshot expectations not changing for any of the themed snapshots, but i'd be nice if it was better conveyed.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
AustinMroz: Grouping these two together does wind up saving a line. I consider it a positive.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
AustinMroz: Needing to apply the error styling outside (so that it overrides the defaults on button) makes me sad and really hurts the cleanliness of the approach here, though I still feel like it's most sustainable approach here. Particularly in the likely event more variants get added.

src/renderer/extensions/vueNodes/components/NodeBadges.vue
AustinMroz: The usual tricks (`last:`, `after:`) don't work well here because the badges can be colored and have rounded border
image

I've kept the div, but moved it to CSS on the div (and made the conditional a bit more explicit)

src/renderer/extensions/vueNodes/components/NodeHeader.vue
AustinMroz: A lovely `amber-400`. I promise I cross referenced our `style.css`, but forgot that, as a default tailwind color, it wouldn't show. This should probably be added as a theme color (`--color-partner`?) but should probably be pulled into a separate PR.

src/renderer/extensions/vueNodes/components/NodeHeader.vue
AustinMroz: It's not introduced in this PR, I promise

(I've still fixed it while I'm here.)

src/renderer/extensions/vueNodes/components/LGraphNode.vue
DrJKL: An alternative would be to pull out the text/icon/handler, type it, compute that set of properties.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
DrJKL: 😛

DrJKL (approved):

src/composables/node/usePriceBadge.ts
DrJKL: I'm for it.

src/components/rightSidePanel/TabError.vue
```diff
@@ -0,0 +1,26 @@
+<script setup lang="ts">
+import Button from '@/components/ui/button/Button.vue'
+import { useCopyToClipboard } from '@/composables/useCopyToClipboard'
+import { t } from '@/i18n'
```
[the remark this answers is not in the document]
DrJKL: @coderabbitai. Recommend vue-i18n's composable instead.

PR #8625, opened 2026-02-05
Fix incorrect widgetValue migration

AustinMroz:
Under a combination of many edge cases, the `widget_values` migration code added in #3326 would cause the progress text on a "Recraft Text to Image" node to incorrectly deserialize into the `control_after_generate`
- widgets_values is of length 1 greater than it should be because progress text serializes
  - It should not, there is no code to deserialize it
- negative_prompt has force_input set and skips serialization
- Migration only applies when `widgets_values` is equal to actual inputs length. The two above edge cases cancel to make this true
- Seed is accounted for when calculating the length of widgets, but not when applying the migration
- Migration occurs even though we track workflow version now and have an accurate way of determining that it can not be needed

The two primary edge cases which cause the bug are both addressed
- `options.serialize` does nothing and has never done anything. I've been guilty of making the same mistake in the ancient past, and want to clean up the misconception where I can.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8625-Fix-incorrect-widgetValue-migration-2fe6d73d365081a683b4c675eaeebb6c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/composables/useProgressTextWidget.ts
+++ b/src/renderer/extensions/vueNodes/widgets/composables/useProgressTextWidget.ts
@@ -42,11 +42,11 @@ export function useTextPreviewWidget(
           widgetValue.value = typeof value === 'string' ? value : String(value)
         },
         getMinHeight: () => options.minHeight ?? 42 + PADDING,
-        serialize: false,
         read_only: true
       },
       type: inputSpec.type
     })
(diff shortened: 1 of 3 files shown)


DrJKL (approved):

christian-byrne (approved):

christian-byrne:
Nicely done

PR #8684, opened 2026-02-06
fix: include subfolder in asset download URL for audio/video files

Myestery:
## Summary

- `getAssetUrl()` was constructing `/view` URLs without the `subfolder` query parameter, causing backend to return "file not found" for assets stored in subfolders (common for audio/video outputs)
- Preview/playback was unaffected because it uses `preview_url` from `ResultItemImpl.url` which correctly includes `subfolder`
- Fixed `getAssetUrl()` to include `subfolder` from `asset.user_metadata` when present
- Simplified download functions to prefer `preview_url` (already correct) with `getAssetUrl` as fallback

## Test plan

- [ ] Generate audio/video output (e.g. via SaveAudio node) that saves to a subfolder
- [ ] Right-click the asset in the assets sidebar and click Download — should download successfully
- [ ] Select multiple audio/video assets and use bulk download — should download all
- [ ] Verify image downloads still work as before
- [ ] Verify cloud environment downloads still work (uses `preview_url` path)

## Release Notes

* **New Features**
  * Added support for organizing and downloading assets from subfolders.

* **Refactor**
  * Improved asset URL generation and download handling for better reliability and performance.

### DIFF
--- a/src/platform/assets/composables/useMediaAssetActions.ts
+++ b/src/platform/assets/composables/useMediaAssetActions.ts
@@ -65,15 +65,8 @@ export function useMediaAssetActions() {
 
     try {
       const filename = targetAsset.name
-      let downloadUrl: string
-
-      // In cloud, use preview_url directly (from cloud storage)
-      // In OSS/localhost, use the /view endpoint
-      if (isCloud && targetAsset.preview_url) {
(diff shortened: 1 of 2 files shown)


viva-jinyi (approved):
LGTM

PR #8689, opened 2026-02-06
fix: jobs stuck in initializing state when failing before execution_start

christian-byrne:
## Summary

Fix jobs getting permanently stuck in "initializing" state when they fail before the `execution_start` WebSocket event fires.

## Changes

- **What**: Added `reconcileInitializingPrompts(activeJobIds)` to `executionStore` that removes orphaned initializing prompt IDs not present in the active jobs set. Called from `queueStore.update()` after fetching Running/Pending jobs, ensuring stale initializing states are cleaned up on every queue poll.

## Review Focus

- The reconciliation delegates to the existing `clearInitializationByPromptIds` to avoid duplicating Set-diffing logic.
- Only runs during `queueStore.update()` which is already a periodic poll — no additional network calls.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8689-fix-jobs-stuck-in-initializing-state-when-failing-before-execution_start-2ff6d73d3650814dbeeeda71c8bb7d43) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/queueStore.ts
+++ b/src/stores/queueStore.ts
@@ -530,6 +530,12 @@ export const useQueueStore = defineStore('queue', () => {
         }
       })
 
+      const activeJobIds = new Set([
+        ...queue.Running.map((j) => j.id),
+        ...queue.Pending.map((j) => j.id)
+      ])
+      executionStore.reconcileInitializingPrompts(activeJobIds)
+
       // Sort by create_time descending and limit to maxItems
       const sortedHistory = [...history]
         .sort((a, b) => b.create_time - a.create_time)
--- a/src/stores/queueStore.ts
+++ b/src/stores/queueStore.ts
@@ -530,11 +530,17 @@ export const useQueueStore = defineStore('queue', () => {
         }
       })
 
-      const activeJobIds = new Set([
-        ...queue.Running.map((j) => j.id),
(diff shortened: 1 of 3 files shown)


src/stores/queueStore.ts
```diff
@@ -530,6 +530,12 @@ export const useQueueStore = defineStore('queue', () => {
         }
       })
 
+      const activeJobIds = new Set([
+        ...queue.Running.map((j) => j.id),
+        ...queue.Pending.map((j) => j.id)
+      ])
+      executionStore.reconcileInitializingPrompts(activeJobIds)
```
benceruleanlu: Potential regression on transient queue fetch failures: `api.getQueue()` returns empty `Running/Pending` on error fallback, so this reconciliation can clear all `initializingPromptIds` even while jobs are still active. Consider gating reconciliation to only run when queue fetch is known-good (e.g. return `{ ok: boolean }` from fetch/getQueue and skip reconcile when `ok === false`).

benceruleanlu (dismissed):
have a slight worry that doesn't seem hard to do

src/stores/queueStore.ts
christian-byrne: Great point, will fix, thanks for pointing it out.

PR #8709, opened 2026-02-07
docs: document subgraph limitations with autogrow and matchtype

christian-byrne:
## Summary

Document why autogrow and matchtype don't work inside subgraphs, covering root cause, symptoms, workarounds, and historical context.

## Changes

- **What**: Add `src/core/graph/subgraph-dynamic-input-limitations.md` explaining the static-vs-dynamic impedance mismatch between subgraph slots (fixed type at creation) and matchtype/autogrow (runtime mutations). Includes Mermaid diagrams, workarounds, and PR history. Link added to `src/lib/litegraph/AGENTS.md` for discoverability.

## Review Focus

- Accuracy of the technical explanation — @AustinMroz authored matchtype/autogrow, @webfiltered authored the subgraph system
- Whether the workarounds section is practical and complete
- File placement: colocated at `src/core/graph/` next to both `subgraph/` and `widgets/dynamicWidgets.ts`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8709-docs-document-subgraph-limitations-with-autogrow-and-matchtype-3006d73d36508134a64ce8c2c49e05f1) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/graph/subgraph-dynamic-input-limitations.md
+++ b/src/core/graph/subgraph-dynamic-input-limitations.md
@@ -34,6 +34,7 @@ The subgraph boundary slot type is frozen at connection time with no subsequent 
 Autogrow hooks `onConnectionsChange` and splices new inputs into the node's `inputs` array when the last slot is connected. These inputs are added to the **internal node**, not to the subgraph's input/output definitions. `SubgraphNode` only listens for events on the subgraph definition — there is no bridge.
 
 Open UX questions (from the autogrow author):
+
 - Should promoting an autogrow connection cause the `SubgraphNode` to autogrow from that slot?
 - How to distinguish "promote this specific input" vs "promote the autogrow behavior"?
 
--- a/src/core/graph/subgraph-dynamic-input-limitations.md
+++ b/src/core/graph/subgraph-dynamic-input-limitations.md
@@ -12,10 +12,10 @@ flowchart LR
     N[Node with matchtype/autogrow]
   end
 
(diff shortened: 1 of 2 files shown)


src/core/graph/subgraph-dynamic-input-limitations.md
```diff
@@ -0,0 +1,53 @@
...
+  classDef dynamic fill:#1a1a2e,stroke:#0f3460,color:#eee
+  class SN,SIN static
+  class N dynamic
+```
+
+## MatchType
+
+`SubgraphInputNode.connectByType` captures the type as a fixed string via `subgraph.addInput(name, String(type))`. `SubgraphNode` mirrors this once via the `input-added` event.
+
+MatchType hooks `onConnectionsChange` and mutates `input.type` at runtime based on the connected source. It reads types across the subgraph boundary via `link.resolve()`, but never writes resolved types back. The `SubgraphNode` retains the original type, so connecting an incompatible type to it can **silently disconnect** internal connections with no visual feedback.
```
AustinMroz: This part is incorrect. The silent disconnection was a hypothetical UX hurdle, not an issue with the current implementation. There is no resolution or reactivity across the subgraph boundary. At the time the first link is made, the type of the subgraph boundry slot is simply frozen.

For example, here, the switch was connected to the subgraphInput first. The subgraph slot create is permanently set to the "*"/any type.
image

If the image is connected first, the subgraph input is correctly typed and the connection is properly blocked outside the subgraph
image

PR #8721, opened 2026-02-07
Support keeping value across dynamic combo options

AustinMroz:
Dynamic Combos will sometimes have options with a level of overlap, either because they differ in options (min/max) or they're only common on a portion of the Dynamic Combo options. This PR provides a means for these shared widgets to maintain there value across option group selections.

TODO
- Currently, the filtering of when updates should apply isn't the greatest. Ideally, value is kept if the defaults of the old and new spec are equal (including if both are undefined), but the default of a widget is not kept after initialization and there is not currently code in place to resolve the input spec of the removed widget.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8721-Support-keeping-value-across-dynamic-combo-options-3006d73d3650817b8d7ae20f68b11b5b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/graph/widgets/dynamicWidgets.ts
+++ b/src/core/graph/widgets/dynamicWidgets.ts
@@ -95,7 +95,9 @@ function dynamicComboWidget(
     const newSpec = value ? options[value] : undefined
 
     const removedInputs = remove(node.inputs, isInGroup)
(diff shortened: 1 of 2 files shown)


DrJKL:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/16508

PR #8723, opened 2026-02-07
feat: add KSampler live previews to assets sidebar jobs

benceruleanlu:
## Summary
Show live KSampler previews on active job cards/list items in the Assets sidebar, while preserving existing fallback behavior.

## Changes
- **What**:
  - Added a prompt-scoped job preview store (`jobPreviewStore`) gated by `Comfy.Execution.PreviewMethod`.
  - Wired `b_preview_with_metadata` handling to map previews by `promptId`.
  - Extended queue job view model with `livePreviewUrl` and consumed it in both sidebar list and grid active job UIs.
  - Cleared prompt previews on execution reset.
  - Added ref-counted shared blob URL lifecycle utility (`objectUrlUtil`) and updated preview stores to retain/release shared URLs so each preview event creates one object URL.
  - Added/updated unit coverage in `useJobList.test.ts` for preview enable/disable mapping.

## Review Focus
- Object URL lifecycle correctness across node previews and job previews (retain/release behavior).
- Preview gating behavior when `Comfy.Execution.PreviewMethod` is `none`.
- Active job UI fallback behavior (`livePreviewUrl` -> `iconImageUrl`).

image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8723-feat-add-KSampler-live-previews-to-assets-sidebar-jobs-3006d73d365081aeb81dd8279bf99f94) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/queue/useJobList.ts
+++ b/src/composables/queue/useJobList.ts
@@ -39,7 +39,6 @@ export type JobListItem = {
   state: JobState
   iconName?: string
   iconImageUrl?: string
-  livePreviewUrl?: string
   showClear?: boolean
   taskRef?: TaskItemImpl
   progressTotalPercent?: number
@@ -260,7 +259,7 @@ export function useJobList() {
         String(executionStore.activePromptId ?? '')
       const showAddedHint = shouldShowAddedHint(task, state)
       const promptKey = taskIdToKey(task.promptId)
-      const livePreviewUrl =
+      const promptPreviewUrl =
         state === 'running' && jobPreviewStore.isPreviewEnabled && promptKey
           ? jobPreviewStore.previewsByPromptId[promptKey]
           : undefined
@@ -283,8 +282,7 @@ export function useJobList() {
         meta: display.secondary,
         state,
         iconName: display.iconName,
-        iconImageUrl: display.iconImageUrl,
-        livePreviewUrl,
+        iconImageUrl: promptPreviewUrl ?? display.iconImageUrl,
         showClear: display.showClear,
         taskRef: task,
         progressTotalPercent:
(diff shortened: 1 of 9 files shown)


christian-byrne (commented):
Can you check how this was done in the past with the Queue sidebar?

benceruleanlu:
> Can you check how this was done in the past with the Queue sidebar?

Deduplicated now

src/stores/jobPreviewStore.ts
```diff
@@ -0,0 +1,61 @@
...
+    if (!enabled) clearAllPreviews()
+  })
+
+  return {
+    previewsByPromptId,
+    isPreviewEnabled,
+    setPreviewUrl,
+    clearPreview,
+    clearAllPreviews
+  }
```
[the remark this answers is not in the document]
benceruleanlu: OK Done

benceruleanlu:
Accidentally assigned before tests finished, will reassign then

src/composables/queue/useJobList.ts
```diff
@@ -256,6 +258,11 @@ export function useJobList() {
         String(task.promptId ?? '') ===
         String(executionStore.activePromptId ?? '')
       const showAddedHint = shouldShowAddedHint(task, state)
+      const promptKey = taskIdToKey(task.promptId)
+      const promptPreviewUrl =
+        state === 'running' && jobPreviewStore.isPreviewEnabled && promptKey
+          ? jobPreviewStore.previewsByPromptId[promptKey]
+          : undefined
```
AustinMroz, line 265: NIT: Having all job items under a single computed means that a preview update on one (order 10/second) will occur on all job outputs (order 100).

From quick testing, nested computeds don't share the magic auto-unwrapping of nested refs,  so it might require changing the iconImageUrl to `MaybeRef`

AustinMroz (approved):
Thinking things over, ref counting for previews is a great idea since it means we now have a means to easily check the number of leaked previews.

(I am seeing previews leak when nodes are removed, but I'm confident that's an existing bug and not a problem with this PR).

src/composables/queue/useJobList.ts
benceruleanlu: Taking a look, will see if I can fix in a followup

PR #8732, opened 2026-02-08
fix: show completion summary banner when QPOV2 is enabled

benceruleanlu:
## Summary

Restore the queue completion summary banner behavior when QPOV2 is enabled, while keeping QPOV2 scoped to completion-summary UI rather than active-job overlay UI.

## Changes

- **What**: Added a `completionSummaryOnly` mode to `QueueProgressOverlay` and wired `TopMenuSection` to use it under QPOV2. Simplified top-menu queue overlay rendering/pressed-state logic by removing tautological legacy guards. Updated `TopMenuSection` tests to cover QPOV2 overlay rendering and prop wiring.

## Review Focus

Confirm that with QPOV2 enabled, only completion summary banner behavior is surfaced (no active job overlay state), and that legacy overlay behavior remains unchanged when QPOV2 is disabled.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8732-fix-show-completion-summary-banner-when-QPOV2-is-enabled-3016d73d3650819c8ce6f36010b0208c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/TopMenuSection.test.ts
+++ b/src/components/TopMenuSection.test.ts
@@ -161,7 +161,7 @@ describe('TopMenuSection', () => {
     expect(queueButton.text()).toContain('3 active')
   })
 
-  it('hides queue progress overlay when QPO V2 is enabled', async () => {
+  it('renders queue progress overlay when QPO V2 is enabled with no active jobs', async () => {
     const pinia = createTestingPinia({ createSpy: vi.fn })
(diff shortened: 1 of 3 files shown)


benceruleanlu:
Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/8740

PR #8574, opened 2026-02-03
fix(vue-nodes): hide slot labels for reroute nodes with empty names

christian-byrne:
## Summary
Fixes reroute node styling in Vue Nodes 2.0 by hiding slot labels when slot names are intentionally empty.

| Before                                                                                                                                        | After                                                                                                                                         |
| --------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
| image | image |

## Problem
Reroute nodes displayed unwanted fallback labels ("Input 0", "Output 0") instead of appearing as minimal connection-only nodes. This happened because:
- Reroute nodes intentionally use empty string (`""`) for slot names
- Slot components used `||` operator for fallback labels, treating `''` as falsy

## Solution
- Add `hasNoLabel` computed property to detect when all label sources (`label`, `localized_name`, `name`) are empty/falsy
- Derive `dotOnly` from either the existing prop OR `hasNoLabel` being true
- When `dotOnly` is true: label text is hidden, padding removed (`lg-slot--dot-only` class), only connection dot visible

Combined with existing `NO_TITLE` support from #7589, reroutes now display as minimal nodes with just connection dots—matching classic reroute appearance.

## Release Notes

* **Bug Fixes**
  * Enhanced input and output slot label handling to automatically conceal labels when unavailable
  * Improved fallback display names for slots with more reliable naming logic

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8574-fix-vue-nodes-hide-slot-labels-for-reroute-nodes-with-empty-names-2fc6d73d365081c38031e260402283d3) by [Unito](https://www.unito.io)

PR #8734, opened 2026-02-08
fix(vueNodes): decrease default size of reroute nodes

### LINKED ISSUE #4704: Please make reroutes smaller
It kind of annoys me that reroutes take up so much space and I think they would look way better if they were smaller. Could this be changed? I think they should be similar size to sockets. They are already different enough visually to not be confused with them, but if it's not enough, maybe we could just increase the size of the gap between the inner circle and the outer ring?

Image

christian-byrne:
## Summary

Remove the 225px minimum width constraint from reroute nodes so they render at their intended ~75×26px size.

## Changes

- **What**: Reroute nodes now bypass the `min-w-[225px]` CSS constraint and bottom padding applied to regular nodes. `resizable: false` is set on the RerouteNode constructor to hide the resize handle. An `isRerouteNode` computed in `LGraphNode.vue` gates these behaviors by checking `nodeData.type === "Reroute"`.

## Review Focus

- Detection mechanism uses `type === "Reroute"` (explicit) rather than `titleMode === NO_TITLE` (semantic but too broad). See PR #8574 as prior art for reroute-specific conditionals.
- PR #7993 (open) modifies `LGraphNode.ts` pos/size setters but does not touch `LGraphNode.vue` template or resize callback — no conflict expected.

Fixes #4704

Reroute nodes now render at ~75px wide instead of being forced to 225px minimum.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8734-fix-vueNodes-decrease-default-size-of-reroute-nodes-3016d73d3650816dbeead93517a52f25) by [Unito](https://www.unito.io)


pythongosssss:
In litegraph mode the node is very hard to move as the clickable area for the slots takes up most of the node

christian-byrne:
I simplified it

Screenshot from 2026-02-13 16-46-37

Screenshot from 2026-02-13 16-46-46

christian-byrne:
Test here: https://pr-2369.testenvs.comfy.org/

pythongosssss:
They still create big then shrink on first connect

https://github.com/user-attachments/assets/35943030-8e84-4be9-b14f-064cd474b336

dante01yoon (dismissed):

dante01yoon:
@christian-byrne Do you want to resolve conflict?

PR #8749, opened 2026-02-08
feat: automatically fit view when loading templates

christian-byrne:
## Description

When loading a template workflow, always call `fitView()` to ensure the template is properly framed within the viewport. This provides a better initial viewing experience for users.

## Problem

Previously, templates with embedded viewport positions (`extra.ds`) would load at arbitrary positions, sometimes showing a blank canvas. Users had to navigate significantly to find the workflow content.

## Solution

Added a single condition in `loadGraphData()` to check if `openSource === 'template'` and force `fitView()` when true, bypassing the saved viewport position.

This change only affects template loading - normal workflow loading behavior remains unchanged.

## Changes

- `src/scripts/app.ts`: Added condition to always call `fitView()` when loading templates

## Testing

- Load various templates from the template selector
- Verify the workflow is fully visible and centered on load
- Verify normal workflow loading still respects saved positions

## Related

- Fixes COM-14156
- Related to COM-10087 (Center view on workflow when loading templates - Done)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8749-feat-automatically-fit-view-when-loading-templates-3016d73d36508167a63de2ae407de7b8) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/templateFitView.spec.ts
+++ b/browser_tests/tests/templateFitView.spec.ts
@@ -1,0 +1,224 @@
+import { expect } from '@playwright/test'
+
+import { comfyPageFixture as test } from '../fixtures/ComfyPage'
+
+/**
+ * Tests that templates are automatically fitted to view when loaded.
+ *
+ * This feature ensures templates with saved off-screen viewport positions
+ * (extra.ds) are always displayed correctly by calling fitView() immediately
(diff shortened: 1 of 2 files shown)


AustinMroz (dismissed):

AustinMroz:
Change itself is minimal and fine. I can reapprove once the failing test is fixed.

PR #8757, opened 2026-02-09
Add support for dragging in multiple workflow files at once

pythongosssss:
## Summary

Allows users to drag in multiple files that are/have embedded workflows and loads each of them as tabs.  
Previously it would only load the first one.

## Changes

- **What**: 
- process all files from drop event
- add defered errors so you don't get errors for non-visible workflows

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8757-Add-support-for-dragging-in-multiple-workflow-files-at-once-3026d73d365081c096e9dfb18ba01253) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM. Will be nice for transferring workflows cloud<>local too.

PR #8772, opened 2026-02-10
Display optional indicator on subgraphNode inputs

AustinMroz:
Display an optional input on subgraphNodes when all links to the subgraph input are optional.
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8772-Display-optional-indicator-on-subgraphNode-inputs-3036d73d365081cc83e8d6034302665f) by [Unito](https://www.unito.io)


pythongosssss (commented):
If you connect to an optional input, then a required, then disconnect the required, it doesn't render as optional

AustinMroz:
Bit of a rabbit hole, but optional status will now update to account for disconnections

And disconnecting single links will now correctly fire input-disconnected events which, most noticeably, means that disconnecting the last link will remove the cloned widget from the parent subgraphNode.

christian-byrne:
Is this PR still needed? Anything I can help with here @AustinMroz ?

AustinMroz:
It is. I never pushed for re-review after addressing the prior feedback and it got sidelined by app mode work. I've rebased it now.

pythongosssss (approved):

PR #8774, opened 2026-02-10
fix: disable control after generate during partial execution

jtydhr88:
## Summary
Passes an isPartialExecution flag through widget beforeQueued/afterQueued callbacks so control-after-generate widgets skip value modifications (randomize, increment, decrement) when the user queues selected output nodes via partial execution.

requested by @christian-byrne in notion

before

https://github.com/user-attachments/assets/3e723087-8849-457b-9f95-b8b5fceab0ed

after

https://github.com/user-attachments/assets/d9816667-51e0-4538-a012-9c84d0944019

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8774-fix-disable-control-after-generate-during-partial-execution-3036d73d365081688ca3d6b0506d69ca) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -1389,11 +1389,13 @@ export class ComfyApp {
           'Comfy.Execution.PreviewMethod'
         )
 
+        const isPartialExecution = !!queueNodeIds?.length
         for (let i = 0; i < batchCount; i++) {
           // Allow widgets to run callbacks before a prompt has been queued
           // e.g. random seed before every gen
           forEachNode(this.rootGraph, (node) => {
-            for (const widget of node.widgets ?? []) widget.beforeQueued?.()
+            for (const widget of node.widgets ?? [])
+              widget.beforeQueued?.({ isPartialExecution })
           })
 
           const p = await this.graphToPrompt(this.rootGraph)
@@ -1449,7 +1451,9 @@ export class ComfyApp {
 
           // Allow widgets to run callbacks after a prompt has been queued
           // e.g. random seed after every gen
-          executeWidgetsCallback(queuedNodes, 'afterQueued')
+          executeWidgetsCallback(queuedNodes, 'afterQueued', {
+            isPartialExecution
+          })
           this.canvas.draw(true, true)
           await this.ui.queue.update()
         }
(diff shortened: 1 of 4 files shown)


src/scripts/app.ts
```diff
@@ -1389,11 +1389,13 @@ export class ComfyApp {
...
         )
 
+        const isPartialExecution = !!queueNodeIds?.length
         for (let i = 0; i < batchCount; i++) {
           // Allow widgets to run callbacks before a prompt has been queued
           // e.g. random seed before every gen
           forEachNode(this.rootGraph, (node) => {
-            for (const widget of node.widgets ?? []) widget.beforeQueued?.()
+            for (const widget of node.widgets ?? [])
+              widget.beforeQueued?.({ isPartialExecution })
```
DrJKL, line 1398: Nit: I should turn on `curly` for this...
```suggestion
            for (const widget of node.widgets ?? []) {
              widget.beforeQueued?.({ isPartialExecution })
            }
```

src/types/litegraph-augmentation.d.ts
```diff
@@ -47,8 +47,8 @@ declare module '@/lib/litegraph/src/types/widgets' {
 
   interface IBaseWidget {
     onRemove?(): void
-    beforeQueued?(): unknown
-    afterQueued?(): unknown
+    beforeQueued?(options?: { isPartialExecution?: boolean }): unknown
+    afterQueued?(options?: { isPartialExecution?: boolean }): unknown
```
DrJKL, line 51: Nit: Want to extract this as an interface?
Maybe something like WidgetQueueLifecycleOptions? That's a little wordy...

DrJKL: `WidgetCallbackOptions`?

DrJKL (approved):

diGitTil:
@christian-byrne @jtydhr88  I would absolutely love to understand the logic behind this change. As an end-user, I cannot comprehend the utility... partial execution is a critical feature to troubleshoot. Now, even selecting the seed node still won't force a beforeQueued call, making it extremely cumbersome to track down workflow failures.

PR #8775, opened 2026-02-10
Add support for values factory function in widget select combo

pythongosssss:
## Summary

Adds support for values factory functions, e.g.
```
this.addWidget(
    "combo",
    "Dynamic",
    "",
    (e) => { },
    {
        values: () => {
            return getSomeValuesHere() 
        }
    }
)
```

Specifically for fixing KJNodes get/set

## Changes

- **What**:  Check if object is a function, if so, calls it.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8775-Add-support-for-values-factory-function-in-widget-select-combo-3036d73d3650819bb4e4f9181445cb1d) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
```diff
@@ -51,25 +53,27 @@ interface Props {
...
+}
+
 const modelValue = defineModel<string | undefined>({
   default(props: Props) {
-    return props.widget.options?.values?.[0] ?? ''
+    const values: unknown = props.widget.options?.values
+    const resolved = typeof values === 'function' ? values() : values
+    return Array.isArray(resolved) ? (resolved[0] ?? '') : ''
   }
 })
```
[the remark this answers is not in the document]
pythongosssss: The defineModel section is hoisted outside the script setup component, so it cant call resolveValues.
I can refactor this if you think its worthwhile, however I figured for 3 lines of code it wasnt worth it.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
```diff
@@ -51,25 +53,27 @@ interface Props {
...
+function resolveValues(values: unknown): string[] {
+  if (typeof values === 'function') return values()
+  if (Array.isArray(values)) return values
+  return []
+}
+
 const modelValue = defineModel<string | undefined>({
   default(props: Props) {
-    return props.widget.options?.values?.[0] ?? ''
+    const values: unknown = props.widget.options?.values
```
simula-r: nit: unknown type

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
```diff
@@ -51,26 +53,27 @@ interface Props {
...
-const selectOptions = computed(() => {
-  const options = props.widget.options
-
-  if (options?.values && Array.isArray(options.values)) {
-    return options.values
-  }
-
-  return []
-})
+const selectOptions = ref(resolveValues(props.widget.options?.values))
```
DrJKL: Should this be a computed?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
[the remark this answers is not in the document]
DrJKL: Defaults in models shouldn't usually contain logic like this, generally.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
```diff
@@ -51,26 +53,27 @@ interface Props {
...
   default(props: Props) {
     const values = props.widget.options?.values
-    return (Array.isArray(values) ? values[0] : undefined) ?? ''
+    const resolved = typeof values === 'function' ? values() : values
+    return Array.isArray(resolved) ? (resolved[0] ?? '') : ''
   }
 })
 
 // Transform compatibility props for overlay positioning
 const transformCompatProps = useTransformCompatOverlayProps()
```
christian-byrne, line 72: suggestion (non-blocking): The old code used `computed()` which tracked `props.widget.options.values` reactively. With `ref()` initialized once and refreshed only on `@show`/`@filter`, if something replaces `widget.options.values` with a new array (not via a factory function), the select won't reflect the change until the dropdown is opened. This is a behavioral regression for the plain-array case.

One option: keep `computed` for the initial/reactive path and layer the refresh logic on top for the function case, e.g. a trigger ref that gets bumped on show/filter and is read inside the computed.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
```diff
@@ -51,26 +53,27 @@ interface Props {
...
 
+function resolveValues(values: unknown): string[] {
+  if (typeof values === 'function') return values()
+  if (Array.isArray(values)) return values
+  return []
+}
+
 const modelValue = defineModel<string | undefined>({
   default(props: Props) {
     const values = props.widget.options?.values
```
christian-byrne, line 64: nitpick (non-blocking): `resolveValues` already handles the `typeof values === 'function'` check. The `default()` function could call `resolveValues(values)` instead of duplicating the typeof branch here.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    mount(WidgetSelectDefault, {
+      props: { widget },
+      global: {
+        plugins: [PrimeVue],
+        components: { SelectPlus }
+      }
+    })
+
+  describe('function-valued options', () => {
+    it('resolves options from a function', () => {
```
christian-byrne, line 66: nitpick (non-blocking): Tests only cover the function-valued path. Adding a case for plain array values and an undefined/empty fallback would give regression coverage for the `computed` -> `ref` change and document the existing behavior.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue
pythongosssss: The defineModel section is hoisted outside the script setup component, so it cant call resolveValues.
I can refactor this if you think its worthwhile, however I figured for 3 lines of code it wasnt worth it.

christian-byrne: fair enough, the defineModel default runs in a hoisted scope so it can't reference resolveValues - not worth restructuring for 3 lines

christian-byrne:
note (non-blocking): `SimplifiedWidget` uses `options?: O` defaulting to `Record<string, any>`, so the `values` property is effectively untyped when consumed by `WidgetSelectDefault.vue`. The litegraph types already have proper typing via `ComboWidgetValues`:

```typescript
type ComboWidgetValues =
  | string[]
  | Record<string, string>
  | ((widget?: IComboWidget, node?: LGraphNode) => string[])
```

But this type information is lost during the simplification step, meaning `resolveValues` has to use `unknown` and runtime checks.

Ideally, `SimplifiedWidget` (or at least the combo specialization) would carry the `ComboWidgetValues` type through so the component gets proper type narrowing instead of runtime `typeof` checks.

Relevant files:
- `src/types/simplifiedWidget.ts` - `SimplifiedWidget<T, O>` generic defaults `O` to `Record<string, any>`
- `src/lib/litegraph/src/types/widgets.ts` - `ComboWidgetValues` and `IComboWidget` already have the right types
- `src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDefault.vue` - consumer that needs the type

@coderabbitai Make an Issue to improve `SimplifiedWidget` typing for combo widget values in a followup PR. The goal is to thread `ComboWidgetValues` (or an equivalent union type) through `SimplifiedWidget` options so that `WidgetSelectDefault.vue` gets type-safe access to `values` without needing `unknown` + runtime typeof checks. Include the file paths above as jumping-off points.

christian-byrne (approved):
LGTM!

PR #8777, opened 2026-02-10
Austin/fix move subgraph input

AustinMroz:
Previously, moving a subgraph input link and re-attaching to the same input slot would result in an invalid link

broken-link

This occurred because:
- A new link is created to which overwrites the target `input.link`
- The previous link is then disconnected, which clears `input.link`

This is solved by instead returning early if the target is the same as the existing link.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8777-Austin-fix-move-subgraph-input-3036d73d365081318de3cccb926f7fe7) by [Unito](https://www.unito.io)


DrJKL (approved):
Playwright failure is the one @benceruleanlu is investigating.

PR #8785, opened 2026-02-10
fix(download): Use content-disposition filename

guill:
When we download an output, we now check if there's a filename defined in the content-disposition and use that if there is.

## Summary
This has been primarily an issue on Comfy Cloud where assets are content-addressed. Before now,
the downloaded files have retained the hash as the filename. With this change, downloaded files
will use the user-supplied filename instead.

## Changes

- **What**: Use content-disposition filename when downloading assets

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8785-fix-download-Use-content-disposition-filename-3046d73d365081ec952ef3c1930e773d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/base/common/downloadUtil.test.ts
+++ b/src/base/common/downloadUtil.test.ts
@@ -1,6 +1,9 @@
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
-import { downloadFile } from '@/base/common/downloadUtil'
+import {
+  downloadFile,
+  extractFilenameFromContentDisposition
+} from '@/base/common/downloadUtil'
 
 let mockIsCloud = false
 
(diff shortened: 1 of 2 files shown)


src/base/common/downloadUtil.test.ts
```diff
@@ -195,5 +202,147 @@ describe('downloadUtil', () => {
...
+      const fetchPromise = fetchMock.mock.results[0].value as Promise<Response>
+      await fetchPromise
+      const blobPromise = blobFn.mock.results[0].value as Promise<Blob>
+      await blobPromise
+      await Promise.resolve()
+      expect(mockLink.download).toBe('my-fallback.png')
+    })
+  })
+
+  describe('extractFilenameFromContentDisposition', () => {
```
DrJKL, line 288: Very nit: This looks like a good place to use `it.for`

DrJKL (approved):

PR #8787, opened 2026-02-11
fix: preserve prototype getter/setter in property instrumentation

Myestery:
## Summary

- Property instrumentation in `LGraphNodeProperties` was shadowing `LGraphNode`'s prototype `shape` getter/setter with a closure-based own accessor, causing `node._shape` to never be set and `renderingShape` to always return the default round shape
- Skip instrumentation when the prototype chain already defines a getter/setter for the property

## Root Cause

`LGraphNodeProperties._instrumentProperty('shape')` checked `hasOwnProperty('shape')` → `false` (defined on prototype), then created an own closure-based accessor. This shadowed the prototype setter that writes `_shape`, so `renderingShape` (which reads `_shape`) always fell back to `NODE_DEFAULT_SHAPE`.

## Test plan

- [ ] Open a workflow in legacy node mode (Nodes 2.0 off)
- [ ] Right-click a node → Shape → Box
- [ ] Verify the node visually changes to box shape
- [ ] Verify shape persists after save/load

Fixes #8532

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8787-fix-preserve-prototype-getter-setter-in-property-instrumentation-3046d73d36508158981acba2db93b5a3) by [Unito](https://www.unito.io)


christian-byrne (commented):
Forwarded from Slack for vis:

The diagnosis definitely seems correct
[4:32 PM]Also the solution seems technically like it will work for the shape case (or, solve the regression at least). But since we are ultimately just skipping shape still, it doesn't solve shape not being instrumented and also kicks down the road the fundamental shortcoming of the property instrumentation implementation we have and its inability to work with get/set. (edited) 
[4:36 PM]For some background, the reason we did this scuffed intstrumentation setup is because we had private #fields on LGraphNodes at the same which can't be accessed by external wrappers (external wrapper approach is what all good libraries do for this like vue). But we have since then actually removed all usage of #private fields, so we can now just simply improve our instrumentation stuff.

christian-byrne (commented):
we should first explore again how hard it is to just make LGraphNodes reactive. Can use some Sentry logging on cloud + codesearch.

christian-byrne:
This PR seems worth still working on @Myestery. WDYT?

Myestery:
Will come back with a new strategy

PR #8805, opened 2026-02-11
Defer vue node layout calculations on hidden browser tabs

pythongosssss:
## Summary

If you load the window in Nodes 2.0, then switch tabs while it is still loading, the position of the nodes is calculated incorrectly due to useElementBounding returning left=0, top=0 for the canvas element in a hidden tab, causing clientPosToCanvasPos to miscalculate node positions from the ResizeObserver measurements

## Changes

- **What**: 
- Store observed elements while document is in hidden state
- Re-observe when tab becomes visible

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8805-Defer-vue-node-layout-calculations-on-hidden-browser-tabs-3046d73d365081019ae6c403c0ac6d1a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
+++ b/src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
@@ -59,9 +59,37 @@ const trackingConfigs: Map<string, ElementTrackingConfig> = new Map([
   ]
 ])
 
+// Elements whose ResizeObserver fired while the tab was hidden
+const deferredElements = new Set<HTMLElement>()
+
+document.addEventListener('visibilitychange', () => {
+  if (document.visibilityState !== 'visible' || deferredElements.size === 0)
+    return
+
+  // Re-observe deferred elements to trigger fresh measurements
+  for (const element of deferredElements) {
+    if (element.isConnected) {
+      resizeObserver.unobserve(element)
+      resizeObserver.observe(element)
+    }
+  }
+  deferredElements.clear()
+})
+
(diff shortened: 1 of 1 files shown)


src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
```diff
@@ -59,9 +59,37 @@ const trackingConfigs: Map<string, ElementTrackingConfig> = new Map([
...
+
+  // Re-observe deferred elements to trigger fresh measurements
+  for (const element of deferredElements) {
+    if (element.isConnected) {
+      resizeObserver.observe(element)
+    }
+  }
+  deferredElements.clear()
+})
+
```
DrJKL: Could you try using [this](https://vueuse.org/core/useDocumentVisibility/#usedocumentvisibility)?

DrJKL: Possibly in combination with [until](https://vueuse.org/shared/until/)

DrJKL (dismissed):

src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
pythongosssss: Changed to use vueuse for visibility but used watch instead of until as if the user switches multiple times we want to handle all

DrJKL (approved):

PR #8807, opened 2026-02-11
Feat/errors tab panel

jaeone94:
## Summary

Add a dedicated **Errors tab** to the Right Side Panel that displays prompt-level, node validation, and runtime execution errors in a unified, searchable, grouped view — replacing the need to rely solely on modal dialogs for error inspection.

## Changes

- **What**:
  - **New components** (`errors/` directory):
    - `TabErrors.vue` — Main error tab with search, grouping by class type, and canvas navigation (locate node / enter subgraph).
    - `ErrorNodeCard.vue` — Renders a single error card with node ID badge, title, action buttons, and error details.
    - `types.ts` — Shared type definitions (`ErrorItem`, `ErrorCardData`, `ErrorGroup`).
  - **`executionStore.ts`** — Added `PromptError` interface, `lastPromptError` ref and `hasAnyError` computed getter. Clears `lastPromptError` alongside existing error state on execution start and graph clear.
  - **`rightSidePanelStore.ts`** — Registered `'errors'` as a valid tab value.
  - **`app.ts`** — On prompt submission failure (`PromptExecutionError`), stores prompt-level errors (when no node errors exist) into `lastPromptError`. On both runtime execution error and prompt error, deselects all nodes and opens the errors tab automatically.
  - **`RightSidePanel.vue`** — Shows the `'errors'` tab (with ⚠ icon) when errors exist and no node is selected. Routes to `TabErrors` component.
  - **`TopMenuSection.vue`** — Highlights the action bar with a red border when any error exists, using `hasAnyError`.
  - **`SectionWidgets.vue`** — Detects per-node errors by matching execution IDs to graph node IDs. Shows an error icon (⚠) and "See Error" button that navigates to the errors tab.
  - **`en/main.json`** — Added i18n keys: `errors`, `noErrors`, `enterSubgraph`, `seeError`, `promptErrors.*`, and `errorHelp*`.
  - **Testing**: 6 unit tests (`TabErrors.test.ts`) covering prompt/node/runtime errors, search filtering, and clipboard copy.
  - **Storybook**: 7 stories (`ErrorNodeCard.stories.ts`) for badge visibility, subgraph buttons, multiple errors, runtime tracebacks, and prompt-only errors.
- **Breaking**: None
- **Dependencies**: None — uses only existing project dependencies (`vue-i18n`, `pinia`, `primevue`)

## Related Work

> **Note**: Upstream PR #8603 (`New bottom button and badges`) introduced a separate `TabError.vue` (singular) that shows per-node errors when a specific node is selected. Our `TabErrors.vue` (plural) provides the **global error overview** — a different scope. The two tabs coexist:
> - `'error'` (singular) → appears when a node with errors is selected → shows only that node's errors
> - `'errors'` (plural) → appears when no node is selected and errors exist → shows all errors grouped by class type
>
> A future consolidation of these two tabs may be desirable after design review.

## Architecture

```
executionStore
├── lastPromptError: PromptError | null     ← NEW (prompt-level errors without node IDs)
├── lastNodeErrors: Record<string, NodeError>  (existing)
├── lastExecutionError: ExecutionError         (existing)
└── hasAnyError: ComputedRef<boolean>       ← NEW (centralized error detection)

TabErrors.vue (errors tab - global view)
├── errorGroups: ComputedRef<ErrorGroup[]>  ← normalizes all 3 error sources
├── filteredGroups                          ← search-filtered view
├── locateNode()                            ← pan canvas to node
├── enterSubgraph()                         ← navigate into subgraph
└── ErrorNodeCard.vue                       ← per-node card with copy/locate actions

types.ts
├── ErrorItem      { message, details?, isRuntimeError? }
├── ErrorCardData  { id, title, nodeId?, errors[] }
└── ErrorGroup     { title, cards[], priority }
```

## Review Focus

1. **Error normalization logic** (`TabErrors.vue` L75–150): Three different error sources (prompt, node validation, runtime) are normalized into a common `ErrorGroup → ErrorCardData → ErrorItem` hierarchy. Edge cases to verify:
   - Prompt errors with known vs unknown types (known types use localized descriptions)
   - Multiple errors on the same node (grouped into one card)
   - Runtime errors with long tracebacks (capped height with scroll)

2. **Canvas navigation** (`TabErrors.vue` L210–250): The `locateNode` and `enterSubgraph` functions navigate to potentially nested subgraphs. The double `requestAnimationFrame` is required due to LiteGraph's asynchronous subgraph switching — worth verifying this timing is sufficient.

3. **Store getter consolidation**: `hasAnyError` replaces duplicated logic in `TopMenuSection` and `RightSidePanel`. Confirm that the reactive dependency chain works correctly (it depends on 3 separate refs).

4. **Coexistence with upstream `TabError.vue`**: The singular `'error'` tab (upstream, PR #8603) and our plural `'errors'` tab serve different purposes but share similar naming. Consider whether a unified approach is preferred.

## Test Results

```
✓ renders "no errors" state when store is empty
✓ renders prompt-level errors (Group title = error message)
✓ renders node validation errors grouped by class_type
✓ renders runtime execution errors from WebSocket
✓ filters errors based on search query
✓ calls copyToClipboard when copy button is clicked

Test Files  1 passed (1)
     Tests  6 passed (6)
```

image
image
image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8807-Feat-errors-tab-panel-3046d73d36508127981ac670a70da467) by [Unito](https://www.unito.io)


src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -57,8 +61,41 @@ watchEffect(() => (widgets.value = widgetsProp))
...
+    if (graphNode && String(graphNode.id) === String(nodeId)) return true
+  }
+
+  return false
+})
+
+/** Switches the right side panel to the Errors tab. */
+function navigateToErrorTab() {
+  rightSidePanelStore.openPanel('errors')
+}
```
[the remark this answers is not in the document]
jaeone94: I modified it in the next commit so that the button to navigate to the relevant global error tab is only visible in the widgets of the global node tab.
[fdb93ba](https://github.com/Comfy-Org/ComfyUI_frontend/pull/8807/commits/fdb93ba61dddc66edbfeede9825cd59f693935ae)

src/stores/executionStore.ts
```diff
@@ -104,6 +104,13 @@ function executionIdToNodeLocatorId(
   return nodeLocatorId
 }
 
+/** Prompt-level errors without node IDs (e.g. invalid_prompt, prompt_no_outputs) */
```
viva-jinyi: nit: Could you move this part to the top of the file?

src/stores/executionStore.ts
```diff
@@ -641,13 +650,23 @@ export const useExecutionStore = defineStore('execution', () => {
     }
   })
 
+  /** Whether any error (node validation, runtime execution, or prompt-level) is present */
+  const hasAnyError = computed(
```
viva-jinyi, line 661:
```
const hasNodeErrors = computed(
  () => Object.keys(lastNodeErrors.value ?? {}).length > 0
)

const hasAnyError = computed(
  () => Boolean(lastExecutionError.value || lastPromptError.value || hasNodeErrors.value)
)
```

viva-jinyi: WDYT?

src/components/rightSidePanel/errors/ErrorNodeCard.vue
```diff
@@ -0,0 +1,120 @@
...
+          v-if="error.details"
+          :class="
+            cn(
+              'rounded-lg bg-secondary-background-hover p-2.5 overflow-y-auto border border-interface-stroke/30',
+              error.isRuntimeError ? 'max-h-[10lh]' : 'max-h-[6lh]'
+            )
+          "
+        >
+          <p
+            class="m-0 text-[11px] text-muted-foreground break-words whitespace-pre-wrap font-mono leading-relaxed"
```
viva-jinyi: Can you use `text-sm` or `text-xs` instead of fixed font size?

src/components/rightSidePanel/errors/ErrorNodeCard.vue
```diff
@@ -0,0 +1,120 @@
...
+          >
+            {{ error.details }}
+          </p>
+        </div>
+
+        <Button
+          variant="secondary"
+          size="sm"
+          class="w-full justify-center gap-2 h-8 text-[11px]"
+          @click="
```
viva-jinyi: I think it would be better to make this into a function

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -57,8 +62,43 @@ watchEffect(() => (widgets.value = widgetsProp))
...
+const rightSidePanelStore = useRightSidePanelStore()
 const { t } = useI18n()
 
+/**
+ * Checks whether the current node has errors from lastNodeErrors or lastExecutionError.
+ * Uses execution ID mapping to match graph node IDs.
+ */
+const nodeHasError = computed(() => {
+  // Only show error indicator in workflow overview (nothing selected on canvas)
+  if (canvasStore.selectedItems.length > 0) return false
```
viva-jinyi: very nit
`if (canvasStore.selectedItems.length > 0 || !targetNode.value) return false`

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,390 @@
+<script setup lang="ts">
```
viva-jinyi, line 99: Could you check this file? 
https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/components/rightSidePanel/TabError.vue
I think we should consolidate the TabErrors.vue and TabError.vue files.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,390 @@
...
+      !targetGraph.isRootGraph && targetGraph instanceof Subgraph
+        ? targetGraph
+        : undefined
+    canvas.setGraph(targetGraph)
+
+    // Ensure DOM and graph state are synchronised
+    await nextTick()
+
+    // Double RAF to wait for LiteGraph's internal canvas frame cycle
+    await new Promise((resolve) =>
```
viva-jinyi: Can't we use a callback that signals subgraph transition completion instead of double `requestAnimationFrame`? This timing-based approach is fragile across different hardware.

src/stores/executionStore.ts
```diff
@@ -104,6 +104,13 @@ function executionIdToNodeLocatorId(
   return nodeLocatorId
 }
 
+/** Prompt-level errors without node IDs (e.g. invalid_prompt, prompt_no_outputs) */
+interface PromptError {
```
viva-jinyi: Wouldn't it be better to define `type` as an enum (or string literal union) instead of a plain `string`? Currently `KNOWN_PROMPT_ERROR_TYPES` in `TabErrors.vue` is a loose `Set<string>` — if a new error type is added, there's no compile-time guarantee that the i18n key exists.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,390 @@
...
+    NodeBadgeMode.None
+)
+
+// ─── Helpers ────────────────────────────────────────────────────────────────
+
+/**
+ * Resolves the display title for a node given its execution ID.
+ * @param nodeId The execution ID of the node.
+ */
+function resolveNodeTitle(nodeId: string): string {
```
viva-jinyi: `resolveNodeTitle` calls `getNodeByExecutionId(app.rootGraph, nodeId)` inside a `computed`. Since `app.rootGraph` is not a Vue reactive object, will this computed actually re-evaluate when the graph changes (e.g. node renamed)?

src/locales/en/main.json
```diff
@@ -2936,7 +2936,23 @@
...
+    "seeError": "See Error",
+    "promptErrors": {
+      "prompt_no_outputs": {
+        "desc": "The workflow does not contain any output nodes (e.g. Save Image, Preview Image) to produce a result."
+      },
+      "no_prompt": {
+        "desc": "The workflow data sent to the server is empty. This may be an unexpected system error."
+      }
+    },
+    "errorHelpPrefix": "For more help, ",
```
viva-jinyi: Instead of splitting the help sentence into separate tokens (`errorHelpPrefix`, `errorHelpOr`, etc.), it would be better to use interpolation variables for `github` and `support`.

src/components/TopMenuSection.vue
```diff
@@ -242,6 +250,8 @@ const shouldShowRedDot = computed((): boolean => {
   return releaseRedDot || shouldShowConflictRedDot.value
 })
 
+const { hasAnyError: hasPromptError } = storeToRefs(executionStore)
```
viva-jinyi: `hasAnyError` is renamed to `hasPromptError` via destructuring, but this computed covers node errors and runtime errors too — not just prompt errors. Why was it renamed?

src/components/rightSidePanel/errors/ErrorNodeCard.vue
```diff
@@ -0,0 +1,120 @@
...
+        class="flex-1 text-sm text-muted-foreground truncate font-medium"
+      >
+        {{ card.nodeTitle }}
+      </span>
+      <Button
+        v-if="card.isSubgraphNode"
+        variant="secondary"
+        size="sm"
+        class="rounded-lg text-sm shrink-0"
+        @click.stop="emit('enterSubgraph', card.nodeId!)"
```
viva-jinyi: nit: Can we use a fallback instead of non-null assertion(`!`)?

src/scripts/app.ts
```diff
@@ -701,6 +702,8 @@ export class ComfyApp {
       } else {
         useDialogService().showExecutionErrorDialog(detail)
       }
+      this.canvas.deselectAll()
```
viva-jinyi: I think deselecting nodes without user intent is not the right UX. If a user is in the middle of configuring a node and an error occurs, having their selection forcefully cleared is disruptive. Errors should be notified, not result in clearing the user's current working context.

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -57,8 +62,43 @@ watchEffect(() => (widgets.value = widgetsProp))
...
+  if (!targetNode.value) return false
+  const nodeId = targetNode.value.id
+
+  // Check lastNodeErrors (validation errors from 400 Bad Request)
+  if (executionStore.lastNodeErrors) {
```
viva-jinyi: `nodeHasError` iterates over all keys in lastNodeErrors and calls `getNodeByExecutionId` (which traverses the graph) for each one. Since `SectionWidgets` is rendered per-node, this results in O(n²) complexity. Could we instead pre-compute a `Set<nodeId>` of error node IDs in the store as a computed, so each component can just do `set.has(myId)` for O(1) lookup?

viva-jinyi (commented):
Our project convention is that code should be self-expressive and comments should be minimized. It would be better to remove decorative separator comments (e.g. `// ─── Sample Data ───...`) and redundant JSDoc that just restates parameter names. Also, In our project, `<template>` goes at the top and `<script>` at the bottom.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Since LiteGraph core doesn't broadcast draw completion, double RAF is our current workaround for canvas synchronization, consistent with our existing patterns. Introducing a callback refers back to LiteGraph core changes, so I suggest tracking this as a separate enhancement.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: I’ve looked into TabError.vue, and I believe it is intended for the node-specific error view when a node is selected.

While they share some similarities, I believe they should remain separate for now because their internal structural hierarchy will differ significantly:

TabErrors.vue
 (Plural): A global overview that groups errors by class_type and includes individual node cards as children.
TabError.vue (Singular): A focused view for a specific node, where error messages themselves will likely serve as the primary group headers.
because it's not just a matter of filtering data but a difference in the component's UI architecture, merging them might introduce unnecessary complexity and conditional branching in the template.

Furthermore, my plan was to focus on the global error overview in this PR and address the node-specific tab in the next one. We can revisit refactoring them into shared sub-components once the patterns for both are fully established.

Does this approach make sense to you, or would you still prefer to see them consolidated at this stage?

src/stores/executionStore.ts
jaeone94: `PromptError.type` is a value received directly from the ComfyUI backend, which can introduce new error types at any time (future updates). 
Strictly narrowing it to an enum or union could actually reduce extensibility, and risks silently dropping errors that the frontend doesn't yet recognize.

Regarding the i18n concern — the code already handles this gracefully: if the error type is in `KNOWN_PROMPT_ERROR_TYPES`, the corresponding i18n key is used; otherwise, the raw `error.message` from the backend is displayed as a fallback (see lines 99–101). So even if a new error type is added without a matching i18n key, users will still see a meaningful error message.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Good observation regarding reactivity. I verified this through testing — in practice, renaming a node requires clicking on it first, which switches the right side panel away from the errors tab. When the user returns to the errors tab, the `errorGroups` computed naturally re-evaluates and resolves the updated title. So while `app.rootGraph` is not reactive, the UI flow inherently ensures the title is always up to date when the errors tab is visible.

src/scripts/app.ts
jaeone94: I understand the concern. However, this error overview is ultimately designed as a **complete replacement for the error modal dialog**.

Consider the scenario: a user is editing a specific node (right side panel shows that node's parameters) and clicks the Run button. If an error occurs but we don't deselect, the right side panel stays on that node's parameter tab — and the user has **no immediate visual feedback** that something went wrong. The error information would be hidden behind a tab they'd have to manually navigate to.

With the current behavior, `deselectAll()` switches the right side panel to the Workflow Overview, which automatically surfaces the Errors tab with a clear summary of what failed. This mirrors the UX intent of the old modal dialog — **errors demand immediate attention** — but in a less intrusive, non-blocking way.

That said, I'm open to exploring alternatives if needed. For example, we could show a toast notification instead and let the user decide when to view the errors. But that would mean users could miss critical errors, which is the exact problem the error modal was solving. What do you think?

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+              @enter-subgraph="enterSubgraph"
+              @copy-to-clipboard="copyToClipboard"
+            />
+          </div>
+        </PropertiesAccordionItem>
+      </div>
+    </div>
+
+    <!-- Fixed Footer: Help Links -->
+    <div class="shrink-0 border-t border-interface-stroke px-4 py-4 min-w-0">
```
AustinMroz:
```suggestion
    <div class="shrink-0 border-t border-interface-stroke p-4 min-w-0">
```

AustinMroz (commented):
Jin Yi was already super thorough here. I only saw a single trivial nit.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Fixed!

src/components/rightSidePanel/errors/TabErrors.vue
viva-jinyi: @AustinMroz since you have more context than me, WDYT?

jaeone94:
I’ve added a commit to register a new setting, Comfy.RightSidePanel.ShowErrorsTab, and wrapped the new features behind this flag to keep them hidden by default. (Apologies for not doing this in the first place!)

Because I wanted to implement a solid baseline where the new UI and core logic work together, this PR has grown quite large. Please let me know if the size is an issue—I’m happy to discuss how to break it down further if you’d prefer!

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: @viva-jinyi  
After a more thorough code review, I’ve decided that it would be better to merge `TabError` into `TabErrors`. To prevent this PR from getting any larger, I will include this change in the next PR.

src/scripts/app.ts
jaeone94: @viva-jinyi 
To address the concern you raised, I’ve discussed this with the team and we are trying out a non-intrusive overlay positioned below the TopMenuSection (PR #8875).

This allows us to notify the user of errors without interrupting their current workflow or forcing a change in node selection. We are currently awaiting feedback from the QA team on this improved UX.

src/stores/executionStore.ts
jaeone94: Done

src/stores/executionStore.ts
jaeone94: Done

src/components/rightSidePanel/errors/ErrorNodeCard.vue
jaeone94: Done

src/components/rightSidePanel/errors/ErrorNodeCard.vue
jaeone94: Done

src/components/rightSidePanel/parameters/SectionWidgets.vue
jaeone94: Done

src/locales/en/main.json
jaeone94: Done

src/components/TopMenuSection.vue
jaeone94: Done

src/components/rightSidePanel/parameters/SectionWidgets.vue
jaeone94: Done. 
The per-node iteration over `lastNodeErrors` with `getNodeByExecutionId` has been replaced with a pre-computed `errorNodeIds` Set in the execution store, giving each component an O(1) `set.has()` lookup. 
The graph traversal now runs once in the store's computed, not per-SectionWidgets instance.

src/components/rightSidePanel/errors/ErrorNodeCard.vue
jaeone94: Done. 
Replaced inline `card.nodeId!` assertions with dedicated handler functions (`handleLocateNode`, `handleEnterSubgraph`) that guard with if (`card.nodeId`) before emitting.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+function resolveNodeTitle(nodeId: string): string {
+  const graphNode = getNodeByExecutionId(app.rootGraph, nodeId)
+  return graphNode?.title || ''
+}
+
+function isSubgraphId(nodeId: string): boolean {
+  return nodeId.includes(':')
+}
+
+const errorGroups = computed<ErrorGroup[]>(() => {
```
christian-byrne: Do you think we could clean up this computed a bit? We could consider doing things like:

1. Moving groupsMap into its own computed value in the module scope, add Map cleanup in unmounted hooks
2. Extract out pure functions like getOrCreateGroup
3. Refactor repeated logic of constructing the groups.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+)
+
+function resolveNodeTitle(nodeId: string): string {
+  const graphNode = getNodeByExecutionId(app.rootGraph, nodeId)
+  return graphNode?.title || ''
+}
+
+function isSubgraphId(nodeId: string): boolean {
+  return nodeId.includes(':')
+}
```
christian-byrne: The codebase should have helpers available already for the "is this node a subgraph node" semantics that we can re-use.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+import { NodeBadgeMode } from '@/types/nodeSource'
+
+import PropertiesAccordionItem from '../layout/PropertiesAccordionItem.vue'
+import FormSearchInput from '@/renderer/extensions/vueNodes/widgets/components/form/FormSearchInput.vue'
+import ErrorNodeCard from './ErrorNodeCard.vue'
+import Button from '@/components/ui/button/Button.vue'
+import type { ErrorCardData, ErrorGroup } from './types'
+
+/** Prompt error types that have predefined localized descriptions. */
+const KNOWN_PROMPT_ERROR_TYPES = new Set(['prompt_no_outputs', 'no_prompt'])
```
christian-byrne: This could be moved into business logic and outside of vue component file.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+/** Whether to show node ID badges, based on the user's LiteGraph Node ID Badge Mode setting */
+const showNodeIdBadge = computed(
+  () =>
+    (settingStore.get('Comfy.NodeBadge.NodeIdBadgeMode') as NodeBadgeMode) !==
+    NodeBadgeMode.None
+)
+
+function resolveNodeTitle(nodeId: string): string {
+  const graphNode = getNodeByExecutionId(app.rootGraph, nodeId)
+  return graphNode?.title || ''
```
christian-byrne: You can use this after you have the reference to the node: https://github.com/Comfy-Org/ComfyUI_frontend/blob/e83e396c096158fe002e2c4fe8989b8e3af3b9dd/src/utils/nodeTitleUtil.ts#L16

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+      cards: Array.from(groupData.cards.values()),
+      priority: groupData.priority
+    }))
+    .sort((a, b) => {
+      if (a.priority !== b.priority) return a.priority - b.priority
+      return a.title.localeCompare(b.title)
+    })
+})
+
+const filteredGroups = computed<ErrorGroup[]>(() => {
```
christian-byrne: nit: we can use our Fuse library, following other components that have search bars.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+            (e) =>
+              e.message.toLowerCase().includes(query) ||
+              e.details?.toLowerCase().includes(query)
+          )
+      )
+
+      return { ...group, cards: matchedCards }
+    })
+    .filter((group) => group.cards.length > 0)
+})
```
christian-byrne: The business logic of searching data should be moved out of the view/component layer (into a typescript business logic module).

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+
+watch(
+  () => rightSidePanelStore.focusedErrorNodeId,
+  (graphNodeId) => {
+    if (!graphNodeId) return
+    for (const group of filteredGroups.value) {
+      const hasMatch = group.cards.some((card) => {
+        if (!card.nodeId) return false
+        const graphNode = getNodeByExecutionId(app.rootGraph, card.nodeId)
+        return graphNode && String(graphNode.id) === graphNodeId
```
christian-byrne: It it possible to store the id on the cards/group data when its constructed to avoid needing to recompute on every reactivity trigger?

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+    await nextTick()
+
+    // Double RAF to wait for LiteGraph's internal canvas frame cycle
+    await new Promise((resolve) =>
+      requestAnimationFrame(() => requestAnimationFrame(resolve))
+    )
+  }
+}
+
+async function locateNode(nodeId: string) {
```
christian-byrne: nit: `focusNode` or `fitViewToNode` seems like a more appropriate name for this function. I would also move this logic out of the component file.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+    string,
+    { priority: number; cards: Map<string, ErrorCardData> }
+  >()
+
+  const getOrCreateGroup = (title: string, priority = 1) => {
+    if (!groupsMap.has(title)) {
+      groupsMap.set(title, { priority, cards: new Map() })
+    }
+    return groupsMap.get(title)!.cards
+  }
```
christian-byrne: Is there a way to do Map access/insert without needing non-null assertion?

src/scripts/app.ts
```diff
@@ -1468,6 +1475,37 @@ export class ComfyApp {
...
+                    type: 'error',
+                    message: respError,
+                    details: ''
+                  }
+                }
+              }
+
+              // Clear selection and open the error panel so the user can immediately
+              // see the error details without extra clicks.
+              this.canvas.deselectAll()
```
christian-byrne: This should go inside the if block below I believe.

src/stores/executionStore.ts
```diff
@@ -47,6 +47,13 @@ interface QueuedPrompt {
   workflow?: ComfyWorkflow
 }
 
+/** Prompt-level errors without node IDs (e.g. invalid_prompt, prompt_no_outputs) */
+interface PromptError {
```
christian-byrne: We should try to define this in https://github.com/Comfy-Org/ComfyUI_frontend/blob/631d4849011a36e7c53b073f3f013f990812e713/src/schemas/apiSchema.ts#L4

src/stores/executionStore.ts
```diff
@@ -641,13 +650,53 @@ export const useExecutionStore = defineStore('execution', () => {
...
 
+  /** Whether a runtime execution error is present */
+  const hasExecutionError = computed(() => !!lastExecutionError.value)
+
+  /** Whether a prompt-level error is present (e.g. invalid_prompt, prompt_no_outputs) */
+  const hasPromptError = computed(() => !!lastPromptError.value)
+
+  /** Whether any node validation errors are present */
+  const hasNodeError = computed(
+    () => !!lastNodeErrors.value && Object.keys(lastNodeErrors.value).length > 0
```
christian-byrne: nit: can use negation of https://es-toolkit.dev/reference/compat/predicate/isEmpty.html to make it more expressive.

src/stores/executionStore.ts
```diff
@@ -641,13 +650,53 @@ export const useExecutionStore = defineStore('execution', () => {
...
+  )
+
+  /** Whether any error (node validation, runtime execution, or prompt-level) is present */
+  const hasAnyError = computed(
+    () => hasExecutionError.value || hasPromptError.value || hasNodeError.value
+  )
+
+  /** Pre-computed Set of graph node IDs (as strings) that have errors. */
+  const errorNodeIds = computed<Set<string>>(() => {
```
christian-byrne: If this is coercing from execution ID to node ID, we lose granularity that I think we actually still want. E.g., what if there is an error on node with execution ID "12:2" and another error on a node with exeuction ID "7:2" -- they would shadow one another?

src/components/rightSidePanel/errors/ErrorNodeCard.vue
```diff
@@ -0,0 +1,111 @@
...
+        class="rounded-lg text-sm shrink-0"
+        @click.stop="emit('enterSubgraph', card.nodeId ?? '')"
+      >
+        {{ t('rightSidePanel.enterSubgraph') }}
+      </Button>
+      <Button
+        variant="textonly"
+        size="icon-sm"
+        class="size-7 text-muted-foreground hover:text-base-foreground shrink-0"
+        @click.stop="emit('locateNode', card.nodeId!)"
```
christian-byrne: Are we actually safe asserting that card.nodeId is never null here?

src/stores/executionStore.ts
christian-byrne, line 1: You may need to clear `lastPromptError` in https://github.com/Comfy-Org/ComfyUI_frontend/blob/631d4849011a36e7c53b073f3f013f990812e713/src/stores/executionStore.ts#L448-L465

src/stores/executionStore.ts
```diff
@@ -641,13 +650,53 @@ export const useExecutionStore = defineStore('execution', () => {
...
+        if (graphNode) ids.add(String(graphNode.id))
+      }
+    }
+
+    // Runtime execution error (WebSocket)
+    if (lastExecutionError.value) {
+      const execNodeId = String(lastExecutionError.value.node_id)
+      const graphNode = getNodeByExecutionId(app.rootGraph, execNodeId)
+      if (graphNode) ids.add(String(graphNode.id))
+    }
```
christian-byrne: nit: this is a great place for some memoization on the nested traversals -- although I would never expect there to be a significant number of errors from one job.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne, line 1: In general I wonder if we can make the code more expressive. It wasn't clear to me until spending a long time reading the semantic of "if no node is selected, use 'errors'; if a node is selected, use argument 'error'". I wonder if we can write the code/APIs in a way that this is immediately obvious to future maintainers.

src/components/rightSidePanel/errors/ErrorNodeCard.stories.ts
```diff
@@ -0,0 +1,170 @@
...
+  title: 'Prompt has no outputs.',
+  errors: [
+    {
+      message:
+        'The workflow does not contain any output nodes (e.g. Save Image, Preview Image) to produce a result.'
+    }
+  ]
+}
+
+// ─── Stories: Badge Visibility ──────────────────────────────────────────────
```
christian-byrne: Can we remove these organizational type of comments.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+import Button from '@/components/ui/button/Button.vue'
+import type { ErrorCardData, ErrorGroup } from './types'
+
+/** Prompt error types that have predefined localized descriptions. */
+const KNOWN_PROMPT_ERROR_TYPES = new Set(['prompt_no_outputs', 'no_prompt'])
+
+const executionStore = useExecutionStore()
+const canvasStore = useCanvasStore()
+const { t } = useI18n()
+const { copyToClipboard: globalCopy } = useCopyToClipboard()
```
christian-byrne: Why do we need to alias the function here?

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+  { immediate: true }
+)
+
+async function copyToClipboard(text: string) {
+  try {
+    await globalCopy(text)
+  } catch (err) {
+    console.error('Failed to copy error details to clipboard:', err)
+  }
+}
```
christian-byrne: Is this needed? See the function being imported already -- does it handle errors well already? https://github.com/Comfy-Org/ComfyUI_frontend/blob/631d4849011a36e7c53b073f3f013f990812e713/src/composables/useCopyToClipboard.ts#L6-L23

If not, we should just improve that shared composable directly, then we don't need to alias it and re-wrap it in this component file here.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+
+  const graphNode = getNodeByExecutionId(app.rootGraph, nodeId)
+  if (!graphNode?.graph) return
+
+  await navigateToGraph(graphNode.graph as LGraph)
+  useLitegraphService().fitView()
+}
+
+const REPO_OWNER = 'comfyanonymous'
+const REPO_NAME = 'ComfyUI'
```
christian-byrne: Business logic constants can be moved out of component file

src/stores/workspace/rightSidePanelStore.ts
```diff
@@ -33,6 +34,7 @@ export const useRightSidePanelStore = defineStore('rightSidePanel', () => {
   const activeTab = ref<RightSidePanelTab>('parameters')
   const isEditingSubgraph = computed(() => activeTab.value === 'subgraph')
   const focusedSection = ref<RightSidePanelSection | null>(null)
+  const focusedErrorNodeId = ref<string | null>(null)
```
christian-byrne, line 42: I'm not sure if the variable name alone really indicates what this API does, what about something like:

```suggestion
  /** Graph node ID to focus in the errors tab. Set by SectionWidgets and cleared by TabErrors after handling. */
  const focusedErrorNodeId = ref<string | null>(null)
```

Alternatively, we could create a more expressive variable name, WDYT?

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -0,0 +1,360 @@
...
+
+// Double RAF is needed to wait for LiteGraph's internal canvas frame cycle
+// after subgraph switching before we can focus on a node bounding box.
+async function navigateToGraph(targetGraph: LGraph) {
+  const canvas = canvasStore.canvas
+  if (!canvas) return
+
+  if (canvas.graph !== targetGraph) {
+    canvas.subgraph =
+      !targetGraph.isRootGraph && targetGraph instanceof Subgraph
```
christian-byrne: Would it be better to just define a (or use existing) `isSubgraph` type guard in the relevant files and import that (rather than importing entire `Subgraph` class in component file)?

src/components/rightSidePanel/errors/ErrorNodeCard.vue
jaeone94: Safe — guarded by v-if="card.nodeId" on line 5 — but replaced the ! assertion with ?? '' to match the pattern on line 23 and avoid non-null assertions entirely

src/renderer/extensions/vueNodes/components/LGraphNode.vue
jaeone94: Good point! The 'error' (single-node) vs 'errors' (all errors) distinction was indeed confusing. This has been resolved in the follow-up PR where 'error' tab is removed entirely and everything is unified under 'errors'. The code you see in the next PR uses:
```
typescript
v-if="hasAnyError && showErrorsTabEnabled"
@click.stop="useRightSidePanelStore().openPanel('errors')"
```
https://github.com/Comfy-Org/ComfyUI_frontend/pull/8875/changes#diff-a7744614cf842e54416047326db79ad81f7c7ab7bfb66ae2b46f5c73ac7d47f2R136-R138

src/stores/executionStore.ts
jaeone94: Good catch! Fixed by scoping `errorNodeIds` (now renamed to `activeGraphErrorNodeIds`) to the currently active graph via `graphNode.graph === activeGraph`. 
This ensures that execution IDs like `"12:2"` and `"7:2"` — which share the same local ID `"2"` — never shadow each other, since only nodes belonging to the graph the user is currently viewing are included in the set.

src/stores/executionStore.ts
jaeone94: Acknowledged. `getNodeByExecutionId` traverses the graph tree per call, but since error counts per job are typically small (as you noted), we'll track this as a future optimization.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Extracted into `useErrorGroups.ts` composable with:
- `buildErrorGroups()` — delegates to three focused functions: `processPromptError`, `processNodeErrors`, `processExecutionError`
- `getOrCreateGroup()` — reusable pure function
- `toSortedGroups()` — Map → sorted array conversion

`TabErrors.vue` now just calls `useErrorGroups(searchQuery, t)` and consumes the result.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Replaced with `isNodeExecutionId` from `@/types/nodeIdentification`.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Moved to `useErrorGroups.ts` as a module-level constant.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Created `resolveNodeInfo()` in `useErrorGroups.ts` that uses `resolveNodeDisplayName` from `nodeTitleUtil.ts`.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. `searchErrorGroups()` in `useErrorGroups.ts` now uses Fuse.js with weighted keys (`searchableNodeId`, `searchableNodeTitle`, `searchableMessage`, `searchableDetails`).

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Extracted as `searchErrorGroups()` pure function in `useErrorGroups.ts`.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Added `graphNodeId` field to `ErrorCardData` type and pre-computed it in `resolveNodeInfo()` when the error groups are constructed, so downstream consumers (e.g. the `focusedErrorNodeId` watcher) can do a simple equality check instead of re-traversing the graph tree.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Renamed to `focusNode` and extracted into `composables/canvas/useFocusNode.ts` composable alongside `enterSubgraph` and `navigateToGraph`.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Changed to `let entry = groupsMap.get(title)` pattern — if undefined, create and set, then return `entry.cards` without `!`.

src/scripts/app.ts
jaeone94: Done. Moved `deselectAll()` inside the conditional block.

src/stores/executionStore.ts
jaeone94: Done. Defined `zPromptError` zod schema in `apiSchema.ts` and exported `PromptError` type. Removed the local interface from `executionStore.ts`.

src/stores/executionStore.ts
jaeone94: Done. Now uses `!isEmpty(lastNodeErrors.value)` with import from `es-toolkit/compat`.

src/stores/executionStore.ts
jaeone94: Done. Added `lastPromptError.value = null` in `resetExecutionState()`.

src/components/rightSidePanel/errors/ErrorNodeCard.stories.ts
jaeone94: Done. Removed all section divider comments (`// ─── Stories: ... ───`).

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Removed the alias — now destructures `copyToClipboard` directly and passes it to the template without wrapping.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Removed `handleCopyToClipboard` wrapper entirely. `useCopyToClipboard` already handles errors with toast notifications internally.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Removed the constants entirely and replaced the hardcoded URL with `staticUrls.githubIssues` from `useExternalLink()` composable.

src/stores/workspace/rightSidePanelStore.ts
jaeone94: Done. Added JSDoc exactly as suggested:
```typescript
/** Graph node ID to focus in the errors tab. Set by SectionWidgets and cleared by TabErrors after handling. */
const focusedErrorNodeId = ref<string | null>(null)
```

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Changed `Subgraph` to type-only import and replaced `instanceof Subgraph` with `isRootGraph` property check + `as Subgraph` cast. No runtime class import needed.

src/components/rightSidePanel/errors/ErrorNodeCard.vue
christian-byrne: Oh, didn't see that. But nice!

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: Nice!

christian-byrne (approved):
LGTM!

PR #8837, opened 2026-02-12
Fix hover state on collapsed bottom button

AustinMroz:
| Before | After |
| ------ | ----- |
| before | after|

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8837-Fix-hover-state-on-collapsed-bottom-button-3056d73d36508139919fc1d750c6b135) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/LGraphNode.vue
+++ b/src/renderer/extensions/vueNodes/components/LGraphNode.vue
@@ -126,8 +126,9 @@
       variant="textonly"
       :class="
         cn(
-          'w-full h-12 rounded-b-2xl -mt-5 pt-7 pb-2 -z-1 text-xs',
-          hasAnyError && 'hover:bg-destructive-background-hover'
+          'w-full h-7 rounded-b-2xl py-2 -z-1 text-xs rounded-t-none',
+          hasAnyError && 'hover:bg-destructive-background-hover',
+          !isCollapsed && '-mt-5 pt-7 h-12'
         )
       "
       as-child


src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -126,8 +126,9 @@
       variant="textonly"
       :class="
         cn(
-          'w-full h-12 rounded-b-2xl -mt-5 pt-7 pb-2 -z-1 text-xs',
-          hasAnyError && 'hover:bg-destructive-background-hover'
+          'w-full h-7 rounded-b-2xl py-2 -z-1 text-xs rounded-t-none',
+          hasAnyError && 'hover:bg-destructive-background-hover',
+          !isCollapsed && '-mt-5 pt-7 h-12'
```
DrJKL, line 131: Just flagging: Negative margins can behave strangely in a few browsers.

DrJKL (approved):

src/renderer/extensions/vueNodes/components/LGraphNode.vue
AustinMroz: Appreciate the flag.
I've now tested this in on chromium, firefox, and a crusty webkit browser that crashloops if you ever make the mistake of adding a loadaudio to the graph. I'm pretty confident the negative margine won't be an issue here.

Though I will hold off on merging for a bit to give design a chance to review.

PR #8845, opened 2026-02-13
feat(vueNodes): support resizing from all four corners

jtydhr88:
## Summary
(Not sure we need this, and I don't know the reason why we only have one cornor support previously, but it is requested by QA reporting in Notion)

Add resize handles at all four corners (NW, NE, SW, SE) of Vue nodes, matching litegraph's multi-corner resize behavior.                                                                 

Vue nodes previously only supported resizing from the bottom-right (SE) corner. This adds handles at all four corners with direction-aware delta math, snap-to-grid support, and minimum size enforcement that keeps the opposite corner anchored. 
The content-driven minimum height is measured from the DOM at resize start to prevent the node from sliding when dragged past its minimum size.

https://github.com/user-attachments/assets/c9d30d93-8243-4c44-a417-5ca6e9978b3b

### DIFF
--- a/src/renderer/extensions/vueNodes/interactions/resize/resizeHandleConfig.ts
+++ b/src/renderer/extensions/vueNodes/interactions/resize/resizeHandleConfig.ts
@@ -1,0 +1,45 @@
+import type { CompassCorners } from '@/lib/litegraph/src/interfaces'
+
+interface ResizeHandle {
+  corner: CompassCorners
+  positionClasses: string
+  cursorClass: string
+  i18nKey: string
+  svgPositionClasses: string
+  svgTransform: string
+}
+
+export const RESIZE_HANDLES: ResizeHandle[] = [
+  {
+    corner: 'SE',
+    positionClasses: '-right-1 -bottom-1',
+    cursorClass: 'cursor-se-resize',
+    i18nKey: 'g.resizeFromBottomRight',
+    svgPositionClasses: 'top-1 left-1',
+    svgTransform: ''
+  },
+  {
+    corner: 'NE',
+    positionClasses: '-right-1 -top-1',
+    cursorClass: 'cursor-ne-resize',
+    i18nKey: 'g.resizeFromTopRight',
+    svgPositionClasses: 'bottom-1 left-1',
+    svgTransform: 'scaleY(-1)'
+  },
+  {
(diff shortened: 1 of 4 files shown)


AustinMroz (approved):
The functionality was removed in #7063 because no one knew of a way to prevent resizing a node past it's minimum size causing the nodes position to instead change.

There's still  some remnants of this behaviour when resizing a node that contains widgets of fixed aspect ratio from the top, but I consider this super nit-picky and a non-blocker.
top-resize

src/renderer/extensions/vueNodes/interactions/resize/resizeHandleConfig.ts
```diff
@@ -0,0 +1,45 @@
...
+  },
+  {
+    corner: 'NW',
+    positionClasses: '-left-1 -top-1',
+    cursorClass: 'cursor-nw-resize',
+    i18nKey: 'g.resizeFromTopLeft',
+    svgPositionClasses: 'bottom-1 right-1',
+    svgTransform: 'scale(-1, -1)'
+  }
+]
```
DrJKL, line 45:
```suggestion
] as const
```

src/renderer/extensions/vueNodes/interactions/resize/resizeHandleConfig.ts
```diff
@@ -0,0 +1,45 @@
+import type { CompassCorners } from '@/lib/litegraph/src/interfaces'
+
+interface ResizeHandle {
+  corner: CompassCorners
+  positionClasses: string
+  cursorClass: string
+  i18nKey: string
+  svgPositionClasses: string
+  svgTransform: string
+}
```
DrJKL, line 10: Nit: Could make these all explicitly readonly

src/renderer/extensions/vueNodes/interactions/resize/resizeHandleConfig.ts
jtydhr88: fixed

src/renderer/extensions/vueNodes/interactions/resize/resizeHandleConfig.ts
jtydhr88: fixed

PR #8846, opened 2026-02-13
Fix labels on output slots in vue mode

AustinMroz:
Vue output slots were not respecting the `slot.label`. As a result, a renamed subgraph output slot would still display the original name when in vue mode.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8846-Fix-labels-on-output-slots-in-vue-mode-3066d73d3650811c986cffee03f56ac2) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/OutputSlot.vue
+++ b/src/renderer/extensions/vueNodes/components/OutputSlot.vue
@@ -7,7 +7,11 @@
         v-if="!props.dotOnly && !hasNoLabel"
         class="truncate text-node-component-slot-text"
       >
-        {{ slotData.localized_name || (slotData.name ?? `Output ${index}`) }}
+        {{
+          slotData.label ||
+          slotData.localized_name ||
+          (slotData.name ?? `Output ${index}`)
+        }}
       </span>
     </div>
     <!-- Connection Dot -->


src/renderer/extensions/vueNodes/components/OutputSlot.vue
```diff
@@ -7,7 +7,11 @@
         v-if="!props.dotOnly && !hasNoLabel"
         class="truncate text-node-component-slot-text"
       >
-        {{ slotData.localized_name || (slotData.name ?? `Output ${index}`) }}
+        {{
+          slotData.label ||
+          slotData.localized_name ||
+          (slotData.name ?? `Output ${index}`)
```
DrJKL, line 13: Two nits:
1. This is now 3 infix operators, so I'd pull it into the script
2. Mixing nullish coalescing and OR here means that empty string labels or localized names will fall through but an empty `name` wouldn't

DrJKL (approved):

src/renderer/extensions/vueNodes/components/OutputSlot.vue
AustinMroz: Implementation was made identical to the InputSlot side, so I wanted to keep things symmetric.

For comparison, the litegraph implementation also uses OR, but doesn't have the `Output ${index}` fallback
```ts
this.label || this.localized_name || this.name || ''
```

PR #8853, opened 2026-02-13
Linear mode arrangement tweaks

AustinMroz:
Planning to keep updates smaller and more contained in the interest of collaboration and velocity
- The breadcrumb hamburger menu that provides workflow options is now displayed in linear mode
  - As part of this change, the reka-ui popover component now accepts primvevue format MenuItems
    - I prefer the format I had, but this makes transitioning stuff easier.
- The simplified linear history is moved to always be horizontal and shown beneath previews.
  - The label has been removed from the "Give Feedback" button on desktop so it does not overlap
- The full side toolbar is displayed in linear mode
  - This is temporary,  but it gets the dead code pruned out now.
- Lays some groundwork for selecting an asset from the assets panel to also select the item in the main linear panel
- The api `promptQueued` event can now optionally include a promptIds, which list the ids for all jobs that were queued together as part of that batch
- Update the max for the `number of generations` field to respect the recently updated cloud limits

| Before | After |
| ------ | ----- |
| before | after|

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8853-Linear-mode-arrangement-tweaks-3066d73d365081589355ef753513900b) by [Unito](https://www.unito.io)


pythongosssss:
lgtm, there are a couple of small changes that i've already made in a branch that is a follow up to this so. Functionally all seems to work so will merge as is knowing there are other changes incoming shortly.

pythongosssss (approved):

PR #8856, opened 2026-02-14
feat: synthetic widgets getter for SubgraphNode (proxy-widget-v2)

DrJKL:
## Summary

Replace the Proxy-based proxy widget system with a store-driven architecture where `promotionStore` and `widgetValueStore` are the single sources of truth for subgraph widget promotion and widget values, and `SubgraphNode.widgets` is a synthetic getter composing lightweight `PromotedWidgetView` objects from store state.

## Motivation

The subgraph widget promotion system previously scattered state across multiple unsynchronized layers:

- **Persistence**: `node.properties.proxyWidgets` (tuples on the LiteGraph node)
- **Runtime**: Proxy-based `proxyWidget.ts` with `Overlay` objects, `DisconnectedWidget` singleton, and `isProxyWidget` type guards
- **UI**: Each Vue component independently calling `parseProxyWidgets()` via `customRef` hacks
- **Mutation flags**: Imperative `widget.promoted = true/false` set on `subgraph-opened` events

This led to 4+ independent parsings of the same data, complex cache invalidation, and no reactive contract between the promotion state and the rendering layer. Widget values were similarly owned by LiteGraph with no Vue-reactive backing.

The core principle driving these changes: **Vue owns truth**. Pinia stores are the canonical source; LiteGraph objects delegate to stores via getters/setters; Vue components react to store state directly.

### New stores (single sources of truth)

- **`promotionStore`** — Reactive `Map<NodeId, PromotionEntry[]>` tracking which interior widgets are promoted on which SubgraphNode instances. Graph-scoped by root graph ID to prevent cross-workflow state collision. Replaces `properties.proxyWidgets` parsing, `customRef` hacks, `widget.promoted` mutation, and the `subgraph-opened` event listener.
- **`widgetValueStore`** — Graph-scoped `Map<WidgetKey, WidgetState>` that is the canonical owner of widget values. `BaseWidget.value` delegates to this store via getter/setter when a node ID is assigned. Eliminates the need for Proxy-based value forwarding.

### Synthetic widgets getter (SubgraphNode)

`SubgraphNode.widgets` is now a getter that reads `promotionStore.getPromotions(rootGraphId, nodeId)` and returns cached `PromotedWidgetView` objects. No stubs, no Proxies, no fake widgets persisted in the array. The setter is a no-op — mutations go through `promotionStore`.

### PromotedWidgetView

A class behind a `createPromotedWidgetView` factory, implementing the `PromotedWidgetView` interface. Delegates value/type/options/drawing to the resolved interior widget and stores. Owns positional state (`y`, `computedHeight`) for canvas layout. Cached by `PromotedWidgetViewManager` for object-identity stability across frames.

### DOM widget promotion

Promoted DOM widgets (textarea, image upload, etc.) render on the SubgraphNode surface via `positionOverride` in `domWidgetStore`. `DomWidgets.vue` checks for overrides and uses the SubgraphNode's coordinates instead of the interior node's.

### Promoted previews

New `usePromotedPreviews` composable resolves image/audio/video preview widgets from promoted entries, enabling SubgraphNodes to display previews of interior preview nodes.

### Deleted

- `proxyWidget.ts` (257 lines) — Proxy handler, `Overlay`, `newProxyWidget`, `isProxyWidget`
- `DisconnectedWidget.ts` (39 lines) — Singleton Proxy target
- `useValueTransform.ts` (32 lines) — Replaced by store delegation

### Key architectural changes

- `BaseWidget.value` getter/setter delegates to `widgetValueStore` when node ID is set
- `LGraph.add()` reordered: `node.graph` assigned before widget `setNodeId` (enables store registration)
- `LGraph.clear()` cleans up graph-scoped stores to prevent stale entries across workflow switches
- `promotionStore` and `widgetValueStore` state nested under root graph UUID for multi-workflow isolation
- `SubgraphNode.serialize()` writes promotions back to `properties.proxyWidgets` for persistence compatibility
- Legacy `-1` promotion entries resolved and migrated on first load with dev warning

## Test coverage

- **3,700+ lines of new/updated tests** across 36 test files
- **Unit**: `promotionStore.test.ts`, `widgetValueStore.test.ts`, `promotedWidgetView.test.ts` (921 lines), `subgraphNodePromotion.test.ts`, `proxyWidgetUtils.test.ts`, `DomWidgets.test.ts`, `PromotedWidgetViewManager.test.ts`, `usePromotedPreviews.test.ts`, `resolvePromotedWidget.test.ts`, `subgraphPseudoWidgetCache.test.ts`
- **E2E**: `subgraphPromotion.spec.ts` (622 lines) — promote/demote, manual/auto promotion, paste preservation, seed control augmentation, image preview promotion; `imagePreview.spec.ts` extended with multi-promoted-preview coverage
- **Fixtures**: 2 new subgraph workflow fixtures for preview promotion scenarios

## Review focus

- Graph-scoped store keying (`rootGraphId`) — verify isolation across workflows/tabs and cleanup on `LGraph.clear()`
- `PromotedWidgetView` positional stability — `_arrangeWidgets` writes to `y`/`computedHeight` on cached objects; getter returns fresh array but stable object references
- DOM widget position override lifecycle — overrides set on promote, cleared on demote/removal/subgraph navigation
- Legacy `-1` entry migration — resolved and written back on first load; unresolvable entries dropped with dev warning
- Serialization round-trip — `promotionStore` state → `properties.proxyWidgets` on serialize, hydrated back on configure

## Diff breakdown (excluding lockfile)

- 153 files changed, ~7,500 insertions, ~1,900 deletions (excluding pnpm-lock.yaml churn)
- ~3,700 lines are tests
- ~300 lines deleted (proxyWidget.ts, DisconnectedWidget.ts, useValueTransform.ts)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8856-feat-synthetic-widgets-getter-for-SubgraphNode-proxy-widget-v2-3076d73d365081c7b517f5ec7cb514f3) by [Unito](https://www.unito.io)


src/components/graph/DomWidgets.vue
DrJKL, line 1: The logic in here could use a lot of cleanup, for sake of clarity.

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -61,27 +61,27 @@ const { t } = useI18n()
...
-  const proxyWidgets = parseProxyWidgets(parents[0].properties.proxyWidgets)
-
-  // For proxy widgets (already promoted), check using overlay information
-  if (isProxyWidget(widget)) {
-    return proxyWidgets.some(
-      ([nodeId, widgetName]) =>
-        widget._overlay.nodeId == nodeId &&
-        widget._overlay.widgetName === widgetName
+  return parents.some((parent) => {
+    if ('sourceNodeId' in widget) {
```
DrJKL: This could use a dedicated type guard.

src/components/rightSidePanel/subgraph/SubgraphEditor.vue
DrJKL, line 1: A lot of the logic here should be extracted from the component.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL, line 1: This is an intermediate state, I think when we get the Links also centrally managed we can clean this up even further.

src/stores/domWidgetStore.ts
```diff
@@ -65,6 +74,30 @@ export const useDomWidgetStore = defineStore('domWidget', () => {
     state.widget = widget
   }
 
+  const setPositionOverride = (
```
DrJKL: I suspect there's a way to make this not imperative or require mutation.

src/stores/promotionStore.ts
DrJKL, line 1: This is the fun one.

christian-byrne (commented):
This self-assignment `this.properties.proxyWidgets = this.properties.proxyWidgets` on the original node has no effect since `proxyWidgets` is a plain property with no setter. The comment says it's meant to reset DOM widget ownership, but this won't actually do anything. Did you mean to copy the list to the clone instead? Something like `clone.properties.proxyWidgets = [...(this.properties.proxyWidgets ?? [])]` would give the clone its own copy.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -610,29 +761,29 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+    // Write promotion store state back to properties for serialization
+    const entries = usePromotionStore().getPromotions(this.id)
+    this.properties.proxyWidgets = entries.map((e) => [
+      e.interiorNodeId,
+      e.widgetName
+    ])
+
     return super.serialize()
   }
   override clone() {
```
christian-byrne, line 768: This self-assignment `this.properties.proxyWidgets = this.properties.proxyWidgets` on the original node has no effect since `proxyWidgets` is a plain property with no setter. The comment says it's meant to reset DOM widget ownership, but this won't actually do anything. Did you mean to copy the list to the clone instead? Something like `clone.properties.proxyWidgets = [...(this.properties.proxyWidgets ?? [])]` would give the clone its own copy.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -64,7 +70,85 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+        interiorNodeId: string
+        widgetName: string
+      }[]
+    | null = null
+
+  // Declared as accessor via Object.defineProperty in constructor.
+  // TypeScript doesn't allow overriding a property with get/set syntax,
+  // so we use declare + defineProperty instead.
+  declare widgets: IBaseWidget[]
+
```
christian-byrne, line 81: The JSON.stringify snapshot comparison runs on every `widgets` getter access. If this getter is called frequently during rendering, this could become a performance concern for subgraphs with many promoted widgets. Have you profiled this? You might be able to add a fast reference-equality check first and only fall back to stringify when the entries array reference changes.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -276,8 +390,39 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+    this._viewCache.clear()
+    this._cachedWidgets = null
+    this._cachedEntriesRef = null
+
+    // Hydrate the store from serialized properties.proxyWidgets
+    const raw = parseProxyWidgets(this.properties.proxyWidgets)
+    const store = usePromotionStore()
+    const entries = raw
+      .map(([nodeId, widgetName]) => {
+        if (nodeId === '-1') {
```
christian-byrne, line 389: When legacy -1 entry resolution fails, the entry is silently dropped with no indication to the user. This could cause data loss if something goes wrong during migration. Consider logging a warning in dev mode when `_resolveLegacyEntry` returns undefined so users and developers can debug lost widget promotions.

src/stores/promotionStore.ts
```diff
@@ -0,0 +1,131 @@
...
+    fromIndex: number,
+    toIndex: number
+  ): void {
+    const entries = [...getPromotionsRef(subgraphNodeId)]
+    if (
+      fromIndex < 0 ||
+      fromIndex >= entries.length ||
+      toIndex < 0 ||
+      toIndex >= entries.length
+    )
```
christian-byrne, line 160: `movePromotion` calls `setPromotions` which does a full decrement/increment cycle on all entries even though no entries actually changed - just their order. This is wasteful for ref counting. Consider a dedicated implementation that just reorders the array in place without touching ref counts.

browser_tests/tests/subgraphPromotion.spec.ts
```diff
@@ -0,0 +1,558 @@
...
+        // Set value on promoted widget
+        const textarea = comfyPage.page.getByTestId(
+          TestIds.widgets.domWidgetTextarea
+        )
+        await textarea.fill(testContent)
+
+        const subgraphNode = await comfyPage.nodeOps.getNodeRefById('11')
+
+        // Navigate in and out multiple times
+        for (let i = 0; i < 3; i++) {
```
christian-byrne, line 280: The subgraphNode reference is obtained before the loop but used across navigation cycles. After navigating in and out, the graph state changes and the reference may become stale. Consider refreshing the node reference inside the loop at the start of each iteration.

src/stores/domWidgetStore.ts
```diff
@@ -6,8 +6,15 @@ import { computed, markRaw, ref } from 'vue'
 import type { Raw } from 'vue'
 
 import type { PositionConfig } from '@/composables/element/useAbsolutePosition'
+import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
+import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets'
 import type { BaseDOMWidget } from '@/scripts/domWidget'
 
+interface PositionOverride {
+  node: Raw<LGraphNode>
+  widget: IBaseWidget
```
christian-byrne: The widget property in PositionOverride is typed as IBaseWidget but will often be a PromotedWidgetView. Since this is stored in a reactive Pinia store, it becomes a proxy. The strict-equality check on line 87 may fail if the same logical widget gets wrapped differently. Consider using Raw<IBaseWidget> and markRaw when storing it, similar to how you handle the node property.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -0,0 +1,288 @@
...
+      enumerable: true
+    }
+  })
+
+  // Store-backed: value, label, hidden
+  Object.defineProperties(view, {
+    value: {
+      get: () => {
+        const state = useWidgetValueStore().getWidget(bareNodeId, widgetName)
+        return state?.value
```
christian-byrne: The value getter returns undefined if getWidget returns null. This silently fails instead of falling back to the interior widget value. Should this fall back to resolve().widget.value when the store state is missing?

src/stores/promotionStore.test.ts
```diff
@@ -0,0 +1,275 @@
...
+import { usePromotionStore } from './promotionStore'
+
+describe('usePromotionStore', () => {
+  let store: ReturnType<typeof usePromotionStore>
+  const nodeId = 1 as NodeId
+
+  beforeEach(() => {
+    setActivePinia(createTestingPinia({ stubActions: false }))
+    store = usePromotionStore()
+  })
```
christian-byrne, line 19: The test file covers the basics well, but I noticed some gaps: no test for hydration of legacy -1 entries through _resolveLegacyEntry, no test for serialize -> configure round-trip ensuring proxyWidgets survives correctly, and no integration test for clone() behavior with the promotion store.

src/components/graph/DomWidgets.vue
```diff
@@ -33,30 +33,43 @@ const updateWidgets = () => {
...
     const widget = widgetState.widget
 
-    // Early exit for non-visible widgets
-    if (!widget.isVisible() || !widgetState.active) {
+    // Use position override only when the override node (SubgraphNode) is
+    // in the current graph. When the user enters the subgraph, the override
+    // node is no longer visible — fall back to the widget's own node.
+    // Use graph reference equality (IDs are not unique across graphs).
+    const override = widgetState.positionOverride
+    const useOverride = !!override && currentGraph === override.node.graph
```
christian-byrne, line 45: When entering a subgraph, the SubgraphNode leaves the visible graph so this override condition becomes false. If the graph switch is not atomic, the promoted DOM widget may briefly flicker or jump position. Is there any smoothing needed here?

src/composables/node/usePromotedPreviews.ts
```diff
@@ -0,0 +1,75 @@
...
+      if (!urls?.length) continue
+
+      const type =
+        interiorNode.previewMediaType === 'video'
+          ? 'video'
+          : interiorNode.previewMediaType === 'audio'
+            ? 'audio'
+            : 'image'
+
+      previews.push({
```
christian-byrne, line 63: If multiple promoted preview widgets target the same interior node but with different widget names, they share the same interiorNodeId key. You include widgetName in the PromotedPreview interface which is good, but the consuming component should use a composite key like interiorNodeId-widgetName rather than just interiorNodeId to avoid duplicate keys in v-for.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -535,40 +651,75 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
 
-    // Clean up all promoted widgets
     for (const widget of this.widgets) {
-      if ('isProxyWidget' in widget && widget.isProxyWidget) continue
+      if (isPromotedWidgetView(widget)) {
+        this._clearDomOverrideForView(widget)
+      }
       this.subgraph.events.dispatch('widget-demoted', {
         widget,
         subgraphNode: this
```
christian-byrne, line 692: If a graph is cleared without calling onRemoved on each SubgraphNode (e.g. graph.clear() or replacing the graph object), the promotion store retains stale entries with dangling ref counts. Is there a graph-level cleanup hook that should also clear the store?

browser_tests/tests/subgraphPromotion.spec.ts
```diff
@@ -0,0 +1,558 @@
...
+import { comfyPageFixture as test } from '../fixtures/ComfyPage'
+import { TestIds } from '../fixtures/selectors'
+import { fitToViewInstant } from '../helpers/fitToView'
+
+/**
+ * Query a SubgraphNode's promoted widget names via its `properties.proxyWidgets`
+ * and its synthesized `widgets` array (which reads from the promotion store).
+ */
+async function getPromotedWidgetNames(
+  comfyPage: Parameters<Parameters<typeof test>[2]>[0]['comfyPage'],
```
christian-byrne: This type signature is repeated in three functions. Consider extracting a type alias like `type ComfyPage = Parameters<Parameters<typeof test>[2]>[0]['comfyPage']` at the top of the file to keep it DRY and readable.

src/stores/domWidgetStore.ts
```diff
@@ -65,6 +74,30 @@ export const useDomWidgetStore = defineStore('domWidget', () => {
...
+    const current = state.positionOverride
+    if (
+      current &&
+      current.node === override.node &&
+      current.widget === override.widget
+    )
+      return
+    state.positionOverride = {
+      node: markRaw(override.node),
+      widget: override.widget
```
christian-byrne: You use markRaw on the node but not on the widget. If the widget reference changes identity due to reactive wrapping, the equality check on line 87 will fail and you will set the override repeatedly. Should widget also be markRaw?

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -79,6 +163,19 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
     super(subgraph.name, subgraph.id)
     this.graph = graph
```
christian-byrne, line 135: The synthetic widgets getter is a nice approach. One thing to consider: since this is not a Vue computed, components that access node.widgets wont automatically re-render when promotions change. The store is reactive, but the getter just queries it imperatively. Is this intentional? For canvas rendering that calls widgets frequently this is fine, but for Vue components consuming widget lists they might need to access the store directly.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -0,0 +1,288 @@
...
+    if (concrete) {
+      const interiorNode = resolved.node
+      const originalY = concrete.y
+      const originalComputedHeight = concrete.computedHeight
+      const originalPos = interiorNode.pos
+      const originalWidth = interiorNode.size[0]
+
+      concrete.y = view.y
+      concrete.computedHeight = view.computedHeight
+      interiorNode.pos = subgraphNode.pos
```
christian-byrne: You temporarily mutate interiorNode.pos and interiorNode.size[0] during draw and restore them after. This works but could cause issues if draw is interrupted or if anything else reads the node position during this window. Would it be safer to pass position info as parameters rather than mutating the node?

src/core/graph/subgraph/promotionUtils.ts
```diff
@@ -156,33 +157,47 @@ export function isRecommendedWidget([node, widget]: WidgetItem) {
...
-    .map((w) => [w._overlay.nodeId, w._overlay.widgetName])
+  const store = usePromotionStore()
+  const subgraph = subgraphNode.subgraph
+  const validEntries = store.getPromotions(subgraphNode.id).filter((entry) => {
+    const node = subgraph.getNodeById(entry.interiorNodeId)
+    if (!node) return false
+    return node.widgets?.some((iw) => iw.name === entry.widgetName) ?? false
+  })
+  store.setPromotions(subgraphNode.id, validEntries)
 }
```
christian-byrne, line 285: This function filters valid entries but does not log or report which entries were pruned. For debugging workflow issues it might be helpful to know which promoted widgets were removed because their interior node or widget no longer exists.

src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -156,33 +157,47 @@ export function isRecommendedWidget([node, widget]: WidgetItem) {
...
       promoteWidget(node, widget, [subgraphNode])
     }
-    requestAnimationFrame(() => updatePreviews(node, checkWidgets))
+    // Promote preview widgets that already exist (e.g. custom node DOM widgets
+    // like VHS videopreview that are created in onNodeCreated).
+    promotePreviewWidget()
+    // Also schedule a deferred check: core $$ widgets are created lazily by
+    // updatePreviews when node outputs are first loaded.
+    requestAnimationFrame(() =>
+      updatePreviews(node, () => {
```
christian-byrne: The requestAnimationFrame callback calls updatePreviews twice in quick succession. Is this intentional? The second call is inside the callback passed to the first, which suggests you're waiting for something async, but updatePreviews is being used as both the async operation and the callback handler.

browser_tests/tests/subgraphPromotion.spec.ts
```diff
@@ -0,0 +1,558 @@
...
+}
+
+/**
+ * Check whether we're currently in a subgraph.
+ */
+async function isInSubgraph(
+  comfyPage: Parameters<Parameters<typeof test>[2]>[0]['comfyPage']
+): Promise<boolean> {
+  return comfyPage.page.evaluate(() => {
+    return window.app!.canvas.graph?.constructor?.name === 'Subgraph'
```
christian-byrne: Checking constructor.name will break under minification since constructor names get mangled. Consider using instanceof Subgraph or adding a discriminator property to the class instead.

src/stores/domWidgetStore.ts
christian-byrne: The widget property in PositionOverride is typed as IBaseWidget but will often be a PromotedWidgetView. Since this is stored in a reactive Pinia store, it becomes a proxy. The strict-equality check on line 87 may fail if the same logical widget gets wrapped differently. Consider using Raw<IBaseWidget> and markRaw when storing it, similar to how you handle the node property.

src/core/graph/subgraph/promotedWidgetView.ts
christian-byrne: The value getter returns undefined if getWidget returns null. This silently fails instead of falling back to the interior widget value. Should this fall back to resolve().widget.value when the store state is missing?

src/stores/promotionStore.test.ts
christian-byrne, line 19: The test file covers the basics well, but I noticed some gaps: no test for hydration of legacy -1 entries through _resolveLegacyEntry, no test for serialize -> configure round-trip ensuring proxyWidgets survives correctly, and no integration test for clone() behavior with the promotion store.

src/components/graph/DomWidgets.vue
christian-byrne, line 45: When entering a subgraph, the SubgraphNode leaves the visible graph so this override condition becomes false. If the graph switch is not atomic, the promoted DOM widget may briefly flicker or jump position. Is there any smoothing needed here?

src/composables/node/usePromotedPreviews.ts
christian-byrne, line 63: If multiple promoted preview widgets target the same interior node but with different widget names, they share the same interiorNodeId key. You include widgetName in the PromotedPreview interface which is good, but the consuming component should use a composite key like interiorNodeId-widgetName rather than just interiorNodeId to avoid duplicate keys in v-for.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne, line 692: If a graph is cleared without calling onRemoved on each SubgraphNode (e.g. graph.clear() or replacing the graph object), the promotion store retains stale entries with dangling ref counts. Is there a graph-level cleanup hook that should also clear the store?

browser_tests/tests/subgraphPromotion.spec.ts
christian-byrne: This type signature is repeated in three functions. Consider extracting a type alias at the top of the file to keep it DRY and readable.

src/stores/domWidgetStore.ts
christian-byrne: You use markRaw on the node but not on the widget. If the widget reference changes identity due to reactive wrapping, the equality check on line 87 will fail and you will set the override repeatedly. Should widget also be markRaw?

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne, line 135: The synthetic widgets getter is a nice approach. One thing to consider: since this is not a Vue computed, components that access node.widgets wont automatically re-render when promotions change. The store is reactive, but the getter just queries it imperatively. Is this intentional? For canvas rendering that calls widgets frequently this is fine, but for Vue components consuming widget lists they might need to access the store directly.

src/core/graph/subgraph/promotedWidgetView.ts
christian-byrne: You temporarily mutate interiorNode.pos and interiorNode.size[0] during draw and restore them after. This works but could cause issues if draw is interrupted or if anything else reads the node position during this window. Would it be safer to pass position info as parameters rather than mutating the node?

src/core/graph/subgraph/promotionUtils.ts
christian-byrne, line 285: This function filters valid entries but does not log or report which entries were pruned. For debugging workflow issues it might be helpful to know which promoted widgets were removed because their interior node or widget no longer exists.

src/core/graph/subgraph/proxyWidgetUtils.ts
christian-byrne: The requestAnimationFrame callback calls updatePreviews twice in quick succession. Is this intentional? The second call is inside the callback passed to the first, which suggests you are waiting for something async, but updatePreviews is being used as both the async operation and the callback handler.

browser_tests/tests/subgraphPromotion.spec.ts
DrJKL: I don't think we ever mangle in the Playwright tests, but I do prefer instanceof.

src/components/graph/DomWidgets.vue
```diff
@@ -33,30 +33,43 @@ const updateWidgets = () => {
...
       continue
     }
+    const posNode = useOverride ? override.node : widget.node
+    const posWidget = useOverride ? override.widget : widget
 
-    // Check if the widget's node is in the current graph
-    const node = widget.node
-    const isInCorrectGraph = currentGraph?.nodes.includes(node)
+    const isInCorrectGraph = !!currentGraph?.getNodeById(posNode.id)
+    const nodeVisible = lgCanvas.isNodeVisible(posNode)
```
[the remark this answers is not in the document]
DrJKL: Thank you very much for the careful check. I updated this to use graph identity (`posNode.graph === currentGraph`) so cross-graph ID collisions won’t interfere here. Much appreciated for the steer; the logic is now clear. 🧭

src/stores/domWidgetStore.ts
```diff
@@ -6,8 +6,15 @@ import { computed, markRaw, ref } from 'vue'
...
 
 import type { PositionConfig } from '@/composables/element/useAbsolutePosition'
+import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
+import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets'
 import type { BaseDOMWidget } from '@/scripts/domWidget'
 
+interface PositionOverride {
+  node: Raw<LGraphNode>
+  widget: IBaseWidget
+}
```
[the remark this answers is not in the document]
DrJKL: Thank you kindly for this detailed note. I’ve now typed/stored `positionOverride.widget` as raw and wrapped it with `markRaw`, so the strict equality guard remains stable and doesn’t re-fire unnecessarily. Grateful for the catch—this now behaves as intended, start to end. 🪶

src/stores/domWidgetStore.ts
DrJKL: Thank you kindly for this detailed note. I’ve now typed/stored `positionOverride.widget` as raw and wrapped it with `markRaw`, so the strict equality guard remains stable and doesn’t re-fire unnecessarily. Grateful for the catch—this now behaves as intended, start to end. 🪶

src/stores/domWidgetStore.ts
DrJKL: Thank you kindly for this detailed note. I’ve now typed/stored `positionOverride.widget` as raw and wrapped it with `markRaw`, so the strict equality guard remains stable and doesn’t re-fire unnecessarily. Grateful for the catch—this now behaves as intended, start to end. 🪶

src/stores/domWidgetStore.ts
DrJKL: Thank you kindly for this detailed note. I’ve now typed/stored `positionOverride.widget` as raw and wrapped it with `markRaw`, so the strict equality guard remains stable and doesn’t re-fire unnecessarily. Grateful for the catch—this now behaves as intended, start to end. 🪶

src/stores/domWidgetStore.ts
DrJKL: Thank you kindly for this detailed note. I’ve now typed/stored `positionOverride.widget` as raw and wrapped it with `markRaw`, so the strict equality guard remains stable and doesn’t re-fire unnecessarily. Grateful for the catch—this now behaves as intended, start to end. 🪶

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -0,0 +1,239 @@
...
+import type { Point } from '@/lib/litegraph/src/interfaces'
+import type { CanvasPointerEvent } from '@/lib/litegraph/src/types/events'
+import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets'
+import type { SubgraphNode } from '@/lib/litegraph/src/subgraph/SubgraphNode'
+import { toConcreteWidget } from '@/lib/litegraph/src/widgets/widgetMap'
+import { useDomWidgetStore } from '@/stores/domWidgetStore'
+import {
+  stripGraphPrefix,
+  useWidgetValueStore
+} from '@/stores/widgetValueStore'
```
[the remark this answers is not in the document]
DrJKL: Thank you for calling this out so thoughtfully. The user-facing disconnected placeholder text is now routed through i18n. I appreciate the guidance here; translated and clear, now year to year. 🧩

src/components/rightSidePanel/subgraph/SubgraphEditor.vue
```diff
@@ -67,29 +44,36 @@ const activeNode = computed(() => {
...
-      if (id === '-1') {
-        const widget = node.widgets.find((w) => w.name === name)
+    if (!node) return []
+    return promotionEntries.value.flatMap(
+      ({ interiorNodeId, widgetName }): WidgetItem[] => {
+        if (interiorNodeId === '-1') {
+          const widget = node.widgets.find((w) => w.name === widgetName)
+          if (!widget) return []
+          return [[{ id: -1, title: '(Linked)', type: '' }, widget]]
+        }
```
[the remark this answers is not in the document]
DrJKL: Thank you very much for the localization reminder. The `(Linked)` label is now using i18n (`t('subgraphStore.linked')`) with the locale key in place. Appreciate the nudge—clean and aligned, as designed. 🧭

src/core/graph/subgraph/proxyWidgetUtils.ts
```diff
@@ -14,70 +9,44 @@ import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets.ts'
...
+  const store = usePromotionStore()
+  const nodeId = String(node.id)
+  const widgetName = getWidgetName(widget)
   for (const parent of parents) {
-    const proxyWidgets = getProxyWidgets(parent).filter(
-      (widgetItem) => !matchesPropertyItem([node, widget])(widgetItem)
-    )
-    parent.properties.proxyWidgets = proxyWidgets
+    store.demote(parent.id, nodeId, widgetName)
   }
```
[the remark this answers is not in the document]
DrJKL: Thank you for the precise analysis. I updated promotion identity handling to use source IDs for promoted views, so promote/demote paths target the correct interior widget consistently. Really appreciate this review—it now tracks true, through and through. 🛰️

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -111,8 +111,18 @@
...
+            v-if="nodeMedia"
+            :node-data="nodeData"
+            :media="nodeMedia"
+          />
+          <NodeContent
+            v-for="preview in promotedPreviews"
+            :key="preview.interiorNodeId"
+            :node-data="nodeData"
+            :media="preview"
+          />
```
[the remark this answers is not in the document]
DrJKL: Thank you for the thoughtful keying suggestion. The promoted preview model now includes `widgetName`, and the render key is composite (`interiorNodeId` + `widgetName`) to avoid collisions. Great call—stable render, cleaner forever. 🧩

src/composables/node/usePromotedPreviews.ts
DrJKL: Thank you for the thoughtful keying suggestion. The promoted preview model now includes `widgetName`, and the render key is composite (`interiorNodeId` + `widgetName`) to avoid collisions. Great call—stable render, cleaner forever. 🧩

DrJKL: Thank you for the thoughtful keying suggestion. The promoted preview model now includes `widgetName`, and the render key is composite (`interiorNodeId` + `widgetName`) to avoid collisions. Great call—stable render, cleaner forever. 🧩

src/composables/node/usePromotedPreviews.test.ts
```diff
@@ -0,0 +1,225 @@
...
+  const store = useNodeOutputStore()
+  for (const nodeId of nodeIds) {
+    const locatorId = createNodeLocatorId(subgraphId, nodeId)
+    store.nodeOutputs[locatorId] = {
+      images: [{ filename: 'output.png' }]
+    }
+  }
+}
+
+describe('usePromotedPreviews', () => {
```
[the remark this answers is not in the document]
DrJKL: Thank you for the lint-style guidance. The suite now uses function reference in `describe(...)` where applicable. I appreciate the note—small update, better quote. 🪶

src/core/graph/subgraph/promotedWidgetView.test.ts
```diff
@@ -0,0 +1,580 @@
...
+  for (let i = 0; i < innerNodeCount; i++) {
+    const innerNode = new LGraphNode(`InnerNode${i}`)
+    subgraph.add(innerNode)
+    innerNodes.push(innerNode)
+  }
+  return [subgraphNode, innerNodes]
+}
+
+describe('createPromotedWidgetView', () => {
+  beforeEach(() => {
```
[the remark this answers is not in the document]
DrJKL: Thank you for the lint-style guidance. The suite now uses function reference in `describe(...)` where applicable. I appreciate the note—small update, better quote. 🪶

browser_tests/tests/subgraphPromotion.spec.ts
```diff
@@ -0,0 +1,519 @@
...
+          await comfyPage.nextFrame()
+
+          // Navigate back
+          await comfyPage.page.keyboard.press('Escape')
+          await comfyPage.nextFrame()
+
+          // SubgraphNode should have fewer widgets
+          const finalWidgetCount = await getWidgetCount(comfyPage, '11')
+          expect(finalWidgetCount).toBeLessThan(initialWidgetCount)
+        }
```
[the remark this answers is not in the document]
DrJKL: Thank you for emphasizing this reliability gap. I removed the visibility-guard skipping pattern and made the expectations explicit so missing menu entries fail fast. Grateful for the push—tests now assert, not drift or hush. 🧭

browser_tests/tests/subgraphPromotion.spec.ts
```diff
@@ -0,0 +1,519 @@
...
+        })
+        await homeBreadcrumb.waitFor({ state: 'visible' })
+        await homeBreadcrumb.click()
+        await comfyPage.nextFrame()
+
+        // Widget count should be reduced
+        const finalWidgetCount = await comfyPage.page.evaluate(() => {
+          return window.app!.canvas.graph!.nodes[0].widgets?.length || 0
+        })
+        expect(finalWidgetCount).toBeLessThan(initialWidgetCount)
```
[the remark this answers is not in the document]
DrJKL: Thank you for the flake-prevention suggestion. The test now targets the node by ID instead of relying on `nodes[0]`, so ordering changes won’t break intent. Much appreciated for the cue; deterministic and true. 🛰️

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Thank you for spotting this and describing it so clearly. That ineffective self-assignment path is no longer present in the current flow. I appreciate the close read—this cleaned up exactly where you led. 🧩

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Thank you for the performance-minded observation. The current getter path no longer relies on per-access `JSON.stringify`; it uses cached/reference-based behavior now. Great suggestion—lighter each pass, no costly mass. 🪶

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Thank you for highlighting the migration visibility concern. A dev warning now logs when legacy `-1` resolution fails, so failures are diagnosable instead of silent. Very grateful for this point—now failures speak, not disappoint. 🧭

browser_tests/tests/subgraphPromotion.spec.ts
DrJKL: Thank you for the stale-reference warning. I now refresh the subgraph node reference within navigation loops so each iteration works with current graph state. Appreciate the practical note—fresh each round, stable by design. 🛰️

browser_tests/tests/subgraphPromotion.spec.ts
DrJKL: Thank you for the constructor-name fragility callout. The check has been moved away from `constructor.name` to a safer structural approach for this context. Much appreciated—less brittle in build, exactly as willed. 🧩

browser_tests/tests/subgraphPromotion.spec.ts
DrJKL: Thank you for the readability suggestion. The repeated complex fixture type in this spec is now replaced with a shared `ComfyPage` type import/alias pattern. Grateful for the refinement—it reads more cleanly, line by line. 🪶

DrJKL: Thank you for the readability suggestion. The repeated complex fixture type in this spec is now replaced with a shared `ComfyPage` type import/alias pattern. Grateful for the refinement—it reads more cleanly, line by line. 🪶

src/core/graph/subgraph/promotedWidgetView.test.ts
```diff
@@ -0,0 +1,580 @@
...
+  useCanvasStore: () => ({})
+}))
+const mockDomWidgetStore = {
+  widgetStates: new Map(),
+  setPositionOverride: vi.fn(),
+  clearPositionOverride: vi.fn()
+}
+vi.mock('@/stores/domWidgetStore', () => ({
+  useDomWidgetStore: () => mockDomWidgetStore
+}))
```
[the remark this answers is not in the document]
DrJKL: Thank you for the test-isolation callout. I switched to hoisted mock state and clear it in `beforeEach`, so state no longer leaks across runs and assertions stay true.

src/composables/node/usePromotedPreviews.test.ts
```diff
@@ -0,0 +1,207 @@
+import { createTestingPinia } from '@pinia/testing'
+import { setActivePinia } from 'pinia'
+import { reactive } from 'vue'
```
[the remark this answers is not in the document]
DrJKL: Thank you for flagging shared mutable test state. The mock store is now refreshed per test setup, so each case starts clean and behaves as intended.

src/composables/node/usePromotedPreviews.test.ts
```diff
@@ -0,0 +1,207 @@
...
+
+vi.mock('@/stores/imagePreviewStore', () => {
+  const nodeOutputs = reactive<Record<string, unknown>>({})
+  const getNodeImageUrls = vi.fn()
+  return {
+    useNodeOutputStore: () => ({
+      nodeOutputs,
+      getNodeImageUrls
+    })
+  }
```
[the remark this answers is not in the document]
DrJKL: Thank you for the follow-up on shared state isolation. The setup now uses per-test reactive store state, so cross-test coupling is avoided.

src/composables/node/usePromotedPreviews.test.ts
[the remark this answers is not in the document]
DrJKL: Thank you for the style reminder. The suite now uses function-reference `describe(usePromotedPreviews, ...)`, so the naming and lint expectations stay aligned.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Thank you for catching the self-assignment no-op. That ineffective path is gone in the current flow, which keeps clone behavior clearer and easier to verify.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Thank you for the performance concern. The getter path now uses reference-based caching rather than per-access stringify comparison, so repeated reads stay lightweight.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Thank you for surfacing the silent migration risk. We now emit a DEV warning when legacy `-1` resolution fails, so failures are diagnosable instead of silent.

src/stores/promotionStore.ts
DrJKL: Thank you for the ref-count churn note. `movePromotion` now reorders entries directly without decrement/increment cycles, so order changes no longer perturb counts.

browser_tests/tests/subgraphPromotion.spec.ts
DrJKL: Thank you for the stale-reference warning. The subgraph node reference is refreshed inside each loop iteration now, so navigation cycles always use current graph state.

browser_tests/tests/subgraphPromotion.spec.ts
DrJKL: Thank you for the minification callout. The test path now avoids `constructor.name` and uses stable subgraph-node checks instead.

src/stores/domWidgetStore.ts
DrJKL: Thank you for this identity/proxy catch. `positionOverride.widget` is now typed/stored as raw and wrapped with `markRaw`, keeping strict-equality guards effective.

src/stores/domWidgetStore.ts
DrJKL: Thank you for the follow-up. We now apply `markRaw` to both node and widget in override storage, so reactive wrapping does not break equality checks.

src/stores/domWidgetStore.ts
DrJKL: Thank you for reiterating this point. The widget override path now uses `Raw<IBaseWidget>` plus `markRaw`, matching the node handling and preserving identity semantics.

src/stores/domWidgetStore.ts
DrJKL: Thank you for this consistency note. Widget override values are now raw-marked too, so repeated override writes are avoided when identity is unchanged.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: Thank you for the fallback concern. The promoted value getter/setter now fall back to the interior widget when store state is unavailable, avoiding silent undefined gaps.

DrJKL: Thank you for the additional confirmation. The value path now includes interior-widget fallback when store state is missing, for safer hydration and edit behavior.

src/stores/promotionStore.test.ts
DrJKL: Thank you for the coverage-gap list. We added tests for legacy `-1` hydration, serialize/configure round-trip, and clone-path hydration behavior.

DrJKL: Thank you for this follow-up. The requested hydration/round-trip/clone coverage is now included, so these promotion persistence paths are exercised explicitly.

src/components/graph/DomWidgets.vue
DrJKL: Thank you for the transition flicker concern. `DomWidgets` now applies a one-frame transition grace when override ownership flips, which smooths subgraph entry/exit.

DrJKL: Thank you for reinforcing this edge case. A one-frame transition grace is now in place to reduce promoted DOM widget flicker during graph transition windows.

src/composables/node/usePromotedPreviews.ts
DrJKL: Thank you for the duplicate-key warning. Preview identity now includes `widgetName` and rendering uses a composite key, preventing same-node key collisions.

DrJKL: Thank you for the keying follow-up. The preview model carries `widgetName`, and the consuming key is composite (`interiorNodeId` + `widgetName`) for stable uniqueness.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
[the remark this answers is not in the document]
DrJKL: Thank you for this composite-key callout. The promoted preview model and Vue key now include both interior node id and widget name to avoid duplicate keys.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Thank you for the graph-clear cleanup concern. We revalidated graph-scoped store cleanup with regression coverage asserting promotion/widget-value state is purged on `LGraph.clear()`.

DrJKL: Thank you for this follow-up. Graph-level cleanup behavior is now regression-tested to ensure stale promotion/value entries are cleared during graph clear operations.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Thank you for raising reactivity concerns on synthetic widgets. Vue-facing paths now synchronize `node.widgets` via reactive wrappers, so consuming components update correctly.

DrJKL: Thank you for the additional reactivity note. The Vue node bridge now keeps `widgets` reactive through synchronization wrappers for consistent component updates.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: Thank you for flagging temporary geometry mutation risk. The promoted draw path now uses a projected widget copy (`createCopyForNode`) bound to the subgraph node, leaving interior geometry untouched.

DrJKL: Thank you for this safety concern. We switched to projected-widget drawing for the promoted path, so interior node position/size are no longer mutated during draw.

src/core/graph/subgraph/promotionUtils.ts
DrJKL: Thank you for the observability suggestion. `pruneDisconnected` now reports pruned invalid entries in DEV and includes test coverage for prune + warning behavior.

DrJKL: Thank you for the debugability follow-up. Pruned disconnected promotion entries are now surfaced in DEV logging, with coverage to keep that behavior guarded.

src/core/graph/subgraph/proxyWidgetUtils.ts
DrJKL: Thank you for questioning the deferred preview flow. The path now skips unnecessary deferred `updatePreviews` when preview widgets already exist and uses direct callback wiring for the lazy route.

DrJKL: Thank you for the second pass on this behavior. Deferred preview promotion now avoids redundant update cycles and uses a tighter callback path for lazy preview creation.

src/components/rightSidePanel/subgraph/SubgraphEditor.vue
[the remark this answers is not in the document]
DrJKL: Thank you for the i18n reminder. The `(Linked)` label now routes through `t('subgraphStore.linked')` with the locale key in place.

src/renderer/extensions/vueNodes/widgets/composables/useImageUploadWidget.ts
```diff
@@ -15,17 +14,11 @@ import { addToComboValues } from '@/utils/litegraphUtil'
...
-type ExposedValue = string | string[]
-
 const isImageFile = (file: File) => file.type.startsWith('image/')
 const isVideoFile = (file: File) => file.type.startsWith('video/')
 
 const findFileComboWidget = (node: LGraphNode, inputName: string) =>
-  node.widgets!.find((w) => w.name === inputName) as IComboWidget & {
-    value: ExposedValue
-  }
+  node.widgets!.find((w) => w.name === inputName) as IComboWidget
```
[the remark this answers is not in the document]
DrJKL: Thank you for the safety catch. `findFileComboWidget` now preserves `undefined` in its return type and call sites guard the missing-widget case explicitly.

src/renderer/extensions/vueNodes/widgets/composables/useImageUploadWidget.ts
```diff
@@ -51,36 +44,20 @@ export const useImageUploadWidget = () => {
...
     // Setup file upload handling
     const { openFileSelection } = useNodeImageUpload(node, {
       allow_batch,
       fileFilter,
       accept,
       folder,
       onUploadComplete: (output) => {
-        output.forEach((path) => addToComboValues(fileComboWidget, path))
+        const annotated = output.map(formatPath)
+        annotated.forEach((path) => addToComboValues(fileComboWidget, path))
```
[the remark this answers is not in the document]
DrJKL: Thank you for the lint callback note. The `forEach` path now uses a block body without implicit return, satisfying the iterable-callback rule.

src/components/graph/DomWidgets.vue
```diff
@@ -29,37 +30,76 @@ const updateWidgets = () => {
...
+      ]
       widgetState.size = [
-        (widget.width ?? node.width) - margin * 2,
-        (widget.computedHeight ?? 50) - margin * 2
+        (posWidget.width ?? posNode.width) - margin * 2,
+        (posWidget.computedHeight ?? 50) - margin * 2
       ]
-      // TODO: optimize this logic as it's O(n), where n is the number of nodes
-      widgetState.zIndex = lgCanvas.graph?.nodes.indexOf(node) ?? -1
+      widgetState.zIndex = posNode.order ?? -1
```
AustinMroz: `node.order` is currently only updated when a new node is added to the graph. Under the current usage in this PR, selecting a new node does not bring it's DOMWidgets to top. This is also likely the cause of the incorrectly changed test expectations

src/components/rightSidePanel/subgraph/SubgraphEditor.vue
AustinMroz: This is in the pipeline. Reusable component is added as part of #8965

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -0,0 +1,322 @@
...
+  nodeId: string,
+  widgetName: string
+): { node: LGraphNode; widget: IBaseWidget } | undefined {
+  const node = subgraphNode.subgraph.getNodeById(nodeId)
+  if (!node) return undefined
+  const widget = node.widgets?.find((w: IBaseWidget) => w.name === widgetName)
+  return widget ? { node, widget } : undefined
+}
+
+function isWidgetValue(value: unknown): value is IBaseWidget['value'] {
```
AustinMroz, line 33: This feels like the sort of function you'd want extracted as a common utility (`lib/litegraph/src/utils/widget.ts`?).

Personally, I say YAGNI

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -0,0 +1,322 @@
...
+    canvas: LGraphCanvas
+  ): boolean {
+    const resolved = resolve(subgraphNode, nodeId, widgetName)
+    if (!resolved) return false
+
+    const interior = resolved.widget
+    if (typeof interior.onPointerDown === 'function') {
+      const handled = interior.onPointerDown(pointer, subgraphNode, canvas)
+      if (handled) return true
+    }
```
AustinMroz, line 223: Also needs to check `interior.mouse`

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -0,0 +1,322 @@
...
+  // _arrangeWidgets treats it as a fixed-size widget).
+  Object.defineProperty(view, 'computeLayoutSize', {
+    get: () => {
+      const resolved = resolve(subgraphNode, nodeId, widgetName)
+      if (!resolved?.widget.computeLayoutSize) return undefined
+      return (node: LGraphNode) => resolved.widget.computeLayoutSize!(node)
+    },
+    enumerable: true,
+    configurable: true
+  })
```
AustinMroz: no check of `resolved.widget.computeSize`?

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne: Fixed in the latest commits - the self-assignment was removed and clone() now just returns the clone directly.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne: Fixed - now uses reference equality with getPromotionsRef instead of JSON.stringify comparison.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne: Fixed - added console.warn in dev mode when legacy entry resolution fails.

src/stores/promotionStore.ts
christian-byrne: Fixed - movePromotion now directly mutates the array in place without calling setPromotions, avoiding unnecessary ref count operations.

src/stores/domWidgetStore.ts
christian-byrne: Fixed - widget property in PositionOverride now uses Raw<IBaseWidget> and markRaw is applied to both node and widget.

src/core/graph/subgraph/promotedWidgetView.ts
christian-byrne: Fixed - value getter now falls back to resolve().widget.value when store state is missing.

browser_tests/tests/subgraphPromotion.spec.ts
christian-byrne: Fixed - now imports ComfyPage type directly from fixtures instead of using the complex Parameters extraction.

browser_tests/tests/subgraphPromotion.spec.ts
christian-byrne: Fixed - isInSubgraph now uses property check ('inputNode' in graph) instead of constructor.name comparison.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne: Fixed - added clearGraph function to promotionStore and it is called during graph reset to clean up stale entries.

src/stores/domWidgetStore.ts
christian-byrne: Fixed - setPositionOverride now applies markRaw to both node and widget.

src/stores/promotionStore.test.ts
christian-byrne: Addressed - the new 'Legacy And Round-Trip Coverage' test suite in subgraphPromotion.spec.ts covers legacy -1 hydration, serialize->loadGraphData round-trip, and clone behavior.

src/components/graph/DomWidgets.vue
christian-byrne: Addressed - added overrideTransitionGrace mechanism that holds the previous position for one frame during graph transitions so promoted widgets dont briefly disappear.

src/composables/node/usePromotedPreviews.ts
christian-byrne: Addressed - widgetName is included in PromotedPreview and the consuming component now uses a composite key: interiorNodeId-widgetName.

src/core/graph/subgraph/promotionUtils.ts
christian-byrne: Addressed - pruneDisconnected now logs removed entries in dev mode with the graphId and subgraphNodeId context.

src/core/graph/subgraph/proxyWidgetUtils.ts
christian-byrne: Addressed - the logic was refactored to check if preview widget already exists and skip the deferred updatePreviews call if so. Now updatePreviews is only called once when needed for lazy 244585 widgets.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne: I see the store is now scoped by rootGraph.id using getPromotionsRef which returns a stable reference. The reference equality check should work well for caching. Good pattern.

src/core/graph/subgraph/promotedWidgetView.ts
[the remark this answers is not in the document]
christian-byrne: Fixed - now uses t('subgraphStore.disconnected') for i18n.

src/components/graph/DomWidgets.vue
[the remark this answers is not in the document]
christian-byrne: Fixed - now uses reference equality: currentGraph === override.node.graph

src/stores/domWidgetStore.ts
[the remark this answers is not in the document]
christian-byrne: Fixed - PositionOverride.widget is now typed as Raw<IBaseWidget> and markRaw is applied.

src/components/rightSidePanel/subgraph/SubgraphEditor.vue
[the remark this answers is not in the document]
christian-byrne: Fixed - now uses t('subgraphStore.linked') for i18n.

src/core/graph/subgraph/promotedWidgetView.test.ts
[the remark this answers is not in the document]
christian-byrne: Fixed - now uses describe(createPromotedWidgetView, ...) with function reference.

src/composables/node/usePromotedPreviews.test.ts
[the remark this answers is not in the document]
christian-byrne: Fixed - now uses describe(usePromotedPreviews, ...) with function reference.

src/core/graph/subgraph/proxyWidgetUtils.ts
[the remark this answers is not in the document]
christian-byrne: Fixed - promoteWidget and demoteWidget now derive nodeId from widget.sourceNodeId for promoted views.

src/core/graph/subgraph/promotedWidgetView.test.ts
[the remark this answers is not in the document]
christian-byrne: Addressed - mocks now use vi.hoisted pattern for per-test state isolation.

src/composables/node/usePromotedPreviews.test.ts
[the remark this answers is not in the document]
christian-byrne: Addressed - mocks now use vi.hoisted pattern for per-test state isolation.

christian-byrne:
## Review Follow-up

I've verified the latest commits address my major concerns:

**Fixed:**
- **Performance**: JSON.stringify removed from widget getter - now uses reference equality  
- **Reactivity**: `markRaw` added for both node and widget in `domWidgetStore.ts` (lines 84-85)
- **Dev warnings**: `console.warn` added for failed legacy entry resolution
- **Ref counting**: `movePromotion` optimized to just reorder without full decrement/increment cycle

**CI Status:** The shard 4 failure is an unrelated flaky screenshot test (`auto-linked-node.png` in `nodeSearchBox.spec.ts`) with 0.01 pixel ratio difference.

**Remaining minor items** (non-blocking, can be addressed in follow-up PRs):
- Some CodeRabbit nitpicks on test patterns
- AustinMroz's comments on `interior.mouse` check and `computeSize` validation

The core architecture looks solid. Ready for maintainer approval.

christian-byrne (dismissed):
The synthetic widgets getter architecture is solid. All major review concerns have been addressed:

- Performance: Reference equality replaces JSON.stringify snapshotting
- Reactivity: markRaw applied to both node and widget in domWidgetStore
- Data integrity: Dev warnings added for failed legacy entry resolution
- Ref counting: movePromotion optimized to simple reorder without churn
- Widget delegation: computeSize and mouse handlers properly forwarded

The CI failure (shard 4) is an unrelated flaky screenshot test in nodeSearchBox.spec.ts.

Remaining CodeRabbit nitpicks are non-blocking and can be addressed in follow-ups.

christian-byrne:
Note that I have not yet read the entire PR myself

christian-byrne (commented):
Nice refactoring in the latest commits. The `PromotedWidgetViewManager` cache reconciliation pattern is clean, and centralizing `resolvePromotedWidgetSource` reduces duplication.

One minor observation on `getProjectedWidget` in `promotedWidgetView.ts` (lines 257-280): the cache invalidation check compares `projectedWidget.type !== resolved.widget.type` after already checking node/widget identity. If the source widget type can change while the widget reference stays the same, this is correct. Otherwise this comparison is redundant since a changed type would imply a different widget reference. Not blocking, just a clarity question.

The `getNodeCount()` fix for CDP serialization is a good catch - avoids serializing the entire reactive node graph across the protocol boundary.

christian-byrne:
Nice refactoring in the latest commits. The `PromotedWidgetViewManager` cache reconciliation pattern is clean, and centralizing `resolvePromotedWidgetSource` reduces duplication.

One minor observation on `getProjectedWidget` in `promotedWidgetView.ts` (lines 257-280): the cache invalidation check compares `projectedWidget.type !== resolved.widget.type` after already checking node/widget identity. If the source widget type can change while the widget reference stays the same, this is correct. Otherwise this comparison is redundant since a changed type would imply a different widget reference. Not blocking, just a clarity question.

The `getNodeCount()` fix for CDP serialization is a good catch - avoids serializing the entire reactive node graph across the protocol boundary.

AustinMroz (approved):
Lots of big and scary changes here, but it needs to be merged early into the release cycle. It has my approval.

DrJKL:
> Lots of big and scary changes here, but it needs to be merged early into the release cycle. It has my approval.



> Lots of big and scary changes here, but it needs to be merged early into the release cycle. It has my approval.

YOLO

PR #8857, opened 2026-02-14
feat: add hideOutputImages flag for nodes with custom preview

jtydhr88:
## Summary

Prerequisite for upcoming native color correction nodes (ColorCorrect, ColorBalance, ColorCurves) which render their own WebGL preview and need to suppress the default output image display while keeping data in the store for downstream nodes.

before
image

after
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8857-feat-add-hideOutputImages-flag-for-nodes-with-custom-preview-3076d73d365081a2aa55d34280601b47) by [Unito](https://www.unito.io)


jtydhr88:
i don't think the failing tests caused by my changes, another two have the same failure tests

src/types/litegraph-augmentation.d.ts
```diff
@@ -177,6 +177,8 @@ declare module '@/lib/litegraph/src/litegraph' {
     isLoading?: boolean
     /** The content type of the node's preview media */
     previewMediaType?: 'image' | 'video' | 'audio' | 'model'
+    /** If true, output images are stored but not rendered below the node */
+    hideOutputImages?: boolean
```
DrJKL, line 181: We might want a more robust version of this later.

DrJKL (approved):

PR #8860, opened 2026-02-14
feat: add CurveEditor component

jtydhr88:
## Summary
Prerequisite for upcoming native color correction nodes (ColorCurves).

Reusable curve editor with monotone cubic Hermite interpolation, drag-to-add/move/delete control points, and SVG-based rendering.
Includes CurvePoint type, LUT generation utility, and useCurveEditor composable for interaction logic.

https://github.com/user-attachments/assets/948352c7-bdf2-40f9-a8f0-35bc2b2f3202

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8860-feat-add-CurveEditor-component-and-d3-shape-dependency-3076d73d3650817f8421f98e349569d0) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/curve/CurveEditor.test.ts
+++ b/src/components/curve/CurveEditor.test.ts
@@ -1,0 +1,113 @@
+import { mount } from '@vue/test-utils'
+import { describe, expect, it } from 'vitest'
+
+import type { CurvePoint } from '@/lib/litegraph/src/types/widgets'
+
+import CurveEditor from './CurveEditor.vue'
+
+function mountEditor(points: CurvePoint[], extraProps = {}) {
+  return mount(CurveEditor, {
+    props: { modelValue: points, ...extraProps }
+  })
+}
+
+function getCurvePath(wrapper: ReturnType<typeof mount>) {
+  return wrapper.find('[data-testid="curve-path"]')
+}
+
+describe('CurveEditor', () => {
+  it('renders SVG with curve path', () => {
+    const wrapper = mountEditor([
+      [0, 0],
+      [1, 1]
+    ])
+    expect(wrapper.find('svg').exists()).toBe(true)
+    const curvePath = getCurvePath(wrapper)
+    expect(curvePath.exists()).toBe(true)
+    expect(curvePath.attributes('d')).toBeTruthy()
+  })
+
+  it('renders a circle for each control point', () => {
+    const wrapper = mountEditor([
+      [0, 0],
+      [0.5, 0.7],
+      [1, 1]
+    ])
+    expect(wrapper.findAll('circle')).toHaveLength(3)
+  })
+
+  it('renders histogram path when provided', () => {
+    const histogram = new Uint32Array(256)
+    for (let i = 0; i < 256; i++) histogram[i] = i + 1
+    const wrapper = mountEditor(
+      [
+        [0, 0],
+        [1, 1]
+      ],
+      { histogram }
+    )
+    const histogramPath = wrapper.find('[data-testid="histogram-path"]')
+    expect(histogramPath.exists()).toBe(true)
+    expect(histogramPath.attributes('d')).toContain('M0,1')
+  })
+
+  it('does not render histogram path when not provided', () => {
+    const wrapper = mountEditor([
+      [0, 0],
+      [1, 1]
+    ])
+    expect(wrapper.find('[data-testid="histogram-path"]').exists()).toBe(false)
+  })
+
+  it('returns empty path with fewer than 2 points', () => {
+    const wrapper = mountEditor([[0.5, 0.5]])
+    expect(getCurvePath(wrapper).attributes('d')).toBe('')
+  })
+
+  it('generates path starting with M and containing L segments', () => {
+    const wrapper = mountEditor([
+      [0, 0],
+      [0.5, 0.8],
+      [1, 1]
+    ])
+    const d = getCurvePath(wrapper).attributes('d')!
+    expect(d).toMatch(/^M/)
+    expect(d).toContain('L')
+  })
+
+  it('curve path only spans the x-range of control points', () => {
+    const wrapper = mountEditor([
+      [0.2, 0.3],
+      [0.8, 0.9]
+    ])
+    const d = getCurvePath(wrapper).attributes('d')!
+    const xValues = d
+      .split(/[ML]/)
+      .filter(Boolean)
+      .map((s) => parseFloat(s.split(',')[0]))
+    expect(Math.min(...xValues)).toBeCloseTo(0.2, 2)
+    expect(Math.max(...xValues)).toBeCloseTo(0.8, 2)
+  })
+
+  it('deletes a point on right-click but keeps minimum 2', async () => {
+    const points: CurvePoint[] = [
+      [0, 0],
+      [0.5, 0.5],
+      [1, 1]
+    ]
+    const wrapper = mountEditor(points)
+    expect(wrapper.findAll('circle')).toHaveLength(3)
+
+    await wrapper.findAll('circle')[1].trigger('pointerdown', {
+      button: 2,
+      pointerId: 1
+    })
+    expect(wrapper.findAll('circle')).toHaveLength(2)
+
+    await wrapper.findAll('circle')[0].trigger('pointerdown', {
+      button: 2,
+      pointerId: 1
+    })
+    expect(wrapper.findAll('circle')).toHaveLength(2)
+  })
+})
--- a/src/components/curve/CurveEditor.vue
+++ b/src/components/curve/CurveEditor.vue
@@ -1,0 +1,103 @@
+<template>
+  <svg
+    ref="svgRef"
+    viewBox="-0.04 -0.04 1.08 1.08"
+    preserveAspectRatio="xMidYMid meet"
+    class="aspect-square w-full cursor-crosshair rounded-[5px] bg-node-component-surface"
+    @pointerdown.stop="handleSvgPointerDown"
+    @contextmenu.prevent.stop
+  >
+    <line
+      v-for="v in [0.25, 0.5, 0.75]"
+      :key="'h' + v"
+      :x1="0"
+      :y1="v"
+      :x2="1"
+      :y2="v"
+      stroke="currentColor"
+      stroke-opacity="0.1"
+      stroke-width="0.003"
+    />
+    <line
+      v-for="v in [0.25, 0.5, 0.75]"
+      :key="'v' + v"
+      :x1="v"
+      :y1="0"
+      :x2="v"
+      :y2="1"
+      stroke="currentColor"
+      stroke-opacity="0.1"
+      stroke-width="0.003"
+    />
+
+    <line
+      x1="0"
+      y1="1"
+      x2="1"
+      y2="0"
+      stroke="currentColor"
+      stroke-opacity="0.15"
+      stroke-width="0.003"
+    />
+
+    <path
+      v-if="histogramPath"
+      data-testid="histogram-path"
+      :d="histogramPath"
+      :fill="curveColor"
+      fill-opacity="0.15"
+      stroke="none"
+    />
+
+    <path
+      data-testid="curve-path"
+      :d="curvePath"
+      fill="none"
+      :stroke="curveColor"
+      stroke-width="0.008"
+      stroke-linecap="round"
+    />
+
+    <circle
+      v-for="(point, i) in modelValue"
+      :key="i"
+      :cx="point[0]"
+      :cy="1 - point[1]"
+      r="0.02"
+      :fill="curveColor"
+      stroke="white"
+      stroke-width="0.004"
+      class="cursor-grab"
+      @pointerdown.stop="startDrag(i, $event)"
+    />
+  </svg>
+</template>
+
+<script setup lang="ts">
+import { computed, useTemplateRef } from 'vue'
+
+import { useCurveEditor } from '@/composables/useCurveEditor'
+import type { CurvePoint } from '@/lib/litegraph/src/types/widgets'
+
+import { histogramToPath } from './curveUtils'
+
+const { curveColor = 'white', histogram } = defineProps<{
+  curveColor?: string
+  histogram?: Uint32Array | null
+}>()
+
+const modelValue = defineModel<CurvePoint[]>({
+  required: true
+})
+
+const svgRef = useTemplateRef<SVGSVGElement>('svgRef')
+
+const { curvePath, handleSvgPointerDown, startDrag } = useCurveEditor({
+  svgRef,
+  modelValue
+})
+
+const histogramPath = computed(() =>
+  histogram ? histogramToPath(histogram) : ''
+)
+</script>
--- a/src/components/curve/curveUtils.test.ts
+++ b/src/components/curve/curveUtils.test.ts
@@ -1,0 +1,141 @@
+import { describe, expect, it } from 'vitest'
+
+import type { CurvePoint } from '@/lib/litegraph/src/types/widgets'
+
+import {
+  createMonotoneInterpolator,
+  curvesToLUT,
+  histogramToPath
+} from './curveUtils'
+
+describe('createMonotoneInterpolator', () => {
+  it('returns 0 for empty points', () => {
+    const interpolate = createMonotoneInterpolator([])
+    expect(interpolate(0.5)).toBe(0)
+  })
+
+  it('returns constant for single point', () => {
+    const interpolate = createMonotoneInterpolator([[0.5, 0.7]])
+    expect(interpolate(0)).toBe(0.7)
+    expect(interpolate(1)).toBe(0.7)
+  })
+
+  it('passes through control points exactly', () => {
+    const points: CurvePoint[] = [
+      [0, 0],
+      [0.5, 0.8],
+      [1, 1]
+    ]
+    const interpolate = createMonotoneInterpolator(points)
+    expect(interpolate(0)).toBeCloseTo(0, 5)
+    expect(interpolate(0.5)).toBeCloseTo(0.8, 5)
+    expect(interpolate(1)).toBeCloseTo(1, 5)
+  })
+
+  it('clamps to endpoint values outside range', () => {
+    const points: CurvePoint[] = [
+      [0.2, 0.3],
+      [0.8, 0.9]
+    ]
+    const interpolate = createMonotoneInterpolator(points)
+    expect(interpolate(0)).toBe(0.3)
+    expect(interpolate(1)).toBe(0.9)
+  })
+
+  it('produces monotone output for monotone input', () => {
+    const points: CurvePoint[] = [
+      [0, 0],
+      [0.25, 0.2],
+      [0.5, 0.5],
+      [0.75, 0.8],
+      [1, 1]
+    ]
+    const interpolate = createMonotoneInterpolator(points)
+
+    let prev = -Infinity
+    for (let x = 0; x <= 1; x += 0.01) {
+      const y = interpolate(x)
+      expect(y).toBeGreaterThanOrEqual(prev)
+      prev = y
+    }
+  })
+
+  it('handles unsorted input points', () => {
+    const points: CurvePoint[] = [
+      [1, 1],
+      [0, 0],
+      [0.5, 0.5]
+    ]
+    const interpolate = createMonotoneInterpolator(points)
+    expect(interpolate(0)).toBeCloseTo(0, 5)
+    expect(interpolate(0.5)).toBeCloseTo(0.5, 5)
+    expect(interpolate(1)).toBeCloseTo(1, 5)
+  })
+})
+
+describe('curvesToLUT', () => {
+  it('returns a 256-entry Uint8Array', () => {
+    const lut = curvesToLUT([
+      [0, 0],
+      [1, 1]
+    ])
+    expect(lut).toBeInstanceOf(Uint8Array)
+    expect(lut.length).toBe(256)
+  })
+
+  it('produces identity LUT for diagonal curve', () => {
+    const lut = curvesToLUT([
+      [0, 0],
+      [1, 1]
+    ])
+    for (let i = 0; i < 256; i++) {
+      expect(lut[i]).toBeCloseTo(i, 0)
+    }
+  })
+
+  it('clamps output to [0, 255]', () => {
+    const lut = curvesToLUT([
+      [0, 0],
+      [0.5, 1.5],
+      [1, 1]
+    ])
+    for (let i = 0; i < 256; i++) {
+      expect(lut[i]).toBeGreaterThanOrEqual(0)
+      expect(lut[i]).toBeLessThanOrEqual(255)
+    }
+  })
+})
+
+describe('histogramToPath', () => {
+  it('returns empty string for empty histogram', () => {
+    expect(histogramToPath(new Uint32Array(0))).toBe('')
+  })
+
+  it('returns empty string when all bins are zero', () => {
+    expect(histogramToPath(new Uint32Array(256))).toBe('')
+  })
+
+  it('returns a closed SVG path for valid histogram', () => {
+    const histogram = new Uint32Array(256)
+    for (let i = 0; i < 256; i++) histogram[i] = i + 1
+    const path = histogramToPath(histogram)
+    expect(path).toMatch(/^M0,1/)
+    expect(path).toMatch(/L1,1 Z$/)
+  })
+
+  it('normalizes using 99.5th percentile to suppress outliers', () => {
+    const histogram = new Uint32Array(256)
+    for (let i = 0; i < 256; i++) histogram[i] = 100
+    histogram[255] = 100000
+    const path = histogramToPath(histogram)
+    // Most bins should map to y=0 (1 - 100/100 = 0) since
+    // the 99.5th percentile is 100, not the outlier 100000
+    const yValues = path
+      .split(/[ML]/)
+      .filter(Boolean)
+      .map((s) => parseFloat(s.split(',')[1]))
+      .filter((y) => !isNaN(y))
+    const nearZero = yValues.filter((y) => Math.abs(y) < 0.01)
+    expect(nearZero.length).toBeGreaterThan(200)
+  })
+})
--- a/src/components/curve/curveUtils.ts
+++ b/src/components/curve/curveUtils.ts
@@ -1,0 +1,124 @@
+import type { CurvePoint } from '@/lib/litegraph/src/types/widgets'
+
+/**
+ * Monotone cubic Hermite interpolation.
+ * Produces a smooth curve that passes through all control points
+ * without overshooting (monotone property).
+ *
+ * Returns a function that evaluates y for any x in [0, 1].
+ */
+export function createMonotoneInterpolator(
+  points: CurvePoint[]
+): (x: number) => number {
+  if (points.length === 0) return () => 0
+  if (points.length === 1) return () => points[0][1]
+
+  const sorted = [...points].sort((a, b) => a[0] - b[0])
+  const n = sorted.length
+  const xs = sorted.map((p) => p[0])
+  const ys = sorted.map((p) => p[1])
+
+  const deltas: number[] = []
+  const slopes: number[] = []
+  for (let i = 0; i < n - 1; i++) {
+    const dx = xs[i + 1] - xs[i]
+    deltas.push(dx === 0 ? 0 : (ys[i + 1] - ys[i]) / dx)
+  }
+
+  slopes.push(deltas[0] ?? 0)
+  for (let i = 1; i < n - 1; i++) {
+    if (deltas[i - 1] * deltas[i] <= 0) {
+      slopes.push(0)
+    } else {
+      slopes.push((deltas[i - 1] + deltas[i]) / 2)
+    }
+  }
+  slopes.push(deltas[n - 2] ?? 0)
+
+  for (let i = 0; i < n - 1; i++) {
+    if (deltas[i] === 0) {
+      slopes[i] = 0
+      slopes[i + 1] = 0
+    } else {
+      const alpha = slopes[i] / deltas[i]
+      const beta = slopes[i + 1] / deltas[i]
+      const s = alpha * alpha + beta * beta
+      if (s > 9) {
+        const t = 3 / Math.sqrt(s)
+        slopes[i] = t * alpha * deltas[i]
+        slopes[i + 1] = t * beta * deltas[i]
+      }
+    }
+  }
+
+  return (x: number): number => {
+    if (x <= xs[0]) return ys[0]
+    if (x >= xs[n - 1]) return ys[n - 1]
+
+    let lo = 0
+    let hi = n - 1
+    while (lo < hi - 1) {
+      const mid = (lo + hi) >> 1
+      if (xs[mid] <= x) lo = mid
+      else hi = mid
+    }
+
+    const dx = xs[hi] - xs[lo]
+    if (dx === 0) return ys[lo]
+
+    const t = (x - xs[lo]) / dx
+    const t2 = t * t
+    const t3 = t2 * t
+
+    const h00 = 2 * t3 - 3 * t2 + 1
+    const h10 = t3 - 2 * t2 + t
+    const h01 = -2 * t3 + 3 * t2
+    const h11 = t3 - t2
+
+    return (
+      h00 * ys[lo] +
+      h10 * dx * slopes[lo] +
+      h01 * ys[hi] +
+      h11 * dx * slopes[hi]
+    )
+  }
+}
+
+/**
+ * Generate a 256-entry lookup table from curve control points.
+ * Points are in [0, 1] space; output is clamped to [0, 255] as Uint8.
+ */
+/**
+ * Convert a 256-bin histogram into an SVG path string.
+ * Normalizes using the 99.5th percentile to avoid outlier spikes.
+ */
+export function histogramToPath(histogram: Uint32Array): string {
+  if (!histogram.length) return ''
+
+  const sorted = Array.from(histogram).sort((a, b) => a - b)
+  const max = sorted[Math.floor(255 * 0.995)]
+  if (max === 0) return ''
+
+  const step = 1 / 255
+  let d = 'M0,1'
+  for (let i = 0; i < 256; i++) {
+    const x = i * step
+    const y = 1 - Math.min(1, histogram[i] / max)
+    d += ` L${x.toFixed(4)},${y.toFixed(4)}`
+  }
+  d += ' L1,1 Z'
+  return d
+}
+
+export function curvesToLUT(points: CurvePoint[]): Uint8Array {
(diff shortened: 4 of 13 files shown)


jtydhr88:
i don't think the failing tests caused by my changes, another two have the same failure tests

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,101 @@
...
+      modelValue.value = newPoints
+    }
+
+    const onUp = () => {
+      dragIndex.value = -1
+      svg.removeEventListener('pointermove', onMove)
+      svg.removeEventListener('pointerup', onUp)
+    }
+
+    svg.addEventListener('pointermove', onMove)
```
christian-byrne, line 126: **[C1] Critical: Event listener leak risk**

Raw `addEventListener`/`removeEventListener` is used for `pointermove` and `pointerup` during drag. If the component unmounts mid-drag (e.g. tab switch, conditional rendering), `onUp` never fires and these listeners leak on the SVG element. There is also no `lostpointercapture` handler, so if the browser cancels pointer capture, `dragIndex` stays stuck.

Codebase convention (e.g. `useNodeResize.ts`) is to use `useEventListener` from `@vueuse/core`, which auto-removes on scope disposal.

**Suggested fix:**
```ts
import { useEventListener } from '@vueuse/core'

const stopMove = useEventListener(svg, 'pointermove', onMove)
const stopUp = useEventListener(svg, 'pointerup', () => {
  dragIndex.value = -1
  stopMove()
  stopUp()
})
```
Also add a `lostpointercapture` handler to reset drag state.

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,101 @@
+import { computed, ref } from 'vue'
+import type { Ref } from 'vue'
+import { curveMonotoneX, line } from 'd3-shape'
```
christian-byrne: **[M1] Major: Dual interpolation — d3-shape may diverge from LUT**

Two independent implementations of monotone cubic Hermite interpolation exist:
1. `curveUtils.ts` — `createMonotoneInterpolator` (Fritsch-Carlson)
2. `d3-shape` — `curveMonotoneX` (different algorithm)

These can produce different curves for the same control points, meaning the SVG preview could diverge from the actual LUT output applied to the image. Also adds `d3-shape` (~268KB source) for ~5 lines of usage.

**Suggested fix:** Remove `d3-shape`. Generate the SVG path by sampling `createMonotoneInterpolator` at sufficient resolution (e.g. 64–128 points) and building a polyline path `d` string. This guarantees visual/functional consistency and removes the dependency.

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,101 @@
...
+
+import type { CurvePoint } from '@/lib/litegraph/src/types/widgets'
+
+interface UseCurveEditorOptions {
+  svgRef: Ref<SVGSVGElement | null>
+  modelValue: Ref<CurvePoint[]>
+}
+
+export function useCurveEditor({ svgRef, modelValue }: UseCurveEditorOptions) {
```
christian-byrne, line 12: **[M2] Major: No unit tests for this composable**

This composable contains all critical interaction logic — coordinate transformation, point addition, point deletion, drag behavior, and boundary clamping — but has zero test coverage. The component tests only verify rendering.

**Suggested fix:** Add `src/composables/useCurveEditor.test.ts` covering:
1. `svgCoords` clamps to [0, 1]
2. `handleSvgPointerDown` adds a point and starts drag
3. Right-click / ctrl+click deletes a point (but not below 2)
4. `startDrag` + pointermove updates point position
5. `curvePath` computed returns a valid SVG path string

Mock the SVG element's `createSVGPoint` / `getScreenCTM`.

src/components/colorcurves/CurveEditor.test.ts
```diff
@@ -0,0 +1,67 @@
...
+      props: {
+        modelValue: [
+          [0, 0],
+          [1, 1]
+        ] as CurvePoint[]
+      }
+    })
+    const svg = wrapper.find('svg')
+    expect(svg.exists()).toBe(true)
+    const curvePath = wrapper.find('path[stroke-width="0.008"]')
```
christian-byrne: **[M3] Major: Tests are change detectors, not behavioral**

Querying by `path[stroke-width="0.008"]` and `path[fill-opacity="0.15"]` (line 52) tests visual attributes, not behavior. Per project guidelines: *"Do not write tests dependent on non-behavioral features (styles, classes)"*. If someone tweaks the stroke width for aesthetics, these tests break.

Also, none of these tests exercise actual user behavior (dragging, adding/removing points, model updates).

**Suggested fix:** Add `data-testid` attributes to the SVG paths (e.g. `data-testid="curve-path"`, `data-testid="histogram-path"`) and query by those instead.

src/components/colorcurves/CurveEditor.test.ts
```diff
@@ -0,0 +1,67 @@
...
+    const points: CurvePoint[] = [
+      [0, 0],
+      [0.5, 0.7],
+      [1, 1]
+    ]
+    const wrapper = mount(CurveEditor, {
+      props: { modelValue: points }
+    })
+    const circles = wrapper.findAll('circle')
+    expect(circles.length).toBe(3)
```
christian-byrne: **[M3 cont.] Assertion style**

Per project convention (vitest-patterns.md): *"Prefer `.toHaveLength()` over `.length.toBe()`"*

```suggestion
    expect(circles).toHaveLength(3)
```

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,101 @@
...
+    const [x, y] = svgCoords(e)
+    const newPoints: CurvePoint[] = [...modelValue.value, [x, y]]
+    modelValue.value = newPoints
+
+    const newIndex = newPoints.length - 1
+    startDrag(newIndex, e)
+  }
+
+  function startDrag(index: number, e: PointerEvent) {
+    if (e.button === 2 || (e.button === 0 && e.ctrlKey)) {
```
christian-byrne, line 89: **[m1] Minor: Endpoint deletion not prevented**

The delete logic prevents removing below 2 points, but doesn't prevent removing endpoint points (x≈0 or x≈1). If a user deletes both endpoints and is left with two arbitrary interior points (e.g. `[0.3, 0.5]` and `[0.7, 0.8]`), the curve won't span the full [0, 1] range, causing clamped extrapolation in LUT generation.

Consider preventing deletion of the first and last points (by x-value), or document this as intentional.

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,101 @@
...
+    if (!ctm) return [0, 0]
+
+    const svgPt = pt.matrixTransform(ctm.inverse())
+    return [
+      Math.max(0, Math.min(1, svgPt.x)),
+      Math.max(0, Math.min(1, 1 - svgPt.y))
+    ]
+  }
+
+  function handleSvgPointerDown(e: PointerEvent) {
```
christian-byrne, line 65: **[m2] Minor: No proximity check before adding points**

Clicking very close to an existing control point adds a new overlapping point instead of dragging the existing one. This makes them impossible to separate.

Consider adding a distance threshold — if the click is within a threshold of an existing point, start dragging that point instead of adding a new one.

src/lib/litegraph/src/types/widgets.ts
```diff
@@ -292,6 +292,8 @@ export interface IBoundingBoxWidget extends IBaseWidget<Bounds, 'boundingbox'> {
   value: Bounds
 }
 
+export type CurvePoint = [x: number, y: number]
```
christian-byrne, line 332: **[m3] Minor: CurvePoint type placement**

`CurvePoint` is a data type for curve control points, not a widget definition. All consumers (`curveUtils.ts`, `useCurveEditor.ts`, `CurveEditor.vue`) are in the `colorcurves` domain.

Consider moving to `src/components/colorcurves/types.ts` or colocating in `curveUtils.ts` to keep `widgets.ts` focused on widget definitions.

src/components/colorcurves/CurveEditor.vue
```diff
@@ -0,0 +1,113 @@
...
+})
+
+const svgRef = useTemplateRef<SVGSVGElement>('svgRef')
+
+const { curvePath, handleSvgPointerDown, startDrag } = useCurveEditor({
+  svgRef,
+  modelValue
+})
+
+const histogramPath = computed(() => {
```
christian-byrne: **[m4] Minor: Histogram path logic not extracted for testability**

This computed contains non-trivial percentile-based normalization (99.5th percentile via sort). It's inline and cannot be tested in isolation — `curveUtils.test.ts` doesn't cover it.

Consider extracting to `histogramToPath(histogram: Uint32Array): string` in `curveUtils.ts` and adding tests for the normalization logic.

christian-byrne (commented):
Looking good

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,101 @@
...
+      .curve(curveMonotoneX)
+
+    return lineGenerator(points) ?? ''
+  })
+
+  function svgCoords(e: PointerEvent): [number, number] {
+    const svg = svgRef.value
+    if (!svg) return [0, 0]
+
+    const pt = svg.createSVGPoint()
```
christian-byrne: **[N1] Nitpick: Deprecated `createSVGPoint()` API**

`SVGSVGElement.createSVGPoint()` is deprecated in SVG2. Consider replacing with:
```ts
const pt = new DOMPoint(e.clientX, e.clientY)
const svgPt = pt.matrixTransform(ctm.inverse())
```

src/components/curve/CurveEditor.vue
```diff
@@ -0,0 +1,113 @@
...
+      fill="none"
+      :stroke="curveColor"
+      stroke-width="0.008"
+      stroke-linecap="round"
+    />
+
+    <circle
+      v-for="(point, i) in modelValue"
+      :key="i"
+      :cx="point[0]"
```
christian-byrne, line 64: **[N2] Nitpick: Array index as `:key`**

Using `i` as `:key` can cause incorrect DOM reuse when points are added/removed from the middle. Consider a stable key like `` :key="`${point[0]}-${point[1]}`" ``.

src/composables/useCurveEditor.ts
jtydhr88: fixed

src/components/colorcurves/CurveEditor.vue
jtydhr88: fixed

src/composables/useCurveEditor.ts
jtydhr88: fixed

src/components/colorcurves/CurveEditor.test.ts
jtydhr88: fixed

src/components/colorcurves/CurveEditor.test.ts
jtydhr88: fixed

src/composables/useCurveEditor.ts
jtydhr88: fixed

src/composables/useCurveEditor.ts
jtydhr88: fixed

src/composables/useCurveEditor.ts
jtydhr88: fixed

src/components/curve/CurveEditor.vue
```diff
@@ -0,0 +1,103 @@
...
+      :cy="1 - point[1]"
+      r="0.02"
+      :fill="curveColor"
+      stroke="white"
```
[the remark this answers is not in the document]
jtydhr88: Not an issue in practice — CurveEditor is only rendered inside node widgets which always have a dark background (bg-node-component-surface). There's no light theme in this project.     
The stroke="white" provides contrast against the curveColor fill to keep control points visible; currentColor wouldn't work here since it's the text color, not guaranteed to contrast   with the fill. 
Also, CSS variables in SVG attributes (not style properties) have inconsistent support.

src/components/curve/curveUtils.ts
```diff
@@ -0,0 +1,124 @@
...
+  const step = 1 / 255
+  let d = 'M0,1'
+  for (let i = 0; i < 256; i++) {
+    const x = i * step
+    const y = 1 - Math.min(1, histogram[i] / max)
+    d += ` L${x.toFixed(4)},${y.toFixed(4)}`
+  }
+  d += ' L1,1 Z'
+  return d
+}
```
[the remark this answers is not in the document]
jtydhr88: Internal function — only called with 256-bin output from computeHistogram. 
Per project guidelines, we validate at system boundaries, not between internal functions with a single call site.

src/components/curve/CurveEditor.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  return mount(CurveEditor, {
+    props: { modelValue: points, ...extraProps }
+  })
+}
+
+function getCurvePath(wrapper: ReturnType<typeof mount>) {
+  return wrapper.find('[data-testid="curve-path"]')
+}
+
+describe('CurveEditor', () => {
```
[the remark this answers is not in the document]
jtydhr88: There is no vitest/prefer-describe-function-title rule in this project's config. describe('CurveEditor', ...) is the standard convention used throughout the codebase.

src/components/curve/CurveEditor.vue
```diff
@@ -0,0 +1,103 @@
...
+
+import { useCurveEditor } from '@/composables/useCurveEditor'
+import type { CurvePoint } from '@/lib/litegraph/src/types/widgets'
+
+import { histogramToPath } from './curveUtils'
+
+const { curveColor = 'white', histogram } = defineProps<{
+  curveColor?: string
+  histogram?: Uint32Array | null
+}>()
```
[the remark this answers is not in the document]
jtydhr88: Same as the stroke="white" comment — no light theme, always rendered inside dark node backgrounds. 
The default 'white' represents the RGB master channel color (red/green/blue channels pass explicit colors). 'currentColor' would lose this semantic meaning.

src/components/curve/CurveEditor.test.ts
christian-byrne, line 1: @coderabbitai make an Issue to do a followup refactor converting/moving all files/folders to DDD and not using technical layers like top level src/composables and src/components. All things related to rendering should be in the renderer domain. Any pure utils that need to be re-used can be in the base folder. Graph domain can be in core/graph. Workspace/chrome UI domain can be in src/workspace.

src/components/curve/curveUtils.test.ts
christian-byrne, line 1: @coderabbitai make a followup issue to add more extensive unit test coverage to curveUtils. Since it's just a module of pure functions doing somewhat complex math/geometry calculations, we will naturally want an extremely large test suite given the number of permutations. Any time we add a new geometry/math engine that is manually coded, we want a ton of test coverage. These are the types of code that are very easy to write tests for and be able to assert all deterministic expectations match the actual calculated results. We should also consider and cover all edge cases.

src/components/curve/curveUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+  const interpolate = createMonotoneInterpolator(points)
+
+  for (let i = 0; i < 256; i++) {
+    const x = i / 255
+    const y = interpolate(x)
+    lut[i] = Math.max(0, Math.min(255, Math.round(y * 255)))
+  }
+
+  return lut
+}
```
christian-byrne, line 120: Since `x = i / 255` is strictly increasing, we can walk the curve segments linearly instead of performing a binary search for every sample.

```ts
export function curvesToLUT(points: CurvePoint[]): Uint8Array {
  const lut = new Uint8Array(256)
  const { xs, ys, slopes } = precomputeInterpolator(points)

  const n = xs.length
  let lo = 0
  let hi = 1

  for (let i = 0; i < 256; i++) {
    const x = i / 255

    while (hi < n - 1 && x > xs[hi]) {
      lo++
      hi++
    }

    const dx = xs[hi] - xs[lo]
    if (dx === 0) {
      lut[i] = Math.round(ys[lo] * 255)
      continue
    }

    const t = (x - xs[lo]) / dx
    const t2 = t * t
    const t3 = t2 * t

    const h00 = 2 * t3 - 3 * t2 + 1
    const h10 = t3 - 2 * t2 + t
    const h01 = -2 * t3 + 3 * t2
    const h11 = t3 - t2

    const y =
      h00 * ys[lo] +
      h10 * dx * slopes[lo] +
      h01 * ys[hi] +
      h11 * dx * slopes[hi]

    lut[i] = Math.max(0, Math.min(255, Math.round(y * 255)))
  }

  return lut
}
```

@coderabbitai make a followup Issue to replace the binary search With monotonic segment walk

src/components/curve/curveUtils.ts
```diff
@@ -0,0 +1,120 @@
+import type { CurvePoint } from '@/lib/litegraph/src/types/widgets'
+
+/**
+ * Monotone cubic Hermite interpolation.
+ * Produces a smooth curve that passes through all control points
+ * without overshooting (monotone property).
+ *
+ * Returns a function that evaluates y for any x in [0, 1].
+ */
+export function createMonotoneInterpolator(
```
christian-byrne, line 10: I think we can precompute per-segment data and reduce the number of operations a magnitude of order.

```ts
interface Segment {
  x0: number
  dx: number
  y0: number
  y1: number
  m0dx: number
  m1dx: number
}

function buildSegments(xs: Float64Array, ys: Float64Array, slopes: Float64Array): Segment[] {
  const segments: Segment[] = []

  for (let i = 0; i < xs.length - 1; i++) {
    const dx = xs[i + 1] - xs[i]
    segments.push({
      x0: xs[i],
      dx,
      y0: ys[i],
      y1: ys[i + 1],
      m0dx: dx * slopes[i],
      m1dx: dx * slopes[i + 1],
    })
  }

  return segments
}
```

@coderabbitai make a followup Issue to precompute the segment constants in this fashion to reduce the number of array reads and multiplies.

src/components/curve/curveUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+  points: CurvePoint[]
+): (x: number) => number {
+  if (points.length === 0) return () => 0
+  if (points.length === 1) return () => points[0][1]
+
+  const sorted = [...points].sort((a, b) => a[0] - b[0])
+  const n = sorted.length
+  const xs = sorted.map((p) => p[0])
+  const ys = sorted.map((p) => p[1])
```
christian-byrne, line 19: Modern JS engines / JIT compilers gain pretty extreme perf benefits when you make it easier for them to vectorize operations. You can do that by using typed arrays (which have more numeric stability). Example:

Replace dynamic JS arrays with `Float64Array`.

```ts
const xs = new Float64Array(n)
const ys = new Float64Array(n)

for (let i = 0; i < n; i++) {
  xs[i] = sorted[i][0]
  ys[i] = sorted[i][1]
}
```

Same for the below with:

```ts
const deltas = new Float64Array(n - 1)
const slopes = new Float64Array(n)
```

@coderabbitai make a followup Issue to apply these optimizations and make the code more JIT friendly

src/components/curve/curveUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Convert a 256-bin histogram into an SVG path string.
+ * Normalizes using the 99.5th percentile to avoid outlier spikes.
+ */
+export function histogramToPath(histogram: Uint32Array): string {
+  if (!histogram.length) return ''
+
+  const sorted = Array.from(histogram).sort((a, b) => a - b)
+  const max = sorted[Math.floor(255 * 0.995)]
```
christian-byrne, line 95: We are re-sorting the historgram on every call 256 times. I believe we can use a QuickSelect (nth-element style) to find percentile without sorting all 256 elements:

```ts
function percentile(arr: Uint32Array, p: number): number {
  const copy = new Uint32Array(arr)
  const k = Math.floor(arr.length * p)
  return quickSelect(copy, k)
}
```

@coderabbitai Make a followup Issue to apply QuickSelect alg in this way.

src/components/curve/curveUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+  if (max === 0) return ''
+
+  const step = 1 / 255
+  let d = 'M0,1'
+  for (let i = 0; i < 256; i++) {
+    const x = i * step
+    const y = 1 - Math.min(1, histogram[i] / max)
+    d += ` L${x.toFixed(4)},${y.toFixed(4)}`
```
christian-byrne, line 103: String concat inside the hotpath is suboptimal. We can use Array join for better perf:

```ts
const parts: string[] = ['M0,1']

for (let i = 0; i < 256; i++) {
  parts.push(` L${x.toFixed(4)},${y.toFixed(4)}`)
}

parts.push(' L1,1 Z')
return parts.join('')
```

---

Likewise, we should just generally avoid `toFixed` in the hot path (manual roudning is faster), but this is extremely marginal. Would work by replacing

```ts
x.toFixed(4)
```

With:

```ts
Math.round(x * 10000) / 10000
```

(Or precompute `step = 1 / 255` and accumulate `x += step`).

---

lastly, there's another easy win here by hoisting the repeated step calculations outside the loop. Instead of computing inside the loop:

```ts
const step = 1 / 255
const x = i * step
```

Use:

```ts
let x = 0
const step = 1 / 255

for (let i = 0; i < 256; i++) {
  ...
  x += step
}
```

---

@coderabbitai Make a followup Issue to apply these optimizations

src/components/curve/curveUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Returns a function that evaluates y for any x in [0, 1].
+ */
+export function createMonotoneInterpolator(
+  points: CurvePoint[]
+): (x: number) => number {
+  if (points.length === 0) return () => 0
+  if (points.length === 1) return () => points[0][1]
+
+  const sorted = [...points].sort((a, b) => a[0] - b[0])
```
christian-byrne, line 16: Extreme nit: we could bake into the API contract that the data will be sorted then avoid the array copy with points.sort(). Personally I would not do this, but just noting.

src/components/curve/curveUtils.ts
```diff
@@ -0,0 +1,120 @@
...
+    const y = 1 - Math.min(1, histogram[i] / max)
+    d += ` L${x.toFixed(4)},${y.toFixed(4)}`
+  }
+  d += ' L1,1 Z'
+  return d
+}
+
+export function curvesToLUT(points: CurvePoint[]): Uint8Array {
+  const lut = new Uint8Array(256)
+  const interpolate = createMonotoneInterpolator(points)
```
christian-byrne, line 111: nit: If this runs during dragging, is it worth avoiding rebuilding interpolator every frame.

```ts
if (pointsChanged) {
  cachedInterpolator = createMonotoneInterpolator(points)
}
```

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    const onMove = (ev: PointerEvent) => {
+      if (dragIndex.value < 0) return
+      const [x, y] = svgCoords(ev)
+      const newPoints = [...modelValue.value]
+      newPoints[dragIndex.value] = [x, y]
+      modelValue.value = newPoints
+    }
+
+    const endDrag = () => {
```
christian-byrne, line 115: Right now we are sorting in both the `sortedPoints` computed (copy + sort) AND in `endDrag` (`sortByX`).

Possible improvements:

1. Do NOT maintain `modelValue` sorted while dragging; only keep a sorted *view*
- Keep `modelValue` in insertion/selection order (stable indices).
- Use `sortedPoints` for rendering / interpolator only.
- On endDrag: do **nothing** (or only sort if you truly require sorted model storage).

Benefits:
- avoids reordering indices, avoids potential UX bugs, avoids extra sorts.

2. Maintain sorted storage, but then remove `sortedPoints` sorting
- Enforce invariant: `modelValue.value` is always sorted by x.
- Then `sortedPoints` becomes just `modelValue.value` (no copy/sort).
- But you must then update dragging logic to not rely on stable indices.

@coderabbitai make a followup Issue for this.

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+    }
+    return nearest
+  }
+
+  function handleSvgPointerDown(e: PointerEvent) {
+    if (e.button !== 0) return
+
+    const [x, y] = svgCoords(e)
+
+    const nearby = findNearestPoint(x, y)
```
christian-byrne, line 70: The issue with this is that `findNearestPoint` iterates `modelValue.value`, but `curvePath` uses `sortedPoints`. If `modelValue` isn't already sorted, we have:

- render order ≠ drag order
- and you pay for sorting + potential mismatch logic

Possible solutions:

- Either enforce `modelValue` always sorted (and then drop `sortedPoints` sort),
- Or treat dragIndex as an index into `modelValue` only and accept that rendering uses a sorted copy (but then be careful about point selection mapping).

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    dragIndex.value = index
+    const svg = svgRef.value
+    if (!svg) return
+
+    svg.setPointerCapture(e.pointerId)
+
+    const onMove = (ev: PointerEvent) => {
+      if (dragIndex.value < 0) return
+      const [x, y] = svgCoords(ev)
```
christian-byrne, line 106: We should consider precomputing and reusing SVG coordinate transform during active drags. `svgCoords()` currently does:

- `svg.getScreenCTM()`
- `ctm.inverse()`
- `new DOMPoint(...).matrixTransform(...)`

That’s relatively expensive if called every pointermove.

**Possible Solution:** At `startDrag`, compute `const inv = svg.getScreenCTM()?.inverse()` once and reuse inside `onMove`.
- Recompute only on `lostpointercapture` / endDrag or if you detect resize/zoom.

Example approach:
- In `startDrag`, create `let inv = svg.getScreenCTM()?.inverse()`
- In `onMove`, use `inv` if present
- Optionally refresh `inv` at most once per animation frame (see next item)

---

@coderabbitai make a follow Issue for this optimization.

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+      modelValue.value = sortByX(modelValue.value)
+      svg.removeEventListener('pointermove', onMove)
+      svg.removeEventListener('pointerup', endDrag)
+      svg.removeEventListener('lostpointercapture', endDrag)
+      cleanupDrag = null
+    }
+
+    cleanupDrag = endDrag
+
+    svg.addEventListener('pointermove', onMove)
```
christian-byrne, line 126: Suggestion: Throttle pointermove updates to animation frames (RAF)
Right now every `pointermove` clones the array and updates `modelValue`, which can trigger Vue recomputes *many times per frame* on high-polling devices.

**Fix:** Use `requestAnimationFrame` batching:
- Store latest pointer position in locals
- If no RAF pending, schedule one
- In RAF callback: apply update once (clone array once)

This often provides the biggest UI perf win in drag interactions. @coderabbitai make a followup Issue to implement this

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    const interpolate = createMonotoneInterpolator(points)
+    const xMin = points[0][0]
+    const xMax = points[points.length - 1][0]
+    const segments = 128
+    const parts: string[] = []
+    for (let i = 0; i <= segments; i++) {
+      const x = xMin + (xMax - xMin) * (i / segments)
+      const y = 1 - interpolate(x)
+      parts.push(`${i === 0 ? 'M' : 'L'}${x.toFixed(4)},${y.toFixed(4)}`)
```
christian-byrne, line 28: Using `toFixed` like this is somewhat suboptimal as with other comments. We can use raw numbers of SVG path can tolerate more precision or pre-round. Or just only round `x` once since `x` steps are uniform (same suggestion as before). @coderabbitai make a followup Issue for this.

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+  })
+
+  const curvePath = computed(() => {
+    const points = sortedPoints.value
+    if (points.length < 2) return ''
+
+    const interpolate = createMonotoneInterpolator(points)
+    const xMin = points[0][0]
+    const xMax = points[points.length - 1][0]
+    const segments = 128
```
christian-byrne, line 23: Suggestion: precompute uniform x positions for the path
`segments=128` and the x positions are a linear ramp.

**Fix:** Precompute an array of normalized t values `[0..1]` or x positions once:
- `const ts = new Float64Array(segments+1)` built once
- In loop: `x = xMin + (xMax - xMin) * ts[i]`

This is small, but it reduces per-recompute arithmetic and makes the loop simpler for the JIT.

@coderabbitai make a followup Issue for this.

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+    ]
+  }
+
+  function findNearestPoint(x: number, y: number): number {
+    const threshold = 0.04
+    let nearest = -1
+    let minDist = threshold
+    for (let i = 0; i < modelValue.value.length; i++) {
+      const [px, py] = modelValue.value[i]
+      const dist = Math.hypot(px - x, py - y)
```
christian-byrne: nit: `Math.hypot` does a sqrt; you can compare squared distances instead.

Replace:
- `dist = Math.hypot(dx, dy)`
with:
- `dist2 = dx*dx + dy*dy`
and compare to `threshold*threshold`.

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+    const newIndex = newPoints.length - 1
+    startDrag(newIndex, e)
+  }
+
+  function sortByX(points: CurvePoint[]): CurvePoint[] {
+    return [...points].sort((a, b) => a[0] - b[0])
+  }
+
+  function startDrag(index: number, e: PointerEvent) {
+    if (e.button === 2 || (e.button === 0 && e.ctrlKey)) {
```
christian-byrne, line 89: nit: In startDrag, you delete when e.button === 2 || (e.button === 0 && e.ctrlKey). But in handleSvgPointerDown, you early-return on e.ctrlKey only in the "add new point" path, not in the "select existing point" path. This means ctrl+click on an existing point deletes, but ctrl+click on empty space does nothing. That might be intended, but it’s not obvious. Consider explicitly documenting it or using a clearer UX gesture (e.g., Alt-click to delete).

Also, do we need to use preventDefault to prevent browser right click default context menu?

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+  modelValue: Ref<CurvePoint[]>
+}
+
+export function useCurveEditor({ svgRef, modelValue }: UseCurveEditorOptions) {
+  const dragIndex = ref(-1)
+  let cleanupDrag: (() => void) | null = null
+
+  const sortedPoints = computed(() => {
+    const points = modelValue.value
+    if (!Array.isArray(points)) return []
```
christian-byrne: We declared that `modelValue: Ref<CurvePoint[]>`, so this check should not be necessary. If it's actually possible for modelValue to be undefined or null, then we should jsut update the types.

src/composables/useCurveEditor.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    const newIndex = newPoints.length - 1
+    startDrag(newIndex, e)
+  }
+
+  function sortByX(points: CurvePoint[]): CurvePoint[] {
+    return [...points].sort((a, b) => a[0] - b[0])
+  }
+
+  function startDrag(index: number, e: PointerEvent) {
```
christian-byrne, line 86: Should we be calling cleanupDrag?.() here to ensure idempotency?

src/renderer/extensions/vueNodes/widgets/composables/useCurveWidget.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const rawWidget = node.addWidget(
+      'curve',
+      spec.name,
+      [...defaultValue],
+      () => {},
+      {
+        serialize: true,
+        canvasOnly: false
+      }
```
christian-byrne: nit: these are both the default options right? So this options object can be removed?

src/components/curve/curveUtils.ts
christian-byrne: Since `x = i / 255` is strictly increasing, we can walk the curve segments linearly instead of performing a binary search for every sample.

```ts
export function curvesToLUT(points: CurvePoint[]): Uint8Array {
  const lut = new Uint8Array(256)
  const { xs, ys, slopes } = precomputeInterpolator(points)

  const n = xs.length
  let lo = 0
  let hi = 1

  for (let i = 0; i < 256; i++) {
    const x = i / 255

    while (hi < n - 1 && x > xs[hi]) {
      lo++
      hi++
    }

    const dx = xs[hi] - xs[lo]
    if (dx === 0) {
      lut[i] = Math.round(ys[lo] * 255)
      continue
    }

    const t = (x - xs[lo]) / dx
    const t2 = t * t
    const t3 = t2 * t

    const h00 = 2 * t3 - 3 * t2 + 1
    const h10 = t3 - 2 * t2 + t
    const h01 = -2 * t3 + 3 * t2
    const h11 = t3 - t2

    const y =
      h00 * ys[lo] +
      h10 * dx * slopes[lo] +
      h01 * ys[hi] +
      h11 * dx * slopes[hi]

    lut[i] = Math.max(0, Math.min(255, Math.round(y * 255)))
  }

  return lut
}
```

@coderabbitai make a followup Issue to replace the binary search With monotonic segment walk

src/composables/useCurveEditor.ts
jtydhr88: Acoording to adobe photoshop, ctrl+click is how to remove point from curve https://helpx.adobe.com/photoshop/using/curves-adjustment.html

Right-click context menu is already suppressed via @contextmenu.prevent.stop on the SVG element in CurveEditor.vue (line 8).

src/composables/useCurveEditor.ts
christian-byrne: Suggestion: precompute uniform x positions for the path
`segments=128` and the x positions are a linear ramp.

**Fix:** Precompute an array of normalized t values `[0..1]` or x positions once:
- `const ts = new Float64Array(segments+1)` built once
- In loop: `x = xMin + (xMax - xMin) * ts[i]`

This is small, but it reduces per-recompute arithmetic and makes the loop simpler for the JIT.

@coderabbitai make a followup Issue for this.

src/lib/litegraph/src/types/widgets.ts
christian-byrne: @coderabbitai Make an issue to do this in a followup PR.

christian-byrne (approved):
LGTM!

Issue #6523, opened 2025-11-01
Panning stops over text widgets on trackpad


benceruleanlu:
### Prerequisites

- [x] I am running the latest version of ComfyUI
- [x] I have searched existing issues to make sure this isn't a duplicate
- [x] I have tested with all custom nodes disabled ([see how](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled))

### What happened?

When you pan around in standard navigation, the string widgets (text input and note node) eats the pan

### Steps to Reproduce

Pan with standard navigation on laptops

### How is this affecting you?

Feature doesn't work as expected

### ComfyUI Frontend Version

1.32.0

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-6523-Panning-stops-over-widgets-on-trackpad-29e6d73d36508151886ccdfbf5344e76) by [Unito](https://www.unito.io)

PR #8863, opened 2026-02-14
fix: restore mouse-wheel scrolling in preview-as-text outputs

christian-byrne:
## Summary

Restore mouse-wheel scrolling for read-only preview widgets (PreviewAny plaintext and markdown modes), broken by the focus-gated wheel capture in #6597.

## Changes

- **What**: Remove `disabled` attribute from read-only textareas (keep `readonly`) so they can receive focus and capture wheel events. Add `data-capture-wheel` and `tabindex` to WidgetMarkdown display div.
- **Root cause**: `disabled` elements cannot receive focus in browsers. The focus-gated `wheelCapturedByFocusedElement()` from #6597 always evaluated to false for disabled textareas, forwarding all wheel events to the canvas.

## Review Focus

- Verify that removing `disabled` while keeping `readonly` does not allow unintended editing
- Confirm `tabindex="0"` on the markdown display div does not cause unexpected tab-order issues
- Ensure trackpad panning over unfocused widgets (#6523) still works correctly

Fixes COM-14812

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8863-fix-restore-mouse-wheel-scrolling-in-preview-as-text-outputs-3076d73d365081719bf5e453235bb2b5) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
@@ -16,7 +16,6 @@
       :class="cn(WidgetInputBaseClass, 'size-full text-xs resize-none')"
       :placeholder
       :readonly="isReadOnly"
-      :disabled="isReadOnly"
       fluid
       data-capture-wheel="true"
       @pointerdown.capture.stop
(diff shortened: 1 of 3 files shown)


src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -16,7 +16,6 @@
       :class="cn(WidgetInputBaseClass, 'size-full text-xs resize-none')"
       :placeholder
       :readonly="isReadOnly"
-      :disabled="isReadOnly"
```
DrJKL, line 19: I think this was for display purposes

DrJKL (approved):

PR #8864, opened 2026-02-14
feat: add Free subscription tier support

huntcsg:
## Summary

Add frontend support for a Free subscription tier — login/signup page restructuring, telemetry instrumentation, and tier-aware billing gating.

## Changes

- **What**: 
  - Restructure login/signup pages: OAuth buttons promoted as primary sign-in method, email login available via progressive disclosure
  - Add Free tier badge on Google sign-up button with dynamic credit count from remote config
  - Add `FREE` subscription tier to type system (tier pricing, tier rank, registry types)
  - Add `isFreeTier` computed to `useSubscription()`
  - Disable credit top-up for Free tier users (dialogService, purchaseCredits, popover CTA)
  - Show subscription/upgrade dialog instead of top-up dialog when Free tier user hits out-of-credits
  - Add funnel telemetry: `trackLoginOpened`, enrich `trackSignupOpened` with `free_tier_badge_shown`, track email toggle clicks

## Review Focus

- Tier gating logic: Free tier users should see "Upgrade" instead of "Add Credits" and never reach the top-up flow
- Telemetry event design for Mixpanel funnel analysis
- Progressive disclosure UX on login/signup pages

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8864-feat-add-Free-subscription-tier-support-3076d73d36508133b84ec5f0a67ccb03) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/cloud/onboarding/CloudLoginView.vue
+++ b/src/platform/cloud/onboarding/CloudLoginView.vue
@@ -2,7 +2,7 @@
   <div class="flex h-full items-center justify-center p-8">
     <div class="max-w-screen p-2 lg:w-96">
       <!-- Header -->
-      <div class="mt-6 mb-8 flex flex-col gap-4">
+      <div class="mb-8 flex flex-col gap-4">
         <h1 class="my-0 text-xl leading-normal font-medium">
           {{ t('auth.login.title') }}
         </h1>
@@ -23,7 +23,7 @@
       <template v-if="!showEmailForm">
         <p
           v-if="showFreeTierBadge"
-          class="mb-4 text-center text-xs text-muted-foreground"
+          class="mb-4 text-center text-sm text-muted-foreground"
         >
           {{
             freeTierCredits
--- a/src/platform/cloud/onboarding/CloudLoginView.vue
+++ b/src/platform/cloud/onboarding/CloudLoginView.vue
@@ -138,7 +138,7 @@ const {
   showFreeTierBadge,
   switchToEmailForm,
   switchToSocialLogin
-} = useFreeTierOnboarding('login')
+} = useFreeTierOnboarding()
 
 const navigateToSignup = async () => {
   await router.push({ name: 'cloud-signup', query: route.query })
--- a/src/platform/cloud/onboarding/CloudLoginView.vue
+++ b/src/platform/cloud/onboarding/CloudLoginView.vue
@@ -183,7 +183,9 @@ const signInWithEmail = async (values: SignInData) => {
 
 onMounted(() => {
   if (isCloud) {
-    telemetry?.trackLoginOpened({ free_tier_badge_shown: showFreeTierBadge })
+    telemetry?.trackLoginOpened({
+      free_tier_badge_shown: showFreeTierBadge.value
+    })
   }
 })
 </script>
--- a/src/platform/cloud/onboarding/CloudLoginView.vue
+++ b/src/platform/cloud/onboarding/CloudLoginView.vue
@@ -61,29 +61,27 @@
         </div>
 
         <div class="mt-6 text-center">
-          <button
-            class="cursor-pointer border-none bg-transparent text-sm text-muted-foreground underline hover:text-base-foreground"
-            @click="onShowEmailForm"
+          <Button
+            variant="muted-textonly"
+            class="text-sm underline"
+            @click="switchToEmailForm"
           >
             {{ t('auth.login.useEmailInstead') }}
-          </button>
+          </Button>
         </div>
       </template>
 
       <template v-else>
-        <Message severity="warn" class="mb-4">
-          {{ t('auth.signup.emailNotEligibleForFreeTier') }}
-        </Message>
-
         <CloudSignInForm :auth-error="authError" @submit="signInWithEmail" />
 
         <div class="mt-4 text-center">
-          <button
-            class="cursor-pointer border-none bg-transparent text-sm text-muted-foreground underline hover:text-base-foreground"
-            @click="onBackToSocialLogin"
+          <Button
+            variant="muted-textonly"
+            class="text-sm underline"
+            @click="switchToSocialLogin"
           >
             {{ t('auth.login.backToSocialLogin') }}
-          </button>
+          </Button>
         </div>
       </template>
 
@@ -112,16 +110,16 @@
 
 <script setup lang="ts">
 import Message from 'primevue/message'
-import { computed, onMounted, ref } from 'vue'
+import { onMounted, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { useRoute, useRouter } from 'vue-router'
 
 import Button from '@/components/ui/button/Button.vue'
 import { useFirebaseAuthActions } from '@/composables/auth/useFirebaseAuthActions'
 import CloudSignInForm from '@/platform/cloud/onboarding/components/CloudSignInForm.vue'
+import { useFreeTierOnboarding } from '@/platform/cloud/onboarding/composables/useFreeTierOnboarding'
 import { getSafePreviousFullPath } from '@/platform/cloud/onboarding/utils/previousFullPath'
 import { isCloud } from '@/platform/distribution/types'
-import { remoteConfig } from '@/platform/remoteConfig/remoteConfig'
 import { useTelemetry } from '@/platform/telemetry'
 import { useToastStore } from '@/platform/updates/common/toastStore'
 import type { SignInData } from '@/schemas/signInSchema'
@@ -132,21 +130,15 @@ const route = useRoute()
 const authActions = useFirebaseAuthActions()
 const isSecureContext = globalThis.isSecureContext
 const authError = ref('')
-const showEmailForm = ref(false)
-const freeTierCredits = computed(() => remoteConfig.value.free_tier_credits)
-const showFreeTierBadge = !localStorage.getItem('comfy:hasAccount')
 const toastStore = useToastStore()
 const telemetry = useTelemetry()
-
-const onShowEmailForm = () => {
-  showEmailForm.value = true
-  telemetry?.trackUiButtonClicked({ button_id: 'login_use_email_instead' })
-}
-
-const onBackToSocialLogin = () => {
-  showEmailForm.value = false
-  telemetry?.trackUiButtonClicked({ button_id: 'login_back_to_social_login' })
-}
+const {
+  showEmailForm,
+  freeTierCredits,
+  showFreeTierBadge,
+  switchToEmailForm,
+  switchToSocialLogin
+} = useFreeTierOnboarding('login')
 
 const navigateToSignup = async () => {
   await router.push({ name: 'cloud-signup', query: route.query })
--- a/src/platform/cloud/onboarding/CloudLoginView.vue
+++ b/src/platform/cloud/onboarding/CloudLoginView.vue
@@ -20,36 +20,72 @@
         {{ t('auth.login.insecureContextWarning') }}
       </Message>
 
-      <!-- Form -->
-      <CloudSignInForm :auth-error="authError" @submit="signInWithEmail" />
-
-      <!-- Divider -->
-      <Divider align="center" layout="horizontal" class="my-8">
-        <span class="text-muted">{{ t('auth.login.orContinueWith') }}</span>
-      </Divider>
-
-      <!-- Social Login Buttons -->
-      <div class="flex flex-col gap-6">
-        <Button
-          type="button"
-          class="h-10 bg-[#2d2e32]"
-          variant="secondary"
-          @click="signInWithGoogle"
+      <template v-if="!showEmailForm">
+        <p
+          v-if="showFreeTierBadge"
+          class="mb-4 text-center text-xs text-muted-foreground"
         >
-          <i class="pi pi-google mr-2"></i>
-          {{ t('auth.login.loginWithGoogle') }}
-        </Button>
+          {{
+            freeTierCredits
+              ? t('auth.login.freeTierDescription', {
+                  credits: freeTierCredits
+                })
+              : t('auth.login.freeTierDescriptionGeneric')
+          }}
+        </p>
 
-        <Button
-          type="button"
-          class="h-10 bg-[#2d2e32]"
-          variant="secondary"
-          @click="signInWithGithub"
-        >
-          <i class="pi pi-github mr-2"></i>
-          {{ t('auth.login.loginWithGithub') }}
-        </Button>
-      </div>
+        <!-- OAuth Buttons (primary) -->
+        <div class="flex flex-col gap-4">
+          <div class="relative">
+            <Button type="button" class="h-10 w-full" @click="signInWithGoogle">
+              <i class="pi pi-google mr-2"></i>
+              {{ t('auth.login.loginWithGoogle') }}
+            </Button>
+            <span
+              v-if="showFreeTierBadge"
+              class="absolute -top-2.5 -right-2.5 rounded-full bg-yellow-400 px-2 py-0.5 text-[10px] font-bold whitespace-nowrap text-gray-900"
+            >
+              {{ t('auth.login.freeTierBadge') }}
+            </span>
+          </div>
+
+          <Button
+            type="button"
+            class="h-10 bg-[#2d2e32]"
+            variant="secondary"
+            @click="signInWithGithub"
+          >
+            <i class="pi pi-github mr-2"></i>
+            {{ t('auth.login.loginWithGithub') }}
+          </Button>
+        </div>
+
+        <div class="mt-6 text-center">
+          <button
+            class="cursor-pointer border-none bg-transparent text-sm text-muted-foreground underline hover:text-base-foreground"
+            @click="onShowEmailForm"
+          >
+            {{ t('auth.login.useEmailInstead') }}
+          </button>
+        </div>
+      </template>
+
+      <template v-else>
+        <Message severity="warn" class="mb-4">
+          {{ t('auth.signup.emailNotEligibleForFreeTier') }}
+        </Message>
+
+        <CloudSignInForm :auth-error="authError" @submit="signInWithEmail" />
+
+        <div class="mt-4 text-center">
+          <button
+            class="cursor-pointer border-none bg-transparent text-sm text-muted-foreground underline hover:text-base-foreground"
+            @click="onBackToSocialLogin"
+          >
+            {{ t('auth.login.backToSocialLogin') }}
+          </button>
+        </div>
+      </template>
 
       <!-- Terms & Contact -->
       <p class="mt-5 text-sm text-gray-600">
@@ -75,9 +111,8 @@
 </template>
 
 <script setup lang="ts">
-import Divider from 'primevue/divider'
 import Message from 'primevue/message'
-import { ref } from 'vue'
+import { computed, onMounted, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { useRoute, useRouter } from 'vue-router'
 
@@ -85,6 +120,9 @@ import Button from '@/components/ui/button/Button.vue'
 import { useFirebaseAuthActions } from '@/composables/auth/useFirebaseAuthActions'
 import CloudSignInForm from '@/platform/cloud/onboarding/components/CloudSignInForm.vue'
 import { getSafePreviousFullPath } from '@/platform/cloud/onboarding/utils/previousFullPath'
+import { isCloud } from '@/platform/distribution/types'
+import { remoteConfig } from '@/platform/remoteConfig/remoteConfig'
+import { useTelemetry } from '@/platform/telemetry'
 import { useToastStore } from '@/platform/updates/common/toastStore'
 import type { SignInData } from '@/schemas/signInSchema'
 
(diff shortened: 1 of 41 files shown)


src/platform/cloud/onboarding/composables/useFreeTierOnboarding.ts
```diff
@@ -0,0 +1,41 @@
...
+      return false
+    }
+  })()
+  const telemetry = useTelemetry()
+
+  function switchToEmailForm() {
+    showEmailForm.value = true
+    telemetry?.trackUiButtonClicked({
+      button_id: `${source}_use_email_instead`
+    })
```
christian-byrne: FYI this telemetry event is currently disabled due to it making us exceed our Mixpanel quota heavily. It can be re-enabled in the dynamic config by name though.

src/platform/cloud/subscription/constants/tierPricing.ts
```diff
@@ -49,12 +56,17 @@ const FOUNDER_MONTHLY_PRICE = 20
...
   const pricing = TIER_PRICING[tierKey]
   return isYearly ? pricing.yearly : pricing.monthly
 }
 
-export function getTierCredits(tierKey: TierKey): number {
+export function getTierCredits(tierKey: 'free'): number | null
+export function getTierCredits(tierKey: Exclude<TierKey, 'free'>): number
+export function getTierCredits(tierKey: TierKey): number | null
+export function getTierCredits(tierKey: TierKey): number | null {
+  if (tierKey === 'free') return remoteConfig.value.free_tier_credits ?? null
```
christian-byrne: https://ampcode.com/threads/T-019c829c-6d9a-7470-9299-3b9fda2d099f

> Verdict: There IS one callsite where TS narrows to 'free' — tierBenefits.ts line 26, inside if (isFree). TypeScript 4.4+ narrows through the aliased const boolean, so the getTierCredits(tierKey: 'free'): number | null overload is being selected there, and the credits !== null check on line 27 is correctly informed by that return type.
> 
> However, this is arguably still over-engineered — even without the overload, the base signature (tierKey: TierKey): number | null already returns number | null, so the null check would still be required. The overload only adds value for the Exclude<TierKey, 'free'> → number case, which lets callers skip the null check. But looking at the two SubscriptionPanelContent* callsites (lines 547/267) that pass a general TierKey, they already do a credits === null check anyway, so they get no benefit either.
> 
> The overloads are effectively dead complexity. The only overload doing real work would be Exclude<TierKey, 'free'> → number, and the callers that could benefit from it (TransitionPreview, AddPaymentPreview) already narrow the type themselves via as casts or prop types that exclude 'free'. They'd get number from the base signature too since TS doesn't use overloads to infer from argument types — it matches top-down. So yes, the overloads could be safely removed and replaced with the single (tierKey: TierKey): number | null signature.

christian-byrne (commented):
Nit: The current structure for the 3-way dialog routing feels a bit awkward — `show()` was the original entry point, but now the PR renames it to `showPricingTable`, creates a new `show` that conditionally calls it, and then leaks `showPricingTable` to callers who re-implement the free-tier branching logic themselves (e.g. line 465 in `SubscriptionPanelContentWorkspace.vue`: `isFreeTierPlan.value ? showPricingTable() : showSubscriptionDialog()`). This defeats the purpose of the composable owning the routing.

Suggestion — single `show()` with a flag:

```typescript
function show(options?: {
  reason?: SubscriptionDialogReason
  skipFreeTierDialog?: boolean  // for "Upgrade" button in panel
}) {
  if (isFreeTier.value && !options?.skipFreeTierDialog) {
    showFreeTierDialog(options)
    return
  }
  showPricingTable(options)  // private, not exported
}

return { show, hide }  // no showPricingTable leak
```

Then `SubscriptionPanelContentWorkspace.vue` line 465 simplifies to `show({ skipFreeTierDialog: true })`, and all routing logic stays inside the composable.

src/platform/cloud/onboarding/CloudLoginView.vue
```diff
@@ -20,36 +20,70 @@
...
+              v-if="showFreeTierBadge"
+              class="absolute -top-2.5 -right-2.5 rounded-full bg-yellow-400 px-2 py-0.5 text-[10px] font-bold whitespace-nowrap text-gray-900"
+            >
+              {{ t('auth.login.freeTierBadge') }}
+            </span>
+          </div>
+
+          <Button
+            type="button"
+            class="h-10 bg-[#2d2e32]"
```
christian-byrne, line 49: nit: Check if we can use design system tokens for the color rather than hard-coding it. Totally optional as this code was pre-existing.

src/platform/cloud/onboarding/composables/useFreeTierOnboarding.ts
```diff
@@ -0,0 +1,41 @@
...
+  const showEmailForm = ref(false)
+  const freeTierCredits = computed(() => remoteConfig.value.free_tier_credits)
+  // Evaluated once at mount — localStorage won't change mid-session
+  const showFreeTierBadge = (() => {
+    try {
+      return !localStorage.getItem(HAS_ACCOUNT_KEY)
+    } catch {
+      return false
+    }
+  })()
```
christian-byrne: Consider something like:

```typescript
import { useLocalStorage } from '@vueuse/core'

export function useFreeTierOnboarding(source: 'login' | 'signup') {
  const hasAccount = useLocalStorage(HAS_ACCOUNT_KEY, null)
  const showFreeTierBadge = computed(() => !hasAccount.value)
  // ...
}
```

src/platform/cloud/onboarding/CloudSignupView.vue
```diff
@@ -22,42 +22,80 @@
...
+        <!-- OAuth Buttons (primary) -->
+        <div class="flex flex-col gap-4">
+          <div class="relative">
+            <Button type="button" class="h-10 w-full" @click="signInWithGoogle">
+              <i class="pi pi-google mr-2"></i>
+              {{ t('auth.signup.signUpWithGoogle') }}
+            </Button>
+            <span
+              v-if="showFreeTierBadge"
+              class="absolute -top-2.5 -right-2.5 rounded-full bg-yellow-400 px-2 py-0.5 text-[10px] font-bold whitespace-nowrap text-gray-900"
```
christian-byrne, line 45: nit: Try to find correct semantic design token rather than text-gray-900 color token

src/platform/cloud/onboarding/CloudSignupView.vue
```diff
@@ -68,30 +106,29 @@
...
-            class="cursor-pointer text-blue-400 no-underline"
-            target="_blank"
-            rel="noopener noreferrer"
-          >
-            {{ t('cloudWaitlist_contactLink') }}</a
-          >.
-        </p>
-      </div>
+      </p>
+      <p class="mt-2 text-sm text-gray-600">
```
christian-byrne, line 113: nit: don't think text-gray doesn't anything here

src/stores/firebaseAuthStore.ts
```diff
@@ -108,6 +111,13 @@ export const useFirebaseAuthStore = defineStore('firebaseAuth', () => {
   onAuthStateChanged(auth, (user) => {
     currentUser.value = user
     isInitialized.value = true
+    if (user) {
+      try {
+        localStorage.setItem(HAS_ACCOUNT_KEY, '1')
+      } catch {
+        // Ignore localStorage errors (e.g., in private browsing mode)
+      }
+    }
```
christian-byrne: Would also recommend useLocalStorage if it should be reactive otherwise using the setStorageValue helpers

christian-byrne (commented):
## PR Review: Free Tier Support

Overall this is a solid PR with clear separation of concerns. The free tier routing, telemetry additions, and pricing table updates are well-structured. A few issues worth addressing below, ranging from dead code to a potential bug in the legacy panel.

**Login/auth semantics**: Verified — error handling, `isInChina` gating, and redirect/routing logic are all preserved correctly. The `isInChina` check now only appears in the email-form branch, which is correct since the restriction only applies to email signup (OAuth was always available to China users even before this PR).

**Snake_case usage**: Confirmed consistent with existing telemetry conventions — all Mixpanel property names in this codebase use snake_case (`is_new_user`, `user_id`, `button_id`, etc). The new `free_tier_badge_shown`, `SubscriptionDialogReason` values, and `button_id` strings all follow the established pattern correctly.

**Telemetry**: The new `trackLoginOpened` follows the exact same dispatch pattern as the existing `trackSignupOpened`. Implementation is consistent.

src/platform/cloud/subscription/constants/tierPricing.ts
```diff
@@ -49,12 +56,17 @@ const FOUNDER_MONTHLY_PRICE = 20
...
   if (tierKey === 'founder') return FOUNDER_MONTHLY_PRICE
   const pricing = TIER_PRICING[tierKey]
   return isYearly ? pricing.yearly : pricing.monthly
 }
 
-export function getTierCredits(tierKey: TierKey): number {
+export function getTierCredits(tierKey: 'free'): number | null
+export function getTierCredits(tierKey: Exclude<TierKey, 'free'>): number
+export function getTierCredits(tierKey: TierKey): number | null
+export function getTierCredits(tierKey: TierKey): number | null {
```
christian-byrne, line 65: **Nit: `getTierCredits` overloads are dead complexity.** Audited every callsite:

- The `Exclude<TierKey, 'free'> → number` overload is never actually leveraged — callers that exclude `'free'` already narrow via `as` casts or prop types (e.g., `CheckoutTierKey = Exclude<TierKey, 'free' | 'founder'>`).
- The `'free' → number | null` overload is only hit in `tierBenefits.ts:26` inside an `if (isFree)` block, but even there the overload adds no value because the implementation signature already returns `number | null`.

Could safely be a single signature:
```typescript
export function getTierCredits(tierKey: TierKey): number | null {
```

src/platform/cloud/onboarding/composables/useFreeTierOnboarding.ts
christian-byrne: **Should use `useLocalStorage` from VueUse instead of raw `localStorage.getItem()`.** The codebase consistently uses `useLocalStorage` in ~10 other places (`apiKeyAuthStore`, `ComfyActionbar`, `NodeLibrarySidebarTab`, `TopMenuSection`, etc.).

Current approach has two issues:
1. **Not reactive** — if the user logs in during the session (which sets `HAS_ACCOUNT_KEY` in `firebaseAuthStore.ts`), `showFreeTierBadge` remains stale since it's evaluated once at mount via IIFE.
2. **Manual try/catch** — `useLocalStorage` handles this automatically.

Suggested:
```typescript
import { useLocalStorage } from '@vueuse/core'

const hasAccount = useLocalStorage(HAS_ACCOUNT_KEY, null)
const showFreeTierBadge = computed(() => !hasAccount.value)
```
This also makes it reactive if the user logs in during the same session.

src/platform/telemetry/types.ts
```diff
@@ -27,6 +27,14 @@ export interface AuthMetadata {
   utm_source?: string
   utm_medium?: string
   utm_campaign?: string
+  free_tier_badge_shown?: boolean
```
christian-byrne: **`free_tier_badge_shown` was added to `AuthMetadata` but no `trackAuth()` call ever passes it.** All four `trackAuth()` calls in `firebaseAuthStore.ts` only pass `method`, `is_new_user`, and `user_id`. This field is dead code in `AuthMetadata` — either wire it up in the auth tracking calls or remove it from this interface.

(The same field in `AuthPageOpenedMetadata` is correctly used by `trackSignupOpened` and `trackLoginOpened`.)

src/platform/cloud/onboarding/composables/useFreeTierOnboarding.ts
```diff
@@ -0,0 +1,41 @@
+import { computed, ref } from 'vue'
+
+import { remoteConfig } from '@/platform/remoteConfig/remoteConfig'
+import { useTelemetry } from '@/platform/telemetry'
+import { HAS_ACCOUNT_KEY } from '@/stores/firebaseAuthStore'
+
+export function useFreeTierOnboarding(source: 'login' | 'signup') {
```
christian-byrne: **Missing tests.** This new composable contains business logic (badge visibility based on localStorage, form toggling, telemetry tracking) but has no unit tests. Similarly, `FreeTierDialogContent.vue` is a new component with conditional rendering logic (reason-based title/subtitle switching) and no tests.

At minimum, `useFreeTierOnboarding` should have tests covering:
- `showFreeTierBadge` returns `true` when `HAS_ACCOUNT_KEY` is absent
- `showFreeTierBadge` returns `false` when `HAS_ACCOUNT_KEY` is set
- `switchToEmailForm` / `switchToSocialLogin` toggle `showEmailForm` and fire telemetry with correct `button_id`

src/platform/cloud/subscription/components/FreeTierDialogContent.vue
```diff
@@ -0,0 +1,122 @@
...
+}>()
+
+defineEmits<{
+  close: [subscribed: boolean]
+  upgrade: []
+}>()
+
+const { formattedRenewalDate } = useSubscription()
+
+const freeTierCredits = computed(() => remoteConfig.value.free_tier_credits)
```
christian-byrne: **Nit: `freeTierCredits` is computed independently in three places.**

1. Here: `computed(() => remoteConfig.value.free_tier_credits)`
2. `useFreeTierOnboarding.ts:10`: same expression
3. `getTierCredits('free')` in `tierPricing.ts:69`: reads `remoteConfig.value.free_tier_credits ?? null`

Consider using `getTierCredits('free')` here (it already exists for this purpose) to avoid the three-way duplication. Or expose it from `useSubscription` as a shared computed.

src/platform/cloud/subscription/components/SubscriptionPanelContentLegacy.vue
```diff
@@ -213,9 +213,10 @@ import {
   DEFAULT_TIER_KEY,
   TIER_TO_KEY,
   getTierCredits,
-  getTierFeatures,
   getTierPrice
 } from '@/platform/cloud/subscription/constants/tierPricing'
+import type { TierBenefit } from '@/platform/cloud/subscription/utils/tierBenefits'
+import { getCommonTierBenefits } from '@/platform/cloud/subscription/utils/tierBenefits'
```
christian-byrne, line 219: **Bug: Legacy panel is missing free-tier guards that the workspace variant has.**

The workspace variant (`SubscriptionPanelContentWorkspace.vue`) correctly hides `Manage Payment` and `More Options` buttons with `v-if="!isFreeTierPlan"`. But this legacy variant has no such guards on its `Manage Subscription` button (line 35) or `Upgrade Plan` button (line 48).

Since `isActiveSubscription` is `true` for free-tier users (they have an active FREE subscription), these buttons will render:
- **Manage Subscription** (line 35) → attempts `accessBillingPortal()` which will likely fail or show an empty portal for free-tier users.
- The `Upgrade Plan` button calls `show()` which correctly routes through the free-tier dialog, so that one is fine functionally, but could use a label change from "Upgrade" to "Subscribe" for free-tier users.

Suggestion: Add `isFreeTier` guards matching the workspace variant, or at minimum handle the billing portal call gracefully for free-tier users.

src/platform/cloud/subscription/components/SubscriptionBenefits.vue
```diff
@@ -13,6 +13,13 @@
         {{ $t('subscription.benefits.benefit2') }}
       </span>
     </div>
+
+    <div class="flex items-center gap-2 py-2">
+      <i class="pi pi-check text-xs text-text-primary" />
+      <span class="text-sm text-text-primary">
```
christian-byrne, line 23: **Minor: Benefit copy is now upsell-oriented, which is slightly off in one of its two usage contexts.**

This component is shared between:
1. `FreeTierDialogContent.vue` — upsell from free → paid (✅ "More monthly credits" makes sense)
2. `SubscriptionRequiredDialogContent.vue` — first-time subscription prompt (⚠️ "More" implies the user already has some credits, which may not be true for users without any subscription)

Low priority since the wording still broadly works, but worth noting if copy is being finalized.

src/scripts/app.ts
```diff
@@ -706,12 +705,9 @@ export class ComfyApp {
           'Payment Required: Please add credits to your account to use this node.'
         )
       ) {
-        const { isActiveSubscription } = useBillingContext()
-        if (isActiveSubscription.value) {
-          useDialogService().showTopUpCreditsDialog({
-            isInsufficientCredits: true
-          })
-        }
+        useDialogService().showTopUpCreditsDialog({
```
christian-byrne, line 743: **Behavioral change worth noting (not a bug):** The old code silently swallowed `Payment Required` errors for users without an active subscription — the `if (isActiveSubscription.value)` guard meant non-subscribers saw nothing when hitting a payment wall. The new code always calls `showTopUpCreditsDialog`, which now redirects free-tier and non-subscribed users to the subscription required dialog.

This is better UX since free-tier users who exhaust credits now see a clear upgrade path instead of a silent no-op. Just confirming this behavioral change is intentional.

src/platform/cloud/onboarding/composables/useFreeTierOnboarding.ts
huntcsg: I think we can gather the metrics other ways (just see increase in proportion of signups with Google), and enrich the user with their auth provider (if it doesn't already exist).

I'll drop the additional telemetry I think.

src/platform/cloud/onboarding/CloudLoginView.vue
huntcsg: We have on the todolist to audit and rework onboarding flow altogether, so I'll probably leave as is. I did do a pass on using the design system tokens where possible.

src/platform/cloud/subscription/components/SubscriptionPanelContentLegacy.vue
huntcsg: 🙏 
I'm testing on an account that's in workspace mode (all ephemeral envs)!

src/platform/cloud/subscription/constants/tierPricing.ts
huntcsg: > Done. Removed the 3 overload signatures, leaving a single getTierCredits(tierKey: TierKey): number | null. The 3 callsites that previously relied on the non-null overload now use ?? 0 — safe since they already narrow to paid tiers where null is unreachable.

src/platform/telemetry/types.ts
huntcsg: I am adding a commit that re-works the telemetry for this feature.

Basically we want to track:

* how seeing the badge affects signup
* free tier -> subscribe

but within free tier -> subscribe there's three flows:

* ran out of credits -> modal CTA -> subscribe
* attempted to add credits -> modal CTA -> subscribe
* was managing plan and clicked upgrade

We'll track those properly now.

src/scripts/app.ts
huntcsg: I'm running this by Alex, but this is intentional in the new Free Tier context.

src/platform/cloud/onboarding/CloudSignupView.vue
huntcsg: Preexisting. Ignoring.

src/platform/cloud/onboarding/CloudSignupView.vue
huntcsg: This is related to the badge which I think is NOT color/theme reactive. I think that's ok. (It seems most/all of cloud onboarding pages are not color/theme reactive.

src/platform/cloud/subscription/components/SubscriptionBenefits.vue
huntcsg: I'm changing this to be aware of whether is_free_tier, as I think the "More monthly credits" is important for free tier, but doesn't make sense if not free tier.

src/stores/firebaseAuthStore.ts
huntcsg: Will do!

src/platform/cloud/onboarding/composables/useFreeTierOnboarding.ts
```diff
@@ -0,0 +1,41 @@
...
+  // - HAS_ACCOUNT_KEY: set by onAuthStateChanged after first login
+  // - Comfy.PreviousWorkflow: set on every workflow save (covers existing
+  //   users before HAS_ACCOUNT_KEY was introduced)
+  // Reactive via useLocalStorage so the badge hides if the user signs in
+  // during the current session.
+  const hasAccount = useLocalStorage<string | null>(HAS_ACCOUNT_KEY, null)
+  const previousWorkflow = useLocalStorage<string | null>(
+    'Comfy.PreviousWorkflow',
+    null
+  )
```
christian-byrne: we have a newUserService for detecting whether this is the first ever session.

src/platform/cloud/subscription/components/SubscriptionBenefits.vue
huntcsg: See above.

christian-byrne (dismissed):
LGTM

christian-byrne (approved):

PR #8877, opened 2026-02-14
fix: prevent scroll event leak in TextPreviewWidget

### LINKED ISSUE #3990: [Bug]: Scroll event leak after scrolling to the top of a text widget
### Frontend Version

https://github.com/user-attachments/assets/130005f3-34cd-48d7-951d-69768ccec06e

### Expected Behavior

When scrolling the text (text being scrollable), the scroll event should never be sent into the canvas layer

If text is not scrollable (smallthe  amount of text), then scroll should work only in canvas

### Actual Behavior

Event current leaks after scrolling to the top or bottom of a text widget

### Steps to Reproduce

Add Preview Any, and add a bunch of text, then scroll

### Debug Logs

```powershell
None
```

### Browser Logs

None

### Setting JSON

None

_No response_

### Other Information

_No response_

omair445:
hey! noticed this issue was causing some annoying behavior when scrolling through text widgets.

## what was happening
when you'd scroll to the top or bottom of a text preview widget, the scroll event would leak through to the canvas layer and cause unwanted zooming/panning. pretty frustrating when you're just trying to read some text.

## the fix
added a wheel event handler that's a bit smarter about when to stop propagation:
- if the text is scrollable and you're scrolling within bounds → stops the event from reaching canvas
- if you're at the top/bottom and keep scrolling → lets it through so canvas zoom still works
- if there's not enough text to scroll → events pass through normally

tested it out and it feels much more natural now. you can scroll through your text without the canvas freaking out, but you can still zoom the canvas when you need to.

closes #3990

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8877-fix-prevent-scroll-event-leak-in-TextPreviewWidget-3076d73d36508143ae96d564006e12bc) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/widgets/TextPreviewWidget.vue
+++ b/src/components/graph/widgets/TextPreviewWidget.vue
@@ -1,6 +1,8 @@
 <template>
   <div
+    ref="containerRef"
     class="relative max-h-[200px] min-h-[28px] w-full overflow-y-auto rounded-lg px-4 py-2 text-xs"
+    @wheel="handleWheel"
   >
     <div class="flex items-center gap-2">
       <div class="flex flex-1 items-center gap-2 break-all">
@@ -13,7 +15,7 @@
 
 <script setup lang="ts">
 import Skeleton from 'primevue/skeleton'
-import { computed, onMounted, ref, watch } from 'vue'
+import { computed, onMounted, ref, useTemplateRef, watch } from 'vue'
 
 import type { NodeId } from '@/lib/litegraph/src/litegraph'
 import { useExecutionStore } from '@/stores/executionStore'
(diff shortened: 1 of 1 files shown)


omair445:
hey good points, both fixed:

- added a 1px tolerance for `isAtTop`/`isAtBottom` checks to handle fractional `scrollTop` values — didn't realize that could be an issue but makes total sense
- switched to a function declaration instead of the arrow function

pushed in f9ae7e3.

omair445:
Hey everyone, I am very very sorry I got stuck in work-related stuff and haven't been able to follow up on this. I am happy if anyone else can take care of this from here or if the maintainers want to push the final bits. Apologies again for the delay!

christian-byrne (commented):
Thanks for the contribution. For context, this was something we have gone back and forth on. We felt like the current behavior is generally what people wanted at the time -- but this was long ago and a decision made by previous maintainers. Would you be able to explain more about the current UI/UX and the desired UI/UX and perhaps some examples from other apps?

I'll tag the designers for their input on this PR as well. Have a great weekend!

LittleSound (approved):

LittleSound:
I tried to fix the lint errors

PR #8880, opened 2026-02-14
fix(queue): address follow-up review comments from #8740

benceruleanlu:
## Summary
Address follow-up review feedback from #8740 by consolidating completed-banner thumbnail fields and tightening queue count sanitization.

## Changes
- **What**:
  - Consolidated completed notification thumbnail data to `thumbnailUrls` only by removing legacy `thumbnailUrl` support from the queue banner notification type and renderer.
  - Updated queue notification banner stories to use `thumbnailUrls` consistently.
  - Simplified `sanitizeCount` to treat any non-positive or non-number value as the fallback count (`1`).

## Review Focus
Confirm the completed notification payload shape is now consistently `thumbnailUrls` and no consumer relies on `thumbnailUrl`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8880-fix-queue-address-follow-up-review-comments-from-8740-3076d73d365081719a70d860691c5f05) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/queue/QueueNotificationBanner.stories.ts
+++ b/src/components/queue/QueueNotificationBanner.stories.ts
@@ -52,7 +52,7 @@ export const Completed: Story = {
   args: args({
     type: 'completed',
     count: 1,
-    thumbnailUrl: thumbnail('4dabf7')
+    thumbnailUrls: [thumbnail('4dabf7')]
   })
 }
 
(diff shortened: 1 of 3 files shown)


viva-jinyi (approved):
Nice!!

PR #8886, opened 2026-02-15
fix: drag-and-drop screenshot creates LoadImage node instead of showing error

### LINKED ISSUE #7752: Why doesn't Comfyui support dragging and dropping images to import after upgrading to the latest version? Can only click to upload images?
Why doesn't Comfyui support dragging and dropping images to import after upgrading to the latest version? Can only click to upload images?        

---

*This issue is transferred from: https://github.com/comfyanonymous/ComfyUI/issues/11485*

Original issue was created by @xiatiantuo at 2025-12-24T04:17:32.000Z

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-7752-Why-doesn-t-Comfyui-support-dragging-and-dropping-images-to-import-after-upgrading-to--2d36d73d365081c0bff8e4bccbc3cfec) by [Unito](https://www.unito.io)

christian-byrne:
## Summary

Fix drag-and-drop of local screenshots onto the canvas failing to create a LoadImage node.

## Changes

- **What**: Replace `_.isEmpty(workflowData)` check in `handleFile` with a check for workflow-relevant keys (`workflow`, `prompt`, `parameters`, `templates`). PNG screenshots often contain non-workflow `tEXt` metadata (e.g. `Software`, `Creation Time`) which made `_.isEmpty()` return `false`, skipping the LoadImage fallback and showing an error instead.

## Review Focus

The root cause is that `getPngMetadata` extracts all `tEXt`/`iTXt` PNG chunks indiscriminately. Rather than filtering at the parser level (which could break extensions relying on arbitrary metadata), the fix checks for workflow-relevant keys before deciding whether to treat the file as a workflow or a plain image.

Fixes #7752

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8886-fix-drag-and-drop-screenshot-creates-LoadImage-node-instead-of-showing-error-3086d73d3650817d86c5f1386aa041c2) by [Unito](https://www.unito.io)


christian-byrne:
This brings the code/logic back to how it was prior to https://github.com/Comfy-Org/ComfyUI_frontend/commit/31e405abc8dd1adcc8246a5eedf594ca78abae79. It's checking for the workflow tag specifically in the data to determine if a workflow is inside rather than just a non empty set of tags.

DrJKL (dismissed):

PR #8896, opened 2026-02-15
fix: eliminate visual shaking when adjusting crop region

jtydhr88:
## Summary
Replace the dual-image approach (dimmed  + background-image in crop box) with a single  and a box-shadow overlay on the crop box. The previous approach required two independently positioned images to be pixel-perfectly aligned, which caused visible jitter from sub-pixel rounding differences, especially when zoomed in.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8896-fix-eliminate-visual-shaking-when-adjusting-crop-region-3086d73d365081309703d3ab0d4ad44d) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM

PR #8898, opened 2026-02-15
fix: resolve errors when converting ImageCrop node to subgraph

jtydhr88:
## Summary

- Pass callback directly to addWidget instead of null to eliminate 'addWidget without a callback or property assigned' warning
- Serialize object widget values as plain objects in LGraphNode.serialize to prevent DataCloneError when structuredClone encounters Vue reactive proxies

before
https://github.com/user-attachments/assets/af20bd84-3e0e-4eca-b095-eaf4d5bb6884

after

https://github.com/user-attachments/assets/5a17772e-04bc-4f3e-abec-78c540e0efa3

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8898-fix-resolve-errors-when-converting-ImageCrop-node-to-subgraph-3086d73d365081b2ae34db31225683ad) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphNode.ts
+++ b/src/lib/litegraph/src/LGraphNode.ts
@@ -965,8 +965,12 @@ export class LGraphNode
       o.widgets_values = []
       for (const [i, widget] of widgets.entries()) {
         if (widget.serialize === false) continue
-        // @ts-expect-error #595 No-null
-        o.widgets_values[i] = widget ? widget.value : null
+        const val = widget?.value
+        // Ensure object values are plain (not reactive proxies) for structuredClone compatibility.
+        o.widgets_values[i] =
+          val != null && typeof val === 'object'
+            ? JSON.parse(JSON.stringify(val))
+            : (val ?? null)
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -965,8 +965,12 @@ export class LGraphNode
       o.widgets_values = []
       for (const [i, widget] of widgets.entries()) {
         if (widget.serialize === false) continue
-        // @ts-expect-error #595 No-null
-        o.widgets_values[i] = widget ? widget.value : null
+        const val = widget?.value
+        // Ensure object values are plain (not reactive proxies) for structuredClone compatibility.
+        o.widgets_values[i] =
```
DrJKL, line 970: Could this use [toRaw](https://vuejs.org/api/reactivity-advanced.html#toraw)?

src/renderer/extensions/vueNodes/widgets/composables/useBoundingBoxWidget.ts
```diff
@@ -45,7 +45,15 @@ export const useBoundingBoxWidget = (): ComfyWidgetConstructorV2 => {
       widgetType,
       name,
       { ...defaultValue },
-      null,
+      () => {
+        for (let i = 0; i < fields.length; i++) {
```
DrJKL, line 49: I'm always a little wary when I see indices used for multiple arrays like this. If they're always meant to be in sync, they should usually be a single array with tuples or structured contents.

DrJKL (approved):

PR #8908, opened 2026-02-16
fix: update asset video previews for card and list

benceruleanlu:
## Summary
Render generated video previews in list items using a real video element (instead of an image element (this caused errors before)) and include a custom play button with dimming per [the designs](https://www.figma.com/design/LVilZgHGk5RwWOkVN6yCEK/Queue-Progress-Modal?node-id=3928-39270&m=dev).

## Changes
- **What**:
  - List item preview path now renders a `video` element when `isVideoPreview` is true.
  - Video list preview uses `preload="metadata"`, `muted`, `playsinline`, and `pointer-events-none` so row click behavior stays unchanged.
  - Kept the custom overlay/play affordance and increased overlay dimming from `bg-black/10` to `bg-black/15`.
  - Updated tests for `AssetsListItem`, `MediaVideoTop`, and `AssetsSidebarListView`.

## Review Focus
- Confirm list item click behavior still opens/selects asset (no inline playback interaction).
- Confirm video list previews now show actual video frame path instead of broken image fallback.

## Limitation
Backend does not currently provide a dedicated poster/thumbnail image for video outputs in the job preview payload. In the frontend today, we can either show a video icon placeholder, or load/render the full video itself to obtain a preview frame.

image
image

### DIFF
--- a/src/platform/assets/components/AssetsListItem.vue
+++ b/src/platform/assets/components/AssetsListItem.vue
@@ -35,12 +35,24 @@
         :icon-class="iconClass"
         :icon-aria-label="iconAriaLabel"
       >
-        <img
-          v-if="previewUrl"
-          :src="previewUrl"
-          :alt="previewAlt"
-          class="size-full object-cover"
-        />
+        <div v-if="previewUrl" class="relative size-full">
+          <template v-if="isVideoPreview">
+            <video
+              :src="previewUrl"
+              preload="metadata"
+              muted
+              playsinline
+              class="pointer-events-none size-full object-cover"
+            />
+            <VideoPlayOverlay size="sm" />
+          </template>
+          <img
+            v-else
+            :src="previewUrl"
+            :alt="previewAlt"
+            class="size-full object-cover"
+          />
+        </div>
         <div v-else class="flex size-full items-center justify-center">
           <i
             aria-hidden="true"
@@ -119,6 +131,8 @@ import { useProgressBarBackground } from '@/composables/useProgressBarBackground
 import Button from '@/components/ui/button/Button.vue'
 import { cn } from '@/utils/tailwindUtil'
 
+import VideoPlayOverlay from './VideoPlayOverlay.vue'
+
 const emit = defineEmits<{
   'stack-toggle': []
 }>()
@@ -130,6 +144,7 @@ const {
   iconAriaLabel,
   iconClass,
   iconWrapperClass,
+  isVideoPreview = false,
   primaryText,
   secondaryText,
   stackCount,
@@ -144,6 +159,7 @@ const {
   iconAriaLabel?: string
   iconClass?: string
   iconWrapperClass?: string
+  isVideoPreview?: boolean
   primaryText?: string
   secondaryText?: string
   stackCount?: number
--- a/src/platform/assets/components/MediaVideoTop.test.ts
+++ b/src/platform/assets/components/MediaVideoTop.test.ts
@@ -4,15 +4,20 @@ import { describe, expect, it, vi } from 'vitest'
 import type { AssetMeta } from '../schemas/mediaAssetSchema'
 import MediaVideoTop from './MediaVideoTop.vue'
 
-const createVideoAsset = (src: string): AssetMeta => ({
-  id: 'video-1',
-  name: 'clip.mp4',
-  asset_hash: null,
-  mime_type: 'video/mp4',
-  tags: [],
-  kind: 'video',
-  src
-})
+function createVideoAsset(
+  src: string,
+  mimeType: AssetMeta['mime_type'] = 'video/mp4'
+): AssetMeta {
+  return {
+    id: 'video-1',
+    name: 'clip.mp4',
+    asset_hash: null,
+    mime_type: mimeType,
+    tags: [],
+    kind: 'video',
+    src
+  }
+}
 
 describe('MediaVideoTop', () => {
   it('renders playable video with darkened paused overlay and play icon', () => {
@@ -29,8 +34,20 @@ describe('MediaVideoTop', () => {
     expect(wrapper.find('source').attributes('src')).toBe(
       'https://example.com/thumb.jpg'
     )
+    expect(wrapper.find('source').attributes('type')).toBe('video/mp4')
     expect(wrapper.find('.bg-black\\/15').exists()).toBe(true)
     expect(wrapper.find('.icon-\\[lucide--play\\]').exists()).toBe(true)
+  })
+
+  it('does not render source element when src is empty', () => {
+    const wrapper = mount(MediaVideoTop, {
+      props: {
+        asset: createVideoAsset('')
+      }
+    })
+
+    expect(wrapper.find('video').exists()).toBe(true)
+    expect(wrapper.find('source').exists()).toBe(false)
   })
 
   it('emits playback events and hides paused overlay while playing', async () => {
--- a/src/platform/assets/components/MediaVideoTop.test.ts
+++ b/src/platform/assets/components/MediaVideoTop.test.ts
@@ -1,0 +1,85 @@
+import { mount } from '@vue/test-utils'
+import { describe, expect, it, vi } from 'vitest'
+
+import type { AssetMeta } from '../schemas/mediaAssetSchema'
+import MediaVideoTop from './MediaVideoTop.vue'
+
+const createVideoAsset = (src: string): AssetMeta => ({
+  id: 'video-1',
+  name: 'clip.mp4',
+  asset_hash: null,
+  mime_type: 'video/mp4',
+  tags: [],
+  kind: 'video',
+  src
+})
+
+describe('MediaVideoTop', () => {
+  it('renders playable video with darkened paused overlay and play icon', () => {
+    const wrapper = mount(MediaVideoTop, {
+      props: {
+        asset: createVideoAsset('https://example.com/thumb.jpg')
+      }
+    })
+
+    const video = wrapper.find('video')
+    const videoElement = video.element as HTMLVideoElement
+    expect(video.exists()).toBe(true)
+    expect(videoElement.controls).toBe(false)
+    expect(wrapper.find('source').attributes('src')).toBe(
+      'https://example.com/thumb.jpg'
+    )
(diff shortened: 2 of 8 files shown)


benceruleanlu:
* **#8914** <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8914?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#8989](https://github.com/Comfy-Org/ComfyUI_frontend/pull/8989) <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8989?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#8990](https://github.com/Comfy-Org/ComfyUI_frontend/pull/8990) <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8990?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#8908** <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8908?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8908?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#8957** <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8957?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

src/platform/assets/components/VideoPlayOverlay.vue
```diff
@@ -0,0 +1,43 @@
...
+          buttonSizeClass
+        )
+      "
+    >
+      <i
+        aria-hidden="true"
+        :class="cn('icon-[lucide--play] text-white', iconSizeClass)"
+      />
+    </div>
+  </div>
```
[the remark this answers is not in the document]
benceruleanlu: K, fixed

src/platform/assets/components/MediaVideoTop.vue
```diff
@@ -10,7 +10,7 @@
       muted
       loop
       playsinline
-      :poster="asset.preview_url"
+      :poster="asset.preview_url || asset.src || ''"
```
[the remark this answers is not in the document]
benceruleanlu: Checked backend, yes its an issue for videos right now, no good poster value exists, so I'm removing it entirely.

src/platform/assets/components/MediaVideoTop.test.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+    })
+
+    expect(wrapper.find('video').exists()).toBe(true)
+    expect(wrapper.find('source').attributes('src')).toBe(
+      'https://example.com/thumb.jpg'
+    )
+    expect(wrapper.find('.bg-black\\/10').exists()).toBe(true)
+    expect(wrapper.find('.icon-\\[lucide--play\\]').exists()).toBe(true)
+  })
```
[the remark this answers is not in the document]
benceruleanlu: Checked backend, yes its an issue for videos right now, no good poster value exists, so I'm removing it entirely.

viva-jinyi:
Can you update the storybook file as well?

benceruleanlu:
> Can you update the storybook file as well?

I already updated the storybook file to keep compatibility. The existing storybooks should be sufficient since they already cover video in list and card form. Did you have anything specific in mind?

viva-jinyi (dismissed):
LGTM

benceruleanlu:
### Merge activity

* **Feb 21, 12:42 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8908).
* **Feb 21, 12:44 AM UTC**: [Graphite](https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8908) rebased this pull request as part of a merge.
* **Feb 21, 12:51 AM UTC**: [Graphite](https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8908) couldn't merge this PR because it was not satisfying all requirements (PR does not have required approvals).
* **Feb 21, 8:59 AM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8908).
* **Feb 21, 8:59 AM UTC**: @benceruleanlu merged this pull request with [Graphite](https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/8908).

benceruleanlu:
Going to merge on previous approval, it was just rebased

src/platform/assets/components/AssetsListItem.vue
```diff
@@ -35,12 +35,24 @@
...
-        />
+        <div v-if="previewUrl" class="relative size-full">
+          <template v-if="isVideoPreview">
+            <video
+              :src="previewUrl"
+              preload="metadata"
+              muted
+              playsinline
+              class="pointer-events-none size-full object-cover"
+            />
```
christian-byrne, line 46: A few suggestions for this video element — it's a 32×32 thumbnail inside a virtualized list:

- **Add `aria-label`** — the `` branch has `:alt="previewAlt"` but the video has no accessible name. Screen readers won't announce what this video represents. Something like `:aria-label="previewAlt"` keeps parity with the image path.
- **Add `disablePictureInPicture` and `disableRemotePlayback`** — prevents the browser from surfacing PiP badges or cast icons on a non-interactive thumbnail.

src/platform/assets/components/MediaVideoTop.vue
```diff
@@ -5,20 +5,21 @@
...
       muted
       loop
       playsinline
-      :poster="asset.preview_url"
       class="relative size-full object-contain transition-transform duration-300 group-hover:scale-105 group-data-[selected=true]:scale-105"
-      @click.stop
+      @click.stop="onVideoClick"
       @play="onVideoPlay"
       @pause="onVideoPause"
       @ended="onVideoEnded"
```
christian-byrne: This handler is unreachable — `loop` on line 12 causes the browser to seek back to the start on completion instead of firing the `ended` event (per the HTML spec, `ended` only fires when `loop` is false). You can remove the `@ended` binding and the `onVideoEnded` function.

src/platform/assets/components/MediaVideoTop.vue
```diff
@@ -5,20 +5,21 @@
...
     >
       <source :src="asset.src || ''" />
```
christian-byrne: Two things on this `<source>`:

1. **`|| ''` is harmful** — when `asset.src` is nullish, an empty-string `src` resolves against the current page URL, causing the browser to fetch the HTML document as video. That wastes a request and logs a console error. A `v-if="asset.src"` guard on the `<source>` (or on the whole `<video>`) would be safer.
2. **Add a `type` attribute** (e.g. `type="video/mp4"`) — lets the browser skip codec probing and decide immediately whether it can play the format, which speeds up playback start.

src/platform/assets/components/MediaVideoTop.vue
```diff
@@ -5,20 +5,21 @@
     @mouseleave="isHovered = false"
   >
     <video
+      ref="videoElement"
       :controls="shouldShowControls"
       preload="metadata"
       muted
       loop
       playsinline
```
christian-byrne, line 13: Two more attributes worth adding:

- **`disablePictureInPicture`** and **`disableRemotePlayback`** — this is an inline preview player, and browser-initiated PiP prompts or Chromecast/AirPlay icons would be unexpected here. Both are simple boolean attributes.
- **`:poster` was removed** — without it the video shows a blank/black frame until `preload="metadata"` fetches the first frame, which can be a visible flash on slower connections. If `asset.preview_url` provides a lightweight image, keeping it as `poster` gives an instant visual at zero extra cost while metadata loads.

benceruleanlu: There is no suitable poster url unfortunately, the backend doesn't send one.

benceruleanlu: About the pip/remote stuff, I don't know if it's really unexpected? I don't see a good reason to disable

benceruleanlu: I guess on average the user wouldn't want to, but... it is just a control, and I think browsers have a detection alg for this to not show on certain videos

src/platform/assets/components/MediaVideoTop.vue
benceruleanlu: SGTM, adding in a followup

src/platform/assets/components/MediaVideoTop.vue
benceruleanlu: SGTM, adding in a followup

src/platform/assets/components/AssetsListItem.vue
benceruleanlu: the img here is just the decorative play button, I don't think we need an aria label for it
the player already has controls disabled when paused, so it seems unnecessary for hardening

src/platform/assets/components/MediaVideoTop.vue
benceruleanlu: actually I can just do it here since I need another review anyway

src/platform/assets/components/MediaVideoTop.vue
benceruleanlu: actually I can just do it here since I need another review anyway

benceruleanlu: Resolved by deleting the on end code

src/platform/assets/components/MediaVideoTop.vue
benceruleanlu: added the v-if and type for assets.mime_type

christian-byrne (approved):
LGTM

PR #8909, opened 2026-02-16
feat: download missing model files automatically

teddav:
There's currently no way to download all model files automatically. You either need to click on the "Download" button (and then pick the correct directory), or "Copy URL" and download manually in the correct dir.
This PR adds the possibility to download all files directly.

- Adds bulk “Download all missing models” flow for workflow missing-model warnings.
- Adds per-file download state updates (running/completed/failed/skipped/canceled) streamed from backend to frontend.
- Adds per-file cancel support during active downloads.
- Keeps download status visible in the missing-model dialog while downloads are in progress.
- Adds backend endpoint support for canceling individual missing-model download tasks.

The linked PR in the backend: https://github.com/Comfy-Org/ComfyUI/pull/12487

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8909-download-missing-model-files-automatically-3096d73d3650812daa68ec2fe678ea45) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/MissingModelsWarning.vue
+++ b/src/components/dialog/content/MissingModelsWarning.vue
@@ -34,6 +34,17 @@
       </template>
     </i18n-t>
   </div>
+  <div class="mb-2 flex justify-end">
+    <Button
+      variant="secondary"
+      size="sm"
+      :loading="isBulkDownloading"
+      :disabled="downloadableModels.length === 0 || isBulkDownloading"
+      @click="downloadAllMissingModels"
+    >
+      {{ t('missingModelsDialog.downloadAll') }}
+    </Button>
+  </div>
   <ListBox :options="missingModels" class="comfy-missing-models">
     <template #option="{ option }">
       <Suspense v-if="isDesktop">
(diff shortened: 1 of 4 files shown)


christian-byrne (commented):
This is a good idea for a feature and from what I can see the frontend and backend code is well implemented. Naturally, we would need to merge the backend PR first in order to proceed with the frontend implemention here. On the backend in the past, we did attempt to implement native model downloading, but ran into issues with needing to support retries, timeouts, resuming, parallelization, proxies/mirrors, safety/security, and so on.

For a large feature like this, my recommendation is to reach out on Discord or email to one of the primary backend maintainers (comfyanonymous, Kosinkadink, guill, etc.) and ask how they feel about supporting this and seeing what they would want to see for this to be shippable.

I do seriously appreciate the high quality contribution - you can also find me on Discord (username cbyrne). Cheers!

src/components/dialog/content/MissingModelsWarning.vue
```diff
@@ -65,6 +102,10 @@ import NoResultsPlaceholder from '@/components/common/NoResultsPlaceholder.vue'
 import { isDesktop } from '@/platform/distribution/types'
 import { useSettingStore } from '@/platform/settings/settingStore'
 import { useSettingsDialog } from '@/platform/settings/composables/useSettingsDialog'
+import { useToastStore } from '@/platform/updates/common/toastStore'
+import { api } from '@/scripts/api'
+import type { MissingModelDownloadItem } from '@/scripts/api'
+import type { MissingModelDownloadWsMessage } from '@/schemas/apiSchema'
```
christian-byrne, line 108: `@/schemas/apiSchema` is deprecated on current `main` (flagged by the repo's lint sweep as `deprecated-apiSchema-import`). Would it be worth sourcing `MissingModelDownloadWsMessage` from wherever the newer schema module re-exports the WS message types instead, so this doesn't need a follow-up cleanup pass once the deprecated module is removed?

PR #8917, opened 2026-02-16
fix: show stop state for active instant run button

benceruleanlu:
Switch the Run (Instant) actionbar button into a stop-state while instant auto-queue is actively running, so users can explicitly stop that mode from the same control.

Figma context: https://www.figma.com/design/LVilZgHGk5RwWOkVN6yCEK/Queue-Progress-Modal?node-id=3381-6181&m=dev

https://github.com/user-attachments/assets/a4aca6ab-eb0c-41a2-9f05-3af7ecf2bedd

### DIFF
--- a/src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
+++ b/src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
@@ -36,11 +36,10 @@
 </template>
 
 <script setup lang="ts">
-import { until } from '@vueuse/core'
 import { storeToRefs } from 'pinia'
 import type { MenuItem } from 'primevue/menuitem'
 import SplitButton from 'primevue/splitbutton'
-import { computed, ref } from 'vue'
+import { computed } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 import Button from '@/components/ui/button/Button.vue'
@@ -56,7 +55,11 @@ import { graphHasMissingNodes } from '@/workbench/extensions/manager/utils/graph
 import BatchCountEdit from '../BatchCountEdit.vue'
 
 const workspaceStore = useWorkspaceStore()
-const { mode: queueMode, batchCount } = storeToRefs(useQueueSettingsStore())
+const {
+  mode: queueMode,
+  batchCount,
+  instantAutoQueueActive
+} = storeToRefs(useQueueSettingsStore())
 const { activeJobsCount } = storeToRefs(useQueueStore())
 
 const nodeDefStore = useNodeDefStore()
@@ -73,6 +76,7 @@ const queueModeMenuItemLookup = computed(() => {
       tooltip: t('menu.disabledTooltip'),
       command: () => {
         queueMode.value = 'disabled'
+        instantAutoQueueActive.value = false
       }
     },
     change: {
@@ -84,6 +88,7 @@ const queueModeMenuItemLookup = computed(() => {
           button_id: 'queue_mode_option_run_on_change_selected'
         })
         queueMode.value = 'change'
+        instantAutoQueueActive.value = false
       }
     }
   }
@@ -97,6 +102,7 @@ const queueModeMenuItemLookup = computed(() => {
           button_id: 'queue_mode_option_run_instant_selected'
         })
         queueMode.value = 'instant'
+        instantAutoQueueActive.value = false
       }
     }
   }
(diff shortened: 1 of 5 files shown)


src/locales/en/main.json
```diff
@@ -894,6 +894,8 @@
     "runWorkflowFront": "Run workflow (Queue at front)",
     "runWorkflowDisabled": "Workflow contains unsupported nodes (highlighted red). Remove these to run the workflow.",
     "run": "Run",
+    "stopRunInstant": "Stop Run (Instant)",
+    "stopRunInstantTooltip": "Stop running",
```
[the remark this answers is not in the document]
benceruleanlu: Design wanted this specifically.

christian-byrne (commented):
LGTM

comfydesigner:
I think it might make more sense for "Stop Run (Instant)" to revert back to "Run (Instant)" instead of "Run". That way a user can skip a click to start the Run (Instant) again

benceruleanlu:
> I think it might make more sense for "Stop Run (Instant)" to revert back to "Run (Instant)" instead of "Run". That way a user can skip a click to start the Run (Instant) again

Yeah this makes sense. I fixed that

benceruleanlu:
image
Snapshot consistently failing, no clue why, seems unrelated, will just regen snapshot, failed like 3-4 times in a row so I'm assuming its a snapshot issue

benceruleanlu:
3-4 times in a row not including retries in the same CI

christian-byrne:
@benceruleanlu that is also happening on main, can ignore for this PR IMO

src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
```diff
@@ -143,6 +174,25 @@ const queueButtonTooltip = computed(() => {
 
 const commandStore = useCommandStore()
 const queuePrompt = async (e: Event) => {
+  if (showStopInstantPresentation.value) {
+    isStoppingInstant.value = true
```
christian-byrne: Wouldn't it be possible to queue up like 50 jobs in normal mode => then switch to instant mode => then press queue again. So then the 5000 timeout won't be adequate?

benceruleanlu: By design, it would go back to the standard run, because under the hood, the "instant" button display relies on queuemode being set to instant. I can just refactor it to decouple state with button UI, it just increases scope of this PR while blocking the stop state addition

benceruleanlu: So I refactored it now to be instant on and instant paused explicitly, which matches standard expectations better and resolves your worry

src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
```diff
@@ -97,6 +102,7 @@ const queueModeMenuItemLookup = computed(() => {
           button_id: 'queue_mode_option_run_instant_selected'
         })
         queueMode.value = 'instant'
+        instantAutoQueueActive.value = false
```
christian-byrne: I don't understand the logic anymore. Why is instantAutoQueueActive getting set to false while the queueMode is being set to "instant"? I don't get how these two variables are distinct

benceruleanlu:
Good question. These two fields are intentionally separate:

- `mode` = the selected queue strategy (`disabled` / `change` / `instant`).
- `instantAutoQueueActive` = whether the instant requeue loop is currently armed/running.

So when the dropdown option sets `mode = 'instant'`, we also set `instantAutoQueueActive = false` to avoid the mode switch itself implicitly starting/continuing auto-requeue.

`instantAutoQueueActive` only flips to `true` when the user explicitly clicks the main Run button while in instant mode, and flips back to `false` on Stop (or mode switch). `autoQueueService` then requeues only when both are true (`mode === 'instant' && instantAutoQueueActive`).

benceruleanlu:
Please read this thread as well: https://github.com/Comfy-Org/ComfyUI_frontend/pull/8917#discussion_r2820690281

christian-byrne (commented):
**Suggestion: Collapse `mode` + `instantAutoQueueActive` into a single enum**

The two-variable design is confusing because `mode === 'instant'` doesn't mean instant is active — you need to also check `instantAutoQueueActive`. This creates implicit coupling and invalid state combos.

Instead, use one discriminated enum:

```ts
type QueueMode = 'disabled' | 'change' | 'instant-idle' | 'instant-running'
```

- Dropdown selects `'instant-idle'` (instant selected but not yet armed)
- Clicking Run transitions to `'instant-running'`  
- Clicking Stop transitions back to `'instant-idle'`
- `autoQueueService` gates requeue on `mode === 'instant-running'` only

Existing checks like `mode === 'instant'` become `mode.startsWith('instant')` or a helper `isInstantMode(mode)`.

This eliminates the confusing separation, makes invalid states unrepresentable, and the stop/start lifecycle is self-documenting from the enum value alone.

benceruleanlu:
> **Suggestion: Collapse `mode` + `instantAutoQueueActive` into a single enum**
> 
> The two-variable design is confusing because `mode === 'instant'` doesn't mean instant is active — you need to also check `instantAutoQueueActive`. This creates implicit coupling and invalid state combos.
> 
> Instead, use one discriminated enum:
> 
> ```ts
> type QueueMode = 'disabled' | 'change' | 'instant-idle' | 'instant-running'
> ```
> 
> * Dropdown selects `'instant-idle'` (instant selected but not yet armed)
> * Clicking Run transitions to `'instant-running'`
> * Clicking Stop transitions back to `'instant-idle'`
> * `autoQueueService` gates requeue on `mode === 'instant-running'` only
> 
> Existing checks like `mode === 'instant'` become `mode.startsWith('instant')` or a helper `isInstantMode(mode)`.
> 
> This eliminates the confusing separation, makes invalid states unrepresentable, and the stop/start lifecycle is self-documenting from the enum value alone.

OK, implemented

christian-byrne (approved):

PR #8938, opened 2026-02-17
fix: preserve PrimitiveNode control_after_generate value on copy-paste

### LINKED ISSUE #1757: [Bug]: Primitive node attached to seed input of Ksampler switching from "randomize" to "fixed" when copying nodes
### Frontend Version

v1.4.13

### Expected Behavior

Workflow where the seed widget of the Ksampler node has been converted to input and a primitive node is attached to input the seed.
The primitive node is set to "randomize" to generate a random seed each run.
Copying and pasting these nodes should retain the "randomize" state of the primitive node.

Screenshot1

### Actual Behavior

When pasted, the primitive node switches to "fixed"

Screenshot2

### Steps to Reproduce

Copy and paste a workflow or part of a workflow (including the primitive and Ksampler nodes) where the seed widget of the Ksampler node has been converted to input and a primitive node is attached to input a randomized seed. The state of the primitive node will change from randomize to fixed.

### Debug Logs

```powershell
-
```

### Browser Logs

-

### Setting JSON

{
    "Comfy.NodeLibrary.Bookmarks": [],
    "Comfy.WidgetControlMode": "before",
    "Comfy.UseNewMenu": "Disabled",
    "Comfy.PromptFilename": false,
    "Comfy.LinkRenderMode": 2
}

Mozilla Firefox

### Other

_No response_

dante01yoon:
## Summary

- PrimitiveNode's `control_after_generate` widget value (increment/decrement/randomize) was reset to "fixed" on copy-paste
- Persist the control value via node `properties` in `onSerialize` so it survives the clone/serialize/configure pipeline
- Read from `properties` first with fallback to `widgets_values` for backward compatibility with existing workflows

- Fixes #1757

## Bug cause

The `control_after_generate` widget is created by `addValueControlWidgets` with `{ serialize: false, canvasOnly: true }` in `options`. However, this only sets `widget.options.serialize`, not `widget.serialize` (the top-level property that litegraph's `serialize()` checks). As a result, the widget value IS included in `widgets_values` during normal serialization.

The problem occurs specifically during **copy-paste**:

1. `clone()` calls `serialize()` on the original node, then creates a **new PrimitiveNode** and calls `configure(data)` on it
2. The new PrimitiveNode has **no widgets** — widgets are created dynamically by `mergeIfValid` only when the node is connected
3. When `serialize()` is called on this widget-less clone, `widgets_values` becomes an **empty array** `[]`
4. On paste, `mergeIfValid` reads `this.widgets_values?.[1]` → `undefined` → defaults to `"fixed"`

## Fix

- **`onSerialize`**: Save the control widget value to `serializedNode.properties` under `__primitiveControlAfterGenerate`. Since `properties` are cloned before `onSerialize` is called, and `configure()` restores properties during paste, the value survives the full clone → serialize → configure pipeline.
- **`mergeIfValid`**: Read from `this.properties` first, falling back to `this.widgets_values?.[1]` for backward compatibility with existing saved workflows.

## Test plan

- [x] Unit tests for `primitiveControlValue` utility functions
- [x] E2E test: load workflow with PrimitiveNode set to "increment", copy-paste, verify pasted node retains "increment"
- [ ] Manual: create PrimitiveNode → connect to seed → change control to "increment" → copy-paste → verify value preserved


christian-byrne (commented):
I think the primary issue is that when a PrimitiveNode is copied, LGraphCanvas calls item.clone()?.serialize(). The clone() step serializes the original (which has widgets, so widgets_values is captured correctly), creates a fresh PrimitiveNode, and calls configure(data) on it (storing widgets_values on the instance). But PrimitiveNode creates its widgets dynamically only on connection, so the clone has no this.widgets.

When serialize() is then called on that clone, the base LGraphNode.serialize() checks if (widgets && this.serialize_widgets) and skips the entire widgets_values block because there are no widgets to iterate.

The seed value still survives because when the pasted PrimitiveNode reconnects to the pasted KSampler, it copies the KSampler's own widget value (widget.value = theirWidget.value). But control_after_generate is created separately by addValueControlWidgets, which reads from this.widgets_values?.[1] - the value that was lost.

I wonder if we can verify that this is what's happening and then see if we can simplify this PR by just overriding serialize() on PrimitiveNode to fall back to this.widgets_values (set during configure()) when the base implementation omits it.

PR #8943, opened 2026-02-17
fix: detect video output from data in Nodes 2.0

Myestery:
## Summary

- Fixes SaveWebM node showing "Error loading image" in Vue nodes mode
- Extracts `isAnimatedOutput`/`isVideoOutput` utility functions from inline logic in `unsafeUpdatePreviews` so both the litegraph canvas renderer and Vue nodes renderer can detect video output directly from execution data
- Uses output-based detection in `imagePreviewStore.isImageOutputs` to avoid applying image preview format conversion to video files

## Background

In Vue nodes mode, `nodeMedia` relied on `node.previewMediaType` to determine if output is video. This property is only set via `onDrawBackground` → `unsafeUpdatePreviews` in the litegraph canvas path, which doesn't run in Vue nodes mode. This caused webm output to render via `` instead of `<video>`.

## Before

https://github.com/user-attachments/assets/36f8a033-0021-4351-8f82-d19e3faa80c2

## After

https://github.com/user-attachments/assets/6558d261-d70e-4968-9637-6c24532e23ac
## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] `pnpm test:unit` passes (4500 tests)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8943-fix-detect-video-output-from-data-in-Vue-nodes-mode-30a6d73d365081e98e91d6d1dcc88785) by [Unito](https://www.unito.io)


christian-byrne (commented):
Looks good, do you want to add a browser test?

Myestery:
> Looks good, do you want to add a browser test?

I added a unit test rather, looks more appropriate here

christian-byrne:
> > Looks good, do you want to add a browser test?
> 
> I added a unit test rather, looks more appropriate here

I disagree, because this is behavior that spans across a ton of disconnected parts of the app and can easily break if any of those change.

christian-byrne (approved):
LGTM

PR #8945, opened 2026-02-18
fix: asset browser filters stick when navigating categories

Myestery:
## Summary

File format and base model filters in the asset browser persisted when navigating to categories that don't contain matching assets, showing empty results with no way to clear the filter.

## Changes

- **What**: Apply the same scope-aware filtering pattern from the template selector dialog. Selected filters that don't exist in the current category become inactive (excluded from filtering) but are preserved so they reactivate when navigating back. Uses writable computeds in `AssetFilterBar` (matching `WorkflowTemplateSelectorDialog`) and active filter intersection in `useAssetBrowser` (matching `useTemplateFiltering`).

## Before

https://github.com/user-attachments/assets/5c61e844-7ea0-489c-9c44-e0864dc916bc

## After

https://github.com/user-attachments/assets/8372e174-107c-41e2-b8cf-b7ef59fe741b

## Review Focus

The pattern mirrors `selectedModelObjects`/`selectedUseCaseObjects` in `WorkflowTemplateSelectorDialog.vue` and `activeModels`/`activeUseCases` in `useTemplateFiltering.ts`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8945-fix-asset-browser-filters-stick-when-navigating-categories-30b6d73d365081609ac5c3982a1a03fc) by [Unito](https://www.unito.io)


viva-jinyi (approved):
LGTM

PR #8951, opened 2026-02-18
feat: workflow sharing and ComfyHub publish flow

DrJKL:
## Summary

Add workflow sharing by URL and a multi-step ComfyHub publish wizard, gated by feature flags and an optional profile gate.

## Changes

- **What**: Share dialog with URL generation and asset warnings; ComfyHub publish wizard (Describe → Examples → Finish) with thumbnail upload and tags; profile gate flow; shared workflow URL loader with confirmation dialog
- **Dependencies**: None (new `sharing/` module under `src/platform/workflow/`)

## Review Focus

- Three new feature flags: `workflow_sharing_enabled`, `comfyhub_upload_enabled`, `comfyhub_profile_gate_enabled`
- Share service API contract and stale-share detection (`workflowShareService.ts`)
- Publish wizard and profile gate state management
- Shared workflow URL loading and query-param preservation

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8951-feat-share-workflow-by-URL-30b6d73d3650813ebbfafdad775bfb33) by [Unito](https://www.unito.io)


src/platform/workflow/sharing/constants/comfyHubTags.ts
```diff
@@ -0,0 +1,81 @@
...
+  { tag: 'Normal Map', count: 1 },
+  { tag: 'OpenPose', count: 1 },
+  { tag: 'Pose transfer', count: 1 },
+  { tag: 'Replacement', count: 1 },
+  { tag: 'Sound Effects', count: 1 },
+  { tag: 'Speech to Text', count: 1 },
+  { tag: 'Text Generation', count: 1 },
+  { tag: 'Turbo', count: 1 },
+  { tag: 'Video Extension', count: 1 },
+  { tag: 'Voice Isolation', count: 1 }
```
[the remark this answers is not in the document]
DrJKL: Tags are a schema concept and are not localized.

src/platform/workflow/sharing/constants/comfyHubTags.ts
```diff
@@ -0,0 +1,81 @@
...
+ * Curated tag options for ComfyHub workflow publishing.
+ * Sourced from https://github.com/Comfy-Org/workflow_templates/blob/main/templates/index.json
+ *
+ * To regenerate: pnpm dlx tsx temp/scripts/extract-comfyhub-tags.ts
+ */
+const COMFY_HUB_TAG_FREQUENCIES = [
+  { tag: 'API', count: 143 },
+  { tag: 'Video', count: 102 },
+  { tag: 'Image', count: 98 },
+  { tag: 'Text to Image', count: 61 },
```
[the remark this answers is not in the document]
DrJKL: @comfyui-wiki ?

src/composables/useFeatureFlags.ts
```diff
@@ -130,6 +133,27 @@ export function useFeatureFlags() {
           false
         )
       )
+    },
+    get workflowSharingEnabled() {
+      return resolveFlag(
+        ServerFeatureFlag.WORKFLOW_SHARING_ENABLED,
+        remoteConfig.value.workflow_sharing_enabled,
+        true
```
DrJKL: Will probably want to make this false before merging.

src/platform/workflow/sharing/composables/useComfyHubProfileGate.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (cachedUserId.value === currentUserId) {
+      return
+    }
+
+    hasProfile.value = null
+    profile.value = null
+    cachedUserId.value = currentUserId
+  }
+
+  async function fetchProfile(options?: {
```
christian-byrne: issue: Race condition -- `fetchProfile()` uses module-level singleton refs but has no deduplication guard. Concurrent callers (e.g., `onMounted` + `checkProfile` in different components) can race, causing one call's failure to overwrite another's success. Consider adding an in-flight promise guard (`if (inflightPromise) return inflightPromise`) or converting to a Pinia store with serialized actions.

src/platform/workflow/sharing/components/ShareWorkflowDialogContent.vue
```diff
@@ -0,0 +1,401 @@
...
+const {
+  state: publishResult,
+  isLoading: isPublishing,
+  execute: handlePublish
+} = useAsyncState(
+  async (): Promise<WorkflowPublishResult | null> => {
+    const workflow = workflowStore.activeWorkflow
+    if (!workflow) return null
+
+    const publishableAssets = await shareService.getShareableAssets()
```
christian-byrne: issue: `handlePublish` re-calls `getShareableAssets()` instead of reusing `assetInfo.value`. If the graph changed between dialog open and publish click, the user acknowledged one set of assets but a different set gets published. Reuse `assetInfo.value` or clear `acknowledged` when assets change.

src/platform/workflow/sharing/composables/useSharedWorkflowUrlLoader.ts
```diff
@@ -0,0 +1,183 @@
...
+      toast.add({
+        severity: 'error',
+        summary: t('g.error'),
+        detail: t('shareWorkflow.loadFailed'),
+        life: 5000
+      })
+      return 'failed'
+    }
+
+    if (result.action === 'copy-and-open' && nonOwnedAssets.length > 0) {
```
christian-byrne, line 157: issue: When `importPublishedAssets` fails (line 161), a toast is shown but execution falls through to return `'loaded'` on line 177. The caller treats this as full success, but the workflow was opened without required assets. Return a distinct `'partial'` status or make the toast sticky.

src/platform/workflow/sharing/composables/useComfyHubProfileGate.ts
```diff
@@ -0,0 +1,132 @@
...
+      const nextProfile = mapHubProfileResponse(await response.json())
+      if (!nextProfile) {
+        hasProfile.value = false
+        profile.value = null
+        return null
+      }
+      hasProfile.value = true
+      profile.value = nextProfile
+      return nextProfile
+    } catch {
```
christian-byrne: suggestion: Bare `catch {}` silently returns `null` and sets `hasProfile = false`. Network errors become indistinguishable from "user has no profile," potentially blocking users who do have profiles from publishing. This codebase has a shared `useErrorHandling` composable (`src/composables/useErrorHandling.ts`) with `wrapWithErrorHandlingAsync` and `toastErrorHandler` that automatically surfaces errors as toasts. Consider using that here to distinguish 404 (no profile) from transient network failures.

src/platform/workflow/sharing/components/publish/ComfyHubPublishWizardContent.vue
```diff
@@ -0,0 +1,141 @@
...
+    onPublish()
+    return
+  }
+
+  isResolvingPublishAccess.value = true
+  try {
+    let profileExists: boolean
+    try {
+      profileExists = await checkProfile()
+    } catch (error) {
```
christian-byrne, line 128: suggestion: The inner `catch` on `checkProfile()` logs to console and returns silently -- the user sees no feedback, the button just stops working. This codebase has `useErrorHandling().toastErrorHandler()` (in `src/composables/useErrorHandling.ts`) which maps errors to user-friendly toast messages. Consider using it here instead of bare `console.error` + silent return.

src/platform/workflow/sharing/components/publish/ComfyHubExamplesStep.vue
```diff
@@ -0,0 +1,140 @@
...
+    emit('update:selectedExampleIds', [...selectedExampleIds, id])
+  }
+}
+
+function addImages(files: FileList) {
+  const newImages: ExampleImage[] = Array.from(files)
+    .filter((f) => f.type.startsWith('image/'))
+    .map((file) => ({
+      id: uuidv4(),
+      url: URL.createObjectURL(file),
```
christian-byrne, line 124: suggestion: `URL.createObjectURL(file)` blob URLs leak when images are removed individually -- they're only revoked on dialog unmount. Revoke removed URLs when updating the image list (e.g., in the remove handler). Worth adding a unit test that asserts `URL.revokeObjectURL` is called when an image is removed, to guard against regression.

src/platform/workflow/sharing/components/publish/ComfyHubThumbnailStep.vue
```diff
@@ -0,0 +1,389 @@
...
+  'update:comparisonAfterFile': [value: File | null]
+}>()
+
+const { t } = useI18n()
+
+function isThumbnailType(value: string): value is ThumbnailType {
+  return value === 'image' || value === 'video' || value === 'imageComparison'
+}
+
+function handleThumbnailTypeChange(value: unknown) {
```
christian-byrne, line 214: suggestion: `handleThumbnailTypeChange` emits null for files but doesn't revoke existing `comparisonPreviewUrls`. Repeated switching accumulates leaked blob URLs. Call `clearAllPreviews()` at the top of `handleThumbnailTypeChange`.

src/platform/workflow/sharing/components/profile/ComfyHubCreateProfileForm.vue
```diff
@@ -0,0 +1,188 @@
...
+        <label for="profile-picture" class="text-sm text-muted-foreground">
+          {{ $t('comfyHubProfile.chooseProfilePicture') }}
+        </label>
+        <label
+          class="flex size-13 cursor-pointer items-center justify-center overflow-hidden rounded-full bg-gradient-to-b from-green-600/50 to-green-900"
+        >
+          <input
+            id="profile-picture"
+            type="file"
+            accept="image/*"
```
christian-byrne, line 32: suggestion: Profile picture, thumbnail, comparison images, and example images all accept `image/*` without file size checks. A user can select multi-GB files causing browser memory exhaustion. Validate file size before accepting (e.g., 10MB limit) and show a user-friendly error.

src/platform/workflow/sharing/components/OpenSharedWorkflowDialogContent.vue
```diff
@@ -0,0 +1,147 @@
...
+
+const nonOwnedAssets = computed(
+  () => sharedWorkflow.value?.assets.filter((a) => !a.in_library) ?? []
+)
+
+const hasAssets = computed(() => nonOwnedAssets.value.length > 0)
+
+const workflowName = computed(() => {
+  if (!sharedWorkflow.value) return ''
+  if (sharedWorkflow.value.name) return sharedWorkflow.value.name
```
christian-byrne, line 168: suggestion (non-blocking): Workflow `name` from the shared workflow API response is only validated as `z.string()`. A crafted name with path separators (`../../malicious/path.json`) could influence where the workflow is stored. Consider stripping path separators and limiting length.

src/platform/workflow/management/stores/workflowStore.ts
```diff
@@ -279,14 +297,21 @@ export const useWorkflowStore = defineStore('workflow', () => {
...
-  const createTemporary = (path?: string, workflowData?: ComfyWorkflowJSON) => {
+  async function createTemporary(
+    path?: string,
+    workflowData?: ComfyWorkflowJSON
+  ): Promise<ComfyWorkflow> {
     const fullPath = getUnconflictedPath(
       ComfyWorkflow.basePath + (path ?? 'Unsaved Workflow.json')
     )
 
+    const normalizedWorkflowData = workflowData
```
christian-byrne, line 301: suggestion (non-blocking): `createTemporary()` and `createNewTemporary()` now silently inject a UUID `id` into workflow data via `ensureWorkflowId()`. Extensions relying on serialized content matching their input will silently break. Consider documenting this new behavior.

src/platform/workflow/sharing/composables/useComfyHubPublishWizard.ts
```diff
@@ -0,0 +1,69 @@
+import { useStepper } from '@vueuse/core'
```
christian-byrne, line 1: suggestion (non-blocking): No tests for `useComfyHubPublishWizard` -- step navigation logic, `canGoNext` validation, and default form data initialization are untested.

src/platform/workflow/sharing/components/publish/ComfyHubPublishWizardContent.vue
```diff
@@ -0,0 +1,141 @@
+<template>
```
christian-byrne, line 1: suggestion (non-blocking): No tests for the publish gate logic in `ComfyHubPublishWizardContent` -- feature flag check, profile gate, double-click guard, and `isPublishDisabled` have zero coverage.

src/platform/workflow/sharing/components/publish/ComfyHubPublishDialog.test.ts
```diff
@@ -0,0 +1,139 @@
+import { flushPromises, mount } from '@vue/test-utils'
```
christian-byrne, line 1: suggestion (non-blocking): Both `useComfyHubProfileGate` and `useComfyHubPublishWizard` are fully mocked in tests. Tests verify mock wiring, not real behavior. Consider testing with real composable implementations.

src/platform/workflow/sharing/components/ShareWorkflowDialogContent.test.ts
```diff
@@ -0,0 +1,456 @@
+import { flushPromises, mount } from '@vue/test-utils'
```
christian-byrne, line 1: suggestion (non-blocking): Missing error/edge case tests: `publishWorkflow` rejection, `getShareableAssets` failure, null `activeWorkflow`, and temporary workflow handling are all uncovered.

src/platform/workflow/sharing/components/publish/ComfyHubDescribeStep.vue
```diff
@@ -0,0 +1,192 @@
+<template>
```
christian-byrne, line 1: suggestion (non-blocking): Core wizard step components (`ComfyHubDescribeStep`, `ComfyHubThumbnailStep`, `ComfyHubExamplesStep`) have zero test coverage for form fields, file upload, and image selection logic.

src/platform/workflow/sharing/components/ShareWorkflowDialogContent.vue
```diff
@@ -0,0 +1,401 @@
+<template>
```
christian-byrne, line 1: suggestion (non-blocking): At 401 lines, this is a large SFC mixing dialog state machine, workflow saving, publish status fetching, asset loading, path building, and rendering. Consider extracting state machine and orchestration into a composable (e.g., `useShareWorkflowState`).

src/platform/workflow/sharing/composables/useSharedWorkflowUrlLoader.ts
```diff
@@ -0,0 +1,183 @@
...
+            root: {
+              class: 'rounded-2xl overflow-hidden w-full sm:w-176 max-w-full'
+            }
+          }
+        }
+      })
+    })
+  }
+
+  async function loadSharedWorkflowFromUrl(): Promise<SharedWorkflowUrlLoadStatus> {
```
christian-byrne, line 100: suggestion (non-blocking): `loadSharedWorkflowFromUrl` has high cyclomatic complexity (~12 branches in ~80 lines), mixing URL management, dialog orchestration, graph loading, and asset importing. Consider extracting into `resolveShareParam()`, dialog call, and `applySharedWorkflow(result)`.

src/composables/useFeatureFlags.ts
christian-byrne: question: `workflowSharingEnabled` defaults to `true` while all other flags default to `false` (opt-in). This exposes the share button in environments without backend support. Is the opt-out default intentional? If so, worth a code comment explaining the rationale.

DrJKL: Nah, let's go straight to live. It's cloud gated elsewhere already.

src/platform/workflow/sharing/components/publish/ComfyHubPublishWizardContent.vue
DrJKL: A silent catch that logged to empty air —
The publish button froze with no excuse.
Now \	oastErrorHandler\ tells users what went where,
And the inner try/catch puts that feedback to use.

src/platform/workflow/sharing/components/profile/ComfyHubCreateProfileForm.vue
DrJKL: A multi-gig upload could crash the tab,
No guard at all on any file input.
Now \isFileTooLarge\ checks each grab,
With toast and limits — oversized files get the boot.

src/platform/workflow/sharing/components/OpenSharedWorkflowDialogContent.vue
DrJKL: The \../../malicious\ name you feared might sneak on through?
It's caught in \shareSchemas\ at the Zod transform gate.
\sanitizeWorkflowName\ strips the separators from view,
Truncates to 200 chars — no traversal to debate.

src/composables/useFeatureFlags.ts
DrJKL: The lone \	rue\ default broke the opt-in pact,
While every other flag was \alse\ by creed.
Now \
esolveFlag()\ keeps the contract intact,
And \alse\ default aligns — consistency decreed.

src/platform/workflow/sharing/composables/useComfyHubPublishWizard.ts
DrJKL: A composable bare, with zero tests in sight —
Step logic, \canGoNext\, all unverified and free.
Fourteen tests now guard each path, each flight,
From default form data to profile-step decree.

src/platform/workflow/sharing/components/publish/ComfyHubPublishWizardContent.vue
DrJKL: The gate had guards but none were ever checked,
No test to prove the double-click was barred.
Eleven tests now verify each aspect,
From flag bypass to error-toast bombard.

src/platform/workflow/sharing/components/ShareWorkflowDialogContent.test.ts
DrJKL: Ten happy paths but not a single storm,
No test for when the publish call would fail.
Six edge-case tests now break the sunny norm —
Null workflows, rejections — each unhappy trail.

src/platform/workflow/sharing/composables/useComfyHubProfileGate.ts
DrJKL: The bare catch was wrong, it swallowed every sin,
So network glitches looked like "no profile found."
Now `toastErrorHandler` reports what happened within,
And `hasProfile` stays `null` — retry on solid ground.

src/platform/workflow/sharing/components/publish/ComfyHubThumbnailStep.vue
DrJKL: Those leaky blobs piled up with every switch,
Comparison URLs left dangling, never freed.
Now `useObjectUrl` handles every stitch —
Set the ref to null, revocation's guaranteed.

src/platform/workflow/sharing/composables/useComfyHubProfileGate.ts
DrJKL: We will probably want at least one store for Hub things.

src/platform/workflow/sharing/composables/useSharedWorkflowUrlLoader.ts
DrJKL: The assets are still usable, they're just not added to the user's library.

src/platform/workflow/sharing/composables/useComfyHubProfileGate.ts
DrJKL: The singleton refs could race when two mounts collide,
So `inflightFetch` now guards the single flight.
Concurrent callers share one promise ride —
Dedup before the fetch, the guard sits tight.

Fixed in `8419ed4` — `fetchProfile()` now deduplicates concurrent calls via an in-flight promise guard. Also added a TODO to migrate to a Pinia store.

src/platform/workflow/sharing/components/ShareWorkflowDialogContent.vue
DrJKL: The user acknowledged set A, but set B got sent —
A fresh `getShareableAssets()` call could drift.
Now `assetInfo.value` is what the publish meant,
The acknowledged payload and the published gift.

Fixed in `8419ed4` — `handlePublish` now reuses `assetInfo.value` instead of re-fetching, ensuring the published assets match what the user acknowledged.

src/platform/workflow/sharing/composables/useSharedWorkflowUrlLoader.ts
DrJKL: The import could fail but `'loaded'` still returned,
A partial success masked as victory complete.
Now `'loaded-without-assets'` marks the lesson learned —
The toast still warns, the status stays concrete.

Fixed in `8419ed4` — `loadSharedWorkflowFromUrl` now returns `'loaded-without-assets'` when `importPublishedAssets` fails, distinguishing partial success from full success.

src/platform/workflow/sharing/components/publish/ComfyHubExamplesStep.vue
DrJKL: Your forward-looking eye has spotted well the seam,
But `addImages` only grows — no remove path to leak.
When removal lands, we'll revoke each orphaned stream,
For now the dialog's close reclaims each URL unique.

src/platform/workflow/sharing/components/publish/ComfyHubDescribeStep.vue
DrJKL: These steps are mostly wiring — props in, emits out,
With form-field bindings too thin for unit test.
The wizard's logic wears its test-suite like a scout,
And E2E will prove these forms pass every quest.

src/platform/workflow/sharing/components/ShareWorkflowDialogContent.vue
DrJKL: Four hundred lines for save, publish, stale, and share —
A multi-tab state machine, each block defined.
Extraction's cost (six refs to pass) outweighs the care,
But noted for the future — when more states are signed.

src/platform/workflow/sharing/composables/useSharedWorkflowUrlLoader.ts
DrJKL: The branches number eight, not twelve as claimed,
A pipeline staged with exits, clean and straight.
Each step returns or falls through, neatly framed —
More splitting here would just add boilerplate.

src/platform/workflow/sharing/constants/comfyHubTags.ts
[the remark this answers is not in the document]
comfyui-wiki: This means I need to clean up some tags, right?

browser_tests/tests/nodeSearchBox.spec.ts-snapshots/added-node-no-connection-chromium-linux.png
AustinMroz, line 1: This is a flake test update, should be ksampler

src/components/TopMenuSection.vue
```diff
@@ -217,6 +236,25 @@ onMounted(() => {
   }
 })
 
+function importShareDialog() {
```
AustinMroz: Since I'lll be reusing this code for adding share buttons in other places, I'd appreciate it if the super cool lazy import code was extracted into a different file. :heart_hands:

src/locales/en/main.json
```diff
@@ -2998,7 +2998,131 @@
...
+  },
+  "shareWorkflow": {
+    "shareLinkTab": "Share",
+    "publishToHubTab": "Publish",
+    "loadingTitle": "Share workflow",
+    "unsavedTitle": "Save workflow first",
+    "unsavedDescription": "You must save your workflow before sharing. Save it now to continue.",
+    "saveButton": "Save workflow",
+    "saving": "Saving...",
+    "workflowNameLabel": "Workflow name",
```
AustinMroz, line 3013: A lot of these keys feel like they should be on higher level. The PR even adds 2 keys for "Workflow name".

But I think localization key cleanup is better handled as a followup PR.

src/platform/workflow/sharing/components/ShareWorkflowDialogContent.vue
```diff
@@ -0,0 +1,400 @@
...
+    : normalizedFilename
+}
+
+async function refreshDialogState() {
+  const workflow = workflowStore.activeWorkflow
+
+  if (!workflow || workflow.isTemporary || workflow.isModified) {
+    dialogState.value = 'unsaved'
+    if (workflow) {
+      workflowName.value = stripJsonExtension(workflow.filename)
```
AustinMroz, line 302: Will likely need to revisit this to also strip `.app.json`, but this is minor, doesn't need to be fixed by you, and probably better handled in a future App-fix PRs

AustinMroz (commented):
Didn't have time to finish review, so I'm positing with current comments.

EDIT: Finished reviewing the rest and had no further feedback

src/locales/en/main.json
DrJKL: @coderabbitai Open an issue to clean up our localization key structure.

src/components/TopMenuSection.vue
DrJKL: Extraction completed 🫡

src/platform/workflow/sharing/components/publish/ComfyHubExamplesStep.vue
christian-byrne: fair enough - addImages only grows, no removal handler exists to leak from. dialog unmount revokes all URLs. will add revocation when removal lands.

src/platform/workflow/management/stores/workflowStore.ts
christian-byrne: note (non-blocking): this is a good callout for documentation. the behavior change is intentional - createTemporary injects a UUID so workflows have stable identity for sharing. tagging for followup.

src/platform/workflow/sharing/components/publish/ComfyHubPublishDialog.test.ts
christian-byrne: note (non-blocking): agreed the mocks are heavy - worth refactoring tests to use real composable implementations in a followup.

src/platform/workflow/sharing/components/publish/ComfyHubDescribeStep.vue
christian-byrne: makes sense - these are thin prop-binding components, E2E coverage is the right layer for them.

src/platform/workflow/sharing/components/ShareWorkflowDialogContent.vue
christian-byrne: yeah the extraction cost (6+ refs to pass) isn't worth it at this size. noted for when more states arrive.

src/platform/workflow/sharing/composables/useSharedWorkflowUrlLoader.ts
christian-byrne: recounted - you're right, it's 8 sequential exit points not 12 interleaved branches. the pipeline pattern reads cleanly as-is.

src/platform/workflow/sharing/components/ShareWorkflowDialogContent.vue
christian-byrne: note (non-blocking): agreed this should be handled in an App-fix PR, not here.

@coderabbitai Make an Issue to also strip .app.json extension in stripJsonExtension (or a dedicated helper) in a followup PR

christian-byrne (approved):
Approving — solid feature PR. Architecture is sound, dedup guards work correctly, test coverage is good (124 tests across 10 files), and all prior review feedback has been addressed.

The items below are all non-blocking but worth tracking for followup.

## Security hardening

- **`ShareAssetThumbnail.vue` L46-59** — `normalizedPreviewUrl` has no protocol allowlist; `catch` block returns raw `previewUrl`. Add protocol guard (`http:`, `https:`, `blob:`) and return `null` from catch.
- **`shareSchemas.ts` L17-22** — `sanitizeWorkflowName` strips `/\:` but not null bytes (`\0`) or control characters (U+0001-U+001F). Add `.replaceAll(/[\x00-\x1f]/g, "")`.

## Accessibility

- **`ShareWorkflowDialogContent.vue` L11-29, L40-44, L129-134** — Tab buttons have `role="tab"` + `aria-selected` but missing `aria-controls`. Tabpanels lack `id` attrs. Incomplete WAI-ARIA tabs pattern.

## Vue patterns

- **`ComfyHubThumbnailStep.vue` L239-252, L335-344** — `thumbnailOptions` and `comparisonSlots` call `t()` at setup time (not reactive to locale change). Wrap in `computed()` like `ComfyHubDescribeStep.vue` L146 does.
- **`ComfyHubThumbnailStep.vue` L256, L264** — `isVideoFile` should be a `computed` derived from `thumbnailFile` instead of a manually-synced ref.
- **`ShareWorkflowDialogContent.vue` L129-131** — `v-if` + `v-show` on publish panel is redundant. Use one or the other.

## Robustness

- **`useComfyHubProfileGate.ts` L57-59** — `catch` block does not set `hasProfile.value = false`, leaving it `null` (indeterminate for direct UI reads).
- **`ShareWorkflowDialogContent.vue` L194-202** — Stale detection compares local fs mtime with hub server timestamp. Add a comment documenting the same-clock assumption.
- **`workflowShareService.ts` L106, L131, L174** — `response.json()` after ok check has no try-catch. A 200 with non-JSON body throws unhelpful SyntaxError.

## Styling

- **`ComfyHubCreateProfileForm.vue` L44** — `text-white` hardcoded instead of design token.

## Test coverage gaps

- No test for `loadGraphData` throwing (`useSharedWorkflowUrlLoader`)
- No test for `publishWorkflow` HTTP error (`workflowShareService`)
- No test for `getPublishStatus` non-404 error (e.g., 500)
- `SharedWorkflowLoadError.isRetryable` branching untested
- `validateFileSize.ts` has zero tests
- `getShareableAssets(includingPublic: true)` branch untested

---

@coderabbitai Create a single tracking issue assigned to @DrJKL covering all the items listed above as followup work for this PR. Title it "followup: share-by-url hardening and test coverage" and organize by the same sections (security, accessibility, vue patterns, robustness, styling, test gaps).

PR #8953, opened 2026-02-18
feat: warn user that Persist will be disabled when turning off AutoSave

Yourz:
## Summary

Prompt user with a confirmation dialog when setting AutoSave to off, informing them that Persist will also be disabled.

## Changes

- **What**:
  - Added `onChange` handler to `Comfy.Workflow.AutoSave` setting that shows a confirm dialog when switching to `off` while Persist is enabled. Default action (confirm/close/dismiss) disables Persist;
  - secondary button keeps Persist enabled. Changed `Comfy.Workflow.Persist` default value from `true` to `false`.

image

## Review Focus

- Dialog dismiss behavior: closing via X, ESC, or clicking outside all trigger the default action (disable Persist). This is intentional per requirements.
- Persist default changed from true to false to align with AutoSave defaulting to off.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8953-feat-warn-user-that-Persist-will-be-disabled-when-turning-off-AutoSave-30b6d73d3650818f8eb2e5eeff3046ca) by [Unito](https://www.unito.io)


src/locales/en/main.json
```diff
@@ -3079,5 +3079,11 @@
       "duplicateName": "A secret with this name already exists",
       "duplicateProvider": "A secret for this provider already exists"
     }
+  },
+  "autoSaveOffPersistWarning": {
+    "title": "Disable Persist?",
+    "body": "Turning off Auto Save will also disable workflow persistence. Your workflow state will no longer be restored on page reload.",
```
christian-byrne, line 3085: Is this true? The persist setting and autosave setting are separate. When autosave is disabled, the persistence should still work actually, just not be saving to the server.

christian-byrne:
Will close for now, please re-open if [comment above](https://github.com/Comfy-Org/ComfyUI_frontend/pull/8953#discussion_r2825174780) is wrong.

PR #8955, opened 2026-02-18
fix: remove cloud guard for clear history menu

benceruleanlu:
## Summary
- remove the cloud-only visibility guard for the queue overlay clear-history (`...`) menu
- keep existing clear-history behavior unchanged; this only affects menu visibility

This was added a long time ago when there was a bug where the clear history items call actually deleted the assets. After asset updates in recent times, this is now safe as purely a history deletion, while keeping the asset.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8955-fix-remove-cloud-guard-for-clear-history-menu-30b6d73d3650815992abf3daabc6f32c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/queue/QueueOverlayHeader.vue
+++ b/src/components/queue/QueueOverlayHeader.vue
@@ -36,7 +36,7 @@
         <i class="icon-[lucide--list-x] size-4" />
       </Button>
     </div>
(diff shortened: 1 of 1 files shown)


benceruleanlu:
Just bringing this into the main PR with a comment

PR #8956, opened 2026-02-18
feat: node-specific error tab with selection-aware grouping and error overlay

jaeone94:
## Summary
Enhances the error panel with node-specific views: single-node selection shows errors grouped by message in compact mode, container nodes (subgraph/group) expose child errors via a badge and "See Error" button, and a floating ErrorOverlay appears after execution failure with a deduplicated summary and quick navigation to the errors tab.

## Changes
- **Consolidate error tab**: Remove `TabError.vue`; merge all error display into `TabErrors.vue` and drop the separate `error` tab type from `rightSidePanelStore`
- **Selection-aware grouping**: Single-node selection regroups errors by message (not `class_type`) and renders `ErrorNodeCard` in compact mode
- **Container node support**: Detect child-node errors in subgraph/group nodes via execution ID prefix matching; show error badge and "See Error" button in `SectionWidgets`
- **ErrorOverlay**: New floating card shown after execution failure with deduplicated error messages, "Dismiss" and "See Errors" actions; `isErrorOverlayOpen` / `showErrorOverlay` / `dismissErrorOverlay` added to `executionStore`
- **Refactor**: Centralize error ID collection in `executionStore` (`allErrorExecutionIds`, `hasInternalErrorForNode`); split `errorGroups` into `allErrorGroups` (unfiltered) and `tabErrorGroups` (selection-filtered); move `ErrorOverlay` business logic into `useErrorGroups`

## Review Focus
- `useErrorGroups.ts`: split into `allErrorGroups` / `tabErrorGroups` and the new `filterBySelection` parameter flow
- `executionStore.ts`: `hasInternalErrorForNode` helper and `allErrorExecutionIds` computed
- `ErrorOverlay.vue`: integration with `executionStore` overlay state and `useErrorGroups`

## Screenshots
image
image
image


src/stores/executionStore.ts
```diff
@@ -291,6 +317,7 @@ export const useExecutionStore = defineStore('execution', () => {
   function handleExecutionStart(e: CustomEvent<ExecutionStartWsMessage>) {
     lastExecutionError.value = null
     lastPromptError.value = null
+    isErrorOverlayOpen.value = false
```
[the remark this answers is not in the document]
jaeone94: Fixed in e02d58d

src/stores/executionStore.ts
```diff
@@ -654,20 +742,49 @@ export const useExecutionStore = defineStore('execution', () => {
...
+  const allErrorExecutionIds = computed<string[]>(() => {
+    const ids: string[] = []
+    if (lastNodeErrors.value) {
+      ids.push(...Object.keys(lastNodeErrors.value))
+    }
+    if (lastExecutionError.value) {
+      ids.push(String(lastExecutionError.value.node_id))
+    }
+    return ids
+  })
```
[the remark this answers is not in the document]
jaeone94: Fixed in e02d58d

src/components/error/ErrorOverlay.vue
```diff
@@ -0,0 +1,100 @@
...
+  () => isErrorOverlayOpen.value && totalErrorCount.value > 0
+)
+
+function dismiss() {
+  executionStore.dismissErrorOverlay()
+}
+
+function seeErrors() {
+  if (canvasStore.canvas) {
+    canvasStore.canvas.deselectAll()
```
christian-byrne: Is it possible for this button to be pushed when a single node is selected already?

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -162,6 +154,11 @@ async function contactSupport() {
     is_external: true,
     source: 'error_dialog'
   })
-  await useCommandStore().execute('Comfy.ContactSupport')
+  try {
+    await useCommandStore().execute('Comfy.ContactSupport')
+  } catch (error) {
+    console.error(error)
+    useToastStore().addAlert(t('rightSidePanel.contactSupportFailed'))
+  }
```
christian-byrne: nit: we likely don't need a try/catch or error-handling for the contact support command.

src/composables/canvas/useFocusNode.ts
```diff
@@ -29,20 +35,44 @@ async function navigateToGraph(targetGraph: LGraph) {
...
+    const parts = parseNodeExecutionId(nodeId)
+    const parentId = parts && parts.length > 1 ? String(parts[0]) : null
+    const parentNode = parentId
+      ? app.rootGraph.getNodeById(Number(parentId))
+      : null
+
+    if (parentNode && isGroupNode(parentNode) && parentNode.graph) {
+      await navigateToGraph(parentNode.graph as LGraph)
+      canvasStore.canvas?.animateToBounds(parentNode.boundingRect)
+      return
```
christian-byrne: Some of the logic/actions we are doing here are repeated elsewhere - I wonder if we should move them to be next to the other utils in graphTraversalUtils, executionStore, workflowStore, etc.? And then only have to import them and not duplicate code. WDYT?

src/scripts/api.ts
```diff
@@ -830,7 +830,20 @@ export class ComfyApi extends EventTarget {
     })
 
     if (res.status !== 200) {
-      throw new PromptExecutionError(await res.json())
+      const text = await res.text()
```
christian-byrne, line 834: question: is there an approach where we don't need to do this? In theory we can just check `isCloud` at any point in the codebase and don't have to add the state into the error object itself right? Wondering if it will make things cleaner (could be wrong).

src/stores/executionStore.ts
```diff
@@ -49,6 +48,33 @@ interface QueuedPrompt {
...
+  const jsonStart = exceptionMessage.indexOf('{')
+  if (jsonStart === -1) return null
+
+  try {
+    const parsed: unknown = JSON.parse(exceptionMessage.substring(jsonStart))
+    return isCloudValidationError(parsed) ? parsed : null
+  } catch {
+    return null
+  }
+}
```
christian-byrne: What's the actual thing we are matching on here? I am somewhat confused because I don't know the different error messages/types that motivated this code.

src/stores/executionStore.ts
```diff
@@ -392,19 +420,80 @@ export const useExecutionStore = defineStore('execution', () => {
...
+    clearInitializationByPromptId(detail.prompt_id)
+    resetExecutionState(detail.prompt_id)
+
+    if (hasNodeErrors) {
+      lastNodeErrors.value = node_errors
+    } else if (promptError) {
+      lastPromptError.value = promptError
+    }
+    return true
+  }
```
christian-byrne, line 450: nit: I think some of this logic can be moved into a separate module (e.g., utils file) which will make the logic more testable. We essentially have a somewhat complex decision tree here which is the perfect type of thing to write unit tests for. The extracting will also make the decision tree much more clear for future maintainers. The stateful stuff (e.g., setting lastNodeErrors and lastPromptError will of course will be in executionStore).a

src/stores/executionStore.ts
```diff
@@ -392,19 +420,80 @@ export const useExecutionStore = defineStore('execution', () => {
...
+
+  function handleCloudValidationError(
+    detail: ExecutionErrorWsMessage
+  ): boolean {
+    const extracted = tryExtractValidationError(detail.exception_message)
+    if (!extracted) return false
+
+    const { error, node_errors } = extracted
+    const hasNodeErrors = node_errors && Object.keys(node_errors).length > 0
```
christian-byrne: nit: It seems wrong that we would be keeping our own `hasNodeErrors` in the function scope while also maintaining `hasNodeErrors` computed in the outer store scope. Does it indicate that we could improve the code flow (it's okay if not)?

src/stores/executionStore.ts
```diff
@@ -661,20 +750,52 @@ export const useExecutionStore = defineStore('execution', () => {
...
+      const nodeId = lastExecutionError.value.node_id
+      if (nodeId !== null && nodeId !== undefined) {
+        ids.push(String(nodeId))
+      }
+    }
+    return ids
+  })
+
+  /** Total count of all individual errors */
+  const totalErrorCount = computed(() => {
```
christian-byrne: nit: this seems like it wants to be separate computeds that derive from each other rather than a single computed with multiple conditionals and for loops.

src/stores/executionStore.ts
```diff
@@ -661,20 +750,52 @@ export const useExecutionStore = defineStore('execution', () => {
...
+    if (lastNodeErrors.value) {
+      ids.push(...Object.keys(lastNodeErrors.value))
+    }
+    if (lastExecutionError.value) {
+      const nodeId = lastExecutionError.value.node_id
+      if (nodeId !== null && nodeId !== undefined) {
+        ids.push(String(nodeId))
+      }
+    }
+    return ids
```
christian-byrne, line 720: This logic is essentiall repeated in the `totalErrorCount`. If we had intermediate computeds, we could just read the `.count` of them instead of having to traverse twice in the same way.

src/stores/executionStore.ts
```diff
@@ -49,6 +48,33 @@ interface QueuedPrompt {
...
+}
+
+function tryExtractValidationError(
+  exceptionMessage: string
+): CloudValidationError | null {
+  const jsonStart = exceptionMessage.indexOf('{')
+  if (jsonStart === -1) return null
+
+  try {
+    const parsed: unknown = JSON.parse(exceptionMessage.substring(jsonStart))
```
christian-byrne: What about `exceptionMessage.substring(jsonStart, exceptionMessage.lastIndexOf('}') + 1)`? Is there a reason to not do it deterministically?

src/components/error/ErrorOverlay.vue
christian-byrne, line 1: This component might want accessibility things like `role="status"` or `aria-live=polite` -- or whatever the correct ones are for this type of component/UX.

christian-byrne (commented):
Looking good!

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Good call — removed the `try/catch` and the unused `contactSupportFailed` localization string + `useToastStore` import.

src/composables/canvas/useFocusNode.ts
jaeone94: Agreed! Extracted a `getRootParentNode(rootGraph, executionId)` utility in `graphTraversalUtil.ts`. It retrieves the root parent node from a hierarchical execution ID (e.g. `"123:456"`) without coupling to a specific type check. Callers now import `getRootParentNode` and apply their own `isGroupNode()` / `isSubgraphNode()` checks, eliminating the duplicated parsing logic.

src/scripts/api.ts
jaeone94: You're right — we don't need `isCloud` baked into the error object. Unified the type to a single `'server_error'` in `api.ts`. The environment-specific i18n key (`server_error_local` / `server_error_cloud`) is now resolved at display time in `processPromptError` using `isCloud`, keeping the data layer environment-agnostic.

src/stores/executionStore.ts
jaeone94: Good question! In cloud, prompt submissions don't return errors synchronously — they go through an async job queue and errors arrive via WebSocket as `execution_error` events. The problem is that validation errors (the same `{ error, node_errors }` shape the local backend returns as HTTP 400) get wrapped inside `exception_message` as a plain-text string like:

`"Failed to send prompt request: ... 400: {\"error\":{\"type\":\"prompt_outputs_failed_validation\",...},\"node_errors\":{...}}"`

So `tryExtractValidationError` finds the first `{` in that string and parses the trailing JSON back into the structured `{ error, node_errors }` format that the rest of the error pipeline expects. This is a workaround for the cloud backend embedding structured data inside a string field. I've added JSDoc comments with an example to make this clearer.

src/stores/executionStore.ts
jaeone94: Good idea — extracted all pure logic into `src/utils/executionErrorUtil.ts` with a new `classifyCloudValidationError` function that returns a discriminated union (`{ kind: 'nodeErrors', ... }` | `{ kind: 'promptError', ... }` | `null`). `handleCloudValidationError` in `executionStore.ts` is now a thin state layer. Added 18 unit tests in `executionErrorUtil.test.ts` covering all exported functions.

src/stores/executionStore.ts
jaeone94: These two actually serve different purposes at different layers:

- **Util (`classifyCloudValidationError`):** Checks `node_errors` at _parse time_ to classify the raw data — "does this embedded JSON contain node errors?" This determines the return type (`nodeErrors` vs `promptError`).
- **Store (`hasNodeError` computed):** Checks `lastNodeErrors` as _reactive state_ for the UI — "should we show node error indicators right now?"

The util classifies incoming data; the store reflects current state. They happen to check a similar condition but at completely different points in the pipeline, so I think the current flow is appropriate.

src/stores/executionStore.ts
jaeone94: Nice suggestion — split into three per-category computeds (`promptErrorCount`, `nodeErrorCount`, `executionErrorCount`) with `totalErrorCount` now derived from their sum.

src/stores/executionStore.ts
jaeone94: The `totalErrorCount` split (into `promptErrorCount`, `nodeErrorCount`, `executionErrorCount`) is done — thanks for the suggestion!

For `allErrorExecutionIds` though, it collects **execution ID strings** (`Object.keys`), while `nodeErrorCount` sums **error item counts** (`Object.values → errors.length`). A single node can have multiple errors, so `allErrorExecutionIds.length` ≠ `nodeErrorCount`. They traverse the same source but extract fundamentally different data, so one can't be derived from the other without keeping the original.

src/stores/executionStore.ts
jaeone94: Good point — updated to use both `indexOf('{')` and `lastIndexOf('}')` for deterministic boundary extraction. More robust against potential trailing content.

src/components/error/ErrorOverlay.vue
jaeone94: Added `role="alert"` and `aria-live="assertive"` to the error overlay container.

christian-byrne (dismissed):
LGTM!

christian-byrne:
The browser tests that check execution error behaviors will need to be updated, otherwise looks good to merge.

christian-byrne (approved):

PR #8963, opened 2026-02-18
fix: reload template workflows when locale changes

Myestery:
## Summary
- Templates were fetched once with the initial locale and cached behind an `isLoaded` guard. Changing language updated i18n UI strings but never re-fetched locale-specific template data (names, descriptions) from the server.
- Extracts core template fetching into `fetchCoreTemplates()` and adds a `watch` on `i18n.global.locale` to re-fetch when the language changes.

## Test plan
- [ ] Open the templates panel
- [ ] Change language in settings (e.g. English -> French)
- [ ] Verify template names and descriptions update without a page refresh
- [ ] Verify initial load still works correctly

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8963-fix-reload-template-workflows-when-locale-changes-30b6d73d36508178a2f8c2c8947b5955) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/workflow/templates/repositories/workflowTemplatesStore.ts
+++ b/src/platform/workflow/templates/repositories/workflowTemplatesStore.ts
@@ -1,5 +1,5 @@
 import { defineStore } from 'pinia'
-import { computed, ref, shallowRef } from 'vue'
+import { computed, ref, shallowRef, watch } from 'vue'
 
 import { i18n, st } from '@/i18n'
 import { isCloud } from '@/platform/distribution/types'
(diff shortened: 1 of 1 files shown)


christian-byrne:
Wait so how do I reproduce the initial bug? I just change languages and check network to see that a new request is not made?

Myestery:
> Wait so how do I reproduce the initial bug? I just change languages and check network to see that a new request is not made?

you replicate by changing the language and seeing that the templates are still stuck on the old language

christian-byrne (approved):
LGTM

PR #8967, opened 2026-02-18
test: fix flaky 'Can drag node' screenshot test

christian-byrne:
## Summary

Fix intermittently failing 'Can drag node' screenshot test that blocks CI on main and all PR branches.

## Changes

- **What**: Add `nextFrame()` waits after switching `Comfy.UseNewMenu` from `Top` to `Disabled` in the `beforeEach` hook. The setting change removes the top bar, causing the canvas to resize. Without waiting, the hardcoded drag coordinates can miss the node entirely (resulting in a canvas pan instead of a node drag).

## Review Focus

The root cause: `setSetting('Comfy.UseNewMenu', 'Disabled')` triggers a layout shift (top bar disappears → canvas grows vertically). Litegraph needs 1-2 frames to process the canvas resize. The drag starts at hardcoded screen coords `{622, 400}` which only map to the node after the resize settles.


christian-byrne:
I am pretty sure the node is truley not being dragged though, so this seems wrong.

DrJKL (dismissed):

christian-byrne:
What is happening in the test runner?

PR #8508, opened 2026-01-31
Feat/workspaces 6 billing

simula-r:
## Summary

Implements billing infrastructure for team workspaces, separate from legacy personal billing.

## Changes

- **Billing abstraction**: New `useBillingContext` composable that switches between legacy (personal) and workspace billing based on context
- **Workspace subscription flows**: Pricing tables, plan transitions, cancellation dialogs, and payment preview components for workspace billing
- **Top-up credits**: Workspace-specific top-up dialog with polling for payment confirmation
- **Workspace API**: Extended with billing endpoints (subscriptions, invoices, payment methods, credits top-up)
- **Workspace switcher**: Now displays tier badges for each workspace
- **Subscribe polling**: Added polling mechanisms (`useSubscribePolling`, `useTopupPolling`) for async payment flows

## Review Focus

- Billing flow correctness for workspace vs legacy contexts
- Polling timeout and error handling in payment flows

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8508-Feat-workspaces-6-billing-2f96d73d365081f69f65c1ddf369010d) by [Unito](https://www.unito.io)

PR #8408, opened 2026-01-29
fix: use getAuthHeader in createCustomer to support API key auth

christian-byrne:
## Summary

Fixes authentication failure when using API key authentication on staging server after frontend update to 1.33.10.

image

## Changes

- **What**: Changed `createCustomer()` to use `getAuthHeader()` instead of `getFirebaseAuthHeader()`, allowing API key users to authenticate successfully

## Review Focus

- Verify `getAuthHeader()` correctly falls back to API key when no Firebase token exists
- Backend `/customers` endpoint supports `X-API-KEY` header (per cloud PR #1766)

Fixes COM-12398

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8408-fix-use-getAuthHeader-in-createCustomer-to-support-API-key-auth-2f76d73d3650819994e3e6d3ed9f3dfa) by [Unito](https://www.unito.io)

PR #8983, opened 2026-02-19
fix: use getAuthHeader in createCustomer for API key auth support

christian-byrne:
## Summary

Re-apply the fix from PR #8408 that was accidentally reverted by PR #8508 — `createCustomer` must use `getAuthHeader()` (not `getFirebaseAuthHeader()`) so API key authentication works.

## Changes

- **What**: Changed `createCustomer` in `firebaseAuthStore.ts` to use `getAuthHeader()` which falls back through workspace token → Firebase token → API key. Added regression tests covering API key auth, Firebase auth, and no-auth paths.

## Review Focus

This is the same one-line fix from #8408. PR #8508 ("Feat/workspaces 6 billing") overwrote it during merge because it was branched before #8408 landed. The regression test should prevent this from happening again.

Fixes COM-15060

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8983-fix-use-getAuthHeader-in-createCustomer-for-API-key-auth-support-30c6d73d365081c2aab6d5defa5298d6) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/firebaseAuthStore.test.ts
+++ b/src/stores/firebaseAuthStore.test.ts
@@ -99,6 +99,19 @@ vi.mock('@/stores/toastStore', () => ({
 // Mock useDialogService
 vi.mock('@/services/dialogService')
 
+// Mock apiKeyAuthStore
+const mockApiKeyGetAuthHeader = vi.fn().mockReturnValue(null)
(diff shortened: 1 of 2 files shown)


simula-r (approved):
LGTM

PR #8985, opened 2026-02-19
feat(litegraph): use number input type in    canvas prompt dialog

### LINKED ISSUE #4768: [feature request] Use input type "number" on numbers
Originally from comfyanonymous/ComfyUI#1612:

>Whenever you double click a numeric value to show the popup dialog to manually set it, the input type is text, which means you don't get the native up/down arrows and scroll wheel does nothing.
>
>Switching those inputs to type="number" and adding step="0.05" would show the arrows, and allow keyboard up/down arrows > and mouse wheel to change the value in 0.05 increments/decrements. Also, min and max attributes can limit the values > whenever needed. Having to manually type values while you could use the mouse wheel to change them is rather cumbersome.

> ➡ [Input type number on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number)

dante01yoon:
## Summary

  Use `<input type="number">` with
  `min`/`max`/`step` attributes for numeric widget
   prompt dialogs, enabling native browser
  controls (arrows, scroll, keyboard up/down).

  ## Changes

  - **What**: `LGraphCanvas.prompt()` accepts a
  `PromptOptions` object (backward-compatible with
   the existing `boolean` parameter).
  `NumberWidget.onClick` passes `inputType:
  'number'` with `min`/`max`/`step`. Fixed
  `changeTracker.ts` prompt override that was
  dropping the 5th argument, causing the options
  to never reach the original method.

https://github.com/user-attachments/assets/1bd8a1c8-86ba-4618-b041-61c3de653297

- add Edit expression on right click on number  widget
스크린샷 2026-03-03 오후 5 03 36

- slash `/` will change number type input to text type input 

  ## Review Focus

  The `changeTracker.ts` monkey-patch of
  `LGraphCanvas.prototype.prompt` was silently
  discarding the `multilineOrOptions` parameter.
  This was the root cause of `type="number"` not
  being applied despite `NumberWidget` passing the
   correct options.

  ## Test Plan

  - [x] Unit tests added (`NumberWidget.test.ts` —
   5 tests covering center click options, step
  values, arrow area exclusion, unbounded options)
  - [ ] E2E test intentionally omitted: the prompt
   dialog requires canvas-coordinate clicking
  which is brittle, and the unit tests already
  cover the core logic. The `changeTracker.ts` fix
   is a transparent argument passthrough unlikely
  to regress in the same pattern.

  Fixes #4768


christian-byrne (commented):
I think the only issue with this one is that we have this expression evaluator system - so if the number input prevents entering mathematical symbols then that feature might not work - would need to confirm manually though.

dante01yoon:
> I think the only issue with this one is that we have this expression evaluator system - so if the number input prevents entering mathematical symbols then that feature might not work - would need to confirm manually though.

Fair point. One approach can be we get separate input field type for expression which allow text based input for specific field or allow syntax starts with `/` which begin allow text input. I don't think we need expression everywhere

christian-byrne:
> > I think the only issue with this one is that we have this expression evaluator system - so if the number input prevents entering mathematical symbols then that feature might not work - would need to confirm manually though.
> 
> Fair point. One approach can be we get separate input field type for expression which allow text based input for specific field or allow syntax starts with `/` which begin allow text input. I don't think we need expression everywhere

That would be a good solution. But in that case, would we need to audit all the existing node definitions out in the wild to see which do and do not want expression-compatibility?

dante01yoon:
> That would be a good solution. But in that case, would we need to audit all the existing node definitions out in the wild to see which do and do not want expression-compatibility?

Thanks for the feedback! I confirmed that `evaluateInput` is only called from `NumberWidget`'s `canvas.prompt()` callback, and the Vue node number input (`WidgetInputNumberInput.vue`) handles expression evaluation independently — so no node-level audit is needed.

I'm planning a follow-up PR that adds an `fx` toggle button to the prompt dialog. When clicked, it switches the input from `type="number"` to `type="text"`, allowing expression entry (e.g. `1+2*3`). This keeps native number UX (scroll wheel, arrow keys, step) as the default while still supporting expressions on demand. The toggle lives entirely in the dialog UI, so no changes to node definitions are required.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -8489,6 +8517,26 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
...
+              getContextMenuOptions: (opts: {
+                e: CanvasPointerEvent
+                node: LGraphNode
+                canvas: LGraphCanvas
+              }) => IContextMenuValue[]
+            }
+          ).getContextMenuOptions({ e: event, node, canvas: this })
+          if (widgetMenuItems.length) {
+            menu_info.unshift(...widgetMenuItems, null)
+          }
```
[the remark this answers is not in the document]
christian-byrne: suggestion (non-blocking): valid point about defensive coding - the type guard is cleaner and Array.isArray is a good safety net for custom widget implementations.

src/lib/litegraph/src/widgets/NumberWidget.ts
```diff
@@ -71,14 +87,41 @@ export class NumberWidget
...
+    node,
+    canvas
+  }: WidgetEventOptions): IContextMenuValue[] {
+    return [
+      {
+        content: t('widgets.editExpression'),
+        callback: () => {
+          canvas.prompt(
+            'Value',
+            this.value,
```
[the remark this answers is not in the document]
christian-byrne: suggestion (non-blocking): good catch - the existing "Value" prompt title from the old code path was carried over but should use t() for consistency since the file already imports it.

christian-byrne (commented):
**Re: `src/lib/litegraph/src/utils/widget.ts:25`**

issue: `evaluateInput` accepts `Infinity` as a valid result.

`1/0` passes the regex, `eval("1/0")` returns `Infinity`, and `Number(Infinity)` is not `NaN` so it passes the guard. When `max` is undefined, `Infinity` is stored as the widget value.

Consider: `if (!Number.isFinite(newValue)) return undefined`

src/scripts/changeTracker.ts
```diff
@@ -306,14 +307,21 @@ export class ChangeTracker {
       title: string,
       value: string | number,
       callback: (v: string) => void,
-      event: CanvasPointerEvent
+      event: CanvasPointerEvent,
```
christian-byrne, line 319: suggestion (non-blocking): The 5th-argument forwarding fix is the root-cause bugfix for this PR but has no regression test. A future refactor could silently re-drop the argument. Consider adding a unit test that verifies `multilineOrOptions` is forwarded through the wrapped `prompt` call.

src/lib/litegraph/src/widgets/NumberWidget.test.ts
```diff
@@ -0,0 +1,180 @@
...
+  })
+
+  it('passes number input options to canvas.prompt for center click', () => {
+    const canvas = createMockCanvas()
+    const e = createMockEvent(100)
+    const widget = createNumberWidget(node)
+
+    widget.onClick({ e, node, canvas })
+
+    expect(canvas.prompt).toHaveBeenCalledWith(
```
christian-byrne, line 54: suggestion (non-blocking): These tests verify `canvas.prompt` is called with the right args but never invoke the captured callback to exercise `evaluateInput` + `setValue`. The actual evaluation path is untested. Consider extracting the callback from `canvas.prompt.mock.calls[0][2]`, invoking it with a test value, and asserting `widget.value`.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -8489,6 +8517,26 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
       } else {
         // on node
         menu_info = this.getNodeMenuOptions(node)
+
+        const widget = node.getWidgetOnPos(event.canvasX, event.canvasY)
+        if (
+          widget &&
+          'getContextMenuOptions' in widget &&
```
christian-byrne, line 8705: suggestion (non-blocking): `getContextMenuOptions` is duck-typed via `'getContextMenuOptions' in widget` + an inline type assertion. Consider declaring it as an optional method on `IBaseWidget` so the contract is discoverable and type-safe.

src/lib/litegraph/src/widgets/NumberWidget.ts
```diff
@@ -71,14 +87,41 @@ export class NumberWidget
         const parsed = evaluateInput(v)
         if (parsed !== undefined) this.setValue(parsed, { e, node, canvas })
```
christian-byrne, line 88: nitpick (non-blocking): The `evaluateInput` -> `setValue` callback is duplicated between `onClick` (L87-88) and `getContextMenuOptions` (L112-115). Consider extracting a shared helper method.

src/lib/litegraph/src/widgets/NumberWidget.ts
```diff
@@ -1,9 +1,25 @@
...
+ * Returns the value if it's within the safe integer range for
+ * `<input type="number">` step arithmetic, otherwise `undefined`.
+ *
+ * The browser's stepUp/stepDown algorithm computes `value - min` internally.
+ * When `min` exceeds `Number.MAX_SAFE_INTEGER`, that subtraction silently
+ * loses the small delta (e.g. ±1), causing the browser to snap back to the
+ * previous value.  Omitting the attribute avoids the broken constraint.
+ */
+function safeMinMax(v: number | undefined): number | undefined {
+  if (v == null) return undefined
```
christian-byrne, line 19: nitpick (non-blocking): `safeMinMax` doesn't guard against `NaN` -- `Math.abs(NaN) > MAX_SAFE_INTEGER` is `false`, so `NaN` passes through. Consider using `!Number.isFinite(v)` instead.

src/scripts/changeTracker.ts
dante01yoon: Agreed this would be valuable. Will track as a follow-up.

@coderabbitai create issue for follow up regression test

src/lib/litegraph/src/widgets/NumberWidget.test.ts
dante01yoon: Fair point — current tests only assert prompt args, not the callback execution path. Will address in a follow-up.

src/lib/litegraph/src/LGraphCanvas.ts
dante01yoon: Agree, adding it to IBaseWidget would be cleaner. Tracking as follow-up to keep this PR scoped.

src/lib/litegraph/src/widgets/NumberWidget.ts
dante01yoon: Acknowledged. The duplication is minor (2 lines) but extracting a helper is reasonable. Will consider in a follow-up.

src/lib/litegraph/src/widgets/NumberWidget.ts
dante01yoon: `safeMinMax` only receives this.options.min/this.options.max which are typed as number | undefined from node definitions — `NaN` shouldn't reach here in practice. But switching to `!Number.isFinite(v)` is strictly safer and a one-line change. Want me to include it in this PR or leave for follow-up?

christian-byrne: note (non-blocking): leave for follow-up, this is fine as-is for the current PR scope

christian-byrne (dismissed):
all non-blocking feedback acknowledged by author, follow-up issues created for type guard (#9390), localization (#9389), and regression test (#9418). lgtm

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -7025,6 +7047,12 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
       if (e.key == 'Escape') {
         // ESC
         dialog.close()
+      } else if (e.key === '/' && this.type === 'number') {
+        // Switch to text mode for expression editing
+        this.type = 'text'
+        this.removeAttribute('min')
```
christian-byrne, line 7160: suggestion (non-blocking): the `/` key toggle and "Edit Expression" context menu are great, but users who are used to typing expressions like `1+2*3` directly into number fields will hit a `type="number"` input that silently rejects their keystrokes with no feedback. two ideas for discoverability:

1. **placeholder text** - set `placeholder="Press / for expressions"` on the number input so users see the hint immediately when the field is empty

2. **toast on math symbol keypress** - add a keydown handler that detects common expression characters (`+`, `*`, `(`, `)`) being typed into a `type="number"` input and shows a brief toast/notification like "Press / to switch to expression mode" - this catches users in the moment they need it

either of these (or both) would bridge the gap between the improved number UX and the existing expression workflow. could be addressed here or in the planned follow-up PR with the fx toggle button.

dante01yoon: @coderabbitai add issue for discuss/applying suggestion

dante01yoon: try again. create issue

dante01yoon: @coderabbitai add issue for discuss/applying suggestion

PR #8987, opened 2026-02-19
V2 Node Search (+ hidden Node Library changes)

### LINKED ISSUE #201: [DevTask] Improve the filter condition selection experience
Currently if you click the text box, it shows you fuzzy matching results based on current input value in the searchbox. You only get the full list of options if you have an empty searchbox or you click the down arrow button.

image

pythongosssss:
## Summary

Redesigned node search with categories

## Changes

- **What**: Adds a v2 search component, leaving the existing implementation untouched
- It also brings onboard the incomplete node library & preview changes, disabled and behind a hidden setting
- **Breaking**: Changes the 'default' value of the node search setting to v2, adding v1 (legacy) as an option

https://github.com/user-attachments/assets/2ab797df-58f0-48e8-8b20-2a1809e3735f

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8987-V2-Node-Search-hidden-Node-Library-changes-30c6d73d36508160902bcb92553f147c) by [Unito](https://www.unito.io)


src/components/searchbox/v2/NodeSearchContent.vue
```diff
@@ -0,0 +1,287 @@
...
+    selectedCategory.value = category
+    searchQuery.value = ''
+  }
+})
+
+function matchesFilters(node: ComfyNodeDefImpl): boolean {
+  return filters.every(({ filterDef, value }) => filterDef.matches(node, value))
+}
+
+const displayedResults = computed<ComfyNodeDefImpl[]>(() => {
```
christian-byrne, line 200: 🟠 **[M1] Major — Search results list has no virtualization**

The sidebar uses `TreeVirtualizer` for large lists, but the search popover results use a plain `v-for` with no limit. Selecting the "custom" category could render hundreds/thousands of nodes without virtualization, causing frame drops on lower-end machines.

**Suggestion:** Add virtualization (like the sidebar does), or cap displayed results with a limit + "show more" pattern.

src/composables/node/useNodeDragToCanvas.ts
```diff
@@ -0,0 +1,112 @@
...
+import { useLitegraphService } from '@/services/litegraphService'
+import type { ComfyNodeDefImpl } from '@/stores/nodeDefStore'
+
+type DragMode = 'click' | 'native'
+
+const isDragging = ref(false)
+const draggedNode = shallowRef<ComfyNodeDefImpl | null>(null)
+const cursorPosition = ref({ x: 0, y: 0 })
+const dragMode = ref<DragMode>('click')
+let listenersSetup = false
```
christian-byrne, line 13: 🟠 **[M2] Major — Module-level mutable state with HMR/singleton hazards**

`listenersSetup` is a plain `let` boolean that never resets on HMR — if the module is hot-reloaded, listeners are orphaned and `setupGlobalListeners()` silently becomes a no-op because the guard is still `true` while old listeners reference stale closures.

Also, if two component instances mount simultaneously, the second `setupGlobalListeners()` is a no-op and the first `cleanupGlobalListeners()` would remove listeners the second still needs.

**Suggestion:** Use an `AbortController` for listener management, or at minimum a reference counter instead of a boolean flag.

christian-byrne (commented):
## Code Review Summary — PR #8987: V2 Node Search

**Reviewer:** christian-byrne (via automated multi-agent review)

| Severity | Count |
|----------|-------|
| Critical | 1 |
| Major | 7 |
| Minor | 6 |
| Nitpick | 2 |

**Overall:** Well-structured PR with clean V1/V2 separation, good composable extraction, strong test coverage (4,746 tests pass ✅). The main concerns are around performance (no virtualization in search results, Fuse index rebuild cost) and a few code quality items.

---

### 🔴 [C1] Critical — `NodeSearchService` re-instantiated on every `visibleNodeDefs` change

**File:** `src/stores/nodeDefStore.ts:377` (not in diff — commenting here in body)

```ts
const nodeSearchService = computed(
  () => new NodeSearchService(visibleNodeDefs.value)
)
```

This computed rebuilds **5 Fuse indices** (main search + 4 filter indices) every time `visibleNodeDefs` changes — triggered by toggling deprecated/experimental/dev settings, any `nodeDefFilters` change, or new nodeDefs being added. Each reconstruction is O(n log n) × 5.

**Suggestion:** Cache the service instance with `shallowRef` and only rebuild when data identity actually changes, or debounce the reconstruction.

---

### ✅ Verified Clean
- Deleted `draftTypes.ts` has no dangling imports
- `main_category` schema addition is backwards compatible (`.optional()` with fallback)
- Settings properly support V1/V2 coexistence
- Browser test fixture (`ComfyNodeSearchBoxV2.ts`) is well-designed
- All 4,746 unit tests pass

src/components/searchbox/v2/NodeSearchContent.vue
```diff
@@ -0,0 +1,287 @@
...
+
+  return filters.length > 0 ? results.filter(matchesFilters) : results
+})
+
+const hoveredNodeDef = computed(
+  () => displayedResults.value[selectedIndex.value] ?? null
+)
+
+watchEffect(() => {
+  emit('hoverNode', hoveredNodeDef.value)
```
christian-byrne: 🟠 **[M3] Major — `hoverNode` emit fires on every keystroke**

`hoveredNodeDef` depends on `displayedResults` which changes on every search query change, so `watchEffect` fires `hoverNode` for every keystroke — potentially triggering expensive `NodePreviewCard` re-mounts.

**Suggestion:** Use `watch` with explicit comparison, or debounce the emit.

src/components/sidebar/tabs/nodeLibrary/EssentialNodeCard.test.ts
```diff
@@ -0,0 +1,185 @@
...
+      const wrapper = mountComponent(node)
+
+      await wrapper.find('div').trigger('click')
+
+      expect(wrapper.emitted('click')).toBeFalsy()
+    })
+  })
+
+  describe('drag and drop', () => {
+    it('should handle dragstart event', async () => {
```
christian-byrne: 🟠 **[M4] Major — Assertion-free tests provide false confidence**

These four tests (dragstart, dragend, mouseenter, mouseleave at lines 155-184) trigger events but assert nothing — they always pass regardless of whether the handlers actually work.

**Suggestion:** Assert that the mocked `startDrag`/`handleNativeDrop` functions were called with expected arguments, consistent with how `TreeExplorerV2Node.test.ts` does it.

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -32,8 +33,15 @@ export function appendJsonExt(path: string) {
   return path
 }
 
-export function highlightQuery(text: string, query: string) {
+export function highlightQuery(
+  text: string,
+  query: string,
+  sanitize: boolean = true
```
christian-byrne, line 39: 🟠 **[M5] Major — No tests for sanitization behavior**

The new `sanitize` parameter is a good addition, but there are no tests covering the sanitize=true vs sanitize=false paths. Also, the `query` string itself is never sanitized — only `text` is — which could be an XSS vector if query comes from untrusted input and the result is used with `v-html`.

**Suggestion:** Add tests covering both sanitization paths. Consider whether `query` should also be sanitized.

src/services/litegraphService.ts
```diff
@@ -843,8 +844,9 @@ export const useLitegraphService = () => {
 
   function addNodeOnGraph(
     nodeDef: ComfyNodeDefV1 | ComfyNodeDefV2,
-    options: Record<string, unknown> & { pos?: Point } = {}
-  ): LGraphNode {
+    options: Record<string, unknown> & { pos?: Point } = {},
+    addOptions?: GraphAddOptions
+  ): LGraphNode | null {
```
christian-byrne, line 849: 🟠 **[M6] Major — Return type changed to `LGraphNode | null` — callers may not handle null**

This is a breaking interface change. Existing callers that assume a non-null return could encounter null pointer errors at runtime. Worth auditing all callers to ensure they handle the null case.

src/components/node/NodePreviewCard.vue
```diff
@@ -0,0 +1,142 @@
...
+}>()
+
+const previewContainerRef = ref<HTMLElement>()
+const previewWrapperRef = ref<HTMLElement>()
+
+useResizeObserver(previewWrapperRef, (entries) => {
+  const entry = entries[0]
+  if (entry && previewContainerRef.value) {
+    const scaledHeight = entry.contentRect.height * SCALE_FACTOR
+    previewContainerRef.value.style.height = `${scaledHeight + 24}px`
```
christian-byrne: 🟡 **[m1] Minor — Magic number `+ 24` for padding offset**

The `+ 24` corresponds to `p-3` (12px × 2) but is undocumented and will silently break if padding changes.

**Suggestion:** Derive from a constant or compute from the actual padding value.

src/composables/node/useNodePreviewAndDrag.ts
```diff
@@ -0,0 +1,147 @@
...
+    nodePreviewStyle.value = {
+      position: 'fixed',
+      top: `${top}px`,
+      left: `${left}px`,
+      pointerEvents: 'none',
+      zIndex: 1000
+    }
+    isHovered.value = true
+
+    requestAnimationFrame(() => {
```
christian-byrne, line 75: 🟡 **[m2] Minor — Preview position one-frame flash**

The preview is first positioned at `rect.top`, then repositioned in a `requestAnimationFrame` callback after measuring height. This causes a one-frame visual jump.

**Suggestion:** Set `opacity: 0` initially and reveal after rAF positioning, or measure in a hidden state first.

src/stores/workspace/searchBoxStore.ts
```diff
@@ -10,10 +10,14 @@ export const useSearchBoxStore = defineStore('searchBox', () => {
   const settingStore = useSettingStore()
   const { x, y } = useMouse()
 
-  const newSearchBoxEnabled = computed(
+  const useSearchBoxV2 = computed(
     () => settingStore.get('Comfy.NodeSearchBoxImpl') === 'default'
```
christian-byrne, line 14: 🟡 **[m3] Minor — Feature flag naming inconsistency**

The search popover V2 is gated by `Comfy.NodeSearchBoxImpl === 'default'` while the sidebar V2 is gated by `Comfy.NodeLibrary.NewDesign` (in `useNodeLibrarySidebarTab.ts:11`). These are independent settings, so a user could run V2 search with V1 sidebar or vice versa. If they're meant to ship together, consider a single flag or document the independent toggle behavior.

src/components/searchbox/v2/NodeSearchCategorySidebar.vue
```diff
@@ -0,0 +1,132 @@
...
+      ? CATEGORY_SELECTED_CLASS
+      : CATEGORY_UNSELECTED_CLASS
+  )
+}
+
+const selectedCollapsed = ref(false)
+
+function selectCategory(categoryId: string) {
+  if (selectedCategory.value === categoryId) {
+    selectedCollapsed.value = !selectedCollapsed.value
```
christian-byrne, line 126: 🟡 **[m4] Minor — `selectedCollapsed` UX ambiguity**

Clicking an already-selected category toggles `selectedCollapsed` but doesn't change `selectedCategory`, so the results list stays the same while the tree children collapse/expand. This dual-purpose click (collapse tree + keep selection) may confuse users who expect clicking an active category to deselect it.

src/components/searchbox/v2/NodeSearchListItem.vue
```diff
@@ -0,0 +1,135 @@
...
+        v-else-if="showCategory"
+        class="option-category truncate text-sm font-light text-muted"
+      >
+        {{ nodeDef.category.replaceAll('/', ' > ') }}
+      </div>
+    </div>
+    <div v-if="!showDescription" class="flex items-center gap-1">
+      <span
+        v-if="nodeDef.deprecated"
+        class="rounded bg-red-500/20 px-1.5 py-0.5 text-xs text-red-400"
```
christian-byrne, line 49: 🟡 **[m5] Minor — Raw Tailwind color instead of semantic token**

`bg-red-500/20 text-red-400` are hardcoded color values. If the app uses non-neutral-based themes or semantic color tokens, these won't adapt. Consider using semantic theme classes for consistency with the rest of the design system.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -3620,6 +3620,9 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
         oldValue: true,
         newValue: false
       })
+
+      this.state.selectionChanged = true
+      this.onSelectionChange?.(this.selected_nodes)
```
christian-byrne, line 3625: 🟡 **[m6] Minor — New `onSelectionChange` on ghost placement**

This fires a selection change event when a ghost node is confirmed (placed). The behavior is correct, but existing listeners expecting selection changes only from user click/drag may fire unexpectedly. Worth verifying no existing code reacts poorly to this additional notification.

src/components/sidebar/tabs/nodeLibrary/EssentialNodeCard.vue
```diff
@@ -0,0 +1,81 @@
...
+import { computed, inject } from 'vue'
+
+import NodePreviewCard from '@/components/node/NodePreviewCard.vue'
+import { SidebarContainerKey } from '@/components/sidebar/tabs/SidebarTabTemplate.vue'
+import { useNodePreviewAndDrag } from '@/composables/node/useNodePreviewAndDrag'
+import type { ComfyNodeDefImpl } from '@/stores/nodeDefStore'
+import type { RenderedTreeExplorerNode } from '@/types/treeExplorerTypes'
+import { cn } from '@/utils/tailwindUtil'
+
+const props = defineProps<{
```
christian-byrne: 💬 **[N1] Nitpick — `const props = defineProps` instead of destructuring**

This is the only V2 component that uses `const props = defineProps()` instead of reactive destructuring (`const { ... } = defineProps<...>()`). The rest of the PR correctly uses the Vue 3.5 destructured pattern.

christian-byrne (approved):

PR #8992, opened 2026-02-20
feat: add gradient-slider widget for FLOAT inputs

jtydhr88:
## Summary
Add a new 'gradient-slider' display mode for FLOAT widget inputs. Nodes can specify gradient_stops (color stop arrays) to render a colored gradient track behind the slider thumb, useful for color adjustment parameters like hue, saturation, brightness, etc.

- GradientSlider.vue: reusable Reka UI-based gradient slider component
- GradientSliderWidget.ts: litegraph canvas-mode fallback rendering
- WidgetInputNumberGradientSlider.vue: Vue node widget integration
- Schema, registry, and type updates for gradient-slider support

this is prerequisite for color correct and balance

BE changes https://github.com/Comfy-Org/ComfyUI/pull/12536

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8992-feat-add-gradient-slider-widget-for-FLOAT-inputs-30d6d73d36508199b3e8db6a0c213ab4) by [Unito](https://www.unito.io)


src/lib/litegraph/src/interfaces.ts
```diff
@@ -48,6 +48,13 @@ export type SharedIntersection<T1, T2> = {
 
 export type CanvasColour = string | CanvasGradient | CanvasPattern
 
+export type ColorStop = readonly [
```
DrJKL: I like that it's a tuple, but can we do something like:
```ts
interface ColorStop {
  readonly offset: number
  readonly color: readonly [r: number, g: number, b: number]
}
```

DrJKL: Might also want some way to indicate that this should be `[0, 1]`, as much as I hate extra documentation ;-)

src/components/gradientslider/gradients.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function stopsToGradient(stops: ColorStop[]): string {
+  if (!stops.length) return 'transparent'
+  const colors = stops.map(
+    ([offset, r, g, b]) => `rgb(${r},${g},${b}) ${offset * 100}%`
+  )
+  return `linear-gradient(to right, ${colors.join(', ')})`
+}
+
+export function interpolateStops(stops: ColorStop[], t: number): string {
```
DrJKL, line 11: Probably not worth pre-optimizing for, but since this is in the computed and could be called a lot, it's a good target for memoizing or otherwise pre-computing and caching the ranges.

src/components/gradientslider/GradientSlider.vue
```diff
@@ -0,0 +1,90 @@
...
+const pressed = ref(false)
+</script>
+
+<template>
+  <SliderRoot
+    v-model="sliderValue"
+    :min="min"
+    :max="max"
+    :step="step"
+    :disabled="disabled"
```
DrJKL: These can use the shorthand binding.
```suggestion
    :min
    :max
    :step
    :disabled
```

src/components/gradientslider/GradientSlider.vue
```diff
@@ -0,0 +1,90 @@
...
+    :class="
+      cn(
+        'relative flex w-full touch-none items-center select-none',
+        'data-[disabled]:opacity-50'
```
DrJKL, line 57: Can you use the `disabled:` directly? https://tailwindcss.com/docs/hover-focus-and-other-states#required-and-disabled

src/components/gradientslider/GradientSlider.vue
```diff
@@ -0,0 +1,90 @@
...
+    >
+      <SliderThumb
+        :class="
+          cn(
+            'block size-4 shrink-0 cursor-grab rounded-full shadow-md ring-1 ring-black/25',
+            'transition-[color,box-shadow,background-color]',
+            'before:absolute before:-inset-1.5 before:block before:rounded-full before:bg-transparent',
+            'hover:ring-2 hover:ring-black/40 focus-visible:ring-2 focus-visible:ring-black/40 focus-visible:outline-hidden',
+            'disabled:pointer-events-none disabled:opacity-50',
+            { 'cursor-grabbing': pressed }
```
DrJKL: Just double checking, the classes should be merged, so if any of these are already on the Thumb you can omit them here.

src/components/gradientslider/GradientSlider.vue
```diff
@@ -0,0 +1,90 @@
...
+          )
+        "
+        :style="{ backgroundColor: thumbColor, top: '50%' }"
```
DrJKL: Top could be a utility class. Can backgroundColor go through a CSS variable? That part, we already have a mix in the codebase so either is fine.

src/components/gradientslider/GradientSlider.vue
```diff
@@ -0,0 +1,90 @@
...
+        :aria-label="ariaLabel"
```
DrJKL:
```suggestion
        :aria-label
```

src/lib/litegraph/src/widgets/GradientSliderWidget.ts
```diff
@@ -0,0 +1,85 @@
...
+  extends BaseWidget<IGradientSliderWidget>
+  implements IGradientSliderWidget
+{
+  override type = 'gradientslider' as const
+
+  override drawWidget(
+    ctx: CanvasRenderingContext2D,
+    { width, showText = true }: DrawWidgetOptions
+  ) {
+    const { fillStyle, strokeStyle, textAlign } = ctx
```
DrJKL: @coderabbitai Add a rule to _always_ use `.save()` and `.restore()` for the state when using a `CanvasRenderingContext2D`

DrJKL: @coderabbitai also open an issue to audit and cleanup existing uses.

src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumber.vue
```diff
@@ -19,25 +20,24 @@ const modelValue = defineModel<number>({ default: 0 })
 const hasControlAfterGenerate = computed(() => {
   return !!props.widget.controlWidget
 })
+
+const widgetComponent = computed(() => {
+  if (props.widget.type === 'gradientslider')
+    return WidgetInputNumberGradientSlider
+  if (props.widget.type === 'slider') return WidgetInputNumberSlider
+  return WidgetInputNumberInput
```
DrJKL: One more and we'll probably want a `switch`

src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumber.vue
```diff
@@ -19,25 +20,24 @@ const modelValue = defineModel<number>({ default: 0 })
...
+})
 </script>
 
 <template>
   <WidgetWithControl
     v-if="hasControlAfterGenerate"
     v-model="modelValue"
     :widget="widget as SimplifiedControlWidget<number>"
```
DrJKL: Could you double check that this contract is satisfied?
We should find a way to not have to assert the type here.

src/renderer/extensions/vueNodes/widgets/components/WidgetInputNumberGradientSlider.vue
```diff
@@ -0,0 +1,91 @@
...
+  const stops = widget.options?.gradient_stops
+  if (stops && stops.length >= 2) return stops
+  return DEFAULT_GRADIENT_STOPS
+})
+
+const filteredProps = computed(() =>
+  filterWidgetProps(widget.options, STANDARD_EXCLUDED_PROPS)
+)
+
+const p = widget.options?.precision
```
DrJKL: Unlikely to be an issue, but some of these should probably be `computed` since they incorporate values from props.

src/components/gradientslider/GradientSlider.vue
jtydhr88: Reka UI is headless/unstyled — SliderThumb renders a bare <span> with no default classes, so all classes here are ours.

src/components/gradientslider/GradientSlider.vue
jtydhr88: SliderRoot renders as a <span>, not a form element — :disabled pseudo-class doesn't apply. Reka UI sets the data-disabled attribute instead, so data-[disabled]: is the correct selector here.

DrJKL (approved):

PR #8997, opened 2026-02-20
fix: widgets disappear when reordering in SubgraphEditor properties panel

christian-byrne:
## Summary

Fix widgets disappearing when reordering in the SubgraphEditor properties panel (opened from the floating toolbox/toolbar).

## Changes

- **What**: Change `SubgraphEditor.vue`'s `applyNewItemsOrder` to reorder `proxyWidgets.value` directly (raw `[nodeId, widgetName]` tuples) instead of going through `activeWidgets` (a lossy computed that resolves widget items then converts back, dropping entries when nodes/widgets can't be resolved). This matches how `TabSubgraphInputs.vue` already handles reordering correctly.

## Review Focus

The root cause: `activeWidgets.get()` uses `flatMap(mapWidgets)` which returns `[]` for unresolvable nodes, silently dropping entries. The setter then writes this shorter array back. By reordering `proxyWidgets` directly, we skip the lossy resolve→convert round-trip entirely.

This was a day-one bug from PR #6952 (Dec 2025), not a recent regression.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8997-fix-widgets-disappear-when-reordering-in-SubgraphEditor-properties-panel-30d6d73d365081478fb8f02977681b77) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/rightSidePanel/subgraph/SubgraphEditor.vue
+++ b/src/components/rightSidePanel/subgraph/SubgraphEditor.vue
@@ -196,7 +196,7 @@ function setDraggableState() {
     '.draggable-item'
   )
   draggableList.value.applyNewItemsOrder = function () {
-    const reorderedItems = []
+    const reorderedItems: HTMLElement[] = []
 
     let oldPosition = -1
     this.getAllItems().forEach((item, index) => {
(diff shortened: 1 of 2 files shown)


AustinMroz (commented):
In the unhappy case, this change makes it so that we're doing blind reordering on two lists that are known to be desynced.
re-order_00002

Prior to this PR, reordering widgets would result in the node reflecting the state of the reordered list. After this PR, the node will have garbage state unrelated to the the drag operation.

If the goal is for disconnected inputs to not be pruned, then I believe a more thorough rewrite will be required to make activeWidgets instead be an augmented form of proxyWidgets (which may have additional filterable information like node title). I do not think the proposed change here brings us in the right direction.

(Small PSA: I've got [an in progress branch](https://github.com/Comfy-Org/ComfyUI_frontend/tree/austin/draggablelist-component) to move usage of draggleList from vue into a component. Branches will conflict, but have no overlap in functionality.)

christian-byrne:
Closing this — the underlying code has been substantially rewritten since this PR was opened. The `proxyWidgets` module no longer exists (replaced by `promotionStore`), and `SubgraphEditor.vue` was moved to `src/components/rightSidePanel/subgraph/`. Combined with @AustinMroz's review feedback that a more thorough rewrite of `activeWidgets` is needed (rather than blind reordering on desynced lists), and his in-progress `draggablelist-component` branch, rebasing this isn't worthwhile.

If the bug still reproduces on current `main`, it'll need a fresh fix targeting the new `promotionStore.setPromotions()` codepath.

PR #8998, opened 2026-02-20
fix: space bar panning over Vue nodes in standard nav mode

### LINKED ISSUE #7806: Spacebar panning in Nodes 2.0

christian-byrne:
## Summary

Fix spacebar panning over Vue nodes in standard navigation mode, including while dragging a Vue-node link.

## Changes

- **What**: Dispatch a typed `litegraph:read-only-changed` event when `LGraphCanvas.read_only` changes, mirror it in `canvasStore.isReadOnly`, and use that reactive state to disable Vue node pointer handling while the canvas owns the pan gesture.
- Route canvas lock commands and app-mode writes through the `read_only` setter so they cannot bypass reactive synchronization.
- Preserve an active Vue link drag while spacebar panning by letting `LGraphCanvas` enter dragging mode and pausing Vue link pointer processing until panning ends.
- Add focused unit coverage plus browser regressions for space-dragging directly on a Vue node and space-panning during a link drag.

## Review Focus

- The event fires only when read-only state changes; cursor refresh behavior remains unchanged.
- `canvasStore.isReadOnly` is the reactive bridge because the `LGraphCanvas` instance itself is held in a `shallowRef`.
- Link dragging resumes after the pan and still connects to the intended slot.

Fixes #7806

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8998-fix-space-bar-panning-over-Vue-nodes-in-standard-nav-mode-30d6d73d36508185a25be9ce89f5b86e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/core/canvas/canvasStore.ts
+++ b/src/renderer/core/canvas/canvasStore.ts
@@ -56,6 +56,7 @@ export const useCanvasStore = defineStore('canvas', () => {
       setMode(val ? 'app' : 'graph')
     }
   })
+  const isReadOnly = ref(false)
 
   // Set up scale synchronization when canvas is available
   let originalOnChanged: ((scale: number, offset: Point) => void) | undefined =
@@ -139,6 +140,11 @@ export const useCanvasStore = defineStore('canvas', () => {
         }
       )
 
+      isReadOnly.value = newCanvas.read_only
+      newCanvas.onReadOnlyChanged = (value: boolean) => {
+        isReadOnly.value = value
+      }
+
       useEventListener(
         newCanvas.canvas,
         'litegraph:set-graph',
@@ -184,6 +190,7 @@ export const useCanvasStore = defineStore('canvas', () => {
     rerouteSelected,
     appScalePercentage,
     linearMode,
+    isReadOnly,
     updateSelectedItems,
     getCanvas,
     setAppZoomFromPercentage,
--- a/src/renderer/core/canvas/canvasStore.ts
+++ b/src/renderer/core/canvas/canvasStore.ts
@@ -141,9 +141,14 @@ export const useCanvasStore = defineStore('canvas', () => {
       )
 
       isReadOnly.value = newCanvas.read_only
-      newCanvas.onReadOnlyChanged = (value: boolean) => {
-        isReadOnly.value = value
-      }
+
+      useEventListener(
+        newCanvas.canvas,
+        'litegraph:read-only-changed',
+        (event: CustomEvent<{ readOnly: boolean }>) => {
+          isReadOnly.value = event.detail.readOnly
+        }
+      )
 
       useEventListener(
         newCanvas.canvas,
--- a/src/renderer/core/canvas/useCanvasInteractions.ts
+++ b/src/renderer/core/canvas/useCanvasInteractions.ts
@@ -27,9 +27,7 @@ export function useCanvasInteractions() {
    * Whether Vue node components should handle pointer events.
    * Returns false when canvas is in read-only/panning mode (e.g., space key held for panning).
    */
-  const shouldHandleNodePointerEvents = computed(
-    () => !(canvasStore.canvas?.read_only ?? false)
-  )
+  const shouldHandleNodePointerEvents = computed(() => !canvasStore.isReadOnly)
 
   /**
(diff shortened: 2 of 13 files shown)


src/renderer/core/canvas/canvasStore.ts
```diff
@@ -115,6 +116,11 @@ export const useCanvasStore = defineStore('canvas', () => {
   whenever(
     () => canvas.value,
     (newCanvas) => {
+      isReadOnly.value = newCanvas.read_only
+      newCanvas.onReadOnlyChanged = (value: boolean) => {
+        isReadOnly.value = value
+      }
```
christian-byrne: This cannot be the best way to do this.

src/renderer/core/canvas/useCanvasInteractions.ts
```diff
@@ -22,9 +22,7 @@ export function useCanvasInteractions() {
    * Whether Vue node components should handle pointer events.
    * Returns false when canvas is in read-only/panning mode (e.g., space key held for panning).
    */
-  const shouldHandleNodePointerEvents = computed(
-    () => !(canvasStore.canvas?.read_only ?? false)
-  )
+  const shouldHandleNodePointerEvents = computed(() => !canvasStore.isReadOnly)
```
dante01yoon, line 30: Issue: shouldHandleNodePointerEvents was computed from canvasStore.canvas?.read_only, but canvas is a shallowRef — Vue can't track changes to properties on the canvas object itself. When the space key toggles read_only, the computed never re-evaluated, so Vue nodes kept intercepting pointer events instead of allowing canvas panning.

christian-byrne:
Revived after rebase onto current main:

- Rebased onto `main` (resolved conflicts in `canvasStore.ts` from #8856 `useAppMode` integration, and merged with new `canvasStore.test.ts` on main).
- Addressed CodeRabbit nitpick on `LGraphCanvas.ts:401-407` — guarded `onReadOnlyChanged` callback so it only fires when `read_only` actually changes, avoiding redundant reactive updates on repeated space keydown.
- Other CodeRabbit comments referred to `handleWheel` forwarding and `useNodePointerInteractions.test.ts` — those changes are no longer in this PR (scope was narrowed to the reactivity bridge), so those threads are obsolete.
- Local: `canvasStore.test.ts` (6 tests), `useCanvasInteractions.test.ts` (10 tests), all `LGraphCanvas.*.test.ts` (50 tests) — pass. `pnpm typecheck` and `pnpm format` clean. Marked ready for review.

src/renderer/core/canvas/canvasStore.ts
```diff
@@ -131,6 +132,11 @@ export const useCanvasStore = defineStore('canvas', () => {
...
```
christian-byrne: I don't understand --

1. Why not use events?
2. Why not, even better, follow the existing pattern of how this is done with canvas properties or reactive values, especially in canvasStore?

benceruleanlu:
Looking now, sorry for the delay

benceruleanlu (approved):

PR #7094, opened 2025-12-02
Fix skeleton loaders for Image/Video Previews

Myestery:
This pull request refines the loading and error handling logic for both the `VideoPreview.vue` and `ImagePreview.vue` components. The main improvements include making the loading skeleton more accurate and visually consistent, updating how loading and error states are managed when URLs change, and ensuring that the main media elements are hidden while loading. These changes enhance the user experience by providing clearer feedback during media load operations.

**Loading and error state improvements:**

* The loading skeleton in both `VideoPreview.vue` and `ImagePreview.vue` now only appears when loading and no error is present, with updated styling and fixed dimensions for better consistency. (`VideoPreview.vue` [[1]](diffhunk://#diff-17b5c19b4628f22e45570b66a85ed1fc16e931dd368fe420584d487e522ab8aaL29-R41) `ImagePreview.vue` [[2]](diffhunk://#diff-0c0b17c5c68464e0284398ad42b823509d414c9cf297f3bc2aa2b00e0f9c2015L29-R48)
* The main video and image elements are now hidden (using the `invisible` class) while loading, preventing display glitches before the media is ready. (`VideoPreview.vue` [[1]](diffhunk://#diff-17b5c19b4628f22e45570b66a85ed1fc16e931dd368fe420584d487e522ab8aaL29-R41) `ImagePreview.vue` [[2]](diffhunk://#diff-0c0b17c5c68464e0284398ad42b823509d414c9cf297f3bc2aa2b00e0f9c2015L29-R48)
* The loading state (`isLoading`) is now set to `true` whenever new URLs are provided, and reset appropriately when navigating between media items, ensuring accurate feedback to the user. (`VideoPreview.vue` [[1]](diffhunk://#diff-17b5c19b4628f22e45570b66a85ed1fc16e931dd368fe420584d487e522ab8aaL145-R152) `ImagePreview.vue` [[2]](diffhunk://#diff-0c0b17c5c68464e0284398ad42b823509d414c9cf297f3bc2aa2b00e0f9c2015L164-R176) [[3]](diffhunk://#diff-0c0b17c5c68464e0284398ad42b823509d414c9cf297f3bc2aa2b00e0f9c2015L224-R236)

**Code consistency and maintainability:**

[...]

PR #8999, opened 2026-02-20
fix: prevent persistent loading state when cycling batches with identical URLs

christian-byrne:
## Summary

Fix persistent loading/skeleton state when cycling through batch images or videos that share the same URL (common on Cloud).

## Changes

- **What**: In `setCurrentIndex()` for both `ImagePreview.vue` and `VideoPreview.vue`, only start the loader when the target URL differs from the current URL. When batch items share the same URL, the browser doesn't fire a new `load`/`loadeddata` event since `src` didn't change, so the loader was never dismissed.
- Also fixes `VideoPreview.vue` navigation dots using hardcoded `bg-white` instead of semantic `bg-base-foreground` tokens.

## Review Focus

This bug has regressed 3+ times (PRs #6521, #7094, #8366). The regression tests specifically target the root cause — cycling through identical URLs — to prevent future reintroduction.

Fixes https://www.notion.so/comfy-org/Bug-Cycling-through-image-batches-results-in-persistent-loading-state-on-Cloud-30c6d73d3650816e9738d5dbea52c47d

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-8999-fix-prevent-persistent-loading-state-when-cycling-batches-with-identical-URLs-30d6d73d36508180831edbaf8ad8ad48) by [Unito](https://www.unito.io)


simula-r (approved):
Nice. LGTM.

PR #9000, opened 2026-02-20
fix: make cloud input-mapping widget spec-driven instead of node-name-driven

christian-byrne:
## Summary

Replace hardcoded `NODE_MEDIA_TYPE_MAP` (keyed by core node class names) with spec-driven media type detection from `ComboInputSpec` upload flags, fixing custom nodes like VHS `LoadVideoUpload` not getting the cloud input-mapping widget.

## Changes

- **What**: Added `getSpecMediaType()` helper that reads `video_upload`/`image_upload`/`audio_upload` flags from `ComboInputSpec`. Replaced `NODE_PLACEHOLDER_MAP` with `MEDIA_TYPE_PLACEHOLDER_MAP` keyed by media type. Updated `createInputMappingWidget` to accept `mediaType` as a parameter. Removed both hardcoded maps.
- **Dependencies**: VHS side needs `{"video_upload": True}` in its Python `INPUT_TYPES` (separate PR to `Kosinkadink/ComfyUI-VideoHelperSuite`).

## Review Focus

- `getSpecMediaType()` flag priority order: `video > image/animated_image > audio`. Multiple flags set simultaneously is not explicitly handled (returns first match).
- Asset filtering in the Proxy still uses `getMediaTypeFromFilename(asset.name)` — filename-based, not mime-type-based. Kept consistent with prior behavior.

Fixes COM-15053

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9000-fix-make-cloud-input-mapping-widget-spec-driven-instead-of-node-name-driven-30d6d73d3650815995c7cbf71fd2d127) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/composables/useComboWidget.test.ts
+++ b/src/renderer/extensions/vueNodes/widgets/composables/useComboWidget.test.ts
@@ -296,12 +296,24 @@ describe('useComboWidget', () => {
       'a1b2c3d4e5f6789012345678901234567890abcdef1234567890abcdef123456.jpg'
 
     it.each([
-      { nodeClass: 'LoadImage', inputName: 'image' },
-      { nodeClass: 'LoadVideo', inputName: 'video' },
-      { nodeClass: 'LoadAudio', inputName: 'audio' }
+      {
+        nodeClass: 'LoadImage',
+        inputName: 'image',
+        uploadFlag: 'image_upload'
+      },
+      {
+        nodeClass: 'LoadVideo',
+        inputName: 'video',
+        uploadFlag: 'video_upload'
+      },
+      {
+        nodeClass: 'LoadAudio',
+        inputName: 'audio',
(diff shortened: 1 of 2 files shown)


AustinMroz (commented):
The required inputSpec changes cause undesirable side effects on the node. In the case of load video, it causes
- A second upload button to be added which does not support upload progressBars
- Adds a duplicate preview on the node which does not support AdvancedPreview functionality, or node sizing code.
- Changes the file selection widget in vue mode to be a media file widget (a good change!), that also desyncs with the litegraph widget value (bad)

The duplicate button and state desynchronization can *probably* be fixed on the VHS side. Making such a change on the VHS end is better to ensure the transition can be done without an intermediate step with doubled buttons. The loss of upload progress bars would be unfortunate, but should be added to core anyways.

The duplicated previews is more problematic. I would prefer not to have to fight the frontend by forcibly removing the duplicated previews on the VHS side.

image

PR #9002, opened 2026-02-20
Add prompt_id support to progress_text WS messages

christian-byrne:
## Summary

Add frontend support for `prompt_id` in `progress_text` binary WS messages, enabling parallel workflow execution to route progress text to the correct active prompt.

Backend PR: https://github.com/Comfy-Org/ComfyUI/pull/12540

## Changes

- Advertise `supports_progress_text_metadata` in client feature flags
- Decode `prompt_id` from new binary format when flag is active
- Add optional `prompt_id` to `zProgressTextWsMessage` schema
- Filter `progress_text` events by `activePromptId` — skip messages for non-active prompts

## Deployment

Can be deployed independently in any order — feature flag negotiation ensures graceful degradation.

Part of COM-12671

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9002-Add-prompt_id-support-to-progress_text-WS-messages-30d6d73d365081acabbcdac939e0c751) by [Unito](https://www.unito.io)

## Summary by CodeRabbit

* **New Features**
  * Progress text updates can include optional metadata so richer context is available.
* **Bug Fixes / Improvements**
  * Progress updates are now filtered to show only the currently active prompt, reducing cross-talk from concurrent operations and improving update accuracy.

### DIFF
--- a/src/scripts/api.ts
+++ b/src/scripts/api.ts
@@ -631,16 +631,22 @@ export class ComfyApi extends EventTarget {
               let offset = 0
               let promptId: string | undefined
 
-              if (this.getClientFeatureFlags()?.supports_progress_text_metadata) {
+              if (
+                this.getClientFeatureFlags()?.supports_progress_text_metadata
+              ) {
                 const promptIdLength = rawView.getUint32(offset)
                 offset += 4
-                promptId = decoder3.decode(rawData.slice(offset, offset + promptIdLength))
+                promptId = decoder3.decode(
+                  rawData.slice(offset, offset + promptIdLength)
+                )
                 offset += promptIdLength
               }
 
               const nodeIdLength = rawView.getUint32(offset)
               offset += 4
-              const nodeId = decoder3.decode(rawData.slice(offset, offset + nodeIdLength))
+              const nodeId = decoder3.decode(
+                rawData.slice(offset, offset + nodeIdLength)
+              )
               offset += nodeIdLength
               const text = decoder3.decode(rawData.slice(offset))
 
--- a/src/scripts/api.ts
+++ b/src/scripts/api.ts
@@ -624,37 +624,41 @@ export class ComfyApi extends EventTarget {
           let imageMime
           switch (eventType) {
             case 3: {
-              const decoder3 = new TextDecoder()
-              const rawData = event.data.slice(4)
-              const rawView = new DataView(rawData)
+              try {
+                const decoder3 = new TextDecoder()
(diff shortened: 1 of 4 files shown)


src/scripts/api.ts
```diff
@@ -598,15 +591,40 @@ export class ComfyApi extends EventTarget {
...
+            case 3: {
+              const decoder3 = new TextDecoder()
+              const rawData = event.data.slice(4)
+              const rawView = new DataView(rawData)
+
+              let offset = 0
+              let promptId: string | undefined
+
+              if (
+                this.getClientFeatureFlags()?.supports_progress_text_metadata
```
dante01yoon: question:  Isn't it better check the server's advertised feature flags (from the /feature_flags endpoint or initial handshake) instead of the client's local config

src/scripts/api.ts
```diff
@@ -598,15 +591,40 @@ export class ComfyApi extends EventTarget {
...
           let imageMime
           switch (eventType) {
-            case 3:
-              const decoder = new TextDecoder()
-              const data = event.data.slice(4)
-              const nodeIdLength = view.getUint32(4)
+            case 3: {
+              const decoder3 = new TextDecoder()
+              const rawData = event.data.slice(4)
+              const rawView = new DataView(rawData)
```
dante01yoon: Suggestion:  If any of the issues cause format mismatches, or if a message is truncated by network, the getUint32 call can return an absurdly large number (e.g., interpreting ASCII bytes as a 32-bit int). rawData.slice(offset, offset + 1_299_345_920) won't crash but produces garbage. getUint32 on an out-of-bounds offset will throw a RangeError that bubbles up uncaught and kills the entire WS message handler — meaning all subsequent WS messages (progress, previews, execution status) are also lost.

How about wrap the case 3 block in try { ... } catch { console.warn(...) }, and/or validate lengths before slicing.

christian-byrne: good catch, added try/catch in fe41ba4f8

src/scripts/api.ts
christian-byrne: the server already checks supports_feature(sockets_metadata, target_sid, ...) on its side before deciding the binary format, so it won't send the new format unless the client advertised support for it. the client checking its own local config is belt-and-suspenders - the local flag correctly reflects its own decoding capability. using the server's advertised flags here would introduce an unnecessary dependency on the feature_flags handshake timing.

dante01yoon (dismissed):

PR #9004, opened 2026-02-20
fix: skip MatchType recalculation during graph configuration

christian-byrne:
## Summary

Fixes COM-14955: "Bug: Switch node in subgraph causes link disconnection on export"

## Problem

When a MatchType node (like Switch) inside a subgraph is configured/restored, `LGraphNode.configure()` calls `onConnectionsChange` for each input sequentially. The `withComfyMatchType` callback was running before all links were restored, seeing incomplete state and incorrectly computing types, which could cause link disconnection.

## Solution

Add early return when `app.configuringGraph` is true to defer type recalculation until after all links are restored. This pattern is already used throughout the codebase:
- `widgetInputs.ts`
- `rerouteNode.ts`
- `customWidgets.ts`

Post-configure recomputation is handled by the existing `requestAnimationFrame` callback in `applyMatchType`.

## Changes

- `src/core/graph/widgets/dynamicWidgets.ts` - Added 1 line: `if (app.configuringGraph) return`
- `src/core/graph/widgets/matchTypeConfiguring.test.ts` - New test file with 3 tests

## Testing

- All existing tests pass
- Added 3 new tests:
  - `skips type recalculation when configuringGraph is true`
  - `performs type recalculation during normal operation`
  - `connects both inputs with same type`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9004-fix-skip-MatchType-recalculation-during-graph-configuration-30d6d73d365081339088ffd8aebba107) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/graph/widgets/dynamicWidgets.ts
+++ b/src/core/graph/widgets/dynamicWidgets.ts
@@ -277,6 +277,7 @@ function withComfyMatchType(node: LGraphNode): asserts node is MatchTypeNode {
     ) {
       const input = this.inputs[slot]
       if (contype !== LiteGraph.INPUT || !this.graph || !input) return
+      if (app.configuringGraph) return
       const [matchKey, matchGroup] = Object.entries(
         this.comfyDynamic.matchType
       ).find(([, group]) => input.name in group) ?? ['', undefined]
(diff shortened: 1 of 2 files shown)


christian-byrne:
Working on ephemeral 
image

AustinMroz (approved):

PR #8957, opened 2026-02-18
feat: split job history into a dedicated sidebar tab

benceruleanlu:
## Summary

Move queue job history into a dedicated sidebar tab (gated by `Comfy.Queue.QPOV2`) and remove mixed job-history UI from the Assets sidebar so assets and job controls are separated.

## Changes

- **What**:
  - Added `JobHistorySidebarTab` with reusable job UI primitives: `JobFilterTabs`, `JobFilterActions`, `JobAssetsList`, and shared `JobHistoryActionsMenu`.
  - Added reactive `job-history` tab registration in `sidebarTabStore`; prepends above Assets when `Comfy.Queue.QPOV2` is enabled and unregisters cleanly when disabled.
  - Added debounced search to `useJobList` (filters by job title, metadata, and prompt id).
  - Extracted clear-history dialog logic to `useQueueClearHistoryDialog` and reused it from queue overlay and job history tab.
  - Removed active-job rendering and queue-clear controls from assets list/grid/tab views; assets sidebar now focuses on media assets only.
  - Removed the QPOV2 gate from `MediaAssetViewModeToggle` and updated queue/job localized copy.
  - Added and updated tests for queue overlay header actions, job filters, search filtering, sidebar tab registration, and assets sidebar behavior.

## Review Focus

- Verify QPOV2 toggle behavior:
  - `Docked Job History` menu action toggles `Comfy.Queue.QPOV2`.
  - `job-history` tab insertion/removal order and active-tab reset on removal.
- Verify behavior split between tabs:
  - Job controls (cancel/delete/view/filter/search/clear history/clear queue) live in Job History.
  - Assets sidebar loading/empty states and list/grid rendering remain correct after removing active jobs.

## Screenshots (if applicable)
image

PR #9005, opened 2026-02-20
fix: preserve paginated assets during incremental history updates

christian-byrne:
## Summary

Fix assets disappearing from Media Assets panel during infinite scroll when new jobs complete.

## Changes

- **What**: Refactor `assetsStore.updateHistory()` from full-reset to incremental prepend. On subsequent calls, only page 1 is fetched and genuinely new items are merged via `insertAssetSorted()`. A `prependedSinceLastLoad` counter tracks offset drift so `loadMoreHistory()` skips already-loaded items. Extracted `insertAssetSorted()` and `trimToMaxItems()` helpers from the monolithic `fetchHistoryAssets()`.

## Review Focus

- Offset drift arithmetic in `fetchHistoryNextPage()` — ensures pagination skips prepended items to avoid duplicates.
- The `prependedSinceLastLoad` reset is intentionally placed after successful fetch (not before) so retries preserve the drift counter.
- Coordination with #8957 (job history tab split) which touches adjacent code.

## Testing

4 new tests covering: item preservation after pagination, deduplication on overlap, offset adjustment after prepend, and drift counter preservation on fetch failure. All 31 tests pass.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9005-fix-preserve-paginated-assets-during-incremental-history-updates-30d6d73d365081d09023e2a07cab3ca5) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/assetsStore.ts
+++ b/src/stores/assetsStore.ts
@@ -161,6 +161,15 @@ export const useAssetsStore = defineStore('assets', () => {
     return true
   }
 
+  function removeHistoryItems(ids: string[]) {
+    const idSet = new Set(ids)
+    allHistoryItems.value = allHistoryItems.value.filter(
+      (item) => !idSet.has(item.id)
+    )
+    ids.forEach((id) => loadedIds.delete(id))
+    historyAssets.value = allHistoryItems.value
+  }
+
   function trimToMaxItems() {
     if (allHistoryItems.value.length > MAX_HISTORY_ITEMS) {
       const removed = allHistoryItems.value.splice(MAX_HISTORY_ITEMS)
@@ -772,6 +781,7 @@ export const useAssetsStore = defineStore('assets', () => {
     updateInputs,
     updateHistory,
     loadMoreHistory,
+    removeHistoryItems,
 
     // Input mapping helpers
     inputAssetsByFilename,
--- a/src/stores/assetsStore.ts
+++ b/src/stores/assetsStore.ts
@@ -149,9 +149,13 @@ export const useAssetsStore = defineStore('assets', () => {
     loadedIds.add(asset.id)
 
     const assetTime = new Date(asset.created_at ?? 0).getTime()
-    const insertIndex = allHistoryItems.value.findIndex(
-      (item) => new Date(item.created_at ?? 0).getTime() < assetTime
-    )
+    // Sort: newer first; ties broken by lexicographically larger id first
+    // so insertion order is stable across repeated merges.
+    const insertIndex = allHistoryItems.value.findIndex((item) => {
+      const itemTime = new Date(item.created_at ?? 0).getTime()
+      if (itemTime !== assetTime) return itemTime < assetTime
+      return item.id < asset.id
+    })
 
     if (insertIndex === -1) {
       allHistoryItems.value.push(asset)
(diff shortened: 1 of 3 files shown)


benceruleanlu:
Reviewing now

src/stores/assetsStore.ts
```diff
@@ -133,86 +134,112 @@ export const useAssetsStore = defineStore('assets', () => {
...
+   * On subsequent calls (e.g. after job completion), incrementally merges
+   * new items without clearing existing pagination state.
    */
-  const updateHistory = async () => {
+  async function updateHistory() {
     historyLoading.value = true
     historyError.value = null
     try {
-      await fetchHistoryAssets(false)
+      await fetchHistoryIncremental()
```
benceruleanlu, line 276: assetsStore.ts:170 makes updateHistory() append-only. That breaks the existing delete refresh path in useMediaAssetActions.ts:631 after api.deleteItem('history', jobId), the deleted output asset is never removed from historyAssets.

benceruleanlu (dismissed):
LGTM after the updateHistory regression is fixed

src/stores/assetsStore.ts
christian-byrne: fixed in 469c8f3 - added removeHistoryItems(ids) that evicts entries from allHistoryItems and loadedIds, called from the delete path before updateHistory() so deleted output assets are pruned from local state

src/stores/assetsStore.ts
```diff
@@ -133,86 +134,121 @@ export const useAssetsStore = defineStore('assets', () => {
...
-          continue // Skip duplicates
-        }
-        loadedIds.add(asset.id)
+  function removeHistoryItems(ids: string[]) {
+    const idSet = new Set(ids)
+    allHistoryItems.value = allHistoryItems.value.filter(
+      (item) => !idSet.has(item.id)
+    )
+    ids.forEach((id) => loadedIds.delete(id))
+    historyAssets.value = allHistoryItems.value
```
benceruleanlu, line 193: Should this helper also update the pagination cursor for the rows it removes? Right now it drops ids from the loaded window, but `historyOffset` still points at the old next page, so the next `loadMoreHistory()` can skip unseen history rows after successful deletes.

benceruleanlu (commented):
Left one inline comment about pagination bookkeeping after deletes.

src/stores/assetsStore.ts
christian-byrne: Good catch — fixed in 972b14f. `removeHistoryItems()` now decrements `historyOffset` by the number of actually-removed items so `loadMoreHistory()` requests the correct server offset. Added a test that verifies the adjusted offset after deletion.

christian-byrne (commented):
This is not the correct way to approach pagination here. Start over and present the problem and a few alternative solutions. The API should be working for us not against us and we shouldn't have to roll all our own code for this library-solved class of issues/functionalities.

christian-byrne:
Closing as superseded.

This PR hand-rolled offset-drift bookkeeping (`prependedSinceLastLoad`, manual `historyOffset` adjustment) on top of `api.getHistory()` offset pagination. As noted in the last review, that's the wrong layer to solve this — the API should do the work.

That's now happening:

- **#12769** (adopt cursor pagination in the Generated tab jobs walk) rewrites this exact path: offset `fetchHistory` → keyset `fetchHistoryPage({ after })` with server-authoritative `next_cursor` / `has_more`, backed by cursors now live on core (Comfy-Org/ComfyUI#14014) and cloud. This is the drift-free fix this PR was reaching for.
- **#12464** (FE-740) moves the Generated tab off `historyAssets` / `api.getHistory()` to `/api/assets?include_tags=output`, which also fixes deleted outputs lingering.
- Supporting cursor migration: #12774, #12996, #12827, and #12720 (merged).

No unique value remains here, and the offset path this builds on is being removed. Tracking the fix via #12769 + #12464 instead.

PR #9013, opened 2026-02-20
fix: promoted widget labels show widgetName instead of "nodeId: widgetName"

christian-byrne:
## Summary

Promoted/proxied widgets on subgraph nodes showed generic "nodeId: widgetName" labels (e.g., "3: seed") in the LiteGraph renderer. Now they correctly show just the widget name (e.g., "seed").

## Changes

- **What**: Set proxy widget overlay `label` to `widgetName` instead of `name` (which contains the unique `"nodeId: widgetName"` format). The overlay `name` still retains the unique format for internal identification.
- Added 2 tests verifying proxy widget label defaults and user rename behavior.

## Review Focus

- The one-line fix in `proxyWidget.ts` line 157: `label: widgetName` instead of `label: name`
- The overlay `name` property is unchanged — only `label` (used for display) is affected
- No code parses the label string for identification; all lookups use `_overlay.nodeId` and `_overlay.widgetName` directly

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9013-fix-promoted-widget-labels-show-widgetName-instead-of-nodeId-widgetName-30d6d73d365081ca8d2feb68585fc187) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/graph/subgraph/proxyWidget.test.ts
+++ b/src/core/graph/subgraph/proxyWidget.test.ts
@@ -135,7 +135,21 @@ describe('Subgraph proxyWidgets', () => {
     expect(proxyWidget.name).toBe('1: seed')
   })
 
-  test('Proxy widget label reflects user rename', () => {
+  test('Proxy widget label reflects linked widget label', () => {
+    const [subgraphNode, innerNodes] = setupSubgraph(1)
+    innerNodes[0].addWidget('text', 'seed', 'value', () => {})
+    subgraphNode.properties.proxyWidgets = [['1', 'seed']]
+
+    const proxyWidget = subgraphNode.widgets[0]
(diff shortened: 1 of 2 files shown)


AustinMroz (dismissed):
Change is still less than ideal since it doesn't (and there isn't a straight forward means to) use the `widget.label` of the linked widget, but I have no objections to the change.

Also relevant: #8856 is awaiting review and proposes removing `proxyWidget.ts` entirely. I believe it will also address this issue in a more robust manor

christian-byrne:
@AustinMroz I'd like to get this in, even if it will be reworked next week, as it improves toolkit nodes UI/UX considerably. I'll try to add a special case (temporary code) for labels and see what you think.

christian-byrne:
Works with labels now 
image

PR #9016, opened 2026-02-20
feat: add password widget for STRING inputs

### LINKED ISSUE #7740: Obfuscation for API key input text fields

Add an option to enable a security feature for API input fields to the text is covered with ****** like a password.

nodes with API input

Multiple times per day

Minor annoyance

### Current workarounds

_No response_

_No response_

### Additional context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-7740-Obfuscation-for-API-key-input-text-fields-2d26d73d365081748309e0ce1d96ee02) by [Unito](https://www.unito.io)

dante01yoon:
## Summary
The backend has supported display: "password" as a declaration, but the frontend simply ignored it. Before this change, in useStringWidget.ts the code was:

```
const widget = multiline
  ? addMultilineWidget(...)
  : node.addWidget('text', inputSpec.name, defaultVal, () => {}, {})
```
It always created a 'text' widget — no check for display at all. So even when a backend node sent {"display": "password"}, the frontend treated it as a regular text input showing the value in plain text.

- Add frontend support for `display: "password"` on STRING inputs
- Canvas renderer: `PasswordWidget` shows masked dots, opens prompt on click
- Vue renderer: `WidgetInputPassword` uses native `<input type="password">`
- Password widgets skip serialization to prevent secrets in saved workflows

Fixes #7740

## Test plan
스크린샷 2026-02-21 오전 12 07 17
스크린샷 2026-02-21 오전 12 24 29

- [x] Add a custom node with `("STRING", {"display": "password"})` input
- [x] Verify canvas mode shows masked dots and prompt dialog on click
- [x] Verify Vue mode shows native password input
- [x] Verify password value is not serialized when saving workflow
- [x] Verify normal STRING inputs are unaffected

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9016-feat-add-password-widget-for-STRING-inputs-30d6d73d36508101a9fdc6b67c9acc30) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
```diff
@@ -129,13 +129,24 @@ export const useStringWidget = () => {
...
+      : node.addWidget(
+          isPassword ? 'password' : 'text',
+          inputSpec.name,
+          defaultVal,
+          () => {},
+          {}
+        )
+
+    if (isPassword) {
+      widget.serialize = false
```
dante01yoon, line 148: Intentional — this is the correct security behavior, but worth documenting. 
Password value lost on save/reload

christian-byrne (commented):
The code looks quite good. I could not find any custom nodes / extensions that currently set password inputs in this fashion on the backend node definitions (searched with sourcegraph). This API is also not defined in the backend schemas yet. Would we also be pushing to add this to the backend schemas and official ComfyUI APIs?

dante01yoon:
> The code looks quite good. I could not find any custom nodes / extensions that currently set password inputs in this fashion on the backend node definitions (searched with sourcegraph). This API is also not defined in the backend schemas yet. Would we also be pushing to add this to the backend schemas and official ComfyUI APIs?

@christian-byrne 
Yes, a companion backend PR should be added to officially define this API.

The backend currently passes through INPUT_TYPES dicts to the frontend via 
`/object_info` without validation, so the frontend implementation works today — 
any custom node can already set `("STRING", {"display": "password"})` and it 
will be picked up. This is the same pattern used by `display: "color"` on INT 
inputs

However, the V3 API should formally support this. Proposed backend changes:
- Add `StringDisplay` enum (mirroring `NumberDisplay`) with `"password"` value
- Add `display` parameter to `String.Input`
- Update `InputTypeOptions` TypedDict documentation

Since the pass-through mechanism means both sides work independently, I'd 
suggest landing these PRs in parallel rather than blocking one on the other. 
Happy to open the backend PR.

PR #9025, opened 2026-02-20
App mode - App mode toolbar - 3

pythongosssss:
## Summary

Adds a new toolbar for app mode

## Changes

- **What**:  Adds new toolbar with builder mode disabled

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9025-App-mode-App-mode-toolbar-3-30d6d73d365081549af6fdd1937b098f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2501,6 +2501,10 @@
     "later": "Later",
     "noReleaseNotes": "No release notes available."
   },
+  "appModeToolbar": {
+    "appBuilder": "App builder",
+    "apps": "Apps"
+  },
   "breadcrumbsMenu": {
     "duplicate": "Duplicate",
     "enterAppMode": "Enter app mode",
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2501,10 +2501,6 @@
     "later": "Later",
     "noReleaseNotes": "No release notes available."
   },
-  "appModeToolbar": {
-    "appBuilder": "App builder",
-    "apps": "Apps"
-  },
   "breadcrumbsMenu": {
     "duplicate": "Duplicate",
     "enterAppMode": "Enter app mode",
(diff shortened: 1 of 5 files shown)


src/locales/en/main.json
```diff
@@ -2482,6 +2482,10 @@
     "later": "Later",
     "noReleaseNotes": "No release notes available."
   },
+  "appModeToolbar": {
+    "appBuilder": "App builder",
+    "apps": "Apps"
+  },
```
AustinMroz: Nit: I'd prefer if the localization was under the existing `linearMode` category, just in case we go through a 4th rename.

src/views/LinearView.vue
```diff
@@ -126,8 +136,8 @@ const linearWorkflowRef = useTemplateRef('linearWorkflowRef')
           :selected-item
           :selected-output
         />
-        <div ref="topLeftRef" class="absolute z-21 top-4 left-4">
-          <WorkflowActionsDropdown source="app_mode_menu_selected" />
+        <div ref="topLeftRef" class="absolute z-21 top-1 left-1">
+          <AppModeToolbar v-if="!appModeStore.isBuilderMode" />
         </div>
         <div ref="topRightRef" class="absolute z-21 top-4 right-4" />
         <div ref="bottomLeftRef" class="absolute z-20 bottom-4 left-4" />
```
AustinMroz, line 139: Nit: Should probably make the `topRight ` dock point have the same distance. I'm not sure if we want the toolbar to mirror when sidebar location is set to right, but the dockpoint here is also used for the workflow notes popover.

AustinMroz (dismissed):

src/views/LinearView.vue
pythongosssss: I have simplified the positioning of the notes to just have it positioned relative to the (i) button.
The popover already closes on click outside, and having it directly under the button you've just clicked works better imo and removes complexity.

There was also some dead code with a v-if="false" referencing the notesTo which I have removed, but unsure if it was just added in the past accidently or part of some other planned feature

Position (works on left/right/mobile):
image

For the changing toolbar side, I think only having it on the left is fine

AustinMroz (approved):

PR #9035, opened 2026-02-21
fix: intercept window.open in zendesk test to avoid external network dependency

christian-byrne:
The zendesk support test was flaky because it waited for `networkidle` on an external Zendesk page (`support.comfy.org`). External network requests in CI are unreliable — they can timeout, be slow, or redirect.

**Fix:** Intercept `window.open` to capture the URL that would be opened, without actually navigating to the external page. This makes the test deterministic and fast.

- Fixes flaky test: `[chromium] › browser_tests/tests/dialog.spec.ts:339:3 › Support › Should open external zendesk link with OSS tag`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9035-fix-intercept-window-open-in-zendesk-test-to-avoid-external-network-dependency-30e6d73d365081c2a021edd816b8c1d0) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/dialog.spec.ts
+++ b/browser_tests/tests/dialog.spec.ts
@@ -340,17 +340,21 @@ test.describe('Support', () => {
     comfyPage
   }) => {
     await comfyPage.settings.setSetting('Comfy.UseNewMenu', 'Top')
-    const pagePromise = comfyPage.page.context().waitForEvent('page')
+
+    // Prevent loading the external page
+    await comfyPage.page.context().route('https://support.comfy.org/**', (route) =>
+      route.fulfill({ body: '<html></html>', contentType: 'text/html' })
+    )
+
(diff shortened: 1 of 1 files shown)


browser_tests/tests/dialog.spec.ts
```diff
@@ -345,17 +340,24 @@ test.describe('Support', () => {
...
+    // Intercept window.open to capture the URL without loading an external page
+    const openUrlPromise = comfyPage.page.evaluate(
+      () =>
+        new Promise<string>((resolve) => {
+          window.open = (url?: string | URL) => {
+            resolve(String(url))
+            return null
+          }
+        })
+    )
```
christian-byrne: We should use the playwright apis to be more idiomatic.

Myestery (approved):

PR #9037, opened 2026-02-21
[bugfix] Fix node replacements not loading due to feature flag timing

viva-jinyi:
## Summary
- Node replacements were never loaded because `useNodeReplacementStore().load()` was called before `api.init()`, meaning `serverFeatureFlags` was always empty at that point
- Dispatch `feature_flags` as a custom event from `api.ts` and trigger `load()` in response within `addApiUpdateHandlers()`

## Changes
- **`api.ts`**: Dispatch `feature_flags` custom event after storing server feature flags (already typed in `BackendApiCalls`)
- **`app.ts`**: Replace eager `load()` call with `feature_flags` event listener inside `addApiUpdateHandlers()`, consistent with other API event handlers
- **`nodeReplacementStore.ts`**: Use `api.getServerFeature()` directly instead of `useFeatureFlags` composable; remove side effects from store setup
- **`nodeReplacementStore.test.ts`**: Update mocks to match new `api.getServerFeature` usage

## Review Focus
- Initialization ordering: listener registered before `api.init()` ensures no missed events

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9037-bugfix-Fix-node-replacements-not-loading-due-to-feature-flag-timing-30e6d73d36508107ae2cd72e83c01e1a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/nodeReplacement/nodeReplacementStore.test.ts
+++ b/src/platform/nodeReplacement/nodeReplacementStore.test.ts
@@ -3,7 +3,9 @@ import type { NodeReplacementResponse } from './types'
 import { createPinia, setActivePinia } from 'pinia'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
+import { ServerFeatureFlag } from '@/composables/useFeatureFlags'
 import { useSettingStore } from '@/platform/settings/settingStore'
+import { api } from '@/scripts/api'
 import { fetchNodeReplacements } from './nodeReplacementService'
 import { useNodeReplacementStore } from './nodeReplacementStore'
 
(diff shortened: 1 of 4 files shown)


christian-byrne (approved):
LGTM. The event-driven approach is the correct pattern here — it's consistent with all other WebSocket handlers in `addApiUpdateHandlers()`, the `feature_flags` event is already fully typed in `BackendApiCalls`/`ApiEventTypes`, and the one-shot imperative `load()` is a better fit for events than a reactive watch.

Filed #9039 as a follow-up to make `serverFeatureFlags` a Vue `ref` so that `useFeatureFlags()` getters become truly reactive for future consumers.

PR #9045, opened 2026-02-21
feat: add slot context menu with rename for Vue nodes

christian-byrne:
## Summary

Add right-click context menu on slot connection dots in Vue nodes (Nodes 2.0) with a 'Rename slot' option, matching litegraph parity.

## Changes

- **What**: Right-click on input/output slot dots shows a context menu with 'Rename slot' (filtered by nameLocked and widget-input checks). Uses `dialogService.prompt()` for the rename dialog. Menu position tracks canvas pan/zoom.

## Review Focus

- Rename slot filtering logic (nameLocked, widget check) matches litegraph behavior at `LGraphCanvas.ts:8456`
- Split from #8691 to keep the bug fix separate from quick-connections feature work

Fixes https://github.com/Comfy-Org/ComfyUI_frontend/issues/9062

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9045-feat-add-slot-context-menu-with-rename-for-Vue-nodes-30e6d73d365081af9d00f6547cab47b4) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -142,4 +144,13 @@ const { onClick, onDoubleClick, onPointerDown } = useSlotLinkInteraction({
...
 })
+
+function onSlotContextMenu(event: MouseEvent) {
+  if (!props.nodeId) return
+  showSlotMenu(event, {
+    nodeId: props.nodeId,
+    slotIndex: props.index,
+    isInput: true
+  })
+}
```
jaeone94, line 143: `onSlotContextMenu` calls `showSlotMenu` without checking `canRenameSlot`. When the slot has `nameLocked: true` or is widget-backed, `menuItems` computes to `[]`, but PrimeVue's ContextMenu still renders an empty popup container on screen. The RAF-based position sync also starts running unnecessarily.

Unlike `NodeContextMenu` (which always has menu items when triggered), this code path can genuinely produce an empty menu. A guard in either `onSlotContextMenu` or `show()` would prevent this:

```typescript
// Option A: guard in InputSlot/OutputSlot
function onSlotContextMenu(event: MouseEvent) {
  if (!props.nodeId) return
  const ctx = { nodeId: props.nodeId, slotIndex: props.index, isInput: true }
  if (!canRenameSlot(ctx)) return
  showSlotMenu(event, ctx)
}

// Option B: guard in SlotContextMenu.show()
function show(event: MouseEvent, context: SlotMenuContext) {
  if (!canRenameSlot(context)) return
  // ...existing code
}

File: src/renderer/extensions/vueNodes/components/SlotContextMenu.vue [hunk 2]
```diff
@@ -0,0 +1,157 @@
...
+  ]
+})
+
+async function handleRename(ctx: SlotMenuContext) {
+  const newLabel = await dialogService.prompt({
+    title: t('g.rename'),
+    message: t('g.enterNewNamePrompt')
+  })
+  if (newLabel === null) return
+  renameSlot(ctx, newLabel)
```
jaeone94: `handleRename` only guards against `null` (dialog cancelled), but not empty string. `dialogService.prompt()` returns `string | null`, so confirming with empty input passes `""` to `renameSlot`, which sets `slotInfo.label = ""`.

The display still works (templates use `slotData.label || slotData.localized_name || ...` falsy fallback), but the slot now has a ghost label — it "has" a label property set to `""` that is invisible. The litegraph equivalent at `SubgraphIONodeBase.ts:282` explicitly guards against this:

```typescript
if (newName) this.renameSlot(slot, newName)
```

Suggest matching this behavior:

```typescript
if (!newLabel) return
renameSlot(ctx, newLabel)
```

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
```diff
@@ -0,0 +1,157 @@
...
+const canvasStore = useCanvasStore()
+const lgCanvas = canvasStore.getCanvas()
+const { left: canvasLeft, top: canvasTop } = useElementBounding(lgCanvas.canvas)
+
+const worldPosition = ref({ x: 0, y: 0 })
+let lastScale = 0
+let lastOffsetX = 0
+let lastOffsetY = 0
+
+function updateMenuPosition() {
```
jaeone94: nit: The world-to-screen coordinate conversion, RAF sync loop, dirty-checking, and PrimeVue container repositioning are nearly line-for-line identical to `NodeContextMenu.vue` (lines 84-130):

- `useElementBounding(lgCanvas.canvas)` for canvas bounds
- `worldPosition` ref + `lastScale`/`lastOffsetX`/`lastOffsetY` dirty check
- `updateMenuPosition()` with same `menuEl.style.left/top` assignment
- `useRafFn` + `watchEffect` start/stop pattern
- `contextMenu.value as unknown as { container?: HTMLElement }` cast

A shared composable (e.g., `useCanvasMenuPositionSync`) would eliminate this duplication.

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
```diff
@@ -0,0 +1,157 @@
...
+  if (!isOpen.value) return
+
+  const menuInstance = contextMenu.value as unknown as {
+    container?: HTMLElement
+  }
```
jaeone94: nit:
This accesses an undocumented internal property of PrimeVue's ContextMenu. A PrimeVue version bump could silently break position tracking. Same pattern exists in `NodeContextMenu.vue`, so if a shared composable is extracted (Comment R48), this fragility would at least be centralized.

jaeone94 (commented):
I added a few comments.

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
```diff
@@ -0,0 +1,157 @@
...
+})
+
+async function handleRename(ctx: SlotMenuContext) {
+  const newLabel = await dialogService.prompt({
+    title: t('g.rename'),
+    message: t('g.enterNewNamePrompt')
+  })
+  if (newLabel === null) return
+  renameSlot(ctx, newLabel)
+}
```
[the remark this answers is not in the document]
christian-byrne: note: leaving this as-is - the suggested try/catch just re-throws, which adds no value. dialogService is internal code with its own error handling, so wrapping it here would be defensive code against a scenario that can'\''t happen in practice.

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
christian-byrne: good call - the duplication is real. extracting a useCanvasMenuPositionSync composable makes sense as a followup, keeping this PR scoped to the slot rename feature itself.

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
christian-byrne: yeah, fair concern. same pattern exists in NodeContextMenu.vue so at least it'\''s a known tradeoff. if the shared composable from the other thread gets extracted, the fragility would be centralized to one spot.

jaeone94 (dismissed):
LGTM

src/components/graph/SlotContextMenu.vue
```diff
@@ -0,0 +1,157 @@
+<template>
```
christian-byrne, line 1: issue: SlotContextMenu.vue is ~80 lines of copy-pasted infrastructure from NodeContextMenu.vue (worldPosition, lastScale/lastOffsetX/lastOffsetY, updateMenuPosition, useRafFn, watchEffect start/stop, useElementBounding, screen-to-world conversion in show(), the `as unknown as { container }` cast, and the register/deregister singleton pattern). This violates DRY and means any future bugfix to canvas-position-synced menus needs to be applied in two places.

Suggested alternative: Extract a `useCanvasMenuPositionSync` composable that encapsulates the world-position tracking, RAF sync loop, and coordinate conversion. Both NodeContextMenu and SlotContextMenu consume it. This also encapsulates the mutable `let` variables and direct DOM manipulation (`menuEl.style.left`) behind a clean reactive interface.

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
```diff
@@ -0,0 +1,157 @@
...
+      <a v-bind="itemProps.action" class="flex items-center gap-2 px-3 py-1.5">
+        <span class="flex-1">{{ item.label }}</span>
+      </a>
+    </template>
+  </ContextMenu>
+</template>
+
+<script setup lang="ts">
+import { useElementBounding, useRafFn } from '@vueuse/core'
+import ContextMenu from 'primevue/contextmenu'
```
christian-byrne: issue: AGENTS.md states "Avoid new usage of PrimeVue components." This introduces a new `primevue/contextmenu` usage. The codebase is migrating to shadcn/vue + Reka UI. Consider using ContextMenu from Reka UI / radix-vue, or extending the existing NodeContextMenu to support slot items (since slot rename is just one menu item, a shared menu component with pluggable item sources would be simpler than a second menu component).

src/renderer/extensions/vueNodes/composables/slotMenuService.ts
```diff
@@ -0,0 +1,71 @@
...
+  isInput: boolean
+}
+
+interface SlotMenuInstance {
+  show: (event: MouseEvent, context: SlotMenuContext) => void
+  hide: () => void
+  isOpen: Ref<boolean>
+}
+
+let slotMenuInstance: SlotMenuInstance | null = null
```
christian-byrne, line 18: suggestion: The `let slotMenuInstance` module-level mutable singleton is the same pattern as `registerNodeOptionsInstance` in NodeContextMenu. In the existing codebase, module-level `let` is reserved for performance-critical batching (RAF pending sets, shared ResizeObserver). For wiring up a component reference, `provide`/`inject` at the GraphCanvas level would be more idiomatic Vue and avoids hidden global mutable state. If you keep the singleton approach, at minimum this file should not be named `use*` since it's not a composable -- it returns no reactive state and has no lifecycle integration. Something like `slotMenuService.ts` would be more accurate.

src/renderer/extensions/vueNodes/composables/useSlotContextMenu.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export function showSlotMenu(
+  event: MouseEvent,
+  context: SlotMenuContext
+): void {
+  slotMenuInstance?.show(event, context)
+}
+
+export function renameSlot(context: SlotMenuContext, newLabel: string): void {
+  if (!canRenameSlot(context)) return
```
christian-byrne: suggestion (non-blocking): `canRenameSlot()` is called in three places for the same action: (1) InputSlot/OutputSlot before calling `showSlotMenu`, (2) `menuItems` computed in SlotContextMenu.vue, and (3) inside `renameSlot()` itself. The guard in the slot components is sufficient to prevent showing the menu. The guard inside `renameSlot` is reasonable as a defensive check. The middle call in `menuItems` is redundant -- if the menu was shown, the check already passed.

christian-byrne (commented):
The business logic (`canRenameSlot`, `renameSlot`) is solid and well-tested. The main issue is that SlotContextMenu.vue duplicates ~80 lines of canvas-position-synced context menu infrastructure from NodeContextMenu.vue and introduces a new PrimeVue usage.

Suggested alternative approach:
1. Extract a `useCanvasMenuPositionSync` composable from NodeContextMenu.vue (this also improves NodeContextMenu)
2. Either extend NodeContextMenu to accept pluggable item sources (slot rename is just one menu item), or build SlotContextMenu on Reka UI's ContextMenu instead of PrimeVue
3. Wire the menu instance via `provide`/`inject` at the GraphCanvas level instead of a module-level singleton
4. Keep `canRenameSlot`/`renameSlot` as-is -- they're clean

src/components/graph/SlotContextMenu.vue
```diff
@@ -0,0 +1,157 @@
...
+defineExpose({ show, hide, isOpen })
+
+onMounted(() => {
+  registerSlotMenuInstance({ show, hide, isOpen })
+})
+
+onUnmounted(() => {
+  registerSlotMenuInstance(null)
+})
+</script>
```
christian-byrne, line 135: nitpick (non-blocking): SlotContextMenu.vue lives under `src/renderer/extensions/vueNodes/components/` but is mounted at the GraphCanvas.vue level alongside NodeContextMenu (which lives in `src/components/graph/`). Components that are siblings in the template tree should be colocated.

src/renderer/extensions/vueNodes/composables/slotMenuService.test.ts
```diff
@@ -0,0 +1,180 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
```
christian-byrne, line 1: suggestion (non-blocking): Tests cover the pure functions (`canRenameSlot`, `renameSlot`) well, but there is no component test for SlotContextMenu.vue itself, which is where most of the complexity lives (positioning, show/hide lifecycle, menu item construction, dialog integration).

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
```diff
@@ -0,0 +1,131 @@
...
+    <div
+      v-if="isOpen"
+      ref="menuEl"
+      :style="{
+        left: `${screenPosition.x}px`,
+        top: `${screenPosition.y}px`
+      }"
+      class="fixed z-1700 min-w-[220px] rounded-lg border border-border-subtle bg-base-background p-2 shadow-sm"
+      role="menu"
+      @keydown.escape="hide"
```
dante01yoon: Suggestion: The template has @keydown.escape="hide" on the menu div AND useEventListener(window,         'keydown', ...) for Escape. The window listener already catches all Escape presses  regardless of focus. Why don't you remove one?

src/composables/graph/useCanvasAnchoredPosition.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  watchEffect(() => {
+    if (isOpen.value) {
+      startSync()
+    } else {
+      stopSync()
+    }
+  })
+
+  return { screenPosition, anchorToEvent, canvasLeft, canvasTop, lgCanvas }
```
dante01yoon: nit: The return exposes canvasLeft, canvasTop, and lgCanvas which are internal implementation  details. Neither SlotContextMenu nor NodeContextMenu uses them:

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
```diff
@@ -0,0 +1,131 @@
...
+}
+
+async function handleRename() {
+  const ctx = activeContext.value
+  if (!ctx) return
+  hide()
+  const newLabel = await dialogService.prompt({
+    title: t('g.rename'),
+    message: t('g.enterNewNamePrompt')
+  })
```
dante01yoon: suggestion: If the prompt dialog rejects (rather than resolving to null), the error is unhandled

src/components/graph/SlotContextMenu.vue
```diff
@@ -0,0 +1,131 @@
...
+  <Teleport to="body">
+    <div
+      v-if="isOpen"
+      ref="menuEl"
+      :style="{
+        left: `${screenPosition.x}px`,
+        top: `${screenPosition.y}px`
+      }"
+      class="fixed z-1700 min-w-[220px] rounded-lg border border-border-subtle bg-base-background p-2 shadow-sm"
+      role="menu"
```
dante01yoon, line 11: nit: The menu has role="menu" and role="menuitem" but no arrow key navigation or tabindex on menu items. Users expect Up/Down arrow keys and Enter/Space to work in ARIA menus. This is a gap vs the PrimeVue ContextMenu used by NodeContextMenu. At minimum, add tabindex="0" to menu items.

src/components/graph/GraphCanvas.vue
```diff
@@ -104,6 +104,7 @@
     <TitleEditor />
     <SelectionToolbox v-if="selectionToolboxEnabled" />
     <NodeContextMenu />
+    <SlotContextMenu />
```
dante01yoon: suggestion: The component mounts and initializes useCanvasAnchoredPosition (which calls                 
  useCanvasStore().getCanvas() and useElementBounding) even when Vue nodes are disabled.    
  Consider guarding:                                                                          
                                                                                            
```vue
 <SlotContextMenu v-if="shouldRenderVueNodes" />
```

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
christian-byrne: fixed in 6fae638 - removed the template @keydown.escape, kept the window listener since it works regardless of focus

src/composables/graph/useCanvasAnchoredPosition.ts
christian-byrne: fixed in 1604b2b - return now only exposes screenPosition and anchorToEvent

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
christian-byrne: fixed in 6fae638 - added .catch(() => null) for the dialog prompt rejection

src/components/graph/SlotContextMenu.vue
christian-byrne: fixed in 6fae638 - added tabindex="0" to all three menu items

src/components/graph/GraphCanvas.vue
christian-byrne: fixed in 098bf1a - SlotContextMenu now has v-if="shouldRenderVueNodes" guard in GraphCanvas.vue

src/components/graph/SlotContextMenu.vue
christian-byrne: already addressed - useCanvasAnchoredPosition composable at src/composables/graph/useCanvasAnchoredPosition.ts is shared by both NodeContextMenu and SlotContextMenu. the RAF sync, coordinate conversion, and dirty-checking are all centralized there.

src/renderer/extensions/vueNodes/components/SlotContextMenu.vue
christian-byrne: outdated - SlotContextMenu now uses a custom div with role="menu", no PrimeVue dependency

src/renderer/extensions/vueNodes/composables/slotMenuService.ts
christian-byrne: fixed in 9287f1d - renamed to slotMenuService.ts

src/renderer/extensions/vueNodes/composables/useSlotContextMenu.ts
christian-byrne: outdated - code was refactored, the redundant call no longer exists

src/components/graph/SlotContextMenu.vue
christian-byrne: fixed in 098bf1a - moved to src/components/graph/SlotContextMenu.vue, colocated with NodeContextMenu

src/renderer/extensions/vueNodes/composables/slotMenuService.test.ts
christian-byrne: good call - unit tests cover slotMenuService.ts (canRenameSlot, renameSlot) and e2e tests cover the full component flow (rename, disconnect, remove). a dedicated component test for SlotContextMenu.vue positioning/lifecycle would be good followup work.

dante01yoon (dismissed):
Issue addressed. LGTM

src/renderer/extensions/vueNodes/composables/slotMenuService.test.ts
christian-byrne: @coderabbitai Make an Issue to do this in a followup PR on a future date - dedicated component test for SlotContextMenu.vue covering positioning, show/hide lifecycle, and dialog integration.

christian-byrne:
Rebased `feat/slot-context-menu-rename` onto current `main` (was 1601 commits behind; head now `834aed3b6`).

Conflicts resolved during rebase:
- `InputSlot.vue` / `OutputSlot.vue`: import-ordering conflicts only, both sides' imports kept.
- `src/lib/litegraph/src/types/graphTriggers.ts`: kept `main`'s `NodeSlotLabelChangedEvent` (superset shape with optional `slotType`), dropped this branch's duplicate. `NodeSlotErrorsChangedEvent`/`NodeSlotLinksChangedEvent` were intentionally removed from `main` by the ECS migration (`a8abe573c`, #14246) along with their only consumer, so they were not re-added.
- `src/composables/graph/useGraphNodeManager.ts` (+ its test file): this file was deleted entirely by the ECS migration. Several commits in this branch's history modified it to add a `refreshNodeSlots` reactivity fix (force `shallowReactive` inputs/outputs to re-trigger on slot label change via splice/spread). That fix has **no live equivalent post-migration** — `NodeState.inputs`/`outputs` are now `shallowReactive` arrays owned by `nodeDataStore.ts`, and nothing currently re-triggers them on a label-only change (`node:slot-label:changed` today only has one listener, in the App Builder's `useResolvedSelectedInputs.ts`, which doesn't touch `NodeState`). I dropped the deleted-file hunks rather than inventing new store-layer code blind during a rebase; this is a genuine gap worth a follow-up, not a rebase artifact.
- `src/renderer/extensions/vueNodes/composables/slotMenuService.ts`: `NodeId` import path moved from `@/lib/litegraph/src/LGraphNode` to `@/types/nodeId` (no longer re-exported from the old location).
- `slotMenuService.test.ts`: `NodeId` is now a branded type; updated raw string literals to `toNodeId('...')` per the convention used elsewhere in the codebase (e.g. `useImageCrop.test.ts`).

Also removed a now-dangling `useSlotElementTracking` import in `InputSlot.vue`/`OutputSlot.vue` — that composable was removed from `main` by #16001 and was never referenced in these two files to begin with.

## Evidence
- `pnpm typecheck` — clean (`vue-tsc --noEmit`, exit 0)
- `pnpm vitest run src/renderer/extensions/vueNodes/composables/slotMenuService.test.ts src/renderer/extensions/vueNodes/components/InputSlot.test.ts src/renderer/extensions/vueNodes/components/OutputSlot.test.ts` — 40 passed
- `pnpm exec eslint <changed files>` — clean
- Feature diff vs `origin/main` confirmed equivalent to the pre-rebase diff vs the old base (same file set, same net additions), modulo the `useGraphNodeManager.ts` drop noted above.

Follow-up: the slot-label reactivity gap (`NodeState.inputs`/`outputs` not re-triggering on label-only rename post-ECS-migration) should get its own fix/ticket rather than being silently reintroduced here.

christian-byrne:
Filed the reactivity gap noted above as [#16642](https://github.com/Comfy-Org/ComfyUI_frontend/issues/16642).

christian-byrne:
Checked mergeability for `act-fe-9045-conflict-1eda8f77`: no merge conflict exists. GitHub reports `mergeable: MERGEABLE` (`mergeStateStatus: BLOCKED` is the draft/review gate, not a conflict). Local verification: `git merge origin/main --no-commit --no-ff` against current PR head `834aed3b65` (pushed today, later than the row's declared conflict SHA `2501d6cc65`) applied cleanly with zero conflict markers — only an auto-merge on `src/locales/en/main.json`.

No rebase/resolution action needed; closing out the conflict-lane row as verified-clean rather than forcing a no-op push.

christian-byrne:
Accounting corrected: 23/23 hunks.
#8691 remains open, not `main`.

This supersession receipt now accounts for every hunk in closed source head `834aed3b658f64ec8803bed263dc7fc20ca408c2` against open survivor head `40f7b89ce9d91f9cd76e85cd2c531199ee0f5ec8`. The 23 rows total the complete source diff: **+1,073/-74, zero unaccounted hunks**.

| ID | Source hunk | Size | #8691 disposition |
|---|---|---:|---|
| B1 | `browser_tests/tests/vueNodes/interactions/node/slotContextMenu.spec.ts` `@@ -0,0 +1,209 @@` | +209/-0 | DROP-TEST: all six browser tests absent. |
| G1 | `src/components/graph/GraphCanvas.vue` `@@ -117,0 +118 @@` | +1/-0 | CARRIED-SEMANTIC: menu mounts from `SelectionToolbox.vue` instead. |
| G2 | `src/components/graph/GraphCanvas.vue` `@@ -156,0 +158 @@` | +1/-0 | CARRIED-SEMANTIC: import moved to `SelectionToolbox.vue`. |
| N1 | `src/components/graph/NodeContextMenu.vue` `@@ -49 +49 @@` | +1/-1 | DROP-REFACTOR. |
| N2 | `src/components/graph/NodeContextMenu.vue` `@@ -53,0 +54 @@` | +1/-0 | DROP-REFACTOR. |
| N3 | `src/components/graph/NodeContextMenu.vue` `@@ -62 +62,0 @@` | +0/-1 | DROP-REFACTOR. |
| N4 | `src/components/graph/NodeContextMenu.vue` `@@ -79 +79 @@` | +1/-1 | DROP-REFACTOR. |
| N5 | `src/components/graph/NodeContextMenu.vue` `@@ -81,14 +81 @@` | +1/-14 | DROP-REFACTOR. |
| N6 | `src/components/graph/NodeContextMenu.vue` `@@ -96 +82,0 @@` | +0/-1 | DROP-REFACTOR. |
| N7 | `src/components/graph/NodeContextMenu.vue` `@@ -102,37 +88,2 @@` | +2/-37 | DROP-REFACTOR. |
| N8 | `src/components/graph/NodeContextMenu.vue` `@@ -220,19 +171 @@` | +1/-19 | DROP-REFACTOR. |
| S1 | `src/components/graph/SlotContextMenu.vue` `@@ -0,0 +1,135 @@` | +135/-0 | PARTITIONED: shell/core rename carried semantically; Disconnect, Remove, custom markup, i18n dialog, and hardening dropped. |
| A1 | `src/composables/graph/useCanvasAnchoredPosition.ts` `@@ -0,0 +1,77 @@` | +77/-0 | CARRIED-INLINED / DROP-ABSTRACTION: equivalent tracking stays local; no shared extraction. |
| L1 | `src/locales/en/main.json` `@@ -19,0 +20 @@` | +1/-0 | DROP-FEATURE: Disconnect Links locale. |
| L2 | `src/locales/en/main.json` `@@ -24,0 +26 @@` | +1/-0 | DROP-FEATURE: Remove Slot locale. |
| I1 | `src/renderer/extensions/vueNodes/components/InputSlot.vue` `@@ -33,0 +34 @@` | +1/-0 | CARRIED: context-menu event. |
| I2 | `src/renderer/extensions/vueNodes/components/InputSlot.vue` `@@ -65,0 +67,5 @@` | +5/-0 | PARTITIONED: menu dispatch carried; action preflight/dead import dropped. |
| I3 | `src/renderer/extensions/vueNodes/components/InputSlot.vue` `@@ -130,0 +137,7 @@` | +7/-0 | PARTITIONED: guard/context/dispatch carried; empty-menu preflight dropped. |
| O1 | `src/renderer/extensions/vueNodes/components/OutputSlot.vue` `@@ -22,0 +23 @@` | +1/-0 | CARRIED: context-menu event. |
| O2 | `src/renderer/extensions/vueNodes/components/OutputSlot.vue` `@@ -36,0 +38,5 @@` | +5/-0 | PARTITIONED: menu dispatch carried; action preflight/dead import dropped. |
| O3 | `src/renderer/extensions/vueNodes/components/OutputSlot.vue` `@@ -114,0 +121,7 @@` | +7/-0 | PARTITIONED: guard/context/dispatch carried; empty-menu preflight dropped. |
| T1 | `src/renderer/extensions/vueNodes/composables/slotMenuService.test.ts` `@@ -0,0 +1,468 @@` | +468/-0 | DROP-TEST: all 34 source tests absent; #8691's 18 tests cover Connect To, not rename. |
| V1 | `src/renderer/extensions/vueNodes/composables/slotMenuService.ts` `@@ -0,0 +1,147 @@` | +147/-0 | PARTITIONED: context/show/core rename carried; Disconnect, Remove, preflight, refresh, shared lookup, and direct-call guard dropped. |

Material intentionally excluded by the survivor choice: six browser tests; all 34 source unit tests, including 12 rename tests; Disconnect Links; Remove Slot; shared NodeContextMenu positioning; localized dialog/custom markup treatment; empty-menu preflight; and the direct-call rename authorization guard.

The full semantic partitions, pinned merge bases, patch hashes, and verification commands are in the [program report](https://github.com/christian-byrne/in-app-agent-program/blob/main/reports/audit/evfail-32-fe9045-diff-to-zero.md). Program receipt: [`42069ee3`](https://github.com/christian-byrne/in-app-agent-program/commit/42069ee390cfd8eca943568839c90597a4db5f30).

#8691 is still open, unapproved, blocked, and outside the merge queue. Its successful checks completed on 2026-09-02 and are stale relative to current `main`; they are not merge-readiness evidence.

</details>

PR #9046, opened 2026-02-21
fix: subgraph unpacking creates extra link to seed widget

christian-byrne:
## Summary

Fix subgraph unpacking creating spurious links to widget inputs (e.g. seed) when the subgraph contains ComfySwitchNode with duplicate internal links.

## Changes

- **What**: Two fixes in `_unpackSubgraphImpl`:
  1. Strip links from serialized node data **before** `configure()` so `onConnectionsChange` doesn't resolve subgraph-internal link IDs against the parent graph's link map (which may contain unrelated links with colliding numeric IDs).
  2. Deduplicate links by `(origin, origin_slot, target, target_slot)` before reconnecting, preventing repeated disconnect/reconnect cycles on widget inputs that cause slot index drift.

## Review Focus

- The link-stripping before `configure()` mirrors what `LGraphNode.clone()` already does — nodes should be configured without stale link references when links will be recreated separately.
- Deduplication is defensive against malformed subgraph data; the duplicate links in the reproduction workflow likely originated from a prior serialization bug.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9046-fix-subgraph-unpacking-creates-extra-link-to-seed-widget-30e6d73d36508125a5fefa1309485516) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/assets/subgraphs/subgraph-duplicate-links.json
+++ b/browser_tests/assets/subgraphs/subgraph-duplicate-links.json
@@ -1,0 +1,183 @@
+{
+  "id": "a1b2c3d4-e5f6-7890-abcd-ef1234567890",
+  "revision": 0,
+  "last_node_id": 2,
+  "last_link_id": 0,
+  "nodes": [
+    {
+      "id": 2,
+      "type": "e5fb1765-aaaa-bbbb-cccc-ddddeeee0001",
+      "pos": [600, 400],
+      "size": [200, 100],
+      "flags": {},
+      "order": 0,
+      "mode": 0,
+      "inputs": [],
+      "outputs": [
+        {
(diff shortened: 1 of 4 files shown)


AustinMroz (approved):
I believe the diagnosis is correct, but the solution leaves me concerned. Having taken some time to think things over, I don't think this PR causes or exacerbates the areas of concern so it has my approval.

Nodes are loaded as if they have no connections, and then connected after. The specific example that comes to mind is primitives: Loading them with no connection means that their control_after_generate won't deserialize correctly, but it appears this was already broken prior to this PR.

Is `dedupedNewLinks` actually filtering out links? If so, that sounds like an underlying bug in need of solving.

PR #9047, opened 2026-02-21
App mode mobile redesign

AustinMroz:
Reworks the app mode display for mobile devices. Adds multiple bottom tabs that can be swiped between.

AnimateDiff_00005

To be handled in followup PRs
- Nicer error display
- Support for even smaller screens
- UX improvements for the 'Outputs' pane
  - Was postponed to minimize conflicts with non-mobile development.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9047-App-mode-mobile-redesign-30e6d73d365081388e4adea4df886522) by [Unito](https://www.unito.io)


src/views/LinearView.vue
```diff
@@ -76,7 +77,8 @@ const bottomRightRef = useTemplateRef('bottomRightRef')
 const linearWorkflowRef = useTemplateRef('linearWorkflowRef')
 </script>
 <template>
-  <div class="absolute w-full h-full">
+  <MobileDisplay v-if="mobileDisplay" />
+  <div v-else class="absolute w-full h-full">
```
pythongosssss, line 78: Line 92 contains a v-if="mobileDisplay" that is now dead code

src/components/common/DropdownMenu.vue
```diff
@@ -0,0 +1,74 @@
...
+      <DropdownMenuContent
+        side="bottom"
+        :side-offset="5"
+        :collision-padding="10"
+        v-bind="$attrs"
+        :class="contentClass"
+      >
+        <slot :item-class>
+          <DropdownItem
+            v-for="(item, index) in entries!"
```
pythongosssss: entries is optional in the prop definition but `!` assumes it isn't. Either make prop required or ensure entries doesnt crash here

src/renderer/extensions/linearMode/MobileDisplay.vue
```diff
@@ -0,0 +1,226 @@
...
+        <div class="overflow-y-auto contain-size h-full w-screen absolute">
+          <LinearControls mobile @navigate-assets="activeIndex = 1" />
+        </div>
+        <div
+          class="w-screen absolute h-full bg-base-background left-[100vw] top-0 flex flex-col"
+        >
+          <LinearPreview mobile />
+        </div>
+        <AssetsSidebarTab
+          class="h-full w-screen absolute bg-base-background [&>:first-child>:nth-child(odd)]:hidden left-[200vw] top-0"
```
pythongosssss: I think we should have a prop for whatever this is hiding `[&>:first-child>:nth-child(odd)]:hidden`

src/renderer/extensions/linearMode/LinearControls.vue
```diff
@@ -285,8 +254,70 @@ defineExpose({ runButtonClick })
...
+            <template #button>
+              <Button size="lg" class="-mr-3 pr-7">
+                <i v-if="batchCount == 1" class="icon-[lucide--chevron-down]" />
+                <div v-else class="tabular-nums" v-text="`${batchCount}x`" />
+              </Button>
+            </template>
+            <WidgetInputNumberInput
+              v-model="batchCount"
+              :widget="batchCountWidget"
+              class="*:[.min-w-0]:w-40 **:[.h-7]:h-10 **:[.h-7_button.w-8]:w-12"
```
pythongosssss: This class looks very fragile, could we instead add a size prop or pass through a more specific targeted prop

src/renderer/extensions/linearMode/MobileDisplay.vue
```diff
@@ -0,0 +1,226 @@
...
+              class="shrink-0 icon-[lucide--chevron-down] bg-muted-foreground"
+            />
+          </div>
+        </template>
+      </Popover>
+      <CurrentUserButton v-if="isLoggedIn" class="[&_i]:hidden" />
+    </header>
+    <div class="size-full contain-content rounded-b-4xl">
+      <div
+        :class="cn('size-full relative', !isSwiping && 'transition-all')"
```
pythongosssss: transition-[translate] ?

src/renderer/extensions/linearMode/MobileDisplay.vue
```diff
@@ -0,0 +1,226 @@
...
+              class="truncate contain-size h-full w-full"
+              v-text="workflowStore.activeWorkflow?.filename"
+            />
+            <i
+              class="shrink-0 icon-[lucide--chevron-down] bg-muted-foreground"
+            />
+          </div>
+        </template>
+      </Popover>
+      <CurrentUserButton v-if="isLoggedIn" class="[&_i]:hidden" />
```
pythongosssss: Should be a prop `[&_i]:hidden`

pythongosssss (changes requested):

src/components/common/DropdownMenu.vue
AustinMroz: It would be ideal if there was a way for the spec to define that either entries, or a slot is required, but adding a default to entries is reasonable.

src/renderer/extensions/linearMode/MobileDisplay.vue
AustinMroz: It's both temp and kludge, but I'm cautious to make changes to the Assets tab directly when I've heard there's already an in progress rework to accomplish this same change.

I think I'll just pull it out for now.  Having a slightly cramped display is better than allowing this change into core.

src/renderer/extensions/linearMode/LinearControls.vue
AustinMroz: I've specifically cleaned up this one. I should likely do the same for the "selected input widgets", but that'd be better handled in a future PR.

src/renderer/extensions/linearMode/MobileDisplay.vue
AustinMroz: And the prop was already implemented. How shameful of me.

src/renderer/extensions/linearMode/MobileDisplay.vue
```diff
@@ -0,0 +1,228 @@
...
+    </header>
+    <div class="size-full contain-content rounded-b-4xl">
+      <div
+        :class="
+          cn('size-full relative', !isSwiping && 'transition-[translate]')
+        "
+        :style="{ translate }"
+      >
+        <div class="overflow-y-auto contain-size h-full w-screen absolute">
+          <LinearControls mobile @navigate-assets="activeIndex = 1" />
```
pythongosssss: Looks like index 1 is outputs not assets

src/renderer/extensions/linearMode/MobileDisplay.vue
```diff
@@ -0,0 +1,228 @@
...
+  disableTextSelect: true,
+  onSwipeEnd() {
+    if (
+      !sliderWidth.value ||
+      Math.abs(distanceX.value) / sliderWidth.value < 0.4
+    )
+      return
+    if (distanceX.value < 0)
+      activeIndex.value = Math.max(activeIndex.value - 1, 0)
+    else activeIndex.value = Math.min(activeIndex.value + 1, 2)
```
pythongosssss: minor nit: changing this to tabs.length-1 would futureproof this and easy for a future dev to overlook

pythongosssss (approved):

PR #9050, opened 2026-02-21
feat: add job history and assets sidebar badge behavior

benceruleanlu:
## Summary
Add sidebar badge behavior for queue/asset visibility updates:
- Job History tab icon shows active jobs count (`queued + running`) only when the Job History panel is closed.
- Assets tab icon no longer mirrors active jobs; when QPO V2 is enabled it now shows the number of assets added since the last time Assets was opened.
- Opening Assets clears the unseen added-assets badge count.

## Changes
- Added `iconBadge` logic to Job History sidebar tab.
- Replaced Assets sidebar badge source with new unseen-assets counter logic.
- Added `assetsSidebarBadgeStore` to track unseen asset additions from history updates and reset on Assets open.
- Added/updated unit tests for both sidebar tab composables and the new store behavior.

https://github.com/user-attachments/assets/33588a2a-c607-4fcc-8221-e7f11c3d79cc

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9050-fix-add-job-history-and-assets-sidebar-badge-behavior-30e6d73d365081c38297fe6aac9cd34c) by [Unito](https://www.unito.io)


src/stores/workspace/assetsSidebarBadgeStore.ts
```diff
@@ -0,0 +1,76 @@
...
+        queueStore.historyTasks.map((task) => task.jobId)
+      )
+    }
+
+    watch(
+      () => queueStore.historyTasks,
+      (historyTasks) => {
+        if (!hasInitializedHistory.value) {
+          hasInitializedHistory.value = true
+          markCurrentHistoryAsSeen()
```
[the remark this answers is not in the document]
benceruleanlu: Fixed with hydration flag

src/stores/workspace/assetsSidebarBadgeStore.ts
```diff
@@ -0,0 +1,96 @@
...
+
+const getAddedAssetCount = (task: TaskItemImpl): number => {
+  if (typeof task.outputsCount === 'number' && task.outputsCount > 0) {
+    return task.outputsCount
+  }
+
+  return task.previewOutput ? 1 : 0
+}
+
+export const useAssetsSidebarBadgeStore = defineStore(
```
christian-byrne: There's a temptation to want to refactor this such that the places where we call `/jobs` handles reconciliation of client jobs records with new server records and then exposes some reactive `newlySeenJobs` in a really efficient manner -- preventing us from needing to traverse here on every update. However, this will all be getting changed when we migrate to `/assets`, so I am not convinced it's worth doing if this code already works and is acceptable on performance.

That said, the counting of assets seems like it could happen when we already traverse the jobs in the job/queue store, do you want to scope out whether that's worth doing?

src/stores/workspace/assetsSidebarBadgeStore.ts
```diff
@@ -0,0 +1,96 @@
+import { defineStore } from 'pinia'
+import { ref, watch } from 'vue'
+
+import type { TaskItemImpl } from '@/stores/queueStore'
+import { useQueueStore } from '@/stores/queueStore'
+import { useSidebarTabStore } from '@/stores/workspace/sidebarTabStore'
+
+const getAddedAssetCount = (task: TaskItemImpl): number => {
+  if (typeof task.outputsCount === 'number' && task.outputsCount > 0) {
+    return task.outputsCount
```
christian-byrne: nit: if the server returns `outputs_count: 0`, this falls through to the preview check and returns 1 if preview exists. is that intentional? if 0 means "unknown" a comment here would help, otherwise might want to trust the server's 0.

```suggestion
  if (typeof task.outputsCount === 'number') {
    return Math.max(task.outputsCount, 0)
  }
```

src/stores/workspace/assetsSidebarBadgeStore.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+export const useAssetsSidebarBadgeStore = defineStore(
+  'assetsSidebarBadge',
+  () => {
+    const queueStore = useQueueStore()
+    const sidebarTabStore = useSidebarTabStore()
+
+    const unseenAddedAssetsCount = ref(0)
+    const countedHistoryAssetsByJobId = ref(new Map<string, number>())
```
christian-byrne, line 23: This Map only gets pruned when `markCurrentHistoryAsSeen()` runs (init + tab open). between tab opens, job IDs that age out of historyTasks stay in the Map. probably never matters with 64 max items but worth being aware of.

src/stores/workspace/assetsSidebarBadgeStore.ts
```diff
@@ -0,0 +1,96 @@
...
+        if (!hasInitializedHistory.value) {
+          hasInitializedHistory.value = true
+          markCurrentHistoryAsSeen()
+          return
+        }
+
+        const isAssetsTabOpen = sidebarTabStore.activeSidebarTabId === 'assets'
+        const countedAssetsByJobId = countedHistoryAssetsByJobId.value
+
+        for (const task of historyTasks) {
```
christian-byrne, line 56: re-traverses all ~64 tasks on every history update even though typically 1-2 changed. fine at this scale but a short-circuit (e.g. comparing array identity / length before iterating) could skip work in the common no-change case.

src/composables/sidebarTabs/useJobHistorySidebarTab.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  })
+}))
+
+vi.mock('@/stores/workspace/sidebarTabStore', () => ({
+  useSidebarTabStore: () => ({
+    activeSidebarTabId: mockActiveSidebarTabId.value
+  })
+}))
+
+describe('useJobHistorySidebarTab', () => {
```
christian-byrne, line 26: minor: no `beforeEach` resetting the hoisted mock values. works today since each test sets both explicitly, but makes future test additions fragile.

src/stores/workspace/assetsSidebarBadgeStore.test.ts
```diff
@@ -0,0 +1,137 @@
...
+          type: 'output',
+          nodeId: '1',
+          mediaType: 'images'
+        }
+      : undefined
+  } as JobListItem)
+
+describe('useAssetsSidebarBadgeStore', () => {
+  beforeEach(() => {
+    setActivePinia(createPinia())
```
christian-byrne: project testing docs recommend `createTestingPinia({ stubActions: false })` over raw `createPinia()` - the integration approach is valid here since you need real watchers, just flagging for consistency.

christian-byrne (commented):
Took a pass. logic is solid overall, nice job on the hydration pattern. Few small things inline.

christian-byrne:
one thought — `queueStore.update()` already diffs old vs new history via `existingByJobId` (line 557-569), which is essentially the same work the badge store watcher re-does on every update. we could expose a reconciliation result from the queue store like:

```ts
const historyReconciliation = shallowRef<{
  added: TaskItemImpl[]
  changed: { prev: TaskItemImpl; next: TaskItemImpl }[]
}>({ added: [], changed: [] })
```

then the badge store just sums the diff instead of iterating all 64 tasks with its own Map.

that said, not even sure if its worth doing here. the current approach is ~64 items × Map lookup which is negligible perf-wise, the badge store is nicely self-contained and easy to delete when we migrate to /assets, and adding reconciliation state to queueStore is investment into plumbing we're about to replace. just wanted to flag the option in case we end up needing that diff data for something else before the migration.

src/stores/workspace/assetsSidebarBadgeStore.ts
benceruleanlu: Good catch, I applied your suggestion manually

src/stores/workspace/assetsSidebarBadgeStore.ts
benceruleanlu: True, changed it to recaclulate on history update

src/stores/workspace/assetsSidebarBadgeStore.ts
benceruleanlu: K, added a cheap equality check for this case

src/composables/sidebarTabs/useJobHistorySidebarTab.test.ts
benceruleanlu: K, added a beforeEach

src/stores/workspace/assetsSidebarBadgeStore.test.ts
benceruleanlu: K, switched to it

christian-byrne (approved):
LGTM

PR #9051, opened 2026-02-21
fix: make serverFeatureFlags reactive via Vue ref

### LINKED ISSUE #9039: Make serverFeatureFlags reactive via Vue ref
## Summary

`ComfyApi.serverFeatureFlags` is a plain `Record<string, unknown>` class property. The `useFeatureFlags()` composable wraps it in `reactive()` with getters, but since the underlying data is non-reactive, Vue cannot track mutations — the getters never re-trigger when flags change via WebSocket.

## Proposed Change

Make `serverFeatureFlags` a Vue `ref` so that `useFeatureFlags()` getters become truly reactive.

### Touch points in `src/scripts/api.ts`

1. **Declaration** (L343): `serverFeatureFlags: Record<string, unknown> = {}` → `serverFeatureFlags = ref<Record<string, unknown>>({})`
2. **Assignment** (L698): `this.serverFeatureFlags = msg.data` → `this.serverFeatureFlags.value = msg.data`
3. **`serverSupportsFeature`** (L1294): `this.serverFeatureFlags` → `this.serverFeatureFlags.value`
4. **`getServerFeature`** (L1304): same
5. **`getServerFeatures`** (L1312): same

### Tests to update

- `src/scripts/api.featureFlags.test.ts` — direct assignments like `api.serverFeatureFlags = {}` become `.value =`
- `src/composables/useFeatureFlags.test.ts` — should add tests verifying reactivity (watch/computed triggers on flag change)

### Benefits

- `useFeatureFlags()` getters will actually trigger Vue reactivity when flags arrive or change
- Components using `flags.nodeReplacementsEnabled` in templates/computed properties will update automatically
- Eliminates the class of timing bugs fixed in #9037 for future consumers

### Context

- PR #9037 fixed a specific instance of this timing issue for node replacements using an event-driven approach, which is correct for that imperative one-shot `load()` use case
- A previous attempt to create a full `featureFlagsStore.ts` Pinia store with version tracking exists on branch `fix-manager-feature-flags-reactive` (commit `8e07126`) but was never merged
- The `ref` approach is much simpler than a full store — just 5 lines changed in `api.ts`
- The reactive foundation composable (`useFeatureFlags`) was added in #4817

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9039-Make-serverFeatureFlags-reactive-via-Vue-ref-30e6d73d36508183a5fad5c16baf9233) by [Unito](https://www.unito.io)

dante01yoon:
## Summary

- Fixes #9039
`serverFeatureFlags` was a plain object, so Vue's reactivity system could not
track changes. When flags arrived via WebSocket, components using
`useFeatureFlags()` did not re-render because the getters internally called
`api.getServerFeature()` which read non-reactive data.

## Changes

- `src/scripts/api.ts`: Convert `serverFeatureFlags` from plain object to
  `ref<Record<string, unknown>>({})`. Update all internal accesses to use
  `.value` (6 locations).
- `src/scripts/api.featureFlags.test.ts`: Update direct accesses to `.value`.
  Add reactivity test verifying `computed` reacts to flag changes.
- `browser_tests/tests/featureFlags.spec.ts`: Update all `page.evaluate()`
  accesses to use `.value` (8 locations).

## Why this works

`useFeatureFlags()` wraps each flag in a `reactive()` getter that calls
`api.getServerFeature()`. That method now reads `this.serverFeatureFlags.value`,
which Vue tracks as a dependency. No changes needed in `useFeatureFlags.ts` or
any of the 28 consumer files.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9051-fix-make-serverFeatureFlags-reactive-via-Vue-ref-30e6d73d365081b3bfe2c151d24ede1b) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/featureFlags.spec.ts
+++ b/browser_tests/tests/featureFlags.spec.ts
@@ -37,12 +37,9 @@ test.describe('Feature Flags', { tag: ['@slow', '@settings'] }, () => {
 
       // Monitor for server feature flags
       const checkInterval = setInterval(() => {
-        if (
-          window.app?.api?.serverFeatureFlags &&
-          Object.keys(window.app.api.serverFeatureFlags).length > 0
-        ) {
-          window.__capturedMessages!.serverFeatureFlags =
-            window.app.api.serverFeatureFlags
+        const flags = window.app?.api?.serverFeatureFlags?.value
+        if (flags && Object.keys(flags).length > 0) {
+          window.__capturedMessages!.serverFeatureFlags = flags
           clearInterval(checkInterval)
         }
       }, 100)
@@ -96,7 +93,7 @@ test.describe('Feature Flags', { tag: ['@slow', '@settings'] }, () => {
   }) => {
     // Get the actual server feature flags from the backend
     const serverFlags = await comfyPage.page.evaluate(() => {
-      return window.app!.api.serverFeatureFlags
+      return window.app!.api.serverFeatureFlags.value
     })
 
     // Verify we received real feature flags from the backend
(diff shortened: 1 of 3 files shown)


christian-byrne (approved):
I think this should be good, although none of the consumers will be taking advantage of the reactivity yet. Don't think it will break anything or change the public api at all.

PR #9057, opened 2026-02-21
fix: resolve bookmark lookup for subgraph blueprints

Yourz:
## Summary

Bookmarked subgraph blueprint nodes were not appearing in the favorites tree because `buildBookmarkTree` looked up nodes only in `nodeDefsByName`, which excludes subgraph blueprints.

## Changes

- **What**: Added `allNodeDefsByName` computed in `nodeDefStore` that includes both regular nodes and subgraph blueprints. Updated `nodeBookmarkStore.buildBookmarkTree` to use it for bookmark resolution.

## Review Focus

- `allNodeDefsByName` iterates over `nodeDefs` (which merges `subgraphBlueprints` + `nodeDefsByName`). Confirm this doesn't introduce performance concerns for large node sets.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9057-fix-resolve-bookmark-lookup-for-subgraph-blueprints-30e6d73d365081259bcae9d0c197de43) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/nodeBookmarkStore.ts
+++ b/src/stores/nodeBookmarkStore.ts
@@ -53,7 +53,7 @@ export const useNodeBookmarkStore = defineStore('nodeBookmark', () => {
         const parts = bookmark.split('/')
         const name = parts.pop() ?? ''
         const category = parts.join('/')
-        const srcNodeDef = nodeDefStore.nodeDefsByName[name]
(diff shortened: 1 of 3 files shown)


christian-byrne (approved):
LGTM

PR #9067, opened 2026-02-21
feat: add feature flag to disable Essentials tab in node library

christian-byrne:
## Summary

Add a `node_library_essentials_enabled` feature flag to gate the Essentials tab, allowing the rest of the new node library to ship while the Essentials tab is finalized.

## Changes

- **What**: New feature flag (`node_library_essentials_enabled`) that hides the Essentials tab in the node library sidebar and search category sidebar. Defaults to `true` in dev/nightly builds, `false` in production. Overridable via remote config or server feature flags. Falls back to the "All" tab if a user previously had Essentials selected.

**Disabled UI**

image

**Enabled UI**

image

## Review Focus

- Feature flag pattern follows existing conventions (e.g. `linearToggleEnabled`)
- Fallback behavior when essentials tab was previously selected by user

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9067-feat-add-feature-flag-to-disable-Essentials-tab-in-node-library-30e6d73d36508103b3cad9fc5d260611) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useFeatureFlags.ts
+++ b/src/composables/useFeatureFlags.ts
@@ -22,7 +22,8 @@ export enum ServerFeatureFlag {
   LINEAR_TOGGLE_ENABLED = 'linear_toggle_enabled',
   TEAM_WORKSPACES_ENABLED = 'team_workspaces_enabled',
   USER_SECRETS_ENABLED = 'user_secrets_enabled',
-  NODE_REPLACEMENTS = 'node_replacements'
+  NODE_REPLACEMENTS = 'node_replacements',
+  NODE_LIBRARY_ESSENTIALS_ENABLED = 'node_library_essentials_enabled'
 }
 
 /**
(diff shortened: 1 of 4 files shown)


src/composables/useFeatureFlags.ts
```diff
@@ -118,6 +119,17 @@ export function useFeatureFlags() {
     },
     get nodeReplacementsEnabled() {
       return api.getServerFeature(ServerFeatureFlag.NODE_REPLACEMENTS, false)
+    },
+    get nodeLibraryEssentialsEnabled() {
+      if (isNightly || import.meta.env.DEV) return true
```
benceruleanlu, line 124: Was this supposed to be in the souce? This makes it so that it's non-overridable in nightly/dev

benceruleanlu (approved):

src/composables/useFeatureFlags.ts
christian-byrne: Yes, I want to keep the new feature enabled in nightly and dev, but disable it in production builds.

PR #9071, opened 2026-02-22
Make the vue toggle ring surround toggle

AustinMroz:
| Before | After |
| ------ | ----- |
| before | after|

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9071-Make-the-vue-toggle-ring-surround-toggle-30f6d73d365081e0a427cf270ef2763a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/layout/WidgetLayoutField.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/layout/WidgetLayoutField.vue
@@ -1,17 +1,26 @@
 <script setup lang="ts">
+import { computed } from 'vue'
+
 import type { SimplifiedWidget } from '@/types/simplifiedWidget'
 import { useHideLayoutField } from '@/types/widgetTypes'
 import { cn } from '@/utils/tailwindUtil'
 
-const { rootClass } = defineProps<{
+const { widget, rootClass } = defineProps<{
   widget: Pick<
     SimplifiedWidget<string | number | undefined>,
     'name' | 'label' | 'borderStyle'
   >
   rootClass?: string
+  noBorder?: boolean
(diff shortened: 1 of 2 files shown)


src/renderer/extensions/vueNodes/widgets/components/layout/WidgetLayoutField.vue
```diff
@@ -1,16 +1,25 @@
...
 import { useHideLayoutField } from '@/types/widgetTypes'
 import { cn } from '@/utils/tailwindUtil'
 
-defineProps<{
+const { widget } = defineProps<{
   widget: Pick<
     SimplifiedWidget<string | number | undefined>,
     'name' | 'label' | 'borderStyle'
   >
+  noBorder?: boolean
```
DrJKL, line 14: Is there a way we could put this into the widget options? I feel like custom node devs might want to be able to use this.

AustinMroz: :thinking: Since this only applies on WidgetLayoutField, there's not currently any way for a custom node to have a sub-component that adds the border back.

If the option can only be used to disable accessibility (rather than a means of augmenting it as it is here), then I think that's unwise to encourage.

DrJKL: It's more a principle of: Any display specific modifications should ideally be available without having to add frontend code.

christian-byrne (dismissed):
LGTM

christian-byrne (approved):
Rebased onto main, resolved conflicts, all CI checks passing. LGTM.

PR #9072, opened 2026-02-22
feat: add display name mappings for Essentials tab nodes

christian-byrne:
## Summary

Add frontend-only display name mappings for nodes shown in the Essentials tab, plus parse the new `essentials_category` field from the backend.

## Changes

- **What**: Created `src/constants/essentialsDisplayNames.ts` with a static mapping of node names to user-friendly display names (e.g. `CLIPTextEncode` → "Text", `ImageScale` → "Resize Image"). Regular nodes use exact name matching; blueprint nodes use prefix matching since their filenames include model-specific suffixes. Integrated into `NodeLibrarySidebarTab.vue`'s `renderedRoot` computed for leaf node labels with fallback to `display_name`. Added `essentials_category` (z.string().optional()) to the node def schema and `ComfyNodeDefImpl` to parse the field already sent by the backend (PR #12357).

## Review Focus

Display names are resolved only in the Essentials tab tree view (`NodeLibrarySidebarTab.vue`), not globally, to avoid side effects on search, bookmarks, or other views. Blueprint prefix matching is ordered longest-first so more specific prefixes (e.g. `image_inpainting_`) match before shorter ones (e.g. `image_edit`).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9072-feat-add-display-name-mappings-for-Essentials-tab-nodes-30f6d73d3650817c9acdc9b0315ed0be) by [Unito](https://www.unito.io)


christian-byrne:
Do not merge, tweaking some things

src/constants/essentialsDisplayNames.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+const EXACT_NAME_MAP: Record<string, string> = {
+  // Basics
+  LoadImage: 'Load Image',
+  SaveImage: 'Save Image',
+  LoadVideo: 'Load Video',
+  SaveVideo: 'Save Video',
+  Load3D: 'Load 3D model',
+  SaveGLB: 'Save 3D Model',
+  CLIPTextEncode: 'Text',
```
christian-byrne: This should be the multiline string primitive not CLIPTextEncode. See: https://github.com/Comfy-Org/ComfyUI/pull/12553 and read it in full.

src/constants/essentialsDisplayNames.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  // Image Tools
+  ImageBatch: 'Batch Image',
```
christian-byrne: This should be BatchImagesNode not ImageBatch which is deprecated. Also, please spin up a PR to ComfyUI to move the ESSENTIALS_CATEGORY = "Image Tools" from ImageBatch to BatchImagesNode as well (and any other necessary changes).

src/constants/essentialsDisplayNames.ts
```diff
@@ -0,0 +1,102 @@
+import type { ComfyNodeDefImpl } from '@/stores/nodeDefStore'
+
+const BLUEPRINT_PREFIX = 'SubgraphBlueprint.'
+
+/**
+ * Static mapping of node names to their Essentials tab display names.
+ * Only includes nodes that currently exist (status ✅ in the tracker).
```
christian-byrne: Don't put emojis in code comments.

src/constants/essentialsDisplayNames.ts
christian-byrne, line 1: We need a browser_test since the blueprints and nodes are actually served by the real backend in those tests, so we can verify e.g. by just calling object_info and global_blueprints.

src/constants/essentialsDisplayNames.ts
```diff
@@ -0,0 +1,102 @@
...
+  Load3D: 'Load 3D model',
+  SaveGLB: 'Save 3D Model',
+  CLIPTextEncode: 'Text',
+
+  // Image Tools
+  ImageBatch: 'Batch Image',
+  ImageCrop: 'Crop Image',
+  ImageScale: 'Resize Image',
+  ImageRotate: 'Rotate',
+  ImageBlur: 'Blur',
```
christian-byrne: ImageBlur should be using the subgraph blueprint GLSL shader blueprint, not the native ImageBlur. If ImageBlur node in ComfyUI has essentials category then we need a new PR to remove that there and then make sure the blur subgraph blueprint has the right category.

src/constants/essentialsDisplayNames.ts
```diff
@@ -0,0 +1,102 @@
...
+  ImageInvert: 'Invert',
+  Canny: 'Canny',
+  RecraftRemoveBackgroundNode: 'Remove Background',
+  'image compare': 'Image compare',
```
christian-byrne: Are we sure the node name is "image compare"? Can you verify in ComfyUI

src/constants/essentialsDisplayNames.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  // Video Tools
+  GetVideoComponents: 'Extract frame',
```
christian-byrne: This should be VideoSlice not GetVideoComponents. Similar to the other nodes, if ComfyUI has the essentials / main category fields on GetVideoComponents, then we need to make a pr to move those fields to the correct VideoSlice node.

src/constants/essentialsDisplayNames.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  // Video Generation
+  KlingLipSyncAudioToVideoNode: 'Lipsync',
+
+  // Text Generation
+  OpenAIChatNode: 'Text generation (LLM)',
+
+  // 3D
+  TencentTextToModelNode: 'Text to 3D model',
+  TencentImageToModelNode: 'Image to 3D Model',
```
christian-byrne: We should also include the other 3D api nodes available in ComfyUI that do these, to allow future changes in backend and still displaying the expected name. Also see what other instances of api node (model provider bound) are in these mappings and see if there's alternative providers in ComfyUI and suggest other nodes we can add alternatives for here to further future-proof.

src/constants/essentialsDisplayNames.ts
```diff
@@ -0,0 +1,102 @@
...
+  SaveAudio: 'Save Audio'
+}
+
+/**
+ * Blueprint prefix patterns mapped to display names.
+ * Entries are matched by checking if the blueprint filename
+ * (after removing the SubgraphBlueprint. prefix) starts with the key.
+ * Ordered longest-first so more specific prefixes match before shorter ones.
+ */
+const BLUEPRINT_PREFIX_MAP: [prefix: string, displayName: string][] = [
```
christian-byrne: Please verify by calling real server APIs that these will be the actual names returned and make a browser_test.

src/components/sidebar/tabs/NodeLibrarySidebarTab.vue
christian-byrne, line 1: Please verify the backned is actually serving and adding essentials_category to blueprints.

src/constants/essentialsDisplayNames.ts
christian-byrne: Addressed — using `PrimitiveStringMultiline` as specified.

src/constants/essentialsDisplayNames.ts
christian-byrne: Addressed — using `BatchImagesNode` (not `ImageBatch`).

src/constants/essentialsDisplayNames.ts
christian-byrne: Addressed — no emojis in code comments.

src/constants/essentialsDisplayNames.ts
christian-byrne: Removed the browser test — CI backend doesn't serve `essentials_category` so the test can't validate mappings against real data. The unit tests in `essentialsDisplayNames.test.ts` (34 tests) cover the mapping logic thoroughly.

src/constants/essentialsDisplayNames.ts
christian-byrne: Addressed — `ImageBlur` removed from mappings. It should use the subgraph blueprint GLSL shader blueprint instead.

src/constants/essentialsDisplayNames.ts
christian-byrne: Verified — using `ImageCompare` which is the correct node name in ComfyUI.

src/constants/essentialsDisplayNames.ts
christian-byrne: Addressed — using `'Video Slice'` which is the correct node name.

src/constants/essentialsDisplayNames.ts
christian-byrne: Addressed — added all 3D API node alternatives: `TencentTextToModelNode`, `TencentImageToModelNode`, `MeshyTextToModelNode`, `MeshyImageToModelNode`, `TripoTextToModelNode`, `TripoImageToModelNode`. Also added both `KlingLipSyncAudioToVideoNode` and `KlingLipSyncTextToVideoNode` for lipsync.

src/constants/essentialsDisplayNames.ts
christian-byrne: Node names verified against ComfyUI source. Removed the browser test since CI backend doesn't serve `essentials_category`.

src/components/sidebar/tabs/NodeLibrarySidebarTab.vue
christian-byrne: The backend serves `essentials_category` on nodes via `object_info`. Blueprints get their category via `getEssentialsCategory()` which was added to main. The display name mappings here are frontend-only overrides.

src/components/sidebar/tabs/NodeLibrarySidebarTabV2.vue
```diff
@@ -112,6 +112,7 @@ import {
 import { computed, nextTick, onMounted, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
 
+import { resolveEssentialsDisplayName } from '@/constants/essentialsDisplayNames'
```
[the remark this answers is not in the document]
christian-byrne: Fixed — `fillNodeInfo` now accepts a `{ useEssentialsLabels?: boolean }` option (defaults to `false`). Only `renderedEssentialRoot` passes `{ useEssentialsLabels: true }`. The All and Custom tabs use the default `display_name`.

benceruleanlu (approved):

PR #9088, opened 2026-02-22
refactor: migrate 10 components from manual image loading to useImage composable

christian-byrne:
## Summary

Replace manual image-loading patterns (`@load`/`@error` handlers, `imageBroken` refs, `new Image()`) with the `useImage` composable from `@vueuse/core` for a consistent reactive image-loading pattern across components and one composable.

> **Note:** This PR was rebased and rewritten on top of `main` to address all prior review feedback and to drop files whose target state has changed on `main`. See "Scope changes vs. earlier revision" below.

## Changes

- **What:** Migrate components + 1 composable to `useImage`. Remove manual `ref` + `@error`/`@load` handlers in favor of `useImage`'s reactive `error`, `isReady`, and `state`. Update test files to mock `useImage` at the module level instead of triggering DOM events.
- **Tier 1 (simple swap):** `UserAvatar`, `PackBanner`, `linearMode/ImagePreview`
- **Tier 2 (medium):** `vueNodes/LivePreview` (with dimension caching), `vueNodes/ImagePreview` (error-only migration; keeps `@load` for `syncLegacyNodeImgs`)
- **Tier 3 (composable):** `useImageCrop` + `WidgetImageCrop` template cleanup

## Reviewer feedback addressed in this rewrite

- **`UserAvatar.vue`** — wrap the `useImage` source in `computed(() => ({ src: photoUrl, alt: ariaLabel }))` so it re-runs when destructured props change.
- **`vueNodes/LivePreview.vue`** — cache last good `naturalWidth`/`Height` in refs that update only when `isReady` fires. Avoids the placeholder text flickering back to "Calculating dimensions" when `imageUrl` changes during live preview streaming.
- **`vueNodes/ImagePreview.vue`** — keep the `@load` handler on the rendered `` so `syncLegacyNodeImgs` receives the actual DOM `` element. `useImage`'s `state` is an off-DOM `Image()` and is not a substitute for the rendered element.
- **`WidgetImageCrop.vue`** — do **not** add `brightness-50` to the image. The crop overlay already darkens the surroundings via `shadow-[0_0_0_9999px_rgba(0,0,0,0.5)]`; adding `brightness-50` would double-darken.
- **`UserAvatar.test.ts`** — preserve the `@testing-library/vue` (`render`, `screen`) behavioral test style. Mocks `useImage` at the module level to control the `error` ref instead of triggering DOM events.
- **General rule applied throughout** — any time a prop is part of the `useImage` source, the source is wrapped in `computed()` so reactivity is preserved.

## Review focus

- `useImage` creates an internal `Image()` element for load/error detection; the template `` loads separately. Browsers cache identical URLs, so there's no double network fetch.
- `vueNodes/ImagePreview` carousel: `useImage` auto-resets `error` when the reactive source changes, replacing the manual resets previously in `setCurrentIndex` and the `imageUrls` watcher.
- `useImageCrop`: load/error detection moved from `@load`/`@error` template events to `watch(isReady)` / `watch(imageLoadError)`. Timing is equivalent (both fire on `Image.onload`). The `imageCropLoadingAfterUrlChange` helper and its tests were dropped — the behaviour is now expressed inline.
- `PackBanner`: the template now uses `v-if`/`v-else` branches on `isImageError` instead of a ternary on `:src` for clearer reading order.

## Scope changes vs. earlier revision

The following files were in the earlier revision of this PR but are **dropped** because `main` now diverges:

- `src/components/common/ComfyImage.vue` — already migrated on `main` (PR #11759 et al.) using the canonical `useImage(computed(() => ({ src, alt })))` pattern.
- `src/components/queue/job/QueueAssetPreview.vue` — deleted on `main` in PR #11621 ("Remove unused queue job components").
- `src/platform/assets/components/Media3DTop.vue` — refactored on `main` to use intersection-observer–driven async loading, with no `@error` handler left to migrate.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9088-refactor-migrate-10-components-from-manual-image-loading-to-useImage-composable-30f6d73d36508165bf8dd96ad38d1b86) by [Unito](https://www.unito.io)


christian-byrne (commented):
~Could we split this PR into vertical slices or stack it with graphite?~ Actually this wouldn't really make sense to do here.

viva-jinyi (dismissed):
👍

src/components/common/ComfyImage.vue
```diff
@@ -46,10 +39,7 @@ const {
   alt?: string
 }>()
 
-const imageBroken = ref(false)
-const handleImageError = () => {
-  imageBroken.value = true
-}
+const { error: imageBroken } = useImage({ src, alt })
```
viva-jinyi: **Breaking:** The current main branch already uses `useImage(computed(() => ({ src, alt })))` with a `computed()` wrapper. This PR removes the reactivity wrapper, so if `src` or `alt` change, `useImage` won't detect it.

```suggestion
const { error: imageBroken } = useImage(computed(() => ({ src, alt })))
```

src/components/common/UserAvatar.vue
```diff
@@ -6,22 +6,22 @@
...
 const { photoUrl, ariaLabel } = defineProps<{
   photoUrl?: string | null
   ariaLabel?: string
 }>()
 
-const imageError = ref(false)
-const handleImageError = () => {
-  imageError.value = true
-}
+const { error: imageError } = useImage({
```
viva-jinyi: **Reactivity issue:** `photoUrl` and `ariaLabel` are destructured props — this creates a static object at setup time. If `photoUrl` changes (e.g., user updates avatar), `useImage` won't re-run. Wrap in `computed()`:

```suggestion
const { error: imageError } = useImage(
  computed(() => ({ src: photoUrl ?? '', alt: ariaLabel ?? '' }))
)
```

src/components/queue/job/QueueAssetPreview.vue
```diff
@@ -40,26 +38,25 @@
...
   name: string
   timeLabel?: string
 }>()
 
 defineEmits(['image-click'])
 
-const imgRef = ref<HTMLImageElement | null>(null)
-const width = ref<number | null>(null)
-const height = ref<number | null>(null)
+const { state, isReady } = useImage({ src: imageUrl, alt: name })
```
viva-jinyi: Same reactivity issue — `imageUrl` and `name` are destructured props, creating a static object. Wrap in `computed()` for reactivity:

```suggestion
const { state, isReady } = useImage(computed(() => ({ src: imageUrl, alt: name })))
```

src/components/imagecrop/WidgetImageCrop.vue
```diff
@@ -28,9 +28,7 @@
         :src="imageUrl"
         :alt="$t('imageCrop.cropPreviewAlt')"
         draggable="false"
-        class="block size-full object-contain select-none"
-        @load="handleImageLoad"
-        @error="handleImageError"
+        class="block size-full object-contain brightness-50 select-none"
```
viva-jinyi: **Visual regression:** The crop overlay already uses `shadow-[0_0_0_9999px_rgba(0,0,0,0.5)]` to darken the area outside the crop box. Adding `brightness-50` on the image itself causes double-darkening — the crop area appears at 50% brightness and the outside at ~25%. Remove `brightness-50`.

```suggestion
        class="block size-full object-contain select-none"
```

src/renderer/extensions/vueNodes/components/LivePreview.vue
```diff
@@ -26,37 +24,28 @@
...
-  if (img.naturalWidth && img.naturalHeight) {
-    actualDimensions.value = `${img.naturalWidth} x ${img.naturalHeight}`
-  }
-}
-
-const handleImageError = () => {
-  imageError.value = true
-  actualDimensions.value = null
-}
+const actualDimensions = computed(() => {
```
viva-jinyi: **Flicker regression:** The original code intentionally kept previous dimensions when the URL changes to avoid flashing "Calculating dimensions" during live preview. Since `useImage` uses `resetOnExecute: true` by default, every URL change resets `state` to `undefined`, causing `actualDimensions` to immediately return `null` and flash the calculating text on every frame.

Consider keeping `actualDimensions` as a `ref` and updating it from a `watch(isReady, ...)` — only clearing it when the image actually errors.

src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -193,37 +198,36 @@ watch(
...
-    actualDimensions.value = `${img.naturalWidth} x ${img.naturalHeight}`
+watch(imageIsReady, (ready) => {
+  if (ready) {
+    stopDelayedLoader()
+    showLoader.value = false
+    if (imageState.value?.naturalWidth && imageState.value?.naturalHeight) {
+      actualDimensions.value = `${imageState.value.naturalWidth} x ${imageState.value.naturalHeight}`
+    }
+    if (props.nodeId && imageState.value) {
+      nodeOutputStore.syncLegacyNodeImgs(
```
viva-jinyi: The original code passed `event.target` (the actual `` DOM element) to `syncLegacyNodeImgs`. This now passes `imageState.value`, which is the off-DOM `Image()` object created internally by `useImage`. If downstream code relies on the image being in the DOM (e.g., as a canvas source with CORS, or reading computed styles), this could cause subtle issues.

src/components/common/UserAvatar.test.ts
```diff
@@ -3,12 +3,27 @@ import type { ComponentProps } from 'vue-component-type-helpers'
...
-import { createApp, nextTick } from 'vue'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { createApp, nextTick, ref } from 'vue'
 import { createI18n } from 'vue-i18n'
 
 import UserAvatar from './UserAvatar.vue'
 
+const mockImageError = ref(false)
+
+vi.mock('@vueuse/core', async () => {
```
viva-jinyi, line 14: **Test quality concerns:**

1. **Framework regression:** The existing test on main uses `@testing-library/vue` (`render`, `screen`, `fireEvent`) for user-centric behavioral tests. This PR replaces it with `@vue/test-utils` + module-level mocks. Per project guidelines: "prefer @testing-library/vue for user-centric, behavioral tests."

2. **Tests now test the mock, not the behavior:** Setting `mockImageError.value = true` and asserting the UI changed only tests Vue's reactivity system. The original tests triggered actual DOM events (`avatar.vm.$emit('error')`) which tested the full integration path. If someone breaks the `useImage` integration (e.g., forgets to pass `src`), these tests would still pass.

viva-jinyi (dismissed):

christian-byrne (commented):
There are many merge conflicts with `main`

src/components/common/UserAvatar.test.ts
```diff
@@ -3,12 +3,27 @@ import type { ComponentProps } from 'vue-component-type-helpers'
...
+  return {
+    ...actual,
+    useImage: () => ({
+      state: ref(undefined),
+      error: mockImageError,
+      isReady: ref(false),
+      isLoading: ref(false)
+    })
+  }
+})
```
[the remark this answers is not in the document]
christian-byrne: Addressed in the rewrite (commit c3ec808). The new test uses `vi.hoisted()` for the mock state. The ref itself is initialized inside the `vi.mock` factory (since `vi.hoisted`'s synchronous body runs before any `vue` import is evaluated), and `beforeEach` resets the value to keep tests isolated.

src/components/queue/job/QueueAssetPreview.vue
```diff
@@ -40,26 +38,25 @@
...
-  if (!el) return
-  width.value = el.naturalWidth || null
-  height.value = el.naturalHeight || null
-}
+const width = computed(() =>
+  isReady.value ? (state.value?.naturalWidth ?? null) : null
+)
+const height = computed(() =>
+  isReady.value ? (state.value?.naturalHeight ?? null) : null
+)
```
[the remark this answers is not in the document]
christian-byrne: Dropped from this PR — `QueueAssetPreview.vue` was deleted on `main` in #11621 ("Remove unused queue job components"), so there's nothing to migrate.

src/renderer/extensions/vueNodes/components/ImagePreview.test.ts
```diff
@@ -2,7 +2,7 @@ import { createTestingPinia } from '@pinia/testing'
 import type { VueWrapper } from '@vue/test-utils'
 import { mount } from '@vue/test-utils'
 import { afterEach, describe, expect, it, vi } from 'vitest'
-import { nextTick } from 'vue'
+import { nextTick, ref } from 'vue'
```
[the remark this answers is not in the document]
christian-byrne: Addressed in the rewrite (commit 6194a4f). The new test uses `vi.hoisted()` to declare the mock holder, with the actual `ref` created inside the `vi.mock` factory and reset in `beforeEach`. There's no longer any module-scoped mutable state outside the hoisted holder.

src/workbench/extensions/manager/components/manager/packBanner/PackBanner.vue
```diff
@@ -45,8 +48,10 @@ const { nodePack } = defineProps<{
...
 }>()
 
-const isImageError = ref(false)
-
 const showDefaultBanner = computed(() => !nodePack.banner_url && !nodePack.icon)
 const imgSrc = computed(() => nodePack.banner_url || nodePack.icon)
+
+const { error: isImageError } = useImage(
+  computed(() => ({ src: imgSrc.value ?? '', alt: nodePack.name + ' banner' }))
+)
```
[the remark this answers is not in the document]
christian-byrne: Acknowledged. The hardcoded `' banner'` string predates this PR (it was already on `main` as `:alt="nodePack.name + ' banner'"`), so I'm leaving the i18n fix out of scope here to keep the migration focused. Happy to file this as a follow-up if you'd like.

src/components/common/ComfyImage.vue
christian-byrne: You're right — `ComfyImage.vue` already has the canonical `useImage(computed(() => ({ src, alt })))` pattern on `main`. The rewrite drops this file from the PR entirely; no behavioural change to `ComfyImage`.

src/components/common/UserAvatar.vue
christian-byrne: Fixed in the rewrite (commit c3ec808). The source is now wrapped in `computed(() => ({ src: photoUrl ?? '', alt: ariaLabel ?? '' }))` so `useImage` re-runs whenever the destructured props change. Test coverage is preserved with the existing behavioral assertions.

src/components/queue/job/QueueAssetPreview.vue
christian-byrne: N/A — `QueueAssetPreview.vue` was deleted on `main` in #11621 and is dropped from this PR.

src/components/imagecrop/WidgetImageCrop.vue
christian-byrne: Fixed in the rewrite (commit 4ef4aae). The migrated `WidgetImageCrop.vue` no longer adds `brightness-50` — the only template change is removing the obsolete `@load`/`@error` handlers. The shadow overlay continues to provide the darkening, so there is no double-darken.

src/renderer/extensions/vueNodes/components/LivePreview.vue
christian-byrne: Fixed in the rewrite (commit d27e1ef). `LivePreview.vue` now caches the last good `naturalWidth`/`Height` in refs that are updated only when `isReady` fires. `useImage`'s reset on source change no longer flashes "Calculating dimensions" mid-streaming. There's a new test ("keeps last good dimensions when imageUrl changes (no flicker)") covering this.

src/renderer/extensions/vueNodes/components/ImagePreview.vue
christian-byrne: Fixed in the rewrite (commit 6194a4f). The migration is intentionally partial: `useImage` is used only for error detection (replacing the `imageError` ref), while the `@load` handler stays on the rendered `` so `syncLegacyNodeImgs` continues to receive the actual DOM `` element via `event.target`. `imageState.value` (the off-DOM `Image()`) is never passed downstream.

src/components/common/UserAvatar.test.ts
christian-byrne: Fixed in the rewrite (commit c3ec808). The test now uses `@testing-library/vue` (`render`, `screen`) again, matching the original behavioral style and the project guideline. `useImage` is mocked at the module level only because the `@error` handler no longer exists on the rendered ``; the integration path through the component template (Avatar slot/icon swap based on `hasAvatar`) is still exercised via real rendering and `screen.getBy*` assertions.

christian-byrne:
@coderabbitai review

src/components/common/UserAvatar.vue
```diff
@@ -11,22 +11,24 @@
...
 const { photoUrl, ariaLabel } = defineProps<{
   photoUrl?: string | null
   ariaLabel?: string
 }>()
 
-const imageError = ref(false)
-const handleImageError = () => {
-  imageError.value = true
-}
+const { error: imageError } = useImage(
```
DrJKL, line 27: The most defensible photo in the album: PrimeVue `Avatar` hides its internal ``, so you genuinely can't bind `@error` to it — probing with `useImage` is a reasonable way to recover the error state. Worth knowing it does load the avatar a second time, but for a small cached avatar that's negligible. No change requested; flagging so the tradeoff is a choice, not an accident.

src/renderer/extensions/vueNodes/components/LivePreview.vue
```diff
@@ -26,37 +24,46 @@
...
-    // Reset error state when URL changes, but keep previous dimensions
-    // to avoid flickering "Calculating dimensions" text during live preview
-    imageError.value = false
-  }
+const {
+  state: imageState,
+  isReady,
+  error
+} = useImage(
+  computed(() => ({ src: props.imageUrl, alt: t('g.liveSamplingPreview') }))
```
DrJKL: **Hot path — main concern.** Live sampling preview swaps `imageUrl` many times/sec, and `useImage` fires a separate `new Image().src = ...` for every frame on top of the rendered ``. These are usually `type=temp` URLs that don't cache, so this roughly doubles network + decode on the streaming path. The native `@load`/`@error` this replaced were free and ran on the element already on screen. Suggest keeping native events here even if the rest of the album uses `useImage` — the consistency isn't worth doubling the live-preview load.

src/renderer/extensions/vueNodes/components/LivePreview.vue
```diff
@@ -26,37 +24,46 @@
...
 
 interface LivePreviewProps {
   imageUrl: string
 }
 
 const props = defineProps<LivePreviewProps>()
 
-const actualDimensions = ref<string | null>(null)
-const imageError = ref(false)
+const { t } = useI18n()
```
DrJKL: `useI18n` is pulled in only to feed `alt: t('g.liveSamplingPreview')` into `useImage` (line 44) — but `useImage`'s `alt` is never rendered anywhere, so it's a phantom arg. The template `` already sets its own `:alt`. If you keep `useImage`, drop the `alt` and this `useI18n` import goes with it.

src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -271,6 +269,20 @@ const gridCols = computed(() => {
   return Math.max(Math.round(Math.sqrt(imageUrls.length * bias)), 1)
 })
 
+// Use useImage for error detection only. Load handling stays on the rendered
```
DrJKL: This photo only half-fits the album, and the comment says why: `@load` has to stay on the real `` (you need the DOM element for `syncLegacyNodeImgs` + dimensions), so `useImage` is here *purely* to catch `@error`. Net result: the image loads twice **and** the native `@error` you removed would have caught the same failure on the element already loading. Suggest reverting error handling to native `@error="handleImageError"` here and dropping `useImage` from this file — it's the one spot where the old pattern was strictly better.

src/workbench/extensions/manager/components/manager/packBanner/PackBanner.vue
```diff
@@ -21,21 +21,24 @@
       ></div>
       <!-- image -->
       <img
```
DrJKL, line 23: The single `` with conditional `:src`/`:class` became two `v-if`/`v-else` imgs plus a `useImage` probe (line 55). The previous `@error="isImageError = true"` on the one rendered img was correct and shorter, with no extra fetch. This is more template for the same behavior. Reasonable to revert to the single-img + native `@error` shape; if you keep `useImage`, at least the two-img split could collapse back to one `` with a computed `:src`/`:class`.

src/renderer/extensions/linearMode/ImagePreview.test.ts
```diff
@@ -0,0 +1,126 @@
...
+import type { Ref } from 'vue'
+import { nextTick } from 'vue'
+import { createI18n } from 'vue-i18n'
+
+const useImageMock = vi.hoisted(() => ({
+  state: null as Ref<HTMLImageElement | undefined> | null,
+  isReady: null as Ref<boolean> | null
+}))
+
+vi.mock('@vueuse/core', async () => {
```
DrJKL, line 13: Same `vi.hoisted` + `vi.mock('@vueuse/core')` `useImage` stub is hand-rolled in 6 specs, each a slightly different shape (`error`-only here vs `state`+`isReady`+`error` in LivePreview). If the `useImage` pattern survives review, extract one `mockUseImage()` test helper so the album has a single negative.

Secondary: these tests now flip `useImageMock.error.value` directly instead of `fireEvent.error(img)`, so they assert "component reads the mock's error ref" rather than "component handles a broken image" — coupled to `useImage`'s internals (don't-mock-what-you-don't-own). If native events come back per the notes above, these revert to real `fireEvent` and get stronger for free.

src/renderer/extensions/linearMode/ImagePreview.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  const actual = await vi.importActual('@vueuse/core')
```
DrJKL: Nit: typing `importActual` lets you drop the `as Record<string, unknown>` cast on line 18 (`...actual` then type-checks).
```suggestion
  const actual = await vi.importActual<typeof import('@vueuse/core')>('@vueuse/core')
```

DrJKL (commented):
Nice consolidation — you've taken a scattered set of one-off `@load`/`@error` handlers (loose photos floating around five components) and collated them into one cohesive `useImage` pattern (an album). The shared shape reads well and the test scaffolding is consistent across files.

My one structural concern is the mounting medium: `useImage` constructs its own `new Image()` to probe the same `src` the template already renders, so each image now loads twice. For most of these it's a fine tradeoff; for `LivePreview` (streaming, swaps URL many times/sec) it's worth a second look. Inline notes below, grouped by where they land. Most are small; nothing here blocks the cohesion win.

Nice cleanup not called out inline: deleting `imageCropLoadingAfterUrlChange` + its 5 change-detector tests and inlining the watch shrinks the exported surface. 👍

src/renderer/extensions/vueNodes/components/LivePreview.vue
DrJKL: @christian-byrne This is still the one I'm not sure about.

DrJKL (dismissed):
I think the Preview ones will be worth watching for performance issues or network load.

christian-byrne:
@coderabbitai review

christian-byrne:
Rebased onto current `main`; conflicts resolved; focused checks green.

Resolved the `ImagePreview` conflict by keeping the PR's `useImage` error state while preserving main's image-load-failure telemetry in the new error watcher. Kept the PR's reactive VueUse test mocks and reset them between tests in both affected suites.

Verification:
- `pnpm typecheck`
- 7 affected Vitest files, 95 tests passed

New head: [`ef2921fd0f`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/ef2921fd0f)

</details>

christian-byrne:
Rebased onto current `main` to resolve merge conflicts. No conflicts encountered during rebase (clean fast-forward through 12 commits); pushed as `4347839a7a` (force-with-lease).

## Evidence

- `git rebase origin/main` — clean, zero conflicts
- `pnpm typecheck` (`vue-tsc --noEmit`) — passed
- Affected vitest specs (7 files matching the 7 changed source files) — 95/95 passed:
  - `src/components/common/UserAvatar.test.ts`
  - `src/components/imagecrop/WidgetImageCrop.test.ts`
  - `src/composables/useImageCrop.test.ts`
  - `src/renderer/extensions/linearMode/ImagePreview.test.ts`
  - `src/renderer/extensions/vueNodes/components/ImagePreview.test.ts`
  - `src/renderer/extensions/vueNodes/components/LivePreview.test.ts`
  - `src/workbench/extensions/manager/components/manager/packBanner/PackBanner.test.ts`
- Post-push `gh pr view` confirms `mergeable: MERGEABLE`.

src/components/common/UserAvatar.vue
christian-byrne: Kept `useImage` in `UserAvatar`: PrimeVue `Avatar` still does not expose its internal image error event, so the small cached-avatar probe remains the intentional tradeoff.

src/renderer/extensions/vueNodes/components/LivePreview.vue
christian-byrne: Restored native rendered-element load/error events in [`5bb8c393f3`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/5bb8c393f3), so live previews no longer issue a second probe request.

src/renderer/extensions/vueNodes/components/LivePreview.vue
christian-byrne: Removed the unused probe `alt` and its `useI18n` dependency in [`9161a83e9e`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/9161a83e9e).

src/renderer/extensions/vueNodes/components/ImagePreview.vue
christian-byrne: Restored native `@error` on the rendered Vue-node image and removed the `useImage` probe in [`e706ab9271`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/e706ab9271). The test now dispatches the real DOM error event.

src/workbench/extensions/manager/components/manager/packBanner/PackBanner.vue
christian-byrne: Collapsed this back to one rendered image with native `@error` and removed the probe in [`377b72b071`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/377b72b071). The fallback test now exercises the DOM event.

src/renderer/extensions/linearMode/ImagePreview.test.ts
christian-byrne: The surviving Vue-node and pack-banner mocks are gone in [`e706ab9271`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/e706ab9271) and [`377b72b071`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/377b72b071); both tests now dispatch native image errors. The live-preview mock was removed with its probe earlier.

src/renderer/extensions/linearMode/ImagePreview.test.ts
christian-byrne: Applied the typed `vi.importActual` form in [`8c5814e0fb`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/8c5814e0fb), removing the record cast.

PR #9094, opened 2026-02-22
feat: add serverCapabilities module and migrate consumers

### LINKED ISSUE #9079: Deliver server feature flags via global object instead of WebSocket
## Problem

Server feature flags are static constants set at server startup (`comfy_api/feature_flags.py`), yet they're delivered via WebSocket message exchange:

```
WS open → client sends flags → server responds with flags
```

This causes:

- **Race conditions** — Components mount before the WS message arrives. Already caused bugs (PR #9037, #9039) requiring `ref()` conversion and event listener workarounds.
- **Unnecessary complexity** — Reactive wrappers, event listeners, and timing-sensitive init logic for data that never changes.
- **Wasted round-trip** — Two WS messages for data the server knows at startup.

## Proposal

Deliver server feature flags as a **global object** available before app initialization, eliminating all timing issues.

### Server flags are static

```python
# comfy_api/feature_flags.py
SERVER_FEATURE_FLAGS = {
    "supports_preview_metadata": True,
    "max_upload_size": args.max_upload_size * 1024 * 1024,
    "extension": {"manager": {"supports_v4": True}},
    "node_replacements": True,
}
```

These values are set once at process start and never change.

### Client → Server direction stays on WS

The server reads client-sent `supports_preview_metadata` per-connection to decide image preview format (`progress.py`, `main.py`). This is per-connection capability negotiation and must remain on WebSocket. Only the **server → client** direction changes.

### No HTML template exists

The backend serves `index.html` as a static file (`web.FileResponse`), no templating engine.

### Option A: Inject `<script>` into HTML response

Modify `get_root` to read `index.html` and inject `<script>window.__SERVER_FEATURE_FLAGS__=...</script>` into `<head>`.

- (+) Zero additional requests, data available before any JS
- (-) HTML string manipulation is fragile, mixes concerns

### Option B: Dynamic `/app-data.js` endpoint

Serve a JS file that sets global variables:

```python
@routes.get("/app-data.js")
async def get_app_data(request):
    data = json.dumps(feature_flags.get_server_features())
    js = f'window.__SERVER_FEATURE_FLAGS__={data};'
    return web.Response(text=js, content_type='application/javascript')
```

Frontend loads it as a blocking script in `<head>`:

```html
<script src="/app-data.js"></script>
```

- (+) Clean separation, extensible (can add version/system info later), easy to test
- (-) One additional blocking HTTP request (tiny payload, same origin)

**Recommendation: Option B** — cleaner, more extensible, and the request overhead is negligible.

### Backend

1. Add `/app-data.js` endpoint serving `window.__SERVER_FEATURE_FLAGS__`
2. Remove server flag response from WS `feature_flags` handler (keep client → server direction)
3. Optionally deprecate `GET /features` REST endpoint

### Frontend

1. Add `<script src="/app-data.js"></script>` to `index.html`
2. `useFeatureFlags.ts` — read from `window.__SERVER_FEATURE_FLAGS__` instead of `api.getServerFeature()`; no reactivity needed since data is static
3. `api.ts` — remove `serverFeatureFlags` ref, remove `feature_flags` WS case, remove `getServerFeature()`/`getServerFeatures()`/`serverSupportsFeature()` methods
4. `app.ts` — remove `feature_flags` event listener, call `nodeReplacementStore.load()` directly during init
5. Add `Window.__SERVER_FEATURE_FLAGS__` type declaration

## Migration Path

1. **Backend PR**: Add `/app-data.js`, stop sending server flags on WS
2. **Frontend PR**: Read from global object, remove WS reception logic, simplify composable and api

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9079-Deliver-server-feature-flags-via-global-object-instead-of-WebSocket-30f6d73d36508123a370c126862ca9e0) by [Unito](https://www.unito.io)

dante01yoon:
## Summary
- Introduces `src/services/serverCapabilities.ts` — fetches server capabilities via `GET /api/features` before app mount with retry logic (3 attempts)
- `getServerCapability(key, default)` provides dot-notation access with dev override support
- Internal consumers (`useFeatureFlags`, `nodeReplacementStore`, `useManagerState`) now use `getServerCapability()` instead of `api.getServerFeature()`
- `app.ts`: direct `load()` call instead of `feature_flags` event listener
- `api.ts` methods (`getServerFeature`, `serverSupportsFeature`, etc.) marked `@deprecated`; still functional via WS for extension compatibility

**Part 1 of 2** for #9079

## Test plan
- [x] 63 unit tests pass (serverCapabilities, consumer tests, api feature flag tests)
- [x] `pnpm typecheck`, `pnpm lint`, `pnpm knip` clean

Fixes #9079


src/services/serverCapabilities.ts
```diff
@@ -0,0 +1,41 @@
+import { get } from 'es-toolkit/compat'
+
+import { isCloud } from '@/platform/distribution/types'
+import { getDevOverride } from '@/utils/devFeatureFlagOverride'
+
+const EMPTY: Readonly<Record<string, unknown>> = Object.freeze({})
+const MAX_RETRIES = 2
+
+let capabilities: Readonly<Record<string, unknown>> = EMPTY
```
christian-byrne, line 9: suggestion (non-blocking): Module-level mutable state without reset capability makes testing harder. Consider using a Pinia store (e.g. `useServerCapabilitiesStore` or adding to `bootstrapStore`) which would provide `$reset()` for tests and align with how other app state is managed.

src/services/serverCapabilities.ts
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export async function initServerCapabilities(): Promise<void> {
+  const url = `${getApiBase()}/api/features`
+
+  for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
+    try {
+      const res = await fetch(url, { cache: 'no-store' })
+      if (res.ok) {
+        capabilities = Object.freeze(await res.json())
```
christian-byrne, line 24: suggestion: The JSON response is assigned without schema validation. If a Zod schema exists (or could be created) for server capabilities, validating the response would guard against malformed server responses affecting security-sensitive flags like `teamWorkspacesEnabled`.

src/main.ts
```diff
@@ -14,13 +14,16 @@ import { VueFire, VueFireAuth } from 'vuefire'
...
 import router from '@/router'
+import { initServerCapabilities } from '@/services/serverCapabilities'
 import { useBootstrapStore } from '@/stores/bootstrapStore'
 
 import App from './App.vue'
 // Intentionally relative import to ensure the CSS is loaded in the right order (after litegraph.css)
 import './assets/css/style.css'
 import { i18n } from './i18n'
 
+await initServerCapabilities()
```
christian-byrne, line 32: question: Does the top-level `await initServerCapabilities()` block app mount? With `MAX_RETRIES=2` and no per-request timeout, this could delay initialization on slow/failing networks. Would it be safer to run this in parallel with non-dependent setup, or defer with defaults until resolved?

src/scripts/app.ts
```diff
@@ -766,9 +766,7 @@ export class ComfyApp {
       releaseSharedObjectUrl(blobUrl)
     })
 
-    api.addEventListener('feature_flags', () => {
-      void useNodeReplacementStore().load()
-    })
+    void useNodeReplacementStore().load()
```
christian-byrne, line 785: question: This changes the timing fix from commit 2b69d7b49 which used an event listener to ensure feature flags loaded before calling `useNodeReplacementStore().load()`. The new approach calls `load()` directly since `initServerCapabilities()` runs at startup. Is this timing guaranteed safe? Also, was there a test covering the original bugfix, or should there be a browser_test to guard against timing regressions here?

src/main.ts
christian-byrne, line 32: nitpick (non-blocking): No integration tests verify app startup when `initServerCapabilities()` fails or times out. A browser_test (Playwright) could catch regressions in the startup flow.

src/composables/useFeatureFlags.ts
```diff
@@ -36,22 +36,22 @@ function resolveFlag<T>(
...
+  return remoteConfigValue ?? getServerCapability(flagKey, defaultValue)
 }
 
 /**
- * Composable for reactive access to server-side feature flags
+ * Composable for reactive access to feature flags
  */
 export function useFeatureFlags() {
   const flags = reactive({
     get supportsPreviewMetadata() {
```
christian-byrne, line 51: nitpick (non-blocking): The flags object mixes three access patterns: direct `getServerCapability()`, `resolveFlag()` with remoteConfig, and custom inline logic. A brief comment indicating which pattern applies to which category of flag would help future contributors.

src/services/serverCapabilities.ts
```diff
@@ -0,0 +1,41 @@
...
+        return
+      }
+    } catch {
```
christian-byrne: nitpick (non-blocking): The empty catch block silently retries on any error including JSON parse errors. Consider logging in debug mode or differentiating retryable (network) vs non-retryable (parse) errors.

src/composables/useFeatureFlags.test.ts
```diff
@@ -6,16 +6,12 @@ import {
...
 import * as distributionTypes from '@/platform/distribution/types'
-import { api } from '@/scripts/api'
+import * as serverCapabilities from '@/services/serverCapabilities'
 
-// Mock the API module
-vi.mock('@/scripts/api', () => ({
-  api: {
-    getServerFeature: vi.fn()
-  }
+vi.mock('@/services/serverCapabilities', () => ({
```
christian-byrne, line 11: question: These tests primarily verify mocks return what mocks were told to return. Is there a specific integration scenario that would benefit from testing without mocking serverCapabilities, or is the current approach sufficient for this layer?

src/scripts/api.ts
```diff
@@ -1311,35 +1303,23 @@ export class ComfyApi extends EventTarget {
     return (await axios.get(this.apiURL('/i18n'))).data
   }
 
-  /**
-   * Checks if the server supports a specific feature.
-   * @param featureName The name of the feature to check (supports dot notation for nested values)
-   * @returns true if the feature is supported, false otherwise
-   */
+  /** @deprecated Use `getServerCapability()` from `@/services/serverCapabilities` */
```
christian-byrne, line 1368: question: These methods are marked `@deprecated` with migration to `getServerCapability()`. Could you clarify: are they deprecated because they rely on WebSocket-delivered feature flags being phased out in favor of REST, or is there another reason? Understanding this helps know when it's safe to remove them.

src/scripts/api.ts
```diff
@@ -339,10 +338,8 @@ export class ComfyApi extends EventTarget {
     return { ...defaultClientFeatureFlags }
   }
 
-  /**
-   * Feature flags received from the backend server.
-   */
-  serverFeatureFlags = ref<Record<string, unknown>>({})
+  /** @deprecated Use `getServerCapability()` from `@/services/serverCapabilities` */
+  serverFeatureFlags: Record<string, unknown> = {}
```
christian-byrne, line 353: issue: `serverFeatureFlags` was changed from a Vue ref (commit de131133b) to a plain object. If any code paths still reactively watch `api.serverFeatureFlags`, this could break reactivity. Have all consumers been migrated away from reactive access?

src/services/serverCapabilities.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    it('returns value for existing key', () => {
+      expect(getServerCapability('supports_preview_metadata')).toBe(true)
+    })
+
+    it('returns default value for missing key', () => {
+      expect(getServerCapability('non_existent', 'fallback')).toBe('fallback')
+    })
+
+    it('supports dot notation for nested values', () => {
```
christian-byrne, line 118: suggestion (non-blocking): Consider adding edge case tests for deeply nested paths that don't exist, e.g. `getServerCapability('extension.non_existent.deep.path')`.

src/main.ts
dante01yoon: @christian-byrne This is something I've considered. Yes, top-level await blocks app mount — however, I think it makes sense that client-side rendered apps should run after all external dependencies are injected. (We already do this for cloud builds with refreshRemoteConfig in main.ts.)

I think there are two options:

1. Non-blocking, parallel feature flag fetching (AS-IS)
No render blocking, but code must be defensive enough to handle all timing cases (before/after feature flag fetch).

2. Inject dependencies before render — top-level await, or inject via script tag

Top-level await isn't the real concern here — we could avoid it by using a separate script tag, but it would still block render. For example, the Python server could generate app-data.js with all feature flags:

```html
<script type="module" src="app-data.js"></script>
<script type="module" src="app.js"></script>
The advantage of this approach is that devs can focus purely on feature logic without needing to handle empty/loading data states. I think this is more suitable for client-side rendered apps.

```

src/services/serverCapabilities.ts
dante01yoon: Good point on testability. The reason I chose a plain module over a Pinia store is that initServerCapabilities() must run before Pinia is created (line 49 of main.ts — `createPinia()` comes after). This is intentional: server capabilities should be available as a low-level dependency that doesn't require the Vue runtime.

For testing, I've used vi.mock on the module which works well — but I'm open to adding a resetForTesting() export if that would help.

src/services/serverCapabilities.ts
dante01yoon: Agreed this would add safety. Currently the server response is a flat/shallow object and consumers already provide defaults via `getServerCapability(key, defaultValue)`, so malformed data degrades gracefully. That said, adding a Zod schema is a reasonable follow-up — I'll create an issue to track it.

src/scripts/app.ts
dante01yoon: Yes — `initServerCapabilities()` is awaited in `main.ts` before the app mounts, so by the time `app.ts` runs and calls `load()`, capabilities are already populated. The previous event listener pattern was needed because WS-delivered feature flags arrived asynchronously; with the new REST-based fetch-before-mount approach, the data is guaranteed available.

Regarding test coverage: the existing unit tests in `serverCapabilities.test.ts` and `nodeReplacementStore.test.ts` cover the data flow. A browser_test for the full startup sequence is a good idea — I'll consider adding one.

src/main.ts
dante01yoon: Good suggestion. The unit tests cover retry + fallback to empty object, but a Playwright test covering the full startup-with-failure path would be more robust. I'll add this as a follow-up.

src/composables/useFeatureFlags.ts
dante01yoon: Fair observation. The three patterns exist because flags have different resolution needs:
- **Direct**: server-only flags (e.g. `supportsPreviewMetadata`)
- **resolveFlag()**: flags that can be overridden by remoteConfig or localStorage (e.g. `modelUploadButtonEnabled`)
- **Inline**: flags with extra conditional logic like `isNightly` checks

Added brief comments grouping these categories.

src/services/serverCapabilities.ts
dante01yoon: Good catch. Updated to differentiate `SyntaxError` (JSON parse failure) from network errors — parse errors now break out of the retry loop immediately with a warning. Network errors and non-ok responses continue to retry silently. (`console.debug` is not allowed by lint rules, so per-attempt logging is omitted.)

src/composables/useFeatureFlags.test.ts
dante01yoon: The value of these tests is verifying the **wiring** — that each flag getter calls `getServerCapability` with the correct path and default. They also cover the `resolveFlag` priority chain (localStorage > remoteConfig > server). The actual integration between `useFeatureFlags` and the real `serverCapabilities` module is implicitly tested through the browser tests that exercise the full app. Given the composable is essentially a mapping layer, I think this level of coverage is appropriate.

src/scripts/api.ts
dante01yoon: Exactly. The WS-based `feature_flags` event will continue to fire for backward compatibility with extensions, but the canonical source is now the REST `GET /api/features` endpoint fetched at startup. The deprecated methods still work via the WS path, but new code should use `getServerCapability()`. We'll remove the deprecated methods once we confirm no active extensions depend on them.

src/scripts/api.ts
dante01yoon: All internal consumers have been migrated to `getServerCapability()` in this PR. The `serverFeatureFlags` property on `api` is kept only for extension compatibility — since extensions accessed it via `api.serverFeatureFlags.value`, changing it to a plain object means `.value` will return `undefined`, which is technically a break. However, extensions should not have been relying on reactivity of this internal property. I'll double-check if any known extensions access this reactively and update the migration note if needed.

src/services/serverCapabilities.test.ts
dante01yoon: Good idea — added a test case for \`getServerCapability('extension.non_existent.deep.path')\`. It returns \`undefined\` as expected (handled by \`es-toolkit/compat\`'s \`get()\`).

src/services/serverCapabilities.ts
christian-byrne: fair point - init running before createPinia() makes this the right approach

src/services/serverCapabilities.ts
christian-byrne: sounds good, the default fallback pattern provides reasonable safety for now

src/main.ts
christian-byrne: @coderabbitai Make an Issue to do this in a followup PR on a future date

src/composables/useFeatureFlags.ts
christian-byrne: comments added, looks good

src/services/serverCapabilities.ts
christian-byrne: fixed in f43e1a2 - SyntaxError now breaks out of retry loop

src/scripts/api.ts
christian-byrne: confirmed all internal consumers migrated in this PR - the coderabbitai thread on this line has a good suggestion for expanding the JSDoc to note the ref-to-plain-object change for extension authors

src/services/serverCapabilities.test.ts
christian-byrne: verified - test added on line 122

src/scripts/api.ts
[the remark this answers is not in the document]
christian-byrne: @dante01yoon this one's worth doing - the expanded JSDoc would save extension authors some debugging time when they hit the ref-to-plain-object change. the suggested diff from coderabbit looks good as-is

src/services/serverCapabilities.test.ts
```diff
@@ -0,0 +1,131 @@
...
+  })
+
+  describe('getServerCapability', () => {
+    it('returns default value when called before init', () => {
+      expect(getServerCapability('some_key', 'fallback')).toBe('fallback')
+    })
+
+    beforeEach(async () => {
+      await initServerCapabilities()
+    })
```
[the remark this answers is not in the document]
christian-byrne: @dante01yoon heads up - the beforeEach at line 102 runs before all tests in the describe block, so the "before init" test on line 98 is actually running after init. it passes by accident since `some_key` isn't in the mock data. moving that test outside the inner describe (or into a sibling describe without the beforeEach) would fix it

src/services/serverCapabilities.ts
```diff
@@ -0,0 +1,41 @@
...
+      // Retry on network errors
+    }
+  }
+
+  console.warn('Failed to fetch server capabilities after retries')
+  capabilities = EMPTY
```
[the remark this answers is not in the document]
christian-byrne: @dante01yoon since this blocks app mount, adding an AbortController timeout per attempt would be a nice safety net - even something generous like 5s would prevent indefinite hangs on stalled networks

src/composables/useFeatureFlags.ts
```diff
@@ -36,23 +36,26 @@ function resolveFlag<T>(
...
+      return getServerCapability(ServerFeatureFlag.SUPPORTS_PREVIEW_METADATA)
     },
     get maxUploadSize() {
-      return api.getServerFeature(ServerFeatureFlag.MAX_UPLOAD_SIZE)
+      return getServerCapability(ServerFeatureFlag.MAX_UPLOAD_SIZE)
     },
     get supportsManagerV4() {
-      return api.getServerFeature(ServerFeatureFlag.MANAGER_SUPPORTS_V4)
+      return getServerCapability(ServerFeatureFlag.MANAGER_SUPPORTS_V4)
     },
```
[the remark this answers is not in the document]
christian-byrne: @dante01yoon quick one - supportsPreviewMetadata and supportsManagerV4 return undefined when missing, which makes them tri-state. passing `false` as the default would keep them clean booleans

christian-byrne (dismissed):
Low-risk mechanical migration — extracts server feature flag access into a standalone `serverCapabilities` module with eager HTTP fetch. Old API surface preserved as deprecated. Well-tested, clean diff.

christian-byrne (changes requested):
Two CI failures need fixing before this can merge:

1. **Unit tests** (`ECONNREFUSED :3000`) - `initServerCapabilities()` is making real `fetch()` calls during unit tests. The module-level `await initServerCapabilities()` in `main.ts` may be triggering when other tests import from the app entry graph without mocking `fetch`.

2. **Playwright E2E** (`featureFlags.spec.ts`) - Multiple tests reference `api.serverFeatureFlags.value` (Vue ref), but this PR changed it to a plain object, so `.value` is `undefined`. Tests also poll for WS-based `feature_flags` delivery which no longer populates the flags. These tests need updating to match the new HTTP-fetch architecture.

christian-byrne:
Are these test failures expected?

src/services/serverCapabilities.ts
```diff
@@ -0,0 +1,60 @@
...
+import { getDevOverride } from '@/utils/devFeatureFlagOverride'
+
+const EMPTY: Readonly<Record<string, unknown>> = Object.freeze({})
+const MAX_RETRIES = 2
+
+let capabilities: Readonly<Record<string, unknown>> = EMPTY
+
+function getApiBase(): string {
+  return isCloud ? '' : location.pathname.split('/').slice(0, -1).join('/')
+}
```
christian-byrne: this would now be duplicated here and in api.ts. If the issue is direction of imports, we can move it to somewhere in the `base` folder

christian-byrne (commented):
Previously we ran into issues in which the dynamic config needs to be loaded before api module is initialized (e.g., on cloud we start on the login views and api.ts is initialized in GraphView), it seems that this aspect is not going to break from this PR but can you confirm?

dante01yoon (commented):
@christian-byrne Thanks for the review.

**Re: duplicated `getApiBase()`** — Moved the shared logic into `src/platform/distribution/types.ts` (alongside `isCloud` which it depends on). Both `serverCapabilities.ts` and `api.ts` now import `getApiBase()` from there. Couldn't place it in `base/` due to the layer architecture constraint (`base/` cannot import from `platform/`).

**Re: dynamic config loading order** — Confirmed safe. `initServerCapabilities()` only depends on:
1. `isCloud` — compile-time constant (`__DISTRIBUTION__`), not dynamic config
2. `fetch` — browser built-in, no app-level setup needed

It runs at `main.ts:32` before `refreshRemoteConfig()` (line 43), but has zero dependency on remote config values. The `api.ts` constructor (which uses the same `getApiBase()`) is instantiated later in `GraphView`, well after both `initServerCapabilities()` and `refreshRemoteConfig()` have completed. So the cloud login → GraphView → api init flow remains intact.

src/scripts/api.ts
```diff
@@ -345,9 +345,12 @@ export class ComfyApi extends EventTarget {
...
 
   /**
-   * Feature flags received from the backend server.
+   * @deprecated Use `getServerCapability()` from `@/services/serverCapabilities`.
+   * Note: This field was previously a Vue `Ref`; accessing `.value` or using it as
+   * a reactive source no longer works. Migrate to `getServerCapability()` for
+   * both value access and capability observation.
    */
-  serverFeatureFlags = ref<Record<string, unknown>>({})
+  serverFeatureFlags: Record<string, unknown> = {}
```
christian-byrne, line 353: issue (blocking): serverFeatureFlags was previously ref<Record<string, unknown>>({}) and is now a plain object. This is a runtime contract break for ecosystem extensions and one in-tree consumer:

- api.serverFeatureFlags.value now returns undefined.
- watch(api.serverFeatureFlags, ...) and computed(() => api.serverFeatureFlags.value...) no longer fire.
- The WS handler at api.ts:753 reassigns the whole object on every message, dropping any test or extension override.
- src/platform/assets/utils/assetPreviewUtil.ts (api.getServerFeature('assets', false)) loses its previous reactivity guarantees because the underlying field is no longer reactive.
- This reverses the bugfix from #9051 ("fix: make serverFeatureFlags reactive via Vue ref").
- Codesearch in the custom-node ecosystem confirms downstream consumers using .value and watch() on this exact field.

The JSDoc admits .value no longer works, but no runtime warning exists, so consumers will silently observe stale or undefined values.

Suggested fix: either keep the field as a Ref with a getter shim, or expose a deprecated `get value()` accessor that returns the plain object so existing reads keep working. At minimum, log a one-shot console.warn when ecosystem code reads .value (Proxy-based, gated by a Set to avoid spam).

src/main.ts
```diff
@@ -18,13 +18,20 @@ import {
...
+  setServerCapability
+} from '@/services/serverCapabilities'
 import { useBootstrapStore } from '@/stores/bootstrapStore'
 
 import App from './App.vue'
 // Intentionally relative import to ensure the CSS is loaded in the right order (after litegraph.css)
 import './assets/css/style.css'
 import { i18n } from './i18n'
 
+await initServerCapabilities()
```
christian-byrne, line 32: issue (blocking): window.__setServerCapability is exposed unconditionally with no import.meta.env.DEV / mode guard. In production builds, any extension or XSS payload can flip security-sensitive flags:

```
window.__setServerCapability('user_secrets_enabled', true)
window.__setServerCapability('team_workspaces_enabled', true)
window.__setServerCapability('private_models_enabled', true)
window.__setServerCapability('max_upload_size', Number.MAX_SAFE_INTEGER)
```

useFeatureFlags.ts:103-107 explicitly notes team_workspaces_enabled is a race-condition guard ("This ensures we never use workspace tokens when the feature is disabled"). Forcing it true from an extension defeats that guard.

The companion devFeatureFlagOverride.ts is correctly gated by `if (!import.meta.env.DEV) return undefined`. The `?` in the Window type only marks the property optional in TypeScript. At runtime the property is set on every page load in every distribution.

Suggested fix: gate behind `if (import.meta.env.DEV || import.meta.env.MODE === 'test')`, or expose only via a Vite `define`-replaced flag (e.g. `__E2E__`) so tree-shaking removes it from prod bundles. Optionally add a build-time assertion that window.__setServerCapability === undefined in production output.

src/scripts/app.ts
```diff
@@ -782,9 +782,7 @@ export class ComfyApp {
...
```
christian-byrne, line 785: issue (blocking): the WS feature_flags listener that re-invoked load() was removed, and load() is now called once synchronously. This changes the failure semantics:

Old flow: WS feature_flags event -> updates api.serverFeatureFlags -> 'feature_flags' listener -> useNodeReplacementStore().load().

New flow: load() reads from getServerCapability(), which is populated only by the HTTP /api/features fetch in initServerCapabilities(). The WS handler at api.ts:753 still updates the deprecated api.serverFeatureFlags but never propagates into the new capabilities map.

Failure scenario: backend serves WS fine but /api/features 5xx for all 3 attempts. capabilities = EMPTY, getServerCapability(NODE_REPLACEMENTS, false) === false, store early-returns and never retries even when the WS later delivers flags. This reverses bugfix #9037.

Two parallel sources of truth (HTTP-fed serverCapabilities and WS-fed api.serverFeatureFlags) can also drift, so an extension calling deprecated api.getServerFeature() may see one value while internal consumers see another.

Suggested fix: have the WS handler in api.ts call setServerCapability() to merge into capabilities (single source of truth) and re-dispatch a 'feature_flags' event so dependents can react. Alternatively, have deprecated api.getServerFeature() shim delegate to getServerCapability().

src/main.ts
```diff
@@ -18,13 +18,20 @@ import {
...
+import {
+  initServerCapabilities,
+  setServerCapability
+} from '@/services/serverCapabilities'
 import { useBootstrapStore } from '@/stores/bootstrapStore'
 
 import App from './App.vue'
 // Intentionally relative import to ensure the CSS is loaded in the right order (after litegraph.css)
 import './assets/css/style.css'
 import { i18n } from './i18n'
```
christian-byrne, line 31: issue (blocking): top-level `await initServerCapabilities()` blocks createApp(), router, Pinia, remoteConfig and window.app for up to ~15s (3 sequential 5s fetches). Backends without /api/features, intercepting proxies, or transient 5xx eat the full retry budget on every page load.

Worse: in serverCapabilities.ts the AbortController 5s timeout only covers fetch(). After clearTimeout(timeoutId) on line 22, `await res.json()` runs with no timeout. A stalled chunked-transfer body parse hangs app boot indefinitely (blank page, no error).

Suggested fix: (a) move clearTimeout into a finally that runs after res.json(), or attach the same controller.signal to res.json() via a follow-up wrapper; (b) race the entire try block against a wall-clock budget; (c) ideally do not block boot on this fetch at all: render with defaults and resolve flags reactively, since the consumers already check defaults.

src/services/serverCapabilities.ts
```diff
@@ -0,0 +1,56 @@
...
+
+const EMPTY: Readonly<Record<string, unknown>> = Object.freeze({})
+const MAX_RETRIES = 2
+
+let capabilities: Readonly<Record<string, unknown>> = EMPTY
+
+export async function initServerCapabilities(): Promise<void> {
+  const url = `${getApiBase()}/api/features`
+
+  for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
```
christian-byrne, line 14: issue: the PR description says "retry logic (3 attempts, exponential backoff)" but there is no `await sleep(...)` between iterations. All 3 attempts fire within milliseconds, which defeats the point of retrying transient backend glitches and amplifies thundering-herd risk if many tabs reconnect at once.

Suggested fix: add `await new Promise(r => setTimeout(r, 250 * 2 ** attempt))` between iterations (skip on the last attempt), or update the PR description to match the implementation.

src/services/serverCapabilities.ts
```diff
@@ -0,0 +1,56 @@
...
+    try {
+      const controller = new AbortController()
+      const timeoutId = setTimeout(() => controller.abort(), 5000)
+      const res = await fetch(url, {
+        cache: 'no-store',
+        signal: controller.signal
+      })
+      clearTimeout(timeoutId)
```
christian-byrne, line 22: suggestion (non-blocking): when fetch() rejects (network error, abort), control jumps to catch and clearTimeout is never called. Across 3 retries that leaks 3 zombie timers. With Vitest fake timers used elsewhere, this can also leak across tests.

Move clearTimeout(timeoutId) into a finally block:

```ts
const controller = new AbortController()
const timeoutId = setTimeout(() => controller.abort(), 5000)
try {
  const res = await fetch(url, { cache: 'no-store', signal: controller.signal })
  if (res.ok) {
    capabilities = Object.freeze(await res.json())
    return
  }
} catch (error) {
  if (error instanceof SyntaxError) {
    console.warn('[serverCapabilities] Invalid JSON response, skipping retries')
    break
  }
} finally {
  clearTimeout(timeoutId)
}
```

src/composables/useFeatureFlags.ts
```diff
@@ -160,7 +168,7 @@ export function useFeatureFlags() {
   })
 
   const featureFlag = <T = unknown>(featurePath: string, defaultValue?: T) =>
-    computed(() => api.getServerFeature(featurePath, defaultValue))
+    computed(() => getServerCapability(featurePath, defaultValue))
```
christian-byrne, line 171: issue: this computed reads from a module-level `let capabilities` in serverCapabilities.ts that is not reactive. Vue memoizes the result of the computed and never re-evaluates it. Post-mount calls to window.__setServerCapability() (used by every Playwright spec changed in this PR) silently do not invalidate, and the reactive({ get xxx() {...} }) getters above have the same problem: components depending on flags.supportsManagerV4 etc. will not re-render when capabilities change.

Suggested fix: back capabilities with a `shallowRef<Record<string, unknown>>` in serverCapabilities.ts. setServerCapability() writes .value (still frozen), getServerCapability() reads .value. This restores the reactive contract previously held by the Ref-based serverFeatureFlags and makes this computed and the getter chain react to setServerCapability() in tests.

src/composables/useFeatureFlags.ts
```diff
@@ -39,23 +39,29 @@ function resolveFlag<T>(
...
   const flags = reactive({
+    // Direct server-only flags — resolved via getServerCapability() only
     get supportsPreviewMetadata() {
-      return api.getServerFeature(ServerFeatureFlag.SUPPORTS_PREVIEW_METADATA)
+      return getServerCapability(
+        ServerFeatureFlag.SUPPORTS_PREVIEW_METADATA,
+        false
+      )
     },
     get maxUploadSize() {
```
christian-byrne, line 57: question: this PR changes the default-value semantics for some flags. Old: `api.getServerFeature(SUPPORTS_PREVIEW_METADATA)` returns undefined when missing. New: `getServerCapability(SUPPORTS_PREVIEW_METADATA, false)` returns false. Same change applied to supportsManagerV4 (now defaults to false), but maxUploadSize on this line was left without a default and still returns undefined. Is this inconsistency intentional?

Consumers using `if (flags.supportsPreviewMetadata === undefined)` (e.g. "not yet known, optimistic default") will now take the `=== false` branch. If intentional, worth documenting in the PR description and grepping consumers for `=== undefined` checks; if not, restore the prior undefined defaults for parity, or pick one convention and apply it to maxUploadSize too.

src/services/serverCapabilities.ts
```diff
@@ -0,0 +1,56 @@
+import { get } from 'es-toolkit/compat'
```
christian-byrne, line 1: question: the @deprecated JSDoc on api.serverSupportsFeature / getServerFeature / getServerFeatures instructs extensions to "Use getServerCapability() from @/services/serverCapabilities". Custom nodes load as separate scripts and can only reach src/* modules through the window.comfyAPI bridge. Is `services/serverCapabilities` exported via window.comfyAPI['services/serverCapabilities']? If not, the migration target is unreachable for ecosystem extensions and the deprecation message points at a non-public path.

If it is not auto-exposed: either explicitly export it on window.comfyAPI, or rewrite the JSDoc to point at a path extensions can actually import. Migration notes in docs/FEATURE_FLAGS.md (or equivalent) would also help maintainers.

src/scripts/api.ts
```diff
@@ -1369,35 +1365,23 @@ export class ComfyApi extends EventTarget {
     return (await axios.get(this.apiURL('/i18n'))).data
   }
 
-  /**
-   * Checks if the server supports a specific feature.
-   * @param featureName The name of the feature to check (supports dot notation for nested values)
-   * @returns true if the feature is supported, false otherwise
-   */
+  /** @deprecated Use `getServerCapability()` from `@/services/serverCapabilities` */
   serverSupportsFeature(featureName: string): boolean {
```
christian-byrne, line 1369: suggestion (non-blocking): @deprecated JSDoc only surfaces in TS hints, not at runtime. Ecosystem authors loading these methods from window.comfyAPI will not see them. A one-shot console.warn (gated by a Set so it does not spam) at the top of each deprecated method, pointing at getServerCapability(), would give maintainers a chance to migrate before eventual removal.

src/main.ts
```diff
@@ -18,13 +18,20 @@ import {
...
 
+await initServerCapabilities()
+window.__setServerCapability = setServerCapability
+
 /**
  * CRITICAL: Load remote config FIRST for cloud builds to ensure
```
christian-byrne, line 36: suggestion (non-blocking): the pre-existing `// CRITICAL: Load remote config FIRST` comment warns that remoteConfig must be loaded before everything else (because every other module reads window.__CONFIG__). The new top-level `await initServerCapabilities()` runs before that block. It happens to be safe today because getApiBase() only depends on __DISTRIBUTION__ and location, but the ordering inverts the explicit comment guidance.

Either move the call to after the remoteConfig block, or update the CRITICAL comment so the new invariant ("server capabilities also load before remote config") is explicit.

src/services/serverCapabilities.ts
```diff
@@ -0,0 +1,56 @@
...
+    try {
+      const controller = new AbortController()
+      const timeoutId = setTimeout(() => controller.abort(), 5000)
+      const res = await fetch(url, {
+        cache: 'no-store',
+        signal: controller.signal
+      })
+      clearTimeout(timeoutId)
+      if (res.ok) {
+        capabilities = Object.freeze(await res.json())
```
christian-byrne, line 24: suggestion (non-blocking): Object.freeze is shallow. Nested values (e.g. extension.manager.supports_v4) live in inner objects that remain mutable. A consumer calling `getServerCapability('extension')` can mutate `.manager.supports_v4` directly and silently change future reads without going through setServerCapability.

Either deep-freeze recursively, or add a brief comment that nested objects are not frozen and must be treated as read-only.

src/platform/distribution/types.ts
```diff
@@ -16,6 +16,10 @@ const DISTRIBUTION: Distribution = __DISTRIBUTION__
 export const isDesktop = DISTRIBUTION === 'desktop'
 export const isCloud = DISTRIBUTION === 'cloud'
 
+export function getApiBase(): string {
+  return isCloud ? '' : location.pathname.split('/').slice(0, -1).join('/')
```
christian-byrne, line 20: question: on a deep-linked first load (e.g. /workflows/my-flow), this returns "/workflows", producing `GET /workflows/api/features` -> 404, which then exhausts initServerCapabilities() retries and falls back to EMPTY for ~15s of every cold load. The previous inline expression in api.ts had the same shape but only ran once at construction; this version now also runs in the new initServerCapabilities() top-level await before any router redirect. Was deep-link cold load already broken on the old code path, or does this PR newly make it block boot? If the former, a follow-up issue is probably enough; if the latter, anchoring to "/" or computing from <base href> would be safer.

christian-byrne (changes requested):
Thanks for tackling this migration. Posting one batched review covering 14 findings (4 critical, 9 major, 1 minor question) from a multi-agent review pass.

Headline concerns are the contract breaks for the custom-node ecosystem (C1, C3) and the unconditional production exposure of window.__setServerCapability (C2). Detail in inline comments.

One additional finding that does not anchor cleanly inline:

suggestion (non-blocking): src/scripts/api.featureFlags.test.ts had `describe('Reactivity')` and `describe('Dev override via localStorage')` blocks asserting contract behavior on api.getServerFeature. Since api.getServerFeature is still exported (deprecated but functional, and still on the extension surface), at least the dev-override tests should remain to prevent future regressions on the extension-visible contract. The reactivity tests can be re-targeted at the new module if reactivity is restored per the M3 comment on useFeatureFlags.ts:171.

PR #9095, opened 2026-02-22
feat: remove legacy WS feature_flags code and stale types

### LINKED ISSUE #9079: Deliver server feature flags via global object instead of WebSocket

dante01yoon:
## Summary
- Remove WS `feature_flags` case from api.ts message handler
- Replace `serverFeatureFlags` with plain stub object
- Convert deprecated methods to thin shims delegating to `getServerCapability()`
- Remove `FeatureFlagsWsMessage` type from apiSchema
- Remove stale types from browser test globals
- Simplify E2E feature flag tests to match REST-based flow

**Part 2 of 2** for #9079 — depends on #9094

## Test plan
- [x] All unit tests pass
- [x] E2E tests updated: client→server WS send, REST `/api/features` endpoint, client flags immutability
- [x] `pnpm typecheck`, `pnpm lint`, `pnpm knip` clean

Fixes #9079

### DIFF
--- a/browser_tests/tests/featureFlags.spec.ts
+++ b/browser_tests/tests/featureFlags.spec.ts
@@ -7,21 +7,16 @@ test.beforeEach(async ({ comfyPage }) => {
 })
 
 test.describe('Feature Flags', { tag: ['@slow', '@settings'] }, () => {
-  test('Client and server exchange feature flags on connection', async ({
+  test('Client sends feature flags on WebSocket connection', async ({
     comfyPage
   }) => {
-    // Navigate to a new page to capture the initial WebSocket connection
     const newPage = await comfyPage.page.context().newPage()
 
-    // Set up monitoring before navigation
     await newPage.addInitScript(() => {
-      // This runs before any page scripts
       window.__capturedMessages = {
-        clientFeatureFlags: null,
-        serverFeatureFlags: null
+        clientFeatureFlags: null
       }
 
-      // Capture outgoing client messages
       const originalSend = WebSocket.prototype.send
       WebSocket.prototype.send = function (data) {
         try {
@@ -29,40 +24,22 @@ test.describe('Feature Flags', { tag: ['@slow', '@settings'] }, () => {
           if (parsed.type === 'feature_flags') {
             window.__capturedMessages!.clientFeatureFlags = parsed
           }
-        } catch (e) {
+        } catch {
           // Not JSON, ignore
         }
         return originalSend.call(this, data)
       }
-
-      // Monitor for server feature flags
-      const checkInterval = setInterval(() => {
-        const flags = window.app?.api?.serverFeatureFlags?.value
-        if (flags && Object.keys(flags).length > 0) {
-          window.__capturedMessages!.serverFeatureFlags = flags
-          clearInterval(checkInterval)
-        }
-      }, 100)
-
-      // Clear after 10 seconds
-      setTimeout(() => clearInterval(checkInterval), 10000)
     })
 
-    // Navigate to the app
     await newPage.goto(comfyPage.url)
 
-    // Wait for both client and server feature flags
     await newPage.waitForFunction(
-      () =>
-        window.__capturedMessages!.clientFeatureFlags !== null &&
-        window.__capturedMessages!.serverFeatureFlags !== null,
+      () => window.__capturedMessages!.clientFeatureFlags !== null,
       { timeout: 10000 }
     )
 
-    // Get the captured messages
     const messages = await newPage.evaluate(() => window.__capturedMessages)
 
-    // Verify client sent feature flags
     expect(messages!.clientFeatureFlags).toBeTruthy()
     expect(messages!.clientFeatureFlags).toHaveProperty('type', 'feature_flags')
     expect(messages!.clientFeatureFlags).toHaveProperty('data')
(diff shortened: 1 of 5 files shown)


christian-byrne:
## Draft Review — Architectural Feedback

**Verdict: 🟡 Needs course correction**

### What's looking good
- Tight, well-scoped removal aligned with the REST migration in #9094 — net −400 LOC speaks for itself
- Shims for `serverSupportsFeature` / `getServerFeature` cleanly delegate to `getServerCapability()`, preserving dev override behaviour through the existing `getDevOverride` path
- Unit + E2E tests rewritten to assert the new REST contract (`/api/features`) and client-side immutability without leaning on the removed WS message

### Architectural concerns
1. **`serverFeatureFlags` shape is backward-incompatible.** The field was `Record<string, unknown>`; it's now `{ value: Record<string, unknown> }`. Extensions reading `api.serverFeatureFlags.foo` will silently break (they'd need `.value.foo`). The `.value` wrapper appears to mimic a Vue `Ref`, but no consumer in this repo reads it that way (the browser test polling that did was deleted in this PR). I'd either keep the stub as a plain `{}` (preserves the old type contract) or drop the field entirely with a clear deprecation note — the `{ value: {} }` shape is a third, novel shape that helps no one.
2. **`getServerFeatures()` now returns `{}` unconditionally.** Reasonable as a deprecation step, but worth either (a) exposing a `getAllServerCapabilities()` from `serverCapabilities` and delegating, or (b) JSDoc-noting explicitly that introspection is gone, so extension authors aren't surprised by an empty object.
3. **Stale type declarations in `src/types/index.ts`** (lines 67–76) — `CapturedMessages.serverFeatureFlags` and the `AppReadiness` interface are still declared on `Window`, mirroring what was correctly removed from `browser_tests/types/globals.d.ts`. These should go too.
4. **Stale docs.** `docs/FEATURE_FLAGS.md` and `docs/SETTINGS_SEQUENCE_DIAGRAM.md` still document the WS `feature_flags` round-trip and `serverFeatureFlags` mutation patterns. Either update them to describe the REST flow or mark the legacy doc deprecated and link to the new capabilities service.

### Suggested direction
The removal is the right move. Before marking ready: collapse `serverFeatureFlags` to either a plain `{}` or omit it entirely (pick one and document the back-compat story for extensions), purge the leftover globals in `src/types/index.ts`, and refresh the two docs. Conflict against `main` is expected since this stacks on #9094 — rebase once that lands.

---
*Early-stage review focused on direction — line-by-line feedback will come when the PR is marked ready for review.*

PR #9107, opened 2026-02-23
fix: replace as-unknown-as casts with safer patterns

Myestery:
## Summary

- Replace 83 `as unknown as` double casts with safer alternatives across 33 files
- Use `as Partial<X> as X` pattern where TypeScript allows it
- Create/reuse factory functions from `litegraphTestUtils.ts` for mock objects
- Widen `getWorkflowDataFromFile` return type to include `ComfyMetadata` directly
- Reduce total `as unknown as` count from ~153 to 71

The remaining 71 occurrences are genuinely necessary due to cross-schema casts, generic variance, missing index signatures, Float64Array-to-tuple conversions, and DOM type incompatibilities.

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] All affected unit tests pass

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9107-fix-replace-as-unknown-as-casts-with-safer-patterns-3106d73d3650815cb5bcd613ad635bd7) by [Unito](https://www.unito.io)

### DIFF
--- a/apps/desktop-ui/src/views/MaintenanceView.stories.ts
+++ b/apps/desktop-ui/src/views/MaintenanceView.stories.ts
@@ -77,7 +77,7 @@ const createMockElectronAPI = () => {
 }
 
 const ensureElectronAPI = () => {
-  const globalWindow = window as unknown as { electronAPI?: unknown }
+  const globalWindow = window as { electronAPI?: unknown }
   if (!globalWindow.electronAPI) {
     globalWindow.electronAPI = createMockElectronAPI()
   }
--- a/browser_tests/fixtures/helpers/SubgraphHelper.ts
+++ b/browser_tests/fixtures/helpers/SubgraphHelper.ts
@@ -88,7 +88,7 @@ export class SubgraphHelper {
 
             if (node.onPointerDown) {
               node.onPointerDown(
-                event as unknown as CanvasPointerEvent,
+                event as Partial<CanvasPointerEvent> as CanvasPointerEvent,
                 app.canvas.pointer,
                 app.canvas.linkConnector
               )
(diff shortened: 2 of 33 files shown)


christian-byrne (commented):
Could you explain why this pattern is safer? It's because the initial `as Partial` checks that the object is at least structurally typed as the argument to Partial? Which would affect any overlapping named properties and base type? But then when we assert, does asserting from a structural match to the end type differ from asserting from the top type to the end type in any way as it relates to downstream type inference?

Myestery:
> Could you explain why this pattern is safer? It's because the initial `as Partial` checks that the object is at least structurally typed as the argument to Partial? Which would affect any overlapping named properties and base type? But then when we assert, does asserting from a structural match to the end type differ from asserting from the top type to the end type in any way as it relates to downstream type inference?

 as Partial<X> as X does one thing that as unknown as X doesn't: the first cast (as Partial<X>) is checked by TypeScript. It    
  will fail if your object has a property that exists on X but with an incompatible type. So it catches a specific class of bugs 
 ie you can't accidentally have { name: 42 } cast to a type where name is string.         

It doesn't catch missing properties, and it doesn't change anything about how the resulting type is used  downstream

christian-byrne (approved):
LGTM, seems you also removed some unnecessary type assertions which is also good.

christian-byrne:
> > Could you explain why this pattern is safer? It's because the initial `as Partial` checks that the object is at least structurally typed as the argument to Partial? Which would affect any overlapping named properties and base type? But then when we assert, does asserting from a structural match to the end type differ from asserting from the top type to the end type in any way as it relates to downstream type inference?
> 
> as Partial as X does one thing that as unknown as X doesn't: the first cast (as Partial) is checked by TypeScript. It will fail if your object has a property that exists on X but with an incompatible type. So it catches a specific class of bugs ie you can't accidentally have { name: 42 } cast to a type where name is string.
> 
> It doesn't catch missing properties, and it doesn't change anything about how the resulting type is used downstream

Thank you for that info

PR #9121, opened 2026-02-23
fix: resolve subgraph node slot link misalignment during workflow load

christian-byrne:
## Summary

Fix subgraph node slot connector links appearing misaligned after workflow load, caused by a transform desync between LiteGraph's internal canvas transform and the Vue TransformPane's CSS transform.

## Changes

- **What**: Changed `syncNodeSlotLayoutsFromDOM` to use DOM-relative measurement (slot position relative to its parent `[data-node-id]` element) instead of absolute canvas-space conversion via `clientPosToCanvasPos`. This makes the slot offset calculation independent of the global canvas transform, eliminating the frame-lag desync that occurred when `fitView()` updated `lgCanvas.ds` before the Vue CSS transform caught up.
- **Cleanup**: Removed the unreachable fallback path that still used `clientPosToCanvasPos` when the parent node element wasn't found (every slot element is necessarily a child of a `[data-node-id]` element — if `closest()` fails the element is detached and measuring is meaningless). This also removed the `conv` parameter from `syncNodeSlotLayoutsFromDOM` and `flushScheduledSlotLayoutSync`, and the `useSharedCanvasPositionConversion` import.
- **Test**: Added a Playwright browser test that loads a subgraph workflow with `workflowRendererVersion: "LG"` (triggering the 1.2x scale in `ensureCorrectLayoutScale`) as a template (triggering `fitView`), and verifies all slot connector positions are within bounds of their parent node element.

## Review Focus

- The core change is in `useSlotElementTracking.ts` — the new measurement approach uses `getBoundingClientRect()` on both the slot and its parent node element, then divides by `currentScale` to get canvas-space offsets. This is simpler and more robust than the previous approach.
- SubgraphNodes were disproportionately affected because they are relatively static and don't often trigger `ResizeObserver`-based re-syncs that would eventually correct stale offsets.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9121-fix-resolve-subgraph-node-slot-link-misalignment-during-workflow-load-3106d73d365081eca413c84f2e0571d6) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -48,27 +47,41 @@ export function flushScheduledSlotLayoutSync() {
...
+  // TransformPane's CSS transform during workflow loading (e.g., after
+  // fitView or ensureCorrectLayoutScale). Both the slot and node elements
+  // share the same DOM transform, so their pixel difference divided by the
+  // effective scale yields a correct canvas-space offset regardless of
+  // whether the TransformPane has flushed its latest transform to the DOM.
+  const nodeEl = node.slots
+    .values()
+    .next()
+    .value?.el.closest('[data-node-id]') as HTMLElement | null
+  const nodeRect = nodeEl?.getBoundingClientRect()
```
christian-byrne, line 136: This defeats the entire purpose of having a cached rect, and introduces possibility of thrashing. Why not just try to fix the stale cache the correct way? Is it not possible?

pythongosssss:
@DrJKL Passing to you as I think this may be better fixed with the layout changes you've been working on

src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
DrJKL: The `cachedOffset` is still computed and reused — `updateNodeSlotsFromCache` (the hot path for node moves) never touches the DOM. `syncNodeSlotLayoutsFromDOM` is the cold path (mount, resize) where full DOM measurement is already expected.

The one extra `getBoundingClientRect()` per node is negligible: the function already calls it N times (once per slot via `getSlotElementRect` at [L142](https://github.com/Comfy-Org/ComfyUI_frontend/blob/fix/subgraph-slot-link-misalignment/src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts#L142)). Adding 1 per node outside the loop doesn't change the cost profile.

**Why `clientPosToCanvasPos` can't be fixed here**: it reads `lgCanvas.ds` directly ([useCanvasPositionConversion.ts#L22](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/composables/element/useCanvasPositionConversion.ts#L22)), which `fitView` updates synchronously. But `getBoundingClientRect()` returns positions based on the TransformPane CSS transform, which lags by a frame (synced via `useRafFn` in [TransformPane.vue#L58-66](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/renderer/core/layout/transform/TransformPane.vue#L58-L66)). New scale/offset ÷ old screen positions = wrong result. The DOM-relative approach cancels this out because both slot and node elements share the same (stale) CSS transform.

**Future work**: A camera-synced conversion function (using `useTransformState().camera` instead of `lgCanvas.ds`) would let us drop the node-element lookup entirely. `camera` is updated via the same RAF as the CSS transform, so it always matches the DOM. This would require exposing a `clientPosToCanvasPos` variant that reads `camera` state — worth doing as part of a broader cleanup of the coordinate conversion utilities, but out of scope for this bugfix.

DrJKL (approved):

PR #9122, opened 2026-02-23
fix: open image in new tab on cloud fetches as blob to avoid GCS auto-download

christian-byrne:
## Summary

Fix "Open Image" on cloud opening a new tab that auto-downloads the asset instead of displaying it inline.

## Changes

- **What**: Add `openFileInNewTab()` to `downloadUtil.ts` that fetches cross-origin URLs as blobs before opening in a new tab, avoiding GCS `Content-Disposition: attachment` redirects. Opens the blank tab synchronously to preserve user-gesture activation (avoiding popup blockers), then navigates to a blob URL once the fetch completes. Blob URLs are revoked after 60s or immediately if the tab was closed. Update both call sites (`useImageMenuOptions` and `litegraphService`) to use the new function.

## Review Focus

- The synchronous `window.open('', '_blank')` before the async fetch is intentional to preserve user-gesture context and avoid popup blockers.
- Blob URL revocation strategy: 60s timeout for successful opens, immediate revoke if tab was closed, tab closed on fetch failure.
- Shared `fetchAsBlob()` helper is also used by the existing `downloadViaBlobFetch`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9122-fix-open-image-in-new-tab-on-cloud-fetches-as-blob-to-avoid-GCS-auto-download-3106d73d365081a3bfa6eb7d77fde99f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/base/common/downloadUtil.test.ts
+++ b/src/base/common/downloadUtil.test.ts
@@ -2,7 +2,8 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 import {
   downloadFile,
-  extractFilenameFromContentDisposition
+  extractFilenameFromContentDisposition,
+  openFileInNewTab
 } from '@/base/common/downloadUtil'
 
 let mockIsCloud = false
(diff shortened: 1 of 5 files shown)


src/base/common/downloadUtil.test.ts
christian-byrne, line 1: Instead of opening in a new tab and having to deal with all the weird constraints of that, lets just use our lightbox (Galleria) that is used when clicking "inspect" in right click context menu of the asset cards in the assets panel -- and then re-use (or first refactor/extract then re-use) the same code and utils.

christian-byrne: @coderabbitai make an Issue to do this improvement in a followup PR.

AustinMroz (dismissed):

PR #9123, opened 2026-02-23
fix: preserve input asset previews across execution updates

christian-byrne:
## Summary

Fix input asset previews (images/videos) disappearing from LoadImage/LoadVideo nodes after execution and a browser tab switch.

## Changes

- **What**: Guard `setOutputsByLocatorId` in `imagePreviewStore` to preserve existing input-type preview images (`type: 'input'`) when the incoming execution output has no images. Execution outputs with actual images still overwrite as expected.

## Review Focus

- The guard only applies when existing output is an input preview (`type === 'input'` for all images) AND incoming output has no images — this is the exact scenario where execution clobbers upload widget previews.
- Root cause: execution results from the backend overwrite the upload widget's synthetic preview for LoadImage/LoadVideo nodes (which produce no output images). Combined with the deferred resize-observer re-observation from PR #8805, returning to a hidden tab reads the now-empty store entry.

### DIFF
--- a/src/stores/nodeOutputStore.test.ts
+++ b/src/stores/nodeOutputStore.test.ts
@@ -229,6 +229,30 @@ describe('nodeOutputStore input preview preservation', () => {
 
     expect(store.nodeOutputs[executionId]?.images).toBeUndefined()
   })
+
+  it('should pass through non-image fields while preserving input preview images', () => {
+    const store = useNodeOutputStore()
+    const executionId = '5'
+
+    const inputPreview = createMockOutputs([
+      { filename: 'example.png', subfolder: '', type: 'input' }
+    ])
+    store.setNodeOutputsByExecutionId(executionId, inputPreview)
+
+    const videoOutput: ExecutedWsMessage['output'] = {
+      video: [{ filename: 'output.mp4', subfolder: '', type: 'output' }]
+    }
+    store.setNodeOutputsByExecutionId(executionId, videoOutput)
+
+    expect(store.nodeOutputs[executionId]?.images).toHaveLength(1)
+    expect(store.nodeOutputs[executionId]?.images?.[0].filename).toBe(
+      'example.png'
+    )
+    expect(store.nodeOutputs[executionId]?.video).toHaveLength(1)
+    expect(store.nodeOutputs[executionId]?.video?.[0].filename).toBe(
+      'output.mp4'
+    )
+  })
 })
 
 describe('nodeOutputStore getPreviewParam', () => {
--- a/src/stores/nodeOutputStore.test.ts
+++ b/src/stores/nodeOutputStore.test.ts
@@ -151,6 +151,86 @@ describe('nodeOutputStore restoreOutputs', () => {
   })
 })
 
+describe('nodeOutputStore input preview preservation', () => {
+  beforeEach(() => {
+    setActivePinia(createTestingPinia({ stubActions: false }))
+    vi.clearAllMocks()
+    app.nodeOutputs = {}
+    app.nodePreviewImages = {}
+  })
+
+  it('should preserve input preview when execution sends empty output', () => {
+    const store = useNodeOutputStore()
+    const executionId = '3'
+
+    const inputPreview = createMockOutputs([
+      { filename: 'example.png', subfolder: '', type: 'input' }
+    ])
+    store.setNodeOutputsByExecutionId(executionId, inputPreview)
+
+    expect(store.nodeOutputs[executionId]?.images).toHaveLength(1)
+
+    const emptyExecutionOutput = createMockOutputs()
+    store.setNodeOutputsByExecutionId(executionId, emptyExecutionOutput)
+
+    expect(store.nodeOutputs[executionId]?.images).toHaveLength(1)
+    expect(store.nodeOutputs[executionId]?.images?.[0].filename).toBe(
+      'example.png'
+    )
+  })
+
+  it('should preserve input preview when execution sends output with empty images array', () => {
+    const store = useNodeOutputStore()
+    const executionId = '3'
+
+    const inputPreview = createMockOutputs([
+      { filename: 'example.png', subfolder: '', type: 'input' }
+    ])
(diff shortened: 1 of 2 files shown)


src/stores/nodeOutputStore.ts
```diff
@@ -139,6 +155,20 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
+    // Preserve input preview images (from upload widgets) when execution
+    // sends outputs with no images. Without this guard, execution results
+    // overwrite the upload widget's preview, causing LoadImage/LoadVideo
+    // nodes to lose their preview after execution + tab switch.
+    const incomingImages = (outputs as ExecutedWsMessage['output'])?.images
+    const hasIncomingImages =
+      Array.isArray(incomingImages) && incomingImages.length > 0
+    if (
+      !hasIncomingImages &&
+      isInputPreviewOutput(app.nodeOutputs[nodeLocatorId])
```
christian-byrne, line 174: issue: the early `return` drops all incoming output fields, not just images. If the incoming output contains other data (audio, video, text, animated flags, or passthrough fields), those updates will be silently ignored when an input preview exists. Consider merging the existing images into the incoming output instead of returning early, e.g.:
```ts
if (!hasIncomingImages && isInputPreviewOutput(app.nodeOutputs[nodeLocatorId])) {
  outputs = {
    ...outputs,
    images: (app.nodeOutputs[nodeLocatorId] as ExecutedWsMessage['output']).images
  }
}
```

src/stores/nodeOutputStore.ts
```diff
@@ -139,6 +155,20 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
     // (e.g., two LoadImage nodes selecting the same image)
     if (outputs == null) return
 
+    // Preserve input preview images (from upload widgets) when execution
+    // sends outputs with no images. Without this guard, execution results
+    // overwrite the upload widget's preview, causing LoadImage/LoadVideo
```
christian-byrne, line 164: question: does this guard handle intentional "clear preview" scenarios? If a user removes an uploaded asset and the widget sends an output with `images: []` or no images to clear the preview, this guard would incorrectly preserve the old preview. Is there a separate mechanism for clearing previews, or should explicit clears be handled differently?

src/stores/nodeOutputStore.test.ts
```diff
@@ -151,6 +151,86 @@ describe('imagePreviewStore restoreOutputs', () => {
...
+
+    const tempOutput = createMockOutputs([
+      { filename: 'temp.png', subfolder: '', type: 'temp' }
+    ])
+    store.setNodeOutputsByExecutionId(executionId, tempOutput)
+
+    const emptyOutput = createMockOutputs()
+    store.setNodeOutputsByExecutionId(executionId, emptyOutput)
+
+    expect(store.nodeOutputs[executionId]?.images).toBeUndefined()
```
christian-byrne, line 230: suggestion (non-blocking): consider adding a test for video outputs. The `zOutputs` schema shows `video` is a separate array field - if LoadVideo stores previews under `video` rather than `images`, this fix may not cover that case.

src/stores/nodeOutputStore.ts
christian-byrne: fixed in 719a44d - changed from early return to merge, preserving input preview images while passing through all other fields (audio, video, text, etc.)

src/stores/nodeOutputStore.ts
christian-byrne: intentional clears go through setNodeOutputs (the widget path in useImageUploadWidget), not setNodeOutputsByExecutionId (the execution path), so this guard doesn't interfere. added a clarifying comment in 719a44d.

src/stores/nodeOutputStore.test.ts
christian-byrne: good call - added a test for video output passthrough in 719a44d. the merge approach now naturally handles this since video fields pass through while input preview images are preserved.

src/stores/nodeOutputStore.ts
```diff
@@ -143,6 +159,27 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
+    // nodes to lose their preview after execution + tab switch.
+    // Note: intentional preview clears go through setNodeOutputs (widget
+    // path), not setNodeOutputsByExecutionId, so this guard does not
+    // interfere with user-initiated clears.
+    const incomingImages = (outputs as ExecutedWsMessage['output'])?.images
```
AustinMroz: Nit: outputs is guarded as non null, but still uses optional chaining(`?.`)

AustinMroz (dismissed):

src/stores/nodeOutputStore.ts
```diff
@@ -143,6 +159,27 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
+    const hasIncomingImages =
+      Array.isArray(incomingImages) && incomingImages.length > 0
+    if (
+      !hasIncomingImages &&
+      isInputPreviewOutput(app.nodeOutputs[nodeLocatorId])
+    ) {
+      outputs = {
+        ...outputs,
+        images: (app.nodeOutputs[nodeLocatorId] as ExecutedWsMessage['output'])
```
DrJKL: This cast is happening a lot, do we have an insufficient interface somewhere?

src/stores/nodeOutputStore.ts
christian-byrne: fixed in 83f7f79 - removed the unnecessary optional chaining

src/stores/nodeOutputStore.ts
christian-byrne: good catch - the cast was actually unnecessary here. `app.nodeOutputs` is already typed as `Record<string, NodeExecutionOutput>` and `NodeExecutionOutput` is the same type as `ExecutedWsMessage['\''output'\'']` (both are `z.infer<typeof zOutputs>` from apiSchema.ts). removed the cast entirely in 5a42abd

PR #9124, opened 2026-02-23
fix: promoted textarea widgets in subgraphs no longer permanently read-only

christian-byrne:
## Summary

Promoted textarea widgets inside subgraphs were permanently read-only because `disabled: true` was unconditionally injected when the widget's input slot was linked.

## Changes

- **What**: In `NodeWidgets.vue`, extract `isPromotedOnOwningNode` check and skip the `disabled: true` override for promoted widgets whose slot is internally linked to a SubgraphInput node. Reuse the same variable for the existing `borderStyle` condition (was duplicated). Add 4 unit tests for `WidgetTextarea` read-only behavior. Remove dead `readonly` prop parameter from test helper.

## Review Focus

The `isPromotedOnOwningNode` guard uses a `nodeId` comparison (`String(widgetState?.nodeId) === String(nodeId)`) to distinguish promoted widgets on their owning node (should be interactive) from proxy widgets on the SubgraphNode (should remain disabled). This fix benefits all promoted widget types (combos, numbers, etc.), not just textareas.

Fixes #8818

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9124-fix-promoted-textarea-widgets-in-subgraphs-no-longer-permanently-read-only-3106d73d3650816b8c08e417c7d10464) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/NodeWidgets.vue
+++ b/src/renderer/extensions/vueNodes/components/NodeWidgets.vue
@@ -192,19 +192,23 @@ const processedWidgets = computed((): ProcessedWidget[] => {
     // Get value from store (falls back to undefined if not registered)
     const value = widgetState?.value as WidgetValue
 
-    // Build options from store state, with slot-linked override for disabled
+    // Build options from store state, with slot-linked override for disabled.
+    // Promoted widgets inside a subgraph are always linked to SubgraphInput,
+    // but should remain interactive — skip the disabled override for them.
     const storeOptions = widgetState?.options ?? {}
-    const widgetOptions = slotMetadata?.linked
-      ? { ...storeOptions, disabled: true }
-      : storeOptions
+    const isPromotedOnOwningNode =
+      widgetState?.promoted && String(widgetState?.nodeId) === String(nodeId)
+    const widgetOptions =
+      slotMetadata?.linked && !isPromotedOnOwningNode
+        ? { ...storeOptions, disabled: true }
+        : storeOptions
 
     // Derive border style from store metadata
-    const borderStyle =
-      widgetState?.promoted && String(widgetState?.nodeId) === String(nodeId)
-        ? 'ring ring-component-node-widget-promoted'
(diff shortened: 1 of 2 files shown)


christian-byrne (commented):
I believe this is a fundamental misunderstanding of the bug and the desired UX

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -192,19 +192,23 @@ const processedWidgets = computed((): ProcessedWidget[] => {
...
-    // Build options from store state, with slot-linked override for disabled
+    // Build options from store state, with slot-linked override for disabled.
+    // Promoted widgets inside a subgraph are always linked to SubgraphInput,
+    // but should remain interactive — skip the disabled override for them.
     const storeOptions = widgetState?.options ?? {}
-    const widgetOptions = slotMetadata?.linked
-      ? { ...storeOptions, disabled: true }
-      : storeOptions
+    const isPromotedOnOwningNode =
+      widgetState?.promoted && String(widgetState?.nodeId) === String(nodeId)
```
dante01yoon, line 200: I might be wrong, but IMHO,

`nodeId` (from `nodeData.id`) includes the graph prefix in subgraphs (e.g. `"3:5"`), while `widgetState?.nodeId` from the store is typically the bare ID (e.g. `"5"`). This means the check would always be `false` inside subgraphs, so promoted widgets could still end up disabled.

Since `bareWidgetId` on line 189 already strips the prefix, would something like this work?

```ts
const isPromotedOnOwningNode =
  widgetState?.promoted && String(widgetState?.nodeId) === bareWidgetId
```
That way both sides are compared as bare IDs regardless of graph depth.

christian-byrne:
Closing — need to re-evaluate against the proxy-widget-v2 refactor (#8856) that just landed, and re-examine the actual bug (reactivity on disconnect, not permanent read-only).

PR #9134, opened 2026-02-23
fix: align run controls with queue modal design

benceruleanlu:
## Summary
- move queue batch controls to the left of the run button
- align run control styling to the Figma queue modal spec using PrimeVue PT/Tailwind (secondary background on batch + dropdown, primary run button)
- normalize control heights to match actionbar buttons and tighten dropdown hit area
- update run typography/spacing and replace all three chevrons (dropdown + batch up/down) with the requested SVG

Design: https://www.figma.com/design/LVilZgHGk5RwWOkVN6yCEK/Queue-Progress-Modal?node-id=3845-23904&m=dev

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9134-fix-align-run-controls-with-queue-modal-design-3106d73d36508160afcedbcfe4b98291) by [Unito](https://www.unito.io)


src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
```diff
@@ -1,21 +1,41 @@
...
+        >
+          <path
+            d="M0.650391 0.649902L3.65039 3.6499L6.65039 0.649902"
+            stroke="#8A8A8A"
+            stroke-width="1.3"
+            stroke-linecap="round"
+            stroke-linejoin="round"
+          />
+        </svg>
+      </template>
```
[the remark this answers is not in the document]
benceruleanlu: Done

src/components/actionbar/BatchCountEdit.vue
```diff
@@ -9,26 +9,78 @@
...
+        >
+          <path
+            d="M0.650391 0.649902L3.65039 3.6499L6.65039 0.649902"
+            stroke="#8A8A8A"
+            stroke-width="1.3"
+            stroke-linecap="round"
+            stroke-linejoin="round"
+          />
+        </svg>
+      </template>
```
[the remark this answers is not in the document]
benceruleanlu: Done

src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
```diff
@@ -136,6 +141,41 @@ const queueButtonSeverity = computed(() =>
...
+        padding: '0.5rem 1rem'
+      }
+    },
+    icon: {
+      class: 'text-white'
+    },
+    label: {
+      class: 'text-white font-light'
+    }
+  },
```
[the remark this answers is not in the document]
benceruleanlu: image

src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
```diff
@@ -136,6 +141,41 @@ const queueButtonSeverity = computed(() =>
   isStopInstantAction.value ? 'danger' : 'primary'
 )
 
+const splitButtonPt = computed(() => ({
```
christian-byrne: The complexity of the passthrough here signals to me that we might want to use our own component / Reka UI component. The primevue pt is also not going to be reactive. WDYT?

DrJKL: https://www.shadcn-vue.com/docs/components/button-group

benceruleanlu: Yeah I'll just refactor to Alex Brown's suggested component or another

benceruleanlu: Made the change, it's a large refactor

benceruleanlu:
Going to hold off on regenerating since this touches like almost every snapshot until code approval

src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
```diff
@@ -1,48 +1,84 @@
...
       data-testid="queue-button"
+      :data-variant="queueButtonVariant"
       @click="queuePrompt"
     >
-      <template #icon>
-        <i :class="iconClass" />
-      </template>
-      <template #item="{ item }">
+      <i :class="[iconClass, 'size-4']" />
+      {{ queueButtonLabel }}
```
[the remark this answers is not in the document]
benceruleanlu: Addressed in 173257af5. Updated the icon binding to use `cn(iconClass, 'size-4')` in `ComfyQueueButton.vue`.

src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
```diff
@@ -1,48 +1,84 @@
...
+              v-tooltip="{
+                value: item.tooltip,
+                showDelay: 600
+              }"
+              :variant="
+                item.key === selectedQueueMode ? 'primary' : 'secondary'
+              "
+              size="sm"
+              :aria-label="item.tooltip"
+              :class="queueMenuItemButtonClass"
```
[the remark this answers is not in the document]
benceruleanlu: Addressed in 173257af5. Removed `:aria-label="item.tooltip"` from the menu button so visible `item.label` remains the accessible name.

src/components/ui/button-group/ButtonGroup.vue
```diff
@@ -0,0 +1,24 @@
...
+
+<template>
+  <Primitive
+    :as
+    :class="
+      cn('inline-flex items-stretch overflow-hidden rounded-md', className)
+    "
+  >
+    <slot />
+  </Primitive>
```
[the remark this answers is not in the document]
benceruleanlu: Addressed in 173257af5. `ButtonGroup.vue` now captures `...restProps`, forwards them via `useForwardProps(restProps)`, and binds them onto `Primitive`.

src/components/actionbar/TinyChevronIcon.vue
```diff
@@ -0,0 +1,27 @@
...
+    viewBox="0 0 8 5"
+    fill="none"
+    aria-hidden="true"
+  >
+    <path
+      d="M0.650391 0.649902L3.65039 3.6499L6.65039 0.649902"
+      class="text-muted-foreground"
+      stroke="currentColor"
+      stroke-width="1.3"
+      stroke-linecap="round"
```
christian-byrne, line 16: question: `text-muted-foreground` on the `<path>` element locks the chevron color regardless of parent context. For example, the step buttons in `BatchCountEdit` set `text-base-foreground` but that has no effect on the chevron because the path overrides via its own class. If the muted color is per the Figma spec for all usages, this is fine - but it means hover states or variant colors on parent buttons won't flow through to the icon. Was this intentional, or should the class live on the SVG root (or be omitted entirely) so `currentColor` inherits from the parent?

src/components/actionbar/TinyChevronIcon.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <svg
+    :class="rotateUp && 'rotate-180'"
+    style="width: 8px; height: 5px; min-width: 8px; min-height: 5px"
```
christian-byrne: nitpick (non-blocking): the SVG uses inline `style` for dimensions but Tailwind for color. For consistency you could use Tailwind arbitrary values instead: `class="h-[5px] w-[8px] min-h-[5px] min-w-[8px]"`.

benceruleanlu: I just fixed it in this PR. The issue was that the button component itself was overriding sizing, I just added not height not width so any SVG with their own declarations will be prioritized

[&_svg:not([width]):not([height])]:size-4

src/components/actionbar/TinyChevronIcon.vue
benceruleanlu: It wasn't really intentional, just moved that to the implementations and off of the base component

src/components/actionbar/ComfyRunButton/ComfyQueueButton.vue
christian-byrne: verified - PrimeVue ButtonGroup replaced with custom Reka UI Primitive-based component using useForwardProps, and ComfyQueueButton now uses Reka UI DropdownMenu. clean refactor

src/components/actionbar/TinyChevronIcon.vue
christian-byrne: nice - currentColor is the right call, verified the path now uses stroke=currentColor so parent context flows through

src/components/actionbar/TinyChevronIcon.vue
christian-byrne: fixed in this PR already - SVG now uses Tailwind classes (h-[5px] w-[8px] etc). issue #9236 still tracks any remaining cleanup

christian-byrne (dismissed):
LGTM but will need to remove then regenerate snapshots.

Issue #8136, opened 2026-01-17
Retention of frontend state upon reloading the GUI mid-generation


c0dezer019:
### Is there an existing issue for this?

- [x] I have searched the existing issues and checked the recent builds/commits

### What problem are you experiencing?

Reloading the GUI in ComfyUI_frontend during an active generation causes the user interface to lose sync with the current job and node progress. TAE (Task/Activity/Event) preview no longer updates, and the UI cannot properly restore precise ongoing task/progress state until completion or restart. Restarting a long-running job can significantly impact work, especially if waiting until job completion results in a bad outcome that could have been stopped before it finished. And due to loss of TAE preview, I'd have no way of knowing if it was going right or going wrong, making cancelling a gamble.

Context: I'm on an older computer that requires running in CPU mode, since GPU does not have support for pytorch. Therefor, gens are usually significantly longer than average, so it's not merely just a matter of going "Well, I'll just wait the 4 minutes and see what I get."

### When does this problem occur?

- This occurs whenever the web UI is refreshed while a job is actively generating.
- The UI immediately shows active tasks in the queue but does not regain proper tracking of ongoing node execution or resume live TAE updates.
- Impacts monitoring or debugging of long-running jobs, especially in cases of accidental page reloads, or a reload becomes necessary mid-generation.

### How often do you encounter this problem?

Multiple times per day

### How much does this problem affect your workflow?

Significantly slows down my work

### Current workarounds

Monitor progress through the task queue, but have to wait for ongoing jobs to finish before status tracking is correct again. No way to restore live node progress/TAE preview mid-job, which may mean loss of context or missing issues.

### Ideas for solutions (Optional)

Perhaps the UI could fetch the current node/task/TAE state from the backend and provide full state restoration after a reload, minimizing operational blind spots.

### Additional context

- Reproducible in multiple browsers.
- Users relying on live feedback/progress through TAE lose visibility after reload.
- No browser console errors are shown for this condition.
- Example: Start a generation, reload page, progress/TAE lost until job ends.
- May impact collaborative or batch automation workflows if UI state is lost during processing.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-8136-Retention-of-frontend-state-upon-reloading-the-GUI-mid-generation-2eb6d73d3650815c8812e322aca32ec6) by [Unito](https://www.unito.io)

PR #9136, opened 2026-02-23
feat: periodically re-poll queue progress state

Myestery:
## Summary
- Add `useQueuePolling` composable that polls `queueStore.update()` every 5s while jobs are active
- Calls `update()` immediately on creation so the UI is current after page reload
- Uses `useIntervalFn` + `watch` pattern (same as `assetDownloadStore`) to pause/resume based on `activeJobsCount`

## Related Issue
- Related to #8136

## QA
- Queue a prompt, reload page mid-execution, verify queue UI updates every ~5s
- Verify polling stops when queue empties
- Verify polling resumes when new jobs are queued

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9136-feat-periodically-re-poll-queue-progress-state-3106d73d36508119a32fc5b9c8eda21c) by [Unito](https://www.unito.io)


src/composables/queue/useQueuePolling.ts
```diff
@@ -0,0 +1,27 @@
...
+    { immediate: false }
+  )
+
+  watch(
+    () => queueStore.activeJobsCount > 0,
+    (hasActiveJobs) => {
+      if (hasActiveJobs) resume()
+      else pause()
+    },
+    { immediate: true }
```
christian-byrne: Don't think it's needed since `update` is already triggered on connection of the websocket which handles the "sync on reconnect" necessity.

src/composables/queue/useQueuePolling.ts
```diff
@@ -0,0 +1,27 @@
+import { useIntervalFn } from '@vueuse/core'
```
christian-byrne: Would also like to see logic like:

- when `update` is invoked (i.e., /jobs are requested), reset the timer back to 5 seconds.


Additionally, I think we can pause polling if more than 1-2 jobs is active because the main issue we are fixing is not getting the _last_ status or execution_success message (e.g., because it was dropped or the connection dropped momentarily). If there is more than 1-2 jobs, we can count on the server eventually sending a new message and don't need to poll. Put simply: we are only trying to solve/guarantee eventual consistency, not overall synchronization.

christian-byrne (commented):
Can we move the files to src/platform/remote/comfyui if it's specific to the ComfyUI server? Or src/base/remote/queue/utils (example) if generalizable to any remote server's queue?

src/composables/queue/useQueuePolling.ts
```diff
@@ -0,0 +1,27 @@
...
+import { watch } from 'vue'
+
+import { useQueueStore } from '@/stores/queueStore'
+
+const POLL_INTERVAL_MS = 5_000
```
christian-byrne: Would prefer something slightly higher like 8 sec.

src/composables/queue/useQueuePolling.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function useQueuePolling() {
+  const queueStore = useQueueStore()
+
+  void queueStore.update()
+
+  const { pause, resume } = useIntervalFn(
```
christian-byrne: nit: we can consider soft exponential backoff (in some capacity) that resets when the job count changes or jobs transition state (i.e., queue state has actually changed)

src/platform/remote/comfyui/useQueuePolling.ts
```diff
@@ -0,0 +1,45 @@
...
+const BACKOFF_MULTIPLIER = 1.5
+
+export function useQueuePolling() {
+  const queueStore = useQueueStore()
+  const delay = ref(BASE_INTERVAL_MS)
+
+  const { start, stop } = useTimeoutFn(
+    () => {
+      delay.value = Math.min(delay.value * BACKOFF_MULTIPLIER, MAX_INTERVAL_MS)
+      void queueStore.update()
```
christian-byrne, line 18: The timeout can fire while an external update is already in-flight (`isLoading` is true — e.g. from a WebSocket event), which means we'd kick off a redundant concurrent `update()` call. The store's `requestId` dedupe prevents stale state writes, but we'd still be making an unnecessary network round-trip.

I think both the callback and `scheduleNextPoll` should check `isLoading` before doing work:

```suggestion
      if (queueStore.activeJobsCount !== 1 || queueStore.isLoading) return
      delay.value = Math.min(delay.value * BACKOFF_MULTIPLIER, MAX_INTERVAL_MS)
      void queueStore.update()
```

And in `scheduleNextPoll`:
```ts
function scheduleNextPoll() {
  stop()
  if (queueStore.activeJobsCount === 1 && !queueStore.isLoading) start()
}
```

That way we never schedule or fire a poll while an update is already happening — the `isLoading` completion watcher will pick it back up once the in-flight update finishes.

src/platform/remote/comfyui/useQueuePolling.ts
```diff
@@ -0,0 +1,45 @@
...
+  )
+
+  // Reschedule after any update completes (whether from polling or
+  // WebSocket events) to avoid redundant requests.
+  watch(
+    () => queueStore.isLoading,
+    (loading, wasLoading) => {
+      if (wasLoading && !loading) scheduleNextPoll()
+    }
+  )
```
christian-byrne, line 46: There's a subtle edge case here — this watcher uses the default `flush: 'pre'`, which means Vue batches it. If `isLoading` flips `false → true → false` within the same synchronous execution (e.g. a near-instant failure, or in tests), the watcher callback could run once with `(false, false)` and never trigger the reschedule.

Adding `{ flush: 'sync' }` ensures the watcher fires on every individual transition so we never miss the completion:

```suggestion
    (loading, wasLoading) => {
      if (wasLoading && !loading) scheduleNextPoll()
    },
    { flush: 'sync' }
```

The callback is tiny (just calling `scheduleNextPoll`) so there's no re-entrancy concern.

src/platform/remote/comfyui/useQueuePolling.test.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  it('stops polling when activeJobsCount drops to 0', async () => {
+    store.activeJobsCount = 1
+    wrapper = mountUseQueuePolling()
+
+    store.activeJobsCount = 0
+    await vi.advanceTimersByTimeAsync(16_000)
+
+    expect(store.update).not.toHaveBeenCalled()
+  })
```
christian-byrne, line 86: Would be good to add a test case for "timeout fires while an update is already in-flight" — set `isLoading = true` before the timer fires and assert that `update` is not called. This covers the guard suggested in the main file and makes sure we don't send redundant requests when a WS-triggered update is already running.

christian-byrne (approved):
LGTM!

PR #9141, opened 2026-02-24
fix: refresh image previews on media upload nodes when refreshing node definitions

### LINKED ISSUE #2082: [Feature Request]: Refresh node definition should also refresh file previews

- [X] I have searched the existing issues and checked the recent builds/commits

Recently in [this discussion](https://github.com/comfyanonymous/ComfyUI/discussions/6240) in the ComfyUI repo, @DocChopper mentioned the fact that after you edit and save the image from an external software, the preview still displays the previous state, even after refreshing node definition:

https://github.com/user-attachments/assets/7d171449-a451-4039-b175-34ca9f778291

Even if you don't refresh, the output will match the current state of the image (check [this](https://github.com/comfyanonymous/ComfyUI/discussions/6240#discussioncomment-11678436)), so the problem is only related to the preview displayed on the node.

Obviously if you refresh the browser page or select again the image, the preview will be updated, but I think that when you press "R" to refresh node definition, the previews should be updated as well to match the current state of the files. 

### Proposed workflow

1. Edit and save the image in an external software
2. Go to ComfyUI and hit "R" to refresh
3. The updated preview is displayed

### Additional information

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-2082-Feature-Request-Refresh-node-definition-should-also-refresh-file-previews-16e6d73d365081c197aff09664e107f4) by [Unito](https://www.unito.io)

Myestery:
## Summary
- When pressing `R` to refresh node definitions, image previews on LoadImage/LoadVideo nodes now update to reflect external file changes
- Re-triggers the combo widget callback to regenerate preview URLs with a fresh cache-busting `&rand=` parameter
- Extracts `isMediaUploadComboInput` from `uploadImage.ts` to `nodeDefSchema.ts` as a shared utility

- Fixes #2082

https://github.com/user-attachments/assets/d18d69ae-6ecd-448d-8d7c-76b2c49fdea5

## Test plan
- [ ] Open a workflow with a LoadImage node and select an image
- [ ] Edit and save the image externally (e.g. in an image editor)
- [ ] Press `R` to refresh node definitions
- [ ] Verify the preview updates to show the edited image

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -43,7 +43,8 @@ import type {
 import {
   type ComfyNodeDef as ComfyNodeDefV1,
   isComboInputSpecV1,
-  isComboInputSpecV2
+  isComboInputSpecV2,
+  isMediaUploadComboInput
 } from '@/schemas/nodeDefSchema'
 import {
   type BaseDOMWidget,
@@ -1822,6 +1823,7 @@ export class ComfyApp {
       this.registerNodeDef(nodeId, defs[nodeId])
     }
     // Refresh combo widgets in all nodes including those in subgraphs
+    const nodeOutputStore = useNodeOutputStore()
     forEachNode(this.rootGraph, (node) => {
       const def = defs[node.type]
       // Allow primitive nodes to handle refresh
@@ -1853,6 +1855,18 @@ export class ComfyApp {
             }
           }
         }
+      }
+
+      // Re-trigger previews on media upload nodes (e.g. LoadImage)
+      // to bust browser cache when files are edited externally
+      const hasMediaUpload = Object.values(def.input.required ?? {}).some(
+        isMediaUploadComboInput
+      )
+      if (hasMediaUpload && nodeOutputStore.getNodeOutputs(node)) {
+        const comboWidget = node.widgets?.find(
+          (w) => w.type === 'combo' && w.callback
+        )
+        comboWidget?.callback?.(comboWidget.value)
       }
     })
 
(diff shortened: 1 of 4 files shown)


src/scripts/app.ts
```diff
@@ -1854,6 +1856,18 @@ export class ComfyApp {
           }
         }
       }
+
+      // Re-trigger previews on media upload nodes (e.g. LoadImage)
+      // to bust browser cache when files are edited externally
+      const hasMediaUpload = Object.values(def.input.required ?? {}).some(
+        isMediaUploadComboInput
+      )
```
christian-byrne: Should we use the existing type guards for this? https://github.com/Comfy-Org/ComfyUI_frontend/blob/c25f9a0e939145c155fbbd0fef24da4b02a40326/src/utils/litegraphUtil.ts#L65-L76

src/scripts/app.ts
```diff
@@ -1854,6 +1856,18 @@ export class ComfyApp {
...
+      if (hasMediaUpload && nodeOutputStore.getNodeOutputs(node)) {
+        const comboWidget = node.widgets?.find(
+          (w) => w.type === 'combo' && w.callback
+        )
+        comboWidget?.callback?.(comboWidget.value)
```
christian-byrne: Can you check if it will work to just use the `showPreview` method from useNodeImage or useNodeVideo? Or check if we can do this by triggering reactivty from the store?

Myestery: Went with triggering reactivity from the store and added a refreshNodeOutputs method that re-assigns   
  the output object with a shallow copy, which causes the nodeMedia computed in LGraphNode.vue to      
  re-evaluate and call getNodeImageUrls with a fresh &rand= param. Much cleaner than invoking widget   
  callbacks directly.

christian-byrne (approved):
LGTM

PR #9144, opened 2026-02-24
fix: prevent auth errors for API key users on local/desktop

DrJKL:
- Guard subscription status auto-fetch with isCloud check in useSubscription.ts to prevent Firebase auth errors on local/desktop
- Guard fetchBalance call in CurrentUserPopoverLegacy.vue for API key users since fetchBalance requires Firebase auth
- Add test to verify subscription status is not fetched when not on cloud
- Update CurrentUserPopoverLegacy test mock to include isApiKeyLogin

Fixes API key authentication showing "User not authenticated" errors when users click profile icon or try to refresh balance on local/desktop versions.

Slack: https://comfy-organization.slack.com/archives/C0A4XMHANP3/p1771896736700429?thread_ts=1771289819.003239&cid=C0A4XMHANP3

https://claude.ai/code/session_012jjZuGQbnnfcciowSwb4dc

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9144-fix-prevent-auth-errors-for-API-key-users-on-local-desktop-3116d73d365081b68095e405e24fa7f1) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/topbar/CurrentUserPopoverLegacy.vue
+++ b/src/components/topbar/CurrentUserPopoverLegacy.vue
@@ -162,8 +162,13 @@ const emit = defineEmits<{
 
 const { buildDocsUrl, docsPaths } = useExternalLink()
 
-const { userDisplayName, userEmail, userPhotoUrl, handleSignOut } =
-  useCurrentUser()
+const {
+  userDisplayName,
+  userEmail,
+  userPhotoUrl,
+  handleSignOut,
+  isApiKeyLogin
+} = useCurrentUser()
 const authActions = useFirebaseAuthActions()
 const authStore = useFirebaseAuthStore()
 const settingsDialog = useSettingsDialog()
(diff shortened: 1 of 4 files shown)


src/components/topbar/CurrentUserPopoverLegacy.vue
```diff
@@ -244,6 +249,8 @@ const handleSubscribed = async () => {
 }
 
 onMounted(() => {
-  void authActions.fetchBalance()
+  if (!isApiKeyLogin.value) {
+    void authActions.fetchBalance()
+  }
```
christian-byrne, line 254: Actually the fetchBalance does work with API key header.

christian-byrne (commented):
I think it can be closed because it is done in https://github.com/Comfy-Org/ComfyUI_frontend/pull/9142

PR #9149, opened 2026-02-24
feat: select group children on click

christian-byrne:
## Summary

Add a setting to select all children (nodes, reroutes, nested groups) when clicking a group on the canvas.

## Changes

- **What**: New `LiteGraph.Group.SelectChildrenOnClick` boolean setting (default: `false`). When enabled, selecting a group cascades `select()` to all its `_children`, and deselecting cascades `deselect()`. Recursion handles nested groups naturally. No double-move risk — the drag handler already uses `skipChildren=true`. The setting is wired via `onChange` to `canvas.groupSelectChildren`, keeping litegraph free of platform imports.

## Review Focus

- The select/deselect cascading in `LGraphCanvas.select()` / `deselect()` — verify no infinite recursion risk with deeply nested groups.
- The `groupSelectChildren` property is set via the setting's `onChange` callback on `LGraphCanvas.active_canvas` — confirm this covers canvas re-creation scenarios.

N/A — behavioral change behind a setting toggle.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9149-feat-select-group-children-on-click-3116d73d365081a1a7b8c82dea95b242) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphCanvas.groupSelection.test.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.groupSelection.test.ts
@@ -1,5 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
+import type { CanvasPointerEvent } from '@/lib/litegraph/src/types/events'
+
 import {
   LGraph,
   LGraphCanvas,
@@ -288,6 +290,40 @@ describe('LGraphCanvas group selection', () => {
         expect(n.selected).toBe(false)
         expect(canvas.selectedItems.has(n)).toBe(false)
       }
+    })
+  })
+
+  describe('processSelect modifier-click deselect', () => {
+    beforeEach(() => {
+      canvas.groupSelectChildren = true
+    })
+
+    it('modifier-click deselects only the group, not its children', () => {
+      canvas.select(group)
+      expect(group.selected).toBe(true)
+      expect(nodeA.selected).toBe(true)
+      expect(nodeB.selected).toBe(true)
+
+      const shiftEvent = { shiftKey: true } as CanvasPointerEvent
+      canvas.processSelect(group, shiftEvent)
+
+      expect(group.selected).toBe(false)
+      expect(canvas.selectedItems.has(group)).toBe(false)
+      expect(nodeA.selected).toBe(true)
+      expect(nodeB.selected).toBe(true)
+      expect(canvas.selectedItems.has(nodeA)).toBe(true)
+      expect(canvas.selectedItems.has(nodeB)).toBe(true)
+    })
+
+    it('ctrl-click deselects only the group, not its children', () => {
+      canvas.select(group)
+
+      const ctrlEvent = { ctrlKey: true } as CanvasPointerEvent
+      canvas.processSelect(group, ctrlEvent)
+
+      expect(group.selected).toBe(false)
+      expect(nodeA.selected).toBe(true)
+      expect(nodeB.selected).toBe(true)
     })
   })
 
--- a/src/lib/litegraph/src/LGraphCanvas.groupSelection.test.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.groupSelection.test.ts
@@ -1,11 +1,10 @@
-import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 import {
   LGraph,
   LGraphCanvas,
   LGraphGroup,
-  LGraphNode,
-  LiteGraph
+  LGraphNode
 } from '@/lib/litegraph/src/litegraph'
 
 vi.mock('@/renderer/core/layout/store/layoutStore', () => ({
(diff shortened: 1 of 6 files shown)


pythongosssss (commented):
When trying to delete a group + all children, the first thing that happens is the group is removed, which triggers deselection of all children so only the root group is removed

src/platform/settings/constants/coreSettings.ts
```diff
@@ -1243,5 +1247,19 @@ export const CORE_SETTINGS: SettingParams[] = [
...
+  },
+  {
+    id: 'LiteGraph.Group.SelectChildrenOnClick',
+    category: ['LiteGraph', 'Group', 'SelectChildrenOnClick'],
+    name: 'Select group children on click',
+    tooltip:
+      'When enabled, clicking a group selects all nodes and items inside it',
+    type: 'boolean',
+    defaultValue: false,
+    onChange(newValue) {
```
christian-byrne: issue: the `onChange` callback reads `LGraphCanvas.active_canvas`, but at registration time (`CORE_SETTINGS.forEach(settingStore.addSetting)` at GraphCanvas.vue:498), the canvas hasn't been created yet (`comfyApp.setup` at line 512). If a user has previously stored this setting as `true`, `onChange` fires with `active_canvas === null` and the property is never applied. The user would need to toggle the setting off/on each page load.

The established pattern for applying settings to canvas instance properties is a `watchEffect` in `useLitegraphSettings.ts` (see `zoom_speed`, `show_info`, `links_render_mode`, etc.), which reactively re-runs when `canvasStore.canvas` becomes available. Consider moving the canvas binding there and dropping the `onChange`.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -4414,6 +4416,11 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
 
     if (item instanceof LGraphGroup) {
       item.recomputeInsideNodes()
```
christian-byrne, line 4402: suggestion (non-blocking): `recomputeInsideNodes` on the outer group already recursively calls `recomputeInsideNodes` on nested groups (via the `containedGroups` loop in LGraphGroup.ts:286). Then when `select` recurses into the inner group, it calls `recomputeInsideNodes` a second time on that same group. Not a correctness issue, but a redundant O(n) pass over nodes/reroutes/groups for each nesting level.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -4388,6 +4390,24 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
...
+      if (this.groupSelectChildren) {
+        // Iterative traversal avoids stack overflow on deeply nested groups
+        // and skips redundant recomputeInsideNodes (already done recursively)
+        const stack: Positionable[] = [...item._children]
```
pythongosssss: nit: would be nice to dedupe this traversal with a shared helper for use in select/deselect

pythongosssss:
@christian-byrne how do you think this should work for shift/ctrl clicking on the group? 
Scenario:
- Group with nodes in it
- Click group header to select nodes
- Shift/ctrl click group header <--- i'd expect this just to deselect the group
- All nodes & group are deselected <--- I wanted to delete all nodes in the group (I can still click them manually)

pythongosssss (dismissed):
lgtm, just nonblocking question on group modifier click behavior

src/lib/litegraph/src/LGraphCanvas.groupSelection.test.ts
```diff
@@ -0,0 +1,327 @@
...
+
+      // Pre-select the inner group before selecting the outer group
+      canvas.select(innerGroup)
+      expect(innerGroup.selected).toBe(true)
+      expect(innerNode.selected).toBeFalsy()
+
+      canvas.select(group)
+
+      expect(innerNode.selected).toBe(true)
+      expect(canvas.selectedItems.has(innerNode)).toBe(true)
```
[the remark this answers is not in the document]
christian-byrne: The analysis is incorrect here. `innerNode` is NOT in `innerGroup._children` in this test environment because `boundingRect` is never computed (no canvas render cycle runs `calculateBounding`), so the node'\''s centre point stays at default `[0, 0]` and `containsCentre` returns `false` during `recomputeInsideNodes`. The `toBeFalsy()` assertion is correct - the node is intentionally not a child of the group in this test setup.

src/platform/settings/constants/coreSettings.ts
```diff
@@ -1243,5 +1243,15 @@ export const CORE_SETTINGS: SettingParams[] = [
...
     experimental: true,
     versionAdded: '1.40.0'
+  },
+  {
+    id: 'LiteGraph.Group.SelectChildrenOnClick',
+    category: ['LiteGraph', 'Group', 'SelectChildrenOnClick'],
+    name: 'Select group children on click',
+    tooltip:
+      'When enabled, clicking a group selects all nodes and items inside it',
+    type: 'boolean',
```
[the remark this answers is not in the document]
christian-byrne: Settings strings defined in `coreSettings.ts` are auto-generated during the build process, so i18n via `vue-i18n` is not applicable here. Please note this pattern for future reviews of this file.

christian-byrne:
Re: shift/ctrl-click behavior — good catch. Agreed that modifier-click should only toggle the group itself, keeping children selected. This matches the standard multi-select toggle convention (Figma does the same: shift-click removes only the clicked item from the selection).

Updated `processSelect` so modifier-click deselects only the group wrapper without cascading to children. The cascade on `select` still works as before.

pythongosssss (approved):
lgtm

PR #9151, opened 2026-02-24
feat: add batch image navigation to ImageCompare node

### LINKED ISSUE #9098: Image comparer (Nodes 2.0) only shows the first image in a batch
### Prerequisites

- [x] I have custom nodes enabled

### What happened?

As the title says, the image comparer Vue 2 node does not show any of the other images than the first pair when the batch is more than 1.
Whereas rgthree image comparer allowed switching to the different images in the batch.
And as we know, the rgthree node does not work in Nodes 2.0.

### Steps to Reproduce

1. Create a simple image workflow that generates 2 images, with a batch size >1
2. Add image comparer and connect the 2 images
3. Try to see the other images from the batch - no way of doing so as far as I can see

### ComfyUI Frontend Version

1.42.2

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9098-Image-comparer-Nodes-2-0-only-shows-the-first-image-in-a-batch-30f6d73d36508126b235c36b873d61fb) by [Unito](https://www.unito.io)

jtydhr88:
## Summary

Add batch image navigation to the ImageCompare node so users can browse all images in a batch instead of only seeing the first one.                                                    
                                                                                                                                                                                         
## Changes
The backend already returns all batch images in a_images/b_images arrays, but the frontend only used index [0]. Now all images are mapped to URLs and a navigation bar with prev/next controls appears above the comparison slider when either side has more than one image. A/B sides navigate independently. Extracted a reusable BatchNavigation component for the index selector UI.

fix https://github.com/Comfy-Org/ComfyUI_frontend/issues/9098

https://github.com/user-attachments/assets/a801cc96-9182-4b0d-a342-4e6107290f47

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9151-feat-add-batch-image-navigation-to-ImageCompare-node-3116d73d365081498be6d401773619a3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetImageCompare.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetImageCompare.vue
@@ -1,6 +1,33 @@
 <template>
-  <div ref="containerRef" class="relative size-full min-h-32 overflow-hidden">
-    <div v-if="beforeImage || afterImage" class="relative size-full">
+  <div class="flex size-full min-h-32 flex-col overflow-hidden">
+    <div
+      v-if="showBatchNav"
+      class="flex shrink-0 justify-between px-2 py-1 text-xs"
+      data-testid="batch-nav"
+    >
+      <BatchNavigation
+        v-model="beforeIndex"
+        :count="beforeBatchCount"
+        data-testid="before-batch"
+      >
+        <template #label>{{ $t('imageCompare.batchLabelA') }}</template>
+      </BatchNavigation>
+      <div v-if="beforeBatchCount <= 1" />
+
+      <BatchNavigation
+        v-model="afterIndex"
+        :count="afterBatchCount"
+        data-testid="after-batch"
+      >
+        <template #label>{{ $t('imageCompare.batchLabelB') }}</template>
+      </BatchNavigation>
+    </div>
+
+    <div
+      v-if="beforeImage || afterImage"
+      ref="containerRef"
+      class="relative min-h-0 flex-1"
+    >
       <img
         v-if="afterImage"
         :src="afterImage"
@@ -25,7 +52,7 @@
       />
     </div>
 
-    <div v-else class="flex size-full items-center justify-center">
+    <div v-else class="flex min-h-0 flex-1 items-center justify-center">
       {{ $t('imageCompare.noImages') }}
     </div>
   </div>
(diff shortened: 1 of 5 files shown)


christian-byrne (commented):
## Review

Clean PR overall — good test coverage, proper i18n, sensible component extraction.

### Visibility coupling between BatchNavigation and parent spacer (suggestion)

`BatchNavigation.vue` internally uses `v-if="count > 1"` to hide itself, and the parent `WidgetImageCompare.vue` uses `<div v-if="beforeBatchCount <= 1" />` as a layout spacer that implicitly mirrors that same condition. These are inverse conditions across two files that must stay in sync — if `BatchNavigation` ever changes when it renders (e.g. shows a label even for count=1), the parent layout silently breaks.

Cleaner pattern: lift visibility control to the parent with `v-if`/`v-else`, and let `BatchNavigation` always render when mounted:

```html
<BatchNavigation v-if="beforeBatchCount > 1" v-model="beforeIndex" :count="beforeBatchCount" ... />
<div v-else />
```

Then remove the `v-if="count > 1"` from `BatchNavigation.vue`'s root element. This keeps the layout intent explicit in one place and makes `BatchNavigation` a simpler, more reusable component.

src/renderer/extensions/vueNodes/widgets/components/WidgetImageCompare.vue
```diff
@@ -61,14 +92,48 @@ watch([elementX, elementWidth, isOutside], ([x, width, outside]) => {
...
+  () => parsedValue.value?.afterImages,
+  () => {
+    afterIndex.value = 0
+  }
+)
+
 const beforeImage = computed(() => {
   const value = props.widget.value
-  return typeof value === 'string' ? value : value?.before || ''
+  if (typeof value === 'string') return value
```
christian-byrne, line 129: nit: extract to `isSingleImage()` function and use in both beforeImage and afterImage to be more self-documenting and not repeat code.

christian-byrne (approved):
LGTM with nits

src/renderer/extensions/vueNodes/widgets/components/WidgetImageCompare.vue
christian-byrne: @coderabbitai make an issue to do this in a followup pr

PR #9150, opened 2026-02-24
refactor: replace withDefaults with Vue 3.5 props destructuring

Myestery:
## Summary
- Replace all `withDefaults(defineProps<...>())` with Vue 3.5 reactive props destructuring across 14 components
- Update `props.xxx` references to use destructured variables directly in script and template

- Fixes #2334

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9150-refactor-replace-withDefaults-with-Vue-3-5-props-destructuring-3116d73d365081e7a721db3369600671) by [Unito](https://www.unito.io)

PR #9155, opened 2026-02-24
refactor: remove redundant `= false` defaults on boolean props

christian-byrne:
## Summary

Remove redundant `= false` defaults from destructured `defineProps` on boolean props, since Vue's [Boolean casting](https://vuejs.org/guide/components/props.html#boolean-casting) already defaults absent boolean props to `false`.

## Changes

- **What**: Remove 10 unnecessary `= false` default values across 8 components

## Review Focus

Vue compiles `defineProps<{ myBool?: boolean }>()` to `{ myBool: { type: Boolean, required: false } }`. Boolean casting means absent boolean props are already `false` — the explicit `= false` in JS destructuring defaults never triggers.

Follow-up to #9150.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9155-refactor-remove-redundant-false-defaults-on-boolean-props-3116d73d36508196af0bcba53257720a) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/breadcrumb/SubgraphBreadcrumbItem.vue
+++ b/src/components/breadcrumb/SubgraphBreadcrumbItem.vue
@@ -78,7 +78,7 @@ interface Props {
   isActive?: boolean
 }
 
-const { item, isActive = false } = defineProps<Props>()
+const { item, isActive } = defineProps<Props>()
 
 const nodeDefStore = useNodeDefStore()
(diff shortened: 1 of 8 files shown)


Myestery (approved):
None of the removed `= false` defaults are breaking changes, since all usages rely on truthy/falsy checks where undefined behaves the same as false, including the menuHovered case.

PR #9159, opened 2026-02-24
[feat] Add view settings toggles for asset grouping and display options

viva-jinyi:
## Summary

Add three toggles to the media asset view settings menu: "Group assets by job", "Show asset names", and "Show asset details", with localStorage persistence and proper loading states.

## Changes

- **What**: Add `groupByJob`, `showAssetNames`, `showAssetDetails` toggles to `MediaAssetSettingsMenu`, wired through `MediaAssetFilterBar` to `AssetsSidebarTab` via `defineModel` / `v-model`
- **What**: `useUngroupedAssets` composable resolves multi-output jobs into individual assets when grouping is disabled, using `computedAsync` + `useCachedRequest` for memoized async resolution with proper loading skeleton
- **What**: `showAssetNames` / `showAssetDetails` use `provide`/`inject` with typed `InjectionKey<Ref<boolean>>` to bypass VirtualGrid scoped-slot reactivity limitation
- **What**: All settings persisted via `useStorage` (`Comfy.Assets.Sidebar.*`)

## Review Focus

- `useUngroupedAssets` uses a sync `computed` wrapper over `computedAsync` to avoid microtick delay when `groupByJob=true` (default). When `groupByJob=false`, `isResolving` ref drives `showLoadingState` to display skeleton during async resolution.
- `computedAsync` callback returns `[]` (not `assets.value`) when `groupByJob=true` to avoid unnecessary dependency tracking on `assets` — prevents re-evaluation on every search/filter/sort change.

[screen-capture.webm](https://github.com/user-attachments/assets/fe15db16-6806-40eb-9217-45fa14722df3)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9159-feat-Add-view-settings-toggles-for-asset-grouping-and-display-options-3116d73d365081089967c5ca3ce35c0e) by [Unito](https://www.unito.io)


src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -88,19 +90,24 @@
...
         <AssetsSidebarListView
           v-if="isListView"
-          :asset-items="listViewAssetItems"
+          :asset-items="
+            groupByJob ? listViewAssetItems : ungroupedListViewItems
+          "
           :is-selected="isSelected"
-          :selectable-assets="listViewSelectableAssets"
+          :selectable-assets="
+            groupByJob ? listViewSelectableAssets : ungroupedAssets
```
benceruleanlu: nit: Naming seems odd here

groupByJob ? listViewSelectableAssets : ungroupedAssets

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -102,16 +110,21 @@ const {
...
   isSelected: (assetId: string) => boolean
   isStackExpanded: (asset: AssetItem) => boolean
   toggleStack: (asset: AssetItem) => Promise<void>
   assetType?: 'input' | 'output'
+  groupByJob?: boolean
 }>()
 
+const showAssetNames = inject(ShowAssetNamesKey, ref(true))
+const showAssetDetails = inject(ShowAssetDetailsKey, ref(true))
+
```
benceruleanlu: nit: Do we need inject? Why not standard props?

src/platform/assets/composables/useUngroupedAssets.ts
```diff
@@ -0,0 +1,70 @@
...
+import { getOutputAssetMetadata } from '@/platform/assets/schemas/assetMetadataSchema'
+import type { AssetItem } from '@/platform/assets/schemas/assetSchema'
+import { resolveOutputAssetItems } from '@/platform/assets/utils/outputAssetUtil'
+
+export function useUngroupedAssets(
+  assets: Ref<AssetItem[]>,
+  groupByJob: Ref<boolean>
+) {
+  const { call: cachedResolve, cancel } = useCachedRequest((jobId: string) => {
+    const asset = assets.value.find((a) => {
```
benceruleanlu: signal is ignored here, can we implement it?

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -391,21 +414,30 @@ const baseAssets = computed(() => {
...
+  ungroupedAssets.value.map((asset) => ({
+    key: `asset-${asset.id}`,
+    asset
+  }))
+)
+
 const visibleAssets = computed(() => {
-  if (!isListView.value) return displayAssets.value
+  if (!isListView.value) return ungroupedAssets.value
+  if (!groupByJob.value) return ungroupedAssets.value
```
benceruleanlu: nit: can we ||?
```suggestion
if (!isListView.value || !groupByJob.value) return ungroupedAssets.value
```

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -425,14 +457,24 @@ const isFolderLoading = computed(
   () => isInFolderView.value && folderLoading.value
 )
 
-const showLoadingState = computed(
-  () =>
-    (loading.value || isFolderLoading.value) && displayAssets.value.length === 0
-)
+const showLoadingState = computed(() => {
+  if (
```
benceruleanlu: nit: readability maybe:

```ts
  const showLoadingState = computed(() => {
    const isPrimaryLoading =
      (loading.value || isFolderLoading.value) &&
      filteredAssets.value.length === 0

    const isUngroupedResolvingEmpty =
      !groupByJob.value &&
      isResolving.value &&
      ungroupedAssets.value.length === 0

    return isPrimaryLoading || isUngroupedResolvingEmpty
  })
```

src/components/sidebar/tabs/AssetsSidebarTab.vue
```diff
@@ -391,21 +414,30 @@ const baseAssets = computed(() => {
 const { searchQuery, sortBy, mediaTypeFilters, filteredAssets } =
   useMediaAssetFiltering(baseAssets)
 
-const displayAssets = computed(() => {
-  return filteredAssets.value
-})
+const { ungroupedAssets, isResolving } = useUngroupedAssets(
+  filteredAssets,
+  groupByJob
+)
```
benceruleanlu, line 422: Filtering then ungrouping can hide valid child outputs

comfydesigner:
Design wise this looks good to me, in the video and from testing. I need to handle this case: 
image

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: Done — extracted `listViewVisibleAssets` computed.

src/components/sidebar/tabs/AssetsSidebarListView.vue
viva-jinyi: Done — switched to props with `true` defaults. `provide` is kept for `MediaAssetCard` (GridView).

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: Done.

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: Done.

src/platform/assets/components/MediaAssetSettingsMenu.vue
```diff
@@ -30,6 +30,42 @@
...
+      variant="textonly"
+      class="w-full"
+      @click="showAssetDetails = !showAssetDetails"
+    >
+      <span>{{ $t('sideToolbar.mediaAssets.showAssetDetails') }}</span>
+      <i
+        class="icon-[lucide--check] ml-auto size-4"
+        :class="!showAssetDetails && 'opacity-0'"
+      />
+    </Button>
```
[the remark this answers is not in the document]
viva-jinyi: Done — added `:aria-pressed` to all three toggle buttons.

src/platform/assets/composables/useUngroupedAssets.ts
viva-jinyi: Implemented. Signal now propagates through the full call chain:
`useCachedRequest` → `resolveOutputAssetItems` → `getJobDetail` → `api.getJobDetail` → `fetchJobDetail` → `fetchApi`

src/platform/assets/composables/useUngroupedAssets.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+  )
+
+  const isResolving = ref(false)
+
+  const resolvedAssets = computedAsync(
+    async (onCancel) => {
+      if (groupByJob.value) return []
+
+      onCancel(() => cancel())
```
benceruleanlu: `onCancel(() => cancel())` can poison the per-job cache with partial results.\n\nWhen this cancellation aborts an in-flight resolve, `fetchJobDetail` currently catches aborts and returns `undefined`, so `resolveOutputAssetItems` falls back to `metadata.allOutputs` (often incomplete) and that value gets cached by `useCachedRequest`. After a quick search/filter change, the same job can stay permanently under-expanded for the rest of the session.\n\nCan we avoid caching aborted runs (for example by propagating abort distinctly and skipping cache writes for aborts), or avoid calling `cancel()` here so in-flight requests can complete and populate the cache with full outputs?

benceruleanlu (changes requested):
Found one correctness issue around cancellation + cache behavior in `useUngroupedAssets`; left an inline comment with reproduction context and suggested direction.

benceruleanlu:
Oops, that was accidental, let me take a look

Edit: it seems accurate, can you TAL? Also, there are some failing tests

src/components/sidebar/tabs/AssetsSidebarTab.vue
benceruleanlu: @viva-jinyi what was the resolution on this? It still looks to be an issue, unless it's intended behaviour

benceruleanlu (commented):
https://github.com/Comfy-Org/ComfyUI_frontend/pull/9159#discussion_r2881544619

This still looks true, is it intended/not a concern?

src/components/sidebar/tabs/AssetsSidebarTab.vue
viva-jinyi: If filtering and ungrouping are used at the same time, items can disappear. But supporting both features together properly would require a large refactor of the Media Asset Panel, moving it from the Job API to the Asset API, so I’m still weighing whether it’s worth doing.

benceruleanlu (commented):
My concern is that deleting a single asset deletes all assets from the job, which is destructive and I can imagine a creative not having a great time seeing the wrong work get mass deleted. Does the API not support this yet? What do you think? Maybe this isn't as high of a concern?

PR #9169, opened 2026-02-24
feat: enhance manager dialog with initial pack id support

jaeone94:
## Summary
Adds `initialPackId` support to the manager dialog so callers can deep-link directly to a specific node pack — pre-filling the search query, switching to packs search mode, and auto-selecting the matching pack once results load.

## Changes
- **ManagerDialog.vue**: Added `initialPackId` prop; wires it into `useRegistrySearch` (forces `packs` mode and pre-fills query) and uses VueUse `until()` to auto-select the target pack and open the right panel once `resultsWithKeys` is populated (one-shot, never re-triggers). Also fixes a latent bug where the effective initial tab (resolving the persisted tab) was not used when determining the initial search mode and query — previously `initialTab` (the raw prop) was checked directly, which would produce incorrect pre-fill when no tab prop was passed but a Missing tab was persisted.
- **useManagerDialog.ts**: Threads `initialPackId` through `show()` into the dialog props
- **useManagerState.ts**: Exposes `initialPackId` in `openManager` options and passes it to `managerDialog.show()`; also removes a stale fallback `show(ManagerTab.All)` call that was redundant for the legacy-only error path

### Refactor: remove `executionIdUtil.ts` and distribute its functions
- **`getAncestorExecutionIds` / `getParentExecutionIds`** → moved to `src/types/nodeIdentification.ts`: both are pure `NodeExecutionId` string operations with no external dependencies, consistent with the existing `parseNodeExecutionId` / `createNodeExecutionId` helpers already in that file
- **`buildSubgraphExecutionPaths`** → moved to `src/platform/workflow/validation/schemas/workflowSchema.ts`: operates entirely on `ComfyNode[]` and `SubgraphDefinition` (both defined there), and `isSubgraphDefinition` is already co-located in the same file
- Tests redistributed accordingly: ancestor/parent ID tests into `nodeIdentification.test.ts`, `buildSubgraphExecutionPaths` tests into `workflowSchema.test.ts`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9169-feat-enhance-manager-dialog-with-initial-pack-id-support-3116d73d365081f7b6a3cbfb2f2755bf) by [Unito](https://www.unito.io)


src/workbench/extensions/manager/components/manager/ManagerDialog.vue
```diff
@@ -475,6 +485,20 @@ watch(
   }
 )
 
+// Auto-select the pack matching initialPackId once displayPacks is populated.
+watch(
```
viva-jinyi: Suggestion: Can you replace watch with VueUse's until? This is a one-shot operation, but watch stays alive for the dialog's lifetime and can re-trigger after tab changes clear the selection.

jaeone94: Good call — switched to `until` in the latest commit. It resolves as soon as the pack appears in results and never re-triggers, which is exactly the semantics we need here.

viva-jinyi (approved):
LGTM

PR #9170, opened 2026-02-24
feat: add performance testing infrastructure with CDP metrics

christian-byrne:
## Summary

Add a permanent, non-failing performance regression detection system using Chrome DevTools Protocol metrics, with automatic PR commenting.

## Changes

- **What**: Performance testing infrastructure — `PerformanceHelper` fixture class using CDP `Performance.getMetrics` to collect `RecalcStyleCount`, `LayoutCount`, `LayoutDuration`, `TaskDuration`, `JSHeapUsedSize`. Adds `@perf` Playwright project (Chromium-only, single-threaded, 60s timeout), 4 baseline perf tests, CI workflow with sticky PR comment reporting, and `perf-report.js` script for generating markdown comparison tables.

## Review Focus

- `PerformanceHelper` uses `page.context().newCDPSession(page)` — CDP is Chromium-only, so perf metrics are not collected on Firefox. This is intentional since CDP gives us browser-level style recalc/layout counts that `performance.mark/measure` cannot capture.
- The CI workflow uses `continue-on-error: true` so perf tests never block merging.
- Baseline comparison uses `dawidd6/action-download-artifact` to download metrics from the target branch, following the same pattern as `pr-size-report.yaml`.

## Stack

This is the foundation PR for the Firefox performance fix stack:
1. **→ This PR: perf testing infrastructure**
2. `perf/fix-cursor-cache` — cursor style caching (depends on this)
3. `perf/fix-subgraph-svg` — SVG pre-rasterization (depends on this)
4. `perf/fix-clippath-raf` — RAF batching for clip-path (depends on this)

PRs 2-4 are independent of each other.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9170-feat-add-performance-testing-infrastructure-with-CDP-metrics-3116d73d3650817cb43def6f8e9917f8) by [Unito](https://www.unito.io)


.github/workflows/ci-perf-report.yaml
```diff
@@ -0,0 +1,106 @@
...
+      pull-requests: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+
+      - name: Setup Node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 22
```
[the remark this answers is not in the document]
christian-byrne: This is a non-issue. The project uses `.pinact.yaml` which explicitly exempts official GitHub actions from SHA pinning:

```yaml
ignore_actions:
  - name: actions/checkout
    ref: v6
  - name: actions/setup-node
    ref: v6
```

The `validate-pins` CI check passes because of these exemptions. SHA-pinning official GitHub actions provides marginal security benefit while significantly hurting maintainability.

scripts/perf-report.ts
```diff
@@ -0,0 +1,122 @@
...
+      '## ⚡ Performance Report\n\nNo perf metrics found. Perf tests may not have run.\n'
+    )
+    process.exit(0)
+  }
+
+  const current: PerfReport = JSON.parse(readFileSync(CURRENT_PATH, 'utf-8'))
+
+  const baseline: PerfReport | null = existsSync(BASELINE_PATH)
+    ? JSON.parse(readFileSync(BASELINE_PATH, 'utf-8'))
+    : null
```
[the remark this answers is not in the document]
christian-byrne: Won't fix for v1. The JSON files are written by our own code (`JSON.stringify`) and by CI artifact download. The `existsSync` check already handles the common failure mode (missing file). If the file exists but contains malformed JSON, the report step fails — but the report job doesn't gate merging, so this is low-risk. Happy to add defensive parsing in a follow-up if it becomes an issue in practice.

scripts/perf-report.ts
```diff
@@ -0,0 +1,125 @@
...
+
+      const layoutDelta = calcDelta(base.layouts, m.layouts)
+      lines.push(
+        `| ${m.name}: layouts | ${base.layouts} | ${m.layouts} | ${formatDeltaCell(layoutDelta)} |`
+      )
+
+      const taskDelta = calcDelta(base.taskDurationMs, m.taskDurationMs)
+      lines.push(
+        `| ${m.name}: task duration | ${base.taskDurationMs.toFixed(0)}ms | ${m.taskDurationMs.toFixed(0)}ms | ${formatDeltaCell(taskDelta)} |`
+      )
```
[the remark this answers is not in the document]
christian-byrne: Intentional omission. Heap delta is excluded from the baseline comparison table because JS heap measurements are highly non-deterministic — they vary significantly between runs due to GC timing, memory pressure, and allocation patterns. Showing percent-change deltas on heap would generate misleading noise. The absolute value in the no-baseline path serves as a rough sanity check; the comparison path focuses on style recalcs, layouts, and task duration which are far more stable and actionable.

browser_tests/helpers/perfReporter.ts
```diff
@@ -0,0 +1,46 @@
...
+  if (tempFiles.length === 0) return
+
+  const measurements: PerfMeasurement[] = tempFiles.map((f) =>
+    JSON.parse(readFileSync(join(TEMP_DIR, f), 'utf-8'))
+  )
+
+  const report: PerfReport = {
+    timestamp: new Date().toISOString(),
+    gitSha: process.env.GITHUB_SHA ?? 'local',
+    branch: process.env.GITHUB_HEAD_REF ?? 'local',
```
DrJKL: Probably fine for now, but this might be annoying if we try to make the action that would invoke this callable.

browser_tests/fixtures/helpers/PerformanceHelper.ts
```diff
@@ -0,0 +1,83 @@
...
+      return metrics.find((m) => m.name === name)?.value ?? 0
+    }
+    return {
+      RecalcStyleCount: get('RecalcStyleCount'),
+      RecalcStyleDuration: get('RecalcStyleDuration'),
+      LayoutCount: get('LayoutCount'),
+      LayoutDuration: get('LayoutDuration'),
+      TaskDuration: get('TaskDuration'),
+      JSHeapUsedSize: get('JSHeapUsedSize'),
+      Timestamp: get('Timestamp')
```
DrJKL, line 62: You could do this fancier, but I'm kind of glad you didn't 😁

browser_tests/fixtures/helpers/PerformanceHelper.ts
```diff
@@ -0,0 +1,83 @@
...
+    return {
+      name,
+      durationMs: (after.Timestamp - before.Timestamp) * 1000,
+      styleRecalcs: after.RecalcStyleCount - before.RecalcStyleCount,
+      styleRecalcDurationMs:
+        (after.RecalcStyleDuration - before.RecalcStyleDuration) * 1000,
+      layouts: after.LayoutCount - before.LayoutCount,
+      layoutDurationMs: (after.LayoutDuration - before.LayoutDuration) * 1000,
+      taskDurationMs: (after.TaskDuration - before.TaskDuration) * 1000,
+      heapDeltaBytes: after.JSHeapUsedSize - before.JSHeapUsedSize
```
DrJKL: Given the repetition, maybe fancier could help. v2 improvement?

DrJKL (dismissed):

browser_tests/helpers/perfReporter.ts
christian-byrne: Good call — updated `writePerfReport()` to accept optional `gitSha` and `branch` params (defaulting to env vars) so it's callable from other contexts.

browser_tests/fixtures/helpers/PerformanceHelper.ts
christian-byrne: Thanks! Keeping it straightforward.

browser_tests/fixtures/helpers/PerformanceHelper.ts
christian-byrne: Added a `delta()` helper to `stopMeasuring()` — removes the repeated `after.X - before.X` pattern while keeping the field mapping explicit.

PR #9173, opened 2026-02-24
fix: batch updateClipPath via requestAnimationFrame

christian-byrne:
## Summary

Batch `getBoundingClientRect()` calls in `updateClipPath` via `requestAnimationFrame` to avoid forced synchronous layout.

## Changes

- **What**: Wrap the layout-reading portion of `updateClipPath` in `requestAnimationFrame()` with cancellation. Multiple rapid calls within the same frame are coalesced into a single layout read. Eliminates ~1,053 forced synchronous layouts per profiling session.

## Review Focus

- `getBoundingClientRect()` forces synchronous layout. When interleaved with style mutations (from PrimeVue `useStyle`, cursor writes, Vue VDOM patching), this creates layout thrashing — especially in Firefox where Stylo aggressively invalidates the entire style cache.
- The RAF wrapper coalesces all calls within a frame into one, reading layout only once per frame. The `cancelAnimationFrame` ensures only the latest parameters are used.
- `willChange: 'clip-path'` is included to hint the browser to optimize clip-path animations.

## Stack

4 of 4 in Firefox perf fix stack. Depends on #9170.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9173-fix-batch-updateClipPath-via-requestAnimationFrame-3116d73d3650810392f7fba7ea5ceb6f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/element/useDomClipping.test.ts
+++ b/src/composables/element/useDomClipping.test.ts
@@ -1,0 +1,224 @@
+import { describe, expect, it, vi, beforeEach, afterEach } from 'vitest'
+
+import { useDomClipping } from './useDomClipping'
+
+function createMockElement(rect: {
+  left: number
+  top: number
+  width: number
+  height: number
+}): HTMLElement {
+  return {
+    getBoundingClientRect: vi.fn(
+      () =>
+        ({
+          ...rect,
+          x: rect.left,
+          y: rect.top,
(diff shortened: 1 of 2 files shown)


pythongosssss (approved):
lgtm

PR #9177, opened 2026-02-24
fix: prevent infinite node resize loop in Vue mode

DrJKL:
## Summary

Fix infinite node resize loop in Vue mode where textarea widgets caused nodes to grow ~33px per frame indefinitely.

## Changes

- **What**: Two feedback loops broken in the LiteGraph↔Vue layout sync:
  1. `_arrangeWidgets()` in LiteGraph's draw loop was calling `setSize()` every frame with its own computed widget height, which disagreed with Vue's DOM-measured height. Guarded with `!LiteGraph.vueNodesMode`.
  2. `useLayoutSync` was calling `setSize()` which triggers the size setter → writes back to layoutStore with `source=Canvas` → `handleLayoutChange` updates CSS vars → ResizeObserver fires → loop. Changed to direct array assignment (matching the existing position sync pattern).

## Review Focus

- The `_arrangeWidgets` guard: in Vue mode, the DOM/ResizeObserver is the source of truth for node sizing, so LiteGraph should not grow nodes via `setSize()`. Verify no Vue-mode features depend on this growth path.
- The `useLayoutSync` change: `liteNode.size[0] = ...` modifies `_size` via the getter without triggering the setter, avoiding the Canvas-source bounce. `onResize` is still called. Verify no downstream code relies on the setter side effects when syncing from layout store.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9177-fix-prevent-infinite-node-resize-loop-in-Vue-mode-3116d73d365081e4ad88f1cfad51df18) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphNode.ts
+++ b/src/lib/litegraph/src/LGraphNode.ts
@@ -4187,7 +4187,12 @@ export class LGraphNode
     // Ref: https://github.com/Comfy-Org/ComfyUI_frontend/issues/2652
     // TODO: Move the layout logic before drawing of the node shape, so we don't
     // need to trigger extra round of rendering.
-    if (y > bodyHeight) {
+    // In Vue mode, the DOM is the source of truth for node sizing — the
+    // ResizeObserver feeds measurements back to the layout store.  Allowing
+    // LiteGraph to also call setSize() here creates an infinite feedback loop
+    // (LG grows node → CSS min-height increases → textarea fills extra space →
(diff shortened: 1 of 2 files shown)


AustinMroz (approved):

AustinMroz:
I can't think of any nodes that modify their own size for any reason other than to 'shrinkwrap' to the minimum size required by widgets which is a non-issue in vue mode. I am confident the change is appropriate and safe.

PR #9183, opened 2026-02-25
refactor(mask editor): make 'clear mask' much easier

kaili-yang:
## Summary

Added a "Clear mask" button directly to the Image Preview component, allowing users to clear masks with a single click without needing to open the full Mask Editor dialog.

## Changes

Added a new eraser icon button that clears the image mask in the background using the existing MaskEditor

## Screenshots 

https://github.com/user-attachments/assets/07d3b0f9-b7cd-4a14-b94d-5e43eef7089a

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9183-refactor-mask-editor-make-clear-mask-much-easier-3126d73d365081ee82ffef59dbe8aed0) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -227,6 +246,65 @@ const handleEditMask = () => {
...
+    const imgCtx = imgCanvas.getContext('2d')
+    imgCtx?.drawImage(baseImage, 0, 0)
+
+    const maskCanvas = document.createElement('canvas')
+    maskCanvas.width = width
+    maskCanvas.height = height
+
+    const rgbCanvas = document.createElement('canvas')
+    rgbCanvas.width = width
+    rgbCanvas.height = height
```
jtydhr88: I am not sure we should call canvas here directly.
can you check how we clear mask in mask editor? 

Image

src/composables/maskeditor/useMaskEditor.ts
```diff
@@ -42,7 +54,55 @@ export function useMaskEditor() {
...
+    isClearingMask.value = true
+
+    const dataStore = useMaskEditorDataStore()
+    const editorStore = useMaskEditorStore()
+    const loader = useMaskEditorLoader()
+    const saver = useMaskEditorSaver()
+    const canvasTools = useCanvasTools()
+
+    try {
+      await loader.loadFromNode(node)
```
christian-byrne, line 89: issue (blocking): `clearMask()` operates on uninitialized canvases and will always fail.

`loader.loadFromNode(node)` populates `dataStore.inputData` but does NOT mount `MaskEditorContent`, so `editorStore.maskCanvas`, `editorStore.imgCanvas`, and `editorStore.maskCtx` are never initialized. This means:
1. `canvasTools.clearMask()` silently no-ops (guarded by `if (maskCtx && maskCanvas)`)
2. `saver.save()` throws `'Canvas not initialized'` because `prepareOutputData()` checks for the canvases

Every "Clear mask" click will throw an error and never actually clear the mask. You'll need to either programmatically create/initialize the canvases in headless mode, or use a different approach (e.g. generating a blank mask image directly and saving it).

src/components/maskeditor/dialog/TopBarHeader.vue
```diff
@@ -106,10 +106,6 @@
         <button :class="textButtonClass" @click="onInvert">
           {{ t('maskEditor.invert') }}
         </button>
```
christian-byrne, line 108: issue: Removing the Clear button from the Mask Editor dialog is a UX regression. Users who open the full Mask Editor can no longer clear the mask from within it. The new ImagePreview "Clear mask" button should be additive -- both entry points should exist. Consider restoring the Clear button here.

src/composables/maskeditor/useMaskEditor.ts
```diff
@@ -2,8 +2,20 @@ import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
...
+import { useMaskEditorDataStore } from '@/stores/maskEditorDataStore'
+import { useMaskEditorStore } from '@/stores/maskEditorStore'
+import { useMaskEditorLoader } from '@/composables/maskeditor/useMaskEditorLoader'
+import { useMaskEditorSaver } from '@/composables/maskeditor/useMaskEditorSaver'
+import { useCanvasTools } from '@/composables/maskeditor/useCanvasTools'
+import { useToast } from 'primevue/usetoast'
+import { useI18n } from 'vue-i18n'
+import { ref } from 'vue'
+
+const isClearingMask = ref(false)
```
christian-byrne, line 14: suggestion: `isClearingMask` is declared at module scope here (shared across ALL callers) AND also as a local `ref` in `ImagePreview.vue:170`. Two competing sources of truth.

The module-scope ref means if two `ImagePreview` instances exist, one clearing blocks the other. The component-local ref can get stuck at `true` if `clearMask()` returns early (e.g. the dialog-open guard returns before reaching `finally`).

Pick one location. If the guard should be global (only one clear at a time), use only the module-scope ref and return it from the composable. If per-component, remove the module-scope ref.

src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -227,6 +241,28 @@ const handleEditMask = () => {
...
+  if (!node) return
+
+  isClearingMask.value = true
+
+  try {
+    await maskEditor.clearMask(node)
+  } catch (error) {
+    useToast().add({
+      severity: 'error',
+      summary: 'Error',
```
christian-byrne: suggestion (non-blocking): `summary: 'Error'` is hardcoded English (should be an i18n key), and `t('g.errorLoadingImage')` is semantically wrong for a mask-clearing error.

src/renderer/extensions/vueNodes/components/ImagePreview.vue
```diff
@@ -227,6 +241,28 @@ const handleEditMask = () => {
...
+  if (!props.nodeId || isClearingMask.value) return
+  const node = app.rootGraph?.getNodeById(Number(props.nodeId))
+  if (!node) return
+
+  isClearingMask.value = true
+
+  try {
+    await maskEditor.clearMask(node)
+  } catch (error) {
+    useToast().add({
```
christian-byrne: suggestion (non-blocking): `useToast()` is called inside the async `handleClearMask` handler rather than at component setup time. The composable is already called at the top of `useMaskEditor` -- consider reusing that instance or calling `useToast()` at setup scope in this component and referencing it here.

src/composables/maskeditor/useMaskEditor.ts
```diff
@@ -42,7 +54,55 @@ export function useMaskEditor() {
...
+    const canvasTools = useCanvasTools()
+
+    try {
+      await loader.loadFromNode(node)
+
+      if (!dataStore.inputData) throw new Error('Failed to load image data')
+
+      canvasTools.clearMask()
+      await saver.save()
+    } catch (error) {
```
christian-byrne: nitpick (non-blocking): `catch (error) { throw error }` has no effect -- the error is just re-thrown unchanged. Remove the `catch` block and keep only `try`/`finally`.

christian-byrne (changes requested):

src/composables/maskeditor/useMaskEditor.ts
kaili-yang: Yeah, I chose to create/initialize the canvases in headless mode. Fixed already.

src/renderer/extensions/vueNodes/components/ImagePreview.test.ts
```diff
@@ -54,7 +55,8 @@ describe('ImagePreview', () => {
           createTestingPinia({
             createSpy: vi.fn
           }),
-          i18n
+          i18n,
+          ToastService
         ],
```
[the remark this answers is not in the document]
christian-byrne: note (non-blocking): agreed this should have behavioral test coverage for the clear-mask flow, but shouldn't block this PR.

christian-byrne:
This looks good but it has some conflicts

PR #9189, opened 2026-02-25
feat: show loading spinner and uploading filename during image upload

dante01yoon:
## Summary
- Show a canvas-based loading spinner on image upload nodes (LoadImage) during file upload via drag-drop, paste, or file picker
- Display the uploading file's name immediately in the filename dropdown instead of showing the previous file's name
- Show the uploading audio file's name immediately in the audio widget during upload

## Changes
- **`useNodeImageUpload.ts`**: Add `isUploading` flag and `onUploadStart` callback to the upload lifecycle; clear `node.imgs` during upload to prevent stale previews
- **`useImagePreviewWidget.ts`**: Add `renderUploadSpinner` that draws an animated arc spinner on the canvas when `node.isUploading` is true; guard against empty `imgs` array
- **`useImageUploadWidget.ts`**: Set `fileComboWidget.value` to the new filename on upload start; clear `node.imgs` on combo widget change
- **`uploadAudio.ts`**: Set `audioWidget.value` to the new filename on upload start
- **`litegraph-augmentation.d.ts`**: Add `isUploading` property to `LGraphNode`

https://github.com/user-attachments/assets/818ce529-cb83-428a-8c98-dd900a128343

## Test plan
- [x] Upload an image via file picker on LoadImage node — spinner shows during upload, filename updates immediately
- [x] Drag-and-drop an image onto LoadImage node — same behavior
- [x] Paste an image onto LoadImage node — same behavior
- [x] Change the dropdown selection on LoadImage — old preview clears, new image loads
- [x] Upload an audio file — filename updates immediately in the widget

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9189-feat-show-loading-spinner-and-uploading-filename-during-image-upload-3126d73d365081e4af27cd7252f34298) by [Unito](https://www.unito.io)


src/composables/node/useNodeImageUpload.ts
```diff
@@ -90,10 +91,21 @@ export const useNodeImageUpload = (
...
+    options.onUploadStart?.(files)
+
+    try {
+      const paths = await Promise.all(files.map(handleUpload))
+      const validPaths = paths.filter((p): p is string => !!p)
+      if (validPaths.length) onUploadComplete(validPaths)
+      return validPaths
+    } finally {
+      node.isUploading = false
+      node.graph?.setDirtyCanvas(true)
```
christian-byrne, line 117: Is this line necessary? I believe the onUploadComplete path will trigger a re-render already but could be wrong

src/renderer/extensions/vueNodes/widgets/composables/useImageUploadWidget.ts
```diff
@@ -59,6 +59,11 @@ export const useImageUploadWidget = () => {
       fileFilter,
       accept,
       folder,
+      onUploadStart: (files) => {
+        if (files.length > 0) {
+          fileComboWidget.value = files[0].name
+        }
+      },
```
christian-byrne, line 71: question: Is this just so we update the combo value faster? The widget value will still get updated in the code paths that will follow after upload. I think that's a good idea, but just clarifying. An issue though with this: what if the upload fails? Then we wouldn't want the eager write to widget value, right?

src/composables/node/useNodeImageUpload.ts
```diff
@@ -90,10 +91,21 @@ export const useNodeImageUpload = (
...
   const handleUploadBatch = async (files: File[]) => {
-    const paths = await Promise.all(files.map(handleUpload))
-    const validPaths = paths.filter((p): p is string => !!p)
-    if (validPaths.length) onUploadComplete(validPaths)
-    return validPaths
+    if (node.isUploading) return []
+    node.isUploading = true
+    node.imgs = undefined
+    node.graph?.setDirtyCanvas(true)
+    options.onUploadStart?.(files)
```
christian-byrne: since this is called outside the finally, block, if it fails, the finally block never runs and node.isUploading will be stuck at true right? Which would block further uploads.

src/renderer/extensions/vueNodes/widgets/composables/useImagePreviewWidget.ts
```diff
@@ -43,6 +43,31 @@ function scheduleDeferredImageRender() {
...
+  ctx.save()
+  ctx.strokeStyle = LiteGraph.NODE_TEXT_COLOR
+  ctx.lineWidth = 3
+  ctx.lineCap = 'round'
+  ctx.beginPath()
+  ctx.arc(centerX, centerY, radius, angle, angle + Math.PI * 1.5)
+  ctx.stroke()
+  ctx.restore()
+
+  node.graph?.setDirtyCanvas(true)
```
christian-byrne, line 73: nit: dirtying on every frame might not be necessary.

src/composables/node/useNodeImageUpload.ts
dante01yoon: Yes, it's needed for the all-fail case. If every file fails, `validPaths` is empty, 
so `onUploadComplete` is never called and no re-render is triggered. Without this, 
the spinner would stay visible after a failed upload.

src/renderer/extensions/vueNodes/widgets/composables/useImageUploadWidget.ts
dante01yoon: Good catch — will move onUploadStart inside the try block and restore the previous 
widget value on failure.

src/composables/node/useNodeImageUpload.ts
dante01yoon: You're right. Will move onUploadStart (and the other setup lines) inside the try block.

src/renderer/extensions/vueNodes/widgets/composables/useImagePreviewWidget.ts
dante01yoon: Agreed — will remove. A static spinner frame is sufficient since the upload completion 
already triggers a re-render.

src/renderer/extensions/vueNodes/widgets/composables/useImagePreviewWidget.ts
```diff
@@ -43,6 +43,29 @@ function scheduleDeferredImageRender() {
...
+  node: LGraphNode,
+  shiftY: number,
+  computedHeight: number | undefined
+) {
+  const dw = node.size[0]
+  const dh = computedHeight ?? 220
+  const centerX = dw / 2
+  const centerY = shiftY + dh / 2
+  const radius = 16
+  const angle = ((Date.now() % 1000) / 1000) * Math.PI * 2
```
christian-byrne: The spinner animation currently relies on `set_canvas_dirty_on_mouse_event` to trigger redraws via mouse movement, which means it works during drag-and-drop but will freeze if the user uploads via paste (Ctrl+V) or file picker and then does not move their mouse -- nothing is continuously dirtying the canvas between the start and end `setDirtyCanvas` calls.

Consider having `renderUploadSpinner` call `node.graph?.setDirtyCanvas(true)` at the end so each draw schedules the next frame, or start a small RAF loop when `isUploading` flips true.

src/renderer/extensions/vueNodes/widgets/composables/useImagePreviewWidget.ts
```diff
@@ -43,6 +43,29 @@ function scheduleDeferredImageRender() {
...
+
+  ctx.save()
+  ctx.strokeStyle = LiteGraph.NODE_TEXT_COLOR
+  ctx.lineWidth = 3
+  ctx.lineCap = 'round'
+  ctx.beginPath()
+  ctx.arc(centerX, centerY, radius, angle, angle + Math.PI * 1.5)
```
christian-byrne: nit: `Math.PI * 2` and `Math.PI * 1.5` are recomputed every call. Consider extracting as module-level constants, e.g. `const TWO_PI = Math.PI * 2` and `const ARC_LENGTH = Math.PI * 1.5`.

src/renderer/extensions/vueNodes/widgets/composables/useImagePreviewWidget.ts
```diff
@@ -43,6 +43,29 @@ function scheduleDeferredImageRender() {
...
+  computedHeight: number | undefined
+) {
+  const dw = node.size[0]
+  const dh = computedHeight ?? 220
+  const centerX = dw / 2
+  const centerY = shiftY + dh / 2
+  const radius = 16
+  const angle = ((Date.now() % 1000) / 1000) * Math.PI * 2
+
+  ctx.save()
```
christian-byrne, line 62: nit: once the animation issue above is fixed, this runs every frame. The caller's per-node save/restore in `drawFrontCanvas` (lines 4917/4929) already handles cleanup, so you could set the properties directly and skip the state stack push/pop.

src/composables/node/useNodeImageUpload.ts
```diff
@@ -90,10 +92,26 @@ export const useNodeImageUpload = (
   }
 
   const handleUploadBatch = async (files: File[]) => {
-    const paths = await Promise.all(files.map(handleUpload))
-    const validPaths = paths.filter((p): p is string => !!p)
-    if (validPaths.length) onUploadComplete(validPaths)
-    return validPaths
+    if (node.isUploading) return []
```
christian-byrne: If a user tries to upload while another is in progress, this silently returns `[]` with no feedback. Consider disabling the upload button/drop zone while `isUploading` is true, or showing a toast so the user understands why nothing happened.

src/composables/node/useNodeImageUpload.ts
```diff
@@ -90,10 +92,26 @@ export const useNodeImageUpload = (
...
+    node.isUploading = true
+
+    try {
+      node.imgs = undefined
+      node.graph?.setDirtyCanvas(true)
+      options.onUploadStart?.(files)
+
+      const paths = await Promise.all(files.map(handleUpload))
```
christian-byrne, line 107: nit: not introduced by this PR, but worth noting -- if the fetch hangs indefinitely, the spinner stays forever since there's no timeout or abort mechanism. An `AbortController` with a reasonable timeout could be added separately.

@coderabbitai please make an Issue to do this in a separate PR on a future date

src/renderer/extensions/vueNodes/widgets/composables/useImageUploadWidget.ts
```diff
@@ -54,11 +54,21 @@ export const useImageUploadWidget = () => {
       createAnnotatedPath(value, { rootFolder: image_folder })
 
     // Setup file upload handling
+    let previousValue = fileComboWidget.value
```
christian-byrne: nit: `let previousValue` is safe here because the `isUploading` guard prevents concurrent uploads on the same node, but as a style improvement you could scope the rollback value more tightly per-upload. For example, capture it as a const inside `onUploadStart` and thread a rollback function to `onUploadError`:

```typescript
let rollback: (() => void) | undefined
onUploadStart: (files) => {
  if (files.length > 0) {
    const prev = fileComboWidget.value
    fileComboWidget.value = files[0].name
    rollback = () => { fileComboWidget.value = prev }
  }
},
onUploadError: () => {
  rollback?.()
  rollback = undefined
}
```

This keeps the captured value immutable and scoped to the upload that produced it.

src/renderer/extensions/vueNodes/widgets/composables/useImagePreviewWidget.ts
```diff
@@ -43,6 +43,29 @@ function scheduleDeferredImageRender() {
   })
 }
 
+function renderUploadSpinner(
```
christian-byrne, line 49: nit: when `vueNodesMode` is enabled, `drawNode` skips all canvas rendering (LGraphCanvas line 5396-5403), so this spinner won't be visible. In a follow-up PR, a Vue-based spinner for the image preview overlay would be needed for Nodes 2.0 compatibility.

@coderabbitai please make an Issue to do this in a separate PR on a future date

src/extensions/core/uploadAudio.ts
```diff
@@ -232,6 +232,7 @@ app.registerExtension({
 
         const handleUpload = async (files: File[]) => {
           if (files?.length) {
+            audioWidget.value = files[0].name
```
christian-byrne, line 239: This optimistically sets `audioWidget.value` which is great UX, but unlike the image upload path there's no rollback if the upload fails. If the server returns non-200, the widget shows the bare filename of a file that doesn't exist on the server. Consider adding the same `previousValue` rollback pattern used in `useImageUploadWidget`.

src/composables/node/useNodeImageUpload.ts
```diff
@@ -90,10 +92,26 @@ export const useNodeImageUpload = (
   }
 
   const handleUploadBatch = async (files: File[]) => {
-    const paths = await Promise.all(files.map(handleUpload))
-    const validPaths = paths.filter((p): p is string => !!p)
-    if (validPaths.length) onUploadComplete(validPaths)
-    return validPaths
+    if (node.isUploading) return []
+    node.isUploading = true
```
christian-byrne, line 100: The upload lifecycle (`isUploading` flag, dedup guard, error rollback, `imgs` clearing) is untested. Even lightweight tests for `useNodeImageUpload` with a mocked `api` would help catch regressions on the flag management and rollback logic.

src/renderer/extensions/vueNodes/widgets/composables/useImagePreviewWidget.ts
christian-byrne: @coderabbitai please make an Issue to do this in a separate PR on a future date

christian-byrne (approved):
LGTM!

PR #9195, opened 2026-02-25
feat(settings): improve search to include nav items and show all results

dante01yoon:
## Summary
- Settings search now matches sidebar navigation items (Keybinding, About, Extension, etc.) and navigates to the corresponding panel
- Search results show all matching settings across all categories instead of filtering to only the first matching category
- Search result group headers display parent category prefix (e.g. "LiteGraph › Node") for clarity

## Test plan
- [x] Search "Keybinding" → sidebar highlights and navigates to Keybinding panel
- [x] Search "badge" → shows all 4 badge settings (3 LiteGraph + 1 Comfy)
- [x] Search "canvas" → shows results from all categories
- [x] Clear search → returns to default category
- [x] Unit tests pass (`pnpm test:unit`)
스크린샷 2026-02-25 오후 3 01 05

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9195-feat-settings-improve-search-to-include-nav-items-and-show-all-results-3126d73d3650814dbf3ce1d59ad962cf) by [Unito](https://www.unito.io)


christian-byrne (commented):
nit: the HTML entity works fine, though you might consider a CSS/Tailwind approach or a named constant if this separator pattern shows up elsewhere in the codebase.

src/platform/settings/components/SettingGroup.vue
```diff
@@ -2,6 +2,15 @@
...
     <Divider v-if="divider" />
     <h3>
+      <span v-if="group.category" class="text-muted">
+        {{
+          $t(
+            `settingsCategories.${normalizeI18nKey(group.category)}`,
+            group.category
+          )
+        }}
+        &#8250;
```
christian-byrne, line 12: nit: the HTML entity works fine, though you might consider a CSS/Tailwind approach or a named constant if this separator pattern shows up elsewhere in the codebase.

src/platform/settings/composables/useSettingSearch.ts
```diff
@@ -99,30 +120,42 @@ export function useSettingSearch() {
...
       }
     })
 
-    return Object.entries(groupedSettings).map(([label, settings]) => ({
-      label,
-      settings
-    }))
+    return Object.entries(groupedSettings).map(
+      ([key, { category, settings }]) => ({
+        label: activeCategory === null ? key.split('/')[1] : key,
```
christian-byrne, line 146: This split approach assumes subcategory names won't contain a forward slash. Seems fine for now since category names are controlled, but worth keeping in mind if that ever changes.

src/platform/settings/components/SettingDialog.vue
```diff
@@ -119,16 +120,30 @@ const authActions = useFirebaseAuthActions()
...
-    })
-  activeCategoryKey.value = firstMatch?.id ?? null
-})
+const searchableNavItems = computed(() =>
+  navGroups.value.flatMap((g) =>
+    g.items.map((item) => ({
+      key: item.id,
+      label: item.id,
+      translatedLabel: item.label
+    }))
```
christian-byrne, line 128: Just confirming: both key and label are set to item.id, while translatedLabel gets item.label. Is that the intended mapping? The naming is a bit confusing since item.label is the translated one but you're calling it translatedLabel here.

src/platform/settings/components/SettingGroup.vue
dante01yoon: Thanks for the note! Since this separator is only used in this one place, I think the HTML entity is clear enough as-is. If it starts appearing elsewhere we can extract it then.

src/platform/settings/composables/useSettingSearch.ts
dante01yoon: Good call-out. Category names are internally controlled strings right now so there's no risk, but worth keeping in mind if that ever changes.

src/platform/settings/components/SettingDialog.vue
dante01yoon: Good catch. label was redundant with key here. Simplified SearchableNavItem to just { key, label } where label is the displayed (translated) name. Tests updated and passing.

christian-byrne (approved):
LGTM

PR #9198, opened 2026-02-25
fix: add GLSLShader to canvas image preview node types

christian-byrne:
## Summary

Add `GLSLShader` to `CANVAS_IMAGE_PREVIEW_NODE_TYPES` so GLSL shader previews are promoted through subgraph nodes.

## Changes

- Add `'GLSLShader'` to the `CANVAS_IMAGE_PREVIEW_NODE_TYPES` set in `src/composables/node/useNodeCanvasImagePreview.ts`

## Context

GLSLShader node previews were not showing on parent subgraph nodes because `CANVAS_IMAGE_PREVIEW_NODE_TYPES` only included `PreviewImage` and `SaveImage`. The `$$canvas-image-preview` pseudo-widget was never created for GLSLShader nodes, so the promotion system had nothing to promote. This degraded the UX of all 12 shipped GLSL blueprint subgraphs — users couldn't see shader output previews without expanding the subgraph.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9198-fix-add-GLSLShader-to-canvas-image-preview-node-types-3126d73d3650817dbe9beab4bdeaa414) by [Unito](https://www.unito.io)

PR #9200, opened 2026-02-25
feat: GLSLPreviewEngine + GLSL utility functions

christian-byrne:
## Summary

Standalone WebGL2 rendering engine for client-side GLSL shader preview, with utility functions that mirror the backend `nodes_glsl.py` detection logic.

## Changes

- **What**: New `GLSLPreviewEngine` class (OffscreenCanvas + WebGL2), `glslUtils.ts` (detectOutputCount, detectPassCount, hasVersionDirective), and unit tests
- **GLSLPreviewEngine**: Fullscreen triangle via `gl_VertexID` (no VBO), ping-pong FBOs for multi-pass rendering, MRT via `gl.drawBuffers()`, blob output via `canvas.convertToBlob()`
- **glslUtils**: Pure functions ported from backend Python to TypeScript, regex-based detection matching `_detect_output_count()` and `_detect_pass_count()`

## Review Focus

- WebGL2 resource lifecycle (context loss, texture cleanup, FBO teardown in `dispose()`)
- Ping-pong FBO logic for multi-pass shaders
- Engine tests are WebGL2-gated (`describe.skip` in happy-dom) — they run in real browser environments

## Stacked PR

PR 2 of 3. Stacked on #9198 (fix: GLSLShader preview promotion).
PR 3: `feat/glsl-live-preview` (composable + LGraphNode.vue integration)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9200-feat-GLSLPreviewEngine-GLSL-utility-functions-3126d73d3650812fadc6df4a26387d0e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/glsl/useGLSLRenderer.ts
+++ b/src/renderer/glsl/useGLSLRenderer.ts
@@ -1,6 +1,6 @@
 import { onScopeDispose } from 'vue'
 
-import { detectOutputCount, detectPassCount } from '@/renderer/glsl/glslUtils'
+import { detectPassCount } from '@/renderer/glsl/glslUtils'
 
 const VERTEX_SHADER_SOURCE = `#version 300 es
 out vec2 v_texCoord;
@@ -73,7 +73,6 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
     length: maxInputs
   }).fill(null)
   const uniformLocations = new Map<string, WebGLUniformLocation | null>()
-  let outputCount = 1
   let passCount = 1
   let disposed = false
 
@@ -171,25 +170,29 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
   function init(width: number, height: number): boolean {
     if (disposed) return false
 
-    canvas = new OffscreenCanvas(width, height)
-    const ctx = canvas.getContext('webgl2', {
-      alpha: true,
-      premultipliedAlpha: false,
-      preserveDrawingBuffer: true
-    })
-    if (!ctx) return false
+    try {
+      canvas = new OffscreenCanvas(width, height)
+      const ctx = canvas.getContext('webgl2', {
+        alpha: true,
+        premultipliedAlpha: false,
+        preserveDrawingBuffer: true
+      })
+      if (!ctx) return false
 
-    gl = ctx
-    gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, true)
-    vertexShader = compileShader(gl, gl.VERTEX_SHADER, VERTEX_SHADER_SOURCE)
-    initPingPongFBOs(gl, width, height)
-    return true
+      gl = ctx
+      gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, true)
+      vertexShader = compileShader(gl, gl.VERTEX_SHADER, VERTEX_SHADER_SOURCE)
+      initPingPongFBOs(gl, width, height)
+      return true
+    } catch {
+      dispose()
+      return false
+    }
   }
 
   function compileFragment(source: string): CompileResult {
     if (disposed || !gl) return { success: false, log: 'Engine disposed' }
 
-    outputCount = detectOutputCount(source)
     passCount = Math.min(detectPassCount(source), MAX_PASSES)
 
     if (fragmentShader) {
@@ -261,9 +264,15 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
     image: HTMLImageElement | ImageBitmap
   ): void {
     if (disposed || !gl) return
+    if (index < 0 || index >= maxInputs) {
+      throw new Error(
+        `Input index ${index} out of range (max ${maxInputs - 1})`
+      )
+    }
 
     if (inputTextures[index]) {
       gl.deleteTexture(inputTextures[index])
+      inputTextures[index] = null
     }
 
     const texture = gl.createTexture()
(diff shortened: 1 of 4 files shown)


src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -0,0 +1,395 @@
...
+    const resLoc = uniformLocations.get('u_resolution')
+    if (resLoc != null) {
+      gl.uniform2f(resLoc, canvas.width, canvas.height)
+    }
+
+    for (let i = 0; i < MAX_INPUTS; i++) {
+      const loc = uniformLocations.get(`u_image${i}`)
+      if (loc != null) {
+        gl.activeTexture(gl.TEXTURE0 + i)
+        gl.bindTexture(gl.TEXTURE_2D, inputTextures[i] ?? createEmptyTexture())
```
pythongosssss: this createEmptyTexture isn't destroyed

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -0,0 +1,395 @@
...
+    const h = canvas!.height
+    const pixels = new Uint8ClampedArray(w * h * 4)
+    gl!.readPixels(0, 0, w, h, gl!.RGBA, gl!.UNSIGNED_BYTE, pixels)
+
+    flipVertically(pixels, w, h)
+    return new ImageData(pixels, w, h)
+  }
+
+  async function toBlob(): Promise<Blob> {
+    return canvas!.convertToBlob({ type: 'image/png' })
```
pythongosssss: I found using PNG here caused massive performance impacts encoding vs using a high quality ~92-95 jpg

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -0,0 +1,395 @@
...
+    const texture = gl.createTexture()
+    if (!texture) return
+
+    gl.activeTexture(gl.TEXTURE0 + index)
+    gl.bindTexture(gl.TEXTURE_2D, texture)
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE)
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE)
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR)
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR)
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, image)
```
pythongosssss, line 298: I think to match backend behavior, you want to add a 
`gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, true)` before calling texImage2D to flip the image automatically, then you can remove the manual flip in readPixels

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -0,0 +1,395 @@
...
+      const loc = uniformLocations.get(`u_image${i}`)
+      if (loc != null) {
+        gl.activeTexture(gl.TEXTURE0 + i)
+        gl.bindTexture(gl.TEXTURE_2D, inputTextures[i] ?? createEmptyTexture())
+        gl.uniform1i(loc, i)
+      }
+    }
+
+    const prevPassUnit = MAX_INPUTS
+    const prevPassLoc = uniformLocations.get('u_prevPass')
```
pythongosssss, line 323: u_prevPass is different behavior to the backend, where each iteration overwrites the previous input.
One thing to note is that the backend only does this for a single input, we should probably update it to pass through all so you can have the initial texture, plus some multistep operation texture, then combine in last step

src/renderer/glsl/useGLSLRenderer.ts
pythongosssss: specifically in Firefox & Safari, Chrome wasn't too bad

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -0,0 +1,395 @@
...
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE)
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE)
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.LINEAR)
+    gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.LINEAR)
+    gl.texImage2D(gl.TEXTURE_2D, 0, gl.RGBA, gl.RGBA, gl.UNSIGNED_BYTE, image)
+
+    inputTextures[index] = texture
+  }
+
+  function render(): void {
```
pythongosssss, line 303: Are you going to be rendering full size images here, or constrained? Running the operation on a 4k+ image will may heavy impact on the browser responsiveness even when debounced

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -0,0 +1,395 @@
...
+void main() {
+    vec2 verts[3] = vec2[](vec2(-1, -1), vec2(3, -1), vec2(-1, 3));
+    v_texCoord = verts[gl_VertexID] * 0.5 + 0.5;
+    gl_Position = vec4(verts[gl_VertexID], 0, 1);
+}
+`
+
+const MAX_INPUTS = 5
+const MAX_FLOAT_UNIFORMS = 5
+const MAX_INT_UNIFORMS = 5
```
pythongosssss: I think you should be able to extract these max uniforms from the node definition so it doesn't get out of sync if we update the backend node

src/renderer/glsl/useGLSLRenderer.ts
pythongosssss: The current implementation still makes it incompatible with the current backend that is expecting u_image0 to be the updated input

src/renderer/glsl/useGLSLRenderer.ts
christian-byrne: Constrained — PR3 caps the preview at 1024px on the longest dimension via `MAX_PREVIEW_DIMENSION` in `useGLSLPreview.ts`, so even 4K+ inputs get scaled down before rendering.

src/renderer/glsl/useGLSLRenderer.ts
christian-byrne: Acknowledged — the current frontend uses ping-pong FBOs for `u_prevPass` rather than overwriting `u_image0` in-place like the backend. I've added a code comment documenting this divergence (line 314-315). Once the backend is updated to pass through all inputs for multi-step operations, we can align both sides. For now the preview-only path works correctly with the ping-pong approach.

jtydhr88 (approved):
LGTM

PR #9211, opened 2026-02-25
fix: move active jobs button into actionbar

benceruleanlu:
## Summary

Move the top menu `N active` queue button into `ComfyActionbar` so it stays attached to the actionbar when docked, dragged, or floating.

## Changes

- Moved the `queue-overlay-toggle` button UI from `TopMenuSection.vue` into `ComfyActionbar.vue`
- Moved queue button behavior and context menu handling (`toggle`, right-click clear queue, active count badge/label) into `ComfyActionbar.vue`
- Removed now-unused queue button state/handlers/imports from `TopMenuSection.vue`

image

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9211-fix-move-active-jobs-button-into-actionbar-3126d73d365081ceb553c172db479e3b) by [Unito](https://www.unito.io)


comfydesigner:
Nice, thanks for  reminding me that the locked subscribe button should be updated at some point in the future too

src/components/actionbar/ComfyActionbar.vue
```diff
@@ -42,6 +42,38 @@
         >
           <i class="icon-[lucide--x] size-4" />
         </Button>
+        <Button
```
christian-byrne, line 45: issue (if-minor): `type="destructive"` is not a valid prop on the custom `Button` component - it only accepts `variant`. This means the button silently falls back to the default `secondary` variant instead of rendering with destructive styling. This is a pre-existing bug carried over from `TopMenuSection.vue`, but since you're touching this code anyway, a one-word fix would be nice:

```suggestion
          variant="destructive"
```

benceruleanlu: If it's not a valid prop, shouldn't we just explicitly move it to secondary instead? Unsure why a button to open job history sidebar tab or QPO is supposed to be destructive

benceruleanlu: Yeah it makes the button red, I just made it secondary explicitly

src/components/actionbar/ComfyActionbar.vue
```diff
@@ -318,11 +360,52 @@ watch(isDragging, (dragging) => {
 const cancelJobTooltipConfig = computed(() =>
   buildTooltipConfig(t('menu.interrupt'))
 )
+const queueHistoryTooltipConfig = computed(() =>
+  buildTooltipConfig(t('sideToolbar.queueProgressOverlay.viewJobHistory'))
+)
```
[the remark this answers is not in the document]
christian-byrne: @coderabbitai make an Issue to fix this in a followup.

christian-byrne (approved):
LGTM leaving the tooltip issue noted above for a followup.

PR #9213, opened 2026-02-25
feat: show missing node packs in Errors Tab with install support

jaeone94:
## Summary

Surfaces missing node pack information in the Errors Tab, grouped by registry pack, with one-click install support via ComfyUI Manager.

## Changes

- **What**: Errors Tab now groups missing nodes by their registry pack and shows a `MissingPackGroupRow` with pack name, node/pack counts, and an Install button that triggers Manager installation. A `MissingNodeCard` shows individual unresolvable nodes that have no associated pack. `useErrorGroups` was extended to resolve missing node types to their registry packs using the `/api/workflow/missing_nodes` endpoint. `executionErrorStore` was refactored to track missing node types separately from execution errors and expose them reactively.
- **Breaking**: None

## Review Focus

- `useErrorGroups.ts` — the new `resolveMissingNodePacks` logic fetches pack metadata and maps node types to pack IDs; edge cases around partial resolution (some nodes have a pack, some don't) produce both `MissingPackGroupRow` and `MissingNodeCard` entries
- `executionErrorStore.ts` — the store now separates `missingNodeTypes` state from `errors`; the deferred-warnings path in `app.ts` now calls `setMissingNodeTypes` so the Errors Tab is populated even when a workflow loads without executing

https://github.com/user-attachments/assets/97f8d009-0cac-4739-8740-fd3333b5a85b

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9213-feat-show-missing-node-packs-in-Errors-Tab-with-install-support-3126d73d36508197bc4bf8ebfd2125c8) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/rightSidePanel/errors/TabErrors.vue
+++ b/src/components/rightSidePanel/errors/TabErrors.vue
@@ -27,6 +27,7 @@
           :key="group.title"
           :collapse="collapseState[group.title] ?? false"
           class="border-b border-interface-stroke"
+          :size="group.type === 'missing_node' ? 'lg' : 'default'"
           @update:collapse="collapseState[group.title] = $event"
         >
           <template #label>
@@ -39,17 +40,46 @@
                   {{ group.title }}
                 </span>
                 <span
-                  v-if="group.cards.length > 1"
+                  v-if="group.type !== 'missing_node' && group.cards.length > 1"
                   class="text-destructive-background-hover"
                 >
                   ({{ group.cards.length }})
                 </span>
               </span>
+              <Button
+                v-if="
+                  group.type === 'missing_node' &&
+                  missingNodePacks.length > 0 &&
+                  shouldShowInstallButton
+                "
+                variant="secondary"
+                size="sm"
+                class="shrink-0 mr-2 h-8 rounded-lg text-sm"
+                :disabled="isInstallingAll"
+                @click.stop="installAll"
+              >
+                <DotSpinner v-if="isInstallingAll" duration="1s" :size="12" />
+                {{
+                  isInstallingAll
+                    ? t('rightSidePanel.missingNodePacks.installing')
+                    : t('rightSidePanel.missingNodePacks.installAll')
+                }}
+              </Button>
             </div>
           </template>
 
-          <!-- Cards in Group (default slot) -->
-          <div class="px-4 space-y-3">
+          <!-- Missing Node Packs -->
+          <MissingNodeCard
+            v-if="group.type === 'missing_node'"
+            :show-info-button="shouldShowManagerButtons"
+            :show-node-id-badge="showNodeIdBadge"
+            :missing-pack-groups="missingPackGroups"
+            @locate-node="handleLocateMissingNode"
+            @open-manager-info="handleOpenManagerInfo"
+          />
+
+          <!-- Execution Errors -->
+          <div v-else class="px-4 space-y-3">
             <ErrorNodeCard
               v-for="card in group.cards"
               :key="card.id"
@@ -108,12 +138,18 @@ import { useExternalLink } from '@/composables/useExternalLink'
 import { useSettingStore } from '@/platform/settings/settingStore'
 import { useTelemetry } from '@/platform/telemetry'
 import { useRightSidePanelStore } from '@/stores/workspace/rightSidePanelStore'
+import { useManagerState } from '@/workbench/extensions/manager/composables/useManagerState'
+import { ManagerTab } from '@/workbench/extensions/manager/types/comfyManagerTypes'
 import { NodeBadgeMode } from '@/types/nodeSource'
 
 import PropertiesAccordionItem from '../layout/PropertiesAccordionItem.vue'
 import FormSearchInput from '@/renderer/extensions/vueNodes/widgets/components/form/FormSearchInput.vue'
 import ErrorNodeCard from './ErrorNodeCard.vue'
+import MissingNodeCard from './MissingNodeCard.vue'
 import Button from '@/components/ui/button/Button.vue'
+import DotSpinner from '@/components/common/DotSpinner.vue'
+import { usePackInstall } from '@/workbench/extensions/manager/composables/nodePack/usePackInstall'
+import { useMissingNodes } from '@/workbench/extensions/manager/composables/nodePack/useMissingNodes'
 import { useErrorGroups } from './useErrorGroups'
 
 const { t } = useI18n()
@@ -122,6 +158,11 @@ const { focusNode, enterSubgraph } = useFocusNode()
 const { staticUrls } = useExternalLink()
 const settingStore = useSettingStore()
 const rightSidePanelStore = useRightSidePanelStore()
+const { shouldShowManagerButtons, shouldShowInstallButton, openManager } =
+  useManagerState()
+const { missingNodePacks } = useMissingNodes()
+const { isInstalling: isInstallingAll, installAllPacks: installAll } =
+  usePackInstall(() => missingNodePacks.value)
 
 const searchQuery = ref('')
 
@@ -136,7 +177,9 @@ const {
   filteredGroups,
   collapseState,
   isSingleNodeSelected,
-  errorNodeCache
+  errorNodeCache,
+  missingNodeCache,
+  missingPackGroups
 } = useErrorGroups(searchQuery, t)
 
 /**
@@ -165,6 +208,19 @@ watch(
 
 function handleLocateNode(nodeId: string) {
   focusNode(nodeId, errorNodeCache.value)
+}
+
+function handleLocateMissingNode(nodeId: string) {
+  focusNode(nodeId, missingNodeCache.value)
+}
+
+function handleOpenManagerInfo(packId: string) {
+  const isKnownToRegistry = missingNodePacks.value.some((p) => p.id === packId)
+  if (isKnownToRegistry) {
+    openManager({ initialTab: ManagerTab.Missing, initialPackId: packId })
+  } else {
+    openManager({ initialTab: ManagerTab.All, initialPackId: packId })
+  }
 }
 
 function handleEnterSubgraph(nodeId: string) {
--- a/src/components/rightSidePanel/errors/TabErrors.vue
+++ b/src/components/rightSidePanel/errors/TabErrors.vue
@@ -37,7 +37,11 @@
                   class="icon-[lucide--octagon-alert] size-4 text-destructive-background-hover shrink-0"
                 />
                 <span class="text-destructive-background-hover truncate">
-                  {{ group.title }}
+                  {{
+                    group.type === 'missing_node'
+                      ? `${group.title} (${missingPackGroups.filter((g) => g.packId !== null).length})`
+                      : group.title
+                  }}
                 </span>
                 <span
                   v-if="group.type !== 'missing_node' && group.cards.length > 1"
--- a/src/components/rightSidePanel/errors/TabErrors.vue
+++ b/src/components/rightSidePanel/errors/TabErrors.vue
@@ -39,12 +39,12 @@
                 <span class="text-destructive-background-hover truncate">
                   {{
                     group.type === 'missing_node'
-                      ? `${group.title} (${missingPackGroups.filter((g) => g.packId !== null).length})`
+                      ? `${group.title} (${missingPackGroups.length})`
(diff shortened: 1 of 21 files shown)


src/components/rightSidePanel/errors/useErrorGroups.ts
```diff
@@ -343,14 +375,134 @@ export function useErrorGroups(
...
+
+      for (const nodeType of toResolve) {
+        const pack = await inferPackFromNodeName.call(nodeType.type)
+        asyncResolvedIds.value = new Map(asyncResolvedIds.value).set(
+          nodeType.type,
+          pack?.id ?? null
+        )
+      }
+    },
+    { immediate: true }
```
christian-byrne, line 435: The sequential `for...of` + `await` resolves node types one at a time, and if `pendingTypes` changes mid-flight (e.g. user loads a new workflow), the old watch callback keeps making API calls since there's no cancellation.

Vue 3.5's `onCleanup` (3rd argument to `watch`) is the canonical pattern for this - it's already used elsewhere in the codebase for AbortController cleanup (e.g. `useConflictDetection`). Combined with `Promise.all` for concurrency:

```ts
watch(
  pendingTypes,
  async (pending, _, onCleanup) => {
    const toResolve = pending.filter(
      (n) => !asyncResolvedIds.value.has(n.type)
    )
    if (!toResolve.length) return

    let cancelled = false
    onCleanup(() => { cancelled = true })

    const updated = new Map(asyncResolvedIds.value)
    for (const n of toResolve) updated.set(n.type, RESOLVING)
    asyncResolvedIds.value = updated

    const results = await Promise.all(
      toResolve.map(async (n) => ({
        type: n.type,
        packId: (await inferPackFromNodeName.call(n.type))?.id ?? null
      }))
    )
    if (cancelled) return

    const final = new Map(asyncResolvedIds.value)
    for (const r of results) final.set(r.type, r.packId)
    asyncResolvedIds.value = final
  },
  { immediate: true }
)
```

src/scripts/app.ts
```diff
@@ -1100,6 +1108,15 @@ export class ComfyApp {
...
       useMissingNodesDialog().show({ missingNodeTypes })
     }
+
+    // For now, we'll make them coexist.
+    // Once the Node Replacement feature is implemented in TabErrors
+    // we'll remove the modal display and direct users to the error tab.
+    const executionErrorStore = useExecutionErrorStore()
+    executionErrorStore.setMissingNodeTypes(missingNodeTypes)
+    if (useSettingStore().get('Comfy.RightSidePanel.ShowErrorsTab')) {
+      executionErrorStore.showErrorOverlay()
```
christian-byrne: This same block (setMissingNodeTypes + conditional showErrorOverlay + the "coexist" comment) also appears in `workflowService.ts:showDeferredWarnings`. Consider extracting it into a shared helper on the store, e.g. `executionErrorStore.surfaceMissingNodes(types)`, so the coexistence logic lives in one place and is easier to remove later.

src/stores/executionErrorStore.ts
```diff
@@ -20,19 +27,60 @@ import {
 } from '@/utils/graphTraversalUtil'
 
 /**
- * Store dedicated to execution error state management.
- *
- * Extracted from executionStore to separate error-related concerns
- * (state, computed properties, graph flag propagation, overlay UI)
- * from execution flow management (progress, queuing, events).
+ * @knipIgnoreUsedByStackedPR
```
christian-byrne: nit: this `@knipIgnoreUsedByStackedPR` looks like it may be left over from an earlier iteration, since `MissingNodesError` is used directly in this PR by the `missingNodesError` ref. Safe to remove if so.

src/components/rightSidePanel/errors/useErrorGroups.ts
```diff
@@ -343,14 +375,134 @@ export function useErrorGroups(
...
+          const idA = typeof a === 'string' ? '' : String(a.nodeId ?? '')
+          const idB = typeof b === 'string' ? '' : String(b.nodeId ?? '')
+          return idA.localeCompare(idB, undefined, { numeric: true })
+        }),
+        isResolving
+      }))
+  })
+
+  /** Builds an ErrorGroup from missingNodesError. Returns [] when none present. */
+  function buildMissingNodeGroups(): ErrorGroup[] {
```
christian-byrne, line 499: Curious about the design here: `buildMissingNodeGroups()` creates a placeholder `ErrorCardData` with `id: '__missing_nodes__'`, but `TabErrors.vue` renders `MissingNodeCard` for this group type, which ignores `group.cards` entirely. Would it make sense to make `cards` optional on `ErrorGroup` (or use a discriminated union) so the dead data doesn't need to exist? Totally fine if you'd rather keep it simple for now and revisit if the type evolves further.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -36,20 +37,53 @@
                   class="icon-[lucide--octagon-alert] size-4 text-destructive-background-hover shrink-0"
                 />
                 <span class="text-destructive-background-hover truncate">
-                  {{ group.title }}
+                  {{
+                    group.type === 'missing_node'
+                      ? `${group.title} (${missingPackGroups.filter((g) => g.packId !== null).length})`
```
christian-byrne: The count here shows `missingPackGroups.filter((g) => g.packId !== null).length`, which counts resolved packs and excludes the "Unknown pack" group. If a user has 5 missing nodes all in the unknown group, they'd see "Missing Node Packs (0)". Is that the intended UX, or should it show the total group count?

src/stores/executionErrorStore.ts
```diff
@@ -57,6 +106,40 @@ export const useExecutionErrorStore = defineStore('executionError', () => {
     lastPromptError.value = null
   }
 
+  function setMissingNodeTypes(types: MissingNodeType[]) {
```
christian-byrne, line 115: This is a nice feature addition! One thought on test coverage: the new store functionality (`setMissingNodeTypes`, `hasMissingNodes`, `isContainerWithMissingNode`, the deduplication logic) and the `useErrorGroups` additions (`buildMissingNodeGroups`, `missingPackGroups`, async pack resolution) don't seem to have test coverage yet. The existing store tests only cover `getNodeErrors` and `slotHasError`.

At minimum, unit tests for `setMissingNodeTypes` (especially the deduplication logic with string vs object entries, nodeId vs type keying) and `missingPackGroups` would give good coverage of the core data flow. Component tests for `MissingNodeCard`/`MissingPackGroupRow` could follow later if you prefer to keep the PR focused.

christian-byrne (commented):
Took a look through the full diff. Overall the feature design is solid - grouping missing nodes by pack, async registry resolution, and install support are all well thought-out. Left a few comments below.

src/components/rightSidePanel/errors/useErrorGroups.ts
jaeone94: I've applied both suggestions: `onCleanup` sets a `cancelled` flag to discard stale results, and `Promise.all` now fires all resolutions concurrently. The `asyncResolvedIds` map is updated in a single assignment after all results arrive, which also removes the extra reactive triggers that the sequential version was causing.

src/scripts/app.ts
jaeone94: Done. Added `surfaceMissingNodes(types: MissingNodeType[])` to the store, which calls `setMissingNodeTypes` and conditionally calls `showErrorOverlay`. Both call sites in `app.ts` and `workflowService.ts` now delegate to it, and the "coexist" comment is placed directly above the modal block to make it clear which part is slated for removal once Node Replacement is implemented.

src/stores/executionErrorStore.ts
jaeone94: Correct — those annotations were temporary guards added while the dependent PR was still open. Now that everything has landed they're no longer needed. Removed all three occurrences.

src/components/rightSidePanel/errors/useErrorGroups.ts
jaeone94: Good point — that placeholder card was dead data. I went with a discriminated union rather than just `cards?` because there are really only two active types right now (`missing_model` is not yet implemented): `execution` always has `cards`, and `missing_node` never does. The union expresses that constraint at the type level, so all the optional-chaining guards and `?? []` fallbacks in the consuming code could be removed in favour of straightforward `group.type === 'execution'` narrowing.

```typescript
export type ErrorGroup =
  | { type: 'execution'; title: string; cards: ErrorCardData[]; priority: number }
  | { type: 'missing_node'; title: string; priority: number }
```

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Not the intended UX — I hadn't considered the all-unknown case. The original filter (`packId !== null`) was meant to count only packs the user could directly act on, but `(0)` is clearly confusing when there is a visible row on screen.

Changed to `missingPackGroups.length` so the count reflects the number of rows actually rendered. The unknown group, when present, now counts as 1.

src/stores/executionErrorStore.ts
jaeone94: Added 11 unit tests for `setMissingNodeTypes` to the existing `executionStore.test.ts`, covering all three deduplication strategies:

- **string entries** — deduplicated by value (group nodes)
- **object entries with `nodeId`** — deduplicated by `nodeId`; distinct `nodeId`s are kept as separate rows even when `type` is the same
- **object entries without `nodeId`** — fall back to `type` as the dedup key

Also covers `hasMissingNodes`, empty-array reset, and the output shape of `missingNodesError`. Component tests for `MissingNodeCard` / `MissingPackGroupRow` can follow in a separate PR to keep this one focused.

@coderabbitai please make a component test plan for `MissingNodeCard` and `MissingPackGroupRow`, covering the key rendering and interaction scenarios (e.g. install button visibility based on manager state, pack resolution loading state, locate-node and open-manager-info event emissions, and the collapse/expand transition of `MissingPackGroupRow`).

src/components/rightSidePanel/errors/useErrorGroups.ts
jaeone94: Update: a follow-up review from @coderabbitai caught two additional edge cases in the implementation above, so it was revised once more:

RESOLVING state leak on cancellation — the original onCleanup only set the cancelled flag, meaning entries written as RESOLVING before the await were never cleaned up if the callback was cancelled mid-flight. Those entries would then permanently block retries since .has() returned true. Fixed by also deleting any RESOLVING markers we set, and changing the filter from .has() to .get() === undefined.

Promise.all → Promise.allSettled — if any single pack lookup threw, Promise.all would reject the entire batch and leave the remaining types stuck in RESOLVING. Replaced with Promise.allSettled so each lookup is handled independently; rejected entries fall back to null (treated as Unknown pack).

src/stores/executionErrorStore.ts
jaeone94: @coderabbitai, could you create an issue for this and assign it to me?

src/components/rightSidePanel/errors/useErrorGroups.ts
```diff
@@ -343,14 +379,130 @@ export function useErrorGroups(
...
+        }))
+      )
+      if (cancelled) return
+
+      const final = new Map(asyncResolvedIds.value)
+      for (const r of results) final.set(r.type, r.packId)
+      asyncResolvedIds.value = final
+    },
+    { immediate: true }
+  )
```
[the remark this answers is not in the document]
jaeone94: addressed 4183a34

src/stores/executionErrorStore.ts
christian-byrne: Thanks for adding the `setMissingNodeTypes` / deduplication tests - those look thorough. A couple of items that would be good to track for a follow-up:

1. **Remaining test coverage**: `missingPackGroups` computed, `buildMissingNodeGroups`, `isContainerWithMissingNode`, and the two new components (`MissingNodeCard`, `MissingPackGroupRow`) are still untested.
2. **Props destructuring style**: The new components use `props.xxx` instead of Vue 3.5 reactive destructuring (`const { group } = defineProps<...>()`) per project conventions.

christian-byrne (approved):
LGTM!

PR #9214, opened 2026-02-25
fix: show inline progress in QPOV2 despite stale overlay flag

benceruleanlu:
## Summary

Fix inline queue progress being hidden in QPOV2 mode when a stale `Comfy.Queue.History.Expanded` setting remains true from legacy queue overlay usage.

## Changes

- Update actionbar inline progress hide condition to respect queue-overlay expansion only when QPOV2 is disabled
- Update top menu inline progress summary hide condition with the same gate
- Keep legacy behavior unchanged for non-QPOV2 queue overlay mode

## Testing

- `pnpm exec eslint src/components/actionbar/ComfyActionbar.vue src/components/TopMenuSection.vue` ✅
- `pnpm typecheck` ✅

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9214-fix-show-inline-progress-in-QPOV2-despite-stale-overlay-flag-3126d73d36508170ac27fbb26826dca9) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/TopMenuSection.vue
+++ b/src/components/TopMenuSection.vue
@@ -127,13 +127,15 @@
         <div
           class="pointer-events-none absolute left-0 right-0 top-full mt-1 flex justify-end pr-1"
         >
-          <QueueInlineProgressSummary :hidden="isQueueOverlayExpanded" />
+          <QueueInlineProgressSummary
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):
LGTM!

PR #9230, opened 2026-02-26
fix: migrate DOM widgets to ComponentWidgetImpl and add element value bridge

### LINKED ISSUE #9194: DOM widget values: architectural mismatch between store ownership and DOM delegation
## DOM widget values: architectural mismatch between store ownership and DOM delegation

PR #9166 hotfixed the immediate bug (DOM widgets showing `undefined` in Vue nodes), but the underlying architectural tension remains. This issue tracks the design work to resolve it properly.

### Context

When #8594 introduced `WidgetValueStore`, it created a centralized reactive store where Vue nodes read widget values. This works perfectly for standard widgets (number, combo, toggle) because their values live directly in `_state.value`. But DOM widgets (textarea/customtext, markdown, audio) store their values in the DOM element itself, accessed through `options.getValue()`/`options.setValue()` — the store never gets the real value.

The root cause is in [`BaseWidget.setNodeId()`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/8c3738fb/src/lib/litegraph/src/widgets/BaseWidget.ts#L137-L145) which spreads `this._state` (where `.value` is `undefined` for DOM widgets) into `registerWidget()`. Meanwhile, the actual value lives behind an [`Object.defineProperty` hack](https://github.com/Comfy-Org/ComfyUI_frontend/blob/8c3738fb/src/scripts/domWidget.ts#L419-L427) on the widget instance that delegates to `options.getValue()`.

```mermaid
flowchart LR
    subgraph "What Vue reads"
        store["widgetValueStore\n.value = undefined ❌"]
    end
    subgraph "What execution reads"
        odp["widget.value\n(Object.defineProperty)\n→ options.getValue()\n→ inputEl.value ✅"]
    end
    vue["NodeWidgets.vue"] --> store
    exec["executionUtil.ts"] --> odp
```

### Current state (hotfix)

#9166 [overrides `setNodeId`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/8c3738fb/src/scripts/domWidget.ts#L155-L166) in `BaseDOMWidgetImpl` to snapshot the DOM-resolved value into the store at registration time. This fixes the bug but the value is a point-in-time copy — the store and DOM element can drift.

### The deeper issue

There are currently **two parallel rendering paths** for the same widget type, and they disagree about who owns the value.

**Path A** — `useStringWidget.ts` [creates a raw `<textarea>`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/8c3738fb/src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts#L20-L40) via `document.createElement` and registers it through `addDOMWidget`. The DOM element owns the value. The widget [manually syncs with the store](https://github.com/Comfy-Org/ComfyUI_frontend/blob/8c3738fb/src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts#L28-L39) in `getValue`/`setValue` (there's even a TODO on line 12 about this).

**Path B** — The widget registry [maps `customtext` → `WidgetTextarea`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/8c3738fb/src/renderer/extensions/vueNodes/widgets/registry/widgetRegistry.ts#L120-L125), a Vue component that uses [`defineModel<string>()`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/8c3738fb/src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue#L58). `NodeWidgets.vue` [binds it with `v-model`](https://github.com/Comfy-Org/ComfyUI_frontend/blob/8c3738fb/src/renderer/extensions/vueNodes/components/NodeWidgets.vue#L59) reading from the store. Here the store owns the value and the DOM is just a view.

So both a raw `<textarea>` AND `WidgetTextarea.vue` exist for the same widget. Path B is already the correct architecture — the store owns the value, the Vue component reads via `modelValue`, user input writes back via `update:modelValue`. The problem is that Path A creates the raw element first and registers `undefined` in the store, which Path B then faithfully renders.

```mermaid
flowchart TD
    subgraph pathA["Path A: Legacy DOM widget"]
        usw["useStringWidget.ts"] -->|"document.createElement('textarea')"| raw["Raw &lt;textarea&gt;\n(owns value)"]
        raw -->|"addDOMWidget()"| odp2["Object.defineProperty hack"]
        odp2 -->|"options.getValue()"| raw
    end
    subgraph pathB["Path B: Vue component"]
        reg["widgetRegistry.ts"] -->|"customtext → textarea"| wt["WidgetTextarea.vue\ndefineModel&lt;string&gt;()"]
        wt -->|"v-model"| store2["widgetValueStore\n(owns value)"]
    end
    nw["NodeWidgets.vue"] -->|"getComponent('customtext')"| wt
    nw -->|"widgetState?.value"| store2
    pathA -.->|"setNodeId registers undefined"| store2
```

### Possible approaches

I've been thinking about a few directions, roughly ordered from least to most aligned with where we want the architecture to go (ECS-style centralized store, consistent with the `widgetValueStore` and `proxy-widget-v2` / #8856 direction).

**1. Delegate value source on store entries (~30 lines, 2 files)**

Add a discriminated union to `WidgetState` that tells the store *how* to resolve a value:

```typescript
type ValueSource<T = unknown> =
  | { type: 'delegate'; get: () => T; set: (v: T) => void }

// in registerWidget:
if (state.valueSource?.type === 'delegate') {
  Object.defineProperty(registered, 'value', {
    get: () => state.valueSource.get(),
    set: (v) => state.valueSource.set(v),
    enumerable: true, configurable: true
  })
}
```

DOM widgets would pass `{ type: 'delegate', get: () => options.getValue(), set: (v) => options.setValue(v) }` at registration. The store resolves at read time instead of snapshotting. This is structurally similar to [Vue's `ComputedRefImpl`](https://github.com/vuejs/core/blob/355d60624a6d3a06330e09a75daf0d572ead35e0/packages/reactivity/src/computed.ts#L131-L142) — both delegate `.value` to a pluggable getter/setter — but without dependency tracking since DOM values aren't reactive sources Vue can track.

Tradeoff: the store becomes a proxy/router for delegate widgets rather than a data owner. It fixes the symptom robustly but the store still doesn't own DOM widget values. Not true ECS.

**2. Migrate DOM widgets to Vue components with `modelValue` (the north star)**

Since `WidgetTextarea.vue` already exists and already uses `defineModel`, the cleanest path is to stop creating raw `<textarea>` elements in `useStringWidget.ts` for the Nodes 2.0 path. Instead, create a `ComponentWidgetImpl` and let the store own the value from the start. The raw element path stays for legacy canvas rendering only.

This eliminates the `getValue`/`setValue` indirection, the `Object.defineProperty` hack, and the manual store sync. The store is the single source of truth, the Vue component reads via `modelValue`, user input writes back. True ECS: entities are just keys, the store owns all data, Vue's reactivity is the system.

The complexity here is feature parity — `useStringWidget.ts` adds wheel event handling, middle-click panning, trackpad gestures, and spellcheck config that `WidgetTextarea.vue` would need to absorb. Also need to handle the legacy canvas path which still needs the raw DOM element.

**3. The approach we explicitly don't want**

For completeness: the v1 design explored a `WidgetValueComponent` interface with `PlainValueComponent` and `DOMValueComponent` classes, factory methods on `BaseWidget`, and `subscribe()`/`dispose()` lifecycle. This is Strategy pattern, not ECS — smart objects with behavior attached to widget entities instead of dumb data in a centralized store. ~200 lines across 6 new files for something the delegate approach does in 30 lines and the north star eliminates entirely.

### My recommendation

Ship approach 1 (delegate) as a follow-up to the hotfix — it's small, handles edge cases the snapshot misses, and is independently valuable. Then pursue approach 2 (migrate to Vue components) per-widget-type, starting with textarea/customtext since `WidgetTextarea.vue` already exists. The delegate type can be removed once all our DOM widgets have Vue component equivalents. Extension-created DOM widgets continue to use `addDOMWidget` → `WidgetDOM.vue` unchanged.

This aligns with the direction established by `widgetValueStore` (#8594) and the subgraph proxy-widget-v2 work (#8856) — centralizing widget state in queryable stores rather than scattering it across widget instances and DOM elements.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9194-DOM-widget-values-architectural-mismatch-between-store-ownership-and-DOM-delegation-3126d73d365081b69555d7d033b7dfd7) by [Unito](https://www.unito.io)

christian-byrne:
## Summary

Migrate internal DOM widgets (textarea, markdown) from raw DOM element creation via `addDOMWidget` to `ComponentWidgetImpl`, and add an element value bridge for extension compatibility.

## Changes

- **What**:
  - New `useDomValueBridge` composable — a Vue `customRef` bridged to a DOM element `.value` via `Object.defineProperty` interception + `input` event listener. Wired into `addDOMWidget` so extension widgets that write `widget.inputEl.value = "x"` (bypassing `widget.value`) sync to the store.
  - `useStringWidget` rewritten to use `ComponentWidgetImpl` + `WidgetTextarea.vue` instead of creating an orphaned raw `<textarea>` via `addDOMWidget`. Removes ~92 lines of manual DOM setup, event listeners, and store sync code.
  - `useMarkdownWidget` rewritten to use `ComponentWidgetImpl` + `WidgetMarkdown.vue` instead of Tiptap editor + raw DOM. Removes all Tiptap dependencies from the project.
  - `WidgetTextarea.vue` gains spellcheck setting support and `data-testid`.
- **Dependencies**: Removed 9 Tiptap packages (`@tiptap/core`, `@tiptap/extension-link`, `@tiptap/extension-table`, `@tiptap/extension-table-cell`, `@tiptap/extension-table-header`, `@tiptap/extension-table-row`, `@tiptap/pm`, `@tiptap/starter-kit`, `tiptap-markdown`)
- **No feature loss from Tiptap removal**: The old `useMarkdownWidget.ts` created a Tiptap editor with `editable: false` — it was only used as a markdown-to-HTML renderer, not as a WYSIWYG editor. Editing was done via a raw `<textarea>` appended alongside it. `WidgetMarkdown.vue` (which predates this PR) already replaces that same pattern using `marked` with GFM support for rendering + a `<textarea>` for editing. Table rendering is preserved via `marked`'s `gfm: true` option, and link rendering is handled by the custom `marked` renderer in `markdownRendererUtil.ts`.

## Review Focus

- The two-strategy architecture: **Store OWNS** values for internal widgets (ComponentWidgetImpl — pure Vue reactivity). **Store DELEGATES** values for extension widgets (DOMWidgetImpl — best-effort sync via the element value bridge).
- `useDomValueBridge` intercepts both programmatic writes (`element.value = "x"`) and user input events. It chains through any existing `Object.defineProperty` on the element (e.g. from extensions).
- Widget `type` preserved: `customtext` for textarea, `MARKDOWN` for markdown — ensures registry mappings and extension compatibility.
- Audio widget migration (`AUDIO_UI`) intentionally deferred — the `<audio>` element is the API surface (`.src`, `.classList`, playback methods), not just a rendering detail.

Fixes #9194

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9230-fix-migrate-DOM-widgets-to-ComponentWidgetImpl-and-add-element-value-bridge-3136d73d365081f09f3dc0dc159c10f1) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
```diff
@@ -1,123 +1,32 @@
...
-    if (canScrollY) {
-      event.stopPropagation()
-      return
-    }
-
-    // If textarea can't scroll vertically, pass to canvas
-    event.preventDefault()
-    app.canvas.processMouseWheel(event)
-  })
+  addWidget(node, widget as BaseDOMWidget<object | string>)
```
DrJKL: Is it not really a `BaseDOMWidget<object | string>`?

src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
```diff
@@ -1,123 +1,32 @@
...
 import { isStringInputSpec } from '@/schemas/nodeDef/nodeDefSchemaV2'
 import type { InputSpec } from '@/schemas/nodeDef/nodeDefSchemaV2'
-import { app } from '@/scripts/app'
+import { ComponentWidgetImpl, addWidget } from '@/scripts/domWidget'
+import type { BaseDOMWidget } from '@/scripts/domWidget'
 import type { ComfyWidgetConstructorV2 } from '@/scripts/widgets'
-import { useWidgetValueStore } from '@/stores/widgetValueStore'
 
-const TRACKPAD_DETECTION_THRESHOLD = 50
+import WidgetTextarea from '../components/WidgetTextarea.vue'
```
DrJKL: I never finished prohibiting ancestor segments in imports. 🪦

src/scripts/domWidget.ts
```diff
@@ -412,5 +413,24 @@ LGraphNode.prototype.addDOMWidget = function <
...
   })
 
+  if (
+    element instanceof HTMLInputElement ||
+    element instanceof HTMLTextAreaElement
+  ) {
+    const bridgeScope = effectScope()
+    bridgeScope.run(() => {
+      const bridgedValue = useDomValueBridge(element)
+      watch(bridgedValue, (newVal) => {
```
DrJKL: Conditional watchers can be risky, especially in contexts like this where they're both conditional and created through a call that can be outside of a Vue lifecycle scope.

DrJKL: Oh, nevermind, Just looked a few lines up 🤦🏻

DrJKL: Still a little worried, but when am I not?

src/scripts/domWidget.ts
```diff
@@ -412,5 +413,24 @@ LGraphNode.prototype.addDOMWidget = function <
...
+        if (widget.value !== newVal) {
+          widget.value = newVal as V
+        }
+      })
+    })
+
+    widget.onRemove = useChainCallback(widget.onRemove, () => {
+      bridgeScope.stop()
+    })
```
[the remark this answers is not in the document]
DrJKL: Trust the rabbit on this one. The subgraph graph callback behavior is maddening.

src/composables/element/useDomValueBridge.ts
```diff
@@ -0,0 +1,70 @@
...
+    set(v: string) {
+      prevSet.call(element, v)
+      notifyChange?.()
+    }
+  })
+
+  function onInput() {
+    notifyChange?.()
+  }
+  element.addEventListener('input', onInput)
```
DrJKL: How about useEventListener?

src/composables/element/useDomValueBridge.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  onScopeDispose(() => {
+    element.removeEventListener('input', onInput)
+    if (existingDescriptor) {
+      Object.defineProperty(element, 'value', existingDescriptor)
+    } else {
+      delete (element as unknown as Record<string, unknown>).value
```
DrJKL: 👀

src/composables/element/useDomValueBridge.ts
DrJKL, line 1: We should come up with a clean API for extension authors so we can add a proper deprecation warning here. How to wire up your custom widget to the widgetValueStore properly.

src/composables/element/useDomValueBridge.ts
christian-byrne: Switched to `useEventListener` from `@vueuse/core` in c060787 — handles cleanup automatically via scope disposal.

src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
christian-byrne: Yep, that is what it is — line 28 already has `widget as BaseDOMWidget<object | string>`. This was pre-existing from the original code.

src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
christian-byrne: Acknowledged — not something to address in this PR but would be a nice lint rule to have.

src/composables/element/useDomValueBridge.ts
christian-byrne: Yeah the cast is unfortunate but necessary — `delete` only works on index-signature types, and `HTMLTextAreaElement` does not have one. This restores the element to its prototype descriptor.

src/composables/element/useDomValueBridge.ts
christian-byrne: Agreed, that deserves its own effort. This PR keeps the existing escape hatch working but a proper public API for extension widget authors would be the right long-term fix.

@coderabbitai Make an Issue to design a clean public API for extension authors to wire custom widgets to the widgetValueStore, replacing the current value-descriptor pattern with a supported integration point.

christian-byrne: Valid follow-up. Filed #11889 to design the public extension API for value-store integration and add a proper deprecation warning here once we have a migration target. Resolving.

christian-byrne:
@DrJKL back to you — addressed both review comments:

- 60a005ce: switched to absolute `@/` import for `WidgetTextarea` (re #discussion_r2857074955)
- dd1c1ae9: widened `addWidget` generic to `<V extends object | string>(node, widget: BaseDOMWidget<V>)` so the `BaseDOMWidget<object | string>` cast at the call site is gone (re #discussion_r2857071676)

Filed #11889 as the follow-up for designing a clean extension API + deprecation warning for raw `inputEl.value` writes (the `useDomValueBridge.ts` thread).

Heads up: PR is currently CONFLICTING with main. Conflicts span binary snapshot images under `browser_tests/`, `package.json` / `pnpm-lock.yaml` / `pnpm-workspace.yaml`, and source files including `useStringWidget.ts`, `useMarkdownWidget.ts`, `WidgetTextarea.vue`, and `WidgetTextarea.test.ts`. Out of scope for this review-comment pass — flagging so it gets handled separately before merge.

src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
```diff
@@ -1,154 +1,47 @@
...
-      const widgetState = widgetStore.getWidget(graphId, node.id, name)
-
-      return (widgetState?.value as string) ?? inputEl.value
-    },
-    setValue(v: string) {
-      inputEl.value = v
-      const graphId = resolveNodeRootGraphId(node, app.rootGraph.id)
-      const widgetState = widgetStore.getWidget(graphId, node.id, name)
-      if (widgetState) widgetState.value = v
+  const widget = new ComponentWidgetImpl({
```
DrJKL: 🔴 **Critical — the `inputEl` shim has flown the coop**

> The shim that once mapped `inputEl` to `element`,
> A whisper of warning, polite and decent —
> Is gone, and with it, `defineDeprecatedProperty` too,
> Forty-plus repos may break before they know what to do.
>
> If `widget.inputEl.value = x` lives in the wild,
> This silent removal will hurt the downstream child.
> Restore the alias, or in the description make clear
> A migration is coming, brace your ecosystem, dear.

src/scripts/domWidget.ts
```diff
@@ -411,5 +413,37 @@ LGraphNode.prototype.addDOMWidget = function <
     }
   })
 
+  if (
```
DrJKL: 🔴 **Critical — the bridge installs on _every_ input element**

> For every input or textarea passed through this door,
> `Object.defineProperty` rewrites what was there before.
> Extensions that hand us elements with their own clever bind
> May find their reactive form models silently behind.
>
> Consider opting in — gate this on widget intent,
> Not a global rewrite of every DOM event.

src/scripts/domWidget.ts
```diff
@@ -411,5 +413,37 @@ LGraphNode.prototype.addDOMWidget = function <
...
+    }
+
+    function stopBridge() {
+      bridgeScope?.stop()
+      bridgeScope = null
+    }
+
+    startBridge()
+
+    this.onAdded = useChainCallback(this.onAdded, startBridge)
```
DrJKL: 🔴 **Critical — start at construction _and_ chain onto `onAdded`?**

> `startBridge()` runs immediately — fine, all is well,
> Then chains onto `onAdded` to ring the bell.
> A guard prevents the double-mount on the very first round,
> But across copy/paste, subgraph collapse, deeper bound,
>
> The closures in `prev*` may capture descriptors stale,
> And `onScopeDispose` restores the wrong veil.
> Capture inside the scope, or drop the chain at start —
> The lifecycle's tangled; pick one role apart.

src/scripts/domWidget.ts
```diff
@@ -411,5 +413,37 @@ LGraphNode.prototype.addDOMWidget = function <
...
+
+    const valueElement = element as HTMLInputElement | HTMLTextAreaElement
+
+    function startBridge() {
+      if (bridgeScope?.active) return
+      bridgeScope = effectScope()
+      bridgeScope.run(() => {
+        const bridgedValue = useDomValueBridge(valueElement)
+        watch(bridgedValue, (newVal) => {
+          if (widget.value !== newVal) {
```
DrJKL: 🟡 **Major — identity check on `V extends object | string` is suspect**

> `widget.value !== newVal` — for strings, this is fair,
> But `V` admits objects, where identity won't compare.
> Here the element guarantees a textual cast,
> So narrow the branch to `string`, and let the `as V` pass.

```suggestion
        watch(bridgedValue, (newVal) => {
          if (widget.value !== newVal) {
            widget.value = newVal
          }
```

src/composables/element/useDomValueBridge.ts
```diff
@@ -0,0 +1,78 @@
...
+      set(v: string) {
+        element.value = v
+        trigger()
+      }
+    }))
+  }
+
+  let notifyChange: (() => void) | undefined
+
+  const ref = customRef<string>((track, trigger) => {
```
DrJKL: 🟡 **Major — `customRef` `track()` does not actually track DOM mutations**

> Notifications fire from the setter we installed in place,
> And the `input` event, when typing leaves its trace.
> But `setRangeText`, `setAttribute('value', …)`, or IME's quiet end —
> None will trigger; the ref ignores them, my friend.
>
> If `canOverride` is false, no notify fires at all,
> Yet callers still believe the ref is reactive after the call.
> Document the boundary, or warn when the bridge can't bind —
> Silent staleness is the worst bug to find.

src/composables/element/useDomValueBridge.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export function useDomValueBridge(element: ValueElement): Ref<string> {
+  const proto = Object.getPrototypeOf(element)
+  const nativeDescriptor = Object.getOwnPropertyDescriptor(proto, 'value')
+  const existingDescriptor = Object.getOwnPropertyDescriptor(element, 'value')
+
+  const prevGet = existingDescriptor?.get ?? nativeDescriptor?.get
+  const prevSet = existingDescriptor?.set ?? nativeDescriptor?.set
+
+  if (!prevGet || !prevSet) {
```
DrJKL: 🟡 **Major — fallback branch never wires `input` listener**

> The early-return ref forgets to subscribe to `input`,
> So typing into the element will not begin to spin it.
> For real `<input>`/`<textarea>`, this branch is dead air —
> The prototype always has a `value` descriptor there.
>
> Remove it, or wire the listener so the fallback's not in vain.
> Dead code today is tomorrow's debugging pain.

src/composables/element/useDomValueBridge.test.ts
```diff
@@ -0,0 +1,184 @@
+import { effectScope, watch } from 'vue'
```
DrJKL: 🟡 **Major — deleted tests have no equivalent replacement**

> Two test files vanished: markdown and string,
> Coverage of callbacks, pointer-events, ctrl+wheel, the works.
> This new test covers the bridge in isolation, true —
> But middle-click forwarding? `onRemove` idempotency? Adieu.
>
> Either prove the new components carry the load,
> Or restore the regression tests down the road.
> Repo guidance is clear: don't delete to make green —
> Behavioral coverage must remain seen.

src/extensions/core/previewAny.ts
```diff
@@ -58,15 +58,16 @@ useExtensionService().registerExtension({
         showValueWidget.options.hidden = true
         showValueWidget.options.read_only = true
         showValueWidget.options.serialize = false
-        showValueWidget.element.readOnly = true
+        if (showValueWidget.element) showValueWidget.element.readOnly = true
```
DrJKL: 🟡 **Major — defensive `if (…element)` masks a type regression**

> Before this change, `element` was required, never undefined,
> Now `if (…element)` guards it — what changed in the bind?
> If the migration to `ComponentWidgetImpl` made it optional too,
> Narrow once, assign once, or warn loudly on null — don't just skew.
>
> Silent guards on what should never be missing
> Are how regressions hide and bugs go fishing.

src/renderer/extensions/vueNodes/widgets/composables/useMarkdownWidget.ts
```diff
@@ -1,138 +1,47 @@
...
 import type { InputSpec } from '@/schemas/nodeDef/nodeDefSchemaV2'
 import { app } from '@/scripts/app'
+import { ComponentWidgetImpl, addWidget } from '@/scripts/domWidget'
+import type { BaseDOMWidget } from '@/scripts/domWidget'
 import type { ComfyWidgetConstructorV2 } from '@/scripts/widgets'
 import { useWidgetValueStore } from '@/stores/widgetValueStore'
 
-// TODO: This widget manually syncs with widgetValueStore via getValue/setValue.
-// Consolidate with useStringWidget into shared helpers (domWidgetHelpers.ts).
+import WidgetMarkdown from '../components/WidgetMarkdown.vue'
```
DrJKL: 🟢 **Minor — verify `WidgetMarkdown.vue` has no tiptap lingering**

> Tiptap is dropped from `package.json` and the catalog true,
> But `WidgetMarkdown.vue` isn't in this review.
> Confirm that component imports no tiptap left behind,
> Else the build will break at runtime, leaving users blind.

src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
```diff
@@ -1,154 +1,47 @@
...
+    node,
+    name,
+    component: WidgetTextarea,
+    inputSpec,
+    type: 'customtext',
+    props: { placeholder: opts.placeholder ?? name },
+    options: {
+      minNodeSize: [400, 200],
+      getValue: () =>
+        (useWidgetValueStore().getWidget(
```
DrJKL: 🟢 **Minor — hoist `useWidgetValueStore()` once, not twice per call**

> Each `getValue`/`setValue` resolves the Pinia store anew,
> The prior code hoisted it — that habit was true.
> Same correctness, cheaper, clearer to read —
> Lift it out of the closure, plant a single seed.

```suggestion
  const widgetStore = useWidgetValueStore()
  const widget = new ComponentWidgetImpl({
    node,
    name,
    component: WidgetTextarea,
    inputSpec,
    type: 'customtext',
    props: { placeholder: opts.placeholder ?? name },
    options: {
      minNodeSize: [400, 200],
      getValue: () =>
        (widgetStore.getWidget(
```

src/scripts/domWidget.ts
```diff
@@ -354,9 +356,9 @@ export class ComponentWidgetImpl<
   }
 }
 
-export const addWidget = <W extends BaseDOMWidget<object | string>>(
+export const addWidget = <V extends object | string>(
```
DrJKL, line 335: 🟢 **Minor — `addWidget` generic widened, but call sites still cast**

> The relaxation from `<W extends BaseDOMWidget…>` to `<V…>` is fine,
> Yet `useMarkdownWidget` still casts `as BaseDOMWidget<object | string>` down the line.
> The cast is a smell — tighten via overload to fit each shape,
> So callers don't reach for the assertion escape.

src/scripts/domWidget.ts
```diff
@@ -411,5 +413,37 @@ LGraphNode.prototype.addDOMWidget = function <
     }
   })
 
+  if (
+    element instanceof HTMLInputElement ||
+    element instanceof HTMLTextAreaElement
+  ) {
+    let bridgeScope: ReturnType<typeof effectScope> | null = null
+
+    const valueElement = element as HTMLInputElement | HTMLTextAreaElement
```
DrJKL: 🟢 **Minor — redundant cast; `instanceof` already narrowed**

> The `instanceof` checks above narrow the union clean,
> So the `as HTMLInputElement | HTMLTextAreaElement` is unseen needed.
> Drop the assertion, let TypeScript's narrowing speak —
> Less noise, same safety, code that's more sleek.

```suggestion
    const valueElement = element
```

src/utils/widgetPropFilter.ts
```diff
@@ -18,7 +18,9 @@ export const INPUT_EXCLUDED_PROPS = [
   ...STANDARD_EXCLUDED_PROPS,
   'inputClass',
   'inputStyle',
-  'read_only'
+  'read_only',
+  'getValue',
```
DrJKL: 🟢 **Minor — exclude list grows, but no test guards it**

> Adding `getValue`/`setValue` is right — they're functions that would crash
> When `v-bind`'s validator gives them a sash.
> But no test asserts the list strips them away —
> One small assertion stops the regression's play.

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -104,6 +107,11 @@ function handleContextMenu(e: MouseEvent) {
   e.preventDefault()
 }
 
+const settingStore = useSettingStore()
+const spellcheck = computed(() =>
```
DrJKL: 🟢 **Minor — `spellcheck` recomputed on every render**

> The old code grabbed the setting once when the widget was made,
> Now it's a `computed` — fine if the store's reactive trade.
> Verify `'Comfy.TextareaWidget.Spellcheck'` truly emits on change,
> Or this `computed` is overhead without anything to arrange.

DrJKL (changes requested):
Multi-angle review — delivered in verse, as requested.

Seventeen findings across critical, major, and minor categories.
See inline rhymes below.

src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
christian-byrne: Current head makes the compatibility boundary explicit: internal textarea widgets now use ComponentWidgetImpl, while extension-owned raw elements remain DOMWidget instances. The PR description documents that split, and #11889 tracks the extension API/deprecation follow-up for raw inputEl writes.

src/scripts/domWidget.ts
christian-byrne: The reviewed global bridge was removed during the main reconciliation. Current src/scripts/domWidget.ts contains no useDomValueBridge path and addWidget only registers lifecycle callbacks, so it no longer rewrites every input descriptor.

src/scripts/domWidget.ts
christian-byrne: The bridge start/onAdded/onScopeDispose lifecycle was removed during reconciliation. Current addWidget has one registration callback and no effect scope or descriptor restoration closure.

src/scripts/domWidget.ts
christian-byrne: The reviewed bridge watcher no longer exists on the current head. Internal text widgets now use ComponentWidgetImpl<string>, so the object-identity comparison path is gone.

src/composables/element/useDomValueBridge.ts
christian-byrne: useDomValueBridge was removed during the main reconciliation, including the customRef contract discussed here. Current internal widgets use component v-model state; extension DOM widgets keep their existing element-owned contract.

src/composables/element/useDomValueBridge.ts
christian-byrne: The fallback branch was removed with useDomValueBridge during reconciliation, so there is no longer a non-reactive fallback claiming to provide this behavior.

src/composables/element/useDomValueBridge.test.ts
christian-byrne: Behavioral coverage is present on the current head: useMarkdownWidget.test.ts, useStringWidget.test.ts, WidgetTextarea.test.ts, and forwardMiddleButtonToCanvas.test.ts all remain. The focused widget suite is being rerun on Node 25 for this pass.

src/extensions/core/previewAny.ts
christian-byrne: The defensive previewAny element guard is no longer part of the PR diff after reconciliation with main. Current previewAny behavior matches main, so the type regression discussed here is absent.

src/renderer/extensions/vueNodes/widgets/composables/useMarkdownWidget.ts
christian-byrne: Verified the current WidgetMarkdown path has no Tiptap imports, and the Tiptap packages are removed from package.json and the workspace catalog. The component uses the existing markdown renderer path.

src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
christian-byrne: The current useStringWidget no longer performs direct widget-store reads in getValue/setValue closures. ComponentWidgetImpl owns the value, so the repeated store resolution discussed here is gone.

src/scripts/domWidget.ts
christian-byrne: Addressed on the current head: addWidget is generic over V and both useMarkdownWidget and useStringWidget pass ComponentWidgetImpl<string> directly, with no BaseDOMWidget assertion at either call site.

src/scripts/domWidget.ts
christian-byrne: The reviewed valueElement assertion was removed with the bridge integration during reconciliation. There is no equivalent cast on the current head.

src/utils/widgetPropFilter.ts
christian-byrne: The getValue/setValue exclusion-list change is no longer part of the PR diff after reconciliation with main, so this regression surface is absent from the current branch.

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
christian-byrne: Verified settingStore.get reads the reactive settingValues ref, and WidgetTextarea wraps that read in computed, so setting changes invalidate the computed value. WidgetTextarea.test.ts also covers the rendered spellcheck binding.

src/composables/element/useDomValueBridge.test.ts
christian-byrne: Correction: the Node 25 focused rerun completed green: 3 test files, 26 tests. `pnpm typecheck` also passed.

christian-byrne:
@coderabbitai review

christian-byrne:
Rebased onto current `main` (post FE #16577) to clear stale red CI. A linear `git rebase origin/main` hit unresolvable per-commit conflicts across `package.json`, `pnpm-lock.yaml`, and the DOM-widget composables (20 commits ahead of merge-base), so used `git merge origin/main` instead, which resolved cleanly with no conflict markers — preserves the intended diff in a single merge commit rather than fighting a 20-commit replay.

<details>
<summary>Evidence</summary>

- `git merge origin/main --no-edit` — clean, zero conflict markers (verified via grep across all changed files)
- `pnpm typecheck` — clean
- `pnpm exec vitest run` on all touched non-e2e test files (`WidgetTextarea.test.ts`, `useMarkdownWidget.test.ts`, `useStringWidget.test.ts`, `domWidget.test.ts`, `glslPreviewStoreContract.test.ts`) — 31/31 passed
- Pushed `de81aec2ab7973c21a8ac055590422c45eb39f76` (fast-forward from `6205f32c78`)

</details>

src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
```diff
@@ -1,67 +1,32 @@
...
-        widgetState.value = v
-        return
-      }
-      widgetStore.registerWidget(id, {
-        type: 'customtext',
-        value: v,
-        options: widget.options ?? {}
-      })
-    }
+  const widget = new ComponentWidgetImpl<string>({
```
DrJKL, line 18: This still removes the `widget.inputEl` compatibility alias restored on main by #9759, including the warning added by #9808. Since #13851 already fixed the persistence issue, please close this old PR and re-propose any remaining migration against current main.

DrJKL (commented):
Verdict: superseded/recommend close

Current main has superseded this fix and compatibility work. PR #13851 fixed custom-text DOM-widget value persistence, #9759 restored `widget.inputEl` compatibility, #9808 added its deprecation warning, and #12617 replaced much of the ownership architecture this branch targeted. Merging this branch would now remove compatibility deliberately restored on main.

Of 19 prior substantive findings, 16 are addressed, two are partially addressed, and the `inputEl` compatibility finding remains open. The remaining Tiptap/component migration should be proposed separately against current architecture.

Checks in an isolated install: typecheck, format, Knip, and 27 targeted tests passed; scoped ESLint had no errors. Maintainability finding: the new string-widget tests mainly assert construction and store plumbing, while current main already owns the persistence behavior.

src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
christian-byrne: Agreed. Recommending close. Not closing it myself, that is your call.

One correction to the premise, which makes the case stronger rather than weaker: this branch is not stale-from-February. Head `de81aec` is a merge of main dated 2026-09-02, so the diff is against a September main that already contains #9759, #9808 and #13851. The alias removal is a deliberate deletion of current main's code, not a rebase artifact.

Inventory of all 13 files against `origin/main` @ `39c0078`:

**Would regress, do not land**

- `useStringWidget.ts` deletes `defineDeprecatedProperty(widget, 'inputEl', 'element', ...)` (#9759 + #9808) and the `getValue`/`setValue` bridge into `widgetValueStore` (#13851). Both are current behavior on main.
- `package.json` / `pnpm-workspace.yaml` / `pnpm-lock.yaml` drop all 8 `@tiptap/*` plus `tiptap-markdown`. As of #16151, merged to main today, tiptap has a second consumer: `src/workbench/extensions/agent/components/agent/composer/inlinePrompt.ts` and `InlinePromptEditor.vue`. Rebasing this onto current main breaks the build.
- `multilineTextarea.ts` un-exports `createMultilineInputElement` and `bindMultilineTextareaWidget`. Only safe because `useStringWidget` stopped calling them. On main they are the live path.
- `useStringWidget.test.ts` replaces 4 behavioral tests with 1 smoke test, dropping store-write, callback-on-input, middle-click / ctrl+wheel canvas forwarding, and listener teardown coverage. `useMarkdownWidget.test.ts` loses 109 lines the same way.
- `domWidgetTestUtils.ts` removes `createMockDOMWidgetNode`, which the #13851 contract test uses.

**Already superseded, or no longer in this diff at all**

- `useDomValueBridge.ts` and its test: gone, deleted by `6205f32` (the Aug 26 ECS reconcile). The "element value bridge" in the PR title no longer exists in this PR.
- `src/utils/widgetPropFilter.ts` `getValue`/`setValue` exclusions: also gone from the diff, and not on main.
- `src/extensions/core/previewAny.ts`: no longer touched, and main's `previewAny.ts` no longer references `element` at all.
- The persistence bug the PR was opened for: fixed by #13851.

So 5 of the 13 outdated threads (the bridge itself, `Object.defineProperty` on every element, `customRef` not tracking `setRangeText`/IME, the dead fallback branch, the prop-filter exclusion) describe code that exists in neither this branch nor main. Nothing to carry forward from them.

**Still novel and worth re-proposing**

One hunk, roughly 8 lines: `WidgetTextarea.vue` plus its test, adding `comfy-multiline-input` and `:spellcheck` to the Vue-node textarea. That is a real gap on main today, independent of any migration:

- `GraphCanvas.vue:373-385` re-applies `Comfy.TextareaWidget.Spellcheck` via `document.querySelectorAll('textarea.comfy-multiline-input')`
- `multilineTextarea.ts:21,25` sets the class and the initial spellcheck, but only on the legacy DOM textarea
- `WidgetTextarea.vue` on main has neither, so the setting is a no-op for Vue-node textareas, and `browser_tests/tests/domWidget.spec.ts` and `nodeDisplay.spec.ts` match nothing there

Follow-up PR against current main: add `comfy-multiline-input` to the textarea class list, bind `:spellcheck` from `settingStore.get('Comfy.TextareaWidget.Spellcheck')`, plus the one test from this branch. Nothing else.

Two optional smaller carries, neither worth a PR alone:

- `addWidget` generic widened from `<W extends BaseDOMWidget<object | string>>` to `<V extends object | string>(node, widget: BaseDOMWidget<V>)`, which removes a cast at call sites
- `glslPreviewStoreContract.test.ts` rewritten to use a real `LGraph`/`LGraphNode` instead of `createMockDOMWidgetNode`

**One old thread is still true of this branch**: the deleted-tests one. It is not true of main.

The `ComponentWidgetImpl` migration itself, if still wanted, should be a fresh PR that keeps `inputEl`, keeps the store bridge, and leaves tiptap alone, which means it shares essentially nothing with this diff. Reviving this branch costs more than starting over. Also relevant for whoever does: this branch's history carries `Amp-Thread-ID:` trailers and `Co-authored-by: Amp`, which the repo convention bans, so fresh commits are needed regardless.

Reassigning to you for the close.

PR #9239, opened 2026-02-26
feat: support accumulate toggle for SaveImage/PreviewImage nodes

christian-byrne:
Read the `accumulate` widget value in the `executed` event handler and pass `merge: true` to the output store when enabled.

Depends on backend PR: https://github.com/Comfy-Org/ComfyUI/pull/12647

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9239-feat-support-accumulate-toggle-for-SaveImage-PreviewImage-nodes-3136d73d365081c0af9ac612956fd4d8) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -717,11 +717,15 @@ export class ComfyApp {
       const nodeOutputStore = useNodeOutputStore()
       const executionId = String(detail.display_node || detail.node)
 
+      const node = getNodeByExecutionId(this.rootGraph, executionId)
+      const accumulate = node?.widgets?.find(
+        (w) => w.name === 'accumulate'
+      )?.value
+
       nodeOutputStore.setNodeOutputsByExecutionId(executionId, detail.output, {
-        merge: detail.merge
+        merge: detail.merge || !!accumulate
       })
 
-      const node = getNodeByExecutionId(this.rootGraph, executionId)
       if (node && node.onExecuted) {
         node.onExecuted(detail.output)
       }


src/scripts/app.ts
```diff
@@ -717,11 +717,15 @@ export class ComfyApp {
       const nodeOutputStore = useNodeOutputStore()
       const executionId = String(detail.display_node || detail.node)
 
+      const node = getNodeByExecutionId(this.rootGraph, executionId)
+      const accumulate = node?.widgets?.find(
```
pythongosssss, line 721: note: low chance of this having unintended accumulation for other nodes with an 'accumulate' widget that are also output nodes, but their 'accumulate' has a different meaning

pythongosssss: Since the executed ws message already supports "merge", would there be a different way to achieve this by having the backend nodes set the merge flag instead of checking the widget value? Not sure how hard that would be in reality

pythongosssss: non blocking, just for consideration.

src/scripts/app.ts
pythongosssss, line 1: no new tests for this behavior

src/scripts/app.ts
```diff
@@ -717,11 +717,15 @@ export class ComfyApp {
...
+        (w) => w.name === 'accumulate'
+      )?.value
```
pythongosssss, line 723: I think `useWidgetValueStore` is the prefered way to do this now

christian-byrne:
Thanks for the review! I've addressed all the feedback by moving the merge logic to the backend in the [backend PR](https://github.com/Comfy-Org/ComfyUI/pull/12647):

- **Backend merge flag**: Instead of the frontend reading widget values, `execution.py` now checks `inputs.get('accumulate') is True` and sets `merge: True` in the `executed` websocket message. The frontend's existing `detail.merge` support handles it with no code changes needed.
- **Widget collision concern addressed**: Since we no longer scan for widgets named 'accumulate' on the frontend, there's zero risk of unintended merge behavior on unrelated nodes.
- **Tests added**: Unit tests for the node definitions and merge flag derivation logic are in the backend PR.
- **No `useWidgetValueStore` needed**: Moving to the backend eliminates frontend widget access entirely.

Closing this PR since no frontend changes are required.

christian-byrne:
Following up now that the backend counterpart is being revived as Comfy-Org/ComfyUI#16304 (rebase of #12647, closed by mistake, not a design rejection).

Per this PR's own closing comment, the design moved to backend-driven: the backend sends `merge: true` on the `executed` websocket event, and `nodeOutputStore.ts`'s existing `detail.merge` handling (`merge?: boolean` at src/stores/nodeOutputStore.ts:54, consumed at line 196) already covers it with zero frontend changes needed.

I rebased this branch's follow-up commit onto current main to confirm: after rebase, the branch has a completely empty diff against main (`git diff origin/main HEAD --stat` produces no output). The small widget-scanning fallback this branch once added on top of #9239's approach cancels out exactly against removing it, landing back at what main already has. So there is nothing to open a new frontend PR for — main already supports the feature end-to-end once #16304 lands on the backend. Closing this out as no-op-needed rather than reopening.

PR #9243, opened 2026-02-26
feat: add executionIdToNodeLocatorId call counter behind feature flag

StoneCypher:
## Summary

Add dev-only instrumentation to count how many times executionIdToNodeLocatorId is called per execution run, providing a baseline measurement for evaluating the impact of caching this function's results, in a later PR.

- **What**: Minor, just adds a feature flagged log and two counters
- **Breaking**: Nothing
- **Dependencies**: None

Gated behind the feature flag `ff:expose_executionId_to_node_locator_id_cache_counters`.  Enable in your browser console:

```js
  localStorage.setItem(
    'ff:expose_executionId_to_node_locator_id_cache_counters',
    'true'
  )
```

On execution completion, logs the total call count to console.warn. No reload needed to toggle.

**feat**: Allow measurement of caching impact, behind a flag

**impact**: None unless activated, then just some logging

## Changes

- **What**: 
- **Breaking**: 
- **Dependencies**: 

## Review Focus

Trivial, just set the flag and look at some workflow with subgraphs

## Screenshots

Login problems due to cross-opener policy are preventing me from taking screenshots from a local dev build at this time

## Thread

There isn't one.  I don't have access to AmpCode or Unito.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9243-feat-add-executionIdToNodeLocatorId-call-counter-behind-feature-flag-3136d73d3650814593e3fdc7e5ad1f73) by [Unito](https://www.unito.io)

PR #9244, opened 2026-02-26
Cache execution id to node locator id mappings

StoneCypher:
## Summary

nodeLocationProgressStates runs executionIdToNodeLocatorId for every
ancestor prefix of every node's display_node_id on every WebSocket
progress update. Each call traverses the subgraph hierarchy via
getNodeById. For nested subgraphs with many executing nodes, this
results in O(N × D²) graph lookups per progress tick, where N is
the number of executing nodes and D is the nesting depth.

Add a plain Map cache inside the execution store that memoizes
executionIdToNodeLocatorId results by execution ID string. The
cache persists across computed re-evaluations within a single
execution run, reducing subsequent progress updates to O(1)
lookups per execution ID.

Cache invalidation:
- Cleared at execution start (handleExecutionStart) to ensure
  fresh graph state for each new run
- Cleared at execution end (resetExecutionState) to prevent
  stale entries leaking across runs

The cache stores strings only (no graph node references), with
typical size of ~50-200 entries per run, so memory impact is
negligible.

- **What**: Caches a mapping of ids to ids, preventing an exponential re-scan of subgraphs
- **Breaking**: Nothing
- **Dependencies**: None

You will need to set the feature flag `ff:expose_executionId_to_node_locator_id_cache_counters` from the underlying counter PR if you want to measure the impact.

```js
  localStorage.setItem(
    'ff:expose_executionId_to_node_locator_id_cache_counters',
    'true'
  )
```

## Review Focus

Before pulling this PR, pull https://github.com/Comfy-Org/ComfyUI_frontend/pull/9243 , set the feature flag, and run a workflow with nested subgraphs, then look in console to see a cache measurement.

Next, pull this PR and load the same workflow.  Note the massive reduction in visits.  

## Screenshots

Login problems due to cross-opener policy are preventing me from taking screenshots from a local dev build at this time

## Thread

There isn't one.  I don't have access to AmpCode or Unito.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9244-Cache-execution-id-to-node-locator-id-mappings-3136d73d365081e680eae3c891480ee7) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/executionStore.test.ts
+++ b/src/stores/executionStore.test.ts
@@ -136,6 +136,140 @@ describe('useExecutionStore - NodeLocatorId conversions', () => {
   })
 })
 
+describe('useExecutionStore - nodeLocationProgressStates caching', () => {
+  let store: ReturnType<typeof useExecutionStore>
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+    mockNodeExecutionIdToNodeLocatorId.mockReset()
+    mockNodeIdToNodeLocatorId.mockReset()
+    mockNodeLocatorIdToNodeExecutionId.mockReset()
+
+    setActivePinia(createTestingPinia({ stubActions: false }))
+    store = useExecutionStore()
+  })
+
+  it('should resolve execution IDs to locator IDs for subgraph nodes', () => {
+    const mockSubgraph = {
+      id: 'a1b2c3d4-e5f6-7890-abcd-ef1234567890',
+      nodes: []
+    }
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):
LGTM — clean caching optimization for executionIdToNodeLocatorId lookups. Cache is properly scoped to execution lifecycle (cleared at start and reset). All CI green, 3 new unit tests cover cache behavior.

PR #9249, opened 2026-02-26
fix: cap nodeProgressStatesByJob to prevent unbounded memory growth

StoneCypher:
This has semi-significant performance impact if you use the same workflow for more than a day.  I left Comfy running with a mouse jiggler and this reduced my instance to a crawl after about an hour.

`nodeProgressStatesByJob` accumulated an entry for every job that ever executed during a session. In long-running sessions this grew without bound, since entries were only removed for the active job on `resetExecutionState`.

Add `MAX_PROGRESS_JOBS` (1000) and `evictOldProgressJobs()`, called after each `handleProgressState` update. When the map exceeds the limit, the oldest entries (by ES2015+ insertion order) are pruned — keeping only the most recent 1000. This mirrors the pattern used by assetsStore's `MAX_HISTORY_ITEMS`.

Also adds tests for:
- nodeLocationProgressStates computed reactivity (recomputes on
  wholesale replacement, produces new references)
- Eviction behavior (retains below limit, evicts oldest above limit,
  preserves most recent, no-op when updating existing job)
- API event handler wiring via captured apiEventHandlers map

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9249-fix-cap-nodeProgressStatesByJob-to-prevent-unbounded-memory-growth-3136d73d365081e49b36d8ade0d4dd6e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/executionStore.test.ts
+++ b/src/stores/executionStore.test.ts
@@ -181,7 +181,6 @@ describe('useExecutionStore - nodeLocationProgressStates caching', () => {
     store = useExecutionStore()
   })
 
-
   it('should resolve execution IDs to locator IDs for subgraph nodes', () => {
     const mockSubgraph = {
       id: 'a1b2c3d4-e5f6-7890-abcd-ef1234567890',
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):
LGTM — clean memory leak fix. MAX_PROGRESS_JOBS eviction prevents unbounded growth in long-running sessions. All CI green.

PR #9250, opened 2026-02-26
fix: remove workspace switching confirmation dialog

christian-byrne:
## Summary

Remove the workspace switching confirmation dialog since switching workspaces no longer discards unsaved changes.

## Changes

- **What**: Remove `hasUnsavedChanges` check, `dialogService.confirm` call, and unused imports (`useI18n`, `useWorkflowStore`, `useDialogService`) from `useWorkspaceSwitch`. Rename `switchWithConfirmation` to `switchWorkspace`. Update callers (`WorkspaceSwitcherPopover.vue`, `InviteAcceptedToast.vue`). Remove `workspace.unsavedChanges` i18n entries from all 12 locale files. Simplify tests to cover core switching behavior only.

## Review Focus

The confirmation dialog was showing inaccurate information (warning about discarding unsaved changes when that no longer happens). This is a pure removal with no new behavior.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9250-fix-remove-workspace-switching-confirmation-dialog-3136d73d365081d3b959da22e8f151d1) by [Unito](https://www.unito.io)


src/platform/workspace/components/toasts/InviteAcceptedToast.vue
```diff
@@ -33,10 +33,10 @@ import { useWorkspaceSwitch } from '@/platform/workspace/composables/useWorkspac
 
 const { t } = useI18n()
 const toast = useToast()
-const { switchWithConfirmation } = useWorkspaceSwitch()
+const { switchWorkspace } = useWorkspaceSwitch()
 
 function viewWorkspace(workspaceId: string) {
-  void switchWithConfirmation(workspaceId)
+  void switchWorkspace(workspaceId)
   toast.removeGroup('invite-accepted')
```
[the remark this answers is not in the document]
christian-byrne: Fixed in d5453bf — `viewWorkspace` now awaits `switchWorkspace` and only dismisses the toast on success.

src/platform/workspace/components/toasts/InviteAcceptedToast.vue
```diff
@@ -33,10 +33,12 @@ import { useWorkspaceSwitch } from '@/platform/workspace/composables/useWorkspac
...
-const { switchWithConfirmation } = useWorkspaceSwitch()
+const { switchWorkspace } = useWorkspaceSwitch()
 
-function viewWorkspace(workspaceId: string) {
-  void switchWithConfirmation(workspaceId)
-  toast.removeGroup('invite-accepted')
+async function viewWorkspace(workspaceId: string) {
+  const success = await switchWorkspace(workspaceId)
+  if (success) {
+    toast.removeGroup('invite-accepted')
```
dante01yoon, line 41: nit: When switchWorkspace returns false (i.e. the switch fails), nothing happens — no error feedback, and the toast stays visible. The user clicks "View Workspace", it silently fails, and the toast just sits there. WDYT?

src/platform/workspace/composables/useWorkspaceSwitch.ts
```diff
@@ -1,49 +1,25 @@
...
     if (activeWorkspace.value?.id === workspaceId) {
       return true
     }
 
-    if (hasUnsavedChanges()) {
-      const confirmed = await dialogService.confirm({
-        title: t('workspace.unsavedChanges.title'),
-        message: t('workspace.unsavedChanges.message'),
-        type: 'dirtyClose'
-      })
```
dante01yoon, line 29: Question: Because the confirm/unsaved-changes were removed, the guarantee that “local changes remain safe when switching workspaces” is no longer covered by tests. Is it intended?

dante01yoon (commented):
leave nit-pick and question

benceruleanlu:
Sorry for missing the review window on this, will leave it to @dante01yoon 🙌

src/platform/workspace/components/toasts/InviteAcceptedToast.vue
christian-byrne: good catch - added an error toast when switchWorkspace returns false in 3726d63

src/platform/workspace/composables/useWorkspaceSwitch.ts
christian-byrne: yes, intentional. the confirmation dialog was the only mechanism guarding unsaved changes during workspace switch, and removing it is the point of this PR. the existing tests in useWorkspaceSwitch.test.ts cover the new behavior (direct switch without confirmation). workspace switching is a server-side operation - local workflow state is orthogonal and managed by the workflow change tracker, not the workspace switcher.

PR #9253, opened 2026-02-26
feat: add node replacement UI to Errors Tab

jaeone94:
## Summary

Adds a node replacement UI to the Errors Tab so users can swap missing nodes with compatible alternatives directly from the error panel, without opening a separate dialog.

## Changes

- **What**: New `SwapNodesCard` and `SwapNodeGroupRow` components render swap groups in the Errors Tab; each group shows the missing node type, its instances (with locate buttons), and a Replace button. Added `useMissingNodeScan` composable to scan the graph for missing nodes and populate `executionErrorStore`. Added `removeMissingNodesByType()` to `executionErrorStore` so replaced nodes are pruned from the error list reactively.

### Bug 1: Replaced nodes render as empty shells until page refresh

`replaceWithMapping()` directly mutates `_nodes[idx]`, bypassing the Vue rendering pipeline entirely. Because the replacement node reuses the same ID, `vueNodeData` retains the stale entry from the old placeholder (`hasErrors: true`, empty widgets/inputs). `graph.setDirtyCanvas()` only repaints the LiteGraph canvas and has no effect on Vue.

**Fix**: After `replaceWithMapping()`, manually call `nodeGraph.onNodeAdded?.(newNode)` to trigger `handleNodeAdded` in `useGraphNodeManager`, which runs `extractVueNodeData(newNode)` and updates `vueNodeData` correctly. Also added a guard in `handleNodeAdded` to skip `layoutStore.createNode()` when a layout for the same ID already exists, preventing a duplicate `spatialIndex.insert()`.

### Bug 2: Missing node error list overwritten by incomplete server response

Two compounding issues: (A) the server's `missing_node_type` error only reports the *first* missing node — the old handler parsed this and called `surfaceMissingNodes([singleNode])`, overwriting the full list collected at load time. (B) `queuePrompt()` calls `clearAllErrors()` before the API request; if the subsequent rescan used the stale `has_errors` flag and found nothing, the missing nodes were permanently lost.

**Fix**: Created `useMissingNodeScan.ts` which scans `LiteGraph.registered_node_types` directly (not `has_errors`). The `missing_node_type` catch block in `app.ts` now calls `rescanAndSurfaceMissingNodes(this.rootGraph)` instead of parsing the server's partial response.

## Review Focus

- `handleReplaceNode` removes the group from the store only when `replaceNodesInPlace` returns at least one replaced node — should we always clear, or only on full success?
- `useMissingNodeScan` re-scans on every execution-error change; confirm no performance concerns for large graphs with many subgraphs.

## Screenshots 

https://github.com/user-attachments/assets/78310fc4-0424-4920-b369-cef60a123d50

https://github.com/user-attachments/assets/3d2fd5e1-5e85-4c20-86aa-8bf920e86987

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9253-feat-add-node-replacement-UI-to-Errors-Tab-3136d73d365081718d4ddfd628cb4449) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/rightSidePanel/errors/SwapNodeGroupRow.vue
+++ b/src/components/rightSidePanel/errors/SwapNodeGroupRow.vue
@@ -1,0 +1,148 @@
+<template>
+  <div class="flex flex-col w-full mb-4">
+    <!-- Type header row: type name + chevron -->
+    <div class="flex h-8 items-center w-full">
+      <p
+        class="flex-1 min-w-0 text-sm font-medium overflow-hidden text-ellipsis whitespace-nowrap text-foreground"
+      >
+        {{ `${group.type} (${group.nodeTypes.length})` }}
+      </p>
+
+      <Button
+        variant="textonly"
+        size="icon-sm"
+        :class="
+          cn(
+            'size-8 shrink-0 transition-transform duration-200 hover:bg-transparent',
+            { 'rotate-180': expanded }
+          )
+        "
+        :aria-label="
+          expanded
+            ? t('rightSidePanel.missingNodePacks.collapse', 'Collapse')
+            : t('rightSidePanel.missingNodePacks.expand', 'Expand')
+        "
+        @click="toggleExpand"
+      >
+        <i
+          class="icon-[lucide--chevron-down] size-4 text-muted-foreground group-hover:text-base-foreground"
+        />
+      </Button>
+    </div>
+
+    <!-- Sub-labels: individual node instances, each with their own Locate button -->
+    <TransitionCollapse>
+      <div
+        v-if="expanded"
+        class="flex flex-col gap-0.5 pl-2 mb-2 overflow-hidden"
+      >
+        <div
+          v-for="nodeType in group.nodeTypes"
+          :key="getKey(nodeType)"
+          class="flex h-7 items-center"
+        >
+          <span
+            v-if="
+              showNodeIdBadge &&
+              typeof nodeType !== 'string' &&
+              nodeType.nodeId != null
+            "
+            class="shrink-0 rounded-md bg-secondary-background-selected px-2 py-0.5 text-xs font-mono text-muted-foreground font-bold mr-1"
+          >
+            #{{ nodeType.nodeId }}
+          </span>
+          <p class="flex-1 min-w-0 text-xs text-muted-foreground truncate">
+            {{ getLabel(nodeType) }}
+          </p>
+          <Button
+            v-if="typeof nodeType !== 'string' && nodeType.nodeId != null"
+            variant="textonly"
+            size="icon-sm"
+            class="size-6 text-muted-foreground hover:text-base-foreground shrink-0 mr-1"
+            :aria-label="t('rightSidePanel.locateNode', 'Locate Node')"
+            @click="handleLocateNode(nodeType)"
+          >
+            <i class="icon-[lucide--locate] size-3" />
+          </Button>
+        </div>
+      </div>
+    </TransitionCollapse>
+
+    <!-- Description rows: what it is replaced by -->
+    <div class="flex flex-col text-[13px] mb-2 mt-1 px-1 gap-0.5">
+      <span class="text-muted-foreground">{{
+        t('nodeReplacement.willBeReplacedBy', 'This node will be replaced by:')
+      }}</span>
+      <span class="font-bold text-foreground">{{
+        group.newNodeId ?? t('nodeReplacement.unknownNode', 'Unknown')
+      }}</span>
+    </div>
+
+    <!-- Replace Action Button -->
+    <div class="flex items-start w-full pt-1 pb-1">
+      <Button
+        variant="secondary"
+        size="md"
+        class="flex flex-1 w-full"
+        @click="handleReplaceNode"
+      >
+        <i class="icon-[lucide--repeat] size-4 text-foreground shrink-0 mr-1" />
+        <span class="text-sm text-foreground truncate min-w-0">
+          {{ t('nodeReplacement.replaceNode', 'Replace Node') }}
+        </span>
+      </Button>
+    </div>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { ref } from 'vue'
+import { cn } from '@/utils/tailwindUtil'
+import { useI18n } from 'vue-i18n'
+import Button from '@/components/ui/button/Button.vue'
+import TransitionCollapse from '@/components/rightSidePanel/layout/TransitionCollapse.vue'
+import type { MissingNodeType } from '@/types/comfy'
+import type { SwapNodeGroup } from './useErrorGroups'
+import { useNodeReplacement } from '@/platform/nodeReplacement/useNodeReplacement'
+import { useExecutionErrorStore } from '@/stores/executionErrorStore'
+
+const props = defineProps<{
+  group: SwapNodeGroup
+  showNodeIdBadge: boolean
+}>()
+
+const emit = defineEmits<{
+  'locate-node': [nodeId: string]
+}>()
+
+const { t } = useI18n()
+const { replaceNodesInPlace } = useNodeReplacement()
+const executionErrorStore = useExecutionErrorStore()
+
+const expanded = ref(false)
+
+function toggleExpand() {
+  expanded.value = !expanded.value
+}
+
+function getKey(nodeType: MissingNodeType): string {
+  if (typeof nodeType === 'string') return nodeType
+  return nodeType.nodeId != null ? String(nodeType.nodeId) : nodeType.type
+}
+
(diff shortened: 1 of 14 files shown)


src/components/rightSidePanel/errors/SwapNodeGroupRow.vue
```diff
@@ -0,0 +1,150 @@
...
+    emit('locate-node', String(nodeType.nodeId))
+  }
+}
+
+function handleReplaceNode() {
+  const replaced = replaceNodesInPlace(props.group.nodeTypes)
+  if (replaced.length > 0) {
+    executionErrorStore.removeMissingNodesByType([props.group.type])
+  }
+}
```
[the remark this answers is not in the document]
jaeone94: The Replace button is only rendered inside SwapNodeGroupRow, which is only displayed when swapNodeGroups contains groups with validated replacement mappings from nodeReplacementStore. There's no UI path where the user can click "Replace" for a node type that has no replacement — the button simply won't exist. An empty return from replaceNodesInPlace in this context would only happen if the graph state changed between render and click (e.g., the user manually deleted the node), which is a transient edge case that doesn't warrant a toast.

src/components/rightSidePanel/errors/useErrorGroups.ts
```diff
@@ -495,18 +503,53 @@ export function useErrorGroups(
...
-      {
+    const groups: ErrorGroup[] = []
+
+    if (swapNodeGroups.value.length > 0) {
+      groups.push({
+        type: 'swap_nodes' as const,
+        title: 'Swap Nodes',
+        priority: 0
+      })
+    }
```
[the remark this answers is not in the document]
jaeone94: Addressed in commit https://github.com/Comfy-Org/ComfyUI_frontend/commit/5aa4baf1161b6059f64cdb2df6e4395f8351de4c

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -229,6 +265,15 @@ function handleOpenManagerInfo(packId: string) {
   }
 }
 
+function handleReplaceAll(swapNodeGroups: SwapNodeGroup[]) {
+  const allNodeTypes = swapNodeGroups.flatMap((g) => g.nodeTypes)
+  const replaced = replaceNodesInPlace(allNodeTypes)
```
christian-byrne, line 269: issue: `handleReplaceAll` removes all swap group types from the error list even when `replaceNodesInPlace` only partially succeeds. `replaced` contains only the type strings that were actually swapped, so this should be `executionErrorStore.removeMissingNodesByType(replaced)` instead of building a separate `typesToRemove` from all groups.

src/composables/useMissingNodeScan.ts
```diff
@@ -0,0 +1,44 @@
+import { LiteGraph } from '@/lib/litegraph/src/litegraph'
```
christian-byrne, line 1: nitpick (non-blocking): this file is placed in `src/composables/` with a `use` prefix, but it exports a plain function, not a Vue composable. More importantly, we're moving toward domain-driven folder organization (e.g., `src/workspace/`, `src/platform/`) rather than technical-layer directories like `src/composables/`. Consider placing this closer to its domain, for example under `src/platform/nodeReplacement/` alongside the existing replacement logic it supports.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -229,6 +265,15 @@ function handleOpenManagerInfo(packId: string) {
   }
```
christian-byrne, line 264: nitpick (non-blocking): `handleReplaceAll(swapNodeGroups: SwapNodeGroup[])` takes a parameter that shadows the `swapNodeGroups` from `useErrorGroups`. Since the template always passes the composable's value, the parameter is redundant - the function could reference the outer scope directly, matching how `handleInstallAll` uses `missingNodePacks` without a parameter.

src/components/rightSidePanel/errors/SwapNodeGroupRow.vue
```diff
@@ -0,0 +1,150 @@
+<template>
```
christian-byrne, line 1: note (non-blocking): all three new files in this PR (`SwapNodeGroupRow.vue`, `SwapNodesCard.vue`, `useMissingNodeScan.ts`) are placed in technical-layer directories (`src/components/`, `src/composables/`). We're moving toward domain-driven folder organization where related logic is co-located by feature - e.g., `src/platform/nodeReplacement/` already exists and would be a natural home for these. No action needed in this PR, just flagging the direction for future work.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Good catch! Fixed — `handleReplaceAll` now passes replaced (the actually-swapped types) directly to removeMissingNodesByType instead of building a separate typesToRemove from all groups. Also removed the redundant parameter while at it.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Good point. Removed the parameter — handleReplaceAll() now references the composable's swapNodeGroups.value directly, consistent with how handleInstallAll uses missingNodePacks.

src/composables/useMissingNodeScan.ts
jaeone94: Agreed — this should live closer to its domain. src/platform/nodeReplacement/ makes sense as the target. Will address in the follow-up issue.

src/components/rightSidePanel/errors/SwapNodeGroupRow.vue
jaeone94: Noted — will co-locate these with the existing src/platform/nodeReplacement/ in the follow-up issue alongside useMissingNodeScan.ts.

christian-byrne (approved):
Previous feedback addressed. The partial-replace bug is fixed, parameter shadowing removed, and the new `await nodeReplacementStore.load()` ensures replacement data is available before workflow processing. DDD reorganization tracked in #9254 and #9255. LGTM.

src/components/rightSidePanel/errors/SwapNodeGroupRow.vue
```diff
@@ -0,0 +1,150 @@
...
+<script setup lang="ts">
+import { ref } from 'vue'
+import { cn } from '@/utils/tailwindUtil'
+import { useI18n } from 'vue-i18n'
+import Button from '@/components/ui/button/Button.vue'
+import TransitionCollapse from '@/components/rightSidePanel/layout/TransitionCollapse.vue'
+import type { MissingNodeType } from '@/types/comfy'
+import type { SwapNodeGroup } from './useErrorGroups'
+import { useNodeReplacement } from '@/platform/nodeReplacement/useNodeReplacement'
+import { useExecutionErrorStore } from '@/stores/executionErrorStore'
```
christian-byrne, line 107: suggestion (non-blocking): this component directly imports and calls `useNodeReplacement()` and `useExecutionErrorStore()` to perform replacements. The sibling `MissingPackGroupRow.vue` follows the emit pattern instead - it emits `locateNode`/`openManagerInfo` and lets the parent handle store interactions. Consider emitting a `'replace'` event and handling the replace+remove logic in `TabErrors.vue` (which already has `handleReplaceAll` with the same pattern). This keeps replacement logic centralized and consistent with the component guideline to prefer emit for state changes.

src/stores/executionErrorStore.ts
```diff
@@ -112,6 +112,17 @@ export const useExecutionErrorStore = defineStore('executionError', () => {
...
+  /** Remove specific node types from the missing nodes list (e.g. after replacement). */
+  function removeMissingNodesByType(typesToRemove: string[]) {
+    if (!missingNodesError.value) return
+    const removeSet = new Set(typesToRemove)
+    const remaining = missingNodesError.value.nodeTypes.filter((node) => {
+      const nodeType = typeof node === 'string' ? node : node.type
+      return !removeSet.has(nodeType)
+    })
+    setMissingNodeTypes(remaining)
+  }
```
christian-byrne, line 124: suggestion (non-blocking): the PR adds several new testable functions (`removeMissingNodesByType`, `scanMissingNodes`, the `swapNodeGroups` computed) without corresponding tests. At minimum, `removeMissingNodesByType` (set-based filtering) and the `swapNodeGroups` grouping logic in `useErrorGroups.ts` would benefit from unit tests to catch regressions if the filtering/grouping criteria change.

christian-byrne (approved):

src/stores/executionErrorStore.ts
jaeone94: @coderabbitai retry

PR #9257, opened 2026-02-26
fix(node-search): add API Nodes category to v2 search

Ni-zav:
## Summary
- Add explicit `API Nodes` source category to V2 Node Search sidebar when API nodes are present.
- Add category filtering logic so selecting `API Nodes` shows only nodes with `api_node = true`.
- Add tests for sidebar visibility and API category filtering behavior.

## Why
The bounty list includes a task where API Nodes do not appear in the Node Search Bar. This change provides an explicit and discoverable path in V2 node search.

## Changes
- `src/components/searchbox/v2/NodeSearchCategorySidebar.vue`
  - Add `hasApiNodes` computed.
  - Add `api` category entry (`g.API Nodes`) when API nodes exist.
- `src/components/searchbox/v2/NodeSearchContent.vue`
  - Add `case 'api'` filter branch (`n.api_node`).
- Tests:
  - `src/components/searchbox/v2/NodeSearchCategorySidebar.test.ts`
  - `src/components/searchbox/v2/NodeSearchContent.test.ts`
  - `src/components/searchbox/v2/__test__/testUtils.ts`

## Validation Notes
- Environment had local runtime/tooling constraints during vitest execution (Node/toolchain mismatch and local hook issues).
- Type-level editor diagnostics show no errors in modified files.
- Test cases were added/updated to cover the new behavior.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9257-fix-node-search-add-API-Nodes-category-to-v2-search-3136d73d3650812e92f5ca5f895f5f8e) by [Unito](https://www.unito.io)


pythongosssss:
Hi @Ni-zav the node search is currently going through another round of design iterations, and as such the current implementation in this PR will no longer fit with the updated design where the non-category items are being removed from the sidebar

Ni-zav:
> Hi @Ni-zav the node search is currently going through another round of design iterations, and as such the current implementation in this PR will no longer fit with the updated design where the non-category items are being removed from the sidebar

got it. i am sorry for not fitting with the current development progress

comfydesigner:
@Ni-zav No need to be sorry, your request is still valid - "Partner"/API Nodes category has been added as a quick filter to the list at the top, which should make it easier to find the specific node you're looking for when the filter is enabled

Screenshot 2026-03-23 at 9 38 25 AM

Screenshot 2026-03-23 at 9 38 39 AM

Ni-zav:
> @Ni-zav No need to be sorry, your request is still valid - "Partner"/API Nodes category has been added as a quick filter to the list at the top, which should make it easier to find the specific node you're looking for when the filter is enabled

I see, thanks for the implementation. even though i am late to fit in the PR, this would at least help me in my work!

src/components/searchbox/v2/NodeSearchCategorySidebar.vue
```diff
@@ -79,11 +79,18 @@ const hasEssentialNodes = computed(() =>
...
+  nodeDefStore.visibleNodeDefs.some((n) => n.api_node)
+)
+
 const sourceCategories = computed(() => {
   const categories = []
   if (flags.nodeLibraryEssentialsEnabled && hasEssentialNodes.value) {
     categories.push({ id: 'essentials', label: t('g.essentials') })
   }
+  if (hasApiNodes.value) {
+    categories.push({ id: 'api', label: t('g.API Nodes') })
```
christian-byrne, line 92: issue (blocking): `g.API Nodes` is referenced here but not defined in any of the locale files under `src/locales/*/main.json` (only in the test fixture in `__test__/testUtils.ts`). With `missingWarn` enabled in `src/i18n.ts`, this fires a runtime warning in every locale, and non-English users see the raw fallback string. The `lobe-i18n locale` script generates other locales from `src/locales/en/main.json`, so the entry needs to be added there for translations to be produced.

christian-byrne, line 92: suggestion: all 296 existing `g.*` keys in `src/locales/en/main.json` are camelCase with no spaces. `g.API Nodes` breaks that pattern. Consider renaming to `g.apiNodes` with value `"API Nodes"` and updating both this call site and the matching entry in `__test__/testUtils.ts` (which CodeRabbit already flagged separately).

christian-byrne (commented):
issue (blocking): main already implements this feature under a different name. `RootCategory.PartnerNodes` is wired up in `src/components/searchbox/v2/rootCategories.ts`, `NodeSearchContent.vue` already filters `allNodes.filter((n) => n.api_node)` for it (around lines 228-229), auto-detects partner nodes when any `n.api_node` is present (line 171), and `NodeSearchFilterBar.vue` adds a `t('g.partner')` button gated on `hasPartnerNodes` (lines 135-136).

Could you rebase against main, confirm whether the existing PartnerNodes wiring covers the original use case, and either close this PR or scope it down to whatever delta remains? The current approach (string id `'api'`, key `g.API Nodes`, separate sidebar entry) does not align with how main now models this.

PR #9259, opened 2026-02-26
Support selection of app inputs and outputs from vue mode

AustinMroz:
- The input and output indicators are now plugged directly into the `LGraphNode.vue` template. Care was taken to make implementation to have low cost for performance and complexity when not in app mode setup.
- Context menu event handlers are added to each widget in vue mode instead of resolving the target widget of an event
- Swap the nodeId passed by `useGraphNodeManager` to not include the locator id. This id was never used and was incorrect since it didn't resolve across nested subgraphs.
- Continued bug fixes for app mode as a whole.

Known issue: There is disparity of nodeId between litegraph (which references the widget in the root graph) and vue (which promotes the original widget). Efforts to reconcile are ongoing.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9259-Support-selection-app-inputs-and-outputs-from-vue-mode-3136d73d365081ae8e56e35bf6322409) by [Unito](https://www.unito.io)


AustinMroz:
The single test failure is a known flake test.

While I do not suspect that the changes in this PR would be related to the test, the failure is not expected and suspicious.

src/renderer/extensions/linearMode/LinearControls.vue
```diff
@@ -69,7 +68,11 @@ const mappedSelections = computed(() => {
       ([id]) => id === nodeId
     ).map(([, widgetName]) => widgetName)
     unprocessedInputs = unprocessedInputs.slice(inputGroup.length)
-    const node = app.rootGraph.getNodeById(nodeId)
+    const node =
+      app.rootGraph.getNodeById(nodeId) ??
```
pythongosssss, line 72: Should this use getNodeByLocatorId

src/renderer/extensions/linearMode/AppInput.vue
```diff
@@ -0,0 +1,60 @@
...
+  return id == nodeId && name === widgetName
+}
+function togglePromotion() {
+  if (isPromoted.value) remove(appModeStore.selectedInputs, matchesThis)
+  else appModeStore.selectedInputs.push([id, name])
+}
+</script>
+<template>
+  <div
+    v-if="mode === 'builder:select'"
```
pythongosssss: nit: to be consistent with AppOutput use isSelectMode

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -50,6 +50,10 @@
     @dragleave="handleDragLeave"
     @drop.stop.prevent="handleDrop"
   >
+    <AppOutput
+      v-if="lgraphNode?.constructor?.nodeData?.output_node"
```
pythongosssss: nit: moving the app mode check here would prevent needing to render the component for every node when 99% it wont be used

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -54,21 +53,24 @@
...
-          :node-type="nodeType"
-          :class="
-            cn(
-              'col-span-2',
-              widget.hasError && 'text-node-stroke-error font-bold'
-            )
-          "
-          @update:model-value="widget.updateHandler"
-        />
+        <AppInput :id="widget.id" :name="widget.name">
```
pythongosssss: would also be nice to not have the AppInput component created for every widget

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -54,21 +53,24 @@
...
+            :node-id="nodeData?.id != null ? String(nodeData.id) : ''"
+            :node-type="nodeType"
+            :class="
+              cn(
+                'col-span-2',
+                widget.hasError && 'text-node-stroke-error font-bold'
+              )
+            "
+            @update:model-value="widget.updateHandler"
+            @contextmenu="widget.handleContextMenu"
```
pythongosssss, line 71: Does the context menu do anything in builder mode (e.g. let you rename widgets?) or should it be disabled

src/stores/widgetValueStore.ts
```diff
@@ -15,14 +15,6 @@ import type {
  */
 type WidgetKey = `${NodeId}:${string}`
 
-/**
- * Strips graph/subgraph prefixes from a scoped node ID to get the bare node ID.
- * e.g., "graph1:subgraph2:42" → "42"
- */
-export function stripGraphPrefix(scopedId: NodeId | string): NodeId {
```
pythongosssss, line 22: Slightly worried about the removal of this, but I can't actively see issues with it, do you know why it was done in the first place?

src/renderer/extensions/linearMode/LinearControls.vue
AustinMroz: No, `selectedInputs` stores `NodeIds`, not `LocatorIds`. I'd be nice if I didn't have to crawl the entire graph, but from discussing things with Alex, it's easiest to just eat the infrequent performance cost until the proposed `nodeStore` is implemented.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
AustinMroz: I've been conflicted on this, but currently lean towards agreement. `useAppMode` being a composable results in ~5 unused computeds per instantiation and ~25 per node which I really dislike.

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
AustinMroz: It is a point of immense frustration to me and the primary reason I pushed so hard against putting App mode content into the vue node component. I'd be strongly against bringing the `v-if` outside since it requires duplicating the NodeWidgets description, but I've pulled `useAppMode` and `appModeStore` up a level.

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
AustinMroz: No specific use, so I agree. It's been disabled.

src/stores/widgetValueStore.ts
AustinMroz: I spoke with Alex about this. There was no existing usage of the subgraph prefixing and the subgraph prefix not being recursively resolved meant it wasn't correct and would never be usable anyways.

Per the  law of YAGNI, it was decided that the code should be removed.

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
AustinMroz: Nevermind, linter doesn't let you mutate props, so it's not worth the added complexity compared to operating on the store directly

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -28,7 +28,6 @@
       <div
         v-if="!widget.hidden && (!widget.advanced || showAdvanced)"
         class="lg-node-widget group col-span-full grid grid-cols-subgrid items-stretch"
-        :data-widget-name="widget.name"
```
pythongosssss, line 31: is this removal intentional?

AustinMroz: Yes. This was previously added by me in #8431. Since the introduction of the widgetValueStore, `widget.name` is not actually the name of the widget and can not be used for resolving widgets in litegraph. This became "a non-viable footgun to implement behaviours in a way strongly discouraged by Alex". With the rework to context menus included in this PR, it no longer serves a purpose and is intentionally removed.

pythongosssss (approved):

PR #9260, opened 2026-02-26
Render app builder in arrange mode

pythongosssss:
## Summary

Adds app builder in arrange/preview mode with re-orderable widgets, maintaining size (as much as possible) between the select + preview steps  

## Changes

- **What**: 
- Extract sidebar size constants for sharing between canvas splitter + app mode
- Add widget list using DraggableList and render inert WidgetItems

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9260-Render-app-builder-in-arrange-mode-3136d73d365081ef875acab683d01d9e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/views/LinearView.vue
+++ b/src/views/LinearView.vue
@@ -6,6 +6,7 @@ import { storeToRefs } from 'pinia'
 import { computed, useTemplateRef } from 'vue'
 import { useI18n } from 'vue-i18n'
 
+import AppBuilder from '@/components/builder/AppBuilder.vue'
 import AppModeToolbar from '@/components/appMode/AppModeToolbar.vue'
 import ExtensionSlot from '@/components/common/ExtensionSlot.vue'
 import ModeToggle from '@/components/sidebar/ModeToggle.vue'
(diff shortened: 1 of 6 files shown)


AustinMroz:
Looks like a another doubled implementation with 6acb3a1. Your implementation seems much shorter in terms of line count. There's some style between the preview stage and the post-creation display, but the linear mode implementation will need another pass anyways (since widgets are no longer grouped by nodes) and it may just be easiest to plug the needed functionality in to the right side panel :thinking:

src/views/LinearView.vue
```diff
@@ -157,3 +210,21 @@ const linearWorkflowRef = useTemplateRef('linearWorkflowRef')
     </Splitter>
   </div>
 </template>
+
+<style scoped>
```
AustinMroz, line 214: There's an obligatory complaint about wanting to avoid style blocks, but I'm not sure selectors this complex can be implemented with tailwind.

AustinMroz (approved):
Code is fine and I'm approving as is.

I'll eat the rebase on my end to keep things moving, if there's a change of decision later, the implementations can always be swapped in a future PR

src/views/LinearView.vue
pythongosssss: Yeah, I don't think this is really possible otherwise

PR #9262, opened 2026-02-26
[refactor] Type createNode options parameter

### LINKED ISSUE #9276: Narrow CreateNodeOptions to explicit Pick of creation-time properties
## Context

PR #9262 introduced `CreateNodeOptions` to replace the untyped `Dictionary<unknown>` parameter in `createNode`. The current implementation uses `Partial<Omit<LGraphNode, 'constructor' | 'inputs' | 'outputs'>>` as the base, which exposes the entire `LGraphNode` surface (hundreds of properties and methods like `computeSize`, `onNodeCreated`, `connect`, etc.) as valid creation-time options.

## Proposed Refinement

Narrow `CreateNodeOptions` to a `Pick` of properties that actually make sense as creation-time options (e.g., `pos`, `properties`, `flags`, `size`, `mode`, `title`, `color`, `bgcolor`).

## Key Considerations

This is a deliberate API boundary decision that requires:

1. **Ecosystem audit**: Catalog which properties callers actually pass today. Initial scan of ~50 custom node repos shows almost all pass only `pos`, with one internal caller passing `outputs`.

2. **Unsupported properties**: Decide what happens when someone needs a property not in the `Pick` set. Do they cast? Do you add it? This creates an ongoing maintenance surface.

3. **Migration path**: Consider deprecation/migration path if any external TypeScript consumers were relying on the current permissiveness.

## References

- Original PR: #9262
- Discussion: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9262#discussion_r2862184707
- Requested by: @christian-byrne

This refinement deserves its own focused PR with explicit sign-off on the allowed property set.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9276-Narrow-CreateNodeOptions-to-explicit-Pick-of-creation-time-properties-3146d73d3650810d850ad5de9c493188) by [Unito](https://www.unito.io)

Myestery:
## Summary
Narrow `CreateNodeOptions` from `Partial<Omit<LGraphNode, ...>>` (exposing hundreds of properties/methods) to an explicit interface listing only creation-time properties.

## Changes
- Replace `Partial<Omit<LGraphNode, 'constructor' | 'inputs' | 'outputs'>>` with explicit `CreateNodeOptions` interface containing only: `pos`, `size`, `properties`, `flags`, `mode`, `color`, `bgcolor`, `boxcolor`, `title`, `shape`, `inputs`, `outputs`
- Rename local `CreateNodeOptions` in `createModelNodeFromAsset.ts` to `ModelNodeCreateOptions` to avoid collision

## Ecosystem verification
GitHub code search across ~50 repos confirms only `pos` and `outputs` are used externally. All covered by the narrowed interface.

Fixes #9276
Fixes #4740

### DIFF
--- a/src/lib/litegraph/src/LiteGraphGlobal.ts
+++ b/src/lib/litegraph/src/LiteGraphGlobal.ts
@@ -10,7 +10,13 @@ import { Reroute } from './Reroute'
 import { InputIndicators } from './canvas/InputIndicators'
 import { LabelPosition, SlotDirection, SlotShape, SlotType } from './draw'
 import { Rectangle } from './infrastructure/Rectangle'
-import type { Dictionary, ISlotType, Rect, WhenNullish } from './interfaces'
+import type {
+  CreateNodeOptions,
+  Dictionary,
+  ISlotType,
+  Rect,
+  WhenNullish
+} from './interfaces'
 import { distance, isInsideRectangle, overlapBounding } from './measure'
 import { SubgraphIONodeBase } from './subgraph/SubgraphIONodeBase'
 import { SubgraphSlot } from './subgraph/SubgraphSlotBase'
@@ -525,7 +531,7 @@ export class LiteGraphGlobal {
   createNode(
     type: string,
     title?: string,
-    options?: Dictionary<unknown>
+    options?: CreateNodeOptions
   ): LGraphNode | null {
     const base_class = this.registered_node_types[type]
     if (!base_class) {
@@ -561,10 +567,7 @@ export class LiteGraphGlobal {
 
     // extra options
     if (options) {
-      for (const i in options) {
-        // @ts-expect-error #577 Requires interface
-        node[i] = options[i]
-      }
+      Object.assign(node, options)
     }
 
     // callback
--- a/src/lib/litegraph/src/interfaces.ts
+++ b/src/lib/litegraph/src/interfaces.ts
@@ -3,13 +3,17 @@ import type { CanvasPointerEvent } from '@/lib/litegraph/src/types/events'
 import type { TWidgetValue } from '@/lib/litegraph/src/types/widgets'
 
 import type { ContextMenu } from './ContextMenu'
-import type { LGraphNode, NodeId } from './LGraphNode'
+import type { LGraphNode, NodeId, NodeProperty } from './LGraphNode'
 import type { LLink, LinkId } from './LLink'
 import type { Reroute, RerouteId } from './Reroute'
 import type { SubgraphInput } from './subgraph/SubgraphInput'
 import type { SubgraphInputNode } from './subgraph/SubgraphInputNode'
 import type { SubgraphOutputNode } from './subgraph/SubgraphOutputNode'
-import type { LinkDirection, RenderShape } from './types/globalEnums'
+import type {
+  LGraphEventMode,
+  LinkDirection,
+  RenderShape
+} from './types/globalEnums'
 import type { IBaseWidget } from './types/widgets'
 
 export type Dictionary<T> = { [key: string]: T }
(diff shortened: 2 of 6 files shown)


src/lib/litegraph/src/interfaces.ts
```diff
@@ -373,6 +373,14 @@ export interface INodeOutputSlot extends INodeSlot {
   slot_index?: number
 }
 
+/** Options for {@link LiteGraphGlobal.createNode}. Shallow-copied onto the new node. */
```
christian-byrne, line 380: question: `src/platform/assets/utils/createModelNodeFromAsset.ts` already defines a local `interface CreateNodeOptions { position?: Point }` with a completely different shape, and it imports from the same `@/lib/litegraph/src/litegraph` barrel. Now that this new `CreateNodeOptions` is exported from that barrel, any future import will shadow or conflict with the local one. Worth renaming one of them (e.g. the local one to `ModelNodeCreateOptions`) or acknowledging the collision.

src/lib/litegraph/src/interfaces.ts
```diff
@@ -373,6 +373,14 @@ export interface INodeOutputSlot extends INodeSlot {
...
+export interface CreateNodeOptions extends Partial<
```
christian-byrne: nitpick (non-blocking): `constructor` is not a member of `keyof LGraphNode` (TypeScript class instance types don't include it), so `Omit<LGraphNode, 'constructor' | ...>` is equivalent to `Omit<LGraphNode, 'inputs' | 'outputs'>`. Keeping it is harmless as a guard, just noting it's a no-op today.

src/lib/litegraph/src/LiteGraphGlobal.ts
```diff
@@ -561,10 +567,7 @@ export class LiteGraphGlobal {
 
     // extra options
     if (options) {
-      for (const i in options) {
-        // @ts-expect-error #577 Requires interface
-        node[i] = options[i]
-      }
+      Object.assign(node, options)
```
christian-byrne, line 570: note (non-blocking): the old `for...in` loop traversed the prototype chain, while `Object.assign` only copies own enumerable properties. That's actually safer behavior - just flagging the subtle semantic difference in case any caller relied on inherited properties (unlikely with plain object literals).

src/lib/litegraph/src/interfaces.ts
christian-byrne, line 380: suggestion (non-blocking): the current `Partial<Omit<LGraphNode, ...>>` base is a good pragmatic first step - it replaces the fully untyped `Dictionary<unknown>` without breaking anyone. That said, it exposes the entire `LGraphNode` surface (hundreds of properties and methods like `computeSize`, `onNodeCreated`, `connect`, etc.) as valid creation-time options.

A future refinement could narrow this to a `Pick` of properties that actually make sense as creation-time options (e.g. `pos`, `properties`, `flags`, `size`, `mode`, `title`, `color`, `bgcolor`). But that's a real decision with weight to it - it would be the first time this API explicitly declares "these are the supported creation options" rather than being a pass-through bag. That means:

- An audit of the ecosystem to catalog which properties callers actually pass today (I did a scan across ~50 custom node repos for this review - almost all pass only `pos`, one internal caller passes `outputs`, and that's it).
- A decision about what happens when someone needs a property that's not in the `Pick` set. Do they cast? Do you add it? That's an ongoing maintenance surface.
- A deprecation/migration path if any external TS consumers were relying on the permissiveness.

So this is less of a "quick tidy-up" and more of a deliberate API boundary decision. Worth doing eventually, but deserves its own focused PR with explicit sign-off on the allowed set.

Myestery: Done:  renamed to ModelNodeCreateOptions in the previous commit

src/lib/litegraph/src/interfaces.ts
Myestery: the interface no longer extends Partial<Omit<LGraphNode, ...>>, it's an explicit property list

src/lib/litegraph/src/LiteGraphGlobal.ts
Myestery: Acknowledged: all callers pass plain object literals, so the behavior change is safe. The stricter semantics are actually preferable here

src/lib/litegraph/src/interfaces.ts
Myestery: Addressed in this PR: CreateNodeOptions is now an explicit interface with only creation-time properties. This closes  #9276

christian-byrne (approved):

PR #9263, opened 2026-02-26
[fix] Replace eval() with safe math expression parser

### LINKED ISSUE #8032: Replace eval() with math expression library (e.g., Math.js)
## Description

The current implementation of mathematical expression evaluation in number widgets uses `eval()` to parse and evaluate user input expressions. This should be replaced with a dedicated math expression library for better security and functionality.

## Context

- Current implementation: `evaluateInput()` function in `src/lib/litegraph/src/utils/widget.ts`
- Previous testing was done with [expr-eval](https://github.com/silentmatt/expr-eval)
- [Math.js](https://mathjs.org/docs/reference/functions/evaluate.html) is suggested as a potentially better option

## Related

- PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/7759
- Comment: https://github.com/Comfy-Org/ComfyUI_frontend/pull/7759#discussion_r2687812989
- Requested by: @DrJKL

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-8032-Replace-eval-with-math-expression-library-e-g-Math-js-2e76d73d365081f7b07ac1a93ab15110) by [Unito](https://www.unito.io)

### LINKED ISSUE #9272: Add edge case tests for math parser number formats
## Context

Follow-up to PR #9263 which replaced eval() with a safe math expression parser.

Requested by @christian-byrne in https://github.com/Comfy-Org/ComfyUI_frontend/pull/9263#discussion_r2862184333

## Description

The math parser test suite should be extended to explicitly cover number format edge cases that are common gotchas in math parsers:

- Trailing dot: `123.` (should evaluate to 123)
- Trailing dot in expression: `123.+3` (should evaluate to 126)
- Lone dot: `.` (should return undefined)
- Double dot: `123..` (should return undefined)

These edge cases are covered in expr-eval's test suite and would improve confidence in the parser's robustness.

## Location

`src/lib/litegraph/src/utils/mathParser.test.ts`

## References

- PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9263
- Comment: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9263#discussion_r2862184333

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9272-Add-edge-case-tests-for-math-parser-number-formats-3146d73d3650813789e0e8ed5762e1c0) by [Unito](https://www.unito.io)

### LINKED ISSUE #9273: Add explicit test cases for unary operators after binary operators in mathParser
## Description

Add explicit parameterized test cases for unary operators appearing after binary operators in `src/lib/litegraph/src/utils/mathParser.test.ts`.

This pattern (`2*-3`, `1+-2`, `2--3`, `-2*-3`) is the #1 source of bugs in hand-rolled math parsers and is most likely to regress. While the current parser implementation handles these correctly, only `2*-3` is covered implicitly in the complex expression block.

## Test Cases to Add

```ts
['1+-2', -1],
['2--3', 5],
['-2*-3', 6],
['-(2+3)*-(4+5)', 45],
```

## References

- PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9263
- Comment: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9263#discussion_r2862184337
- Requested by: @christian-byrne

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9273-Add-explicit-test-cases-for-unary-operators-after-binary-operators-in-mathParser-3146d73d3650819286deefd1fd00790f) by [Unito](https://www.unito.io)

### LINKED ISSUE #9274: Normalize negative zero to positive zero in mathParser
## Description

The math expression parser in `src/lib/litegraph/src/utils/mathParser.ts` can produce `-0` from expressions like `-0`, `0 * -1`, or `0 / -1`. While `-0 === 0` is true in JavaScript, the `expr-eval` library explicitly normalizes negative zero to positive zero (`nstack[0] === 0 ? 0 : result`) because `-0` has caused real issues downstream in some contexts.

## Suggested Change

Add normalization at the end of `evaluateMathExpression`:
```typescript
return result === 0 ? 0 : result
```

## Context

- Originated from PR #9263 which replaced `eval()` with the custom math parser
- Comment discussion: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9263#discussion_r2862184330
- Suggestion by @christian-byrne
- Non-blocking improvement for future PR

## References

- PR: #9263
- File: `src/lib/litegraph/src/utils/mathParser.ts`

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9274-Normalize-negative-zero-to-positive-zero-in-mathParser-3146d73d3650818b9b68d9e0b7dad7f6) by [Unito](https://www.unito.io)

Myestery:
## Summary
Replace `eval()` in `evaluateInput()` with a custom recursive descent math parser, eliminating a security concern and enabling the `no-eval` lint rule.

## Changes
- **New**: `mathParser.ts` — recursive descent parser for `+`, `-`, `*`, `/`, `%`, `()`, decimals, unary operators. Zero new dependencies.
- **Modified**: `widget.ts` — replaced `eval()` call with `evaluateMathExpression()`, use `isFinite()` instead of `isNaN()` to reject `Infinity`
- **Modified**: `.oxlintrc.json` — `no-eval` rule changed from `"off"` to `"error"`
- **Tests**: 59 parser tests + 23 integration tests covering complex expressions, edge cases, and invalid input

## Review Feedback Addressed
- Renamed `unit()` → `primary()` for clarity
- Added modulo (`%`) operator support
- Normalized negative zero to positive zero
- Added depth limit (200) for nested parentheses
- Used `isFinite()` instead of `isNaN()` to reject `Infinity`/`-Infinity`
- Added tests for edge-case number formats, unary-after-binary operators, modulo, depth limits, scientific/hex notation, and `Infinity`

Fixes #8032
Fixes #9272
Fixes #9273
Fixes #9274
Fixes #9275

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9263-fix-Replace-eval-with-safe-math-expression-parser-3136d73d3650812f9f8dea21d1ea4f06) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/utils/mathParser.test.ts
+++ b/src/lib/litegraph/src/utils/mathParser.test.ts
@@ -1,0 +1,93 @@
+import { describe, expect, test } from 'vitest'
+
+import { evaluateMathExpression } from '@/lib/litegraph/src/utils/mathParser'
+
+describe('evaluateMathExpression', () => {
+  test.each([
+    ['2+3', 5],
+    ['10-4', 6],
+    ['3*7', 21],
+    ['15/3', 5]
+  ])('basic arithmetic: %s = %d', (input, expected) => {
+    expect(evaluateMathExpression(input)).toBe(expected)
+  })
+
+  test.each([
+    ['2+3*4', 14],
+    ['(2+3)*4', 20],
+    ['10-2*3', 4],
+    ['10/2+3', 8]
+  ])('operator precedence: %s = %d', (input, expected) => {
+    expect(evaluateMathExpression(input)).toBe(expected)
+  })
+
+  test.each([
+    ['3.14*2', 6.28],
+    ['.5+.5', 1],
+    ['1.5+2.5', 4],
+    ['0.1+0.2', 0.1 + 0.2]
+  ])('decimals: %s', (input, expected) => {
+    expect(evaluateMathExpression(input)).toBe(expected)
+  })
+
+  test.each([
+    [' 2 + 3 ', 5],
+    ['  10  -  4  ', 6],
+    [' ( 2 + 3 ) * 4 ', 20]
+  ])('whitespace handling: "%s" = %d', (input, expected) => {
+    expect(evaluateMathExpression(input)).toBe(expected)
+  })
+
+  test.each([
+    ['((2+3))', 5],
+    ['(1+(2*(3+4)))', 15],
+    ['((1+2)*(3+4))', 21]
+  ])('nested parentheses: %s = %d', (input, expected) => {
+    expect(evaluateMathExpression(input)).toBe(expected)
+  })
+
+  test.each([
+    ['-5', -5],
+    ['-(3+2)', -5],
+    ['--5', 5],
+    ['+5', 5],
+    ['-3*2', -6],
+    ['2*-3', -6]
+  ])('unary operators: %s = %d', (input, expected) => {
+    expect(evaluateMathExpression(input)).toBe(expected)
+  })
+
+  test.each([
+    ['2 /2+3 * 4.75- -6', 21.25],
+    ['2 / (2 + 3) * 4.33 - -6', 7.732],
+    ['12* 123/-(-5 + 2)', 492],
+    ['((80 - (19)))', 61],
+    ['(1 - 2) + -(-(-(-4)))', 3],
+    ['1 - -(-(-(-4)))', -3],
+    ['12* 123/(-5 + 2)', -492],
+    ['12 * -123', -1476],
+    ['((2.33 / (2.9+3.5)*4) - -6)', 7.45625],
+    ['123.45*(678.90 / (-2.5+ 11.5)-(80 -19) *33.25) / 20 + 11', -12042.760875],
+    [
+      '(123.45*(678.90 / (-2.5+ 11.5)-(((80 -(19))) *33.25)) / 20) - (123.45*(678.90 / (-2.5+ 11.5)-(((80 -(19))) *33.25)) / 20) + (13 - 2)/ -(-11) ',
+      1
+    ]
+  ])('complex expression: %s', (input, expected) => {
+    expect(evaluateMathExpression(input)).toBeCloseTo(expected as number)
+  })
+
+  test.each(['', 'abc', '2+', '(2+3', '2+3)', '()', '*3', '2 3'])(
+    'invalid input returns undefined: "%s"',
+    (input) => {
+      expect(evaluateMathExpression(input)).toBeUndefined()
+    }
+  )
+
+  test('division by zero returns Infinity', () => {
+    expect(evaluateMathExpression('1/0')).toBe(Infinity)
+  })
+
+  test('0/0 returns NaN', () => {
+    expect(evaluateMathExpression('0/0')).toBeNaN()
+  })
+})
--- a/src/lib/litegraph/src/utils/mathParser.test.ts
+++ b/src/lib/litegraph/src/utils/mathParser.test.ts
@@ -25,7 +25,9 @@ describe('evaluateMathExpression', () => {
     ['3.14*2', 6.28],
     ['.5+.5', 1],
     ['1.5+2.5', 4],
-    ['0.1+0.2', 0.1 + 0.2]
+    ['0.1+0.2', 0.1 + 0.2],
+    ['123.', 123],
+    ['123.+3', 126]
   ])('decimals: %s', (input, expected) => {
     expect(evaluateMathExpression(input)).toBe(expected)
   })
@@ -52,7 +54,11 @@ describe('evaluateMathExpression', () => {
     ['--5', 5],
     ['+5', 5],
     ['-3*2', -6],
-    ['2*-3', -6]
+    ['2*-3', -6],
+    ['1+-2', -1],
+    ['2--3', 5],
+    ['-2*-3', 6],
+    ['-(2+3)*-(4+5)', 45]
   ])('unary operators: %s = %d', (input, expected) => {
     expect(evaluateMathExpression(input)).toBe(expected)
   })
@@ -76,7 +82,7 @@ describe('evaluateMathExpression', () => {
     expect(evaluateMathExpression(input)).toBeCloseTo(expected as number)
   })
 
-  test.each(['', 'abc', '2+', '(2+3', '2+3)', '()', '*3', '2 3'])(
+  test.each(['', 'abc', '2+', '(2+3', '2+3)', '()', '*3', '2 3', '.', '123..'])(
     'invalid input returns undefined: "%s"',
     (input) => {
       expect(evaluateMathExpression(input)).toBeUndefined()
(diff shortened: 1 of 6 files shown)


src/lib/litegraph/src/utils/mathParser.ts
```diff
@@ -0,0 +1,111 @@
...
+
+  function peek(): Token | undefined {
+    return tokens[pos]
+  }
+
+  function consume(): Token {
+    return tokens[pos++]
+  }
+
+  function unit(): number | undefined {
```
christian-byrne: nitpick (non-blocking): the function name `unit` is a bit ambiguous in a math/widget context. in recursive descent parsers the lowest-level production is conventionally called `primary` or `atom` (expr-eval uses `parseAtom`, most textbooks use `primary`).

src/lib/litegraph/src/utils/widget.ts
```diff
@@ -12,14 +14,10 @@ export function getWidgetStep(options: IWidgetOptions<unknown>): number {
...
-      input = eval(input)
-    } catch {
-      // Ignore eval errors
-    }
+  const result = evaluateMathExpression(input)
+  if (result !== undefined) {
+    if (isNaN(result)) return undefined
+    return result
   }
   const newValue = Number(input)
```
christian-byrne, line 22: question: when `evaluateMathExpression` returns `undefined`, the `Number(input)` fallback means things like `"1e5"`, `"0xff"`, and `"Infinity"` are still accepted as valid widget values. the old regex guard would have also let these fall through to `Number()`, so behavior is preserved - just want to confirm this is the intended contract.

this also creates a subtle inconsistency: `"1e5"` evaluates to `100000` (via the fallback), but `"1e5+1"` returns `undefined` (the tokenizer rejects `e`). this matches the old `eval()` behavior since the old regex also excluded `e`, but might surprise users. worth a code comment documenting this?

src/lib/litegraph/src/utils/mathParser.ts
```diff
@@ -0,0 +1,111 @@
...
+      const right = factor()
+      if (right === undefined) return undefined
+      left = op === '+' ? left + right : left - right
+    }
+    return left
+  }
+
+  const result = expr()
+  if (result === undefined || pos !== tokens.length) return undefined
+  return result
```
christian-byrne: suggestion (non-blocking): expr-eval explicitly normalizes negative zero to positive zero at the end of evaluation (`nstack[0] === 0 ? 0 : result`) because `-0` caused real issues downstream. the parser can produce `-0` from expressions like `-0`, `0 * -1`, or `0 / -1`. while `-0 === 0` is true in JS, consider normalizing here: `return result === 0 ? 0 : result`.

src/lib/litegraph/src/utils/mathParser.test.ts
```diff
@@ -0,0 +1,93 @@
...
+  )
+
+  test('division by zero returns Infinity', () => {
+    expect(evaluateMathExpression('1/0')).toBe(Infinity)
+  })
+
+  test('0/0 returns NaN', () => {
+    expect(evaluateMathExpression('0/0')).toBeNaN()
+  })
+})
```
christian-byrne, line 121: nitpick (non-blocking): expr-eval's test suite covers several number format edge cases the parser handles correctly but aren't tested explicitly here: trailing dot (`123.`), trailing dot in expression (`123.+3 = 126`), lone dot (`.` → undefined), and double dot (`123..` → undefined). might be worth adding a few of these since they're common gotchas in math parsers.

src/lib/litegraph/src/utils/widget.ts
```diff
@@ -12,14 +14,10 @@ export function getWidgetStep(options: IWidgetOptions<unknown>): number {
...
-  if (/^[\d\s.()*+/-]+$/.test(input)) {
-    // Solve the equation if possible
-    try {
-      input = eval(input)
-    } catch {
-      // Ignore eval errors
-    }
+  const result = evaluateMathExpression(input)
+  if (result !== undefined) {
+    if (isNaN(result)) return undefined
```
christian-byrne: issue: `evaluateInput('1/0')` returns `Infinity` and `evaluateInput('-1/0')` returns `-Infinity`. the `isNaN` check here doesn't catch these. for a widget numeric value, `Infinity` will likely cause problems downstream (layout calculations, serialization, etc.). consider using `isFinite(result)` instead - it rejects both `NaN` and `±Infinity` in one check:

```ts
if (!isFinite(result)) return undefined
```

the existing test on line 33 (`'division by zero returns Infinity'`) would need updating to expect `undefined` instead.

src/lib/litegraph/src/utils/mathParser.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (!t) return undefined
+
+    if (t.type === 'number') {
+      consume()
+      return t.value
+    }
+
+    if (t.type === 'op' && t.value === '(') {
+      consume()
+      const result = expr()
```
christian-byrne, line 55: suggestion (non-blocking): deeply nested parentheses (e.g. 3000 levels of `(((...)))`) cause a stack overflow since `unit()` → `expr()` → `factor()` → `unary()` → `unit()` recurse for each nesting level. verified this crashes around depth ~3000. for a widget input this is unlikely, but a cheap depth counter (increment on `(`, bail if > ~200) would harden it against pathological input.

src/lib/litegraph/src/utils/mathParser.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    ['+5', 5],
+    ['-3*2', -6],
+    ['2*-3', -6]
+  ])('unary operators: %s = %d', (input, expected) => {
+    expect(evaluateMathExpression(input)).toBe(expected)
+  })
+
+  test.each([
+    ['2 /2+3 * 4.75- -6', 21.25],
+    ['2 / (2 + 3) * 4.33 - -6', 7.732],
```
christian-byrne, line 68: suggestion (non-blocking): the pattern of unary operators appearing after binary operators (`2*-3`, `1+-2`, `2--3`, `-2*-3`) is the #1 source of bugs in hand-rolled math parsers per multiple parsing references - it's the exact case that trips up nearly every tutorial implementation. the parser handles all of these correctly (nice recursive `unary()` design), but only `2*-3` is covered implicitly in the complex expression block. consider adding explicit parameterized tests since these are the cases most likely to regress:

```ts
['1+-2', -1],
['2--3', 5],
['-2*-3', 6],
['-(2+3)*-(4+5)', 45],
```

src/lib/litegraph/src/utils/mathParser.ts
```diff
@@ -0,0 +1,111 @@
+type Token = { type: 'number'; value: number } | { type: 'op'; value: string }
+
+function tokenize(input: string): Token[] | undefined {
+  const tokens: Token[] = []
+  const re = /(\d+(?:\.\d*)?|\.\d+)|([+\-*/()])/g
```
christian-byrne: question: the old `eval()` supported `%` (modulo), which the new parser doesn't. was this intentional? if a user previously typed `10 % 3` in a widget, it would have worked via eval. adding it would be trivial - just include `%` in the regex character class here and handle it alongside `*`/`/` in `factor()`.

Myestery: yeah, adding now

src/lib/litegraph/src/utils/widget.ts
Myestery: okay

src/lib/litegraph/src/utils/widget.ts
Myestery: Yes, this is intentional, it preserves the behavior from the old eval()-based code. The Number() fallback     handles formats like scientific notation (1e5) that the math parser doesn't cover.

src/lib/litegraph/src/utils/widget.ts
christian-byrne: verified - both the math result path and the Number() fallback now use isFinite, and the tests confirm 1/0 and 0/0 return undefined. nice.

src/lib/litegraph/src/utils/mathParser.ts
christian-byrne: confirmed - % is in the tokenizer regex, handled in factor(), and has test coverage. looks good.

src/lib/litegraph/src/utils/widget.ts
christian-byrne: makes sense, and the isFinite check on the Number() fallback now rejects Infinity/NaN which addresses the main concern. the 1e5 vs 1e5+1 inconsistency is inherited from the old code so fine to leave as-is.

christian-byrne (approved):
LGTM

PR #9215, opened 2026-02-25
fix: open target panel when toggling Docked Job History

benceruleanlu:
## Summary
Make the Docked Job History toggle deterministic so it opens the expected UI target in both directions.

## Changes
- Update `JobHistoryActionsMenu` toggle behavior:
  - When currently docked (`Comfy.Queue.QPOV2=true`), disable docked mode and explicitly open floating QPO (`Comfy.Queue.History.Expanded=true`)
  - When currently floating (`Comfy.Queue.QPOV2=false`), enable docked mode and open the `job-history` sidebar tab
- Add/adjust unit tests in `QueueOverlayHeader.test.ts` to verify both toggle directions and target panel behavior

## Testing
- `pnpm exec eslint src/components/queue/JobHistoryActionsMenu.vue src/components/queue/QueueOverlayHeader.test.ts`
- `pnpm typecheck`
- `pnpm test:unit -- src/components/queue/QueueOverlayHeader.test.ts`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9215-fix-open-target-panel-when-toggling-Docked-Job-History-3126d73d3650810eb409ff38e3a521f3) by [Unito](https://www.unito.io)

PR #9265, opened 2026-02-27
fix: make docked job history toggle persistence-safe

benceruleanlu:
## Summary
Follow-up to #9215 to keep Docked Job History toggle behavior deterministic even when settings persistence fails.

## Changes
- Close the actions popover immediately when toggling Docked Job History.
- Use settingStore.setMany(...) when switching from docked to floating mode.
- Set sidebarTabStore.activeSidebarTabId = 'job-history' before persisting when switching from floating to docked mode.
- Wrap persistence calls in try/catch without rollback so locally-applied UI state remains deterministic.
- Expand QueueOverlayHeader tests to cover setMany, popover close behavior, and persistence-failure paths.

## Testing
- pnpm test:unit -- src/components/queue/QueueOverlayHeader.test.ts
- pnpm typecheck
- pnpm exec eslint src/components/queue/JobHistoryActionsMenu.vue src/components/queue/QueueOverlayHeader.test.ts
- pnpm lint (fails in this branch due pre-existing stylelint errors in generated apps/desktop-ui/dist/**/*.css files, unrelated to this change)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9265-fix-make-docked-job-history-toggle-persistence-safe-3146d73d3650818f86c4dfdd57669abd) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/queue/JobHistoryActionsMenu.vue
+++ b/src/components/queue/JobHistoryActionsMenu.vue
@@ -20,7 +20,7 @@
             class="w-full justify-between text-sm font-light"
             variant="textonly"
             size="sm"
-            @click="onToggleDockedJobHistory"
+            @click="onToggleDockedJobHistory(close)"
           >
             <span class="flex items-center gap-2">
               <i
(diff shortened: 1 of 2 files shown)


src/components/queue/JobHistoryActionsMenu.vue
```diff
@@ -100,14 +100,22 @@ const onClearHistoryFromMenu = (close: () => void) => {
...
+const onToggleDockedJobHistory = async (close: () => void) => {
+  close()
+
+  try {
+    if (isQueuePanelV2Enabled.value) {
+      await settingStore.setMany({
+        'Comfy.Queue.QPOV2': false,
+        'Comfy.Queue.History.Expanded': true
+      })
+      return
```
christian-byrne, line 112: nitpick (non-blocking): the catch block silently swallows persistence errors. Since the local UI state is already applied before the await, this is safe for users - but a console.warn or even a Sentry track alert here would make it easier to diagnose persistence issues in production without affecting the UX.

christian-byrne (approved):

PR #9266, opened 2026-02-27
fix: subgraph output slot labels not updating in v2 renderer

artokun:
## Summary

Custom names set on subgraph output nodes are ignored in the v2 renderer — it always shows the data type name (e.g. "texts") instead of the user-defined label. Works correctly in v1.

## Changes

- **What**: Made `outputs` in `extractVueNodeData` reactive via `shallowReactive` + `defineProperty` (matching the existing `inputs` pattern). Added a `node:slot-label:changed` graph trigger that `SubgraphNode` fires when input/output labels are renamed, so the Vue layer picks up the change.

## Review Focus

- The `outputs` reactivity mirrors `inputs` exactly — same `shallowReactive` + setter pattern. The new trigger event forces `shallowReactive` to detect the deep property change by re-assigning the array.
- Also handles input label renames for consistency, even though the current bug report is output-specific.

## Screenshots

**v1 — output correctly shows custom label "output_text":**
Screenshot 2026-02-26 at 4 43 00 PM

**v2 before fix — output shows type name "texts" instead of custom label:**
Screenshot 2026-02-26 at 4 43 30 PM

**v2 after fix — output correctly shows "output_text":**
Screenshot 2026-02-26 at 5 14 44 PM

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9266-fix-subgraph-output-slot-labels-not-updating-in-v2-renderer-3146d73d365081979327fd775a6ef62b) by [Unito](https://www.unito.io)


src/composables/graph/useGraphNodeManager.test.ts
```diff
@@ -240,6 +240,34 @@ describe('Widget slotMetadata reactivity on link disconnect', () => {
...
+
+    // Simulate what SubgraphNode does: set the label, then fire the trigger
+    node.outputs[0].label = 'custom_label'
+    graph.trigger('node:slot-label:changed', { nodeId: node.id })
+
+    await nextTick()
+
+    // Re-read after trigger — the outputs array was replaced
+    const updatedData = vueNodeData.get(String(node.id))
+    expect(updatedData?.outputs?.[0].label).toBe('custom_label')
```
christian-byrne: suggestion: The test verifies raw data after the trigger but doesn't confirm that a Vue `computed` or `watch` depending on `outputs[0].label` actually fires. Since the fix's purpose is making `shallowReactive` detect the deep property change, verifying a watcher fires is the behavioral proof. There's already a strong precedent at lines 147-183 using `computed()` + `watch()` for this pattern.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -310,6 +310,15 @@ export function extractVueNodeData(node: LGraphNode): VueNodeData {
       reactiveInputs.splice(0, reactiveInputs.length, ...v)
     }
   })
```
christian-byrne, line 414: suggestion (non-blocking): The `outputs` defineProperty omits explicit `configurable: true` and `enumerable: true`, unlike the `widgets` pattern at L291-292 which sets both. Currently safe because LGraphNode class field initializers default to writable+configurable+enumerable, but being explicit would be consistent and guard against class definition changes.

src/composables/graph/useGraphNodeManager.test.ts
```diff
@@ -240,6 +240,34 @@ describe('Widget slotMetadata reactivity on link disconnect', () => {
   })
 })
 
+describe('Subgraph output slot label reactivity', () => {
```
christian-byrne, line 246: suggestion (non-blocking): The test only covers output slot labels, but the handler also re-assigns `inputs`. SubgraphNode fires the trigger for both `renaming-input` and `renaming-output`. A parallel test for input slot labels would prevent the inputs branch from silently regressing.

src/lib/litegraph/src/types/graphTriggers.ts
```diff
@@ -23,10 +23,16 @@ interface NodeSlotLinksChangedEvent {
   linkId: number
 }
 
+interface NodeSlotLabelChangedEvent {
```
christian-byrne, line 26: suggestion (non-blocking): `NodeSlotLabelChangedEvent` only carries `nodeId`, so the handler re-assigns both `inputs` and `outputs` arrays even when only one changed. Adding an optional `slotType` field (like `NodeSlotLinksChangedEvent` has) would allow targeted updates and avoid unnecessary reactive churn.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -651,6 +674,9 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
         case 'node:slot-links:changed':
           triggerHandlers['node:slot-links:changed'](event)
           break
+        case 'node:slot-label:changed':
```
christian-byrne, line 792: note (non-blocking): The `onTrigger` switch now has 4 cases each doing identical `triggerHandlers[event.type](event)`. Pre-existing pattern -- no action needed in this PR, just flagging for awareness.

src/composables/graph/useGraphNodeManager.test.ts
```diff
@@ -240,6 +240,34 @@ describe('Widget slotMetadata reactivity on link disconnect', () => {
...
+    expect(nodeData?.outputs?.[0].label).toBeUndefined()
+
+    // Simulate what SubgraphNode does: set the label, then fire the trigger
+    node.outputs[0].label = 'custom_label'
+    graph.trigger('node:slot-label:changed', { nodeId: node.id })
+
+    await nextTick()
+
+    // Re-read after trigger — the outputs array was replaced
+    const updatedData = vueNodeData.get(String(node.id))
```
christian-byrne: nitpick (non-blocking): `updatedData` re-reads from `vueNodeData.get(...)` with a comment "Re-read after trigger", but since `vueNodeData` is `reactive()`, the original `nodeData` reference already reflects updates. Asserting against `nodeData` directly would also verify the reactive Map propagation.

src/composables/graph/useGraphNodeManager.test.ts
```diff
@@ -240,6 +240,34 @@ describe('Widget slotMetadata reactivity on link disconnect', () => {
...
 
+describe('Subgraph output slot label reactivity', () => {
+  beforeEach(() => {
+    setActivePinia(createTestingPinia({ stubActions: false }))
+  })
+
+  it('updates output slot label when node:slot-label:changed is triggered', async () => {
+    const graph = new LGraph()
+    const node = new LGraphNode('test')
+    node.addOutput('original_name', 'STRING')
```
christian-byrne, line 254: nitpick (non-blocking): The test uses a single output. A follow-up case with multiple outputs where only one label changes would verify the other labels remain untouched.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -637,6 +646,20 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
         if (slotLinksEvent.slotType === NodeSlotType.INPUT) {
           refreshNodeSlots(String(slotLinksEvent.nodeId))
         }
+      },
+      'node:slot-label:changed': (slotLabelEvent) => {
+        const nodeId = String(slotLabelEvent.nodeId)
+        const nodeRef = nodeRefs.get(nodeId)
+        if (!nodeRef) return
```
christian-byrne, line 768: nitpick (non-blocking): The `if (!nodeRef) return` guard is untested. A quick "does not throw for unknown nodeId" test would cover that branch.

christian-byrne (dismissed):

artokun (commented):
Pulled/rebased and pushed `d02b4a3`.

Addressed the scoped nits here: explicit descriptors, input coverage, `slotType` on the slot-label trigger, and a no-throw guard test. I left the `onTrigger` switch dedupe note alone for scope.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -219,6 +219,9 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
         if (input._widget) {
           input._widget.label = newName
         }
+        this.graph?.trigger('node:slot-label:changed', {
+          nodeId: this.id
+        })
```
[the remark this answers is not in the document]
artokun: Yep. Tightened this on top of the rebased branch. `node:slot-label:changed` now carries optional `slotType`, `SubgraphNode` emits input vs output, and the handler only refreshes the affected side.

src/composables/graph/useGraphNodeManager.test.ts
artokun: Covered now. Added a second output and assert the untouched one stays untouched.

src/composables/graph/useGraphNodeManager.ts
artokun: Good call. Made the `widgets` / `inputs` / `outputs` descriptors explicit for `configurable` + `enumerable`.

src/composables/graph/useGraphNodeManager.test.ts
artokun: Tried the watcher/computed route first, but `vueNodeData` is a reactive `Map` holding plain entries, so watching `outputs[0].label` directly here does not actually subscribe the way it looks like it should. Kept this at the map re-read level instead.

src/composables/graph/useGraphNodeManager.ts
artokun: Added the unknown-`nodeId` no-throw case.

src/lib/litegraph/src/types/graphTriggers.ts
artokun: Also did this. `node:slot-label:changed` now carries optional `slotType`, and `SubgraphNode` emits `INPUT` vs `OUTPUT` so we only refresh the affected side.

src/composables/graph/useGraphNodeManager.ts
artokun: Yep, saw it. Leaving the switch dedupe alone for this PR.

src/composables/graph/useGraphNodeManager.test.ts
artokun: I checked this while tightening the test. The stable `nodeData` ref does not actually prove the contract I care about here, so I kept the explicit re-read after the trigger.

src/composables/graph/useGraphNodeManager.test.ts
artokun: Yep, added the parallel input-label case.

src/composables/graph/useGraphNodeManager.test.ts
christian-byrne: fair point - the reactive Map semantics here don't actually let a watcher on outputs[0].label fire the way it reads, so the re-read assertion is the right level to test at

christian-byrne (approved):
all feedback addressed cleanly - slotType discriminant, explicit descriptors, and test coverage all look good

artokun:
@DrJKL Backport PRs ready for review:
- core/1.41: #10727
- cloud/1.41: #10728

Both typecheck clean and affected tests pass. Conflicts resolved by keeping target branch versions of `extractVueNodeData` and `_invalidatePromotedViewsCache`.

PR #9269, opened 2026-02-27
fix: node replacement fails after execution and modal sync

jaeone94:
## Summary

Fixes two bugs in the node replacement flow: placeholder detection failing after workflow execution or pack reinstallation, and missing UI sync in the Errors Tab when replacements are applied from the modal dialog.

## Changes

- **Placeholder detection**: Node placeholder detection now matches against `targetTypes` (derived from the replaceable node list built at workflow load time) instead of relying on `has_errors` flag or `registered_node_types` lookup. This ensures replacement works reliably after execution (where `has_errors` gets cleared) and after pack reinstallation (where the type becomes registered).
- **Modal → Errors Tab sync**: Added `executionErrorStore.removeMissingNodesByType()` call in `MissingNodesContent.vue` after replacement, so the Errors Tab reflects changes immediately without requiring a page reload.

## Review Focus

- `collectAllNodes` predicate change in `useNodeReplacement.ts`: now uses `targetTypes.has(originalType)` to find nodes by their original serialized type. This is independent of runtime state like `has_errors` or `registered_node_types`.
- `executionErrorStore.removeMissingNodesByType` call timing in `MissingNodesContent.vue` — runs synchronously after `replaceNodesInPlace` resolves, before auto-close logic.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9269-fix-node-replacement-fails-after-execution-and-modal-sync-3146d73d365081218398c961639b450f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/MissingNodesContent.vue
+++ b/src/components/dialog/content/MissingNodesContent.vue
@@ -234,6 +234,7 @@ import { isCloud } from '@/platform/distribution/types'
 import type { NodeReplacement } from '@/platform/nodeReplacement/types'
 import { useNodeReplacement } from '@/platform/nodeReplacement/useNodeReplacement'
 import { useDialogStore } from '@/stores/dialogStore'
+import { useExecutionErrorStore } from '@/stores/executionErrorStore'
 import type { MissingNodeType } from '@/types/comfy'
 import { cn } from '@/utils/tailwindUtil'
 import { useMissingNodes } from '@/workbench/extensions/manager/composables/nodePack/useMissingNodes'
@@ -245,6 +246,7 @@ const { missingNodeTypes } = defineProps<{
 const { missingCoreNodes } = useMissingNodes()
 const { replaceNodesInPlace } = useNodeReplacement()
 const dialogStore = useDialogStore()
+const executionErrorStore = useExecutionErrorStore()
 
 interface ProcessedNode {
   label: string
@@ -338,6 +340,11 @@ function handleReplaceSelected() {
   }
   replacedTypes.value = nextReplaced
   selectedTypes.value = nextSelected
+
+  // Sync with execution error store so the Errors Tab updates immediately
+  if (result.length > 0) {
+    executionErrorStore.removeMissingNodesByType(result)
+  }
 
   // Auto-close when all replaceable nodes replaced and no non-replaceable remain
   const allReplaced = replaceableNodes.value.every((n) =>
--- a/src/components/dialog/content/MissingNodesContent.vue
+++ b/src/components/dialog/content/MissingNodesContent.vue
@@ -341,7 +341,10 @@ function handleReplaceSelected() {
   replacedTypes.value = nextReplaced
   selectedTypes.value = nextSelected
 
-  // Sync with execution error store so the Errors Tab updates immediately
+  // replaceNodesInPlace() handles canvas rendering via onNodeAdded(),
+  // but the modal only updates its own local UI state above.
+  // Without this call the Errors Tab would still list the replaced nodes
+  // as missing because executionErrorStore is not aware of the replacement.
   if (result.length > 0) {
     executionErrorStore.removeMissingNodesByType(result)
   }
--- a/src/platform/nodeReplacement/useNodeReplacement.ts
+++ b/src/platform/nodeReplacement/useNodeReplacement.ts
@@ -226,16 +226,22 @@ export function useNodeReplacement() {
       useWorkflowStore().activeWorkflow?.changeTracker ?? null
     changeTracker?.beforeChange()
 
+    // Target types come from node_replacements fetched at workflow load time
+    // and the missing nodes detected at that point — not from the current
+    // registered_node_types. This ensures replacement still works even if
+    // the user has since installed the missing node pack.
+    const targetTypes = new Set(
+      selectedTypes.map((t) => (typeof t === 'string' ? t : t.type))
+    )
+
     try {
-      const placeholders = collectAllNodes(
-        graph,
-        (n) =>
-          !!n.last_serialization &&
-          !(
-            (n.last_serialization.type ?? n.type ?? '') in
-            LiteGraph.registered_node_types
-          )
-      )
+      const placeholders = collectAllNodes(graph, (n) => {
+        if (!n.last_serialization) return false
+        // Prefer the original serialized type; fall back to the live type
+        // for nodes whose serialization predates the type field.
+        const originalType = n.last_serialization.type ?? n.type
+        return !!originalType && targetTypes.has(originalType)
+      })
 
       for (const node of placeholders) {
         const match = findMatchingType(node, selectedTypes)
--- a/src/platform/nodeReplacement/useNodeReplacement.ts
+++ b/src/platform/nodeReplacement/useNodeReplacement.ts
@@ -229,7 +229,12 @@ export function useNodeReplacement() {
     try {
       const placeholders = collectAllNodes(
         graph,
-        (n) => !!n.has_errors && !!n.last_serialization
+        (n) =>
+          !!n.last_serialization &&
+          !(
+            (n.last_serialization.type ?? n.type ?? '') in
+            LiteGraph.registered_node_types
+          )
       )
 
       for (const node of placeholders) {


src/platform/nodeReplacement/useNodeReplacement.ts
```diff
@@ -229,7 +229,12 @@ export function useNodeReplacement() {
     try {
       const placeholders = collectAllNodes(
         graph,
-        (n) => !!n.has_errors && !!n.last_serialization
+        (n) =>
+          !!n.last_serialization &&
+          !(
+            (n.last_serialization.type ?? n.type ?? '') in
```
christian-byrne: question: the fallback chain `n.last_serialization.type ?? n.type ?? ''` means if both are undefined/null, this checks `'' in LiteGraph.registered_node_types`. Is that the intended behavior? Also, if `last_serialization` exists but has no `.type` property, it falls through to `n.type`. useMissingNodeScan.ts uses identical logic so this is probably fine, just wanted to confirm the edge cases are covered.

src/platform/nodeReplacement/useNodeReplacement.ts
```diff
@@ -229,7 +229,12 @@ export function useNodeReplacement() {
     try {
       const placeholders = collectAllNodes(
         graph,
-        (n) => !!n.has_errors && !!n.last_serialization
+        (n) =>
```
christian-byrne: suggestion (non-blocking): the predicate is a bit dense with nested ternary and negation. Consider extracting for clarity:
```ts
const nodeType = n.last_serialization?.type ?? n.type ?? ''
return !!n.last_serialization && !(nodeType in LiteGraph.registered_node_types)
```

src/platform/nodeReplacement/useNodeReplacement.ts
```diff
@@ -229,7 +229,12 @@ export function useNodeReplacement() {
     try {
       const placeholders = collectAllNodes(
```
christian-byrne: question: if a node pack is reinstalled after load, nodes of that type will no longer be detected as placeholders since they'll appear in `registered_node_types`. Is this acceptable, or are there scenarios where a "re-registered" node should still be replaced?

src/platform/nodeReplacement/useNodeReplacement.ts
```diff
@@ -229,7 +229,12 @@ export function useNodeReplacement() {
     try {
```
christian-byrne, line 237: suggestion (non-blocking): the detection logic change from `has_errors` to `registered_node_types` check should have test coverage. Consider adding a test that verifies placeholder detection works when `has_errors` is false but node type is unregistered - the exact scenario this PR fixes. The existing test mock also has `registered_node_types: {}` which doesn't validate that registered types are correctly excluded.

src/components/dialog/content/MissingNodesContent.vue
```diff
@@ -339,6 +341,11 @@ function handleReplaceSelected() {
   replacedTypes.value = nextReplaced
   selectedTypes.value = nextSelected
```
christian-byrne, line 343: nitpick (non-blocking): the comment could clarify *why* the sync is needed - e.g., "so the Errors Tab reflects the fix immediately without requiring workflow re-execution."

src/platform/nodeReplacement/useNodeReplacement.ts
jaeone94: Good call. The previous approach used `registered_node_types` to identify missing nodes, which required a fallback chain (`n.last_serialization.type ?? n.type ?? ''`) to avoid `undefined in obj` errors. I've reworked the detection to match against `targetTypes`, which is derived from the replaceable node list built at workflow load time (from the `node_replacements` API response and the missing nodes detected at that point). This eliminates the fallback chain entirely — `!!originalType` guards against `undefined`/`null`/`''` before the `Set.has()` call, and `Set.has(undefined)` safely returns `false` rather than throwing.

 ```typescript
 const placeholders = collectAllNodes(graph, (n) = {
   if (!n.last_serialization) return false
   const originalType = n.last_serialization.type ?? n.type
   return !!originalType && targetTypes.has(originalType)
 })
 ```

src/platform/nodeReplacement/useNodeReplacement.ts
jaeone94: Applied — extracted `originalType` as a named variable and added comments explaining why we prefer `last_serialization.type` over `n.type`. The entire predicate was also restructured as part of switching from `registered_node_types` lookup to `targetTypes` matching.

 ```typescript
 const placeholders = collectAllNodes(graph, (n) = {
   if (!n.last_serialization) return false
   // Prefer the original serialized type; fall back to the live type
   // for nodes whose serialization predates the type field.
   const originalType = n.last_serialization.type ?? n.type
   return !!originalType && targetTypes.has(originalType)
 })
 ```

src/components/dialog/content/MissingNodesContent.vue
jaeone94: Updated. The comment now explains the specific gap: `replaceNodesInPlace()` handles canvas rendering via `onNodeAdded()`, but the modal only manages its own local UI state (`replacedTypes` / `selectedTypes`). Without this `removeMissingNodesByType()` call, `executionErrorStore` wouldn't know about the replacement, so the Errors Tab would continue listing the replaced nodes as missing.

 ```typescript
 // replaceNodesInPlace() handles canvas rendering via onNodeAdded(),
 // but the modal only updates its own local UI state above.
 // Without this call the Errors Tab would still list the replaced nodes
 // as missing because executionErrorStore is not aware of the replacement.
 if (result.length  0) {
   executionErrorStore.removeMissingNodesByType(result)
 }

src/platform/nodeReplacement/useNodeReplacement.ts
jaeone94: Good question — I looked into this and you're right, there was a real issue. The previous implementation checked `registered_node_types`, which means if a user reinstalled the pack in the meantime, the node type would already be registered and `collectAllNodes` wouldn't find it. The UI would still show the "Replace" button, but clicking it would silently do nothing.

 To address this, I changed the approach to match against `targetTypes`, which is derived from the replaceable node list built at workflow load time — from the `node_replacements` API response and the missing nodes detected at that point. Since this data is determined at load time rather than queried from the current `registered_node_types`, replacement still works even if the user has since installed the missing node pack. Replacement typically upgrades to a newer/better node, so it should remain available regardless of installation state.

 ```typescript
 // Target types come from node_replacements fetched at workflow load time
 // and the missing nodes detected at that point — not from the current
 // registered_node_types. This ensures replacement still works even if
 // the user has since installed the missing node pack.
 const targetTypes = new Set(
   selectedTypes.map((t) = (typeof t === 'string' ? t : t.type))
 )
 ```

src/platform/nodeReplacement/useNodeReplacement.ts
christian-byrne: nice, the targetTypes approach is cleaner - eliminates the edge case entirely

src/platform/nodeReplacement/useNodeReplacement.ts
christian-byrne: looks good, much more readable now

src/platform/nodeReplacement/useNodeReplacement.ts
christian-byrne: makes sense - deriving from load-time data instead of runtime registration is the right call here

src/components/dialog/content/MissingNodesContent.vue
christian-byrne: the comment reads well now, thanks for spelling out the executionErrorStore gap

christian-byrne (approved):
LGTM!

christian-byrne:
note (non-blocking): the `collectAllNodes` predicate matches `originalType` against unsanitized `targetTypes`, but the fallback path (`n.last_serialization.type` is undefined) uses `n.type` which is sanitized - so types containing `& < > " ' \` =` would silently miss the `Set.has()` check. Super rare edge case (old serializations + special chars in type names), but could be closed with `targetTypes.has(originalType) || targetTypes.has(sanitizeNodeName(originalType))` if it ever comes up.

PR #9282, opened 2026-02-27
fix: stabilize nested subgraph promoted widget resolution

DrJKL:
## Summary

Fix multiple issues with promoted widget resolution in nested subgraphs, ensuring correct value propagation, slot matching, and rendering for deeply nested promoted widgets.

## Changes

- **What**: Stabilize nested subgraph promoted widget resolution chain
  - Use deep source keys for promoted widget values in Vue rendering mode
  - Resolve effective widget options from the source widget instead of the promoted view
  - Stabilize slot resolution for nested promoted widgets
  - Preserve combo value rendering for promoted subgraph widgets
  - Prevent subgraph definition deletion while other nodes still reference the same type
  - Clean up unused exported resolution types

## Review Focus

- `resolveConcretePromotedWidget.ts` — new recursive resolution logic for deeply nested promoted widgets
- `useGraphNodeManager.ts` — option extraction now uses `effectiveWidget` for promoted widgets
- `SubgraphNode.ts` — unpack no longer force-deletes definitions referenced by other nodes

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9282-fix-stabilize-nested-subgraph-promoted-widget-resolution-3146d73d365081208a4fe931bb7569cf) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LGraph.ts
```diff
@@ -1071,13 +1071,23 @@ export class LGraph
...
+            candidate.isSubgraphNode() &&
+            candidate.type === node.subgraph.id
+        )
+      )
+
+      if (!hasRemainingReferences) {
+        forEachNode(node.subgraph, (innerNode) => {
+          innerNode.onRemoved?.()
+          innerNode.graph?.onNodeRemoved?.(innerNode)
+        })
```
christian-byrne, line 1088: issue (non-blocking): when the last reference to a subgraph node is removed, the old code also deleted `rootGraph.subgraphs` entries for nested SubgraphNodes via `forEachNode`. That recursive cleanup was dropped here, so nested subgraph definitions remain orphaned in `rootGraph.subgraphs` until the workflow is reloaded.

Serialization already filters by `usedSubgraphIds`, so this isn't a correctness bug, but it is a session-lifetime memory leak for workflows that repeatedly create/remove deeply nested subgraphs, and the `allGraphs` scan in this same method grows monotonically.

I realize there's a lot of context and recent PRs around this area, and this is likely an accepted tradeoff for this PR to fix the immediate bugs. For the long term though, we need a better architecture for subgraph definition lifecycle management - probably something that reconciles definitions centrally rather than relying on node removal callbacks. Worth tracking as a follow-up.

src/lib/litegraph/src/subgraph/PromotedWidgetViewManager.ts
```diff
@@ -70,17 +80,40 @@ export class PromotedWidgetViewManager<TView> {
...
+    }
+    return true
+  }
+
+  private makeKey(
+    interiorNodeId: string,
+    widgetName: string,
+    viewKey?: string
+  ): string {
+    return viewKey ?? `${interiorNodeId}:${widgetName}`
```
christian-byrne: issue (non-blocking): `makeKey` returns `viewKey ?? \`${interiorNodeId}:${widgetName}\`` - when `viewKey` is provided, `interiorNodeId` and `widgetName` are completely ignored. This means:

1. `remove(id, name)` generates `${id}:${name}` but cannot find entries stored under a `viewKey`
2. `_removePromotedView` works around this by calling both `remove()` and `removeByViewKey()`, but that's a fragile implicit contract
3. During `reconcile`'s stale-key cleanup (line 54-56), entries stored under a viewKey cannot be evicted by their `(id, name)` pair

Consider either incorporating `interiorNodeId:widgetName` as a prefix in the viewKey path, or renaming the parameter to `keyOverride` and documenting the dual-keying contract explicitly.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -356,13 +387,30 @@ function drawDisconnectedPlaceholder(
   y: number,
   H: number
 ) {
+  const backgroundColor = readDesignToken(
```
christian-byrne, line 394: suggestion (non-blocking): `readDesignToken` calls `getComputedStyle(document.documentElement).getPropertyValue(token)` four times per `drawDisconnectedPlaceholder` invocation. This is on the canvas draw path, so for N disconnected promoted widgets it's 4N `getComputedStyle` calls per frame. `getComputedStyle` can force style recalculation.

Since these CSS custom properties only change on theme switch, consider caching the values at module level - e.g. a lazy `let cachedTokens: Record<string, string> | null` that's populated on first call and invalidated via a `matchMedia` listener or theme-change event.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -73,21 +74,240 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+
+      seenEntryKeys.add(entryKey)
+      return true
+    })
+  }
+
   private _getPromotedViews(): PromotedWidgetView[] {
     const store = usePromotionStore()
     const entries = store.getPromotionsRef(this.rootGraph.id, this.id)
+    const linkedEntries = this._getLinkedPromotionEntries()
```
christian-byrne, line 146: suggestion (non-blocking): `_getPromotedViews()` calls `_getLinkedPromotionEntries()` on every `widgets` getter access. That method iterates all inputs, resolves links via `resolveSubgraphInputLink` (which does `.find()` on slots, iterates `linkIds`, resolves links, and does another `.find()` on inner inputs), then deduplicates. The `widgets` getter is a very hot path - rendering, layout, serialization all hit it.

While `PromotedWidgetViewManager.reconcile()` caches the view objects, the entry key computation and link resolution preceding it runs unconditionally every time. Consider caching `_getLinkedPromotionEntries()` and invalidating only when inputs/links actually change (via the existing `input-connected`/`input-disconnected`/`removing-input` event handlers).

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -193,11 +197,68 @@ function safeWidgetMapper(
...
+
+    return {
+      canvasOnly: widget.options.canvasOnly,
+      advanced: widget.advanced,
+      hidden: widget.options.hidden,
+      read_only: widget.options.read_only
+    }
+  }
+
+  function resolveSourceSeedByInputName(inputName: string): {
```
christian-byrne: suggestion: `resolveSourceSeedByInputName` is a near-exact copy of `SubgraphNode._resolveLinkedPromotionByInputName` (same branching logic, same `resolveSubgraphInputLink` call, just different field names: `sourceNodeId` vs `interiorNodeId`). This is change amplification - any future change to resolution semantics must be updated in both places.

Consider extracting a shared resolver function in `resolveSubgraphInputLink.ts` that both call sites use.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -73,21 +74,240 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+    interiorNodeId: string
+    widgetName: string
+  }> = []
 
   // Declared as accessor via Object.defineProperty in constructor.
   // TypeScript doesn't allow overriding a property with get/set syntax,
   // so we use declare + defineProperty instead.
   declare widgets: IBaseWidget[]
 
+  private _resolveLinkedPromotionByInputName(
```
christian-byrne, line 103: suggestion (non-blocking): this PR adds ~220 lines of private promotion merge/sync infrastructure (`_resolveLinkedPromotionByInputName`, `_getLinkedPromotionEntries`, `_syncPromotions`, `_buildPromotionMergeState`, plus 8 more helpers). The class now mixes graph node concerns (rendering, serialization, events) with promotion state reconciliation.

Since the project is moving towards ECS-style centralized state, consider extracting this reconciliation logic into a standalone composable or pure-function module (e.g. `usePromotionReconciler(node)` or a set of pure functions operating on immutable promotion state) rather than coupling it to the class. That would align better with the Vue/functional/immutable direction and make the logic independently testable.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -193,11 +197,68 @@ function safeWidgetMapper(
...
     try {
+      const promotedInputName = isPromotedWidgetView(widget)
+        ? node.inputs?.find((input) => {
+            if (input.name === widget.name) return true
+            if (input._widget === widget) return true
+            return false
+          })?.name
+        : undefined
+      const displayName = promotedInputName ?? widget.name
+      const promotedSourceSeed =
```
christian-byrne: suggestion (non-blocking): `safeWidgetMapper` now interleaves widget-to-view-model mapping, promoted widget resolution, and store identity computation. `SafeWidgetData` has grown to 4 identity fields (`nodeId`, `storeNodeId`, `name`, `storeName`).

Similar to the SubgraphNode comment - as the project moves toward ECS-style centralized state, consider extracting the promoted widget identity resolution into a pure helper (e.g. `resolvePromotedWidgetIdentity(node, widget) -> { nodeId, storeNodeId, name, storeName, displayName }`). The mapper would then just assemble final data from pre-resolved identities, keeping concerns separated.

Also a naming nit: `promotedSourceSeed` sounds like it relates to seed randomization (a first-class concept in ComfyUI). It's actually the starting point for resolving the concrete promoted widget. Same for `resolveSourceSeedByInputName`. Consider `promotedSource` / `resolvePromotedSourceByInputName`.

src/core/graph/subgraph/resolveSubgraphInputLink.ts
```diff
@@ -0,0 +1,53 @@
+import type { INodeInputSlot } from '@/lib/litegraph/src/interfaces'
```
christian-byrne, line 1: suggestion (non-blocking): this new function has non-trivial logic (reverse iteration, link resolution, lazy widget caching) and is used in two critical call sites, but has no dedicated unit tests. The behavior is only exercised indirectly through integration-level subgraph tests.

Consider adding a focused `resolveSubgraphInputLink.test.ts` covering: non-subgraph node returns `undefined`, missing input slot, broken link, and the lazy `getTargetWidget` caching.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -215,36 +276,52 @@ function safeWidgetMapper(
...
-            advanced: widget.advanced,
-            hidden: widget.options.hidden,
-            read_only: widget.options.read_only
-          }
-        : undefined
+      const options = extractWidgetDisplayOptions(widget)
       const subgraphId = node.isSubgraphNode() && node.subgraph.id
 
+      const resolvedSourceResult =
+        isPromotedWidgetView(widget) && promotedSourceSeed
```
christian-byrne: suggestion (non-blocking): the new nested resolution paths in `safeWidgetMapper` (`storeNodeId`, `storeName`, `effectiveWidget` type resolution) have no dedicated tests. A promoted widget whose concrete source is two layers deep should produce the correct `storeNodeId`, `storeName`, and `type` in `SafeWidgetData`.

src/lib/litegraph/src/subgraph/PromotedWidgetViewManager.ts
```diff
@@ -70,17 +80,40 @@ export class PromotedWidgetViewManager<TView> {
     this.invalidateMemoizedList()
   }
 
+  removeByViewKey(
```
christian-byrne, line 83: suggestion (non-blocking): the new `viewKey` parameter and `removeByViewKey` method aren't covered by existing tests. This is the mechanism that prevents cache collisions when two inputs share one concrete widget name - worth having direct coverage.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -103,26 +97,28 @@ class PromotedWidgetView implements IPromotedWidgetView {
...
+  get computedDisabled(): boolean {
+    return this._computedDisabled
   }
 
-  set computedDisabled(_value: boolean | undefined) {}
+  set computedDisabled(value: boolean | undefined) {
+    this._computedDisabled = value ?? false
+  }
 
   get type(): IBaseWidget['type'] {
```
christian-byrne, line 111: suggestion (non-blocking): within a single draw cycle, the getters `type`, `options`, `tooltip`, `linkedWidgets`, `hidden`, `computeLayoutSize`, `computeSize`, and `draw()` each independently call `resolveConcrete()` -> `traversePromotedWidgetChain()`. For deeply nested subgraphs (the feature this PR enables), each call is O(D x W) where D = nesting depth and W = widgets per node.

A few concrete approaches to avoid redundant traversals:
- **Dirty-flag pattern** (similar to Vue's internal computed): cache the resolved result with a version counter, invalidate it when the subgraph structure changes (node added/removed, link connected/disconnected). Getters read from cache; mutations bump the version.
- **Per-frame cache**: set a `_resolvedConcreteCache` field at the start of `draw()`, clear it at the end. Other getters check the cache first.
- **Lazy computed**: use a Vue `computed()` ref internally (since this is already in a Vue app context) that depends on a reactive trigger bumped by structural changes.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -464,24 +679,68 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
         break
       }
     }
+
+    this._syncPromotions()
```
christian-byrne, line 692: nitpick (non-blocking): sonarjs flags this method at cognitive complexity 18 (threshold 15). Extracting the legacy entry resolution loop into a helper would bring it under the threshold.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -73,21 +74,240 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+      inputName: string
+      interiorNodeId: string
+      widgetName: string
+    }> = []
+
+    for (const input of this.inputs) {
+      const resolved = this._resolveLinkedPromotionByInputName(input.name)
+      if (!resolved) continue
+
+      input.widget ??= { name: input.name }
```
christian-byrne: suggestion (non-blocking): `_getLinkedPromotionEntries()` is called from the `widgets` getter but mutates `input.widget` (`??=` and unconditional `.name =` assignment). This violates the expectation that getters are side-effect free. If any code reads `widgets` in a read-only context (serialization, the reference-counting scan in `LGraph.remove`), it will silently mutate input state.

Consider moving the `input.widget` initialization into `_setWidget` or `_syncPromotions` where mutations are expected.

src/core/graph/subgraph/resolveConcretePromotedWidget.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  return { node, widget }
+}
+
+export function resolvePromotedWidgetLookupTarget(
+  hostNode: SubgraphNode,
+  nodeId: string,
+  widgetName: string
+): { nodeId: string; widgetName: string } {
+  const result = traversePromotedWidgetChain(hostNode, nodeId, widgetName)
```
christian-byrne: question: when `traversePromotedWidgetChain` returns a failure (missing-node, missing-widget, max-depth-exceeded), this falls back to the original promoted widget's coordinates (`{ nodeId, widgetName }`). For transient states during graph editing, this means `getWidgetState()` in `PromotedWidgetView` will look up a store entry that may not exist or belongs to a different widget.

Would it be safer to return `undefined` on failure and have callers handle the missing state explicitly (e.g. showing the disconnected placeholder), or use the last successfully resolved coordinates before the failure?

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -244,15 +243,44 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   private resolve(): { node: LGraphNode; widget: IBaseWidget } | undefined {
```
christian-byrne: suggestion (non-blocking): `resolve()` vs `resolveConcrete()` don't communicate the depth difference - one resolves at the immediate host level, the other traverses the full nested chain to the leaf widget. The public API already uses clearer names (`resolvePromotedWidgetAtHost` vs `resolveConcretePromotedWidget`). Consider aligning the private methods, e.g. `resolveAtHost()` / `resolveDeepest()`.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -73,21 +74,240 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+        (entry, index) =>
+          entry.interiorNodeId !== entries[index]?.interiorNodeId ||
+          entry.widgetName !== entries[index]?.widgetName
+      )
+    if (!hasChanged) return
+
+    store.setPromotions(this.rootGraph.id, this.id, mergedEntries)
+  }
+
+  private _buildPromotionMergeState(
```
christian-byrne: suggestion (non-blocking): `_buildPromotionMergeState` returns a 4-field object (`displayNameByViewKey`, `mergedEntries`, `reconcileEntries`, `shouldPersistLinkedOnly`) consumed by two callers that each destructure different subsets. It mixes reconciliation concerns (view identity) with persistence concerns (store sync).

Consider splitting into two focused functions - one for view reconciliation (returning `reconcileEntries` + `displayNameByViewKey`) and one for store sync (returning `mergedEntries` + whether to persist). Each caller would get exactly what it needs.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -73,21 +74,240 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+    widgetName: string
+  }> = []
 
   // Declared as accessor via Object.defineProperty in constructor.
   // TypeScript doesn't allow overriding a property with get/set syntax,
   // so we use declare + defineProperty instead.
   declare widgets: IBaseWidget[]
 
+  private _resolveLinkedPromotionByInputName(
+    widgetName: string
```
christian-byrne: nitpick (non-blocking): the parameter is named `widgetName` but per the method name (`_resolveLinkedPromotionByInputName`) and call site (`this._resolveLinkedPromotionByInputName(input.name)`), it receives an input slot name. Consider renaming to `inputName`.

src/core/graph/subgraph/resolveSubgraphInputLink.ts
```diff
@@ -0,0 +1,53 @@
...
+  resolve: (context: SubgraphInputLinkContext) => TResult | undefined
+): TResult | undefined {
+  if (!node.isSubgraphNode()) return undefined
+
+  const inputSlot = node.subgraph.inputNode.slots.find(
+    (slot) => slot.name === inputName
+  )
+  if (!inputSlot) return undefined
+
+  for (let index = inputSlot.linkIds.length - 1; index >= 0; index -= 1) {
```
christian-byrne, line 23: suggestion (non-blocking): the reverse iteration over `inputSlot.linkIds` seems intentional (prefer the most recently connected link?), but there's no comment explaining the ordering expectation. A brief inline comment would help future readers understand why reverse order matters here.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -645,6 +924,19 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
     }
   }
 
+  private _removePromotedView(view: PromotedWidgetView): void {
```
christian-byrne, line 936: suggestion (non-blocking): `_removePromotedView` calls both `remove()` (composite key) and `removeByViewKey()` (viewKey) for the same entry. It's unclear from reading this whether both are always needed or if this is defensive cleanup. A brief comment explaining the dual-removal strategy would help - e.g. "views may be cached under either key format depending on whether they were created via reconcile or getOrCreate".

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -73,21 +74,240 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+
+  private _buildPromotionMergeState(
+    entries: Array<{ interiorNodeId: string; widgetName: string }>,
+    linkedEntries: Array<{
+      inputName: string
+      interiorNodeId: string
+      widgetName: string
+    }>
+  ): {
+    displayNameByViewKey: Map<string | undefined, string>
```
christian-byrne: nitpick (non-blocking): `Map<string | undefined, string>` allows `undefined` as a key, implying a meaningful `.get(undefined)` case. Consider using a sentinel string (e.g. empty string) for fallback entries, or narrowing to `Map<string, string>` and skipping the lookup for entries without a viewKey.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -244,15 +243,44 @@ class PromotedWidgetView implements IPromotedWidgetView {
...
+    const lookupTarget = this.resolveStateLookupTarget()
     return useWidgetValueStore().getWidget(
       this.graphId,
-      this.bareNodeId,
+      lookupTarget.nodeId,
+      lookupTarget.widgetName
+    )
+  }
+
+  private resolveStateLookupTarget(): { nodeId: NodeId; widgetName: string } {
```
christian-byrne: suggestion (non-blocking): `resolveStateLookupTarget()` triggers full chain traversal on every `value` get/set, but the lookup target rarely changes (only when subgraph structure changes).

A pattern similar to how Vue's `computed()` works internally would fit well here: cache the result with a dirty flag, and invalidate it only when the subgraph structure actually changes (node added/removed, link connected/disconnected). Something like:

```ts
private _lookupTargetDirty = true
private _cachedLookupTarget?: { nodeId: NodeId; widgetName: string }

private resolveStateLookupTarget() {
  if (!this._lookupTargetDirty && this._cachedLookupTarget) {
    return this._cachedLookupTarget
  }
  // ... existing resolution logic ...
  this._lookupTargetDirty = false
  this._cachedLookupTarget = result
  return result
}
```

With the dirty flag bumped from the structural change events the SubgraphNode already listens to.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -73,21 +74,240 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+        if (!targetWidget) return undefined
+        return {
+          interiorNodeId: String(inputNode.id),
+          widgetName: targetWidget.name
+        }
+      }
+    )
+  }
+
+  private _getLinkedPromotionEntries(): Array<{
```
christian-byrne: nitpick (non-blocking): the type `{ inputName: string; interiorNodeId: string; widgetName: string }` is repeated inline across 7 method signatures. Consider extracting a named type alias (e.g. `LinkedPromotionEntry`) to reduce change amplification.

src/components/graph/widgets/DomWidget.vue
```diff
@@ -110,13 +110,17 @@ watch(
       updateDomClipping()
     }
 
+    const override = widgetState.positionOverride
+    const isDisabled = override
```
christian-byrne, line 114: nitpick (non-blocking): this new disabled-state logic is only covered by the browser-level E2E test (`subgraphPromotion.spec.ts`), not by a focused component-level test. Consider adding one that verifies `opacity: 0.5` and `pointerEvents: 'none'` when `positionOverride.widget.computedDisabled` is true.

src/core/graph/subgraph/resolvePromotedWidgetSource.ts
```diff
@@ -1,29 +1,22 @@
...
-interface ResolvedPromotedWidgetSource {
-  node: LGraphNode
-  widget: IBaseWidget
-}
-
 export function resolvePromotedWidgetSource(
   hostNode: LGraphNode,
   widget: IBaseWidget
-): ResolvedPromotedWidgetSource | undefined {
+): ResolvedPromotedWidget | undefined {
```
christian-byrne, line 10: note (non-blocking): `resolvePromotedWidgetSource` now resolves through nested promoted widget chains (via `resolveConcretePromotedWidget`) instead of only resolving the immediate source. Extensions calling this function may now receive a deeper interior node/widget pair than before. The structural type shape is unchanged, but the behavioral contract is different. If any extension needs the shallow (one-level) resolution, they should use the new `resolvePromotedWidgetAtHost` export.

src/core/graph/subgraph/promotedWidgetView.test.ts
```diff
@@ -786,6 +953,443 @@ function createFakeCanvasContext() {
...
+      0,
+      LiteGraph.NODE_WIDGET_HEIGHT,
+      false
+    )
+
+    const renderedText = fillText.mock.calls.map((call) => call[0])
+    expect(renderedText).toContain('a')
+  })
+
+  test('draw shows value through two input-based promotion layers', () => {
```
christian-byrne, line 1043: nitpick (non-blocking): the nested promotion tests ("draw shows value through two...", "value updates propagate...", "state lookup resolves...", etc.) repeat nearly identical 15-line setup blocks to create a two-level nested subgraph. Extracting a shared `createTwoLevelNestedSubgraph()` helper would reduce setup noise and make the unique assertion in each test stand out.

src/core/graph/subgraph/promotedWidgetView.test.ts
```diff
@@ -786,6 +953,443 @@ function createFakeCanvasContext() {
...
+    const subgraphB = createTestSubgraph({
+      inputs: [{ name: 'b_input', type: '*' }]
+    })
+    subgraphB.add(subgraphNodeA)
+    subgraphNodeA._internalConfigureAfterSlots()
+    subgraphB.inputNode.slots[0].connect(subgraphNodeA.inputs[0], subgraphNodeA)
+
+    const subgraphNodeB = createTestSubgraphNode(subgraphB, { id: 22 })
+    const widgetValueStore = useWidgetValueStore()
+    const getWidgetSpy = vi.spyOn(widgetValueStore, 'getWidget')
```
christian-byrne: nitpick (non-blocking): this test uses `vi.spyOn(widgetValueStore, 'getWidget')` to assert that `getWidget` was called with the concrete inner node's ID. This is a change-detector test - if the implementation changes how it retrieves widget state (caching, batching), this breaks even though behavior is correct. The value propagation tests nearby already verify the same semantics more robustly by checking actual values.

src/core/graph/subgraph/resolveConcretePromotedWidget.test.ts
```diff
@@ -0,0 +1,343 @@
...
+import {
+  resolveConcretePromotedWidget,
+  resolvePromotedWidgetAtHost,
+  resolvePromotedWidgetLookupTarget
+} from '@/core/graph/subgraph/resolveConcretePromotedWidget'
+import type { LGraphNode } from '@/lib/litegraph/src/litegraph'
+import type { SubgraphNode } from '@/lib/litegraph/src/subgraph/SubgraphNode'
+import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets'
+
+type MockGraphNode = {
```
christian-byrne: nitpick (non-blocking): these tests use hand-rolled `MockGraphNode`/`MockSubgraph` types instead of the `createTestSubgraph`/`createTestSubgraphNode` helpers used elsewhere. This keeps tests fast but creates a divergence risk - if `SubgraphNode` gains new properties that affect `isPromotedWidgetView` or `isSubgraphNode` checks, these mocks won't catch regressions. Consider adding at least one integration-level test using real fixtures (the nested promotion tests in `promotedWidgetView.test.ts` may already cover this).

src/lib/litegraph/src/subgraph/PromotedWidgetViewManager.ts
DrJKL: There once was a key with no base,
That wandered all over the place.
Now node-widget leads,
Then viewKey succeeds,
And cache ops all run in one space.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: A placeholder painted each frame,
With style reads again and again.
Now token values stay,
In a cache by the way,
So redraws are far less of a strain.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: A getter kept chasing each link,
On every tiny re-think.
Now entries are cached,
With invalidates matched,
So hot paths no longer must clink.

src/composables/graph/useGraphNodeManager.ts
DrJKL: Two resolvers marched side by side,
With branching and checks duplicated wide.
Now one helper knows,
Where the true target goes,
And both callers use that one guide.

src/core/graph/subgraph/resolveSubgraphInputLink.ts
DrJKL: A resolver with subtle intent,
Had tests that were only indirect.
Now stale links are tried,
Missing inputs denied,
And widget caching is fully checked.

src/composables/graph/useGraphNodeManager.ts
DrJKL: Through two nested graphs it would roam,
To find where the values call home.
Now tests verify clear,
storeNodeId is near,
And storeName and type match the dome.

src/lib/litegraph/src/subgraph/PromotedWidgetViewManager.ts
DrJKL: When two promoted views looked the same,
viewKey had to carry the flame.
Now tests prove they split,
And targeted deletes hit,
So cache identity stays tame.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: A widget kept diving too deep,
For type, draw, and state it would leap.
Now revision-based cache,
Makes repeated calls dash,
And traversal cost no longer steep.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: A collector that promised to read,
Was mutating input in speed.
That write has been moved,
So purity's proved,
And getters do less than they need.

src/core/graph/subgraph/resolveSubgraphInputLink.ts
DrJKL: It looped from the end to the start,
But why was not clear to the heart.
Now comment declares,
Newest link gets the shares,
And readers can follow that art.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: A map with undefined in play,
Suggested a fallback that may.
Now key type is tight,
All strings, clean and right,
And lookups are clearer today.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: Lookup kept traversing each call,
Though structure changed rarely at all.
Now target is cached,
By revision it's matched,
And churn has grown notably small.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: A type that repeated too long,
In signatures all through the song.
Now alias is named,
And repetition is tamed,
So edits are safer and strong.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: @coderabbitai Try to make the issue again.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Merged winds now split clean,
Views reconcile, stores persist,
Each task has one home.

src/composables/graph/useGraphNodeManager.ts
DrJKL: One path names the source,
Mapper breathes in simpler lines,
Identity stays.

src/core/graph/subgraph/promotedWidgetView.test.ts
DrJKL: Spyglass set aside,
Behavior sings what is true,
Value proves the path.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Two keys swept with care,
One note guards forgotten trails,
Cache ghosts fade to still.

src/components/graph/widgets/DomWidget.vue
DrJKL: Disabled bloom dims,
Opacity, touch held still,
Small test keeps the truth.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: Two depths, two clear names,
Host and deepest now spoken,
Intent lights the way.

src/core/graph/subgraph/promotedWidgetView.test.ts
DrJKL: Nested dawn once cloned,
One helper now sets the stage,
Assertions stand bright.

browser_tests/tests/interaction.spec.ts
```diff
@@ -171,6 +171,7 @@ test.describe('Node Interaction', () => {
 
   test('Can drag node', { tag: '@screenshot' }, async ({ comfyPage }) => {
     await comfyPage.nodeOps.dragTextEncodeNode2()
+    await comfyPage.nextFrame()
```
DrJKL, line 174: This is a stabilization fix for the test, the DOM widget z-index is inconsistent at point of screenshot otherwise.

DrJKL: Might be worth coming back and adding a proper wait condition later.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: **F**rame remembers what frenzy forgot.
**R**esolution lands once, then echoes.
**A**cross getters, draw, tooltip, type—same pulse.
**M**icro-storms of traversal collapse into one weather cell.
**E**ffort bends; behavior stays.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: **L**ookup no longer chases every footstep.
**O**ne deep resolve feeds state reads and writes.
**O**verhead thins into a single contour.
**K**inetic UI keeps its breath.
**U**nnecessary traversals dissolve.
**P**ath stays correct, just quieter.

src/core/graph/subgraph/resolveConcretePromotedWidget.ts
DrJKL: **U**nmoored fallbacks retired.
**N**ull is now a signal, not a shrug.
**D**eep resolution fails loud by absence.
**E**rrant coordinates no longer impersonate truth.
**F**ailure becomes form: undefined, and the placeholder breathes.

src/core/graph/subgraph/resolveConcretePromotedWidget.test.ts
DrJKL: **F**ixtures replace stage props.
**I**ntegration texture enters the test grain.
**X**-ray mocks give way to living subgraph bones.
**T**he helpers (createTestSubgraph, createTestSubgraphNode) anchor reality.
**U**nder drift risk, coverage hardens.
**R**egression now meets a truer mirror.
**E**dges hold.

PR #9287, opened 2026-02-27
Fix essentials nodes not being marked core

AustinMroz:
In adding an essentials cateogory for nodes, #8987 introduced a regression where core nodes which are also essential are marked as being from a `nodes` custom node instead of being marked core. Since the essentials designation should pre-empt core and custom nodes can choose to mark themself as essential, the getter for `isCoreNode` is updated to instead repeat the existing check for if a node is core.

| Before | After |
| ------ | ----- |
| before | after|

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9287-Fix-essentials-nodes-not-being-marked-core-3146d73d365081fca2a0f8bdc2baf01a) by [Unito](https://www.unito.io)


AustinMroz:
Manually updated the incorrect snapshot. The other test is a known unrelated flake.

pythongosssss (approved):

PR #9294, opened 2026-02-28
feat: wrap CURVE widget value with typed format

jtydhr88:
## Summary
Send CURVE values as { __type: 'CURVE', value: [...] } instead of { __value__: [...] } to avoid ambiguity with link detection and enable external tools to identify the data type.

change requested by @guill

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9294-feat-wrap-CURVE-widget-value-with-typed-format-3156d73d365081bf8e5de59527e2d3ce) by [Unito](https://www.unito.io)

### DIFF
--- a/src/utils/executionUtil.ts
+++ b/src/utils/executionUtil.ts
@@ -109,7 +109,7 @@ export const graphToPrompt = async (
         // execution.
         inputs[widget.name] = Array.isArray(widgetValue)
           ? widget.type === 'curve'
-            ? { __type: 'CURVE', value: widgetValue }
+            ? { __type__: 'CURVE', __value__: widgetValue }
             : { __value__: widgetValue }
           : widgetValue
       }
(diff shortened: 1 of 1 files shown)


src/utils/executionUtil.ts
```diff
@@ -108,9 +108,9 @@ export const graphToPrompt = async (
         // The backend automatically unwraps the object to an array during
         // execution.
         inputs[widget.name] = Array.isArray(widgetValue)
-          ? {
-              __value__: widgetValue
-            }
+          ? widget.type === 'curve'
+            ? { __type: 'CURVE', value: widgetValue }
```
guill: We should probably be consistent and do `__type__` and `__value__` like we do below.

guill (dismissed):
Requested a change, but proactively approving so you don't have to wait for me after you make the change.

guill (approved):

PR #9295, opened 2026-02-28
feat: comprehensive subscription funnel telemetry

huntcsg:
## Summary

Fixes telemetry gaps across the subscription/upgrade funnel identified in a comprehensive audit.

### Modal tracking
- Add `modal_name` property to `subscription_required_modal_opened` events (`free_tier_upsell` vs `pricing_table`) so each dialog self-identifies
- Fire `subscribe_now_button_clicked` when user clicks through FreeTierDialog → PricingTable
- Fire second `modal_opened` impression when PricingTable opens in the two-step flow
- Add `entry_point` tracking to settings upgrade (`settings_upgrade_plan`) and popover entries (`popover_upgrade`, `popover_plans_and_pricing`)

### current_tier threading
- Add `current_tier` to `run_button_clicked` events
- Add `current_tier` to `add_api_credit_button_clicked` events (previously had no metadata)
- Add `previous_tier` to workspace `begin_checkout` events (legacy already had it)

https://github.com/user-attachments/assets/80b7593d-7247-4bff-bb1e-fd06bda82b6f

### Other fixes
- Fix TopbarSubscribeButton to fire `subscribe_clicked` instead of `run_button_clicked`
- Wire up workspace topup success tracking in `billingOperationStore`
- Add `subscription_succeeded` tracking to legacy dialog polling
- Remove redundant `modal_opened` from `useSubscription` (now centralized in `useSubscriptionDialog`)
- Centralize all modal/button tracking in `useSubscriptionDialog` composable

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9295-feat-comprehensive-subscription-funnel-telemetry-3156d73d3650813db6a4d1675c36bc9d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/actionbar/ComfyRunButton/CloudRunButtonWrapper.vue
+++ b/src/components/actionbar/ComfyRunButton/CloudRunButtonWrapper.vue
@@ -1,19 +1,16 @@
 <template>
-  <component
-    :is="currentButton"
-    :key="isActiveSubscription ? 'queue' : 'subscribe'"
-  />
+  <div class="relative">
+    <ComfyQueueButton />
+    <SubscribeToRunButton
+      v-if="!isActiveSubscription"
+      class="absolute inset-0 z-10"
+    />
+  </div>
 </template>
 <script setup lang="ts">
-import { computed } from 'vue'
-
 import ComfyQueueButton from '@/components/actionbar/ComfyRunButton/ComfyQueueButton.vue'
 import { useBillingContext } from '@/composables/billing/useBillingContext'
 import SubscribeToRunButton from '@/platform/cloud/subscription/components/SubscribeToRun.vue'
 
 const { isActiveSubscription } = useBillingContext()
-
-const currentButton = computed(() =>
-  isActiveSubscription.value ? ComfyQueueButton : SubscribeToRunButton
-)
 </script>
--- a/src/components/dialog/content/setting/LegacyCreditsPanel.vue
+++ b/src/components/dialog/content/setting/LegacyCreditsPanel.vue
@@ -137,7 +137,7 @@ const authStore = useFirebaseAuthStore()
 const authActions = useFirebaseAuthActions()
 const commandStore = useCommandStore()
 const telemetry = useTelemetry()
-const { isActiveSubscription } = useBillingContext()
+const { isActiveSubscription, subscription } = useBillingContext()
 const loading = computed(() => authStore.loading)
 const balanceLoading = computed(() => authStore.isFetchingBalance)
 
@@ -160,7 +160,9 @@ watch(
 
 const handlePurchaseCreditsClick = () => {
   // Track purchase credits entry from Settings > Credits panel
-  useTelemetry()?.trackAddApiCreditButtonClicked()
+  useTelemetry()?.trackAddApiCreditButtonClicked({
+    current_tier: subscription.value?.tier?.toLowerCase()
+  })
   dialogService.showTopUpCreditsDialog()
 }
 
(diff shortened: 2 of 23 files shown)


christian-byrne:
## Draft Review — Architectural Feedback

**Verdict: 🟡 Needs course correction**

Direction is solid — pulling modal/button tracking into `useSubscriptionDialog`, threading `current_tier`/`previous_tier`/`entry_point`, and renaming `monthly_subscription_succeeded` → `subscription_succeeded` are all the right moves. A few structural items worth resolving before this goes ready:

### Architectural concerns

1. **Centralization is partial — `modal_opened` is dropped on the legacy path.** Removing the `modal_opened` call from `useSubscription.showSubscriptionDialog` orphans every caller that flows through `useBillingContext.showSubscriptionDialog()` (e.g. `useCoreCommands` run-when-unsubscribed, the new `TopbarSubscribeButton`). That path opens `SubscriptionRequiredDialogContent` via `legacyShowSubscriptionDialog` and never calls `useSubscriptionDialog.show()`, so it now silently fires no `modal_opened`. Either route the legacy entry through `useSubscriptionDialog.show()` or fire `modal_opened` from `SubscriptionRequiredDialogContent`'s `onMounted`.

2. **`entry_point` should be a string-literal union, not `string`.** Loose typing on a telemetry property invites typos that only surface in Mixpanel. Define `type SubscriptionEntryPoint = 'settings_upgrade_plan' | 'popover_upgrade' | 'popover_plans_and_pricing' | ...` and use it on `SubscriptionDialogOptions` and `SubscriptionMetadata`.

3. **`subscription.value?.tier?.toLowerCase()` is duplicated 8+ times.** Expose a `currentTier` computed on `useBillingContext` (or `useSubscription`) so call sites just spread `{ current_tier }` — keeps the lowercase normalization in one place.

4. **Possible double-fire of `subscription_succeeded`.** Both `billingOperationStore` (`operation.type === 'subscription'`) and `SubscriptionRequiredDialogContentWorkspace.handleAddCreditCard` / `handleConfirmTransition` fire it on the workspace path. Worth confirming the success states are mutually exclusive, or moving the call to a single source of truth (the store seems like the right place).

5. **Renaming `MONTHLY_SUBSCRIPTION_SUCCEEDED` → `SUBSCRIPTION_SUCCEEDED` is an event-name break.** Confirm with the data team that downstream dashboards/funnels are migrated, or dual-fire for one release cycle.

6. **Duplicated subscribe-CTA styling.** The new `TopbarSubscribeButton.vue` and `CurrentUserPopoverLegacy.vue` both inline the same `var(--color-subscription-button-gradient)` style block. Per AGENTS.md, prefer a shared Button variant or class over inline `:style` — this CTA appears 2–3 times now and will keep growing.

7. **Overlay pattern leaves `ComfyQueueButton` focusable.** `<div class="relative"><ComfyQueueButton /><SubscribeToRunButton class="absolute inset-0 z-10" /></div>` (also `LinearControls.vue`) keeps the queue button in the tab order under the overlay. Add `inert`/`aria-hidden` to the underlying button when the subscribe overlay is shown.

8. **PR scope.** Title says "telemetry", but this also ships a UX feature (always-visible subscribe button across topbar / linear / cloud-run wrapper). Worth splitting into two PRs — telemetry can ship/revert independently and the UI change deserves its own design review.

### Suggested direction

Land #1 first (it's a real regression), pull #6/#7 into a separate UX PR, then #2–#5 are quick refactors that keep the centralization actually centralized. Overall this is a meaningful funnel improvement — these are tightening passes, not a rethink.

---
*This is an early-stage review focused on direction — detailed line-by-line feedback will come when the PR is marked ready for review.*

PR #9298, opened 2026-02-28
feat: use cloud backend thumbnail resize for image previews

dante01yoon:
## Summary

- In cloud mode, large generated images (4K, 8K+) cause browser freezing when loaded at full resolution for preview display
- The cloud backend (ingest service) now supports a `res` query parameter on `/api/view` that returns server-side resized JPEG (quality 80, max 512px) instead of redirecting to the full-size GCS original
- This PR adds `&res=512` to all image preview URLs in cloud mode, reducing browser decode overhead from tens of MB to tens of KB
- Downloads still use the original resolution (no `res` param)
- No impact on localhost/desktop builds (`isCloud` compile-time constant)

### without `?res`

302 -> png downloads
스크린샷 2026-02-28 오후 6 53 03
스크린샷 2026-02-28 오후 6 52 52

200 -> jpeg

### with `?res`
스크린샷 2026-02-28 오후 6 51 55

### Changes

- **New utility**: `getCloudResParam(filename?)` returns `&res=512` in cloud mode for image files, empty string otherwise
- **Core stores**: `imagePreviewStore` appends `res` to node output URLs; `queueStore.ResultItemImpl` gets a `previewUrl` getter (separates preview from download URLs)
- **Applied to**: asset browser thumbnails, widget dropdown previews, linear mode indicators, image compare node, background image upload

### Intentionally excluded

- Downloads (`getAssetUrl`) — need original resolution
- Mask editor — needs pixel-accurate data
- Audio/video/3D files — `res` only applies to raster images
- Execution-in-progress previews — use WebSocket blob URLs, not `/api/view`

## Test plan

- [x] Unit tests for `getCloudResParam()` (5 tests: cloud/non-cloud, image/non-image, undefined filename)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] All 5332 unit tests pass
- [x] Manual verification on cloud.comfy.org: `res=512` returns 200 with resized JPEG; without `res` returns 302 redirect to GCS PNG original


src/platform/distribution/cloudPreviewUtil.ts
```diff
@@ -0,0 +1,32 @@
...
+const IMAGE_EXTENSIONS = new Set([
+  'png',
+  'jpg',
+  'jpeg',
+  'webp',
+  'gif',
+  'bmp',
+  'tiff',
+  'tif'
+])
```
DrJKL: Do we not have these enumerated somewhere already? I feel like a lot of this could be more general.

src/platform/distribution/cloudPreviewUtil.ts
```diff
@@ -0,0 +1,32 @@
...
+ * Returns `&res=N` for cloud image preview URLs.
+ * The cloud backend resizes images server-side to prevent
+ * the frontend from loading very large images.
+ *
+ * Returns empty string when not in cloud mode or for non-image files.
+ */
+export function getCloudResParam(filename?: string): string {
+  if (!isCloud) return ''
+  if (filename && !isImageFilename(filename)) return ''
+  return `&res=${CLOUD_PREVIEW_RES}`
```
DrJKL: Until/unless we need to coordinate this across different modules, I'd inline the number here.

src/renderer/extensions/linearMode/LinearControls.vue
```diff
@@ -89,7 +90,7 @@ function getDropIndicator(node: LGraphNode) {
     iconClass: 'icon-[lucide--image]',
     imageUrl: filename
       ? api.apiURL(
-          `/view?${new URLSearchParams(resultItem)}${app.getPreviewFormatParam()}`
+          `/view?${new URLSearchParams(resultItem)}${app.getPreviewFormatParam()}${getCloudResParam(String(filename))}`
```
DrJKL: It feels like we should be structuring this such that resultItem has _all_ of the query parameters we want to append

src/components/common/BackgroundImageUpload.vue
```diff
@@ -88,7 +89,7 @@ const handleFileUpload = async (event: Event) => {
           type: 'input',
           subfolder: 'backgrounds'
         })
-        modelValue.value = `/api/view?${params.toString()}`
+        modelValue.value = `/api/view?${params.toString()}${getCloudResParam(file.name)}`
```
DrJKL: See below, but I think we should be standardizing to fit new options into the URLSearchParams instead of mixing it so that some are in there and some are generated separately and appended

src/platform/distribution/cloudPreviewUtil.test.ts
```diff
@@ -0,0 +1,50 @@
+import { describe, expect, it, vi } from 'vitest'
+
+let mockIsCloud = false
```
DrJKL: You'll probably want to hoist this.

src/platform/distribution/cloudPreviewUtil.test.ts
```diff
@@ -0,0 +1,50 @@
...
+vi.mock('./types', () => ({
+  get isCloud() {
+    return mockIsCloud
+  }
+}))
+
+const { getCloudResParam } = await import('./cloudPreviewUtil')
```
DrJKL: Why not just import this normally?

DrJKL (changes requested):
This makes me think we're missing a more general place to construct the query params for all of these. Would you mind fixing that while you're here to make subsequent changes like this a lot simpler?

src/platform/distribution/cloudPreviewUtil.test.ts
dante01yoon: Switched to static import with vi.hoisted().

src/platform/distribution/cloudPreviewUtil.ts
dante01yoon: Good catch — replaced with getMediaTypeFromFilename from shared-frontend-utils.

src/platform/distribution/cloudPreviewUtil.ts
dante01yoon: Agreed, inlined it.

src/renderer/extensions/linearMode/LinearControls.vue
dante01yoon: Refactored — the function now takes a URLSearchParams and sets res on it directly. All call sites build params via URLSearchParams first, then call appendCloudResParam(params, filename).

src/components/common/BackgroundImageUpload.vue
dante01yoon: Addressed together with above — same pattern applied here.

src/platform/distribution/cloudPreviewUtil.test.ts
dante01yoon: Done — using vi.hoisted() now.

src/platform/distribution/cloudPreviewUtil.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  }
+}))
+
+import { appendCloudResParam } from './cloudPreviewUtil'
+
+function buildParams(filename?: string): URLSearchParams {
+  const params = new URLSearchParams()
+  appendCloudResParam(params, filename)
+  return params
+}
```
DrJKL, line 17:
```suggestion
import { describe, expect, it, vi } from 'vitest'

import { appendCloudResParam } from './cloudPreviewUtil'

const mockIsCloud = vi.hoisted(() => ({ value: false }))

vi.mock('./types', () => ({
  get isCloud() {
    return mockIsCloud.value
  }
}))

function buildParams(filename?: string): URLSearchParams {
  const params = new URLSearchParams()
  appendCloudResParam(params, filename)
  return params
}
```

src/platform/distribution/cloudPreviewUtil.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  it('does not set res for non-image files in cloud mode', () => {
+    mockIsCloud.value = true
+    for (const name of [
+      'video.mp4',
+      'video.webm',
+      'audio.mp3',
+      'audio.wav',
+      'model.glb',
+      'icon.svg'
+    ]) {
```
DrJKL: Could you use `it.for` here?

src/stores/imagePreviewStore.ts
```diff
@@ -118,10 +119,13 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
     const rand = app.getRandParam()
     const previewParam = getPreviewParam(node, outputs)
+    const isImage = isImageOutputs(node, outputs)
+    const firstFilename = outputs.images[0]?.filename
 
     return outputs.images.map((image) => {
-      const imgUrlPart = new URLSearchParams(image)
-      return api.apiURL(`/view?${imgUrlPart}${previewParam}${rand}`)
+      const params = new URLSearchParams(image)
+      if (isImage) appendCloudResParam(params, firstFilename)
```
DrJKL, line 127: @coderabbitai Create an issue to bring the previewParam and rand into alignment with the rest of the parameters.

DrJKL (dismissed):

DrJKL (approved):

src/extensions/core/imageCompare.ts
```diff
@@ -25,9 +26,10 @@ useExtensionService().registerExtension({
 
       const { a_images: aImages, b_images: bImages } = output
       const rand = app.getRandParam()
+      const resParam = getCloudResParam()
```
christian-byrne, line 29: suggestion (non-blocking): `getCloudResParam()` is called without a filename here, which bypasses the image-extension guard and unconditionally returns `&res=512` in cloud mode. Every other call site passes a filename for the safety check. Consider moving it inside `toUrl` and passing the filename:

```ts
const toUrl = (params: Record<string, string>) =>
  api.apiURL(`/view?${new URLSearchParams(params)}${getCloudResParam(params.filename)}${rand}`)
```

src/stores/imagePreviewStore.ts
```diff
@@ -118,10 +119,13 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
 
     const rand = app.getRandParam()
     const previewParam = getPreviewParam(node, outputs)
+    const resParam = isImageOutputs(node, outputs)
```
christian-byrne, line 122: suggestion (non-blocking): `resParam` is computed from `outputs.images[0]?.filename` then applied to every image in the `.map()`. If a batch ever contains mixed file types, the first file's extension governs whether `&res=512` is appended to all of them. Consider moving `getCloudResParam` inside the `.map()` callback so each image gets its own per-filename check - the cost is trivial (a Set lookup).

src/renderer/extensions/linearMode/LinearControls.vue
```diff
@@ -92,7 +90,7 @@ function getDropIndicator(node: LGraphNode) {
...
```
christian-byrne, line 93: nitpick (non-blocking): `String(filename)` converts a possibly-undefined value to the string `"undefined"`, which then fails the image extension check and returns `''`. This works but by accident - `getCloudResParam` already accepts `string | undefined`. Could pass `filename` directly since the surrounding ternary already guards the undefined case.

src/platform/distribution/cloudPreviewUtil.test.ts
christian-byrne, line 3: suggestion (non-blocking): `let mockIsCloud = false` is global mutable state without a `beforeEach` reset. Per the project's AGENTS.md convention, consider using `vi.hoisted()` for mock control variables:

```ts
const mockIsCloud = vi.hoisted(() => ({ value: false }))
vi.mock('./types', () => ({
  get isCloud() { return mockIsCloud.value }
}))
```

src/stores/queueStore.ts
```diff
@@ -91,6 +92,11 @@ export class ResultItemImpl {
     return api.apiURL('/view?' + this.urlParams)
   }
 
+  get previewUrl(): string {
```
christian-byrne, line 95: nitpick (non-blocking): the `previewUrl` getter gates on `this.isImage` (gif/webp suffixes or `mediaType === 'images'`), then calls `getCloudResParam` which has its own `IMAGE_EXTENSIONS` set with different coverage. The two classification systems diverge - e.g. `isImageBySuffix` only covers gif/webp while `IMAGE_EXTENSIONS` has 8 entries. This works today because `isImage` also falls through to `mediaType`, but the split makes it harder to reason about which files get the `&res` param.

christian-byrne (commented):
Also worth noting: `OutputHistoryItem.vue:20` renders a 40x40 thumbnail via `output.url` (full resolution). Now that `previewUrl` exists to separate preview from download URLs, this seems like a candidate for `output.previewUrl`. Same for `assetsStore.test.ts:133` which still maps `preview_url: output.url` in the mock helper.

src/stores/queueStore.ts
dante01yoon: @christian-byrne Thank you for feedbacks. I'll address in separete PR

PR #9299, opened 2026-02-28
fix: handle clipboard errors in Copy Image and useCopyToClipboard

christian-byrne:
## Summary

Fix unhandled promise rejection ("Document is not focused") in Copy Image and improve clipboard fallback reliability.

## Changes

- **What**: Two clipboard fixes:
  1. `litegraphService.ts`: The "Copy Image" context menu passed async `writeImage` as a callback to `canvas.toBlob()` without awaiting — errors became unhandled promise rejections reported in [Sentry CLOUD-FRONTEND-STAGING-AQ](https://comfy-org.sentry.io/issues/6948073569/). Extracted `convertToPngBlob` helper that wraps `toBlob` in a proper Promise so errors propagate to the existing outer try/catch and surface as a user-facing toast instead of a silent Sentry error.
  2. `useCopyToClipboard.ts`: Replaced `useClipboard({ legacy: true })` with explicit modern→legacy fallback that checks `document.execCommand('copy')` return value. VueUse's `legacyCopy` sets `copied.value = true` regardless of whether `execCommand` succeeded, causing false success toasts.

## Review Focus

- The `convertToPngBlob` helper does the same canvas→PNG work as the old inline code but properly awaited
- The happy path (PNG clipboard write succeeds first try) is unchanged
- No public API surface changes — verified zero custom node dependencies via ecosystem code search

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9299-fix-handle-clipboard-errors-in-Copy-Image-and-useCopyToClipboard-3156d73d3650817c8608cba861ee64a9) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useCopyToClipboard.test.ts
+++ b/src/composables/useCopyToClipboard.test.ts
@@ -1,4 +1,4 @@
-import { ref } from 'vue'
+import { computed, ref } from 'vue'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 const mockCopy = vi.fn()
@@ -8,7 +8,7 @@ vi.mock('@vueuse/core', () => ({
   useClipboard: vi.fn(() => ({
     copy: mockCopy,
     copied: ref(false),
-    isSupported: ref(true)
+    isSupported: computed(() => true)
   }))
 }))
 
@@ -31,7 +31,7 @@ describe('useCopyToClipboard', () => {
     vi.mocked(useClipboard).mockReturnValue({
       copy: mockCopy,
       copied: ref(false),
-      isSupported: ref(true),
+      isSupported: computed(() => true),
       text: ref('')
     })
   })
@@ -77,7 +77,7 @@ describe('useCopyToClipboard', () => {
     vi.mocked(useClipboard).mockReturnValue({
       copy: mockCopy,
       copied: ref(false),
-      isSupported: ref(false),
+      isSupported: computed(() => false),
       text: ref('')
     })
     document.execCommand = vi.fn(() => true)
--- a/src/composables/useCopyToClipboard.test.ts
+++ b/src/composables/useCopyToClipboard.test.ts
@@ -1,0 +1,94 @@
+import { ref } from 'vue'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+const mockCopy = vi.fn()
+const mockToastAdd = vi.fn()
+
+vi.mock('@vueuse/core', () => ({
+  useClipboard: vi.fn(() => ({
+    copy: mockCopy,
+    copied: ref(false),
+    isSupported: ref(true)
+  }))
+}))
+
+vi.mock('primevue/usetoast', () => ({
+  useToast: vi.fn(() => ({
+    add: mockToastAdd
+  }))
+}))
+
+vi.mock('@/i18n', () => ({
+  t: (key: string) => key
+}))
+
+import { useClipboard } from '@vueuse/core'
+import { useCopyToClipboard } from '@/composables/useCopyToClipboard'
+
+describe('useCopyToClipboard', () => {
+  beforeEach(() => {
+    vi.resetAllMocks()
+    vi.mocked(useClipboard).mockReturnValue({
+      copy: mockCopy,
+      copied: ref(false),
+      isSupported: ref(true),
+      text: ref('')
+    })
+  })
+
+  it('shows success toast when modern clipboard succeeds', async () => {
+    mockCopy.mockResolvedValue(undefined)
+
+    const { copyToClipboard } = useCopyToClipboard()
+    await copyToClipboard('hello')
+
+    expect(mockCopy).toHaveBeenCalledWith('hello')
+    expect(mockToastAdd).toHaveBeenCalledWith(
+      expect.objectContaining({ severity: 'success' })
+    )
+  })
+
+  it('falls back to legacy when modern clipboard fails', async () => {
+    mockCopy.mockRejectedValue(new Error('Not allowed'))
+    document.execCommand = vi.fn(() => true)
+
+    const { copyToClipboard } = useCopyToClipboard()
+    await copyToClipboard('hello')
+
+    expect(document.execCommand).toHaveBeenCalledWith('copy')
+    expect(mockToastAdd).toHaveBeenCalledWith(
+      expect.objectContaining({ severity: 'success' })
+    )
+  })
+
+  it('shows error toast when both modern and legacy fail', async () => {
+    mockCopy.mockRejectedValue(new Error('Not allowed'))
+    document.execCommand = vi.fn(() => false)
+
+    const { copyToClipboard } = useCopyToClipboard()
+    await copyToClipboard('hello')
+
+    expect(mockToastAdd).toHaveBeenCalledWith(
+      expect.objectContaining({ severity: 'error' })
+    )
+  })
+
+  it('falls through to legacy when isSupported is false', async () => {
+    vi.mocked(useClipboard).mockReturnValue({
+      copy: mockCopy,
+      copied: ref(false),
+      isSupported: ref(false),
+      text: ref('')
+    })
+    document.execCommand = vi.fn(() => true)
+
+    const { copyToClipboard } = useCopyToClipboard()
+    await copyToClipboard('hello')
+
+    expect(mockCopy).not.toHaveBeenCalled()
+    expect(document.execCommand).toHaveBeenCalledWith('copy')
+    expect(mockToastAdd).toHaveBeenCalledWith(
+      expect.objectContaining({ severity: 'success' })
+    )
+  })
+})
--- a/src/composables/useCopyToClipboard.ts
+++ b/src/composables/useCopyToClipboard.ts
@@ -29,7 +29,7 @@ export function useCopyToClipboard() {
     try {
       if (isSupported.value) {
         await copy(text)
-        success = copied.value
+        success = true
       }
     } catch {
       // Modern clipboard API failed, fall through to legacy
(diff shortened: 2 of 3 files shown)


src/services/litegraphService.ts
```diff
@@ -74,6 +74,49 @@ import { getOrderedInputSpecs } from '@/workbench/utils/nodeDefOrderingUtil'
 import { useExtensionService } from './extensionService'
 import { useMaskEditor } from '@/composables/maskeditor/useMaskEditor'
 
+async function convertToPngBlob(
```
dante01yoon: nit: The convertToPngBlob helper takes a Blob but the name says "convert to PNG" — consider  naming clarity

src/services/litegraphService.ts
```diff
@@ -74,6 +74,49 @@ import { getOrderedInputSpecs } from '@/workbench/utils/nodeDefOrderingUtil'
...
+  blob: Blob,
+  width: number,
+  height: number
+): Promise<Blob> {
+  const canvas = $el('canvas', { width, height }) as HTMLCanvasElement
```
dante01yoon, line 83: nit: depending on the implementation of $el, this may or may not append the canvas to the DOM. If it does append it, there's no cleanup. If it just creates a detached  element, it's fine (GC handles it). Worth verifying that $el doesn't append.

dante01yoon (dismissed):
approve with minor comments

src/composables/useCopyToClipboard.ts
```diff
@@ -3,34 +3,59 @@ import { useToast } from 'primevue/usetoast'
...
-        })
-      } else {
-        toast.add({
-          severity: 'error',
-          summary: t('g.error'),
-          detail: t('clipboard.errorMessage')
-        })
+      if (isSupported.value) {
+        await copy(text)
+        success = copied.value
```
dante01yoon: **[quality] medium Priority**

**Issue**: `copied.value` is read synchronously after `await copy(text)`, coupling success detection to VueUse's internal ref-update timing. VueUse's `useClipboard` sets `copied` to `true` then resets it after a timeout — if the timing ever changes (e.g. nextTick), this silently breaks.

**Suggestion**: Since `copy()` rejects on failure and resolves on success, simply set `success = true` after the await:
```ts
if (isSupported.value) {
  await copy(text)
  success = true
}
```
This makes the success signal self-contained and removes the dependency on VueUse's ref lifecycle.

src/composables/useCopyToClipboard.ts
```diff
@@ -3,34 +3,59 @@ import { useToast } from 'primevue/usetoast'
...
+    textarea.remove()
+  }
+}
+
 export function useCopyToClipboard() {
-  const { copy, copied } = useClipboard({ legacy: true })
+  const { copy, copied, isSupported } = useClipboard()
   const toast = useToast()
 
   async function copyToClipboard(text: string) {
```
dante01yoon, line 26: **[quality] medium Priority**

**Issue**: No unit tests exist for `useCopyToClipboard`. This PR significantly changes the composable's behavior (switching from VueUse's built-in legacy mode to a custom modern→legacy fallback chain), but there are no regression tests to validate:
1. Modern clipboard succeeds → success toast
2. Modern clipboard fails → legacy fallback succeeds → success toast
3. Both fail → error toast
4. `isSupported` is false → falls through to legacy

**Context**: The composable is used in ~10 places across the codebase and the PR description notes VueUse's `legacyCopy` had a false-positive bug — exactly the kind of subtle behavior that tests catch.

**Suggestion**: Add a `useCopyToClipboard.test.ts` covering these four paths. The mocking surface is small (VueUse's `useClipboard`, `document.execCommand`, PrimeVue's `useToast`).

src/composables/useCopyToClipboard.ts
```diff
@@ -3,34 +3,59 @@ import { useToast } from 'primevue/usetoast'
 
 import { t } from '@/i18n'
 
+function legacyCopy(text: string): boolean {
+  const textarea = document.createElement('textarea')
+  textarea.setAttribute('readonly', '')
+  textarea.value = text
+  textarea.style.position = 'absolute'
+  textarea.style.opacity = '0'
```
dante01yoon: **[quality] low Priority**

**Issue**: The textarea is positioned with `opacity: '0'` which still occupies space in the accessibility tree and can receive focus in some screen readers. The more common pattern uses `left: '-9999px'` or `clip-path: inset(50%)` to fully move it off-screen.

**Suggestion**: Minor robustness improvement:
```ts
textarea.style.position = 'fixed'
textarea.style.left = '-9999px'
textarea.style.top = '-9999px'
```

dante01yoon (commented):
# Comprehensive PR Review

This review is generated by Claude. It may not always be accurate, as with human reviewers. If you believe that any of the comments are invalid or incorrect, please state why for each. For others, please implement the changes in one way or another.

## Review Summary

**PR**: fix: handle clipboard errors in Copy Image and useCopyToClipboard (#9299)
**Impact**: 99 additions, 65 deletions across 2 files

### Issue Distribution
- Critical: 0
- High: 0
- Medium: 2
- Low: 1

### Category Breakdown
- Architecture: 0 issues
- Security: 0 issues
- Performance: 0 issues
- Code Quality: 3 issues

## Key Findings

### Architecture & Design

The PR is well-scoped — two closely-related clipboard fixes in two files. The extraction of `convertToPngBlob` as a standalone helper is a good separation of concerns. It moves canvas-based PNG conversion out of the deeply nested context menu callback, making both pieces more readable and testable.

The explicit modern→legacy fallback in `useCopyToClipboard` is architecturally sound — it replaces VueUse's `{ legacy: true }` mode which silently swallowed `execCommand` failures, causing false success toasts.

### Code Quality

1. **`copied.value` coupling** (medium): Reading VueUse's internal `copied` ref to determine success is fragile. Since `copy()` rejects on failure, `success = true` after the await is simpler and self-contained.
2. **Missing tests** (medium): The composable is used in ~10 components and the behavior change is non-trivial. Unit tests covering the four fallback paths would prevent regression.
3. **Textarea positioning** (low): Using `opacity: '0'` instead of off-screen positioning for the legacy copy textarea.

### Security Considerations

No security concerns. The clipboard API usage is standard. The `legacyCopy` function creates and immediately removes a textarea — no persistent DOM injection.

### Performance Impact

Negligible. The `convertToPngBlob` helper creates a transient canvas (verified `$el` does NOT append to DOM unless `parent` prop is passed). `ImageBitmap.close()` is properly called in the finally block, preventing memory leaks from retained GPU resources.

## Positive Observations

- Eliminates 8 `@ts-expect-error` suppressions — significant type safety improvement
- The `toBlob` callback is now properly wrapped in a Promise, fixing the root Sentry issue (fire-and-forget async callback)
- `ImageBitmap` cleanup via `.close()` is correctly handled in a `finally` block with a feature check
- Error messages in the helper are descriptive and actionable
- The happy path (PNG clipboard write succeeds first try) is unchanged — no risk to the common case

## References

- [VueUse useClipboard source](https://github.com/vueuse/vueuse/blob/main/packages/core/useClipboard/index.ts) — confirms `copied` ref timing behavior
- [Sentry CLOUD-FRONTEND-STAGING-AQ](https://comfy-org.sentry.io/issues/6948073569/) — root issue

## Next Steps

1. Address `copied.value` → `success = true` simplification
2. Add unit tests for `useCopyToClipboard` covering the four fallback paths
3. Consider i18n for the pre-existing hardcoded `'Copy Image'` string (follow-up)

---

_This is a comprehensive automated review. For architectural decisions requiring human judgment, please request additional manual review._

src/services/litegraphService.ts
christian-byrne: renamed to `reencodeAsPngBlob` in 37e1134

src/services/litegraphService.ts
christian-byrne: checked - `$el` (src/scripts/ui.ts) only appends to the DOM when a `parent` prop is passed (lines 87-89). the call here is `$el('canvas', { width, height })` with no `parent`, so it creates a detached element that gets GC'd normally. no cleanup needed.

src/composables/useCopyToClipboard.ts
christian-byrne: good catch, changed to `success = true` after awaiting `copy()` - no longer depends on VueUse ref timing. c125ab2

src/composables/useCopyToClipboard.ts
christian-byrne: added tests in src/composables/useCopyToClipboard.test.ts covering 4 paths: modern clipboard success, legacy fallback on modern failure, both fail showing error toast, and isSupported=false falling through to legacy. 5f1f510

src/composables/useCopyToClipboard.ts
christian-byrne: updated to `position: fixed; left: -9999px; top: -9999px` - moves it completely off-screen and out of the accessibility tree. e55b7c6

dante01yoon (dismissed):
Issue addressed. please check lint failed

src/composables/useCopyToClipboard.test.ts
```diff
@@ -0,0 +1,94 @@
...
+
+describe('useCopyToClipboard', () => {
+  beforeEach(() => {
+    vi.resetAllMocks()
+    vi.mocked(useClipboard).mockReturnValue({
+      copy: mockCopy,
+      copied: ref(false),
+      isSupported: ref(true),
+      text: ref('')
+    })
```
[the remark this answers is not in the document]
christian-byrne: fixed in 4e82ab9 - changed all `ref()` mocks to `computed()` to match the `ComputedRef` type

src/composables/useCopyToClipboard.ts
```diff
@@ -3,34 +3,60 @@ import { useToast } from 'primevue/usetoast'
...
+  try {
+    return document.execCommand('copy')
+  } finally {
+    textarea.remove()
+  }
+}
+
 export function useCopyToClipboard() {
-  const { copy, copied } = useClipboard({ legacy: true })
+  const { copy, copied, isSupported } = useClipboard()
```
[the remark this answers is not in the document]
christian-byrne: fixed in 34a803b - removed unused `copied` destructuring

PR #9302, opened 2026-02-28
fix: return undefined for muted node output resolution

christian-byrne:
## Summary

Muted (NEVER mode) subgraph nodes throw "No inner node DTO found" during prompt serialization because `resolveOutput()` falls through to subgraph resolution for nodes whose inner DTOs were never registered.

## Changes

- **What**: Add early return in `ExecutableNodeDTO.resolveOutput()` for `NEVER` mode nodes, matching the existing `BYPASS` mode guard. Add 5 tests covering muted, bypassed, and normal mode resolution.

## Review Focus

The fix is a single-line early return. The key insight is that `graphToPrompt` in `executionUtil.ts` correctly skips `getInnerNodes()` for muted/bypassed nodes, so their inner DTOs are never in the map — but `resolveOutput()` was missing the corresponding guard for `NEVER` mode.

Fixes #8986

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9302-fix-return-undefined-for-muted-node-output-resolution-3156d73d3650811e9697c7281f11cf96) by [Unito](https://www.unito.io)


Myestery (dismissed):

jaeone94 (commented):
`ExecutableGroupNodeDTO` overrides `resolveOutput` entirely without calling `super.resolveOutput()`. The override has a `BYPASS` guard (line 28) but **no `NEVER` guard**. 

If a downstream node traces a link back through a muted group node, the override will attempt to resolve internal links via `this.node.outputs[slot]?.links?.at(0)` and throw `"Failed to get link for group node..."` instead of returning `undefined`.

This is the exact same class of bug being fixed for `ExecutableNodeDTO` — just in the group node variant. The same defensive guard should be added:

`src/utils/executableGroupNodeDto.ts`, line 26
```typescript
override resolveOutput(slot: number, type: ISlotType, visited: Set<string>) {
    // Muted nodes produce no output
    if (this.mode === LGraphEventMode.NEVER) return

    // Temporary duplication: Bypass nodes are bypassed using the first input with matching type
    if (this.mode === LGraphEventMode.BYPASS) {
```

christian-byrne:
good catch - added the NEVER mode guard to `ExecutableGroupNodeDTO.resolveOutput()` in 7316077, along with a test case for group node muting. same pattern as the `ExecutableNodeDTO` fix.

jaeone94 (approved):
LGTM

mmartial:
Thank you. 
Is there a way to confirm that the fix is in place by trying either to update Comfy or using a specific value for `--front-end-version` ?

I see 1.42.4 does not have it yet and `--front-end-version Comfy-Org/ComfyUI_frontend@git ` shows that I am on 1.41.18

christian-byrne:
After around 2pm PST tomorrow there will be a new nightly version you can verify with using `--front-end-version Comfy-Org/ComfyUI_frontend@latest`

mmartial:
> After around 2pm PST tomorrow there will be a new nightly version you can verify with using `--front-end-version Comfy-Org/ComfyUI_frontend@latest`

Thank you. I will test then, and report.

DemonAlone:
@christian-byrne, I have tried the  latest (1.42.4), there is no effect at all.

mmartial:
> @christian-byrne, I have tried the latest (1.42.4), there is no effect at all.

If you look at the delta between 1.42.4 and main, it is in main and not yet in a published release.
Christian mentioned we should be able to test in a few hours using the `@latest` solution.

mmartial:
I see 1.42.5 is out, I do not have the DTO error anymore. Thank you

DemonAlone:
> I see 1.42.5 is out, testing

I have already tried it, it works for me.

PR #9303, opened 2026-02-28
fix: RAF-batch WebSocket progress events

christian-byrne:
## What

RAF-batch WebSocket `progress` and `progress_state` events to coalesce multiple updates per animation frame.

## Why

Each WebSocket progress message immediately updates reactive state, triggering Vue re-renders. Profiler shows 339 markers from `onProgressUpdate`. During rapid execution, multiple progress events per frame cause redundant reactive update → watcher → setDirty cycles.

## How

Defer reactive state updates to `requestAnimationFrame` callbacks. Multiple progress events within the same frame are coalesced — only the latest value is applied. Pending RAFs are cancelled on `unbindExecutionEvents`.

## Verification

- [x] Unit tests verifying RAF batching behavior (7 new tests)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

## Perf Impact

Reduces per-frame reactive updates during workflow execution from N (one per WebSocket message) to 1.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9303-fix-RAF-batch-WebSocket-progress-events-3156d73d365081959d93e85a4b8d38c4) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/executionStore.test.ts
+++ b/src/stores/executionStore.test.ts
@@ -1,8 +1,9 @@
 import { setActivePinia } from 'pinia'
-import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { nextTick } from 'vue'
 
 import { app } from '@/scripts/app'
+import { api } from '@/scripts/api'
 import { MAX_PROGRESS_JOBS, useExecutionStore } from '@/stores/executionStore'
 import { useExecutionErrorStore } from '@/stores/executionErrorStore'
 import { useMissingNodesErrorStore } from '@/platform/nodeReplacement/missingNodesErrorStore'
@@ -460,14 +461,21 @@ describe('useExecutionStore - nodeProgressStatesByJob eviction', () => {
     handler(
       new CustomEvent('progress_state', { detail: { nodes, prompt_id: jobId } })
     )
+    // Flush the RAF so the batched update is applied immediately
+    vi.advanceTimersByTime(16)
   }
 
   beforeEach(() => {
+    vi.useFakeTimers()
     vi.clearAllMocks()
     apiEventHandlers.clear()
     setActivePinia(createTestingPinia({ stubActions: false }))
     store = useExecutionStore()
     store.bindExecutionEvents()
+  })
+
+  afterEach(() => {
+    vi.useRealTimers()
   })
 
   it('should retain entries below the limit', () => {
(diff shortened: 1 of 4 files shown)


src/stores/executionStore.ts
```diff
@@ -211,6 +211,17 @@ export const useExecutionStore = defineStore('execution', () => {
...
+    if (_progressRafId !== null) {
+      cancelAnimationFrame(_progressRafId)
+      _progressRafId = null
+      _pendingProgress = null
+    }
+    if (_progressStateRafId !== null) {
+      cancelAnimationFrame(_progressStateRafId)
+      _progressStateRafId = null
+      _pendingProgressState = null
+    }
```
jaeone94: This RAF cleanup only exists in `unbindExecutionEvents`, but `resetExecutionState()` (line 465) also clears `_executingNodeProgress.value` and `nodeProgressStates.value` without cancelling pending RAFs. This creates a new race condition:

1. A progress WebSocket event arrives and schedules a RAF
2. In the same frame, an execution-end event fires and calls `resetExecutionState()`, clearing all state
3. On the next frame, the pending RAF callback fires and writes stale progress data back into the cleared state

The same applies to `handleExecuting` (line 264) which clears `_executingNodeProgress.value` without cancelling `_progressRafId`.

`unbindExecutionEvents` already has the correct cleanup pattern (lines 214-224). The same cancellation should happen in `resetExecutionState` (and `handleExecuting` for `_progressRafId`).

src/stores/executionStore.ts
```diff
@@ -265,8 +276,24 @@ export const useExecutionStore = defineStore('execution', () => {
...
+    if (_progressStateRafId === null) {
+      _progressStateRafId = requestAnimationFrame(() => {
+        _progressStateRafId = null
+        if (_pendingProgressState) {
+          _applyProgressState(_pendingProgressState)
+          _pendingProgressState = null
+        }
+      })
+    }
+  }
```
jaeone94, line 576: The codebase already has `src/utils/rafBatch.ts` which exports `createRafBatch` — a reusable RAF batching utility with `schedule`, `cancel`, `flush`, and `isScheduled` methods. It's already used in `useSlotLinkInteraction.ts` and `useSlotElementTracking.ts`.

The four raw `let` variables (`_pendingProgress`, `_progressRafId`, `_pendingProgressState`, `_progressStateRafId`) and the manual `requestAnimationFrame`/`cancelAnimationFrame` calls could be replaced with two `createRafBatch` instances. This would:

- Reduce boilerplate
- Simplify cleanup in `unbindExecutionEvents` and `resetExecutionState`

src/stores/executionStore.test.ts
```diff
@@ -442,3 +464,171 @@ describe('useExecutionErrorStore - setMissingNodeTypes', () => {
     expect(store.missingNodesError?.nodeTypes).toEqual(input)
   })
 })
+
+describe('useExecutionStore - RAF batching', () => {
```
jaeone94, line 1489: nit: 
Related to Comment about  RAF cleanup: there's no test covering the scenario where a progress event is pending when execution ends (via `handleExecutionSuccess`/`handleExecutionError`/`handleExecutionInterrupted`). Adding a test like "pending RAF is discarded when execution completes" would document the expected behavior and catch the race condition.

src/stores/executionStore.ts
christian-byrne: Fixed in 222845d5b — replaced the four raw `let` variables and manual `requestAnimationFrame`/`cancelAnimationFrame` calls with two `createRafBatch` instances from `src/utils/rafBatch.ts`.

src/stores/executionStore.test.ts
christian-byrne: Fixed in 139ffcc41 — added 4 tests covering pending RAF discard on `execution_success`, `execution_error`, and `execution_interrupted` for both progress and progress_state events.

src/stores/executionStore.ts
christian-byrne: Fixed in 69a077241 — added `progressBatch.cancel()` / `progressStateBatch.cancel()` calls in both `resetExecutionState` and `handleExecuting` to prevent stale RAF callbacks from writing data back after state is cleared.

jaeone94 (dismissed):
Solid update. All three issues from my previous review are addressed:

1. **Race condition fixed** — `cancelPendingProgressUpdates()` is now called in `resetExecutionState`, `handleExecuting`, and `unbindExecutionEvents`. The stale-RAF-writes-back-after-clear scenario is no longer possible.

2. **Uses existing utility** — Instead of four raw `let` variables with manual `requestAnimationFrame`/`cancelAnimationFrame`, the code now builds on `createRafBatch`. The new `createRafCoalescer<T>` wrapper adds a clean last-write-wins buffer with type safety. Cleanup is a single `.cancel()` call — harder to get wrong.

3. **Tests cover the edge cases** — The new tests verify RAF discard on `execution_success`, `execution_error`, and `execution_interrupted`, plus unbind cleanup. The coalescer utility has its own isolated test suite.

The structure is noticeably better than the initial version. Splitting the generic coalescing logic into `rafBatch.ts` and keeping the store focused on business logic is a good separation. The cancel points are thorough and the test coverage matches.

LGTM.

christian-byrne (commented):
This needs a lot more tests, including e2e tests, and qa. And we need to consider whether there's a better, more builtin, or library provided way to do the RAF with the higher order private function approach which is awkward and hard to read.

src/stores/executionStore.test.ts
```diff
@@ -453,16 +454,23 @@ describe('useExecutionStore - nodeProgressStatesByJob eviction', () => {
     handler(
       new CustomEvent('progress_state', { detail: { nodes, prompt_id: jobId } })
     )
+    // Flush the RAF so the batched update is applied immediately
+    vi.advanceTimersByTime(16)
```
AustinMroz: Probably silly since the test works, but is `16` sufficient?  `16` is strictly less than `1000/60`.

Should this be `vi.advanceTimersToNextFrame()`?

src/stores/executionStore.test.ts
```diff
@@ -1298,6 +1306,312 @@ describe('useMissingNodesErrorStore - setMissingNodeTypes', () => {
   })
 })
 
+describe('useExecutionStore - RAF batching', () => {
+  let store: ReturnType<typeof useExecutionStore>
+
+  function getRegisteredHandler(eventName: string) {
+    const calls = vi.mocked(api.addEventListener).mock.calls
+    const call = calls.find(([name]) => name === eventName)
+    return call?.[1] as (e: CustomEvent) => void
```
AustinMroz:
```suggestion
    return call![1]!
```

AustinMroz (dismissed):
Lots of potential for edge cases with custom nodes, but I think the changes are safe.
- Only `progress` and `progressState` are affected
- The raw events sent that pass through `api` aren't affected, only consumers on the other side of the `executionStore`
- On the boundry of node execution, a progress event on a new node may block the progress event on a prior node, but I think this is always fine.
  - `progress_text` would be particularly vulnerable here, but it's not batched

jaeone94 (dismissed):
I re-reviewed this PR at the latest HEAD. Overall, the implementation looks solid and the test coverage is strong. LGTM.

One non-blocking test suggestion: it may be worth adding a regression test that queues a pending legacy `progress` update, fires an `executing` event before the RAF is flushed, and then verifies that the stale progress is not applied.

The existing `executing` test verifies that `_executingNodeProgress` is cleared synchronously, but it would still pass if the newly added `progressCoalescer.cancel()` call were removed. A pending-RAF test would directly cover that cancellation behavior.

One very narrow edge case I noticed—not suggesting a code change for this PR:

Since `_applyProgressState` performs preview cleanup on `running`, last-write-wins batching can skip that cleanup if `running → finished` occurs before the next RAF.

For this to be observable, a previous run must have left a preview, the next run must produce no new preview, and it must finish within the same RAF window. This seems rare enough that handling it may be unnecessary.

src/stores/executionStore.test.ts
christian-byrne: good catch, 16 was cutting it close against 1000/60. switched to vi.advanceTimersToNextFrame() in 358af36a

src/stores/executionStore.test.ts
christian-byrne: applied in 15e573b3

PR #9304, opened 2026-02-28
fix: avoid forced layout in renderInfo by using canvas.height

christian-byrne:
## What

Replace `canvas.offsetHeight` with `canvas.height / devicePixelRatio` in `renderInfo` to avoid forced synchronous layout.

## Why

`renderInfo` is called ~2,631 times in a typical session. Each call reads `this.canvas.offsetHeight`, which forces the browser to flush pending style/layout changes synchronously. With PrimeVue injecting styles dynamically and Vue patching the DOM, there are almost always pending mutations — converting every canvas-only `renderInfo` call into a forced layout.

## How

`canvas.height` is the DPR-scaled internal resolution (set in `resizeCanvas` as `cssHeight * devicePixelRatio`). Dividing by `devicePixelRatio` yields the same CSS pixel value as `offsetHeight` without triggering layout.

## Verification

- [x] Unit test: verifies `offsetHeight` is not accessed when y is provided
- [x] Unit test: verifies fallback uses `canvas.height / devicePixelRatio`
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] All litegraph tests pass (538 passed)

## Perf Impact

Eliminates ~2,631 forced synchronous layouts per session from the canvas info panel.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9304-fix-avoid-forced-layout-in-renderInfo-by-using-canvas-height-3156d73d36508171973dda289b30d5ee) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphCanvas.renderInfo.test.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.renderInfo.test.ts
@@ -1,0 +1,54 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import {
+  LGraph,
+  LGraphCanvas
+} from '@/lib/litegraph/src/litegraph'
+
+describe('LGraphCanvas.renderInfo', () => {
+  let lgCanvas: LGraphCanvas
+  let ctx: CanvasRenderingContext2D
+
+  beforeEach(() => {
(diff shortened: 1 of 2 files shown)


pythongosssss:
lgtm

PR #9315, opened 2026-03-01
feat: add ever-present upgrade button for free-tier users

huntcsg:
## Summary

Add persistent upgrade CTAs for free-tier users: a topbar button and "Upgrade to add credits" replacing "Add Credits" in popovers and settings panels.

## Changes

- **What**:
  - New `TopbarSubscribeButton` component in both GraphCanvas and LinearView topbars, visible only to free-tier users
  - Profile popover (legacy + workspace): free-tier users see "Upgrade to add credits" instead of "Add Credits", linking directly to the pricing table
  - Manage Plan settings (legacy + workspace): same replacement — free-tier users see "Upgrade to add credits" instead of "Add Credits"
  - Paid-tier users retain the original "Add Credits" behavior in all locations
  - All upgrade buttons go directly to the pricing table (one-step flow)

## Review Focus

- The `isFreeTier` conditional gating on the buttons — ensure free-tier users see upgrade CTAs and paid users see normal Add Credits
- Layout in Manage Plan panels uses `flex flex-col gap-3` to stack the upgrade button below the usage history link instead of side-by-side

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9315-feat-add-ever-present-upgrade-button-for-free-tier-users-3166d73d365081228cdfe6a67fec6aec) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/topbar/CurrentUserPopoverLegacy.vue
+++ b/src/components/topbar/CurrentUserPopoverLegacy.vue
@@ -46,6 +46,16 @@
         class="icon-[lucide--circle-help] cursor-help text-base text-muted-foreground mr-auto"
       />
       <Button
+        v-if="isFreeTier"
+        variant="gradient"
+        size="sm"
+        data-testid="upgrade-to-add-credits-button"
+        @click="handleUpgradeToAddCredits"
+      >
+        {{ $t('subscription.upgradeToAddCredits') }}
+      </Button>
+      <Button
+        v-else
         variant="secondary"
         size="sm"
         class="text-base-foreground"
@@ -61,7 +71,7 @@
         :fluid="false"
         :label="$t('subscription.subscribeToComfyCloud')"
         size="sm"
-        variant="gradient"
+        button-variant="gradient"
         @subscribed="handleSubscribed"
       />
     </div>
@@ -170,6 +180,7 @@ const settingsDialog = useSettingsDialog()
 const dialogService = useDialogService()
 const {
   isActiveSubscription,
+  isFreeTier,
   subscriptionTierName,
   subscriptionTier,
   fetchStatus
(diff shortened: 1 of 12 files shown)


src/components/topbar/CurrentUserPopoverLegacy.vue
```diff
@@ -46,6 +46,21 @@
...
       />
       <Button
+        v-if="isFreeTier"
+        variant="primary"
+        size="sm"
+        :style="{
+          background: 'var(--color-subscription-button-gradient)',
+          color: 'var(--color-white)',
+          borderColor: 'transparent'
+        }"
```
DrJKL: Nit: Could you use tailwind classes for these?

src/components/topbar/TopbarSubscribeButton.vue
```diff
@@ -0,0 +1,30 @@
...
+  <Button
+    v-if="isFreeTier"
+    class="mr-2 shrink-0 whitespace-nowrap"
+    variant="primary"
+    size="sm"
+    :style="{
+      background: 'var(--color-subscription-button-gradient)',
+      color: 'var(--color-white)',
+      borderColor: 'transparent'
+    }"
```
DrJKL: Same here.
Also, could this component be used in both places?

DrJKL: Actually, given how many times we repeat this, this should probably be a Button Variant.

DrJKL (changes requested):

src/components/topbar/TopbarSubscribeButton.vue
DrJKL: That should simplify all of the usages to just be variant="subscribe-gradient" or something similar, maybe with different sizes

AustinMroz (approved):
Got a little nerd-sniped trying to do everything inline, but I'm happy with the latest batch of style changes.

DrJKL (approved):

PR #9318, opened 2026-03-01
fix: restore correct workflow on page reload

### LINKED ISSUE #9317: fix: incorrect workflow loaded from draft storage on page reload
## Bug

When a user refreshes the page, a different workflow loads instead of the one they were working on.

### Root Cause

Two interacting bugs in `useWorkflowPersistenceV2.ts`:

1. **Draft cleanup removes active workflow's draft** (line 124-127): Saved, unmodified workflows have their drafts removed during persistence. On reload, the session path lookup fails and `fallbackToLatestDraft` loads a completely different workflow.

2. **Tab restoration never activates the correct workflow** (line 274-277): `restoreWorkflowTabsState()` calls `openWorkflowsInBackground()` which only adds paths to the tab bar but never calls `openWorkflow()` on the active tab.

### Reproduction

1. Open two workflows: A (saved, unmodified) as active tab, B (modified) as background tab
2. Wait for persistence debounce (512ms)
3. Refresh the page
4. Workflow B is loaded instead of A

### Related Issues

- https://github.com/Comfy-Org/ComfyUI/issues/12628
- https://github.com/Comfy-Org/ComfyUI/issues/10225
- https://github.com/comfyanonymous/ComfyUI/issues/5095
- https://github.com/Comfy-Org/ComfyUI_frontend/issues/5465
- https://github.com/Comfy-Org/desktop/issues/1435
- https://github.com/Comfy-Org/ComfyUI_frontend/issues/5848

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9317-fix-incorrect-workflow-loaded-from-draft-storage-on-page-reload-3166d73d3650814b875eca665ba27534) by [Unito](https://www.unito.io)

dante01yoon:
## Summary

- Fixes incorrect workflow loading after page refresh when the active workflow is saved and unmodified
- Adds saved-workflow fallback in `loadPreviousWorkflowFromStorage()` before falling back to latest draft
- Calls `openWorkflow()` in `restoreWorkflowTabsState()` to activate the correct tab after restoration

- Fixes #9317

## Test plan

- [x] Unit tests for saved-workflow fallback (draft missing, saved workflow exists)
- [x] Unit tests for correct tab activation after restoration
- [x] Unit tests for existing behavior preservation (draft preferred over saved, no-session-path fallback)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes
- [x] All 117 persistence tests pass

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9318-fix-restore-correct-workflow-on-page-reload-3166d73d365081ba9139f7c23c917aa4) by [Unito](https://www.unito.io)


dante01yoon:
The FCM test is out of scope this PR, should track in separate issue.
https://github.com/Comfy-Org/ComfyUI_frontend/issues/9319

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
```diff
@@ -275,6 +283,15 @@ export function useWorkflowPersistenceV2() {
...
     })
+
+    // Activate the correct workflow at storedActiveIndex
+    const activePath = storedWorkflows[storedActiveIndex]
+    if (activePath) {
+      const workflow = workflowStore.getWorkflowByPath(activePath)
+      if (workflow) {
+        await useWorkflowService().openWorkflow(workflow)
+      }
+    }
```
DrJKL: Very nit: I like to minimize nesting when feasible.
```suggestion
    if (!activePath) return

    const workflow = workflowStore.getWorkflowByPath(activePath)
    if (!workflow) return
    
    await useWorkflowService().openWorkflow(workflow)
```

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
```diff
@@ -131,19 +131,27 @@ export function useWorkflowPersistenceV2() {
...
     const sessionPath = tabState.getActivePath()
-    if (
-      sessionPath &&
-      (await draftStore.loadPersistedWorkflow({
-        workflowName: null,
-        preferredPath: sessionPath
-      }))
-    ) {
-      return true
+    if (sessionPath) {
```
DrJKL: If you invert this, I know it doesn't follow the 1,2,3 declaration order, but it would reduce nesting and make it clear that the persistence path is directly tied to a lack of session data.

DrJKL (dismissed):

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
dante01yoon: Thanks for the suggestion! Applied the early return pattern in 221c737.

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
dante01yoon: Good call on minimizing nesting! Flipped to an early return — the three-step fallback logic now flows flat without additional nesting. Applied in 221c737.

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
```diff
@@ -133,19 +133,32 @@ export function useWorkflowPersistenceV2() {
...
       return true
     }
 
-    // 2. Fall back to most recent draft
+    // 2. No draft — try loading saved workflow by path
+    const savedWorkflow = workflowStore.getWorkflowByPath(sessionPath)
+    if (savedWorkflow) {
+      await useWorkflowService().openWorkflow(savedWorkflow)
+      return true
+    }
```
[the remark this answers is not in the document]
dante01yoon: Addressed the CodeRabbit major review feedback:

useWorkflowPersistenceV2.ts — keep draft fallback when openWorkflow rejects

Wrapped the openWorkflow(savedWorkflow) call in a try/catch so that a rejection falls through to the step-3 latest-draft recovery path instead of propagating up to initializeWorkflow and loading a blank workflow.

Commit: 60a98d16d

src/renderer/extensions/vueNodes/widgets/components/layout/WidgetLayoutField.vue
DrJKL, line 1: Merge artifact?

dante01yoon: This was likely auto-formatted by `pnpm format` (oxfmt) when editing nearby files. Just a Tailwind class reorder + Tailwind 4 syntax update (`has-[:focus-visible]:` → `has-focus-visible:`). No behavioral change.

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
```diff
@@ -255,10 +275,11 @@ export function useWorkflowPersistenceV2() {
...
     const storedActiveIndex = storedTabState?.activeIndex ?? -1
 
-    tabStateRestored = true
-
     const isRestorable = storedWorkflows.length > 0 && storedActiveIndex >= 0
-    if (!isRestorable) return
+    if (!isRestorable) {
+      tabStateRestored = true
+      return
+    }
```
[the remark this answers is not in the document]
dante01yoon: Line 309 already guards against an out-of-range index with `if (!activePath) return`, so the worst case is just skipping tab activation — no crash or partial restore. The extra  validation would be over-engineering for a self-managed stored value.

src/renderer/extensions/vueNodes/widgets/components/layout/WidgetLayoutField.vue
DrJKL: I wonder why it didn't format Kaili's PR

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
```diff
@@ -132,19 +132,39 @@ export function useWorkflowPersistenceV2() {
   const debouncedPersist = debounce(persistCurrentWorkflow, PERSIST_DEBOUNCE_MS)
 
   const loadPreviousWorkflowFromStorage = async () => {
-    // 1. Try session pointer (for tab restoration)
     const sessionPath = tabState.getActivePath()
+    if (!sessionPath) {
+      return await draftStore.loadPersistedWorkflow({
+        workflowName: null,
+        fallbackToLatestDraft: true
```
christian-byrne: The early-return restructuring duplicates the `fallbackToLatestDraft` call in two places (here and at the end as "last resort"). The original structure was cleaner — just insert the saved-workflow step as step 2 without restructuring:

```typescript
const loadPreviousWorkflowFromStorage = async () => {
  const sessionPath = tabState.getActivePath()

  // 1. Try draft for session path
  if (sessionPath && await draftStore.loadPersistedWorkflow({
    workflowName: null,
    preferredPath: sessionPath
  })) return true

  // 2. Try saved workflow by path (draft may not exist for saved+unmodified workflows)
  if (sessionPath) {
    const saved = workflowStore.getWorkflowByPath(sessionPath)
    if (saved) {
      await useWorkflowService().openWorkflow(saved)
      return true
    }
  }

  // 3. Fall back to most recent draft
  return await draftStore.loadPersistedWorkflow({
    workflowName: null,
    fallbackToLatestDraft: true
  })
}
```

The try/catch around `openWorkflow` is unnecessary — `initializeWorkflow` already wraps the entire call in try/catch and falls back to `loadDefaultWorkflow`.

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
```diff
@@ -280,6 +301,21 @@ export function useWorkflowPersistenceV2() {
...
+
+    // Activate the correct workflow at storedActiveIndex
+    const activePath = storedWorkflows[storedActiveIndex]
+    if (!activePath) return
+
+    const workflow = workflowStore.getWorkflowByPath(activePath)
+    if (!workflow) return
+
+    try {
+      await useWorkflowService().openWorkflow(workflow)
```
christian-byrne, line 302: The `openWorkflow` addition is the right fix, but the guard chain can be simplified:

```typescript
tabStateRestored = true

const activePath = storedWorkflows[storedActiveIndex]
const workflow = activePath ? workflowStore.getWorkflowByPath(activePath) : null
if (workflow) {
  await useWorkflowService().openWorkflow(workflow)
}
```

The try/catch here is also unnecessary — if `openWorkflow` fails, the previously loaded workflow from `initializeWorkflow` stays active, which is acceptable fallback behavior. Swallowing the error silently makes debugging harder.

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.test.ts
```diff
@@ -0,0 +1,360 @@
+import { createTestingPinia } from '@pinia/testing'
```
christian-byrne, line 1: Thanks for adding tests for the new behaviors! One suggestion: the mock setup here is nearly identical to the existing V1 persistence test file — it might be worth extracting shared test fixtures to reduce the boilerplate. Also, tests like "prefers draft over saved workflow" and "falls back to latest draft" cover existing behavior that's already implicitly tested — the new tests could focus on just the two new code paths (saved-workflow fallback and tab activation) to keep things tight. Happy to help with this if useful!

christian-byrne (changes requested):
The two bugs are real and the fixes are directionally correct — nice catch. Requesting some simplifications to reduce the diff surface and keep it minimal.

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
dante01yoon: Good call — simplified to the structure you suggested. Removed the early-return split and the try/catch since `initializeWorkflow` already wraps the whole thing.

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
dante01yoon: Agreed, consolidated the guards into a single ternary and dropped the try/catch.

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.test.ts
dante01yoon: Removed the two tests that were re-verifying existing behavior (`prefers draft over saved workflow`, `falls back to latest draft`). For the shared fixture extraction — opened #9775 to track that separately since it touches multiple test files.

christian-byrne (approved):
LGTM

MaggotHATE:
Hi! The issue still exists and wasn't fixed properly.

PR #9321, opened 2026-03-01
fix: propagate widget disabled state to Vue node components

jtydhr88:
## Summary

Widgets with `widget.disabled = true` (e.g. display-only counters in custom nodes) were editable in Vue node mode despite being correctly greyed out in litegraph mode. The disabled state from the widget store was not being merged into the options passed to Vue widget components.

Before

https://github.com/user-attachments/assets/6957dd86-6eb9-4edb-93ee-50fc5aa5350f

After

https://github.com/user-attachments/assets/d954006f-d7e6-4e7c-9b3c-bcabed0e6260

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9321-fix-propagate-widget-disabled-state-to-Vue-node-components-3166d73d365081a7936aeabe81eb6e15) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/components/NodeWidgets.vue
+++ b/src/renderer/extensions/vueNodes/components/NodeWidgets.vue
@@ -204,9 +204,11 @@ const processedWidgets = computed((): ProcessedWidget[] => {
     // Get value from store (falls back to undefined if not registered)
     const value = widgetState?.value as WidgetValue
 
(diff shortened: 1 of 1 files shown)


DrJKL (approved):
Could you double check the behavior of this with promoted widgets in subgraphs? Especially nested ones.

PR #9322, opened 2026-03-01
fix: rewrite Ogg parser for multi-page packets (#9256)

### LINKED ISSUE #9256: [Bug] Failed to load workflows from long Opus/Ogg files (SyntaxError due to OggS page breaks)
### Description
When dragging and dropping an `.opus` or `.ogg` file containing a saved workflow, if the metadata (or the file itself) is very large, a `SyntaxError: Unexpected token 'O', "OggS"... is not valid JSON` occurs and the workflow fails to load.

### Cause
Ogg container specification splits data into pages (typically around 64KB). When a long workflow JSON is embedded in the OpusTags (Vorbis Comment), `OggS` magic byte headers are inserted into the middle of the JSON string across page boundaries. 
The current regex-based string matching in `src/scripts/metadata/ogg.ts` does not account for these interleaved page headers, causing it to extract a corrupted JSON string containing `OggS`.

### Proposed Solution
Rewrite `getOggMetadata` in `ogg.ts` to properly demux the Ogg container, aggregate the OpusTags packets by discarding the `OggS` page headers and segment tables, and decode the binary data before parsing the JSON.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9256-Bug-Failed-to-load-workflows-from-long-Opus-Ogg-files-SyntaxError-due-to-OggS-page--3136d73d36508107829bd264aed210a3) by [Unito](https://www.unito.io)

xeinherjer-dev:
The previous implementation used naive regex matching on binary data converted to strings. This approach failed entirely when a large prompt caused the `OpusTags` packet to span multiple Ogg pages.

This commit replaces the fragile string search with a robust binary parser that strictly follows the Ogg logical bitstream and Vorbis Comments specifications. It correctly handles Ogg lacing and reassembles segmented packets, ensuring that large JSON metadata is successfully extracted.

Also adds unit tests (ogg.test.ts) to verify the correct parsing of both standard and multi-page Ogg files.

## Summary

Replaced the fragile string-based Ogg metadata extraction with a robust binary parser to successfully load workflows from large multi-page Ogg files.

## Changes

- **What**: 
  - Completely rewrote the Ogg metadata parser to correctly handle Ogg lacing and reassemble segmented `OpusTags` multi-page packets.
  - Refactored file reading logic to use the modern `file.arrayBuffer()` API, replacing `FileReader` and manual Promises to reduce overhead and improve processing efficiency.
  - Added comprehensive unit tests (`ogg.test.ts`).

## Review Focus

Please review the binary parsing logic (specifically the handling of Ogg lacing and multi-page segments) and the newly added test cases in `ogg.test.ts`.

Fixes #9256

- **Unit Test Coverage**: Verified 100% test coverage for `ogg.ts` with 8 test cases covering both standard and multi-page Ogg files.
- **Manual Verification**: Confirmed that dragging and dropping a large multi-page Ogg file successfully loads the expected workflow without errors.

image

## Reference Diagram
ogg_opus_data_flow_en_1772437600957

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9322-fix-rewrite-Ogg-parser-for-multi-page-packets-9256-3166d73d3650819dbaa6e33f958a434a) by [Unito](https://www.unito.io)


xeinherjer-dev:
Note: The PR size increased to XL because I added test codes to cover the new JSON parsing fixes and maintain 100% coverage.

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,155 @@
...
+function extractOpusTags(data: Uint8Array, decoder: TextDecoder): Uint8Array[] {
+  const OGG_HEADER_SIZE = 27 // Base size of the Ogg page header (from magic number to just before segment count)
+  const OGG_PAGE_SEGMENTS_OFFSET = 26 // Offset position where the number of segments in the page is stored
+  const OGG_MAX_SEGMENT_SIZE = 255 // Ogg spec: the maximum size of a single segment is 255 bytes
+
+  const segments: Uint8Array[] = []
+  let offset = 0
+  let inOpusTags = false
+
+  while (offset + OGG_HEADER_SIZE < data.length) {
```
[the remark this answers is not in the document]
xeinherjer-dev: It doesn't change the actual outcome since a 27-byte page has 0 segments anyway, but I'll update it.

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,155 @@
 export async function getOggMetadata(file: File) {
```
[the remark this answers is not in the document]
xeinherjer-dev: Instead of creating a new OggMetadata type, I used the existing ComfyMetadata type to keep things consistent.

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,160 @@
...
+        if (segmentLength < OGG_MAX_SEGMENT_SIZE) {
+          pageProcessingEnded = true
+          break
+        }
+      }
+    }
+
+    if (pageProcessingEnded) break
+    offset = dataOffset
+  }
```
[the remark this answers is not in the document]
xeinherjer-dev: Opus files (RFC 7845) contain exactly one logical stream, so interleaved pages from other streams cannot occur. This parser only handles ComfyUI-generated Opus files, making the multi-stream hardening unnecessary.

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,160 @@
...
+      const key = text.substring(0, separatorIndex)
+      const value = text.substring(separatorIndex + 1)
+      if (key === 'prompt') {
+        try {
+          result.prompt = JSON.parse(value)
+        } catch (e) {
+          console.warn('Ogg metadata parsing failed for prompt:', e)
+        }
+      } else if (key === 'workflow') {
+        try {
```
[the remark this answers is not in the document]
xeinherjer-dev: Both "prompt" and "workflow" keys are lowercase by construction:
"workflow" originates from a JS object property in 
api.ts (extra_pnginfo: { workflow }), written as-is by the backend.
"prompt" is a hardcoded lowercase string literal in the backend Python code.
Neither key can appear in non-lowercase form in ComfyUI-generated Opus files, so case-insensitive matching is not needed. Dismissing.

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,201 @@
...
+ * JSON-encoded prompt and workflow data.
+ *
+ * @param file - The Ogg/Opus file to extract metadata from.
+ * @returns A promise resolving to a {@link ComfyMetadata} object containing
+ *   the parsed `prompt` and `workflow` fields, or `undefined` for each if
+ *   not found or if parsing fails.
+ */
+export async function getOggMetadata(file: File): Promise<ComfyMetadata> {
+  // Read the entire file into memory (Opus files are generally small enough)
+  const arrayBuffer = await file.arrayBuffer()
```
christian-byrne: issue: `file.arrayBuffer()` reads the entire file into memory. Every other metadata parser in this codebase caps reads -- `gltf.ts` at 1 MB, `ebml.ts` at 2 MB, `isobmff.ts` at 64 MB. Ogg metadata lives in the first couple of pages, so reading a full-length audio file is unnecessary.

```ts
const MAX_READ_BYTES = 2 * 1024 * 1024
const arrayBuffer = await file.slice(0, MAX_READ_BYTES).arrayBuffer()
```

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,201 @@
...
-    if (page.match('OggS\u0000')) oggs++
-    header += page
-    if (oggs > 1) break
+    readIndex += commentLength
+
+    const separatorIndex = text.indexOf('=')
+    if (separatorIndex !== -1) {
+      const key = text.substring(0, separatorIndex)
+      const value = text.substring(separatorIndex + 1)
+      if (key === 'prompt') {
```
christian-byrne, line 210: suggestion: RFC 7845 sec 5.2 says Vorbis Comment field names are case-insensitive ASCII. ComfyUI's backend writes lowercase so this works today, but third-party encoders conventionally use uppercase (`TITLE`, `ARTIST`). A `.toLowerCase()` on the key costs nothing and makes this robust.

```ts
const key = text.substring(0, separatorIndex).toLowerCase()
```

src/scripts/metadata/ogg.test.ts
```diff
@@ -0,0 +1,381 @@
...
+  packetView.setUint32(pos, commentKeys.length, true)
+  pos += 4
+  for (const entryBytes of commentEntries) {
+    packetView.setUint32(pos, entryBytes.length, true)
+    pos += 4
+    packetData.set(entryBytes, pos)
+    pos += entryBytes.length
+  }
+
+  // Construct Ogg Pages for Metadata
```
christian-byrne, line 56: suggestion: The test helper builds OpusTags as page 0, but a real Ogg/Opus file always has an OpusHead identification header as page 0 and OpusTags starting at page 1+. Prepending a minimal OpusHead page would make these tests validate that the parser correctly skips non-OpusTags pages before finding the metadata, which is the real-world code path.

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,201 @@
...
+      const segmentLength = data[lengthsOffset + i]
+
+      // Bounds check: ensure the segment data lies within the available data
+      if (dataOffset + segmentLength > data.length) break
+
+      const segment = data.subarray(dataOffset, dataOffset + segmentLength)
+      dataOffset += segmentLength
+
+      if (!inOpusTags) {
+        const segmentMagic = decoder.decode(segment.subarray(0, 8))
```
christian-byrne: suggestion (non-blocking): `segment.subarray(0, 8)` clamps safely per the Uint8Array spec so this won't crash on short segments, but a guard like `segmentLength >= 8` before the decode would make the intent explicit and avoid a pointless decode+compare on tiny segments.

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,201 @@
...
+export async function getOggMetadata(file: File): Promise<ComfyMetadata> {
+  // Read the entire file into memory (Opus files are generally small enough)
+  const arrayBuffer = await file.arrayBuffer()
+  const data = new Uint8Array(arrayBuffer)
+  const decoder = new TextDecoder('utf-8')
+
+  // Sequentially trace Ogg pages to extract segments of the 'OpusTags' packet containing metadata
+  const segments = extractOpusTags(data, decoder)
+  if (segments.length === 0) {
+    console.error(
```
christian-byrne: suggestion (non-blocking): A file without OpusTags is not an error -- it's just a non-ComfyUI Ogg file. The sibling parsers (`gltf.ts`, `ebml.ts`) silently return empty objects for files without matching metadata. Consider removing the `console.error` or downgrading to `console.warn` to match the codebase pattern.

src/scripts/metadata/ogg.ts
xeinherjer-dev: Fixed in 3e0e366011b6e0aa640a364e89509416b0e384fe

src/scripts/metadata/ogg.ts
xeinherjer-dev: Fixed in 3e0e366011b6e0aa640a364e89509416b0e384fe

src/scripts/metadata/ogg.test.ts
xeinherjer-dev: Fixed in 3e0e366011b6e0aa640a364e89509416b0e384fe

src/scripts/metadata/ogg.ts
xeinherjer-dev: Fixed in 3e0e366011b6e0aa640a364e89509416b0e384fe

src/scripts/metadata/ogg.ts
xeinherjer-dev: Fixed in 3e0e366011b6e0aa640a364e89509416b0e384fe

xeinherjer-dev:
**Summary of changes to address feedback:**
1. **Memory Optimization**: Limited file reading to 2MB (previously reading the entire file) to align with other metadata parsers.
2. **RFC Compliance**: Added `.toLowerCase()` to Vorbis Comment keys to ensure case-insensitive matching per RFC 7845.
3. **Guard Clause**: Added a `segmentLength >= 8` check before decoding to avoid unnecessary processing on tiny segments.
4. **Log Level Adjustment**: Downgraded the missing [OpusTags] log from `console.error` to `console.warn` for consistency with other parsers.
5. **Test Robustness**: Updated the test helper to include an `OpusHead` identification header (Page 0), ensuring the parser correctly handles real-world Ogg/Opus structures.
Verified with 14/14 tests passing and 100% coverage on `ogg.ts`.

christian-byrne (approved):
code looks good - clean rewrite with proper Ogg page traversal, thorough bounds checking, and comprehensive test coverage. all review feedback addressed.

christian-byrne (changes requested):
Rescinding approval - the parser rewrite itself is solid and already wired via parser.ts, but the PR is missing E2E coverage that the other media format PRs include (e.g., PR #4420 for AVIF):

1. Add a `workflow.ogg` test asset to `browser_tests/assets/workflowInMedia/` (a real Ogg/Opus file with embedded ComfyUI prompt+workflow metadata using core nodes only)
2. Add `'workflow.ogg'` to the `fileNames` array in `browser_tests/tests/loadWorkflowInMedia.spec.ts`

This ensures the full drag-and-drop → metadata extraction → workflow load path is covered end-to-end, not just the parser in isolation.

christian-byrne (commented):
Some extra guidance on creating the test asset:

**Generating `workflow.ogg`:**
The easiest way is to use ComfyUI itself — run any workflow that uses a `SaveAudio` node (which outputs Opus/Ogg). The backend embeds `prompt` and `workflow` as Vorbis Comments in the OpusTags packet automatically via `extra_pnginfo`.

**Important: use core nodes only.** The `workflow.avif` test asset is currently commented out in the test file because it uses non-core nodes, which triggers a missing nodes dialog and breaks the screenshot comparison. To avoid this, build a simple workflow using only built-in nodes (e.g., `EmptyLatentImage` → `KSampler` → something that produces audio, or just any minimal core-only workflow).

**Files to change:**
1. Add the generated `.ogg` file to `browser_tests/assets/workflowInMedia/workflow.ogg`
2. Add `'workflow.ogg'` to the `fileNames` array in `browser_tests/tests/loadWorkflowInMedia.spec.ts` (line ~27, after `'workflow.svg'`)

**Verifying locally:**
```bash
pnpm test:browser:local -- --grep 'workflow.ogg'
```

See PR #4420 for the full pattern (that PR added AVIF support with the same set of changes).

xeinherjer-dev:
I've confirmed that the E2E test for workflow.ogg passes locally without any issues, and I've pushed the updates.

I also included a fix for the MIME type mapping in DragDropHelper.ts to ensure .ogg files are correctly identified as audio/ogg during the tests.

image

xeinherjer-dev:
I forced a push to clean up the commit history.
Please review ef4fdf6.

xeinherjer-dev:
Resolved merge conflicts. Updated the tests to align with the recent changes in browser_tests/fixtures/helpers/mimeTypeUtil.ts.

xeinherjer-dev:
@christian-byrne 
This PR now conflicts with main due to #11307. As shown in the screenshot, the target issue remains unresolved.
image

Please let me know if this fix is still needed. If so, I will update the branch and resolve the conflicts.

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,204 @@
...
+ * JSON-encoded prompt and workflow data.
+ *
+ * @param file - The Ogg/Opus file to extract metadata from.
+ * @returns A promise resolving to a {@link ComfyMetadata} object containing
+ *   the parsed `prompt` and `workflow` fields, or `undefined` for each if
+ *   not found or if parsing fails.
+ */
+export async function getOggMetadata(file: File): Promise<ComfyMetadata> {
+  // Read the beginning of the file (Opus files metadata is usually in the first couple of pages)
+  const MAX_READ_BYTES = 2 * 1024 * 1024
```
christian-byrne, line 31: question: the 2MB MAX_READ_BYTES cap is reasonable for typical metadata, but if OpusTags ever exceeds it the file is silently truncated and parsing returns empty. Worth documenting the tradeoff, or falling back to reading more bytes if the OpusTags packet is still open at the cap?

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,204 @@
...
+  const OGG_HEADER_SIZE = 27 // Base size of the Ogg page header (from magic number to just before segment count)
+  const OGG_PAGE_SEGMENTS_OFFSET = 26 // Offset position where the number of segments in the page is stored
+  const OGG_MAX_SEGMENT_SIZE = 255 // Ogg spec: the maximum size of a single segment is 255 bytes
+
+  const segments: Uint8Array[] = []
+  let offset = 0
+  let inOpusTags = false
+
+  while (offset + OGG_HEADER_SIZE <= data.length) {
+    const pageSignature = decoder.decode(data.subarray(offset, offset + 4))
```
christian-byrne: nitpick (non-blocking): decoder.decode(data.subarray(offset, offset+4)) per page just to compare against 'OggS' allocates a string each iteration. A direct byte compare (data[offset]===0x4f && data[offset+1]===0x67 && data[offset+2]===0x67 && data[offset+3]===0x53) avoids the per-page allocation. Same applies to the 'OpusTags' check at line 91. Pure micro-perf.

src/scripts/metadata/ogg.ts
```diff
@@ -1,30 +1,204 @@
...
+
+    if (dataOffset > data.length) break
+
+    let pageProcessingEnded = false
+
+    for (let i = 0; i < pageSegmentsCount; i++) {
+      const segmentLength = data[lengthsOffset + i]
+
+      // Bounds check: ensure the segment data lies within the available data
+      if (dataOffset + segmentLength > data.length) break
```
christian-byrne: suggestion (non-blocking): when this bounds check fires, the inner for breaks but pageProcessingEnded stays false, so the outer while advances offset = dataOffset into the middle of a truncated segment. The next iteration's 'OggS' magic check normally catches it, so behavior is safe in practice. Setting pageProcessingEnded = true (or breaking the outer loop) on this bounds failure would make the intent explicit and avoid relying on the next-page magic check as a guard.

src/scripts/metadata/ogg.ts
xeinherjer-dev: I added a console.warn so it no longer fails silently when the OpusTags packet is truncated at the 2MB cap.

src/scripts/metadata/ogg.ts
xeinherjer-dev: I updated the code to use direct byte comparisons to avoid the per-page allocations. I also extracted this into a compareMagic function to keep it readable.

src/scripts/metadata/ogg.ts
xeinherjer-dev: I've added explicit bounds checking during segment processing to prevent infinite loops on corrupted files.

xeinherjer-dev:
@christian-byrne 
I have addressed the review comments.
I also merged the latest main to resolve the conflicts with #11307. Since I incorporated some test cases from it, I added a try...catch block to the main code to handle them properly.

xeinherjer-dev:
I have read and agree to the Contributor License Agreement

PR #9327, opened 2026-03-01
refactor: use the gradient directly instead of with a custom utility.

DrJKL:
## Summary

Little simpler.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9327-refactor-use-the-gradient-directly-instead-of-with-a-custom-utility-3166d73d36508179876af0ef8cea35b7) by [Unito](https://www.unito.io)

### DIFF
--- a/packages/design-system/src/css/style.css
+++ b/packages/design-system/src/css/style.css
@@ -199,9 +199,7 @@
       #3e1ffc 65.17%,
       #009dff 103.86%
     ),
-    linear-gradient(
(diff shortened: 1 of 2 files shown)


AustinMroz (approved):
Strangely, this PR does not function on my primary browser and results in an unstyled button. Excepting this, the code is a significant improvement and works on the all 3 other browsers I tried.

Code has my approval, but is worth keeping an eye on.

PR #9330, opened 2026-03-02
fix: release webcam stream when WebcamCapture node is removed

jtydhr88:
## Summary

The MediaStream was not stopped on node removal, keeping the camera active. Hoist the stream reference and stop all tracks in onRemoved.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9330-fix-release-webcam-stream-when-WebcamCapture-node-is-removed-3176d73d365081d2ae66e938ee58b713) by [Unito](https://www.unito.io)

### DIFF
--- a/src/extensions/core/webcamCapture.ts
+++ b/src/extensions/core/webcamCapture.ts
@@ -24,9 +24,11 @@ app.registerExtension({
         const video = document.createElement('video')
         video.style.height = video.style.width = '100%'
 
+        let stream: MediaStream | null = null
+
         const loadVideo = async () => {
           try {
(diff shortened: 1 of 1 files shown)


src/extensions/core/webcamCapture.ts
```diff
@@ -61,6 +63,13 @@ app.registerExtension({
           }
         }
 
+        const originalOnRemoved = node.onRemoved
```
Myestery, line 66: We want to cater for other cases.
Eg, when its set to on run, this is sufficient. However when its set to the other option, it makes sense to turn off the stream after the image is captured

PR #9332, opened 2026-03-02
fix: respect 'always snap to grid' when auto-scale layout from nodes 1.0 to 2.0

woct0rdho:
## Summary

Previously when I switch from nodes 1.0 to 2.0, positions and sizes of nodes do not follow 'always snap to grid'. You can guess what a mess it is for people relying on snap to grid to retain sanity. This PR fixes it.

## Changes

In `ensureCorrectLayoutScale`, we call `snapPoint` after the position and the size are updated.

We also need to ensure that the snapped size is larger than the minimal size required by the content, so I've added 'ceil' mode to `snapPoint`, and the patch is larger than I thought first.

I'd happily try out nodes 2.0 once this is addressed :)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9332-fix-respect-always-snap-to-grid-when-auto-scale-layout-from-nodes-1-0-to-2-0-3176d73d365081f5b6bcc035a8ffa648) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LGraph.ts
```diff
@@ -2542,7 +2542,15 @@ export class LGraph
...
-          this.getNodeById(id)?.configure(nodeData)
+          const node = this.getNodeById(id)
+          node?.configure(nodeData)
+
+          if (LiteGraph.alwaysSnapToGrid && node) {
+            const snapTo = this.getSnapToGridSize()
+            if (node.snapToGrid(snapTo)) {
+              node.pos = [node.pos[0], node.pos[1]]
+            }
+          }
```
christian-byrne, line 2608: issue: `node.snapToGrid(snapTo)` only snaps position, not size. The other snap sites in this PR (`ensureCorrectLayoutScale` and `app.ts`) snap both position and size (using `ceil` for sizes). This means workflows loaded/configured with `alwaysSnapToGrid` enabled will have snapped positions but unsnapped sizes.

Consider adding size snapping here as well, e.g.:
```ts
if (LiteGraph.alwaysSnapToGrid && node) {
  const snapTo = this.getSnapToGridSize()
  if (node.snapToGrid(snapTo)) {
    node.pos = [node.pos[0], node.pos[1]]
  }
  snapPoint(node.size, snapTo, 'ceil')
}
```

src/lib/litegraph/src/LGraph.ts
```diff
@@ -2542,7 +2542,15 @@ export class LGraph
...
         // configure nodes afterwards so they can reach each other
         for (const [id, nodeData] of nodeDataMap) {
-          this.getNodeById(id)?.configure(nodeData)
+          const node = this.getNodeById(id)
+          node?.configure(nodeData)
+
+          if (LiteGraph.alwaysSnapToGrid && node) {
+            const snapTo = this.getSnapToGridSize()
+            if (node.snapToGrid(snapTo)) {
+              node.pos = [node.pos[0], node.pos[1]]
```
christian-byrne, line 2605: suggestion (non-blocking): `node.pos = [node.pos[0], node.pos[1]]` looks like a no-op but is needed because `snapToGrid` mutates the internal `_pos` array in-place (bypassing the setter), and this reassignment triggers the `pos` setter to sync to the Vue layout store. A brief comment would prevent future confusion or accidental removal.

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
```diff
@@ -38,6 +40,10 @@ export function ensureCorrectLayoutScale(
     return
   }
 
+  const alwaysSnapToGrid = settingStore.get('pysssss.SnapToGrid')
+  const snapSize =
+    settingStore.get('Comfy.SnapToGrid.GridSize') || LiteGraph.CANVAS_GRID_SIZE
```
christian-byrne: suggestion (non-blocking): The grid size resolution here (`settingStore.get('Comfy.SnapToGrid.GridSize') || LiteGraph.CANVAS_GRID_SIZE`) differs from `LGraph.getSnapToGridSize()` which has additional fallback-to-1 logic when `alwaysSnapToGrid` is true and grid size is 0. In practice both resolve the same due to `useLitegraphSettings.ts` syncing them, but using `getSnapToGridSize()` (or matching its logic) would be more consistent.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -2945,6 +2945,9 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
 
             // Enforce minimum size
             const min = node.computeSize()
+            if (this._snapToGrid) {
+              snapPoint(min, this._snapToGrid, 'ceil')
```
christian-byrne, line 2965: nitpick (non-blocking): `newBounds.size` is snapped with `'round'` (above) but `min` is snapped with `'ceil'` here. This is correct -- minimums should round up to avoid clipping -- but a brief comment explaining the asymmetry would help future readers.

src/lib/litegraph/src/measure.ts
```diff
@@ -351,11 +351,15 @@ export function createBounds(
  * @returns `true` if snapTo is truthy, otherwise `false`
  * @remarks `NaN` propagates through this function and does not affect return value.
  */
-export function snapPoint(pos: Point | Rect, snapTo: number): boolean {
+export function snapPoint(
+  pos: Point | Rect,
+  snapTo: number,
+  method: 'round' | 'ceil' | 'floor' = 'round'
+): boolean {
```
christian-byrne, line 358: nitpick (non-blocking): The existing `snapPoint` test only covers the default `'round'` behavior. Consider adding test cases for the new `'ceil'` and `'floor'` methods.

woct0rdho:
Hi @christian-byrne , thank you for the review! I've implemented the requested changes.

woct0rdho:
I've merged the main branch with #9680 which changed `ensureCorrectLayoutScale` a lot. Please check again.

christian-byrne (approved):
All review feedback addressed. Clean, focused change - snap-to-grid logic is additive with safe defaults, good test coverage for new ceil/floor modes.

christian-byrne (approved):
LGTM

christian-byrne (commented):
Re-checked after the merge with #9680. The refactored ensureCorrectLayoutScale merged cleanly - snap logic (applySnap helper, unprojectPosSize with graph param, ceil for sizes) is correctly layered on the new simplified structure. Typecheck passes.

PR #9333, opened 2026-03-02
feat: add collapse/expand all toggle to right panel tabs

jaeone94:
## Summary
Adds a collapse/expand all toggle button to all parameter and error tabs in the right side panel, letting users quickly collapse or expand all accordion sections at once.

## Changes
- **New component**: `CollapseToggleButton.vue` — a reusable icon button (list-collapse / list-tree icon) with tooltip, bound via `v-model`
- **Error tab**: Toggle collapses/expands all error groups; per-group state is now managed through `isSectionCollapsed` / `setSectionCollapsed` helpers
- **Nodes tab** (`TabNodes.vue`): Per-node `collapseMap`; toggle overrides per-section state; defaults to collapsed (nodes tab default was already collapsed)
- **Normal inputs tab** (`TabNormalInputs.vue`): Per-node `collapseMap` + `advancedCollapsed`; toggle covers both normal and advanced sections; defaults to collapsed when multiple nodes selected
- **Subgraph inputs tab** (`TabSubgraphInputs.vue`): Toggle covers both main and advanced inputs sections
- **Global parameters tab** (`TabGlobalParameters.vue`): Toggle bound to single `SectionWidgets`
- **i18n**: Added `g.collapseAll` and `g.expandAll` keys

## Review Focus
- `isAllCollapsed` getter in each tab: reads from the same per-section state so the toggle accurately reflects current state rather than being independently tracked
- `TabNormalInputs`: multi-node selection default collapse behaviour is preserved through `isSectionCollapsed` fallback logic

## Screenshots
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9333-feat-add-collapse-expand-all-toggle-to-right-panel-tabs-3176d73d36508123ba22d6e81983bb1b) by [Unito](https://www.unito.io)


src/components/rightSidePanel/parameters/TabNormalInputs.vue
```diff
@@ -100,21 +132,22 @@ const advancedLabel = computed(() => {
       :node
       :label
       :widgets
-      :collapse="isMultipleNodesSelected && !isSearching"
+      :collapse="isSectionCollapsed(String(node.id))"
```
dante01yoon: Question: The `!isSearching` guard was dropped entirely. Now if a user has multiple nodes selected (sections default collapsed), and they type a search query, the sections stay collapsed. The search filters widgets correctly behind the scenes, but the user sees collapsed accordions with no visible results — it looks like "no matches found" even though matches exist.

src/components/rightSidePanel/parameters/TabGlobalParameters.vue
```diff
@@ -117,9 +119,11 @@ onBeforeUnmount(() => {
       :searcher
       :update-key="favoritedWidgets"
     />
+    <CollapseToggleButton v-model="isAllCollapsed" />
   </div>
   <SectionWidgets
     ref="sectionWidgetsRef"
+    v-model:collapse="isAllCollapsed"
```
dante01yoon: Question:  If the user clicks the collapse toggle to collapse the section, then starts typing a search query, the section remains collapsed. The filtered favorites are computed correctly but rendered inside a collapsed accordion — invisible to the user? 

`TabNodes.vue` already handles this correctly with `:collapse="isSectionCollapsed(...) && !isSearching"`. The inconsistency means search works differently depending on which tab you're in, which is confusing.

dante01yoon (commented):
few thoughts

src/components/rightSidePanel/parameters/TabNormalInputs.vue
jaeone94: Great catch, you're absolutely right. The `isSectionCollapsed` function falls back to
`isMultipleNodesSelected` when no explicit state is set in `collapseMap`, so sections remain
collapsed during search even though the widgets are correctly filtered behind the scenes.

Fixed in this PR by adding the `&& !isSearching` guard back to the `:collapse` binding,
matching the existing pattern in `TabNodes.vue`:

```vue
:collapse="isSectionCollapsed(String(node.id)) && !isSearching"
```

src/components/rightSidePanel/parameters/TabGlobalParameters.vue
jaeone94: Valid issue — `isAllCollapsed` was a plain `ref` with no awareness of `isSearching`, so a
manually collapsed section would stay collapsed regardless of an active search query.

Fixed by splitting `v-model:collapse` into a one-way `:collapse` binding with the
`&& !isSearching` guard, while merging the existing `@update:collapse` handler
(`nextTick(setDraggableState)`) and the new state assignment into a single arrow function
to avoid a duplicate-attribute lint error:

```vue
:collapse="isAllCollapsed && !isSearching"
@update:collapse="(v) => { isAllCollapsed = v; nextTick(setDraggableState) }"
```

All three parameter tabs (`TabNodes`, `TabNormalInputs`, `TabGlobalParameters`) now share
consistent search-expand behavior.

LittleSound (changes requested):
- This button is meaningless when only one node is displayed or no node is displayed. It is recommended to hide it.

- This button is invalid when searching, and it is recommended to fix it.

dante01yoon (dismissed):
LGTM

src/components/rightSidePanel/parameters/TabGlobalParameters.vue
```diff
@@ -111,23 +113,47 @@ onBeforeUnmount(() => {
...
+    <Transition
+      enter-active-class="transition-all duration-300 ease-in-out"
+      enter-from-class="max-w-0 opacity-0 ml-0"
+      enter-to-class="max-w-10 opacity-100 ml-2"
+      leave-active-class="transition-all duration-300 ease-in-out"
+      leave-from-class="max-w-10 opacity-100 ml-2"
+      leave-to-class="max-w-0 opacity-0 ml-0"
+    >
+      <div
+        v-if="!isSearching && 1 > 1"
```
LittleSound: What is this? `v-if="1 > 1"`

src/components/rightSidePanel/parameters/TabNodes.vue
```diff
@@ -39,12 +75,30 @@ async function searcher(query: string) {
...
+      class="flex-1"
     />
+    <Transition
+      enter-active-class="transition-all duration-300 ease-in-out"
+      enter-from-class="max-w-0 opacity-0 ml-0"
+      enter-to-class="max-w-10 opacity-100 ml-2"
+      leave-active-class="transition-all duration-300 ease-in-out"
+      leave-from-class="max-w-10 opacity-100 ml-2"
+      leave-to-class="max-w-0 opacity-0 ml-0"
+    >
```
LittleSound: The Transition template is repeated 5 times. The identical 8-line Transition class definition is copied and pasted into 5 tabs. Consider placing the animation inside the button component?

src/components/rightSidePanel/parameters/TabNodes.vue
```diff
@@ -17,7 +19,7 @@ const workflowStore = useWorkflowStore()
 
 const nodes = computed((): LGraphNode[] => {
   // Depend on activeWorkflow to trigger recomputation when workflow changes
-  void workflowStore.activeWorkflow?.path
+  Boolean(workflowStore.activeWorkflow?.path)
   return (canvasStore.canvas?.graph?.nodes ?? []) as LGraphNode[]
```
LittleSound, line 23: Boolean(...) creates a value and then discards it, doing the same thing as void, both of which are used to trigger Vue's reactive dependency tracking. void is the more idiomatic way to write this, so this change is meaningless.

src/components/rightSidePanel/parameters/TabNodes.vue
```diff
@@ -30,6 +32,40 @@ const searchedWidgetsSectionDataList = shallowRef<NodeWidgetsListList>(
...
+function setSectionCollapsed(nodeId: string, collapsed: boolean) {
+  collapseMap[nodeId] = collapsed
+}
+
+const isAllCollapsed = computed({
+  get() {
+    return searchedWidgetsSectionDataList.value.every(
+      ({ node }) => isSectionCollapsed(String(node.id)) === true
+    )
+  },
```
LittleSound, line 61: This `=== true` is inconsistent; `TabNodes.vue:59` still has `=== true`, while `TabNormalInputs.vue:83` does not.

LittleSound (changes requested):

src/components/rightSidePanel/parameters/TabNormalInputs.vue
```diff
@@ -76,12 +119,35 @@ const advancedLabel = computed(() => {
...
+      <div
+        v-if="
+          !isSearching &&
+          widgetsSectionDataList.length +
+            advancedWidgetsSectionDataList.length >
+            1
+        "
+        class="overflow-hidden flex items-center"
+      >
+        <CollapseToggleButton v-model="isAllCollapsed" />
```
LittleSound: There is no gap between the search box and the button.
Image

src/components/rightSidePanel/parameters/TabGlobalParameters.vue
jaeone94: Good catch! That was a mistake left from development — `1 > 1` is always `false`, which completely hides the button and makes this a dead code block. Since `TabGlobalParameters` only ever shows a single section, the `CollapseToggleButton` is not needed here at all. I've removed the entire block along with the `isAllCollapsed` ref and the unused import.

Also cleaned up two follow-on issues this removal exposed:
- Extracted the inline `() => nextTick(setDraggableState)` from the template into an explicit `onCollapseUpdate()` function in the script, so `nextTick` is visibly used in the script rather than only in the template.
- Removed the hardcoded `:collapse="false"` prop from `SectionWidgets` — since the prop is just the Vue 3 default (`undefined` / falsy), passing it explicitly adds noise without meaning.

src/components/rightSidePanel/parameters/TabNodes.vue
jaeone94: Totally agree, thanks for pointing this out. I've moved the `<Transition>` wrapper and the `<div v-if>` logic directly into CollapseToggleButton.vue and added a `show` prop to control visibility. Each tab now just passes `:show="..."` — no more copy-pasted transition blocks.

src/components/rightSidePanel/parameters/TabNodes.vue
jaeone94: You're right, my bad. The `Boolean(...)` line was redundant since `void` on the line above already handles the reactive dependency tracking. I've removed the `Boolean(...)` call and kept only the `void`.

src/components/rightSidePanel/parameters/TabNodes.vue
jaeone94: Fixed. Removed the unnecessary `=== true` comparison to make it consistent with TabNormalInputs.vue. Since isSectionCollapsed already returns a `boolean`, the explicit comparison was redundant.

src/components/rightSidePanel/parameters/TabNormalInputs.vue
jaeone94: Fixed! Since the `<Transition>` + `<div>` wrapper has now been moved into CollapseToggleButton.vue, I've added `ml-2` directly to the inner `div` inside the component, so the gap is consistently applied whenever the button is visible.

christian-byrne (approved):
Confirmed all of LittleSound's review comments are addressed, LGTM generally. Assigned for second review but can also merge now proactively and address incoming comments from 2nd round in a followup should they occur.

LittleSound (approved):
LGTM

PR #9334, opened 2026-03-02
feat(error-groups): sort execution error cards by node execution ID

jaeone94:
## Summary

Sort execution error cards within each error group by their node execution ID in ascending numeric order, ensuring consistent and predictable display order.

## Changes

- **What**: Added `compareExecutionId` utility to `src/types/nodeIdentification.ts` that splits node IDs on `:` and compares segments numerically left-to-right; applied it as a sort comparator when building `ErrorGroup.cards` in `useErrorGroups.ts`

## Review Focus

- The comparison treats missing segments as `0`, so `"1"` sorts before `"1:20"` (subgraph nodes follow their parent); confirm this ordering matches user expectations
- All comparisons are purely numeric — non-numeric segment values would sort as `NaN` (treated as `0`)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9334-feat-error-groups-sort-execution-error-cards-by-node-execution-ID-3176d73d365081e1b3e4e4fa8831fe16) by [Unito](https://www.unito.io)

### DIFF
--- a/src/types/nodeIdentification.ts
+++ b/src/types/nodeIdentification.ts
@@ -161,7 +161,10 @@ export function compareExecutionId(
   const idA = parse(a)
   const idB = parse(b)
   for (let i = 0; i < Math.max(idA.length, idB.length); i++) {
-    const diff = (idA[i] ?? 0) - (idB[i] ?? 0)
+    const segA = idA[i] ?? 0
+    const segB = idB[i] ?? 0
+    const diff =
+      (Number.isNaN(segA) ? 0 : segA) - (Number.isNaN(segB) ? 0 : segB)
     if (diff !== 0) return diff
   }
   return 0
--- a/src/types/nodeIdentification.ts
+++ b/src/types/nodeIdentification.ts
@@ -146,3 +146,23 @@ export function getParentExecutionIds(
 ): NodeExecutionId[] {
   return getAncestorExecutionIds(executionId).slice(0, -1)
 }
+
+/**
+ * Compare two NodeExecutionIds for ascending numeric sort order.
+ * Splits each ID by ":" and compares segments numerically left to right.
+ *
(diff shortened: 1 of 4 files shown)


src/types/nodeIdentification.ts
```diff
@@ -146,3 +146,26 @@ export function getParentExecutionIds(
...
   return getAncestorExecutionIds(executionId).slice(0, -1)
 }
+
+/**
+ * Compare two NodeExecutionIds for ascending numeric sort order.
+ * Splits each ID by ":" and compares segments numerically left to right.
+ *
+ * Example order: "1" < "1:20" < "2" < "10:11:12"
+ */
+export function compareExecutionId(
```
dante01yoon, line 156: `compareExecutionId` has no dedicated unit tests in `nodeIdentification.test.ts` edge cases (empty string, undefined, non-numeric segments) are only indirectly covered through the integration tests in `useErrorGroups.test.ts`.

jaeone94: You're right. I only wrote integration tests and forgot to add dedicated
unit tests for `compareExecutionId`.

Added a dedicated `describe('compareExecutionId')` block covering:

- Basic ascending sort of simple numeric IDs
- Left-to-right segment comparison for nested IDs (matching the documented
  example order: `"1" < "1:20" < "2" < "10:11:12"`)
- Missing trailing segments defaulting to `0` (`idA[i] ?? 0` path)
- `undefined` inputs (`id ?? ''` fallback path)
- Empty string inputs
- Non-numeric segments hitting the `Number.isNaN(seg) ? 0 : seg` guard

All 28 tests in the file pass.

dante01yoon (approved):

christian-byrne (approved):

PR #9335, opened 2026-03-02
App builder item - cap max width and truncate

pythongosssss:
## Summary

Currently they overflow, this adds truncation

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9335-App-builder-item-cap-max-width-and-truncate-3176d73d365081d1b461cf2758b04ec2) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/builder/IoItem.vue
+++ b/src/components/builder/IoItem.vue
@@ -32,9 +32,12 @@ const entries = computed(() => {
 })
 </script>
 <template>
-  <div class="p-2 my-2 rounded-lg flex items-center-safe">
-    <span class="mr-auto" v-text="title" />
-    <span class="text-muted-foreground mr-2 text-end" v-text="subTitle" />
+  <div class="p-2 my-2 rounded-lg flex items-center-safe min-w-0 gap-0.5">
+    <span class="max-w-3/4 truncate mr-auto" v-text="title" />
+    <span
+      class="text-muted-foreground mr-2 text-end truncate"
+      v-text="subTitle"
+    />
     <Popover :entries>
       <template #button>
         <Button variant="muted-textonly">
--- a/src/components/builder/IoItem.vue
+++ b/src/components/builder/IoItem.vue
@@ -33,9 +33,12 @@ const entries = computed(() => {
 </script>
 <template>
   <div class="p-2 my-2 rounded-lg flex items-center-safe gap-2">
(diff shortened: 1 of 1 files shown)


src/components/builder/IoItem.vue
```diff
@@ -32,9 +32,12 @@ const entries = computed(() => {
...
 <template>
-  <div class="p-2 my-2 rounded-lg flex items-center-safe">
-    <span class="mr-auto" v-text="title" />
-    <span class="text-muted-foreground mr-2 text-end" v-text="subTitle" />
+  <div class="p-2 my-2 rounded-lg flex items-center-safe min-w-0 gap-0.5">
+    <span class="max-w-3/4 truncate mr-auto" v-text="title" />
+    <span
+      class="text-muted-foreground mr-2 text-end truncate"
+      v-text="subTitle"
+    />
```
AustinMroz: Perhaps a hybrid approach like this?
```suggestion
  <div class="p-2 my-2 rounded-lg flex items-center-safe gap-0.5">
    <span class="max-w-3/5 truncate mr-auto shrink-1" v-text="title" />
    <span
      class="text-muted-foreground mr-2 text-end truncate grow-1 basis-0"
      v-text="subTitle"
    />
```
image

I've lowered the `max-w` to `3/5` to account for the popover button.

AustinMroz (commented):
I took a stab at fixing this over in #9259 (using shrink with varying weight), but was rather unhappy with how it would function with "short widget name with long node title". I was hopeful that your approach would fix things, but it looks like it still struggles with the humble "CLIP Text Encode".

image

src/components/builder/IoItem.vue
pythongosssss: https://github.com/user-attachments/assets/06c0dadc-21ce-47fe-9122-b1e7d2599a0d

What do you think of this resizing?
The problem with the previous approach was when the title does fit fully with a short subtitle, it was still truncated
image
vs 
image

pythongosssss: Maybe not enough preservation of title vs subtitle?

pythongosssss: Increased weight of title a bit more

prev:
image

curr:
image

AustinMroz (approved):

src/components/builder/IoItem.vue
AustinMroz: Very clever. I like this.

PR #9338, opened 2026-03-02
feat: App mode saving rework

pythongosssss:
## Summary

Change app mode changes to be written directly to the workflow on change instead of requiring explicit save via builder.
Temporary: Adds `.app.json` file extension to app files for identification since we don't currently have a way to identify them with metadata
Removes app builder save dialog and replaces it with default mode selection

## Changes

- **What**: 
- ensure all save locations handle app mode
- remove dirtyLinearData and flushing

- **Breaking**: 
- if people are relying on workflow names and are converting to/from app mode in the same workflow, they will gain/lose the `.app` part of the extension

image

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9338-feat-App-mode-saving-rework-3176d73d3650813f9ae1f6c5a234da8c) by [Unito](https://www.unito.io)


packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -33,6 +33,35 @@ export function appendJsonExt(path: string) {
...
+export type WorkflowSuffix = 'json' | 'app.json'
+
+export function getWorkflowSuffix(
+  suffix: string | null | undefined
+): WorkflowSuffix {
+  return suffix === 'app.json' ? 'app.json' : 'json'
+}
+
+export function appendWorkflowJsonExt(path: string, isApp: boolean): string {
+  return ensureWorkflowSuffix(path, isApp ? 'app.json' : 'json')
```
AustinMroz: Not blocking, but the disparity of when, and how infrequently the EXT constants are used isn't great.

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -96,20 +125,27 @@ export function formatCommitHash(value: string): string {
...
-  if (fullFilename.includes('.')) {
+  const lower = fullFilename.toLowerCase()
+  if (lower.endsWith(APP_JSON_EXT)) {
+    return {
+      filename: fullFilename.slice(0, -APP_JSON_EXT.length),
+      suffix: 'app.json'
+    }
+  }
+  const dotIndex = fullFilename.lastIndexOf('.')
+  if (dotIndex > 0) {
```
AustinMroz: Can invert the conditional to further polish to
```suggestion
  if (dotIndex <= 0) return { filename: fullFilename, suffix: null }
```
but I'm happy enough with the current performance improvements.

src/components/builder/DefaultViewDialogContent.vue
```diff
@@ -87,17 +66,18 @@ import BuilderDialog from './BuilderDialog.vue'
...
 
-const { defaultFilename, onSave, onClose } = defineProps<{
-  defaultFilename: string
-  onSave: (filename: string, openAsApp: boolean) => void
-  onClose: () => void
+const { initialOpenAsApp = true } = defineProps<{
+  initialOpenAsApp?: boolean
+}>()
+
+const emit = defineEmits<{
```
AustinMroz:
```suggestion
defineEmits<{
```
Also very nit, you can do an empty `defineEmits` and use `$emit` in the template

src/components/builder/useAppSetDefaultView.ts
```diff
@@ -0,0 +1,59 @@
...
+import DefaultViewDialogContent from './DefaultViewDialogContent.vue'
+
+const DIALOG_KEY = 'builder-default-view'
+
+export function useAppSetDefaultView() {
+  const workflowStore = useWorkflowStore()
+  const dialogService = useDialogService()
+  const dialogStore = useDialogStore()
+
+  const settingView = ref(false)
```
AustinMroz: Seems kinda smelly to be reimplementing the visibility guards here when `dialogStore`/`dialogService` already check for duplicate dialogues through the group key and provides a `isDialogOpen`

AustinMroz (commented):
This one should really have a lot more deletions. Shouldn't `syncLinearMode` be removed entirely?

packages/shared-frontend-utils/src/formatUtil.ts
pythongosssss: added consts for JSON and APP_JSON for the non dot-prefixed ones

src/platform/workflow/core/services/workflowService.ts
```diff
@@ -103,36 +111,28 @@ export const useWorkflowService = () => {
...
-        itemList: [newPath]
-      })
-
-      if (res !== true) return false
+      if ((await confirmOverwrite(newPath)) !== true) return false
 
       if (!isSelfOverwrite) {
         const deleted = await deleteWorkflow(existingWorkflow, true)
         if (!deleted) return false
       }
```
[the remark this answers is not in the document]
pythongosssss: This is a pre-existing pattern for workflow overwriting, however you are correct that we should clean this up, but not in this PR. Can you add a new issue for resolving this, probably by allowing for the overwrite flag to be passed through when saving the workflow

AustinMroz (approved):

PR #9358, opened 2026-03-03
fix: add focus mode keyboard shortcut

Myestery:
## Summary

Add default keybinding Ctrl+Shift+F for `Workspace.ToggleFocusMode`, which existed as a command but had no keyboard shortcut.

## Changes

- **What**: Add `Ctrl+Shift+F` keybinding for `Workspace.ToggleFocusMode` in `CORE_KEYBINDINGS`; add unit test verifying the binding

## Review Focus

Straightforward single keybinding addition. `Ctrl+Shift+F` was previously unbound.

Fixes #6883


comfydesigner:
Thank you, this should help. The keybind for Focus mode in the panel below should automatically be updated as well, correct? I'm seeing that whatever I bind it to, the keys are reflected properly.

Screenshot 2026-03-13 at 3 18 04 PM

PR #9360, opened 2026-03-03
[style] Update error/subgraph node footer design with layered overlay approach

jaeone94:
## Summary

Refactors the error and subgraph node footer UI by extracting a dedicated `NodeFooter` component and replacing the CSS `outline` approach with a layered border overlay for selection/executing state indicators.

## Changes

- **What**: Extracted `NodeFooter.vue` from `LGraphNode.vue` to encapsulate the footer tab logic (subgraph enter, error, advanced inputs). Replaced CSS `outline` with an absolutely-positioned border overlay div for selection and executing state. Added a separate root border overlay div for the node body border. Removed unused `isTransparent` function from `colorUtil.ts`.
- **Dependencies**: None

## Review Focus

- The layered overlay approach (`absolute -inset-[3px] border-3`) for selection/executing outlines vs the previous `outline-3` approach — ensures the outline renders outside the node bounds correctly including the footer area
- `NodeFooter` handles 4 cases: subgraph+error (dual tabs), error only, subgraph only, advanced inputs — verify edge cases render correctly
- Resize handle bottom offset adjustments for nodes with footers (`hasFooter`)

## Screenshots
image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9360-style-Update-error-subgraph-node-footer-design-with-layered-overlay-approach-3186d73d365081b2ac31f166f4d1944a) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/components/LGraphNode.test.ts
```diff
@@ -184,16 +201,27 @@ describe('LGraphNode', () => {
...
     const wrapper = mountLGraphNode({ nodeData: mockNodeData })
 
+    // Root div should have the executing class
     expect(wrapper.classes()).toContain('outline-node-stroke-executing')
+
+    // The layered outline overlay should be present
+    const overlay = wrapper.find('.border-node-stroke-executing')
+    expect(overlay.exists()).toBe(true)
+    expect(overlay.classes()).toContain('border-3')
   })
```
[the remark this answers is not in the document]
jaeone94: addressed in [5b357b5](https://github.com/Comfy-Org/ComfyUI_frontend/pull/9360/commits/5b357b5f03eced4e14a53e00aa3e38874e54360c)

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -600,6 +573,34 @@ const shapeClass = computed(() => {
...
+    case RenderShape.BOX:
+      return 'rounded-none'
+    case RenderShape.CARD:
+      return isExpanded
+        ? 'rounded-tl-[23px] rounded-br-[23px] rounded-tr-none rounded-bl-none'
+        : 'rounded-tl-[19px] rounded-br-[19px] rounded-tr-none rounded-bl-none'
+    default:
+      return isExpanded ? 'rounded-[23px]' : 'rounded-[19px]'
+  }
+})
```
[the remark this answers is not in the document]
jaeone94: The default case is intentional and correct here. The only node shapes exposed to users via the right-click context menu are Default, Box, and Card — all other RenderShape variants (CIRCLE, HollowCircle, ARROW, GRID, etc.) are used internally for canvas slot rendering (e.g., optional input dots), not as selectable node shapes.

This pattern is also consistent with the existing beforeShapeClass computed property that has been present since the original implementation — it too only handles BOX and CARD explicitly, letting everything else fall through to default.

Adding explicit cases for the remaining variants would produce dead code that is never reached in practice.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -563,34 +541,47 @@ const { latestPreviewUrl, shouldShowPreviewImg } = useNodePreviewState(
...
+
+const footerRootBorderBottomClass = computed(() =>
+  hasFooter.value ? '-bottom-8' : 'bottom-0'
+)
+
+const footerResizeHandleBottomClass = computed(() => {
+  if (!hasFooter.value) return ''
+  return hasAnyError.value ? 'bottom-[-31px]' : 'bottom-[-33px]'
 })
```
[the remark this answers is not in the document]
jaeone94: `hasErrorFooter` is not the right condition here.

**The red ring and border inset are both driven by `hasAnyError`, not `hasErrorFooter`.**

Even when `showErrorsTabEnabled` is `false`, the node still renders the red ring:

```html
<div :class="cn(..., hasAnyError && 'ring-4 ring-destructive-background')">
```

And the outer gray border's inset expansion is also tied to `hasAnyError`:

```html
<div :class="cn(..., hasAnyError ? '-inset-1' : 'inset-0', footerRootBorderBottomClass)">
```

With the `hasErrorFooter` approach, when `hasAnyError=true` and `showErrorsTabEnabled=false` on a plain node (`hasFooter=false`), `footerRootBorderBottomClass` returns `'bottom-0'`, which **overrides and cancels** the `-4px` bottom set by `-inset-1`. The gray border snaps to `0px` while the red ring still extends `4px` outward — causing visible misalignment.

Similarly, the 1px difference in `footerStateOutlineBottomClass` and `footerResizeHandleBottomClass` comes from the selection outline's own inset expanding from `-inset-[3px]` to `-inset-[7px]` based on `hasAnyError`. Since this visual expansion is unconditional on `showErrorsTabEnabled`, the compensating offsets must also be gated on `hasAnyError`.

The `hasFooter` computed already handles the "is the error tab rendered?" concern for footer **visibility**. The offset classes serve a separate geometry concern and need to stay `hasAnyError`-based.

viva-jinyi (dismissed):
LGTM

src/utils/colorUtil.ts
```diff
@@ -21,15 +21,6 @@ export interface ColorAdjustOptions {
   opacity?: number
 }
 
-export function isTransparent(color: string) {
```
AustinMroz, line 24: Removing this section of code introduces a regression with custom nodes like the rgthree label node.

| Before | After |
| ------ | ----- |
| before | after|

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -563,34 +541,47 @@ const { latestPreviewUrl, shouldShowPreviewImg } = useNodePreviewState(
...
+
+const bodyRoundingClass = computed(() => {
+  switch (nodeData.shape) {
+    case RenderShape.BOX:
+      return 'rounded-none'
+    case RenderShape.CARD:
+      // Rounding tl-none and br-2xl as per user's final manual correction
+      return 'rounded-tl-none rounded-br-2xl rounded-tr-none rounded-bl-none'
+    default:
+      return 'rounded-b-2xl'
```
AustinMroz, line 590: Very nit. No need to explicity set rounded-none. Same for for the other shape classes.
(and the comment should probably be removed)

```suggestion
    case RenderShape.BOX:
      return ''
    case RenderShape.CARD:
      return 'rounded-br-2xl'
    default:
      return 'rounded-b-2xl'
```

AustinMroz (commented):
Nothing blocking from me, 

There's some subtle changes in node sizing I'm not too happy about. My best guess is that it's a consequence of the node sizing restrictions being applied on a sub-element.
image


Introduces a regression where node borders draw above node slots

| Before | After |
| ------ | ----- |
| before | after|

src/renderer/extensions/vueNodes/components/LGraphNode.vue
jaeone94: Thank you for pointing that out! I've removed all unnecessary comments and classes !

src/utils/colorUtil.ts
jaeone94: Ah! I see. I restored the isTransparent function and made the default border and selection border invisible! (And I also made the badges invisible in the Vue node.)

jaeone94: And I adjusted the z-index of the selected border so it would appear below the slot.

jaeone94: Also , about the some subtle changes in node sizing
The sizing difference was caused by a structural change in how borders are applied.

In main, the border sits directly on the root container div, so its 1px contributes to the overall layout size. In this PR, the border was moved to an absolutely-positioned overlay (the "Root Border Overlay"), which means it no longer participates in layout flow. Since the node-inner-wrapper — which now owns the min-h / w sizing constraints — had no border of its own, the node ended up ~2px narrower/shorter (1px per side).

Fixed by adding border border-solid border-transparent to the node-inner-wrapper, so it occupies the same 1px of layout space without visually showing an extra border.

jaeone94: Thank you so much for meticulously reviewing the parts I missed. This PR seems much better than before.

AustinMroz (dismissed):
Looks like test failures are flakes

viva-jinyi (dismissed):

PR #9362, opened 2026-03-04
Split selection into an inputs and outputs step

AustinMroz:
When building an app, selecting inputs and selecting outputs are now 2 separate steps. This prevents confusion where clicking on the widget of an output node will select that widget instead of the entire output.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9362-Split-selection-into-an-inputs-and-outputs-step-3196d73d36508187b4a1e51c73f1c54c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/builder/BuilderToolbar.vue
+++ b/src/components/builder/BuilderToolbar.vue
@@ -6,14 +6,7 @@
     <div
       class="inline-flex items-center gap-1 rounded-2xl border border-border-default bg-base-background p-2 shadow-interface"
     >
-      <template
-        v-for="(step, index) in [
-          selectInputsStep,
-          selectOutputsStep,
-          arrangeStep
-        ]"
-        :key="step.id"
-      >
+      <template v-for="(step, index) in steps" :key="step.id">
         <button
           :class="
             cn(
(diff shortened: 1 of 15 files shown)


src/locales/en/main.json
```diff
@@ -3358,16 +3358,18 @@
...
+    "outputs": "Outputs",
+    "inputsDescription": "Choose inputs",
+    "outputsDescription": "Choose outputs",
     "arrange": "Preview",
     "arrangeDescription": "Review app layout",
     "defaultView": "Set a default view",
     "defaultViewDescription": "Choose how this opens",
     "connectOutput": "Connect an output",
     "connectOutputBody1": "Your app needs at least one output to be connected before it can be saved.",
     "connectOutputBody2": "Switch to the 'Select' step and click on output nodes to add them here.",
```
pythongosssss: This one also needs updating

src/components/builder/BuilderToolbar.vue
pythongosssss: this should be 3 now - probably pull the array out of the loop, then use that.length

pythongosssss: and here

pythongosssss (changes requested):

pythongosssss (approved):

PR #9364, opened 2026-03-04
feat: support disabled state and upstream value display for curve widget

jtydhr88:
## Summary
When the curve widget has an incoming connection, display the upstream node's curve value in read-only mode with reduced opacity and hidden control points.

this is improvment for Curve Editor node and Curve to Sigmas Node [(here)](https://github.com/Comfy-Org/ComfyUI/pull/12757)

https://github.com/user-attachments/assets/b010b16c-7509-42ab-bce3-7258fe462567

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9364-feat-support-disabled-state-and-upstream-value-display-for-curve-widget-3196d73d3650813d8065fd4663798634) by [Unito](https://www.unito.io)


src/components/curve/WidgetCurve.vue
```diff
@@ -1,16 +1,39 @@
...
   ]
 })
+
+const isDisabled = computed(() => !!widget.options?.disabled)
+
+const upstreamValue = useLinkedWidgetValue(nodeId, widget.name, 'curve')
+
+const effectivePoints = computed(() =>
+  isDisabled.value && upstreamValue.value
+    ? (upstreamValue.value as CurvePoint[])
```
DrJKL, line 36: This seems like a dangerous cast. Can we do more validation here?

src/composables/useLinkedWidgetValue.ts
DrJKL, line 1: I get that this is functional, but I worry that it could create some really weird behaviors to have more cross-widget and now cross-widget+link data binding.

src/composables/useLinkedWidgetValue.ts
```diff
@@ -0,0 +1,50 @@
...
+import { computed } from 'vue'
+
+import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
+import { useWidgetValueStore } from '@/stores/widgetValueStore'
+
+/**
+ * Returns the upstream widget value for a connected input slot.
+ * Matches by output slot name first, then falls back to widget type.
+ */
+export function useLinkedWidgetValue(
```
christian-byrne, line 11: issue: This composable may be unnecessary. `NodeWidgets.vue` already sets `options.disabled = true` when an input slot is linked (L229-232: `slotMetadata?.linked`), and the `v-model` value from the widget store (`widgetState?.value`) should already reflect the current upstream value. If that's the case, `WidgetCurve` just needs to read the existing `modelValue` when disabled -- no graph traversal needed.

Additionally, the `computed` here accesses non-reactive litegraph internals (`getNodeById`, `getLink`, `inputs.find`) through `canvasStore.canvas?.graph`. These are plain JS objects, not Vue reactive proxies, so the computed will not re-evaluate when links are connected or disconnected -- only when `canvasStore.canvas` itself is replaced.

Could you verify whether `widgetState.value` already contains the upstream value for a slot-linked curve widget? If it does, this composable can be removed entirely.

src/components/curve/WidgetCurve.vue
christian-byrne, line 36: issue (non-blocking): `upstreamValue.value as CurvePoint[]` casts `unknown` without runtime validation. If the upstream widget holds a value that isn't a `CurvePoint[]`, `CurveEditor` will receive invalid data. A guard would be safer:
```ts
if (Array.isArray(upstream) && upstream.length >= 2) {
  return upstream as CurvePoint[]
}
```

src/composables/useLinkedWidgetValue.ts
```diff
@@ -0,0 +1,50 @@
...
+  nodeId: string,
+  widgetName: string,
+  widgetType: string
+): ComputedRef<unknown | undefined> {
+  const canvasStore = useCanvasStore()
+  const widgetValueStore = useWidgetValueStore()
+
+  return computed(() => {
```
christian-byrne, line 19: suggestion: The graph traversal here reads non-reactive litegraph internals (`graph.getNodeById`, `graph.getLink`, `node.inputs.find`) inside a Vue `computed`. Since these are plain JS objects and not reactive proxies, the computed only re-evaluates when `canvasStore.canvas` changes (the only tracked dependency), not when links are connected/disconnected. This will cause stale upstream values after rewiring.

If this composable is still needed (see other comment), consider introducing a reactive graph-revision counter that increments on graph mutations and is read inside this computed to trigger re-evaluation.

src/composables/useLinkedWidgetValue.ts
christian-byrne, line 11: suggestion (non-blocking): No tests cover the new composable or the disabled behavior in `CurveEditor`. `useLinkedWidgetValue` has multiple code paths (name match, type fallback, null guards) that would benefit from unit tests.

jtydhr88:
use https://github.com/Comfy-Org/ComfyUI_frontend/pull/9851 instead

PR #9365, opened 2026-03-04
fix: prevent Vue node collapse when dragging from top-left corner

kaili-yang:
## Summary

Suppress the browser click event that fires after pointerup when a drag operation ends over the collapse button, matching litegraph behavior.

## Screenshots
Verify that dragging on interactive elements does not activate them, like `delete` or `collapse ` icon.
Here's an example for `collapse ` icon.

before

https://github.com/user-attachments/assets/747b405f-1b31-4f31-bc18-888922accf23

after

https://github.com/user-attachments/assets/9e55c53c-6130-40c2-9976-0991f78f7791


src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -138,6 +138,7 @@ export function useNodePointerInteractions(
       safeDragEnd(event)
 
       if (wasDragging) {
+        suppressNextClick(event)
```
christian-byrne: issue: `event.currentTarget` may reference a detached DOM node after `safeDragEnd()`.

`safeDragEnd(event)` is called on line 138 before `suppressNextClick(event)` on line 141. `safeDragEnd` -> `endDrag` can trigger Vue re-renders that replace the DOM element. If the element is replaced, `event.currentTarget` points to a detached node, and the capture listener attaches to it -- it never fires, so no click suppression occurs (the `if (!target) return` silently fails).

Consider capturing the target before `safeDragEnd`:
```ts
if (hasDraggingStarted || wasDragging) {
  const nodeEl = event.currentTarget as HTMLElement | null
  safeDragEnd(event)
  if (wasDragging) {
    if (nodeEl) suppressNextClick(nodeEl)
    return
  }
}
```
(with `suppressNextClick` signature changed to accept `HTMLElement`)

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -153,6 +154,24 @@ export function useNodePointerInteractions(
...
 
+  /**
+   * Register a one-shot capture-phase listener to swallow it so
+   * interactive children (e.g. collapse button) don't trigger.
+   */
+  function suppressNextClick(event: PointerEvent) {
+    const target = event.currentTarget as HTMLElement | null
+    if (!target) return
+
+    target.addEventListener(
```
christian-byrne, line 171: issue: The `{ once: true }` listener leaks if no click fires after drag.

Browsers only synthesize a `click` after `pointerup` if pointerdown and pointerup target the same element without significant movement. After a drag, this condition often isn't met, so no `click` fires and the one-shot listener persists. The next legitimate click on that node (collapse, widget, combo box) gets silently swallowed.

A `setTimeout` cleanup would fix this:
```ts
function suppressNextClick(target: HTMLElement) {
  const handler = (e: Event) => {
    e.stopPropagation()
    e.preventDefault()
  }
  target.addEventListener('click', handler, { capture: true, once: true })
  setTimeout(() => target.removeEventListener('click', handler, { capture: true }), 300)
}
```

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -153,6 +154,24 @@ export function useNodePointerInteractions(
     }
   }
 
+  /**
```
christian-byrne, line 160: suggestion (non-blocking): No test coverage for the new `suppressNextClick` behavior.

The existing test file has thorough coverage for drag, selection, context menu, and modifier keys, but nothing for click suppression after drag. Key scenarios to consider: (1) click suppressed immediately after drag, (2) click NOT suppressed after non-drag pointer interaction, (3) listener cleanup when no click fires.

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
kaili-yang: Thanks for the suggestion! It's been applied and the issue is fixed.

PR #9368, opened 2026-03-04
Misc app mode fixes

AustinMroz:
A working branch of smaller app mode fixes. Can be merged at any time and I'll make a new branch.
- Selected inputs and outputs can now be re-ordered when clicking on label text
- 3d outputs once again display correctly
  - Some padding has been added to the side so that control buttons don't overlap with the floating app sidebar controls
- A "Share" button placeholder has been added to the menu, but is disabled
- Adds a workaround for canvas read_only state being disabled when 'space' is pressed.
  - This one is particularly hacky, and can be pulled out if problematic
- Fix download all only downloading the first output

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9368-Misc-app-mode-fixes-3196d73d365081eab02ad1e693784707) by [Unito](https://www.unito.io)


src/stores/appModeStore.ts
```diff
@@ -61,9 +61,20 @@ export const useAppModeStore = defineStore('appMode', () => {
     { deep: true }
   )
 
+  let unwatch: () => void | undefined
   watch(
     () => mode.value === 'builder:select',
```
pythongosssss, line 66: This is a bit horrible, but without guards in readOnly itself i'm not sure there is a better solution

pythongosssss (approved):

PR #9369, opened 2026-03-04
Evan/project marketplace

stackoverfloweth:
## Summary                                                                    
   
  Adds a "My Templates" section to the workflow templates modal with a full     
  publish wizard for submitting marketplace templates, including metadata entry,
   media uploads, and review/submission flow.

  ## Changes

  - **What**: Marketplace template publishing wizard (3-step: metadata → media →
   review), "My Templates" nav item with dashboard stats, template details right
   panel with edit/delete, workflow preview thumbnails before install, and
  supporting services for custom node/model detection and VRAM estimation
  - **Breaking**: None
  - **Dependencies**: None (uses existing Reka UI, PrimeVue, chart.js)

  ### Marketplace publishing (`src/platform/marketplace/`)
  - 3-step wizard: metadata (workflow selection, requirements auto-detection,
  description, difficulty, categories, tags, license), media
  (gallery/comparison/video upload variants with progress tracking), review
  (card preview + status management)
  - Mock API layer backed by localStorage (`mockKVStore`) for templates, media,
  tags, categories, and author data — ready to swap for real endpoints
  - Schema validation via Zod for each wizard step
  - Per-file upload progress indicators and error toasts on media step
  - Tag autocomplete with debounced server suggestions
  - "My Templates" dashboard with download/favorite/rating stats and bar chart
  - Template details side panel with edit, delete, and "revert to draft" actions

  ### Template browser enhancements
  - Workflow graph preview thumbnail shown in details panel before install
  - `WorkflowJsonDataSource` extracts minimap-compatible node/link data from raw
   workflow JSON
  - Approved marketplace templates appear in the main template grid

  ### Supporting utilities
  - `workflowCustomNodeDetectionService` — detects non-core node types from a
  graph
  - `workflowModelDetectionService` — detects required models via model-to-node
  mapping
  - `vramUtil` — extracts VRAM snapshot from system stats
  - `mockKVStore` — namespaced localStorage wrapper with typed collections

  ### Accessibility
  - Breadcrumb step navigation changed from `<a>` (unfocusable without href) to
  `<button>`
  - Media upload remove buttons visible on `focus-visible` (not just hover)
  - Focus ring on media mode variant selector buttons

  ## Review Focus

  - Mock API design in `services/` — are the interfaces right for the real
  backend?
  - Wizard state management via provide/inject (`usePublishTemplateWizard`) —
  appropriate scope?
  - `WorkflowJsonDataSource` link parsing handles both array-tuple and object
  formats from LiteGraph — edge cases?
- much of the complexity stems from `src/components/custom/widget/WorkflowTemplateSelectorDialog.vue` , which was not a simple component to begin with
- some lightweight composables, should be using @kitbag/query or tanstack to manage cache
- defined `MarketplaceTemplate` as a superset of `TemplateInfo`. That way we can still satisfy any components that expect to be given a template (like the cards)
    - added a key for my type guard
    - couldn’t be symbol for JSON reasons

## Cut from scope / Future enhancements

- Versioning
    - system defaults new templates to 1.0.0, not editable by user
    - if a user chooses to revert a previously published template, we would want to give the author a way to dictate (major|minor|patch) as well as offer some markdown release notes that could be displayed to user
    - too many unknowns, do we physically keep old versions and let users specify which they want to install? how can we confidently show user what changed? Do we have a diff for graphs?
    - current paradigm appears to be just part of name, we at least improved that.
- User feedback
    - users should have a means of flagging templates that are inappropriate, broken, or to request changes.
    - there’s a lack of infrastructure to support this. As I understand it, there is no way to tie a particular workflow back to the template that was installed to create the workflow.
- Admin view
    - unsure where it would physically live [platform.comfy.org/admin](http://platform.comfy.org/admin)?
    - this means there is no way to change status, request changes
    - would be easy to fast-follow, let users start submitting while we work on moderation tools
- Rich text editor
    - users can use markdown as part of the description field but it abdicates responsibility to the user to format properly. This is almost certainly a new dependency, which I wouldn’t want to add without first having a discussion with the team
    - doesn’t limit what the user can do, just how convenient they can do it
- Assets manager
    - opted for a pretty lightweight file manager
    - mostly to avoid additional state inside a modal with increasing scope
    - give metadata about assets (where they came from)
    - allow user to drag and drop, to upload AND to assign to open slots
    - allow user to upload many at once
    - add guardrails for file sizes, run through 3rd party to shrink, etc

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9369-Evan-project-marketplace-3196d73d36508175bdabe7792ff97238) by [Unito](https://www.unito.io)


dante01yoon (commented):
Suggestion: 
- I think your PR is too large to review, how about split into small chunks of PRs and link each other? 
- Why don't you first handle coderabbitai's  review for critical/major one?

src/platform/marketplace/services/templateApi.ts
```diff
@@ -0,0 +1,119 @@
...
+    return submitTemplate({
+      ...body,
+      ...draft
+    })
+  }
+
+  const template = collection.get(id)
+
+  if (!template) {
+    throw new Error('Template not found')
```
christian-byrne, line 107: suggestion: (non-blocking) would it be cleaner to report this expected lookup failure with `reportError()` and return `null`, matching the nullable return contract, instead of introducing `throw new Error(...)`? The same treatment is needed for the status guard at line 111 to satisfy `comfy/no-new-error-throw`.

src/utils/vramUtil.ts
```diff
@@ -0,0 +1,19 @@
+import type { SystemStats } from '@/schemas/apiSchema'
```
christian-byrne, line 1: suggestion: (non-blocking) could this import `SystemStatsResponse` from `@comfyorg/ingest-types` and use that generated response type for `extractVramSnapshot`? `@/schemas/apiSchema` is deprecated by the current-main lint rule, and the generated package exports `SystemStatsResponse`.

christian-byrne (commented):
Exact-head lint pass for the two current-main preflight findings.

PR #9382, opened 2026-03-04
fix: split perf report workflow for fork PR support

christian-byrne:
## Summary

Perf report workflow fails on fork PRs because `GITHUB_TOKEN` is read-only for forks, causing "Resource not accessible by integration" on the PR comment step.

## Changes

- **What**: Split `ci-perf-report.yaml` into a data-collection workflow + a `workflow_run`-triggered reporter (`pr-perf-report.yaml`), matching the existing `ci-size-data`/`pr-size-report` pattern. Added fork PR permissions guidance to `.github/AGENTS.md`.
- **ci-perf-report.yaml**: Removed the `report` job and `pull-requests: write` permission. Added PR metadata (number + base branch) artifact upload.
- **pr-perf-report.yaml** (new): Triggered by `workflow_run` on the perf workflow. Downloads metrics + metadata artifacts, generates report, posts PR comment with write permissions from the default-branch context.

## Review Focus

- The two-workflow split follows the same pattern as `ci-size-data.yaml` → `pr-size-report.yaml`, which already works for fork PRs.
- The `workflow_run` trigger runs in the base repo context per [GitHub Security Lab guidance](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/), so it safely has write permissions even for fork PRs.
- AGENTS.md guidance documents this pattern to prevent recurrence.

Fixes the failure seen in https://github.com/Comfy-Org/ComfyUI_frontend/actions/runs/22684230751/job/65763595989?pr=9380

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9382-fix-split-perf-report-workflow-for-fork-PR-support-3196d73d365081b29b35ed354e7789e2) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/pr-perf-report.yaml
+++ b/.github/workflows/pr-perf-report.yaml
@@ -33,13 +33,38 @@ jobs:
           run_id: ${{ github.event.workflow_run.id }}
           path: temp/perf-meta/
 
-      - name: Set PR number
-        id: pr-number
-        run: echo "content=$(cat temp/perf-meta/number.txt)" >> $GITHUB_OUTPUT
+      - name: Resolve and validate PR metadata
+        id: pr-meta
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const fs = require('fs');
+            const artifactPr = Number(fs.readFileSync('temp/perf-meta/number.txt', 'utf8').trim());
+            const artifactBase = fs.readFileSync('temp/perf-meta/base.txt', 'utf8').trim();
 
-      - name: Set base branch
-        id: pr-base
-        run: echo "content=$(cat temp/perf-meta/base.txt)" >> $GITHUB_OUTPUT
+            // Resolve PR from trusted workflow context
+            let pr = context.payload.workflow_run.pull_requests?.[0];
+            if (!pr) {
+              const { data: prs } = await github.rest.pulls.list({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                state: 'open',
+              });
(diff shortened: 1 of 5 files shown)


.github/workflows/pr-perf-report.yaml
```diff
@@ -0,0 +1,95 @@
...
+          path: temp/perf-meta/
+
+      - name: Resolve and validate PR metadata
+        id: pr-meta
+        uses: actions/github-script@v8
+        with:
+          script: |
+            const fs = require('fs');
+            const artifactPr = Number(fs.readFileSync('temp/perf-meta/number.txt', 'utf8').trim());
+            const artifactBase = fs.readFileSync('temp/perf-meta/base.txt', 'utf8').trim();
```
Myestery, line 44: Both pr-perf-report.yaml and updated pr-size-report.yaml validate artifactPr, but still trust artifactBase from untrusted artifacts and use it to fetch baseline data.
It's safer is to derive base from trusted PR context (pr.base.ref) after PR resolution instead of artifact content

Myestery (changes requested):
Coderabbit has some nice suggestions about finding the related PR the right way without risking undefined behaviour by loading just 30 PRs
Also, we can improve the way we reference whats base due to possibility of forks

.github/workflows/pr-perf-report.yaml
christian-byrne: fixed in e9814c8 - both workflows now validate artifactBase against pr.base.ref and use the trusted value from the API response

PR #9388, opened 2026-03-04
Support search filtering to dynamic input types

AustinMroz:
Previously, MatchType and Autogrow inputs would not be considered when filtering searchbox entires. For example, "Batch Images" would not show as a suggestion would dragging a noodle from a "Load Image" node.

This is resolved by adding a step during nodeDef registration to precalculate a list of all input types. This may have performance implications.
- Search filtering should be more performant
- Initial node registration will be slower
- There's additional memory cost to store this information on every node.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9388-Support-search-filtering-to-dynamic-input-types-3196d73d365081d9939eff5e167a7e83) by [Unito](https://www.unito.io)


christian-byrne (dismissed):

christian-byrne (dismissed):

Myestery (approved):
LGTM, was able to verify we don't really the the whitespace trimming suggested by coderabbit

PR #9395, opened 2026-03-04
Fix localization on share and hide entry

AustinMroz:
A placeholder share entry was added in #9368, but the localization for this share label was then removed in #9361.

This localization is re-added in a location that is less likely to be overwritten and the menu item is set to hidden. I'll manually connect it to the workflow sharing feature flag in a followup PR after that has been merged.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9395-Fix-localization-on-share-and-hide-entry-3196d73d36508146a343f625a5327bdd) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useWorkflowActionsMenu.ts
+++ b/src/composables/useWorkflowActionsMenu.ts
@@ -189,11 +189,10 @@ export function useWorkflowActionsMenu(
 
     addItem({
       id: 'share',
-      label: t('menuLabels.Share'),
+      label: t('breadcrumbsMenu.share'),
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):

AustinMroz:
Timeline has become a little funky since workflow sharing is now released. Will merge this PR as is (to make sure the fix for displaying a raw localization key gets in), and then batch the "conditional display of menu item by feature flag" into a bulk app-fixes PR.

Test is known flake.

PR #9403, opened 2026-03-05
fix: skip redundant appScalePercentage updates during zoom/pan

christian-byrne:
## What
Add equality check before updating `appScalePercentage` reactive ref.

## Why
Firefox profiler shows 586 `setElementText` markers from continuous text interpolation updates during zoom/pan. The rounded percentage value often doesn't change between events.

## How
Extract `updateAppScalePercentage()` helper with equality guard — compares new rounded value to current before assigning to the ref.

## Perf Impact
Expected: eliminates ~90% of `setElementText` markers during zoom/pan

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9403-fix-skip-redundant-appScalePercentage-updates-during-zoom-pan-31a6d73d3650812db8f2d68ac73c95b0) by [Unito](https://www.unito.io)


src/renderer/core/canvas/canvasStore.ts
```diff
@@ -43,6 +43,12 @@ export const useCanvasStore = defineStore('canvas', () => {
 
   // Reactive scale percentage that syncs with app.canvas.ds.scale
   const appScalePercentage = ref(100)
+  const updateAppScalePercentage = (scale: number) => {
+    const newValue = Math.round(scale * 100)
+    if (newValue !== appScalePercentage.value) {
```
pythongosssss: Vue already checks for equality before triggering reactivity, are you sure this is helping?

christian-byrne: you're right — Vue's ref setter uses `Object.is()` internally before triggering effects ([source](https://github.com/vuejs/core/blob/main/packages/reactivity/src/ref.ts#L141-L163)), so the manual `!==` guard was redundant for a primitive number.

removed it in f0735c7. the PR now just extracts `updateAppScalePercentage` as a named function to centralize the `Math.round` logic (was duplicated in 3 call sites).

benceruleanlu (approved):
The performance improvement disappeared! Also the CR comment on app.canvas.ds seems to be right, can you TAL?

PR #9404, opened 2026-03-05
fix: cache ctx.measureText results to avoid redundant calls in draw loop

christian-byrne:
## What
Add a per-frame text measurement cache for all hot-path ctx.measureText() calls.

## Why
drawTruncatingText() in BaseWidget calls ctx.measureText() per widget per frame with zero caching. For a 50-node workflow at 60fps: ~78,000-243,000 measureText calls/sec. Text labels rarely change between frames.

## How
Global Map<string, number> cache keyed by font+text, cleared once per frame at the start of drawFrontCanvas(). Replaces direct ctx.measureText() calls in BaseWidget.drawTruncatingText, draw.ts truncateTextToWidth/drawTextInArea, LGraphBadge.getWidth, LGraphButton.getWidth, and textUtils.truncateText.

## Perf Impact
Expected: ~95% reduction in measureText calls (only cache misses on first frame and value changes). Firefox has slower measureText than Chrome, so this disproportionately benefits Firefox.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9404-fix-cache-ctx-measureText-results-to-avoid-redundant-calls-in-draw-loop-31a6d73d3650814e9cdac16949c55cb7) by [Unito](https://www.unito.io)

### DIFF
--- a/.claude/skills/perf-fix-with-proof/SKILL.md
+++ b/.claude/skills/perf-fix-with-proof/SKILL.md
@@ -117,7 +117,7 @@ Implement the fix. The `@perf` test from PR1 is now on main and will run automat
 
 The `ci-perf-report.yaml` posts a comment like:
 
-```
+```markdown
 ## ⚡ Performance Report
 
 | Metric | Baseline | PR (n=3) | Δ | Sig |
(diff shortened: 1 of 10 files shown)


pythongosssss:
There look to be a couple of missed calls to measureText e.g. `LGraphCanvas._measureText` `LGraphNode.measure` , `LGraphCanvas.drawLinkTooltip`, if these were intentionally skipped lgtm

PR #9406, opened 2026-03-05
feat: Warn when binding browser-reserved shortcuts

### LINKED ISSUE #1087: [Keybinding] Show conflict keybinding with browser keybindings
Related PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/1081

Keybindings such as `Ctrl + t` are reserved by the browser for default behaviors. We should hint users that these keys are reserved by the browser.

Myestery:
## Summary

Show a non-blocking warning in the keybinding edit dialog when users try to bind shortcuts that browsers intercept (e.g. Ctrl+T, Ctrl+W, F12).

## Changes

- **What**: Add `RESERVED_BY_BROWSER` set of known browser-intercepted shortcuts, `isBrowserReserved` getter on `KeyComboImpl`, and a warning `<Message>` in the keybinding edit dialog. Users can still save the binding.

## Review Focus

Whether the list of browser-reserved shortcuts is comprehensive enough, and whether a non-blocking warning (vs blocking) is the right UX choice.

## Before

https://github.com/user-attachments/assets/5abfc062-5ed1-4fcd-b394-ff98221d82a8

## After

https://github.com/user-attachments/assets/12a49e24-051f-4579-894a-164dbf1cb7b7

Fixes #1087

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9406-feat-Warn-when-binding-browser-reserved-shortcuts-31a6d73d36508162a021e88ab76914f6) by [Unito](https://www.unito.io)


src/platform/keybindings/keyCombo.ts
```diff
@@ -61,6 +61,10 @@ export class KeyComboImpl implements KeyCombo {
     return this.shift && this.modifierCount === 1
   }
 
+  get isBrowserReserved(): boolean {
+    return RESERVED_BY_BROWSER.has(this.toString())
+  }
```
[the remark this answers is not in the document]
Myestery: - toString() calls getKeySequences() which pushes this.key directly (the raw event.key value)        
  - RESERVED_BY_BROWSER entries use lowercase letters like Ctrl + t, Ctrl + w, etc.                    
  - When a user presses Ctrl+T, event.key is "t" (lowercase) because the browser reports the key       
  without shift capitalization when Ctrl is held                                                       
  - When Shift is also held (e.g. Ctrl + Shift + n), the set explicitly includes that combination, and
  event.key is still "n" (lowercase) because Shift is a separate modifier

  The equals() method uses toUpperCase() for comparison because it compares two combos that may have
  been constructed differently. But toString() and the reserved set are consistently using event.key
  values, which are lowercase for letter keys with Ctrl held.

  No fix needed — the casing is already consistent between toString() output and the
  RESERVED_BY_BROWSER entries.

src/components/bottomPanel/BottomPanel.vue
```diff
@@ -25,15 +25,13 @@
                   class: {
                     'p-3 rounded-lg': true,
                     'pointer-events-none':
-                      bottomPanelStore.bottomPanelTabs.length === 1
-                  },
-                  style: {
-                    color: 'var(--fg-color)',
-                    backgroundColor:
+                      bottomPanelStore.bottomPanelTabs.length === 1,
+                    'bg-secondary-background text-secondary-foreground':
```
Myestery, line 29: Fixed a minor UI issue on keyboard tab control on light mode

src/platform/keybindings/keyCombo.ts
```diff
@@ -61,6 +61,10 @@ export class KeyComboImpl implements KeyCombo {
     return this.shift && this.modifierCount === 1
   }
 
+  get isBrowserReserved(): boolean {
+    return RESERVED_BY_BROWSER.has(this.toString())
```
pythongosssss: I think this has the same issue as isReservedByTextInput where because we are checking case sensitively the check can incorrectly pass based on e.g. user pressing shift or having caps lock on
We should either check actual objects for equality or make it case insensitive

src/components/dialog/content/setting/KeybindingPanel.vue
```diff
@@ -59,7 +59,15 @@
...
         <template #body="slotProps">
-          <div class="truncate" :title="slotProps.data.id">
+          <div
+            class="flex items-center gap-1.5 truncate"
+            :title="slotProps.data.id"
+          >
+            <i
+              v-if="slotProps.data.keybinding?.combo.isBrowserReserved"
+              v-tooltip="$t('g.browserReservedKeybindingTooltip')"
+              class="pi pi-exclamation-triangle shrink-0 text-amber-400"
```
pythongosssss: I think we should use lucid icon here, and is there a design token equivalent for amber e.g. warning

src/components/dialog/content/setting/keybinding/EditKeybindingContent.vue
```diff
@@ -0,0 +1,58 @@
...
+      :placeholder="$t('g.enterYourKeybind')"
+      :aria-label="$t('g.enterYourKeybind')"
+      autocomplete="off"
+      autofocus
+      @keydown.stop.prevent="captureKeybinding"
+    />
+    <div class="min-h-12">
+      <p
+        v-if="dialogState.newCombo?.isBrowserReserved"
+        class="m-0 text-sm text-destructive-background"
```
pythongosssss, line 22: should these be destructive or warning?

src/components/dialog/content/setting/keybinding/EditKeybindingFooter.vue
```diff
@@ -0,0 +1,72 @@
...
+
+import Button from '@/components/ui/button/Button.vue'
+import { KeybindingImpl } from '@/platform/keybindings/keybinding'
+import { useKeybindingService } from '@/platform/keybindings/keybindingService'
+import { useKeybindingStore } from '@/platform/keybindings/keybindingStore'
+import { useDialogStore } from '@/stores/dialogStore'
+
+import type { EditKeybindingDialogState } from '@/composables/useEditKeybindingDialog'
+
+const DIALOG_KEY = 'edit-keybinding'
```
pythongosssss: Duplicated key should be imported from the dialog composable

src/components/dialog/content/setting/keybinding/EditKeybindingContent.vue
Myestery: destructive according to design

src/components/dialog/content/setting/keybinding/EditKeybindingContent.vue
```diff
@@ -0,0 +1,66 @@
...
+
+const { dialogState, onUpdateCombo } = defineProps<{
+  dialogState: EditKeybindingDialogState
+  commandLabel: string
+  onUpdateCombo: (combo: KeyComboImpl) => void
+}>()
+
+const keybindingStore = useKeybindingStore()
+
+const existingKeybindingOnCombo = computed<KeybindingImpl | null>(() => {
```
pythongosssss: you look to compute this in both the content and footer, can you move it to the composable and pass in as a prop instead

Myestery: Done — moved `existingKeybindingOnCombo` to the composable as a single `computed` and pass it as a prop to both content and footer.

pythongosssss (approved):

PR #9408, opened 2026-03-05
refactor: unify image classification and fix cloud preview param handling

### LINKED ISSUE #9386: refactor: unify image classification between ResultItemImpl.isImage and appendCloudResParam IMAGE_EXTENSIONS
## Background

Flagged during review of PR #9298 (https://github.com/Comfy-Org/ComfyUI_frontend/pull/9298#discussion_r2886142886) by @christian-byrne.

## Problem

The `previewUrl` getter on `ResultItemImpl` (in `src/stores/queueStore.ts`) gates on `this.isImage` before calling `appendCloudResParam`. However, the two classification systems have diverging coverage:

- **`ResultItemImpl.isImage`**: Returns `true` if `isImageBySuffix` (gif/webp only) **OR** `mediaType === 'images'` (and not video/audio suffix).
- **`appendCloudResParam` / `IMAGE_EXTENSIONS`** (in `src/platform/distribution/cloudPreviewUtil.ts`): Uses its own independent set of 8 extensions — png, jpg, jpeg, webp, gif, bmp, tiff, tif.

For example, `isImageBySuffix` only covers `.gif` and `.webp`, while `IMAGE_EXTENSIONS` covers 8 extensions. The current behavior works because `isImage` falls through to `mediaType === 'images'`, but the dual classification logic makes it difficult to reason about which files actually receive the `&res=512` query parameter.

## Goal

Consolidate image classification into a single source of truth so that `ResultItemImpl.isImage` and `appendCloudResParam` agree on what constitutes an image. Consider whether `ResultItemImpl.isImage` should delegate to the same `IMAGE_EXTENSIONS` set used by `cloudPreviewUtil`, or vice versa.

## Relevant files

- `src/stores/queueStore.ts` — `ResultItemImpl.isImage`, `ResultItemImpl.isImageBySuffix`, `ResultItemImpl.previewUrl`
- `src/platform/distribution/cloudPreviewUtil.ts` — `appendCloudResParam`, `IMAGE_EXTENSIONS`

## References

- PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9298
- Comment: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9298#discussion_r2886142886

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9386-refactor-unify-image-classification-between-ResultItemImpl-isImage-and-appendCloudRes-3196d73d365081de999bdfa2eaa710b3) by [Unito](https://www.unito.io)

dante01yoon:
## Summary

Addresses review feedback from PR #9298 and resolves the divergence between `ResultItemImpl.isImage` and `appendCloudResParam`'s image classification.

### Changes

- **Unify suffix-based classification**: Replace narrow `isImageBySuffix` (gif/webp only), `isVideoBySuffix` (webm/mp4), and `isAudioBySuffix` with `getMediaTypeFromFilename()` from shared-frontend-utils, using the same `IMAGE_EXTENSIONS` set (png, jpg, jpeg, gif, webp, bmp, avif, tif, tiff) that `appendCloudResParam` uses
- **imageCompare.ts**: Pass `record.filename` to `appendCloudResParam` (was called without filename, bypassing image-extension guard)
- **imagePreviewStore.ts**: Use per-image `image.filename` instead of first image's filename for all images in batch
- **LinearControls.vue**: Use `resultItem.filename` (already a string) instead of `String(filename)` which converts undefined to `"undefined"`

### Related review comments

- [imageCompare.ts — missing filename](https://github.com/Comfy-Org/ComfyUI_frontend/pull/9298#discussion_r2886137498)
- [imagePreviewStore.ts — per-image filename](https://github.com/Comfy-Org/ComfyUI_frontend/pull/9298#discussion_r2886138718)
- [LinearControls.vue — String(filename)](https://github.com/Comfy-Org/ComfyUI_frontend/pull/9298#discussion_r2886140159)
- [queueStore.ts — diverging image classification](https://github.com/Comfy-Org/ComfyUI_frontend/pull/9298#discussion_r2886142886)

## Test plan

- [x] 66 unit tests pass (queueStore + cloudPreviewUtil)
- [x] `pnpm typecheck` passes
- [x] `pnpm lint` passes

- Fixes #9386


src/stores/queueStore.ts
```diff
@@ -159,11 +159,11 @@ export class ResultItemImpl {
   }
 
   get isVideoBySuffix(): boolean {
-    return this.isWebm || this.isMp4
+    return getMediaTypeFromFilename(this.filename) === 'video'
   }
 
   get isImageBySuffix(): boolean {
-    return this.isGif || this.isWebp
+    return getMediaTypeFromFilename(this.filename) === 'image'
```
christian-byrne, line 161: suggestion (non-blocking): Now that `isImageBySuffix` delegates to `getMediaTypeFromFilename`, the `isGif` and `isWebp` getters appear to be dead code -- they were only consumed by the old `isImageBySuffix` implementation. Consider removing them in a followup cleanup.

src/stores/queueStore.ts
```diff
@@ -159,11 +159,11 @@ export class ResultItemImpl {
   }
 
   get isVideoBySuffix(): boolean {
-    return this.isWebm || this.isMp4
+    return getMediaTypeFromFilename(this.filename) === 'video'
```
christian-byrne, line 157: suggestion (non-blocking): `isVideoBySuffix` now includes `.mov` and `.avi` via `VIDEO_EXTENSIONS`, but `htmlVideoType` (line 110-127) only maps `.webm` and `.mp4` to MIME types. A `.avi` file will pass `isVideo` and render in `ResultVideo.vue` with `type=undefined` -- browsers can't play AVI, so the user would see a broken video player. Consider either extending `htmlVideoType` to cover the new extensions, or scoping `VIDEO_EXTENSIONS` down to browser-playable formats.

src/stores/queueStore.ts
dante01yoon: Good catch — removed both in 9d92cb9.

src/stores/queueStore.ts
dante01yoon: Added `video/quicktime` MIME type for `.mov` in `htmlVideoType` (9d92cb9). Safari supports it natively, Chrome has partial support.

For `.avi`, left `htmlVideoType` returning `undefined` since no modern browser can decode it — the `<source>` element will attempt playback without a type hint and fail gracefully.

christian-byrne (approved):
LGTM!

PR #9409, opened 2026-03-05
feat: Add PostHog telemetry provider

robinjhuang:
Add PostHog as a telemetry provider for cloud builds so custom events can be correlated with session recordings. Follows the same pattern as MixpanelTelemetryProvider with dynamic import, event queuing, and disabled events from remote config. Tree-shaken away in OSS builds.

The posthog-js package uses Apache-2.0 (verified from its LICENSE file) but declares it as "SEE LICENSE IN LICENSE" in package.json, which   
  the license checker can't parse.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9409-feat-Add-PostHog-telemetry-provider-31a6d73d3650818b8e86c772c6551099) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/ci-oss-assets-validation.yaml
+++ b/.github/workflows/ci-oss-assets-validation.yaml
@@ -100,6 +100,7 @@ jobs:
             --production \
             --summary \
             --excludePackages '@comfyorg/comfyui-frontend;@comfyorg/design-system;@comfyorg/registry-types;@comfyorg/shared-frontend-utils;@comfyorg/tailwind-utils;@comfyorg/comfyui-electron-types' \
+            --clarificationsFile .github/license-clarifications.json \
             --onlyAllow 'MIT;MIT*;Apache-2.0;BSD-2-Clause;BSD-3-Clause;ISC;0BSD;BlueOak-1.0.0;Python-2.0;CC0-1.0;Unlicense;(MIT OR Apache-2.0);(MIT OR GPL-3.0);(Apache-2.0 OR MIT);(MPL-2.0 OR Apache-2.0);CC-BY-4.0;CC-BY-3.0;GPL-3.0-only'; then
             echo ''
             echo '✅ All production dependency licenses are approved!'
--- a/src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
+++ b/src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
@@ -1,29 +1,23 @@
-import type { OverridedMixpanel } from 'mixpanel-browser'
+import type { PostHog } from 'posthog-js'
 import { watch } from 'vue'
 
 import { useCurrentUser } from '@/composables/auth/useCurrentUser'
-import {
-  checkForCompletedTopup as checkTopupUtil,
-  clearTopupTracking as clearTopupUtil,
-  startTopupTracking as startTopupUtil
-} from '@/platform/telemetry/topupTracker'
-import type { AuditLog } from '@/services/customerEventsService'
-
-import { getExecutionContext } from '../../utils/getExecutionContext'
+import { remoteConfig } from '@/platform/remoteConfig/remoteConfig'
+import type { RemoteConfig } from '@/platform/remoteConfig/types'
 
 import type {
   AuthMetadata,
-  CreditTopupMetadata,
   EnterLinearMetadata,
   ExecutionContext,
-  ExecutionTriggerSource,
   ExecutionErrorMetadata,
   ExecutionSuccessMetadata,
+  ExecutionTriggerSource,
   HelpCenterClosedMetadata,
   HelpCenterOpenedMetadata,
   HelpResourceClickedMetadata,
   NodeSearchMetadata,
   NodeSearchResultMetadata,
+  PageViewMetadata,
   PageVisibilityMetadata,
   RunButtonProperties,
   SettingChangedMetadata,
@@ -41,9 +35,8 @@ import type {
   WorkflowCreatedMetadata,
   WorkflowImportMetadata
 } from '../../types'
-import { remoteConfig } from '@/platform/remoteConfig/remoteConfig'
-import type { RemoteConfig } from '@/platform/remoteConfig/types'
 import { TelemetryEvents } from '../../types'
+import { getExecutionContext } from '../../utils/getExecutionContext'
 import { normalizeSurveyResponses } from '../../utils/surveyNormalization'
 
 const DEFAULT_DISABLED_EVENTS = [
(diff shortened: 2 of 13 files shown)


robinjhuang:
@benceruleanlu is this a flaky test? Or a bug I introduced

image

benceruleanlu:
> @benceruleanlu is this a flaky test? Or a bug I introduced
> 
> image

Flaky. Snapshot tests in FE have been the biggest piece of tech debt for months.

christian-byrne (commented):
Looks good, can you add a query term to ensure it never gets included in the OSS distribution? Just 1-2 lines in this file https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/.github/workflows/ci-dist-telemetry-scan.yaml.

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
```diff
@@ -0,0 +1,379 @@
...
+
+  trackSettingChanged(metadata: SettingChangedMetadata): void {
+    this.trackEvent(TelemetryEvents.SETTING_CHANGED, metadata)
+  }
+
+  trackUiButtonClicked(metadata: UiButtonClickMetadata): void {
+    this.trackEvent(TelemetryEvents.UI_BUTTON_CLICKED, metadata)
+  }
+
+  trackPageView(pageName: string): void {
```
christian-byrne: issue: `trackPageView` drops the `properties` parameter defined in the `TelemetryProvider` interface.

The interface declares `trackPageView?(pageName: string, properties?: PageViewMetadata)` (see `types.ts:399`), and `TelemetryRegistry.dispatch` forwards both arguments. This implementation only accepts `pageName`, silently discarding `path`, `referrer`, and `title` metadata. `PageViewMetadata` also needs to be added to the type imports from `../../types`.

```suggestion
  trackPageView(pageName: string, properties?: PageViewMetadata): void {
    this.captureRaw(TelemetryEvents.PAGE_VIEW, { page_name: pageName, ...properties })
  }
```

.github/workflows/ci-oss-assets-validation.yaml
```diff
@@ -99,7 +99,7 @@ jobs:
           if npx license-checker-rseidelsohn@4 \
             --production \
             --summary \
-            --excludePackages '@comfyorg/comfyui-frontend;@comfyorg/design-system;@comfyorg/registry-types;@comfyorg/shared-frontend-utils;@comfyorg/tailwind-utils;@comfyorg/comfyui-electron-types' \
+            --excludePackages '@comfyorg/comfyui-frontend;@comfyorg/design-system;@comfyorg/registry-types;@comfyorg/shared-frontend-utils;@comfyorg/tailwind-utils;@comfyorg/comfyui-electron-types;posthog-js' \
```
christian-byrne: suggestion: `--excludePackages` bypasses all license checking for `posthog-js` and its transitive dependencies. posthog-js uses `"license": "SEE LICENSE IN LICENSE"` in its package.json (which is Apache-2.0), so the issue is just a non-SPDX string that the checker can't parse.

Consider using `--clarificationsFile` with a JSON file that maps `posthog-js` to `Apache-2.0` instead. That way transitive deps still get checked and CI will catch it if posthog-js ever changes their license. Example:

```json
{ "posthog-js@*": { "licenses": "Apache-2.0" } }
```

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
```diff
@@ -0,0 +1,379 @@
...
+
+  trackApiCreditTopupSucceeded(): void {
+    this.trackEvent(TelemetryEvents.API_CREDIT_TOPUP_SUCCEEDED)
+  }
+
+  trackRunButton(options?: {
+    subscribe_to_run?: boolean
+    trigger_source?: string
+  }): void {
+    this.captureRaw(TelemetryEvents.RUN_BUTTON_CLICKED, options)
```
christian-byrne: question (non-blocking): Several events like `trackRunButton` (2 fields) and `trackWorkflowExecution` (0 fields) send significantly less data than MixpanelTelemetryProvider's equivalents (which send ~19 fields via `getExecutionContext()`). Is there a decision framework for which events get rich properties vs minimal session-recording-correlation-only data? Asking because it would help future contributors know whether to add properties when implementing new events here.

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
```diff
@@ -0,0 +1,379 @@
...
+        void import('posthog-js')
+          .then((posthogModule) => {
+            this.posthog = posthogModule.default
+            this.posthog!.init(apiKey, {
+              api_host:
+                window.__CONFIG__?.posthog_api_host || 'https://ph.comfy.org',
+              autocapture: false,
+              capture_pageview: false,
+              capture_pageleave: false,
+              persistence: 'localStorage+cookie'
```
christian-byrne, line 107: question (non-blocking): `persistence: 'localStorage+cookie'` is more persistent than Mixpanel's `'cookie'` setting -- localStorage survives cookie clearing. Is the dual persistence intentional here, or would `'cookie'` be sufficient for the session-recording use case?

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
```diff
@@ -0,0 +1,379 @@
...
+            })
+            this.isInitialized = true
```
christian-byrne, line 109: question (non-blocking): `isInitialized` is set synchronously after `posthog.init()` inside the `.then()` callback, while Mixpanel defers it to the `loaded` callback (which fires after internal SDK setup). PostHog's `init()` is synchronous so this is likely fine, but was the timing difference considered?

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
robinjhuang: Fixed

.github/workflows/ci-oss-assets-validation.yaml
robinjhuang: Done!

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
robinjhuang: I looked this up and by default they track both so things work across subdomains. 
> By default, we store all this information in both a cookie and localStorage, which means PostHog can identify your users across subdomains.

https://posthog.com/docs/libraries/js/persistence

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
robinjhuang: I think it's fine because init is synchronous.

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
robinjhuang: great point, I refactored getExecutionContext into a shared function and call it from both places now.

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
christian-byrne: confirmed - PageViewMetadata is now imported and trackPageView accepts the properties parameter

.github/workflows/ci-oss-assets-validation.yaml
christian-byrne: looks good, clarificationsFile is in place with the Apache-2.0 mapping

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
christian-byrne: nice, shared getExecutionContext is a clean approach - both providers now get the same rich properties

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
christian-byrne: makes sense, thanks for looking it up

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
christian-byrne: agreed, synchronous init means no race condition here

christian-byrne (approved):
LGTM!

Issue #9410, opened 2026-03-05
[Bug] Missing models dialog lost Civitai file size fetch and HF gated model redirect


christian-byrne:
## Description

PR #9014 removed `useDownload.ts` and `useCivitaiModel.ts` during the missing models dialog redesign. Two behaviors were not carried over to the new implementation:

### 1. Civitai model file sizes no longer resolve

**Before:** `useDownload` detected Civitai URLs and called the [Civitai public API](https://developer.civitai.com/docs/api/public-rest) (`/api/v1/model-versions/{id}`) to get `sizeKB` from file metadata, since `HEAD` requests to Civitai download URLs do not return `content-length`.

**After:** `MissingModelsContent.vue` only does a `HEAD` request and silently swallows failures. Civitai models will never show a file size.

### 2. Gated HuggingFace models (e.g. Flux) no longer redirect to repo page

**Before:** `useDownload.triggerBrowserDownload()` checked `if (url.includes('huggingface.co') && error.value)` and redirected to the HF repo page (via `downloadUrlToHfRepoUrl()`) so users could accept the TOS/license agreement first.

**After:** `downloadModel()` in `missingModelsUtils.ts` always links directly to the download URL. For gated models like Flux, this will fail silently instead of guiding the user to accept the license.

## Relevant old code

- `src/composables/useDownload.ts` (removed in #9014)
- `src/composables/useCivitaiModel.ts` (removed in #9014)

## Relevant new code

- `src/components/dialog/content/missingModelsUtils.ts`
- `src/components/dialog/content/MissingModelsContent.vue`

## Steps to Reproduce

1. Load a workflow that references a Civitai-hosted model → file size column will be blank
2. Load a workflow that references a gated HF model (e.g. Flux) → download button will fail instead of opening HF repo page

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9410-Bug-Missing-models-dialog-lost-Civitai-file-size-fetch-and-HF-gated-model-redirect-31a6d73d365081d8abc6c66b2592b408) by [Unito](https://www.unito.io)

viva-jinyi:
#9415

christian-byrne:
Fixed in https://github.com/Comfy-Org/ComfyUI_frontend/pull/9415

PR #9415, opened 2026-03-05
[feat] Add model metadata fetching with loading skeleton and gated repo support

viva-jinyi:
## Summary
- Fetch file sizes via HEAD requests (HuggingFace) and Civitai API with caching and request deduplication
- Show skeleton loader while metadata is loading
- Display "Accept terms" link for gated HuggingFace models instead of download button

## Changes
- **`missingModelsUtils.ts`**: Add `fetchModelMetadata` with Civitai API support, HuggingFace gated repo detection, in-memory cache, and inflight request deduplication
- **`MissingModelsContent.vue`**: Add Skeleton loading state, gated model "Accept terms" link, extract `showSkeleton` helper
- **`missingModelsUtils.test.ts`**: Tests for HEAD/Civitai fetching, gated repo detection, caching, and deduplication
- **`main.json`**: Add `acceptTerms` i18n key

## Related Issues
https://github.com/Comfy-Org/ComfyUI_frontend/issues/9410
https://github.com/Comfy-Org/ComfyUI_frontend/issues/9412

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9415-feat-Add-model-metadata-fetching-with-loading-skeleton-and-gated-repo-support-31a6d73d36508127859efa0b3847505e) by [Unito](https://www.unito.io)


src/components/dialog/content/missingModelsUtils.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+vi.mock('@/utils/formatUtil', () => ({
+  isCivitaiModelUrl: (url: string) =>
+    url.includes('civitai.com/api/download/models/'),
+  downloadUrlToHfRepoUrl: (url: string) => {
+    const match = /^(.*?)(?:\/resolve\/|\/blob\/|$)/.exec(new URL(url).pathname)
+    const repoPath = match?.[1]?.replace(/^\//, '') || ''
+    return `https://huggingface.co/${repoPath}`
+  }
+}))
```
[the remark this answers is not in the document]
viva-jinyi: Good catch on the narrow mock. However, rather than widening the mock to match production patterns, I think the cleanest fix is to drop the mock entirely and use the real `isCivitaiModelUrl` implementation via `vi.importActual`. This avoids the mock drifting from production in the future.

Will apply this approach.

src/components/dialog/content/missingModelsUtils.ts
```diff
@@ -81,3 +82,92 @@ export function downloadModel(
...
+const inflight = new Map<string, Promise<ModelMetadata>>()
+
+async function fetchCivitaiMetadata(url: string): Promise<ModelMetadata> {
+  try {
+    const modelId = new URL(url).pathname.split('/').pop()
+    if (!modelId) return { fileSize: null, gatedRepoUrl: null }
+
+    const apiUrl = `https://civitai.com/api/v1/model-versions/${modelId}`
+    const res = await fetch(apiUrl)
+    if (!res.ok) return fetchHeadMetadata(url)
```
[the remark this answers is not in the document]
viva-jinyi: Valid point. The current `pathname.split('/').pop()` would extract a model ID (not a model-version ID) for `/api/v1/models/:id` URLs, leading to incorrect API calls.

Will fix by explicitly matching only the version-ID path patterns (`/api/download/models/:id` and `/api/v1/models-versions/:id`) and falling back to HEAD for unrecognized paths like `/api/v1/models/:id`.

src/components/dialog/content/missingModelsUtils.ts
```diff
@@ -81,3 +82,92 @@ export function downloadModel(
...
+    const size = response.headers.get('content-length')
+    return {
+      fileSize: size ? parseInt(size, 10) : null,
+      gatedRepoUrl: null
+    }
+  } catch {
+    if (url.includes('huggingface.co')) {
+      return { fileSize: null, gatedRepoUrl: downloadUrlToHfRepoUrl(url) }
+    }
+    return { fileSize: null, gatedRepoUrl: null }
```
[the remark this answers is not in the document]
viva-jinyi: Agreed — treating all non-OK responses as gated is too broad. Will narrow it to 401/403 only and remove gated detection from the catch path.

viva-jinyi:
Re: the suggestion to replace `onMounted` with `watch` for metadata fetching —

This dialog is controlled via `v-if` and gets fully mounted/unmounted each time it opens. The `missingModels` prop won't change while the dialog is open, so `onMounted` is sufficient and simpler. Adding `watch` + `runId` would be unnecessary complexity (YAGNI).

The other 3 review items are valid and will be addressed.

src/components/dialog/content/MissingModelsContent.vue
```diff
@@ -29,14 +29,24 @@
...
             <span
-              v-if="model.isDownloadable && fileSizes.get(model.url)"
-              class="text-xs text-muted-foreground"
+              v-else-if="model.isDownloadable && fileSizes.get(model.url)"
+              class="pl-1.5 text-xs text-muted-foreground"
             >
               {{ formatSize(fileSizes.get(model.url)) }}
             </span>
+            <a
+              v-if="gatedModelUrls.has(model.url)"
```
christian-byrne: issue: Broken v-if/v-else-if chain -- gated link renders alongside file size.

Line 40 starts a new `v-if` instead of continuing the skeleton/fileSize chain from line 32. When a model is both gated AND has a resolved fileSize, the size span and "Accept terms" link both render simultaneously.

Suggested fix: change `v-if` to `v-else-if`:
```html
<a v-else-if="gatedModelUrls.has(model.url)" ...>
```

src/components/dialog/content/missingModelsUtils.ts
```diff
@@ -81,3 +82,99 @@ export function downloadModel(
...
+
+const metadataCache = new Map<string, ModelMetadata>()
+const inflight = new Map<string, Promise<ModelMetadata>>()
+
+async function fetchCivitaiMetadata(url: string): Promise<ModelMetadata> {
+  try {
+    const pathname = new URL(url).pathname
+    const versionIdMatch =
+      pathname.match(/^\/api\/download\/models\/(\d+)$/) ??
+      pathname.match(/^\/api\/v1\/models-versions\/(\d+)$/)
```
christian-byrne, line 108: issue: `/api/v1/models/:id` URLs pass `isCivitaiModelUrl` but no regex here handles them.

`isCivitaiModelUrl` (formatUtil.ts) accepts three patterns: `/api/download/models/:id`, `/api/v1/models/:id`, and `/api/v1/models-versions/:id`. But only the first and third have regexes here -- so `/api/v1/models/:id` falls through to `fetchHeadMetadata`, which is the broken HEAD path this PR fixes.

Additionally, `/api/v1/models/:id` is a model ID, not a model-version ID, so calling `/api/v1/model-versions/{id}` would return wrong data. Either add a third regex with the correct API endpoint (`/api/v1/models/{id}`), or exclude that pattern from `isCivitaiModelUrl`.

src/components/dialog/content/missingModelsUtils.ts
```diff
@@ -81,3 +82,99 @@ export function downloadModel(
...
+    if (!versionIdMatch) return fetchHeadMetadata(url)
+
+    const [, modelVersionId] = versionIdMatch
+    const apiUrl = `https://civitai.com/api/v1/model-versions/${modelVersionId}`
+    const res = await fetch(apiUrl)
+    if (!res.ok) return fetchHeadMetadata(url)
+
+    const data: CivitaiModelVersionResponse = await res.json()
+    const matchingFile = data.files?.find(
+      (file) => file.downloadUrl && url.startsWith(file.downloadUrl)
```
christian-byrne: issue: `url.startsWith(file.downloadUrl)` is backwards.

Civitai API downloadUrls often include query params (e.g. `?type=Model&format=SafeTensor`), making them longer than the bare input URL. This means `url.startsWith(file.downloadUrl)` will be false when the downloadUrl has a suffix the input URL lacks.

Consider: `file.downloadUrl.startsWith(url)` or compare after stripping query params.

src/components/dialog/content/missingModelsUtils.ts
```diff
@@ -81,3 +82,99 @@ export function downloadModel(
...
+const inflight = new Map<string, Promise<ModelMetadata>>()
+
+async function fetchCivitaiMetadata(url: string): Promise<ModelMetadata> {
+  try {
+    const pathname = new URL(url).pathname
+    const versionIdMatch =
+      pathname.match(/^\/api\/download\/models\/(\d+)$/) ??
+      pathname.match(/^\/api\/v1\/models-versions\/(\d+)$/)
+
+    if (!versionIdMatch) return fetchHeadMetadata(url)
```
christian-byrne: suggestion: Falling back to `fetchHeadMetadata` for Civitai URLs defeats the purpose.

HEAD requests to Civitai download URLs don't return `content-length` -- that's the regression this PR fixes. When the regex doesn't match or the API fails, the fallback silently returns `{ fileSize: null, gatedRepoUrl: null }` via HEAD anyway.

Consider returning `{ fileSize: null, gatedRepoUrl: null }` directly for known Civitai URLs instead of making a doomed HEAD request.

src/components/dialog/content/MissingModelsContent.vue
```diff
@@ -150,23 +163,30 @@ const totalDownloadSize = computed(() =>
     .reduce((total, model) => total + (fileSizes.get(model.url) ?? 0), 0)
 )
 
+const gatedModelUrls = reactive(new Map<string, string>())
+
+function showSkeleton(model: ProcessedModel): boolean {
+  return (
+    model.isDownloadable &&
+    loading.value &&
+    !fileSizes.get(model.url) &&
```
christian-byrne: suggestion (non-blocking): `!fileSizes.get(model.url)` is falsy when fileSize === 0.

If a file legitimately has size 0, `showSkeleton()` would remain true forever. `!fileSizes.has(model.url)` avoids the edge case.

src/components/dialog/content/MissingModelsContent.vue
```diff
@@ -150,23 +163,30 @@ const totalDownloadSize = computed(() =>
...
-      try {
-        const response = await fetch(url, { method: 'HEAD' })
-        if (!response.ok) return
-        const size = response.headers.get('content-length')
-        if (size) fileSizes.set(url, parseInt(size, 10))
-      } catch {
-        // Silently skip size fetch failures
-      }
+      const metadata = await fetchModelMetadata(url)
+      if (metadata.fileSize) fileSizes.set(url, metadata.fileSize)
```
christian-byrne: suggestion (non-blocking): `if (metadata.fileSize)` skips fileSize === 0 the same way.

Same falsy-zero issue as `showSkeleton`. Use `metadata.fileSize !== null` for a stricter check.

src/components/dialog/content/missingModelsUtils.test.ts
```diff
@@ -0,0 +1,151 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest'
+
+import { fetchModelMetadata } from './missingModelsUtils'
+
+const fetchMock = vi.fn()
+vi.stubGlobal('fetch', fetchMock)
+
+vi.mock('@/utils/formatUtil', async () => {
+  return await vi.importActual('@/utils/formatUtil')
+})
```
christian-byrne: nitpick (non-blocking): `vi.mock('@/utils/formatUtil', async () => await vi.importActual(...))` is a passthrough that re-imports the actual module. It can be removed unless you plan to spy on specific exports.

browser_tests/assets/missing/missing_models_real_urls.json
```diff
@@ -0,0 +1,90 @@
+{
```
christian-byrne: question: This fixture isn't referenced by any test in this PR. Is it wired to a browser test not included here, or can it be removed?

src/components/dialog/content/MissingModelsContent.vue
viva-jinyi: Good catch — fixed. Changed to `v-else-if` so the gated link properly chains with the conditions above and doesn't render simultaneously with the download button.

src/components/dialog/content/missingModelsUtils.ts
viva-jinyi: Agreed — the regex now properly matches `/api/download/models/:id` and `/api/v1/model-versions/:id`. For unrecognized Civitai URL patterns (like `/api/v1/models/:id`), we now return null directly instead of falling through to a HEAD request.

src/components/dialog/content/missingModelsUtils.ts
viva-jinyi: You're right, this was a real bug. Civitai's `downloadUrl` can include query parameters (e.g. `?type=Model&format=SafeTensor`), so `url.startsWith(file.downloadUrl)` would fail. Fixed to `file.downloadUrl.startsWith(url)`.

src/components/dialog/content/missingModelsUtils.ts
viva-jinyi: Agreed — the HEAD fallback was unnecessarily optimistic. Now returns `{ fileSize: null, gatedRepoUrl: null }` directly when the version ID pattern doesn't match or the API call fails.

src/components/dialog/content/MissingModelsContent.vue
viva-jinyi: Fixed — `fileSizes.get()` returns `0` which is falsy, so the skeleton would never disappear for zero-size entries. Changed to `fileSizes.has()` to check presence instead of truthiness.

src/components/dialog/content/MissingModelsContent.vue
viva-jinyi: Fixed — `if (metadata.fileSize)` would skip storing `0`. Changed to `if (metadata.fileSize !== null)` for an explicit null check.

src/components/dialog/content/missingModelsUtils.test.ts
viva-jinyi: Removed — it was a no-op mock that just re-imported the actual module.

browser_tests/assets/missing/missing_models_real_urls.json
viva-jinyi: This was a fixture I created for local testing. Removed it — good catch.

christian-byrne (approved):
LGTM

PR #9426, opened 2026-03-05
fix: stabilize flaky screenshot tests that rely on search box timings

christian-byrne:
## Summary

Fixes flaky screenshot test that was identified during the PR #9400 snapshot update, where baselines regenerated without any code changes affecting them.

**Root cause:** `fillAndSelectFirstNode('KSampler')` selected `nth(0)` from the autocomplete dropdown, but search result ordering is non-deterministic when the search box opens via link release with filter chips. Sometimes 'Preview Image' appeared as the first result instead of 'KSampler', causing a completely different node to be added.

**Fix:** Added an `exact: true` option to `fillAndSelectFirstNode` that uses an `aria-label` selector to click the specific matching result instead of blindly selecting the first item.

- Fixes #4658

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9426-fix-stabilize-flaky-screenshot-tests-for-search-results-and-image-preview-31a6d73d365081598167ce285416995c) by [Unito](https://www.unito.io)


benceruleanlu (approved):
Note: loadWorkflowInMedia.spec.ts wasn't modified here but was mentioned in the PR description. There are no new snapshots here. Stabilization LGTM.

christian-byrne:
> Note: loadWorkflowInMedia.spec.ts wasn't modified here but was mentioned in the PR description. There are no new snapshots here. Stabilization LGTM.

Good point, thank you. Update the PR descript/title.

---

@coderabbitai this was the part of the PR description I removed, we didn't end up doing it in this PR:

> ### 2. `no-workflow-webp` (loadWorkflowInMedia.spec.ts)
> 
> **Root cause:** Dropping `no_workflow.webp` (which has no embedded workflow) creates a LoadImage node and uploads the file. The image preview then loads asynchronously via the `/view` API. The screenshot was taken before the upload and preview load completed, causing the preview to sometimes show the loaded image and sometimes show a black/empty area.
> 
> **Fix:** Wait for the upload response and the subsequent `/view` response before taking the screenshot, ensuring the image preview always renders before comparison.

Can you make a followup/tracking Issue to do this later and assign it to me?

PR #9437, opened 2026-03-05
fix(ErrorNodeCard): show error message body in compact (single-node) mode

jaeone94:
## Summary
Remove the `!compact` condition that was preventing the error message body from being displayed in compact (single-node error) mode.

## Changes
- **What**: Removed `&& !compact` guard from `v-if="error.message && !compact"` in `ErrorNodeCard.vue` so the error message body is always shown when present, regardless of display mode

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9437-fix-ErrorNodeCard-show-error-message-body-in-compact-single-node-mode-31a6d73d3650814683f7e6983534a4ad) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/rightSidePanel/errors/ErrorNodeCard.vue
+++ b/src/components/rightSidePanel/errors/ErrorNodeCard.vue
@@ -49,7 +49,7 @@
       >
         <!-- Error Message -->
         <p
-          v-if="error.message && !compact"
+          v-if="error.message"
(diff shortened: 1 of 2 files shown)


viva-jinyi (approved):
LGTM

christian-byrne (approved):

PR #9439, opened 2026-03-05
feat/fix: App mode QA updates

pythongosssss:
## Summary

Various fixes from app mode QA

## Changes

- **What**: 
- fix: prevent inserting nodes from workflow/apps sidebar tabs
- fix: hide json extension in workflow tab
- fix: hide apps nav button in apps tab when already in apps mode
- fix: center text on arrange page
- fix: prevent IoItems from "jumping" due to stale transform after drag and drop op
- fix: refactor side panels and add custom stable pixel based sizing
- fix: make outputs/inputs lists in app builder scrollable
- fix: fix rerun not working correctly

- feat: add text to interrupt button
- feat: add enter app mode button to builder toolbar
- feat: add tooltip to download button on linear view
- feat: show last output of workflow in arrange tab if available
- feat: show download count in download all button, hide if only 1 asset to download

## Review Focus

- Rerun - I am not sure why it was triggering widget actions, removing it seemed like the correct fix
- useStablePrimeVueSplitter - this is a workaround for the fact it uses percent sizing, I also tried switching to reka-ui splitters, but they also only support % sizing in our version [pixel based looks to have been added in a newer version, will log an issue to upgrade & replace splitters with this]

image
image
image
image
image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9439-feat-fix-App-mode-QA-updates-31a6d73d365081b38337d63207b88817) by [Unito](https://www.unito.io)


src/renderer/extensions/linearMode/ImagePreview.vue
```diff
@@ -1,19 +1,27 @@
...
 const imageRef = useTemplateRef('imageRef')
 const width = ref('')
 const height = ref('')
 </script>
 <template>
-  <ZoomPane v-if="!mobile" v-slot="slotProps" class="flex-1 w-full">
+  <ZoomPane
+    v-if="!mobile"
+    v-slot="slotProps"
+    :class="cn('flex-1 w-full', attrs.class as string)"
```
AustinMroz: You can use `$attrs` instead of the importing `useAttrs`
```suggestion
    :class="cn('flex-1 w-full', $attrs.class as string)"
```

As a whole, I'm confused for why the attrs changes are needed in the first place. Both from my understanding of vue, and quick testing, the padding changes in LinearArrange will still correctly propagate downwards.

AustinMroz (dismissed):
Splitter changes are spooky, but work well from my testing.

---

Didn't know that reka-UI was adding pixel sizes. That's super exciting.

When I last attempted a migration, there was some bugginess where "setting a size well bellow the minimum content size for an area (as we did with 1-98-1) would require a panel be dragged the distance of this discrepency before resizing would visually occur."

Since pixel sizes would solve this, we may be able to migrate in the future.

src/renderer/extensions/linearMode/ImagePreview.vue
pythongosssss: Fixed

AustinMroz (approved):

pythongosssss:
@coderabbitai add a new issue to look at upgrading reka-ui and replacing PrimeVue splitters so we can use pixel sizing https://reka-ui.com/docs/components/splitter#pixel-sizing

AustinMroz (approved):

PR #9442, opened 2026-03-05
test: regenerate screenshot expectations

DrJKL:
The pixels fade, the tests turn red,
Old screenshots haunt us from the dead.
A branch is born, a label placed,
And golden images are replaced.

The shards spin up, four workers strong,
To right what rendering got wrong.
When CI turns green, the deed is done—
New expectations, freshly won.


Myestery (approved):
Nice poem

PR #9446, opened 2026-03-05
feat: add syntax highlighting for comments in ClipTextEncode textarea

### LINKED ISSUE #697: [Feature Request]: Simple syntax highlighting for comments in prompt or other windows

- [X] I have searched the existing issues and checked the recent builds/commits

In case I forget later, using comments instead of deleting prompts, is the only way to experiment. But it can get a bit busy and difficult to find certain pieces. This can be mitigated by lots of breaks and spaces, but I'd rather really like to tell at a glance which prompts are actually active at any time.

### Proposed workflow

1. Comment is made
2. It's green.
3. Profit.

### Additional information

_No response_

sonnybox:
# feat: add syntax highlighting for comments in ClipTextEncode (#697)

- Highlights line and multiline comments in green
- Highlights incomplete comments in yellow (stray '/*' or '*/')
- Adds experimental toggle in settings to enable/disable (off by default)
- For nodes 2.0 only
- Closes #697

## Summary

Adds syntax highlighting for comments in the textarea of CLIP text encoding nodes.

## Changes

- **What**: 
  - Created a new `useTextareaHighlighting` composable that parses text into a lightweight AST to identify comments.
  - Modified `WidgetTextarea.vue` to add a overlay `div` that mirrors the textarea content with `<span>` highlighting.
  - Added a new experimental setting `Comfy.TextareaWidget.HighlightClipComments` to toggle the feature.
  - Updated `coreSettings.ts` to support the new configuration option.
- **Dependencies**: None.

## Review Focus

- **AST Parsing Logic**: The logic in `useTextareaHighlighting.ts` handles nested text and validates block comment pairs. It provides a "yellow" state for unclosed / stray delimiters.
- **Scroll Synchronization**: The `syncScroll` function tracks the native textarea during scrolling.
- **UI**: The overlay uses `pointer-events-none` and matches font/padding/width. It should have no effect on typing experience.

### Textarea Nodes
<p align="center">
  syntax-highlight-1
  <br>
  syntax-highlight-2
  <br>
  syntax-highlight-3
</p>

### Settings Toggle
<p align="center">
  settings-toggle
</p>

---
**Fixes #697**

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9446-feat-add-syntax-highlighting-for-comments-in-ClipTextEncode-textarea-31a6d73d36508126aa1bfa1b88e04b42) by [Unito](https://www.unito.io)


christian-byrne:
This is a high quality PR. Would you take a look at the conflicts when you get a moment.

I have some questions as well:

- Which encoders or models support comments (i.e., strip them) / annotations? Or is this a ComfyUI feature already that I forgot about at some point (likely)? Excuse my ignorance there.
- Would it be cleaner to unify the div/span overlay with a single component and v-if/v-else approach? "No" as an answer is fine.

christian-byrne:
FYI GitHub actions is currently experiencing issues, not necessarily a fault of your code that the CI checks are failing/stalling.

sonnybox:
> Which encoders or models support comments (i.e., strip them) / annotations? Or is this a ComfyUI feature already that I forgot about at some point (likely)? Excuse my ignorance there.

I've accidentally discovered ComfyUI supported commenting through the wiki: [https://comfyui-wiki.com](https://comfyui-wiki.com/en/tutorial/basic/stable-diffusion-prompt-basic#comment-support). I however did not check if this is universally implemented in all CLIP Text Encode nodes, I've only tested the generic one:

| All commented | Dog | Cat |
| :---: | :---: | :---: |
| image | image | image |

> Would it be cleaner to unify the div/span overlay with a single component and v-if/v-else approach? "No" as an answer is fine.

As far as I know we cannot apply spans in textareas so we need the textarea to type in and then another element to render the colored text, so I would say no. 

If I understand you correctly, making it a switch would mean we lose the ability to edit while seeing the highlights (I also thought about only rendering the highlights when the user isn't focused in the textarea). I thought about creating a new 'textarea' like widget without 'textarea' itself but I think that is over complicating things.

src/renderer/extensions/vueNodes/widgets/composables/useTextareaHighlighting.ts
```diff
@@ -0,0 +1,130 @@
...
+        } else {
+          // Unmatched closing bracket
+          flushText()
+          const node: AstNode = {
+            type: 'block',
+            valid: false,
+            children: [...current.children]
+          }
+          node.children.push({ type: 'text', value: '*/', children: [] })
+          current.children = [node]
```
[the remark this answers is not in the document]
sonnybox: This is desired, foo should be yellow `foo /* ok */ bar */` to mark where you could add a opening `/*`.

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -27,17 +27,37 @@
...
       @contextmenu.capture.stop
     />
+    <div
+      v-if="isClipTextEncode"
+      ref="overlayRef"
+      aria-hidden="true"
+      :class="
+        cn(
+          'absolute inset-0 z-11 pointer-events-none',
+          'text-xs whitespace-pre-wrap wrap-break-word font-[monospace]',
```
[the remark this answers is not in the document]
sonnybox: wrap-break-word exists in tailwindcss 4.2: https://tailwindcss.com/docs/overflow-wrap#wrapping-mid-word

sonnybox:
@christian-byrne I had trouble finding where the textarea font is set, for now I just set the textarea and overlay to `font-mono`, let me know where I can find the real font class!

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -27,17 +27,37 @@
...
+          'border border-transparent px-2.75 pb-5',
+          !hideLayoutField ? 'pt-5' : 'pt-2',
+          'overflow-y-auto',
+          '[scrollbar-color:transparent_transparent] [&::-webkit-scrollbar-thumb]:bg-transparent [&::-webkit-scrollbar-track]:bg-transparent'
+        )
+      "
+      v-html="sanitizedHighlightedText"
```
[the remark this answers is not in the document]
sonnybox: Addressed in latest commit

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -17,27 +17,49 @@
...
+          !hideLayoutField && 'pt-5',
+          isClipTextEncode &&
+            'selection:text-transparent selection:bg-blue-500/50'
         )
       "
       :placeholder
       :readonly="isReadOnly"
+      :style="isClipTextEncode ? '-webkit-text-fill-color: transparent;' : ''"
       data-capture-wheel="true"
+      @scroll.passive="syncScroll"
```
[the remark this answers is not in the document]
sonnybox: Disabled highlighting while composing

christian-byrne (commented):
Overall this is a nice feature. One additional note: the `useTextareaHighlighting` composable has non-trivial AST parsing logic with multiple branches (line comments, block comments, unmatched delimiters) but no unit tests. Consider adding tests for the parser, especially edge cases like nested delimiters and unmatched markers.

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -17,22 +17,43 @@
...
       @contextmenu.capture.stop
     />
+    <div
+      v-if="showHighlight"
+      ref="overlayRef"
+      aria-hidden="true"
+      :class="
+        cn(
+          'pointer-events-none absolute inset-0',
+          'overflow-y-auto font-[monospace] text-xs wrap-break-word whitespace-pre-wrap',
```
christian-byrne: issue: The overlay div uses `font-[monospace]` but the `<Textarea>` component inherits the default sans-serif system font (no `font-family` set in the base `Textarea.vue`). Since the overlay is absolutely positioned over the textarea, this font mismatch will cause highlighted spans to drift out of alignment with the actual text as content grows.

Either add `font-[monospace]` to the textarea as well (when `showHighlight` is true), or remove it from the overlay so both use the same inherited font.

src/renderer/extensions/vueNodes/widgets/composables/useTextareaHighlighting.ts
```diff
@@ -0,0 +1,130 @@
...
+          i += 2
```
christian-byrne: issue (non-blocking): When an unmatched `*/` is encountered with no open `/*`, this wraps all of `current.children` (all previously parsed nodes) into an invalid block. This causes all preceding text to render yellow, not just the stray `*/` marker.

For example, `hello world */` would highlight everything yellow rather than only the `*/`.

Consider only wrapping the `*/` text itself in the invalid block node instead of the preceding children.

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -49,7 +70,9 @@
 </template>
 
 <script setup lang="ts">
-import { computed, useId } from 'vue'
+import { computed, useId, ref, watch, nextTick } from 'vue'
+import dompurify from 'dompurify'
```
christian-byrne: suggestion (non-blocking): All other usages in the codebase use `import { default as DOMPurify } from 'dompurify'` (PascalCase). This import uses lowercase `dompurify` which is inconsistent.

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -63,17 +86,38 @@ import {
...
 const modelValue = defineModel<string>({ default: '' })
 
 const hideLayoutField = useHideLayoutField()
 const { copyToClipboard } = useCopyToClipboard()
 
+const settingStore = useSettingStore()
+
+const isComposing = ref(false)
+
+const isClipTextEncode = computed(() => {
```
christian-byrne: nitpick (non-blocking): `isClipTextEncode` name is misleading since it also checks the setting, not just the node type. A name like `shouldHighlight` or `isHighlightEnabled` would better convey that it combines both conditions.

src/renderer/extensions/vueNodes/widgets/composables/useTextareaHighlighting.ts
```diff
@@ -0,0 +1,130 @@
...
+        flushText()
+        const node: AstNode = { type: 'line', value: '//', children: [] }
+        current.children.push(node)
+        i += 2
+
+        let lineText = ''
+        while (i < text.length && text[i] !== '\n') {
+          lineText += text[i]
+          i++
+        }
```
christian-byrne: nitpick (non-blocking): The character-by-character string building (`lineText += text[i]`, `currentText += text[i]`) could use `indexOf` to find the next delimiter and `slice` to grab the substring in one step, which would be more efficient on large prompts.

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
sonnybox: In my system they inherit different fonts, not sure if this is consistent behavior. Just to make sure, I only apply a font face only when highlighting is in use, in this case I choose font-mono to both so the theme can set the font.

src/renderer/extensions/vueNodes/widgets/composables/useTextareaHighlighting.ts
sonnybox: That’s a good point. I initially wanted it this way to act as a clear visual indicator for the user to end commenting, but this can be confusing. I'll update it so only the stray */ is wrapped in the invalid block.

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
sonnybox: I used inspector and it looks like the base font is just 'monospace' so I will use that instead

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
sonnybox: Updated

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
sonnybox: Good point, updated

src/renderer/extensions/vueNodes/widgets/composables/useTextareaHighlighting.ts
sonnybox: Updated

PR #9461, opened 2026-03-06
fix: track nodePreviewImages in usePromotedPreviews for GLSL live preview propagation

christian-byrne:
## Summary

Fix GLSL shader live preview propagation to SubgraphNode parents, broken after the promoted widget rewrite in #8856.

## Changes

- **What**: Add `nodePreviewImages` as a reactive dependency in `usePromotedPreviews`'s computed property. The GLSL live preview renderer writes to `nodePreviewImages` (via `setNodePreviewsByNodeId`), not `nodeOutputs`, so the computed never re-evaluated. The gate now passes if either store has data. `getNodeImageUrls` already handles the fallback correctly (checks previews first, then outputs).

## Review Focus

The fix is a one-line gate change (`||` instead of only checking `nodeOutputs`). Both reactive refs are read to establish Vue dependency tracking, matching the existing pattern.

N/A — preview rendering is identical to pre-#8856 behavior; this restores broken reactivity.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9461-fix-track-nodePreviewImages-in-usePromotedPreviews-for-GLSL-live-preview-propagation-31b6d73d3650818d9c5ff4b9fb5349ad) by [Unito](https://www.unito.io)


AustinMroz (dismissed):

christian-byrne:
Closed in favor of https://github.com/Comfy-Org/ComfyUI_frontend/pull/10165

PR #9463, opened 2026-03-06
fix: allow URI drops to bubble from Vue nodes to document handler

christian-byrne:
## Summary

Fix URI drops (e.g. dragging `` thumbnails) onto Vue-rendered nodes by letting unhandled drops bubble to the document-level `text/uri-list` fallback in `app.ts`.

## Changes

- **What**: Removed unconditional `.stop` modifier from `@drop` in `LGraphNode.vue`. `stopPropagation()` is now called conditionally — only when `onDragDrop` returns `true` (file drop handled). Made `handleDrop` synchronous since `onDragDrop` returns a plain boolean.

## Review Focus

The key insight is that `onDragDrop` (from `useNodeDragAndDrop`) returns `false` synchronously for URI drags (no files in `DataTransfer`), so the event must bubble to reach the document handler that fetches the URI. The original `async` + `await` pattern would have deferred `stopPropagation` past the synchronous propagation phase, so `handleDrop` is now synchronous.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9463-fix-allow-URI-drops-to-bubble-from-Vue-nodes-to-document-handler-31b6d73d36508196a1b3f17e7e4837a9) by [Unito](https://www.unito.io)


browser_tests/tests/loadWorkflowInMedia.spec.ts
```diff
@@ -67,5 +67,37 @@ test.describe(
...
+
+      const dropPosition = {
+        x: box!.x + box!.width / 2,
+        y: box!.y + box!.height / 2
+      }
+
+      await comfyPage.dragDrop.dragAndDropURL(
+        'https://comfyanonymous.github.io/ComfyUI_examples/hidream/hidream_dev_example.png',
+        { dropPosition, preserveNativePropagation: true }
+      )
```
dante01yoon: Question:  If GitHub Pages is down, the file is moved, or CI has restricted network access, this test will flake. The existing URL drop test in the same file (Load workflow from URL) has the same pattern — but this new test compounds the risk. Using page.route() to intercept the fetch and serve a local fixture would make it hermetic.

christian-byrne: Excellent idea. Yeah if we had merged this as-is, this almost certainly would have been a flaky test in the long-run. Good call.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -32,7 +32,7 @@
     @contextmenu="handleContextMenu"
     @dragover.prevent="handleDragOver"
     @dragleave="handleDragLeave"
-    @drop.stop.prevent="handleDrop"
+    @drop.prevent="handleDrop"
```
dante01yoon, line 36: **Issue**: `onDragDrop` is typed as `Promise<boolean> | boolean` (see `litegraph-augmentation.d.ts:149`). The synchronous truthiness check `if (handled)` would evaluate to `true` for **any** Promise return — even `Promise.resolve(false)` — because a Promise object is truthy. This would incorrectly call `stopPropagation()` and break URI drop bubbling for any extension that returns a Promise from `onDragDrop`.

**Context**: Today only `useNodeDragAndDrop` assigns `onDragDrop` (returns plain boolean), but the type contract allows async implementations. The document handler in `app.ts:588` properly uses `await`.

**Suggestion**: Use strict equality to exclude Promise objects:
```ts
const handled = node.onDragDrop(event)
if (handled === true) event.stopPropagation()
```
This is safe because a `Promise<boolean>` will never be `=== true`, only a plain `true` will match.

src/renderer/extensions/vueNodes/components/LGraphNode.test.ts
```diff
@@ -251,4 +253,48 @@ describe('LGraphNode', () => {
...
+      const onDragDrop = vi.fn().mockReturnValue(true)
+      mockData.mockLgraphNode = {
+        onDragDrop,
+        onDragOver: vi.fn(),
+        isSubgraphNode: () => false
+      }
+
+      const wrapper = mountLGraphNode({ nodeData: mockNodeData })
+
+      const parentListener = vi.fn()
```
dante01yoon, line 310: nit-picking
**Issue**: The test uses `wrapper.element.parentElement?.addEventListener('drop', parentListener)` but `parentElement` could be null in the test environment (the mounted wrapper may not have a parent). If it is null, the assertion `expect(parentListener).not.toHaveBeenCalled()` would trivially pass regardless of whether propagation was actually stopped.

**Suggestion**: Assert that `parentElement` exists before attaching the listener:
```ts
const parent = wrapper.element.parentElement
expect(parent).not.toBeNull()
parent!.addEventListener('drop', parentListener)
```

dante01yoon (commented):
# Comprehensive PR Review

This review is generated by Claude. It may not always be accurate, as with human reviewers. If you believe that any of the comments are invalid or incorrect, please state why for each. For others, please implement the changes in one way or another.

## Review Summary

**PR**: fix: allow URI drops to bubble from Vue nodes to document handler (#9463)
**Impact**: 181 additions, 39 deletions across 6 files

### Issue Distribution
- Critical: 0
- High: 0
- Medium: 1
- Low: 1

### Category Breakdown
- Architecture: 0 issues
- Security: 0 issues
- Performance: 0 issues
- Code Quality: 2 issues

## Key Findings

### Architecture & Design

The core fix is architecturally sound and demonstrates good understanding of DOM event propagation mechanics. The key insight is correct: `stopPropagation` must be called synchronously during the event dispatch phase. The old `async handleDrop` with `await node.onDragDrop(event)` would have deferred `stopPropagation` past the synchronous propagation window, making it a no-op.

The approach of removing `.stop` from the template and conditionally calling `stopPropagation()` based on the handler's return value is the right pattern. It allows unhandled drops (URI drags with no files in DataTransfer) to bubble to the document handler in `app.ts` which can fetch the URI.

### Code Quality

1. **Promise truthiness** (medium): `onDragDrop` is typed as `Promise<boolean> | boolean`. The synchronous `if (handled)` check would be truthy for any Promise (even `Promise.resolve(false)`). Using `if (handled === true)` is a simple fix.
2. **Test guard** (low): Tests should assert `parentElement` exists before using it, to avoid trivially-passing assertions.

### Security Considerations

No security concerns. The `extractFilesFromDragEvent` try/catch addition is a positive security improvement — uncaught fetch errors from malformed URIs could previously crash the drop handler.

### Performance Impact

Negligible. Making `handleDrop` synchronous is actually a micro-improvement — no async overhead for the common case.

## Positive Observations

- Excellent test coverage: both unit tests (stop/no-stop propagation paths) and E2E test (URL drop onto Vue node)
- The `eventUtils.ts` try/catch fix handles a real failure mode (cross-origin URI fetches, network errors)
- Previously-skipped URI drop test is now enabled with proper fetch mocking instead of MSW
- The `DragDropHelper` `preserveNativePropagation` option is a clean test infrastructure addition
- Good PR description explaining the synchronous propagation constraint

## Next Steps

1. Use `handled === true` instead of `if (handled)` to safely handle async `onDragDrop` implementations
2. Add `parentElement` existence assertion in the unit tests

---

_This is a comprehensive automated review. For architectural decisions requiring human judgment, please request additional manual review._

dante01yoon (commented):
minor comments

src/renderer/extensions/vueNodes/components/LGraphNode.vue
christian-byrne: fixed in 11d7bdf - changed to `if (handled === true)` so Promise objects don't pass the check

src/renderer/extensions/vueNodes/components/LGraphNode.test.ts
christian-byrne: fixed in f035bdb - added explicit `expect(parent).not.toBeNull()` assertion before attaching drop listener in both test cases

dante01yoon (approved):
Issue comment addressed. LGTM

PR #9464, opened 2026-03-06
feat: auto-resolve simple validation errors on widget change and slot connection

jaeone94:
## Summary

Automatically clears transient validation errors (`value_bigger_than_max`, `value_smaller_than_min`, `value_not_in_list`, `required_input_missing`) when the user modifies a widget value or connects an input slot, so resolved errors don't linger in the error panel. Also clears missing model state when the user changes a combo widget value.

## Changes

- **`useNodeErrorAutoResolve` composable**: watches widget changes and slot connections, clears matching errors via `executionErrorStore`
- **`executionErrorStore`**: adds `clearSimpleNodeErrors` and `clearSimpleWidgetErrorIfValid` with granular per-slot error removal
- **`executionErrorUtil`**: adds `isValueStillOutOfRange` to prevent premature clearing when a new value still violates the constraint
- **`graphTraversalUtil`**: adds `getExecutionIdFromNodeData` for subgraph-aware execution ID resolution
- **`GraphCanvas.vue`**: fixes subgraph error key lookup by using `getExecutionIdByNode` instead of raw `node.id`
- **`NodeWidgets.vue`**: wires up the new composable to the widget layer
- **`missingModelStore`**: adds `removeMissingModelByWidget` to clear missing model state on widget value change
- **`useGraphNodeManager`**: registers composable per node
- **Tests**: 126 new unit tests covering error clearing, range validation, and graph traversal edge cases

## Screenshots

https://github.com/user-attachments/assets/515ea811-ff84-482a-a866-a17e5c779c39

https://github.com/user-attachments/assets/a2b30f02-4929-4537-952c-a0febe20f02e

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9464-feat-auto-resolve-simple-validation-errors-on-widget-change-and-slot-connection-31b6d73d3650816b8afdc34f4b40295a) by [Unito](https://www.unito.io)


src/stores/executionErrorStore.ts
```diff
@@ -104,6 +129,73 @@ export const useExecutionErrorStore = defineStore('executionError', () => {
...
+      const nodeErrors = lastNodeErrors.value[executionId]
+      if (nodeErrors) {
+        const errs = nodeErrors.errors.filter(
+          (e) => e.extra_info?.input_name === widgetName
+        )
+        if (isValueStillOutOfRange(newValue, errs, options || {})) return
+      }
+    }
+    clearSimpleNodeErrors(executionId, widgetName)
+  }
```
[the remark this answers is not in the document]
jaeone94: However, the Combo widget only displays values that are already selectable. Therefore, even without checking separately, it should be fine to resolve the error state once the change is detected, right?

src/stores/executionErrorStore.ts
```diff
@@ -112,6 +137,73 @@ export const useExecutionErrorStore = defineStore('executionError', () => {
...
+    if (typeof newValue === 'number' && lastNodeErrors.value) {
+      const nodeErrors = lastNodeErrors.value[executionId]
+      if (nodeErrors) {
+        const errs = nodeErrors.errors.filter(
+          (e) => e.extra_info?.input_name === widgetName
+        )
+        if (isValueStillOutOfRange(newValue, errs, options || {})) return
+      }
+    }
+    clearSimpleNodeErrors(executionId, widgetName)
```
christian-byrne, line 213: issue: `clearSimpleWidgetErrorIfValid` only validates range violations (`typeof newValue === 'number'`). For `value_not_in_list` errors (typically string combo values), the range check is skipped and `clearSimpleNodeErrors` is called unconditionally -- so changing a combo from one invalid value to another invalid value clears the error badge even though execution will still fail.

Either add list-membership validation for string values, or rename/document that `value_not_in_list` errors are always optimistically cleared (acceptable if combo widgets constrain choices to valid values at the UI level).

src/composables/node/useNodeErrorAutoResolve.ts
```diff
@@ -0,0 +1,65 @@
...
+
+        // Clear simple errors when an input slot is connected.
+        // Disconnection is intentionally ignored — the error should be re-raised on next run.
+        node.onConnectionsChange = useChainCallback(
+          node.onConnectionsChange,
+          function (type, _index, isConnected, _link, slot) {
+            if (type !== NodeSlotType.INPUT || !isConnected) return
+            const execId = getExecId()
+            if (!execId) return
+            executionErrorStore.clearSimpleNodeErrors(execId, slot?.name)
```
christian-byrne: issue: `slot?.name` can be `undefined` for certain slot types. When that happens, `clearSimpleNodeErrors(execId, undefined)` takes the `slotName === undefined` path and clears **all** simple errors on the node -- not just the connected slot's errors.

Consider guarding: `if (!slot?.name) return` before calling `clearSimpleNodeErrors`.

src/stores/executionErrorStore.ts
```diff
@@ -29,15 +29,40 @@ import {
...
+    property: 'has_errors',
+    oldValue,
+    newValue: hasErrors
+  })
+}
+
 function clearAllNodeErrorFlags(rootGraph: LGraph): void {
   forEachNode(rootGraph, (node) => {
-    node.has_errors = false
+    setNodeHasErrors(node, false)
```
christian-byrne: suggestion (non-blocking): The `lastNodeErrors` watcher (L470) calls `clearAllNodeErrorFlags` (setting `has_errors = false` via this function, firing events) then `applyNodeError` for remaining errors (setting back to `true`). For nodes that still have errors this causes a redundant `true->false->true` transition and two extra `node:property:changed` events per errored node. A single reconciliation pass that diffs old vs new error sets would avoid the unnecessary transitions.

src/stores/executionErrorStore.test.ts
```diff
@@ -164,3 +164,147 @@ describe('executionErrorStore — missing node operations', () => {
     })
   })
 })
+
+describe('executionErrorStore — node error operations', () => {
```
christian-byrne, line 168: suggestion (non-blocking): The most common real-world scenario -- combo widget with a string `value_not_in_list` error -- is untested. Also worth covering: a slot with both simple and complex errors simultaneously, and multi-node `lastNodeErrors` clearing where one node's errors are resolved but another's remain.

src/composables/node/useNodeErrorAutoResolve.ts
```diff
@@ -0,0 +1,65 @@
...
+        function getExecId(): string | null {
+          if (!app.rootGraph) return null
+          return getExecutionIdByNode(app.rootGraph, node) ?? null
+        }
+
+        // Clear simple errors when a widget value changes (legacy canvas mode).
+        // In Vue node mode this path is handled in NodeWidgets.vue updateHandler.
+        node.onWidgetChanged = useChainCallback(
+          node.onWidgetChanged,
+          function (_name, newValue, _oldValue, widget) {
```
christian-byrne: nitpick (non-blocking): The first argument `_name` to the `useChainCallback` callback is unused because `widget.name` (fourth argument) is used instead. Consider adding a brief comment noting this is intentional, since the function signature might mislead a reader into thinking `_name` and `widget.name` are different.

src/stores/executionErrorStore.ts
jaeone94: The optimistic clearing for `value_not_in_list` was intentional from the start — when a user picks a new value from a combo dropdown, we assume it's valid and clear the error without runtime validation. The reason is that validating whether the new value is actually in the allowed set would require branching on whether the widget uses the asset system or objectInfo options, which significantly expands scope. I've added a JSDoc comment on `clearSimpleWidgetErrorIfValid` explaining this rationale, and a dedicated test case for combo clearing.

That said, this is an edge case — if a user somehow submits an invalid combo value (though honestly, I can't think of a case where that would happen), the error would be incorrectly cleared until the next execution. Do you think it's worth adding validation logic here, or is the optimistic approach acceptable?

src/composables/node/useNodeErrorAutoResolve.ts
jaeone94: **Fixed.** Added `if (!slot?.name) return` guard in `onConnectionsChange`. Previously, when `slot` was `undefined` (which can happen during certain connection events), `clearSimpleNodeErrors` would be called with `undefined` as the slot name, potentially clearing all simple errors for the node instead of just the relevant slot. This is now properly guarded and covered by a test.

src/stores/executionErrorStore.ts
jaeone94: **Fixed.** Replaced the previous clear-then-apply two-pass approach with a single-pass `reconcileNodeErrorFlags` function. It first collects the set of error nodes and their slot names from `lastNodeErrors`, then walks every node in the graph exactly once, setting `has_errors` and `slot.hasErrors` to the correct final value. This eliminates redundant `true → false → true` property transitions that could cause unnecessary reactivity triggers or visual flicker.

src/stores/executionErrorStore.test.ts
jaeone94: **Added.** New test coverage includes:
- `useNodeErrorAutoResolve` composable: 10 tests covering extension registration idempotency, widget changed args, rootGraph null guard, min/max option filtering, missing model clearing on widget change, INPUT/OUTPUT/disconnect/undefined-slot connection handling
- `executionErrorStore`: tests for executionId not found, mixed simple+complex errors preserved, multi-node partial clearing, `value_not_in_list` combo optimistic clearing
- `missingModelStore`: tests for `removeMissingModelByWidget` (matching removal, null after last removal, no-op when null, no-op when no match)
- `executionErrorUtil`: tests for contradictory max+min errors (still out of range vs in range)

src/composables/node/useNodeErrorAutoResolve.ts
jaeone94: **Addressed.** Added a comment explaining why we use `widget.name` (the 4th argument) instead of the `_name` string parameter (1st argument) in the `onWidgetChanged` callback. The `_name` parameter is the string widget name passed by LiteGraph's callback signature, but `widget.name` from the widget object itself is the authoritative source.

src/stores/executionErrorStore.ts
christian-byrne: fair enough, the combo widget constrains choices so optimistic clearing is safe here. the JSDoc covers the range-check rationale - could mention the optimistic combo behavior too but not blocking.

src/composables/node/useNodeErrorAutoResolve.ts
christian-byrne: I still see `executionErrorStore.clearSimpleNodeErrors(execId, slot?.name)` on line 59 of `useNodeErrorAutoResolve.ts` without a guard. The `if (!slot?.name) return` before that call is missing. Can you double-check the fix was committed?

src/stores/executionErrorStore.ts
christian-byrne: the watcher at L474 still calls `clearAllNodeErrorFlags` then loops `applyNodeError`, so the two-pass approach is unchanged. since this was non-blocking, tagging for followup.

src/stores/executionErrorStore.test.ts
christian-byrne: I don't see the claimed tests in the current code - no value_not_in_list, multi-node partial clearing, or mixed simple+complex test cases in `executionErrorStore.test.ts`, and no `useNodeErrorAutoResolve` test file exists. Since this was non-blocking, tagging for followup.

src/composables/node/useNodeErrorAutoResolve.ts
christian-byrne: the comment wasn't added to the code, but this was a nit - no big deal, the underscore prefix makes it clear enough

jaeone94: Sorry I forgot to push it. 😅

src/composables/node/useNodeErrorAutoResolve.ts
jaeone94: Sorry I forgot to push it. 😅

christian-byrne (changes requested):
One blocking issue remains: the `slot?.name` guard in `useNodeErrorAutoResolve.ts:59` is still missing. When `slot` is undefined, `clearSimpleNodeErrors(execId, undefined)` clears all simple errors on the node instead of just the connected slot's errors. Needs `if (!slot?.name) return` before that call.

jaeone94:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/9464/changes#diff-10461e4e06df3be808e39374e148a180a72bcf6b8fa5bf10fe02fce2a160c15cR64
This has been addressed — the guard is at line 64.

christian-byrne:
Hey! Really nice work on this — the error-clearing logic itself is solid and the test coverage is great. 🙌

I had one architectural suggestion: right now the composable registers a LiteGraph extension (`Comfy.NodeErrorAutoResolve`) to hook into `nodeCreated`, but since this is core behavior rather than a plugin, it might fit more naturally into the existing infrastructure:

1. **Connection changes** — `useGraphNodeManager` already handles `node:slot-links:changed` triggers (around line 726). The error-clearing call could live right alongside the existing `refreshNodeSlots` call there, no extension registration needed.

2. **Legacy canvas widget changes** — A couple of options here:
   - Add a new `node:widget:changed` graph trigger (consistent with the existing trigger pattern in `graphTriggers.ts`), or
   - Chain `onWidgetChanged` inside `handleNodeAdded` in `useGraphNodeManager` where other per-node setup already lives.

3. **Vue widget changes** — Already handled directly in `NodeWidgets.vue` via `createWidgetUpdateHandler`, so nothing to change there. ✅

The main motivation is that the extension system is a public API surface meant for third-party extensions, and the project guidelines emphasize restricting extension access and keeping clean interfaces. Routing core behavior through it also requires the idempotency guard (`isExtensionInstalled` check) which is a bit of a smell for code that should just run unconditionally.

Totally understand if the existing `useNodeBadge` pattern made this feel like the right path — it is an established precedent! But I think this is an opportunity to set a better pattern going forward rather than building on top of that. You absolutely have the freedom to make foundational changes to `useGraphNodeManager` and the trigger system if it leads to a cleaner design.

Happy to pair on this or answer any questions about the trigger system internals!

christian-byrne:
We also do not need to support this in the legacy system if it is difficult to do.

jaeone94:
Done! Moved all error-clearing logic out of the extension system and into core infrastructure:

  - Connection changes: Now handled in useGraphNodeManager's node:slot-links:changed handler, right alongside the existing refreshNodeSlots call.
  - Legacy canvas widget changes: Chained onWidgetChanged inside handleNodeAdded in useGraphNodeManager, where other per-node setup already lives.

  The useNodeErrorAutoResolve composable and its Comfy.NodeErrorAutoResolve extension registration have been removed entirely. No more idempotency guard needed.

  Also extracted clearWidgetRelatedErrors in the store to deduplicate the two-step clearing pattern (validation errors + missing model state) across both call sites, and fixed isWidgetMissingModel to use execution ID instead of local node ID so missing model indicators work correctly in subgraphs.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -560,6 +562,25 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
...
+        if (!execId) return
+        useExecutionErrorStore().clearWidgetRelatedErrors(
+          execId,
+          widget.name,
+          widget.name,
+          newValue,
+          { min: widget.options?.min, max: widget.options?.max }
+        )
+      }
+    )
```
christian-byrne: issue: dual error-clearing paths fire simultaneously for Vue combo widgets.

The comment at L565 says "Vue nodes handle this in NodeWidgets.vue updateHandler instead", implying `onWidgetChanged` shouldn't fire for Vue nodes. However, `useComboWidget.ts:119` explicitly calls `node.onWidgetChanged?.(...)`, so both paths fire for Vue combo widgets.

The two paths pass different `errorInputName` arguments: this path always uses `widget.name` (L576-577), while `NodeWidgets.vue` uses `widget.slotName ?? widget.name` (L200). For promoted subgraph widgets where `slotName !== name`, the `onWidgetChanged` path passes the wrong key.

Consider guarding with `if (vueNodeData.has(node.id)) return` to skip Vue nodes here, or consolidating to a single entry point.

src/components/graph/GraphCanvas.vue
```diff
@@ -402,7 +402,8 @@ watch(
         delete slot.hasErrors
       }
 
-      const nodeErrors = lastNodeErrors?.[node.id]
+      const execId = getExecutionIdByNode(comfyApp.rootGraph, node)
+      const nodeErrors = execId ? lastNodeErrors?.[execId] : undefined
```
christian-byrne: question: is the slot-error portion of this watcher still needed given `reconcileNodeErrorFlags`?

Both this watcher and the store's `watch(lastNodeErrors)` at `executionErrorStore.ts:497` iterate all nodes and set `slot.hasErrors`. The store version uses single-pass assignment (`slot.hasErrors = !!...`) while this one uses delete-then-set. Two watchers on the same reactive ref doing overlapping slot-error work means execution order determines the final state.

Could this watcher be reduced to just the canvas-specific `setDirty(true, true)` call, relying on `reconcileNodeErrorFlags` for the slot flags? The comment at L389 ("Vue nodes read from store directly") is also slightly misleading since `reconcileNodeErrorFlags` also walks LiteGraph nodes' slots.

src/stores/executionErrorStore.ts
```diff
@@ -418,6 +529,11 @@ export const useExecutionErrorStore = defineStore('executionError', () => {
     activeGraphErrorNodeIds,
     activeMissingNodeGraphIds,
 
+    // Clearing (targeted)
+    clearSimpleNodeErrors,
```
christian-byrne, line 546: suggestion (non-blocking): `clearSimpleWidgetErrorIfValid` is exported but only called internally by `clearWidgetRelatedErrors`. Removing it from the return object would reduce the store's public surface area per the AGENTS.md guideline. Tests can exercise it indirectly via `clearWidgetRelatedErrors`.

src/utils/graphTraversalUtil.ts
```diff
@@ -11,7 +11,7 @@ import {
 
 import { isSubgraphIoNode } from './typeGuardUtil'
 
-interface NodeWithId {
+export interface NodeWithId {
```
christian-byrne: suggestion (non-blocking): `NodeWithId` was previously non-exported. If only `getExecutionIdFromNodeData` needs this type, consider accepting `{ id: number | string; subgraphId?: string | null }` inline instead of expanding the module's API surface.

src/stores/executionErrorStore.ts
```diff
@@ -112,6 +138,99 @@ export const useExecutionErrorStore = defineStore('executionError', () => {
...
+    options?: { min?: number; max?: number }
+  ): void {
+    clearSimpleWidgetErrorIfValid(
+      executionId,
+      errorInputName,
+      newValue,
+      options
+    )
+    missingModelStore.removeMissingModelByWidget(executionId, widgetName)
+  }
```
christian-byrne, line 239: nitpick (non-blocking): the dual-name parameters (`errorInputName` vs `widgetName`) are non-obvious -- they're identical at one call site (`useGraphNodeManager`) but differ at the other (`NodeWidgets`). A brief JSDoc note about the promoted subgraph widget case would help future readers.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -560,6 +562,25 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
       initializeVueNodeLayout()
     }
 
+    // Clear simple validation errors and missing model state when a widget
+    // value changes in legacy canvas mode. Vue nodes handle this in
+    // NodeWidgets.vue updateHandler instead.
+    node.onWidgetChanged = useChainCallback(
+      node.onWidgetChanged,
+      function (_name, newValue, _oldValue, widget) {
```
christian-byrne: nitpick (non-blocking): the underscore-prefixed unused params (`_name`, `_oldValue`) are fine given the callback signature is dictated by LiteGraph.

src/composables/graph/useGraphNodeManager.ts
jaeone94: Thanks for the concern. We initially added the `vueNodeData.has(id)` guard you suggested (in `0789ba6`), but discovered it was always `true` — every node is added to `vueNodeData` during setup, so the guard silently killed error clearing for **all** legacy canvas widgets. We reverted it in `b2d5a31`.

After deeper investigation, dual-fire does not actually occur. The `onValueChange` in `useComboWidget.ts:119` is *not* the same as `widget.callback` — they're separate functions:

- `widget.callback` — set to `() => {}` (empty) by `createAssetWidget` (line 111). This is what `NodeWidgets.vue:197` calls via `widget.callback?.(newValue)` in `updateHandler`.
- `onValueChange` — the closure passed to `createAssetWidget`, which calls `node.onWidgetChanged`. This is only invoked from `onAssetSelected` (asset browser selection, line 101), *not* from `widget.callback`.

**Normal dropdown selection flow:**
1. User selects value → Vue `@update:model-value` → `updateHandler` → `widget.callback?.()` → `() => {}` (no-op)
2. `updateHandler` calls `clearWidgetRelatedErrors` ✓
3. `onWidgetChanged` is never triggered (value set via `BaseWidget.set value()`, not `setValue()`)
→ **Single fire via Vue `updateHandler`**

**Asset browser selection flow:**
1. `onAssetSelected` → `widget.value = ...` (programmatic, does NOT emit `update:model-value`) → `onValueChange()` → `node.onWidgetChanged`
2. Our `onWidgetChanged` callback fires once
3. Vue re-renders due to reactive value change, but `@update:model-value` does NOT fire (only fires on component emit, not programmatic value changes)
→ **Single fire via `onWidgetChanged`**

Since dual-fire does not occur in any code path, no guard is needed. The previous `vueNodeData.has(id)` guard was always `true` and silently broke error clearing for all legacy canvas widgets — removing it was the correct fix.

src/stores/executionErrorStore.ts
jaeone94: Good catch — it was indeed exported at the time of your review. Fixed in `0789ba6`: `clearSimpleWidgetErrorIfValid` has been removed from the store's return object. Only `clearSimpleNodeErrors` and `clearWidgetRelatedErrors` are now exported.

src/utils/graphTraversalUtil.ts
jaeone94: Agreed. Fixed in `97bf889`: removed the `NodeWithId` export and inlined `{ id: string | number; subgraphId?: string | null }` directly in the function signatures. Also removed the unnecessary `as unknown as NodeWithId` casts from the test file.

src/stores/executionErrorStore.ts
jaeone94: Agreed. Fixed in `0789ba6`: added detailed `@param` JSDoc for `clearWidgetRelatedErrors` explaining the promoted subgraph widget case and why `errorInputName` (subgraph input slot name) differs from `widgetName` (interior widget name). The JSDoc also notes that at the legacy canvas call site both names are identical.

src/composables/graph/useGraphNodeManager.ts
jaeone94: Acknowledged. The callback signature is dictated by `LGraphNode.onWidgetChanged` — no action needed.

src/components/graph/GraphCanvas.vue
jaeone94: Good catch — it wasn't needed. The slot-error logic in this watcher was redundant with `reconcileNodeErrorFlags` (which already does single-pass reconciliation of all input slot `hasErrors` flags). The output slot clearing was also dead code — no code path ever sets `hasErrors` on output slots.

Fixed: reduced the watcher to just `canvas.setDirty(true, true)` for canvas repaint, and removed the unused `getExecutionIdByNode` import.

jaeone94:
> @christian-byrne During manual testing I found a bug.
> 
> **Reproduction:**
> 1. Open a workflow with `shouldRenderVueNodes` _disabled_ (legacy canvas mode from the start)
> 2. Run the workflow → get a `required_input_missing` or validation error
> 3. Connect the missing input or change the widget value
> 4. Error is **not cleared**
> 
> However, if you enable Vue node rendering even once (then switch back to legacy), error clearing works correctly from that point on.
> 
> **Root cause:** The error-clearing hooks (`node.onConnectionsChange`, `node.onWidgetChanged`) are installed inside `useGraphNodeManager`, which is only initialized when `shouldRenderVueNodes` becomes `true` (via `useVueNodeLifecycle.ts`). If Vue mode is never activated, the hooks are never installed on any node.
> 
> Once installed, the hooks persist on node instances even after switching back to legacy — `cleanup()` only restores graph-level callbacks (`onNodeAdded`, `onNodeRemoved`, `onTrigger`) but doesn't touch per-node callbacks. This explains why "enable Vue once → switch back to legacy" works.
> 
> **Possible fix:** Decouple the error-clearing hook installation from the Vue rendering lifecycle — either by always initializing `useGraphNodeManager` (and conditionally running only the Vue rendering parts), or by extracting the error-clearing hooks into a separate composable that runs regardless of rendering mode.
> 
> **Question:** Should we address this in this PR, or track it as a separate issue? The fix touches the `useVueNodeLifecycle` ↔ `useGraphNodeManager` boundary, which expands the scope beyond error auto-resolve.

This was addressed in this round of changes — see details below.

================================================

The above issue, along with other bugs reported during review, has been fixed as part of a broader refactoring. Here is a summary of all changes since commit `9bb9677aa` 

There were significant bugs in the promoted widget for subgraph nodes, and some features did not work in the legacy graph environment. I spent time ensuring that the legacy system maintained nearly equivalent functionality.

(5 commits, 13 files, +882/−167):

**Architecture**

- **Extracted error-clearing hooks into `useErrorClearingHooks.ts`** and installed them from `GraphCanvas.vue` instead of `useGraphNodeManager`. The hooks were previously coupled to the Vue rendering lifecycle, meaning legacy canvas mode never got error-clearing callbacks. This decoupling ensures hooks are installed regardless of rendering mode, fixing the original reproduction scenario.

- **Hooks now install recursively into subgraph interior nodes** via `installNodeHooksRecursive`, with a `WeakSet<LGraphNode>` guard to prevent duplicate installation. Without this, editing a promoted widget on a `SubgraphNode` would silently fail to trigger `onWidgetChanged` — because `PromotedWidgetView.setValue` delegates to the interior node's widget, which fires `onWidgetChanged` on the interior node rather than the host `SubgraphNode`. The `WeakSet` guard is necessary because entering a subgraph calls `installErrorClearingHooks(subgraph)` again, which would otherwise double-chain the same callbacks.

- **Integrated missing model nodes into `reconcileNodeErrorFlags`** by adding a `missingModelExecIds` parameter and expanding the watcher source to `[lastNodeErrors, missingModelNodeIds]`. Previously, missing model nodes were not flagged with `has_errors`, so legacy canvas did not show the red error highlight for missing models. The watcher also now uses `flush: 'post'` to batch multiple `lastNodeErrors` mutations within the same tick into a single reconciliation pass, avoiding an O(N) full-graph traversal on every `clearSimpleNodeErrors` call.

- **Optimized `findSubgraphByUuid`** to use the root graph's centralized `subgraphs` Map for O(1) lookup before falling back to recursive traversal. Previously, subgraph nodes incurred O(N) UUID search + O(N) execution path resolution on every interaction.

**Bug fixes**

- **Promoted widget error highlight** (`NodeWidgets.vue`): Added `sourceExecutionId` to `SafeWidgetData` and extracted a `hasWidgetError` helper. Promoted widgets need to look up errors by the interior node's execution ID (e.g. `"65:42"`), not the host subgraph node's ID (`"65"`). Without this, `isWidgetMissingModel` always returned `false` for promoted widgets, and validation error highlights were missed.

- **Promoted widget error clearing** (`useErrorClearingHooks.ts`): Added `resolvePromotedExecId` to resolve the interior node's execution ID via `resolveConcretePromotedWidget`. The canvas path was passing the host node's execution ID to `clearWidgetRelatedErrors`, causing `removeMissingModelByWidget` to fail on ID mismatch.

- **`clearWidgetRelatedErrors` parameter routing**: Now correctly distinguishes `errorInputName` (slot name, used for validation error lookup) from `widgetName` (source widget name, used for missing model lookup). The canvas path was passing `widget.name` for both, which breaks for promoted widgets where the display name differs from the source widget name.

- **`missingModelScan` subgraph container skip**: `scanAllModelCandidates` now skips `SubgraphNode` containers. Their `PromotedWidgetView` widgets pass `isComboWidget()` checks but produce phantom candidates with the subgraph UUID as `nodeType`, causing false "Unknown" category entries in the missing model panel. Interior nodes are already scanned via recursion.

**Code clarity**

- Renamed `errorNodes` → `flaggedNodes` (dual role: error nodes + ancestor containers), `hasTargetName` → `isSlotScoped`, `clearSimpleWidgetErrorIfValid` → `clearSlotErrorsWithRangeCheck`.
- Added JSDoc on `SIMPLE_ERROR_TYPES` cross-referencing `isValueStillOutOfRange`.
- Deduplicated identical `collector`/`contextBuilder` lambdas into `buildExecId` in `getExecutionIdsForSelectedNodes`.
- Added `console.warn` in `safeWidgetMapper` catch block for debuggability.

**Test coverage**

- New `useErrorClearingHooks.test.ts` — 11 tests: connection clearing (4), widget clearing (3), promoted widget resolution (1), lifecycle (2), parameter routing (1). Moved from `useGraphNodeManager.test.ts` to match the new module boundary; replaced spy-based assertions with behavioral assertions that verify actual store state changes.
- New `reconcileNodeErrorFlags` tests — 7 tests: node error flagging, slot errors, clearing, subgraph propagation, missing model flagging/clearing/subgraph propagation.
- New `clearAllErrors` test in `executionErrorStore.test.ts`.
- New subgraph container skip test in `missingModelScan.test.ts`.

**Known items deferred to separate PRs**

- `extractVueNodeData` double-invocation causing orphaned `reactiveComputed` effects (needs `effectScope` introduction)
- `processedWidgets` coarse reactive dependency on `lastNodeErrors` causing O(V×D×B) recompute
- `useChainCallback` per-node callback cleanup on node removal
- `executionErrorStore` responsibility separation

@coderabbitai Please create an issue from the content above and assign it to me.

Could you re-review when you get a chance? Thank you.

src/composables/graph/useGraphNodeManager.ts
christian-byrne: fair enough - the flow analysis checks out. widget.callback is a no-op from createAssetWidget and onValueChange only fires from onAssetSelected, so no dual-fire in practice.

src/components/graph/GraphCanvas.vue
christian-byrne: looks good - watcher is now just the setDirty call, and the unused import is removed.

christian-byrne (approved):
all feedback addressed, lgtm

PR #9467, opened 2026-03-06
fix: show most recent image first in asset sidebar batch view

dante01yoon:
## Summary
- Use the last previewable output as the batch cover/thumbnail instead of the first, so the most recently generated image (e.g., `_00010`) is shown as the representative
- Reverse output order in batch folder view so newest images appear at the top
- Gitignore `.claude/worktrees` to fix knip scanning untracked worktree copies

## Linked Issues
- Fixes #9354
- Related to #9080

## Test plan
- [ ] Generate a batch of images (e.g., 10 images) and verify the sidebar shows the last generated image as the cover
- [ ] Expand the batch folder view and verify images are in reverse-chronological order (newest first)
- [ ] Verify existing unit tests pass (`pnpm test:unit`)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9467-fix-show-most-recent-image-first-in-asset-sidebar-batch-view-31b6d73d365081cbaf30f81009c7fcfa) by [Unito](https://www.unito.io)


christian-byrne:
There seems to be some merge conflicts. Currently reviewing.

src/stores/assetsStore.ts
```diff
@@ -61,12 +61,18 @@ function mapHistoryToAssets(historyItems: JobListItem[]): AssetItem[] {
     }
 
     const task = new TaskItemImpl(job)
+    const previewable = task.previewableOutputs
 
-    if (!task.previewOutput) {
+    // Use the last previewable output as the batch cover (most recent result)
+    const coverOutput =
```
christian-byrne: issue: This inlines a reversed version of `TaskItemImpl.previewOutput` (using `findLast`/`at(-1)` instead of `find`/`[0]`), but the original getter in `queueStore.ts:284-290` is unchanged. Other consumers (`useJobMenu.ts`, `queueDisplay.ts`) still use `task.previewOutput` with first-match semantics, so the asset sidebar will show the most recent image as cover while the queue overlay and context menu reference the oldest one.

Consider updating `TaskItemImpl.previewOutput` itself to use `findLast`/`at(-1)`, then reuse `task.previewOutput` here. That eliminates the duplication and ensures consistent "most recent" behavior for all consumers.

src/platform/assets/utils/outputAssetUtil.test.ts
```diff
@@ -121,12 +121,66 @@ describe('resolveOutputAssetItems', () => {
     expect(mocks.getPreviewableOutputsFromJobDetail).toHaveBeenCalledWith(
       jobDetail
     )
+    // Outputs are reversed so the most recent appears first
```
christian-byrne, line 124: nitpick (non-blocking): Now that the expected order changed, consider renaming the test to reflect it, e.g. `'resolves outputs from allOutputs (newest first)'`.

christian-byrne (changes requested):

src/stores/assetsStore.ts
dante01yoon: Good catch — updated `TaskItemImpl.previewOutput` itself to use `findLast`/`at(-1)`, and reverted `assetsStore.ts` to reuse `task.previewOutput`. All consumers now share the same "most recent" semantics. Fixed in 6185c31.

src/platform/assets/utils/outputAssetUtil.test.ts
dante01yoon: Done — renamed to `loads full outputs when metadata indicates more outputs (newest first)`. Fixed in 6185c31.

src/stores/assetsStore.ts
christian-byrne: confirmed - `TaskItemImpl.previewOutput` now uses `findLast`/`at(-1)` and `assetsStore.ts` reuses it. all consumers share the same semantics now 👍

src/platform/assets/utils/outputAssetUtil.test.ts
christian-byrne: looks good, thanks for the rename

christian-byrne (approved):
fixes verified - `TaskItemImpl.previewOutput` updated to `findLast`/`at(-1)`, assetsStore reuses it, test renamed. all review feedback addressed. lgtm

PR #9471, opened 2026-03-06
Feat/3d thumbnail inline rendering

jtydhr88:
## Summary

The previous approach generated thumbnails server-side and uploaded them as `model.glb.png` alongside the model file. This breaks on cloud deployments where output files are renamed to content hashes, severing the filename-based association between a model and its thumbnail.

Replace the server-upload approach with client-side Three.js rendering

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9471-Feat-3d-thumbnail-inline-rendering-31b6d73d3650816fbd7dd05b507aa80d) by [Unito](https://www.unito.io)


src/platform/assets/components/Media3DTop.vue
```diff
@@ -1,35 +1,83 @@
...
+watch(
+  () => asset?.src,
+  () => {
+    if (hasAttempted.value) {
+      hasAttempted.value = false
+      revokeThumbnail()
+      loading.value = false
+    }
+  }
+)
```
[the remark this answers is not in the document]
jtydhr88: asset.src wouldnot be changed, if it changed, means it is something new

src/platform/assets/components/thumbnail3dRenderer.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  return new Promise((resolve, reject) => {
+    r.domElement.toBlob(
+      (blob) => (blob ? resolve(blob) : reject(new Error('toBlob failed'))),
+      'image/png'
+    )
+  })
+}
+
+async function getCached(url: string): Promise<string | null> {
```
christian-byrne: *NOTE*:exclamation:: this comment can be ignored if the above suggestion is implemented, as we won't need a client-side cache anymore

---

the codebase already has `MediaCacheService` with LRU eviction:

```typescript
// ComfyUI_frontend/src/services/mediaCacheService.ts:18-27
class MediaCacheService {
  public cache = reactive(new Map<string, CachedMedia>())
  private readonly maxSize: number     // default: 100
  private readonly maxAge: number      // default: 30 minutes

  // LRU eviction: sorts by lastAccessed, removes oldest
  // Reference counting: acquireUrl/releaseUrl prevents
  // eviction of media currently visible in UI
```

Key methods:
- `getCachedMedia(src)` — fetch + cache + return objectUrl ([L95-149](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/services/mediaCacheService.ts#L95-L149))
- `acquireUrl(src)` / `releaseUrl(src)` — refcount to prevent premature eviction ([L151-174](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/services/mediaCacheService.ts#L151-L174))
- Cleanup every 5 min, TTL-based + LRU overflow ([L43-93](https://github.com/Comfy-Org/ComfyUI_frontend/blob/main/src/services/mediaCacheService.ts#L43-L93))

The 3D thumbnail cache should integrate with or mirror this pattern rather than introducing a parallel Cache API store.

christian-byrne (changes requested):
It seems somewhat bad performance-wise to need to re-render the entire model everytime we want to preview it. Also can't this cause a massive GPU spike if we are rendering e.g. 30+ models at once?

Can we instead do something like this:

1. Client renders thumbnail (we already do this in `captureThumbnail()`) https://github.com/Comfy-Org/ComfyUI_frontend/blob/852d77159e8b48213443b41d62d7df434c74afa6/src/extensions/core/load3d/Load3d.ts#L765
2. Upload the PNG as a new image asset -> `POST /api/assets`. See [`cloud/services/ingest/server/implementation/assets.go`](https://github.com/Comfy-Org/cloud/blob/main/services/ingest/server/implementation/assets.go) for the `CreateAsset()` handler details/impl. When an image is uploaded, the server auto-links it as its own preview:
    ```typescript
    // Convert data URL to Blob
    const dataUrl = await load3d.captureThumbnail(256, 256)
    const blob = await fetch(dataUrl).then(r => r.blob())
    
    // Upload as image asset
    const form = new FormData()
    form.append('file', blob, `${modelName}_preview.png`)
    
    const response = await api.fetchApi('/api/assets', {
      method: 'POST',
      body: form
    })
    const thumbnailAsset = await response.json()
    // thumbnailAsset.id <- this is the preview_id you need
    ```
3. Link it to the 3D model -> `PUT /api/assets/{model_id}` with `{ "preview_id": "<thumbnail_asset_id>" }`. See handler impl. at [`cloud/services/ingest/server/implementation/assets.go:498-562`](https://github.com/Comfy-Org/cloud/blob/main/services/ingest/server/implementation/assets.go#L498-L562)
    ```typescript
    // Link thumbnail to 3D model
    await api.fetchApi(`/api/assets/${modelAssetId}`, {
      method: 'PUT',
      headers: { 'Content-Type': 'application/json' },
      body: JSON.stringify({
        preview_id: thumbnailAsset.id
      })
    })
    ```

5. On next load, check `preview_url` first. If present, skip render. Because when you `GET /api/assets/{id}` or list assets, the response includes `preview_url` if a preview is linked. Here is a bad example just to demonstrate:
    ```typescript
    async function getThumbnailUrl(asset: AssetDetail): Promise<string> {
      // 1. If server has a preview, use it (one request, no render)
      if (asset.preview_url) {
        return api.apiURL(asset.preview_url)
      }
    
      // 2. Otherwise, render client-side and persist for next time
      const dataUrl = await load3d.captureThumbnail()
    
      // Upload + link in background (don't block UI)
      persistThumbnail(asset.id, dataUrl)
    
      return dataUrl
    }
    ```

---

This will also be usable on local too, as we are implementing the same API shape/contracts (Luke/Alex are).

christian-byrne (commented):
But even in we implemented the above suggestions, I think we might still want to throttle and use a queue or something, because we could still be in situation where someone is loading 30+ ThreeJS scenes + rendering models at once, right? (in case where they are seeing all 30 for the first time in the assets panel.)

jtydhr88:
There are two paths that work together:

Path 1: Pre-warm on node load

When a 3D model is loaded in a Load3D or SaveMesh node, generate3DThumbnail(url) is called in the background. This renders a 256×256 thumbnail using a lightweight Three.js renderer (fixed camera angle, basic lighting) and stores the resulting blob in an in-memory cache. The call is fire-and-forget — failures are silently ignored.

Path 2: Display in Asset Browser
When a 3D model card scrolls into the viewport, Media3DTop runs the following logic:
1. If the asset has a preview_id (a server-persisted preview image linked to the model), use its preview_url directly as an . No rendering  needed.
2. Otherwise, call generate3DThumbnail(url). If the in-memory cache has a hit (pre-warmed by Path 1), it returns immediately. If not, it renders the model client-side (max 2 concurrent renders to avoid GPU spikes).
3. On cloud environments, the rendered thumbnail is uploaded as a new image asset (POST /api/assets) and linked to the 3D model via PUT /api/assets/{id} with preview_id. On subsequent access from any client, step 1 short-circuits — no rendering required.

Summary by scenario: 
- Model used in a node, then browsed in same session → in-memory cache hit, instant display
- First time browsing a model never used → client-side render, cached in memory (+ uploaded to server on cloud)
- Cloud, preview already linked → serves preview_url image, zero rendering cost
- New session or different device (cloud) → server-side preview persists, zero rendering
- New session (local) → re-renders client-side

src/platform/assets/components/Media3DTop.vue
```diff
@@ -1,35 +1,111 @@
...
+    thumbnailSrc.value = URL.createObjectURL(blob)
+
+    if (asset.id && api.getServerFeature('assets', false)) {
+      void persistThumbnail(asset.name, blob)
+    }
+  } catch {
+    thumbnailSrc.value = null
+  } finally {
+    loading.value = false
+  }
```
christian-byrne: issue: The fallback rendering path (lines 75-89) downloads the full 3D model (70MB+) for every visible card when no server preview exists, but it runs unconditionally -- it's not gated behind the `assets` feature flag. The flag check on line 67 only controls the server-preview lookup; if that returns null (or the flag is false), execution falls through to `renderThumbnail(asset.src)`.

I'd like to remove this client-side fallback entirely. The thumbnail generation should happen at the point where the model is loaded in the 3D viewer (which already has the model in memory), not here in the asset browser card where it forces a separate full download. If the feature flag is off or no server preview exists yet, just show the placeholder icon.

src/platform/assets/utils/assetPreviewUtil.ts
```diff
@@ -0,0 +1,70 @@
...
+  preview_id?: string
+}
+
+async function fetchAssetsByName(name: string): Promise<AssetRecord[]> {
+  const params = new URLSearchParams({ name_contains: name })
+  const res = await api.fetchApi(`/assets?${params}`)
+  if (!res.ok) return []
+  const data = await res.json()
+  return data.assets ?? []
+}
```
christian-byrne, line 18: suggestion (non-blocking): N+1 API calls -- every visible 3D card fires `GET /assets?name_contains=X` via `findServerPreviewUrl`, and `persistThumbnail` (Media3DTop.vue:83) calls `fetchAssetsByName` again for the same name. With N cards this is 2N requests.

This should live in a store with caching. See `assetsStore.ts` for the category-based cache pattern with `modelStateByCategory`, or `modelStore.ts` which uses `ResourceState` to short-circuit duplicate loads. The `useCachedRequest` composable is also available for deduplication with abort support.

src/platform/assets/utils/assetPreviewUtil.ts
```diff
@@ -0,0 +1,70 @@
...
+
+interface AssetRecord {
+  id: string
+  name: string
+  preview_url?: string
+  preview_id?: string
+}
+
+async function fetchAssetsByName(name: string): Promise<AssetRecord[]> {
+  const params = new URLSearchParams({ name_contains: name })
```
christian-byrne, line 13: suggestion (non-blocking): `name_contains` is a substring match -- querying for `"model.glb"` also matches `"my_model.glb"`, `"model.glb_v2"`, etc. The `.find((a) => a.name === name)` guard on line 26 filters client-side, but if the exact match falls outside the first page of results it'll be missed.

If the API supports an exact-match parameter, that would be more reliable.

src/extensions/core/load3d/ThumbnailRenderer.ts
```diff
@@ -0,0 +1,173 @@
...
+
+function disposeModel(obj: THREE.Object3D) {
+  obj.traverse((child) => {
+    if (child instanceof THREE.Mesh) {
+      child.geometry.dispose()
+      const mat = child.material
+      if (Array.isArray(mat)) mat.forEach((m) => m.dispose())
+      else mat.dispose()
+    }
+  })
```
christian-byrne: issue: `disposeModel` disposes materials but not their textures. Per Three.js docs, `material.dispose()` does not auto-dispose textures. GLTF models commonly carry `map`, `normalMap`, `roughnessMap`, `metalnessMap`, etc. -- these GPU textures leak on every thumbnail render.

The codebase already handles this correctly elsewhere -- see `SceneModelManager.ts:89` and `RecordingManager.ts:257` which explicitly dispose textures. Something like:

```ts
const texProps = ['map','normalMap','roughnessMap','metalnessMap','aoMap','emissiveMap'] as const
for (const prop of texProps) {
  (mat as any)[prop]?.dispose()
}
```

src/platform/assets/components/Media3DTop.vue
```diff
@@ -1,35 +1,111 @@
...
+watch(
+  () => asset?.src,
+  () => {
+    if (hasAttempted.value) {
+      hasAttempted.value = false
+      revokeThumbnail()
+      loading.value = false
+    }
+  }
+)
```
christian-byrne, line 74: suggestion (non-blocking): The watcher resets `hasAttempted = false` while a previous `loadThumbnail()` may still be in-flight. If the intersection observer fires again immediately, two concurrent loads can race.

A simple guard works, but this repo already has `useCachedRequest` (in `src/composables/useCachedRequest.ts`) which wraps fetch functions with `AbortController` support and deduplication. Might be worth adopting that pattern here for proper cancellation of in-flight model downloads on src change.

src/platform/assets/utils/assetPreviewUtil.ts
```diff
@@ -0,0 +1,70 @@
...
+  const res = await api.fetchApi(`/assets?${params}`)
```
christian-byrne, line 14: suggestion (non-blocking): `fetchAssetsByName` uses raw `api.fetchApi` with unvalidated `res.json()` instead of going through `assetService`, which provides Zod validation. The `AssetRecord` interface here also duplicates the `AssetItem` type.

Consider adding a `getAssetsByName` method to `assetService` and using it here.

src/extensions/core/load3d/ThumbnailRenderer.ts
jtydhr88: client-side render will be removed, no need this fix

src/platform/assets/components/Media3DTop.vue
jtydhr88: removed

src/platform/assets/components/Media3DTop.vue
christian-byrne: non-blocking, fine to address later

christian-byrne (approved):
all feedback addressed. race condition nit tracked as followup issue via coderabbitai. lgtm

PR #9491, opened 2026-03-06
fix: Add timeout and abort mechanism for image upload (#9226)

### LINKED ISSUE #9226: Add timeout and abort mechanism for image upload
## Context
Reported by @christian-byrne in PR #9189: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9189#discussion_r2856396644

## Problem
Currently, if the fetch request in `handleUpload` (src/composables/node/useNodeImageUpload.ts) hangs indefinitely, the loading spinner will remain visible forever since there's no timeout or abort mechanism. This leaves the node in an uploading state with no way to recover.

## Suggested Solution
Add an `AbortController` with a reasonable timeout to the upload fetch request. This would allow the upload to be aborted after a configured duration, cleaning up the `isUploading` state and providing user feedback about the timeout.

## Related Files
- `src/composables/node/useNodeImageUpload.ts`

## Acceptance Criteria
- Add `AbortController` to upload fetch requests
- Implement reasonable timeout duration
- Handle timeout gracefully by resetting `node.isUploading` and notifying the user
- Ensure `onUploadError` callback is invoked on timeout

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9226-Add-timeout-and-abort-mechanism-for-image-upload-3136d73d3650812a9745d9e299acfc85) by [Unito](https://www.unito.io)

Myestery:
Closes #9226

## Summary

Image uploads had no timeout or abort mechanism, meaning a stalled upload could hang indefinitely with no user feedback. This adds a 2-minute timeout using `AbortController` and shows a user-friendly toast message when the upload times out.

## Changes

- `src/composables/node/useNodeImageUpload.ts`: Added `AbortController` with a 120-second timeout to the `uploadFile` function. The abort signal is passed to `fetchApi`. In the `handleUpload` error handler, `AbortError` is now caught separately to display a localized timeout message.
- `src/locales/en/main.json`: Added `uploadTimedOut` i18n translation key.

---
Automated by coderabbit-fixer

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9491-fix-Add-timeout-and-abort-mechanism-for-image-upload-9226-31b6d73d365081d7a7d7f7016f3a71c6) by [Unito](https://www.unito.io)


src/composables/node/useNodeImageUpload.ts
```diff
@@ -28,11 +29,17 @@ const uploadFile = async (
...
+  const timeoutId = setTimeout(() => controller.abort(), UPLOAD_TIMEOUT_MS)
+
   const resp = await api.fetchApi('/upload/image', {
     method: 'POST',
-    body
+    body,
+    signal: controller.signal
   })
 
+  clearTimeout(timeoutId)
```
christian-byrne: issue: `clearTimeout` is unreachable on the error path. If `fetchApi()` throws (network error, abort, etc.), execution jumps to `handleUpload`'s catch block and this `clearTimeout` is never reached -- the 120s timer keeps running and calls `controller.abort()` on a dead controller.

Wrap the fetch + response handling in `try/finally` so the timer is always cleared:
```ts
const controller = new AbortController()
const timeoutId = setTimeout(() => controller.abort(), UPLOAD_TIMEOUT_MS)

try {
  const resp = await api.fetchApi('/upload/image', {
    method: 'POST',
    body,
    signal: controller.signal
  })
  // ... rest of response handling ...
} finally {
  clearTimeout(timeoutId)
}
```

src/composables/node/useNodeImageUpload.ts
```diff
@@ -88,7 +95,11 @@ export const useNodeImageUpload = (
       if (!path) return
       return path
     } catch (error) {
-      useToastStore().addAlert(String(error))
+      if (error instanceof DOMException && error.name === 'AbortError') {
```
christian-byrne: suggestion: The codebase has an existing `isAbortError()` utility in `@/utils/typeGuardUtil`, used consistently in comfyManagerService, customerEventsService, comfyRegistryService, and releaseService. Consider using it here for consistency:
```ts
import { isAbortError } from '@/utils/typeGuardUtil'

if (isAbortError(error)) {
  useToastStore().addAlert(t('g.uploadTimedOut'))
} else {
  useToastStore().addAlert(String(error))
}
```

src/composables/node/useNodeImageUpload.ts
```diff
@@ -28,11 +29,17 @@ const uploadFile = async (
   if (isPasted) body.append('subfolder', 'pasted')
   if (formFields.type) body.append('type', formFields.type)
 
+  const controller = new AbortController()
```
christian-byrne: suggestion (non-blocking): You could simplify this with `AbortSignal.timeout(UPLOAD_TIMEOUT_MS)` instead of the manual AbortController + setTimeout pattern. It's supported in all modern browsers, removes the need for `clearTimeout`, and throws `TimeoutError` instead of `AbortError` -- making the intent unambiguous and future-proofing against user-cancel scenarios.
```ts
const resp = await api.fetchApi('/upload/image', {
  method: 'POST',
  body,
  signal: AbortSignal.timeout(UPLOAD_TIMEOUT_MS)
})
```
This also eliminates the `clearTimeout` issue from my other comment entirely.

src/composables/node/useNodeImageUpload.ts
christian-byrne: nitpick (non-blocking): Currently only the timeout calls `abort()`, so catching `AbortError` as "timed out" is fine. But if a cancel button is added later, users would see "Upload timed out" when they intentionally cancelled. Using `AbortSignal.timeout()` from the other suggestion would sidestep this since it throws `TimeoutError` instead.

src/composables/node/useNodeImageUpload.ts
Myestery: Done — switched to `AbortSignal.timeout(UPLOAD_TIMEOUT_MS)` which eliminates the AbortController, setTimeout, and clearTimeout entirely.

src/composables/node/useNodeImageUpload.ts
Myestery: Fixed by adopting `AbortSignal.timeout()` — no more manual timer to clean up.

src/composables/node/useNodeImageUpload.ts
Myestery: Since `AbortSignal.timeout()` throws `TimeoutError` (not `AbortError`), `isAbortError()` doesn't apply here. Updated the check to `error.name === 'TimeoutError'` instead.

Myestery: Addressed — `AbortSignal.timeout()` throws `TimeoutError`, so future user-cancel (`AbortError`) won't be misreported as a timeout.

christian-byrne (approved):
LGTM

PR #9516, opened 2026-03-07
feat: replace loading indicator with C logo fill loader and pre-Vue splash screen

viva-jinyi:
## Summary

Replace the initial loading indicator with a C logo fill animation and horizontal progress bar, shown as a pre-Vue splash screen to eliminate flash of unstyled content.

## Changes

- **What**:
  - Replace `LogoComfyWaveLoader` with `LogoCFillLoader` (left-to-right fill + progress bar)
  - Inline SVG splash loader in `index.html` visible before JS loads
  - Cache theme hint (`comfy-theme-hint`) in localStorage for flash-free dark/light splash
  - Unify loading into single splash that persists until `workspaceStore.spinner` becomes false
  - Remove redundant loader from `WorkspaceAuthGate` (splash at z-9999 covers it)
  - Update `WorkspaceAuthGate` tests for removed loading indicator DOM

## Review Focus

- Splash removal timing: watch detects `true→false` transition of `spinner` to avoid premature removal
- Theme hint is a localStorage cache only; server settings remain source of truth
- Responsive sizing uses `min()` with vw fallback for small viewports

[screen-capture (1).webm](https://github.com/user-attachments/assets/05f28135-8dcf-49da-83a7-922fe99a5f7b)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9516-feat-replace-loading-indicator-with-C-logo-fill-loader-and-pre-Vue-splash-screen-31c6d73d3650816196fbf7287579c32c) by [Unito](https://www.unito.io)


DrJKL:
Can we do just the theme part and not the C? I kind of love the wave

index.html
```diff
@@ -46,14 +46,111 @@
...
+      }
+      @media (prefers-reduced-motion: reduce) {
+        #splash-loader .wave-group,
+        #splash-loader .wave-path {
+          animation: none;
+        }
+        #splash-loader .wave-group {
+          transform: translateY(-80px);
+        }
+      }
```
DrJKL: We could probably put these into a separate small CSS file, assuming it loads relatively quickly and then gets cached.

src/views/GraphView.vue
```diff
@@ -132,6 +132,14 @@ watch(
     } else {
       document.body.classList.add(DARK_THEME_CLASS)
     }
+    try {
+      localStorage.setItem(
+        'comfy-theme-hint',
+        newTheme.light_theme ? 'light' : 'dark'
+      )
```
DrJKL: I was thinking we could put the actual base foreground and background values into localstorage in the color palette service.

index.html
viva-jinyi: Done — extracted to `public/splash.css`.

src/views/GraphView.vue
viva-jinyi: Done — now caching the actual `bg-color` / `fg-color` values from `completedActivePalette.colors.comfy_base` into `comfy-splash-bg` / `comfy-splash-fg` in localStorage. The splash script reads those directly and falls back to CSS `prefers-color-scheme` when no cache exists.

viva-jinyi:
@DrJKL Totally agree — I love the wave too, so I already swapped index.html to use it! Also addressed your review comments:

- Splash CSS → extracted to `public/splash.css`
- Now caching actual bg/fg color values in localStorage (`comfy-splash-bg`, `comfy-splash-fg`) instead of a light/dark hint
- Added explicit light-mode CSS fallback + `prefers-color-scheme` media query so first load respects the system theme

src/components/loader/LogoCFillLoader.vue
DrJKL, line 1: Is this still used?

viva-jinyi: No, I will delete it

viva-jinyi:
I’m investigating because the white or gray screen, which we had previously prevented from showing, is appearing again.

viva-jinyi:
@DrJKL Addressed all review comments:

- **Splash CSS** → extracted to `public/splash.css`
- **Actual theme colors cached** → `colorPaletteService` now stores real `bg-color`/`fg-color` in localStorage (`comfy-splash-bg`/`comfy-splash-fg`), splash script reads them with `matchMedia` fallback for first load
- **Removed `comfy-theme-hint`** from GraphView since the palette service handles it now

Also fixed a regression from the main merge — `immediate: true` on the splash removal watch caused it to fire before `spinner` was set to `true`, removing the splash prematurely.

viva-jinyi:
All tests passed! Can you review it again?

src/assets/palettes/light.json
```diff
@@ -53,6 +53,7 @@
     "comfy_base": {
       "fg-color": "#222",
       "bg-color": "#DDD",
+      "splash-bg-color": "#FFFFFF",
```
DrJKL: I'd rather not have extra single-use tokens, if we can help it...

viva-jinyi: But if we don’t create a new token for this, the background color in the light theme becomes gray, so you end up seeing a black wave on a gray screen, which is NOT pretty.

DrJKL: Given that it's just for light mode, I _would_ be okay special casing that single color in the splash context only.

viva-jinyi: Are you suggesting we add a conditional only for light mode?

DrJKL: Yeah, I'd rather that than having to duplicate the background in the rest of them. I could also see an override where the special color is optional and we cascade to the default when unset.

viva-jinyi: Sounds good — removed the splash-bg-color token from all palettes and the schema. Now the splash background falls back to bg-color by default, and only the light theme is special-cased to #FFFFFF in `colorPaletteService.ts` (via light_theme flag). No extra tokens needed!

DrJKL (approved):

PR #9521, opened 2026-03-07
chore: standardize on Node 24

benceruleanlu:
## Summary

Standardize the repo's Node contract on 24 while centralizing workflow resolution through `.nvmrc` so local setup, CI, and package metadata stay aligned from one version file.

## Changes

- **What**: Add `package.json` `engines.node = 24.x`, switch every `actions/setup-node` workflow in the repo to `node-version-file: '.nvmrc'`, and update contributor and Playwright docs to point to `.nvmrc` as the Node source of truth.

## Review Focus

The workflow behavior should be unchanged apart from sourcing the Node version from `.nvmrc` instead of repeating literals like `20`, `22`, `24.x`, or `lts/*`. GitHub's formatter also moved the new `engines` block to the package metadata section near the end of `package.json`.

### DIFF
--- a/.github/workflows/release-biweekly-comfyui.yaml
+++ b/.github/workflows/release-biweekly-comfyui.yaml
@@ -82,7 +82,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v6
         with:
-          node-version: lts/*
+          node-version: '24.x'
 
       - name: Install dependencies
         working-directory: frontend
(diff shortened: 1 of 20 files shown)


benceruleanlu:
Three workflows were not on 24 before, but 22, 20, and 20. Don't know why, checked PRs, nothing found.

.github/workflows/release-biweekly-comfyui.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v6
         with:
-          node-version: lts/*
+          node-version-file: '.nvmrc'
```
DrJKL:
```suggestion
          node-version-file: 'frontend/.nvmrc'
```

DrJKL (approved):

PR #9531, opened 2026-03-07
fix: Prevent corruption of workflow data due to checkState during graph loading

pythongosssss:
## Summary

During workflow loading, the workflow data & active workflow object can be out of sync, meaning any checkState calls will overwrite data into the wrong workflow.

Recreation steps:
* Open 2-3 workflows
* Enter builder mode > select step
* Select some different inputs on each
* Quickly tap the shift key (this triggers checkState) while switching tabs
* After a while, you'll see the wrong inputs on the workflows

Alternatively, register an extension that guarantees to call checkState during the bad phase, run this in browser devtools and switch tabs:
```
window.app.registerExtension({
  name: 'bad',
  async afterConfigureGraph() {
    window.app.extensionManager.workflow.activeWorkflow.changeTracker.checkState()
  }
})
```

## Changes

- **What**: 
- Add loading graph flag
- Prevent checkState calls while loading
- Prevent app mode data sync while loading

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9531-fix-Prevent-corruption-of-workflow-data-due-to-checkState-during-graph-loading-31c6d73d365081e2ab91d9145bf1d025) by [Unito](https://www.unito.io)

### DIFF
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -9,6 +9,7 @@ import { layoutStore } from '@/renderer/core/layout/store/layoutStore'
 import { flushScheduledSlotLayoutSync } from '@/renderer/extensions/vueNodes/composables/useSlotElementTracking'
 
 import { st, t } from '@/i18n'
+import { ChangeTracker } from '@/scripts/changeTracker'
 import type { IContextMenuValue } from '@/lib/litegraph/src/interfaces'
 import {
   LGraph,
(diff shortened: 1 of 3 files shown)


pythongosssss:
@coderabbitai Log an issue to write a unit test to prevent a regression for this issue.
We should be able to recreate it for ensuring the test is valid by disabling the `ChangeTracker.isLoadingGraph` fix, then in the test register a custom extension that calls `changeTracker.checkState()` in the `onAfterGraphConfigured` hook, switch tabs, observe that the workflow is overwritten with the other tab data, then re-enable the fix and ensure test passes

pythongosssss:
@coderabbitai Log an issue to see if we can rearchitect this to avoid getting into the desynced workflow/graph data state completley

DrJKL (approved):

AustinMroz (approved):
PRs like this make me very happy for `git diff -w`/(hide whitespace)

PR #9549, opened 2026-03-08
fix: restore queue job details popover

benceruleanlu:
## Summary

Restore the existing job details popover in the expanded queue overlay after the list row UI moved to `AssetsListItem`.

## Changes

- **What**: Add delayed hover-driven job details popover support back to `JobAssetsList`, including teleported positioning, row-to-popover hover handoff, and focused component coverage for the timing behavior.

## Review Focus

Please verify the hover transition between the queue row and the teleported popover, especially around positioning and whether the popover remains reachable in the top menu queue overlay.

N/A

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9549-fix-restore-queue-job-details-popover-31d6d73d3650815eaf45dcc2a14d3dee) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/queue/job/JobAssetsList.test.ts
+++ b/src/components/queue/job/JobAssetsList.test.ts
@@ -273,4 +273,34 @@ describe('JobAssetsList', () => {
     await nextTick()
     expect(wrapper.findComponent(JobDetailsPopoverStub).exists()).toBe(false)
   })
+
+  it('shows the new popover after the previous row hides while the next row stays hovered', async () => {
+    vi.useFakeTimers()
+    const firstJob = buildJob({ id: 'job-1' })
+    const secondJob = buildJob({ id: 'job-2', title: 'Job 2' })
+    const wrapper = mountJobAssetsList([firstJob, secondJob])
+    const firstRow = wrapper.find('[data-job-id="job-1"]')
+    const secondRow = wrapper.find('[data-job-id="job-2"]')
+
+    await firstRow.trigger('mouseenter')
+    await vi.advanceTimersByTimeAsync(200)
+    await nextTick()
+    expect(wrapper.findComponent(JobDetailsPopoverStub).props('jobId')).toBe(
+      'job-1'
+    )
+
+    await firstRow.trigger('mouseleave')
+    await secondRow.trigger('mouseenter')
+
+    await vi.advanceTimersByTimeAsync(150)
+    await nextTick()
+    expect(wrapper.findComponent(JobDetailsPopoverStub).exists()).toBe(false)
+
+    await vi.advanceTimersByTimeAsync(50)
+    await nextTick()
+
+    const popover = wrapper.findComponent(JobDetailsPopoverStub)
+    expect(popover.exists()).toBe(true)
+    expect(popover.props('jobId')).toBe('job-2')
+  })
 })
--- a/src/components/queue/job/JobAssetsList.test.ts
+++ b/src/components/queue/job/JobAssetsList.test.ts
@@ -165,6 +165,26 @@ describe('JobAssetsList', () => {
     expect(wrapper.emitted('viewItem')).toBeUndefined()
   })
 
+  it('emits viewItem from the View button for completed jobs without preview output', async () => {
+    const job = buildJob({
+      iconImageUrl: undefined,
+      taskRef: createTaskRef()
+    })
+    const wrapper = mountJobAssetsList([job])
+    const jobRow = wrapper.find(`[data-job-id="${job.id}"]`)
+
+    await jobRow.trigger('mouseenter')
+    const viewButton = wrapper
+      .findAll('button')
+      .find((button) => button.text() === 'menuLabels.View')
+    expect(viewButton).toBeDefined()
+
+    await viewButton!.trigger('click')
+    await nextTick()
+
+    expect(wrapper.emitted('viewItem')).toEqual([[job]])
+  })
+
   it('shows and hides the job details popover with hover delays', async () => {
     vi.useFakeTimers()
     const job = buildJob()
--- a/src/components/queue/job/JobAssetsList.test.ts
+++ b/src/components/queue/job/JobAssetsList.test.ts
@@ -92,11 +92,11 @@ const mountJobAssetsList = (jobs: JobListItem[]) => {
   })
 }
 
-afterEach(() => {
-  vi.useRealTimers()
-})
-
 describe('JobAssetsList', () => {
+  afterEach(() => {
+    vi.useRealTimers()
+  })
+
   it('emits viewItem on preview-click for completed jobs with preview', async () => {
     const job = buildJob()
     const wrapper = mountJobAssetsList([job])
--- a/src/components/queue/job/JobAssetsList.test.ts
+++ b/src/components/queue/job/JobAssetsList.test.ts
@@ -247,4 +247,30 @@ describe('JobAssetsList', () => {
     await nextTick()
     expect(wrapper.findComponent(JobDetailsPopoverStub).exists()).toBe(false)
   })
+
+  it('clears the previous popover when hovering a new row briefly and leaving the list', async () => {
+    vi.useFakeTimers()
+    const firstJob = buildJob({ id: 'job-1' })
+    const secondJob = buildJob({ id: 'job-2', title: 'Job 2' })
+    const wrapper = mountJobAssetsList([firstJob, secondJob])
+    const firstRow = wrapper.find('[data-job-id="job-1"]')
+    const secondRow = wrapper.find('[data-job-id="job-2"]')
+
+    await firstRow.trigger('mouseenter')
+    await vi.advanceTimersByTimeAsync(200)
+    await nextTick()
+    expect(wrapper.findComponent(JobDetailsPopoverStub).props('jobId')).toBe(
+      'job-1'
+    )
+
+    await firstRow.trigger('mouseleave')
+    await secondRow.trigger('mouseenter')
+    await vi.advanceTimersByTimeAsync(100)
+    await nextTick()
+    await secondRow.trigger('mouseleave')
+
+    await vi.advanceTimersByTimeAsync(150)
+    await nextTick()
+    expect(wrapper.findComponent(JobDetailsPopoverStub).exists()).toBe(false)
+  })
 })
--- a/src/components/queue/job/JobAssetsList.test.ts
+++ b/src/components/queue/job/JobAssetsList.test.ts
@@ -1,11 +1,21 @@
 import { mount } from '@vue/test-utils'
-import { describe, expect, it, vi } from 'vitest'
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { defineComponent, nextTick } from 'vue'
 
 import type { JobGroup, JobListItem } from '@/composables/queue/useJobList'
 import type { JobListItem as ApiJobListItem } from '@/platform/remote/comfyui/jobs/jobTypes'
 import { ResultItemImpl, TaskItemImpl } from '@/stores/queueStore'
 
 import JobAssetsList from './JobAssetsList.vue'
+
+const JobDetailsPopoverStub = defineComponent({
+  name: 'JobDetailsPopover',
+  props: {
+    jobId: { type: String, required: true },
+    workflowId: { type: String, default: undefined }
+  },
+  template: '<div class="job-details-popover-stub" />'
+})
 
 vi.mock('vue-i18n', () => {
   return {
@@ -46,6 +56,7 @@ const createTaskRef = (preview?: ResultItemImpl): TaskItemImpl => {
     create_time: Date.now(),
     preview_output: null,
     outputs_count: preview ? 1 : 0,
+    workflow_id: 'workflow-1',
     priority: 0
   }
   const flatOutputs = preview ? [preview] : []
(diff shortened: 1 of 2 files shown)


src/components/queue/job/JobAssetsList.vue
```diff
@@ -8,78 +8,102 @@
...
+              :aria-label="t('g.delete')"
+              @click.stop="emitDeleteItem(job)"
+            >
+              <i class="icon-[lucide--trash-2] size-4" />
+            </Button>
+            <Button
+              v-else-if="job.state === 'completed'"
+              variant="textonly"
+              size="sm"
+              @click.stop="emitViewItem(job)"
```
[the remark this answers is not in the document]
benceruleanlu: Fixed in `1227cda68` and carried forward in the latest branch state. The explicit `View` button now emits again for completed jobs even when there is no preview output, which restores the pre-change behavior. I also added a regression test in `JobAssetsList.test.ts` covering the completed-without-preview path.

src/components/queue/job/JobAssetsList.vue
```diff
@@ -90,11 +114,97 @@ const emit = defineEmits<{
...
+  clearShowTimer()
+  activeDetails.value = null
+  activeRowElement.value = null
+  popoverPosition.value = null
+}
+
+const scheduleDetailsShow = (job: JobListItem, rowElement: HTMLElement) => {
+  clearHideTimer()
+  clearShowTimer()
+  activeRowElement.value = rowElement
```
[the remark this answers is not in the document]
benceruleanlu: Fixed in `f0281e4bd`. The hide timer now tracks which job it belongs to, so briefly hovering a second row no longer cancels the first row’s pending hide unless that second row actually becomes the active popover. I also added a regression test for the A → B → exit sequence in `JobAssetsList.test.ts`.

src/components/queue/job/JobAssetsList.vue
```diff
@@ -90,11 +114,104 @@ const emit = defineEmits<{
...
+const scheduleDetailsShow = (job: JobListItem, rowElement: HTMLElement) => {
+  clearShowTimer()
+  activeRowElement.value = rowElement
+  showTimer.value = window.setTimeout(() => {
+    activeDetails.value = {
+      jobId: job.id,
+      workflowId: job.taskRef?.workflowId
+    }
+    showTimer.value = null
+    void nextTick(updatePopoverPosition)
```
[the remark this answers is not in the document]
benceruleanlu: Fixed in `8da207973`. The delayed show now re-applies `activeRowElement` when the timeout fires, so if row A hides while row B is still waiting to open, B still has a valid anchor when `updatePopoverPosition()` runs. I also added a regression test in `JobAssetsList.test.ts` for the A → B handoff where B stays hovered long enough to open.

christian-byrne:
How critical is this? What does it need to be backported to?

benceruleanlu:
> How critical is this? What does it need to be backported to?

I don't think it's that critical, maybe backported to last minor version but it's not that serious, just a feature.

src/components/queue/job/JobAssetsList.vue
```diff
@@ -90,11 +114,105 @@ const emit = defineEmits<{
...
+  }, 200)
+}
+
+const scheduleDetailsHide = (jobId?: string) => {
+  if (!jobId) return
+
+  clearShowTimer()
+  if (hideTimerJobId.value && hideTimerJobId.value !== jobId) {
+    return
+  }
```
christian-byrne, line 177: issue: The guard `if (hideTimerJobId.value && hideTimerJobId.value !== jobId) return` silently drops hide requests when the jobId doesn't match the pending timer. This can leave the popover stuck in a rapid-switch scenario: hover row A (popover shows) → hover row B briefly → leave row B. `onJobLeave('B')` calls `scheduleDetailsHide('B')`, but `hideTimerJobId` is `'A'`, so B's hide is dropped. If A's timer was already cleared by `onJobEnter` -> `scheduleDetailsShow` -> `clearShowTimer()` (which doesn't clear the hide timer), no cleanup fires.

Consider always clearing and rescheduling:
```ts
function scheduleDetailsHide(jobId?: string) {
  if (!jobId) return
  clearShowTimer()
  clearHideTimer()
  hideTimerJobId.value = jobId
  hideTimer.value = window.setTimeout(() => { ... }, 150)
}
```
And call `clearHideTimer()` at the top of `onJobEnter` before `scheduleDetailsShow`.

src/components/queue/job/JobAssetsList.vue
```diff
@@ -8,78 +8,102 @@
...
 
+import JobDetailsPopover from '@/components/queue/job/JobDetailsPopover.vue'
 import Button from '@/components/ui/button/Button.vue'
 import type { JobGroup, JobListItem } from '@/composables/queue/useJobList'
 import AssetsListItem from '@/platform/assets/components/AssetsListItem.vue'
 import { iconForJobState } from '@/utils/queueDisplay'
 import { isActiveJobState } from '@/utils/queueUtil'
 
-defineProps<{ displayedJobGroups: JobGroup[] }>()
+const props = defineProps<{ displayedJobGroups: JobGroup[] }>()
```
christian-byrne: suggestion (non-blocking): AGENTS.md prefers reactive props destructuring over `const props = defineProps<...>()`. Consider:
```ts
const { displayedJobGroups } = defineProps<{ displayedJobGroups: JobGroup[] }>()
```
Then reference `displayedJobGroups` directly in the watcher at line 280.

src/components/queue/job/JobAssetsList.vue
```diff
@@ -90,11 +114,105 @@ const emit = defineEmits<{
...
 const { t } = useI18n()
 const hoveredJobId = ref<string | null>(null)
+const activeDetails = ref<{ jobId: string; workflowId?: string } | null>(null)
+const activeRowElement = ref<HTMLElement | null>(null)
+const popoverPosition = ref<{ top: number; right: number } | null>(null)
+const hideTimer = ref<number | null>(null)
+const hideTimerJobId = ref<string | null>(null)
+const showTimer = ref<number | null>(null)
+
+const clearHideTimer = () => {
```
christian-byrne, line 123: suggestion (non-blocking): AGENTS.md rule 24 says "Do not use function expressions if it's possible to use function declarations instead." All ~15 new functions in this component use `const fn = () => {}`. Consider converting to `function` declarations. (I see QueueJobItem.vue uses the same pattern, so this may be worth aligning across both in a follow-up.)

src/components/queue/job/JobAssetsList.vue
```diff
@@ -90,11 +114,105 @@ const emit = defineEmits<{
 
 const { t } = useI18n()
 const hoveredJobId = ref<string | null>(null)
+const activeDetails = ref<{ jobId: string; workflowId?: string } | null>(null)
```
christian-byrne, line 116: suggestion (non-blocking): The timer-based hover logic (show/hide scheduling, position calculation, clear/reset) is ~60-70% structurally duplicated with QueueJobItem.vue. A `usePopoverHover` composable could eliminate ~80 lines here and guarantee consistent behavior. Could be a follow-up PR.

src/components/queue/job/JobAssetsList.vue
```diff
@@ -8,78 +8,102 @@
...
-        :is-video-preview="isVideoPreviewJob(job)"
-        :preview-alt="job.title"
-        :icon-name="job.iconName ?? iconForJobState(job.state)"
-        :icon-class="getJobIconClass(job)"
-        :primary-text="job.title"
-        :secondary-text="job.meta"
-        :progress-total-percent="job.progressTotalPercent"
-        :progress-current-percent="job.progressCurrentPercent"
-        @mouseenter="hoveredJobId = job.id"
+        class="relative"
```
christian-byrne: nitpick (non-blocking): `class="relative"` on this wrapper div appears unused -- the popover is teleported to body with `position: fixed`, so nothing positions relative to this element. Safe to remove.

src/components/queue/job/JobAssetsList.vue
```diff
@@ -90,11 +114,105 @@ const emit = defineEmits<{
...
+  clearHideTimer()
+  clearShowTimer()
+  activeDetails.value = null
+  activeRowElement.value = null
+  popoverPosition.value = null
+}
+
+const scheduleDetailsShow = (job: JobListItem, rowElement: HTMLElement) => {
+  clearShowTimer()
+  activeRowElement.value = rowElement
```
christian-byrne: nitpick (non-blocking): `activeRowElement.value = rowElement` is set here (L161, before the timer), again inside the timer callback (L163), and also in `onJobEnter` (L207). The L161 pre-timer assignment is redundant since L207 already handles it before calling this function.

src/components/queue/job/JobAssetsList.test.ts
```diff
@@ -71,10 +82,20 @@ const mountJobAssetsList = (jobs: JobListItem[]) => {
...
+    global: {
+      stubs: {
+        teleport: true,
+        JobDetailsPopover: JobDetailsPopoverStub
+      }
+    }
   })
 }
 
+afterEach(() => {
```
christian-byrne: nitpick (non-blocking): `afterEach(vi.useRealTimers)` is at module scope outside the `describe` block. It works, but runs after every test even those not using fake timers. Moving it inside the `describe` or co-locating with `vi.useFakeTimers()` calls is more conventional.

src/components/queue/job/JobAssetsList.vue
benceruleanlu: I'm a bit confused, in this scenario it seems like everything went expectedly because B has nothing to hide. There is a regression test I added for this A → B → leave path as well. leaving row A schedules A’s hide, and entering row B does not clear that hide timer because onJobEnter only clears the show timer. That means the existing A hide still fires and cleans up the stale popover.

src/components/queue/job/JobAssetsList.vue
benceruleanlu: Made the change

src/components/queue/job/JobAssetsList.vue
benceruleanlu: Removed

src/components/queue/job/JobAssetsList.test.ts
benceruleanlu: Moved

src/components/queue/job/JobAssetsList.vue
benceruleanlu: Removed

src/components/queue/job/JobAssetsList.vue
benceruleanlu: Will do in the same followup

src/components/queue/job/JobAssetsList.vue
benceruleanlu: Will do in the same followup

src/components/queue/job/JobAssetsList.vue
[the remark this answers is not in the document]
christian-byrne: verified - `emitCompletedViewItem` emits unconditionally for completed jobs now, and a regression test covers the no-preview path

src/components/queue/job/JobAssetsList.vue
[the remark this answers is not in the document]
christian-byrne: verified - hideTimerJobId tracking prevents the stale popover scenario, and the A->B->exit regression test passes

src/components/queue/job/JobAssetsList.vue
```diff
@@ -8,78 +8,102 @@
...
-        :icon-class="getJobIconClass(job)"
-        :primary-text="job.title"
-        :secondary-text="job.meta"
-        :progress-total-percent="job.progressTotalPercent"
-        :progress-current-percent="job.progressCurrentPercent"
-        @mouseenter="hoveredJobId = job.id"
+        class="relative"
+        :data-job-id="job.id"
+        @mouseenter="onJobEnter(job, $event)"
         @mouseleave="onJobLeave(job.id)"
```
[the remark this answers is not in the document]
christian-byrne: valid point - the existing QueueJobItem.vue also lacks keyboard paths for its popover, so this should be addressed holistically across both components

src/components/queue/job/JobAssetsList.vue
```diff
@@ -90,11 +114,97 @@ const emit = defineEmits<{
...
+  const rowElement = activeRowElement.value
+  if (!rowElement) return
+
+  const rect = rowElement.getBoundingClientRect()
+  const gap = 8
+  popoverPosition.value = {
+    top: rect.top,
+    right: window.innerWidth - rect.left + gap
+  }
+}
```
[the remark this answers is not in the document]
christian-byrne: same issue exists in QueueJobItem.vue - worth fixing in both at once

src/components/queue/job/JobAssetsList.vue
christian-byrne: you are right - I was conflating the show and hide timers. B has no popover to hide, so the early return is correct. A's hide timer is unaffected by `onJobEnter(B)` since `scheduleDetailsShow` only clears the show timer. the test confirms the sequence works.

src/components/queue/job/JobAssetsList.vue
christian-byrne: @coderabbitai Make an Issue to do this in a followup PR on a future date

src/components/queue/job/JobAssetsList.vue
christian-byrne: @coderabbitai Make an Issue to do this in a followup PR on a future date

christian-byrne (approved):
all feedback addressed and verified in code. three non-blocking items tagged for followup issues (keyboard a11y, scroll repositioning, function declarations + composable extraction). lgtm

PR #9551, opened 2026-03-08
fix: update FormDropdown filteredItems when items prop changes

christian-byrne:
## Problem

The Load Video node's Media Assets panel dropdown didn't show videos when switching between asset types. The `filteredItems` ref stayed stale because the update chain relied on a computed `itemsKey` (string of IDs) — if the key didn't change (e.g., switching between two empty arrays or identical item sets), the watcher never fired.

## Solution

- Replace computed `itemsKey` with a `Symbol` ref that always produces a unique value on change
- Add a `watch` on the `items` prop that immediately re-runs the searcher with cancellation support
- Add unit tests covering prop changes with same IDs, different names, and empty↔non-empty transitions

## Testing

- `pnpm test:unit -- FormDropdown.test.ts` — 3/3 pass
- All quality gates pass (format, lint, typecheck)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9551-fix-update-FormDropdown-filteredItems-when-items-prop-changes-31d6d73d3650812bb325e5c31036a5a0) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
 import Popover from 'primevue/popover'
-import { computed, ref, useTemplateRef } from 'vue'
+import { computed, ref, useTemplateRef, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 
 import { useToastStore } from '@/platform/updates/common/toastStore'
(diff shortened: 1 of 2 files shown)


AustinMroz (approved):

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -101,10 +101,30 @@ const maxSelectable = computed(() => {
   return 1
 })
 
-const itemsKey = computed(() => items.map((item) => item.id).join('|'))
+const itemsKey = ref<symbol>(Symbol())
 
 const filteredItems = ref<FormDropdownItem[]>([])
 
+watch(
```
DrJKL, line 108: https://vueuse.org/core/computedAsync/#oncancel ?

DrJKL: Open to following up on this later too :-)

PR #9622, opened 2026-03-08
fix: 3D asset disappears when switching to image output in app mode

christian-byrne:
## Summary

Fix 3D asset disappearing when switching between 3D and image outputs in app mode — missing `onUnmounted` cleanup leaked WebGL contexts.

## Changes

- **What**: Add `onUnmounted` hook to `Preview3d.vue` that calls `viewer.cleanup()`, releasing the WebGL context when Vue destroys the component via its v-if chain. Add unit tests covering init, cleanup on unmount, and remount behavior.

## Review Focus

When switching outputs in app mode, Vue's v-if chain destroys and recreates `Preview3d`. Without `onUnmounted` cleanup, the old `Load3d` instance (WebGL context, RAF loop, ResizeObserver) leaks. After ~8-16 toggles, the browser's WebGL context limit is exhausted and new 3D viewers silently fail to render.


src/renderer/extensions/linearMode/flattenNodeOutput.ts
```diff
@@ -1,20 +1,38 @@
 import type { NodeExecutionOutput, ResultItem } from '@/schemas/apiSchema'
+import { resultItemType } from '@/schemas/apiSchema'
 import { ResultItemImpl } from '@/stores/queueStore'
 
+const EXCLUDED_KEYS = new Set(['animated'])
+
+function isResultItemLike(item: unknown): item is ResultItem {
```
DrJKL: Why aren't we able to just use the Zod schema?

src/renderer/extensions/linearMode/flattenNodeOutput.ts
```diff
@@ -1,20 +1,38 @@
...
-  if (nodeOutput.gifs) knownOutputs.gifs = nodeOutput.gifs as ResultItem[]
-  if (nodeOutput['3d']) knownOutputs['3d'] = nodeOutput['3d'] as ResultItem[]
-
-  return Object.entries(knownOutputs).flatMap(([mediaType, outputs]) =>
-    outputs.map(
-      (output) => new ResultItemImpl({ ...output, mediaType, nodeId })
+  return Object.entries(nodeOutput)
+    .filter(([key, value]) => !EXCLUDED_KEYS.has(key) && Array.isArray(value))
+    .flatMap(([mediaType, items]) =>
+      (items as unknown[])
```
DrJKL: I like that we have the filter, I don't love that we have to cast to `unknown[]` first, makes me think we're missing something in the type upstream.

DrJKL (dismissed):

src/stores/resultItemParsing.ts
```diff
@@ -0,0 +1,52 @@
...
+ * `filename` and `subfolder` as optional (matching the wire format), but
+ * a ResultItemImpl needs both to construct a valid URL.
+ */
+function isResultItem(item: unknown): item is ResultItem {
+  if (!item || typeof item !== 'object' || Array.isArray(item)) return false
+
+  const candidate = item as Record<string, unknown>
+
+  if (typeof candidate.filename !== 'string') return false
+  if (typeof candidate.subfolder !== 'string') return false
```
christian-byrne: issue: The new `isResultItem` guard is significantly stricter than the old `isResultItemLike` from `jobOutputCache.ts`. The old guard accepted items with *any subset* of `{filename, subfolder, type}` (at least one present). The new guard requires *both* `filename` AND `subfolder` as strings.

For example, an item like `{ filename: 'no-subfolder.png' }` was previously accepted by the job output cache path -- it would get `subfolder: ''` via `ResultItemImpl` constructor's `?? ''` fallback. Now it's silently dropped.

The test change in `queueStore.test.ts` (flatOutputs length from 1 to 0) confirms this is an observable behavioral change. While arguably more correct, it could break existing custom node outputs that only send `filename` without `subfolder`. Should this strictness change be called out explicitly in the PR description so reviewers are aware it's intentional?

src/renderer/extensions/linearMode/flattenNodeOutput.ts
```diff
@@ -1,20 +1,10 @@
...
-  if (nodeOutput.gifs) knownOutputs.gifs = nodeOutput.gifs as ResultItem[]
-  if (nodeOutput['3d']) knownOutputs['3d'] = nodeOutput['3d'] as ResultItem[]
-
-  return Object.entries(knownOutputs).flatMap(([mediaType, outputs]) =>
-    outputs.map(
-      (output) => new ResultItemImpl({ ...output, mediaType, nodeId })
-    )
-  )
+  return parseNodeOutput(nodeId, nodeOutput)
 }
```
christian-byrne, line 10: issue (non-blocking): The old `flattenNodeOutput` used a hardcoded allowlist (`audio`, `images`, `video`, `gifs`, `3d`). The new `parseNodeOutput` uses a denylist (excludes only `animated` and `text`). This means any new custom node output key (e.g., `meshes`, `point_clouds`, `thumbnails`, `masks`) will now be included in linear mode results, whereas before they were silently ignored.

This broadens the surface area and could cause unexpected items to appear in the linear mode preview. Was this change intentional? If so, it's worth noting in the PR description since it's a semantic change, not just a refactor.

src/stores/resultItemParsing.ts
```diff
@@ -0,0 +1,52 @@
+import type { NodeExecutionOutput, ResultItem } from '@/schemas/apiSchema'
+import { resultItemType } from '@/schemas/apiSchema'
+import { ResultItemImpl } from '@/stores/queueStore'
+
+const METADATA_KEYS = new Set(['animated', 'text'])
```
christian-byrne, line 5: suggestion (non-blocking): `METADATA_KEYS` excludes `text`, but the old `queueStore.calculateFlatOutputs()` did NOT exclude `text` -- it only excluded `animated` (via `_.omit` in the constructor). The old queue store would create a `ResultItemImpl` from `text: "hello"` with `filename: ''`, which would then fail `filterPreviewable`. The new behavior (filtering `text` out early) is better.

However, `this.outputs` in the `TaskItemImpl` constructor still does `_.omit(nodeOutputs, 'animated')`, so `animated` is now filtered twice (once in the constructor, once in `parseNodeOutput`). The redundant `_.omit` could be cleaned up, or at least documented to avoid confusion about which layer is responsible for the filtering.

src/stores/resultItemParsing.ts
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  return true
+}
+
+export function parseNodeOutput(
+  nodeId: string | number,
+  nodeOutput: NodeExecutionOutput
+): ResultItemImpl[] {
+  return Object.entries(nodeOutput)
```
christian-byrne, line 34: question: `Object.entries(nodeOutput)` iteration order depends on property insertion order, whereas the old `flattenNodeOutput` had a fixed allowlist order (`audio`, `images`, `video`, `gifs`, `3d`). Could the changed ordering affect which item becomes the default `previewOutput`?

`TaskItemImpl.previewOutput` picks the first `type === 'output'` item from `previewableOutputs`. If a node outputs both `images` and `video`, the order they appear in `flatOutputs` may now differ from before, potentially changing which preview is shown by default.

src/stores/resultItemParsing.ts
```diff
@@ -0,0 +1,52 @@
+import type { NodeExecutionOutput, ResultItem } from '@/schemas/apiSchema'
+import { resultItemType } from '@/schemas/apiSchema'
+import { ResultItemImpl } from '@/stores/queueStore'
```
christian-byrne, line 3: nitpick (non-blocking): There's a circular module dependency: `resultItemParsing.ts` imports `ResultItemImpl` from `queueStore.ts`, and `queueStore.ts` imports `parseTaskOutput` from `resultItemParsing.ts`. This works at runtime because the imports are used at call time (not during module initialization), but it introduces fragility.

Consider moving `ResultItemImpl` to a separate module, or having `parseNodeOutput` return plain objects and letting callers construct `ResultItemImpl`.

src/renderer/extensions/linearMode/Preview3d.vue
```diff
@@ -13,11 +13,16 @@ const containerRef = useTemplateRef('containerRef')
...
+  [containerRef, () => modelUrl],
+  async () => {
+    if (!containerRef.value || !modelUrl) return
 
-  await viewer.value.initializeStandaloneViewer(containerRef.value, modelUrl)
-})
+    viewer.value.cleanup()
+    await viewer.value.initializeStandaloneViewer(containerRef.value, modelUrl)
+  },
+  { flush: 'post' }
```
christian-byrne, line 24: note (non-blocking): The actual bug fix (adding `onUnmounted` cleanup + `viewer.value.cleanup()` call before reinit in the watcher) looks correct and well-tested. The `{ flush: 'post' }` option ensures the DOM element exists before init. Tests cover mount, unmount, remount, and prop change. No concerns here.

src/renderer/extensions/linearMode/flattenNodeOutput.ts
christian-byrne: The Zod schema (`zResultItem`) has `filename` and `subfolder` as `.optional()` — matching the wire format where custom nodes can omit them. But `ResultItemImpl` needs both to build a valid URL (`/view?filename=X&subfolder=Y&type=Z`).

`zResultItem.safeParse()` would pass items like `{ type: 'output' }` (no filename), which would produce broken URLs downstream. The `isResultItem` guard is intentionally stricter: it validates the structural shape the Zod schema can't enforce without breaking the wire format.

We could use both (`safeParse` + stricter check), but it'd just be redundant — the manual guard already covers the Zod checks.

src/renderer/extensions/linearMode/flattenNodeOutput.ts
christian-byrne: Agreed the cast is unfortunate. The root cause is `NodeExecutionOutput` uses `.passthrough()` on the Zod schema, so non-standard output keys (e.g. `gifs`, `3d`, custom node keys) have type `unknown` in the inferred TS type. Only the known keys (`audio`, `images`, `video`) get `ResultItem[]` typing.

When we iterate with `Object.entries()`, the values are typed as the union of all property types, which includes `unknown`. The `unknown[]` cast + `isResultItem` guard is the type-narrowing path for that.

To fix upstream, we'd need to change the schema to something like `z.record(z.string(), z.array(zResultItem))` for the passthrough portion, but that would over-constrain `animated` (boolean array) and `text` (string). The current approach keeps the schema accurate to the wire format and handles the type narrowing at the consumption site.

src/stores/resultItemParsing.ts
christian-byrne: Fixed in 753585727. Softened `isResultItem` to only require `filename` (string). `subfolder` is now optional — `ResultItemImpl` constructor falls back to `''`. This restores compatibility with custom nodes that send `{ filename: 'foo.png' }` without `subfolder`.

src/stores/resultItemParsing.ts
christian-byrne: Fixed in 1b8a3fb73. Removed the redundant `_.omit(nodeOutputs, 'animated')` from `TaskItemImpl` constructor — `METADATA_KEYS` in `parseNodeOutput` already handles this. Also removed the now-unused `es-toolkit/compat` import.

src/renderer/extensions/linearMode/flattenNodeOutput.ts
christian-byrne: Intentional. The denylist approach is needed to support non-standard output keys from custom nodes (the original bug — `a_images`, `b_images` from image compare nodes were silently dropped by the old allowlist). The `isResultItem` guard + `supportsPreview` filter ensures only valid previewable items make it through, so unknown keys with non-ResultItem arrays are safely excluded.

src/stores/resultItemParsing.ts
christian-byrne: Good catch. In practice, `previewOutput` picks the first `type === 'output'` item, so order only matters when a node outputs multiple media types with `type: 'output'`. The old fixed order was `audio, images, video, gifs, 3d` — JS `Object.entries()` follows insertion order, which for server JSON is typically the order the Python dict was built. For most nodes this is a single media type. The risk is low, but worth noting — if this causes issues we can add an explicit priority sort.

src/stores/resultItemParsing.ts
christian-byrne: Acknowledged. The cycle is safe at runtime (call-time imports only), but fragile. Clean fix would be extracting `ResultItemInit` to a separate types module and having `parseNodeOutput` return plain objects. Leaving as a followup since the fix requires touching all consumers (`jobOutputCache`, `flattenNodeOutput`, `queueStore`).

DrJKL (approved):

src/renderer/extensions/linearMode/Preview3d.vue
```diff
@@ -13,11 +13,16 @@ const containerRef = useTemplateRef('containerRef')
 
 const viewer = ref(useLoad3dViewer())
 
-watch([containerRef, () => modelUrl], async () => {
-  if (!containerRef.value || !modelUrl) return
+watch(
```
DrJKL, line 16: Non-blocking: this async `watch` can race when `modelUrl` changes quickly. We call `cleanup()` and then await a new init, but there is no invalidation/cancellation for prior runs. An earlier run can complete after a later one and reinitialize/register listeners unexpectedly. Consider using `onCleanup` (or a local generation token) to ignore stale async completions.

src/stores/resultItemParsing.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Requires `filename` (string) since ResultItemImpl needs it for a valid URL.
+ * `subfolder` is optional here — ResultItemImpl constructor falls back to ''.
+ */
+function isResultItem(item: unknown): item is ResultItem {
+  if (!item || typeof item !== 'object' || Array.isArray(item)) return false
+
+  const candidate = item as Record<string, unknown>
+
+  if (typeof candidate.filename !== 'string') return false
```
DrJKL, line 18: Non-blocking: `isResultItem` no longer validates `subfolder` type. If a passthrough payload includes `subfolder: { ... }`, it now passes and later serializes as `subfolder=[object Object]` in URLs. Consider restoring a guard like `candidate.subfolder === undefined || typeof candidate.subfolder === 'string'`.

DrJKL (commented):
Non-blocking review notes from manual pass.

PR #9630, opened 2026-03-09
fix: support non-standard output keys in app mode preview

christian-byrne:
## Summary

Nodes with non-standard output keys (e.g. ImageCompare outputs `a_images`/`b_images`) show no preview in App Mode because `flattenNodeOutput` used a hardcoded allowlist of 5 keys.

## Changes

- **What**: Replace hardcoded key checks (`images`, `audio`, `video`, `gifs`, `3d`) with dynamic `Object.entries()` iteration, excluding only `animated`, and validating items with `isResultItemLike` — matching the existing pattern in `jobOutputCache.ts`
- **Tests**: Added cases for non-standard keys, excluded keys, non-ResultItem arrays, and partial objects

## Review Focus

- `isResultItemLike` guard validates `filename` (string), `subfolder` (string), and optionally `type` via zod — sufficient to avoid false positives from arbitrary output data

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9630-fix-support-non-standard-output-keys-in-app-mode-preview-31e6d73d365081228c43c7a434600392) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/linearMode/flattenNodeOutput.test.ts
+++ b/src/renderer/extensions/linearMode/flattenNodeOutput.test.ts
@@ -82,4 +82,41 @@ describe(flattenNodeOutput, () => {
     const result = flattenNodeOutput(['1', output])
     expect(result).toEqual([])
   })
+
+  it('flattens non-standard output keys with ResultItem-like values', () => {
+    const output = makeOutput({
+      a_images: [{ filename: 'before.png', subfolder: '', type: 'output' }],
+      b_images: [{ filename: 'after.png', subfolder: '', type: 'output' }]
+    } as unknown as Partial<NodeExecutionOutput>)
+
+    const result = flattenNodeOutput(['10', output])
+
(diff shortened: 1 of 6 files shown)


christian-byrne (commented):
This is really just emblematic of a larger problem, not specific to app mode. The solution here is not the greatest then, as it could be solved by changing how outputs/results are modeled in the data and processed generally.

christian-byrne:
Closing in favor of #9622, which is a strict superset of this PR. Both PRs fix the same root cause (hardcoded output key allowlist → dynamic `Object.entries()` + `isResultItemLike` validation in the parsing layer), which resolves both the ImageCompare non-standard keys bug (`a_images`/`b_images`) and the 3D asset disappearing bug. #9622 additionally includes the Preview3d.vue `onUnmounted` cleanup fix and an ADR documenting the `.passthrough()` schema design decision.

PR #9642, opened 2026-03-09
fix: virtual scroll pagination not working in media asset list view

viva-jinyi:
## Summary

Fix virtual scroll pagination not triggering in media asset panel list view.

## Changes

- **What**: `VirtualGrid` in `AssetsSidebarListView` was missing `maxColumns=1` and had an incorrect default item height (200px vs actual ~48px). Without `maxColumns`, `cols` was calculated as `floor(containerWidth / 200)` (e.g. 2), causing the row count to be halved and `isNearEnd` to never fire correctly. Added `:max-columns="1"` and `:default-item-height="48"` to fix pagination. Added regression tests to `VirtualGrid.test.ts`.

## Review Focus

- The root cause: `VirtualGrid.cols` computed as `floor(width/200)` instead of `1` for single-column list layout, breaking spacer heights and `approach-end` detection.
- Test covers both the fix (approach-end fires with maxColumns=1) and the bug reproduction (does not fire without it).

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9642-fix-virtual-scroll-pagination-not-working-in-media-asset-list-view-31e6d73d365081a4863cf7bd1405b80b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/common/VirtualGrid.test.ts
+++ b/src/components/common/VirtualGrid.test.ts
@@ -155,6 +155,93 @@ describe('VirtualGrid', () => {
     wrapper.unmount()
   })
 
+  it('emits approach-end for single-column list when scrolled near bottom', async () => {
+    const items = createItems(50)
+    mockedWidth.value = 400
+    mockedHeight.value = 600
+    mockedScrollY.value = 0
+
+    const wrapper = mount(VirtualGrid<TestItem>, {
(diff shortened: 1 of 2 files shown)


viva-jinyi:
Closing fork PR. Reopening from Comfy-Org branch directly.

PR #9645, opened 2026-03-09
feat: replace PrimeVue AutoComplete with SearchAutocomplete in ManagerDialog

viva-jinyi:
## Summary

Replace legacy PrimeVue `AutoCompletePlus` with a new `SearchAutocomplete` component built on Reka UI, matching the `SearchInput` design system.

## Changes

- **What**: Add `SearchAutocomplete` component extending `SearchInput` with dropdown suggestions, IME composition handling, and generic typed `optionLabel` support. Replace `AutoCompletePlus` usage in `ManagerDialog`.
- **Dependencies**: None (uses existing Reka UI Combobox primitives)

## Review Focus

- `SearchAutocomplete` feature parity with the replaced `AutoCompletePlus` (suggestions, option selection, IME handling)
- Dropdown styling and positioning via Reka UI `ComboboxContent`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9645-feat-replace-PrimeVue-AutoComplete-with-SearchAutocomplete-in-ManagerDialog-31e6d73d36508117ba0bef3d30dd0863) by [Unito](https://www.unito.io)


DrJKL:
@coderabbitai review full

viva-jinyi:
@DrJKL Can you review it again?

src/components/ui/search-input/SearchAutocomplete.vue
```diff
@@ -0,0 +1,222 @@
...
+  }
+)
+
+watchDebounced(
+  modelValue,
+  (value: string) => {
+    emit('search', value)
+  },
+  { debounce: debounceTime }
+)
```
DrJKL: This goes back to my "You probably don't need a watch"
https://vueuse.org/shared/refdebounced/

src/components/ui/search-input/SearchAutocomplete.vue
```diff
@@ -0,0 +1,222 @@
...
+  if (isComposing.value) {
+    e.preventDefault()
+    e.stopPropagation()
+  }
+}
+
+watch(
+  () => suggestions,
+  (items) => {
+    isOpen.value = items.length > 0 && !!modelValue.value
```
DrJKL, line 210: Can isOpen be a computed?

src/components/ui/search-input/SearchAutocomplete.vue
```diff
@@ -0,0 +1,222 @@
...
+)
+
+function clearSearch() {
+  modelValue.value = ''
+  focus()
+}
+
+function suggestionLabel(item: T): string {
+  if (optionLabel && typeof item === 'object' && item !== null) {
+    return String((item as Record<string, unknown>)[optionLabel])
```
DrJKL: Is item missing a constraint? You might be able to avoid the type assertion.

src/components/ui/search-input/SearchAutocomplete.vue
```diff
@@ -0,0 +1,222 @@
+<template>
+  <ComboboxRoot
+    v-model="modelValue"
+    v-model:open="isOpen"
+    :ignore-filter="true"
+    :disabled="disabled"
```
DrJKL:
```suggestion
    ignore-filter
    :disabled
```

DrJKL (dismissed):

src/components/ui/search-input/SearchAutocomplete.vue
viva-jinyi: Applied. Changed to `ignore-filter` (bare boolean) and `:disabled` (Vue 3.4+ same-name shorthand).

src/components/ui/search-input/SearchAutocomplete.vue
viva-jinyi: Good catch. Extracted a shared `getItemProperty(item, key)` helper that uses `item[key]` directly (since `key` is `keyof T & string`), eliminating the `as Record<string, unknown>` assertion.

src/components/ui/search-input/SearchAutocomplete.vue
viva-jinyi: `isOpen` cannot be a `computed` because it is used with `v-model:open` on `ComboboxRoot` (which requires a writable ref) and is also manually set to `false` in `onSelectSuggestion` after a selection is made. A computed would not support either of these writes.

src/components/ui/search-input/SearchAutocomplete.vue
viva-jinyi: Applied. The only consumer (ManagerDialog) does not use `@search` — debouncing is already handled in `useRegistrySearch` via `watchDebounced(searchQuery, ...)`. Removed `search` emit, `debounceTime` prop, and all associated watch/debounce logic from the component entirely.

viva-jinyi:
@DrJKL Can you approve again?

DrJKL (approved):

PR #9649, opened 2026-03-09
perf: detect pointer drag in useTransformSettling for pan optimization

LittleSound:
## Summary

When many nodes are rendered in the transform container, both zoom and pan can cause FPS drops because the browser re-rasterizes all visible content at the new transform. `will-change: transform` tells the browser to keep the layer as a GPU texture and skip re-rasterization during active interaction, restoring visual quality only after settling.

- Add pointer drag detection so `will-change: transform` covers pan in addition to zoom. Without this, dragging with 256+ nodes causes jank as the browser re-rasterizes the entire layer on every frame of the pan.
- Fix settleDelay from 16ms to 256ms. At 16ms the debounce fires between consecutive wheel events (~50ms apart on a physical mouse), causing `will-change` to toggle on/off rapidly. Each toggle forces the browser to promote/demote the compositor layer, which is more expensive than not having the optimization at all.
- Replace scoped CSS with Tailwind `will-change-transform`.
- Remove per-node `will-change: transform` on `.lg-node`. Promoting each node to its own compositor layer (256 nodes = 256 GPU textures) increases memory pressure and compositing overhead, making performance worse than a single promoted container.
- Previously, the virtual DOM of Nodes was updated during zooming and dragging, but now this update is avoided through some techniques.
- Using the 3D versions of scale and translate can provide a smoother experience when dealing with a large number of nodes.

## Test plan
- [x] Manual: verify during both zoom and pan
- [x] Manual: compare pan FPS with 256 nodes before/after

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9649-perf-detect-pointer-drag-in-useTransformSettling-for-pan-optimization-31e6d73d3650818bb2c3ccd01a465140) by [Unito](https://www.unito.io)


benceruleanlu (commented):
Ctrl+Shift drag-zoom returns before LiteGraph calls pointer.down(e), so the canvas never captures the pointer. Since usePointerDrag() still sets isPointerDown on the initial pointerdown and waits for pointerup/pointercancel on the same target to clear it, releasing off-canvas can leave the state stuck.

benceruleanlu (dismissed):
Wow this is great, I actually can feel it on my mac!

minor nit, can you make sure that left click node selection in standard navigation mode also does not trigger this as it does not pan?

LittleSound:
> minor nit, can you make sure that left click node selection in standard navigation mode also does not trigger this as it does not pan?


I didn't encounter this problem in my test.

src/renderer/core/layout/transform/useTransformSettling.ts
```diff
@@ -50,35 +50,70 @@ export function useTransformSettling(
...
+  target: MaybeRefOrGetter<HTMLElement | null | undefined>,
+  onDrag: () => void,
+  eventOptions: AddEventListenerOptions
+) {
+  const isPointerDown = ref(false)
+
+  useEventListener(
+    target,
+    'pointerdown',
+    () => {
```
christian-byrne: suggestion (non-blocking): `pointerdown` fires for every mouse button. Right-clicking (context menu) or middle-clicking on the canvas will set `isPointerDown = true`, and the next `pointermove` triggers `markInteracting()` -- promoting a compositor layer for 256ms unnecessarily.

Consider filtering by the primary button:
```ts
useEventListener(
  target,
  'pointerdown',
  (e: PointerEvent) => {
    if (e.button === 0) isPointerDown.value = true
  },
  eventOptions
)
```

src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -33,7 +27,32 @@ const { transformStyle, syncWithCanvas } = useTransformState()
...
+ * These values change every animation frame during zoom or pan.
+ * If they were bound in the template, Vue would diff the entire
+ * TransformPane vnode—including all child node slots—on every frame,
+ * causing expensive vdom patch work across the full node list.
+ * Mutating the DOM directly limits the update to a single element.
+ */
+
+watch([transformStyle, transformPaneRef], ([newStyle, el]) => {
+  if (el) {
+    Object.assign(el.style, newStyle)
```
christian-byrne, line 48: nitpick (non-blocking): `Object.assign(el.style, newStyle)` only sets properties, it never removes old ones. Right now `transformStyle` always returns the same two keys so there's no bug, but if the computed shape ever changes the stale values will persist. Explicit assignment (`el.style.transform = ...`) makes the contract clearer.

src/renderer/core/layout/transform/useTransformSettling.ts
```diff
@@ -50,35 +50,70 @@ export function useTransformSettling(
...
+
+/**
+ * Calls `onDrag` on each pointermove while a pointer is held down.
+ */
+function usePointerDrag(
+  target: MaybeRefOrGetter<HTMLElement | null | undefined>,
+  onDrag: () => void,
+  eventOptions: AddEventListenerOptions
+) {
+  const isPointerDown = ref(false)
```
christian-byrne: nitpick (non-blocking): `isPointerDown` is a single boolean, so with two touches finger-A-up clears the flag even while finger-B is still down. The debounce self-corrects on the next `pointermove`, and multi-touch on desktop canvas apps is uncommon, so this is low-impact. A `pointerCount` ref (increment on `pointerdown`, decrement on `pointerup`/`pointercancel`) would be more correct if you ever need it.

src/renderer/core/layout/transform/TransformPane.vue
LittleSound: To remove styles, undefined should be assigned. the current approach is more reasonable and was written after considering future scalability.

christian-byrne (approved):
LGTM

PR #9656, opened 2026-03-09
fix: Upgrade reka-ui and replace PrimeVue Splitters with pixel sizing support (#9508)

Myestery:
Closes #9508

## Summary

PrimeVue Splitter components used percentage-based sizing, which caused panel sizes to jump/reset when toggling sidebars, switching tabs, or changing panel visibility. This was because PrimeVue's `initializePanels()` recalculated percentage-based flex values on re-render, losing the user's intended pixel sizes.

Replaced PrimeVue `Splitter`/`SplitterPanel` with reka-ui's `SplitterGroup`/`SplitterPanel`/`SplitterResizeHandle`, which support persistent layout via `auto-save-id`.

## Changes

- **pnpm-workspace.yaml** / **pnpm-lock.yaml**: Bumped reka-ui from ^2.5.0 to ^2.9.0
- **src/components/LiteGraphCanvasSplitterOverlay.vue**: Replaced PrimeVue `Splitter`/`SplitterPanel` with reka-ui `SplitterGroup`/`SplitterPanel`/`SplitterResizeHandle`. Added explicit `order` and `default-size` props, moved resize handle visibility into conditional `SplitterResizeHandle` components, and updated scoped styles to target reka-ui's `data-state` attributes
- **src/views/LinearView.vue**: Same migration from PrimeVue to reka-ui splitter components. Removed `useStablePrimeVueSplitterSizer` workaround (no longer needed). Added `auto-save-id` for persistent layout state
- **src/composables/useStablePrimeVueSplitterSizer.ts**: Deleted — this workaround for PrimeVue's percentage-based sizing is no longer needed with reka-ui
- **src/composables/useStablePrimeVueSplitterSizer.test.ts**: Deleted — tests for removed composable

### Follow-up fixes (CodeRabbit review)

- Stabilized `auto-save-id` to use `splitterRefreshKey` (encodes sidebar location, panel visibility, builder mode) so resized widths persist correctly across page refreshes
- Included `sidebarLocation` in LinearView `auto-save-id` to prevent saved widths from being applied to reversed panel positions when sidebar switches sides
- Added `w-1` to all `SplitterResizeHandle` elements so drag handles are visible and easier to grab
- Reduced `CENTER_PANEL_SIZE` from 80% to 60% so 3-panel layouts (left + center + right) sum to 100% without normalization


pythongosssss:
- still uses % sizing instead of pixels, open both panels, resize one, other also changes size
- resize handles area is visually larger than the actual handle (hover like 5px to the left of the bar, it'll show the resize arrow, cursor but you cant resize there)
- looks like you can resize to < min size, then resizing offset is incorrect (resize right panel smaller, move cursor all the way to window edge, then drag cursor to the left making panel larger, the panel grows faster than your mouse moves)
- e2e tests failing as it looks like some right space spacing has gone
- needs new tests for px scaling behavior

PR #9657, opened 2026-03-09
fix: hide empty actionbar container and relocate error border to floating actionbar

jaeone94:
## Summary
When the actionbar is floating and has no docked buttons, the container is now hidden (zero-width, transparent border) to avoid showing an empty rounded box. Additionally, the error/destructive border is now applied to the floating actionbar panel itself (via `ComfyActionbar`) instead of the container, so it appears in the correct location when floating.

## Changes
- **TopMenuSection**: Added `hasDockedButtons` and `isActionbarContainerEmpty` computed properties to detect when the docked container has no visible buttons; `actionbarContainerClass` computed hides the container by collapsing it when empty and floating, while preserving the legacy drop zone via `:has(.border-dashed)` CSS selector
- **TopMenuSection**: Error border (`border-destructive-background-hover`) is now only applied to the docked container when the actionbar is **not** floating
- **ComfyActionbar**: Accepts new `hasAnyError` prop and applies the error border to the floating panel's `panelClass` when floating

## Review Focus
- The `has-[.border-dashed]` CSS selector restores the container visuals when a legacy drag-target element is present inside it — verify this works as expected
- Error border placement: docked mode shows border on container, floating mode shows border on the fixed panel

## Screenshots

https://github.com/user-attachments/assets/75caabac-e391-4bfd-b4dc-62d564e55d37

### DIFF
--- a/src/components/TopMenuSection.vue
+++ b/src/components/TopMenuSection.vue
@@ -114,7 +114,7 @@
 </template>
 
 <script setup lang="ts">
-import { useLocalStorage } from '@vueuse/core'
+import { useLocalStorage, useMutationObserver } from '@vueuse/core'
 import { storeToRefs } from 'pinia'
 import { computed, onMounted, ref } from 'vue'
 import { useI18n } from 'vue-i18n'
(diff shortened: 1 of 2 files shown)


viva-jinyi (approved):
LGTM

PR #9671, opened 2026-03-09
Restore hiding of linked inputs in app mode

AustinMroz:
As a temporary fix for widgets being incorrectly hidden, #9669 allowed all disabled widgets to be displayed.

This PR provides a more robust implementation to derive whether the widget, as would be displayed from the root graph, is disabled.

Potential regression:
- Drag drop handlers are applied on node, not widgets. A subgraph containing a "Load Image" node, does not allow dragging and dropping an image onto the subgraphNode in order to load it. Because app mode widgets would display from the original owning node prior to this PR, these drag/drop handlers would apply. Placing "Load Image" nodes. I believe this change makes behavior more consistent, but it warrants consideration.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9671-Restore-hiding-of-linked-inputs-in-app-mode-31e6d73d365081688e37fbb931f3af68) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/linearMode/LinearControls.vue
+++ b/src/renderer/extensions/linearMode/LinearControls.vue
@@ -73,7 +73,6 @@ const mappedSelections = computed(() => {
   const processedInputs: ReturnType<typeof nodeToNodeData>[] = []
   while (unprocessedInputs.length) {
     const [node] = unprocessedInputs[0]
-    if (!node) continue
     const inputGroup = takeWhile(unprocessedInputs, ([n]) => n === node).map(
       ([, widget]) => widget
     )
@@ -92,7 +91,7 @@ const mappedSelections = computed(() => {
       return !inputGroup.some(
         (subWidget) =>
           isPromotedWidgetView(subWidget) &&
-          subWidget.sourceNodeId === storeNodeId &&
+          subWidget.sourceNodeId == storeNodeId &&
           subWidget.sourceWidgetName === vueWidget.storeName
       )
     })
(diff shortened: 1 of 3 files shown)


src/renderer/extensions/linearMode/LinearControls.vue
```diff
@@ -63,21 +64,45 @@ useEventListener(
...
     unprocessedInputs = unprocessedInputs.slice(inputGroup.length)
-    const node = resolveNode(nodeId)
+    //FIXME: hide widget if owning node bypassed
     if (node?.mode !== LGraphEventMode.ALWAYS) continue
 
     const nodeData = nodeToNodeData(node)
-    remove(nodeData.widgets ?? [], (w) => !inputGroup.includes(w.name))
+    remove(nodeData.widgets ?? [], (w) => !!w.slotMetadata?.linked)
+    if (node.isSubgraphNode()) {
+      remove(nodeData.widgets ?? [], (vueWidget) => {
```
DrJKL: Non-blocking, it does feel like there's a common filter function that could be extracted to avoid the branches here.

DrJKL (dismissed):

src/renderer/extensions/linearMode/LinearControls.vue
AustinMroz: Was conflicted on this as well. With hindsight, it's definitely cleaner. :+1:

DrJKL (approved):

PR #9678, opened 2026-03-09
Use preview downscaling in fewer places

AustinMroz:
Thumbnail downscaling is currently being used in more places than it should be. 
- Nodes which display images will display incorrect resolution indicators
  image image

  This is particularly confusing with output nodes, which claim the output is not of the intended resolution
- The "Download Image" and "Open Image" context menu actions will incorrectly download the downscaled thumbnail.
- The assets panel will incorrectly display the thumbnail resolution as the resolution of the output
- The lightbox (zoom) of an image will incorrectly display a downscaled thumbnail.

This PR is a quick workaround to staunch the major problems
- Nodes always display full previews.
- Resolution downscaling is applied on the assert card, not on the assetItem itself
  - Due to implementation, this means that asset cards will still incorrectly show the resolution of the thumbnail instead of the size of the full image.


src/platform/distribution/cloudPreviewUtil.ts
```diff
@@ -17,3 +17,8 @@ export function appendCloudResParam(
   if (filename && getMediaTypeFromFilename(filename) !== 'image') return
   params.set('res', '512')
 }
+
+export function appendCloudResUrl(url: string): string {
+  if (!isCloud) return url
+  return url + '&res=512'
```
DrJKL: The URLSearchParams piece is a guard for when the asset already has a `res` set in the URL sent by the server

AustinMroz: Appreciate the catch. I had hoped to make this a smaller PR than reworking the assetsSchema to pass both.

Will start on the nicer fix.

DrJKL (approved):

PR #9679, opened 2026-03-09
fix: keep job details popover on-screen in sidebar

benceruleanlu:
## Summary

Keep the restored job details popover visible when the job history sidebar is docked on the left edge of the workspace.

## Changes

- **What**: Replace the fixed `right`-based hover popover positioning with viewport-aware `left` positioning so the popover opens on the side with available space, reuse that logic in both `JobAssetsList` and `QueueJobItem`, and add coverage for left-edge/right-edge placement plus the job history sidebar integration.

## Review Focus

Please verify the hover popover opens on-screen for left-docked job history, and that queue overlay / legacy queue row behavior still matches the intended hover handoff.

N/A

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9679-fix-keep-job-details-popover-on-screen-in-sidebar-31e6d73d3650816d9e7ffb0749430218) by [Unito](https://www.unito.io)


DrJKL:
Should we switch to using [a full popover component](https://shadcn-vue.com/docs/components/popover)?

benceruleanlu:
> Should we switch to using [a full popover component](https://shadcn-vue.com/docs/components/popover)?

Yes, but the correct shared primitive here is HoverCard rather than our click Popover; I can refactor toward that if you want this PR to take on the extra scope.

christian-byrne (commented):
Can we add a test?

christian-byrne:
added unit tests for `getHoverPopoverPosition` in 220218ee6 - 7 test cases covering boundary clamping, position offsets, and edge cases. all 20 tests pass across the 3 test files.

christian-byrne (approved):
merge conflict resolved, all 21 tests pass across 3 files, CI green. lgtm

PR #9680, opened 2026-03-09
fix: simplify ensureCorrectLayoutScale and fix link sync during Vue node drag

DrJKL:
## Summary

Fix node layout drift from repeated `ensureCorrectLayoutScale` scaling, simplify it to a pure one-time normalizer, and fix links not following Vue nodes during drag.

## Changes

- **What**:
  - `ensureCorrectLayoutScale` simplified to a one-time normalizer: unprojects legacy Vue-scaled coordinates back to canonical LiteGraph coordinates, marks the graph as corrected, and does nothing else. No longer touches the layout store, syncs reroutes, or changes canvas scale.
  - Removed no-op calls from `useVueNodeLifecycle.ts` (a renderer version string was passed where an `LGraph` was expected).
  - `layoutStore.finalizeOperation` now calls `notifyChange` synchronously instead of via `setTimeout`. This ensures `useLayoutSync`'s `onChange` callback pushes positions to LiteGraph `node.pos` and calls `canvas.setDirty()` within the same RAF frame as a drag update, fixing links not following Vue nodes during drag.
- **Tests**: Added tests for `ensureCorrectLayoutScale` (idempotency, round-trip, unknown-renderer no-op) and `graphRenderTransform` (project/unproject round-trips, anchor caching).

## Review Focus

- The `setTimeout(() => this.notifyChange(change), 0)` → `this.notifyChange(change)` change in `layoutStore.ts` is the key fix for the drag-link-sync bug. The listener (`useLayoutSync`) only writes to LiteGraph, not back to the layout store, so synchronous notification is safe.
- `ensureCorrectLayoutScale` no longer has any side effects beyond normalizing coordinates and setting `workflowRendererVersion` metadata.


src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -917,8 +917,9 @@ class LayoutStoreImpl implements LayoutStore {
       }
     })
 
-    // Notify listeners (after transaction completes)
-    setTimeout(() => this.notifyChange(change), 0)
+    // Notify listeners synchronously so Layout→LiteGraph sync
+    // (useLayoutSync) pushes positions within the same frame.
+    this.notifyChange(change)
```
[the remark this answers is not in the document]
DrJKL: The `change` object's `source` is set at construction (line 840: `source: operation.source`), and `notifyChange` passes that immutable `LayoutChange` to listeners — no listener reads `this.currentSource`. The code already does the second suggested approach (immutable context object). False positive.

---

_A rabbit reviewed my source,_
_And flagged a "mutable" course._
_But `change.source` is sealed,_
_No shared state revealed —_
_The bunny was chasing a horse._ 🐇

DrJKL:
@coderabbitai review

src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
```diff
@@ -126,8 +156,35 @@ const resizeObserver = new ResizeObserver((entries) => {
...
+      elementsNeedingFreshMeasurement.has(element)
+    const hasMatchingCachedNodeMeasurement =
+      previousMeasurement != null &&
+      previousMeasurement.nodeId === nodeId &&
+      nodeLayout != null &&
+      isBoundsEqual(previousMeasurement.bounds, nodeLayout.bounds)
+
+    // ResizeObserver emits entries where nothing changed (e.g. initial observe).
+    // Skip expensive DOM reads when this exact element/node already measured at
+    // the same normalized bounds and size.
```
AustinMroz, line 176: I've tried multiple browsers, but I'm unable to replicate a scenario where entries which have not actually undergone a resize are included.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.test.ts
AustinMroz, line 1: Needing to make changes to formdropdown tests as a result of node resize code seems really out of scope.

AustinMroz (dismissed):
I'm not seeing any glaring issues from reading over the code.

As mentioned offline, I'm still able to trivially produce errors in node sizing by reloading while the browser tab is not in the foreground. This seems to be a browser specific issue as I can not reproduce in Firefox. Of note useDocumentVisibility reports that the document is visible during resizeObserver trigger.
image

I am specifically concerned that a 'code simplification PR' is adding ~6x the code it's removing.

---

In terms of a future ideal, is there ever a time that nodes should resize or move position without either
1. Code that mutates the position of the node
2. A user releasing a drag handle

Is having a resizeObserver at all actually needed?

browser_tests/tests/rendererToggleStability.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+  'Renderer toggle stability',
+  { tag: ['@node', '@canvas'] },
+  () => {
+    test('node positions do not drift when toggling between Vue and LiteGraph renderers', async ({
+      comfyPage
+    }) => {
+      const TOGGLE_COUNT = 5
+
+      const initialPositions = await getAllNodePositions(comfyPage)
+      expect(initialPositions.length).toBeGreaterThan(0)
```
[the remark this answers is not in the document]
DrJKL: Default workflow is fine for now.

browser_tests/tests/rendererToggleStability.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+        const afterVue = await getAllNodePositions(comfyPage)
+
+        for (const initial of initialPositions) {
+          const current = afterVue.find((n) => n.id === initial.id)
+          expect(
+            current,
+            `node ${initial.id} missing after Vue toggle ${i + 1}`
+          ).toBeDefined()
+          expect(current!.x).toBeCloseTo(initial.x, 1)
+          expect(current!.y).toBeCloseTo(initial.y, 1)
```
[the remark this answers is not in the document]
DrJKL: We want to be strict. We're hardcore like that.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.test.ts
DrJKL: 🤷🏻

src/renderer/core/layout/transform/graphRenderTransform.ts
```diff
@@ -0,0 +1,45 @@
+import type { LGraph } from '@/lib/litegraph/src/LGraph'
+import { createBounds } from '@/lib/litegraph/src/measure'
+import type { Bounds, Point } from '@/renderer/core/layout/types'
+
+export const RENDER_SCALE_FACTOR = 1.2
+
+export const MIN_NODE_WIDTH = 225
+
+const anchorCache = new WeakMap<LGraph, Point>()
```
christian-byrne, line 9: suggestion: `anchorCache` is compute-once per `LGraph` reference and never invalidates. That's correct for the current one-shot normalizer, but `getGraphRenderAnchor` is a public export -- a future caller adding/moving nodes and re-calling it would silently get a stale anchor.

Consider adding a `@remarks` JSDoc noting the compute-once invariant, or a `clearAnchorCache(graph)` escape hatch.

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -986,6 +1013,8 @@ class LayoutStoreImpl implements LayoutStore {
       this.linkSegmentLayouts.clear()
       this.slotLayouts.clear()
       this.rerouteLayouts.clear()
+      this.pendingGlobalChanges = []
```
christian-byrne, line 1017: suggestion: `initializeFromLiteGraph` resets ynodes, spatial indexes, slot/link/reroute layouts, `pendingGlobalChanges`, and `isGlobalDispatchQueued`, but `nodeChangeListeners` is not cleared. If a graph is re-initialized while Vue components still hold subscriptions from the previous graph (before `onUnmounted` fires), stale listener callbacks could fire against new node IDs that collide with old subscriptions.

`nodeRefs`/`nodeTriggers` are intentionally retained (comment on line 1004), but `nodeChangeListeners` callbacks capture component-scoped state and should be cleared:
```ts
this.nodeChangeListeners.clear()
```

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
```diff
@@ -1,165 +1,104 @@
...
-      const scaledX = originX + relativeX * scaleFactor
-      const scaledY = originY + relativeY * scaleFactor
-
-      const scaledWidth = oldWidth * scaleFactor
-      const scaledHeight = oldHeight * scaleFactor
-
-      ioNode.pos = [scaledX, scaledY]
-      ioNode.size = [scaledWidth, scaledHeight]
+    ]) {
+      unprojectPosSize(ioNode, anchor)
```
christian-byrne: suggestion (non-blocking): `'inputNode' in graph` can be true while `graph.inputNode` is `null` (e.g., freshly created subgraph). The cast + `unprojectPosSize` would throw accessing `.pos` on null. Consider guarding:

```suggestion
      if (ioNode) unprojectPosSize(ioNode, anchor)
```

browser_tests/tests/rendererToggleStability.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { ComfyPage } from '../fixtures/ComfyPage'
+import type { Position } from '../fixtures/types'
+
+type NodeSnapshot = { id: number } & Position
+
+async function getAllNodePositions(
+  comfyPage: ComfyPage
+): Promise<NodeSnapshot[]> {
+  return comfyPage.page.evaluate(() =>
+    window.app!.graph.nodes.map((n) => ({
```
christian-byrne, line 14: suggestion (non-blocking): `getAllNodePositions` captures `pos` but not `size`. The drift bug this test guards against could also affect node dimensions via the old scale factor. Adding `width: n.size[0], height: n.size[1]` to the snapshot (and asserting on them) would strengthen the regression gate.

src/renderer/extensions/vueNodes/layout/useNodeDrag.test.ts
```diff
@@ -0,0 +1,173 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
```
christian-byrne, line 1: suggestion (non-blocking): `endDrag` has significant cleanup logic (cancelling RAF, snap-to-grid, clearing drag state) and is not covered by any test here. A broken `endDrag` would leak `rafId` or skip snap. Consider adding a test that calls `startDrag` -> `handleDrag` -> `endDrag` and asserts RAF is cancelled and final position reflects snap.

src/renderer/core/layout/transform/graphRenderTransform.ts
DrJKL: Implemented in src/renderer/core/layout/transform/graphRenderTransform.ts: added `@remarks` on `getGraphRenderAnchor` clarifying it’s compute-once per graph instance by design. The cache still doesn’t invalidate itself, but now it at least introduces itself before surprising anyone.

src/renderer/core/layout/store/layoutStore.ts
DrJKL: Implemented in src/renderer/core/layout/store/layoutStore.ts: `initializeFromLiteGraph` now clears `nodeChangeListeners`. Old listeners were attending the new graph uninvited, so now they’re escorted out at initialization.

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
DrJKL: Implemented in src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts: added a null guard before calling `unprojectPosSize` on subgraph IO nodes. If an IO node is null, we now let it continue being null in peace.

browser_tests/tests/rendererToggleStability.spec.ts
DrJKL: Implemented in browser_tests/tests/rendererToggleStability.spec.ts: snapshots/assertions now include `width` and `height` alongside `x/y`. Position drift had company; now size drift does too.

src/renderer/extensions/vueNodes/layout/useNodeDrag.test.ts
DrJKL: Implemented in src/renderer/extensions/vueNodes/layout/useNodeDrag.test.ts: added `startDrag -> handleDrag -> endDrag` coverage verifying RAF cancellation and snap finalization via `batchUpdateNodeBounds`. Cleanup now has a witness.

src/renderer/core/layout/sync/useLayoutSync.ts
```diff
@@ -26,45 +105,23 @@ export function useLayoutSync() {
...
-          liteNode.onResize?.(liteNode.size)
-        }
-      }
+      // Topology-only changes (links, reroutes) don't need LiteGraph
+      // node writeback — link rendering reads from the store directly.
+      if (change.nodeIds.length === 0) return
 
-      // Trigger single redraw for all changes
-      canvas.setDirty(true, true)
+      change.nodeIds.forEach((nodeId) => pendingNodeIds.add(nodeId))
```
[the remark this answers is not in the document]
DrJKL: Are we using biome in this project?

src/renderer/core/layout/transform/graphRenderTransform.ts
christian-byrne: fixed in afd9445 - @remarks added documenting the compute-once invariant

src/renderer/core/layout/store/layoutStore.ts
christian-byrne: fixed in afd9445 - nodeChangeListeners.clear() added on line 1008

src/renderer/extensions/vueNodes/layout/ensureCorrectLayoutScale.ts
christian-byrne: fixed in afd9445 - null guard added before unprojectPosSize call

src/renderer/extensions/vueNodes/layout/useNodeDrag.test.ts
christian-byrne: fixed in afd9445 - endDrag test added covering RAF cancellation and snap finalization

browser_tests/tests/rendererToggleStability.spec.ts
christian-byrne: note: size assertions were added in afd9445 but removed in d157906 to stabilize the test - position-only is the right scope for now since this test guards the drift bug specifically

src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts
christian-byrne: note (non-blocking): ResizeObserver does fire on initial observe per the spec (the first callback always includes the observed element even if no resize occurred). The guard here is a perf optimization to skip redundant DOM reads in that case. Harmless if it never triggers in practice, but it does match the spec behavior.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.test.ts
christian-byrne: note (non-blocking): the mock simplification (defineComponent -> plain objects) was needed because the resize tracking changes affected how component mounting works in tests. The changes are test-only and add a missing coverage case. Agreed it is scope-adjacent - happy to split out if you feel strongly, but it should not block merge.

christian-byrne (approved):
All blocking feedback addressed and verified in code. Two non-blocking threads from AustinMroz remain open (ResizeObserver guard justification, FormDropdown test scope) - replied with rationale, neither blocks merge.

PR #9686, opened 2026-03-10
Mobile input tweaks

AustinMroz:
- Buttons are marked as `touch-manipulation` so double-tapping on them doesn't initiate a zoom.
- Move scrubable inputs to usePointerSwipe
  - Strangely, swipe direction was inverted on mobile. This solves the issue and simplifies code
  - Moves event handlers into the scrubbable input component
- Make the slightly bigger buttons only apply when on mobile.
- Updates the workflows dropdown to have a check by the activeWorkflow and truncate workflow names
- Displays dropzones (for the image preview) on mobile, but disables the prompt to drag and drop an image if none is selected.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9686-Mobile-input-tweaks-31f6d73d3650811d9025d0cd1ac58534) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/common/DropdownItem.vue
+++ b/src/components/common/DropdownItem.vue
@@ -54,11 +54,12 @@ defineProps<{ itemClass: string; contentClass: string; item: MenuItem }>()
     :disabled="toValue(item.disabled) ?? !item.command"
     @select="item.command?.({ originalEvent: $event, item })"
   >
-    <i class="size-5" :class="item.icon" />
-    {{ item.label }}
+    <i class="size-5 shrink-0" :class="item.icon" />
+    <div class="mr-auto truncate" v-text="item.label" />
+    <i v-if="item.checked" class="icon-[lucide--check] shrink-0" />
     <div
-      v-if="item.new"
-      class="ml-auto flex items-center rounded-full bg-primary-background px-1 text-xxs leading-none font-bold"
+      v-else-if="item.new"
+      class="flex shrink-0 items-center rounded-full bg-primary-background px-1 text-xxs leading-none font-bold"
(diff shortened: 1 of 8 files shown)


src/components/common/DropdownItem.vue
```diff
@@ -54,11 +54,12 @@ defineProps<{ itemClass: string; contentClass: string; item: MenuItem }>()
     :disabled="toValue(item.disabled) ?? !item.command"
     @select="item.command?.({ originalEvent: $event, item })"
   >
-    <i class="size-5" :class="item.icon" />
-    {{ item.label }}
+    <i class="size-5 shrink-0" :class="item.icon" />
+    <div class="mr-auto truncate" v-text="item.label" />
```
DrJKL, line 58: Uncommon pattern 😄

src/locales/en/main.json
```diff
@@ -3181,6 +3181,7 @@
     "deleteAllAssets": "Delete all assets from this run",
     "hasCreditCost": "Requires additional credits",
     "viewGraph": "View node graph",
+    "mobileNoWorkflow": "This workflow hasn't been built for app mode. Try a different one.",
```
DrJKL, line 3184: Will this be a link?

DrJKL (approved):

src/locales/en/main.json
AustinMroz: Wasn't planning on it.

Since building an app on mobile will be absolutely awful, I want to lean towards not allowing entry to app mode (/ or at least that tab) for workflows that aren't setup. But preventing access like that feels like it'd be a little too risky a change for the current weeik.

PR #9697, opened 2026-03-10
fix: detect missing nodes when registry API fails to resolve packs

viva-jinyi:
## Summary

Nodes without `cnr_id` metadata that fail registry lookup (404) were silently dropped, causing the Missing Nodes tab to show "No Missing Nodes" incorrectly.

## Changes

- **What**: Track nodes that are missing locally AND unresolvable via registry as "unresolved nodes". Show them in a dedicated Unresolved Nodes tab (only visible when such nodes exist).
- `useWorkflowPacks`: Collect `unresolvedNodeNames` — nodes where `inferPack` returns null and `nodeDefStore` has no definition
- `useMissingNodes`: Include `unresolvedNodeNames` in `hasMissingNodes` check
- `ManagerDialog`: Add conditional Unresolved Nodes nav item with badge count
- `useManagerDisplayPacks`: Handle `ManagerTab.Unresolved` (returns empty array, no pack cards)
- New `UnresolvedNodesMessage.vue` component for the tab content

## Review Focus

- The boundary between "unresolved" (registry 404 + not in nodeDefStore) vs installed nodes that simply aren't in the registry
- Whether the Unresolved tab UX is appropriate (no actionable install button, informational only)

[screen-capture.webm](https://github.com/user-attachments/assets/50d037f5-c828-4198-9d0b-ecc22f33fc90)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9697-fix-detect-missing-nodes-when-registry-API-fails-to-resolve-packs-31f6d73d365081ba9af8f3f7ee331b1a) by [Unito](https://www.unito.io)


src/workbench/extensions/manager/composables/nodePack/useWorkflowPacks.ts
```diff
@@ -111,12 +112,34 @@ const _useWorkflowPacks = () => {
...
+    const resolvedPacks: WorkflowPack[] = []
+    const unresolved: string[] = []
+
+    for (let i = 0; i < packs.length; i++) {
+      const pack = packs[i]
+      if (pack) {
+        resolvedPacks.push(pack)
+      } else {
+        const node = nodes[i]
+        const nodeName = node.type
```
Myestery: is it intentional to use node type as a key here instead of node name?

viva-jinyi: Yes — this follows the same convention as `inferPack` (line 60) in this file, where `node.type` is used as the node's identifier for registry lookups. In LiteGraph, node.type is effectively the node class name (e.g. "KSampler"), which is what the registry API and `nodeDefStore.nodeDefsByName` both use as their key. The variable is named nodeName to match the existing pattern.

src/workbench/extensions/manager/composables/nodePack/useWorkflowPacks.ts
```diff
@@ -111,12 +112,43 @@ const _useWorkflowPacks = () => {
...
+      unresolvedNodeNames.value = []
+      return
+    }
+
+    const nodes = mapAllNodes(app.rootGraph, (node) => node)
+    const packPromises = nodes.map((node) => workflowNodeToPack(node))
     const packs = await Promise.all(packPromises)
-    workflowPacks.value = packs.filter((pack) => pack !== undefined)
+
+    const resolvedPacks: WorkflowPack[] = []
```
AustinMroz, line 125: Nit: Function looks logically correct, but I'm not super happy about the implementation.
I'd probably have all the logic be inside a ` await Promise.all(mapAllNodes(...))` that does the pack lookup and pushes to `resolvedPacks`/`unresolved`

It also seems a little clunky how workflowPacks is stored and utilized. I'd assume it be more performant to collect entries to a set directly, but that looks like it would be out of scope for this PR.

src/workbench/extensions/manager/components/manager/UnresolvedNodesMessage.vue
```diff
@@ -0,0 +1,26 @@
...
+      {{ $t('manager.unresolvedNodes.title') }}
+    </h3>
+    <p class="text-center text-sm text-muted-foreground">
+      {{ $t('manager.unresolvedNodes.message') }}
+    </p>
+    <ul class="mt-2 flex flex-col gap-1">
+      <li
+        v-for="name in nodeNames"
+        :key="name"
+        class="rounded-sm bg-comfy-input px-3 py-1.5 font-mono text-sm"
```
AustinMroz: The styling for the list also appears a little out of place, but design is well outside my area of expertise.

Maybe place a `bg-secondary-background rounded-lg` for the `ul` and no background for the `li`? :thinking:

AustinMroz (dismissed):

src/locales/en/main.json
```diff
@@ -404,6 +405,10 @@
     "noDescription": "No description available",
     "installSelected": "Install Selected",
     "installAllMissingNodes": "Install All",
+    "unresolvedNodes": {
+      "title": "Unresolved Missing Nodes",
+      "message": "The following nodes are not installed and could not be found in the registry."
+    },
```
DrJKL, line 411: We already have a task to clean up the i18n keys' organization, not going to pester the bunny.

src/workbench/extensions/manager/components/manager/ManagerDialog.vue
```diff
@@ -309,7 +315,17 @@ const navItems = computed<(NavItemData | NavGroupData)[]>(() => [
...
+      ...(unresolvedNodeNames.value.length > 0
+        ? [
+            {
+              id: ManagerTab.Unresolved,
+              label: t('manager.nav.unresolvedNodes'),
+              icon: 'icon-[lucide--help-circle]',
+              badge: unresolvedNodeNames.value.length
+            }
+          ]
+        : [])
```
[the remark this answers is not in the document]
viva-jinyi: Already addressed in commit 07feca8 — added a `watch(navItems, ...)` that resets `selectedNavId` to `ManagerTab.Missing` when the current tab no longer exists in navItems (line 356-360).

src/workbench/extensions/manager/composables/nodePack/useWorkflowPacks.ts
```diff
@@ -111,12 +112,34 @@ const _useWorkflowPacks = () => {
...
+   * Nodes that have no local definition and no registry match are tracked
+   * as unresolved so downstream consumers can surface them to the user.
    */
   const getWorkflowPacks = async () => {
     if (!app.rootGraph) return []
-    const packPromises = mapAllNodes(app.rootGraph, workflowNodeToPack)
+
+    const nodes = mapAllNodes(app.rootGraph, (node) => node)
+    const packPromises = nodes.map((node) => workflowNodeToPack(node))
     const packs = await Promise.all(packPromises)
```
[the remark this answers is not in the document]
viva-jinyi: Already addressed in commit 07feca8 — the early return now resets both `workflowPacks.value = []` and `unresolvedNodeNames.value = []` before returning.

src/workbench/extensions/manager/composables/nodePack/useWorkflowPacks.ts
```diff
@@ -111,12 +112,34 @@ const _useWorkflowPacks = () => {
...
+    for (let i = 0; i < packs.length; i++) {
+      const pack = packs[i]
+      if (pack) {
+        resolvedPacks.push(pack)
+      } else {
+        const node = nodes[i]
+        const nodeName = node.type
+        if (nodeName && !nodeDefStore.nodeDefsByName[nodeName]) {
+          unresolved.push(nodeName)
+        }
```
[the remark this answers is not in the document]
viva-jinyi: Already addressed in commit 07feca8 — added `getWorkflowNodePackId(node) === undefined` check so nodes with `cnr_id` or `aux_id` metadata are excluded from the unresolved list.

src/workbench/extensions/manager/composables/nodePack/useWorkflowPacks.ts
viva-jinyi: Good call — refactored in 7a7733e to use a single `await Promise.all(mapAllNodes(..., async (node) => { ... }))` callback that handles both pack resolution and unresolved classification inline, eliminating the separate `nodes` array and index-based correlation.

src/workbench/extensions/manager/components/manager/UnresolvedNodesMessage.vue
viva-jinyi: Applied in 7a7733e — moved the background to `ul` with `bg-secondary-background rounded-lg p-2` and removed the per-item background from `li`.

src/workbench/extensions/manager/components/manager/UnresolvedNodesMessage.vue
```diff
@@ -0,0 +1,26 @@
...
+    <p class="text-center text-sm text-muted-foreground">
+      {{ $t('manager.unresolvedNodes.message') }}
+    </p>
+    <ul class="mt-2 flex flex-col gap-1 rounded-lg bg-secondary-background p-2">
+      <li
+        v-for="name in nodeNames"
+        :key="name"
+        class="px-3 py-1.5 font-mono text-sm"
+      >
+        {{ name }}
```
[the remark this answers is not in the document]
viva-jinyi: Node name is not that long.

src/workbench/extensions/manager/composables/nodePack/useWorkflowPacks.ts
```diff
@@ -111,12 +112,39 @@ const _useWorkflowPacks = () => {
...
+            !nodeDefStore.nodeDefsByName[nodeName]
+          ) {
+            unresolved.push(nodeName)
+          }
+        }
+      })
+    )
+
+    workflowPacks.value = resolvedPacks
+    unresolvedNodeNames.value = [...new Set(unresolved)]
```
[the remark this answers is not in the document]
viva-jinyi: This issue does not exist in the current code. The nodeDefsByName store is always populated before any component can invoke getWorkflowPacks(), because:

1. comfyApp.setup() awaits registerNodes() → updateVueAppNodeDefs() → populates nodeDefsByName
2. comfyAppReady is set true only after setup() completes
3. All components using useMissingNodes/useWorkflowPacks are gated behind v-if="comfyAppReady"

No change is needed — adding a hydration flag would be unnecessary complexity (YAGNI).

AustinMroz (approved):

PR #9714, opened 2026-03-10
feat: Node search UX updates

pythongosssss:
## Summary

Addresses feedback from the initial v2 node search implementation for improved UI and UX

## Changes

- **What**: 
- add root filter buttons
- remove all extra tree categories leaving only "Most relevant"
- replace input/output selection with popover
- replace price badge with one from node header
- add chevrons and additional styling to category tree
- hide empty categories
- fix bug with hovering selecting item under mouse automatically
- fix tailwind merge with custom sizes removing them
- keyboard navigation
- general tidy/refactor/test

https://github.com/user-attachments/assets/db798dfa-e248-4b48-bb56-2fa7b6c5f65f

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9714-feat-Node-search-UX-updates-31f6d73d365081cebd96c4253ad1ca53) by [Unito](https://www.unito.io)


src/components/searchbox/v2/NodeSearchContent.vue
```diff
@@ -113,57 +123,102 @@ const emit = defineEmits<{
...
+function onSelectCategory(category: string) {
+  if (rootFilter.value === category) {
+    rootFilter.value = null
+  } else {
+    rootFilter.value = category
+  }
+  selectedCategory.value = 'most-relevant'
+  searchQuery.value = ''
   nextTick(() => searchInputRef.value?.focus())
 }
```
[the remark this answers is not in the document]
pythongosssss: We'll leave it as is and see how the UX and feedback is

Yourz (dismissed):

src/components/node/CreditBadge.vue
```diff
@@ -0,0 +1,28 @@
+<template>
+  <span
+    :class="
+      cn(
+        'flex h-5 shrink-0 items-center bg-component-node-widget-background p-1 text-xs',
+        rest ? 'rounded-l-full pr-1' : 'rounded-full'
+      )
+    "
+  >
+    <i class="icon-[lucide--component] h-full bg-amber-400" />
```
viva-jinyi, line 10: can we use font size instead of h-full?

src/components/node/NodePreviewCard.vue
```diff
@@ -18,21 +23,21 @@
       <!-- Category Path -->
       <p
         v-if="showCategoryPath && nodeDef.category"
-        class="-mt-1 text-xs text-muted-foreground"
+        class="-mt-1 truncate text-xs text-muted-foreground"
       >
-        {{ nodeDef.category.replaceAll('/', ' > ') }}
+        {{ nodeDef.category.replaceAll('/', ' / ') }}
```
viva-jinyi: I think it'd be better to use computed!

src/components/node/NodePreviewCard.vue
```diff
@@ -18,21 +23,21 @@
...
+      <div class="flex flex-wrap gap-2 overflow-hidden empty:hidden">
+        <NodePricingBadge class="max-w-full truncate" :node-def="nodeDef" />
         <NodeProviderBadge :node-def="nodeDef" />
       </div>
 
       <!-- Description -->
       <p
         v-if="nodeDef.description"
-        class="m-0 text-[11px] leading-normal font-normal text-muted-foreground"
+        class="m-0 max-h-[30vh] overflow-y-auto text-[11px] leading-normal font-normal text-muted-foreground"
```
viva-jinyi: can we use text-xs instead of using text-[11px]?

src/components/searchbox/v2/NodeSearchCategorySidebar.vue
```diff
@@ -1,105 +1,134 @@
...
-        type="button"
-        :data-testid="`category-${preset.id}`"
-        :aria-current="selectedCategory === preset.id || undefined"
-        :class="categoryBtnClass(preset.id)"
-        @click="selectCategory(preset.id)"
+        as-child
       >
-        {{ preset.label }}
-      </button>
+        <button
```
viva-jinyi: I think we can use common button component

src/components/searchbox/v2/NodeSearchCategoryTreeNode.vue
```diff
@@ -1,32 +1,66 @@
...
-      :key="child.key"
-      :node="child"
-      :depth="depth + 1"
-      :selected-category="selectedCategory"
-      :selected-collapsed="selectedCollapsed"
-      @select="$emit('select', $event)"
-    />
-  </template>
+    <RovingFocusItem as-child>
+      <button
```
viva-jinyi: it would be better to be replaced with button component

src/components/searchbox/NodeSearchBoxPopover.vue
```diff
@@ -7,7 +7,7 @@
       :pt="{
         root: {
           class: useSearchBoxV2
-            ? 'w-4/5 min-w-[32rem] max-w-[56rem] border-0 bg-transparent mt-[10vh] max-md:w-[95%] max-md:min-w-0 overflow-visible'
+            ? 'w-4/5 min-w-[32rem] max-w-[56rem] border-0 bg-transparent max-md:w-[95%] max-md:min-w-0 overflow-visible'
```
viva-jinyi: I think `w-full max-w-[56rem] min-w-[32rem] max-md:min-w-0 bg-transparent border-0` would work too?

viva-jinyi (changes requested):

src/components/node/CreditBadge.vue
pythongosssss: This was taken directly from `src/renderer/extensions/vueNodes/components/NodeHeader.vue`, i'd rather not change it in case there is some issue with it rendering there

src/components/node/NodePreviewCard.vue
pythongosssss: 👍

src/components/searchbox/v2/NodeSearchCategorySidebar.vue
pythongosssss: Using the Button component means needing to fight against the styling it adds or adding a new variant that would only be used in this context - I think it adds more overhead for little benefit to use it here

src/components/searchbox/v2/NodeSearchCategoryTreeNode.vue
pythongosssss: Using the Button component means needing to fight against the styling it adds or adding a new variant that would only be used in this context - I think it adds more overhead for little benefit to use it here

src/components/searchbox/NodeSearchBoxPopover.vue
pythongosssss: It's a bit different, as that then removes the spacing around at smaller screen sizes, but that's probably not a bad thing, so done 👍

src/components/node/NodePreviewCard.vue
pythongosssss: I'm not aware of the specifics of this, but since this was recently redesigned i'd imagine this was a specific part of the design

viva-jinyi: We’ve defined all text sizes using size tokens, so for things like color and size we typically rely on what’s already defined in Tailwind. Would it be okay to just use text-sm here?

src/components/searchbox/v2/NodeSearchCategorySidebar.vue
viva-jinyi: Isn’t this basically just a text-only button? Couldn’t we just add a text-only variant to the Button component?

viva-jinyi:
I think we could make the code shorter here by using a shared button component for the buttons in this section. But everyone has a different bar for this, so let’s hear other people’s opinions too. @DrJKL What do you think?

src/components/node/NodePreviewCard.vue
pythongosssss: I'd say we should go with what product have said to use, but as stated, I am not aware of the specific for why 11px was chosen and it is a pre-existing styling choice for this unique component.

https://github.com/Comfy-Org/ComfyUI_frontend/commit/824f8bb2814db1627c138f11d8042899ff97ff28#diff-e708b804b619de1b1aba97d98a71dad92af0d07e7b6b646b8bf5aaf4bb99c013
@Yourz it looks like you updated it to use 11px while implementing the component, was this due to a request from product/design, is it fine to change it?

Yourz: yeah. this is based on the design feedback from julien

comfydesigner: Screenshot 2026-03-23 at 9 44 09 AM

The description text of the node preview right? To keep things more consistent, I'd push for text-xs, but I'm not aware if Julien had any concerns about the text being too large when it was 12pt. 

I tried text-xs just now and it looks alright to me

viva-jinyi (commented):
👍

src/components/searchbox/v2/NodeSearchListItem.vue
```diff
@@ -2,46 +2,77 @@
...
+              aria-hidden="true"
+              class="flex size-[18px] shrink-0 items-center justify-center rounded-full bg-secondary-background-hover/80"
+            >
+              <ComfyLogo :size="10" mode="fill" color="currentColor" />
+            </span>
+            <span
+              v-else-if="showSourceBadge && isCustom"
+              :class="badgePillClass"
+            >
+              <span class="truncate text-[10px]">
```
viva-jinyi: The design system defines `--text-2xs: 0.625rem` (= 10px). Use the semantic token instead of an arbitrary value.

```suggestion
              <span class="truncate text-2xs">
```

src/components/searchbox/v2/NodeSearchFilterBar.vue
```diff
@@ -1,22 +1,43 @@
...
+      @toggle="(v) => emit('toggleFilter', tf.chip.filter, v)"
+      @clear="emit('clearFilterGroup', tf.chip.filter.id)"
+      @escape-close="emit('focusSearch')"
+    >
+      <button type="button" :class="chipClass(false, tf.values.length > 0)">
+        <span v-if="tf.values.length > 0" class="flex items-center">
+          <span
+            v-for="val in tf.values.slice(0, MAX_VISIBLE_DOTS)"
+            :key="val"
+            class="-mx-[2px] text-lg leading-none"
```
viva-jinyi, line 32: Tailwind has `-mx-0.5` for 2px. No need for an arbitrary value.

```suggestion
            class="-mx-0.5 text-lg leading-none"
```

src/components/searchbox/v2/NodeSearchContent.vue
```diff
@@ -176,67 +243,70 @@ const searchResults = computed(() => {
   })
 })
 
-const effectiveCategory = computed(() =>
-  searchQuery.value ? 'most-relevant' : selectedCategory.value
-)
+const effectiveCategory = computed(() => selectedCategory.value)
```
viva-jinyi: This computed is now a trivial pass-through — it just wraps `selectedCategory` with no transformation. Then `sidebarCategory` wraps it again in a writable computed. Both indirections can be removed; use `selectedCategory` directly.

src/components/searchbox/v2/NodeSearchCategorySidebar.vue
```diff
@@ -114,28 +117,82 @@ const categoryTree = computed<CategoryNode[]>(() => {
...
+    hideChevrons ? 'pl-3' : 'pl-9',
     selectedCategory.value === id
       ? CATEGORY_SELECTED_CLASS
       : CATEGORY_UNSELECTED_CLASS
   )
 }
 
-const selectedCollapsed = ref(false)
+const expandedCategory = ref(selectedCategory.value)
+let lastEmittedCategory = ''
```
viva-jinyi, line 172: This is a non-reactive mutable `let` variable used to coordinate between `selectCategory`/`collapseCategory` and the `watch`. Per project guidelines: "Avoid mutable state, prefer immutability." Consider using a `ref` to make the state tracking explicit and reactive.

src/components/searchbox/v2/NodeSearchTypeFilterPopover.vue
```diff
@@ -0,0 +1,175 @@
...
+const searchFilterRef = ref<InstanceType<typeof ListboxFilter>>()
+
+function onOpenChange(isOpen: boolean) {
+  if (!isOpen) searchQuery.value = ''
+}
+
+const selectedSet = computed(() => new Set(selectedValues))
+
+function onSelectionChange(value: AcceptableValue) {
+  const newValues = value as string[]
```
viva-jinyi, line 142: Use `ListboxRoot<string>` generic type parameter instead of casting. This provides proper typing without the assertion.

```suggestion
  const newValues = value as string[]
```

→ Configure `ListboxRoot` with `<string>` generic and this cast becomes unnecessary.

viva-jinyi (changes requested):

src/components/searchbox/v2/NodeSearchTypeFilterPopover.vue
pythongosssss: I don't think this is possible, the reka ui emit is typed as AcceptableValue

src/components/searchbox/v2/NodeSearchCategorySidebar.vue
pythongosssss: This shouldnt be reactive, this is used to prevent immediate changes from being undone vs changes from the parent, i'm not sure there is a cleaner solution

src/components/searchbox/v2/NodeSearchCategoryTreeNode.vue
pythongosssss: Done

src/components/searchbox/v2/NodeSearchCategorySidebar.vue
pythongosssss: Done

src/components/searchbox/v2/NodeSearchContent.vue
```diff
@@ -1,107 +1,129 @@
...
+  isCustomNode,
+  isEssentialNode
+} from '@/types/nodeSource'
+import type { FuseFilter, FuseFilterWithValue } from '@/utils/fuseUtil'
 import { cn } from '@/utils/tailwindUtil'
 
+const sourceCategoryFilters: Record<string, (n: ComfyNodeDefImpl) => boolean> =
+  {
+    essentials: isEssentialNode,
+    comfy: (n) => !isCustomNode(n),
```
dante01yoon: comfy used to mean NodeSourceType.Core, but !isCustomNode() now also matches Essentials, Blueprint, and Unknown. That makes the Comfy root filter overlap the dedicated Essentials/Blueprint filters and can surface unclassified nodes in the Comfy bucket. Can we keep this scoped to core nodes instead?


```suggestion
    comfy: (n) => n.nodeSource.type === NodeSourceType.Core,
```

src/components/searchbox/v2/NodeSearchListItem.vue
```diff
@@ -2,46 +2,81 @@
...
+          class="shrink-0 rounded-sm bg-secondary-background px-1.5 py-0.5 text-xs text-muted-foreground"
           v-html="highlightQuery(nodeDef.name, currentQuery)"
         />
 
-        <NodePricingBadge :node-def="nodeDef" />
-        <NodeProviderBadge v-if="nodeDef.api_node" :node-def="nodeDef" />
+        <template v-if="showDescription">
+          <div class="flex-1" />
+          <div class="flex shrink-0 items-center gap-1">
+            <span
```
dante01yoon, line 28: This changes the badge semantics from “show explicit source info for non-core, non-unknown nodes” to “show a generic Comfy badge for everything that isn’t custom”. As a result, essentials, blueprints, and unknown nodes are now presented like first-party Comfy nodes, which is misleading. Should this stay limited to actual core nodes and keep explicit labels for the other source types?

pythongosssss:
Addressed comments, also included two other fixes
1. Remove blank categories (e.g. a//b or a/b/)
2. Fix blueprints not adding as draggable ghost node

src/components/searchbox/v2/NodeSearchCategorySidebar.vue
```diff
@@ -114,28 +118,84 @@ const categoryTree = computed<CategoryNode[]>(() => {
...
       ? CATEGORY_SELECTED_CLASS
       : CATEGORY_UNSELECTED_CLASS
   )
 }
 
-const selectedCollapsed = ref(false)
+const expandedCategory = ref(selectedCategory.value)
+// Skips the watch when selectCategory/collapseCategory set selectedCategory,
+// so their expandedCategory toggle isn't immediately undone.
+let lastEmittedCategory = ''
```
christian-byrne, line 172: question: `lastEmittedCategory` is a plain `let` used as a guard to prevent the watcher from undoing `selectCategory`/`collapseCategory` changes. The empty-string reset on line 178 is subtle -- would a `ref` flag or a `nextTick` guard make the intent clearer?

packages/shared-frontend-utils/src/formatUtil.ts
```diff
@@ -74,10 +74,14 @@ export function highlightQuery(
...
-  // Escape special regex characters in the query string
-  const escapedQuery = query.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
-
-  const regex = new RegExp(`(${escapedQuery})`, 'gi')
+  // Escape special regex characters, then join with optional
+  // whitespace so cross-word matches (e.g. "geto" → "imaGE TO") are
+  // highlighted correctly.
+  const pattern = Array.from(query)
+    .map((ch) => ch.replace(/[.*+?^${}()|[\]\\]/g, '\\$&'))
+    .join('\\s*')
```
christian-byrne, line 82: suggestion: (non-blocking) `\s*` between every character means a query like `"hi"` could match across arbitrary whitespace, e.g. `"t h i n g"`. Would `\s?` (at most one whitespace char) reduce false positives while still supporting the cross-word case?

src/components/searchbox/v2/NodeSearchContent.vue
```diff
@@ -176,67 +244,68 @@ const searchResults = computed(() => {
...
+// Check if any tree category has children (for chevron visibility)
+const anyTreeCategoryHasChildren = computed(() =>
+  rootFilteredNodeDefs.value.some((n) => n.category.includes('/'))
+)
+
+function getMostRelevantResults(baseNodes: ComfyNodeDefImpl[]) {
+  if (searchQuery.value || filters.length > 0) {
+    const searched = searchResults.value
+    if (!rootFilter.value) return searched
+    const rootSet = new Set(baseNodes.map((n) => n.name))
```
christian-byrne, line 263: nitpick: (non-blocking) `new Set(baseNodes.map(...))` is re-created on every `getMostRelevantResults` call. Since `baseNodes` comes from a computed, this Set could be lifted to its own `computed` to avoid re-allocation on each access.

src/components/searchbox/v2/NodeSearchTypeFilterPopover.vue
```diff
@@ -0,0 +1,175 @@
...
+const searchQuery = ref('')
+const searchFilterRef = ref<InstanceType<typeof ListboxFilter>>()
+
+function onOpenChange(isOpen: boolean) {
+  if (!isOpen) searchQuery.value = ''
+}
+
+const selectedSet = computed(() => new Set(selectedValues))
+
+function onSelectionChange(value: AcceptableValue) {
```
christian-byrne, line 141: nitpick: (non-blocking) Since `ListboxRoot` has `selection-behavior="toggle"`, the diff between old and new values is always exactly one item. A simpler approach could use `@select` or compare lengths to determine if it was an add or remove, avoiding the `find` calls.

dante01yoon (dismissed):

dante01yoon:
please resolve conflict

dante01yoon (changes requested):
Two things still look under-covered in the current diff:

1. `src/components/searchbox/v2/NodeSearchListItem.vue` narrows the description-mode source badge from the old “any non-core, non-unknown source” behavior to just `isCore` / `isCustom`. That means essentials and blueprint results lose their source label in mixed-result views, but `NodeSearchListItem.test.ts` only exercises the core/custom cases right now.

2. `packages/shared-frontend-utils/src/formatUtil.ts` now builds the matcher with `\s*` between every query character, so the new cross-word highlighting also spans arbitrary whitespace runs and line breaks, not just the intended single word boundary. `formatUtil.test.ts` adds the positive `ge to` case, but there is still no boundary test protecting against those broader matches. Because `highlightQuery()` is shared by both the v2 and legacy search result items, this behavior change is wider than just the new UI path.

dante01yoon (approved):

PR #9724, opened 2026-03-10
fix: make zPreviewOutput accept text-only job outputs

luke-mino-altherr:
## Summary

Fixes Zod validation crash when the jobs batch contains text-only preview outputs (e.g. from LLM nodes), which caused the entire Assets sidebar to show nothing.

## Changes

- **What**: Made `filename`, `subfolder`, and `type` optional in `zPreviewOutput` and added `.passthrough()` for extra fields like `content`. Text-only jobs are safely filtered out downstream by `supportsPreview`.
- Added tests in `fetchJobs.test.ts` verifying a mixed batch (image + text-only + no-preview) parses successfully.
- Added test in `assetsStore.test.ts` verifying text-only jobs are skipped without breaking sibling image jobs. Improved `TaskItemImpl` mock to realistically handle media types.

## Review Focus

- The `zPreviewOutput` schema now uses `.passthrough()` to allow extra fields from new preview types (like `content` on text previews). This is consistent with `zRawJobListItem` and `zExecutionError` which also use `.passthrough()`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9724-fix-make-zPreviewOutput-accept-text-only-job-outputs-31f6d73d36508119a7aef99f9b765ecd) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/remote/comfyui/jobs/jobTypes.ts
+++ b/src/platform/remote/comfyui/jobs/jobTypes.ts
@@ -18,14 +18,16 @@ const zJobStatus = z.enum([
   'cancelled'
 ])
 
-const zPreviewOutput = z.object({
-  filename: z.string(),
-  subfolder: z.string(),
-  type: resultItemType,
-  nodeId: z.string(),
-  mediaType: z.string(),
-  display_name: z.string().optional()
-})
+const zPreviewOutput = z
+  .object({
+    filename: z.string().optional(),
+    subfolder: z.string().optional(),
+    type: resultItemType.optional(),
+    nodeId: z.string(),
+    mediaType: z.string(),
+    display_name: z.string().optional()
+  })
+  .passthrough()
 
 /**
  * Execution error from Jobs API.
--- a/src/stores/assetsStore.test.ts
+++ b/src/stores/assetsStore.test.ts
@@ -72,6 +72,8 @@ vi.mock('@/stores/modelToNodeStore', () => ({
 }))
 
 // Mock TaskItemImpl
+const PREVIEWABLE_MEDIA_TYPES = new Set(['images', 'video', 'audio'])
+
 vi.mock('@/stores/queueStore', () => ({
   TaskItemImpl: class {
     public flatOutputs: Array<{
@@ -91,19 +93,28 @@ vi.mock('@/stores/queueStore', () => ({
         }
       | undefined
     public jobId: string
+    public outputsCount: number | null
 
     constructor(public job: JobListItem) {
       this.jobId = job.id
-      this.flatOutputs = [
-        {
+      this.outputsCount = job.outputs_count ?? null
+      const preview = job.preview_output
+      const isPreviewable =
+        !!preview?.filename && PREVIEWABLE_MEDIA_TYPES.has(preview.mediaType)
+      if (preview && isPreviewable) {
+        const item = {
           supportsPreview: true,
-          filename: 'test.png',
-          subfolder: '',
-          type: 'output',
-          url: 'http://test.com/test.png'
+          filename: preview.filename!,
+          subfolder: preview.subfolder ?? '',
+          type: preview.type ?? 'output',
+          url: `http://test.com/${preview.filename}`
         }
-      ]
-      this.previewOutput = this.flatOutputs[0]
+        this.flatOutputs = [item]
+        this.previewOutput = item
+      } else {
+        this.flatOutputs = []
+        this.previewOutput = undefined
+      }
     }
 
     get previewableOutputs() {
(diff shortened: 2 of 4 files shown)


src/platform/remote/comfyui/jobs/jobTypes.ts
```diff
@@ -18,14 +18,16 @@ const zJobStatus = z.enum([
...
+const zPreviewOutput = z
+  .object({
+    filename: z.string().optional(),
+    subfolder: z.string().optional(),
+    type: resultItemType.optional(),
+    nodeId: z.string(),
+    mediaType: z.string(),
+    display_name: z.string().optional()
+  })
+  .passthrough()
```
dante01yoon, line 30: Nit(non-blocking): The schema change in jobTypes.ts makes filename, subfolder, and type optional. This is safe at the Zod layer, and the new tests in `fetchJobs.test.ts` and `assetsStore.test.ts` cover parsing and asset-level filtering well.

However, I think there's a coverage gap in the real `TaskItemImpl` path. When a text-only job arrives with preview_output present (truthy), the synthetic output construction in `queueStore.ts` will create a ResultItemImpl with filename=''. This item ends up in flatOutputs but is filtered out by supportsPreview — so it's functionally safe today.

The concern is that the assetsStore.test.ts mock of TaskItemImpl reimplements the filtering logic (PREVIEWABLE_MEDIA_TYPES.has(...)) independently from the real `ResultItemImpl.supportsPreview`. If those two diverge in the future, the test would still pass while the real code breaks.

Suggestion: Add one test in `queueStore.test.ts` exercising the real `TaskItemImpl` with a text-only preview_output:
```ts
it('should produce no previewable outputs for text-only preview_output', () => {
  const job: JobListItem = {
    id: 'text-job',
    status: 'completed',
    create_time: 1000,
    priority: 1,
    preview_output: {
      nodeId: '5',
      mediaType: 'text'
    } as JobListItem['preview_output']
  }

  const task = new TaskItemImpl(job)

  expect(task.flatOutputs).toHaveLength(1)
  expect(task.flatOutputs[0].filename).toBe('')
  expect(task.previewableOutputs).toHaveLength(0)
  expect(task.previewOutput).toBeUndefined()
})
```
This ensures the real code path is covered, not just the mock's re-implementation.

dante01yoon (dismissed):

src/stores/assetsStore.test.ts
```diff
@@ -91,19 +93,28 @@ vi.mock('@/stores/queueStore', () => ({
...
+      const isPreviewable =
+        !!preview?.filename && PREVIEWABLE_MEDIA_TYPES.has(preview.mediaType)
+      if (preview && isPreviewable) {
+        const item = {
           supportsPreview: true,
-          filename: 'test.png',
-          subfolder: '',
-          type: 'output',
-          url: 'http://test.com/test.png'
+          filename: preview.filename!,
```
DrJKL, line 107: Does it need the bang?

src/stores/assetsStore.test.ts
```diff
@@ -200,6 +211,33 @@ describe('assetsStore - Refactored (Option A)', () => {
...
+        {
+          id: 'text-only-job',
+          status: 'completed',
+          create_time: 2000,
+          priority: 2000,
+          preview_output: {
+            content: 'some generated text',
+            nodeId: '5',
+            mediaType: 'text'
+          } as JobListItem['preview_output']
```
DrJKL: Might be safer to do `satisfies`, to add a contract check.

src/stores/queueStore.test.ts
```diff
@@ -191,6 +191,23 @@ describe('TaskItemImpl', () => {
     })
   })
 
+  it('should produce no previewable outputs for text-only preview_output', () => {
+    const job: JobListItem = {
+      ...createHistoryJob(0, 'text-job'),
+      preview_output: {
+        nodeId: '5',
+        mediaType: 'text'
+      } as JobListItem['preview_output']
```
DrJKL: Same here, unless you're intentionally violating the type.

DrJKL (dismissed):

src/stores/assetsStore.test.ts
```diff
@@ -72,6 +72,8 @@ vi.mock('@/stores/modelToNodeStore', () => ({
 }))
 
 // Mock TaskItemImpl
+const PREVIEWABLE_MEDIA_TYPES = new Set(['images', 'video', 'audio'])
+
```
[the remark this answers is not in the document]
DrJKL: @coderabbitai Open an issue to flesh out the assetsStore tests to be more comprehensive (and also to not redefine the types separately within the test module).

DrJKL (approved):

PR #9730, opened 2026-03-11
feat/mask-editor-brush-step-size

kaili-yang:
## Summary

Fix #9727
FYI #9534

## GIMP Source

Accroding to GIMP( like open-source Adobe Photoshop)`/composables/maskeditor/useBrushDrawing.ts`, 
```
const brushStepSizeSliderValue = computed({
  get: () => {
    if (rawStepSizeSliderValue.value !== null) {
      const cachedSize = Math.round(Math.pow(100, rawStepSizeSliderValue.value))
      if (cachedSize === brushStepSize.value) {
        return rawStepSizeSliderValue.value
      }
    }
    return Math.log(brushStepSize.value) / Math.log(100)
  },
  set: (value: number) => {
    rawStepSizeSliderValue.value = value
    const size = Math.round(Math.pow(100, value))
    store.setBrushStepSize(size)
  }
})
```

## Screenshot

https://github.com/user-attachments/assets/971e7bd9-9690-475f-b214-33d06939d2ef

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9730-feat-mask-editor-brush-step-size-3206d73d36508198b934e0fb374eb5a9) by [Unito](https://www.unito.io)


jtydhr88 (approved):
LGTM

src/composables/maskeditor/useBrushDrawing.ts
```diff
@@ -924,7 +924,8 @@ export function useBrushDrawing(initialSettings?: {
       }
 
       // Calculate target spacing based on step size percentage
-      const stepPercentage = store.brushSettings.stepSize / 100
+      const stepPercentage =
+        Math.pow(100, store.brushSettings.stepSize / 100) / 100
```
DrJKL, line 928: Not for this PR, but it seems like there's a lot of math going on in these functions that could be extracted. Both for sharing logic but also maybe performance.

DrJKL (approved):

christian-byrne (approved):

PR #9733, opened 2026-03-11
fix: update PostHog api_host fallback domain

robinjhuang:
Update the PostHog `api_host` fallback from `ph.comfy.org` to `t.comfy.org`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9733-fix-update-PostHog-api_host-fallback-domain-3206d73d36508107a5d1e1fdfd3ccaec) by [Unito](https://www.unito.io)


benceruleanlu (approved):
lgtm if t. is right

PR #9741, opened 2026-03-11
fix(minimap): minimap re-render/perf issue

kaili-yang:
## Summary

Fix #9732

To clarify how preventing the 60 FPS object assignment solves the `vue-i18n` (intlify) issue, here is the complete chain reaction leading to the performance loop:

1. The Root Cause: In  `useMinimapViewport.ts`, `useRafFn` acts as a timer bound to the browser's **refresh rate** (60 FPS). In the original code, it unconditionally executed the `viewportTransform.value = { ... } `assignment 60 times a second.

2. Vue's Reactivity Interception: Because `viewportTransform` is a reactive variable (`ref`), updating it causes its corresponding **computed** property (`viewportStyles`) to register a data dependency update.

3. Forced Re-rendering: The `<template> ` in `MiniMap.vue` is bound to `:style="viewportStyles"`. Since the dependent value changed, Vue's Virtual DOM decides: "I must re-render the entire `MiniMap.vue` interface **60 times** per second to ensure the element positions are up-to-date!"

4. The Victim Emerges: Inside the template of `MiniMap.vue`, there are several internationalization translation functions: `<button :aria-label="$t('g.settings')" ...> <button :aria-label="$t('g.close')" ...> `In Vue, whenever a component re-renders, all functions within its template (including `$t()`) must be re-evaluated. Because the component was being forced to re-render **60 times** per second, and there are approximately **6 calls** to `$t()` within this UI, it multiplied into 60 × 6 = **360** intlify compilation and evaluate events per second.

## Solution
Only assemble objects and hand them over to Vue for rendering when the mouse is actually dragging the canvas.

By extracting the math into **stack-allocated** primitive variables `(x, y, w, h) `and strictly comparing them, it completely halts the CPU burn at the source with minimal runtime overhead.

## Screenshot

before
image

after
image

### DIFF
--- a/src/renderer/extensions/minimap/composables/useMinimapViewport.ts
+++ b/src/renderer/extensions/minimap/composables/useMinimapViewport.ts
@@ -90,12 +90,16 @@ export function useMinimapViewport(
     const centerOffsetX = (width - bounds.value.width * scale.value) / 2
     const centerOffsetY = (height - bounds.value.height * scale.value) / 2
 
-    viewportTransform.value = {
-      x: (worldX - bounds.value.minX) * scale.value + centerOffsetX,
-      y: (worldY - bounds.value.minY) * scale.value + centerOffsetY,
-      width: viewportWidth * scale.value,
-      height: viewportHeight * scale.value
-    }
+    const x = (worldX - bounds.value.minX) * scale.value + centerOffsetX
+    const y = (worldY - bounds.value.minY) * scale.value + centerOffsetY
+    const w = viewportWidth * scale.value
+    const h = viewportHeight * scale.value
+
+    const curr = viewportTransform.value
+    if (curr.x === x && curr.y === y && curr.width === w && curr.height === h)
+      return
+
+    viewportTransform.value = { x, y, width: w, height: h }
   }
 
   const updateBounds = () => {
--- a/src/renderer/extensions/minimap/composables/useMinimapViewport.test.ts
+++ b/src/renderer/extensions/minimap/composables/useMinimapViewport.test.ts
@@ -187,6 +187,46 @@ describe('useMinimapViewport', () => {
     expect(transform.height).toBeCloseTo(viewportHeight * 0.5) // 300 * 0.5 = 150
   })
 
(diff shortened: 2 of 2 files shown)


src/renderer/extensions/minimap/composables/useMinimapViewport.ts
```diff
@@ -90,12 +90,16 @@ export function useMinimapViewport(
...
+    const x = (worldX - bounds.value.minX) * scale.value + centerOffsetX
+    const y = (worldY - bounds.value.minY) * scale.value + centerOffsetY
+    const w = viewportWidth * scale.value
+    const h = viewportHeight * scale.value
+
+    const curr = viewportTransform.value
+    if (curr.x === x && curr.y === y && curr.width === w && curr.height === h)
+      return
+
+    viewportTransform.value = { x, y, width: w, height: h }
```
dante01yoon, line 102: Nit: This PR fixes a real performance issue with clear results, but per project guidelines `("Write tests for all changes, especially bug fixes to catch future regressions")`, please add a unit test for `updateViewport` that asserts `viewportTransform.value` retains the same object reference when called repeatedly with unchanged canvas state, and assigns a new object when `ds.offset` or `ds.scale` changes — this ensures the shallow equality guard doesn't regress back to unconditional assignment in future refactors.

dante01yoon (commented):
Thanks for the contribution. The symptom is valid, but I do not think `vue-i18n` is the root cause here.

The repeated `intlify` events look more like a downstream effect of the minimap’s RAF-based reactive updates. `useMinimapViewport` was assigning a new `viewportTransform` object every frame, which forces `MiniMap.vue` to re-render, and the template’s `$t(...)` calls then show up in the timeline.

So this seems better categorized as a minimap re-render/perf issue rather than an i18n bug.

src/renderer/extensions/minimap/composables/useMinimapViewport.ts
kaili-yang: A unit test is added and the name of the issue has been changed. Thanks for the review. ❤️

jtydhr88 (approved):
LGTM

PR #9743, opened 2026-03-11
[feat] Surface missing models in Errors tab (Cloud)

jaeone94:
## Summary
When a workflow is loaded with missing models, users currently have no way to identify or resolve them from within the UI. This PR adds a full missing-model detection and resolution pipeline that surfaces missing models in the Errors tab, allowing users to install or import them without leaving the editor.

### Missing Model Detection
- Scan all COMBO widgets across root graph and subgraphs for model-like filenames during workflow load
- Enrich candidates with embedded workflow metadata (url, hash, directory) when available
- Verify asset-supported candidates against the asset store asynchronously to confirm installation status
- Propagate missing model state to `executionErrorStore` alongside existing node/prompt errors

### Errors Tab UI — Model Resolution
- Group missing models by directory (e.g. `checkpoints`, `loras`, `vae`) with collapsible category cards
- Each model row displays:
  - Model name with copy-to-clipboard button
  - Expandable list of referencing nodes with locate-on-canvas button
  - **Library selector**: Pick an alternative from the user's existing models to substitute the missing model with one click
  - **URL import**: Paste a Civitai or HuggingFace URL to import a model directly; debounced metadata fetch shows filename and file size before confirming; type-mismatch warnings (e.g. importing a LoRA into checkpoints directory) are surfaced with an "Import Anyway" option
  - **Upgrade prompt**: In cloud environment, free-tier subscribers are shown an upgrade modal when attempting URL import
- Separate "Import Not Supported" section for custom-node models that cannot be auto-resolved
- Status card with live download progress, completion, failure, and category-mismatch states

### Canvas Integration
- Highlight nodes and widgets that reference missing models with error indicators
- Propagate missing-model badges through subgraph containers so issues are visible at every graph level

### Code Cleanup
- Simplify `surfacePendingWarnings` in workflowService, remove stale widget-detected model merging logic
- Add `flattenWorkflowNodes` utility to workflowSchema for traversing nested subgraph structures
- Extract `MissingModelUrlInput`, `MissingModelLibrarySelect`, `MissingModelStatusCard` as focused single-responsibility components

## Testing
- Unit tests for scan pipeline (`missingModelScan.test.ts`): enrichment, skip-installed, subgraph flattening
- Unit tests for store (`missingModelStore.test.ts`): state management, removal helpers
- Unit tests for interactions (`useMissingModelInteractions.test.ts`): combo select, URL input, import flow, library confirm
- Component tests for `MissingModelCard` and error grouping (`useErrorGroups.test.ts`)
- Updated `workflowService.test.ts` and `workflowSchema.test.ts` for new logic

## Review Focus
- Missing model scan + enrichment pipeline in `missingModelScan.ts`
- Interaction composable `useMissingModelInteractions.ts` — URL metadata fetch, library install, upload fallback
- Store integration and canvas-level error propagation

## Screenshots 

https://github.com/user-attachments/assets/339a6d5b-93a3-43cd-98dd-0fb00681b66f

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9743-feat-Surface-missing-models-in-Errors-tab-Cloud-3206d73d365081678326d3a16c2165d8) by [Unito](https://www.unito.io)


src/components/rightSidePanel/errors/useErrorGroups.ts
```diff
@@ -559,14 +569,72 @@ export function useErrorGroups(
...
+
+  function buildMissingModelGroups(): ErrorGroup[] {
+    if (!missingModelGroups.value.length) return []
+    return [
+      {
+        type: 'missing_model' as const,
+        title: `${t('rightSidePanel.missingModels.missingModelsTitle')} (${missingModelGroups.value.reduce((count, group) => count + group.models.length, 0)})`,
+        priority: 2
+      }
+    ]
```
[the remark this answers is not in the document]
jaeone94: This is intentional. Missing node, swap node, and missing model are always displayed regardless of the search value. This is because failing to resolve them constitutes a critical error that renders the workflow execution itself meaningless.

src/platform/missingModel/missingModelScan.ts
```diff
@@ -0,0 +1,386 @@
...
+  for (const r of settled) {
+    if (r.status === 'rejected') {
+      console.warn(
+        '[Missing Model Pipeline] checkModelInstalled failed:',
+        r.reason
+      )
+      continue
+    }
+    if (r.value) enriched.push(r.value)
+  }
```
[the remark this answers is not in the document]
jaeone94: The current behavior is intentional.

  When checkModelInstalled throws, Promise.allSettled captures it as { status: 'rejected' }, and lines 228-233 already handle this by logging a warning and
  skipping the candidate. This is the right approach because:

  1. If we can't determine whether a model is installed, we shouldn't surface it as missing — that would produce false positives for models that are actually
  installed.
  2. Adding isMissing: undefined to the candidate list would leave non-asset-supported models (custom node models) permanently unresolved, since they have no
  secondary verification path via verifyAssetSupportedCandidates.
  3. This is consistent with the same principle applied in verifyAssetSupportedCandidates, where candidates whose asset refresh failed are also skipped rather
  than classified.
  4. Introducing an installCheckError field would expand the MissingModelCandidate type surface with no consumer — nothing in the UI or pipeline reads or acts on
  such a field.

  The current console.warn + skip is the appropriate error handling for this case.

src/platform/missingModel/components/MissingModelCard.vue
```diff
@@ -0,0 +1,78 @@
...
+
+<script setup lang="ts">
+import { useI18n } from 'vue-i18n'
+import type { MissingModelGroup } from '@/platform/missingModel/types'
+import MissingModelRow from '@/platform/missingModel/components/MissingModelRow.vue'
+
+defineProps<{
+  missingModelGroups: MissingModelGroup[]
+  showNodeIdBadge: boolean
+}>()
```
viva-jinyi: very nit
```suggestion
const { missingModelGroups, showNodeIdBadge } = defineProps<{
  missingModelGroups: MissingModelGroup[]
  showNodeIdBadge: boolean
}>()
```

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -33,7 +33,9 @@
           :collapse="isSectionCollapsed(group.title) && !isSearching"
           class="border-b border-interface-stroke"
           :size="
-            group.type === 'missing_node' || group.type === 'swap_nodes'
+            group.type === 'missing_node' ||
+            group.type === 'swap_nodes' ||
+            group.type === 'missing_model'
```
viva-jinyi: very nit: can we replace it to computed?

src/platform/missingModel/components/MissingModelCard.vue
```diff
@@ -0,0 +1,78 @@
...
+      <!-- Asset unsupported group notice -->
+      <div
+        v-if="!group.isAssetSupported"
+        class="flex items-start gap-1.5 px-0.5 py-1 pl-2"
+      >
+        <i
+          aria-hidden="true"
+          class="mt-0.5 icon-[lucide--info] size-3.5 shrink-0 text-muted-foreground"
+        />
+        <span class="text-[11px] leading-tight text-muted-foreground">
```
viva-jinyi: I think we can use tailwind class instead

src/platform/missingModel/components/MissingModelLibrarySelect.vue
```diff
@@ -0,0 +1,51 @@
...
+      <template #dropdownicon>
+        <i class="icon-[lucide--chevron-down] size-3.5 text-muted-foreground" />
+      </template>
+    </SelectPlus>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { useI18n } from 'vue-i18n'
+import SelectPlus from '@/components/primevueOverride/SelectPlus.vue'
```
viva-jinyi: We decided to stop using `PrimeVue` and move to `Reka UI` instead.

src/platform/missingModel/components/MissingModelRow.vue
```diff
@@ -0,0 +1,222 @@
...
+            canConfirm ? 'bg-primary/10 hover:bg-primary/15' : 'opacity-20'
+          )
+        "
+        @click="
+          confirmLibrarySelect(
+            modelKey,
+            model.name,
+            model.referencingNodes,
+            directory
+          )
```
viva-jinyi: can you replace it with the handler like `handleLibrarySelect`?

src/platform/missingModel/components/MissingModelUrlInput.vue
```diff
@@ -0,0 +1,161 @@
...
+            tabindex: 0,
+            onKeydown: (e: KeyboardEvent) => {
+              if (e.key === 'Enter' || e.key === ' ') {
+                e.preventDefault()
+                showUploadDialog()
+              }
+            }
+          }
+        : {}
+    "
```
viva-jinyi: This v-bind is getting a bit complex with the conditional object and inline handler. It might be cleaner to move this into a computed prop and reference it from the template.

src/platform/missingModel/components/MissingModelUrlInput.vue
```diff
@@ -0,0 +1,161 @@
...
+
+  <TransitionCollapse>
+    <div v-if="urlMetadata[modelKey]" class="flex flex-col gap-2">
+      <div class="flex items-center gap-2 px-0.5 pt-0.5">
+        <span class="text-foreground min-w-0 truncate text-xs font-bold">
+          {{ urlMetadata[modelKey]?.filename }}
+        </span>
+        <span
+          v-if="(urlMetadata[modelKey]?.content_length ?? 0) > 0"
+          class="shrink-0 rounded-sm bg-secondary-background-selected px-1.5 py-0.5 text-[10px] font-medium text-muted-foreground"
```
viva-jinyi: I think you can use `text-sm` or `text-xs`?

src/platform/missingModel/components/MissingModelUrlInput.vue
```diff
@@ -0,0 +1,161 @@
...
+          {{ formatSize(urlMetadata[modelKey]?.content_length ?? 0) }}
+        </span>
+      </div>
+
+      <div v-if="typeMismatch" class="flex items-start gap-1.5 px-0.5">
+        <i
+          aria-hidden="true"
+          class="mt-0.5 icon-[lucide--triangle-alert] size-3 shrink-0 text-warning-background"
+        />
+        <span class="text-[11px] leading-tight text-warning-background">
```
viva-jinyi: Same here!

src/platform/missingModel/components/MissingModelUrlInput.vue
```diff
@@ -0,0 +1,161 @@
...
+          @click="handleImport(modelKey, directory)"
+        >
+          <i
+            aria-hidden="true"
+            :class="
+              urlImporting[modelKey]
+                ? 'icon-[lucide--loader-circle] size-4 animate-spin'
+                : 'icon-[lucide--download] size-4'
+            "
+          />
```
viva-jinyi: Question: Doesn’t the button already have a loading state prop?

src/platform/missingModel/components/MissingModelCard.vue
jaeone94: Done. Applied reactive props destructuring as suggested.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Done. Extracted to a `getGroupSize()` function using a `Set` for the full-size group types. Chose a plain function over `computed` since it takes a parameter (the group) and the Set lookup is O(1).

src/platform/missingModel/components/MissingModelCard.vue
jaeone94: Done. Replaced `text-[11px]` with `text-xs`.

src/platform/missingModel/components/MissingModelLibrarySelect.vue
jaeone94: Done. Migrated from PrimeVue `SelectPlus` to the project's Reka UI `Select` wrapper. Also added a `#prepend` slot to `SelectContent.vue` to support placing a search filter input above the scroll buttons. The search now uses `useFuse` from `@vueuse/integrations` for fuzzy matching, which handles large option lists (e.g. 100+ LoRAs) better than simple substring matching.

src/platform/missingModel/components/MissingModelRow.vue
jaeone94: Done. Extracted the inline `confirmLibrarySelect(...)` call into a `handleLibrarySelect()` function.

src/platform/missingModel/components/MissingModelUrlInput.vue
jaeone94: Done. Extracted to `upgradePromptAttrs` computed property.

src/platform/missingModel/components/MissingModelUrlInput.vue
jaeone94: Done. Replaced `text-[10px]` with `text-xs`.

src/platform/missingModel/components/MissingModelUrlInput.vue
jaeone94: Done. Replaced `text-[11px]` with `text-xs`.

src/platform/missingModel/components/MissingModelUrlInput.vue
jaeone94: Yes, it does. Replaced the manual spinner icon toggle with `:loading="urlImporting[modelKey]"` on the Button component.

viva-jinyi (approved):
LGTM

PR #9745, opened 2026-03-11
fix: switching tabs in app mode clearing outputs

pythongosssss:
## Summary

- remove reset on exiting app mode and instead cleanup at specific stages instead of a reset all
- more job<->workflow specificity updates
- ensure pending data is cleared up and doesnt leak over time

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9745-fix-switching-tabs-in-app-mode-clearing-outputs-3206d73d365081038cb0c83f0d953e71) by [Unito](https://www.unito.io)


AustinMroz (approved):
Tests are wonderfully thorough.

There will probably need to be some significant changes to make this code work with concurrent execution (when it happens), but it's not worth worrying about now.

PR #9748, opened 2026-03-11
fix: add background to running job rows

benceruleanlu:
## Summary

Add the missing surface fill for running job rows in the queue progress panel using the existing semantic background token.

## Changes

- Apply `bg-secondary-background` to running rows in `JobAssetsList` while preserving the existing hover state.
- Reuse the existing `secondary-background` / `secondary-background-hover` tokens instead of introducing new single-use job-card tokens.

## Validation

- `pnpm test:unit -- src/components/queue/job/JobAssetsList.test.ts`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm format:check`

https://github.com/user-attachments/assets/a926ede6-99e8-4f5a-b164-f9cf3cd124a7

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9748-fix-add-background-to-running-job-rows-3206d73d365081519559dfe3a9cf2037) by [Unito](https://www.unito.io)

### DIFF
--- a/packages/design-system/src/css/style.css
+++ b/packages/design-system/src/css/style.css
@@ -15,7 +15,7 @@
 @plugin "./lucideStrokePlugin.js";
 
 /* Safelist dynamic comfy icons for node library folders */
-@source inline("icon-[comfy--{ai-model,bfl,bria,bytedance,credits,extensions-blocks,file-output,gemini,grok,hitpaw,ideogram,image-ai-edit,kling,ltxv,luma,magnific,mask,meshy,minimax,moonvalley-marey,node,openai,pin,pixverse,play,recraft,rodin,runway,sora,stability-ai,template,tencent,topaz,tripo,veo,vidu,wan,wavespeed,workflow}]");
+@source inline("icon-[comfy--{ai-model,bfl,bria,bytedance,credits,elevenlabs,extensions-blocks,file-output,gemini,grok,hitpaw,ideogram,image-ai-edit,kling,ltxv,luma,magnific,mask,meshy,minimax,moonvalley-marey,node,openai,pin,pixverse,play,recraft,reve,rodin,runway,sora,stability-ai,template,tencent,topaz,tripo,veo,vidu,wan,wavespeed,workflow}]");
 
 /* Safelist dynamic comfy icons for essential nodes (kebab-case of node names) */
 @source inline("icon-[comfy--{save-image,load-video,save-video,load-3-d,save-glb,image-batch,batch-images-node,image-crop,image-scale,image-rotate,image-blur,image-invert,canny,recraft-remove-background-node,kling-lip-sync-audio-to-video-node,load-audio,save-audio,stability-text-to-audio,lora-loader,lora-loader-model-only,primitive-string-multiline,get-video-components,video-slice,tencent-text-to-model-node,tencent-image-to-model-node,open-ai-chat-node,subgraph-blueprint-canny-to-video-ltx-2-0,subgraph-blueprint-pose-to-video-ltx-2-0,preview-image,image-and-mask-preview,layer-mask-mask-preview,mask-preview,image-preview-from-latent}]");
@@ -1418,15 +1418,6 @@ audio.comfy-audio.empty-audio-widget {
   transition:
     opacity 0.1s ease,
     font-size 0.1s ease;
-}
-
-/* Performance optimization during canvas interaction */
-.transform-pane--interacting .lg-node * {
-  transition: none !important;
-}
-
-.transform-pane--interacting .lg-node {
-  will-change: transform;
 }
 
 /* ===================== Mask Editor Styles ===================== */
--- a/packages/design-system/src/css/style.css
+++ b/packages/design-system/src/css/style.css
@@ -15,7 +15,7 @@
 @plugin "./lucideStrokePlugin.js";
 
 /* Safelist dynamic comfy icons for node library folders */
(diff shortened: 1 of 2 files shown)


comfydesigner:
Nice!

packages/design-system/src/css/style.css
```diff
@@ -305,6 +305,8 @@
   --component-node-widget-background-highlighted: var(--color-ash-500);
   --component-node-widget-promoted: var(--color-purple-700);
   --component-node-widget-advanced: var(--color-azure-400);
+  --interface-panel-job-card-surface: var(--secondary-background);
+  --interface-panel-job-card-hover: var(--secondary-background-hover);
```
DrJKL, line 308: I'd like to avoid adding more single use tokens if we can help it. Is there an existing one that we can expand semantically to cover these?

src/components/queue/job/JobAssetsList.vue
```diff
@@ -103,6 +104,14 @@ const isCancelable = (job: JobListItem) =>
 const isFailedDeletable = (job: JobListItem) =>
   job.showClear !== false && job.state === 'failed'
 
+const getJobRowClass = (job: JobListItem) =>
+  cn(
+    'w-full shrink-0 cursor-default text-text-primary transition-colors',
+    job.state === 'running'
+      ? 'bg-interface-panel-job-card-surface hover:bg-interface-panel-job-card-hover'
+      : 'hover:bg-secondary-background-hover'
+  )
```
DrJKL: nit: If it's only used in one place, I prefer to keep the `cn` in the template.

DrJKL: Option:
```suggestion
const getJobRowClass = (job: JobListItem) =>
  cn(
    'w-full shrink-0 cursor-default text-text-primary transition-colors hover:bg-secondary-background-hover',
    job.state === 'running'
      && 'bg-interface-panel-job-card-surface hover:bg-interface-panel-job-card-hover'
  )
```

packages/design-system/src/css/style.css
comfydesigner: `secondary-background` and `secondary-background-hover` could be used instead of the single use tokens 'interface-panel-job-card-...'

Works in light theme as well. I'll remove the latter component tokens from our figma color system and just use the semantic ones

src/components/queue/job/JobAssetsList.test.ts
```diff
@@ -143,4 +143,33 @@ describe('JobAssetsList', () => {
...
+    expect(runningItem.classes()).toContain(
+      'hover:bg-interface-panel-job-card-hover'
+    )
+    expect(completedItem.classes()).not.toContain(
+      'bg-interface-panel-job-card-surface'
+    )
+    expect(completedItem.classes()).toContain(
+      'hover:bg-secondary-background-hover'
+    )
+  })
```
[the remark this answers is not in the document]
benceruleanlu: Removed the style-coupled class assertion instead of replacing it with a prop/state tautology. The background change is now left out of unit assertions, and the file keeps behavioral coverage for the actual `JobAssetsList` interactions.

packages/design-system/src/css/style.css
benceruleanlu: Updated to use `secondary-background` / `secondary-background-hover` and removed the `interface-panel-job-card-*` tokens.

src/components/queue/job/JobAssetsList.vue
benceruleanlu: Inlined the `cn()` call in the template in the rebased version.

benceruleanlu: Applied in the current version, with the semantic token change from the other thread: the shared hover class stays inline and running rows add `bg-secondary-background`.

packages/design-system/src/css/style.css
```diff
@@ -1420,6 +1420,15 @@ audio.comfy-audio.empty-audio-widget {
...
 
+/* Performance optimization during canvas interaction */
+.transform-pane--interacting .lg-node * {
+  transition: none !important;
+}
+
+.transform-pane--interacting .lg-node {
+  will-change: transform;
+}
+
```
jaeone94: `.transform-pane--interacting` CSS rules are dead code that resurrects already-removed legacy

The `.transform-pane--interacting` CSS class was **already removed from `main`** in commit `67e363fc3` by Rizumi Ayaka ("fix: legacy issues with transform-pane--interacting css class"). The removal was intentional — the approach was replaced with Tailwind utility classes (`will-change-transform` / `will-change-auto`) toggled directly on the container element.

Current `TransformPane.vue` (lines 51-56):
```ts
watch([isInteracting, transformPaneRef], ([interacting, el]) => {
  if (el) {
    el.classList.toggle('will-change-transform', interacting)
    el.classList.toggle('will-change-auto', !interacting)
  }
})
```

**No code in the codebase applies the `transform-pane--interacting` class.** Merging this PR would re-introduce dead CSS that was deliberately removed. These rules should be dropped from this PR entirely.

jaeone94 (commented):
This PR contains three distinct changes under a single commit:

1. **Icon safelist cleanup** — removal of `elevenlabs` and `reve` from the `@source` safelist (`style.css:18`)
2. **Canvas interaction performance CSS** — dead `.transform-pane--interacting` rules (`style.css:1423-1430`)
3. **Running job row background** — the stated fix (`JobAssetsList.vue:19-24`)

Consider removing (2) entirely and splitting (1) into a separate commit or PR for easier review and bisection.

### 3. `!important` + wildcard selector (`style.css:1425`)

```css
.transform-pane--interacting .lg-node * {
  transition: none !important;
}
```

Even if this rule were kept, the project guideline discourages `!important`, and the `*` wildcard combined with `!important` is particularly broad. This is moot given Finding 1, but worth noting for any future alternative approach.

### Recommendation

**Remove the `.transform-pane--interacting` CSS rules (lines 1423-1430) from this PR.** The `JobAssetsList.vue` change looks good to merge once the dead CSS is dropped.

packages/design-system/src/css/style.css
benceruleanlu: whoops had an outdated checkout, 1 and 2 are unintentional and have been reverted, merging 3 now

PR #9749, opened 2026-03-11
fix: dismiss queue history menus on pointerdown

benceruleanlu:
## Summary

Dismiss the queue history asset menu on outside pointerdown and simplify the overlay dismissal implementation so it no longer depends on PrimeVue instance internals.

## Changes

- **What**: Close the job history menu on outside `pointerdown`, preserve same-trigger toggle behavior, and extract shared dismissal logic into `useDismissableOverlay` for the queue and media asset menus.

## Review Focus

Verify the menu closes when drag-panning starts outside the overlay, but remains stable when interacting with the menu trigger itself.

https://github.com/user-attachments/assets/f5c51460-23b8-4fc9-9515-363387d96fbf

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9749-fix-dismiss-queue-history-menus-on-pointerdown-3206d73d365081d2b696fc5624a04bae) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/queue/job/JobContextMenu.vue
+++ b/src/components/queue/job/JobContextMenu.vue
@@ -1,7 +1,7 @@
 <template>
   <Popover
     ref="jobItemPopoverRef"
-    :dismissable="true"
+    :dismissable="false"
     :close-on-escape="true"
     unstyled
     :pt="{
--- a/src/components/queue/job/JobContextMenu.vue
+++ b/src/components/queue/job/JobContextMenu.vue
@@ -12,8 +12,11 @@
         ]
       }
     }"
+    @show="isVisible = true"
+    @hide="onHide"
   >
     <div
+      ref="contentRef"
       class="flex min-w-56 flex-col items-stretch rounded-lg border border-interface-stroke bg-interface-panel-surface px-2 py-3 font-inter"
     >
       <template v-for="entry in entries" :key="entry.key">
@@ -45,9 +48,10 @@
 
 <script setup lang="ts">
 import Popover from 'primevue/popover'
-import { ref } from 'vue'
+import { nextTick, ref } from 'vue'
 
 import Button from '@/components/ui/button/Button.vue'
+import { useDismissableOverlay } from '@/composables/useDismissableOverlay'
 import type { MenuEntry } from '@/composables/queue/useJobMenu'
 
 defineProps<{ entries: MenuEntry[] }>()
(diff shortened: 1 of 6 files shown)


src/platform/assets/components/MediaAssetContextMenu.vue
```diff
@@ -74,34 +74,22 @@ const emit = defineEmits<{
...
-useEventListener(window, 'pointerdown', dismissIfOutside, { capture: true })
-useEventListener(window, 'scroll', dismissIfOutside, {
-  capture: true,
-  passive: true
+useDismissableOverlay({
+  isOpen: isVisible,
+  getOverlayEl: () => document.getElementById(contextMenuId),
+  onDismiss: hide,
+  dismissOnScroll: true
 })
```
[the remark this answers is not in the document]
benceruleanlu: It is rendered with v-bind="ptmi('root')", so it's there, I think things about the context menu is out of scope for this PR, I do agree that we should refactor to a repo-local context-menu abstraction later.

src/composables/useDismissableOverlay.ts
DrJKL, line 1: Can we use the popover from Reka/shadcn-vue? https://reka-ui.com/docs/components/popover

benceruleanlu: > Can we use the popover from Reka/shadcn-vue? https://reka-ui.com/docs/components/popover

I think that’s a broader refactor than this PR is trying to do. This change is just fixing pointerdown dismissal and removing the PrimeVue-internals dependency. I’m happy to do a follow-up to evaluate migrating these menus to Reka primitives, but I’d rather not widen this PR unless you want to explicitly redirect it in that direction.

src/composables/useDismissableOverlay.ts
```diff
@@ -0,0 +1,66 @@
...
+          return
+        }
+
+        onDismiss()
+      },
+      {
+        capture: true,
+        passive: true
+      }
+    )
```
jaeone94: The `dismissIfOutside` handler (lines 27-45) correctly checks `overlay.contains(event.target)` before dismissing, but the scroll handler (lines 53-58) unconditionally calls `onDismiss()` without the same containment check.

The previous implementation in `MediaAssetContextMenu.vue` used `dismissIfOutside` for both `pointerdown` and `scroll`, which included the containment guard. This means scrolling inside a scrollable context menu will now dismiss it — a regression from the previous behavior.

**Suggested fix:** Apply the same containment check to the scroll handler:

```ts
useEventListener(
  window,
  'scroll',
  (event: Event) => {
    if (!toValue(isOpen)) return
    if (isNode(event.target) && isInside(event.target, getOverlayEl())) return
    onDismiss()
  },
  { capture: true, passive: true }
)
```

src/components/queue/job/JobContextMenu.vue
jaeone94: `:dismissable="true"` is explicitly set on the Popover, which registers PrimeVue's built-in click-outside handler. The new `useDismissableOverlay` composable (lines 74-79) adds a second dismissal path via `pointerdown` in the capture phase. Both listeners fire on the same outside interaction, creating a double-dismiss race.

**Suggested fix:** Set `:dismissable="false"` on the Popover now that `useDismissableOverlay` fully owns dismissal.

src/platform/assets/components/MediaAssetContextMenu.vue
```diff
@@ -74,34 +74,22 @@ const emit = defineEmits<{
...
-
-useEventListener(window, 'pointerdown', dismissIfOutside, { capture: true })
-useEventListener(window, 'scroll', dismissIfOutside, {
-  capture: true,
-  passive: true
+useDismissableOverlay({
+  isOpen: isVisible,
+  getOverlayEl: () => document.getElementById(contextMenuId),
+  onDismiss: hide,
+  dismissOnScroll: true
```
jaeone94, line 92: Nit: The refactor introduces a `useId()` (line 83) + `document.getElementById(contextMenuId)` (line 90) element lookup pattern. This is a different integration path from what the composable unit tests exercise (which use direct element references). A minimal test covering the `pt.root.id` pass-through and `getElementById` lookup would add confidence that the wiring works correctly.

jaeone94 (commented):
I tested it out, and it works beautifully. I left a few comments.

src/composables/useDismissableOverlay.ts
benceruleanlu: Fixed in `a4bc686fb`. I changed `useDismissableOverlay()` so `scroll` now goes through the same outside/containment guard as `pointerdown`, which preserves the old behavior for scrolling inside the overlay. I also added a regression test for inside-overlay scroll in `src/composables/useDismissableOverlay.test.ts`.

src/platform/assets/components/MediaAssetContextMenu.vue
benceruleanlu: Added in `a4bc686fb`. There is now a focused `MediaAssetContextMenu` test that exercises the `pt.root.id` passthrough plus the `document.getElementById(contextMenuId)` lookup path and verifies outside `pointerdown` dismissal end-to-end.

src/components/queue/job/JobContextMenu.vue
benceruleanlu: Applied in `a4bc686fb`. `JobContextMenu` now sets `:dismissable="false"` so `useDismissableOverlay()` is the single owner of outside dismissal, while keeping the PR’s `pointerdown` behavior and same-trigger handling.

jaeone94 (approved):
LGTM

PR #9758, opened 2026-03-11
feat: add server-side PostHog config overrides

robinjhuang:
Add posthog_config to RemoteConfig so any PostHog init parameter can be overridden via /api/features. Client-side defaults are applied first, then server config is spread on top.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9758-feat-add-server-side-PostHog-config-overrides-3206d73d36508123ad82c31187e69e49) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
+++ b/src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
@@ -101,6 +101,7 @@ export class PostHogTelemetryProvider implements TelemetryProvider {
         void import('posthog-js')
           .then((posthogModule) => {
             this.posthog = posthogModule.default
+            const serverConfig = remoteConfig.value?.posthog_config ?? {}
             this.posthog!.init(apiKey, {
               api_host:
                 window.__CONFIG__?.posthog_api_host || 'https://t.comfy.org',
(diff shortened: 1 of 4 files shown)


src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
```diff
@@ -109,9 +110,8 @@ export class PostHogTelemetryProvider implements TelemetryProvider {
               capture_pageview: false,
               capture_pageleave: false,
               persistence: 'localStorage+cookie',
-              debug:
-                window.__CONFIG__?.posthog_debug ??
-                import.meta.env.VITE_POSTHOG_DEBUG === 'true'
+              debug: import.meta.env.VITE_POSTHOG_DEBUG === 'true',
+              ...serverConfig
```
dante01yoon, line 114: `posthog_config` is only read once during `posthog.init()`, but `/features` is refreshed again after auth and later during polling. Those later config updates never reach the active PostHog instance because the `remoteConfig` watcher only updates `telemetry_disabled_events`.

If `posthog_config` is intended to be truly server-driven, I think we should reapply it when `remoteConfig` changes (for example via `posthog.set_config(...)` after init) and add a regression test for the post-init refresh case. If it is intentionally bootstrap-only, that contract should be made explicit.

robinjhuang: For this use case I think it's fine to update only when the user refreshes the page. Don't think posthog cleanly supports dynamically updating settings without page refresh anyways. WDYT?

robinjhuang: okay nvm, you can use `set_config` to dynamically update configuration. 

But for our use case I think updating on page refresh is sufficient.

dante01yoon: Got it. Make sense

dante01yoon (approved):

christian-byrne (approved):

PR #9759, opened 2026-03-12
fix: restore widget.inputEl backward compatibility for custom nodes

### LINKED ISSUE #12893: comfyui-frontend-package==1.41.15 Is breaking a lot of workflow and nodes.
### Custom Node Testing

### Expected Behavior

Load everything correctly.

### Actual Behavior

commit: abc87d36693b007bdbdab5ee753ccea6326acb34
Most of my workflow have this issue at load
Comfyui Cannot read properties of undefined (reading 'addEventListener')
They are completely different from each other but have the same bug.
Some custom nodes are loaded correctly but when I choose to add them to the canvas, nothing happen.
Downgrading to commit from 5 hours ago 980621da83267beffcb84839a27101b7092256e7 that use comfyui-frontend-package==1.39.19 fix the issue.

### Steps to Reproduce

Update to commit abc87d36693b007bdbdab5ee753ccea6326acb34 and try to open your custom workflows.

### Debug Logs

```powershell
no available
```

### Other

_No response_

dante01yoon:
스크린샷 2026-03-12 오전 9 55 12

See [issue](https://github.com/Comfy-Org/ComfyUI/issues/12893#issuecomment-4043141962) for reproduce 
## Summary

Restores `widget.inputEl` assignment on STRING multiline widgets that was removed in commit a7c211516 (PR #8594) when it was renamed to `widget.element`.

### Root cause

In PR #8594 (`WidgetValueStore`), the multiline STRING widget changed `widget.inputEl = inputEl` to `widget.element = inputEl`. This removed a property that custom node extensions treat as a de facto public API.

### Impact

[ComfyUI-Custom-Scripts](https://github.com/pythongosssss/ComfyUI-Custom-Scripts) (~3k stars, one of the most widely used custom node packs) patches `ComfyWidgets.STRING` globally to attach `TextAreaAutoComplete` to every STRING widget. The autocomplete setup accesses `widget.inputEl` to call `addEventListener` and set `readOnly`. Since the property is now `undefined`, this causes:

- `TypeError: Cannot read properties of undefined (reading 'addEventListener')`
- `TypeError: Cannot set properties of undefined (setting 'readOnly')`

Because the extension patches the **global** `ComfyWidgets.STRING` constructor, **all workflows fail to load** — including stock ComfyUI workflows — as long as the extension is installed. Users do not need to use any custom nodes in their workflow to be affected.

- Fixes Comfy-Org/ComfyUI#12893

## Test plan

- Install comfyui-custom-scripts and verify workflows with multiline STRING widgets load correctly
- Verify `widget.element` still works as before
- Verify `widget.inputEl` is accessible for backward compatibility

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
+++ b/src/renderer/extensions/vueNodes/widgets/composables/useStringWidget.ts
@@ -40,6 +40,7 @@ function addMultilineWidget(
   })
 
   widget.element = inputEl
+  widget.inputEl = inputEl
   widget.options.minNodeSize = [400, 200]
 
   inputEl.addEventListener('input', (event) => {


Myestery (approved):
LGTM
I was able to verify that this was the root cause

PR #9766, opened 2026-03-12
fix: prevent WebGLRenderer leak in app mode 3D preview

jtydhr88:
## Summary
Reuse the Load3d instance when switching between 3D results in app mode instead of creating a new WebGLRenderer each time. Add onUnmounted cleanup to Preview3d to release WebGL resources when the component is removed.

before

https://github.com/user-attachments/assets/c9818d10-941f-4994-9b48-2710c88454e7

after

https://github.com/user-attachments/assets/36361763-6800-4bc8-8089-14d64b7fcd16

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9766-fix-prevent-WebGLRenderer-leak-in-app-mode-3D-preview-3216d73d365081e19305d7255b71bc49) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useLoad3dViewer.ts
+++ b/src/composables/useLoad3dViewer.ts
@@ -398,7 +398,7 @@ export const useLoad3dViewer = (node?: LGraphNode) => {
       setupAnimationEvents()
     } catch (error) {
       console.error('Error initializing standalone 3D viewer:', error)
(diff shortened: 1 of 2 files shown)


AustinMroz (approved):
Thanks!

PR #9769, opened 2026-03-12
test: [DO NOT MERGE] verify URI drop regression test fails on main

christian-byrne:
## Purpose

This PR exists solely to verify that the new browser test for URI drops onto Vue nodes **fails on main** (without the fix from #9463).

If the Playwright test `Load workflow from URL dropped onto Vue node` fails, it proves the test is a valid regression test.

**Close this PR after verifying CI results. Do not merge.**

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9769-test-DO-NOT-MERGE-verify-URI-drop-regression-test-fails-on-main-3216d73d365081728cb7c1475928f1f9) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/fixtures/helpers/DragDropHelper.ts
+++ b/browser_tests/fixtures/helpers/DragDropHelper.ts
@@ -24,13 +24,15 @@ export class DragDropHelper {
       url?: string
       dropPosition?: Position
       waitForUpload?: boolean
+      preserveNativePropagation?: boolean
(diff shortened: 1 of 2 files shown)


christian-byrne:
Verification complete: the browser test correctly fails on main (without the fix), confirming it catches the regression. Closing as this was only for CI validation — the test lives on PR #9463.

PR #9786, opened 2026-03-12
feat: add focusedJobId for concurrent job execution support

christian-byrne:
## Summary

Add `focusedJobId` to `executionStore` to support concurrent job execution, where multiple jobs run simultaneously but only one is "focused" on the canvas.

## Changes

- **What**: New `focusedJobId` ref in `executionStore` that tracks which running job the user is viewing. `nodeProgressStates` stays in sync with the focused job. `isIdle` now checks `runningJobIds` (derived from `nodeProgressStatesByJob`) instead of `activeJobId`. Added `useConcurrentExecution` composable gated by remote config + user setting. Added `concurrent_execution_enabled` and `max_concurrent_jobs` to `RemoteConfig`. Added cloud settings for concurrent execution toggle and onboarding.
- **Dependencies**: None

## Review Focus

- `focusedJobId` vs `activeJobId` semantics: `activeJobId` = last job to receive `execution_start` from WS. `focusedJobId` = user-controlled "which job shows on canvas". Both exist because concurrent execution means multiple jobs run at once.
- `isIdle` change: now derived from `runningJobIds.length === 0` instead of `!activeJobId`. This is more correct for concurrent execution where one job can finish while others still run.
- `nodeProgressStatesByJob` stores per-job progress keyed by prompt_id, and `nodeProgressStates` is kept in sync with whichever job is focused.
- Cloud-only feature: gated by `concurrent_execution_enabled` from remote config + `Comfy.Cloud.ConcurrentExecution` user setting.

Part 1 of 2 — core store infrastructure. Consumer migration follows in #9785.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9786-feat-add-focusedJobId-for-concurrent-job-execution-support-3216d73d365081d89142d1c58a9f725f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/settings/constants/coreSettings.ts
+++ b/src/platform/settings/constants/coreSettings.ts
@@ -1268,5 +1268,21 @@ export const CORE_SETTINGS: SettingParams[] = [
     type: 'boolean',
     defaultValue: false,
     versionAdded: '1.42.0'
+  },
+  {
+    id: 'Comfy.Cloud.ConcurrentExecution',
+    name: 'Run jobs in parallel',
+    tooltip:
+      'When enabled, multiple workflow runs execute concurrently instead of queuing sequentially.',
+    type: isCloud ? 'boolean' : 'hidden',
+    defaultValue: true,
+    versionAdded: '1.42.0'
+  },
+  {
+    id: 'Comfy.Cloud.ConcurrentExecution.OnboardingSeen',
+    name: 'Concurrent execution onboarding dialog seen',
+    type: 'hidden',
+    defaultValue: false,
+    versionAdded: '1.42.0'
   }
 ]
--- a/src/platform/settings/constants/coreSettings.ts
+++ b/src/platform/settings/constants/coreSettings.ts
@@ -1274,7 +1274,7 @@ export const CORE_SETTINGS: SettingParams[] = [
     name: 'Run jobs in parallel',
     tooltip:
       'When enabled, multiple workflow runs execute concurrently instead of queuing sequentially.',
-    type: isCloud ? 'boolean' : 'hidden',
+    type: 'boolean',
     defaultValue: true,
     versionAdded: '1.42.0'
   },
--- a/src/stores/executionStore.ts
+++ b/src/stores/executionStore.ts
@@ -263,7 +263,7 @@ export const useExecutionStore = defineStore('execution', () => {
     // Auto-focus the first job, or if the current focused job is no longer running
     if (
       !focusedJobId.value ||
-      !nodeProgressStatesByJob.value[focusedJobId.value]
+      !queuedJobs.value[focusedJobId.value]
     ) {
       focusedJobId.value = activeJobId.value
     }
@@ -390,6 +390,7 @@ export const useExecutionStore = defineStore('execution', () => {
   }
 
   function handleProgress(e: CustomEvent<ProgressWsMessage>) {
+    if (e.detail.prompt_id !== focusedJobId.value) return
     _executingNodeProgress.value = e.detail
   }
 
@@ -540,6 +541,7 @@ export const useExecutionStore = defineStore('execution', () => {
 
   function setFocusedJob(jobId: string | null) {
     focusedJobId.value = jobId
+    _executingNodeProgress.value = null
     if (jobId) {
       nodeProgressStates.value = nodeProgressStatesByJob.value[jobId] ?? {}
     } else {
--- a/src/stores/executionStore.ts
+++ b/src/stores/executionStore.ts
@@ -60,6 +60,7 @@ export const useExecutionStore = defineStore('execution', () => {
 
   const clientId = ref<string | null>(null)
   const activeJobId = ref<string | null>(null)
+  const focusedJobId = ref<string | null>(null)
   const queuedJobs = ref<Record<NodeId, QueuedJob>>({})
   // This is the progress of all nodes in the currently executing workflow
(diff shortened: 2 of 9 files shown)


src/platform/settings/constants/coreSettings.ts
```diff
@@ -1257,5 +1257,21 @@ export const CORE_SETTINGS: SettingParams[] = [
     defaultValue: true,
     experimental: true,
     versionAdded: '1.40.0'
+  },
+  {
+    id: 'Comfy.Cloud.ConcurrentExecution',
+    name: 'Run jobs in parallel',
+    tooltip:
+      'When enabled, multiple workflow runs execute concurrently instead of queuing sequentially.',
+    type: isCloud ? 'boolean' : 'hidden',
```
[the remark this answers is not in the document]
christian-byrne: All settings in coreSettings.ts use hardcoded English for `name` and `tooltip` — there's no i18n pattern established for settings labels in this file. Will address i18n for all settings in a separate effort rather than introducing a one-off inconsistency here.

src/stores/executionStore.ts
```diff
@@ -300,16 +317,19 @@ export const useExecutionStore = defineStore('execution', () => {
...
+      // If we have progress for the currently executing node, update it for backwards compatibility
+      if (executingNodeId.value && nodes[executingNodeId.value]) {
+        const nodeState = nodes[executingNodeId.value]
+        _executingNodeProgress.value = {
+          value: nodeState.value,
+          max: nodeState.max,
+          prompt_id: nodeState.prompt_id,
+          node: nodeState.display_node_id || nodeState.node_id
+        }
       }
```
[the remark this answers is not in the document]
christian-byrne: Fixed in ad4b8734c — `handleProgress` now gates on `focusedJobId`, and `setFocusedJob` resets `_executingNodeProgress` when focus changes.

src/stores/executionStore.ts
```diff
@@ -170,20 +171,28 @@ export const useExecutionStore = defineStore('execution', () => {
...
   })
 
-  const isIdle = computed<boolean>(() => !activeJobId.value)
+  const isIdle = computed<boolean>(() => runningJobIds.value.length === 0)
 
   const nodesExecuted = computed<number>(() => {
-    if (!activeJob.value) return 0
-    return Object.values(activeJob.value.nodes).filter(Boolean).length
+    if (!focusedJob.value) return 0
+    return Object.values(focusedJob.value.nodes).filter(Boolean).length
```
benceruleanlu: Now that `nodesExecuted` / `executionProgress` are derived from `focusedJob`, the execution event handlers can no longer update completion state through `activeJob`. In the first concurrent case where job A stays focused and job B starts later, `execution_cached` / `executed` for A will either be ignored or applied to B because `activeJobId` has already moved on. That leaves the focused job’s total progress frozen and can corrupt the last-started job’s node map. These handlers need to update `queuedJobs[e.detail.prompt_id]` instead of `activeJob`.

benceruleanlu (changes requested):
Blocking concurrency regression left inline. `executionProgress` now reads from `focusedJob`, but completion events still mutate `activeJob`, so a focused job can stop advancing as soon as another job starts.

benceruleanlu:
https://github.com/Comfy-Org/ComfyUI_frontend/pull/9785 is closed, is it supposed to be in draft state?

christian-byrne:
> #9785 is closed, is it supposed to be in draft state?

Replaced by https://github.com/Comfy-Org/ComfyUI_frontend/pull/9789

src/stores/executionStore.ts
christian-byrne: fixed in d433fcf - both handleExecutionCached and handleExecuted now look up the job via queuedJobs[e.detail.prompt_id] instead of activeJob

src/stores/executionStore.ts
benceruleanlu, line 1: handleExecutionInterrupted and handleExecutionSuccess seem to mix e.detail.prompt_id and activeJobId, this looks like a bug

benceruleanlu (dismissed):
approach lgtm but needs to go with https://github.com/Comfy-Org/ComfyUI_frontend/pull/9789 and there are conflicts and what looks like a minor bug to me

src/stores/executionStore.ts
christian-byrne: good catch - both handlers were using activeJobId instead of e.detail.prompt_id, which would break during concurrent execution. fixed in bab3fa8: handleExecutionInterrupted now uses the event's prompt_id for clearInitializationByJobId, and handleExecutionSuccess uses it for telemetry tracking too.

benceruleanlu:
Taking a look now

benceruleanlu (commented):
Left one inline comment about the `executing` handler still clearing state using `activeJobId` even though concurrent execution now allows multiple live jobs.

src/stores/executionStore.ts
benceruleanlu, line 1: One thing I’d still want to resolve before concurrent execution rolls out: the `executing` path in this store is still effectively single-job. That event reaches the store without a `prompt_id`, so if two jobs overlap and an older job emits the terminal `executing` event first, we can only clear whatever is currently in `activeJobId`. It would be safer to either preserve `prompt_id` on `executing` or stop using that handler to tear down per-job state.

benceruleanlu (commented):
Left one file-level note on `executionStore.ts` about the `executing` path still clearing state via `activeJobId` in the concurrent model.

PR #9787, opened 2026-03-12
fix: resolve 'No inner node DTO found' error when muting subgraphs

dante01yoon:
## Summary

Fix "No inner node DTO found for id [X:Y:Z]" error that occurs when executing a workflow containing muted subgraphs or group nodes. This is a regression from v1.39.2 → v1.39.3 that blocks workflow execution for many users relying on muted subgraphs.

## Symptoms

- Red popup error: `"No inner node DTO found for id [outer:inner]"` when queueing a prompt
- Workflow fails to execute entirely
- Triggered when a muted (mode=NEVER) subgraph/group node has outputs connected to downstream nodes (e.g. context switch nodes)
- Workaround: using bypass instead of mute, or downgrading to frontend v1.39.2
- Affects custom nodes like rgthree context/context-switch, but is not limited to any specific custom node

## Root Cause

In `executionUtil.ts` (lines 68-73), `graphToPrompt` correctly skips registering inner nodes of muted subgraphs — muted nodes shouldn't execute, so their inner node DTOs are intentionally not added to `nodeDtoMap`:

```typescript
if (node.mode === LGraphEventMode.NEVER || node.mode === LGraphEventMode.BYPASS) {
  continue  // skips dto.getInnerNodes() registration
}
```

However, when a downstream node's input is connected to the muted node's output, `resolveOutput()` still tries to resolve into those inner nodes:
- `ExecutableNodeDTO._resolveSubgraphOutput()` looks up the inner node DTO in `nodesByExecutionId` → not found → throws at line 399
- `ExecutableGroupNodeDTO.resolveOutput()` calls `getInnerNodes()` → fails similarly at line 62

The fix adds an early return for NEVER mode in `resolveOutput()` of both classes, matching the existing pattern used for BYPASS mode. Muted nodes return `undefined` (no output), which is the correct semantic — they produce nothing.

## Changes

- **What**: Add `if (this.mode === LGraphEventMode.NEVER) return` early exit in `resolveOutput()` for both `ExecutableNodeDTO` and `ExecutableGroupNodeDTO`

## Review Focus

The fix mirrors how BYPASS mode is already handled — both short-circuit before reaching inner-node resolution. The only difference is that BYPASS delegates to `resolveInput()` (pass-through), while NEVER returns `undefined` (no output).

Fixes #8986

### DIFF
--- a/src/lib/litegraph/src/subgraph/ExecutableNodeDTO.test.ts
+++ b/src/lib/litegraph/src/subgraph/ExecutableNodeDTO.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, it, vi } from 'vitest'
 
 import {
   LGraph,
+  LGraphEventMode,
   LGraphNode,
   ExecutableNodeDTO
 } from '@/lib/litegraph/src/litegraph'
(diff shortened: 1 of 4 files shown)


christian-byrne (commented):
There is also this one: https://github.com/Comfy-Org/ComfyUI_frontend/pull/9302

dante01yoon:
Closing as duplicate of #9302 which has already been merged. The production code changes are identical and the fix is already in main.

PR #9808, opened 2026-03-12
fix: add deprecation warning for widget.inputEl on STRING multiline widgets

### LINKED ISSUE #12893: comfyui-frontend-package==1.41.15 Is breaking a lot of workflow and nodes.

christian-byrne:
## Summary

Add a deprecation warning when custom nodes access `widget.inputEl` on STRING multiline widgets, directing them to use `widget.element` instead.

## Changes

- **What**: Add a reusable `defineDeprecatedProperty` helper in `feedback.ts` that creates an ODP getter/setter proxy from a deprecated property to its replacement, logging via the existing `warnDeprecated` utility (deduplicates: warns once per unique message per session). Use it to deprecate `widget.inputEl` → `widget.element`.

## Review Focus

- `defineDeprecatedProperty` is generic and can be reused for future property deprecations across the codebase.
- `warnDeprecated` already handles deduplication via a `Set`, so heavy access patterns (e.g. custom nodes reading `widget.inputEl` in tight loops) won't spam.
- `enumerable: false` keeps the deprecated alias out of `Object.keys()` / `for...in` / `JSON.stringify`.

Fixes Comfy-Org/ComfyUI#12893


christian-byrne:
### Custom Node Ecosystem Impact Assessment

Searched the ComfyUI custom node ecosystem for `widget.inputEl` usage patterns. Found 50+ matches across multiple repos including:

- **rgthree-comfy**: `widget.inputEl.readOnly = true` (property access)
- **ComfyUI-AllergicPack**: `widget.inputEl.style`, `widget.inputEl.rows`, truthiness checks (`if (widget.inputEl)`)
- **ComfyUI-A1111-cond**: `widget.inputEl.style.display`, `widget.inputEl.readOnly`, caching to local var (`const textarea = widget.inputEl`)
- **comfyui-craftgear-nodes**: `const inputEl = widget.inputEl`, then using the cached reference
- **bv_nodepack**: Has a TS augmentation declaring `inputEl?: T` on the widget type

**All patterns are fully compatible with the ODP getter/setter approach.** The getter returns the real element, so property access, style mutation, truthiness checks, and caching to a local variable all work identically. The only behavioral difference is a one-time deprecation warning per session (deduplicated by `warnDeprecated`).

src/lib/litegraph/src/utils/feedback.ts
```diff
@@ -25,3 +25,34 @@ export function warnDeprecated(message: string, source?: object): void {
...
+    get() {
+      warnDeprecated(message)
+      return this[currentKey]
+    },
+    set(value: unknown) {
+      warnDeprecated(message)
+      this[currentKey] = value
+    },
+    configurable: true,
+    enumerable: true
```
dante01yoon: issue: The PR description and SKILL.md both claim:                                                 
  > enumerable: false keeps the deprecated alias out of Object.keys() / for...in /  JSON.stringify                                                                              
                                                                                              
  But the actual code at feedback.ts:55 has `enumerable: true`:                                 
   ```                                                                                           
      configurable: true,
      enumerable: true   // ← contradicts PR description and SKILL.md                         
   ```                                                                                           
  This should be enumerable: false to match the stated design intent. Deprecated aliases      
  shouldn't pollute enumeration.

dante01yoon (commented):
Suggestion: `defineDeprecatedProperty` is a new reusable utility with non-trivial behavior (getter/setter proxy, dedup via warnDeprecated). Per project guidelines ("Write tests for all changes"), it needs unit tests. Suggested coverage:                          
                                                                                              
  - Getter returns the current property's value                                               
  - Setter writes through to the current property
  - warnDeprecated is called on get and set                                                   
  - Deduplication works (only warns once)                         
  - Property is not enumerable (once enumerable: false is fixed)                              
  - Property is configurable

src/lib/litegraph/src/utils/feedback.ts
christian-byrne: fixed in 01fcfec59 - changed enumerable to false

dante01yoon (approved):

PR #9832, opened 2026-03-12
feat: bake frontend commit hash into build

huntcsg:
## Summary

Bake the frontend git commit hash into the build so it no longer needs to be fetched from the server via `/api/system_stats`.

## Changes

- **What**: Add `__COMFYUI_FRONTEND_COMMIT__` build-time constant (via Vite `define`) sourced from `git rev-parse HEAD` at build time. Falls back to `"unknown"` if git is unavailable. `SystemStatsPanel` uses this baked-in value for the "Frontend Version" row in cloud mode instead of the server-provided `comfyui_frontend_version` field.

## Testing
Confirmed to display the actual commit.
Screenshot 2026-03-12 at 7 09 52 PM

## Review Focus

- The `getDisplayValue` override for `comfyui_frontend_version` — cleanest way to swap the data source without restructuring the column system.
- No cloud-side changes needed: the `sync-frontend-build` workflow already checks out the frontend repo at the exact commit ref, so `git rev-parse HEAD` returns the correct hash.

### DIFF
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -1,6 +1,7 @@
 import { sentryVitePlugin } from '@sentry/vite-plugin'
 import tailwindcss from '@tailwindcss/vite'
 import vue from '@vitejs/plugin-vue'
+import { execSync } from 'child_process'
 import { config as dotenvConfig } from 'dotenv'
 import type { IncomingMessage, ServerResponse } from 'http'
 import { Readable } from 'stream'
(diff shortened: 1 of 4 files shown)


vite.config.mts
```diff
@@ -55,6 +56,16 @@ const DISTRIBUTION: 'desktop' | 'localhost' | 'cloud' =
 // Can be overridden via IS_NIGHTLY env var for testing
 const IS_NIGHTLY = process.env.IS_NIGHTLY === 'true'
 
+// Resolve git commit hash at build time: git → FRONTEND_COMMIT_HASH env var → 'unknown'
+let GIT_COMMIT = process.env.FRONTEND_COMMIT_HASH || ''
+if (!GIT_COMMIT) {
+  try {
+    GIT_COMMIT = execSync('git rev-parse HEAD').toString().trim()
```
christian-byrne: suggestion (non-blocking): `execSync` has no timeout, so if git hangs (e.g., a credential helper prompt on a misconfigured system), the build blocks indefinitely. Consider adding a timeout:
```js
GIT_COMMIT = execSync('git rev-parse HEAD', { timeout: 5000 }).toString().trim()
```

vite.config.mts
```diff
@@ -55,6 +56,16 @@ const DISTRIBUTION: 'desktop' | 'localhost' | 'cloud' =
 // Can be overridden via IS_NIGHTLY env var for testing
 const IS_NIGHTLY = process.env.IS_NIGHTLY === 'true'
 
+// Resolve git commit hash at build time: git → FRONTEND_COMMIT_HASH env var → 'unknown'
+let GIT_COMMIT = process.env.FRONTEND_COMMIT_HASH || ''
```
christian-byrne, line 63: question: Is the `FRONTEND_COMMIT_HASH` env var override intended for CI pipelines? If so, would it be worth documenting it (e.g., in a comment or README) so future contributors know about the escape hatch?

huntcsg: Documented it is for build environments that don't have git (e.g. a docker build container).

christian-byrne (approved):

PR #9849, opened 2026-03-13
fix: inline splash CSS to prevent SPA fallback breakage on cloud environments

viva-jinyi:
## Summary

Inline splash screen CSS into `index.html` to fix broken loading animation on cloud/ephemeral environments.

## Changes

- **What**: On cloud/ephemeral environments (e.g. `fe-pr-*.testenvs.comfy.org`), SPA fallback serves `index.html` for unknown paths. The `<link href="splash.css">` request resolves to `/cloud/splash.css`, which the server does not find as a static file — so it returns `index.html` with `200 OK`. The browser receives HTML instead of CSS, the CSS parser silently ignores it, and the splash screen renders without any styles or animations.
  - Inlined `splash.css` directly into `index.html` `<style>` block — eliminates the external request entirely
  - Moved `splash.css` to `src/assets/` for content-hashed Vite processing as source of truth
  - Removed `public/splash.css`

## Review Focus

- The inline CSS is byte-for-byte identical to the original `public/splash.css`
- `src/assets/splash.css` preserved as canonical source for future changes

[screen-capture (1).webm](https://github.com/user-attachments/assets/06729641-d1fd-47aa-9dd4-4acd28c2cfcf)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9849-fix-inline-splash-CSS-to-prevent-SPA-fallback-breakage-on-cloud-environments-3226d73d365081418741eb0944a74977) by [Unito](https://www.unito.io)


jaeone94 (approved):
LGTM !

dante01yoon (approved):
Note: It's worthwhile mention that root cause is happening from devops, not a frontend -- There should be handled by cloud which static file should be downloaded properly

PR #9850, opened 2026-03-13
fix: show download icon alongside file size in missing models dialog

viva-jinyi:
## Summary

Fix download icon not appearing when file size is successfully fetched in the missing models dialog.

## Changes

- **What**: Restructured the `v-if/v-else-if` chain in `MissingModelsContent.vue` so that file size and download icon render together instead of being mutually exclusive. Previously, a successful file size fetch would prevent the download button from rendering.

## Review Focus

The file size span and download/gated-link are now inside a shared `<template v-else-if="model.isDownloadable">` block. File size uses `v-if` (independent), while gated link and download button remain `v-if/v-else` (mutually exclusive with each other).

[screen-capture.webm](https://github.com/user-attachments/assets/f2f04d52-265b-4d05-992e-0ffe9bf64026)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9850-fix-show-download-icon-alongside-file-size-in-missing-models-dialog-3226d73d365081fd943bcfdedda87c73) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/dialog/content/MissingModelsContent.vue
+++ b/src/components/dialog/content/MissingModelsContent.vue
@@ -30,31 +30,33 @@
           </div>
           <div class="flex shrink-0 items-center gap-2">
             <Skeleton v-if="showSkeleton(model)" class="ml-1.5 h-4 w-12" />
-            <span
-              v-else-if="model.isDownloadable && fileSizes.get(model.url)"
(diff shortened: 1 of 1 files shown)


jaeone94 (approved):
LGTM

dante01yoon (approved):

PR #9851, opened 2026-03-13
feat: reactive upstream value display for disabled curve and imagecrop widgets

jtydhr88:
Replacement for https://github.com/Comfy-Org/ComfyUI_frontend/pull/9364

## Summary
Add a generic mechanism for widgets to reactively display upstream linked values when disabled, with concrete implementations for curve and imagecrop widgets.   

## Changes

 - What: When a widget input is linked to an upstream node, the widget enters a disabled state and displays the upstream node's current value reactively. This is built as a two-layer system:
 - Infrastructure layer: Resolve link origin info (originNodeId, originOutputName) in buildSlotMetadata, pass it through SimplifiedWidget.linkedUpstream, and provide a generic useUpstreamValue composable that reads upstream values from widgetValueStore.
 - Widget layer: Each widget type provides its own ValueExtractor to interpret upstream data. singleValueExtractor handles simple type-matched values (e.g. CurvePoint[]); boundsExtractor composes a Bounds object from either a single upstream widget or four individual x/y/width/height number widgets.
- Curve widget: shows upstream curve points in read-only mode
- ImageCrop widget: shows upstream bounding box with disabled crop handles and number inputs
- CurveEditor and WidgetBoundingBox: gain disabled prop support

## Adapting future widgets
The system is designed so that any widget needing upstream value display only needs to:
1. Accept widget: SimplifiedWidget as a prop (provides linkedUpstream automatically)
2. Call useUpstreamValue(() => widget.linkedUpstream, extractor) with a suitable extractor
3. Use singleValueExtractor(typeGuard) for single-value types, or write a custom ValueExtractor for composite cases like boundsExtractor            
4. Compute an effectiveValue that switches between upstream and local based on disabled state

No infrastructure changes are needed — linkedUpstream is already populated for all widget types that have a corresponding input slot.

## Review Focus
- The buildSlotMetadata helper is shared between extractVueNodeData and refreshNodeSlots — verify the graph ref is reliably available in both paths 
- boundsExtractor composing from 4 individual number widgets (x/y/width/height) — this handles the BBox→ImageCrop case where upstream exposes separate widgets rather than a single Bounds object

## Screenshots
https://github.com/user-attachments/assets/dbc57a44-c5df-44f0-acce-d347797ee8fb

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9851-feat-reactive-upstream-value-display-for-disabled-curve-and-imagecrop-widgets-3226d73d36508134b386ddc9b9f1266b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/curve/curveUtils.ts
+++ b/src/components/curve/curveUtils.ts
@@ -1,5 +1,18 @@
 import type { CurvePoint } from './types'
 
+export function isCurvePointArray(value: unknown): value is CurvePoint[] {
+  return (
+    Array.isArray(value) &&
+    value.every(
+      (p) =>
+        Array.isArray(p) &&
+        p.length === 2 &&
+        typeof p[0] === 'number' &&
+        typeof p[1] === 'number'
+    )
+  )
+}
+
 /**
  * Monotone cubic Hermite interpolation.
  * Produces a smooth curve that passes through all control points
--- a/src/components/imagecrop/WidgetImageCrop.vue
+++ b/src/components/imagecrop/WidgetImageCrop.vue
@@ -36,26 +36,37 @@
 
       <div
         v-if="imageUrl && !isLoading"
-        class="absolute box-content cursor-move border-2 border-white shadow-[0_0_0_9999px_rgba(0,0,0,0.5)]"
+        :class="
+          cn(
+            'absolute box-content cursor-move border-2 border-white shadow-[0_0_0_9999px_rgba(0,0,0,0.5)]',
+            isDisabled && 'pointer-events-none opacity-60'
+          )
+        "
         :style="cropBoxStyle"
         @pointerdown="handleDragStart"
         @pointermove="handleDragMove"
         @pointerup="handleDragEnd"
       />
 
-      <div
-        v-for="handle in resizeHandles"
-        v-show="imageUrl && !isLoading"
-        :key="handle.direction"
-        :class="['absolute', handle.class]"
-        :style="handle.style"
-        @pointerdown="(e) => handleResizeStart(e, handle.direction)"
-        @pointermove="handleResizeMove"
-        @pointerup="handleResizeEnd"
-      />
+      <template v-for="handle in resizeHandles" :key="handle.direction">
+        <div
+          v-show="imageUrl && !isLoading"
+          :class="
+            cn(
+              'absolute',
+              handle.class,
+              isDisabled && 'pointer-events-none opacity-60'
+            )
+          "
+          :style="handle.style"
+          @pointerdown="(e) => handleResizeStart(e, handle.direction)"
+          @pointermove="handleResizeMove"
+          @pointerup="handleResizeEnd"
+        />
+      </template>
     </div>
 
-    <div class="flex shrink-0 items-center gap-2">
+    <div v-if="!isDisabled" class="flex shrink-0 items-center gap-2">
       <label class="text-xs text-muted-foreground">
         {{ $t('imageCrop.ratio') }}
       </label>
@@ -90,12 +101,16 @@
       </Button>
     </div>
 
-    <WidgetBoundingBox v-model="modelValue" class="shrink-0" />
+    <WidgetBoundingBox
+      v-model="effectiveBounds"
+      :disabled="isDisabled"
+      class="shrink-0"
+    />
   </div>
 </template>
 
 <script setup lang="ts">
-import { useTemplateRef } from 'vue'
+import { computed, useTemplateRef } from 'vue'
 
 import WidgetBoundingBox from '@/components/boundingbox/WidgetBoundingBox.vue'
 import Button from '@/components/ui/button/Button.vue'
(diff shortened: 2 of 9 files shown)


src/composables/useUpstreamValue.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (outputName) {
+      const matched = widgets.find((w) => w.name === outputName)
+      if (matched && isValid(matched.value)) return matched.value
+    }
+    const valid = widgets.filter((w) => isValid(w.value))
+    return valid.length === 1 ? valid[0].value : undefined
+  }
+}
+
+function isBoundsObject(value: unknown): boolean {
```
christian-byrne, line 48: issue: `isBoundsObject` returns `boolean` instead of a type predicate, unlike `isCurvePointArray` which correctly returns `value is CurvePoint[]`. This forces downstream code to use `as Bounds` casts (e.g. `WidgetImageCrop.vue:151`).

Change signature to `function isBoundsObject(value: unknown): value is Bounds` and the cast in `WidgetImageCrop` becomes unnecessary.

src/components/curve/curveUtils.ts
```diff
@@ -1,5 +1,18 @@
 import type { CurvePoint } from './types'
 
+export function isCurvePointArray(value: unknown): value is CurvePoint[] {
```
christian-byrne, line 3: suggestion: `isCurvePointArray` accepts empty arrays as valid because `[].every(...)` returns `true`. The curve editor defaults to `[[0,0],[1,1]]` -- an empty array would render no curve and `createMonotoneInterpolator` falls back to `() => 0`. Consider adding `value.length >= 2` if a sub-2-point curve is invalid for display, or document that empty is intentionally valid.

src/components/imagecrop/WidgetImageCrop.vue
```diff
@@ -105,17 +120,41 @@ import SelectItem from '@/components/ui/select/SelectItem.vue'
...
 })
 
+const isDisabled = computed(() => !!widget.options?.disabled)
+
+const upstreamValue = useUpstreamValue(
+  () => widget.linkedUpstream,
+  boundsExtractor()
+)
+
+const effectiveBounds = computed({
```
christian-byrne, line 148: suggestion (non-blocking): `effectiveBounds` getter returns upstream values when disabled, but the setter always writes to `modelValue`. If anything writes while disabled, the write succeeds but is invisible (getter ignores it). Probably fine since handles have `pointer-events-none`, but a guard in the setter would make the intent explicit:

```ts
set: (v) => { if (!isDisabled.value) modelValue.value = v }
```

src/composables/useUpstreamValue.ts
```diff
@@ -0,0 +1,83 @@
...
+  )
+}
+
+export function boundsExtractor(): ValueExtractor {
+  const single = singleValueExtractor(isBoundsObject)
+  return (widgets, outputName) => {
+    const singleResult = single(widgets, outputName)
+    if (singleResult) return singleResult
+
+    const getNum = (name: string): number | undefined => {
```
christian-byrne, line 65: question: The fallback path in `boundsExtractor` looks up widgets by hardcoded names `'x'`, `'y'`, `'width'`, `'height'`. The PR mentions "BBox->ImageCrop" as the use case -- are these widget names guaranteed by BBox nodes, or could other upstream nodes use different naming (e.g. `left`/`top`/`w`/`h`)? If the contract is fixed, a brief comment noting the expected upstream widget names would help.

src/composables/useUpstreamValue.ts
```diff
@@ -0,0 +1,83 @@
+import { computed } from 'vue'
```
christian-byrne, line 1: suggestion (non-blocking): This composable has meaningful logic in the extractor factories (`singleValueExtractor`, `boundsExtractor`, `isBoundsObject`) that would be straightforward to unit test. Consider adding tests, especially for edge cases like missing widgets or partial bounds.

src/composables/useUpstreamValue.ts
```diff
@@ -0,0 +1,83 @@
...
+  type: string
+  value?: unknown
+}
+
+type ValueExtractor = (
+  widgets: UpstreamWidget[],
+  outputName: string | undefined
+) => unknown | undefined
+
+export function useUpstreamValue(
```
christian-byrne, line 18: nitpick (non-blocking): `useUpstreamValue` returns `ComputedRef<unknown>`, requiring `as CurvePoint[]` / `as Bounds` casts at every call site. A generic signature like `useUpstreamValue<T>(getLinkedUpstream, extractValue: (...) => T | undefined): ComputedRef<T | undefined>` would eliminate the casts. Also relates to the `isBoundsObject` type predicate comment above.

src/composables/useUpstreamValue.ts
```diff
@@ -0,0 +1,83 @@
+import { computed } from 'vue'
+
+import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
+import { useWidgetValueStore } from '@/stores/widgetValueStore'
+import type { LinkedUpstreamInfo } from '@/types/simplifiedWidget'
+
+interface UpstreamWidget {
```
christian-byrne, line 7: nitpick (non-blocking): `UpstreamWidget` has `{ name, type, value }` which looks like it mirrors `WidgetState` from `widgetValueStore`. Could this reuse the existing type instead?

christian-byrne (approved):

PR #9858, opened 2026-03-13
feat: display original image dimensions in cloud mode

viva-jinyi:
## Summary

In cloud mode, the Assets panel shows 512px thumbnail dimensions instead of the original image size because `/api/view` uses `res=512`.

This reads `X-Original-Width` and `X-Original-Height` response headers (added in Comfy-Org/cloud#2830) to display accurate original dimensions.

## Changes

- **`MediaImageTop.vue`**: In cloud mode, fetch the image URL and read original dimension headers before falling back to `naturalWidth`/`naturalHeight`.

## Dependencies

- Comfy-Org/cloud#2830 (backend: include original image dimensions in resized view response headers)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9858-feat-display-original-image-dimensions-in-cloud-mode-3226d73d36508195a1f0c68c43f707c9) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/assets/components/MediaImageTop.vue
+++ b/src/platform/assets/components/MediaImageTop.vue
@@ -20,6 +20,9 @@
 
 <script setup lang="ts">
 import { useImage, whenever } from '@vueuse/core'
+import { ref } from 'vue'
+
+import { isCloud } from '@/platform/distribution/types'
 
 import type { AssetMeta } from '../schemas/mediaAssetSchema'
 import { getAssetDisplayName } from '../utils/assetMetadataUtils'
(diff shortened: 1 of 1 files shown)


christian-byrne (commented):
This seems to somewhat defeat the purpose of having the thumbnails, as now we are still requesting the potentially massive image but then also requesting the thumbnail. Let's consider some ways to do this by having the backend play nice.

Also is the `src` being set to the thumbnail url in these cases - which would just yield the thumbnail dimensions always? I could be wrong about that but that is what it looks like at first glance.

christian-byrne:
https://comfy-organization.slack.com/archives/C08V9NDB3B4/p1773418577238849

PR #9865, opened 2026-03-13
fix: clear stale progress bar on SubgraphNode after navigation

christian-byrne:
## Problem

When navigating back from a subgraph to the root graph, the SubgraphNode can retain a stale progress bar. This happens because the progress watcher in `GraphCanvas.vue` watches `[nodeLocationProgressStates, canvasStore.canvas]`, but neither value changes reference during subgraph navigation:

- `nodeLocationProgressStates` is already `{}` (execution completed while viewing the subgraph)
- `canvasStore.canvas` is a `shallowRef` set once at startup — only `canvas.graph` changes (via `setGraph()`)

**Reproduction** (from PR #4382 comment thread by @guill):
1. Create a subgraph with a KSampler
2. Execute the workflow
3. While progress bar is halfway, enter the subgraph
4. Wait for execution to complete
5. Navigate back to root graph
6. Progress bar is stuck at 50%

## Root Cause

`canvasStore.canvas` is a `shallowRef` — subgraph navigation mutates `canvas.graph` (a nested property) via `LGraphCanvas.setGraph()`, which doesn't trigger a shallow watch. The watcher never re-fires to clear stale `node.progress` values.

## Fix

Add `canvasStore.currentGraph` to the watcher's dependency array. This is already a `shallowRef` in `canvasStore` that's updated on every `litegraph:set-graph` event. Zero overhead, precise targeting.

## Context

- Original discussion: https://github.com/Comfy-Org/ComfyUI_frontend/pull/4382/files/BASE..868e047272f6c5d710db7e607b8997d4c243490f#r2202024855
- PR #9248 correctly removed `deep: true` from this watcher but missed the subgraph edge case
- `deep: true` was the wrong fix — `canvasStore.currentGraph` is the precise solution

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9865-fix-clear-stale-progress-bar-on-SubgraphNode-after-navigation-3226d73d3650811c8f9de612d81ef98a) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/subgraphProgressClear.spec.ts
+++ b/browser_tests/tests/subgraphProgressClear.spec.ts
@@ -1,0 +1,77 @@
+import { expect } from '@playwright/test'
+
+import { comfyPageFixture as test } from '../fixtures/ComfyPage'
+
+test.describe(
+  'Subgraph progress clear on navigation',
+  { tag: ['@subgraph'] },
+  () => {
+    test('Stale progress is cleared on subgraph node after navigating back', async ({
+      comfyPage
+    }) => {
+      await comfyPage.workflow.loadWorkflow('subgraphs/basic-subgraph')
+      await comfyPage.nextFrame()
+
+      // Find the subgraph node
+      const subgraphNodeId = await comfyPage.page.evaluate(() => {
+        const graph = window.app!.canvas.graph!
+        const subgraphNode = graph.nodes.find(
+          (n) => typeof n.isSubgraphNode === 'function' && n.isSubgraphNode()
+        )
+        return subgraphNode ? String(subgraphNode.id) : null
+      })
+      expect(subgraphNodeId).not.toBeNull()
+
+      // Simulate a stale progress value on the subgraph node.
+      // This happens when:
+      //   1. User views root graph during execution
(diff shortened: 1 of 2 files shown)


browser_tests/tests/subgraphProgressClear.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+      const subgraphNode = await comfyPage.nodeOps.getNodeRefById(
+        subgraphNodeId!
+      )
+      await subgraphNode.navigateIntoSubgraph()
+
+      // Verify we're inside the subgraph
+      const inSubgraph = await comfyPage.page.evaluate(() => {
+        const graph = window.app!.canvas.graph
+        return !!graph && 'inputNode' in graph
+      })
```
DrJKL, line 55: There's a lot of evaluates in here. Possible target for extraction/abstraction?

DrJKL (approved):

browser_tests/tests/subgraphProgressClear.spec.ts
christian-byrne: @coderabbitai add a tracking ISsue to do this

PR #9840, opened 2026-03-12
feat: show ComfyUI context menu on textarea widget right-click

christian-byrne:
## Summary

Right-clicking a textarea widget (e.g. text node) shows the browser's native context menu instead of ComfyUI's context menu, preventing access to promote/un-promote options in subgraphs.

## Changes

- **What**: Replace `@contextmenu.capture.stop` on `WidgetTextarea.vue`'s `<Textarea>` with a handler implementing double-right-click toggling: first right-click shows ComfyUI's context menu, second right-click (while menu is open) allows browser native menu. Exposes `isNodeOptionsOpen()` from `useMoreOptionsMenu.ts` to check menu state.

## Review Focus

The capture-phase handler in `WidgetTextarea.vue` only changes `contextmenu` handling — pointer event modifiers (`pointerdown/move/up.capture.stop`) that prevent canvas panning are untouched. The double-right-click pattern matches Notion/YouTube behavior for editable text fields.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9840-fix-show-ComfyUI-context-menu-on-textarea-widget-right-click-3216d73d36508102b4c9c13a5915bc48) by [Unito](https://www.unito.io)

PR #9891, opened 2026-03-13
test: add browser test for textarea right-click context menu in subgraph

christian-byrne:
## Summary

Add E2E test coverage for the textarea widget right-click context menu inside subgraphs.

The fix was shipped in #9840 — this PR adds the missing browser test.

## Test

- Loads a subgraph workflow with a CLIPTextEncode (textarea) node
- Navigates into the subgraph
- Right-clicks the textarea DOM element
- Asserts that the ComfyUI "Promote Widget" context menu option appears

## Related

- Fixes the test gap from #9840
- Notion ticket: d7a53160-e1e1-42bb-a5ac-c0c2702c629c

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9891-test-add-browser-test-for-textarea-right-click-context-menu-in-subgraph-3226d73d365081a4be51f89b5d505361) by [Unito](https://www.unito.io)


browser_tests/tests/subgraphPromotion.spec.ts
```diff
@@ -405,6 +405,46 @@ test.describe(
...
+        await comfyPage.canvas.click({
+          position: widgetPos,
+          button: 'right',
+          force: true
+        })
+        await comfyPage.nextFrame()
+
+        const promoteEntry = comfyPage.page
+          .locator('.litemenu-entry, .p-contextmenu .p-menuitem-text')
+          .filter({ hasText: /Promote Widget/ })
```
Myestery: /Promote Widget/ also matches Un-Promote Widget, so this assertion can pass on the wrong menu state.

Myestery (dismissed):
Approved with a nit

browser_tests/tests/subgraphPromotion.spec.ts
christian-byrne: @coderabbitai add a tracking Issue to fix this

christian-byrne (commented):
There's one failing test. it likely needs a logic change to account for new feature

browser_tests/tests/subgraphPromotion.spec.ts
christian-byrne: good catch - tracking issue created in #9898, will tighten the regex in a followup

Myestery (dismissed):

PR #9893, opened 2026-03-13
fix: reconcile stale progress after reconnect

benceruleanlu:
## Summary

Reconcile stale execution state after reconnect so the run progress UI clears when the active job finished while the client was disconnected.

## Changes

- **What**: Add queue/history-based reconnect reconciliation for missed terminal websocket events, factor targeted tracked-job cleanup into a shared execution-store removal helper used by both normal resets and reconnect repair, and add regression coverage for the stale progress cleanup path.
- **Dependencies**: None

## Review Focus

- Recovery intentionally uses `active job missing from queue + same job present in history`, not `queue is empty`, because websocket `status` only carries `queue_remaining`.
- `api.getQueue()` / `fetchJobsRaw()` still collapse fetch failures into empty snapshots. That is a code smell and future refactor target; this change keeps the bug fix narrow by relying on positive history evidence instead of trusting empty queue alone.
- Shared cleanup is now centralized in `removeTrackedJobState()` so `resetExecutionState()` and reconnect repair cannot drift.

N/A

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9893-fix-reconcile-stale-progress-after-reconnect-3226d73d365081cbba96ef022635cc46) by [Unito](https://www.unito.io)


src/stores/executionStore.ts
```diff
@@ -433,24 +433,73 @@ export const useExecutionStore = defineStore('execution', () => {
...
+    if (
+      options?.clearAllNodeProgressStates ||
+      _executingNodeProgress.value?.prompt_id === normalizedJobId
+    ) {
+      _executingNodeProgress.value = null
+    }
+
+    if (wasActiveJob) {
+      activeJobId.value = null
+      if (options?.clearPromptError) {
```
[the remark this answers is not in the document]
benceruleanlu: Addressed in `ea691eaea`.

The mismatch case was real: `resetExecutionState(promptId)` could globally clear progress while leaving a stale `activeJobId` behind when the terminal event arrived for a different prompt after reconnect. The fix now clears the terminal prompt state and, if needed, also removes the mismatched stale active job in the same reset path. I also added a regression test covering that reconnect scenario.

benceruleanlu:
Addressed the review feedback in `ea691eaea`.

- Fixed the stale `activeJobId` mismatch in `resetExecutionState()` when a terminal event arrives for a different prompt after reconnect.
- Added a regression test for that reconnect path.
- Also took the small `queueStore` simplification nit and removed the unnecessary `String(activeJobId)` conversion.

Validation:
- `pnpm test:unit -- src/stores/executionStore.test.ts src/stores/queueStore.test.ts`
- `pnpm exec eslint src/stores/executionStore.ts src/stores/queueStore.ts src/stores/executionStore.test.ts src/stores/queueStore.test.ts`
- `pnpm typecheck`

src/stores/queueStore.ts
```diff
@@ -580,6 +581,17 @@ export const useQueueStore = defineStore('queue', () => {
...
+
+      const activeJobId = executionStore.activeJobId
+      const missedTerminalExecution =
+        activeJobId !== null &&
+        !activeJobIds.has(activeJobId) &&
+        nextHistoryTasks.some((task) => task.jobId === activeJobId)
+
+      if (missedTerminalExecution) {
+        executionStore.clearTrackedJob(activeJobId)
+      }
```
christian-byrne, line 593: issue: The `missedTerminalExecution` check is not guarded by `queueHasData`. When `getQueue()` returns empty arrays on a transient error, `activeJobIds` is an empty Set, so `!activeJobIds.has(activeJobId)` is trivially true. Combined with a successful history fetch, this would incorrectly call `clearTrackedJob` for a job that is actually still running.

The comment on lines 548-550 explicitly warns about this pattern for `reconcileInitializingJobs` -- the same guard should apply here.

Wrapping in `if (queueHasData)` or adding `queueHasData` to the condition would fix it.

src/stores/executionStore.ts
```diff
@@ -488,25 +488,86 @@ export const useExecutionStore = defineStore('execution', () => {
...
-    executionErrorStore.clearPromptError()
+  }
+
+  /**
+   * Clear tracked execution state for a single job.
+   * Used to recover from missed terminal websocket events after reconnects.
+   */
+  function clearTrackedJob(jobId: string) {
+    removeTrackedJobState(jobId)
   }
```
christian-byrne, line 571: issue: `clearTrackedJob` calls `removeTrackedJobState`, which does not clear `executionIdToLocatorCache`. Compare with `resetExecutionState` (line 537) which calls `executionIdToLocatorCache.clear()` at the top. After reconnect reconciliation, stale cache entries could map wrong `NodeLocatorId` values if a new job reuses node execution IDs.

Consider adding `executionIdToLocatorCache.clear()` in the `wasActiveJob` branch of `removeTrackedJobState`, or directly in `clearTrackedJob`.

src/stores/queueStore.ts
```diff
@@ -580,6 +581,17 @@ export const useQueueStore = defineStore('queue', () => {
       if (!isHistoryUnchanged) {
         historyTasks.value = nextHistoryTasks
       }
+
+      const activeJobId = executionStore.activeJobId
```
christian-byrne, line 585: suggestion: `activeJobId` is read after `Promise.all([getQueue(), getHistory()])` resolves. During the async gap, a websocket `execution_start` event could update `activeJobId` to a newly-started job. The history check partially mitigates this (a just-started job wouldn't be in history), but snapshotting before the async work would be safer:

```ts
const snapshotActiveJobId = executionStore.activeJobId
// ... before Promise.all or immediately after the stale-request guard
```

src/stores/queueStore.ts
```diff
@@ -580,6 +581,17 @@ export const useQueueStore = defineStore('queue', () => {
...
+      const missedTerminalExecution =
+        activeJobId !== null &&
+        !activeJobIds.has(activeJobId) &&
+        nextHistoryTasks.some((task) => task.jobId === activeJobId)
```
christian-byrne, line 589: question (non-blocking): `nextHistoryTasks` is capped at `maxHistoryItems` (default 64). If a user is disconnected long enough for >64 jobs to complete, the stale job's history entry would be evicted and reconciliation would silently fail forever -- the UI would show a permanently "running" job. Is that an acceptable tradeoff, or would a fallback (e.g., clearing jobs older than some threshold) be worth considering?

src/stores/executionStore.ts
```diff
@@ -488,25 +488,86 @@ export const useExecutionStore = defineStore('execution', () => {
...
+
+    nodeProgressStates.value = options?.clearAllNodeProgressStates
+      ? {}
+      : Object.fromEntries(
+          Object.entries(nodeProgressStates.value).filter(
+            ([_, state]) => state.prompt_id !== normalizedJobId
+          )
+        )
+
+    delete queuedJobs.value[normalizedJobId]
```
christian-byrne, line 515: nitpick (non-blocking): Minor inconsistency within `removeTrackedJobState` -- `nodeProgressStatesByJob` uses spread-and-replace for reactivity (lines 503-505) while `queuedJobs` uses direct `delete` (line 515). Both work in Vue 3's reactive system, but the mixed strategy in a single function is slightly confusing.

src/stores/queueStore.test.ts
```diff
@@ -564,6 +565,36 @@ describe('useQueueStore', () => {
       // Should preserve array identity when history is unchanged
       expect(store.historyTasks).toBe(initialHistoryTasks)
     })
+
+    it('clears stale execution tracking when the active job appears in history', async () => {
+      const executionStore = useExecutionStore()
+      const clearTrackedJobSpy = vi.spyOn(executionStore, 'clearTrackedJob')
```
christian-byrne, line 571: suggestion (non-blocking): The tests cover the case where the active job appears in history (cleared) and where it's missing from history (not cleared), but there's no test for the normal case: active job is still present in the `Running` queue. That's the path most directly related to the `queueHasData` guard issue above -- a regression there could falsely clear a running job.

src/stores/executionStore.test.ts
```diff
@@ -430,6 +430,108 @@ describe('useExecutionStore - reconcileInitializingJobs', () => {
...
+      'node-2': activeNodeState
+    })
+    expect(store.nodeProgressStatesByJob).toEqual({
+      'job-2': { 'node-2': activeNodeState }
+    })
+    expect(store.initializingJobIds).toEqual(new Set(['job-2']))
+    expect(store._executingNodeProgress).toBeNull()
+  })
+
+  it('clears a stale active job when a different prompt completes', () => {
```
christian-byrne, line 495: nitpick (non-blocking): This test ("clears a stale active job when a different prompt completes") fires `execution_success`, which routes through `handleExecutionSuccess` -> `resetExecutionState`, not `clearTrackedJob`. The test is valid but the suite name is misleading -- it's really testing the `resetExecutionState` path for stale active jobs.

benceruleanlu:
Closing as superseded by #12067, which merged the same reconnect recovery into `main` on May 11.

#12067 handles the missed-terminal-event window at the actual WebSocket `reconnected` boundary: it refreshes the authoritative queue, preserves the previous queue snapshot on fetch failure, and clears stale execution state through the existing complete reset path. It also includes unit and browser coverage for both sides of the behavior—clearing finished-job progress and preserving a job that is still running—and was backported to the 1.44 release branches.

There is no unique behavior left in this PR worth carrying forward. Its queue-plus-history approach is now both redundant and weaker: the branch is conflicting with current `main`, history is capped, and the last review identified unresolved false-clear/cache-cleanup concerns. Current focused tests for the merged implementation pass (158 passed, 1 skipped across the reconnect, execution-store, and queue-store suites).

PR #9750, opened 2026-03-11
feat: add Nodes 2.0 alignment tools

benceruleanlu:
## Summary

Add Nodes 2.0 drag alignment snapping and expose richer multi-node arrangement actions from the right-click menu.

## Changes

- **What**: add an opt-in `Comfy.Canvas.AlignNodesWhileDragging` setting, snap Vue-node drags to nearby rendered node edges and centers with visible guides, add horizontal/vertical center alignment actions, and gate arrangement actions to the multi-node right-click menu instead of the selection-toolbox More Options menu.
- **Dependencies**: none

## Review Focus

- Rendered-bounds snapping behavior during Vue-node drag
- Context-menu-only gating for alignment/distribution actions
- Drag-guide lifecycle and grid-snap precedence

## Verification

- `pnpm test:unit -- src/renderer/extensions/vueNodes/layout/nodeAlignmentSnap.test.ts src/renderer/extensions/vueNodes/layout/useNodeDrag.test.ts src/composables/graph/useMoreOptionsMenu.test.ts src/composables/graph/useSelectionMenuOptions.test.ts`
- `pnpm typecheck`
- `pnpm typecheck:browser`
- `pnpm format:check`
- `pnpm lint`
- Added a Playwright spec for the Vue-node alignment path; local execution was blocked here because `http://localhost:8188` was unavailable (`ECONNREFUSED` on `/api/users`).

## Screenshots (if applicable)

- None

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9750-feat-add-Nodes-2-0-alignment-tools-3206d73d3650811dbc4dfd87f37c30cd) by [Unito](https://www.unito.io)

PR #9894, opened 2026-03-13
feat: add Nodes 2.0 drag snapping

benceruleanlu:
## Summary
- add an opt-in `Comfy.Canvas.AlignNodesWhileDragging` setting for Nodes 2.0
- snap dragged Vue-node selections to nearby unselected node edges and centers
- render transient alignment guides during drag and clear them on drag end/cancel
- prune snap candidates by proximity using the layout spatial index
- add scoped unit coverage plus a browser spec for drag snapping only

## Scope
This PR is intentionally limited to drag snapping. It does not include the broader alignment/distribution context-menu work from the earlier draft.

Supersedes #9750 for review purposes.

## Implementation Notes
- snapping uses rendered node bounds, including the title bar, so guides align with the visible Vue node box
- the snap threshold remains zoom-normalized (`8px / zoom`)
- candidate lookup is now proximity-based rather than scanning all unselected nodes each drag frame
- guide rendering is gated to Vue mode only and guide-only updates only dirty the foreground canvas

## Verification
- `pnpm install`
- `pnpm typecheck`
- `pnpm typecheck:browser`
- `pnpm vitest run src/renderer/extensions/vueNodes/layout/nodeAlignmentSnap.test.ts src/renderer/extensions/vueNodes/layout/useNodeDrag.test.ts src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts`
- `pnpm exec eslint src/components/rightSidePanel/settings/TabGlobalSettings.vue src/lib/litegraph/src/LGraphCanvas.ts src/platform/settings/constants/coreSettings.ts src/renderer/core/layout/store/layoutStore.ts src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts src/renderer/extensions/vueNodes/layout/nodeAlignmentSnap.ts src/renderer/extensions/vueNodes/layout/nodeAlignmentSnap.test.ts src/renderer/extensions/vueNodes/layout/useNodeDrag.ts src/renderer/extensions/vueNodes/layout/useNodeDrag.test.ts src/schemas/apiSchema.ts browser_tests/tests/vueNodes/interactions/node/dragSnapping.spec.ts --no-cache`
- `pnpm exec oxfmt --check src/components/rightSidePanel/settings/TabGlobalSettings.vue src/lib/litegraph/src/LGraphCanvas.ts src/platform/settings/constants/coreSettings.ts src/renderer/core/layout/store/layoutStore.ts src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.test.ts src/renderer/extensions/vueNodes/layout/nodeAlignmentSnap.ts src/renderer/extensions/vueNodes/layout/nodeAlignmentSnap.test.ts src/renderer/extensions/vueNodes/layout/useNodeDrag.ts src/renderer/extensions/vueNodes/layout/useNodeDrag.test.ts src/schemas/apiSchema.ts browser_tests/tests/vueNodes/interactions/node/dragSnapping.spec.ts src/locales/en/settings.json`

## Notes
- I did not run the Playwright spec here because the local ComfyUI server dependency was unavailable in this environment.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9894-feat-add-nodes-2-drag-snapping-3226d73d3650813a9957e76706e34b5c) by [Unito](https://www.unito.io)

### DIFF
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -5007,6 +5007,7 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
 
       if (!LiteGraph.vueNodesMode || !this.overlayCtx) {
         this._drawConnectingLinks(ctx)
+        this._drawVueDragAlignmentGuides(ctx)
       } else {
         this._drawOverlayLinks()
       }
@@ -5111,7 +5112,8 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
 
     octx.clearRect(0, 0, overlayCanvas.width, overlayCanvas.height)
 
-    if (!this.linkConnector.isConnecting) return
+    const hasDragGuides = layoutStore.vueDragSnapGuides.value.length > 0
+    if (!this.linkConnector.isConnecting && !hasDragGuides) return
 
     octx.save()
 
@@ -5120,9 +5122,37 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
 
     this.ds.toCanvasContext(octx)
 
-    this._drawConnectingLinks(octx)
+    if (this.linkConnector.isConnecting) {
+      this._drawConnectingLinks(octx)
+    }
+    this._drawVueDragAlignmentGuides(octx)
 
     octx.restore()
+  }
+
+  private _drawVueDragAlignmentGuides(ctx: CanvasRenderingContext2D): void {
+    const guides = layoutStore.vueDragSnapGuides.value
+    if (!guides.length) return
+
+    const scale = this.ds.scale || 1
+    ctx.save()
+    ctx.beginPath()
+    ctx.lineWidth = 1 / scale
+    ctx.strokeStyle = '#ff4d4f'
+    ctx.setLineDash([6 / scale, 4 / scale])
+
+    for (const guide of guides) {
+      if (guide.axis === 'vertical') {
+        ctx.moveTo(guide.coordinate, guide.start)
+        ctx.lineTo(guide.coordinate, guide.end)
+      } else {
+        ctx.moveTo(guide.start, guide.coordinate)
+        ctx.lineTo(guide.end, guide.coordinate)
+      }
+    }
+
+    ctx.stroke()
+    ctx.restore()
   }
 
   /** Get the target snap / highlight point in graph space */
--- a/src/lib/litegraph/src/LGraphCanvas.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.ts
@@ -5007,7 +5007,9 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
 
       if (!LiteGraph.vueNodesMode || !this.overlayCtx) {
         this._drawConnectingLinks(ctx)
-        this._drawVueDragAlignmentGuides(ctx)
+        if (LiteGraph.vueNodesMode) {
+          this._drawVueDragAlignmentGuides(ctx)
+        }
       } else {
         this._drawOverlayLinks()
       }
--- a/src/renderer/core/layout/store/layoutStore.ts
+++ b/src/renderer/core/layout/store/layoutStore.ts
@@ -9,7 +9,6 @@ import { computed, customRef, ref } from 'vue'
 import type { ComputedRef, Ref } from 'vue'
 import * as Y from 'yjs'
 
-import type { NodeAlignmentGuide } from '@/renderer/extensions/vueNodes/layout/nodeAlignmentSnap'
 import { removeNodeTitleHeight } from '@/renderer/core/layout/utils/nodeSizeUtil'
 
 import { ACTOR_CONFIG } from '@/renderer/core/layout/constants'
@@ -31,6 +30,7 @@ import type {
   LinkSegmentLayout,
   MoveNodeOperation,
   MoveRerouteOperation,
+  NodeAlignmentGuide,
   NodeBoundsUpdate,
   NodeId,
   NodeLayout,
--- a/src/renderer/core/layout/store/layoutStore.ts
+++ b/src/renderer/core/layout/store/layoutStore.ts
@@ -9,6 +9,7 @@ import { computed, customRef, ref } from 'vue'
 import type { ComputedRef, Ref } from 'vue'
 import * as Y from 'yjs'
 
+import type { NodeAlignmentGuide } from '@/renderer/extensions/vueNodes/layout/nodeAlignmentSnap'
 import { removeNodeTitleHeight } from '@/renderer/core/layout/utils/nodeSizeUtil'
 
 import { ACTOR_CONFIG } from '@/renderer/core/layout/constants'
@@ -144,6 +145,7 @@ class LayoutStoreImpl implements LayoutStore {
 
   // Vue dragging state for selection toolbox (public ref for direct mutation)
   public isDraggingVueNodes = ref(false)
+  public vueDragSnapGuides = ref<NodeAlignmentGuide[]>([])
   // Vue resizing state to prevent drag from activating during resize
   public isResizingVueNodes = ref(false)
 
--- a/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
+++ b/src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
@@ -12,7 +12,7 @@ import { useNodeDrag } from '@/renderer/extensions/vueNodes/layout/useNodeDrag'
 export function useNodePointerInteractions(
   nodeIdRef: MaybeRefOrGetter<string>
 ) {
-  const { startDrag, endDrag, handleDrag } = useNodeDrag()
+  const { startDrag, endDrag, handleDrag, cancelDrag } = useNodeDrag()
   // Use canvas interactions for proper wheel event handling and pointer event capture control
   const { forwardEventToCanvas, shouldHandleNodePointerEvents } =
     useCanvasInteractions()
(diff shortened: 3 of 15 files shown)


src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -5007,6 +5007,7 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
 
       if (!LiteGraph.vueNodesMode || !this.overlayCtx) {
         this._drawConnectingLinks(ctx)
+        this._drawVueDragAlignmentGuides(ctx)
```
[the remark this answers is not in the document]
benceruleanlu: Fixed in `94fb623fc`. The core issue was real: the fallback path could still call `_drawVueDragAlignmentGuides(ctx)` when `vueNodesMode` was false. I gated that call to Vue mode only. I did not gate it on overlay availability, since we still want guides to render in Vue mode when `overlayCtx` is unavailable and we fall back to drawing on the main canvas.

src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
```diff
@@ -120,11 +139,12 @@ function useNodeDragIndividual() {
         x: canvasWithDelta.x - canvasOrigin.x,
         y: canvasWithDelta.y - canvasOrigin.y
       }
+      const snappedCanvasDelta = maybeResolveAlignmentSnap(event, canvasDelta)
```
christian-byrne, line 150: question (non-blocking): `handleDrag` returns early when a rAF is already pending (line 117), which means subsequent `pointermove` events are dropped and the RAF callback closes over the first event's `clientX`/`clientY` rather than the latest. During fast mouse movement this can cause positional lag.

This is a pre-existing pattern not introduced by this PR -- is there a reason the latest event isn't stored and read inside the RAF callback? e.g. capturing `latestEvent = event` and reading from that inside `requestAnimationFrame`.

src/renderer/core/layout/store/layoutStore.ts
```diff
@@ -9,6 +9,7 @@ import { computed, customRef, ref } from 'vue'
 import type { ComputedRef, Ref } from 'vue'
 import * as Y from 'yjs'
 
+import type { NodeAlignmentGuide } from '@/renderer/extensions/vueNodes/layout/nodeAlignmentSnap'
```
christian-byrne: issue: `renderer/core` imports a type from `renderer/extensions`:
```ts
import type { NodeAlignmentGuide } from '@/renderer/extensions/vueNodes/layout/nodeAlignmentSnap'
```
This creates an upward dependency from the core layer into an extension. Consider moving the `NodeAlignmentGuide` interface to `renderer/core/layout/types.ts` (where `Bounds`, `Point`, etc. already live) so the dependency arrow stays `extensions -> core`.

src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
```diff
@@ -250,4 +273,141 @@ function useNodeDragIndividual() {
...
+function expandBounds(bounds: Bounds, padding: number): Bounds {
+  return {
+    x: bounds.x - padding,
+    y: bounds.y - padding,
+    width: bounds.width + padding * 2,
+    height: bounds.height + padding * 2
+  }
+}
+
+function translateBounds(bounds: Bounds, delta: Point): Bounds {
```
christian-byrne: suggestion: `translateBounds` is duplicated identically here and in `nodeAlignmentSnap.ts:219`. Extract it to a shared utility (e.g. alongside `expandBounds`/`mergeBounds` or into a geometry utils file) to keep a single source of truth.

src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
```diff
@@ -250,4 +273,141 @@ function useNodeDragIndividual() {
...
+    event: PointerEvent,
+    canvasDelta: Point
+  ): Point {
+    if (!settingStore.get('Comfy.Canvas.AlignNodesWhileDragging')) {
+      updateDragSnapGuides([])
+      return canvasDelta
+    }
+
+    if (
+      shouldSnap(event) ||
```
christian-byrne: nitpick (non-blocking): Inside `maybeResolveAlignmentSnap`, `shouldSnap(event)` reads as "should do snapping" but it actually checks for grid snap -- so `true` here *disables* alignment snapping. A brief inline comment (e.g. `// Grid snap active -- skip alignment snapping`) would clarify the intent for future readers.

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
```diff
@@ -102,6 +102,7 @@ export function useNodePointerInteractions(
 
   function cleanupDragState() {
     layoutStore.isDraggingVueNodes.value = false
```
christian-byrne, line 104: question (non-blocking): `onContextmenu` (line 166) calls `cleanupDragState()` instead of `endDrag()`, which means `stopShiftSync` listeners are never torn down. This is a pre-existing pattern, but this PR adds `vueDragSnapGuides` cleanup here without the corresponding `endDrag` call -- so the snap guides are cleared but shift listeners still leak. Would it make sense to call `safeDragEnd` here (similar to `onPointercancel`) to clean up everything?

src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
```diff
@@ -250,4 +273,141 @@ function useNodeDragIndividual() {
...
+        continue
+      }
+
+      candidates.push(getRenderedNodeBounds(layout))
+    }
+
+    return candidates
+  }
+
+  function updateDragSnapGuides(guides: NodeAlignmentGuide[]) {
```
christian-byrne, line 348: suggestion (non-blocking): `updateDragSnapGuides([])` unconditionally calls `setDirty(true)` even when guides are already empty. Consider guarding with a length check to avoid redundant canvas repaints:
```ts
function updateDragSnapGuides(guides: NodeAlignmentGuide[]) {
  if (guides.length === 0 && layoutStore.vueDragSnapGuides.value.length === 0) return
  layoutStore.vueDragSnapGuides.value = guides
  canvasStore.canvas?.setDirty(true)
}
```

src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
benceruleanlu: good point that was just the old shape I made it use the latest event now

src/renderer/core/layout/store/layoutStore.ts
benceruleanlu: true! merged it into types

src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
benceruleanlu: done, moved to geometry.ts in utils folder

src/renderer/core/layout/store/layoutStore.ts
artokun: Noting this here for coordination: the same core->extensions layering issue also shows up in [#10224](https://github.com/Comfy-Org/ComfyUI_frontend/pull/10224). The concrete spots are [`useSubgraphDragBridge.ts` lines 19-21](https://github.com/Comfy-Org/ComfyUI_frontend/blob/01af1a39653a0f728ba7273df4fd4881adae14af/src/renderer/core/canvas/links/useSubgraphDragBridge.ts#L19-L21), where `renderer/core` imports runtime helpers from `renderer/extensions/vueNodes/*`, plus the runtime hookup in [`GraphCanvas.vue` line 166](https://github.com/Comfy-Org/ComfyUI_frontend/blob/01af1a39653a0f728ba7273df4fd4881adae14af/src/components/graph/GraphCanvas.vue#L166) and [`GraphCanvas.vue` line 468](https://github.com/Comfy-Org/ComfyUI_frontend/blob/01af1a39653a0f728ba7273df4fd4881adae14af/src/components/graph/GraphCanvas.vue#L468).

That PR is a separate feature, but it looks like the same structural cleanup. If we want to restore snapping/handoff from the canvas side into Vue Nodes 2.0, I am happy to use the `#10224` bridge as the source for that path rather than duplicate the same canvas->Vue bridge work in both PRs.

src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
benceruleanlu: Renamed to shouldSnapToGrid

src/renderer/extensions/vueNodes/composables/useNodePointerInteractions.ts
benceruleanlu: agree, added safeDragCancel there that wraps cancelDrag and cleanupDragState

src/renderer/extensions/vueNodes/layout/useNodeDrag.ts
benceruleanlu: K, guarded with !guides.length && !layoutStore.vueDragSnapGuides.value.length

christian-byrne:
Removed assignment -- can you find someone else?

benceruleanlu:
> Removed assignment -- can you find someone else?

We can push this until after the test overhaul is done

artokun:
@christian-byrne @benceruleanlu i can take this over, it's related to other work I'm doing

artokun:
@christian-byrne can you assign to me, that way it's easier for me to track, will need fork and open a new PR but at least it lets others know I'm working on it

DrJKL:
Thank you for building and iterating on Nodes 2.0 drag snapping, including addressing the review feedback. The work was deferred with a plan for a new takeover PR, and the layout and geometry architecture has since been substantially centralized in #14110, leaving this branch conflicted and several key target files obsolete. If drag snapping is still desired, please open a focused implementation against current main so it can use the present layout APIs and test structure.

PR #9896, opened 2026-03-13
fix: stabilize subgraph promoted widget identity and rendering

DrJKL:
## Summary

Fix subgraph promoted widget identity/rendering so on-node widgets stay correct through configure/hydration churn, duplicate names, and linked+independent coexistence.

## Changes

- **Subgraph promotion reconciliation**: stabilize linked-entry identity by subgraph slot id, preserve deterministic linked representative selection, and prune stale alias/fallback entries without dropping legitimate independent promotions.
- **Promoted view resolution**: bind slot mapping by promoted view object identity (`getSlotFromWidget` / `getWidgetFromSlot`) to avoid same-name collisions.
- **On-node widget rendering**: harden `NodeWidgets` identity and dedup to avoid visual aliasing, prefer visible duplicates over hidden stale entries, include type/source execution identity, and avoid collapsing transient unresolved entries.
- **Mapping correctness**: update `useGraphNodeManager` promoted source mapping to resolve by input target only when the promoted view is actually bound to that input.
- **Subgraph input uniqueness**: ensure empty-slot promotion creates unique input names (`seed`, `seed_1`, etc.) for same-name multi-source promotions.
- **Safety fix**: guard against undefined canvas in slot-link interaction.
- **Tests/fixtures**: add focused regressions for fixture path `subgraph_complex_promotion_1`, linked+independent same-name cases, duplicate-name identity mapping, dedup behavior, and input-name uniqueness.

## Review Focus

Validate behavior around transient configure/hydration states (`-1` id to concrete id), duplicate-name promotions, linked representative recovery, and that dedup never hides legitimate widgets while still removing true duplicates.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9896-fix-stabilize-subgraph-promoted-widget-identity-and-rendering-3226d73d365081c8a1e8d0a5a22e826d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/graph/subgraph/matchPromotedInput.ts
+++ b/src/core/graph/subgraph/matchPromotedInput.ts
@@ -1,0 +1,18 @@
+import type { IBaseWidget } from '@/lib/litegraph/src/types/widgets'
+
+type PromotedInputLike = {
+  name: string
+  _widget?: IBaseWidget
+}
+
+export function matchPromotedInput(
+  inputs: PromotedInputLike[] | undefined,
+  widget: IBaseWidget
+): PromotedInputLike | undefined {
+  if (!inputs) return undefined
+
+  return (
+    inputs.find((input) => input._widget === widget) ??
+    inputs.find((input) => input.name === widget.name)
+  )
+}
--- a/src/core/graph/subgraph/matchPromotedInput.ts
+++ b/src/core/graph/subgraph/matchPromotedInput.ts
@@ -11,8 +11,9 @@ export function matchPromotedInput(
 ): PromotedInputLike | undefined {
   if (!inputs) return undefined
 
-  return (
-    inputs.find((input) => input._widget === widget) ??
-    inputs.find((input) => input.name === widget.name)
-  )
+  const exactMatch = inputs.find((input) => input._widget === widget)
+  if (exactMatch) return exactMatch
+
+  const sameNameMatches = inputs.filter((input) => input.name === widget.name)
+  return sameNameMatches.length === 1 ? sameNameMatches[0] : undefined
 }
--- a/src/core/graph/subgraph/promotedWidgetView.ts
+++ b/src/core/graph/subgraph/promotedWidgetView.ts
@@ -136,13 +136,17 @@ class PromotedWidgetView implements IPromotedWidgetView {
     const linkedWidgets = this.getLinkedInputWidgets()
     if (linkedWidgets.length > 0) {
       const widgetStore = useWidgetValueStore()
+      let didUpdateState = false
       for (const linkedWidget of linkedWidgets) {
         const state = widgetStore.getWidget(
           this.graphId,
           linkedWidget.nodeId,
           linkedWidget.widgetName
         )
-        if (state) state.value = value
+        if (state) {
+          state.value = value
+          didUpdateState = true
+        }
       }
 
       const resolved = this.resolveDeepest()
@@ -152,10 +156,13 @@ class PromotedWidgetView implements IPromotedWidgetView {
           stripGraphPrefix(String(resolved.node.id)),
           resolved.widget.name
         )
-        if (resolvedState) resolvedState.value = value
+        if (resolvedState) {
+          resolvedState.value = value
+          didUpdateState = true
+        }
       }
 
-      return
+      if (didUpdateState) return
     }
 
     const state = this.getWidgetState()
--- a/src/core/graph/subgraph/promotedWidgetView.ts
+++ b/src/core/graph/subgraph/promotedWidgetView.ts
@@ -1,4 +1,4 @@
-import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
+import type { LGraphNode, NodeId } from '@/lib/litegraph/src/LGraphNode'
 import type { LGraphCanvas } from '@/lib/litegraph/src/LGraphCanvas'
 import type { CanvasPointer } from '@/lib/litegraph/src/CanvasPointer'
 import type { Point } from '@/lib/litegraph/src/interfaces'
@@ -13,6 +13,7 @@ import {
   stripGraphPrefix,
   useWidgetValueStore
 } from '@/stores/widgetValueStore'
+import type { WidgetState } from '@/stores/widgetValueStore'
 import {
   resolveConcretePromotedWidget,
   resolvePromotedWidgetAtHost
@@ -37,6 +38,12 @@ type LegacyMouseWidget = IBaseWidget & {
 
 function hasLegacyMouse(widget: IBaseWidget): widget is LegacyMouseWidget {
   return 'mouse' in widget && typeof widget.mouse === 'function'
+}
+
+function hasWidgetNode(
+  widget: IBaseWidget
+): widget is IBaseWidget & { node: LGraphNode } {
+  return 'node' in widget && !!widget.node
 }
 
 const designTokenCache = new Map<string, string>()
@@ -125,12 +132,31 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   get value(): IBaseWidget['value'] {
+    const linkedState = this.getLinkedInputWidgetStates()[0]
+    if (linkedState && isWidgetValue(linkedState.value))
+      return linkedState.value
+
     const state = this.getWidgetState()
     if (state && isWidgetValue(state.value)) return state.value
     return this.resolveAtHost()?.widget.value
   }
 
   set value(value: IBaseWidget['value']) {
+    const linkedWidgets = this.getLinkedInputWidgets()
+    if (linkedWidgets.length > 0) {
+      const widgetStore = useWidgetValueStore()
+      for (const linkedWidget of linkedWidgets) {
+        const state = widgetStore.getWidget(
+          this.graphId,
+          linkedWidget.nodeId,
+          linkedWidget.widgetName
+        )
+        if (state) state.value = value
+        if (isWidgetValue(value)) linkedWidget.widget.value = value
+      }
+      return
+    }
+
     const state = this.getWidgetState()
     if (state) {
       state.value = value
@@ -278,6 +304,9 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   private getWidgetState() {
+    const linkedState = this.getLinkedInputWidgetStates()[0]
+    if (linkedState) return linkedState
+
     const resolved = this.resolveDeepest()
     if (!resolved) return undefined
     return useWidgetValueStore().getWidget(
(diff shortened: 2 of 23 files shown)


src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -241,12 +239,48 @@ function hasWidgetError(
...
+    ? `${stableIdentityRoot}:${storeWidgetName}:${slotNameForIdentity}:${widget.type}`
+    : undefined
+  const renderKey =
+    dedupeIdentity ??
+    `transient:${String(nodeId ?? '')}:${storeWidgetName}:${slotNameForIdentity}:${widget.type}:${index}`
+
+  return {
+    dedupeIdentity,
+    renderKey
+  }
```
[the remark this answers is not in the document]
DrJKL: The nodes should all have unique IDs.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -131,6 +135,38 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   set value(value: IBaseWidget['value']) {
+    const linkedWidgets = this.getLinkedInputWidgets()
```
christian-byrne, line 138: issue: `get value()` reads from `getLinkedInputWidgetStates()[0]` (the first linked state), while `set value()` writes to ALL linked widgets plus the deepest-resolved widget. If `getLinkedInputWidgets()` returns entries in non-deterministic order, getter and setter operate on different state objects.

Also, if some `getWidget()` calls return undefined during the setter loop, partial updates occur -- some linked states get the new value, others don't, and the `didUpdateState` flag prevents the fallback path from running.

Consider making the getter read from the same canonical state the setter writes to as its primary target.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -287,6 +326,57 @@ class PromotedWidgetView implements IPromotedWidgetView {
     )
   }
 
+  private getLinkedInputWidgets(): Array<{
```
christian-byrne, line 329: suggestion: `getLinkedInputWidgets()` performs a linear scan of all inputs with nested calls to `matchPromotedInput()`, `getConnectedWidgets()`, and `filter(hasWidgetNode)`. It's called from `getWidgetState()` -> `value` getter on every render frame, but unlike `resolveDeepest()` (which has `cachedDeepestByFrame`), it has no caching. Consider adding frame-based caching similar to `cachedDeepestByFrame`.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -91,82 +99,201 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+      this._linkedEntriesCache = {
+        version: this._cacheVersion,
+        hasMissingBoundSourceWidget,
+        entries: deduplicatedEntries
+      }
+
     return deduplicatedEntries
   }
 
+  private _hasMissingBoundSourceWidget(): boolean {
```
christian-byrne, line 226: suggestion: `_hasMissingBoundSourceWidget()` iterates all inputs with `getNodeById` + widget array scan, and it's called as a cache-freshness signal on every `widgets` getter access (i.e., every frame) from both `_getLinkedPromotionEntries()` and `_getPromotedViews()`. This creates a hybrid cache invalidation model (reactive heuristic bypassing `_cacheVersion`) that's hard to reason about. Consider either making it part of the dirty-flag invalidation (set on relevant events) or removing it from the cache key and relying solely on `_cacheVersion`.

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -256,10 +290,76 @@ const processedWidgets = computed((): ProcessedWidget[] => {
...
+        mergedOptions,
+        widgetState,
+        isVisible: visible
+      })
+      continue
+    }
+
+    const existingWidget = uniqueWidgets[existingIndex]
+    if (existingWidget && !existingWidget.isVisible && visible) {
+      uniqueWidgets[existingIndex] = {
```
christian-byrne, line 348: issue: When a duplicate is found and the replacement is visible but the original wasn't, the entire entry is replaced including `identity` -- which may have a different `renderKey`. Since `renderKey` is used as Vue's `:key`, this forces Vue to destroy and recreate the component rather than patch in-place. Consider preserving the original `renderKey` when replacing the dedup entry.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -316,11 +548,12 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
   private _makePromotionViewKey(
-    inputName: string,
+    inputKey: string,
     interiorNodeId: string,
-    widgetName: string
+    widgetName: string,
+    inputName = ''
   ): string {
-    return `${inputName}:${interiorNodeId}:${widgetName}`
+    return `${inputKey}:${interiorNodeId}:${widgetName}:${inputName}`
```
christian-byrne: suggestion: The key format `${inputKey}:${interiorNodeId}:${widgetName}:${inputName}` uses `:` as delimiter, but node IDs can contain colons (e.g., `"65:18"` for subgraph execution paths), creating ambiguous key boundaries. Consider using a delimiter that cannot appear in any component (e.g., `\x1F` unit separator) or `JSON.stringify([...])`.

src/core/graph/subgraph/widgetRenderKey.test.ts
```diff
@@ -0,0 +1,34 @@
...
+
+function createWidget(overrides: Partial<IBaseWidget> = {}): IBaseWidget {
+  return {
+    name: 'seed',
+    type: 'number',
+    ...overrides
+  } as IBaseWidget
+}
+
+describe(getStableWidgetRenderKey, () => {
```
christian-byrne, line 15: suggestion (non-blocking): The test file covers key stability and uniqueness but doesn't test the promoted vs widget prefix branch -- the function's primary differentiator (`isPromotedWidgetView(widget) ? 'promoted' : 'widget'`). Consider adding a test with a mock `PromotedWidgetView` verifying the key starts with `'promoted:'`.

src/core/graph/subgraph/matchPromotedInput.ts
```diff
@@ -0,0 +1,18 @@
...
+}
+
+export function matchPromotedInput(
+  inputs: PromotedInputLike[] | undefined,
+  widget: IBaseWidget
+): PromotedInputLike | undefined {
+  if (!inputs) return undefined
+
+  return (
+    inputs.find((input) => input._widget === widget) ??
```
christian-byrne: suggestion (non-blocking): The name-fallback path (`inputs.find((input) => input.name === widget.name)`) can match the wrong input when there are duplicate names -- the exact scenario this PR is fixing. The first name match wins, which may be incorrect. Consider documenting the constraint that callers must scope inputs to avoid ambiguity, or removing the name fallback.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -91,82 +99,201 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
+  }
+
   private _getPromotedViews(): PromotedWidgetView[] {
     const store = usePromotionStore()
     const entries = store.getPromotionsRef(this.rootGraph.id, this.id)
+    const hasMissingBoundSourceWidget = this._hasMissingBoundSourceWidget()
+    const cachedViews = this._promotedViewsCache
+    if (
+      cachedViews?.version === this._cacheVersion &&
+      cachedViews.entriesRef === entries &&
```
christian-byrne, line 250: suggestion (non-blocking): `cachedViews.entriesRef === entries` uses reference equality, but `getPromotionsRef()` returns `?? []` (in `promotionStore.ts:65`), which creates a new array on every call for nodes with no promotions. This means the cache always misses for those nodes. Consider caching a stable empty-array sentinel in `getPromotionsRef`.

src/lib/litegraph/src/subgraph/__fixtures__/subgraphHelpers.ts
```diff
@@ -209,6 +230,47 @@ export function createTestSubgraphNode(
...
+  graph: LGraph
+  subgraph: Subgraph
+  hostNode: SubgraphNode
+} {
+  const fixture = structuredClone(subgraphComplexPromotion1)
+  const subgraphData = fixture.definitions?.subgraphs?.[0]
+  if (!subgraphData)
+    throw new Error('Expected fixture to contain one subgraph definition')
+
+  LiteGraph.registerNodeType(
```
christian-byrne, line 249: suggestion (non-blocking): `FixtureStringConcatenateNode` is registered globally via `LiteGraph.registerNodeType` but never unregistered. This leaks across test files and could cause ordering-dependent test failures. Consider adding cleanup in `afterEach` or using vitest hooks to unregister after tests complete.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Implemented this one~ ✨ We switched SubgraphNode._makePromotionViewKey to structured encoding so key generation is unambiguous and collision-safe even when segments contain separator-like text (｡•̀ᴗ-)✧.

Added regression coverage for collision-shaped tuples too, so this path is now locked down nicely ದ್ದി(˵ •̀ ᴗ - ˵ ).

src/core/graph/subgraph/matchPromotedInput.ts
DrJKL: Fixed with a stricter fallback in matchPromotedInput \(´｡• ᵕ •｡`)/. We still prioritize exact _widget matching, and now name fallback only applies when the name is uniquely matched; duplicate-name ambiguity now returns undefined instead of guessing ૮꒰ ˶• ༝ •˶꒱ა.

Added tests for both unique fallback and duplicate rejection.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Implemented~ We now return a stable shared empty ref from getPromotionsRef in the empty case, so referential checks stay stable and we avoid unnecessary cache churn (˶ᵔ ᵕ ᵔ˶).

Added a dedicated test for stable empty ref behavior too (*ฅ́˘ฅ̀*).

src/lib/litegraph/src/subgraph/__fixtures__/subgraphHelpers.ts
DrJKL: Great catch, fixed! We added fixture node-type cleanup so Fixture/StringConcatenate registration is torn down cleanly and does not leak across tests ᐢ•˕•ᐢ.

Also added a focused helper cleanup test and wired teardown in the fixture-heavy promoted-widget tests (❀◠‿◠).

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: Re-checked this deeply and kept behavior unchanged for now (｡•ㅅ•｡). The read and write asymmetry is intentional in this flow: representative read path plus fanout write path for linked consistency.

Avoided semantic expansion in this PR and focused on correctness-critical fixes first ♡.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: Very fair perf thought~ (´꒳`)♡ We did not add caching in this pass to keep invalidation complexity low without profiler evidence of a hotspot.

If perf traces show this path is hot, we can add a minimal frame-scoped cache follow-up \(˶ᵔ ᵕ ᵔ˶)/.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: Reviewed this skeptically and decided not to redesign freshness invalidation in this PR ૮₍˶ •. • ⑅₎ა.

Keeping the current model for now to avoid broad architectural risk while landing concrete correctness fixes first.

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
DrJKL: Double-checked this path and did not apply a change ( •̀ᴗ•́ )✧.

In the dedupe replacement branch, dedupeIdentity and renderKey remain aligned, so that specific replacement flow should not force remount by key change.

src/core/graph/subgraph/widgetRenderKey.test.ts
DrJKL: Totally reasonable suggestion~ ♡(˶╹̆ ▿ ╹̆˵)♡

Treated this as non-blocking in this pass and prioritized higher-impact correctness items first. Happy to add this branch-coverage polish in a tiny follow-up \(^_^)/.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: (I will follow up on this in a performance focused pass later)

PR #5265, opened 2025-08-30
[fix] Change keyboard event handling from event.key to event.code

snomiao:
Just vibe coded this, I think we should do more research for backward compatibilities for this drafted PR.

and @iamrajiv may help on improvements :D

cc @christian-byrne 

## Summary

This PR fixes keyboard shortcuts not working on non-English (non-Latin) keyboard layouts by changing the keyboard event handling from `event.key` to `event.code`.

The issue was that `event.key` returns different values depending on the active keyboard layout, while `event.code` always represents the physical key regardless of layout. This means shortcuts like `Ctrl+S` would fail when the user has a Russian or other non-Latin keyboard layout active.

## Changes Made

- **KeyComboImpl.fromEvent()**: Now uses `event.code` instead of `event.key` to create key combinations
- **isModifier check**: Updated to use key codes for modifier keys (`ControlLeft`, `ControlRight`, etc.)
- **getDisplayKey() method**: Added new method to convert key codes back to readable display names
- **Escape key handling**: Updated keybindingService to use `event.code` for Escape key checks
- **KeybindingPanel**: Updated captureKeybinding to use `event.code` for Escape/Enter detection
- **Core keybindings**: Migrated all keybindings from `event.key` format to `event.code` format
- **Tests**: Fixed all test mocks to use `event.code` instead of `event.key`

## Test Plan

- [x] Existing keybinding store tests pass
- [x] Keybinding service escape key tests pass
- [x] TypeScript type checking passes
- [ ] Manual testing with non-Latin keyboard layout (Russian, Japanese, etc.)
- [ ] Verify shortcuts work with different layouts active
- [ ] Test that keybinding UI still displays readable key names

## Breaking Changes

None - this is an internal implementation change. The external API and user experience remain the same.

Fixes #5252

🤖 Generated with [Claude Code](https://claude.ai/code)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5265-fix-Change-keyboard-event-handling-from-event-key-to-event-code-25f6d73d365081e48188eb010e9e24c0) by [Unito](https://www.unito.io)

PR #9900, opened 2026-03-13
fix: resolve keybinding key from event.code when modifiers are held

### LINKED ISSUE #2340: [Bug]: When using a different locale (non-English) and with a different language (non-English) enabled in the system, Keybinding does not work
### Frontend Version

ComfyUI_frontend v1.8.5

### Expected Behavior

When using a different locale (not English) and with a different language (not English) enabled in the system, any key combinations set in the settings should work. It should not be that CTRL+S does not work when using a different locale and a different language set in the system.
I expect that the CTRL+S I set up will always work, in any locales and system languages ​​(in any keyboard layout).

### Actual Behavior

When using a different locale (non-English) and with a different language (non-English) enabled in the system, Keybinding does not work.
More precisely, ONLY those keybindings that are set in the locale language start to work.

### Steps to Reproduce

1) Change the locale in ComfyUI settings to, for example, Russian.
2) Switch the system keyboard layout to Russian.
3) Try using the hotkey CTRL+S, which was set in the Keybindings settings.

### Debug Logs

```powershell
-
```

### Browser Logs

-

### Setting JSON

-

Google Chrome

### Other

I assume this is an intentional change?

However, I really wish this behavior could be disabled, as I often switch between two languages ​​in the system, which is why keybindings simply do not work if they are set in another language.

Also, if this innovation is left as is, it would be very useful to add in this case the ability to add multiple hotkeys for one action.
For example, if I used the "Save Workflow" function, I would set two hotkeys for different locales (English and Russian):
CTRL+S; CTRL+Ы

Myestery:
## Summary

Use `event.code` instead of `event.key` in `KeyComboImpl.fromEvent()` when modifier keys (Ctrl/Alt/Meta) are held, fixing keybinding capture and matching on macOS and non-English keyboard layouts.

## Changes

- **What**: `resolveKeyFromEvent()` extracts the physical key from `event.code` (e.g., `KeyM` → `m`, `Digit1` → `1`) when any modifier is held, instead of using `event.key` which returns locale-dependent or macOS Option-modified characters (e.g., `µ`, `ß`, `ы`). Keybinding capture input is now `readonly` with composition events prevented to stop macOS dead keys (Alt+E/U/I/N) from corrupting subsequent keypresses.
- **Breaking**: Previously saved Alt keybindings that stored special characters (e.g., `µ` for Alt+M) will need to be re-set

## Review Focus

- The fix applies broadly to all `fromEvent` call sites (keybinding capture dialog, keybinding service handler, canvas forwarding)
- Only `Key*` and `Digit*` codes are resolved from `event.code`; other keys (Enter, Escape, F-keys, etc.) fall through to `event.key`

Fixes #5252
Fixes #2340

https://github.com/user-attachments/assets/89ce2334-78a7-4889-9c17-78966171daca

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9900-fix-resolve-keybinding-key-from-event-code-when-modifiers-are-held-3226d73d365081878ec6fa231a573653) by [Unito](https://www.unito.io)


src/platform/keybindings/keyCombo.ts
```diff
@@ -89,6 +89,23 @@ export class KeyComboImpl implements KeyCombo {
...
+ *
+ * Use event.code to resolve the physical key in these cases.
+ */
+function resolveKeyFromEvent(event: KeyboardEvent): string {
+  if ((event.ctrlKey || event.altKey || event.metaKey) && event.code) {
+    const { code } = event
+    if (code.startsWith('Key')) return code.slice(3).toLowerCase()
+    if (code.startsWith('Digit')) return code.slice(5)
+  }
+  return event.key
```
christian-byrne, line 127: issue: `resolveKeyFromEvent` only maps `Key*` and `Digit*` codes. Punctuation codes like `Comma`, `Equal`, `Minus`, `Backquote` fall through to `event.key`, which is still locale-dependent when modifiers are held.

On a Russian keyboard: Ctrl+, produces `event.key: 'б'` but `code: 'Comma'` -- the function returns 'б' instead of ','.

Affected default keybindings:
- `Ctrl+,` (ShowSettingsDialog, code: `Comma`)
- `Alt+=` (ZoomIn, code: `Equal`)
- `Alt+-` (ZoomOut, code: `Minus`)
- `` Ctrl+` `` (ToggleBottomPanel, code: `Backquote`)

A `CODE_TO_KEY` map for common punctuation codes would fix this, or it could be documented as a known limitation for a follow-up.

src/platform/keybindings/keyCombo.ts
```diff
@@ -89,6 +89,23 @@ export class KeyComboImpl implements KeyCombo {
...
+/**
+ * When modifier keys (Ctrl, Alt, Meta) are held, event.key may return
+ * unexpected characters:
+ * - macOS Option (Alt) produces special chars (e.g., Alt+M → µ, Alt+S → ß)
+ * - Non-English layouts return localized chars (e.g., Ctrl+S → Ctrl+ы in Russian)
+ *
+ * Use event.code to resolve the physical key in these cases.
+ */
+function resolveKeyFromEvent(event: KeyboardEvent): string {
+  if ((event.ctrlKey || event.altKey || event.metaKey) && event.code) {
```
christian-byrne: suggestion (non-blocking): the Shift exclusion from the modifier check is correct (Shift alone intentionally changes `event.key`, e.g. Shift+1 produces '!'), but a brief inline comment explaining why would prevent a future maintainer from "fixing" this.

christian-byrne: note (non-blocking): on Dvorak with Alt held, physical C position resolves to 'c' (QWERTY) instead of Dvorak 'j'. This is an acceptable tradeoff -- the i18n fix covers far more users than Dvorak+Alt users, and matches how other editors (VS Code) handle this.

src/platform/keybindings/keyCombo.test.ts
```diff
@@ -3,6 +3,85 @@ import { describe, expect, it } from 'vitest'
...
+        expect(combo.key).toBe(expected)
+        expect(combo.alt).toBe(true)
+      }
+    )
+
+    it.each([
+      { key: '¡', code: 'Digit1', expected: '1' },
+      { key: '€', code: 'Digit2', expected: '2' }
+    ])(
+      'resolves Alt+$code to "$expected" instead of "$key" (macOS)',
```
christian-byrne: nitpick (non-blocking): both `it.each` blocks at lines 24 and 38 share the same description string `'resolves Alt+$code to "$expected" instead of "$key" (macOS)'`. Consider differentiating them (e.g. `"(macOS digits)"` for the second) or merging into one table.

src/platform/keybindings/keyCombo.test.ts
```diff
@@ -3,6 +3,85 @@ import { describe, expect, it } from 'vitest'
...
+      expect(combo.key).toBe('ы')
+    })
+
+    it('uses event.key for non-letter/digit codes with modifier', () => {
+      const combo = KeyComboImpl.fromEvent(
+        mockKeyEvent({ key: 'Enter', code: 'Enter', altKey: true })
+      )
+      expect(combo.key).toBe('Enter')
+      expect(combo.alt).toBe(true)
+    })
```
christian-byrne, line 120: nitpick (non-blocking): no tests for punctuation key codes with modifiers (e.g. `Ctrl+Comma`, `Alt+Equal`). Adding them would have caught the `Key*`/`Digit*`-only mapping gap in `resolveKeyFromEvent`.

src/platform/keybindings/keyCombo.test.ts
```diff
@@ -3,6 +3,85 @@ import { describe, expect, it } from 'vitest'
...
+    )
+
+    it('uses event.key when no modifier is pressed', () => {
+      const combo = KeyComboImpl.fromEvent(
+        mockKeyEvent({ key: 'ы', code: 'KeyS' })
+      )
+      expect(combo.key).toBe('ы')
+    })
+
+    it('uses event.key for non-letter/digit codes with modifier', () => {
```
christian-byrne: nitpick (non-blocking): `Ctrl+Numpad1` has `code: 'Numpad1'` and `key: '1'`. The fallthrough to `event.key` is correct behavior here, but a test case would document that expectation.

christian-byrne (commented):
suggestion (non-blocking): user-customized Alt+letter bindings on macOS previously stored special characters (e.g. Alt+M stored 'µ'). After this change, keypresses resolve to plain letters so those stored bindings won't match. Impact is low since those bindings were already broken, but worth a note in the release changelog.

pythongosssss:
Is it worth looking at using some standard library for keybinding and migrating our definitions?

src/platform/keybindings/keyCombo.ts
Myestery: Addressed in bdf252fb7 — added a `CODE_TO_KEY` map covering `Comma`, `Period`, `Slash`, `Backslash`, `BracketLeft`, `BracketRight`, `Semicolon`, `Quote`, `Backquote`, `Minus`, and `Equal`. Punctuation codes now resolve to their US-layout characters when modifiers are held, matching the same approach used for `Key*`/`Digit*` codes.

src/platform/keybindings/keyCombo.ts
Myestery: Added an inline comment explaining the Shift exclusion rationale.

src/platform/keybindings/keyCombo.test.ts
Myestery: Fixed — second block now says `"resolves digit Alt+$code"` to differentiate.

src/platform/keybindings/keyCombo.test.ts
Myestery: Added `it.each` covering `Comma`, `Period`, `BracketRight`, `Backquote`, `Minus`, and `Equal` with Alt modifier.

src/platform/keybindings/keyCombo.test.ts
Myestery: Added a test case for `Ctrl+Numpad1` documenting the fallthrough to `event.key`.

src/components/dialog/content/setting/keybinding/EditKeybindingFooter.vue
```diff
@@ -0,0 +1,79 @@
...
+  const combo = dialogState.newCombo
+  const commandId = dialogState.commandId
+  if (!combo || !commandId) return
+
+  dialogStore.closeDialog({ key: DIALOG_KEY })
+
+  const updated = keybindingStore.updateKeybindingOnCommand(
+    new KeybindingImpl({ commandId, combo })
+  )
+  if (updated) await keybindingService.persistUserKeybindings()
```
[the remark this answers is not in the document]
Myestery: Fixed — dialog now closes after `persistUserKeybindings()` completes.

src/components/dialog/content/setting/KeybindingPanel.vue
```diff
@@ -59,7 +59,15 @@
...
         <template #body="slotProps">
-          <div class="truncate" :title="slotProps.data.id">
+          <div
+            class="flex items-center gap-1.5 truncate"
+            :title="slotProps.data.id"
+          >
+            <i
+              v-if="slotProps.data.keybinding?.combo.isBrowserReserved"
+              v-tooltip="$t('g.browserReservedKeybindingTooltip')"
+              class="icon-[lucide--triangle-alert] shrink-0 text-warning-background"
```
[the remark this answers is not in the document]
Myestery: No change needed — `toNormalizedString` (line 114) already lowercases single-char keys before lookup, so persisted combos match regardless of the case stored in the keybinding definition.

Myestery:
Looked into this. The main standard libraries (tinykeys, hotkeys-js, mousetrap) handle basic hotkey registration but don't cover several features we rely on:

- **Browser-reserved key detection** with user warnings (23 combos)
- **Text-input context awareness** — suppressing commands while typing in inputs (26 combos)
- **Element scoping** — canvas-only vs global bindings
- **User customization + persistence** — add/remove/reset with settings integration
- **Key code resolution** — the `event.code` handling for macOS Option + non-English layouts that this PR adds

We'd end up wrapping whichever library we pick with most of this logic anyway. The core key resolution is ~130 lines; the rest is integration with our command/settings/dialog stores.

That said, if there's a specific library you had in mind I'm happy to evaluate it more concretely.

pythongosssss:
No specific library in mind.

On Windows using United States International keyboard layout, doing CTRL+ALT+A/E/I/O/U shows áéíóú 
Doing the same in e.g. VScode shows CTRL+ALT+A correctly

@DrJKL last time keybind handling like this were reviewed (#5265) you had some concerns, if you could have a look at this

edit: Looks like vscode uses a native library for its keybindings, when testing using https://vscode.dev/ - the keybinding here for CTRL+ALT+A just shows A - so this may be a browser limitation

src/platform/keybindings/keyCombo.ts
```diff
@@ -89,6 +89,41 @@ export class KeyComboImpl implements KeyCombo {
...
+  Slash: '/',
+  Backslash: '\\',
+  BracketLeft: '[',
+  BracketRight: ']',
+  Semicolon: ';',
+  Quote: "'",
+  Backquote: '`',
+  Minus: '-',
+  Equal: '='
+}
```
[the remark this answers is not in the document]
Myestery: Not applicable. event.code for Space is 'Space' which doesn't match startsWith('Key') or startsWith('Digit'), so it falls through to event.key (' ').

src/platform/keybindings/keyCombo.test.ts
```diff
@@ -3,6 +3,111 @@ import { describe, expect, it } from 'vitest'
...
+      expect(combo.key).toBe('Enter')
+      expect(combo.alt).toBe(true)
+    })
+
+    it('falls through to event.key for unmapped codes like Numpad', () => {
+      const combo = KeyComboImpl.fromEvent(
+        mockKeyEvent({ key: '1', code: 'Numpad1', ctrlKey: true })
+      )
+      expect(combo.key).toBe('1')
+      expect(combo.ctrl).toBe(true)
```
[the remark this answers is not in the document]
Myestery: Already addressed in commit 6fa9f068a

src/platform/keybindings/defaults.ts
```diff
@@ -100,7 +100,7 @@ export const CORE_KEYBINDINGS: Keybinding[] = [
   },
   {
     combo: {
-      key: '+',
+      key: '=',
```
DrJKL, line 102: So key is now keyCode?

Myestery: well, accessing `+` means going through shift, so the way to get here is by binding shift and the base key

DrJKL:
Can you add an integration test for this?

christian-byrne (changes requested):
> Breaking: Previously saved Alt keybindings that stored special characters (e.g., µ for Alt+M) will need to be re-set


Is that actually an acceptable downside? There are also a lot of other downsides of doing this which is why we never did it, like:

- AZERTY / other non-QWERTY layouts break for non-modifier keys. E.g., this would break things for me specifically
- symbols and special characters become non-configurable

It also doesn't fix underlying issue for all layouts. Maybe explore the KeyboardMap API (can't remember exact name).

Myestery:
> > Breaking: Previously saved Alt keybindings that stored special characters (e.g., µ for Alt+M) will need to be re-set
> 
> Is that actually an acceptable downside? There are also a lot of other downsides of doing this which is why we never did it, like:
> 
> * AZERTY / other non-QWERTY layouts break for non-modifier keys. E.g., this would break things for me specifically
> * symbols and special characters become non-configurable
> 
> It also doesn't fix underlying issue for all layouts. Maybe explore the KeyboardMap API (can't remember exact name).

So this only affects non-default keybind settings and yes I think its a downside we want to accept for now. There's almost no way to make these improvements without discarding whatever was set using those weird characters.

Symbols will continue be supported with the normal english set expectations.

jtydhr88:
As discussed we don’t need it

PR #9901, opened 2026-03-13
feat: differentiate personal/team pricing table with two-stage team workspace flow

huntcsg:
## Summary

Differentiates the subscription pricing dialog between personal and team workspaces with distinct visual treatments and a two-stage team workspace upgrade flow.

### Changes

- **Personal pricing dialog**: Shows "P" avatar badge, "Plans for Personal Workspace" header, and "Solo use only – Need team workspace?" banner on each tier card
- **Team pricing dialog**: Shows workspace avatar, "Plans for Team Workspace" header (emerald), green "Invite up to X members" badge, and emerald border on Creator card
- **Two-stage upgrade flow**: "Need team workspace?" → closes pricing → opens CreateWorkspaceDialog → sessionStorage flag → page reload → WorkspaceAuthGate auto-opens team pricing dialog
- **Spacing**: Reduced vertical gaps/padding/font sizes so the table fits without scrolling

### Key decisions

- sessionStorage key `comfy:resume-team-pricing` bridges the page reload during workspace creation
- `onChooseTeam` prop is conditionally passed only to the personal variant
- `resumePendingPricingFlow()` is called from WorkspaceAuthGate after workspace initialization

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9901-feat-differentiate-personal-team-pricing-table-with-two-stage-team-workspace-flow-3226d73d365081e7af60dcca86e83673) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -559,7 +559,14 @@ onMounted(async () => {
 
   // Open create workspace dialog from URL if present (e.g., ?create_workspace=1)
   if (createWorkspaceUrlLoader && flags.teamWorkspacesEnabled) {
-    await createWorkspaceUrlLoader.loadCreateWorkspaceFromUrl()
+    try {
+      await createWorkspaceUrlLoader.loadCreateWorkspaceFromUrl()
+    } catch (error) {
+      console.error(
+        '[GraphCanvas] Failed to load create workspace from URL:',
+        error
+      )
+    }
   }
 
   // Initialize release store to fetch releases from comfy-api (fire-and-forget)
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -191,6 +191,7 @@ import { forEachNode } from '@/utils/graphTraversalUtil'
 import SelectionRectangle from './SelectionRectangle.vue'
 import { isCloud } from '@/platform/distribution/types'
 import { useFeatureFlags } from '@/composables/useFeatureFlags'
+import { useCreateWorkspaceUrlLoader } from '@/platform/workspace/composables/useCreateWorkspaceUrlLoader'
 import { useInviteUrlLoader } from '@/platform/workspace/composables/useInviteUrlLoader'
 
 const { t } = useI18n()
@@ -434,8 +435,9 @@ useEventListener(
 const comfyAppReady = ref(false)
 const workflowPersistence = useWorkflowPersistence()
 const { flags } = useFeatureFlags()
-// Set up invite loader during setup phase so useRoute/useRouter work correctly
+// Set up URL loaders during setup phase so useRoute/useRouter work correctly
 const inviteUrlLoader = isCloud ? useInviteUrlLoader() : null
+const createWorkspaceUrlLoader = isCloud ? useCreateWorkspaceUrlLoader() : null
 useCanvasDrop(canvasRef)
 useLitegraphSettings()
 useNodeBadge()
@@ -553,6 +555,11 @@ onMounted(async () => {
   // WorkspaceAuthGate ensures flag state is resolved before GraphCanvas mounts
   if (inviteUrlLoader && flags.teamWorkspacesEnabled) {
     await inviteUrlLoader.loadInviteFromUrl()
+  }
+
+  // Open create workspace dialog from URL if present (e.g., ?create_workspace=1)
+  if (createWorkspaceUrlLoader && flags.teamWorkspacesEnabled) {
+    await createWorkspaceUrlLoader.loadCreateWorkspaceFromUrl()
   }
 
   // Initialize release store to fetch releases from comfy-api (fire-and-forget)
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2586,7 +2586,8 @@
     "roleOwner": "Owner",
     "roleMember": "Member",
     "createWorkspace": "Create new workspace",
-    "maxWorkspacesReached": "You can only own 10 workspaces. Delete one to create a new one."
+    "maxWorkspacesReached": "You can only own 10 workspaces. Delete one to create a new one.",
+    "failedToSwitch": "Failed to switch workspace"
   },
   "selectionToolbox": {
     "executeButton": {
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2573,7 +2573,7 @@
   "teamWorkspacesDialog": {
     "title": "Team Workspaces",
     "subtitle": "Switch to an existing one or create a new workspace",
-    "subtitleNoWorkspaces": "Create a new team workspace to collaborate",
+    "subtitleNoWorkspaces": "Create a new team workspace to share credits",
     "confirmCallbackFailed": "Workspace created but setup incomplete",
     "yourTeamWorkspaces": "Your team workspaces",
     "switch": "Switch",
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2570,6 +2570,15 @@
       "failedToFetchWorkspaces": "Failed to load workspaces"
     }
   },
+  "teamWorkspacesDialog": {
+    "title": "Team Workspaces",
+    "subtitle": "Switch to an existing one or create a new workspace",
+    "yourTeamWorkspaces": "Your team workspaces",
+    "switch": "Switch",
+    "newWorkspace": "New workspace",
+    "namePlaceholder": "e.g. Marketing Team",
+    "createWorkspace": "Create workspace"
+  },
   "workspaceSwitcher": {
     "switchWorkspace": "Switch workspace",
     "subscribe": "Subscribe",
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2573,6 +2573,8 @@
   "teamWorkspacesDialog": {
     "title": "Team Workspaces",
     "subtitle": "Switch to an existing one or create a new workspace",
+    "subtitleNoWorkspaces": "Create a new team workspace to collaborate",
+    "confirmCallbackFailed": "Workspace created but setup incomplete",
     "yourTeamWorkspaces": "Your team workspaces",
     "switch": "Switch",
     "newWorkspace": "New workspace",
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2253,7 +2253,6 @@
     "topupTimeout": "Top-up verification timed out"
   },
   "subscription": {
-    "chooseBestPlanWorkspace": "Choose the best plan for your workspace",
     "plansFor": "Plans for",
     "personalWorkspace": "Personal Workspace",
     "teamWorkspace": "Team Workspace",
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2253,7 +2253,7 @@
     "topupTimeout": "Top-up verification timed out"
   },
   "subscription": {
-    "plansFor": "Plans for",
+    "plansForWorkspace": "Plans for {workspace}",
     "personalWorkspace": "Personal Workspace",
     "teamWorkspace": "Team Workspace",
     "soloUseOnly": "Solo use only",
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2579,7 +2579,8 @@
     "switch": "Switch",
     "newWorkspace": "New workspace",
     "namePlaceholder": "e.g. Marketing Team",
-    "createWorkspace": "Create workspace"
+    "createWorkspace": "Create workspace",
+    "nameValidationError": "Name must be 1–50 characters using letters, numbers, spaces, or common punctuation."
   },
   "workspaceSwitcher": {
     "switchWorkspace": "Switch workspace",
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -2254,6 +2254,12 @@
   },
   "subscription": {
     "chooseBestPlanWorkspace": "Choose the best plan for your workspace",
+    "plansFor": "Plans for",
+    "personalWorkspace": "Personal Workspace",
+    "teamWorkspace": "Team Workspace",
+    "soloUseOnly": "Solo use only",
+    "needTeamWorkspace": "Need team workspace?",
+    "inviteUpTo": "Invite up to",
     "title": "Subscription",
     "titleUnsubscribed": "Subscribe to Comfy Cloud",
     "comfyCloud": "Comfy Cloud",
--- a/src/platform/cloud/subscription/components/PricingTable.vue
+++ b/src/platform/cloud/subscription/components/PricingTable.vue
@@ -95,6 +95,21 @@
             </div>
           </div>
 
+          <div
+            class="flex h-10 items-center rounded-lg bg-secondary-background px-3"
+          >
+            <span class="text-sm text-muted-foreground">
+              {{ t('subscription.soloUseOnly') }}
+              <span class="mx-1 text-muted-foreground">–</span>
+              <button
+                class="text-primary-foreground cursor-pointer border-none bg-transparent p-0 text-sm font-medium underline hover:text-base-foreground"
+                @click="emit('chooseTeamWorkspace')"
+              >
+                {{ t('subscription.needTeamWorkspace') }}
+              </button>
+            </span>
+          </div>
+
           <div class="flex flex-1 flex-col gap-4 pb-0">
             <div class="flex flex-row items-center justify-between">
               <span
@@ -302,6 +317,10 @@ interface PricingTierConfig {
   customLoRAs: boolean
   isPopular?: boolean
 }
+
+const emit = defineEmits<{
+  chooseTeamWorkspace: []
+}>()
 
 const { t, n } = useI18n()
 
--- a/src/platform/cloud/subscription/components/PricingTable.vue
+++ b/src/platform/cloud/subscription/components/PricingTable.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="flex flex-col gap-8">
+  <div class="flex flex-col gap-6">
     <div class="flex justify-center">
       <SelectButton
         v-model="currentBillingCycle"
@@ -38,7 +38,7 @@
         </template>
       </SelectButton>
     </div>
-    <div class="flex flex-col items-stretch gap-6 xl:flex-row">
+    <div class="flex flex-col items-stretch gap-4 xl:flex-row">
       <div
         v-for="tier in tiers"
         :key="tier.id"
@@ -49,7 +49,7 @@
           )
         "
       >
-        <div class="flex flex-col gap-8 p-8 pb-0">
+        <div class="flex flex-col gap-4 p-6 pb-0">
           <div class="flex flex-row items-center justify-between gap-2">
             <span
               class="font-inter text-base/normal font-bold text-base-foreground"
@@ -67,7 +67,7 @@
             <div class="flex flex-col gap-2">
               <div class="flex flex-row items-baseline gap-2">
                 <span
-                  class="font-inter text-[32px] leading-normal font-semibold text-base-foreground"
+                  class="font-inter text-[28px] leading-normal font-semibold text-base-foreground"
                 >
                   <span
                     v-show="currentBillingCycle === 'yearly'"
@@ -110,7 +110,7 @@
             </span>
           </div>
 
-          <div class="flex flex-1 flex-col gap-4 pb-0">
+          <div class="flex flex-1 flex-col gap-3 pb-0">
             <div class="flex flex-row items-center justify-between">
               <span
                 class="text-foreground font-inter text-sm/normal font-normal"
@@ -194,7 +194,7 @@
             </div>
           </div>
         </div>
-        <div class="flex flex-col p-8">
+        <div class="flex flex-col p-6">
           <Button
             :variant="getButtonSeverity(tier)"
             :disabled="isLoading || isCurrentPlan(tier.key)"
--- a/src/platform/cloud/subscription/components/PricingTable.vue
+++ b/src/platform/cloud/subscription/components/PricingTable.vue
@@ -96,7 +96,7 @@
           </div>
 
           <div
-            class="flex h-10 items-center rounded-lg bg-secondary-background px-3"
+            class="flex h-10 items-center rounded-lg bg-muted-foreground/30 px-3"
           >
             <span class="text-sm text-muted-foreground">
               {{ t('subscription.soloUseOnly') }}
--- a/src/platform/cloud/subscription/components/PricingTable.vue
+++ b/src/platform/cloud/subscription/components/PricingTable.vue
@@ -95,20 +95,20 @@
             </div>
           </div>
 
-          <div
-            class="flex h-10 items-center rounded-lg bg-muted-foreground/30 px-3"
+          <p
+            role="note"
+            :aria-label="t('subscription.soloUseOnly')"
+            class="m-0 flex h-10 items-center rounded-lg bg-muted-foreground/30 px-3 text-sm text-muted-foreground"
           >
-            <span class="text-sm text-muted-foreground">
-              {{ t('subscription.soloUseOnly') }}
-              <span class="mx-1 text-muted-foreground">–</span>
-              <button
-                class="text-primary-foreground cursor-pointer border-none bg-transparent p-0 text-sm font-medium underline hover:text-base-foreground"
-                @click="emit('chooseTeamWorkspace')"
-              >
-                {{ t('subscription.needTeamWorkspace') }}
-              </button>
-            </span>
-          </div>
+            {{ t('subscription.soloUseOnly') }}
+            <span class="mx-1 text-muted-foreground">–</span>
+            <button
+              class="text-primary-foreground cursor-pointer border-none bg-transparent p-0 text-sm font-medium underline hover:text-base-foreground focus-visible:ring-1 focus-visible:outline-none"
+              @click="emit('chooseTeamWorkspace')"
+            >
+              {{ t('subscription.needTeamWorkspace') }}
+            </button>
+          </p>
 
           <div class="flex flex-1 flex-col gap-3 pb-0">
             <div class="flex flex-row items-center justify-between">
--- a/src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
+++ b/src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
@@ -10,7 +10,7 @@
       :aria-label="$t('g.close')"
       @click="handleClose"
     >
-      <i class="pi pi-times text-xl" />
+      <i class="pi pi-times text-xl" aria-hidden="true" />
     </Button>
     <div class="flex flex-col items-center gap-3">
       <div
@@ -70,7 +70,7 @@
       :aria-label="$t('g.close')"
       @click="handleClose"
     >
-      <i class="pi pi-times" />
+      <i class="pi pi-times" aria-hidden="true" />
     </Button>
 
     <div
--- a/src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
+++ b/src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
@@ -15,6 +15,7 @@
     <div class="flex flex-col items-center gap-3">
       <div
         class="flex size-10 items-center justify-center rounded-xl bg-muted-foreground/30 text-lg font-semibold text-white"
+        aria-hidden="true"
       >
         <!-- Decorative initial for "Personal" workspace icon; not user-facing text -->
         P
--- a/src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
+++ b/src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
@@ -16,6 +16,7 @@
       <div
         class="flex size-10 items-center justify-center rounded-xl bg-muted-foreground/30 text-lg font-semibold text-white"
       >
+        <!-- Decorative initial for "Personal" workspace icon; not user-facing text -->
         P
       </div>
       <h2 class="m-0 font-inter text-2xl font-semibold text-base-foreground">
--- a/src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
+++ b/src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
@@ -1,7 +1,7 @@
 <template>
   <div
     v-if="showCustomPricingTable"
-    class="relative flex h-full flex-col gap-6 overflow-y-auto! p-4 pt-8 md:px-16 md:py-8"
+    class="relative flex h-full flex-col gap-6 overflow-y-auto p-4 pt-8 md:px-16 md:py-8"
   >
     <Button
       size="icon"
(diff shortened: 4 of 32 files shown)


src/locales/en/main.json
```diff
@@ -2249,7 +2249,12 @@
     "topupTimeout": "Top-up verification timed out"
   },
   "subscription": {
-    "chooseBestPlanWorkspace": "Choose the best plan for your workspace",
+    "plansForWorkspace": "Plans for {workspace}",
```
christian-byrne, line 2256: issue: The `chooseBestPlanWorkspace` key was removed here, but the same key remains in all 11 non-English locale files (tr, pt-BR, zh, ja, ru, fr, zh-TW, fa, ko, es, ar). This leaves dead translation data that will confuse translators.

src/locales/en/main.json
```diff
@@ -2249,7 +2249,12 @@
...
+    "personalWorkspace": "Personal Workspace",
+    "teamWorkspace": "Team Workspace",
```
christian-byrne, line 2258: suggestion (non-blocking): Six new `subscription.*` keys, the entire `teamWorkspacesDialog` block (9 keys), and `workspaceSwitcher.failedToSwitch` are added in this PR but missing from all non-English locale files. While vue-i18n falls back to English, adding placeholder entries lets translators discover them.

src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
```diff
@@ -245,6 +261,15 @@ const handleSubscribed = () => {
   emit('close', true)
 }
 
+const handleChooseTeam = () => {
```
christian-byrne, line 264: suggestion: The script block has grown to mix polling logic, subscription-status watching, multiple event handlers, and UI state management in a single scope. Consider extracting polling logic into a composable (e.g., `useSubscriptionPolling`) to reduce cognitive complexity and improve testability.

src/platform/cloud/subscription/composables/useSubscriptionDialog.test.ts
```diff
@@ -1,77 +1,196 @@
-import { describe, expect, it, vi } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
```
christian-byrne, line 1: suggestion: This test file was fully rewritten. The original tests from PR #9958 (preventing subscription UI on non-cloud distributions) appear preserved, but the mock setup changed significantly (firebase mocks removed, vi.fn patterns changed). Worth verifying the original bug scenario still passes -- specifically that `showPricingTable()` is a no-op on non-cloud distributions.

src/platform/workspace/components/dialogs/TeamWorkspacesDialogContent.vue
```diff
@@ -0,0 +1,217 @@
...
+  } catch (error) {
+    toast.add({
+      severity: 'error',
+      summary: t('workspaceSwitcher.failedToSwitch'),
+      detail: error instanceof Error ? error.message : t('g.unknownError')
+    })
+  }
+}
+
+async function onCreate() {
```
christian-byrne, line 201: suggestion (non-blocking): `onCreate` has 3 nested try-catch blocks reaching 4 levels of nesting. An early-return pattern or named helpers (e.g., `tryCreateWorkspace`, `tryConfirm`) would flatten this.

src/platform/workspace/components/dialogs/TeamWorkspacesDialogContent.vue
```diff
@@ -0,0 +1,217 @@
...
+            <div class="flex min-w-0 flex-1 flex-col items-start gap-1">
+              <div class="flex items-center gap-1.5">
+                <span
+                  class="truncate text-left text-sm font-medium text-base-foreground"
+                >
+                  {{ workspace.name }}
+                </span>
+                <span
+                  v-if="tierLabels.get(workspace.id)"
+                  class="shrink-0 rounded-full bg-base-foreground px-1 py-0.5 text-[10px] font-bold text-base-background uppercase"
```
christian-byrne, line 65: issue: `text-[10px]` is below the 14pt threshold where WCAG 2.1 AA requires a 4.5:1 contrast ratio minimum. At this size the tier badge text may be effectively unreadable for low-vision users. Consider increasing to at least `text-xs` (12px).

src/platform/workspace/components/dialogs/TeamWorkspacesDialogContent.vue
```diff
@@ -0,0 +1,217 @@
...
+      <h3
+        class="m-0 text-xs font-semibold tracking-wider text-muted-foreground uppercase"
+      >
+        {{ $t('teamWorkspacesDialog.newWorkspace') }}
+      </h3>
+      <div class="flex flex-col gap-2">
+        <label for="workspace-name-input" class="text-sm text-base-foreground">
+          {{ $t('workspacePanel.createWorkspaceDialog.nameLabel') }}
+        </label>
+        <input
```
christian-byrne, line 100: suggestion (non-blocking): The workspace name input silently disables the submit button when invalid without explaining why. Screen reader users receive no feedback about what is wrong. Adding `aria-describedby` pointing to a validation message and `aria-invalid="true"` when invalid would help.

src/platform/workspace/components/PricingTableWorkspace.vue
```diff
@@ -482,7 +506,13 @@ const getAnnualTotal = (tier: PricingTierConfig): number => {
   return plan ? plan.price_cents / 100 : tier.pricing.yearly * 12
 }
 
-const getMaxMembers = (tier: PricingTierConfig): number => getMaxSeats(tier.key)
+const maxMembersByTier = computed(
+  () =>
+    Object.fromEntries(tiers.map((t) => [t.key, getMaxSeats(t.key)])) as Record<
```
christian-byrne, line 511: suggestion (non-blocking): The `.map((t) => ...)` parameter `t` shadows the `t` from `useI18n()`. If someone adds an i18n call inside this callback, it would silently call the tier object. Renaming to `tier` avoids the trap: `tiers.map((tier) => [tier.key, getMaxSeats(tier.key)])`.

src/platform/workspace/components/WorkspaceSwitcherPopover.vue
```diff
@@ -141,27 +142,8 @@ const emit = defineEmits<{
...
-  isYearly: boolean
-): string {
-  if (!tier) return ''
-  const key = tierKeyMap[tier] ?? 'standard'
-  const baseName = t(`subscription.tiers.${key}.name`)
-  return isYearly
-    ? t('subscription.tierNameYearly', { name: baseName })
-    : baseName
-}
+const { formatTierName, getTierLabel: getStoreTierLabel } =
```
christian-byrne: nitpick (non-blocking): The destructured rename `getTierLabel: getStoreTierLabel` followed by a local function also named `getTierLabel` that wraps it creates a naming collision. Consider renaming one to reduce the indirection (e.g., `getWorkspaceTierLabel` from the composable, or `resolveTierLabel` for the local wrapper).

src/platform/workspace/composables/useCreateWorkspaceUrlLoader.ts
```diff
@@ -0,0 +1,71 @@
...
+    router.replace({ query: newQuery }).catch((error) => {
+      console.warn(
+        '[useCreateWorkspaceUrlLoader] Failed to clean URL params:',
+        error
+      )
+    })
+    clearPreservedQuery(NAMESPACE)
+
+    try {
+      await dialogService.showCreateWorkspaceDialog()
```
christian-byrne: question: This calls `dialogService.showCreateWorkspaceDialog()` (the simple dialog), while the rest of this PR introduces `showTeamWorkspacesDialog()` for the team-aware flow. Is it intentional that `?create_workspace=1` opens the simpler dialog rather than the new team workspace dialog?

src/platform/workspace/composables/useWorkspaceTierLabel.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function useWorkspaceTierLabel() {
+  const { t } = useI18n()
+
+  function formatTierName(
+    tier: string | null | undefined,
+    isYearly: boolean
+  ): string {
+    if (!tier) return ''
+    const key = tierKeyMap[tier] ?? 'standard'
```
christian-byrne: suggestion (non-blocking): `tierKeyMap[tier] ?? 'standard'` silently maps any unrecognized tier string to "Standard". Meanwhile `getTierLabel` on line 35 returns `null` for unrecognized tiers -- inconsistent handling. Returning `null` or empty string for unknown tiers (or adding a comment explaining the fallback) would be clearer.

src/platform/workspace/composables/useWorkspaceTierLabel.ts
```diff
@@ -0,0 +1,55 @@
...
+  FOUNDERS_EDITION: 'founder'
+}
+
+interface WorkspaceSubscriptionInfo {
+  isSubscribed: boolean
+  subscriptionPlan: string | null
+  subscriptionTier: SubscriptionTier | null
+}
+
+export function useWorkspaceTierLabel() {
```
christian-byrne, line 20: nitpick (non-blocking): `useWorkspaceTierLabel`, `formatTierName`, and `getTierLabel` are exported and used across multiple components but have no JSDoc. A brief description of purpose and return values would help consumers.

christian-byrne (commented):
**Additional findings on pre-existing lines (not in diff):**

---

`src/components/graph/GraphCanvas.vue:13`

note (non-blocking): Line 9 checks `workflowTabsPosition === 'Topbar'` in a `v-if`, and inside that block line 13 checks `workflowTabsPosition !== 'Topbar'`, which is always false. The inner native drag area div never renders. Pre-existing, not introduced by this PR.

---

`src/platform/cloud/subscription/components/PricingTable.vue:248` / `PricingTableWorkspace.vue:266`

note (non-blocking): `v-html="'&rarr;'"` bypasses Vue's XSS protections unnecessarily. The value is hardcoded and safe, but could be replaced with the Unicode character directly: `<span class="no-underline">→</span>`. Pre-existing in PricingTable.vue; present in PricingTableWorkspace.vue as well.

---

`src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue:191`

note (non-blocking): `computed(() => isSubscriptionEnabled())` wraps a non-reactive function that reads build-time constants. The computed never re-evaluates reactively. A plain `const showCustomPricingTable = isSubscriptionEnabled()` would be equivalent.

---

`src/platform/workspace/components/SubscriptionRequiredDialogContentWorkspace.vue:13`

issue: The `<i class="pi pi-arrow-left">` (line 13) and `<i class="pi pi-times">` (line 23) inside icon-only Buttons are missing `aria-hidden="true"`. Screen readers may announce icon font ligature text. The same icons in `SubscriptionRequiredDialogContent.vue` were fixed in this PR with `aria-hidden="true"` but these were missed.

---

`src/platform/workspace/components/SubscriptionRequiredDialogContentWorkspace.vue:209`

issue: `handleAddCreditCard()` (lines 209-261) and `handleConfirmTransition()` (lines 263-315) have nearly identical ~50-line bodies. Both call `subscribe()` with the same args, handle the same three response-status branches, and have identical catch/finally blocks. Only the fallback error message differs. Extracting a shared helper (e.g., `executeSubscription(errorMessage: string)`) would eliminate the duplication.

---

`src/platform/workspace/components/SubscriptionRequiredDialogContentWorkspace.vue:239`

suggestion (non-blocking): `window.open(response.payment_method_url, '_blank')` (lines 239, 293) opens a URL from the API without client-side validation. If the API were compromised, this could be an open redirect. Consider validating that `payment_method_url` starts with `https://` and belongs to an expected domain.

---

`src/platform/workspace/components/WorkspaceSwitcherPopover.vue:72`

issue: The "Create workspace" action uses a `<div>` with `@click` (line 72/81) instead of a `<button>`. No `role`, `tabindex`, or `@keydown` handler. Keyboard-only users cannot focus or activate this control.

---

`src/platform/workspace/components/WorkspaceSwitcherPopover.vue:34`

issue: Each workspace row `<button>` has no `aria-label`. Screen readers read flattened text content but the button's purpose ("switch to this workspace") is not communicated. Consider adding `:aria-label` describing the action.

src/locales/en/main.json
christian-byrne: note (non-blocking): vue-i18n falls back to English so this doesn't break anything. adding placeholder entries across 11 locales is mechanical work better done separately.

src/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue
christian-byrne: suggestion (non-blocking): agreed this would improve testability, but it is pre-existing complexity not introduced by this PR - better as a followup refactor.

src/platform/cloud/subscription/composables/useSubscriptionDialog.test.ts
christian-byrne: the non-cloud distribution guard test is preserved - the mock setup changed but the test still verifies showPricingTable is a no-op when isCloud is false (see the 'does not show dialog on non-cloud distribution' test case)

src/platform/workspace/components/PricingTableWorkspace.vue
christian-byrne: suggestion (non-blocking): fair point on the shadow risk - the callback only accesses tier properties so it is safe today, but renaming would prevent future confusion.

src/platform/workspace/composables/useWorkspaceTierLabel.ts
christian-byrne: nit (non-blocking): agreed JSDoc would help here, but the function signatures and naming are fairly self-documenting. can add in a followup.

src/platform/workspace/components/dialogs/TeamWorkspacesDialogContent.vue
christian-byrne: note: the 10px badge styling matches WorkspaceSwitcherPopover (line 51) - this is a consistent design pattern for tier badges across the workspace UI. the badge is supplementary info (workspace name is always readable), and the high-contrast combo (bg-base-foreground on bg-base-background with font-bold uppercase) helps compensate. worth revisiting if the design system adds a badge component.

src/platform/workspace/components/WorkspaceSwitcherPopover.vue
```diff
@@ -196,33 +177,12 @@ function getRoleLabel(role: AvailableWorkspace['role']): string {
...
-
-  // Extract tier from plan slug (e.g., "CREATOR_MONTHLY" -> "CREATOR")
-  const tierMatch = Object.keys(tierKeyMap).find((tier) =>
-    planSlug.startsWith(tier)
-  )
-  if (!tierMatch) return null
-
-  const isYearly = planSlug.includes('YEARLY') || planSlug.includes('ANNUAL')
-  return formatTierName(tierMatch, isYearly)
+  return getTierLabel(workspace)
```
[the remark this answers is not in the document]
christian-byrne: this is already handled - line 182 returns currentSubscriptionTierName or null, and the v-if on line 50 hides the badge when null. during initial load the badge just disappears briefly rather than showing stale data, which is the correct behavior.

src/platform/cloud/subscription/components/PricingTable.vue
```diff
@@ -95,7 +95,22 @@
...
+            role="note"
+            :aria-label="t('subscription.soloUseOnly')"
+            class="m-0 flex h-10 items-center rounded-lg bg-muted-foreground/30 px-3 text-sm text-muted-foreground"
+          >
+            {{ t('subscription.soloUseOnly') }}
+            <span class="mx-1 text-muted-foreground">–</span>
+            <button
+              class="text-primary-foreground cursor-pointer border-none bg-transparent p-0 text-sm font-medium underline hover:text-base-foreground focus-visible:ring-1 focus-visible:outline-none"
+              @click="emit('chooseTeamWorkspace')"
+            >
```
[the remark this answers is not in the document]
christian-byrne: nit (non-blocking): this component is not rendered inside a form, and the button has an explicit click handler. the default type=submit only fires when inside a form element, which this is not. low risk, but can add in a followup if desired.

src/platform/cloud/subscription/composables/useSubscriptionDialog.ts
```diff
@@ -42,13 +43,20 @@ export const useSubscriptionDialog = () => {
             import('@/platform/cloud/subscription/components/SubscriptionRequiredDialogContent.vue')
         )
 
+    const personalProps = {
+      onClose: hide,
+      reason: options?.reason,
+      onChooseTeam: () => startTeamWorkspaceUpgradeFlow()
+    }
```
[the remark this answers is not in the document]
christian-byrne: note: the team workspace flow triggers a full page reload (workspace switch), so the reason would need to be persisted in sessionStorage alongside RESUME_PRICING_KEY. the generic pricing dialog is acceptable post-reload since the user is now in a team workspace context where the original paywall reason (out_of_credits etc) may no longer apply. keeping this simple for now.

src/platform/workspace/auth/WorkspaceAuthGate.test.ts
```diff
@@ -142,6 +156,16 @@ describe('WorkspaceAuthGate', () => {
...
 
+    it('calls resumePendingPricingFlow after successful workspace init', async () => {
+      mockTeamWorkspacesEnabled.value = true
+      mockWorkspaceStoreInitState.value = 'ready'
+
+      mountComponent()
+      await flushPromises()
+
+      expect(mockResumePendingPricingFlow).toHaveBeenCalled()
+    })
```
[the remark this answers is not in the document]
christian-byrne: note: the test verifies that resumePendingPricingFlow is called when the store is ready, which is the critical integration point. driving the state transition from the mock would test WorkspaceAuthGate's internal wiring rather than the resume behavior itself. the current approach is simpler and tests the actual contract.

src/platform/workspace/components/dialogs/TeamWorkspacesDialogContent.vue
```diff
@@ -0,0 +1,217 @@
...
+})
+
+function onCancel() {
+  dialogStore.closeDialog({ key: DIALOG_KEY })
+}
+
+async function handleSwitch(workspaceId: string) {
+  try {
+    await switchWorkspace(workspaceId)
+    dialogStore.closeDialog({ key: DIALOG_KEY })
```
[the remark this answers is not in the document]
christian-byrne: note: switchWorkspace triggers a full page reload, so onConfirm (which sets the sessionStorage resume key) needs to run before the switch. the current flow intentionally separates create (where we control the timing and can set the resume key) from switch (where the reload happens immediately). for the switch path, the resume key should be set before calling switchWorkspace. will address in a followup.

src/platform/workspace/composables/useCreateWorkspaceUrlLoader.ts
```diff
@@ -0,0 +1,71 @@
...
+  async function loadCreateWorkspaceFromUrl() {
+    hydratePreservedQuery(NAMESPACE)
+    const mergedQuery = mergePreservedQueryIntoQuery(NAMESPACE, route.query)
+    if (mergedQuery) {
+      await router.replace({ query: mergedQuery })
+    }
+
+    const query = mergedQuery ?? route.query
+    const param = query.create_workspace
+    if (!param || typeof param !== 'string') return
```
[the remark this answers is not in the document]
christian-byrne: note: the early return on invalid values means preserved query stays in sessionStorage, but the hydration at line 38-42 will merge it back next time and the same guard runs again - so it is a no-op loop, not a re-entry issue. the URL param itself is cleaned by the merge step. low risk but could add a clearPreservedQuery before the early return for cleanliness.

src/platform/workspace/composables/useCreateWorkspaceUrlLoader.ts
```diff
@@ -0,0 +1,71 @@
...
+    })
+    clearPreservedQuery(NAMESPACE)
+
+    try {
+      await dialogService.showCreateWorkspaceDialog()
+    } catch (error) {
+      console.error(
+        '[useCreateWorkspaceUrlLoader] Failed to open create workspace dialog:',
+        error
+      )
```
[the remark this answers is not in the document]
christian-byrne: note: cleaning the URL first prevents an infinite retry loop if the dialog component itself has a loading error (lazy-loaded component that consistently fails would cause a reload loop). the console error gives developers visibility. if the dialog fails to load, the user can re-navigate to the URL. this is the safer ordering.

christian-byrne (approved):
all review feedback addressed and verified in code. dead locale keys removed across all 12 locales, onCreate nesting flattened, a11y attributes added to workspace name input, tier label naming cleaned up, URL loader updated to use team dialog. non-blocking items (locale placeholders, polling extraction, JSDoc, variable shadowing, switch-path resume flow) tracked via coderabbit followup issues. lgtm

src/platform/workspace/components/PricingTableWorkspace.vue
christian-byrne, line 511: nit: `t` in `tiers.map((t) => ...)` shadows the `t` from `useI18n()` at line 351. Rename the callback param.

```suggestion
    Object.fromEntries(tiers.map((tier) => [tier.key, getMaxSeats(tier.key)])) as Record<
```

christian-byrne (commented):
Follow-up pass on `SubscriptionRequiredDialogContentWorkspace.vue` and `PricingTableWorkspace.vue`. Three file-level issues below, one inline nit.

---

**`src/platform/workspace/components/SubscriptionRequiredDialogContentWorkspace.vue`**

1. issue: hardcoded English strings in toast messages (lines 170, 180, 195, 252, 255, 306, 309, 330, 333). All user-facing strings must use `t()` per AGENTS.md. The personal variant avoids this by delegating subscribing to child components, but this file inlines the subscribe flow and skips i18n for error toasts.

2. issue: payment redirect URLs are hardcoded to production (`https://www.comfy.org/payment/success` and `/failed`) in 4 places (lines 221-222, 275-276). Staging deploys will redirect users to production. `src/config/comfyApi.ts` already exports `getComfyPlatformBaseUrl()` which respects `__USE_PROD_CONFIG__` and remote config -- derive URLs from that.

3. suggestion: `handleAddCreditCard` (lines 209-261) and `handleConfirmTransition` (lines 263-315) are nearly identical ~50-line functions differing only in the fallback error message string. Extract a shared `executeSubscription(fallbackErrorKey)` helper within this file.

christian-byrne (commented):
A few more things from a closer look at the new composables and dialog flow.

---

**`src/platform/workspace/composables/useWorkspaceTierLabel.ts`**

4\. question: `formatTierName` changed behavior for unknown tiers. The old code in WorkspaceSwitcherPopover had `tierKeyMap[tier] ?? 'standard'` (unknown tiers fall back to showing "Standard"). The new extraction has `if (!key) return ''` (unknown tiers show no label). Is this intentional? If a new tier is added server-side before the frontend ships the matching key, subscribed workspaces with that tier would silently lose their label badge.

**`src/platform/workspace/composables/useCreateWorkspaceUrlLoader.ts`**

5\. issue: URL param cleanup happens before the dialog action. Compare with `useInviteUrlLoader`, which cleans up in a `finally` block after the action completes. Here, `create_workspace` is deleted from the URL and `clearPreservedQuery` is called (lines 48-56) before `showTeamWorkspacesDialog` (line 59). If the dialog import fails, the user can't retry by refreshing. Move the cleanup into a `finally` block after the dialog call to match the invite loader pattern.

**`src/platform/workspace/components/dialogs/TeamWorkspacesDialogContent.vue`**

6\. suggestion: `onCreate` sets `loading.value = false` on two separate paths (line 213, line 226) instead of using `try/finally`. If an unexpected error is thrown between the two try blocks (e.g. `dialogStore.closeDialog` throws), the button stays in a loading state permanently. Wrapping the whole function body in `try/finally` with `loading.value = false` in the finally block is more resilient.

src/components/graph/GraphCanvas.vue
```diff
@@ -434,8 +435,9 @@ useEventListener(
 const comfyAppReady = ref(false)
 const workflowPersistence = useWorkflowPersistence()
 const { flags } = useFeatureFlags()
-// Set up invite loader during setup phase so useRoute/useRouter work correctly
+// Set up URL loaders during setup phase so useRoute/useRouter work correctly
 const inviteUrlLoader = isCloud ? useInviteUrlLoader() : null
+const createWorkspaceUrlLoader = isCloud ? useCreateWorkspaceUrlLoader() : null
```
christian-byrne, line 440: In both of these cases, the "this is a no-op on cloud" should be fully encapsulated and the nuance/weirdness should not be exposed to consumers like GraphCanvas.vue.

The fact that the returned object can be `null` should be expressed in the type.

src/components/graph/GraphCanvas.vue
```diff
@@ -555,6 +557,18 @@ onMounted(async () => {
...
+  if (createWorkspaceUrlLoader && flags.teamWorkspacesEnabled) {
+    try {
+      await createWorkspaceUrlLoader.loadCreateWorkspaceFromUrl()
+    } catch (error) {
+      console.error(
+        '[GraphCanvas] Failed to load create workspace from URL:',
+        error
+      )
+    }
+  }
```
christian-byrne, line 570: Similar to above, this nuance like feature flag, error handling, should all be in the module and not something consumers should need to understand.

src/platform/cloud/subscription/components/PricingTable.vue
```diff
@@ -67,7 +67,7 @@
             <div class="flex flex-col gap-2">
               <div class="flex flex-row items-baseline gap-2">
                 <span
-                  class="font-inter text-[32px] leading-normal font-semibold text-base-foreground"
+                  class="font-inter text-[28px] leading-normal font-semibold text-base-foreground"
```
christian-byrne, line 70: @coderabbitai make an Issue to add a linter rule that enforces tailwind units when one is available for the given `-[npx]` usage.

Not for this PR as it was pre-existing, and should be something enforced in the linter.

src/platform/workspace/composables/useWorkspaceTierLabel.ts
```diff
@@ -0,0 +1,56 @@
...
+import type { SubscriptionTier } from '@/platform/workspace/api/workspaceApi'
+
+const tierKeyMap: Record<string, string> = {
+  FREE: 'free',
+  STANDARD: 'standard',
+  CREATOR: 'creator',
+  PRO: 'pro',
+  FOUNDER: 'founder',
+  FOUNDERS_EDITION: 'founder'
+}
```
christian-byrne, line 12: question: This doesn't exist already in the codebase? If not, it doesn't seem like it belongs in a module that's for view-layer labeling semantics and rather belong in data layer where interfaces and api contracts are defined.

src/platform/cloud/subscription/composables/useSubscriptionDialog.ts
```diff
@@ -101,9 +109,58 @@ export const useSubscriptionDialog = () => {
...
+    dialogService
+      .showTeamWorkspacesDialog(() => {
+        try {
+          sessionStorage.setItem(RESUME_PRICING_KEY, '1')
+        } catch {
+          // sessionStorage may be unavailable
+        }
+      })
+      .catch((error) => {
+        console.error(
```
christian-byrne, line 133: Do you want to have a specific error/tag/context in Sentry for this scenario that you can easily check later?

src/platform/cloud/subscription/composables/useSubscriptionDialog.ts
```diff
@@ -42,13 +43,20 @@ export const useSubscriptionDialog = () => {
...
 
+    const personalProps = {
+      onClose: hide,
+      reason: options?.reason,
+      onChooseTeam: () => startTeamWorkspaceUpgradeFlow()
+    }
+    const workspaceProps = {
+      onClose: hide,
+      reason: options?.reason
+    }
```
christian-byrne, line 54: Decent use-case for `Omit`

src/platform/cloud/subscription/composables/useSubscriptionDialog.ts
```diff
@@ -101,9 +109,58 @@ export const useSubscriptionDialog = () => {
...
+   * Start the two-stage team workspace upgrade flow:
+   * 1. Close the current pricing dialog
+   * 2. Open the create workspace dialog
+   * 3. On successful creation, persist a resume intent so the team pricing
+   *    dialog reopens automatically after the page reload
+   *
+   * Uses sessionStorage (not a store) because the intent must survive
+   * a full page reload triggered by workspace switching.
+   */
+  function startTeamWorkspaceUpgradeFlow() {
```
christian-byrne, line 122: suggestion (non-blocking): The callback-prop chain (`onChooseTeam` -> `handleChooseTeam` -> `startTeamWorkspaceUpgradeFlow` -> `showTeamWorkspacesDialog(onConfirm)` -> sessionStorage -> `resumePendingPricingFlow`) threads state across 4 components and a page reload via a hand-rolled sessionStorage flag. The codebase already has `preservedQueryManager` + URL loaders (`useCreateWorkspaceUrlLoader`, `useInviteUrlLoader`) for exactly this "survive a reload" pattern. Consider using a query param like `?show_pricing=1` instead -- it eliminates `onChooseTeam` prop-threading, the `onConfirm` callback, `RESUME_PRICING_KEY`, and `resumePendingPricingFlow` entirely. The URL is self-describing and debuggable.

src/platform/workspace/auth/WorkspaceAuthGate.vue
```diff
@@ -86,6 +88,14 @@ async function initialize(): Promise<void> {
...
     // Step 5: WORKSPACE MODE - Full initialization
     await initializeWorkspaceMode()
+
+    // Step 6: Resume any pending pricing flow from team workspace creation
+    // Only safe after workspace store initialized successfully — the pricing
+    // dialog reads workspace state to decide which variant to show.
+    const workspaceStore = useTeamWorkspaceStore()
+    if (workspaceStore.initState === 'ready') {
+      subscriptionDialog.resumePendingPricingFlow()
+    }
```
christian-byrne, line 98: issue (non-blocking): If `initializeWorkspaceMode()` throws, `initState` won't be `'ready'` and the sessionStorage flag is never consumed. It persists silently and could fire on an unrelated future page load, or never fire at all -- the user completed the workspace creation flow but the pricing table never appears. Consider always calling `resumePendingPricingFlow()` (it already guards against personal workspace internally), or using the URL param approach where the param is cleaned up unconditionally on load.

src/platform/cloud/subscription/composables/useSubscriptionDialog.ts
```diff
@@ -8,6 +8,7 @@ import { useTeamWorkspaceStore } from '@/platform/workspace/stores/teamWorkspace
 
 const DIALOG_KEY = 'subscription-required'
 const FREE_TIER_DIALOG_KEY = 'free-tier-info'
+const RESUME_PRICING_KEY = 'comfy:resume-team-pricing'
```
christian-byrne, line 11: nitpick (non-blocking): The try/catch sessionStorage read-write-remove pattern here is a manual reimplementation of what `preservedQueryManager` already abstracts (with validation, error handling, and cleanup). If the URL param approach isn't adopted, this could at least use `preservedQueryManager` to stay consistent with the other reload-surviving flows (invite, create_workspace, share, template).

PR #9909, opened 2026-03-14
fix: gate cloud API calls behind Firebase authentication

mattmillerai:
## Summary

- On cloud deployments, the bootstrap sequence was firing authenticated API calls (`/api/users`, `/api/settings`, `/api/userdata`, `/api/i18n`) before verifying the user was authenticated via Firebase, causing repeated 401 responses
- Moves the Firebase auth gate to the top of `startStoreBootstrap()` and waits for both `isInitialized` **and** `isAuthenticated` before making any API calls
- The router guard already redirects unauthenticated users to login; once they authenticate, `isAuthenticated` becomes `true` and the bootstrap proceeds normally

## Test plan

- [ ] Verify on cloud deployment: unauthenticated users see no 401 API errors in the network tab
- [ ] Verify on cloud: after login, settings/i18n/workflows load correctly
- [ ] Verify non-cloud deployments are unaffected (no `isCloud` guard hit)
- [ ] Unit tests pass (`pnpm test:unit -- --run src/stores/bootstrapStore.test.ts`)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9909-fix-gate-cloud-API-calls-behind-Firebase-authentication-3236d73d3650819287e4e4c68623463b) by [Unito](https://www.unito.io)


src/stores/bootstrapStore.ts
```diff
@@ -36,14 +36,17 @@ export const useBootstrapStore = defineStore('bootstrap', () => {
...
-    const userStore = useUserStore()
-    await userStore.initialize()
-
     if (isCloud) {
-      const { isInitialized } = storeToRefs(useFirebaseAuthStore())
+      const { isInitialized, isAuthenticated } = storeToRefs(
+        useFirebaseAuthStore()
+      )
       await until(isInitialized).toBe(true)
+      await until(isAuthenticated).toBe(true)
```
dante01yoon, line 44: Issue:  `await until(isAuthenticated).` will hang forever if the user is unauthenticated. `onAuthStateChanged` fires with `user === null` for unauthenticated users, setting `isInitialized = true` but leaving `isAuthenticated` as `false`. The router guard redirects to login, but `startStoreBootstrap()` remains permanently awaited with no timeout. The existing `isInitialized` gate in the router (router.ts#L143-L146) uses a 16s timeout — this new until should have one too, or use a different approach (e.g., combine with needsLogin flow).

dante01yoon: Suggestion: Add a timeout and bail out gracefully, mirroring the router guard's pattern:

```ts
async function startStoreBootstrap() {
  if (isCloud) {
    const { isInitialized, isAuthenticated } = storeToRefs(
      useFirebaseAuthStore()
    )
    await until(isInitialized).toBe(true)

    // If user is not authenticated after Firebase init,
    // bail out — the router guard will redirect to login.
    if (!isAuthenticated.value) return
  }

  const userStore = useUserStore()
  await userStore.initialize()

  const { needsLogin } = storeToRefs(userStore)
  await until(needsLogin).toBe(false)

  void loadI18n()
  loadAuthenticatedStores()
}
```

src/stores/bootstrapStore.ts
```diff
@@ -36,14 +36,17 @@ export const useBootstrapStore = defineStore('bootstrap', () => {
...
     }
 
+    const userStore = useUserStore()
+    await userStore.initialize()
```
dante01yoon, line 48: Question:  Moving `userStore.initialize()` after the Firebase gate means `non-cloud` deployments still call `userStore.initialize()` first (correct), but the code ordering now implies userStore.initialize() depends on Firebase auth completing in cloud mode. However, `userStore.initialize()` calls api.getUserConfig() — in cloud mode, the API layer in api.ts#L415-L419 already waits for Firebase isInitialized before making requests. This means the original code may not have been the root cause of the 401s — the real issue might be that the API layer only waits for isInitialized (not isAuthenticated). If so, the fix belongs in the API interceptor, not the bootstrap store.

dante01yoon: Suggestion: Instead of reordering the bootstrap store, fix the root cause in src/scripts/api.ts where the Firebase gate only checks isInitialized but not isAuthenticated:

// In api.ts, wherever the Firebase gate exists (~L415-L419):
```ts
if (isCloud) {
  const { isInitialized, isAuthenticated } = storeToRefs(useFirebaseAuthStore())
  await until(isInitialized).toBe(true)
  if (!isAuthenticated.value) {
    throw new Error('User is not authenticated')
  }
}
```
This approach fixes the 401s at the source (the HTTP layer) rather than restructuring the bootstrap sequence. It would also protect all API calls, not just the ones during bootstrap. If the PR author prefers the bootstrap approach, they should document why the API-layer fix was rejected.

dante01yoon (changes requested):
IMHO, The core idea (don't make authenticated API calls before authentication) is correct, but the implementation has a potential deadlock for unauthenticated cloud users. The `await until(isAuthenticated).toBe(true)` needs either a timeout or a bail-out condition. Additionally, investigating whether the fix should live in the API interceptor layer (which already has a Firebase gate) may be a better architectural fit. WDYT?

src/stores/bootstrapStore.ts
mattmillerai: Makes sense, great catch. Thanks! This has been addressed in: [f8d88f2](https://github.com/Comfy-Org/ComfyUI_frontend/pull/9909/commits/f8d88f2c3ec03d9a0eae74cb157d344e4e5c1d1b)

mattmillerai: Good catch on the hang concern. I initially switched to an early return (`if (!isAuthenticated.value) return`), but that introduced a worse issue — after the user logs in on the login page, `startStoreBootstrap()` is never retriggered, so stores/i18n/settings never load.

Reverted back to `await until(isAuthenticated).toBe(true)`. This works correctly because:

- If the user has a persisted session, `onAuthStateChanged` sets both `currentUser` and `isInitialized` in the same tick, so `isAuthenticated` is immediately `true`.
- If the user has no session, the router guard redirects to login. Once they sign in, `onAuthStateChanged` fires with the user, `isAuthenticated` becomes `true`, and the await resolves.
- The "hang" for users who never log in is a single Vue watcher — cleaned up on app teardown.

Also found and fixed an independent bug in `api.ts`'s `waitForAuthInitialization`: it was passing the Pinia-unwrapped `authStore.isInitialized` (a plain boolean) to `until()` without `.toBe(true)`, so the gate was silently no-oping. Fixed with `storeToRefs` + `.toBe(true)`.

See 7aa7c3144.

mattmillerai: Moved the other independent fix to: https://github.com/Comfy-Org/ComfyUI_frontend/pull/10115

src/stores/bootstrapStore.ts
mattmillerai: Good instinct that the API layer was involved. We found that `waitForAuthInitialization` in `api.ts` was actually a complete no-op — `authStore.isInitialized` is unwrapped by Pinia (plain boolean, not a ref), and `until()` without `.toBe()` returns a builder object, not a promise. So `Promise.race` resolved immediately without waiting.

Fixed separately in #10115 (955195e62).

The bootstrap reordering is still the right call — in cloud mode we should explicitly wait for auth before kicking off store initialization, rather than relying solely on each individual API call to gate itself.

mattmillerai: FYI the reordering was addressed here in this commit: dcf561512

dante01yoon (approved):

src/stores/bootstrapStore.ts
```diff
@@ -36,14 +36,17 @@ export const useBootstrapStore = defineStore('bootstrap', () => {
...
-    await userStore.initialize()
-
     if (isCloud) {
-      const { isInitialized } = storeToRefs(useFirebaseAuthStore())
+      const { isInitialized, isAuthenticated } = storeToRefs(
+        useFirebaseAuthStore()
+      )
       await until(isInitialized).toBe(true)
+      await until(isAuthenticated).toBe(true)
     }
```
christian-byrne, line 45: suggestion: The `await until(isAuthenticated).toBe(true)` has no timeout, unlike the router guard which uses `{ timeout: 16_000 }` for `isInitialized` (see `router.ts:146`). If Firebase initializes successfully but the auth state never resolves to `true` (e.g., corrupted session, network failure during token refresh), the bootstrap will hang indefinitely -- no stores load, no UI renders, and no error is surfaced.

Consider adding a timeout here to match the router guard's defensive pattern:
```ts
await until(isAuthenticated).toBe(true, { timeout: 16_000 })
```
With a catch that redirects to the login page or surfaces an error.

src/stores/bootstrapStore.test.ts
```diff
@@ -95,17 +98,23 @@ describe('bootstrapStore', () => {
       mockDistributionTypes.isCloud = true
     })
 
-    it('waits for Firebase auth before loading i18n and settings', async () => {
```
christian-byrne, line 98: nitpick (non-blocking): The test covers the sequential unlock path (init first, then auth), but doesn't cover the fast path where a returning user has a cached Firebase session and both `isInitialized` and `isAuthenticated` are already `true` when bootstrap starts. A small additional test case would increase confidence that cached sessions don't regress.

christian-byrne (approved):
Looks good. Core logic is sound, local users unaffected, CI green. The two suggestions (timeout on isAuthenticated await, cached-session test) are tagged as CodeRabbit followup issues.

PR #9910, opened 2026-03-14
feat: add TBT/frameDuration metrics and new perf test scenarios

christian-byrne:
## Summary

Adds Total Blocking Time (TBT) and frame duration metrics to the performance testing infrastructure, plus three new test scenarios covering zoom, pan, and many-nodes-idle.

### New Metrics
- **`totalBlockingTimeMs`** — Computed from PerformanceObserver `longtask` entries: `sum(duration - 50ms)` for tasks >50ms. Measures main thread blocking.
- **`frameDurationMs`** — Average frame duration via rAF timing (16.67ms = 60fps target). Measures rendering smoothness.

### New Test Scenarios
| Scenario | Description |
|---|---|
| `canvas-zoom-sweep` | 10 zoom-in + 10 zoom-out cycles on default workflow |
| `canvas-pan-many-nodes` | 10 pan sweeps over 100-node workflow |
| `canvas-many-nodes-idle` | 2-second idle measurement with 100 nodes rendered |

### Infrastructure
- `PerformanceHelper.ts`: Installs PerformanceObserver for longtask, collects TBT, measures frame duration via rAF
- `perf-report.ts`: Reports TBT and frame duration in PR comment tables
- `browser_tests/assets/perf/many_nodes_100.json`: 100-node (10×10 grid) test fixture

## Review Focus
- TBT collection clears entries at `startMeasuring()` and reads at `stopMeasuring()` — ensure no race with observer buffering
- Frame duration sampling uses 10 frames — enough for signal without slowing tests

Depends on: #9886, #9887

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9910-feat-add-TBT-frameDuration-metrics-and-new-perf-test-scenarios-3236d73d365081488ae3c594a8bf7cff) by [Unito](https://www.unito.io)


browser_tests/fixtures/helpers/PerformanceHelper.ts
```diff
@@ -63,12 +65,72 @@ export class PerformanceHelper {
...
+      ) as PerformanceEntry[]
+      let tbt = 0
+      for (const entry of entries) {
+        if (entry.duration > 50) {
+          tbt += entry.duration - 50
+        }
+      }
+      return tbt
+    })
+  }
```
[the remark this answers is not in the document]
christian-byrne: Good catch! `performance.clearResourceTimings()` indeed doesn't affect longtask entries. Fixed by:

1. Removing the ineffective `clearResourceTimings()` call
2. Using `observer.takeRecords()` in `startMeasuring()` to drain any previously buffered longtask entries before a new measurement window
3. Using `observer.takeRecords()` in `collectTBT()` as well, so we only get entries accumulated since the last drain

This ensures longtask entries don't bleed across measurement windows.

browser_tests/tests/performance.spec.ts
```diff
@@ -129,4 +131,115 @@ test.describe('Performance', { tag: ['@perf'] }, () => {
...
+        const store = window.app?.extensionManager
+        return store !== undefined
+      },
+      { timeout: 10000 }
+    )
+
+    // Wait for the output widget to populate (execution_success)
+    await expect(async () => {
+      expect(await (await outputNode.getWidget(0)).getValue()).toBe('foo')
+    }).toPass({ timeout: 10000 })
```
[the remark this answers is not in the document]
christian-byrne: Agreed — removed this unnecessary `waitForFunction`. By the time we've loaded a workflow and queued a prompt, `window.app.extensionManager` is long since defined. The `toPass` block below already correctly polls for the execution result.

browser_tests/tests/performance.spec.ts
```diff
@@ -129,4 +131,115 @@ test.describe('Performance', { tag: ['@perf'] }, () => {
...
+
+  test('minimap idle', async ({ comfyPage }) => {
+    // Enable minimap via setting, load workflow, then measure idle cost
+    await comfyPage.settings.setSetting('Comfy.Minimap.Visible', true)
+    await comfyPage.workflow.loadWorkflow('perf/many_nodes_100')
+
+    // Wait for minimap to render
+    await comfyPage.page
+      .locator('.litegraph-minimap')
+      .waitFor({ state: 'visible', timeout: 5000 })
```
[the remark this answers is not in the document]
christian-byrne: There is no minimap TestId in `selectors.ts` — only `toggleMinimapButton` exists (for the toolbar button, not the minimap container itself). The existing `minimap.spec.ts` uses the same raw `.litegraph-minimap` CSS selector in 4 places, so this is consistent with established patterns. The minimap is rendered by litegraph (canvas-based), not by a Vue component with `data-testid`, so a CSS class selector is the correct approach here.

scripts/perf-report.ts
```diff
@@ -292,7 +298,6 @@ function renderNoBaselineReport(
   for (const [testName, prSamples] of prGroups) {
     for (const { key, label, unit } of REPORTED_METRICS) {
       const prMean = meanMetric(prSamples, key)
-      if (prMean === null) continue
       lines.push(`| ${testName}: ${label} | ${formatValue(prMean, unit)} |`)
```
jaeone94, line 296: `meanMetric()` (line 127) can return `null`, but here it's passed directly to `formatValue` (line 109) which calls `.toFixed(0)`. The other two render functions (`renderFullReport` line 161, `renderColdStartReport` line 258) both have `if (prMean === null) continue` guards -- this one seems to be missing the same guard. 
```typescript
const prMean = meanMetric(prSamples, key)
if (prMean === null) continue  // missing?
lines.push(`| ${testName}: ${label} | ${formatValue(prMean, unit)} |`)
```

browser_tests/fixtures/helpers/PerformanceHelper.ts
```diff
@@ -75,12 +77,76 @@ export class PerformanceHelper {
...
+        ? observer.takeRecords()
+        : (performance.getEntriesByType('longtask') as PerformanceEntry[])
+      let tbt = 0
+      for (const entry of entries) {
+        if (entry.duration > 50) {
+          tbt += entry.duration - 50
+        }
+      }
+      return tbt
+    })
```
jaeone94, line 99: `takeRecords()` might not capture all entries from the measurement window

The new `takeRecords()` approach (commit `b0c7429ce`) looks like a good improvement over the previous `getEntriesByType` approach. One concern though: the observer callback is still empty (line 139-141), and `takeRecords()` only returns entries that haven't been dispatched to the callback yet. If the browser has already dispatched some longtask entries to the (empty) callback between `startMeasuring` and `stopMeasuring`, those entries would be lost -- `takeRecords()` in `collectTBT()` (line 89) wouldn't return them.

Would it be more reliable to collect entries inside the callback itself? Something like:

```typescript
// startMeasuring:
(window as Record<string, unknown>).__perfLongtaskEntries = []
const observer = new PerformanceObserver((list) => {
  ;(window as any).__perfLongtaskEntries.push(...list.getEntries())
})

// collectTBT: read from __perfLongtaskEntries
```

This way both dispatched and undispatched entries are captured. What do you think?

browser_tests/fixtures/helpers/PerformanceHelper.ts
```diff
@@ -75,12 +77,76 @@ export class PerformanceHelper {
...
+              return
+            }
+            const total = timestamps[timestamps.length - 1] - timestamps[0]
+            resolve(total / (timestamps.length - 1))
+          }
+        }
+        requestAnimationFrame(tick)
+      })
+    }, sampleFrames)
+  }
```
jaeone94, line 130: nit: The `page.evaluate` (line 106) creates a Promise that only resolves when enough rAF callbacks fire. If rAF stops firing in a headless CI environment, this hangs until Playwright's 30-second default timeout. That seems like a long wait for what should be a sub-second measurement -- would a safety timeout (e.g. 5 seconds) inside the Promise be worth adding?

```typescript
return new Promise<number>((resolve) => {
  const timeout = setTimeout(() => resolve(0), 5000)
  // ... existing rAF logic, with clearTimeout(timeout) in resolve branch
})
```

browser_tests/fixtures/helpers/PerformanceHelper.ts
```diff
@@ -27,6 +27,8 @@ export interface PerfMeasurement {
   jsHeapTotalBytes: number
   scriptDurationMs: number
   eventListeners: number
+  totalBlockingTimeMs: number
```
jaeone94, line 30: nit: The same interface is defined independently in both files. This PR demonstrates the maintenance cost -- both copies had to be updated with `totalBlockingTimeMs` and `frameDurationMs`. I noticed that `browser_tests/helpers/perfReporter.ts` already imports the canonical definition from `PerformanceHelper.ts`, but `scripts/perf-report.ts` can't do the same since they're in separate tsconfig projects.

Would it be worth extracting the interface to a shared location (e.g. a small `.d.ts` or shared types file)? Or at least adding a comment like `// Keep in sync with PerformanceHelper.ts:17` to link the two copies?

browser_tests/tests/performance.spec.ts
jaeone94, line 1: nit: The new test scenarios (lines 135-214: zoom sweep, pan many nodes, many nodes idle, minimap idle) collect metrics and log them but don't call `expect()`. It looks like regression detection is deferred to `scripts/perf-report.ts` rather than done inline -- is that the intended design? If so, a brief comment at the top of the describe block would help future readers understand why there are no assertions.

jaeone94 (commented):
I left a few comments.
Also, it looks like the lint test is failing.

scripts/perf-report.ts
christian-byrne: good catch, added the null guard in 7ce916be7

browser_tests/fixtures/helpers/PerformanceHelper.ts
christian-byrne: yeah you are right - entries dispatched to the callback were being silently dropped. fixed in dc892af9c: the observer callback now accumulates TBT into a shared `__perfLongtaskState.tbtMs` field, and `collectTBT()` flushes any remaining queued entries via `takeRecords()` before reading the total.

browser_tests/fixtures/helpers/PerformanceHelper.ts
christian-byrne: added 5s safety timeout in dc892af9c - resolves to 0 if rAF stops firing so we don'\''t hang for 30s in CI

browser_tests/fixtures/helpers/PerformanceHelper.ts
christian-byrne: valid concern - the two files live in separate tsconfig projects (browser_tests/tsconfig.json vs the root tsconfig) so they can'\''t import from each other directly. a shared .d.ts would be the right fix but feels low priority for now since the interface is small and stable. I'\''ll add a sync comment linking the two copies for now.

browser_tests/tests/performance.spec.ts
christian-byrne: yeah that'\''s the intended design - the perf tests only collect and record raw measurements via recordMeasurement(). regression detection is deferred to scripts/perf-report.ts which compares metrics against baselines and posts GitHub PR comments with delta analysis. good call on the comment though, I'\''ll add a note at the top of the describe block.

browser_tests/fixtures/helpers/PerformanceHelper.ts
christian-byrne: created followup: #9960

browser_tests/tests/performance.spec.ts
christian-byrne: created followup: #9961

jaeone94 (dismissed):
LGTM
All blocker issues have been resolved, and follow-up issues have been created.

PR #9911, opened 2026-03-14
feat: unified PR report combining bundle size and runtime perf

christian-byrne:
## Summary

Replaces two separate PR comment workflows (bundle size + performance) with a single unified report that posts one combined comment per PR.

## Changes

- **What**: New `pr-report.yaml` aggregator workflow triggers on both `CI: Size Data` and `CI: Performance Report` completions. Finds sibling workflow runs by PR head SHA. Renders combined report via `unified-report.js` (shells out to existing `size-report.js` and `perf-report.ts`). Sections show "pending" or "failed" placeholders when data is unavailable.
- **Breaking**: Removes `pr-size-report.yaml` and `pr-perf-report.yaml`. Legacy `<!-- COMFYUI_FRONTEND_SIZE -->` and `<!-- COMFYUI_FRONTEND_PERF -->` comments are auto-cleaned on first run.
- **Dependencies**: None

## Review Focus

- Concurrency key uses `head_sha` so the later-completing workflow cancels the earlier report run, ensuring the final comment always has both sections.
- Stale-run guard: verifies workflow_run SHA matches the live PR head before posting.
- The `workflow_dispatch` re-trigger path from `pr-size-report.yaml` is not carried forward — the unified workflow handles re-trigger naturally via its dual-trigger design.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9911-feat-unified-PR-report-combining-bundle-size-and-runtime-perf-3236d73d365081baac1cce6f0d9244ac) by [Unito](https://www.unito.io)

PR #9939, opened 2026-03-15
feat: add trend visualization with sparklines to perf report

christian-byrne:
## Summary

Add historical trend visualization (ASCII sparklines + directional arrows) to the performance PR report, showing how each metric has moved over recent commits on main.

## Changes

- **What**: New `sparkline()`, `trendDirection()`, `trendArrow()` functions in `perf-stats.ts`. New collapsible "Trend" section in the perf report showing per-metric sparklines, direction indicators, and latest values. CI workflow updated to download historical data from the `perf-data` orphan branch and switched to `setup-frontend` action with `pnpm exec tsx`.

## Review Focus

- The trend section only renders when ≥3 historical data points exist (gracefully absent otherwise)
- `trendDirection()` uses a split-half mean comparison with ±10% threshold — review whether this sensitivity is appropriate
- The `git archive` step in `pr-perf-report.yaml` is idempotent and fails silently if no perf-history data exists yet on the perf-data branch

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9939-feat-add-trend-visualization-with-sparklines-to-perf-report-3246d73d36508125a6fcc39612f850fe) by [Unito](https://www.unito.io)

### DIFF
--- a/.github/workflows/pr-report.yaml
+++ b/.github/workflows/pr-report.yaml
@@ -173,6 +173,19 @@ jobs:
           path: temp/perf-baseline/
           if_no_artifact_found: warn
 
+      - name: Download perf history from perf-data branch
+        if: steps.pr-meta.outputs.skip != 'true' && steps.find-perf.outputs.status == 'ready'
+        continue-on-error: true
+        run: |
+          if git ls-remote --exit-code origin perf-data >/dev/null 2>&1; then
+            git fetch origin perf-data --depth=1
+            mkdir -p temp/perf-history
+            for file in $(git ls-tree --name-only origin/perf-data baselines/ 2>/dev/null | sort -r | head -10); do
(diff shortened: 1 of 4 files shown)


benceruleanlu (commented):
feel free to (force) merge and iterate, but it will need to change with https://github.com/Comfy-Org/ComfyUI_frontend/pull/9911, and PTAL:

> the new history download path is incompatible with the current perf-data branch contents. I verified origin/perf-data currently contains baselines/perf-*.json, while this change archives perf-history/ and the loader only reads temp/perf-history/<dir>/perf-metrics.json.

christian-byrne:
fixed in 0f3ea483b - stacked this PR on #9911 (which replaces pr-perf-report.yaml with pr-report.yaml), added a history download step that reads from baselines/ using git ls-tree + git show, and updated loadHistoricalReports() to handle flat .json files. also fixed the zero-baseline edge case coderabbit flagged.

ready for re-review @benceruleanlu

PR #9953, opened 2026-03-15
draft: add red-green-fix skill for verified bug fix workflow

dante01yoon:
## Summary

- Add a Claude Code skill (`/red-green-fix`) that enforces the red-green commit pattern for bug fixes
- Ensures a failing test is committed first (red CI), then the fix is committed separately (green CI)
- Gives reviewers proof that the test actually catches the bug

## Test Plan

- [ ] Invoke `/red-green-fix <bug description>` in Claude Code and verify the two-step workflow
- [ ] Confirm PR template includes red-green verification table

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9953-draft-add-red-green-fix-skill-for-verified-bug-fix-workflow-3246d73d365081238f44e2343bae419b) by [Unito](https://www.unito.io)


dante01yoon:
## Open question: CI enforcement for red-green pattern?

### What exists: Skill (opt-in, Claude Code only)

The `/red-green-fix` skill guides Claude Code users through the two-step process:
- Step 1: write only the failing test → commit `test:` → push → wait for CI red
- Step 2: write only the fix → commit `fix:` → push → confirm CI green
- Auto-generates a PR with a red/green verification table

This is **opt-in** — it only works for developers using Claude Code and only when they invoke the skill.

### What a CI workflow would do (enforced on all contributors)

A GitHub Action that triggers on PRs with the `bug` label:

**Rules:**
- Checks commit history: a `test:` prefixed commit must appear before the `fix:` prefixed commit
- Both prefixes must be present — a `fix:` alone would fail the check
- Blocks merge if the pattern isn't followed
- Escape hatch: a `skip-red-green` label bypasses the check (for hotfixes, trivial one-liners)

**Key difference from the Skill:**
| | Skill | CI Workflow |
|---|---|---|
| Scope | Claude Code users only | All contributors |
| Enforcement | Guidance (soft) | Merge blocker (hard) |
| Flexibility | Can adapt to context | Strict commit convention |
| When it runs | During development | At PR review time |

### Design question: alignment with `perf-fix-with-proof`

We already have the `/perf-fix-with-proof` skill which follows a very similar pattern but uses **two stacked PRs** instead of two commits in one PR:
- PR1 (test only) → merges to main → establishes CI baseline
- PR2 (fix) → CI compares against baseline → delta proves improvement

The reason perf uses stacked PRs is that `ci-perf-report.yaml` compares against the **base branch baseline** — the test needs to exist on main first.

For bug fixes, the red/green pattern doesn't need a baseline on main, so two commits in one PR works. But should we align the approach across both skills for consistency?

### Questions for the team

1. Is the skill-level guidance enough, or should we gate it in CI?
2. If CI enforcement, should it apply to all `bug` PRs or only certain ones?
3. Any concerns about requiring `test:` / `fix:` commit prefixes?
4. Should we also update the file structure to match existing convention (`skills/red-green-fix/SKILL.md` instead of `skills/red-green-fix.md`)?

PR #9958, opened 2026-03-15
fix: prevent subscription UI from rendering on non-cloud distributions

christian-byrne:
## Summary

Prevent Plans & Pricing dialog, subscription buttons, and cloud-only menu items from appearing on desktop/localhost distributions.

## Changes

- **What**: Add `isCloud` guards to `useSubscriptionDialog.showPricingTable`, `TopbarSubscribeButton`, and `CurrentUserPopoverLegacy` so subscription UI only renders on cloud
- **Tests**: 24 tests across 3 test files (1 modified, 2 new) covering cloud/non-cloud behavior

## Review Focus

- Guard placement in `CurrentUserPopoverLegacy.vue` — multiple `v-if` conditions updated to include `isCloud`
- Early-return in `showPricingTable` as a defense-in-depth measure

Fixes COM-16820

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9958-fix-prevent-subscription-UI-from-rendering-on-non-cloud-distributions-3246d73d365081559a9ee8650409c5b4) by [Unito](https://www.unito.io)


src/components/topbar/CurrentUserPopoverLegacy.vue
```diff
@@ -91,6 +94,7 @@
     </div>
```
benceruleanlu, line 94: question: this seems API node related, which works on desktop and localhost, are we supposed to hide this?

src/components/topbar/CurrentUserPopoverLegacy.vue
```diff
@@ -29,8 +29,11 @@
       </span>
     </div>
 
-    <!-- Credits Section -->
-    <div v-if="isActiveSubscription" class="flex items-center gap-2 px-4 py-2">
+    <!-- Credits Section (cloud only) -->
+    <div
+      v-if="isCloud && isActiveSubscription"
+      class="flex items-center gap-2 px-4 py-2"
```
benceruleanlu, line 35: question: this seems API node related, which works on desktop and localhost, are we supposed to hide this?

benceruleanlu (approved):
code lgtm, wanted to flag two parts I was unsure if we're supposed to hide since api nodes work on local/desktop

src/components/topbar/CurrentUserPopoverLegacy.vue
christian-byrne: this is the "Partner nodes pricing" menu item - it opens the cloud pricing docs page (docs.comfy.org/tutorials/partner-nodes/pricing). it's about cloud billing for partner nodes, not the API node feature itself, so hiding it on non-cloud distributions is correct

src/components/topbar/CurrentUserPopoverLegacy.vue
christian-byrne: this is the credits balance section (shows cloud credit balance + add credits / upgrade buttons). it'\''s cloud billing UI, not the API node feature itself - so gating on isCloud is correct here

PR #9966, opened 2026-03-15
feat: add Copy URL button to missing model rows for OSS

jaeone94:
## Summary
- Add a "Copy URL" text button in missing model row headers for non-asset-supported models in OSS
- Guard both Copy URL and Download buttons with `!isCloud` to prevent display in Cloud environments
- Button design matches the existing "Download All" button style (`variant="secondary" size="sm"`)

## Test plan
- [ ] Verify Copy URL button appears in OSS missing model rows when model has a URL
- [ ] Verify clicking Copy URL copies the download URL to clipboard
- [ ] Verify Copy URL and Download buttons do NOT appear in Cloud environments
- [ ] Verify no visual regression in Cloud missing model UI

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9966-feat-add-Copy-URL-button-to-missing-model-rows-for-OSS-3246d73d365081fcb753cd07bf7c235d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -3437,6 +3437,7 @@
       "customNodeDownloadDisabled": "Cloud environment does not support model imports for custom nodes in this section. Please use standard loader nodes or substitute with a model from the library below.",
       "importNotSupported": "Import Not Supported",
       "copyModelName": "Copy model name",
+      "copyUrl": "Copy URL",
       "confirmSelection": "Confirm selection",
       "locateNode": "Locate node on canvas",
       "cancelSelection": "Cancel selection",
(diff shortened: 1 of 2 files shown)


christian-byrne (approved):

silveroxides:
Considering all the URL in templates are "resolve" url which directly downlods model either way, isn't this kind of superfluous. Even before the button disappeared, it barely had any proper functionality apart from perhaps if you needed to paste it in a different browser. Just do a string replacement on the URL to replace "resolve" with "blob" instead so it actually links to the page with the model file that has the download button already as well as another direct download url copy alongside a bunch of other information.

jaeone94:
you are absolutely right, I will fix that.

PR #9968, opened 2026-03-15
perf: speed up local inference on shared GPU

kubk:
## Summary

Allows to disable UI animations to speed up local inference on shared GPU setups. Gives 7-13% improvement

## Changes

- **What**: New setting under Appearance > General. Applies a CSS class that disables all animations and transitions
- **Breaking**: None since it's opt in

I also considered auto-disabling animations only during generation on shared GPU setups. We could detect that, lmk if you want it to be implemented.

## Screenshots

Mac with Apple Silicon

Screenshot 2026-03-16 at 01 15 03
Screenshot 2026-03-16 at 01 42 34

## How to reproduce:

- Run frontend & backend locally on a device with shared GPU
- Install the model: `cd models/checkpoints &&  curl -L -o v1-5-pruned-emaonly-fp16.safetensors "https://huggingface.co/Comfy-Org/stable-diffusion-v1-5-archive/resolve/main/v1-5-pruned-emao
  nly-fp16.safetensors"` (~2GB)
- Use this workflow: https://cloud.comfy.org/?share=4086706e0018 (also attached): 
[test-workflow.json](https://github.com/user-attachments/files/26006974/test-workflow.json)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9968-perf-speed-up-local-inference-on-shared-GPU-3246d73d365081adaa03c348a7efdc1e) by [Unito](https://www.unito.io)


christian-byrne:
That's a pretty cool idea

src/platform/settings/constants/coreSettings.ts
```diff
@@ -576,6 +576,19 @@ export const CORE_SETTINGS: SettingParams[] = [
     migrateDeprecatedValue: (value: unknown) =>
       value === 'Integrated' ? 'Default' : value
   },
+  {
```
christian-byrne, line 579: suggestion: The setting ID `Comfy.Appearance.ReduceAnimations` and CSS class `reduce-animations` say "reduce", but the user-facing name says "Disable animations" and the tooltip says "Turns off all". Since the implementation fully kills animations (duration: 0.001ms), this is a disable, not a reduce. Since the setting ID is a persisted contract, consider aligning before merge -- either rename ID/class to `DisableAnimations`/`disable-animations`, or soften the UI label to "Reduce animations".

src/platform/settings/constants/coreSettings.ts
```diff
@@ -576,6 +576,19 @@ export const CORE_SETTINGS: SettingParams[] = [
...
+    id: 'Comfy.Appearance.ReduceAnimations',
+    category: ['Appearance', 'General'],
+    name: 'Disable animations',
+    type: 'boolean',
```
christian-byrne, line 583: suggestion: The tooltip says "Turns off all UI animations" but this CSS-only approach won't affect canvas/WebGL rendering, requestAnimationFrame-driven animations, or Vue Transition components using JS hooks. Consider: "Turns off most CSS animations and transitions. Speeds up inference when the display GPU is also used for generation."

src/platform/settings/constants/coreSettings.ts
```diff
@@ -576,6 +576,19 @@ export const CORE_SETTINGS: SettingParams[] = [
     migrateDeprecatedValue: (value: unknown) =>
       value === 'Integrated' ? 'Default' : value
   },
+  {
+    id: 'Comfy.Appearance.ReduceAnimations',
+    category: ['Appearance', 'General'],
+    name: 'Disable animations',
```
christian-byrne, line 582: suggestion (non-blocking): Users with OS-level `prefers-reduced-motion: reduce` don't automatically get this setting enabled. Could initialize the default from `window.matchMedia('(prefers-reduced-motion: reduce)').matches` so users who've already opted in at the OS level get it for free.

src/assets/css/style.css
```diff
@@ -1,5 +1,12 @@
 @import '@comfyorg/design-system/css/style.css';
 
+.reduce-animations *,
+.reduce-animations *::before,
+.reduce-animations *::after {
+  animation-duration: 0.001ms !important;
```
christian-byrne, line 8: nitpick (non-blocking): The `0.001ms` instead of `0s` is clever -- it ensures transitionend/animationend events still fire so JS listeners aren't broken. Worth a CSS comment so future contributors don't "fix" it to `0s`.

DrJKL (approved):

PR #9940, opened 2026-03-15
test: add large-graph perf test with 245-node workflow (backlog N5)

christian-byrne:
## What

Adds a 245-node workflow asset and two `@perf` tests to establish a baseline for large-graph performance regressions (Tier 6 in the performance backlog).

## Why

Backlog item N5: we need CI regression detection for compositor layer management, GPU texture count, and transform pane cost at 245+ nodes. This is PR1 of 2 — establishes baseline metrics on main. Future optimization PRs will show improvement deltas against this baseline.

## Tests Added

- **`large graph idle rendering`** — 120 frames idle with 245 nodes, measures style recalcs, layouts, task duration, heap delta
- **`large graph pan interaction`** — middle-click pan across 245 nodes, stresses compositor layer management and transform recalculation

## Workflow Asset

`browser_tests/assets/large-graph-workflow.json` — 245 nodes (49 pipelines of CheckpointLoader → 2× CLIPTextEncode → KSampler + EmptyLatentImage), 294 links. Minimal structure focused on node count.

## Verification

- [x] `pnpm typecheck:browser` passes
- [x] `pnpm lint` passes (eslint on changed file)
- [x] All link references in JSON validated programmatically

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-9940-test-add-large-graph-perf-test-with-245-node-workflow-backlog-N5-3246d73d365081f6b5d8ddb9a85e6ad0) by [Unito](https://www.unito.io)

PR #10001, opened 2026-03-16
test: add vue renderer perf tests (idle, pan, zoom culling)

christian-byrne:
Add 3 perf tests exercising Vue renderer (Nodes 2.0) with the 245-node large-graph-workflow:

- **vue renderer large graph idle** — enables Vue nodes, idles 120 frames, measures style recalcs/layouts/DOM nodes. Catches containment regressions.
- **vue renderer large graph pan** — middle-click pans 60 frames, measures frameDurationMs/TBT. Catches TransformPane falling back to reactive style bindings.
- **vue renderer zoom out culling** — zooms out 20 steps (triggers size-based culling at <4px), idles 60 frames, zooms back in. Catches viewport culling regressions.

Complements existing LiteGraph renderer tests for direct A/B comparison in the perf report. Uses existing `large-graph-workflow.json` asset from #9940.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10001-test-add-vue-renderer-perf-tests-idle-pan-zoom-culling-3256d73d365081b4b714d8f3b396604c) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/performance.spec.ts
+++ b/browser_tests/tests/performance.spec.ts
@@ -222,78 +222,76 @@ test.describe('Performance', { tag: ['@perf'] }, () => {
     )
   })
 
-  test('vue renderer large graph idle', async ({ comfyPage }) => {
-    await comfyPage.settings.setSetting('Comfy.VueNodes.Enabled', true)
-    await comfyPage.workflow.loadWorkflow('large-graph-workflow')
-    await comfyPage.vueNodes.waitForNodes()
-    await comfyPage.perf.startMeasuring()
+  test.describe('vue renderer large graph', () => {
+    test.beforeEach(async ({ comfyPage }) => {
+      await comfyPage.settings.setSetting('Comfy.VueNodes.Enabled', true)
+      await comfyPage.workflow.loadWorkflow('large-graph-workflow')
+      await comfyPage.vueNodes.waitForNodes()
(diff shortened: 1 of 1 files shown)


jaeone94 (commented):
**Nit:** The three Vue tests repeat the same setup (`setSetting` + `loadWorkflow` + `waitForNodes`). A nested `test.describe` with `beforeEach` could reduce duplication if more Vue perf tests are planned.

**Minor:** `canvasOps.zoom()` already calls `nextFrame()` internally (`CanvasHelper.ts:32`), so the explicit `await comfyPage.nextFrame()` after each zoom call in the zoom-culling test results in two idle frames per step. This may dilute per-frame metrics compared to the pan test which uses raw `mouse.move` without a wrapper.

jaeone94 (commented):
Nit: The three Vue tests repeat the same setup (setSetting + loadWorkflow + waitForNodes). A nested test.describe with beforeEach could reduce duplication if more Vue perf tests are planned.

Minor: canvasOps.zoom() already calls nextFrame() internally (CanvasHelper.ts:32), so the explicit await comfyPage.nextFrame() after each zoom call in the zoom-culling test results in two idle frames per step. This may dilute per-frame metrics compared to the pan test which uses raw mouse.move without a wrapper.

PR #10007, opened 2026-03-16
fix: migrate V1 tab state pointers during V1→V2 draft migration

christian-byrne:
## Problem

When upgrading from V1 to V2 draft persistence, users' open workflow tabs were lost. V1 stored tab state (open paths + active index) in localStorage via `setStorageValue` fallback, but the V1→V2 migration only migrated draft payloads — not these tab state pointers.

This meant that after upgrading, all previously open tabs disappeared and users had to manually reopen their workflows.

## Solution

Add `migrateV1TabState()` to the V1→V2 migration path. After draft payloads are migrated, the function reads the V1 localStorage keys (`Comfy.OpenWorkflowsPaths` and `Comfy.ActiveWorkflowIndex`) and writes them to V2's sessionStorage format via `writeOpenPaths()`.

The `clientId` is threaded from `useWorkflowPersistenceV2` (which has access to `api.clientId`) through to `migrateV1toV2()`.

## Changes

- **`migrateV1toV2.ts`**: Added `migrateV1TabState()` + V1 key constants for tab state
- **`useWorkflowPersistenceV2.ts`**: Pass `api.clientId` to migration call
- **`migrateV1toV2.test.ts`**: Two new tests proving tab state migration works

## Testing

TDD approach — RED commit shows the test failing, GREEN commit shows it passing.

All 123 persistence tests pass.

- Fixes #9974

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10007-fix-migrate-V1-tab-state-pointers-during-V1-V2-draft-migration-3256d73d36508103b619e521c1b603f5) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
+++ b/src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
@@ -53,8 +53,8 @@ export function useWorkflowPersistenceV2() {
   const toast = useToast()
   const { onUserLogout } = useCurrentUser()
 
-  // Run migration on module load
-  migrateV1toV2()
+  // Run migration on module load, passing clientId for tab state migration
+  migrateV1toV2(undefined, api.clientId ?? api.initialClientId ?? undefined)
 
   // Clear workflow persistence storage when user signs out (cloud only)
   onUserLogout(() => {
--- a/src/platform/workflow/persistence/migration/migrateV1toV2.ts
+++ b/src/platform/workflow/persistence/migration/migrateV1toV2.ts
@@ -154,8 +154,13 @@ function migrateV1TabState(workspaceId: string, clientId?: string): void {
     if (!Array.isArray(paths) || paths.length === 0) return
 
     const indexJson = localStorage.getItem(V1_KEYS.activeIndex)
-    const activeIndex =
-      indexJson !== null ? Math.max(0, JSON.parse(indexJson)) : 0
+    let activeIndex = 0
+    if (indexJson !== null) {
+      const parsed = JSON.parse(indexJson)
+      if (typeof parsed === 'number' && Number.isFinite(parsed)) {
+        activeIndex = Math.min(Math.max(0, parsed), paths.length - 1)
+      }
+    }
 
     writeOpenPaths(clientId, { workspaceId, paths, activeIndex })
   } catch {
--- a/src/platform/workflow/persistence/migration/migrateV1toV2.ts
+++ b/src/platform/workflow/persistence/migration/migrateV1toV2.ts
@@ -10,7 +10,12 @@ import type { DraftIndexV2 } from '../base/draftTypes'
 import { upsertEntry, createEmptyIndex } from '../base/draftCacheV2'
 import { hashPath } from '../base/hashUtil'
 import { getWorkspaceId } from '../base/storageKeys'
-import { readIndex, writeIndex, writePayload } from '../base/storageIO'
+import {
+  readIndex,
+  writeIndex,
+  writeOpenPaths,
+  writePayload
+} from '../base/storageIO'
 
 /**
  * V1 draft snapshot structure (from draftCache.ts)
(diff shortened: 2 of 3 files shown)


src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
```diff
@@ -53,8 +53,8 @@ export function useWorkflowPersistenceV2() {
   const toast = useToast()
   const { onUserLogout } = useCurrentUser()
 
-  // Run migration on module load
-  migrateV1toV2()
+  // Run migration on module load, passing clientId for tab state migration
+  migrateV1toV2(undefined, api.clientId ?? api.initialClientId ?? undefined)
```
dante01yoon, line 57: Issue:
`migrateV1toV2` runs at composable init time ("on module load"). At that point, `api.clientId`  
  is typically `undefined` — it's only set later when the WebSocket sid message arrives (line   
  718 in `api.ts`). `api.initialClientId` reads from `sessionStorage.getItem('clientId')`, which is 
  populated from prior sessions — so this works on reload but not on a fresh first-ever 
  session (no prior clientId in sessionStorage). On a fresh session the migration will run
  with `clientId === undefined` and `migrateV1TabState` will return early, silently skipping tab
  state migration.

  This is arguably a narrow edge case (V1→V2 migration on a truly fresh session where         
  localStorage has V1 data but sessionStorage has never had a clientId), but worth documenting
   or handling — `e.g., by deferring tab state migration until clientId is available, or by    
  generating a temporary clientId for the migration.`

src/platform/workflow/persistence/migration/migrateV1toV2.ts
```diff
@@ -116,12 +126,43 @@ export function migrateV1toV2(workspaceId: string = getWorkspaceId()): number {
...
+  try {
+    const pathsJson = localStorage.getItem(V1_KEYS.openPaths)
+    if (!pathsJson) return
+
+    const paths = JSON.parse(pathsJson)
+    if (!Array.isArray(paths) || paths.length === 0) return
+
+    const indexJson = localStorage.getItem(V1_KEYS.activeIndex)
+    const activeIndex =
+      indexJson !== null ? Math.max(0, JSON.parse(indexJson)) : 0
```
dante01yoon: If V1 stored `activeIndex: 5` but only `3` paths exist, this writes an out-of-bounds index into 
  V2. The consumer (restoreWorkflowTabsState) uses storedWorkflows[storedActiveIndex] which   
  would yield undefined, but it's better to clamp here:                                       
```ts                                                            
const activeIndex = indexJson !== null
  ? Math.min(Math.max(0, JSON.parse(indexJson)), paths.length - 1)                          
  : 0                                                                                       
```                                                                                      
Also agree with CodeRabbit that a typeof `parsed === 'number' && Number.isFinite(parsed)`  check would be more robust against corrupted localStorage.

dante01yoon (commented):
Migration is correctly placed after draft payloads are written

src/platform/workflow/persistence/migration/migrateV1toV2.ts
christian-byrne: fixed in 5513605 - activeIndex is now clamped to paths.length - 1 and parsed value is validated with typeof parsed === 'number' && Number.isFinite(parsed) before use

src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
christian-byrne: good catch - this is a valid edge case. it requires a very specific scenario though: a user who has V1 localStorage data but has never had a sessionStorage clientId (truly fresh first-ever session). in practice this means the tab state migration would be skipped on that one session, but (1) the migration is best-effort and wrapped in try/catch, (2) draft data migration still works correctly since it does not depend on clientId, and (3) only the open-tabs restoration is affected - not workflow data. happy to handle it in a follow-up if the team thinks it is worth addressing, e.g. by deferring tab state migration until clientId is available

dante01yoon (approved):
Issue comment addressed. LGTM

PR #10004, opened 2026-03-16
fix: use immediate node ID for nested widget promotion

christian-byrne:
## Summary

Fix widgets promoted through 2+ levels of subgraph nesting showing "disconnected" instead of rendering correctly.

## Changes

- **`getWidgetName()`**: Always return `w.name` instead of branching on `sourceWidgetName` for PromotedWidgetView
- **`promoteWidget()`**: Always use `String(node.id)` (immediate interior node) instead of `widget.sourceNodeId` (deep leaf)
- **`demoteWidget()`**: Same fix as promoteWidget
- **Cleanup**: Remove unused `isPromotedWidgetView` import and dead `CANVAS_IMAGE_PREVIEW_WIDGET` re-export

## Root Cause

When promoting a `PromotedWidgetView` (re-promoting an already-promoted widget from a nested subgraph), `promoteWidget()` stored `widget.sourceNodeId` (the deep leaf node ID) instead of `node.id` (the immediate SubgraphNode). This recorded an `interiorNodeId` not present in the immediate subgraph, so `pruneDisconnected()` removed it on load.

## Tests

- 7 unit tests in `promotionUtils.test.ts` covering promote/demote/getWidgetName with PromotedWidgetView and pruneDisconnected
- E2E test converting KSampler → subgraph → nested subgraph, then promoting seed through two levels

Fixes #9973

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10004-fix-use-immediate-node-ID-for-nested-widget-promotion-3256d73d365081e0a3e1f39f989ea881) by [Unito](https://www.unito.io)

Issue #9979, opened 2026-03-15
Subgraph Functionalities broken (and more)


byteconcepts:
### Custom Node Testing

- [ ] I have tried disabling custom nodes and the issue persists (see [how to disable custom nodes](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled) if you need help)

### Expected Behavior

To work as expected

### Actual Behavior

At least since last Update broken is:

* Copying and Pasting images from one LoadImage to another in Subgraph does not work any more
* List entries from Power Lora Loader (rgthree) not showing up any more in closed Subgraph, but the empty space fro the is reserved, which means empty space for exactly so many items that should have been shown
* In Subgraph shown text fields stay as "ghost" textfields on workspace when subgraph is minified

Tried with completely deleted Browser-data and different Browsers (Chrome and Firefox) on two different OS.
No error messages are shown.

OMG:

After switching to Nodes 2.0:
* List entries from Power Lora Loader (rgthree) show up again on Subgraphs, but are not interactive (can't turn them on or off or change Strength Value)
* List entries from Power Lora Loader (rgthree) DO NOT show any more in the Power Lora Loader (rgthree) in a open Subgraph
* Layout order of Subgraph elements in closed Subgraph not respected any more after a change in "Shown on node" Configuration tab on the right
* Node Properties gone? Where do I get a nodes Properties from now? - When I click on "Node Info" the sidebar Node Tab opens just with the search and the list of all nodes. So how do I get the Info about the Node if I don't know it's name?
* For example: Reroute (rgthree) does not show it's title any more, just a "frontend_only" Label. - How should that work if you don't know by Title where the Reroute is connected to, what it's for?
* Fast Groups Bypasser not working any more although display seems correct

After switching back from Nodes 2.0:
* Suddenly random in a Subgraph contained text boxes showed up in the folded, but not minimized Subgraph that were not configured to show up

### Steps to Reproduce

Try yourself

### Debug Logs

```powershell
There are no error messages shown
```

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9979-Subgraph-Functionalities-broken-and-more-3246d73d365081f7be07d932419c5849) by [Unito](https://www.unito.io)

gifandy:
yes, recent update make it more worse, after 2-3 runs, the subgraphs are messed up. missing wires, and random node shows up with red borders.

[...]

PR #10008, opened 2026-03-16
fix: preserve promotion order through visibility toggle

### LINKED ISSUE #9995: Subgraph: Layout order of elements not respected after changing "Shown on node" configuration
### Description

After modifying the "Shown on node" configuration for subgraph elements in the right-side configuration tab, the layout order of elements in the collapsed subgraph is no longer respected.

### Steps to Reproduce
1. Create a subgraph with multiple elements configured to show on the node
2. Open the "Shown on node" configuration tab on the right panel
3. Change visibility settings for one or more elements
4. Observe that the layout order is no longer correct in the collapsed subgraph view

### Expected Behavior
Element layout order should be preserved (or follow the configured order) after changing visibility settings.

### Actual Behavior
Layout order becomes incorrect/randomized after modifying "Shown on node" configuration.

---
*Split from #9979*

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9995-Subgraph-Layout-order-of-elements-not-respected-after-changing-Shown-on-node-config-3256d73d36508164b9c3dad2b1ddac09) by [Unito](https://www.unito.io)

christian-byrne:
## Summary

Preserve widget/input promotion ordering when toggling visibility (show/hide) in subgraph nodes.

## Changes

- **What**: Replaced destructive promote/demote operations in `promotionStore.ts` with a manifest-based approach using an internal `ManifestEntry` type that extends `PromotionEntry` with a `promoted` boolean flag.

### Root Cause

`promote()` always appended new entries to the end of the array, and `demote()` removed entries entirely. When a user toggled visibility (demote then re-promote), the widget lost its original position:

```
[A, B, C] → demote B → [A, C] → promote B → [A, C, B]  ← wrong
[A, B, C] → demote B → [A, C] → promote B → [A, B, C]  ← expected
```

### Fix Approach

```mermaid
flowchart TD
    subgraph BEFORE["❌ Before (broken)"]
        direction TB
        B1["promote(A, B, C)"] --> B2["entries: [A, B, C]"]
        B2 --> B3["demote(B) — splice out"]
        B3 --> B4["entries: [A, C]"]
        B4 --> B5["promote(B) — append"]
        B5 --> B6["entries: [A, C, B] ← wrong"]
    end

    subgraph AFTER["✅ After (manifest-based)"]
        direction TB
        A1["promote(A, B, C)"] --> A2["manifest: [A✓, B✓, C✓]"]
        A2 --> A3["demote(B) — flag false"]
        A3 --> A4["manifest: [A✓, B✗, C✓]"]
        A4 --> A5["promote(B) — flag true in-place"]
        A5 --> A6["manifest: [A✓, B✓, C✓] ← correct"]
    end

    A4 -->|"getPromotionsRef()"| FILTERED["cached filtered view: [A, C]"]
    A6 -->|"getPromotionsRef()"| FILTERED2["cached filtered view: [A, B, C]"]

    classDef broken fill:#1a0000,stroke:#ff4444,color:#ff9999
    classDef fixed fill:#001a00,stroke:#44ff44,color:#99ff99
    classDef cache fill:#0a0a2a,stroke:#6688ff,color:#aabbff

    class BEFORE broken
    class B1,B2,B3,B4,B5,B6 broken
    class AFTER fixed
    class A1,A2,A3,A4,A5,A6 fixed
    class FILTERED,FILTERED2 cache
```

- `demote()` marks `promoted = false` (preserving position in the internal manifest) instead of removing the entry
- `promote()` finds existing hidden entries and flips `promoted = true` in-place, only appending truly new entries
- `getPromotionsRef()` returns a reference-stable filtered view of only promoted entries (required by downstream caching in `SubgraphNode._getPromotedViews()`)
- `setPromotions()` remains the authoritative bulk reset (replaces entire manifest)
- `movePromotion()` translates visible indices to manifest indices

### Complementary to #10004

This PR changes **how** promote/demote handle ordering (`promotionStore.ts`). PR #10004 changes **what** IDs get passed to promote/demote (`promotionUtils.ts`). Zero file overlap, no semantic conflict — they fix different layers of the same problem.

## Review Focus

- Reference-stable caching in `getPromotionsRef()` — without this, downstream tests that compare entries by reference identity break
- Manifest index translation in `movePromotion()` — visible index must map to correct manifest position

## Test Coverage

- 8 new tests covering ordering preservation through visibility toggle (demote/re-promote cycles, first/last entry, multi-cycle, mixed with new entries, movePromotion with hidden entries, setPromotions replacing manifest, ref-count consistency)
- 0 regressions across 214 downstream tests (subgraph, promotedWidgetView, graphNodeManager, etc.)

Fixes #9995

Related: #9979 (parent umbrella), #9841, #5588, #9985


AustinMroz (commented):
The prior functionality was intended. I'm not opposed to the change, but I'd want to double check with design.

Code itself looks fine. I was concerned about zombie entries (promote a widget -> hide the widget -> delete the node), but it looks like the existing code paths handle this already. Draft PR has my tentative approval.

artokun:
Hey @christian-byrne — nice work identifying this issue! I ran into the same ordering problem while working on the subgraph IO reorder feature (#10227 / #10232) and ended up putting together a fix in #10584.

I took a slightly different approach — rather than refactoring the store's internal data model, I added a lightweight `lastDemotedIndices` map that records where an entry was before demotion and restores it on re-promotion. This keeps the existing public API and `PromotedWidgetSource` shape unchanged, which avoids needing to update all the callers (SubgraphNode, WidgetActions, SectionWidgets, etc.).

While investigating, I also found two other bugs in the same code path that contributed to the symptoms in #9995:

1. `getSourceNodeId()` in `promotionUtils.ts` was falling back to `sourceNodeId` when `disambiguatingSourceNodeId` was `undefined`, which caused `isPromoted` lookups to fail for non-nested subgraphs. This made the sidebar show "Show input" on widgets that were already promoted and allowed duplicate promotions.

2. The `node:slot-label:changed` handler in `useGraphNodeManager.ts` was calling `extractVueNodeData()` a second time, which isn't idempotent and could snapshot a different widget order into the reactive arrays.

Both of those are addressed in #10584 as well.

Would you be open to letting #10584 take this one? I think the manifest approach you explored here is a really clean concept — it just would need the caller-side updates and `disambiguatingSourceNodeId` support to be mergeable. Happy to collaborate if you see anything in my PR that could be improved!

DrJKL:
Thank you for diagnosing the promotion-order issue and developing the manifest-based fix. #10584 now covers the same hide and show ordering bug with additional source-ID and reactivity fixes plus browser coverage, and the original promotion-store paths here have since been refactored out of current main. We are closing this draft in favor of that replacement; if anything remains after #10584, please send a fresh focused PR against the current promotion architecture.

PR #10009, opened 2026-03-16
fix: resolve nodes in subgraphs for image copy/paste and display

### LINKED ISSUE #9993: Subgraph: Copy/paste images between LoadImage nodes inside subgraph broken
### Description

Copying and pasting images from one LoadImage node to another LoadImage node **inside a subgraph** no longer works.

### Steps to Reproduce
1. Create a subgraph with two LoadImage nodes
2. Load an image in one LoadImage node
3. Try to copy and paste the image to the other LoadImage node
4. Observe it does not work

### Expected Behavior
Image should be copied/pasted between LoadImage nodes inside subgraphs as before.

### Actual Behavior
Copy/paste of images between LoadImage nodes in subgraphs silently fails.

### Environment
Tested on Chrome and Firefox across two different operating systems. Reproduced with cleared browser data.

---
*Split from #9979*

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9993-Subgraph-Copy-paste-images-between-LoadImage-nodes-inside-subgraph-broken-3256d73d365081188392fb4351e1c996) by [Unito](https://www.unito.io)

christian-byrne:
## Summary

Replace `app.rootGraph.getNodeById()` with `resolveNode()` so node lookups search subgraphs, fixing broken image copy/paste and display for nodes inside subgraphs.

## Changes

- **What**: Updated 7 call sites across 5 files to use `resolveNode()` from `litegraphUtil.ts` instead of `app.rootGraph.getNodeById()`. The `resolveNode` function already existed and searches both the root graph and all subgraphs. Added a unit test verifying subgraph node resolution in `nodeOutputStore`.

## Review Focus

The fix is mechanical — each call site simply swaps `app.rootGraph?.getNodeById(id)` for `resolveNode(id)`. The `resolveNode` utility iterates `graph.subgraphs` if the node is not found in the root graph.

Fixes #9993

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10009-fix-resolve-nodes-in-subgraphs-for-image-copy-paste-and-display-3256d73d3650814f9467c53999f5d755) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/load3d/Load3D.vue
+++ b/src/components/load3d/Load3D.vue
@@ -80,7 +80,7 @@ import { useLoad3d } from '@/composables/useLoad3d'
 import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
 import { useSettingStore } from '@/platform/settings/settingStore'
 import type { NodeId } from '@/platform/workflow/validation/schemas/workflowSchema'
-import { app } from '@/scripts/app'
+import { resolveNode } from '@/utils/litegraphUtil'
 import type { ComponentWidget } from '@/scripts/domWidget'
 import type { SimplifiedWidget } from '@/types/simplifiedWidget'
 
@@ -101,7 +101,7 @@ if (isComponentWidget(props.widget)) {
   node.value = props.widget.node
 } else if (props.nodeId) {
   onMounted(() => {
-    node.value = app.rootGraph?.getNodeById(props.nodeId!) || null
+    node.value = resolveNode(props.nodeId!) ?? null
   })
 }
 
--- a/src/composables/useImageCrop.ts
+++ b/src/composables/useImageCrop.ts
@@ -4,8 +4,8 @@ import { computed, onMounted, ref, watch } from 'vue'
 
 import type { LGraphNode, NodeId } from '@/lib/litegraph/src/LGraphNode'
 import type { Bounds } from '@/renderer/core/layout/types'
-import { app } from '@/scripts/app'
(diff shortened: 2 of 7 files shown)


AustinMroz (dismissed):

dante01yoon (commented):
Would you mind perform red-green test workflow?

dante01yoon (dismissed):
nit:   Looking at the implementation (litegraphUtil.ts:310-322), it only iterates    graph.subgraphs.values() — not recursively. If nested subgraphs (subgraph within a subgraph) are supported, nodes at depth > 1 would still be missed. This is a pre-existing limitation  of resolveNode, not introduced by this PR, but worth flagging since the PR expands itsusage.

christian-byrne:
re: red-green TDD request from @dante01yoon - added in 1fbdca48b. 7 unit tests covering null/undefined graph, root graph lookup, subgraph lookup, root graph priority, and multi-subgraph search. verified the RED phase: removing the subgraph iteration causes `finds a node inside a subgraph` and `searches across multiple subgraphs` to fail as expected.

re: nit on nested subgraph recursion - good catch. you're right that `resolveNode` only iterates `graph.subgraphs.values()` at depth 1 and won't find nodes at depth > 1. agreed this is a pre-existing limitation of `resolveNode` itself, not introduced by this PR (which only expands its usage to copy/paste and image display). worth tracking as a follow-up if nested subgraphs become more common.

@coderabbitai can you make an Issue to track the above?

dante01yoon (approved):

PR #10013, opened 2026-03-16
test: add regression tests for subgraph slot label propagation

christian-byrne:
## Summary

Add regression tests for subgraph slot label propagation. The OutputSlot.vue fix (adding `slotData.label` to the display template) was already merged via another PR — this adds tests to prevent future regressions.

## Changes

- **What**: Two new test files covering the label/localized_name fallback chain in OutputSlot.vue and SubgraphNode label propagation through configure() and rename event paths.

## Review Focus

Tests only — no production code changes. Verifies that renamed subgraph inputs/outputs display correctly in Nodes 2.0 mode.

Fixes #9998

### DIFF
--- a/src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
+++ b/src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
@@ -71,11 +71,13 @@ describe('OutputSlot', () => {
   it('renders label when present on slotData', () => {
     const wrapper = mountOutputSlot({
       name: 'internal_name',
+      localized_name: 'Localized Name',
       label: 'My Custom Label'
     })
 
     expect(wrapper.text()).toContain('My Custom Label')
     expect(wrapper.text()).not.toContain('internal_name')
+    expect(wrapper.text()).not.toContain('Localized Name')
   })
 
   it('falls back to localized_name when label is absent', () => {
--- a/src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
+++ b/src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
@@ -87,6 +87,7 @@ describe('OutputSlot', () => {
     })
 
     expect(wrapper.text()).toContain('Localized Name')
+    expect(wrapper.text()).not.toContain('internal_name')
   })
 
   it('falls back to name when label and localized_name are absent', () => {
--- a/src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
+++ b/src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
@@ -1,0 +1,101 @@
+import { mount } from '@vue/test-utils'
+import { describe, expect, it, vi } from 'vitest'
+import { defineComponent } from 'vue'
+import { createI18n } from 'vue-i18n'
+
+import type { INodeSlot } from '@/lib/litegraph/src/litegraph'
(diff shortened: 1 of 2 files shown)


src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      directives: { tooltip: {} },
+      stubs: { SlotConnectionDot: SlotConnectionDotStub }
+    }
+  })
+}
+
+describe('OutputSlot', () => {
+  it('renders label when present on slotData', () => {
+    const wrapper = mountOutputSlot({
+      name: 'internal_name',
```
Myestery, line 73: nit: This test is good, but it could be more robust by also providing `localized_name` in the `slotData`. This would ensure that `label` is correctly prioritized over both `localized_name` and `name`.

```typescript
  it('renders label when present on slotData', () => {
    const wrapper = mountOutputSlot({
      name: 'internal_name',
      localized_name: 'Localized Name',
      label: 'My Custom Label'
    })

    expect(wrapper.text()).toContain('My Custom Label')
    expect(wrapper.text()).not.toContain('internal_name')
    expect(wrapper.text()).not.toContain('Localized Name')
  })
```

src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    expect(wrapper.text()).not.toContain('internal_name')
+  })
+
+  it('falls back to localized_name when label is absent', () => {
+    const wrapper = mountOutputSlot({
+      name: 'internal_name',
+      localized_name: 'Localized Name'
+    })
+
+    expect(wrapper.text()).toContain('Localized Name')
```
Myestery, line 89: nit: To make this test more thorough, we can also assert that the `name` property is not being used when `localized_name` is present. This will confirm the fallback priority is correctly implemented.

```typescript
  it('falls back to localized_name when label is absent', () => {
    const wrapper = mountOutputSlot({
      name: 'internal_name',
      localized_name: 'Localized Name'
    })

    expect(wrapper.text()).toContain('Localized Name')
    expect(wrapper.text()).not.toContain('internal_name')
  })
```

Myestery (dismissed):
Approved with nits

src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
christian-byrne: fixed in e3e99d9 - added localized_name to the test data and asserted it's not rendered

src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
christian-byrne: fixed in 42533ad - added negative assertion for name

Myestery (approved):

src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
christian-byrne: @coderabbitai make an Issue to do this in a followup

src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
christian-byrne: @coderabbitai make an Issue to do this in a followup

PR #10017, opened 2026-03-16
test: add perf test for PrimeVue dialog style recalculations

christian-byrne:
Adds a `@perf` test to establish a baseline for PrimeVue `useStyle` style injection overhead (backlog #7).

This is PR 1 of 2. The fix will follow in a separate PR once this baseline is established on main.

## What

Adds `primevue-dialog-open-close` to the performance test suite measuring style recalculations during settings dialog open/close.

## Why

PrimeVue's `useStyle` composable injects `<style>` tags at runtime for each component mount. Firefox profiling showed 4,555 style invalidation markers from this mechanism. Opening a dialog that mounts many PrimeVue components (Dialog, InputText, Select, ToggleSwitch, etc.) triggers a burst of style injections.

Needed to prove the improvement from the upcoming fix for backlog item #7.

## How

The test:
1. Loads default workflow and waits for stabilization
2. Opens the settings dialog (triggers PrimeVue component mounts + style injections)
3. Waits for settlement
4. Closes the dialog
5. Records style recalc count, layout count, and recalc duration

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10017-test-add-perf-test-for-PrimeVue-dialog-style-recalculations-3256d73d365081dcba00e168ae486cc4) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/performance.spec.ts
+++ b/browser_tests/tests/performance.spec.ts
@@ -252,9 +252,7 @@ test.describe('Performance', { tag: ['@perf'] }, () => {
       await comfyPage.nextFrame()
     }
 
-    const m = await comfyPage.perf.stopMeasuring(
-      'primevue-dialog-open-close'
-    )
+    const m = await comfyPage.perf.stopMeasuring('primevue-dialog-open-close')
     recordMeasurement(m)
     console.log(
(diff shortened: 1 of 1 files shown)


christian-byrne:
Closing — investigation concluded that backlog #7 (PrimeVue useStyle) is a startup-only cost (per-component-type first mount, not per-frame). The 4,555 markers are amortized and dwarfed by per-frame bottlenecks. Will resolve naturally as PrimeVue components are replaced with shadcn/vue.

PR #10021, opened 2026-03-16
feat: add Paste Image option to Load Image node context menu

### LINKED ISSUE #9989: Feature Request: Add 'Paste Image' option to Load Image node context menu
### Description

Add a **"Paste Image"** option to the context menu of the **Load Image** node, complementing the existing **"Copy Image"** option that is already present.

### Motivation

Currently, the Load Image node has a "Copy Image" context menu entry but no corresponding "Paste Image" entry. Users expect symmetry between copy and paste operations in the UI. This would provide a reliable alternative to `Ctrl+V` for pasting images (from clipboard) directly into the Load Image node, including images copied from external sources.

### Proposed Behavior

- Right-clicking the Load Image node should show a **"Paste Image"** menu item
- Clicking it should read the current clipboard image and upload it to the node (same as a successful `Ctrl+V` paste)
- Should work with both internally copied images (from other nodes) and externally copied images (from OS clipboard)

### References

- Originally requested by @V1sionVerse in: https://github.com/Comfy-Org/ComfyUI/issues/12918
- Requester: @dante01yoon

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9989-Feature-Request-Add-Paste-Image-option-to-Load-Image-node-context-menu-3256d73d36508160831dc54b6977948f) by [Unito](https://www.unito.io)

dante01yoon:
## Summary

- Adds a **Paste Image** context menu option to nodes that support image pasting (e.g. Load Image), complementing the existing **Copy Image** option
- Reads clipboard image data via `navigator.clipboard.read()` and delegates to `node.pasteFiles()` — same path as `Ctrl+V` paste
- Only shown on nodes that implement `pasteFiles` (e.g. LoadImage)
- Available even when no image is loaded yet (e.g. fresh LoadImage node)
- **Node 2.0 context menu only** — legacy litegraph menu is not supported

- Fixes #9989

스크린샷 2026-03-16 오후 5 34 28

## Test plan

- [x] Right-click a Load Image node (with or without image loaded) → verify "Paste Image" appears
- [x] Copy an image to clipboard → click "Paste Image" → verify image is loaded into the node
- [x] Verify "Paste Image" does not appear on output-only image nodes (e.g. Save Image, Preview Image)
- [x] Unit tests pass: `pnpm vitest run src/composables/graph/useImageMenuOptions.test.ts`


dante01yoon:
@coderabbitai why do you set changes requested?

browser_tests/tests/pasteImageContextMenu.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+    await comfyPage.workflow.loadWorkflow('nodes/single_save_image_node')
+
+    const saveImageNode = (
+      await comfyPage.nodeOps.getNodeRefsByType('SaveImage')
+    )[0]
+
+    const menuOptions = await saveImageNode.getContextMenuOptionNames()
+
+    expect(menuOptions).not.toContain('Paste Image')
+  })
```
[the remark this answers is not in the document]
dante01yoon: Addressed — added `expect(menuOptions).not.toContain('Open Image')` to explicitly verify the image menu section is absent, not just Paste Image. Also removed the legacy litegraph menu support (Node 2.0 only now) and fixed the early-return so Paste Image shows on nodes without images loaded.

src/composables/graph/useImageMenuOptions.ts
```diff
@@ -3,9 +3,14 @@ import { useI18n } from 'vue-i18n'
 import { downloadFile, openFileInNewTab } from '@/base/common/downloadUtil'
 import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
 import { useCommandStore } from '@/stores/commandStore'
+import { pasteClipboardImageToNode } from '@/utils/clipboardUtil'
 
 import type { MenuOption } from './useMoreOptionsMenu'
 
+function canPasteImage(node: LGraphNode): boolean {
+  return typeof node.pasteFiles === 'function'
+}
```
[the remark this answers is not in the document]
dante01yoon: Fixed — `canPasteImage` now takes optional `node?` and uses `node?.pasteFiles`. Also cached the result in a `canPaste` local to avoid duplicate calls.

src/utils/clipboardUtil.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  try {
+    const clipboardItems = await navigator.clipboard.read()
+    for (const item of clipboardItems) {
+      const imageType = item.types.find((type) => type.startsWith('image/'))
+      if (!imageType) continue
+
+      const blob = await item.getType(imageType)
+      const file = new File([blob], 'pasted-image.png', { type: imageType })
```
christian-byrne: nitpick (non-blocking): The filename is hardcoded to `'pasted-image.png'` but `imageType` may be `image/jpeg`, `image/webp`, etc. Consider deriving the extension:

```ts
const ext = imageType.split('/')[1] ?? 'png'
const file = new File([blob], `pasted-image.${ext}`, { type: imageType })
```

src/utils/clipboardUtil.ts
```diff
@@ -0,0 +1,25 @@
+import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
+
+export async function pasteClipboardImageToNode(
+  node: LGraphNode
+): Promise<void> {
+  if (!navigator.clipboard?.read) {
+    console.warn('Clipboard API not available')
```
christian-byrne: suggestion (non-blocking): When the Clipboard API isn't available or `read()` throws, the user gets no visual feedback -- only console output. A toast/notification could help users understand why nothing happened. That said, the existing `copyImage` in this file follows the same console-only pattern, so this is consistent as-is.

src/utils/clipboardUtil.ts
```diff
@@ -0,0 +1,25 @@
+import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
+
+export async function pasteClipboardImageToNode(
```
christian-byrne: question (non-blocking): `usePaste.ts` already has `pasteItemsOnNode` which converts DataTransfer items to files and calls `node.pasteFiles()`. This creates a second clipboard-to-node pathway with subtly different behavior (hardcoded filename, only first image, different File construction). I understand the context menu case needs `navigator.clipboard.read()` since there's no paste event -- but could the file-construction logic be shared between the two paths to keep behavior consistent?

src/utils/clipboardUtil.ts
```diff
@@ -0,0 +1,25 @@
+import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
```
christian-byrne: suggestion (non-blocking): This repo is migrating toward a layered architecture (`base/` -> `platform/` -> `workbench/` -> `renderer/`). New files in flat technical folders like `src/utils/` are discouraged -- this utility could live under something like `src/platform/clipboard/` or alongside its sole consumer in `src/composables/graph/`. Not blocking since lots of existing utils live here, but worth considering for alignment with the migration direction.

src/utils/clipboardUtil.ts
christian-byrne: note (non-blocking): the function was moved to be co-located with its consumer, which is a step in the right direction. unifying the file-construction logic between the context menu path and the paste-event path is a good followup.

src/composables/graph/useImageMenuOptions.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  useCommandStore: () => ({ execute: vi.fn() })
+}))
+
+function mockClipboard(clipboard: Partial<Clipboard> | undefined) {
+  Object.defineProperty(navigator, 'clipboard', {
+    value: clipboard,
+    writable: true,
+    configurable: true
+  })
+}
```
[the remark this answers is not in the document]
christian-byrne: valid point - the clipboard override should be restored in afterEach. the existing copyImage tests elsewhere follow the same console-only pattern though, so this is consistent. worth cleaning up as a followup.

christian-byrne (dismissed):
all feedback addressed - filename extension derived from mime type, utility moved alongside consumer, pasteFile call added. two non-blocking items tagged for followup issues. lgtm

PR #6289, opened 2025-10-26
Fix node sizing in vue mode

AustinMroz:
Before
broken-resize
After
vue-resize

Also add ~~content~~ [contain](https://developer.mozilla.org/en-US/docs/Web/CSS/contain) styling for improved render performance.

Future:
- Update size scaling for WidgetLayoutField widgets.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6289-Fix-node-sizing-in-vue-mode-2986d73d365081ac8fa0da35a635b226) by [Unito](https://www.unito.io)

PR #5169, opened 2025-08-22
feat: v3 style of node body

LittleSound:
## Summary

v3 style of node body. finished the basic style of the node.

## Review Focus

There's a problem with the connection lines position and window size, and I'm not quite sure how to fix it yet, so I'm mainly focusing on how to align design sheets in style.

## Screenshots (if applicable)

CleanShot 2025-08-22 at 19 14 16
CleanShot 2025-08-22 at 19 14 29

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-5169-feat-v3-style-of-node-body-2576d73d3650813cbb5cd3f3a60e4b39) by [Unito](https://www.unito.io)

PR #10022, opened 2026-03-16
fix: resolve node text bleed-through by isolating stacking contexts

### LINKED ISSUE #9988: Pasting Image onto Canvas Failure
### Prerequisites

- [ ] I have custom nodes enabled

### What happened?

When an image is pasted onto the ComfyUI Canvas (i.e., without any other node selected), it is supposed to spawn a Load Image node with the pasted image. As it stands, as of the latest version, this functionality is broken. 

Recently, @dante01yoon created PR #9773 to solve the pasting issue _when a media node is selected before pasting_. However, the issue is still present when a user attempts to paste an image onto the canvas directly. 

### Steps to Reproduce

1. Copy Image
2. Paste on Canvas

### ComfyUI Frontend Version

1.42.5

### Browser

Edge

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-9988-Pasting-Image-onto-Canvas-Failure-3256d73d365081ffa88edf7055277797) by [Unito](https://www.unito.io)

dante01yoon:
## Summary

- Move `contain-layout contain-style` from the outer node container to the inner wrapper to prevent CSS containment from interfering with inter-node compositing
- Add `isolation: isolate` to the outer container for explicit stacking context isolation

- Fixes #9988

## Root Cause

Originally, PR [#5169](https://github.com/Comfy-Org/ComfyUI_frontend/pull/5169) (`57db10f`, 2025-08-23) applied all three containment properties to the outer node container:

```
contain-layout contain-style contain-paint
```

`contain-paint` ensured each node's painting was clipped to its own boundaries, guaranteeing render isolation between sibling nodes.

Later, PR [#6289](https://github.com/Comfy-Org/ComfyUI_frontend/pull/6289) (`6e4471ad`, 2025-10-28) removed `contain-paint` to fix a node sizing issue — likely because `contain-paint` clips content to the element's padding box, which would clip selection overlays and footer tabs that extend beyond the node boundary via negative insets.

Removing `contain-paint` while keeping `contain-layout contain-style` on the outer container broke paint isolation between nodes. Without paint containment, the browser's compositor no longer treated each node as an isolated painting unit, allowing text from lower z-index nodes to visually bleed through higher z-index nodes.

## Fix

1. **Move containment inward**: `contain-layout contain-style` is moved from the outer container to the inner wrapper (`node-inner-wrapper`). This preserves layout/style containment performance benefits while keeping them scoped to the node body — away from the stacking context boundary between sibling nodes.

2. **Explicit isolation**: `isolation: isolate` is added to the outer container. While `transform` + `z-index` already create a stacking context, `isolation: isolate` provides an unambiguous signal to the browser's compositor to treat each node as an isolated compositing unit — replacing the paint isolation role that `contain-paint` previously served, without clipping overlays.

## Red-Green Verification

| Commit | Status | Purpose |
|--------|--------|---------|
| [`feb8555`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/feb855501) `test:` | :red_circle: Red (verified locally) | Test asserts `isolate` on outer container and `contain-layout contain-style` on inner wrapper only — fails against current main |
| [`c47f89e`](https://github.com/Comfy-Org/ComfyUI_frontend/commit/c47f89ed2) `fix:` | :green_circle: Green (CI: Tests Unit passing) | Moves containment to inner wrapper, adds `isolation: isolate` — test passes |

**Local red verification:**
```
AssertionError: expected [ 'group/node', 'lg-node', …(9) ] to include 'isolate'
 ❯ src/renderer/extensions/vueNodes/components/LGraphNode.test.ts:264:26
```

## Test Plan

- [x] CI red on test-only commit (verified locally — test fails without fix)
- [x] CI green on fix commit (CI: Tests Unit passes)
- [ ] Place a String Literal node with long text on the canvas
- [ ] Place a Load Image node (with an image loaded) overlapping the String Literal
- [ ] Verify text from the String Literal does NOT bleed through the Load Image node
- [ ] Verify node selection outlines, error borders, and footer tabs render correctly


dante01yoon:
@coderabbitai check if changes requested still needed

src/renderer/extensions/vueNodes/components/LGraphNode.test.ts
```diff
@@ -251,4 +251,26 @@ describe('LGraphNode', () => {
...
+    // CSS containment must NOT be on the outer container because it
+    // interferes with inter-node stacking compositing.
+    expect(outerClasses).not.toContain('contain-layout')
+    expect(outerClasses).not.toContain('contain-style')
+
+    // CSS containment belongs on the inner wrapper for performance
+    // without affecting stacking between sibling nodes.
+    expect(innerClasses).toContain('contain-layout')
+    expect(innerClasses).toContain('contain-style')
+  })
```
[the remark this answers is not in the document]
DrJKL: Agreed, this is a CSS change detector test.

browser_tests/tests/saveImageAndWebp.spec.ts-snapshots/save-image-and-webm-preview-chromium-linux.png
DrJKL, line 1: If we can't have outlines anymore, should we also remove all of the outline styling?
(Or is this not an intended change?)

dante01yoon: It's not intentional where did you find outline disappeared? @DrJKL

src/renderer/extensions/vueNodes/components/LGraphNode.test.ts
```diff
@@ -251,4 +251,20 @@ describe('LGraphNode', () => {
...
+    // Outer container must have isolation: isolate to prevent text
+    // from lower z-index nodes bleeding through (see #9988).
+    expect(outerClasses).toContain('isolate')
+
+    // CSS containment stays on outer container alongside isolate.
+    // Moving it to the inner wrapper causes the inner wrapper's
+    // stacking context to paint over the Root Border Overlay.
+    expect(outerClasses).toContain('contain-layout')
+    expect(outerClasses).toContain('contain-style')
+  })
```
DrJKL: I'd probably rather no test than a change detector. I think the earlier failures in the Playwright screenshot assertions show that there's some coverage for behavior here.

DrJKL (approved):

PR #10024, opened 2026-03-16
fix: allow URL input for free tier users, gate on import button

jaeone94:
## Summary
- Remove free-tier restriction from the URL input field in `MissingModelUrlInput.vue` so it is always editable
- Move the subscription check (`canImportModels`) to the Import button click handler — free-tier users see the upgrade modal only when they attempt to import
- Extract inline ternary to named `handleImportClick` method for clarity

## Test plan
- [x] Unit tests added (`MissingModelUrlInput.test.ts`) verifying:
  - URL input is always editable regardless of subscription tier
  - Import button calls `handleImport` for paid users
  - Import button calls `showUploadDialog` (upgrade modal) for free-tier users
- [x] Verify URL input is editable for free-tier users on cloud
- [x] Verify clicking Import as free-tier opens the subscription modal
- [x] Verify paid users can import normally without changes

## E2E test rationale
Playwright E2E regression tests are impractical for this change because `MissingModelUrlInput` only renders when `isAssetSupported` is true, which requires `isCloud` — a compile-time constant (`__DISTRIBUTION__`). The OSS test build always sets `isCloud = false`, so the component never renders in the E2E environment. Unit tests with mocked feature flags provide equivalent behavioral coverage.

### DIFF
--- a/src/platform/missingModel/components/MissingModelUrlInput.test.ts
+++ b/src/platform/missingModel/components/MissingModelUrlInput.test.ts
@@ -120,10 +120,11 @@ describe('MissingModelUrlInput', () => {
       mockPrivateModelsEnabled.value = false
       const wrapper = mountComponent()
       const input = wrapper.find('input')
+      input.element.value = 'https://example.com/model.safetensors'
       await input.trigger('input')
       expect(mockHandleUrlInput).toHaveBeenCalledWith(
         MODEL_KEY,
-        expect.any(String)
+        'https://example.com/model.safetensors'
       )
     })
   })
@@ -167,6 +168,16 @@ describe('MissingModelUrlInput', () => {
 
       expect(mockShowUploadDialog).toHaveBeenCalled()
       expect(mockHandleImport).not.toHaveBeenCalled()
+    })
+
+    it('clear button works for free-tier users', async () => {
+      mockPrivateModelsEnabled.value = false
+      const store = useMissingModelStore()
+      store.urlInputs[MODEL_KEY] = 'https://example.com/model.safetensors'
+      const wrapper = mountComponent()
+      const clearBtn = wrapper.find('button[aria-label="Clear URL"]')
+      await clearBtn.trigger('click')
+      expect(mockHandleUrlInput).toHaveBeenCalledWith(MODEL_KEY, '')
     })
   })
 })
--- a/src/platform/missingModel/components/MissingModelUrlInput.test.ts
+++ b/src/platform/missingModel/components/MissingModelUrlInput.test.ts
@@ -1,0 +1,172 @@
+import { mount } from '@vue/test-utils'
+import { createPinia, setActivePinia } from 'pinia'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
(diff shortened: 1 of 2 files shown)


src/platform/missingModel/components/MissingModelUrlInput.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    })
+
+    it('input is editable when privateModelsEnabled is false (free tier)', () => {
+      mockPrivateModelsEnabled.value = false
+      const wrapper = mountComponent()
+      const input = wrapper.find('input')
+      expect(input.attributes('readonly')).toBeUndefined()
+    })
+
+    it('input accepts user typing when privateModelsEnabled is false', async () => {
```
dante01yoon, line 119: Triggering `input` without setting a value means `$event.target.value` is `''` — the test passes but doesn't prove typed text flows through. Consider:

```typescript
const input = wrapper.find('input')
input.element.value = 'https://example.com/model.safetensors'
await input.trigger('input')
expect(mockHandleUrlInput).toHaveBeenCalledWith(
  MODEL_KEY,
  'https://example.com/model.safetensors'
)
```

src/platform/missingModel/components/MissingModelUrlInput.test.ts
```diff
@@ -0,0 +1,172 @@
...
+      await importBtn!.trigger('click')
+
+      expect(mockHandleImport).toHaveBeenCalledWith(MODEL_KEY, 'checkpoints')
+      expect(mockShowUploadDialog).not.toHaveBeenCalled()
+    })
+
+    it('calls showUploadDialog when privateModelsEnabled is false (free tier)', async () => {
+      mockPrivateModelsEnabled.value = false
+      const store = useMissingModelStore()
+      store.urlMetadata[MODEL_KEY] = {
```
dante01yoon, line 156: nit: Missing test -- clear button for free-tier users

The clear button (<Button @click.stop="handleUrlInput(modelKey, '')">) is visible when there's input, but there's no test verifying it works for free-tier users. Since the old code had pointer-events-none on the input area, it's worth confirming the clear button is reachable. A quick test would add confidence:

```typescript
it('clear button works for free-tier users', async () => {
  mockPrivateModelsEnabled.value = false
  const store = useMissingModelStore()
  store.urlInputs[MODEL_KEY] = 'https://example.com/model.safetensors'
  const wrapper = mountComponent()
  const clearBtn = wrapper.find('button[aria-label="Clear URL"]')
  await clearBtn.trigger('click')
  expect(mockHandleUrlInput).toHaveBeenCalledWith(MODEL_KEY, '')
})
```

dante01yoon (commented):
Clean, focused change. Moves the subscription gate from the URL input to the Import button — better UX since free-tier users can preview metadata before hitting the paywall. Template is simpler without the faux-button `upgradePromptAttrs` computed. All CI green.

Two minor test gaps noted inline.

dante01yoon (commented):
nit

dante01yoon (approved):

PR #10026, opened 2026-03-16
tool: add CSS containment audit skill and Playwright diagnostic

dante01yoon:
## Summary

- Add a Playwright-based diagnostic tool (`@audit` tagged) that automatically detects DOM elements where CSS `contain: layout style` would improve rendering performance
- Extend `ComfyPage` fixture and `playwright.config.ts` to support `@audit` tag (excluded from CI, perf infra enabled)
- Add `/contain-audit` skill definition documenting the workflow

## How it works

1. Loads the 245-node workflow in a real browser
2. Walks the DOM tree and scores every element by subtree size and sizing constraints
3. For each high-scoring candidate, applies `contain: layout style` via JS
4. Measures rendering performance (style recalcs, layouts, task duration) before and after
5. Takes before/after screenshots to detect visual breakage
6. Outputs a ranked report to console

## Test plan

- [ ] `pnpm typecheck` passes
- [ ] `pnpm typecheck:browser` passes
- [ ] `pnpm lint` passes
- [ ] Existing Playwright tests unaffected (`@audit` excluded from CI via `grepInvert`)
- [ ] Run `pnpm exec playwright test browser_tests/tests/containAudit.spec.ts --project=chromium` locally with dev server

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10026-tool-add-CSS-containment-audit-skill-and-Playwright-diagnostic-3256d73d365081b29470df164f798f7d) by [Unito](https://www.unito.io)


browser_tests/tests/containAudit.spec.ts
```diff
@@ -0,0 +1,391 @@
...
+      console.log('Already contained elements:')
+      alreadyContained.forEach((c) => {
+        console.log(`  ${c.selector} (subtree: ${c.subtreeSize})`)
+      })
+    }
+
+    expect(results.length).toBeGreaterThan(0)
+  })
+
+  test('measure containment impact with pan interaction', async ({
```
christian-byrne: issue: This pan measurement test (load `large-graph-workflow`, middle-mouse pan with `j * 5, j * 2` offsets across 60 frames, CDP style-recalc/layout metrics) is nearly identical to christian-byrne's open PR #10001 which adds the same test to `performance.spec.ts`. Additionally, PR #9946 (also open, by christian-byrne) already applies `contain-layout contain-style` to the node inner wrapper -- the exact fix this audit is designed to discover. The SKILL.md even references #9946 as its example PR.

Consider coordinating with @christian-byrne's existing perf infrastructure rather than introducing a parallel measurement path. The audit's value proposition is diminished if both the fix (#9946) and the test (#10001) already exist as open PRs.

browser_tests/tests/containAudit.spec.ts
```diff
@@ -0,0 +1,391 @@
...
+        if (isFlexChild) score *= 1.5
+        if (alreadyContained) score = 0
+
+        let selector = el.tagName.toLowerCase()
+        const testId = el.getAttribute('data-testid')
+        if (testId) {
+          selector = `[data-testid="${testId}"]`
+        } else if (el.id) {
+          selector = `#${el.id}`
+        } else if (el.className && typeof el.className === 'string') {
```
christian-byrne: issue: The first-class-name fallback (`el.className.split(' ')[0]`) is fragile with Tailwind utility classes. Many elements share the same first class (e.g., `.flex`, `.relative`), so `querySelectorAll(selector)` on L153 will apply containment to all of them -- not just the intended candidate. This can produce misleading perf numbers and false visual breakage reports.

The second test (L288-335) correctly uses known selectors like `[data-testid="node-inner-wrapper"]` directly. Consider applying the same approach here, or adding a parent/nth-child qualifier to disambiguate.

browser_tests/tests/containAudit.spec.ts
```diff
@@ -0,0 +1,391 @@
...
+      for (let i = 0; i < 60; i++) {
+        await comfyPage.nextFrame()
+      }
+      const withContain = await comfyPage.perf.stopMeasuring(
+        `contain-${candidate.selector}`
+      )
+
+      // Take screenshot with containment applied to detect visual breakage
+      const containScreenshot = await comfyPage.page.screenshot()
+      const visuallyBroken = !baselineScreenshot.equals(containScreenshot)
```
christian-byrne, line 197: suggestion (non-blocking): `baselineScreenshot.equals(containScreenshot)` does byte-level comparison. Subpixel rendering, anti-aliasing, and GPU timing differences will cause false positives (marking candidates as "BREAK" when they're visually identical). Consider using Playwright's `toMatchSnapshot({ maxDiffPixelRatio })` or a pixel-diff library with a tolerance threshold.

playwright.config.ts
```diff
@@ -36,7 +36,7 @@ export default defineConfig({
       name: 'chromium',
       use: { ...devices['Desktop Chrome'] },
       timeout: 15000,
-      grepInvert: /@mobile|@perf/ // Run all tests except those tagged with @mobile or @perf
+      grepInvert: /@mobile|@perf|@audit/ // Run all tests except those tagged with @mobile, @perf, or @audit
```
christian-byrne, line 39: suggestion (non-blocking): `@audit` is excluded from the `chromium` project via `grepInvert` (good), but unlike `@perf` there is no dedicated `audit` project with its own timeout. Running manually with `--project=chromium` inherits the 15s timeout, which will likely be insufficient for the audit workload (10 candidates x 60-frame warmup x 60-frame measurement). Consider adding a dedicated project entry, similar to the existing `performance` project.

.claude/skills/contain-audit/SKILL.md
```diff
@@ -0,0 +1,99 @@
+---
```
christian-byrne, line 1: question (non-blocking): `.claude/skills/` doesn't exist on `main` yet. Is this an accepted pattern for this repo, or should agent skill definitions live elsewhere (e.g., in a separate tooling repo)?

christian-byrne (changes requested):

.claude/skills/contain-audit/SKILL.md
dante01yoon: Good question! I think keeping `.claude/skills/` in this repo is the right call — these skill definitions are tightly coupled to the codebase (they reference specific test files, fixtures, and project conventions), so co-locating them makes it easier to keep them in sync as the code evolves. Moving them to a separate tooling repo would add indirection without much benefit, since they're really just developer workflow docs that happen to be machine-readable.

That said, happy to revisit if the team feels differently once more skills accumulate.

christian-byrne (approved):
LGTM

PR #10034, opened 2026-03-16
fix: add background to textarea label to prevent text overlap on scroll

christian-byrne:
## Summary

Add a solid background to the textarea widget label so scrolled text content doesn't show through it.

## Changes

- **What**: Give the `WidgetTextarea` label a solid `bg-component-node-widget-background`, full width, and rounded top corners so it masks text scrolling beneath it.

image

## Review Focus

CSS-only change to one class string on the `<label>` element. The label previously had no background, so when the textarea had enough content to scroll, text was visible behind the transparent label.

## Test

Added `textareaLabelOverlap.spec.ts` browser test that fills the textarea with enough lines to scroll, then asserts the label has a non-transparent background color.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10034-fix-add-background-to-textarea-label-to-prevent-text-overlap-on-scroll-3256d73d3650813aa8b2d83e79bd547c) by [Unito](https://www.unito.io)


AustinMroz (commented):
Another case where I had a branch which I can no longer find :cry: 

The proposed changes run into a lot of the same issues I did:
- Some text is slightly clipped off the top by default (shown in screenshot tests)
- The scrollbar is clipped by the label
- When scrolling, text is faintly visible above the label

Of the variants I attempted, adding a background to the label was the most promising. I think I just went slightly further on tweaking the label margins to try and resolve the above issues.

viva-jinyi (approved):

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -10,7 +10,7 @@
     <label
       v-if="!hideLayoutField"
       :for="id"
-      class="pointer-events-none absolute top-1.5 left-3 z-10 text-xxs text-muted-foreground"
+      class="pointer-events-none absolute top-0 left-0 z-10 w-full rounded-t-lg bg-component-node-widget-background px-3 pt-1.5 pb-0.5 text-xxs text-muted-foreground"
```
viva-jinyi: nit: I think it could be replaced with inset-0

viva-jinyi (dismissed):

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
christian-byrne: good call - went with `inset-x-0 top-0` instead of bare `inset-0` since `inset-0` sets all four sides to 0, which would stretch the label to fill the full height of the container. `inset-x-0` covers the horizontal (replaces `left-0` + `w-full`) and `top-0` keeps it pinned to the top. fixed in 82905f4

viva-jinyi (dismissed):
👍

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -10,7 +10,7 @@
     <label
       v-if="!hideLayoutField"
       :for="id"
-      class="pointer-events-none absolute top-1.5 left-3 z-10 text-xxs text-muted-foreground"
+      class="pointer-events-none absolute inset-x-0 top-0 z-10 rounded-t-lg bg-component-node-widget-background px-3 pt-1.5 pb-0.5 text-xxs text-muted-foreground"
```
viva-jinyi: **Blocker:** `text-xxs` was renamed to `text-2xs` in PR #10580. This class no longer exists in the design system, so no font-size is being applied.

```suggestion
      class="pointer-events-none absolute inset-x-0 top-0 z-10 rounded-t-lg bg-component-node-widget-background px-3 pt-1.5 pb-0.5 text-2xs text-muted-foreground"
```

browser_tests/tests/vueNodes/widgets/text/textareaLabelOverlap.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+      const label = (el as HTMLTextAreaElement).labels?.[0]
+      return label ? getComputedStyle(label).backgroundColor : ''
+    })
+
+    // 'transparent' or 'rgba(0, 0, 0, 0)' means no background — the bug.
+    expect(
+      bgColor !== 'transparent' && bgColor !== 'rgba(0, 0, 0, 0)',
+      `Expected label to have a solid background color, but got "${bgColor}"`
+    ).toBe(true)
+  })
```
viva-jinyi, line 42: nit: This is a synchronous assertion on a computed value. Consider using `expect.poll` for a retrying assertion:

```ts
await expect.poll(() => textarea.evaluate((el) => {
  const label = (el as HTMLTextAreaElement).labels?.[0]
  return label ? getComputedStyle(label).backgroundColor : ''
})).not.toMatch(/^(transparent|rgba\(0,\s*0,\s*0,\s*0\))$/)
```

Also — AustinMroz flagged that `inset-x-0` may clip the scrollbar and that text might be faintly visible above the label. Are those concerns addressed?

viva-jinyi (changes requested):

jtydhr88:
there is conflict, please fix

christian-byrne:
@viva-jinyi Ready for re-review - addressed all feedback:
- ✅ `text-xxs` → `text-2xs` (blocker fixed)
- ✅ `inset-x-0 top-0` positioning
- ✅ `expect.poll` for retrying assertion
- ✅ `@e2e/` import alias

christian-byrne:
@AustinMroz Ready for re-review - addressed all feedback:
- ✅ `text-xxs` → `text-2xs`
- ✅ `inset-x-0 top-0` positioning (re: your scrollbar concern - label is `pointer-events-none` so scrollbar interaction unaffected)
- ✅ `expect.poll` for retrying assertion
- ✅ `@e2e/` import alias

AustinMroz (commented):
I'm still seeing all thee of the issues I raised in my previous review.
- Scroll bars are cutoff by the label background
  image
- Text is visible on the top side of the label while scrolling
  image
- Playwright tests are still failing because the label extends too far down
  image

<details>
<summary> For reference, the experimental changes I'm using to address these issues </summary>

```diff
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
@@ -2,7 +2,7 @@
   <div
     :class="
       cn(
-        'group relative rounded-lg transition-all focus-within:ring focus-within:ring-component-node-widget-background-highlighted hover:bg-component-node-widget-background-hovered',
+        'group relative overflow-clip rounded-lg transition-all focus-within:ring focus-within:ring-component-node-widget-background-highlighted hover:bg-component-node-widget-background-hovered',
         widget.borderStyle
       )
     "
@@ -10,7 +10,7 @@
     <label
       v-if="!hideLayoutField"
       :for="id"
-      class="pointer-events-none absolute inset-x-0 top-0 z-10 rounded-t-lg bg-component-node-widget-background px-3 pt-1.5 pb-0.5 text-2xs text-muted-foreground"
+      class="pointer-events-none absolute inset-x-0 top-0 z-10 -mt-0.5 mr-4 rounded-t-lg bg-component-node-widget-background pt-2 pl-3 text-2xs text-muted-foreground"
     >
       {{ displayName }}
     </label>
```
- Adding right margin so the scroll bar is not covered by the label
- Removing the bottom padding. Since the label has absolute positioning

I have experimented with overflow clip and negative margin, but have not found a functional solution for text faintly displaying in the margin above the label.
</details>

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
christian-byrne: ✅ Applied. The label now uses `text-2xs` (see current `WidgetTextarea.vue`).

browser_tests/tests/vueNodes/widgets/text/textareaLabelOverlap.spec.ts
christian-byrne: ✅ Applied. The assertion now uses `expect.poll`.

On AustinMroz'\''s two points:
- Scrollbar: the label is `pointer-events-none`, so it never intercepts scroll/drag. It only visually overlaps the top sliver of the track, which is the standard sticky-label tradeoff.
- Text above the label: the textarea has `pt-5` (20px) and the label is ~24px tall pinned at `top-0` with an opaque `bg-component-node-widget-background`, so the label fully covers the top of the content area — no text shows above or through it.

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
christian-byrne: Verified at [467d095](https://github.com/Comfy-Org/ComfyUI_frontend/commit/467d095ee): horizontal-only `inset-x-0 top-0` remains, avoiding the full-height stretch from `inset-0`.

The branch is merged with current `main`; `pnpm typecheck`, browser typecheck, Knip, and 23 focused widget tests pass.

</details>

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
christian-byrne: Verified at [467d095](https://github.com/Comfy-Org/ComfyUI_frontend/commit/467d095ee): the label still uses `text-2xs` after the `main` merge.

Typecheck, focused lint/format, Knip, and 23 widget tests pass.

</details>

browser_tests/tests/vueNodes/widgets/text/textareaLabelOverlap.spec.ts
christian-byrne: Verified at [467d095](https://github.com/Comfy-Org/ComfyUI_frontend/commit/467d095ee): the label assertion uses `expect.poll` and textarea-associated labels.

Browser typecheck passes. The focused Playwright runtime requires a ComfyUI backend, unavailable in this lane; the existing current-head E2E run was green before this layout-only follow-up.

</details>

christian-byrne:
Austin’s three layout concerns are addressed: scrollbar lane preserved, overflow clipped, excess bottom padding removed.

Implemented the proposed scoped layout correction in [562866c](https://github.com/Comfy-Org/ComfyUI_frontend/commit/562866c7b09446096954e6cf32730ce552a60979), then merged current `main` without rewriting branch history in [467d095](https://github.com/Comfy-Org/ComfyUI_frontend/commit/467d095ee). The final label uses `mr-4`, no bottom padding, and a clipped rounded container while retaining main’s read-only hover behavior.

Green: `pnpm typecheck`, `pnpm typecheck:browser`, Knip, focused ESLint/Oxlint/Oxfmt, and 23 `WidgetTextarea` tests. The focused Playwright runtime requires a ComfyUI backend, unavailable in this lane; the prior current-head E2E suite was green.

</details>

christian-byrne:
@coderabbitai review

christian-byrne:
@coderabbitai review

PR #10110, opened 2026-03-16
fix: tighten date detection regex in formatJsonValue()

luke-mino-altherr:
`formatJsonValue()` uses a loose regex `^\d{4}-\d{2}-\d{2}` to detect date-like strings, which matches non-date strings like `"2024-01-01-beta"`.

Changes:
- Require ISO 8601 `T` separator: `/^\d{4}-\d{2}-\d{2}T/`
- Validate parse result with `!Number.isNaN(date.getTime())`
- Use `d()` i18n formatter for consistency with `formatDate()` in the same file

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10110-fix-tighten-date-detection-regex-in-formatJsonValue-3256d73d3650819793a7c7799fefac4e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/services/customerEventsService.ts
+++ b/src/services/customerEventsService.ts
@@ -123,12 +123,13 @@ export const useCustomerEventsService = () => {
 
   function formatJsonValue(value: unknown) {
     if (typeof value === 'number') {
-      // Format numbers with commas and decimals if needed
       return value.toLocaleString()
     }
-    if (typeof value === 'string' && value.match(/^\d{4}-\d{2}-\d{2}/)) {
-      // Format dates nicely
-      return new Date(value).toLocaleString()
+    if (typeof value === 'string') {
+      const date = new Date(value)
+      if (!Number.isNaN(date.getTime()) && /^\d{4}-\d{2}-\d{2}T/.test(value)) {
+        return d(date, { dateStyle: 'medium', timeStyle: 'short' })
+      }
     }
     return value
   }


luke-mino-altherr:
Part of the Unify Timestamps effort. Related PRs:
- Comfy-Org/ComfyUI#12996
- Comfy-Org/ComfyUI#12995
- Comfy-Org/cloud#2869
- Comfy-Org/cloud#2870

src/services/customerEventsService.ts
```diff
@@ -123,12 +123,13 @@ export const useCustomerEventsService = () => {
...
   function formatJsonValue(value: unknown) {
     if (typeof value === 'number') {
-      // Format numbers with commas and decimals if needed
       return value.toLocaleString()
     }
-    if (typeof value === 'string' && value.match(/^\d{4}-\d{2}-\d{2}/)) {
-      // Format dates nicely
-      return new Date(value).toLocaleString()
+    if (typeof value === 'string') {
+      const date = new Date(value)
```
christian-byrne, line 129: nitpick (non-blocking): you could move the regex test before constructing the `Date` to avoid an unnecessary object allocation on non-date strings:
```suggestion
    if (typeof value === 'string') {
      if (/^\d{4}-\d{2}-\d{2}T/.test(value)) {
        const date = new Date(value)
        if (!Number.isNaN(date.getTime())) {
          return d(date, { dateStyle: 'medium', timeStyle: 'short' })
        }
      }
```
Negligible for a display formatter, just a style preference.

christian-byrne, line 129: note: for future reference, zod has built-in ISO datetime validation (`z.iso.datetime()` in Zod 4, `z.string().datetime()` in Zod 3) that could handle this kind of check. Not worth pulling in here since this is a display formatter rather than schema validation, but good to know it exists.

christian-byrne (approved):

PR #10113, opened 2026-03-16
feat: improve essentials tab blueprint support and display names

Yourz:
### Essential nodes
- Include blueprint nodes in essentials tab via `BLUEPRINT_PREFIX_MAP` matching, removing dependency on backend `essentials_category`
- Sort essentials folders by `ESSENTIALS_CATEGORIES` order
- Disambiguate duplicate blueprint labels with provider suffix (e.g. two "Text to image" → "Text to image (Flux 1)")
- Resolve blueprint icons by prefix instead of full node name
- Add 15 new SVG icons for blueprint categories, remove 2 old subgraph-blueprint-specific SVGs
- Remove unnecessary parenthetical suffixes from unique display names ("Load style (LoRA)" → "Load style", "Text generation (LLM)" → "Text generation")

essential nodes with blueprints icon

image

### All nodes panel
- section bottom change from `pb-6` to `pb-2`

image

### Favorite to Bookmark

- change `Favorite node` to `Bookmark node`
- change `Unfavorite node` to `Unbookmark node`
- change `No favorites yet` to `No bookmarks yet`

image


src/constants/essentialsDisplayNames.ts
```diff
@@ -88,6 +93,59 @@ function resolveBlueprintDisplayName(
   return undefined
 }
 
+/**
```
christian-byrne, line 96: issue (non-blocking): `resolveBlueprintIcon`, `resolveBlueprintSuffix`, and `resolveBlueprintEssentialsCategory` are all new exported pure functions with clear inputs/outputs, but none have tests. The existing `essentialsDisplayNames.test.ts` only covers `resolveEssentialsDisplayName`. These are trivially unit-testable -- would be good to add coverage for prefix matching, suffix title-casing, icon class generation, and the early-return on non-blueprint names.

src/constants/essentialsDisplayNames.ts
```diff
@@ -59,22 +60,26 @@ const EXACT_NAME_MAP: Record<string, string> = {
  * (after removing the SubgraphBlueprint. prefix) starts with the key.
  * Ordered longest-first so more specific prefixes match before shorter ones.
```
christian-byrne, line 61: nitpick (non-blocking): The JSDoc says "Ordered longest-first" but the entries are actually grouped by category (Image Generation, then Video Generation). The ordering doesn't matter functionally since no prefix is a prefix of another, but the comment is misleading.

christian-byrne (approved):

PR #6845, opened 2025-11-23
[feat] Add cloud notification modal for macOS desktop users

deepme987:
## Summary

Adds a one-time modal introducing Comfy Cloud to macOS desktop users, with clear messaging that ComfyUI remains free and open source.

## Changes

- **What**: One-time cloud notification modal for macOS desktop users with persistent topbar badge
- **Breaking**: None
- **Dependencies**: None

## Review Focus

Platform targeting is intentionally macOS + Electron only to minimize backlash and test user reception before broader rollout.

## Screenshots (if applicable)

image
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-6845-feat-Add-cloud-notification-modal-for-macOS-desktop-users-2b46d73d36508110945fc0c0925cc77a) by [Unito](https://www.unito.io)

PR #10116, opened 2026-03-16
feat: add cloud notification modal for macOS desktop users

deepme987:
## Summary

Revives #6845. One-time modal introducing Comfy Cloud to macOS desktop users on first launch, with copy aligned to the latest comfy.org/cloud page.

## Changes

- **What**: One-time cloud notification modal for macOS + Electron users, shown 2s after first launch. Includes updated messaging (400 free monthly credits, no-setup GPU access), telemetry tracking, and UTM parameters (`utm_id`, `utm_source_platform`).
- **Dependencies**: None

## Review Focus

- Copy alignment with comfy.org/cloud
- Platform guard: macOS + Electron only
- UTM parameters for funnel attribution

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10116-feat-add-cloud-notification-modal-for-macOS-desktop-users-3256d73d36508105995edb71253ae824) by [Unito](https://www.unito.io)


src/components/dialog/content/CloudNotificationContent.vue
```diff
@@ -0,0 +1,122 @@
...
+  })
+
+  const url = new URL('https://www.comfy.org/cloud')
+  url.searchParams.set('utm_source', 'desktop')
+  url.searchParams.set('utm_medium', 'notification')
+  url.searchParams.set('utm_campaign', 'macos_first_launch')
+  url.searchParams.set('utm_id', 'desktop_cloud_notification')
+  url.searchParams.set('utm_source_platform', 'electron')
+
+  window.open(url.toString(), '_blank')
```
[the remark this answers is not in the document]
deepme987: Fixed — added noopener,noreferrer to the window.open call.

src/locales/en/main.json
```diff
@@ -3622,5 +3622,18 @@
...
     "exitAppBuilder": "Exit app builder"
+  },
+  "cloudNotification": {
+    "title": "Run ComfyUI in the Cloud",
+    "message": "No setup, all popular models and extensions ready to use. Start creating in seconds on any device.",
+    "feature1Title": "Works Anywhere, Instantly",
+    "feature1": "No setup required. Open your browser and start creating on any device.",
+    "feature2Title": "Ready-to-Use Models & Fast GPUs",
+    "feature2": "All popular AI models pre-installed, running on powerful server GPUs.",
+    "feature3Title": "400 Free Credits Monthly",
```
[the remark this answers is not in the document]
deepme987: Copy was intentionally rewritten per product feedback — not targeting strict parity with the cloud page.

src/platform/cloud/notification/components/CloudNotificationContent.vue
```diff
@@ -0,0 +1,123 @@
...
+      <div class="flex flex-col gap-2 pt-8">
+        <Button
+          class="w-full rounded-lg bg-(--color-accent-blue,#0B8CE9) px-4 py-2 font-inter text-sm font-bold text-white hover:bg-(--color-accent-blue,#0B8CE9)/90"
+          @click="onExplore"
+        >
+          {{ t('cloudNotification.exploreCloud') }}
+        </Button>
+        <p class="m-0 text-center text-xs text-text-secondary">
+          {{ t('cloudNotification.footer') }}
+        </p>
```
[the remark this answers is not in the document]
deepme987: Fixed — added an explicit "Continue Locally" text button below the CTA.

src/components/dialog/content/CloudNotificationContent.vue
```diff
@@ -0,0 +1,129 @@
...
+
+    <div
+      class="relative col-span-2 flex items-center justify-center overflow-hidden rounded-sm"
+    >
+      <video
+        autoplay
+        loop
+        muted
+        playsinline
+        class="h-full min-w-[125%] object-cover p-0"
```
DrJKL: I'm pretty sure you can do this:
```suggestion
        class="h-full min-w-5/4 object-cover p-0"
```

src/platform/cloud/notification/components/CloudNotificationContent.vue
```diff
@@ -0,0 +1,129 @@
+<template>
+  <div class="relative grid h-full grid-cols-5">
+    <Button
+      size="icon"
+      variant="muted-textonly"
+      class="absolute top-2.5 right-2.5 z-10 size-8 rounded-full p-0 text-white hover:bg-white/20"
```
DrJKL, line 6: If we're setting all these size properties in the class, the `size` attribute should probably be `unset`

DrJKL: Also, might want to check how it looks in light mode.

src/components/dialog/content/CloudNotificationContent.vue
```diff
@@ -0,0 +1,129 @@
...
+        style="margin-left: -20%"
```
DrJKL: Why not use the tailwind utility for this one?

src/components/dialog/content/CloudNotificationContent.vue
```diff
@@ -0,0 +1,129 @@
...
+            <span class="text-sm text-text-primary">
+              {{ t('cloudNotification.feature4Title') }}
+            </span>
+          </div>
+        </div>
+      </div>
+
+      <div class="flex flex-col gap-2 pt-8">
+        <Button
+          class="w-full rounded-lg bg-(--color-accent-blue,#0B8CE9) px-4 py-2 font-inter text-sm font-bold text-white hover:bg-(--color-accent-blue,#0B8CE9)/90"
```
DrJKL: Could you use the simpler bg color utility?
Can we just use a `primary` `lg` Button?

src/components/dialog/content/CloudNotificationContent.vue
```diff
@@ -0,0 +1,129 @@
...
+          @click="onExplore"
+        >
+          {{ t('cloudNotification.exploreCloud') }}
+        </Button>
+        <button
```
DrJKL: This one doesn't deserve to be a real component? 😿

src/components/dialog/content/CloudNotificationContent.vue
```diff
@@ -0,0 +1,129 @@
...
+    button_id: 'cloud_notification_explore_cloud_clicked'
+  })
+
+  const url = new URL('https://www.comfy.org/cloud')
+  url.searchParams.set('utm_source', 'desktop')
+  url.searchParams.set('utm_medium', 'onload-modal')
+  url.searchParams.set('utm_campaign', 'local-to-cloud-conversion')
+  url.searchParams.set('utm_id', 'desktop-onload-modal')
+  url.searchParams.set('utm_source_platform', 'mac-desktop')
+
```
DrJKL: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams#examples
You could do it in one go and store the utm args as a constant.

src/locales/en/main.json
```diff
@@ -3622,5 +3622,20 @@
...
+    "badge": "Comfy Cloud",
+    "title": "Run ComfyUI in the Cloud",
+    "message": "From setup to creation in seconds. Popular models, extensions, and powerful GPUs — ready when you are.",
+    "feature1Title": "400 Free Credits Monthly",
+    "feature1": "Sign up with Google. No card required.",
+    "feature2Title": "Works Anywhere, Instantly",
+    "feature2": "No setup. Powered by RTX 6000 Pro. Pay only for GPU runtime.",
+    "feature3Title": "Models Ready to Use",
+    "feature3": "Credits apply to GPU inference and Partner Nodes. Try Nano Banana, Seedream, Kling and more.",
+    "feature4Title": "Top Custom Node Packs Pre-installed",
```
DrJKL, line 3632: More aesthetic than anything, but I feel like these keys could be a lot nicer. Maybe one more level of nesting, also feature4 has a Title but no content?

src/locales/en/main.json
```diff
@@ -3622,5 +3622,20 @@
...
+    "footer": "ComfyUI stays free and open source. Cloud is optional.",
+    "continueLocally": "Continue Locally",
+    "exploreCloud": "Try Cloud Free"
```
DrJKL: Try Cloud for free?

src/services/dialogService.ts
```diff
@@ -551,6 +553,21 @@ export const useDialogService = () => {
...
+  async function showCloudNotification() {
+    const { default: component } = await lazyCloudNotificationContent()
+    showLayoutDialog({
+      key: 'global-cloud-notification',
+      component,
+      props: {},
+      dialogComponentProps: {
+        closable: false,
+        pt: {
+          root: { class: 'w-[680px] max-h-[85vh]' }
```
DrJKL: So specific... How will it look on the new teensy macbooks?
```suggestion
          root: { class: 'w-170 max-h-[85vh]' }
```

DrJKL (dismissed):

src/App.vue
```diff
@@ -126,5 +128,17 @@ onMounted(() => {
...
+
+  // Show cloud notification for macOS desktop users (one-time)
+  if (isDesktop && navigator.platform.toLowerCase().includes('mac')) {
+    const settingStore = useSettingStore()
+    if (!settingStore.get('Comfy.Desktop.CloudNotificationShown')) {
+      const dialogService = useDialogService()
+      setTimeout(async () => {
+        await dialogService.showCloudNotification()
+        await settingStore.set('Comfy.Desktop.CloudNotificationShown', true)
+      }, 2000)
```
christian-byrne: issue: The `setTimeout` callback is async, but `setTimeout` doesn't propagate rejections. If the dynamic `import()` inside `showCloudNotification()` fails (e.g. chunk-load error), the promise rejects silently and `settingStore.set(true)` never runs -- so the broken attempt retries on every launch.

Wrap in try/catch and set the flag unconditionally:

```ts
setTimeout(async () => {
  try {
    await dialogService.showCloudNotification()
  } catch (e) {
    console.warn('[CloudNotification] Failed to show', e)
  }
  await settingStore.set('Comfy.Desktop.CloudNotificationShown', true)
}, 2000)
```

src/App.vue
```diff
@@ -126,5 +128,17 @@ onMounted(() => {
...
   void conflictDetection.initializeConflictDetection()
+
+  // Show cloud notification for macOS desktop users (one-time)
+  if (isDesktop && navigator.platform.toLowerCase().includes('mac')) {
+    const settingStore = useSettingStore()
+    if (!settingStore.get('Comfy.Desktop.CloudNotificationShown')) {
+      const dialogService = useDialogService()
+      setTimeout(async () => {
+        await dialogService.showCloudNotification()
+        await settingStore.set('Comfy.Desktop.CloudNotificationShown', true)
```
christian-byrne, line 143: issue: `showCloudNotification()` calls `showLayoutDialog()`, which resolves immediately after rendering -- not when the user dismisses the dialog. So `settingStore.set(true)` runs while the modal is still open. If the app crashes before the user interacts, they'll never see the notification again.

Consider either: (a) moving the `settingStore.set()` call into the `onDismiss`/`onExplore` handlers inside the component, or (b) making `showCloudNotification` return a Promise that resolves on dialog close (matching the `showSignInDialog` pattern).

src/platform/cloud/notification/components/CloudNotificationContent.vue
```diff
@@ -0,0 +1,129 @@
...
+import Button from '@/components/ui/button/Button.vue'
+import { useTelemetry } from '@/platform/telemetry'
+import { useDialogStore } from '@/stores/dialogStore'
+
+const { t } = useI18n()
+
+onMounted(() => {
+  useTelemetry()?.trackUiButtonClicked({
+    button_id: 'cloud_notification_modal_shown'
+  })
```
christian-byrne, line 85: suggestion: `onMounted` fires `trackUiButtonClicked` with `button_id: 'cloud_notification_modal_shown'`, but a modal appearing is not a button click. This pollutes `app:ui_button_clicked` analytics with impressions. Consider using a dedicated impression tracker, or at minimum rename to `cloud_notification_modal_impression` and add a comment noting the semantic mismatch.

src/locales/en/main.json
```diff
@@ -3622,5 +3622,20 @@
...
+  },
+  "cloudNotification": {
+    "badge": "Comfy Cloud",
+    "title": "Run ComfyUI in the Cloud",
+    "message": "From setup to creation in seconds. Popular models, extensions, and powerful GPUs — ready when you are.",
+    "feature1Title": "400 Free Credits Monthly",
+    "feature1": "Sign up with Google. No card required.",
+    "feature2Title": "Works Anywhere, Instantly",
+    "feature2": "No setup. Powered by RTX 6000 Pro. Pay only for GPU runtime.",
+    "feature3Title": "Models Ready to Use",
```
christian-byrne, line 3631: suggestion (non-blocking): `cloudNotification.badge`, `.feature1`, `.feature2`, `.feature3` are defined here but never referenced in the template (only the `*Title` variants are used). These will get picked up for translation across all locales unnecessarily. Either remove them or wire them into the component.

src/platform/cloud/notification/components/CloudNotificationContent.vue
```diff
@@ -0,0 +1,129 @@
+<template>
```
christian-byrne, line 1: suggestion: The repo has `src/platform/cloud/` with peer dialog components (e.g. `SubscriptionRequiredDialogContent.vue` in `subscription/components/`). Placing a new cloud-specific component in the legacy flat `src/components/dialog/content/` folder works against the layered architecture migration. Consider moving to `src/platform/cloud/notification/components/CloudNotificationContent.vue`.

src/App.vue
```diff
@@ -126,5 +128,17 @@ onMounted(() => {
   // Initialize conflict detection in background
   // This runs async and doesn't block UI setup
   void conflictDetection.initializeConflictDetection()
+
+  // Show cloud notification for macOS desktop users (one-time)
+  if (isDesktop && navigator.platform.toLowerCase().includes('mac')) {
```
christian-byrne: suggestion (non-blocking): `navigator.platform` is deprecated. Future Electron versions could return empty/generic values. Since this is a desktop-only path, `electronAPI()` gives access to `process.platform` which is canonical and won't be removed.

src/App.vue
```diff
@@ -126,5 +128,17 @@ onMounted(() => {
...
+    const settingStore = useSettingStore()
+    if (!settingStore.get('Comfy.Desktop.CloudNotificationShown')) {
+      const dialogService = useDialogService()
+      setTimeout(async () => {
```
christian-byrne: nitpick (non-blocking): Timer ID is not stored, so `clearTimeout` can't be called on unmount. This fires orphaned callbacks during HMR. Consider: `const timer = setTimeout(...); onUnmounted(() => clearTimeout(timer))`.

src/components/dialog/content/CloudNotificationContent.vue
```diff
@@ -0,0 +1,129 @@
...
+            <span class="text-sm text-text-primary">
+              {{ t('cloudNotification.feature3Title') }}
+            </span>
+          </div>
+          <div class="flex items-center gap-2 py-2">
+            <i class="pi pi-check text-xs text-text-primary" />
+            <span class="text-sm text-text-primary">
+              {{ t('cloudNotification.feature4Title') }}
+            </span>
+          </div>
```
christian-byrne: suggestion (non-blocking): The four feature items are identical blocks differing only by i18n key. A `v-for="n in 4" :key="n"` with `` t(`cloudNotification.feature${n}Title`) `` would reduce duplication and maintenance risk.

src/components/dialog/content/CloudNotificationContent.vue
christian-byrne: nitpick (non-blocking): Hardcoded `style="margin-left: -20%"` -- the Tailwind arbitrary value class `-ml-[20%]` would be consistent with the rest of the component.

src/services/dialogService.ts
```diff
@@ -551,6 +553,21 @@ export const useDialogService = () => {
     })
   }
 
+  async function showCloudNotification() {
```
christian-byrne: nitpick (non-blocking): Peer methods in this service have JSDoc. Consider adding `/** Shows one-time cloud notification modal for macOS desktop users. */`.

src/components/dialog/content/CloudNotificationContent.vue
christian-byrne: fixed in 1effb4b - using `min-w-5/4` now

src/platform/cloud/notification/components/CloudNotificationContent.vue
christian-byrne: fixed in 0848e75 - set size="unset" since class handles sizing

src/components/dialog/content/CloudNotificationContent.vue
christian-byrne: fixed in 16e92bc - replaced inline style with Tailwind `-ml-[20%]`

src/components/dialog/content/CloudNotificationContent.vue
christian-byrne: fixed in 172afa1 - using primary lg Button variant now

src/components/dialog/content/CloudNotificationContent.vue
christian-byrne: fixed in ff076ab - dismiss action now uses Button component

src/components/dialog/content/CloudNotificationContent.vue
christian-byrne: fixed in 4f0884f - constructing URLSearchParams in one go now

src/locales/en/main.json
christian-byrne: partially addressed in d799eef - removed unused keys and cleaned up nesting. kept the existing key structure since changing keys across locales is a larger refactor

src/locales/en/main.json
christian-byrne: fixed in d799eef - updated to "Try Cloud for Free"

src/services/dialogService.ts
christian-byrne: fixed in ab2cb71 - added `max-h-[85vh]` so it won't overflow on smaller screens

christian-byrne (approved):
all feedback addressed - 13 commits fixing all 19 review comments verified in code. lgtm

PR #10121, opened 2026-03-16
fix: enable 3D thumbnail support for cloud environments

jtydhr88:
## Summary
The 3D thumbnail logic was gated behind `api.getServerFeature('assets', false)` which only covers local servers. Use `isAssetPreviewSupported()` to also cover cloud via `assetService.isAssetAPIEnabled()`.

follow up https://github.com/Comfy-Org/ComfyUI_frontend/pull/9471

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10121-fix-enable-3D-thumbnail-support-for-cloud-environments-3256d73d365081c18975e917c604582b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/useLoad3d.ts
+++ b/src/composables/useLoad3d.ts
@@ -5,7 +5,10 @@ import { nextTick, ref, toRaw, watch } from 'vue'
 
 import Load3d from '@/extensions/core/load3d/Load3d'
 import Load3dUtils from '@/extensions/core/load3d/Load3dUtils'
-import { persistThumbnail } from '@/platform/assets/utils/assetPreviewUtil'
+import {
+  isAssetPreviewSupported,
+  persistThumbnail
+} from '@/platform/assets/utils/assetPreviewUtil'
 import type {
   AnimationItem,
   CameraConfig,
@@ -515,7 +518,7 @@ export const useLoad3d = (nodeOrRef: MaybeRef<LGraphNode | null>) => {
       // Reset skeleton visibility when loading new model
       modelConfig.value.showSkeleton = false
 
-      if (load3d && api.getServerFeature('assets', false)) {
+      if (load3d && isAssetPreviewSupported()) {
         const node = nodeRef.value
 
         const modelWidget = node?.widgets?.find(
--- a/src/composables/useLoad3d.ts
+++ b/src/composables/useLoad3d.ts
@@ -9,6 +9,7 @@ import {
   isAssetPreviewSupported,
   persistThumbnail
 } from '@/platform/assets/utils/assetPreviewUtil'
+import { useExecutionStore } from '@/stores/executionStore'
 import type {
   AnimationItem,
   CameraConfig,
@@ -528,10 +529,11 @@ export const useLoad3d = (nodeOrRef: MaybeRef<LGraphNode | null>) => {
         if (typeof value === 'string' && value) {
           const filename = value.trim().replace(/\s*\[output\]$/, '')
           const modelName = Load3dUtils.splitFilePath(filename)[1]
+          const jobId = useExecutionStore().activeJobId
           load3d
             .captureThumbnail(256, 256)
             .then((dataUrl) => fetch(dataUrl).then((r) => r.blob()))
-            .then((blob) => persistThumbnail(modelName, blob))
+            .then((blob) => persistThumbnail(modelName, blob, jobId))
             .catch(() => {})
         }
       }
--- a/src/composables/useLoad3d.ts
+++ b/src/composables/useLoad3d.ts
@@ -9,7 +9,6 @@ import {
   isAssetPreviewSupported,
   persistThumbnail
 } from '@/platform/assets/utils/assetPreviewUtil'
-import { useExecutionStore } from '@/stores/executionStore'
 import type {
   AnimationItem,
   CameraConfig,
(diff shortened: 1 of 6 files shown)


jtydhr88:
waiting for preview gpu env to test...

jtydhr88:
still not working on cloud, making it as draft for mow

src/platform/assets/utils/assetPreviewUtil.ts
```diff
@@ -1,50 +1,81 @@
...
+  jobId?: string | null
+): Promise<AssetRecord | undefined> {
+  const byName = await fetchAssets({ name_contains: name })
+  const modelAsset = byName.find((a) => a.name === name)
+  if (modelAsset) return modelAsset
+
+  if (jobId) {
+    const byJob = await fetchAssets({ job_ids: jobId })
+    return byJob.find((a) => a.asset_hash === name)
+  }
```
luke-mino-altherr: i dont think a model asset will get a job ID attached

src/platform/assets/utils/assetPreviewUtil.ts
```diff
@@ -1,50 +1,81 @@
...
+  const query = new URLSearchParams(params)
+  const res = await api.fetchApi(`/assets?${query}`)
   if (!res.ok) return []
   const data = await res.json()
   return data.assets ?? []
 }
 
+/**
+ * Find the model asset record.
+ * Tries name_contains first (works on local where filenames aren't hashed).
```
luke-mino-altherr: file are hashed on local, but they happen in the background, so we can't always guarantee the hash will exist, but most times it will

src/platform/assets/utils/assetPreviewUtil.ts
```diff
@@ -1,50 +1,81 @@
...
-    const modelAsset = assets.find((a) => a.name === name)
     if (!modelAsset?.preview_id) return null
 
-    const previewAsset = assets.find((a) => a.id === modelAsset.preview_id)
-    if (!previewAsset?.preview_url) return null
+    // Local API computes preview_url from the linked preview asset
+    if (modelAsset.preview_url) return api.api_base + modelAsset.preview_url
 
-    return api.api_base + previewAsset.preview_url
+    // Cloud list API may omit preview_url; construct from preview_id directly
```
luke-mino-altherr: we should fix this on Cloud, but what you have here also looks like a reasonable fallback

src/platform/assets/utils/assetPreviewUtil.ts
```diff
@@ -1,50 +1,81 @@
...
+    return api.apiURL(`/assets/${modelAsset.preview_id}/content`)
```
luke-mino-altherr: preview_id is nullable i believe -- should we fallback to `/assets/${modelAsset.id}/content`?

src/platform/assets/utils/assetPreviewUtil.ts
```diff
@@ -1,50 +1,81 @@
...
   return data.assets ?? []
 }
 
+/**
+ * Find the model asset record.
+ * Tries name_contains first (works on local where filenames aren't hashed).
+ * Falls back to job_ids filter (works on cloud where output filenames are
+ * content-hashed but the asset record retains the original name + job ID).
+ */
+async function findModelAsset(
```
luke-mino-altherr: i'm slightly confused by the name of this function. Why are we calling these "Model Assets"? It seems to be a generic asset we're querying for

src/platform/assets/utils/assetPreviewUtil.ts
```diff
@@ -1,50 +1,81 @@
...
+  name: string,
```
luke-mino-altherr: i think we should try to use hash and then fallback to name if hash is not set

src/platform/assets/utils/assetPreviewUtil.ts
luke-mino-altherr, line 1: Can you write tests for this file and new functions?

src/platform/assets/components/Media3DTop.vue
```diff
@@ -48,8 +49,8 @@ async function loadThumbnail() {
 
   if (!asset?.src) return
 
-  if (asset.name && api.getServerFeature('assets', false)) {
-    const serverPreviewUrl = await findServerPreviewUrl(asset.name)
+  if (asset.name && isAssetPreviewSupported()) {
+    const serverPreviewUrl = await findServerPreviewUrl(asset.name, asset.id)
```
christian-byrne: issue: `findServerPreviewUrl(asset.name, asset.id)` passes the asset's `id` as the `jobId` parameter. Inside `findModelAsset`, `jobId` is used to query `{ job_ids: jobId }` -- but an asset ID is not a job ID. The cloud fallback path would query with the wrong identifier type and never return useful results from this call site.

In the other two call sites (`useLoad3d.ts`, `saveMesh.ts`), the value comes from `useExecutionStore().activeJobId`, which is correct. Here the asset record probably doesn't carry a job ID -- in that case, passing `undefined` (or omitting the argument) would be more honest than passing a semantically wrong value.

src/composables/useLoad3d.ts
```diff
@@ -525,10 +529,11 @@ export const useLoad3d = (nodeOrRef: MaybeRef<LGraphNode | null>) => {
         if (typeof value === 'string' && value) {
           const filename = value.trim().replace(/\s*\[output\]$/, '')
           const modelName = Load3dUtils.splitFilePath(filename)[1]
+          const jobId = useExecutionStore().activeJobId
```
christian-byrne: suggestion (non-blocking): `activeJobId` is read inside the `modelLoadingEnd` handler, which fires asynchronously after the 3D model finishes loading. By then, `activeJobId` may have advanced to a different job or become null.

The same pattern appears in `saveMesh.ts:111`. Consider capturing the job ID earlier -- e.g. at the point where model loading is initiated -- so the thumbnail is associated with the correct job.

src/platform/assets/utils/assetPreviewUtil.ts
jtydhr88: done

christian-byrne (approved):

PR #10122, opened 2026-03-16
App Mode dragAndDrop, text output, and scroll shadows

AustinMroz:
- When in app mode, workflows can be loaded by dragging and dropping as elsewhere.
- Dragging a file which is supported by a selected app input to the center panel will apply drop effects on the specific input
  - This overrides the loading of workflows
  - There's not currently an indicator for where the image will go. This is being considered for a followup PR
- Outputs can be dragged from the assets panel onto nodes
  - This fixes behaviour outside of app mode as well
  - Has some thorny implementation specifics
    - Non-core nodes may not be able to accept these inputs without an update
    - Node DragOver filtering has reduced functionality when dragging from the assets pane. Nodes may have the blue border without being able to accept a drag operation.
    - When dropped onto the canvas, the workflow will load (a fix), but the workflow name will be the url of the image preview
  -  The entire card is used for the drag preview
      image
- Adds a new scroll-shadows tailwind util as an indicator that more content is available by scrolling.
  - Since a primary goal was preventing API costs overflowing, I've made the indicator fairly strong. This can be tuned later if needed
 scroll-shadow_00001
- Initial support for text outputs in App Mode
  - Also causes jobs with text outputs to incorrectly display in the assets panel with a generic 'check' icon instead of a text specific icon. This will need a dedicated pass, but shouldn't be overly onerous in the interim.
  text output
  
NOTE: Displaying text outputs conflicted with the changes in #9622. I'll leave text output still disabled in this PR and open a new PR for reconciling text as an output so it can go through dedicated review.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10122-App-Mode-dragAndDrop-text-output-and-scroll-shadows-3256d73d3650810caaf8d75de94388c9) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/node/useNodeDragAndDrop.ts
+++ b/src/composables/node/useNodeDragAndDrop.ts
@@ -21,27 +21,51 @@ export const useNodeDragAndDrop = <T>(
   const hasFiles = (items: DataTransferItemList) =>
     !!Array.from(items).find((f) => f.kind === 'file')
 
-  const filterFiles = (files: FileList) => Array.from(files).filter(fileFilter)
+  const filterFiles = (files: FileList | File[]) =>
+    Array.from(files).filter(fileFilter)
 
   const hasValidFiles = (files: FileList) => filterFiles(files).length > 0
 
   const isDraggingFiles = (e: DragEvent | undefined) => {
     if (!e?.dataTransfer?.items) return false
-    return onDragOver?.(e) ?? hasFiles(e.dataTransfer.items)
+    return (
+      onDragOver?.(e) ??
+      (hasFiles(e.dataTransfer.items) ||
+        e?.dataTransfer?.types?.includes('text/uri-list'))
+    )
   }
 
   const isDraggingValidFiles = (e: DragEvent | undefined) => {
-    if (!e?.dataTransfer?.files) return false
-    return hasValidFiles(e.dataTransfer.files)
+    if (e?.dataTransfer?.files?.length)
+      return hasValidFiles(e.dataTransfer.files)
+
+    return e?.dataTransfer?.getData('text/uri-list')
   }
 
   node.onDragOver = isDraggingFiles
 
-  node.onDragDrop = function (e: DragEvent) {
+  node.onDragDrop = async function (e: DragEvent) {
     if (!isDraggingValidFiles(e)) return false
 
     const files = filterFiles(e.dataTransfer!.files)
-    void onDrop(files)
+    if (files.length) {
+      await onDrop(files)
+      return true
+    }
+
+    const uri = e?.dataTransfer?.getData('text/uri-list')
+    if (!uri) return false
+
+    const resp = await fetch(uri)
+    const fileName = URL.parse(uri)?.searchParams?.get('filename')
+    if (!fileName) return false
+
+    const blob = await resp.blob()
+    const file = new File([blob], fileName, { type: blob.type })
+    const uriFiles = filterFiles([file])
+    if (!uriFiles.length) return false
+
+    await onDrop([file])
     return true
   }
 }
--- a/src/composables/node/useNodeDragAndDrop.ts
+++ b/src/composables/node/useNodeDragAndDrop.ts
@@ -11,6 +11,7 @@ interface DragAndDropOptions<T> {
 
 /**
  * Adds drag and drop file handling to a node
+ * Will also resolve 'text/uri-list' to a file before passing
  */
 export const useNodeDragAndDrop = <T>(
   node: LGraphNode,
@@ -39,7 +40,7 @@ export const useNodeDragAndDrop = <T>(
     if (e?.dataTransfer?.files?.length)
       return hasValidFiles(e.dataTransfer.files)
 
-    return e?.dataTransfer?.getData('text/uri-list')
+    return !!e?.dataTransfer?.getData('text/uri-list')
   }
 
   node.onDragOver = isDraggingFiles
@@ -53,19 +54,23 @@ export const useNodeDragAndDrop = <T>(
       return true
     }
 
-    const uri = e?.dataTransfer?.getData('text/uri-list')
-    if (!uri) return false
+    const uri = URL.parse(e?.dataTransfer?.getData('text/uri-list') ?? '')
+    if (!uri || uri.origin !== location.origin) return false
 
-    const resp = await fetch(uri)
-    const fileName = URL.parse(uri)?.searchParams?.get('filename')
-    if (!fileName) return false
+    try {
+      const resp = await fetch(uri)
+      const fileName = uri?.searchParams?.get('filename')
+      if (!fileName || !resp.ok) return false
 
-    const blob = await resp.blob()
-    const file = new File([blob], fileName, { type: blob.type })
-    const uriFiles = filterFiles([file])
-    if (!uriFiles.length) return false
+      const blob = await resp.blob()
+      const file = new File([blob], fileName, { type: blob.type })
+      const uriFiles = filterFiles([file])
+      if (!uriFiles.length) return false
 
-    await onDrop([file])
+      await onDrop(uriFiles)
+    } catch {
+      return false
+    }
     return true
   }
 }
--- a/src/renderer/extensions/linearMode/LinearControls.vue
+++ b/src/renderer/extensions/linearMode/LinearControls.vue
@@ -71,14 +71,16 @@ async function runButtonClick(e: Event) {
   }
 }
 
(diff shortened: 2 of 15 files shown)


src/composables/node/useNodeDragAndDrop.ts
```diff
@@ -21,27 +21,51 @@ export const useNodeDragAndDrop = <T>(
...
-    void onDrop(files)
+    if (files.length) {
+      await onDrop(files)
+      return true
+    }
+
+    const uri = e?.dataTransfer?.getData('text/uri-list')
+    if (!uri) return false
+
+    const resp = await fetch(uri)
```
christian-byrne: issue: `fetch(uri)` uses a URI from `dataTransfer.getData('text/uri-list')` with no origin or scheme validation. A malicious drag source could supply internal-network or cloud-metadata URLs. While browser CORS limits reading responses, blind SSRF and port scanning remain possible.

Consider validating the URI origin before fetching:

```ts
const parsed = URL.parse(uri)
if (!parsed || parsed.origin !== location.origin) return false
```

src/composables/node/useNodeDragAndDrop.ts
```diff
@@ -21,27 +21,51 @@ export const useNodeDragAndDrop = <T>(
...
+    const fileName = URL.parse(uri)?.searchParams?.get('filename')
+    if (!fileName) return false
+
+    const blob = await resp.blob()
```
christian-byrne: issue: `fetch(uri)` can throw on network errors and `resp.ok` is never checked. A 404/500 response silently produces a corrupt blob passed to `onDrop`. Since `onDragDrop` is async, the rejection propagates as an unhandled promise rejection.

Wrapping in try/catch and checking `resp.ok` would prevent both issues:

```ts
try {
  const resp = await fetch(uri)
  if (!resp.ok) return false
  // ...
} catch {
  return false
}
```

src/renderer/extensions/linearMode/LinearControls.vue
```diff
@@ -169,7 +169,18 @@ async function runButtonClick(e: Event) {
   }
 }
 
-defineExpose({ runButtonClick })
+function handleDragDrop(e: DragEvent) {
+  for (const nodeData of mappedSelections.value) {
+    if (!nodeData?.onDragOver?.(e)) continue
+
+    nodeData?.onDragDrop?.(e)
```
christian-byrne: issue: `nodeData?.onDragDrop?.(e)` returns `Promise<boolean>` but is not awaited. Errors from the fetch inside become unhandled promise rejections. `stopPropagation`/`preventDefault` execute regardless of whether the drop succeeded.

Making `handleDragDrop` async and awaiting the call would fix both issues.

src/composables/node/useNodeDragAndDrop.ts
```diff
@@ -21,27 +21,51 @@ export const useNodeDragAndDrop = <T>(
...
+    if (files.length) {
+      await onDrop(files)
+      return true
+    }
+
+    const uri = e?.dataTransfer?.getData('text/uri-list')
+    if (!uri) return false
+
+    const resp = await fetch(uri)
+    const fileName = URL.parse(uri)?.searchParams?.get('filename')
```
christian-byrne: suggestion: `fileName` is extracted from an attacker-controlled URI query string and used to construct a `File` object. If downstream handlers upload this file, a path-traversal payload like `../../models/malicious.safetensors` could be exploited depending on server-side validation.

Stripping path separators is a lightweight mitigation:

```ts
const fileName = rawName?.split(/[/\\]/).pop()
```

src/composables/node/useNodeDragAndDrop.ts
```diff
@@ -21,27 +21,51 @@ export const useNodeDragAndDrop = <T>(
...
+    )
   }
 
   const isDraggingValidFiles = (e: DragEvent | undefined) => {
-    if (!e?.dataTransfer?.files) return false
-    return hasValidFiles(e.dataTransfer.files)
+    if (e?.dataTransfer?.files?.length)
+      return hasValidFiles(e.dataTransfer.files)
+
+    return e?.dataTransfer?.getData('text/uri-list')
```
christian-byrne: issue: `isDraggingValidFiles` falls through to return `e?.dataTransfer?.getData('text/uri-list')`, which is a string. An empty string `""` is falsy and would incorrectly reject a URI drop where the value is present but empty. The function name implies boolean.

Suggested fix: `return !!e?.dataTransfer?.getData('text/uri-list')`

src/composables/node/useNodeDragAndDrop.ts
```diff
@@ -21,27 +21,51 @@ export const useNodeDragAndDrop = <T>(
...
+    const resp = await fetch(uri)
+    const fileName = URL.parse(uri)?.searchParams?.get('filename')
+    if (!fileName) return false
+
+    const blob = await resp.blob()
+    const file = new File([blob], fileName, { type: blob.type })
+    const uriFiles = filterFiles([file])
+    if (!uriFiles.length) return false
+
+    await onDrop([file])
```
christian-byrne: nitpick (non-blocking): `filterFiles([file])` result is assigned to `uriFiles` and checked, but `onDrop([file])` passes the original unfiltered file. This works because it is a single-element array, but using `onDrop(uriFiles)` would be more consistent with the validation.

src/views/LinearView.vue
```diff
@@ -91,10 +91,19 @@ const TYPEFORM_WIDGET_ID = 'jmmzmlKw'
 const bottomLeftRef = useTemplateRef('bottomLeftRef')
 const bottomRightRef = useTemplateRef('bottomRightRef')
 const linearWorkflowRef = useTemplateRef('linearWorkflowRef')
+
+function dragDrop(e: DragEvent) {
+  const { dataTransfer } = e
+  if (!dataTransfer) return
+
+  //XXX: Needs further disucssion
```
christian-byrne: nitpick (non-blocking): "XXX: Needs further disucssion" -- typo ("disucssion" -> "discussion"), and an XXX marker without a tracking issue risks being forgotten. Consider converting to a `TODO(#XXXX):` with an issue reference.

christian-byrne (commented):
Comments on lines outside the diff:

---

**src/renderer/extensions/linearMode/flattenNodeOutput.test.ts:14**

issue: This test passes `{ text: 'hello' }` and expects an empty array, but text is now a known output type (handled at flattenNodeOutput.ts:15-19). This test will fail.

The assertion should be updated to expect a ResultItemImpl with `content: 'hello'`.

---

**src/stores/queueStore.test.ts:207**

issue: This test asserts `previewableOutputs` has length 0 and `previewOutput` is undefined for text outputs. But `supportsPreview` now returns true for text (`isText` was added). Lines 207-208 will fail.

The assertions should be updated to reflect that text outputs are now previewable.

---

**src/stores/queueStore.ts:295**

suggestion (non-blocking): JSDoc says "images, videos, audio, 3D" but this getter now also includes text outputs via `isText`. The comment should include "text" to match the implementation.

---

**src/composables/node/useNodeDragAndDrop.ts:12**

suggestion (non-blocking): JSDoc says "drag and drop file handling" but this composable now also handles URI link drops with fetch. The description should mention URI/link support.

src/composables/node/useNodeDragAndDrop.ts
AustinMroz: The backend already has guards in place to ensure that the resolved path is a subfolder of the input directory (or maaybe output/temp depending on node). I lean slightly towards keeping path so that people who frequently utilize the feature can have better organization of uploads, but I'll keep an eye on it.

src/views/LinearView.vue
AustinMroz: Followup from design is that the current behaviour is desirable, so I've removed the comment entirely.

src/renderer/extensions/linearMode/LinearControls.vue
AustinMroz: This one, I'm real conflicted on. Ideally, the drop operation will bubble through to the listener in `@/scripts/app` if no inputs can handle it, but you can't perform an await prior to event stop/prevent.

I've moved things over to a hybrid approach that postpones the stop/prevent as long as possible, but will still await the result to potentially allow other inputs to attempt. It's an objective improvement even though it's still incomplete.

pythongosssss (approved):
Would be nice to have a progress indicator in the dropzone while it is uploading, but this is  preexisting

AustinMroz:
I actually wrote some code for VHS to do upload progress bars.

Moving that code into core is still on my 'eventually' list.

PR #10128, opened 2026-03-17
🧙 feat: wire ComfyHub publish wizard with profile gate, asset upload, and submission

DrJKL:
## Summary 🎯

Wire the ComfyHub publish flow end-to-end: profile gate, multi-step wizard (describe, examples, finish), asset upload, and workflow submission via hub API.

> *A wizard of steps, from describe to the end,*
> *Upload your assets, your workflows you'll send!*
> *With tags neatly slugged and thumbnails in place,*
> *Your ComfyHub publish is ready to race!* 🏁

## Changes 🔧

- 🌐 **Hub Service** — `comfyHubService` for profile CRUD, presigned asset uploads, and workflow publish
- 📦 **Submission** — `useComfyHubPublishSubmission` orchestrates file uploads → publish in one flow
- 🧙 **Wizard Steps** — Describe (name/description/tags) → Examples (drag-drop reorderable images) → Thumbnail → Finish (profile card + private-asset warnings)
- 🖼️ **ReorderableExampleImage** — Drag-drop *and* keyboard reordering, accessible and fun
- 🏷️ **Tag Normalization** — `normalizeTags` slugifies before publishing
- 🔄 **Re-publish Prefill** — Fetches hub workflow metadata on re-publish, with in-memory cache fallback
- 📐 **Schema Split** — Publish-record schema separated from hub-workflow-metadata schema
- 🙈 **Unlisted Skip** — No hub-detail prefill fetch for unlisted records
- 👤 **Profile Gate** — Username validation in `useComfyHubProfileGate`
- 🧪 **Tests Galore** — New suites for DescribeStep, ExamplesStep, WizardContent, PublishSubmission, comfyHubService, normalizeTags, plus expanded PublishDialog & workflowShareService coverage

## Review Focus 🔍

> *Check the service, the schema, the Zod validation too,*
> *The upload orchestration — does it carry things through?*
> *The prefill fetch strategy: status → detail → cache,*
> *And drag-drop reordering — is it keeping its place?* 🤔

- 🌐 `comfyHubService.ts` — API contract shape, error handling, Zod parsing
- 📦 `useComfyHubPublishSubmission.ts` — Upload-then-publish flow, edge cases (no profile, no workflow)
- 🗂️ `ComfyHubPublishDialog.vue` — Prefill fetch strategy (publish status → hub detail → cache)
- 🖼️ `ReorderableExampleImage.vue` — Drag-drop + keyboard a11y

## Testing 🧪

```bash
pnpm test:unit -- src/platform/workflow/sharing/
pnpm typecheck
```

> *If the tests all turn green and the types all align,*
> *Then merge it on in — this publish flow's fine!* ✅


dante01yoon (commented):
adding nit and some concerns

src/platform/workflow/sharing/components/publish/ComfyHubPublishWizardContent.vue
```diff
@@ -53,7 +60,9 @@
         :is-first-step
         :is-last-step
         :is-publish-disabled
+        :is-publishing="isPublishInFlight"
```
dante01yoon, line 63: **Bug: dead event listener** — `@cancel="onCancel"` is bound here, but `ComfyHubPublishFooter.vue` never emits `cancel`. Its `defineEmits` only declares `back | next | publish`, and the template has no cancel button.

The commit message mentions "Add Cancel button to footer on all steps", but the actual button was dropped. The test stubs paper over this with a fake cancel button.

Either add the cancel button + emit to the footer, or remove this listener.

src/platform/workflow/sharing/composables/useComfyHubPublishSubmission.ts
```diff
@@ -0,0 +1,121 @@
...
+import { useWorkflowShareService } from '@/platform/workflow/sharing/services/workflowShareService'
+import type {
+  ComfyHubApiThumbnailType,
+  ComfyHubPublishFormData,
+  ThumbnailType
+} from '@/platform/workflow/sharing/types/comfyHubTypes'
+import { normalizeTags } from '@/platform/workflow/sharing/utils/normalizeTags'
+
+function mapThumbnailType(type: ThumbnailType): ComfyHubApiThumbnailType {
+  if (type === 'imageComparison') {
```
dante01yoon: **Duplicate: triple thumbnail mapping** — This `mapThumbnailType` function (`imageComparison` → `image_comparison`) is duplicated in:
- `comfyHubService.ts` as `normalizeThumbnailType()`
- `workflowShareService.ts` as `mapApiThumbnailType()`

The submission composable maps it here, then passes it to the service which maps it again (idempotent but wasteful). Pick one canonical location — the service layer seems right since it owns the API contract.

src/platform/workflow/sharing/composables/useComfyHubPublishSubmission.ts
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  return type
+}
+
+function getFileContentType(file: File): string {
+  return file.type || 'application/octet-stream'
+}
+
+function hasExampleContent(formData: ComfyHubPublishFormData): boolean {
```
dante01yoon: **Nit:** `hasExampleContent()` is trivially inlinable — `formData.exampleImages.length > 0` is clear enough at the call site. YAGNI.

src/platform/workflow/sharing/composables/useComfyHubProfileGate.ts
```diff
@@ -15,14 +15,43 @@ const profile = ref<ComfyHubProfile | null>(null)
 const cachedUserId = ref<string | null>(null)
 let inflightFetch: Promise<ComfyHubProfile | null> | null = null
 
-function mapHubProfileResponse(payload: unknown): ComfyHubProfile | null {
-  const result = zHubProfileResponse.safeParse(payload)
-  return result.success ? result.data : null
+function getCurrentWorkspaceId(): string {
```
dante01yoon, line 18: **Design: fragile sessionStorage parsing** — This hand-rolled JSON parse + shape check is brittle. If the workspace data shape changes, this fails silently or throws a generic error.

Consider:
1. A Zod schema (consistent with the rest of the codebase)
2. Or checking if the workspace store already exposes the current workspace ID

src/platform/workflow/sharing/components/publish/ReorderableExampleImage.vue
```diff
@@ -0,0 +1,159 @@
...
+  remove: [id: string]
+  move: [id: string, direction: number]
+  insertFiles: [index: number, files: FileList]
+}>()
+
+function focusVisible(el: HTMLElement) {
+  el.focus({ focusVisible: true } as FocusOptions)
+}
+
+async function handleArrowKey(direction: number, event: KeyboardEvent) {
```
dante01yoon, line 77: **Nit:** `focusVisible` is not part of the standard `FocusOptions` type (hence the `as FocusOptions` cast). It works in Chromium/Electron but the cast silences the type system. Worth a brief comment explaining why.

src/platform/workflow/sharing/schemas/shareSchemas.ts
```diff
@@ -36,9 +45,28 @@ export const zHubProfileResponse = z.preprocess((data) => {
   const d = data as Record<string, unknown>
   return {
     username: d.username,
-    name: d.name,
+    name: d.name ?? d.display_name,
```
dante01yoon, line 48: **Nit: unsafe preprocess cast** — `const d = data as Record<string, unknown>` has no null guard. If `data` is `null` or a non-object primitive, `d.username` will throw at runtime before Zod can catch it.

Add a guard:
```ts
if (!data || typeof data !== 'object') return data
const d = data as Record<string, unknown>
```

dante01yoon (commented):
adding additional comments

src/platform/workflow/sharing/composables/useComfyHubPublishWizard.ts
```diff
@@ -13,22 +19,50 @@ const PUBLISH_STEPS = [
 
 export type ComfyHubPublishStep = (typeof PUBLISH_STEPS)[number]
 
+const cachedPrefills = new Map<string, PublishPrefill>()
```
dante01yoon, line 23: nit: module-level mutable Map outside of Pinia

```ts
const cachedPrefills = new Map<string, PublishPrefill>()
```

This is a module-level singleton cache that:
- Persists across dialog open/close cycles (good)
- Is lost on page refresh (expected, but worth documenting)
- Lives outside Pinia, making it invisible to devtools and inconsistent with the state management approach

The existing profile gate code has the same pattern with a TODO on line ~7 saying *"Migrate to a Pinia store for proper singleton state management"*. This new cache should follow the same migration path.

Not blocking, but consider adding a similar TODO comment so the two get migrated together.

src/platform/workflow/sharing/components/publish/ReorderableExampleImage.vue
dante01yoon, line 77: nit: non-standard `focusVisible` option needs a comment

```ts
el.focus({ focusVisible: true } as FocusOptions)
```

The `focusVisible` property is a recent addition to the spec (Chromium 122+). It's not part of the TypeScript `FocusOptions` type, which is why the `as FocusOptions` cast is needed to silence the compiler.

This works fine in Electron (Chromium-based), but the cast hides a potential issue if this component is ever used in a non-Chromium context. A brief comment explaining the cast would help future readers:

```ts
// focusVisible is a Chromium 122+ extension to FocusOptions
// (not yet in TypeScript's lib.dom.d.ts)
el.focus({ focusVisible: true } as FocusOptions)
```

src/platform/workflow/sharing/schemas/shareSchemas.ts
dante01yoon, line 48: nit: unsafe `as` cast without null guard

```ts
export const zHubProfileResponse = z.preprocess((data) => {
  const d = data as Record<string, unknown>  // <-- no null check
  return {
    username: d.username,
    ...
  }
}, zComfyHubProfile)
```

If `data` is `null`, `undefined`, or a primitive (e.g., a number or string), `d.username` will throw a `TypeError` at runtime (`Cannot read properties of null`). This crashes before Zod can catch and report the validation error.

**Fix:**
```ts
export const zHubProfileResponse = z.preprocess((data) => {
  if (!data || typeof data !== 'object') return data  // let Zod handle invalid input
  const d = data as Record<string, unknown>
  return { ... }
}, zComfyHubProfile)
```

This way, non-object inputs fall through to Zod's object validator, which returns a proper validation error instead of a runtime crash.

dante01yoon (changes requested):
- issue: if issue happens after upload,  asset step is empty 

https://github.com/user-attachments/assets/7ec5192b-fc09-48cc-a02a-67760aa750c1

christian-byrne (commented):
From coderabbit test check:

> PR includes bug-fix work (blob URL revocation, Pinia reactivity, prefill race/overwrite fixes) but lacks browser_tests modifications and explanation for missing e2e regression coverage.

src/platform/workflow/sharing/composables/useComfyHubProfileGate.ts
```diff
@@ -38,14 +67,7 @@ export function useComfyHubProfileGate() {
...
     try {
-      const response = await api.fetchApi('/hub/profile')
-      if (!response.ok) {
-        hasProfile.value = false
-        profile.value = null
-        return null
-      }
-
-      const nextProfile = mapHubProfileResponse(await response.json())
+      const nextProfile = await getMyProfile()
```
christian-byrne, line 70: issue: When `getMyProfile()` throws (network error, 500), the catch block (line 79) returns `null` but does not set `hasProfile.value = false`. Since `isProfileLoading` is computed as `hasProfile.value === null || isFetchingProfile.value`, after the error it stays `true` permanently -- the user sees an infinite loading skeleton with no way to retry.

Add `hasProfile.value = false` in the catch block so the profile-creation form appears instead of an infinite spinner.

src/platform/workflow/sharing/composables/useComfyHubPublishSubmission.ts
```diff
@@ -0,0 +1,117 @@
...
+    const assetIds = getAssetIds(
+      await workflowShareService.getShareableAssets()
+    )
+
+    const thumbnailTokenOrUrl =
+      formData.thumbnailFile && formData.thumbnailType !== 'imageComparison'
+        ? await uploadFileAndGetToken(formData.thumbnailFile)
+        : formData.comparisonBeforeFile
+          ? await uploadFileAndGetToken(formData.comparisonBeforeFile)
+          : undefined
```
christian-byrne: suggestion: The nested ternary for `thumbnailTokenOrUrl` is fragile. If `thumbnailType` is `'image'` but `thumbnailFile` is null while `comparisonBeforeFile` exists (e.g. from prefill), the fallthrough sends the wrong file as the thumbnail. Even if ThumbnailStep currently clears files on type switch, the logic here is independently incorrect.

Consider replacing with an explicit helper:
```ts
function resolveThumbnailFile(
  formData: ComfyHubPublishFormData
): File | undefined {
  if (formData.thumbnailType === 'imageComparison') {
    return formData.comparisonBeforeFile ?? undefined
  }
  return formData.thumbnailFile ?? undefined
}
```
Then: `const thumbnailTokenOrUrl = resolveThumbnailFile(formData) ? await uploadFileAndGetToken(resolveThumbnailFile(formData)!) : undefined`

src/platform/workflow/sharing/components/publish/ComfyHubPublishDialog.vue
```diff
@@ -59,8 +121,72 @@ const {
...
+  () => workflowStore.activeWorkflow?.isTemporary ?? false
+)
+
+function checkNeedsSave() {
+  const workflow = workflowStore.activeWorkflow
+  needsSave.value = !workflow || workflow.isTemporary || workflow.isModified
+  if (workflow) {
+    workflowName.value = workflow.filename.replace(/\.json$/i, '')
+  }
+}
```
christian-byrne, line 142: suggestion: The save-before-proceed flow (`checkNeedsSave`, `handleSave`, `buildWorkflowPath`, name input + save button template) is implemented 3 times with near-identical logic here and in `ShareWorkflowDialogContent.vue` (two tabs). Future changes must be replicated in all 3 locations.

Consider extracting a `useWorkflowSaveGate()` composable + shared `<WorkflowSavePrompt>` component.

src/platform/workflow/sharing/components/publish/ComfyHubPublishWizardContent.vue
```diff
@@ -93,35 +104,48 @@ const {
...
 const isResolvingPublishAccess = ref(false)
+const isPublishInFlight = computed(
+  () => isPublishing || isResolvingPublishAccess.value
+)
 const isPublishDisabled = computed(
-  () => flags.comfyHubProfileGateEnabled && hasProfile.value !== true
+  () =>
+    isPublishInFlight.value ||
+    (flags.comfyHubProfileGateEnabled && hasProfile.value !== true) ||
+    (finishStepRef.value !== null && !finishStepRef.value.isReady)
```
christian-byrne: suggestion: `finishStepRef.value.isReady` accessed via template ref + `defineExpose` breaks unidirectional data flow (props down, events up). When FinishStep isn't rendered (different wizard step), `finishStepRef.value` is null, requiring the fragile null-check on line 134.

Consider using `v-model:ready` instead. In FinishStep:
```ts
const ready = defineModel<boolean>('ready', { default: false })
watch(isReady, (val) => { ready.value = val }, { immediate: true })
```
Then remove `defineExpose({ isReady })`. In WizardContent:
```vue
<ComfyHubFinishStep
  v-else-if="currentStep === 'finish' && hasProfile && profile"
  v-model:ready="finishStepReady"
  v-model:acknowledged="assetsAcknowledged"
  :profile
/>
```
```ts
const finishStepReady = ref(false)
// Replace finishStepRef null-check in isPublishDisabled:
const isPublishDisabled = computed(() =>
  isPublishInFlight.value ||
  (flags.comfyHubProfileGateEnabled && hasProfile.value !== true) ||
  !finishStepReady.value
)
```

src/platform/workflow/sharing/components/publish/ComfyHubPublishDialog.vue
```diff
@@ -75,18 +201,59 @@ function handleRequireProfile() {
...
+  try {
+    await submitToComfyHub(formData.value)
+    const path = workflowStore.activeWorkflow?.path
+    if (path) {
+      cachePublishPrefill(path, formData.value)
+    }
+    onClose()
+  } finally {
+    isPublishing.value = false
+  }
```
christian-byrne, line 226: issue: `handlePublish` has `try/finally` but no `catch`. Errors from `submitToComfyHub` propagate as unhandled promise rejections. The `finally` block resets `isPublishing`, but the user gets no feedback about what went wrong.

Add a catch block that shows a toast error and resets state -- the pattern already exists in the save handler above (line 180).

christian-byrne (commented):
Nice work wiring the full publish flow end-to-end! One thing worth considering: this PR introduces a significant new user journey (publish wizard open -> save -> upload assets -> submit) but has no `useTelemetry()` instrumentation. The codebase already has `trackShareFlow` for the share dialog -- adding similar events for the publish flow (wizard opened, step transitions, publish succeeded/failed) would help measure adoption and debug issues in production.

src/platform/workflow/sharing/components/publish/ComfyHubExamplesStep.vue
```diff
@@ -40,96 +44,149 @@
...
+    const replacedImages = updated.splice(index, newImages.length, ...newImages)
+    for (const img of replacedImages) {
+      if (img.file) URL.revokeObjectURL(img.url)
+    }
+  } else {
+    updated.splice(index, 0, ...newImages)
+  }
+
+  exampleImages.value = updated
+}
```
christian-byrne, line 193: nitpick (non-blocking): `insertImagesAt` silently replaces existing images when at max capacity, but the name implies insertion only. The `maxInsert` calculation with the `remaining <= 0` branch adds non-obvious dual behavior. Consider splitting into `replaceImagesAt` / `insertNewImagesAt` or adding a comment explaining the intent.

src/platform/workflow/sharing/components/publish/ComfyHubPublishDialog.vue
```diff
@@ -75,18 +201,59 @@ function handleRequireProfile() {
...
+      : getCachedPrefill(path)
+    if (prefill) {
+      applyPrefill(prefill)
+    }
+  } catch {
+    const cached = getCachedPrefill(path)
+    if (cached) {
+      applyPrefill(cached)
+    }
+  }
```
christian-byrne: nitpick (non-blocking): `fetchPublishPrefill` swallows errors silently (catch block returns undefined with no logging). A network failure during prefill fetch is completely invisible, making debugging harder in production.

Consider adding `console.warn('Failed to fetch publish prefill:', error)` in the catch. The codebase already uses this pattern alongside toast notifications. If PostHog error tracking is available, tagging this with a breadcrumb would also help triage.

src/platform/workflow/sharing/composables/useComfyHubPublishSubmission.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  async function submitToComfyHub(
+    formData: ComfyHubPublishFormData
+  ): Promise<void> {
+    const username = getUsername(profile.value)
+    const workflowFilename = getWorkflowFilename(
+      workflowStore.activeWorkflow?.path
+    )
+    const assetIds = getAssetIds(
+      await workflowShareService.getShareableAssets()
```
christian-byrne, line 74: nitpick (non-blocking): `getShareableAssets()` (which runs `graphToPrompt` + API call) is invoked on FinishStep mount for asset warnings and again inside `submitToComfyHub` on publish click. Consider passing the already-fetched result through to avoid the redundant call.

dante01yoon: addressed in #10377

src/platform/workflow/sharing/components/publish/ComfyHubPublishDialog.vue
dante01yoon: addressed in #10377

src/platform/workflow/sharing/components/publish/ComfyHubPublishDialog.vue
dante01yoon: addressed in #10377

src/platform/workflow/sharing/components/publish/ComfyHubPublishWizardContent.vue
dante01yoon: addressed in #10377

src/platform/workflow/sharing/components/publish/ComfyHubPublishDialog.vue
christian-byrne: suggestion (non-blocking): agreed, the save gate logic is duplicated across publish and share tabs. worth extracting into a composable + component.

src/platform/workflow/sharing/utils/normalizeTags.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Normalizes a tag to its slug form for the ComfyHub API.
+ * Converts display names like "Text to Image" to "text-to-image".
+ */
+export function normalizeTag(tag: string): string {
+  return tag.trim().toLowerCase().replace(/\s+/g, '-')
+}
```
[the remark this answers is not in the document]
christian-byrne: note: the tags in comfyHubTags.ts are display names (e.g. "Text to Image"), not API identifiers. normalizeTag correctly converts them to slug form ("text-to-image") before API submission. The backend expects slugified tags, so this transformation is intentional and correct.

src/platform/workflow/sharing/components/publish/ComfyHubFinishStep.vue
```diff
@@ -0,0 +1,86 @@
...
+          class="flex size-12 shrink-0 items-center justify-center overflow-hidden rounded-full bg-linear-to-b from-green-600/50 to-green-900"
+        >
+          <img
+            v-if="profile.profilePictureUrl"
+            :src="profile.profilePictureUrl"
+            :alt="profile.username"
+            class="size-full rounded-full object-cover"
+          />
+          <span v-else class="text-base text-white">
+            {{ (profile.name ?? profile.username)[0].toUpperCase() }}
```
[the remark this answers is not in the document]
christian-byrne: nit (non-blocking): valid point - .charAt(0) is safer than [0] for potentially empty strings. username is z.string() which does allow empty.

src/platform/workflow/sharing/components/profile/ComfyHubCreateProfileForm.vue
```diff
@@ -77,6 +77,12 @@
             </span>
             <Input id="profile-username" v-model="username" class="pl-7" />
           </div>
+          <p
+            v-if="showUsernameError"
+            class="text-xs text-destructive-background"
+          >
+            {{ $t('comfyHubProfile.usernameError') }}
+          </p>
```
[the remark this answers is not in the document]
christian-byrne: nit (non-blocking): good a11y suggestion - aria-invalid and aria-describedby should be wired to the username error.

src/platform/workflow/sharing/components/publish/ComfyHubFinishStep.vue
```diff
@@ -0,0 +1,86 @@
...
+const { state: privateAssets, isLoading: isLoadingAssets } = useAsyncState(
+  () => shareService.getShareableAssets(),
+  []
+)
+
+const hasPrivateAssets = computed(() => privateAssets.value.length > 0)
+const isReady = computed(
+  () =>
+    !isLoadingAssets.value && (!hasPrivateAssets.value || acknowledged.value)
+)
```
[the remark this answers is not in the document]
christian-byrne: suggestion (non-blocking): valid concern - if getShareableAssets() fails, isReady could evaluate true. Adding error tracking from useAsyncState would prevent publish progression on failure.

src/platform/workflow/sharing/services/comfyHubService.ts
```diff
@@ -0,0 +1,211 @@
...
+      headers: {
+        'Content-Type': input.contentType
+      },
+      body: input.file
+    })
+
+    if (!response.ok) {
+      throw new Error('Failed to upload file')
+    }
+  }
```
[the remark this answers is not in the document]
christian-byrne: suggestion (non-blocking): agreed, the error message should parse the response body using parseErrorMessage for actionable feedback, matching the pattern in getMyProfile/createProfile.

src/platform/workflow/sharing/components/publish/ComfyHubExamplesStep.test.ts
```diff
@@ -0,0 +1,95 @@
...
+  })
+
+  it('does not emit when moving last image right (boundary)', async () => {
+    const wrapper = mountStep(createImages(3))
+
+    const tiles = wrapper.findAll('[role="listitem"]')
+    await tiles[2].trigger('keydown', { key: 'ArrowRight' })
+
+    expect(wrapper.emitted('update:exampleImages')).toBeFalsy()
+  })
```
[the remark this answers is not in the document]
christian-byrne: note: commit 2cf8e64 fixed the reorder tests but missed these boundary tests. The boundary tests pass without shiftKey because plain Arrow doesn'\''t enter the reorder path at all - so they'\''re testing focus movement boundary, not reorder boundary. Still worth adding shiftKey to properly test the reorder boundary case.

src/platform/workflow/sharing/components/publish/ComfyHubExamplesStep.vue
```diff
@@ -40,96 +44,149 @@
...
+  exampleImages.value = exampleImages.value.filter((img) => img.id !== id)
 }
 
-function addImages(files: FileList) {
-  const newImages: ExampleImage[] = Array.from(files)
+function createExampleImages(files: FileList): ExampleImage[] {
+  return Array.from(files)
     .filter((f) => f.type.startsWith('image/'))
     .filter((f) => !isFileTooLarge(f, MAX_IMAGE_SIZE_MB))
     .map((file) => ({
```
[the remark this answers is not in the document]
christian-byrne: suggestion (non-blocking): agreed, user feedback for rejected files would improve UX. Not blocking for initial launch.

src/platform/workflow/sharing/components/publish/ComfyHubExamplesStep.vue
```diff
@@ -40,96 +44,149 @@
...
+
+  if (newImages.length === 0) return
+  if (remaining <= 0) {
+    const replacedImages = updated.splice(index, newImages.length, ...newImages)
+    for (const img of replacedImages) {
+      if (img.file) URL.revokeObjectURL(img.url)
+    }
+  } else {
+    updated.splice(index, 0, ...newImages)
+  }
```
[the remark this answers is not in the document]
christian-byrne: nit (non-blocking): index is always derived from the tile list position (drag target or keyboard-focused element), so negative values aren'\''t reachable in practice. Adding a clamp would be purely defensive.

dante01yoon (approved):

PR #10131, opened 2026-03-17
fix: show all outputs in FormDropdown for multi-output jobs

synap5e:
## Summary

FormDropdown Outputs tab only showed the first output for multi-output jobs because the Jobs API `/jobs` returns a single `preview_output` per job.

## Changes

- **What**: When history assets include jobs with `outputs_count > 1`, lazily fetch full outputs via `getJobDetail` (cached in `jobOutputCache`) and expand them into individual dropdown items. Single-output jobs are unaffected. Added in-flight guard to prevent duplicate fetches.
- This is a consumer-side workaround in `WidgetSelectDropdown.vue` that becomes a no-op once the backend returns all outputs in the list response (planned Assets API migration).

## Review Focus

- The `resolvedMultiOutputs` shallowRef + watch pattern for async data feeding into a computed. Each `getJobDetail` call is cached by `jobOutputCache` LRU, so no redundant network requests.
- This fix is intentionally temporary — it will be superseded when OSS/cloud both return full outputs from list endpoints.

## No E2E test

E2E coverage is impractical here: reproducing requires a running ComfyUI backend executing a workflow that produces multiple outputs, then inspecting the FormDropdown's Outputs tab. The unit test covers the lazy-loading logic with mocked `getJobDetail` responses.


src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -153,24 +158,106 @@ function assetKindToMediaType(kind: AssetKind): string {
   return kind === 'mesh' ? '3D' : kind
 }
 
+/**
+ * Cache of fully-resolved outputs for multi-output jobs, keyed by jobId.
+ * Populated lazily when history assets include jobs with outputs_count > 1.
+ */
+const resolvedMultiOutputs = shallowRef(new Map<string, ResultItemImpl[]>())
+const pendingJobIds = new Set<string>()
```
christian-byrne: Right now, we’re effectively re-implementing multi-output resolution logic that already exists in the service/utility layer:
- Deduplication via `pendingJobIds`
- Caching via `resolvedMultiOutputs`
- The `shouldLoadFullOutputs` check
- Calling `getJobDetail` + `flattenNodeOutput`

Existing implementations already cover this:

- `resolveOutputAssetItems` (`src/platform/assets/utils/outputAssetUtil.ts`)
  - Handles `shouldLoadFullOutputs(outputCount, allOutputs.length)`
  - Calls `getJobDetail`
  - Uses `getPreviewableOutputsFromJobDetail` with proper fallback
- `useOutputHistory.ts`
  - Demonstrates the composable-based async pattern using `useAsyncState`
  - Includes a `resolvedCache` Map for caching
- `getJobDetail`
  - Already has a 50-entry LRU cache in `jobOutputCache.ts`
  - Makes the component-level `resolvedMultiOutputs` Map both redundant and unbounded

Could this component reuse `resolveOutputAssetItems`, or alternatively extract a shared composable (following the `useOutputHistory` pattern), instead of re-implementing a parallel cache and deduplication layer within the UI component?

christian-byrne (changes requested):

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.test.ts
```diff
@@ -484,6 +484,148 @@ describe('WidgetSelectDropdown cloud asset mode (COM-14333)', () => {
...
+      hasMore: ref(false),
+      isLoadingMore: ref(false)
+    },
+    mockGetJobDetail: vi.fn()
+  }
+})
+
+vi.mock('@/platform/assets/composables/media/useMediaAssets', () => ({
+  useMediaAssets: () => mockMediaAssets
+}))
```
christian-byrne: question: `vi.mock` calls are hoisted by Vitest to file scope, so this `useMediaAssets` mock replaces the module for all describe blocks in the file -- including the earlier tests (lines 254-269, 310-323) that access `outputItems` and previously used the real/auto-mocked implementation. Were those earlier tests verified to still pass correctly with this mock active? If so, it may be worth scoping the mock (e.g., via `vi.doMock` inside a `describe`) to avoid fragile coupling.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -153,24 +158,106 @@ function assetKindToMediaType(kind: AssetKind): string {
...
+      .then((detail) => {
+        if (!detail?.outputs) return
+        const items = Object.entries(detail.outputs).flatMap(flattenNodeOutput)
+        const updated = new Map(resolvedMultiOutputs.value)
+        updated.set(meta.jobId, items)
+        resolvedMultiOutputs.value = updated
+      })
+      .finally(() => {
+        pendingJobIds.delete(meta.jobId)
+      })
```
christian-byrne: question: When `getJobDetail` resolves with `undefined` (network error, 404, etc.), the job is removed from `pendingJobIds` in `.finally()` but never stored in `resolvedMultiOutputs`. If the media list doesn't change afterward (no new watch trigger), the job won't be retried and the user sees only partial outputs with no indication of failure. Is that the intended degradation behavior, or would a retry / user-visible indicator be worth adding?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -153,24 +158,106 @@ function assetKindToMediaType(kind: AssetKind): string {
...
+        continue
+      const path = output.subfolder
+        ? `${output.subfolder}/${output.filename}`
+        : output.filename
+      const annotatedPath = `${path} [output]`
+      if (seen.has(annotatedPath)) continue
+      seen.add(annotatedPath)
+      items.push({
+        id: `output-${annotatedPath}`,
+        preview_url: getMediaUrl(path, 'output'),
```
christian-byrne: question: For resolved multi-output items, `preview_url` is always set to `getMediaUrl(path, 'output')`, while the fallback path (line 235) preserves `asset.preview_url || getMediaUrl(...)`. If the API provides optimized CDN thumbnails via `preview_url`, they'd be lost for expanded multi-output items. Is that intentional (the resolved outputs don't carry their own `preview_url`), or should this mirror the fallback logic?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.test.ts
```diff
@@ -484,6 +484,148 @@ describe('WidgetSelectDropdown cloud asset mode (COM-14333)', () => {
...
+
+vi.mock('@/services/jobOutputCache', () => ({
+  getJobDetail: (...args: unknown[]) => mockGetJobDetail(...args)
+}))
+
+describe('WidgetSelectDropdown multi-output jobs', () => {
```
christian-byrne, line 514: suggestion (non-blocking): The happy path is well tested, but there's no coverage for error/edge cases: `getJobDetail` rejecting, returning `undefined`, returning a job with empty `outputs`, or multiple assets sharing the same `jobId`. These would help document the silent-degradation behavior and guard against regressions.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
synap5e: Good call, resolved

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
synap5e: Added a .catch() for this

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
synap5e: Updated to mirror

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.test.ts
synap5e: Added tests for rejection and multiple assets sharing a jobid. undefined and [] are handled and IMO trivial enough

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.test.ts
synap5e: Verified to pass, hoisted

synap5e:
- Replaced whole-list `shallowRef<FormDropdownItem[]>` with per-job `Map<string, AssetItem[]>` so concurrent resolutions don't overwrite each other
- Removed duplicated `hasAllOutputs` / `shouldLoadFullOutputs` logic — now delegates entirely to `resolveOutputAssetItems`
- Added `.catch()` for error handling with preview fallback
- Added `onCleanup` cancellation guard + `pendingJobIds.clear()` to prevent stale writes and stuck pending jobs
- Moved file-scoped mocks to top of test file alongside other mocks
- Added tests: two concurrent jobs, complete `allOutputs` metadata, and resolver rejection fallback

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -153,24 +156,84 @@ function assetKindToMediaType(kind: AssetKind): string {
...
+          const next = new Map(resolvedByJobId.value)
+          next.set(meta.jobId, resolved)
+          resolvedByJobId.value = next
+        })
+        .catch((error) => {
+          console.warn('Failed to resolve multi-output job', meta.jobId, error)
+        })
+        .finally(() => {
+          pendingJobIds.delete(meta.jobId)
+        })
```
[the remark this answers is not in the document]
synap5e: Resolved. The pendingJobIds.clear() at line 174 was specifically added to fix this. Here's the sequence:

Watch 1 fires, dispatches job-X, adds to pendingJobIds
Media changes → onCleanup sets cancelled = true for watch 1
Watch 2 fires, pendingJobIds.clear() runs, then dispatches job-X again
Watch 1's .then() → sees cancelled, skips write
Watch 2's .then() → not cancelled, writes to resolvedByJobId ✅

Without the .clear(), CodeRabbit's concern would be valid — job-X would be stuck in pendingJobIds and never retried. The .clear() ensures the new watch invocation can re-dispatch any jobs that were in-flight when the previous one was cancelled.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -153,24 +156,82 @@ function assetKindToMediaType(kind: AssetKind): string {
...
+watch(
+  () => outputMediaAssets.media.value,
+  (_, __, onCleanup) => {
+    let cancelled = false
+    onCleanup(() => {
+      cancelled = true
+    })
+    pendingJobIds.clear()
+
+    for (const asset of outputMediaAssets.media.value) {
```
DrJKL: Why re-query it here instead of accepting the value in the callback?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -153,24 +156,82 @@ function assetKindToMediaType(kind: AssetKind): string {
...
+ * Per-job cache of resolved outputs for multi-output jobs.
+ * Keyed by jobId, populated lazily via resolveOutputAssetItems which
+ * fetches full outputs through getJobDetail (itself LRU-cached).
+ */
+const resolvedByJobId = shallowRef(new Map<string, AssetItem[]>())
+const pendingJobIds = new Set<string>()
+
+watch(
+  () => outputMediaAssets.media.value,
+  (_, __, onCleanup) => {
```
DrJKL: Any chance this could be converted to use https://vueuse.org/core/computedAsync/?

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -153,24 +156,82 @@ function assetKindToMediaType(kind: AssetKind): string {
...
 const outputItems = computed<FormDropdownItem[]>(() => {
   if (!['image', 'video', 'audio', 'mesh'].includes(props.assetKind ?? ''))
     return []
 
   const targetMediaType = assetKindToMediaType(props.assetKind!)
-  const outputFiles = outputMediaAssets.media.value.filter(
-    (asset) => getMediaTypeFromFilename(asset.name) === targetMediaType
-  )
+  const seen = new Set<string>()
+  const items: FormDropdownItem[] = []
```
DrJKL, line 213: But it was so nice and functional before 😢

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -153,24 +156,82 @@ function assetKindToMediaType(kind: AssetKind): string {
   return kind === 'mesh' ? '3D' : kind
 }
 
+/**
+ * Per-job cache of resolved outputs for multi-output jobs.
+ * Keyed by jobId, populated lazily via resolveOutputAssetItems which
+ * fetches full outputs through getJobDetail (itself LRU-cached).
+ */
+const resolvedByJobId = shallowRef(new Map<string, AssetItem[]>())
+const pendingJobIds = new Set<string>()
```
DrJKL, line 165: This is a lot of things to keep in the air at once.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
synap5e: Fixed

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
synap5e: Not AFAICT - computedAsnc is for single, we need n parallel jobs independently. I've cleaned this up a bit though.
- Replaced a single shallowRef<FormDropdownItem[]> (whole-list replacement) with shallowRef(new Map<string, AssetItem[]>()) keyed by jobId
- Added onCleanup cancellation guard + pendingJobIds.clear() to handle watch reruns
- Added .catch() for error handling
- Removed duplicated hasAllOutputs check — now delegates entirely to resolveOutputAssetItems

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
synap5e: Given a go at improving this. To be completely honest, my unfamiliarity with FE is showing in this PR 😅

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
DrJKL: @coderabbitai Open an issue to refactor this into a clean pipeline.

src/renderer/extensions/vueNodes/widgets/components/WidgetSelectDropdown.vue
```diff
@@ -153,24 +156,82 @@ function assetKindToMediaType(kind: AssetKind): string {
...
+
+watch(
+  () => outputMediaAssets.media.value,
+  (assets, _, onCleanup) => {
+    let cancelled = false
+    onCleanup(() => {
+      cancelled = true
+    })
+    pendingJobIds.clear()
```
DrJKL, line 174: @coderabbitai Open an issue to try to simplify this using https://vueuse.org/core/computedAsync/

DrJKL (approved):

DrJKL:
@coderabbitai Open an issue to add support in the devtools extension that would allow us to test this in Playwright.

PR #10135, opened 2026-03-17
feat: enhance error tab with full error report and GitHub search

jaeone94:
## Summary

Add full system error report generation, GitHub issue search, Get Help support action, and dedicated runtime error layout to the Errors Tab panel.

## Changes

- **What**: 
  - Runtime errors now display a full diagnostic report (system info, GPU/VRAM, server logs, workflow JSON) in the details section, matching the legacy error dialog's "Show Report" functionality
  - Added "Find on Github" button that searches GitHub issues by error message
  - Added "Get Help" button that executes `Comfy.ContactSupport` command with `trackHelpResourceClicked` telemetry, matching the legacy error dialog's "Help Fix This" flow
  - Added "Copy" button that copies the full visible report content
  - Runtime errors render in a dedicated full-height flex panel outside the accordion for better space utilization; when coexisting with other error types, details are capped at `max-h-[15lh]`
  - Added `nodeTitle` fallback to `card.title` when display name is unavailable
  - Extracted report generation into `useErrorReport` composable
  - `exceptionType` is now passed from the upstream `execution_error` WebSocket event, with a named constant fallback

## Review Focus

- Runtime errors are rendered outside the accordion (`v-if`/`v-else` in TabErrors) to enable a full-height flex layout. The accordion lives inside a scrollable container (`overflow-y-auto`) where `flex-1` cannot constrain child height — hence the dedicated rendering path. When runtime errors coexist with other error types, they fall back to the accordion with `max-h-[15lh]`, controlled via the `fullHeight` prop.
- The composable accepts `MaybeRefOrGetter<ErrorCardData>` for reactivity safety, though in practice components re-key on card change.

## Test Coverage

- Unit tests: 12 tests for `ErrorNodeCard` (report generation, exceptionType passthrough/fallback, copy, fallbacks, GitHub search, Get Help) + 7 tests for `TabErrors`
- E2E test: Verifies "Find on Github" and "Copy" buttons are visible in the Errors Tab after triggering an execution error
- **Note**: E2E coverage for the dedicated runtime error panel (`singleRuntimeErrorCard` path) is not included because runtime execution errors (`lastExecutionError`) are triggered via WebSocket `execution_error` events from the backend, which cannot be reliably simulated in the Playwright test environment. The existing `nodes/execution_error` test workflow produces validation errors (`lastNodeErrors`), not WebSocket runtime errors.

https://github.com/user-attachments/assets/589c5238-490f-47fb-835d-5c45bac5928a


src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -264,6 +285,22 @@ const {
...
 } = useErrorGroups(searchQuery, t)
 
+const singleRuntimeErrorGroup = computed(() => {
+  if (filteredGroups.value.length !== 1) return null
+  const group = filteredGroups.value[0]
+  if (
+    group.type !== 'execution' ||
+    group.cards.length !== 1 ||
+    !group.cards[0].errors.some((e) => e.isRuntimeError)
+  )
```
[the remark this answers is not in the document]
jaeone94: This is not applicable. Runtime errors and validation errors cannot coexist in the same `ErrorCardData`. They are produced by separate processing paths in `useErrorGroups.ts`:
- `processExecutionError()` creates cards with a single `isRuntimeError: true` error item
- `processNodeErrors()` creates cards with validation errors (no `isRuntimeError` flag)

These two functions produce cards keyed by `nodeId` with different `idPrefix` values (`'exec'` vs `'node'`), so they never merge into the same card. Changing to `every()` would be more restrictive without addressing a real scenario, and could prevent the feature from activating in unexpected edge cases.

src/components/rightSidePanel/errors/useErrorReport.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  const displayedDetailsMap = computed(() => {
+    const card = toValue(cardSource)
+    return Object.fromEntries(
+      card.errors.map((error, idx) => [
+        idx,
+        enrichedDetails[idx] ?? error.details
+      ])
+    )
+  })
```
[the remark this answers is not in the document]
jaeone94: This concern is already addressed by Vue's re-keying pattern. `ErrorNodeCard` is rendered with `:key="card.id"` in both `TabErrors.vue` (line 27) and the accordion path (line 155). When the card changes, Vue destroys the old component instance and creates a new one, which triggers a fresh `onMounted`. The `enrichedDetails` reactive object is per-instance and is naturally garbage collected on unmount.

Additionally, the composable accepts `MaybeRefOrGetter<ErrorCardData>` and calls `toValue()` inside the `displayedDetailsMap` computed, so the computed itself is reactive to card changes. Switching to `watch` would cause unnecessary API calls on every prop change, which is undesirable.

src/components/rightSidePanel/errors/useErrorReport.ts
```diff
@@ -0,0 +1,94 @@
...
+    const workflow = app.rootGraph.serialize()
+
+    for (const { error, idx } of uncached) {
+      try {
+        const report = generateErrorReport({
+          exceptionType: card.title,
+          exceptionMessage: error.message,
+          traceback: error.details,
+          nodeId: card.nodeId,
+          nodeType: card.title,
```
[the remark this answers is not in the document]
jaeone94: This was an intentional design decision. Parsing `error.message` by splitting on `': '` is fragile because:
- The colon-space delimiter can appear in the exception message itself (e.g., `ValueError: Expected shape: [1, 3, 512, 512]`)
- Different Python exception formatters may use different delimiters

The full exception type + message is preserved in `exceptionMessage` (e.g., `"RuntimeError: CUDA out of memory"`), so no information is lost in the report. Using `card.title` (the node class type) as `exceptionType` is a safe fallback that avoids parsing-related bugs. If more precise `exceptionType` data is needed in the future, it should be added as a separate field on `ErrorItem` upstream in `useErrorGroups.ts`, not parsed from a formatted string.

src/components/rightSidePanel/errors/useErrorReport.ts
```diff
@@ -0,0 +1,94 @@
...
+      await systemStatsStore.refetchSystemStats()
+    }
+    if (!systemStatsStore.systemStats) return
+
+    let logs: string
+    try {
+      logs = await api.getLogs()
+    } catch {
+      return
+    }
```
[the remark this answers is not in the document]
jaeone94: addressed in [6678843](https://github.com/Comfy-Org/ComfyUI_frontend/pull/10135/commits/667884315ad6302e2132cca26f965dfa216fa0fc)

src/components/rightSidePanel/errors/useErrorReport.ts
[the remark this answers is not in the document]
jaeone94: you are right, that addressed in [d9bec02](https://github.com/Comfy-Org/ComfyUI_frontend/pull/10135/commits/d9bec025bfca5c83a3c8067fbb7d8b97f61780d5)

src/components/rightSidePanel/errors/ErrorNodeCard.test.ts
```diff
@@ -0,0 +1,286 @@
...
+  it('generates report with fallback logs when getLogs fails', async () => {
+    mockGetLogs.mockRejectedValue(new Error('Network error'))
+
+    const wrapper = mountCard(makeRuntimeErrorCard())
+    await flushPromises()
+
+    // Report is still generated with fallback log message
+    expect(mockGenerateErrorReport).toHaveBeenCalledOnce()
+    expect(wrapper.text()).toContain('ComfyUI Error Report')
+  })
```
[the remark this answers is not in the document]
jaeone94: An end-to-end (E2E) test to verify the presence of the GitHub button added in this PR has been included, and I noted in the PR summary that it is difficult to write tests for runtime (execution) errors because they depend on the backend.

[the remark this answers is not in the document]
jaeone94: @coderabbitai Could you please review the requested changes to the PR? Are there any further changes needed?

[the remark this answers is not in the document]
jaeone94: According to the contribution guide, I understand that translations for versions other than en are unnecessary and are automatically translated during the CI process.

src/components/rightSidePanel/errors/useErrorReport.ts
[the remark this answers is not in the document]
jaeone94: @coderabbitai yes please make issue and assign it to me

src/components/rightSidePanel/errors/useErrorReport.ts
```diff
@@ -0,0 +1,94 @@
...
+      logs = 'Failed to retrieve server logs'
+    }
+
+    const workflow = app.rootGraph.serialize()
+
+    for (const { error, idx } of uncached) {
+      try {
+        const report = generateErrorReport({
+          exceptionType: card.title,
+          exceptionMessage: error.message,
```
christian-byrne: issue: `card.title` is used for both `exceptionType` and `nodeType`, but `card.title` holds the ComfyUI node class type (e.g. "KSampler"), not the Python exception class (e.g. "RuntimeError").

`createErrorCard()` in `useErrorGroups.ts` populates `card.title` from `e.node_type` / `nodeError.class_type`. The existing `ErrorDialogContent.vue` correctly maps `e.exception_type` -> `exceptionType` and `e.node_type` -> `nodeType`. With this code, the generated report will show e.g. "KSampler" where "RuntimeError" is expected.

Suggested fix: extend `ErrorCardData` with an `exceptionType` field populated from `e.exception_type` in `processExecutionError`, then use it here. Alternatives: access `executionErrorStore.lastExecutionError.exception_type` directly, or parse the message prefix (fragile).

src/components/rightSidePanel/errors/useErrorReport.ts
```diff
@@ -0,0 +1,94 @@
...
+
+import { api } from '@/scripts/api'
+import { app } from '@/scripts/app'
+import { useSystemStatsStore } from '@/stores/systemStatsStore'
+import { generateErrorReport } from '@/utils/errorReportUtil'
+
+import type { ErrorCardData } from './types'
+
+/** Module-level cache keyed by card ID + error index + error message to avoid stale entries. */
+const reportCache = new Map<string, string>()
```
christian-byrne: issue: Module-level `reportCache` is never evicted during normal operation. Each distinct failed execution adds an entry keyed by `cardId-idx-errorMessage`. Since the store creates new objects from WS messages, cache entries accumulate across the session -- an unbounded memory leak for long-running sessions with many failures.

Consider: (a) clearing the cache on `execution_start`, (b) using a bounded LRU, or (c) removing caching entirely since `generateErrorReport` is a synchronous string-builder and the only async cost is `api.getLogs()` which is called once per mount anyway.

src/components/rightSidePanel/errors/MissingPackGroupRow.vue
```diff
@@ -106,7 +106,7 @@
       <Button
         variant="secondary"
         size="md"
-        class="flex w-full flex-1"
+        class="flex w-full flex-1 rounded-lg"
```
christian-byrne, line 109: nitpick (non-blocking): This `rounded-lg` addition (also at line 164) appears unrelated to the error-report feature. Consider splitting into a separate PR for cleaner history.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -264,6 +286,22 @@ const {
...
+    return null
+  return group
+})
+
+const singleRuntimeErrorCard = computed(
+  () => singleRuntimeErrorGroup.value?.cards[0] ?? null
+)
```
christian-byrne, line 302: question: `singleRuntimeErrorCard` only activates when there's exactly 1 filtered group with 1 card that has a runtime error. If other error groups exist alongside the runtime error (e.g. missing nodes), the dedicated full-height layout won't activate and it falls back to the accordion view. Is that intentional? It seems reasonable but worth confirming the UX intent.

christian-byrne (dismissed):

src/components/rightSidePanel/errors/MissingPackGroupRow.vue
jaeone94: You're right. This is definitely an unrelated fix. However, while working on this PR, the designer pointed out a discrepancy in the button's “round” value, and a new button was included in this PR. When adding the new button, I addressed this issue at the same time to maintain consistency with the other buttons. I think this could be handled in a separate PR, but it's a very simple fix and doesn't seem like it would be a major issue.

src/components/rightSidePanel/errors/useErrorReport.ts
jaeone94: Fixed. Added `exceptionType` field to `ErrorItem` in `types.ts`, populated from `e.exception_type` in `processExecutionError()` in `useErrorGroups.ts`. The composable now reads `error.exceptionType` directly, with a named constant `FALLBACK_EXCEPTION_TYPE` (`'Runtime Error'`) as the fallback when the backend does not provide one.

src/components/rightSidePanel/errors/useErrorReport.ts
jaeone94: Agreed. Removed the cache entirely. As you noted, `generateErrorReport` is a synchronous string builder — the only async cost is `api.getLogs()`, which is a lightweight call that runs once per mount. The simplification eliminates the unbounded memory growth concern and removes the `clearReportCache` test utility that was needed for test isolation.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Yes, this is intentional due to CSS layout constraints.

The `v-if`/`v-else` split is necessary because the accordion lives inside a scrollable container (`overflow-y-auto`), which allows its children to grow indefinitely. Within this context, `flex-1` cannot constrain the height of the details section — the content simply expands and the scroll shifts to the TabErrors level instead.

For runtime errors, the details section displays a full diagnostic report (system info, GPU/VRAM, logs, workflow JSON), which is significantly longer than a typical traceback. We wanted this content to fill the available panel height with its own scrollbar, which requires a flex layout from a non-scrollable parent — hence the dedicated rendering path outside the accordion.

When a runtime error coexists with other error types (e.g., missing nodes), it falls back to the accordion layout. During review, we identified that the runtime error details were excessively tall within the accordion in this mixed-error scenario. This has been addressed by applying `max-h-[15lh]` to runtime error details when rendered inside the accordion, controlled via the `fullHeight` prop (only `true` in the dedicated panel path).

src/components/rightSidePanel/errors/ErrorNodeCard.vue
```diff
@@ -125,10 +173,30 @@ function handleEnterSubgraph() {
...
   )
 }
+
+function handleGetHelp() {
+  telemetry?.trackHelpResourceClicked({
+    resource_type: 'help_feedback',
+    is_external: true,
+    source: 'error_dialog'
+  })
+  useCommandStore().execute('Comfy.ContactSupport')
```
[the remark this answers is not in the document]
jaeone94: Intentionally kept as `'error_dialog'`. The Get Help button replicates the legacy error dialog's "Help Fix This" action. Changing the source to a new value (e.g., `'error_tab'`) would split the metric for what is functionally the same user action, making it appear as though legacy dialog usage dropped while a new unknown source appeared. This would be confusing for analytics consumers. If separate tracking is needed in the future, it should be coordinated with the analytics team to ensure dashboards are updated accordingly.

src/locales/en/main.json
```diff
@@ -104,6 +104,8 @@
     "usageHint": "Usage hint",
     "triggerPhrase": "Trigger phrase",
     "findIssues": "Find Issues",
+    "findOnGithub": "Find on GitHub",
+    "getHelpAction": "Get Help",
```
dante01yoon, line 108: nit: The action buttons are unconditionally rendered for every error item.  Searching GitHub for validation errors like "Required input is missing" or "value not in  list" is unlikely to produce useful results. Consider gating at least "Find on GitHub"  behind error.isRuntimeError, or explain the UX rationale if intentional.

src/components/rightSidePanel/errors/ErrorNodeCard.vue
```diff
@@ -125,10 +170,30 @@ function handleEnterSubgraph() {
...
```
dante01yoon: issue: useCommandStore().execute(...) is called inline in handleGetHelp, while useTelemetry() and useExternalLink() are correctly stored at the top level. This is  inconsistent. While Pinia stores work when called inside handlers, the idiomatic pattern is to call them at setup time

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -264,6 +287,22 @@ const {
   swapNodeGroups
 } = useErrorGroups(searchQuery, t)
 
+const singleRuntimeErrorGroup = computed(() => {
+  if (filteredGroups.value.length !== 1) return null
+  const group = filteredGroups.value[0]
+  if (
+    group.type !== 'execution' ||
+    group.cards.length !== 1 ||
+    !group.cards[0].errors.some((e) => e.isRuntimeError)
```
dante01yoon: issue: group.cards[0].errors.some((e) => e.isRuntimeError) means a card with mixed errors (runtime + validation) would get the full-height layout. Isn't better use `every()` to only   activate the dedicated panel when all errors on the card are runtime errors. I see coderabbitai flagged this already — worth confirming the fix landed.

src/components/rightSidePanel/errors/useErrorReport.ts
```diff
@@ -0,0 +1,75 @@
...
+    const card = toValue(cardSource)
+    const runtimeErrors = card.errors
+      .map((error, idx) => ({ error, idx }))
+      .filter(({ error }) => error.isRuntimeError)
+
+    if (runtimeErrors.length === 0) return
+
+    if (!systemStatsStore.systemStats) {
+      try {
+        await systemStatsStore.refetchSystemStats()
```
dante01yoon, line 44: Suggestion: no cancellation on unmount. The onMounted callback does async work (getLogs, serialize, generateErrorReport) and writes to enrichedDetails without checking if the component is  still mounted. If the user navigates away quickly, the writes are harmless but it's a code smell. why don't you use abort controller or use flag to cancel

dante01yoon (commented):
Solid feature that brings the error tab closer to parity with the legacy error  dialog. Good test coverage

src/locales/en/main.json
jaeone94: This was a deliberate decision made in discussion with the designer. The concern is that hiding the button for validation errors could signal to users that they *shouldn't* search GitHub for those errors, which isn't the intent. Additionally, this button is planned to be replaced in the future with a link to beginner-friendly documentation — the specifics are still being worked out, but the button slot is intentionally kept visible for all error types.

src/components/rightSidePanel/errors/ErrorNodeCard.vue
jaeone94: Good catch — fixed in cdacd4d. Moved `useCommandStore()` to setup scope and referenced via `commandStore` in the handler, consistent with how `useTelemetry()` and `useExternalLink()` are used in the same file.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Good point. Defensively switched to `every()` and refactored the condition into a positive check with a named `isSoleRuntimeError` variable for readability. Fixed in cdacd4d.

src/components/rightSidePanel/errors/useErrorReport.ts
jaeone94: Agreed — fixed in cdacd4d. Added an `onUnmounted` hook that sets a `cancelled` flag, checked after each `await` point (`refetchSystemStats` and `getLogs`) to skip state writes if the component has already been unmounted.

dante01yoon (approved):
Comment / Suggestion well addressed

PR #10158, opened 2026-03-17
feat: add WaveAudioPlayer with waveform visualization and authenticated audio fetch

viva-jinyi:
## Summary

Add a waveform-based audio player component (`WaveAudioPlayer`) replacing the native `<audio>` element, with authenticated API fetch for cloud audio playback.

## Changes

- **What**:
  - Add `useWaveAudioPlayer` composable with waveform visualization from audio data (Web Audio API `decodeAudioData`), playback controls, and seek support
  - Add `WaveAudioPlayer.vue` component with compact (inline waveform + time) and expanded (full transport controls) variants
  - Replace native `<audio>` in `MediaAudioTop.vue` and `ResultAudio.vue` with `WaveAudioPlayer`
  - Use `api.fetchApi()` instead of bare `fetch()` to include Firebase JWT auth headers, fixing 401 errors in cloud environments
  - Add Storybook stories and unit tests

## Review Focus

- The audio URL is fetched via `api.fetchApi()` with auth headers, converted to a Blob URL, then passed to the native `<audio>` element. This avoids 401 Unauthorized in cloud environments where `/api/view` requires authentication.
- URL-to-route extraction logic (`url.includes(apiBase)`) handles both full API URLs and relative paths.

[screen-capture.webm](https://github.com/user-attachments/assets/44e61812-0391-4b47-a199-92927e75f8b4)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10158-feat-add-WaveAudioPlayer-with-waveform-visualization-and-authenticated-audio-fetch-3266d73d365081beab3fc6274c39fcd4) by [Unito](https://www.unito.io)


src/composables/useWaveAudioPlayer.ts
```diff
@@ -0,0 +1,167 @@
...
+    handleWaveformClick
+  }
+}
+
+export function formatTime(seconds: number): string {
+  if (isNaN(seconds) || seconds === 0) return '0:00'
+  const mins = Math.floor(seconds / 60)
+  const secs = Math.floor(seconds % 60)
+  return `${mins}:${secs.toString().padStart(2, '0')}`
+}
```
DrJKL: Can we reuse the one in audioUtils?

src/composables/useWaveAudioPlayer.ts
```diff
@@ -0,0 +1,167 @@
...
+    if (blobUrl.value) URL.revokeObjectURL(blobUrl.value)
+  })
+
+  return {
+    audioRef,
+    waveformRef,
+    audioSrc,
+    bars,
+    loading,
+    isPlaying: playing,
```
DrJKL, line 191: The only one that has to be renamed...

src/composables/useWaveAudioPlayer.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    if (!playing.value) {
+      playing.value = true
+    }
+  }
+
+  watch(
+    src,
+    (url) => {
+      if (url) {
```
DrJKL: When the whole watch body is truthy gated, you can use `whenever`

src/composables/useWaveAudioPlayer.ts
```diff
@@ -0,0 +1,193 @@
...
+      })
+      if (blobUrl.value) URL.revokeObjectURL(blobUrl.value)
+      blobUrl.value = URL.createObjectURL(blob)
+
+      ctx = new AudioContext()
+      const audioBuffer = await ctx.decodeAudioData(arrayBuffer)
+      generateBarsFromBuffer(audioBuffer)
+    } catch {
+      bars.value = generatePlaceholderBars()
+    } finally {
```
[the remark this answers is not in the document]
viva-jinyi: The catch block already provides a UI fallback via `bars.value = generatePlaceholderBars()` — placeholder waveform bars are shown when decoding fails, which is the intended graceful degradation. There's currently no UI to display an error state, and no consumer for an error ref, so adding one would be YAGNI. We can revisit if error reporting becomes a requirement.

src/renderer/extensions/vueNodes/widgets/utils/audioUtils.ts
```diff
@@ -1,16 +1,7 @@
...
- * Format time in MM:SS format
- */
-export function formatTime(seconds: number): string {
-  if (isNaN(seconds) || seconds === 0) return '0:00'
-
-  const mins = Math.floor(seconds / 60)
-  const secs = Math.floor(seconds % 60)
-  return `${mins}:${secs.toString().padStart(2, '0')}`
-}
+export { formatTime } from '@/utils/formatUtil'
```
DrJKL: Instead of re-exporting, can the consumers all just import from formatUtil?

src/components/common/WaveAudioPlayer.vue
```diff
@@ -0,0 +1,226 @@
...
+  barCount = 40,
+  height = 32,
+  align = 'center',
+  variant = 'compact'
+} = defineProps<{
+  src: string
+  barCount?: number
+  height?: number
+  align?: 'center' | 'bottom'
+  variant?: 'compact' | 'expanded'
```
DrJKL, line 187: Worth using `cva`?

src/components/common/WaveAudioPlayer.vue
```diff
@@ -0,0 +1,226 @@
...
+            loading
+              ? 'bg-muted-foreground/20'
+              : index <= playedBarIndex
+                ? 'bg-base-foreground'
+                : 'bg-muted-foreground/40'
+          )
+        "
+        :style="{
+          height: (bar.height / 100) * height + 'px',
+          minHeight: '2px'
```
DrJKL: This one could be baked into the classlist, since it's constant.

src/components/common/WaveAudioPlayer.vue
```diff
@@ -0,0 +1,226 @@
...
+
+    <div
+      :ref="(el) => (waveformRef = el as HTMLElement)"
+      :class="
+        cn(
+          'flex min-w-0 flex-1 cursor-pointer gap-px',
+          align === 'center' ? 'items-center' : 'items-end'
+        )
+      "
+      :style="{ height: height + 'px' }"
```
DrJKL, line 34: For all of the styles derived from height, one option is to use a CSS variable instead that you can set once at a common root and use something like h-(--wave-bar-height) so that you're not mutating styles on as many DOM elements.

DrJKL (dismissed):

src/components/common/WaveAudioPlayer.vue
```diff
@@ -0,0 +1,197 @@
+<template>
+  <!-- Compact: [▶] [waveform] [time] -->
+  <div
+    v-if="variant === 'compact'"
+    :class="
+      cn('flex w-full gap-2', align === 'center' ? 'items-center' : 'items-end')
+    "
+    @pointerdown.stop
+  >
```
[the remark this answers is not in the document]
viva-jinyi: Added `@click.stop` to both compact and expanded containers in 7d7152c.

src/composables/useWaveAudioPlayer.ts
viva-jinyi: Good call — switched to `whenever` from VueUse in 7d7152c.

src/composables/useWaveAudioPlayer.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  const playedBarIndex = computed(() => {
+    if (duration.value === 0) return -1
+    return Math.floor((currentTime.value / duration.value) * barCount) - 1
+  })
+
+  const formattedCurrentTime = computed(() => formatTime(currentTime.value))
+  const formattedDuration = computed(() => formatTime(duration.value))
+
+  const audioSrc = computed(() => blobUrl.value ?? src.value)
```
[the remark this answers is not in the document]
viva-jinyi: Fixed in 7d7152c:
- catch block now revokes and clears `blobUrl` on failure
- `audioSrc` returns empty string when `src` is falsy
- `onUnmounted` increments `decodeRequestId` to invalidate in-flight decodes

src/renderer/extensions/vueNodes/widgets/utils/audioUtils.ts
viva-jinyi: Removed the re-export. AudioPreviewPlayer and WidgetRecordAudio now import `formatTime` directly from `@/utils/formatUtil` (7d7152c).

src/components/common/WaveAudioPlayer.vue
viva-jinyi: Moved to Tailwind `min-h-0.5` class in 7d7152c.

src/components/common/WaveAudioPlayer.vue
viva-jinyi: Considered it, but the current variant logic is simple enough that `cn()` handles it well. Adding `cva` would introduce more indirection without much benefit at this scale.

src/components/common/WaveAudioPlayer.vue
viva-jinyi: Nice idea for the container height. However, individual bar heights are each computed from `bar.height / 100 * height`, so per-element inline styles are still needed there. Keeping it consistent with inline `:style` for both for now.

DrJKL (approved):

viva-jinyi:
@coderabbitai can you approve it?

PR #10168, opened 2026-03-17
fix: make graph canvas toolbar visible on mobile

christian-byrne:
## Summary

Fix the bottom-right graph canvas toolbar (zoom controls, fit-to-view, minimap toggle) not being visible on mobile devices in normal graph mode.

## Problem

PrimeVue applies `overflow: hidden` to all `.p-splitterpanel` elements by default. The `GraphCanvasMenu` component is absolutely positioned (`right-0 bottom-0`) inside the `graph-canvas-panel` SplitterPanel. On mobile viewports, the panel's bounding box can be smaller than the full canvas area, causing the toolbar to be clipped by the `overflow: hidden`.

## Solution

Add `overflow-visible` to the `graph-canvas-panel` SplitterPanel class to override PrimeVue's default `overflow: hidden`. This allows the absolutely-positioned toolbar (and minimap) to remain visible regardless of viewport size.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10168-fix-make-graph-canvas-toolbar-visible-on-mobile-3266d73d36508130b675e839cb748fd5) by [Unito](https://www.unito.io)


christian-byrne:
I'm not sure why the mobile baseline e2e test didn't catch this bug initially, going to add a new one now though.

Edit: done

viva-jinyi (approved):
LGTM

PR #10177, opened 2026-03-17
feat: add tab status indicator (running/done/errored)

pythongosssss:
## Summary

Adds indicator to show outcome of last job per tab, cleared next time the workflow is activated.

## Changes

- **What**: 
- add workflow status tracking to execution store, handling various events
- add icon to tab based on store
- handle race condition where job finishes instantly (e.g. invalid workflow or already executed)

https://github.com/user-attachments/assets/8b1d8d8e-57d4-4ac2-9cc3-0d218d6eb0f7

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10177-feat-add-tab-status-indicator-running-done-errored-3266d73d365081a89f5dfd58487bb065) by [Unito](https://www.unito.io)


src/stores/executionStore.ts
```diff
@@ -574,6 +628,15 @@ export const useExecutionStore = defineStore('execution', () => {
...
     jobIdToSessionWorkflowPath.value = next
+
+    // If this job is still executing, mark the workflow as running.
+    // Handles the race where handleExecutionStart fired before the path
+    // mapping existed (WebSocket arrived before the HTTP response).
+    // The `has(path)` guard prevents overwriting a terminal status written
+    // by a late-arriving WS event for a job that already completed.
+    if (activeJobId.value === jobId && !workflowStatus.value.has(path)) {
+      mutateStatus((m) => m.set(path, 'running'))
+    }
```
christian-byrne: issue (blocking): the recovery path here only writes 'running'. For an instant-finish job (cached, invalid, already executed) the WS sequence `execution_start` → `execution_success`/`_error`/`_interrupted` arrives before `storeJob` runs, so `setWorkflowStatus` no-ops in all three handlers (no path mapping yet). By the time `ensureSessionWorkflowPath` runs from the late HTTP, `activeJobId` has been cleared by `resetExecutionState`, so this branch also no-ops. The terminal status is silently dropped for the very class of jobs the PR description calls out ("invalid workflow or already executed"). Suggested fix: buffer terminal statuses by `jobId` when `setWorkflowStatus` no-ops on missing path, then flush in `ensureSessionWorkflowPath` after the mapping is created. Found by 5 independent reviewers.

christian-byrne (commented):
External PR review for #10177 (feat: add tab status indicator).

Synthesized from 14 review agents across 3 waves: 1 critical, 6 major, 20 minor, 12 nitpicks. Highlights:

- Comment [1] is the headline finding: terminal status is silently dropped for instant-finish jobs (the very race the PR description targets), and the new tests on lines 494-502/539-555 of `executionStore.test.ts` codify this gap rather than catch it. Found independently by 5 agents.
- Comment [5] is a real WCAG 1.4.11 fail: the running spinner uses `text-muted-foreground` which is ~1.35:1 against `--comfy-menu-bg` in the default dark theme.
- Comments [3], [4] are correctness issues (cancel rendered as error; status keyed by mutable `workflow.path`).

Architectural notes (ADRs 0003/0008): no entity classes, World, or graph state were touched; the entity-architecture rules apply to LGraphNode/LGraphCanvas, not to Pinia stores, so this PR is compliant.

Posting as a comment review per author request; reviewer would normally REQUEST_CHANGES on [1] + [5].

src/stores/executionStore.ts
```diff
@@ -574,6 +628,15 @@ export const useExecutionStore = defineStore('execution', () => {
       else break
     }
     jobIdToSessionWorkflowPath.value = next
+
+    // If this job is still executing, mark the workflow as running.
+    // Handles the race where handleExecutionStart fired before the path
+    // mapping existed (WebSocket arrived before the HTTP response).
+    // The `has(path)` guard prevents overwriting a terminal status written
+    // by a late-arriving WS event for a job that already completed.
+    if (activeJobId.value === jobId && !workflowStatus.value.has(path)) {
```
christian-byrne: issue: the `!workflowStatus.value.has(path)` guard is not sufficient on its own. If the user re-queues the same active workflow, the previous run's terminal status is still in the map (the activeWorkflow watcher only clears on path change, not on re-queue), so a fresh running job is blocked from showing 'running' here. Either clear the path's status proactively at the start of `handleExecutionStart`, or pair this guard with the per-jobId pending-status logic for finding 1.

src/stores/executionStore.ts
```diff
@@ -270,6 +320,7 @@ export const useExecutionStore = defineStore('execution', () => {
     e: CustomEvent<ExecutionInterruptedWsMessage>
   ) {
     const jobId = e.detail.prompt_id
+    setWorkflowStatus(jobId, 'failed')
```
christian-byrne: issue: `execution_interrupted` writes 'failed', which renders as `lucide--octagon-alert text-destructive-background` (red error icon). User-initiated interrupts (Stop button) are not failures, so conflating cancel with error is misleading UX. Either introduce a separate 'interrupted' state with a neutral icon (e.g. `lucide--circle-stop`/`text-muted-foreground`), or skip the status write on interrupt.

src/stores/executionStore.ts
```diff
@@ -80,6 +82,53 @@ export const useExecutionStore = defineStore('execution', () => {
 
   const initializingJobIds = ref<Set<string>>(new Set())
 
+  /**
+   * Per-workflow-path execution status for the current session.
```
christian-byrne: issue: the status map keys on `workflow.path`, but `ComfyWorkflow.path` is mutable. `workflowStore.renameWorkflow` updates it in place. After a background tab is renamed, the old path is no longer in `openWorkflows`, the prune watcher drops the entry, and the new path has nothing, so the indicator silently disappears. Consider keying by `workflow.activeState?.id` (the store already maintains `jobIdToWorkflowId`) or by the `ComfyWorkflow` reference in a `WeakMap`.

src/components/topbar/WorkflowTab.vue
```diff
@@ -21,8 +21,27 @@
...
+            v-if="jobState"
+            data-testid="job-state-indicator"
+            role="status"
+            :data-state="jobState"
+            :aria-label="jobStateLabel"
+            :class="
+              cn(
+                'absolute top-1/2 left-1/2 z-10 size-4 -translate-1/2 group-hover:hidden',
+                jobState === 'running' &&
+                  'icon-[lucide--loader-circle] animate-spin text-muted-foreground',
```
christian-byrne: issue (a11y): `text-muted-foreground` resolves to `--color-charcoal-200` (#494a50) against `--comfy-menu-bg` (#353535) in the default dark theme, contrast ratio about 1.35:1, well below the 3:1 required by WCAG 2.1 SC 1.4.11 for graphical UI components. The spinner is effectively invisible on the running state until hover (and `group-hover:hidden` then removes it). 'completed' (~5.6:1) and 'failed' (~3.7:1) both pass. Suggest a higher-contrast token (e.g. `text-foreground` or a dedicated info/accent token) and verifying against both themes.

src/stores/executionStore.test.ts
```diff
@@ -440,6 +453,165 @@ describe('useExecutionStore - reconcileInitializingJobs', () => {
...
+    expect(store.workflowStatus.get('/workflows/a.json')).toBeUndefined()
+  })
+
+  it('does not set running when job was interrupted before HTTP arrives', () => {
+    fireExecutionStart('job-1')
+    fireExecutionInterrupted('job-1')
+
+    callStoreJob('job-1', '/workflows/a.json')
+    expect(store.workflowStatus.get('/workflows/a.json')).toBeUndefined()
+  })
```
christian-byrne, line 635: issue: these tests assert `workflowStatus.get(path)` is `undefined` when WS events arrive before HTTP, i.e. they pin the broken behavior from finding 1 into the suite. If finding 1 is fixed, these tests should be updated to assert the terminal status is set. As written they document the gap rather than catching it.

src/stores/executionStore.test.ts
```diff
@@ -440,6 +453,165 @@ describe('useExecutionStore - reconcileInitializingJobs', () => {
...
+    expect(store.workflowStatus.get('/workflows/a.json')).toBe('completed')
+  })
+
+  it('sets failed on execution_error', () => {
+    callStoreJob('job-1', '/workflows/a.json')
+    fireExecutionStart('job-1')
+    fireExecutionError('job-1')
+
+    expect(store.workflowStatus.get('/workflows/a.json')).toBe('failed')
+  })
```
christian-byrne, line 651: issue: 'sets failed on execution_error' only covers the case where `callStoreJob` already populated the path mapping. The "invalid workflow" path (one of the two scenarios in the PR summary) never fires `execution_start`, so no path mapping exists when `execution_error` arrives. A test that fires `execution_error` first, then `storeJob`, and asserts the workflow ends 'failed' would surface finding 1 today.

src/stores/executionStore.ts
```diff
@@ -652,6 +715,7 @@ export const useExecutionStore = defineStore('execution', () => {
     nodeLocatorIdToExecutionId,
     jobIdToWorkflowId,
     jobIdToSessionWorkflowPath,
-    ensureSessionWorkflowPath
+    ensureSessionWorkflowPath,
```
christian-byrne, line 854: suggestion (non-blocking): `workflowStatus` is exposed as the raw `Map` inside a `shallowRef`. Internal writers correctly funnel through `mutateStatus` (which replaces `.value` to trigger reactivity), but any external consumer that calls `.set(path, ...)` or `.delete(path)` will mutate the underlying Map without triggering subscribers. The sole consumer (`WorkflowTab.vue:189`) only calls `.get(path)`; consider exposing a `getWorkflowStatus(path): WorkflowStatus | undefined` accessor instead.

src/stores/executionStore.ts
```diff
@@ -80,6 +82,53 @@ export const useExecutionStore = defineStore('execution', () => {
...
+  function clearWorkflowStatus(path: string) {
+    if (!workflowStatus.value.has(path)) return
+    mutateStatus((m) => m.delete(path))
+  }
+
+  // Clear status when a workflow becomes active — the user has seen it.
+  watch(
+    () => workflowStore.activeWorkflow?.path,
+    (path) => {
+      if (path) clearWorkflowStatus(path)
```
christian-byrne: suggestion (non-blocking): with the default 'pre' flush, if a status write lands in the same microtask as an `activeWorkflow` change, the order is non-deterministic. A 'completed' write can land after the watcher's clear runs; the indicator is hidden by `isActiveTab` in the template, but the entry persists and reappears on the next tab switch. Re-clearing inside `setWorkflowStatus` when the resolved path equals `workflowStore.activeWorkflow?.path` would close this.

src/stores/executionStore.ts
```diff
@@ -80,6 +82,53 @@ export const useExecutionStore = defineStore('execution', () => {
...
+    }
+  )
+
+  // Prune statuses for workflows that have been closed.
+  watch(
+    () => workflowStore.openWorkflows,
+    (openWorkflows) => {
```
christian-byrne, line 203: suggestion (non-blocking): if a tab closes while a job is running, late `execution_success`/`_error` events re-insert an entry under the now-closed workflow path; nothing prunes it until `openWorkflows` next changes. Consider gating `setWorkflowStatus` on membership in the open-workflow path set before writing.

src/stores/executionStore.ts
christian-byrne, line 854: suggestion (non-blocking): `unbindExecutionEvents` removes WS listeners but does not clear `workflowStatus`. If `bindExecutionEvents` is later called again (reconnect, test rebind), pre-existing terminal statuses persist into the new session. Consider resetting in `unbindExecutionEvents` (or at the top of `bindExecutionEvents`).

src/stores/executionStore.ts
```diff
@@ -574,6 +628,15 @@ export const useExecutionStore = defineStore('execution', () => {
       else break
     }
     jobIdToSessionWorkflowPath.value = next
+
```
christian-byrne: suggestion (non-blocking): when `jobIdToSessionWorkflowPath` exceeds `MAX_SESSION_PATH_ENTRIES` and oldest entries are evicted, a long-running job's mapping can be evicted before its terminal WS event arrives. `setWorkflowStatus` then no-ops and the indicator is silently lost. Consider evicting only mappings for jobs already in a terminal state, or skipping eviction when a job is still pending.

src/stores/executionStore.ts
```diff
@@ -383,6 +435,8 @@ export const useExecutionStore = defineStore('execution', () => {
   }
 
   function handleExecutionError(e: CustomEvent<ExecutionErrorWsMessage>) {
+    setWorkflowStatus(e.detail.prompt_id, 'failed')
```
christian-byrne: question: `setWorkflowStatus(prompt_id, 'failed')` runs unconditionally at the top of `handleExecutionError`, before `handleCloudValidationError` and `handleServiceLevelError` classify. For pre-flight cloud validation errors (no work happened) the tab shows a persistent "failed" indicator alongside the validation toast/dialog. Is that intended? If not, moving this after the cloud/service-level short-circuits would scope the indicator to actual runtime errors.

src/components/topbar/WorkflowTab.vue
```diff
@@ -159,6 +183,22 @@ const isActiveTab = computed(() => {
   return workflowStore.activeWorkflow?.key === props.workflowOption.workflow.key
 })
 
+const jobState = computed(() => {
```
christian-byrne: suggestion (non-blocking): the store names the type/ref/setter `WorkflowStatus`/`workflowStatus`/`setWorkflowStatus`, but the component renames everywhere to `jobState`/`jobStateLabel`/`jobStateI18nKeys`/`data-state`/`data-testid="job-state-indicator"`/`g.running`. Two names for one concept across the boundary forces mental translation and breaks grep. Picking one term end-to-end (suggest `workflowStatus`) would help.

src/components/topbar/WorkflowTab.vue
```diff
@@ -21,8 +21,27 @@
...
+                jobState === 'completed' &&
+                  'icon-[lucide--circle-check] text-success-background',
+                jobState === 'failed' &&
+                  'icon-[lucide--octagon-alert] text-destructive-background'
+              )
+            "
+          />
           <span
-            v-if="shouldShowStatusIndicator"
+            v-else-if="shouldShowStatusIndicator"
```
christian-byrne: suggestion (non-blocking): now that two indicators (job state, unsaved dot) live in the same `<div class="relative">` slot, `shouldShowStatusIndicator` only governs the unsaved dot, and its testid was even renamed to `unsaved-indicator` on line 44. Consider renaming the computed to `shouldShowUnsavedIndicator` to match.

src/stores/executionStore.test.ts
```diff
@@ -440,6 +453,165 @@ describe('useExecutionStore - reconcileInitializingJobs', () => {
...
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+    apiEventHandlers.clear()
+    mockActiveWorkflow.value = null
+    mockOpenWorkflows.value = []
+    setActivePinia(createTestingPinia({ stubActions: false }))
+    store = useExecutionStore()
+    store.bindExecutionEvents()
+  })
```
christian-byrne, line 595: suggestion (non-blocking): `mockActiveWorkflow` and `mockOpenWorkflows` are hoisted at module scope; only this describe's `beforeEach` resets them. Future tests in other describes that mutate these refs will silently leak state across blocks; this is the "shared mutable state between tests" anti-pattern called out in `docs/testing/unit-testing.md`. Consider moving resets to a top-level `beforeEach` or wrapping in a helper.

src/components/topbar/WorkflowTab.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      isLast: false
+    }
+  })
+}
+
+describe('WorkflowTab - job state indicator', () => {
+  beforeEach(() => {
+    mockWorkflowStatus.value = new Map()
+    mockCloseWorkflow.mockClear()
+  })
```
christian-byrne, line 170: suggestion (non-blocking): the parametrized test asserts both `data-state` (internal hook) and `aria-label` (user-facing). Per AGENTS.md ("Do not write tests that are dependent on non-behavioral features like utility classes or styles"), prefer dropping the `data-state` assertion and relying on `screen.getByRole('status', { name: ... })`.

src/stores/executionStore.test.ts
```diff
@@ -440,6 +453,165 @@ describe('useExecutionStore - reconcileInitializingJobs', () => {
...
+    fireExecutionSuccess('job-a')
+    fireExecutionStart('job-b')
+    fireExecutionSuccess('job-b')
+
+    mockActiveWorkflow.value = { path: '/workflows/a.json' }
+    await nextTick()
+
+    expect(store.workflowStatus.has('/workflows/a.json')).toBe(false)
+    expect(store.workflowStatus.get('/workflows/b.json')).toBe('completed')
+  })
```
christian-byrne, line 702: suggestion (non-blocking): the activation-clears-status case only exercises 'completed'. The watcher clears any status, including 'running', so pinning that with a test would catch a future guard like `if (status !== 'running')` slipping in. Parametrize over `['running','completed','failed']` or add a dedicated 'clears running status when workflow becomes active mid-run' case.

src/components/topbar/WorkflowTab.test.ts
```diff
@@ -0,0 +1,202 @@
...
+  it('delegates close to workflow service with the tab workflow', async () => {
+    renderTab()
+    const user = userEvent.setup()
+    await user.click(screen.getByTestId('close-workflow-button'))
+
+    expect(mockCloseWorkflow).toHaveBeenCalledWith(
+      expect.objectContaining({ key: 'test-key' }),
+      expect.anything()
+    )
+  })
```
christian-byrne, line 232: nitpick (non-blocking): `it('delegates close to workflow service…')` sits inside the 'WorkflowTab - job state indicator' describe but only asserts that clicking close calls `mockCloseWorkflow`, pure mock interaction unrelated to this PR's feature. Either remove (predates this PR's scope) or move to a separate describe and assert observable consequences (event emitted, tab unmounted).

src/components/topbar/WorkflowTab.vue
```diff
@@ -21,8 +21,27 @@
           {{ workflowOption.workflow.filename }}
         </span>
         <div class="relative">
+          <i
+            v-if="jobState"
+            data-testid="job-state-indicator"
+            role="status"
```
christian-byrne: suggestion (non-blocking) (a11y): `role="status"` is an implicit polite live region, but the algorithm announces changes to descendant text content, not attribute changes. The `<i>` has no text (Iconify renders via CSS), so AT may not re-announce on `aria-label` change. Also, every non-active tab contributes its own live region; many concurrent regions can lead to overlapping or coalesced announcements. A single shared `aria-live="polite"` region in `WorkflowTabs.vue` (e.g. "{filename} completed") would address both.

src/components/topbar/WorkflowTab.vue
christian-byrne: suggestion (non-blocking) (a11y): `animate-spin` runs indefinitely regardless of `prefers-reduced-motion`. Consider `motion-safe:animate-spin` (or a non-animated alternative under reduced-motion).

src/stores/executionStore.ts
```diff
@@ -257,6 +306,7 @@ export const useExecutionStore = defineStore('execution', () => {
       const path = queuedJobs.value[activeJobId.value]?.workflow?.path
       if (path) ensureSessionWorkflowPath(activeJobId.value, path)
     }
+    setWorkflowStatus(activeJobId.value, 'running')
```
christian-byrne, line 386: nitpick: when path is not yet mapped, `handleExecutionStart` calls `ensureSessionWorkflowPath(jobId, path)` (which sets 'running' at line 637), then immediately calls `setWorkflowStatus(activeJobId.value, 'running')` again here. Two `mutateStatus` allocations and two reactivity triggers per execution start. Consider funneling both call sites through one helper or dropping the unconditional set after the if-block.

src/stores/executionStore.test.ts
```diff
@@ -440,6 +453,165 @@ describe('useExecutionStore - reconcileInitializingJobs', () => {
...
+  }
+
+  function callStoreJob(jobId: string, workflowPath: string) {
+    store.storeJob({
+      nodes: ['1'],
+      id: jobId,
+      workflow: { path: workflowPath } as Parameters<
+        typeof store.storeJob
+      >[0]['workflow']
+    })
```
christian-byrne, line 585: nitpick: `callStoreJob` builds `{ path }` and casts via `as Parameters<typeof store.storeJob>[0]['workflow']`. Future changes to `ComfyWorkflow` adding required fields touched by `storeJob` won't break this test even if they break callers in production. Consider a faithful workflow stub/factory.

src/stores/executionStore.ts
christian-byrne: suggestion (non-blocking): `jobIdToSessionWorkflowPath` (jobId→path) and `workflowStatus` (path→status) are two interlocking shallowRef Maps keyed differently; every mutation hops through one to key the other, and the race-handling logic exists specifically because they can be populated out of order. AGENTS.md flags interlocking maps as a smell. Consider `Map<jobId, { path; status }>` with a path-keyed view via `computed`.

src/stores/executionStore.ts
christian-byrne, line 386: suggestion (non-blocking): 'running' is written from both `handleExecutionStart` and `ensureSessionWorkflowPath`. The `!workflowStatus.value.has(path)` guard at line 637 also encodes a subtle ordering rule (terminal status from late WS must win). Funneling both into one `markRunningIfActive(jobId)` helper called once from `handleExecutionStart` after `ensureSessionWorkflowPath` would consolidate the rule.

src/stores/executionStore.ts
```diff
@@ -80,6 +82,53 @@ export const useExecutionStore = defineStore('execution', () => {
...
+  }
+
+  function clearWorkflowStatus(path: string) {
+    if (!workflowStatus.value.has(path)) return
+    mutateStatus((m) => m.delete(path))
+  }
+
+  // Clear status when a workflow becomes active — the user has seen it.
+  watch(
+    () => workflowStore.activeWorkflow?.path,
```
christian-byrne: question: this watcher encodes a UI semantic ("user looked at the tab, dismiss indicator"). The execution store now imports the workflow management store solely to react to a UI focus event. Would moving the "clear on activate" policy into a thin composable consumed by `WorkflowTab.vue` make sense, keeping the store as a pure source of execution facts?

src/components/topbar/WorkflowTab.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      changeTracker: null,
+      ...overrides
+    }
+  } as unknown as WorkflowTabProps['workflowOption']
+}
+
+function renderTab({
+  workflowOverrides = {},
+  activeWorkflowKey = 'other-key'
+} = {}) {
```
christian-byrne, line 136: suggestion (non-blocking): `as unknown as` is the canonical workaround for the `as any` rule and triggers the same anti-pattern AGENTS.md targets. The test silently accepts a partial `ComfyWorkflow` shape; if the type adds a new required field, the test keeps compiling. Build a proper `ComfyWorkflow` factory or destructure only the fields the component reads with a typed test helper.

src/components/topbar/WorkflowTab.vue
```diff
@@ -159,6 +183,22 @@ const isActiveTab = computed(() => {
...
+  const path = props.workflowOption.workflow.path
+  if (!path || isActiveTab.value) return null
+  return executionStore.workflowStatus.get(path) ?? null
+})
+
+const jobStateI18nKeys: Record<WorkflowStatus, string> = {
+  running: 'g.running',
+  completed: 'g.completed',
+  failed: 'g.failed'
+}
```
christian-byrne, line 187: nitpick (non-blocking): `jobStateI18nKeys` is a static `Record` declared inside `<script setup>`, so it is re-allocated for every WorkflowTab instance. Hoisting to module scope makes it a single allocation.

src/components/topbar/WorkflowTab.vue
```diff
@@ -159,6 +183,22 @@ const isActiveTab = computed(() => {
...
+
+const jobStateLabel = computed(() =>
+  jobState.value ? t(jobStateI18nKeys[jobState.value]) : undefined
```
christian-byrne: nitpick (non-blocking): `jobStateLabel` returns `undefined` while `jobState` returns `null`. Consider one absence sentinel for both (suggest `undefined`).

src/stores/executionStore.ts
```diff
@@ -53,6 +53,8 @@ interface QueuedJob {
  */
 export const MAX_PROGRESS_JOBS = 1000
 
+export type WorkflowStatus = 'running' | 'completed' | 'failed'
```
christian-byrne: nitpick (non-blocking): `WorkflowStatus` is a generic name for a tab-scoped concept. `HubWorkflowStatus` already exists in `packages/ingest-types/src/types.gen.ts` for a different domain; future generated types may collide. Renaming to `WorkflowExecutionStatus` or `TabWorkflowStatus` now is cheap; post-merge it is breaking.

christian-byrne: nitpick (non-blocking): PR title says "errored", code uses `'failed'`. Optional brief comment on the type cross-referencing the user-facing label.

src/components/topbar/WorkflowTab.vue
```diff
@@ -21,8 +21,27 @@
...
+            :class="
+              cn(
+                'absolute top-1/2 left-1/2 z-10 size-4 -translate-1/2 group-hover:hidden',
+                jobState === 'running' &&
+                  'icon-[lucide--loader-circle] animate-spin text-muted-foreground',
+                jobState === 'completed' &&
+                  'icon-[lucide--circle-check] text-success-background',
+                jobState === 'failed' &&
+                  'icon-[lucide--octagon-alert] text-destructive-background'
+              )
```
christian-byrne, line 32: nitpick (non-blocking): inline `cn()` mixes layout classes with three status-conditional icon/color chunks; every new status adds another inline branch. Hoisting a `Record<WorkflowStatus, string>` `jobStateIconClasses` to module scope reduces the template to `cn('...', jobStateIconClasses[jobState])`.

src/stores/executionStore.ts
christian-byrne: nitpick (non-blocking): the 'running' subset of `workflowStatus` is derivable from `runningJobIds` + `jobIdToSessionWorkflowPath`. Either derive 'running' as a computed and store only terminal outcomes here, or add a comment documenting `workflowStatus` as the canonical source for tab indicators.

src/stores/executionStore.ts
```diff
@@ -80,6 +82,53 @@ export const useExecutionStore = defineStore('execution', () => {
...
   const initializingJobIds = ref<Set<string>>(new Set())
 
+  /**
+   * Per-workflow-path execution status for the current session.
+   * Updated by WebSocket event handlers; cleared automatically when the
+   * workflow becomes active or is closed.
+   */
+  const workflowStatus = shallowRef<Map<string, WorkflowStatus>>(new Map())
+
+  function mutateStatus(mutator: (map: Map<string, WorkflowStatus>) => void) {
```
christian-byrne: nitpick (non-blocking): the clone-on-write pattern in `mutateStatus` duplicates the in-place clone in `ensureSessionWorkflowPath`. Worth a follow-up `useShallowMap<K,V>()` extraction; not for this PR.

src/stores/executionStore.ts
```diff
@@ -80,6 +82,53 @@ export const useExecutionStore = defineStore('execution', () => {
...
+    const next = new Map(workflowStatus.value)
+    mutator(next)
+    workflowStatus.value = next
+  }
```
christian-byrne, line 167: nitpick (non-blocking): consider `triggerRef` instead of a full Map clone for single-key writes. Stylistic; either pattern is valid.

src/components/topbar/WorkflowTab.vue
```diff
@@ -21,8 +21,27 @@
           {{ workflowOption.workflow.filename }}
         </span>
         <div class="relative">
+          <i
+            v-if="jobState"
+            data-testid="job-state-indicator"
```
christian-byrne: nitpick (non-blocking): `data-testid="job-state-indicator"` and `data-testid="close-workflow-button"` ship to production but both are queryable via `getByRole(...)` without testids. Optional drop.

src/components/topbar/WorkflowTab.test.ts
```diff
@@ -0,0 +1,202 @@
...
+  })
+}
+
+describe('WorkflowTab - job state indicator', () => {
+  beforeEach(() => {
+    mockWorkflowStatus.value = new Map()
+    mockCloseWorkflow.mockClear()
+  })
+
+  it.for(['running', 'completed', 'failed'] as const)(
```
christian-byrne, line 172: nitpick (non-blocking): describe name 'WorkflowTab - job state indicator' undersells coverage (also covers close behavior). Split into separate describes or rename the outer describe to `WorkflowTab` with nested ones per concern.

src/components/topbar/WorkflowTab.vue
christian-byrne, line 187: nitpick (non-blocking): `jobStateI18nKeys: Record<WorkflowStatus, string>` is exhaustiveness-checked at type level today, but adding a new status without updating the record yields `t(undefined)` and an empty `aria-label`. An inline `satisfies Record<WorkflowStatus, string>` plus a small unit test would pin it.

src/stores/executionStore.ts
christian-byrne: nitpick (non-blocking): `setWorkflowStatus` is placed before `resetExecutionState`, and the latter does not touch `jobIdToSessionWorkflowPath` today. A future cleanup adding path-map eviction to `resetExecutionState` would silently break the indicator. Optional brief comment near these call sites noting the dependency.

pythongosssss: skipped write on interrupt

src/stores/executionStore.ts
pythongosssss: updated to use ComfyWorkflow reference in a WeakMap

src/components/topbar/WorkflowTab.vue
pythongosssss: updated to `text-base-foreground`

src/stores/executionStore.ts
pythongosssss: added `getWorkflowStatus` as suggested

src/stores/executionStore.ts
pythongosssss: Moved

src/stores/executionStore.ts
pythongosssss: Documented

src/components/topbar/WorkflowTab.vue
pythongosssss: Kept close testid, removed indicator

src/composables/useWorkflowStatusDismissal.ts
```diff
@@ -0,0 +1,23 @@
...
+import { useExecutionStore } from '@/stores/executionStore'
+
+export function useWorkflowStatusDismissal() {
+  const workflowStore = useWorkflowStore()
+  const executionStore = useExecutionStore()
+
+  watch(
+    () => {
+      const workflow = workflowStore.activeWorkflow
+      return workflow ? executionStore.getWorkflowStatus(workflow) : undefined
```
christian-byrne: suggestion (non-blocking): the watch source returns the status value of the active workflow rather than the active workflow itself, so it only re-fires when the resolved status changes. It works today because activation clears the active tab's terminal status to undefined (so the prior value is almost always undefined when you land on a terminal tab), but watching activeWorkflow directly would express the intent more clearly and drop the dual activeWorkflow read:

```ts
watch(
  () => workflowStore.activeWorkflow,
  (workflow) => {
    if (workflow && executionStore.getWorkflowStatus(workflow) !== 'running') {
      executionStore.clearWorkflowStatus(workflow)
    }
  },
  { immediate: true }
)
```

christian-byrne (dismissed):
Second-pass review. The blocking and major findings from the first pass are all resolved and now covered by tests: the instant-finish terminal-status race is fixed via the pendingWorkflowStatusByJobId buffer + flushPendingWorkflowStatus on storeJob, re-queue overwrites stale terminal with running, interrupt no longer writes failed, status is keyed by ComfyWorkflow reference instead of mutable path, the spinner uses text-base-foreground (+ motion-safe:animate-spin), and the failed write now runs after the cloud/service-level error guards. Tests flipped from pinning the dropped-status behavior to asserting the terminal status survives.

Approving on the code. One non-blocking suggestion inline. Note: the branch currently conflicts with main in executionStore.ts and executionStore.test.ts; the conflicts are purely additive (telemetry/share-tracking from main vs status-tracking here) so a keep-both rebase resolves them with no logic change.

christian-byrne (approved):
Re-approving on the current head. The earlier approval was auto-dismissed by the main-merge commits; the only delta since is those merges (no feature changes), and CI is green. All first-pass blocking/major findings remain resolved and test-covered.

PR #10186, opened 2026-03-17
feat: fire subscription_success telemetry on subscription activation

christian-byrne:
## Summary

Fire a client-side `subscription_success` event to GTM when a subscription activates, enabling LinkedIn and Meta conversion tracking tags.

## Changes

- **What**: Wire `trackMonthlySubscriptionSucceeded()` into both subscription detection paths (legacy dialog watcher + workspace billingOperationStore). This method existed but had zero call sites (dead code). The event name remains `subscription_success` (not `purchase`) to avoid double-counting with the server-side GA4 Measurement Protocol purchase event and the existing Google Ads Purchase conversion tag in GTM.

## Review Focus

- billingOperationStore only fires telemetry for `subscription` operations, not `topup`.
- Legacy flow: telemetry fires in the existing `isActiveSubscription` watcher, before `emit("close", true)`.
- Workspace flow: telemetry fires in `handleSuccess()` after status update but before billing context refresh.
- No event name change: `subscription_success` was the original name and avoids collision with the server-side `purchase` event.

### DIFF
--- a/src/platform/workspace/stores/billingOperationStore.ts
+++ b/src/platform/workspace/stores/billingOperationStore.ts
@@ -4,6 +4,7 @@ import { computed, ref } from 'vue'
 
 import { useBillingContext } from '@/composables/billing/useBillingContext'
 import { t } from '@/i18n'
+import { useTelemetry } from '@/platform/telemetry'
 import { useToastStore } from '@/platform/updates/common/toastStore'
 import { workspaceApi } from '@/platform/workspace/api/workspaceApi'
 import { useSettingsDialog } from '@/platform/settings/composables/useSettingsDialog'
@@ -132,6 +133,10 @@ export const useBillingOperationStore = defineStore('billingOperation', () => {
 
     updateOperationStatus(opId, 'succeeded', null)
     cleanup(opId)
+
+    if (operation.type === 'subscription') {
+      useTelemetry()?.trackMonthlySubscriptionSucceeded()
+    }
 
     const billingContext = useBillingContext()
     await Promise.all([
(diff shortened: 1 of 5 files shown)


src/platform/workspace/stores/billingOperationStore.ts
```diff
@@ -133,6 +134,10 @@ export const useBillingOperationStore = defineStore('billingOperation', () => {
     updateOperationStatus(opId, 'succeeded', null)
     cleanup(opId)
 
+    if (operation.type === 'subscription') {
+      useTelemetry()?.trackMonthlySubscriptionSucceeded()
```
benceruleanlu, line 138: Could we wire the GTM side of this before merging? These new call sites invoke `trackMonthlySubscriptionSucceeded()`, but `GtmTelemetryProvider` still does not implement that method, so this PR will not actually emit the conversion event to `window.dataLayer`. As written, it looks like this changes Mixpanel/PostHog behavior but not the GTM/LinkedIn/Meta path described in the PR.

benceruleanlu (changes requested):
There’s one blocking issue before this is ready to merge: the new `trackMonthlySubscriptionSucceeded()` call sites do not currently produce the GTM conversion event described in the PR because the GTM provider still does not implement that hook.

benceruleanlu:
Added this on the branch in `e181310bf`.

`GtmTelemetryProvider` now implements `trackMonthlySubscriptionSucceeded()` and pushes `subscription_success` to `window.dataLayer`, and I added a focused unit test covering that event in `GtmTelemetryProvider.test.ts`.

christian-byrne:
Appreciate it, could you approve or do you still have changes you want @benceruleanlu ?

benceruleanlu (approved):
LGTM

PR #10187, opened 2026-03-17
fix: prevent nested SubgraphNode input slots from doubling on reload

jaeone94:
## Summary

- Fix nested SubgraphNode input slots doubling on each page reload
- Root cause: during configure, `_configureSubgraph` recreates `SubgraphInput` objects with new references, and the `input-added` event handler used `===` identity check which failed for these new objects, causing `addInput()` to duplicate inputs
- Add `id`-based fallback matching in the `input-added` handler and rebind `_subgraphSlot` with re-registered listeners

## Changes

**`SubgraphNode.ts:614-622`**: Add UUID `id` fallback to the `===` reference check in the `input-added` event handler. When a stale reference is matched by id, call `_addSubgraphInputListeners()` to update `_subgraphSlot` and re-register listeners on the new `SubgraphInput` object.

**`SubgraphNode.test.ts`**: 2 regression tests for nested subgraph reconfigure scenarios.

## Test plan

- [x] Existing SubgraphNode tests pass (6 passed, 34 skipped)
- [x] New tests verify inputs don't duplicate after single and repeated reconfigure cycles
- [x] Manual: create a subgraph containing another subgraph node, save, reload — input slots should remain unchanged

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10187-fix-prevent-nested-SubgraphNode-input-slots-from-doubling-on-reload-3266d73d3650817286abea52365a626e) by [Unito](https://www.unito.io)


src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -612,9 +612,14 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
         const existingInput = this.inputs.find(
-          (input) => input._subgraphSlot === subgraphInput
+          (input) =>
+            input._subgraphSlot === subgraphInput ||
+            (input._subgraphSlot && input._subgraphSlot.id === subgraphInput.id)
         )
         if (existingInput) {
+          // Rebind to the new SubgraphInput object and re-register listeners
+          // (configure recreates SubgraphInput objects with the same id)
+          this._addSubgraphInputListeners(subgraphInput, existingInput)
```
christian-byrne, line 622: suggestion (non-blocking): `_addSubgraphInputListeners` runs on both the `===` identity-match path and the id-based fallback path. When `===` succeeds the reference hasn't changed, so the rebind is safe (abort-first) but unnecessary. Consider guarding:
```typescript
if (existingInput._subgraphSlot !== subgraphInput) {
  this._addSubgraphInputListeners(subgraphInput, existingInput)
}
```

src/lib/litegraph/src/subgraph/SubgraphNode.test.ts
```diff
@@ -698,6 +698,55 @@ describe('SubgraphNode duplicate input pruning (#9977)', () => {
...
+    // Simulate what happens during nested subgraph configure:
+    // B.configure() calls _configureSubgraph(), which recreates SubgraphInput
+    // objects and dispatches 'input-added' events with new references.
+    const serialized = subgraph.asSerialisable()
+    subgraph.configure(serialized)
+
+    // The SubgraphNode's event listener should recognize existing inputs
+    // by ID and NOT add duplicates.
+    expect(node.inputs).toHaveLength(2)
+    expect(node.inputs.every((i) => i._subgraphSlot)).toBe(true)
```
christian-byrne, line 724: suggestion (non-blocking): The assertion only checks `_subgraphSlot` is truthy. The key behavioral change is that the reference should point to the *new* `SubgraphInput` object after reconfigure. Capturing `_subgraphSlot` references before reconfigure and asserting they changed (while IDs stayed the same) would catch a regression that keeps stale references.

src/lib/litegraph/src/subgraph/SubgraphNode.test.ts
```diff
@@ -698,6 +698,55 @@ describe('SubgraphNode duplicate input pruning (#9977)', () => {
...
+    const node = createTestSubgraphNode(subgraph)
+    expect(node.inputs).toHaveLength(2)
+
+    for (let i = 0; i < 5; i++) {
+      const serialized = subgraph.asSerialisable()
+      subgraph.configure(serialized)
+    }
+
+    expect(node.inputs).toHaveLength(2)
+    expect(node.inputs.map((i) => i.name)).toEqual(['x', 'y'])
```
christian-byrne, line 746: nitpick (non-blocking): The first test asserts `_subgraphSlot` is truthy after reconfigure, but the second test (5 cycles) doesn't -- worth adding for consistency.

src/lib/litegraph/src/subgraph/SubgraphNode.test.ts
```diff
@@ -698,6 +698,55 @@ describe('SubgraphNode duplicate input pruning (#9977)', () => {
...
+    const serialized = subgraph.asSerialisable()
+    subgraph.configure(serialized)
+
+    // The SubgraphNode's event listener should recognize existing inputs
+    // by ID and NOT add duplicates.
+    expect(node.inputs).toHaveLength(2)
+    expect(node.inputs.every((i) => i._subgraphSlot)).toBe(true)
+  })
+
+  it('should not accumulate inputs across multiple reconfigure cycles', () => {
```
christian-byrne, line 727: nitpick (non-blocking): `'should not accumulate inputs across multiple reconfigure cycles'` reads very similarly to the existing test `'should not accumulate duplicate inputs on reconfigure'`. They test different paths (`subgraph.configure` vs `node.configure`) -- maybe reflect that distinction in the name, e.g. `'...across multiple subgraph.configure() cycles'`.

christian-byrne (approved):

PR #10188, opened 2026-03-17
fix: don't resurface missing node warnings on workflow tab switch

christian-byrne:
## Problem

The E2E test "Does not report warning when switching between opened workflows" (`tests/sidebar/workflows.spec.ts:235`) is flaky because the error overlay for missing nodes re-appears after being dismissed when switching back to a workflow with missing nodes.

## Root Cause

`openWorkflow()` always passed `showMissingNodes: true` to `loadGraphData`. This caused `runMissingModelPipeline` to re-set `pendingWarnings` on every workflow switch, and then `showPendingWarnings()` would re-surface the error overlay — even though the user had already dismissed it.

## Fix

Gate `showMissingNodes` by `loadFromRemote` (same pattern already used for `showMissingModels`). When switching between already-opened workflows (`loadFromRemote === false`), missing node warnings are not re-surfaced. This matches the existing behavior for undo/redo, where `changeTracker.ts` already passes `showMissingNodes: false`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10188-fix-don-t-resurface-missing-node-warnings-on-workflow-tab-switch-3266d73d3650811b933ddf67db137b7e) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/workflow/core/services/workflowService.ts
+++ b/src/platform/workflow/core/services/workflowService.ts
@@ -240,7 +240,7 @@ export const useWorkflowService = () => {
       workflow,
       {
         showMissingModels: loadFromRemote,
-        showMissingNodes: true,
+        showMissingNodes: loadFromRemote,
         checkForRerouteMigration: false,
         deferWarnings: true
       }


jaeone94 (commented):
Hi Christian, the change to `true` on that line in the previous PR was intentional. I believed that missing node scanning is a low-cost operation, and practically speaking, when switching tabs, missing nodes are highlighted in red on the canvas but without re-scanning they won't appear in the errors tab — users have no way to find out which node packs are missing without opening the Manager (since missing node pack info won't be displayed again unless the page is refreshed). In particular, the Swap Node feature can only be triggered from the errors tab — even the Manager cannot trigger it. So the intent was to ensure the errors tab always surfaces missing nodes for users without requiring a page reload. However, I should have updated the E2E test accordingly — I apologize for missing that.

If you think this behavior is acceptable, I've just published a PR that updates the E2E test to match: https://github.com/Comfy-Org/ComfyUI_frontend/pull/10190 — would you mind reviewing it?

Alternatively, if you believe it's better to disable missing node scanning on tab switch, I'll approve this PR instead.

jaeone94:
I also recaptured the failure in the “reroute node visual appearance” screenshot comparison test ( `rerouteNodeSize.spec.ts:14` ) in #10190 (I took a snapshot after implementing the missing model, but it appears to have been overwritten by another PR before it was merged into the main branch.)

christian-byrne:
This one specifically just comes from some user complaints we were receiving on GitHub issues. I recalled the original behavior so thought it might have been a regression. I think your points are valid. We can think about it more and bring the discussion to slack for others to weigh in on if you'd like.

jaeone94:
For now, I think we should focus on resolving the current E2E test failure issue and merge #10190 while preserving the existing behavior. It would be best to consult everyone in the temp-error-check channel to determine the way forward.

PR #10190, opened 2026-03-17
test: update E2E to expect missing nodes overlay on tab switch

jaeone94:
## Summary

- Update the E2E test expectation for missing nodes overlay behavior when switching between opened workflows
- `showMissingNodes` was intentionally changed to `true` in `workflowService.ts` so users always see missing node pack info when switching tabs, allowing them to install missing packs without a page reload
- The test previously asserted the overlay should NOT reappear; now it asserts the overlay SHOULD reappear

## Test plan

- [x] CI E2E tests pass with updated expectation

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10190-test-update-E2E-to-expect-missing-nodes-overlay-on-tab-switch-3266d73d36508106a48efff7309bd4e5) by [Unito](https://www.unito.io)


christian-byrne:
Ah woops I just regenerated the test for this. You might need to rebase, pop the commit, then regenerate again. Sorry about that

christian-byrne (approved):

PR #10195, opened 2026-03-17
fix: subgraph promoted widget input label rename

artokun:
## Summary

Promoted primitive subgraph inputs (String, Int) render their link anchor at the header position instead of the widget row. Renaming subgraph input labels breaks the match entirely, causing connections to detach from their widgets visually.

## Changes

- **What**: Fix widget-input slot positioning for promoted subgraph inputs in both LiteGraph and Vue (Nodes 2.0) rendering modes
- `_arrangeWidgetInputSlots`: Removed Vue mode branch that skipped setting `input.pos`. Promoted widget inputs aren't rendered as `<InputSlot>` Vue components (NodeSlots filters them out), so `input.pos` is the only position fallback
- `drawConnections`: Added pre-pass to arrange nodes with unpositioned widget-input slots before link rendering. The background canvas renders before the foreground canvas calls `arrange()`, so positions weren't set on the first frame
- `SubgraphNode`: Sync `input.widget.name` with the display name on label rename and initial setup. The `IWidgetLocator` name diverged from `PromotedWidgetView.name` after rename, breaking all name-based slot↔widget matching (`_arrangeWidgetInputSlots`, `getWidgetFromSlot`, `getSlotFromWidget`)

## Review Focus

- The `_arrangeWidgetInputSlots` rewrite iterates `_concreteInputs` directly instead of building a spread-copy map — simpler and avoids the stale index issue
- `input.widget.name` is now kept in sync with the display name (`input.label ?? subgraphInput.name`). This is a semantic shift from using the raw internal name, but it's required for all name-based matching to work after renames. The value is overwritten on deserialize by `_setWidget` anyway
- The `_widget` fallback in `_arrangeWidgetInputSlots` is a safety net for edge cases where the name still doesn't match (e.g., stale cache)

Fixes #9998

## Screenshots
Screenshot 2026-03-17 at 3 05 32 PM
Screenshot 2026-03-17 at 3 05 23 PM
Screenshot 2026-03-17 at 3 05 16 PM
Screenshot 2026-03-17 at 3 05 06 PM

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10195-fix-subgraph-promoted-widget-input-slot-positions-after-label-rename-3266d73d365081dfa623dd94dd87c718) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/helpers/subgraphTestUtils.ts
+++ b/browser_tests/helpers/subgraphTestUtils.ts
@@ -1,3 +1,5 @@
+import type { Page } from '@playwright/test'
+
 import type { LGraph, Subgraph } from '../../src/lib/litegraph/src/litegraph'
 import { isSubgraph } from '../../src/utils/typeGuardUtil'
 
@@ -13,4 +15,31 @@ export function assertSubgraph(
       'Expected to be in a subgraph context, but graph is not a Subgraph'
     )
   }
+}
+
+/**
+ * Returns the widget-input slot Y position and the node title height
+ * for the promoted "text" input on the SubgraphNode.
+ *
+ * The slot Y should be at the widget row, not the header. A value near
+ * zero or negative indicates the slot is positioned at the header (the bug).
+ */
+export function getTextSlotPosition(page: Page, nodeId: string) {
+  return page.evaluate((id) => {
+    const node = window.app!.canvas.graph!.getNodeById(id)
+    if (!node) return null
+
+    const titleHeight = window.LiteGraph!.NODE_TITLE_HEIGHT
+
+    for (const input of node.inputs) {
+      if (!input.widget || input.type !== 'STRING') continue
+      return {
+        hasPos: !!input.pos,
+        posY: input.pos?.[1] ?? null,
+        widgetName: input.widget.name,
+        titleHeight
+      }
+    }
+    return null
+  }, nodeId)
 }
--- a/browser_tests/tests/subgraph-promoted-slot-position.spec.ts
+++ b/browser_tests/tests/subgraph-promoted-slot-position.spec.ts
@@ -1,7 +1,6 @@
 import { expect } from '@playwright/test'
 
 import { comfyPageFixture as test } from '../fixtures/ComfyPage'
-import { TestIds } from '../fixtures/selectors'
 import { getTextSlotPosition } from '../helpers/subgraphTestUtils'
 
 test.describe(
@@ -71,11 +70,7 @@ test.describe(
       await comfyPage.page.waitForSelector(dialog, { state: 'hidden' })
 
       // Navigate back to parent graph
-      const breadcrumb = comfyPage.page.getByTestId(TestIds.breadcrumb.subgraph)
-      await breadcrumb.waitFor({ state: 'visible', timeout: 5000 })
-      await breadcrumb.getByRole('link').first().click()
-      await comfyPage.nextFrame()
-      await comfyPage.nextFrame()
+      await comfyPage.subgraph.exitViaBreadcrumb()
 
       // Verify slot position is still at the widget row after rename
       const after = await getTextSlotPosition(comfyPage.page, '11')
--- a/browser_tests/tests/subgraph-promoted-slot-position.spec.ts
+++ b/browser_tests/tests/subgraph-promoted-slot-position.spec.ts
@@ -83,8 +83,9 @@ test.describe(
       expect(after!.hasPos).toBe(true)
       expect(after!.posY).toBeGreaterThan(after!.titleHeight)
 
-      // widget.name should have been synced with the new label
-      expect(after!.widgetName).toBe('my_custom_prompt')
+      // widget.name is the stable identity key — it does NOT change on rename.
+      // The display label is on input.label, read via PromotedWidgetView.label.
+      expect(after!.widgetName).not.toBe('my_custom_prompt')
     })
   }
 )
--- a/browser_tests/tests/subgraph-promoted-slot-position.spec.ts
+++ b/browser_tests/tests/subgraph-promoted-slot-position.spec.ts
@@ -1,46 +1,13 @@
-import type { Page } from '@playwright/test'
 import { expect } from '@playwright/test'
 
 import { comfyPageFixture as test } from '../fixtures/ComfyPage'
 import { TestIds } from '../fixtures/selectors'
-
-/**
- * Returns the widget-input slot Y position and the node title height
- * for the promoted "text" input on the SubgraphNode.
- *
- * The slot Y should be at the widget row, not the header. A value near
- * zero or negative indicates the slot is positioned at the header (the bug).
- */
-async function getTextSlotPosition(page: Page, nodeId: string) {
-  return page.evaluate((id) => {
-    const node = window.app!.canvas.graph!.getNodeById(id)
-    if (!node) return null
-
-    const titleHeight = window.LiteGraph!['NODE_TITLE_HEIGHT'] as number
-    const slotHeight = window.LiteGraph!['NODE_SLOT_HEIGHT'] as number
-
-    for (const input of node.inputs) {
-      if (!input.widget || input.type !== 'STRING') continue
-      return {
-        hasPos: !!input.pos,
-        posY: input.pos?.[1] ?? null,
-        widgetName: input.widget.name,
-        titleHeight,
-        slotHeight
-      }
-    }
-    return null
-  }, nodeId)
-}
+import { getTextSlotPosition } from '../helpers/subgraphTestUtils'
 
 test.describe(
   'Subgraph promoted widget-input slot position',
   { tag: '@subgraph' },
   () => {
-    test.beforeEach(async ({ comfyPage }) => {
-      await comfyPage.settings.setSetting('Comfy.UseNewMenu', 'Disabled')
-    })
-
     test('Promoted text widget slot is positioned at widget row, not header', async ({
       comfyPage
     }) => {
--- a/browser_tests/tests/subgraph-promoted-slot-position.spec.ts
+++ b/browser_tests/tests/subgraph-promoted-slot-position.spec.ts
@@ -1,0 +1,123 @@
+import type { Page } from '@playwright/test'
+import { expect } from '@playwright/test'
+
+import { comfyPageFixture as test } from '../fixtures/ComfyPage'
+import { TestIds } from '../fixtures/selectors'
+
+/**
+ * Returns the widget-input slot Y position and the node title height
+ * for the promoted "text" input on the SubgraphNode.
+ *
+ * The slot Y should be at the widget row, not the header. A value near
+ * zero or negative indicates the slot is positioned at the header (the bug).
+ */
+async function getTextSlotPosition(page: Page, nodeId: string) {
+  return page.evaluate((id) => {
+    const node = window.app!.canvas.graph!.getNodeById(id)
+    if (!node) return null
+
+    const titleHeight = window.LiteGraph!['NODE_TITLE_HEIGHT'] as number
+    const slotHeight = window.LiteGraph!['NODE_SLOT_HEIGHT'] as number
+
+    for (const input of node.inputs) {
+      if (!input.widget || input.type !== 'STRING') continue
+      return {
+        hasPos: !!input.pos,
+        posY: input.pos?.[1] ?? null,
+        widgetName: input.widget.name,
+        titleHeight,
+        slotHeight
+      }
+    }
+    return null
+  }, nodeId)
+}
+
+test.describe(
+  'Subgraph promoted widget-input slot position',
+  { tag: '@subgraph' },
+  () => {
+    test.beforeEach(async ({ comfyPage }) => {
+      await comfyPage.settings.setSetting('Comfy.UseNewMenu', 'Disabled')
+    })
+
+    test('Promoted text widget slot is positioned at widget row, not header', async ({
+      comfyPage
+    }) => {
+      await comfyPage.workflow.loadWorkflow(
+        'subgraphs/subgraph-with-promoted-text-widget'
+      )
+
+      // Render a few frames so arrange() runs
+      await comfyPage.nextFrame()
+      await comfyPage.nextFrame()
+
+      const result = await getTextSlotPosition(comfyPage.page, '11')
+      expect(result).not.toBeNull()
+      expect(result!.hasPos).toBe(true)
+
+      // The slot Y position should be well below the title area.
+      // If it's near 0 or negative, the slot is stuck at the header (the bug).
+      expect(result!.posY).toBeGreaterThan(result!.titleHeight)
+    })
+
+    test('Slot position remains correct after renaming subgraph input label', async ({
+      comfyPage
+    }) => {
+      await comfyPage.workflow.loadWorkflow(
+        'subgraphs/subgraph-with-promoted-text-widget'
+      )
+      await comfyPage.nextFrame()
+      await comfyPage.nextFrame()
+
+      // Verify initial position is correct
+      const before = await getTextSlotPosition(comfyPage.page, '11')
+      expect(before).not.toBeNull()
+      expect(before!.hasPos).toBe(true)
+      expect(before!.posY).toBeGreaterThan(before!.titleHeight)
+
+      // Navigate into subgraph and rename the text input
+      const subgraphNode = await comfyPage.nodeOps.getNodeRefById('11')
+      await subgraphNode.navigateIntoSubgraph()
+
+      const initialLabel = await comfyPage.page.evaluate(() => {
+        const graph = window.app!.canvas.graph
+        if (!graph || !('inputNode' in graph)) return null
+        const textInput = graph.inputs?.find(
+          (i: { type: string }) => i.type === 'STRING'
+        )
+        return textInput?.label || textInput?.name || null
+      })
+
+      if (!initialLabel)
+        throw new Error('Could not find STRING input in subgraph')
+
+      await comfyPage.subgraph.rightClickInputSlot(initialLabel)
+      await comfyPage.contextMenu.clickLitegraphMenuItem('Rename Slot')
+      await comfyPage.nextFrame()
+
+      const dialog = '.graphdialog input'
+      await comfyPage.page.waitForSelector(dialog, { state: 'visible' })
+      await comfyPage.page.fill(dialog, '')
+      await comfyPage.page.fill(dialog, 'my_custom_prompt')
+      await comfyPage.page.keyboard.press('Enter')
+      await comfyPage.page.waitForSelector(dialog, { state: 'hidden' })
+
+      // Navigate back to parent graph
+      const breadcrumb = comfyPage.page.getByTestId(TestIds.breadcrumb.subgraph)
+      await breadcrumb.waitFor({ state: 'visible', timeout: 5000 })
+      await breadcrumb.getByRole('link').first().click()
+      await comfyPage.nextFrame()
+      await comfyPage.nextFrame()
+
+      // Verify slot position is still at the widget row after rename
+      const after = await getTextSlotPosition(comfyPage.page, '11')
+      expect(after).not.toBeNull()
+      expect(after!.hasPos).toBe(true)
+      expect(after!.posY).toBeGreaterThan(after!.titleHeight)
+
+      // widget.name should have been synced with the new label
+      expect(after!.widgetName).toBe('my_custom_prompt')
+    })
+  }
+)
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -26,7 +26,8 @@ const config: KnipConfig = {
       project: ['src/**/*.{js,ts}']
     },
     'packages/ingest-types': {
-      project: ['src/**/*.{js,ts}']
+      project: ['src/**/*.{js,ts}'],
+      entry: ['src/index.ts', 'src/zod.gen.ts']
     }
   },
   ignoreBinaries: ['python3'],
@@ -50,7 +51,9 @@ const config: KnipConfig = {
     // Pending integration in stacked PR
     'src/components/sidebar/tabs/nodeLibrary/CustomNodesPanel.vue',
     // Agent review check config, not part of the build
-    '.agents/checks/eslint.strict.config.js'
+    '.agents/checks/eslint.strict.config.js',
+    // Auto-generated registry types with unused exports
+    'packages/registry-types/src/comfyRegistryTypes.ts'
   ],
   vite: {
     config: ['vite?(.*).config.mts']
--- a/knip.config.ts
+++ b/knip.config.ts
@@ -26,8 +26,7 @@ const config: KnipConfig = {
       project: ['src/**/*.{js,ts}']
     },
     'packages/ingest-types': {
-      project: ['src/**/*.{js,ts}'],
-      entry: ['src/index.ts', 'src/zod.gen.ts']
+      project: ['src/**/*.{js,ts}']
     }
   },
   ignoreBinaries: ['python3'],
@@ -51,9 +50,7 @@ const config: KnipConfig = {
     // Pending integration in stacked PR
     'src/components/sidebar/tabs/nodeLibrary/CustomNodesPanel.vue',
     // Agent review check config, not part of the build
-    '.agents/checks/eslint.strict.config.js',
-    // Auto-generated registry types with unused exports
-    'packages/registry-types/src/comfyRegistryTypes.ts'
+    '.agents/checks/eslint.strict.config.js'
   ],
   vite: {
     config: ['vite?(.*).config.mts']
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -92,6 +92,8 @@ export interface SafeWidgetData {
    * execution ID (e.g. `"65:42"` vs the host node's `"65"`).
    */
   sourceExecutionId?: string
+  /** Tooltip text from the resolved widget. */
+  tooltip?: string
   /** For promoted widgets, the display label from the subgraph input slot. */
   promotedLabel?: string
 }
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -92,6 +92,8 @@ export interface SafeWidgetData {
    * execution ID (e.g. `"65:42"` vs the host node's `"65"`).
    */
   sourceExecutionId?: string
+  /** For promoted widgets, the display label from the subgraph input slot. */
+  promotedLabel?: string
 }
 
 export interface VueNodeData {
@@ -352,7 +354,8 @@ function safeWidgetMapper(
           sourceNode && app.rootGraph
             ? (getExecutionIdByNode(app.rootGraph, sourceNode) ?? undefined)
             : undefined,
-        tooltip: widget.tooltip
+        tooltip: widget.tooltip,
+        promotedLabel: isPromotedWidgetView(widget) ? widget.label : undefined
       }
     } catch (error) {
       console.warn(
--- a/src/composables/graph/useGraphNodeManager.ts
+++ b/src/composables/graph/useGraphNodeManager.ts
@@ -808,6 +808,8 @@ export function useGraphNodeManager(graph: LGraph): GraphNodeManager {
         if (slotLabelEvent.slotType !== NodeSlotType.INPUT && nodeRef.outputs) {
           nodeRef.outputs = [...nodeRef.outputs]
         }
+        // Re-extract widget data so promotedLabel reflects the rename
+        vueNodeData.set(nodeId, extractVueNodeData(nodeRef))
       }
     }
 
--- a/src/composables/graph/useVueNodeLifecycle.ts
+++ b/src/composables/graph/useVueNodeLifecycle.ts
@@ -91,13 +91,16 @@ function useVueNodeLifecycleIndividual() {
       // the first legacy drawConnections frame (which may run before
       // drawNode on the foreground canvas).
       const graph = comfyApp.canvas?.graph
-      if (graph) {
-        for (const node of graph._nodes) {
-          try {
-            if (!node.flags.collapsed) node.arrange()
-          } catch {
-            /* skip nodes not fully initialized */
-          }
+      if (!graph) {
+        comfyApp.canvas?.setDirty(true, true)
+        return
+      }
+      for (const node of graph._nodes) {
+        if (node.flags.collapsed) continue
+        try {
+          node.arrange()
+        } catch {
+          /* skip nodes not fully initialized */
         }
       }
 
--- a/src/composables/graph/useVueNodeLifecycle.ts
+++ b/src/composables/graph/useVueNodeLifecycle.ts
@@ -86,6 +86,21 @@ function useVueNodeLifecycleIndividual() {
     () => !shouldRenderVueNodes.value,
     () => {
       disposeNodeManagerAndSyncs()
+
+      // Force arrange() on all nodes so input.pos is computed before
+      // the first legacy drawConnections frame (which may run before
+      // drawNode on the foreground canvas).
+      const graph = comfyApp.canvas?.graph
+      if (graph) {
+        for (const node of graph._nodes) {
+          try {
+            if (!node.flags.collapsed) node.arrange()
+          } catch {
+            /* skip nodes not fully initialized */
+          }
+        }
+      }
+
       comfyApp.canvas?.setDirty(true, true)
     }
   )
--- a/src/core/graph/subgraph/promotedWidgetView.ts
+++ b/src/core/graph/subgraph/promotedWidgetView.ts
@@ -103,7 +103,13 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   get name(): string {
-    return this.identityName ?? this.sourceWidgetName
+    if (this.identityName) return this.identityName
+
+    // Resolve from the SubgraphNode's input by matching source IDs
+    const slot = this.findBoundSubgraphSlot()
+    if (slot?.name) return slot.name
+
+    return this.sourceWidgetName
   }
 
   get y(): number {
@@ -191,13 +197,49 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   get label(): string | undefined {
-    const state = this.getWidgetState()
-    return state?.label ?? this.displayName ?? this.sourceWidgetName
+    // Live-read from the bound subgraph slot's display name
+    const slot = this.findBoundSubgraphSlot()
+    if (slot) {
+      return 'displayName' in slot
+        ? (slot as { displayName: string }).displayName
+        : slot.name
+    }
+    return this.displayName
   }
 
-  set label(value: string | undefined) {
-    const state = this.getWidgetState()
-    if (state) state.label = value
+  /** Finds the SubgraphInput slot bound to this promoted view. */
+  private findBoundSubgraphSlot():
+    | { name: string; label?: string }
+    | undefined {
+    for (const input of this.subgraphNode.inputs ?? []) {
+      const slot = input._subgraphSlot as
+        | { name: string; label?: string }
+        | undefined
+      if (!slot) continue
+
+      // Match by subgraph slot's connected interior node, not by _widget identity
+      const w = input._widget
+      if (w && 'sourceNodeId' in w && 'sourceWidgetName' in w) {
+        if (
+          w.sourceNodeId === this.sourceNodeId &&
+          w.sourceWidgetName === this.sourceWidgetName
+        ) {
+          return slot
+        }
+      }
+
+      // Fallback: match by widget locator name against our composite key
+      if (
+        input.widget?.name === `${this.sourceNodeId}:${this.sourceWidgetName}`
+      ) {
+        return slot
+      }
+    }
+    return undefined
+  }
+
+  set label(_value: string | undefined) {
+    // No-op: label is derived from the subgraph input slot.
   }
 
   get hidden(): boolean {
--- a/src/core/graph/subgraph/promotedWidgetView.ts
+++ b/src/core/graph/subgraph/promotedWidgetView.ts
@@ -50,14 +50,16 @@ export function createPromotedWidgetView(
   nodeId: string,
   widgetName: string,
   displayName?: string,
-  disambiguatingSourceNodeId?: string
+  disambiguatingSourceNodeId?: string,
+  identityName?: string
 ): IPromotedWidgetView {
   return new PromotedWidgetView(
     subgraphNode,
     nodeId,
     widgetName,
     displayName,
-    disambiguatingSourceNodeId
+    disambiguatingSourceNodeId,
+    identityName
   )
 }
 
@@ -88,7 +90,8 @@ class PromotedWidgetView implements IPromotedWidgetView {
     nodeId: string,
     widgetName: string,
     private readonly displayName?: string,
-    readonly disambiguatingSourceNodeId?: string
+    readonly disambiguatingSourceNodeId?: string,
+    private readonly identityName?: string
   ) {
     this.sourceNodeId = nodeId
     this.sourceWidgetName = widgetName
@@ -100,7 +103,7 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   get name(): string {
-    return this.displayName ?? this.sourceWidgetName
+    return this.identityName ?? this.sourceWidgetName
   }
 
   get y(): number {
--- a/src/core/graph/subgraph/promotedWidgetView.ts
+++ b/src/core/graph/subgraph/promotedWidgetView.ts
@@ -202,7 +202,7 @@ class PromotedWidgetView implements IPromotedWidgetView {
 
   get label(): string | undefined {
     const slot = this.getBoundSubgraphSlot()
-    if (slot) return slot.displayName ?? slot.name
+    if (slot) return slot.label ?? slot.displayName ?? slot.name
     return this.displayName
   }
 
@@ -214,7 +214,12 @@ class PromotedWidgetView implements IPromotedWidgetView {
 
   /**
    * Returns the cached bound subgraph slot reference, refreshing only when
-   * the subgraph node's input list has changed (version mismatch).
+   * the subgraph node's input list has changed (length mismatch).
+   *
+   * Note: Using length as the cache key works because the returned reference
+   * is the same mutable slot object. When slot properties (label, name) change,
+   * the caller reads fresh values from that reference.  The cache only needs
+   * to invalidate when slots are added or removed, which changes length.
    */
   private getBoundSubgraphSlot(): SubgraphSlotRef | undefined {
     const version = this.subgraphNode.inputs?.length ?? 0
@@ -292,17 +297,19 @@ class PromotedWidgetView implements IPromotedWidgetView {
     projected.value = this.value
     projected.label = this.label
 
-    projected.drawWidget(ctx, {
-      width: widgetWidth,
-      showText: !lowQuality,
-      suppressPromotedOutline: true,
-      previewImages: resolved.node.imgs
-    })
-
-    projected.y = originalY
-    projected.computedHeight = originalComputedHeight
-    projected.computedDisabled = originalComputedDisabled
-    projected.label = originalLabel
+    try {
+      projected.drawWidget(ctx, {
+        width: widgetWidth,
+        showText: !lowQuality,
+        suppressPromotedOutline: true,
+        previewImages: resolved.node.imgs
+      })
+    } finally {
+      projected.y = originalY
+      projected.computedHeight = originalComputedHeight
+      projected.computedDisabled = originalComputedDisabled
+      projected.label = originalLabel
+    }
   }
 
   onPointerDown(
--- a/src/core/graph/subgraph/promotedWidgetView.ts
+++ b/src/core/graph/subgraph/promotedWidgetView.ts
@@ -199,21 +199,17 @@ class PromotedWidgetView implements IPromotedWidgetView {
   get label(): string | undefined {
     // Live-read from the bound subgraph slot's display name
     const slot = this.findBoundSubgraphSlot()
-    if (slot) {
-      return 'displayName' in slot
-        ? (slot as { displayName: string }).displayName
-        : slot.name
-    }
+    if (slot) return slot.displayName ?? slot.name
     return this.displayName
   }
 
   /** Finds the SubgraphInput slot bound to this promoted view. */
   private findBoundSubgraphSlot():
-    | { name: string; label?: string }
+    | { name: string; displayName?: string }
     | undefined {
     for (const input of this.subgraphNode.inputs ?? []) {
       const slot = input._subgraphSlot as
-        | { name: string; label?: string }
+        | { name: string; displayName?: string }
         | undefined
       if (!slot) continue
 
--- a/src/core/graph/subgraph/promotedWidgetView.ts
+++ b/src/core/graph/subgraph/promotedWidgetView.ts
@@ -279,10 +279,13 @@ class PromotedWidgetView implements IPromotedWidgetView {
     const originalComputedHeight = projected.computedHeight
     const originalComputedDisabled = projected.computedDisabled
 
+    const originalLabel = projected.label
+
     projected.y = this.y
     projected.computedHeight = this.computedHeight
     projected.computedDisabled = this.computedDisabled
     projected.value = this.value
+    projected.label = this.label
 
     projected.drawWidget(ctx, {
       width: widgetWidth,
@@ -294,6 +297,7 @@ class PromotedWidgetView implements IPromotedWidgetView {
     projected.y = originalY
     projected.computedHeight = originalComputedHeight
     projected.computedDisabled = originalComputedDisabled
+    projected.label = originalLabel
   }
 
   onPointerDown(
--- a/src/core/graph/subgraph/promotedWidgetView.ts
+++ b/src/core/graph/subgraph/promotedWidgetView.ts
@@ -203,13 +203,18 @@ class PromotedWidgetView implements IPromotedWidgetView {
   get label(): string | undefined {
     const slot = this.getBoundSubgraphSlot()
     if (slot) return slot.label ?? slot.displayName ?? slot.name
-    return this.displayName
+    // Fall back to persisted widget state (survives save/reload before
+    // the slot binding is established) then to construction displayName.
+    const state = this.getWidgetState()
+    return state?.label ?? this.displayName
   }
 
   set label(value: string | undefined) {
     const slot = this.getBoundSubgraphSlot()
-    if (!slot) return
-    slot.label = value || undefined
+    if (slot) slot.label = value || undefined
+    // Also persist to widget state store for save/reload resilience
+    const state = this.getWidgetState()
+    if (state) state.label = value
   }
 
   /**
--- a/src/core/graph/subgraph/promotedWidgetView.ts
+++ b/src/core/graph/subgraph/promotedWidgetView.ts
@@ -27,6 +27,12 @@ import type { PromotedWidgetView as IPromotedWidgetView } from './promotedWidget
 export type { PromotedWidgetView } from './promotedWidgetTypes'
 export { isPromotedWidgetView } from './promotedWidgetTypes'
 
+interface SubgraphSlotRef {
+  name: string
+  label?: string
+  displayName?: string
+}
+
 function isWidgetValue(value: unknown): value is IBaseWidget['value'] {
   if (value === undefined) return true
   if (typeof value === 'string') return true
@@ -85,6 +91,10 @@ class PromotedWidgetView implements IPromotedWidgetView {
   private cachedDeepestByFrame?: { node: LGraphNode; widget: IBaseWidget }
   private cachedDeepestFrame = -1
 
+  /** Cached reference to the bound subgraph slot, set at construction. */
+  private _boundSlot?: SubgraphSlotRef
+  private _boundSlotVersion = -1
+
   constructor(
     private readonly subgraphNode: SubgraphNode,
     nodeId: string,
@@ -103,13 +113,7 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   get name(): string {
-    if (this.identityName) return this.identityName
-
-    // Resolve from the SubgraphNode's input by matching source IDs
-    const slot = this.findBoundSubgraphSlot()
-    if (slot?.name) return slot.name
-
-    return this.sourceWidgetName
+    return this.identityName ?? this.sourceWidgetName
   }
 
   get y(): number {
@@ -197,45 +201,46 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   get label(): string | undefined {
-    // Live-read from the bound subgraph slot's display name
-    const slot = this.findBoundSubgraphSlot()
+    const slot = this.getBoundSubgraphSlot()
     if (slot) return slot.displayName ?? slot.name
     return this.displayName
   }
 
-  /** Finds the SubgraphInput slot bound to this promoted view. */
-  private findBoundSubgraphSlot():
-    | { name: string; displayName?: string }
-    | undefined {
+  set label(value: string | undefined) {
+    const slot = this.getBoundSubgraphSlot()
+    if (!slot) return
+    slot.label = value || undefined
+  }
+
+  /**
+   * Returns the cached bound subgraph slot reference, refreshing only when
+   * the subgraph node's input list has changed (version mismatch).
+   */
+  private getBoundSubgraphSlot(): SubgraphSlotRef | undefined {
+    const version = this.subgraphNode.inputs?.length ?? 0
+    if (this._boundSlotVersion === version) return this._boundSlot
+
+    this._boundSlot = this.findBoundSubgraphSlot()
+    this._boundSlotVersion = version
+    return this._boundSlot
+  }
+
+  private findBoundSubgraphSlot(): SubgraphSlotRef | undefined {
     for (const input of this.subgraphNode.inputs ?? []) {
-      const slot = input._subgraphSlot as
-        | { name: string; displayName?: string }
-        | undefined
+      const slot = input._subgraphSlot as SubgraphSlotRef | undefined
       if (!slot) continue
 
-      // Match by subgraph slot's connected interior node, not by _widget identity
       const w = input._widget
-      if (w && 'sourceNodeId' in w && 'sourceWidgetName' in w) {
-        if (
-          w.sourceNodeId === this.sourceNodeId &&
-          w.sourceWidgetName === this.sourceWidgetName
-        ) {
-          return slot
-        }
-      }
-
-      // Fallback: match by widget locator name against our composite key
       if (
-        input.widget?.name === `${this.sourceNodeId}:${this.sourceWidgetName}`
+        w &&
+        isPromotedWidgetView(w) &&
+        w.sourceNodeId === this.sourceNodeId &&
+        w.sourceWidgetName === this.sourceWidgetName
       ) {
         return slot
       }
     }
     return undefined
-  }
-
-  set label(_value: string | undefined) {
-    // No-op: label is derived from the subgraph input slot.
   }
 
   get hidden(): boolean {
--- a/src/lib/litegraph/src/LGraphCanvas.drawConnections.test.ts
+++ b/src/lib/litegraph/src/LGraphCanvas.drawConnections.test.ts
@@ -9,6 +9,7 @@ import {
   LiteGraph
 } from '@/lib/litegraph/src/litegraph'
 import { LLink } from '@/lib/litegraph/src/LLink'
+import { createMockCanvas2DContext } from '@/utils/__tests__/litegraphTestUtils'
 
 vi.mock('@/renderer/core/layout/store/layoutStore', () => ({
   layoutStore: {
@@ -25,27 +26,15 @@ vi.mock('@/renderer/core/layout/store/layoutStore', () => ({
   }
 }))
 
-/** Creates a stub CanvasRenderingContext2D with all required methods mocked. */
 function createMockCtx(): CanvasRenderingContext2D {
-  return {
-    save: vi.fn(),
-    restore: vi.fn(),
+  return createMockCanvas2DContext({
     translate: vi.fn(),
     scale: vi.fn(),
-    fillRect: vi.fn(),
-    strokeRect: vi.fn(),
     fillText: vi.fn(),
     measureText: vi.fn().mockReturnValue({ width: 50 }),
-    beginPath: vi.fn(),
-    moveTo: vi.fn(),
-    lineTo: vi.fn(),
-    stroke: vi.fn(),
-    fill: vi.fn(),
     closePath: vi.fn(),
-    arc: vi.fn(),
     rect: vi.fn(),
     clip: vi.fn(),
-    clearRect: vi.fn(),
     setTransform: vi.fn(),
     roundRect: vi.fn(),
     getTransform: vi
@@ -57,10 +46,6 @@ function createMockCtx(): CanvasRenderingContext2D {
     bezierCurveTo: vi.fn(),
     quadraticCurveTo: vi.fn(),
     isPointInStroke: vi.fn().mockReturnValue(false),
-    font: '',
-    fillStyle: '',
-    strokeStyle: '',
-    lineWidth: 1,
     globalAlpha: 1,
     textAlign: 'left' as CanvasTextAlign,
     textBaseline: 'alphabetic' as CanvasTextBaseline,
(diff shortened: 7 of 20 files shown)


artokun:
@coderabbit I accidentally rebased main while you were reviewing, can you re-review?

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+        hasPos: !!input.pos,
+        posY: input.pos?.[1] ?? null,
+        widgetName: input.widget.name,
+        titleHeight,
+        slotHeight
+      }
+    }
+    return null
+  }, nodeId)
+}
```
DrJKL: This could probably be put into the subgraphTestUtils, we have a lot more that we should be moving into the helpers/fixtures soon.

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+
+test.describe(
+  'Subgraph promoted widget-input slot position',
+  { tag: '@subgraph' },
+  () => {
+    test.beforeEach(async ({ comfyPage }) => {
+      await comfyPage.settings.setSetting('Comfy.UseNewMenu', 'Disabled')
+    })
+
```
DrJKL: For a new test, we should probably use the new chrome.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -690,9 +690,8 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
         if (!input) throw new Error('Subgraph input not found')
 
         input.label = newName
-        if (input._widget) {
-          input._widget.label = newName
-        }
+        if (input.widget) input.widget.name = newName
+        if (input._widget) input._widget.label = newName
```
DrJKL: I think we probably want to isolate the input from the widget here, renaming the widget can be risky, given that the name is used as part of the key in the widget value store.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -1108,8 +1114,11 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
     // NOTE: This code creates linked chains of prototypes for passing across
     // multiple levels of subgraphs. As part of this, it intentionally avoids
     // creating new objects. Have care when making changes.
-    input.widget ??= { name: subgraphInput.name }
-    input.widget.name = subgraphInput.name
+    // Use the display name (label ?? internal name) so that slot.widget.name
+    // stays in sync with PromotedWidgetView.name after label renames.
+    const displayWidgetName = input.label ?? subgraphInput.name
+    input.widget ??= { name: displayWidgetName }
```
DrJKL: I wonder if just the label cascade here would be enough to solve the display issue.

DrJKL (commented):
Still testing some of the behavior locally, but here are my initial comments 🙂

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: I had tried this but I ran into some sync issues, but, I did this before my other changes so i'll revisit this next

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: well, I really just want to rename the placeholder so that primitive inputs are easier to discern for subgraph boundary inputs. I'll see what I can do here to propagate things better

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
artokun: I'll move it out, agreed

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -690,9 +690,8 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
         if (!input) throw new Error('Subgraph input not found')
 
         input.label = newName
-        if (input._widget) {
-          input._widget.label = newName
-        }
+        if (input.widget) input.widget.name = newName
```
DrJKL: ### 🟡 Most dangerous consumer has no test coverage

This line syncs `input.widget.name` to the renamed label. The most critical downstream consumer is `widgetInputs.ts:576-586`:

```ts
const name = input.widget.name
const w = this.widgets?.find((w) => w.name === name)
if (!w) {
  this.removeInput(...)  // DESTRUCTIVE
}
```

`onGraphConfigured` runs on **all node types** via prototype chain, including SubgraphNode. If `widget.name` and `PromotedWidgetView.name` ever diverge (stale cache, timing), this code **silently removes the input**.

The PR makes this *safer* by keeping them in sync, but there's no test proving a SubgraphNode with a renamed label survives `onGraphConfigured`. Please add a unit test that:
1. Creates a SubgraphNode with a promoted widget input
2. Renames the label (triggering `renaming-input`)
3. Triggers `onGraphConfigured`
4. Asserts the input is NOT removed

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -1108,8 +1114,11 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
...
-    input.widget ??= { name: subgraphInput.name }
-    input.widget.name = subgraphInput.name
+    // Use the display name (label ?? internal name) so that slot.widget.name
+    // stays in sync with PromotedWidgetView.name after label renames.
+    const displayWidgetName = input.label ?? subgraphInput.name
+    input.widget ??= { name: displayWidgetName }
+    input.widget.name = displayWidgetName
     if (inputWidget) Object.setPrototypeOf(input.widget, inputWidget)
 
     input._widget = view
```
DrJKL: ### 🟡 Document the semantic shift & deduplicate the expression

`input.widget.name` is now a **display name**, not an internal identifier. This is a contract change for `IWidgetLocator.name` that only applies to SubgraphNode — for all other node types, `widget.name` remains the internal widget name. Future code using `widget.name` as an identity key will break subtly.

Also, `input.label ?? subgraphInput.name` appears 3 times in `_setWidget`:
- Line 1104 (`createPromotedWidgetView` call)
- Line 1110 (`_makePromotionViewKey` — but uses `input.label ?? input.name`, a **subtle variant**)
- Line 1119 (here)

The variant on line 1110 uses `input.name` instead of `subgraphInput.name` — is that intentional? If so, add a comment explaining why the cache key diverges. If not, this is a latent bug.

Suggestion: hoist `displayWidgetName` above the `getOrCreate` call and reuse for lines 1104 and 1119.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -4211,40 +4211,27 @@ export class LGraphNode
...
-    // Only set custom pos if not using Vue positioning
-    // Vue positioning calculates widget slot positions dynamically
-    if (!LiteGraph.vueNodesMode) {
-      for (const widget of this.widgets) {
-        const slot = slotByWidgetName.get(widget.name)
-        if (!slot) continue
+      // Prefer the slot's direct _widget binding (1:1 for promoted inputs).
+      // Fall back to name-map lookup for regular nodes without _widget set.
+      const widget = slot._widget ?? widgetByName.get(slot.widget.name)
+      if (!widget) continue
```
DrJKL, line 4236: ### 🟡 Duplicate display labels cause ambiguous fallback

When `slot._widget` is unset, this falls back to `widgetByName.get(slot.widget.name)`. If two promoted inputs on the same SubgraphNode share the same label (I didn't find a uniqueness constraint), this map lookup is ambiguous — the last-inserted widget wins, and the wrong widget's Y coordinate gets used for slot positioning.

The `_widget` path is safe (1:1 identity), but consider either:
1. Enforcing unique labels per subgraph node, or
2. Adding a guard/warning when the name-map lookup is ambiguous

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -5861,6 +5862,22 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
...
+    // canvas, which may render before drawNode has executed for this frame.
+    // In Vue mode, promoted widget inputs lack DOM-registered slot positions
+    // (NodeSlots filters them out), so the fallback needs input.pos set here.
+    for (const node of nodes) {
+      if (node.flags.collapsed || !node.widgets?.length) continue
+      if (!node.inputs?.some((inp) => isWidgetInputSlot(inp) && !inp.pos))
+        continue
+
+      node._setConcreteSlots()
+      node.arrange()
```
DrJKL, line 5967: ### 🟡 Pre-pass perf on hot render path + double iteration

This loop runs on **every frame** for **every node**. The `!inp.pos` guard is self-healing (triggers once), but `.some(isWidgetInputSlot && !inp.pos)` still scans all inputs of all non-collapsed nodes per frame.

Additionally, `graph._nodes` is iterated twice — once in this pre-pass and again in the main connection loop immediately below. The pre-pass could be folded into the top of the existing per-node loop (before `getSlotPosition` is called for the same node), since `arrange()` would complete before any slot position reads.

Alternatively, consider a dirty flag on the node (e.g. `node._widgetSlotsArranged`) set after the first `arrange()` call, reducing per-frame cost to a single boolean check.

src/lib/litegraph/src/LGraphCanvas.drawConnections.test.ts
```diff
@@ -0,0 +1,222 @@
...
+    globalAlpha: 1,
+    textAlign: 'left' as CanvasTextAlign,
+    textBaseline: 'alphabetic' as CanvasTextBaseline,
+    shadowColor: '',
+    shadowBlur: 0,
+    shadowOffsetX: 0,
+    shadowOffsetY: 0,
+    imageSmoothingEnabled: true
+  } as Partial<CanvasRenderingContext2D> as CanvasRenderingContext2D
+}
```
DrJKL, line 58: ### 🔵 `createMockCtx()` duplicates shared utility

A shared `createMockCanvas2DContext` already exists in `src/utils/__tests__/litegraphTestUtils.ts` (~line 263). This 44-line local copy substantially overlaps it. Consider extending the shared factory (it accepts an `overrides` parameter) with the additional methods needed here (`bezierCurveTo`, `getTransform`, `createLinearGradient`) rather than duplicating.

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+ * The slot Y should be at the widget row, not the header. A value near
+ * zero or negative indicates the slot is positioned at the header (the bug).
+ */
+async function getTextSlotPosition(page: Page, nodeId: string) {
+  return page.evaluate((id) => {
+    const node = window.app!.canvas.graph!.getNodeById(id)
+    if (!node) return null
+
+    const titleHeight = window.LiteGraph!['NODE_TITLE_HEIGHT'] as number
+    const slotHeight = window.LiteGraph!['NODE_SLOT_HEIGHT'] as number
```
DrJKL: ### 🔵 `slotHeight` is extracted but never used

`slotHeight` is assigned on this line but never referenced in any assertion or calculation. Remove it to avoid confusion.

Also, prefer dot notation without the `as number` cast — `NODE_TITLE_HEIGHT` and `NODE_SLOT_HEIGHT` are typed as `number` on `LiteGraphGlobal`:
```ts
const titleHeight = window.LiteGraph!.NODE_TITLE_HEIGHT
```

DrJKL (changes requested):
## widget.name semantic shift — deep trace review

The fix is **correct and actually more consistent** than the old behavior — `PromotedWidgetView.name` already returns the display name, so keeping `input.widget.name` as the internal name caused every name-based lookup to diverge after a rename (which is the bug). However, this is a semantic redefinition of `IWidgetLocator.name` for SubgraphNodes that touches ~10 lookup sites, and the most dangerous path (`onGraphConfigured` input removal) has no test coverage.

See inline comments for specifics.

---

### 🟡 Serialization now persists display name to disk (`slotUtils.ts:55`)

(Not changed in this PR but directly affected by it.)

`slot.widget.name` is now the display name for SubgraphNodes, so `inputAsSerialisable` will write e.g. `{ widget: { name: "my_custom_prompt" } }` instead of `{ widget: { name: "text" } }` to the workflow file.

On reload, `_internalConfigureAfterSlots` → `_resolveInputWidget` → `_setWidget` re-establishes the correct value, so this is safe for the **current codebase**. However, workflows saved with new code **cannot be cleanly loaded by older frontend versions** that expect `widget.name` to be the internal name. This is a forward-compatibility concern worth documenting.

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
artokun: I'll add that preference to my pre-commits, better type safety this way. good call

src/lib/litegraph/src/LGraphCanvas.drawConnections.test.ts
artokun: Ah yeah, I struggled with this cause I am worried that we would end up with a overly abstracted overridable method that can break each time we touch it. I opted for a secondary duplicated function as these tests are generally ok to bloat cause they don't get packaged in with the app. I will try and merge these as you wish, I am just worried about making this shared function brittle

src/lib/litegraph/src/LGraphCanvas.ts
artokun: Setting a dirty flag on the pre-pass is a good idea... but I feel that it is out of scope for this PR, I can probably do this on a follow-on, or do it prior to this PR... I was hoping to avoid a cache-adjacent change for this PR. Let me look into this more closely to see if there's another way, like the other comment you made a few comments down to just pass the label.

Chicken and egg problem! This is a brain teaser

src/lib/litegraph/src/LGraphNode.ts
artokun: I had tested this path and didn't see any issues, but I can see where it hets hairy when we start dealing with duplicate keys (which we already handle with appending a _N counter. I will check this path again and write a test to make sure we're still good

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: This is the crux of the issue and how we ended up in this place I think, let me dig into this more and explain my reasoning

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: Good point let me address this directly and write some consumer tests. I think this path may be related to another issue where nodes disappear, but if not directly, it would be where I would look once I get to that other issue later (if not resolved already)

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
artokun: Removed the legacy chrome override so the spec runs against the default UI shell. The assertions here are canvas/subgraph-specific and shouldn't depend on old chrome.

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
artokun: Done — moved `getTextSlotPosition` into `browser_tests/helpers/subgraphTestUtils.ts` alongside `assertSubgraph`. Also dropped the unused `slotHeight` return and the `as number` casts while I was in there.

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
artokun: Cleaned up — removed `slotHeight` (wasn't referenced) and switched to direct property access on `LiteGraph.NODE_TITLE_HEIGHT` without the bracket/cast.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: Dug into this more — turns out we *need* the eager `input.widget.name = newName` here. The `widgets` getter rebuilds the PromotedWidgetView cache immediately after invalidation, so the view's `.name` updates to the new label right away. If the locator doesn't match, `onGraphConfigured` in widgetInputs.ts silently removes the input. Added a comment explaining the constraint and a unit test that proves the input survives the rename.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: Revisited — label cascade alone isn't enough. The `widgets` getter rebuilds the PromotedWidgetView with the new display name immediately on cache invalidation, so the locator (`input.widget.name`) has to be updated in the same handler to stay in sync. Hoisted `displayWidgetName` in `_setWidget` so the intent is clearer.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: Added. New test in `SubgraphNode.test.ts`: creates a SubgraphNode with a promoted widget, renames the label, then asserts `input.widget.name` still resolves to a widget in `node.widgets`. This is the exact check `onGraphConfigured` does before it calls `removeInput`, so if this test passes the input won't get silently nuked.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: Hoisted `displayWidgetName = input.label ?? subgraphInput.name` above the `getOrCreate` call and reuse it for the view factory and the locator assignment. The `input.label ?? input.name` variant on the cache key is intentional — nested subgraphs can remap `input.name` independently of the interior node's name, so the cache key needs the slot-level name. Added a comment explaining the distinction.

src/lib/litegraph/src/LGraphNode.ts
artokun: Added a comment noting the ambiguity. The `_widget` path is the primary lookup now so duplicate labels won't cause issues there. The name-map fallback only fires for regular nodes without `_widget` set, where duplicate widget names shouldn't happen. If we ever see it in the wild we can add a uniqueness constraint on the subgraph side.

src/lib/litegraph/src/LGraphCanvas.ts
artokun: Added a TODO for the dirty flag approach. Agreed it's the right optimization — the `.some()` scan is cheap per-node but adds up at scale. Going to tackle it as a follow-up so this PR stays focused on the positioning fix.

src/lib/litegraph/src/LGraphCanvas.drawConnections.test.ts
artokun: Switched to extending `createMockCanvas2DContext` with the extra methods as overrides. Keeps the shared factory as the base and only adds what `drawConnections` specifically needs (`bezierCurveTo`, `getTransform`, `createLinearGradient`, etc). Should be less brittle this way since the overrides are explicit.

artokun:
@coderabbit addressed @DrJKL feedback, can you do a sanity check? Then I will ping him for re-review

artokun:
> Tip
> 
> For best results, initiate chat on the files or code changes.
> 
> 🧩 Analysis chain
> `@artokun` — here's the sanity check summary. Everything looks solid across all four changed areas:
> 
> ### ✅ `_arrangeWidgetInputSlots` (LGraphNode.ts)
> * Correctly iterates `_concreteInputs` (1:1 with `this.inputs`, correct indices for `_measureSlot`).
> * Uses `slot._widget ?? widgetByName.get(slot.widget.name)` — the direct `_widget` reference is used for promoted inputs, avoiding name-ambiguity entirely. The name-map fallback handles non-promoted widget slots.
> * `slot.pos` is set unconditionally; the Vue-mode branch that was skipping it is gone.
> 
> ### ✅ `SubgraphNode._setWidget` / rename path
> * `displayWidgetName = input.label ?? subgraphInput.name` is computed once and used consistently.
> * The two-line pattern `input.widget ??= { name: displayWidgetName }; input.widget.name = displayWidgetName` correctly handles both first-time setup **and** re-entry after a rename.
> * Rename handler updates both `input.widget.name = newName` and `input._widget.label = newName`, then calls `_invalidatePromotedViewsCache()` — ensuring fresh views with the correct name on next `_setWidget` invocation.
> * `getWidgetFromSlot`/`getSlotFromWidget` remain name-based (`slot.widget.name === widget.name`) but are now safe because `widget.name` is kept in sync throughout.
> 
> ### ✅ `drawConnections` pre-pass (LGraphCanvas.ts)
> * The `.some((inp) => isWidgetInputSlot(inp) && !inp.pos)` gate is efficient: after the first frame where positions are set, no redundant `_setConcreteSlots()`/`arrange()` calls are made.
> 
> ### ✅ Test coverage
> * **`SubgraphNode.test.ts`**: new test `"should keep input.widget.name in sync with widgets after rename (onGraphConfigured safety)"` directly targets the regression vector — it verifies that after a rename, `promotedInput.widget.name` still resolves to a widget in `node.widgets` (which `onGraphConfigured` / `widgetInputs.ts` relies on).
> * **`LGraphCanvas.drawConnections.test.ts`**: 3 tests covering pre-render arrangement, idempotency (no re-arrange when already positioned), and mismatched display-name fallback.
> * **`browser_tests/tests/subgraph-promoted-slot-position.spec.ts`**: Playwright E2E covering both initial slot position and post-rename stability.
> 
> No issues found. The implementation is well-reasoned and the test suite covers the three root causes identified in the PR description. Good to go for `@DrJKL`'s re-review. 🚀
> 
> 🧠 Learnings used

@coderabbit do you approve this PR then? Can you mark it approved for the pre-check hooks if that's the case

src/lib/litegraph/src/subgraph/SubgraphNode.ts
DrJKL: I think that might be a hard blocker. Widgets' names are a part of the node's contract. You should be able to see the issue if you rename a widget so that it collides with another on the same node.

artokun: ok lemme try this and test it some more, thanks for reflagging this

artokun: Addressed — `input.widget.name` is no longer mutated on rename. The identity now flows through `subgraphInput.name` (stable, unique per slot, never changes on label rename):

- `PromotedWidgetView.name` returns `identityName` (the subgraph input's internal name)
- `PromotedWidgetView.label` returns the display name (follows renames)
- `input.widget.name` = `subgraphInput.name` (set once in `_setWidget`, never touched by rename handler)
- Rendering layer uses `widget.label ?? widget.name` everywhere, so the display updates correctly

Added collision tests: two promoted widgets with the same label now keep distinct identity keys — no more silent `removeInput` from `onGraphConfigured`, no ambiguous name-map lookups. The `_widget` binding stays 1:1 regardless of labels.

artokun:
@coderabbit i made a big refactor from using widget.name to widget.label with a change in the promotion path. can you do a new full review before bothering @DrJKL for a re-review?

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: **Summary of the name→label separation (final state)**

**Why:** Renaming a subgraph input's label was mutating `input.widget.name`, which is used as an identity key by `onGraphConfigured` (widgetInputs.ts) and the `_arrangeWidgetInputSlots` name-map. When two promoted inputs shared the same label, the identity keys collided — causing silent `removeInput` calls, wrong slot positions, and widget value store leakage into interior nodes.

**What changed:**

1. **`widget.name` is now the stable identity** — set once in `_setWidget` to `subgraphInput.name`, never mutated on rename. `PromotedWidgetView.name` resolves this via `findBoundSubgraphSlot()` → `slot.name`.

2. **`widget.label` handles display** — `PromotedWidgetView.label` live-reads from `subgraphSlot.displayName` (which is `slot.label ?? slot.name`). The setter is a no-op — renames go through `subgraph.renameInput()` which updates the slot, and the getter picks it up reactively.

3. **No more interior node leakage** — the old `label` setter wrote to the widget value store keyed by the interior node's ID. After disconnecting, the interior node's widget kept the renamed label. Now the label is derived from the subgraph slot, not stored on the interior widget.

4. **Rendering layer updated** — `NodeWidgets.vue` uses `promotedLabel` (from `SafeWidgetData`) for promoted widgets instead of `widgetState?.label` from the store, ensuring the display always matches the subgraph pin's label.

5. **`slotName` added to `PromotionEntry`** — all three type definitions (`SubgraphNode.ts`, `PromotedWidgetViewManager.ts`, `promotionStore.ts`) now carry `slotName?` so the identity flows cleanly through the reconciliation path without type hacks.

**Test coverage:** 3 collision tests (duplicate labels keep unique keys, inputs survive `onGraphConfigured`), updated E2E (widget.name is stable after rename), 405 unit tests + 83 promotedWidgetView tests all passing.

browser_tests/helpers/subgraphTestUtils.ts
```diff
@@ -14,3 +16,30 @@ export function assertSubgraph(
...
+ */
+export function getTextSlotPosition(page: Page, nodeId: string) {
+  return page.evaluate((id) => {
+    const node = window.app!.canvas.graph!.getNodeById(id)
+    if (!node) return null
+
+    const titleHeight = window.LiteGraph!.NODE_TITLE_HEIGHT
+
+    for (const input of node.inputs) {
+      if (!input.widget || input.type !== 'STRING') continue
```
DrJKL, line 35: Feels a little single-use, but I don't know if it's worth generalizing yet either.

artokun: Yeah agreed — it's pretty specific to this one test. If we end up needing slot position checks elsewhere I'll generalize it, but for now it's not worth the abstraction.

artokun:
Also just pushed a fix for the litegraph (non-Vue) draw path — the PromotedWidgetView.draw() method now sets projected.label = this.label before calling projected.drawWidget(), so the subgraph input's display name shows up in legacy canvas mode too. Was only working in Vue because the Vue path uses promotedLabel from SafeWidgetData.

DrJKL:
I'll take another look in the morning, just want to make sure the new tests go in before this.
<3

artokun:
@coderabbitai resume

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -188,13 +197,45 @@ class PromotedWidgetView implements IPromotedWidgetView {
...
+      }
+
+      // Fallback: match by widget locator name against our composite key
+      if (
+        input.widget?.name === `${this.sourceNodeId}:${this.sourceWidgetName}`
+      ) {
+        return slot
+      }
+    }
+    return undefined
```
DrJKL, line 253: 🔴 **Hot-path linear scan**: `findBoundSubgraphSlot()` is called from both `get name()` and `get label()`, which are per-frame render paths. Each call does an O(N) scan of all inputs with dual matching strategies.

The bound slot reference is already available at construction time in `_setWidget` — consider storing it directly on the view (or caching the result) instead of reverse-engineering it every frame.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -188,13 +197,45 @@ class PromotedWidgetView implements IPromotedWidgetView {
...
+          w.sourceWidgetName === this.sourceWidgetName
+        ) {
+          return slot
+        }
+      }
+
+      // Fallback: match by widget locator name against our composite key
+      if (
+        input.widget?.name === `${this.sourceNodeId}:${this.sourceWidgetName}`
+      ) {
```
DrJKL, line 249: 🟡 **Dead fallback**: This composite key format (`${sourceNodeId}:${sourceWidgetName}`) will never match — `_setWidget` sets `input.widget.name = subgraphInput.name` (a plain slot name, never a composite key). This branch is unreachable. Remove it or document when it could actually be hit.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -188,13 +197,45 @@ class PromotedWidgetView implements IPromotedWidgetView {
...
+    for (const input of this.subgraphNode.inputs ?? []) {
+      const slot = input._subgraphSlot as
+        | { name: string; displayName?: string }
+        | undefined
+      if (!slot) continue
+
+      // Match by subgraph slot's connected interior node, not by _widget identity
+      const w = input._widget
+      if (w && 'sourceNodeId' in w && 'sourceWidgetName' in w) {
+        if (
```
DrJKL: 🟡 **Prefer existing type guard**: `'sourceNodeId' in w && 'sourceWidgetName' in w` duplicates `isPromotedWidgetView(w)` which is already imported in this file. Using the guard is safer (checks the symbol marker) and more self-documenting.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -188,13 +197,45 @@ class PromotedWidgetView implements IPromotedWidgetView {
...
+      ) {
+        return slot
+      }
+    }
+    return undefined
+  }
+
+  set label(_value: string | undefined) {
+    // No-op: label is derived from the subgraph input slot.
   }
```
DrJKL: 🟡 **Silent no-op setter**: Callers like `renameWidget` (widgetUtil.ts:76) and the rename handler (SubgraphNode.ts:799) call `widget.label = newName`. This now silently discards the value. The rename happens to work via other paths, but the silent drop violates `IBaseWidget`'s writable contract.

Consider at minimum a `console.warn` in dev mode, or propagating the write to the slot.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -5950,6 +5951,24 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
...
+    // TODO: Add a dirty flag (e.g. node._widgetSlotsArranged) to avoid the
+    // per-frame .some() scan once slots have been positioned.
+    for (const node of nodes) {
+      if (node.flags.collapsed || !node.widgets?.length) continue
+      if (!node.inputs?.some((inp) => isWidgetInputSlot(inp) && !inp.pos))
+        continue
+
+      node._setConcreteSlots()
+      node.arrange()
+    }
```
DrJKL, line 5968: 🔴 **Render-layer band-aid**: This per-frame O(N) scan with `.some()` is expensive and architecturally misplaced — the TODO inline acknowledges this. The root cause is a lifecycle ordering gap (background canvas renders before foreground `drawNode` calls `arrange()`).

Consider either:
1. Eagerly setting `slot.pos` in `_setWidget`/`onGraphConfigured` so it's always available.
2. Adding the dirty flag mentioned in the TODO as part of this PR rather than leaving it as tech debt.

As-is, this runs every frame for every node even after positions are stable (the `!inp.pos` check only skips nodes where *all* widget slots have positions).

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -100,7 +103,13 @@ class PromotedWidgetView implements IPromotedWidgetView {
...
 
   get name(): string {
-    return this.displayName ?? this.sourceWidgetName
+    if (this.identityName) return this.identityName
```
DrJKL: 🟡 **3-way fallback chain**: `identityName` is always `subgraphInput.name` in practice (the only caller is `_setWidget`). The fallback through `findBoundSubgraphSlot()` → `sourceWidgetName` adds complexity for a case that never occurs when `identityName` is always provided.

Consider making `identityName` (or `slotName`) a required parameter and dropping the fallback chain.

src/utils/widgetUtil.ts
```diff
@@ -76,6 +76,12 @@ export function renameWidget(
   widget.label = newLabel || undefined
   if (input) {
     input.label = newLabel || undefined
+
+    const subgraphSlot = (input as { _subgraphSlot?: { label?: string } })
+      ._subgraphSlot
```
DrJKL: 🟢 **Inline type assertion**: `(input as { _subgraphSlot?: { label?: string } })` — works but fragile. Consider importing the actual `SubgraphInput` type (or a narrower interface from the slot module) to avoid shape drift.

knip.config.ts
```diff
@@ -29,10 +29,10 @@ const config: KnipConfig = {
     },
     'packages/ingest-types': {
       project: ['src/**/*.{js,ts}'],
-      entry: ['src/index.ts']
+      entry: ['src/index.ts', 'src/zod.gen.ts']
```
DrJKL: 🟢 **Unrelated change**: Adding `src/zod.gen.ts` as an entry and removing `'generate'` from `ignoreBinaries` don't appear related to this fix. Consider splitting into a separate commit for cleaner history.

DrJKL (commented):
Nice fix for a tricky bug. The core insight — using `subgraphInput.name` as a stable identity instead of the display label — is correct and well-tested. The `_arrangeWidgetInputSlots` simplification is a clear improvement. A few items to address, mostly around the `findBoundSubgraphSlot()` hot path and a dead code fallback.

knip.config.ts
DrJKL: (This is almost certainly a local dev setup issue)

src/core/graph/subgraph/promotedWidgetView.ts
artokun: Fixed in a7b8a94f0 — replaced manual duck-typing check with isPromotedWidgetView(w) type guard.

src/core/graph/subgraph/promotedWidgetView.ts
artokun: Fixed in a7b8a94f0 — replaced the fallback chain in drawConnections with a _widgetSlotsDirty flag check. The flag is set on widget add/remove and cleared after arrange().

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -188,13 +197,45 @@ class PromotedWidgetView implements IPromotedWidgetView {
...
+
+      // Fallback: match by widget locator name against our composite key
+      if (
+        input.widget?.name === `${this.sourceNodeId}:${this.sourceWidgetName}`
+      ) {
+        return slot
+      }
+    }
+    return undefined
+  }
```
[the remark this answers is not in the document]
artokun: Fixed in a7b8a94f0 — getBoundSubgraphSlot() now caches the slot reference with version-based invalidation, eliminating the ambiguous per-call O(N) lookup.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -4206,40 +4206,29 @@ export class LGraphNode
...
-      // For Vue positioning, just measure the slots without setting pos
-      for (const widget of this.widgets) {
-        const slot = slotByWidgetName.get(widget.name)
-        if (!slot) continue
-
-        this._measureSlot(this._concreteInputs[slot.index], slot.index, true)
-      }
+      const offset = LiteGraph.NODE_SLOT_HEIGHT * 0.5
+      slot.pos = [offset, widget.y + offset]
+      this._measureSlot(slot, i, true)
```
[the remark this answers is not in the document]
artokun: Fixed in a7b8a94f0 — removeWidget() now clears input.pos alongside input._widget and input.widget when removing a widget, and sets _widgetSlotsDirty = true so stale geometry is never used.

src/core/graph/subgraph/promotedWidgetView.ts
artokun: Fixed in a7b8a94f0 — added `_boundSlot` cache with version-based invalidation (keyed on `inputs.length`). `get name()` and `get label()` now read from the cache instead of doing per-frame O(N) scans.

src/core/graph/subgraph/promotedWidgetView.ts
artokun: Fixed in a7b8a94f0 — removed the unreachable composite-key fallback branch.

src/core/graph/subgraph/promotedWidgetView.ts
artokun: Fixed in a7b8a94f0 — replaced with `isPromotedWidgetView(w)` guard.

src/core/graph/subgraph/promotedWidgetView.ts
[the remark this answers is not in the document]
artokun: Fixed in a7b8a94f0 — overlaps with #1. The `_boundSlot` cache stores the reference directly, eliminating the ambiguous lookup. Only invalidates when `inputs.length` changes.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -5950,6 +5951,24 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
...
+    // (NodeSlots filters them out), so the fallback needs input.pos set here.
+    // TODO: Add a dirty flag (e.g. node._widgetSlotsArranged) to avoid the
+    // per-frame .some() scan once slots have been positioned.
+    for (const node of nodes) {
+      if (node.flags.collapsed || !node.widgets?.length) continue
+      if (!node.inputs?.some((inp) => isWidgetInputSlot(inp) && !inp.pos))
+        continue
+
+      node._setConcreteSlots()
+      node.arrange()
```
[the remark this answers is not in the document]
artokun: Fixed in a7b8a94f0 — overlaps with #5. The dirty flag approach (`_widgetSlotsDirty`) eliminates the need for `arrange()` in draw time entirely.

artokun:
Added a Playwright DOM assertion test in `browser_tests/tests/subgraph-promoted-widget-dom.spec.ts`. It enables Vue Nodes mode, converts KSampler to a subgraph, verifies the promoted seed widget's DOM bounding box is below the header, then renames the slot label and asserts the widget position hasn't jumped to the header area. This complements the existing canvas-side test in `subgraph-promoted-slot-position.spec.ts`.

src/core/graph/subgraph/promotedWidgetView.ts
artokun: yeah good catch, added a `_boundSlot` cache that invalidates when `inputs.length` changes. no more per-frame scanning from the getters

src/core/graph/subgraph/promotedWidgetView.ts
artokun: removed, you were right that composite key path was never reachable

knip.config.ts
artokun: In other words (i have no idea how to fix >.<)

src/core/graph/subgraph/promotedWidgetView.ts
artokun: switched to `isPromotedWidgetView(w)`, much cleaner

artokun:
@coderabbit can you review and either mark as approved or let me know what else needs a fixin

src/core/graph/subgraph/promotedWidgetView.ts
artokun: agreed the silent no-op was sketchy. setter now writes through to the bound slot's label so callers like `renameWidget` actually work

src/lib/litegraph/src/LGraphCanvas.ts
artokun: took your advice and went with the dirty flag approach. added `_widgetSlotsDirty` on LGraphNode, set it on widget add/remove, check + clear it in `drawConnections`. the `.some()` scan and `arrange()` calls in draw are gone now

src/core/graph/subgraph/promotedWidgetView.ts
artokun: simplified to just `this.identityName ?? this.sourceWidgetName`. the intermediate slot lookup was unnecessary since identityName is always passed in

src/utils/widgetUtil.ts
artokun: imported `ISubgraphInput` properly now, no more inline casting

artokun:
@coderabbit it still says requested changes can you approve it if good?

knip.config.ts
artokun: yeah this is a main issue not specific to this PR. `zod.gen.ts` is in knip's ignore list so it cant trace the zod import. adding it as an entry point was the quickest unblock, affects all fork PRs rebasing onto current main. i.e. I have no idea how to fix it plz help

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -352,7 +354,8 @@ function safeWidgetMapper(
           sourceNode && app.rootGraph
             ? (getExecutionIdByNode(app.rootGraph, sourceNode) ?? undefined)
             : undefined,
-        tooltip: widget.tooltip
+        tooltip: widget.tooltip,
+        promotedLabel: isPromotedWidgetView(widget) ? widget.label : undefined
```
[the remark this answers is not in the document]
artokun: added `tooltip?: string` to the interface, was just missing from the type

src/core/graph/subgraph/promotedWidgetView.ts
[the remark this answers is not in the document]
artokun: this is the same underlying issue as the hot-path scan above. the cached `_boundSlot` reference resolves both — no more ambiguous per-frame lookups

src/lib/litegraph/src/LGraphCanvas.ts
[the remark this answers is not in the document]
artokun: same fix as the dirty flag one above. no more `arrange()` in draw time, the `_widgetSlotsDirty` flag handles it

src/lib/litegraph/src/LGraphNode.ts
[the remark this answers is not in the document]
artokun: `removeWidget()` now clears `input.pos` too, along with setting the dirty flag. should prevent stale anchors sticking around after widget removal

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -27,6 +27,12 @@ import type { PromotedWidgetView as IPromotedWidgetView } from './promotedWidget
 export type { PromotedWidgetView } from './promotedWidgetTypes'
 export { isPromotedWidgetView } from './promotedWidgetTypes'
 
+interface SubgraphSlotRef {
+  name: string
+  label?: string
+  displayName?: string
+}
```
DrJKL, line 34: ### `SubgraphSlotRef` getter/setter asymmetry

The `label` getter (line 205) reads `slot.displayName ?? slot.name`, but `displayName` is never set on this ref, so this always returns `slot.name`. Meanwhile the `label` setter (line 212) writes `slot.label`, which is a property the getter never reads.

This means label renames done through the setter aren't visible through the getter. The rename flow works because `SubgraphNode` sets `input._widget.label = newName` directly (bypassing this getter), but that's fragile — anyone calling `view.label` after `view.label = 'x'` would get the old value.

**Suggestion:** Align the getter to read `slot.label ?? slot.displayName ?? slot.name`, or drop the unused `displayName` field.

```suggestion
interface SubgraphSlotRef {
  name: string
  label?: string
}
```

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -188,13 +201,46 @@ class PromotedWidgetView implements IPromotedWidgetView {
   }
 
   get label(): string | undefined {
-    const state = this.getWidgetState()
-    return state?.label ?? this.displayName ?? this.sourceWidgetName
+    const slot = this.getBoundSubgraphSlot()
+    if (slot) return slot.displayName ?? slot.name
+    return this.displayName
```
DrJKL: This returns `slot.displayName ?? slot.name` but `displayName` is never populated on `SubgraphSlotRef` (only `label` is written via the setter at line 212). After a rename, `slot.label` holds the new name, but the getter ignores it.

Should this be `slot.label ?? slot.displayName ?? slot.name`?

```suggestion
    if (slot) return slot.label ?? slot.displayName ?? slot.name
```

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -188,13 +201,46 @@ class PromotedWidgetView implements IPromotedWidgetView {
...
+   * Returns the cached bound subgraph slot reference, refreshing only when
+   * the subgraph node's input list has changed (version mismatch).
+   */
+  private getBoundSubgraphSlot(): SubgraphSlotRef | undefined {
+    const version = this.subgraphNode.inputs?.length ?? 0
+    if (this._boundSlotVersion === version) return this._boundSlot
+
+    this._boundSlot = this.findBoundSubgraphSlot()
+    this._boundSlotVersion = version
+    return this._boundSlot
```
DrJKL, line 235: ### Cache keyed on `inputs.length` — works by accident

The version is `this.subgraphNode.inputs?.length ?? 0`. When a rename changes `_subgraphSlot.label`, `inputs.length` doesn't change, so this returns the cached `_boundSlot`. It works only because the cached object is the **same reference** whose property was mutated in-place.

This is correct but accidental — if `findBoundSubgraphSlot()` ever returns a copy instead of the live reference, renames will silently break.

**Suggestion:** Add a comment documenting the reference-stability invariant, or use `_cacheVersion` from the parent `SubgraphNode` for explicit invalidation.

src/core/graph/subgraph/promotedWidgetView.ts
```diff
@@ -253,6 +302,7 @@ class PromotedWidgetView implements IPromotedWidgetView {
     projected.y = originalY
     projected.computedHeight = originalComputedHeight
     projected.computedDisabled = originalComputedDisabled
+    projected.label = originalLabel
```
DrJKL: ### Consider try/finally around projected widget save/restore

If `projected.drawWidget()` throws, the projected widget's `y`, `computedHeight`, `computedDisabled`, and `label` are left in the overridden state. The existing save/restore pattern (pre-dating this PR) wasn't wrapped in try/finally, but now that `label` has been added as a fourth property to save/restore, the risk surface is wider. Consider wrapping the block.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -295,6 +295,12 @@ export class LGraphNode
    */
   freeWidgetSpace?: number
 
+  /**
+   * Set to true when widget-backed input slot positions need recalculation.
+   * Cleared after arrange() runs. Avoids per-frame O(N) scans in drawConnections.
+   */
+  _widgetSlotsDirty = true
```
DrJKL: ### Every node starts dirty — first-frame cost

`_widgetSlotsDirty = true` means the `drawConnections` pre-pass will call `_setConcreteSlots()` + `arrange()` on **all** non-collapsed nodes on the first frame. For large graphs (hundreds of nodes, common in ComfyUI), this is a potentially noticeable cost.

Consider:
- Init to `false` and set dirty only in `addCustomWidget` / `removeWidget` (which already do this)
- Or gate the pre-pass on `node.inputs?.some(isWidgetInputSlot)` to skip nodes with no widget-input slots

DrJKL (commented):
## Multi-Angle PR Review

The core fix — separating identity name from display label — is the right approach and is well-tested. The `_arrangeWidgetInputSlots` rewrite is cleaner than the original, and the dirty flag pattern in `drawConnections` is sound.

A few items to address, mostly around the `SubgraphSlotRef` getter/setter asymmetry and naming proliferation (`name`, `label`, `displayName`, `identityName`, `slotName`, `inputName` across the promotion system).

Typecheck and lint both pass. Tests crash with STATUS_STACK_OVERFLOW (exit code 3221225725) on Windows — likely a pre-existing environment issue but worth verifying in CI.

src/lib/litegraph/src/subgraph/PromotedWidgetViewManager.ts
```diff
@@ -1,6 +1,9 @@
 import type { PromotedWidgetSource } from '@/core/graph/subgraph/promotedWidgetTypes'
 
-type ViewManagerEntry = PromotedWidgetSource & { viewKey?: string }
+type ViewManagerEntry = PromotedWidgetSource & {
+  viewKey?: string
+  slotName?: string
```
DrJKL: ### `slotName` unused by `PromotedWidgetViewManager`

`slotName` is added to the entry type but never referenced by any method in this class — it's only consumed by the `CreateView` callback passed in from outside. This widens the manager's API surface for a domain concept it doesn't own.

Consider passing `slotName` via the `CreateView` closure instead, keeping this type focused on what the manager actually needs (`sourceNodeId`, `sourceWidgetName`, `viewKey`).

src/utils/widgetUtil.ts
```diff
@@ -76,8 +78,18 @@ export function renameWidget(
...
+    const subgraphSlot = (input as Partial<ISubgraphInput>)._subgraphSlot
+    if (subgraphSlot) {
+      subgraphSlot.label = newLabel || undefined
+    }
   }
 
+  node.graph?.trigger('node:slot-label:changed', {
+    nodeId: node.id,
+    slotType: NodeSlotType.INPUT
+  })
```
DrJKL, line 92: ### `node:slot-label:changed` now fires for non-subgraph nodes

Previously this event was only dispatched from `SubgraphNode.renaming-input`. Now it fires unconditionally for every `renameWidget` call, including regular (non-subgraph) widgets.

Please verify that all listeners of `node:slot-label:changed` handle non-subgraph nodes gracefully. If any listener assumes `node` is a `SubgraphNode` (e.g. accesses `.subgraph`), this would break.

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+      await comfyPage.nextFrame()
+      await comfyPage.nextFrame()
+
+      const result = await getTextSlotPosition(comfyPage.page, '11')
+      expect(result).not.toBeNull()
+      expect(result!.hasPos).toBe(true)
+
+      // The slot Y position should be well below the title area.
+      // If it's near 0 or negative, the slot is stuck at the header (the bug).
+      expect(result!.posY).toBeGreaterThan(result!.titleHeight)
```
DrJKL, line 27: nit: Magic node ID `'11'` — a comment or named constant indicating which fixture node this refers to would help readability and prevent confusion if the fixture changes.

src/core/graph/subgraph/promotedWidgetView.ts
artokun: good eye, the getter was reading `displayName` which is never set on the ref. fixed to `slot.label ?? slot.displayName ?? slot.name` so it actually picks up what the setter writes

src/core/graph/subgraph/promotedWidgetView.ts
artokun: yep exactly, changed it to `slot.label ?? slot.displayName ?? slot.name`

src/core/graph/subgraph/promotedWidgetView.ts
artokun: you're right it works by coincidence since it's the same mutable ref. added a doc comment explaining this — the cache only needs to bust when slots are added/removed (length change), property mutations on the cached object are visible immediately since we hold the reference

src/core/graph/subgraph/promotedWidgetView.ts
artokun: wrapped it in try/finally now. if drawWidget throws the y, computedHeight, computedDisabled, and label all get restored properly

src/lib/litegraph/src/LGraphNode.ts
artokun: changed initial value to false. flag only gets set when `addCustomWidget` or `removeWidget` is called now, so the first frame doesn't pay the cost for nodes that already have their slots positioned

src/lib/litegraph/src/subgraph/PromotedWidgetViewManager.ts
artokun: agreed, made `reconcile()` generic over `TEntry extends ViewManagerEntry` so callers can pass extra fields without widening the base type. the manager itself only reads sourceNodeId/sourceWidgetName/viewKey

src/utils/widgetUtil.ts
artokun: checked all listeners — `useGraphNodeManager` does an early return if the nodeRef isn't found, so non-subgraph nodes are handled fine. firing it broadly is actually useful since the vue rendering layer uses it to refresh widget labels regardless of node type. added a comment noting this is intentional

browser_tests/tests/subgraph-promoted-slot-position.spec.ts
artokun: fair point, added a comment noting it's the SubgraphNode id from the fixture

jaeone94 (commented):
When switching from legacy to Vue nodes, slot positions are measured from the DOM before the ResizeObserver → layout-sync → CSS-update cycle has converged, causing link endpoints to render at stale (legacy) positions. A manual node resize fixes the issue because it re-triggers `syncNodeSlotLayoutsFromDOM` after the layout has settled.

I checked out this PR and attempted several approaches to fix this:
- Invalidating `_widgetSlotsDirty` on mode switch
- Setting `pendingSlotSync` with immediate `requestSlotLayoutSyncForAllNodes()`
- Nudging node sizes through the layout store to force ResizeObserver
- Listening for `LayoutSource.DOM` changes via `layoutStore.onChange`
- Adding an explicit `onLayoutSettled` event to the ResizeObserver pipeline

None of these worked reliably because the layout convergence requires multiple async cycles (ResizeObserver → layout-sync writeback → CSS update → DOM re-render → ResizeObserver again), and there is currently no single event that signals completion of this cycle.

The only approach that worked was a debounce pattern in `useVueNodeLifecycle.ts` — suppress link rendering with `pendingSlotSync` and defer the definitive slot resync until layout changes stop arriving:

```typescript
function schedulePostMountSlotResync() {
  const SETTLE_MS = 150
  let timer: ReturnType<typeof setTimeout> | null = null

  function doResync() {
    unsubscribe()
    timer = null
    const graph = comfyApp.canvas?.graph
    if (!graph) return

    layoutStore.clearAllSlotLayouts()
    for (const node of graph._nodes) {
      syncNodeSlotLayoutsFromDOM(String(node.id))
    }
    layoutStore.setPendingSlotSync(false)
    comfyApp.canvas?.setDirty(true, true)
  }

  const unsubscribe = layoutStore.onChange(() => {
    if (timer) clearTimeout(timer)
    timer = setTimeout(doResync, SETTLE_MS)
  })

  // Fallback: if no layout changes arrive (e.g. empty graph), clear flag.
  setTimeout(() => {
    if (timer) clearTimeout(timer)
    unsubscribe()
    if (layoutStore.pendingSlotSync) {
      layoutStore.setPendingSlotSync(false)
      comfyApp.canvas?.setDirty(true, true)
    }
  }, 3000)
}
```

With the mode-switch watch updated to:

```typescript
watch(
  () => shouldRenderVueNodes.value,
  (enabled) => {
    layoutStore.clearAllSlotLayouts()
    if (enabled) {
      layoutStore.setPendingSlotSync(true)
      schedulePostMountSlotResync()
    }
  }
)
```

This fixes the issue, but the debounce is a workaround — not an architectural solution. The root cause is that there is no way to know when the multi-cycle layout convergence (ResizeObserver → layout-sync → CSS → DOM → ResizeObserver …) has finished. A single "layout settled" event wouldn't work either, since the first ResizeObserver completion does not mean convergence. 
@DrJKL , do you have thoughts on a cleaner way to handle this, or is this an acceptable pragmatic fix for now?


---------------------------------------------------------

Separate topic — not a blocker for this PR, but worth discussing:

When renaming a promoted widget via the RightSidePanel, the rename currently propagates to the input slot names of the interior nodes connected to that promoted widget inside the subgraph. I understand this is existing behavior on `main` (and previously the promoted widget label itself didn't update at all), but it seems like the rename should only affect the promoted widget's display label — the internal node slot names are an implementation detail and shouldn't change when the user renames the externally-facing label. Would it make sense to decouple these in a follow-up, while this area is being actively worked on?

jaeone94:
~~Correction: OMG I'm still experiencing issues where the debounce pattern doesn't work intermittently... Is this happening to anyone else?~~
I confirmed that the legacy→Vue mode switch link position issue also reproduces on v1.41.19, 
so this is a pre-existing bug unrelated to this PR.

artokun:
This bothers me also but it's outside of the scope of this particular fix. The two others you mentioned, the internal input node rename is a preexisting issue, and the LG to Vue handle to connector re-calculation can also be a separate ticket. I could open two separate follow on tickets (or you are also welcome to as you already provided a solution). My concern is adding another watcher when we really need just an event that propagates once whenever we switch modes.

jaeone94:
I had some time and experimented with a few of the issues I reported in a separate PR: https://github.com/Comfy-Org/ComfyUI_frontend/pull/10395

It includes the same core changes from this PR along with some additional fixes, so feel free to cherry-pick anything that looks useful — no pressure at all.

I agree that the legacy-to-Vue mode switch link breakage is an edge case and a pre-existing issue, so it makes sense to keep it out of scope here. That said, the broken link endpoints on draft workflow restore feels like a more common scenario that many users would hit, and it seems to fit naturally with this PR's intent. If you think it's worth addressing here, the approach I took was to explicitly reset the Vue node lifecycle after `initializeWorkflow` completes — happy to discuss if that's helpful.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -4211,40 +4211,28 @@ export class LGraphNode
...
-      // For Vue positioning, just measure the slots without setting pos
-      for (const widget of this.widgets) {
-        const slot = slotByWidgetName.get(widget.name)
-        if (!slot) continue
-
-        this._measureSlot(this._concreteInputs[slot.index], slot.index, true)
-      }
+      const offset = LiteGraph.NODE_SLOT_HEIGHT * 0.5
+      slot.pos = [offset, widget.y + offset]
+      this._measureSlot(slot, i, true)
```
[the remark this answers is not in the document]
artokun: Already addressed — the lookup order is now `slot._widget ?? widgetByName.get(slot.widget.name)`, preferring the direct binding first. Comment also documents the ambiguity risk with the name-map fallback.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: Correction on my earlier reply — after more iteration I landed on the opposite approach: the `renaming-input` handler now explicitly does **not** touch `input.widget.name`. It only sets `input.label` and `input._widget.label` for display. `widget.name` stays as `subgraphInput.name` (the stable internal key), so `onGraphConfigured` in widgetInputs.ts always resolves correctly. Added a comment in the handler explaining the constraint and tests proving the name stays stable after rename.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: Revisited again — turns out you were right, label cascade *is* enough. The final approach only sets `input.label` and `input._widget.label` in the rename handler, leaving `input.widget.name` untouched. `_setWidget` assigns `input.widget.name = subgraphInput.name` (the stable internal key) on initial setup, and it never changes after that. Display is fully driven through `label`.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
artokun: Update — ended up not needing `displayWidgetName` after all. The final approach keeps `input.widget.name = subgraphInput.name` (stable internal key) and uses `label` exclusively for display. So there's no semantic shift on `IWidgetLocator.name` — it stays as the internal identifier for all node types including SubgraphNode. The cache key variant (`input.label ?? input.name`) is still intentional for nested subgraph remapping, added a comment on that.

artokun:
@jaeone94 cherry-picked two of your fixes — the arrange() force on Vue-to-legacy switch and the re-extract on label change. Both applied cleanly and have Co-Authored-By credit. The draft restore debounce is a good catch too but needs a separate cherry-pick since it's in a larger commit - will follow up on that after this lands. Thanks!

knip.config.ts
DrJKL: This might be a line endings issue.
```sh
git add --renormalize .
```

src/composables/graph/useVueNodeLifecycle.ts
```diff
@@ -86,6 +86,21 @@ function useVueNodeLifecycleIndividual() {
...
+      const graph = comfyApp.canvas?.graph
+      if (graph) {
+        for (const node of graph._nodes) {
+          try {
+            if (!node.flags.collapsed) node.arrange()
+          } catch {
+            /* skip nodes not fully initialized */
+          }
+        }
+      }
```
DrJKL, line 105: Very nit, but I like to flatten when possible.
I'm also wondering if the errors from nodes not being initialized at this point could be a synchronization issue down the road. Maybe we need an explicit readiness gate?
```suggestion
      const graph = comfyApp.canvas?.graph
      if (!graph) {
        comfyApp.canvas?.setDirty(true, true)
        return
      }
      for (const node of graph._nodes) {
        if (node.flags.collapsed) continue
        try {
          node.arrange()
        } catch {
          /* skip nodes not fully initialized */
        }
      }
```

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL: Semantically, what's the difference between label and displayName?

src/composables/graph/useVueNodeLifecycle.ts
artokun: Good call on the flattening — applied. Re: the readiness gate, agreed this is papering over a synchronization gap. The root cause is that `drawConnections` can fire before any `drawNode` pass has run `arrange()`. A proper fix would be a node-level dirty flag (e.g. `_slotLayoutDirty`) that `arrange()` clears and `drawConnections` checks, but that's a bigger refactor. Filed as a follow-up concern.

src/core/graph/subgraph/promotedWidgetView.ts
DrJKL, line 34: Optional:
```suggestion
interface SubgraphSlotRef {
  readonly name: string
  label?: string
  readonly displayName?: string
}
```

DrJKL: Though I wonder if we can reuse a different interface and just narrow it...

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -295,6 +295,12 @@ export class LGraphNode
    */
   freeWidgetSpace?: number
 
+  /**
+   * Set to true when widget-backed input slot positions need recalculation.
+   * Cleared after arrange() runs. Avoids per-frame O(N) scans in drawConnections.
+   */
+  _widgetSlotsDirty = false
```
DrJKL, line 302: Maybe?
```suggestion
  private _widgetSlotsDirty = false
```

src/utils/widgetUtil.ts
```diff
@@ -76,8 +78,19 @@ export function renameWidget(
   widget.label = newLabel || undefined
   if (input) {
     input.label = newLabel || undefined
+
+    const subgraphSlot = (input as Partial<ISubgraphInput>)._subgraphSlot
```
DrJKL, line 82: Any way we can avoid the cast? Are we missing a type guard?

src/utils/widgetUtil.test.ts
```diff
@@ -128,6 +128,21 @@ describe('renameWidget', () => {
     expect(promotedWidget.label).toBe('Renamed')
   })
 
+  it('updates _subgraphSlot.label when input has a subgraph slot', () => {
+    const widget = makeWidget({ name: 'seed' })
+    const subgraphSlot = { label: undefined as string | undefined }
```
DrJKL, line 133: Maybe?
```suggestion
    const subgraphSlot: Partial<ISubgraphInput> = { label: undefined }
```

DrJKL (dismissed):

DrJKL:
Only other thing I'd love to have would be a full Playwright test that goes through the slot renaming start to end.

artokun:
Added in the latest push — `subgraphInputSlotRename.spec.ts` covers the full flow:

1. Load workflow with subgraph (uses a new `test-values-input-subgraph.json` fixture you provided)
2. Verify seed widget visible on parent node (Vue DOM assertion)
3. Verify widget is in body, not header (bounding box check)
4. Enter subgraph, find seed input slot, right-click → Rename Slot → "my_seed"
5. Exit back to parent
6. Verify widget label updated to "my_seed"
7. Verify widget still in body position

Uses Vue nodes for DOM assertions (steps 1-2, 6-7) and litegraph mode for the rename dialog (steps 3-5) since `canvas.prompt()` needs the litegraph context menu. Passes locally alongside the other 3 subgraph tests.

artokun:
Heads up — just pushed a fix for the `appModeWidgetRename` test failures. The `PromotedWidgetView.label` getter was refactored to read from the bound subgraph slot, but I forgot to keep the `widgetValueStore` fallback that app mode relies on for label persistence across save/reload. Been so heads down in the subgraph internals I completely missed the app mode rename flow. Fixed now — label getter falls back to the persisted widget state store before the displayName fallback, and the setter writes to both. All 4 app mode rename tests pass locally.

DrJKL (approved):

PR #10196, opened 2026-03-17
fix: support progressive fallback for deeply nested model directories

deepme987:
## Summary
- Fix `findProvidersWithFallback` to try all parent paths progressively instead of jumping from exact match to top-level segment only. Previously `"a/b/c"` would try `"a/b/c"` then `"a"`, now correctly tries `"a/b/c"` → `"a/b"` → `"a"`
- Use empty key `''` for `DownloadAndLoadCogVideoControlNet` and `DownloadAndLoadCogVideoModel` registrations so `shouldUseAssetBrowser` returns false — these nodes use HuggingFace repo names, not file-based assets, so the asset browser finds nothing and shows an empty dropdown

## Test plan
- [x] All 46 existing `modelToNodeStore` tests pass
- [ ] Verify CogVideo ControlNet and HF model dropdowns show options (after backend deploy)
- [x] Verify "Use" button on CogVideo/ControlNet/* models creates the correct node
- [ ] Verify GGUF asset browser still works


src/stores/modelToNodeStore.ts
```diff
@@ -88,12 +88,13 @@ export const useModelToNodeStore = defineStore('modelToNode', () => {
...
-    const fallback = modelToNodeMap.value[topLevel]
-
-    if (fallback && fallback.length > 0) return fallback
+    // Try progressively shorter parent paths
+    const segments = modelType.split('/')
+    for (let i = segments.length - 1; i >= 1; i--) {
+      const parentPath = segments.slice(0, i).join('/')
+      const fallback = modelToNodeMap.value[parentPath]
+      if (fallback && fallback.length > 0) return fallback
+    }
```
DrJKL: I'd love like 3 unit tests for this (or 1 parameterized one)

src/stores/modelToNodeStore.ts
```diff
@@ -88,12 +88,13 @@ export const useModelToNodeStore = defineStore('modelToNode', () => {
     const exactMatch = modelToNodeMap.value[modelType]
     if (exactMatch && exactMatch.length > 0) return exactMatch
```
DrJKL: If you just start at segments.length, you'll get the same behavior, right?

DrJKL (dismissed):

DrJKL (approved):

PR #10201, opened 2026-03-17
fix: wrap configure() widget restoration in hydration transaction

christian-byrne:
## Summary

Wraps `LGraphNode.configure()`'s widget-value restoration phase in a hydration transaction (`beginHydration` / `commitHydration`), preventing derived-state callbacks (e.g. CustomCombo's `updateCombo`) from firing before all widget values are restored.

### Architecture

```mermaid
flowchart LR
    subgraph core["Core Layer"]
        LGN["LGraphNode.configure()"]
        WVS["widgetValueStore"]
    end

    subgraph extensions["Extension Layer"]
        CW["CustomCombo\n(customWidgets.ts)"]
        PN["PrimitiveNode\n(widgetInputs.ts)"]
    end

    LGN -- "beginHydration /\ncommitHydration" --> WVS
    CW -- "isHydrating() guard" --> WVS
    CW -- "onHydrationComplete()" --> WVS
    PN -- "serialize() override\npreserves widgets_values" --> LGN
```

### Hydration Sequence

```mermaid
sequenceDiagram
    participant Caller as Paste / Load
    participant Node as LGraphNode.configure()
    participant Store as widgetValueStore
    participant Widget as Widget Setters
    participant Combo as CustomCombo.updateCombo

    Caller->>Node: configure(serializedData)
    Node->>Store: beginHydration(nodeId)
    activate Store
    Note over Store: hydratingNodes.add(nodeId)

    loop For each widget
        Node->>Widget: widget.value = restored_value
        Widget->>Combo: updateCombo() triggered
        Combo->>Store: isHydrating(nodeId)?
        Store-->>Combo: true - early return
        Note over Combo: Side effect suppressed
    end

    Node->>Node: onConfigure(info)
    Note over Node: Extensions register onHydrationComplete callbacks

    Node->>Store: commitHydration(nodeId)
    Note over Store: hydratingNodes.delete(nodeId)
    Store->>Combo: fire queued callbacks
    Note over Combo: updateCombo() runs once with all values present
    deactivate Store
```

### Before / After

```mermaid
flowchart TB
    subgraph before["Before: Race Condition"]
        direction TB
        B1["configure starts"] --> B2["widget1.value = optionA"]
        B2 --> B3["updateCombo fires immediately"]
        B3 --> B4["values list incomplete,\ncomboWidget.value reset"]
        B4 --> B5["widget2.value = optionB"]
        B5 --> B6["updateCombo fires again"]
        B6 --> B7["Combo has wrong value"]
    end

    subgraph after["After: Hydration Transaction"]
        direction TB
        A1["configure starts"] --> A2["beginHydration nodeId"]
        A2 --> A3["widget1.value = optionA"]
        A3 --> A4["updateCombo skipped"]
        A4 --> A5["widget2.value = optionB"]
        A5 --> A6["updateCombo skipped"]
        A6 --> A7["commitHydration nodeId"]
        A7 --> A8["updateCombo runs once,\nall values present"]
    end
```

### Changes

- **`LGraphNode.configure()`**: Wraps widget restoration + `onConfigure` in `beginHydration`/`commitHydration` with `try/finally` for exception safety
- **`PrimitiveNode.serialize()`**: Adds override to preserve `widgets_values` when widgets are dynamically disconnected
- **`customWidgets.ts`**: Simplifies CustomCombo's `onConfigure` to use `onHydrationComplete()` instead of manually managing hydration state
- **3 new tests**: Hydration transaction wrapping, `onHydrationComplete` callback firing, and exception safety

### Part of

Stacked on #10010. Implements Design A from the widget state architecture RFC.

### DIFF
--- a/src/extensions/core/customWidgets.ts
+++ b/src/extensions/core/customWidgets.ts
@@ -88,12 +88,12 @@ function onCustomComboCreated(this: LGraphNode) {
         )?.value
       },
       set(v: string) {
-        const state = useWidgetValueStore().getWidget(
+        useWidgetValueStore().getOrCreateWidget(
           app.rootGraph.id,
           node.id,
-          widgetName
-        )
-        if (state) state.value = v
+          widgetName,
+          v
+        ).value = v
         updateCombo()
         if (!node.widgets) return
         const lastWidget = node.widgets.at(-1)
--- a/src/extensions/core/customWidgets.ts
+++ b/src/extensions/core/customWidgets.ts
@@ -91,13 +91,6 @@ function onCustomComboCreated(this: LGraphNode) {
       set(v: string) {
         const store = useWidgetValueStore()
 
-        if (
-          !store.isHydrating(node.id) &&
-          !store.getWidget(app.rootGraph.id, node.id, widgetName)
-        ) {
-          store.beginHydration(node.id)
-        }
-
         store.getOrCreateWidget(
           app.rootGraph.id,
           node.id,
@@ -139,11 +132,7 @@ function onCustomComboCreated(this: LGraphNode) {
   this.onConfigure = useChainCallback(
     this.onConfigure,
     function (this: LGraphNode) {
-      const store = useWidgetValueStore()
-      if (store.isHydrating(this.id)) {
-        store.commitHydration(this.id)
-        updateCombo()
-      }
+      useWidgetValueStore().onHydrationComplete(this.id, updateCombo)
     }
   )
 }
--- a/src/extensions/core/customWidgets.ts
+++ b/src/extensions/core/customWidgets.ts
@@ -64,6 +64,7 @@ function onCustomComboCreated(this: LGraphNode) {
       ).map((w) => `${w.value}`)
     )
     if (app.configuringGraph) return
+    if (useWidgetValueStore().isHydrating(this.id)) return
     if (values.includes(`${comboWidget.value}`)) return
     comboWidget.value = values[0] ?? ''
     comboWidget.callback?.(comboWidget.value)
@@ -88,12 +89,24 @@ function onCustomComboCreated(this: LGraphNode) {
         )?.value
       },
       set(v: string) {
-        useWidgetValueStore().getOrCreateWidget(
+        const store = useWidgetValueStore()
+
+        if (
+          !store.isHydrating(node.id) &&
+          !store.getWidget(app.rootGraph.id, node.id, widgetName)
+        ) {
+          store.beginHydration(node.id)
+        }
+
+        store.getOrCreateWidget(
           app.rootGraph.id,
           node.id,
           widgetName,
           v
         ).value = v
+
+        if (store.isHydrating(node.id)) return
+
         updateCombo()
         if (!node.widgets) return
         const lastWidget = node.widgets.at(-1)
@@ -122,6 +135,17 @@ function onCustomComboCreated(this: LGraphNode) {
     y: 0
   })
   addOption(this)
+
+  this.onConfigure = useChainCallback(
+    this.onConfigure,
+    function (this: LGraphNode) {
+      const store = useWidgetValueStore()
+      if (store.isHydrating(this.id)) {
+        store.commitHydration(this.id)
+        updateCombo()
+      }
+    }
+  )
 }
 
 function onCustomIntCreated(this: LGraphNode) {
--- a/src/lib/litegraph/src/LGraphNode.ts
+++ b/src/lib/litegraph/src/LGraphNode.ts
@@ -8,6 +8,9 @@ import {
 import type { SlotPositionContext } from '@/renderer/core/canvas/litegraph/slotCalculations'
 import { useLayoutMutations } from '@/renderer/core/layout/operations/layoutMutations'
 import { LayoutSource } from '@/renderer/core/layout/types'
+import { getActivePinia } from 'pinia'
+
+import { useWidgetValueStore } from '@/stores/widgetValueStore'
 import { adjustColor } from '@/utils/colorUtil'
 import type { ColorAdjustOptions } from '@/utils/colorUtil'
 import {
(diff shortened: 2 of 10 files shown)


christian-byrne:
## Alignment with ECS ADR (#10420)

cc @DrJKL — this PR implements a concrete step toward the ECS architecture you proposed. Here's how it maps:

### ✅ Strong alignment

| ECS Principle (#10420) | This PR |
|---|---|
| `widgetValueStore` is "the closest thing to a true ECS component store" | Extends it with `beginHydration`/`commitHydration`/`isHydrating`/`onHydrationComplete` — adding lifecycle management to the store, not to the god object |
| "Scattered side effects" (§6 in entity-problems) | Directly solves this for widget callbacks — instead of side effects firing mid-configure, one `commitHydration` batches them |
| "Serialization interleaved with container logic" is a pain point | The hydration transaction creates a clean boundary between *restoration* and *side effects*, exactly the separation ECS wants between data writes and system reactions |
| `WidgetState` as plain data | The new `getOrCreateWidget()` returns the same plain `WidgetState` objects — no methods, no back-references |

### ⚠️ Minor tension

| ECS Principle | Tension |
|---|---|
| "Domain objects should not have direct dependencies on UI framework stores" | `LGraphNode.configure()` now imports `useWidgetValueStore` and `getActivePinia`. The ADR flags this pattern as problematic. In the ECS future, a `SerializationSystem` would own this. |
| Hydration state lives in the store, not as a component | `hydratingNodes` is a `Set<NodeId>` inside the store — in a full ECS, this would be a transient component on the node entity. Pragmatic given the incremental migration strategy. |

### Summary

This is the kind of **incremental extraction** that #10420's migration strategy step 4 calls for: extract one concern (hydration lifecycle) from the class into the centralized store, without breaking existing code. It strengthens `widgetValueStore` as the proto-ECS component store that #10420 already identifies as the most ECS-aligned in the codebase.

src/stores/widgetValueStore.ts
```diff
@@ -76,14 +80,68 @@ export const useWidgetValueStore = defineStore('widgetValue', () => {
...
+
+    const state: WidgetState = {
+      nodeId,
+      name: widgetName,
+      type: 'string',
+      value: defaultValue,
+      options: {}
+    }
+    widgetStates.set(key, state)
+    return widgetStates.get(key) as WidgetState
```
DrJKL: Alternatively, we could wrap this in some form of assertExists... 😉 
```suggestion
    return widgetStates.get(key)!
```

src/stores/widgetValueStore.test.ts
```diff
@@ -155,6 +155,95 @@ describe('useWidgetValueStore', () => {
...
+      store.onHydrationComplete('node-1', () => calls.push('immediate'))
+      expect(calls).toEqual(['immediate'])
+    })
+
+    it('commitHydration is safe to call when not hydrating', () => {
+      const store = useWidgetValueStore()
+      expect(() => store.commitHydration('node-1')).not.toThrow()
+    })
+
+    it('hydration is node-scoped — independent per node', () => {
```
DrJKL, line 232: Wouldn't idempotency checks mean multiple beginHydration calls?

src/stores/widgetValueStore.ts
```diff
@@ -76,14 +80,68 @@ export const useWidgetValueStore = defineStore('widgetValue', () => {
...
+  function isHydrating(nodeId: NodeId): boolean {
+    return hydratingNodes.has(nodeId)
+  }
+
+  function onHydrationComplete(nodeId: NodeId, callback: HydrationCallback) {
+    if (!hydratingNodes.has(nodeId)) {
+      callback()
+      return
+    }
+    const existing = hydrationCallbacks.get(nodeId) ?? []
```
DrJKL, line 127: So we could register the same callback for a node multiple times here, right?

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -891,44 +894,54 @@ export class LGraphNode
...
 
-    if (this.widgets) {
-      for (const w of this.widgets) {
-        if (!w) continue
+    // Hydration transaction: suppress derived-state callbacks (e.g.
+    // CustomCombo's updateCombo) until all widget values are restored.
+    // onConfigure handlers may commit early (CustomCombo does); the
+    // final commitHydration is idempotent in that case.
+    const store = getActivePinia() ? useWidgetValueStore() : null
+    store?.beginHydration(this.id)
```
DrJKL, line 902: Do you know when we'd be constructing Node instances before we'd setup the stores? Or is this just defensive?

src/extensions/core/customWidgets.ts
```diff
@@ -122,6 +128,13 @@ function onCustomComboCreated(this: LGraphNode) {
     y: 0
   })
   addOption(this)
+
+  this.onConfigure = useChainCallback(
+    this.onConfigure,
+    function (this: LGraphNode) {
+      useWidgetValueStore().onHydrationComplete(this.id, updateCombo)
+    }
+  )
```
DrJKL, line 137: I know this is the current pattern, but it makes me sad every time.

src/stores/widgetValueStore.ts
```diff
@@ -76,14 +80,68 @@ export const useWidgetValueStore = defineStore('widgetValue', () => {
...
+    hydratingNodes.add(nodeId)
+  }
+
+  function commitHydration(nodeId: NodeId): void {
+    hydratingNodes.delete(nodeId)
+    const callbacks = hydrationCallbacks.get(nodeId)
+    if (callbacks) {
+      hydrationCallbacks.delete(nodeId)
+      for (const cb of callbacks) cb()
+    }
```
DrJKL: Very nit, but if you flip this into a guard you can reduce the nesting by 1.

DrJKL (approved):

christian-byrne:
## Architectural Audit vs ECS ADR (#10420)

Audited this PR against the ECS ADR. **TL;DR: the hydration transaction concept is ECS-aligned; the integration seams are known debt that will migrate cleanly when systems land.**

### What's good (keep as-is)
- `beginHydration`/`commitHydration`/`isHydrating` in `widgetValueStore` — structurally identical to what a `HydrationSystem` would do
- Node-scoped, idempotent, `try/finally` — right granularity and safety

### Known debt (tracked in #10529)
1. **`getActivePinia()` in `LGraphNode.configure()`** — domain object reaching into Pinia. When `SerializationSystem` exists, it wraps `configure()` externally instead
2. **`useChainCallback(this.onConfigure, ...)` in `customWidgets.ts`** — deepens callback chaining where ADR wants systems. Post-hydration reconciliation will become a system responsibility
3. **Callback queue in widgetValueStore** — the store is drifting toward store+event bus. Queue moves to `HydrationSystem` later

### Why not rework now
- No unified paste/load layer exists yet to move the wrapping to
- Building a bespoke coordinator now creates a throwaway abstraction that the `SerializationSystem` replaces
- The debt is isolated to 2 files, ~10 lines, and trivially deletable when systems land

Full analysis and migration plan: #10529

PR #10203, opened 2026-03-17
fix: replace stale-request guard with single-flight coalescing in queueStore.update()

christian-byrne:
## Summary

`queueStore.update()` previously used a request-ID guard that discarded responses whenever a newer call had been initiated. During concurrent job execution, rapid WebSocket events (`status`, `execution_success`, etc.) created a sustained stream of `update()` calls where:

1. **Every response got discarded** — each returning response saw a higher `updateRequestId`, so data was never applied to the store until the event burst subsided
2. **All HTTP requests still fired** — the guard didn't suppress outgoing calls, just discarded stale responses after the fact

This caused the UI to freeze showing stale job states (e.g. completed jobs still showing as "Running") during bursts, then snap all at once when the storm ended.

## Changes

Replace the request-ID guard with a single-flight coalescing pattern:

- **At most one fetch in flight** at a time — no request spam
- **Every response is applied** — no UI starvation
- **Dirty flag** triggers one re-fetch if calls arrived during flight — eventual consistency

## Testing

Added 5 unit tests covering:
- Concurrent calls coalesce into a single re-fetch
- No response starvation (every response applied)
- No duplicate requests when no concurrent calls
- Loading state transitions through coalesced re-fetches
- Normal sequential behavior preserved

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10203-fix-replace-stale-request-guard-with-single-flight-coalescing-in-queueStore-update-3266d73d365081088656e4e55ca4dbd3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/queueStore.ts
+++ b/src/stores/queueStore.ts
@@ -594,7 +594,6 @@ export const useQueueStore = defineStore('queue', () => {
       isLoading.value = false
       inFlight = false
       if (dirty) {
-        dirty = false
         void update()
       }
     }
--- a/src/stores/queueStore.ts
+++ b/src/stores/queueStore.ts
@@ -488,8 +488,13 @@ export const useQueueStore = defineStore('queue', () => {
   const maxHistoryItems = ref(64)
   const isLoading = ref(false)
 
-  // Scoped per-store instance; incremented to dedupe concurrent update() calls
-  let updateRequestId = 0
+  // Single-flight coalescing: at most one fetch in flight at a time.
+  // If update() is called while a fetch is running, the call is coalesced
+  // and a single re-fetch fires after the current one completes.
+  // This prevents both request spam and UI starvation (where a rapid stream
+  // of calls causes every response to be discarded by a stale-request guard).
+  let inFlight = false
+  let dirty = false
 
   const tasks = computed<TaskItemImpl[]>(
     () =>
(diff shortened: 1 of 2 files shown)


src/stores/queueStore.ts
```diff
@@ -488,8 +488,13 @@ export const useQueueStore = defineStore('queue', () => {
...
   const isLoading = ref(false)
 
-  // Scoped per-store instance; incremented to dedupe concurrent update() calls
-  let updateRequestId = 0
+  // Single-flight coalescing: at most one fetch in flight at a time.
+  // If update() is called while a fetch is running, the call is coalesced
+  // and a single re-fetch fires after the current one completes.
+  // This prevents both request spam and UI starvation (where a rapid stream
+  // of calls causes every response to be discarded by a stale-request guard).
+  let inFlight = false
```
DrJKL, line 496: When would `inFlight` and `isLoading.value` be different?

DrJKL (dismissed):

src/stores/queueStore.ts
```diff
@@ -582,11 +591,11 @@ export const useQueueStore = defineStore('queue', () => {
...
-      // Only clear loading if this is the latest request.
-      // A stale request completing (success or error) should not touch loading state
-      // since a newer request is responsible for it.
-      if (requestId === updateRequestId) {
-        isLoading.value = false
+      isLoading.value = false
+      inFlight = false
+      if (dirty) {
+        dirty = false
+        void update()
```
DrJKL, line 597: I think you could also cut the dirty = false here, since it'd be reset if inFlight is false when you re-trigger update.

src/stores/queueStore.ts
christian-byrne: They're semantically different things that happen to be in sync today:

- `inFlight` is a plain `let` — a synchronous concurrency mutex ("is a fetch currently running?")
- `isLoading` is a reactive `ref` — exposed to the UI for spinners, skeleton states, etc.

Merging them into just `isLoading.value` would work, but it couples concurrency control to a reactive UI ref. If we ever want them to diverge (e.g. keep the spinner visible across a coalesced re-fetch chain), we'd have to re-separate them. The extra `let` is cheap and keeps the intent clear, so I'll leave them as-is.

DrJKL: YAGNI

christian-byrne: :laughing:

src/stores/queueStore.ts
christian-byrne: Done, damn you're good.

PR #10244, opened 2026-03-18
fix: App mode - handle socket/response race when tracking jobs

pythongosssss:
## Summary

Improves the handling of the job tracking where the websocket vs http response are processed in a non-deterministic order

## Changes

- **What**: Update onJobStart watch and guard to watch both the activeJobId and the path mapping to ensure both trigger a start, but do not double start

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10244-fix-App-mode-handle-socket-response-race-when-tracking-jobs-3276d73d365081f2862bd6cdcce3aac7) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/linearMode/linearOutputStore.test.ts
+++ b/src/renderer/extensions/linearMode/linearOutputStore.test.ts
@@ -1312,4 +1312,96 @@ describe('linearOutputStore', () => {
       ).toHaveLength(2)
     })
   })
+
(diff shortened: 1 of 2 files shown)


pythongosssss:
@coderabbitai this is not a good candidate for e2e testing as it requires a race condition when executing jobs, I have added additional tests to cover this race condition

AustinMroz (approved):

Issue #8173, opened 2026-01-20
Switching workflows while inside a subgraph and switching back can return the canvas view to the wrong offset/scale


GrandpaGameHacker:
### Prerequisites

- [x] I am running the latest version of ComfyUI
- [x] I have searched existing issues to make sure this isn't a duplicate
- [x] I have tested with all custom nodes disabled ([see how](https://docs.comfy.org/troubleshooting/custom-node-issues#step-1%3A-test-with-all-custom-nodes-disabled))

### What happened?

Whenever you are working with subgraphs and multiple workflows sometimes when switching back and forth, and entering and exiting different subgraphs, the view of a subgraph will have an incorrect offset and or scale, forcing the user to use the `Fit view (.)` or mini-map functionality to refocus the graph correctly. This happens quite often.

I am not too familiar with the codebase but my theory at a glance is that the `ChangeTracker` class and the `SubgraphNavigationStore` do not handle swapping workflows correctly in regards to subgraphs.

### Steps to Reproduce

1. Create two or more workflows, at least two with subgraphs.
2. Enter a subgraph and swap to other workflows and enter their subgraphs, exit them, any order of combinations until step (3) causes the bug.
3. Swap back to the first subgraph workflow, sometimes the offset and scale for the canvas is incorrect to what it should have been before, either being too zoomed in or completely at the wrong offset.

### How is this affecting you?

Visual/UI issue only

### ComfyUI Frontend Version

1.36.14

### Browser

Chrome/Chromium

### Console Errors

```javascript

```

### Logs

```shell

```

### Additional Context

OS - Windows 11 & Windows 10
NVIDIA 3090 / 3060
Chromium browser (specifically Opera Gx)

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-8173-Switching-workflows-while-inside-a-subgraph-and-switching-back-can-return-the-canvas-v-2ee6d73d365081e8ababd5ea870c9a8f) by [Unito](https://www.unito.io)

Tekaiguy:
I have noticed this too, at least upon initial creation of a subgraph, or after restarting the comfyui session. 'Fit view' is almost always necessary to locate the nodes inside the subgraph to the point where it might as well be automated. However, if the subgraph is more complex (larger) then applying 'fit view' every time could be undesirable so I don't know what the solution is.

[...]

PR #10247, opened 2026-03-18
fix: persist subgraph viewport across navigation and tab switches

artokun:
## Summary

Fix subgraph viewport (zoom + position) drifting when navigating in/out of subgraphs and switching workflow tabs.

## Problem

Three root causes:
1. **First visit**: `restoreViewport()` silently returned on cache miss, leaving canvas at stale position
2. **Cross-workflow leakage**: Cache keyed by bare `graphId` — two workflows with the same subgraph or unsaved workflows shared cache entries
3. **Stale save on tab switch**: `loadGraphData` and `changeTracker.restore()` overwrite `canvas.ds` before the async watcher could save the old viewport

## Solution

1. **Workflow-scoped cache keys**: `${path}#${instanceId}:${graphId}` — WeakMap assigns unique IDs per workflow object, handling unsaved workflows with identical paths
2. **`flush: 'sync'` on activeSubgraph watcher**: Fires immediately during `setGraph()`, BEFORE `loadGraphData`/`changeTracker` can corrupt `canvas.ds`
3. **Cache miss → rAF fitToBounds**: On first visit, computes bounds from `graph._nodes` and calls `ds.fitToBounds()` after the browser has rendered
4. **Workflow switch watcher** (`flush: 'sync'`): Pre-saves viewport under old workflow identity, suppresses `onNavigated` saves during load cycle

Key architectural insight: `setGraph()` never touches `canvas.ds`, but `loadGraphData` and `changeTracker.restore()` both write to it. By using `flush: 'sync'`, the save happens during `setGraph` (before the overwrites).

## Review Focus

- `subgraphNavigationStore.ts` — the three fixes and their interaction
- `flush: 'sync'` watchers — critical for correct save timing
- `suppressNavigatedSave` flag — prevents stale saves during async workflow load

## Breaking Changes

None. Viewport cache is session-only (in-memory LRU). Existing workflows unaffected.

## Demo Video of Fix

https://github.com/user-attachments/assets/71dd4107-a030-4e68-aa11-47fe00101b25

## Test plan

- [x] Unit: save/restore with workflow-scoped keys
- [x] Unit: cache miss doesn't mutate canvas synchronously
- [x] Unit: navigation integration (enter/exit preserves viewport)
- [x] E2E: first subgraph visit has visible nodes
- [x] Manual: enter subgraph → zoom/pan → exit → re-enter → viewport restored
- [x] Manual: tab with subgraph → different tab → back → viewport restored
- [x] Manual: two unsaved workflows → switch between → viewports isolated

- Fixes #10246
- Related: #8173

---
## 🔍 Automated QA Report

| | |
|---|---|
| **Status** | ✅ Complete |
| **Report** | [sno-qa-10247.comfy-qa.pages.dev](https://sno-qa-10247.comfy-qa.pages.dev/) |
| **CI Run** | [View workflow](https://github.com/Comfy-Org/ComfyUI_frontend/actions/runs/23373279990) |

Before/after video recordings with **Behavior Changes** and **Timeline Comparison** tables.

### DIFF
--- a/browser_tests/tests/subgraphViewport.spec.ts
+++ b/browser_tests/tests/subgraphViewport.spec.ts
@@ -37,23 +37,22 @@ test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
     )
     await comfyPage.nextFrame()
 
-    const result = await comfyPage.page.evaluate(() => {
-      const graph = window.app!.canvas.graph!
-      const sgNode = graph._nodes.find(
-        (n: { isSubgraphNode?: () => boolean }) => n.isSubgraphNode?.()
-      ) as { subgraph?: unknown } | undefined
-      if (!sgNode?.subgraph) return { error: 'No subgraph node' }
+    await comfyPage.page.evaluate(() => {
+      const canvas = window.app!.canvas
+      const graph = canvas.graph!
+      const sgNode = graph._nodes.find((n) =>
+        'isSubgraphNode' in n
+          ? (n as unknown as { isSubgraphNode: () => boolean }).isSubgraphNode()
+          : false
+      ) as unknown as { subgraph?: typeof graph } | undefined
+      if (!sgNode?.subgraph) throw new Error('No subgraph node')
 
-      window.app!.canvas.openSubgraph(sgNode.subgraph as never, sgNode as never)
-      return { entered: true }
+      canvas.setGraph(sgNode.subgraph)
     })
-    expect(result).not.toHaveProperty('error')
 
-    await comfyPage.nextFrame()
-    await comfyPage.nextFrame()
-    await comfyPage.nextFrame()
-
-    expect(await comfyPage.page.evaluate(hasVisibleNodeInViewport)).toBe(true)
+    await expect(async () => {
+      expect(await comfyPage.page.evaluate(hasVisibleNodeInViewport)).toBe(true)
+    }).toPass({ timeout: 2000 })
   })
 
   test('first visit fits viewport to subgraph nodes (Vue)', async ({
@@ -65,13 +64,11 @@ test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
     )
     await comfyPage.vueNodes.waitForNodes()
 
-    // Enter the subgraph via Vue nodes helper
     await comfyPage.vueNodes.enterSubgraph('11')
-    await comfyPage.nextFrame()
-    await comfyPage.nextFrame()
-    await comfyPage.nextFrame()
 
-    expect(await comfyPage.page.evaluate(hasVisibleNodeInViewport)).toBe(true)
+    await expect(async () => {
+      expect(await comfyPage.page.evaluate(hasVisibleNodeInViewport)).toBe(true)
+    }).toPass({ timeout: 2000 })
   })
 
   test('viewport is restored when returning to root (Vue)', async ({
@@ -83,36 +80,24 @@ test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
     )
     await comfyPage.vueNodes.waitForNodes()
 
-    // Capture root viewport before entering subgraph
     const rootViewport = await comfyPage.page.evaluate(() => {
       const ds = window.app!.canvas.ds
       return { scale: ds.scale, offset: [...ds.offset] }
     })
 
-    // Enter subgraph
     await comfyPage.vueNodes.enterSubgraph('11')
     await comfyPage.nextFrame()
-    await comfyPage.nextFrame()
 
-    // Verify viewport changed (we're in a different graph now)
-    const subgraphViewport = await comfyPage.page.evaluate(() => {
-      const ds = window.app!.canvas.ds
-      return { scale: ds.scale, offset: [...ds.offset] }
-    })
-    expect(subgraphViewport).not.toEqual(rootViewport)
-
-    // Exit via breadcrumb
     await comfyPage.subgraph.exitViaBreadcrumb()
-    await comfyPage.nextFrame()
-    await comfyPage.nextFrame()
 
-    // Verify root viewport is restored
-    const restoredViewport = await comfyPage.page.evaluate(() => {
-      const ds = window.app!.canvas.ds
-      return { scale: ds.scale, offset: [...ds.offset] }
-    })
-    expect(restoredViewport.scale).toBeCloseTo(rootViewport.scale, 2)
-    expect(restoredViewport.offset[0]).toBeCloseTo(rootViewport.offset[0], 0)
-    expect(restoredViewport.offset[1]).toBeCloseTo(rootViewport.offset[1], 0)
+    await expect(async () => {
+      const restored = await comfyPage.page.evaluate(() => {
+        const ds = window.app!.canvas.ds
+        return { scale: ds.scale, offset: [...ds.offset] }
+      })
+      expect(restored.scale).toBeCloseTo(rootViewport.scale, 2)
+      expect(restored.offset[0]).toBeCloseTo(rootViewport.offset[0], 0)
+      expect(restored.offset[1]).toBeCloseTo(rootViewport.offset[1], 0)
+    }).toPass({ timeout: 2000 })
   })
 })
--- a/browser_tests/tests/subgraphViewport.spec.ts
+++ b/browser_tests/tests/subgraphViewport.spec.ts
@@ -2,8 +2,36 @@ import { expect } from '@playwright/test'
 
 import { comfyPageFixture as test } from '../fixtures/ComfyPage'
 
+function hasVisibleNodeInViewport() {
+  const canvas = window.app!.canvas
+  if (!canvas?.graph?._nodes?.length) return false
+
+  const ds = canvas.ds
+  const cw = canvas.canvas.width / window.devicePixelRatio
+  const ch = canvas.canvas.height / window.devicePixelRatio
+  const visLeft = -ds.offset[0]
+  const visTop = -ds.offset[1]
+  const visRight = visLeft + cw / ds.scale
+  const visBottom = visTop + ch / ds.scale
+
+  for (const node of canvas.graph._nodes) {
+    const [nx, ny] = node.pos
+    const [nw, nh] = node.size
+    if (
+      nx + nw > visLeft &&
+      nx < visRight &&
+      ny + nh > visTop &&
+      ny < visBottom
+    )
+      return true
+  }
+  return false
+}
+
 test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
-  test('first visit fits viewport to subgraph nodes', async ({ comfyPage }) => {
+  test('first visit fits viewport to subgraph nodes (LG)', async ({
+    comfyPage
+  }) => {
     await comfyPage.workflow.loadWorkflow(
       'subgraphs/subgraph-with-promoted-text-widget'
     )
@@ -21,37 +49,70 @@ test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
     })
     expect(result).not.toHaveProperty('error')
 
-    // Wait for rAF fitToBounds to settle
     await comfyPage.nextFrame()
     await comfyPage.nextFrame()
     await comfyPage.nextFrame()
 
-    const hasVisibleNode = await comfyPage.page.evaluate(() => {
-      const canvas = window.app!.canvas
-      if (!canvas?.graph?._nodes?.length) return false
+    expect(await comfyPage.page.evaluate(hasVisibleNodeInViewport)).toBe(true)
+  })
 
-      const ds = canvas.ds
-      const cw = canvas.canvas.width / window.devicePixelRatio
-      const ch = canvas.canvas.height / window.devicePixelRatio
-      const visLeft = -ds.offset[0]
-      const visTop = -ds.offset[1]
-      const visRight = visLeft + cw / ds.scale
-      const visBottom = visTop + ch / ds.scale
+  test('first visit fits viewport to subgraph nodes (Vue)', async ({
+    comfyPage
+  }) => {
+    await comfyPage.settings.setSetting('Comfy.VueNodes.Enabled', true)
+    await comfyPage.workflow.loadWorkflow(
+      'subgraphs/subgraph-with-promoted-text-widget'
+    )
+    await comfyPage.vueNodes.waitForNodes()
 
-      for (const node of canvas.graph._nodes) {
-        const [nx, ny] = node.pos
-        const [nw, nh] = node.size
-        if (
-          nx + nw > visLeft &&
-          nx < visRight &&
-          ny + nh > visTop &&
-          ny < visBottom
-        )
-          return true
-      }
-      return false
+    // Enter the subgraph via Vue nodes helper
+    await comfyPage.vueNodes.enterSubgraph('11')
+    await comfyPage.nextFrame()
+    await comfyPage.nextFrame()
+    await comfyPage.nextFrame()
+
+    expect(await comfyPage.page.evaluate(hasVisibleNodeInViewport)).toBe(true)
+  })
+
+  test('viewport is restored when returning to root (Vue)', async ({
+    comfyPage
+  }) => {
+    await comfyPage.settings.setSetting('Comfy.VueNodes.Enabled', true)
+    await comfyPage.workflow.loadWorkflow(
+      'subgraphs/subgraph-with-promoted-text-widget'
+    )
+    await comfyPage.vueNodes.waitForNodes()
+
+    // Capture root viewport before entering subgraph
+    const rootViewport = await comfyPage.page.evaluate(() => {
+      const ds = window.app!.canvas.ds
+      return { scale: ds.scale, offset: [...ds.offset] }
     })
 
-    expect(hasVisibleNode).toBe(true)
+    // Enter subgraph
+    await comfyPage.vueNodes.enterSubgraph('11')
+    await comfyPage.nextFrame()
+    await comfyPage.nextFrame()
+
+    // Verify viewport changed (we're in a different graph now)
+    const subgraphViewport = await comfyPage.page.evaluate(() => {
+      const ds = window.app!.canvas.ds
+      return { scale: ds.scale, offset: [...ds.offset] }
+    })
+    expect(subgraphViewport).not.toEqual(rootViewport)
+
+    // Exit via breadcrumb
+    await comfyPage.subgraph.exitViaBreadcrumb()
+    await comfyPage.nextFrame()
+    await comfyPage.nextFrame()
+
+    // Verify root viewport is restored
+    const restoredViewport = await comfyPage.page.evaluate(() => {
+      const ds = window.app!.canvas.ds
+      return { scale: ds.scale, offset: [...ds.offset] }
+    })
+    expect(restoredViewport.scale).toBeCloseTo(rootViewport.scale, 2)
+    expect(restoredViewport.offset[0]).toBeCloseTo(rootViewport.offset[0], 0)
+    expect(restoredViewport.offset[1]).toBeCloseTo(rootViewport.offset[1], 0)
   })
 })
--- a/browser_tests/tests/subgraphViewport.spec.ts
+++ b/browser_tests/tests/subgraphViewport.spec.ts
@@ -50,9 +50,11 @@ test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
       canvas.setGraph(sgNode.subgraph)
     })
 
-    await expect(async () => {
-      expect(await comfyPage.page.evaluate(hasVisibleNodeInViewport)).toBe(true)
-    }).toPass({ timeout: 2000 })
+    await expect
+      .poll(() => comfyPage.page.evaluate(hasVisibleNodeInViewport), {
+        timeout: 2000
+      })
+      .toBe(true)
   })
 
   test('first visit fits viewport to subgraph nodes (Vue)', async ({
@@ -66,9 +68,11 @@ test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
 
     await comfyPage.vueNodes.enterSubgraph('11')
 
-    await expect(async () => {
-      expect(await comfyPage.page.evaluate(hasVisibleNodeInViewport)).toBe(true)
-    }).toPass({ timeout: 2000 })
+    await expect
+      .poll(() => comfyPage.page.evaluate(hasVisibleNodeInViewport), {
+        timeout: 2000
+      })
+      .toBe(true)
   })
 
   test('viewport is restored when returning to root (Vue)', async ({
@@ -90,14 +94,21 @@ test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
 
     await comfyPage.subgraph.exitViaBreadcrumb()
 
-    await expect(async () => {
-      const restored = await comfyPage.page.evaluate(() => {
-        const ds = window.app!.canvas.ds
-        return { scale: ds.scale, offset: [...ds.offset] }
+    await expect
+      .poll(
+        () =>
+          comfyPage.page.evaluate(() => {
+            const ds = window.app!.canvas.ds
+            return { scale: ds.scale, offset: [...ds.offset] }
+          }),
+        { timeout: 2000 }
+      )
+      .toEqual({
+        scale: expect.closeTo(rootViewport.scale, 2),
+        offset: [
+          expect.closeTo(rootViewport.offset[0], 0),
+          expect.closeTo(rootViewport.offset[1], 0)
+        ]
       })
-      expect(restored.scale).toBeCloseTo(rootViewport.scale, 2)
-      expect(restored.offset[0]).toBeCloseTo(rootViewport.offset[0], 0)
-      expect(restored.offset[1]).toBeCloseTo(rootViewport.offset[1], 0)
-    }).toPass({ timeout: 2000 })
   })
 })
--- a/browser_tests/tests/subgraphViewport.spec.ts
+++ b/browser_tests/tests/subgraphViewport.spec.ts
@@ -40,7 +40,13 @@ test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
 
       for (const node of canvas.graph._nodes) {
         const [nx, ny] = node.pos
-        if (nx >= visLeft && nx <= visRight && ny >= visTop && ny <= visBottom)
+        const [nw, nh] = node.size
+        if (
+          nx + nw > visLeft &&
+          nx < visRight &&
+          ny + nh > visTop &&
+          ny < visBottom
+        )
           return true
       }
       return false
--- a/browser_tests/tests/subgraphViewport.spec.ts
+++ b/browser_tests/tests/subgraphViewport.spec.ts
@@ -1,0 +1,51 @@
+import { expect } from '@playwright/test'
+
+import { comfyPageFixture as test } from '../fixtures/ComfyPage'
+
+test.describe('Subgraph viewport restoration', { tag: '@subgraph' }, () => {
+  test('first visit fits viewport to subgraph nodes', async ({ comfyPage }) => {
+    await comfyPage.workflow.loadWorkflow(
+      'subgraphs/subgraph-with-promoted-text-widget'
+    )
+    await comfyPage.nextFrame()
+
+    const result = await comfyPage.page.evaluate(() => {
+      const graph = window.app!.canvas.graph!
+      const sgNode = graph._nodes.find(
+        (n: { isSubgraphNode?: () => boolean }) => n.isSubgraphNode?.()
+      ) as { subgraph?: unknown } | undefined
+      if (!sgNode?.subgraph) return { error: 'No subgraph node' }
+
+      window.app!.canvas.openSubgraph(sgNode.subgraph as never, sgNode as never)
+      return { entered: true }
+    })
+    expect(result).not.toHaveProperty('error')
+
+    // Wait for rAF fitToBounds to settle
+    await comfyPage.nextFrame()
+    await comfyPage.nextFrame()
+    await comfyPage.nextFrame()
+
+    const hasVisibleNode = await comfyPage.page.evaluate(() => {
+      const canvas = window.app!.canvas
+      if (!canvas?.graph?._nodes?.length) return false
+
+      const ds = canvas.ds
+      const cw = canvas.canvas.width / window.devicePixelRatio
+      const ch = canvas.canvas.height / window.devicePixelRatio
+      const visLeft = -ds.offset[0]
+      const visTop = -ds.offset[1]
+      const visRight = visLeft + cw / ds.scale
+      const visBottom = visTop + ch / ds.scale
+
+      for (const node of canvas.graph._nodes) {
+        const [nx, ny] = node.pos
+        if (nx >= visLeft && nx <= visRight && ny >= visTop && ny <= visBottom)
+          return true
+      }
+      return false
+    })
+
+    expect(hasVisibleNode).toBe(true)
+  })
+})
--- a/src/stores/subgraphNavigationStore.ts
+++ b/src/stores/subgraphNavigationStore.ts
@@ -6,10 +6,10 @@ import { useRouter } from 'vue-router'
 
 import type { DragAndScaleState } from '@/lib/litegraph/src/DragAndScale'
 import type { Subgraph } from '@/lib/litegraph/src/litegraph'
-import type { ReadOnlyRect } from '@/lib/litegraph/src/interfaces'
 import { useWorkflowStore } from '@/platform/workflow/management/stores/workflowStore'
 import { useWorkflowService } from '@/platform/workflow/core/services/workflowService'
 import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
+import { useLitegraphService } from '@/services/litegraphService'
 import { app } from '@/scripts/app'
 import { findSubgraphPathById } from '@/utils/graphTraversalUtil'
 import { isNonNullish, isSubgraph } from '@/utils/typeGuardUtil'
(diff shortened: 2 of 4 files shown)


src/stores/subgraphNavigationStore.ts
```diff
@@ -100,15 +100,43 @@ export const useSubgraphNavigationStore = defineStore(
...
+        canvas.ds.scale = viewport.scale
+        canvas.ds.offset[0] = viewport.offset[0]
+        canvas.ds.offset[1] = viewport.offset[1]
+        canvas.setDirty(true, true)
+        return
+      }
+
+      // First visit — fit viewport to the subgraph's nodes
+      const items = Array.from(canvas.positionableItems ?? [])
+      if (!items.length) return
```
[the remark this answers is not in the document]
artokun: Added the guard — thanks.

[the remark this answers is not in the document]
artokun: Addressed in the rewrite — the current code uses flush:'sync' for the activeSubgraph watcher and workflow-scoped cache keys. The hasVisibleContent/guard issues are from earlier iterations that were replaced.

src/stores/subgraphNavigationStore.ts
```diff
@@ -158,6 +197,20 @@ export const useSubgraphNavigationStore = defineStore(
...
+      () => workflowStore.activeWorkflow,
+      (_newWorkflow, _oldWorkflow) => {
+        const subgraph = workflowStore.activeSubgraph
+        if (subgraph) {
+          saveViewport(subgraph.id)
+        } else {
+          saveViewport(getCurrentRootGraphId())
+        }
+      }
+    )
```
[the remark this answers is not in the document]
artokun: Addressed in the rewrite — the current code uses flush:'sync' for the activeSubgraph watcher and workflow-scoped cache keys. The hasVisibleContent/guard issues are from earlier iterations that were replaced.

src/stores/subgraphNavigationStore.ts
```diff
@@ -92,24 +107,92 @@ export const useSubgraphNavigationStore = defineStore(
...
+      const ch = canvas.canvas.height / (window.devicePixelRatio || 1)
+      const visLeft = -ds.offset[0]
+      const visTop = -ds.offset[1]
+      const visRight = visLeft + cw / ds.scale
+      const visBottom = visTop + ch / ds.scale
+
+      const [bx, by, bw, bh] = bounds
+      return (
+        bx + bw > visLeft && bx < visRight && by + bh > visTop && by < visBottom
+      )
```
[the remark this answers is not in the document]
artokun: Addressed in the rewrite — the current code uses flush:'sync' for the activeSubgraph watcher and workflow-scoped cache keys. The hasVisibleContent/guard issues are from earlier iterations that were replaced.

src/stores/subgraphNavigationStore.ts
```diff
@@ -23,114 +23,184 @@ export const useSubgraphNavigationStore = defineStore(
...
+          // write to canvas.ds AFTER this runs. Use double-rAF to ensure
+          // we apply AFTER all other viewport mutations settle.
+          requestAnimationFrame(() => {
+            requestAnimationFrame(() => {
+              applyViewport(viewport)
+            })
+          })
+        } else {
+          applyViewport(viewport)
+        }
```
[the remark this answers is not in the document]
artokun: Added isStillTarget() guard on both deferred paths — checks the full workflow-scoped cache key before applying.

browser_tests/tests/subgraphViewport.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      const ch = canvas.canvas.height / window.devicePixelRatio
+      const visLeft = -ds.offset[0]
+      const visTop = -ds.offset[1]
+      const visRight = visLeft + cw / ds.scale
+      const visBottom = visTop + ch / ds.scale
+
+      for (const node of canvas.graph._nodes) {
+        const [nx, ny] = node.pos
+        if (nx >= visLeft && nx <= visRight && ny >= visTop && ny <= visBottom)
+          return true
```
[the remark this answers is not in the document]
artokun: Switched to rect intersection — checks nx+nw/ny+nh overlap with viewport.

src/stores/subgraphNavigationStore.viewport.test.ts
```diff
@@ -15,174 +15,153 @@ const { mockSetDirty } = vi.hoisted(() => ({
...
+
 describe('useSubgraphNavigationStore - Viewport Persistence', () => {
   beforeEach(() => {
     setActivePinia(createTestingPinia({ stubActions: false }))
-    // Reset canvas state
+    rafCallbacks = []
+    vi.stubGlobal('requestAnimationFrame', (cb: FrameRequestCallback) => {
+      rafCallbacks.push(cb)
+      return rafCallbacks.length
+    })
```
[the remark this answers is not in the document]
artokun: Added vi.unstubAllGlobals() in afterEach.

artokun:
@coderabbitai resume

snomiao:
<\!-- MANUAL_QA_STATUS -->
## 🔍 Automated QA Report

| | |
|---|---|
| **Status** | ❓  |
| **Report** | [sno-qa-10247.comfy-qa.pages.dev](https://sno-qa-10247.comfy-qa.pages.dev/) |
| **CI Run** | [View workflow](https://github.com/Comfy-Org/ComfyUI_frontend/actions/runs/23373279990) |

The QA report includes:
- Before/after video recordings
- **Behavior Changes** table (what changed at a high level)
- **Timeline Comparison** table (frame-by-frame diff)
- AI-generated review with issue analysis

_Generated by automated QA pipeline. Will be updated when run completes._

snomiao:
<\!-- MANUAL_QA_STATUS -->
## 🔍 Automated QA Report

| | |
|---|---|
| **Status** | ✅ Complete |
| **Report** | [sno-qa-10247.comfy-qa.pages.dev](https://sno-qa-10247.comfy-qa.pages.dev/) |
| **CI Run** | [View workflow](https://github.com/Comfy-Org/ComfyUI_frontend/actions/runs/23373279990) |

The report includes before/after video recordings with:
- **Behavior Changes** table — high-level summary of what changed
- **Timeline Comparison** table — frame-by-frame diff with timestamps
- AI-generated issue analysis with severity ratings

_Generated by automated QA pipeline._

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,105 +35,176 @@ export const useSubgraphNavigationStore = defineStore(
...
-    /** LRU cache for viewport states. Key: subgraph ID or 'root' for root graph */
+    /** LRU cache for viewport states. Key: `workflowPath:graphId` */
     const viewportCache = new QuickLRU<string, DragAndScaleState>({
       maxSize: VIEWPORT_CACHE_MAX_SIZE
     })
 
-    /**
-     * Get the ID of the root graph for the currently active workflow.
-     * @returns The ID of the root graph for the currently active workflow.
-     */
```
AustinMroz, line 45: Why are docstrings being stripped from untouched functions?

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,105 +35,176 @@ export const useSubgraphNavigationStore = defineStore(
...
+
+      // Cache miss — fit to content after the canvas has the new graph.
+      // rAF fires after layout + paint, when nodes are positioned.
+      const expectedGraphId = graphId
+      requestAnimationFrame(() => {
+        if (getActiveGraphId() !== expectedGraphId) return
+        if (!canvas.graph) return
+
+        try {
+          const nodes = canvas.graph._nodes
```
AustinMroz: Why is the competing implementation with `litegraphService.fitView()`?

src/stores/subgraphNavigationStore.ts
```diff
@@ -147,20 +219,51 @@ export const useSubgraphNavigationStore = defineStore(
...
     }
 
-    // Update navigation stack when opened subgraph changes (also triggers when switching workflows)
+    // ── Watchers ─────────────────────────────────────────────────────
+
+    // Save viewport BEFORE workflow switch corrupts canvas.ds.
+    // Watch the workflow object (not just path) because multiple unsaved
+    // workflows share the same path. flush:'sync' ensures this fires
+    // before loadGraphData overwrites canvas.ds.
+    watch(
```
AustinMroz: This should probably be moved to `workflowService.beforeLoadNewGraph()`

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,105 +35,176 @@ export const useSubgraphNavigationStore = defineStore(
...
+      const canvas = app.canvas
+      if (!canvas) return
+
+      const expectedKey = cacheKey(graphId)
+      const isStillTarget = () => cacheKey(getActiveGraphId()) === expectedKey
+      const viewport = viewportCache.get(expectedKey)
+      if (viewport) {
+        if (suppressNavigatedSave) {
+          requestAnimationFrame(() => {
+            requestAnimationFrame(() => {
```
AustinMroz: Having nested `requestAnimationFrame`s like leaves a real bad taste in my mouth. Wouldn't applying the viewport in `workflowService.afterLoadNewGraph()` be sufficient and remove the need for cyclecounting?

browser_tests/tests/subgraphViewport.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+      const sgNode = graph._nodes.find(
+        (n: { isSubgraphNode?: () => boolean }) => n.isSubgraphNode?.()
+      ) as { subgraph?: unknown } | undefined
+      if (!sgNode?.subgraph) return { error: 'No subgraph node' }
+
+      window.app!.canvas.openSubgraph(sgNode.subgraph as never, sgNode as never)
+      return { entered: true }
+    })
+    expect(result).not.toHaveProperty('error')
+
```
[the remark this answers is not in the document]
artokun: Good point — added a guard in the latest push that checks `canvas.graph === targetGraph` after `openSubgraph()` to verify the navigation actually succeeded.

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,105 +35,176 @@ export const useSubgraphNavigationStore = defineStore(
...
+            minX - pad,
+            minY - pad,
+            maxX - minX + 2 * pad,
+            maxY - minY + 2 * pad
+          ]
+          canvas.ds.fitToBounds(bounds)
+          canvas.setDirty(true, true)
+        } catch {
+          // graph._nodes can throw if graph is mid-transition
+        }
```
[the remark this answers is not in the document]
artokun: The catch already logs the error and falls back to the default viewport. The `_nodes` access failure during mid-transition is transient and resolves on the next navigation. Retrying would risk an infinite loop if the graph is genuinely broken.

src/stores/subgraphNavigationStore.ts
artokun: Good catch — the docstrings were removed during a refactor pass to reduce noise, but they shouldn't have been stripped from functions that weren't otherwise changed. Will restore them.

src/stores/subgraphNavigationStore.ts
artokun: The inline bounds calculation exists because `litegraphService.fitView()` fires asynchronously and doesn't guarantee timing relative to the graph transition. The inline version runs inside the same rAF cycle as the viewport restore, so we know the graph is settled. Happy to switch to `fitView()` if we can confirm it handles the mid-transition edge case (where `_nodes` might be partially populated).

src/stores/subgraphNavigationStore.ts
artokun: Agree that `beforeLoadNewGraph()` is the right place for saving the outgoing viewport. Will move the save there.

src/stores/subgraphNavigationStore.ts
artokun: The nested rAFs are there because `afterLoadNewGraph()` fires before the canvas has fully transitioned — Vue reactivity and litegraph layout both run async. The double-rAF ensures paint + layout have completed. If `afterLoadNewGraph()` can guarantee the graph is fully ready (nodes positioned, canvas updated), then a single call there would be cleaner. Will investigate whether the timing is reliable enough.

artokun:
@coderabbitai review

src/stores/subgraphNavigationStore.ts
artokun: Done — moved the viewport save into `beforeLoadNewGraph()` and it works cleanly. Also removed the nested rAFs from `restoreViewport` — a single deferred `requestAnimationFrame` during workflow switches is enough since `beforeLoadNewGraph` captures the viewport before `loadGraphData` touches `canvas.ds`. Tested locally: enter subgraph → zoom/pan → switch workflow tabs → switch back — viewport restores correctly in all cases.

src/stores/subgraphNavigationStore.ts
artokun: Resolved — the nested rAFs are gone. The refactor to `beforeLoadNewGraph` eliminated the timing issue entirely. Restore now applies directly for subgraph navigation, and defers with a single rAF only during workflow switches (to run after `loadGraphData` applies `extra.ds`).

artokun:
@coderabbitai review

artokun:
@coderabbit if the changes look good can you mark approved please for CI?

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,111 +35,177 @@ export const useSubgraphNavigationStore = defineStore(
...
+          if (!Number.isFinite(minX)) return
+
+          const pad = 80
+          const bounds: ReadOnlyRect = [
+            minX - pad,
+            minY - pad,
+            maxX - minX + 2 * pad,
+            maxY - minY + 2 * pad
+          ]
+          canvas.ds.fitToBounds(bounds)
```
DrJKL: 🔴 **Duplicate lens** — This 15-line min/max loop with padding is a near-exact reimplementation of `createBounds` from `src/lib/litegraph/src/measure.ts:324-345`. Nodes already expose `boundingRect` via `LGraphNode`. The entire block could be replaced with:

```ts
import { createBounds } from '@/lib/litegraph/src/measure'
// ...
const bounds = createBounds(nodes, 80)
if (!bounds) return
canvas.ds.fitToBounds(bounds)
canvas.setDirty(true, true)
```

This also eliminates the need for the bare `catch {}` on L175 since `createBounds` returns `null` for degenerate inputs.

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,111 +35,177 @@ export const useSubgraphNavigationStore = defineStore(
...
+     * workflow switch transition.
      */
+    let isWorkflowSwitching = false
+    // ── Helpers ──────────────────────────────────────────────────────
+
+    /** Build a workflow-scoped cache key unique per workflow instance. */
+    const cacheKey = (graphId: string, workflowRef?: object | null): string => {
+      const wf = workflowRef ?? workflowStore.activeWorkflow
+      const prefix = wf
+        ? `${(wf as { path?: string }).path ?? ''}#${getWorkflowInstanceId(wf)}`
```
DrJKL: 🟡 **Blurry cast** — `activeWorkflow` is `LoadedComfyWorkflow | null` which inherits `path: string` from `UserFile`. The `as { path?: string }` cast drops that type info. Consider typing the `workflowRef` parameter more precisely:

```ts
const cacheKey = (graphId: string, workflowRef?: { path?: string } | null): string => {
```

Or `Pick<ComfyWorkflow, 'path'>` to eliminate the assertion entirely.

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,111 +35,177 @@ export const useSubgraphNavigationStore = defineStore(
...
+    const workflowInstanceIds = new WeakMap<object, number>()
+    let nextInstanceId = 0
+    const getWorkflowInstanceId = (workflow: object): number => {
+      let id = workflowInstanceIds.get(workflow)
+      if (id === undefined) {
+        id = nextInstanceId++
+        workflowInstanceIds.set(workflow, id)
+      }
+      return id
+    }
```
DrJKL: 🟡 **Over-focused lens** — The `WeakMap<object>` + monotonic counter exists to disambiguate unsaved workflows sharing a path. But the LRU cache is ephemeral (not persisted), and the worst case with path-only keys is a stale viewport restore — a cosmetic issue for an edge case that may never happen in practice. Consider starting with `path:graphId` keys and adding the instance discriminator if the collision is observed (YAGNI).

Also, `WeakMap<object>` could be `WeakMap<ComfyWorkflow>` to tighten the aperture on what's allowed in.

src/stores/subgraphNavigationStore.ts
```diff
@@ -229,13 +305,24 @@ export const useSubgraphNavigationStore = defineStore(
...
 
+    /** Save the current viewport for the active graph/workflow. Called by
+     *  workflowService.beforeLoadNewGraph() before the canvas is overwritten. */
+    const saveCurrentViewport = () => {
+      saveViewport(getActiveGraphId())
+      isWorkflowSwitching = true
+      requestAnimationFrame(() => {
+        isWorkflowSwitching = false
+      })
+    }
```
DrJKL, line 271: 🟡 **Shutter stuck open** — `isWorkflowSwitching` is set `true` here and reset via `requestAnimationFrame`. If the tab is backgrounded, browsers throttle/pause rAF, leaving this flag stuck `true` indefinitely. All subsequent `onNavigated` calls would skip viewport saves and unnecessarily defer restores.

Consider a synchronous reset — e.g., clear the flag at the start of `onNavigated` after reading it, or use `setTimeout(fn, 0)` which fires even in backgrounded tabs.

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,111 +35,177 @@ export const useSubgraphNavigationStore = defineStore(
...
       if (isInRootGraph) {
         idStack.value.length = 0
-        restoreViewport(getCurrentRootGraphId())
+        if (isWorkflowSwitching) {
+          // Defer restore until after loadGraphData applies extra.ds
+          requestAnimationFrame(() => {
+            restoreViewport(getCurrentRootGraphId())
+          })
+        } else {
+          restoreViewport(getCurrentRootGraphId())
```
DrJKL: 🟡 **Double exposure** — When `isWorkflowSwitching` is true, `onNavigated` wraps `restoreViewport` in a `requestAnimationFrame`. But `restoreViewport` itself already schedules another `requestAnimationFrame` on cache miss (L146). On the cache-miss + workflow-switch path, that’s two nested rAFs deep — timing-dependent and fragile. The comment says "Defer until after loadGraphData applies extra.ds" — would a single rAF in `restoreViewport` be sufficient if the guard on L147 is correct?

src/stores/subgraphNavigationStore.ts
```diff
@@ -147,20 +214,29 @@ export const useSubgraphNavigationStore = defineStore(
...
+    // ── Watchers ─────────────────────────────────────────────────────
+
+    // React to subgraph navigation (enter/exit/switch).
     watch(
       () => workflowStore.activeSubgraph,
       (newValue, oldValue) => {
         onNavigated(newValue, oldValue)
-      }
+      },
+      { flush: 'sync' }
```
DrJKL, line 194: 🟡 **Rapid-fire shutter** — `{ flush: 'sync' }` fires the watcher synchronously the instant `activeSubgraph` is assigned, before any other watchers or DOM updates from the same state change. If `activeSubgraph` is set *before* the canvas finishes updating, `getCurrentViewport()` captures a stale `ds.state`.

The previous default (`'pre'`) flush waited until the render queue flushed. If sync is truly needed here, a comment explaining the invariant would help future readers understand why the lens must fire before the scene is composed.

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,111 +35,177 @@ export const useSubgraphNavigationStore = defineStore(
...
-     * Restore viewport state for a graph.
-     * @param graphId The graph ID to restore. Use 'root' for root graph, or omit to use current context.
+     * Restore viewport for a graph. On cache miss, fit the canvas to
+     * visible content after the browser has completed layout (rAF).
      */
-    const restoreViewport = (graphId: string) => {
-      const viewport = viewportCache.get(graphId)
-      if (!viewport) return
-
+    /** Apply a viewport state to the canvas. */
```
DrJKL, line 120: nit: **Duplicate viewfinder** — Two adjacent JSDoc comments here:
```
* Restore viewport for a graph. On cache miss, fit the canvas to
* visible content after the browser has completed layout (rAF).
*/
/** Apply a viewport state to the canvas. */
```
The first describes `restoreViewport`, the second describes `applyViewport`. They're back-to-back because `applyViewport` was extracted but the original doc wasn't moved.

src/stores/subgraphNavigationStore.ts
```diff
@@ -229,13 +305,24 @@ export const useSubgraphNavigationStore = defineStore(
...
       activeSubgraph,
       navigationStack,
       restoreState,
       exportState,
       saveViewport,
       restoreViewport,
+      saveCurrentViewport,
       updateHash,
       viewportCache
     }
```
DrJKL, line 284: nit: `viewportCache` is exposed on the store’s public API but only consumed by tests. Exposing the raw LRU gives any consumer full read/write/delete access to internal state. Consider keeping it private and testing through `saveViewport`/`restoreViewport` behavior.

browser_tests/tests/subgraphViewport.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+    await comfyPage.nextFrame()
+
+    const result = await comfyPage.page.evaluate(() => {
+      const graph = window.app!.canvas.graph!
+      const sgNode = graph._nodes.find(
+        (n: { isSubgraphNode?: () => boolean }) => n.isSubgraphNode?.()
+      ) as { subgraph?: unknown } | undefined
+      if (!sgNode?.subgraph) return { error: 'No subgraph node' }
+
+      window.app!.canvas.openSubgraph(sgNode.subgraph as never, sgNode as never)
```
DrJKL: 🟡 **Lens cap on** — `as never` is a total type erasure, no better than `as any` for catching future signature changes. If `openSubgraph` gains a required parameter or changes argument types, this compiles silently and fails at runtime. Prefer specific assertions:

```ts
window.app!.canvas.openSubgraph(
  sgNode.subgraph as Subgraph,
  sgNode as unknown as SubgraphNode
)
```

browser_tests/tests/subgraphViewport.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+      return { entered: true }
+    })
+    expect(result).not.toHaveProperty('error')
+
+    // Wait for rAF fitToBounds to settle
+    await comfyPage.nextFrame()
+    await comfyPage.nextFrame()
+    await comfyPage.nextFrame()
+
```
DrJKL, line 59: 🟡 **Long exposure** — Three sequential `nextFrame()` is a magic-number timing hack. If the implementation changes to use two rAFs (or a `setTimeout`), this breaks silently. Playwright best practices prefer retrying assertions:

```ts
await expect(async () => {
  const visible = await comfyPage.page.evaluate(() => { /* visibility check */ })
  expect(visible).toBe(true)
}).toPass({ timeout: 2000 })
```

src/stores/subgraphNavigationStore.viewport.test.ts
```diff
@@ -18,174 +18,158 @@ const { mockSetDirty } = vi.hoisted(() => ({
...
-      navigationStore.restoreViewport('non-existent')
+      store.restoreViewport('non-existent')
 
-      // Canvas should not change
+      // Should not change canvas synchronously
       expect(mockCanvas.ds.scale).toBe(1)
       expect(mockCanvas.ds.offset).toEqual([0, 0])
       expect(mockSetDirty).not.toHaveBeenCalled()
+      // But should have scheduled a rAF
+      expect(rafCallbacks).toHaveLength(1)
```
DrJKL, line 179: 🟡 **Phantom exposure** — This test asserts a rAF was *scheduled* but never flushes it. The entire cache-miss codepath (bounding-box calculation, `fitToBounds`, `setDirty`, the `getActiveGraphId()` guard) is untested. Consider flushing `rafCallbacks[0](0)` in a follow-up test with `canvas.graph._nodes` populated, and asserting `fitToBounds` was called with the expected padded bounds.

src/stores/subgraphNavigationStore.viewport.test.ts
```diff
@@ -18,174 +18,158 @@ const { mockSetDirty } = vi.hoisted(() => ({
...
+
+      // Key includes instance ID: `path#N:graphId`
+      const keys = [...store.viewportCache.keys()]
+      const matchingKey = keys.find(
+        (k) => k.startsWith('workflows/test.json#') && k.endsWith(':root')
+      )
+      expect(matchingKey).toBeDefined()
+      // Bare key without workflow path should not exist
+      expect(keys.some((k) => k === ':root')).toBe(false)
+    })
```
DrJKL, line 123: nit: This test inspects raw cache keys via `viewportCache.keys()` and pattern-matches the string format — it's testing the encoding, not the behavior. A sharper lens: save a viewport under workflow A, switch `activeWorkflow` to B, save the same `graphId`, then verify restoring under A returns A’s values.

DrJKL (commented):
## Viewport Zoom Review 📷

The core fix here is solid — fitting subgraph content into view on first visit is the right UX. A few places where the lens could be sharpened:

**Highlights:**
- ✅ Workflow-scoped cache keys prevent cross-workflow viewport collisions
- ✅ Fit-to-bounds fallback gives users a sensible default zoom on first subgraph entry
- ✅ `saveCurrentViewport` hook in `beforeLoadNewGraph` captures viewport before canvas overwrite

**Key concerns:**
- 🔴 The manual bounding-box calculation is a duplicate lens — `createBounds` from `measure.ts` already does this exact framing
- 🟡 The `isWorkflowSwitching` flag uses rAF for reset, which can stick in backgrounded tabs
- 🟡 The new fit-to-bounds codepath and `saveCurrentViewport()` lack unit test coverage
- 🟡 `flush: 'sync'` watcher needs justification — it can snap the shutter before the scene is composed

src/stores/subgraphNavigationStore.ts
artokun: Restored all stripped docstrings.

src/stores/subgraphNavigationStore.ts
artokun: Replaced with `litegraphService.fitView()`.

src/stores/subgraphNavigationStore.ts
artokun: Already called from there — `workflowService.beforeLoadNewGraph()` calls `saveCurrentViewport()` at line 387.

src/stores/subgraphNavigationStore.ts
artokun: The single rAF defer is needed because `loadGraphData` applies `extra.ds` during configure, which runs before `afterLoadNewGraph`. The defer waits for that to land before restoring. Not nested rAFs — the `isWorkflowSwitching` flag gates a single deferred restore.

src/stores/subgraphNavigationStore.ts
artokun: Already replaced with `litegraphService.fitView()` in the previous push.

src/stores/subgraphNavigationStore.ts
artokun: Typed `workflowRef` parameter as `{ path?: string } | null`, removed the cast.

src/stores/subgraphNavigationStore.ts
artokun: Removed instance-counter disambiguation. Cache keys are now path-only.

src/stores/subgraphNavigationStore.ts
artokun: Switched to `setTimeout(fn, 0)` — fires even in backgrounded tabs.

src/stores/subgraphNavigationStore.ts
artokun: Collapsed — `onNavigated` now calls `restoreViewport` directly, which handles its own rAF defer on cache miss.

src/stores/subgraphNavigationStore.ts
artokun: Added comment: sync flush captures the outgoing viewport before other watchers mutate the canvas.

src/stores/subgraphNavigationStore.ts
artokun: Fixed in the previous push.

src/stores/subgraphNavigationStore.ts
artokun: Marked `@internal` — kept exposed since many tests rely on direct cache access.

browser_tests/tests/subgraphViewport.spec.ts
artokun: Replaced with specific type assertions and switched to `canvas.setGraph()`.

browser_tests/tests/subgraphViewport.spec.ts
artokun: Switched to `expect().toPass({ timeout: 2000 })` retrying assertions.

src/stores/subgraphNavigationStore.viewport.test.ts
artokun: Added `fitView` cache-miss test that flushes the rAF callback.

src/stores/subgraphNavigationStore.viewport.test.ts
artokun: Rewrote to test behavioral isolation: save under workflow A, save under B, restore under A returns A's values.

artokun:
@coderabbitai review

src/stores/subgraphNavigationStore.ts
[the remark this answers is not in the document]
artokun: Addressed — `viewportKey()` was replaced with `cacheKey()` that takes an explicit `workflowRef` parameter. `saveCurrentViewport()` is now called from `beforeLoadNewGraph()` before the workflow changes.

browser_tests/tests/subgraphViewport.spec.ts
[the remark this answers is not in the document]
artokun: Already using rectangle intersection — checks `nx+nw > visLeft && nx < visRight && ny+nh > visTop && ny < visBottom`.

src/stores/subgraphNavigationStore.ts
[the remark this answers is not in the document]
artokun: Addressed — collapsed nested rAFs. `restoreViewport` now has a single rAF with `getActiveGraphId() !== expectedGraphId` guard on cache miss. No more double-deferred paths.

src/stores/subgraphNavigationStore.viewport.test.ts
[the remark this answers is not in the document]
artokun: Already present — `vi.unstubAllGlobals()` in `afterEach` at line 91.

browser_tests/tests/subgraphViewport.spec.ts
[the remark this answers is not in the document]
artokun: Test rewritten — LG test now uses `canvas.setGraph()` with proper type assertions. Vue test uses `comfyPage.vueNodes.enterSubgraph()` with `toPass()` retrying assertions.

src/stores/subgraphNavigationStore.ts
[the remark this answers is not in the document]
artokun: Addressed — inline bounds calculation replaced with `litegraphService.fitView()` which handles the bounds computation and error cases internally.

artokun:
Addressed CodeRabbit review feedback:

- **Cache key mismatch in LRU test** (minor): Fixed — keys now use workflow-scoped `:` prefix matching production format.
- **Workflow-switch restore key** (major, duplicate): This is safe as-is — when `isWorkflowSwitching` is true, `onNavigated` skips the save. On restore, if the key misses (stale workflow prefix), it falls through to `fitView()` which is correct first-visit behavior. Adding an explicit `workflowRef` to `restoreViewport` would be a defensive hardening but risks changing the timing contract in a way that's hard to test. Noted for follow-up.

AustinMroz (dismissed):
Changes have my approval.  :+1:

src/stores/subgraphNavigationStore.ts
```diff
@@ -34,20 +35,44 @@ export const useSubgraphNavigationStore = defineStore(
...
+      const wf = workflowRef ?? workflowStore.activeWorkflow
+      const prefix = wf?.path ?? ''
+      return `${prefix}:${graphId}`
+    }
+
+    /** ID of the graph currently shown on the canvas. */
+    const getActiveGraphId = (): string => {
+      const canvas = canvasStore.getCanvas()
+      return canvas?.subgraph?.id ?? getCurrentRootGraphId()
+    }
```
DrJKL, line 72: Nit: I almost always prefer function declarations where possible.
```suggestion
    /** Build a workflow-scoped cache key. */
    function buildCacheKey(
      graphId: string,
      workflowRef?: { path?: string } | null
    ): string {
      const wf = workflowRef ?? workflowStore.activeWorkflow
      const prefix = wf?.path ?? ''
      return `${prefix}:${graphId}`
    }

    /** ID of the graph currently shown on the canvas. */
    function getActiveGraphId(): string {
      const canvas = canvasStore.getCanvas()
      return canvas?.subgraph?.id ?? getCurrentRootGraphId()
    }
```

DrJKL: With a verb if it's doing something.

src/stores/subgraphNavigationStore.ts
DrJKL: @coderabbitai Open an issue to fix the tests to not depend on internals and prefer testing via the intended public interfaces.

DrJKL (dismissed):

DrJKL (approved):

PR #10252, opened 2026-03-18
fix: restore is_template tracking for app mode templates

comfyui-wiki:
## Summary

App mode templates (names ending in `.app`, e.g. `templates-qwen_multiangle.app`) were never counted as template executions in Mixpanel because `getExecutionContext` used `activeWorkflow.filename` for the `knownTemplateNames` lookup — but `getFilenameDetails` treats `.app.json` as a compound extension and strips it entirely, leaving `"templates-qwen_multiangle"` instead of `"templates-qwen_multiangle.app"`. The set lookup always returned `false`, so every execution was sent with `is_template: false`.

## Changes

- **Fix**: derive the template lookup key from `fullFilename.replace(/\.json$/i, '')` instead of `filename`, which preserves the `.app` suffix and correctly matches `knownTemplateNames`
- **Also fixes**: `workflow_name`, `getTemplateByName`, and `getEnglishMetadata` calls in the same branch now use the corrected name
- **Tests**: three new cases in `MixpanelTelemetryProvider.test.ts` — regular template, `.app` template (regression), and non-template

## Before / After

| Template name in index | `activeWorkflow.filename` | `fullFilename` → stripped | `is_template` |
|---|---|---|---|
| `flux-dev` | `flux-dev` | `flux-dev` | ✅ true |
| `templates-qwen_multiangle.app` | `templates-qwen_multiangle` ❌ | `templates-qwen_multiangle.app` ✅ | fixed: true |

## Review Focus

The change is confined to `getExecutionContext.ts`. `fullFilename` is always set (it is assigned in `UserFile` constructor from `getPathDetails`), so no null-safety issue.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10252-fix-restore-is_template-tracking-for-app-mode-templates-3276d73d365081d4b998edc62ad010dc) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/telemetry/utils/getExecutionContext.ts
+++ b/src/platform/telemetry/utils/getExecutionContext.ts
@@ -80,20 +80,21 @@ export function getExecutionContext(): ExecutionContext {
   )
 
   if (activeWorkflow?.filename) {
-    const isTemplate = templatesStore.knownTemplateNames.has(
-      activeWorkflow.filename
-    )
+    // Use fullFilename minus .json to reconstruct the template name, which
+    // preserves compound suffixes like ".app" (e.g. "foo.app.json" → "foo.app").
+    // Using just `filename` strips ".app.json" entirely (e.g. "foo"), which
+    // won't match knownTemplateNames entries like "foo.app".
+    const templateName = activeWorkflow.fullFilename.replace(/\.json$/i, '')
+    const isTemplate = templatesStore.knownTemplateNames.has(templateName)
 
     if (isTemplate) {
-      const template = templatesStore.getTemplateByName(activeWorkflow.filename)
+      const template = templatesStore.getTemplateByName(templateName)
 
-      const englishMetadata = templatesStore.getEnglishMetadata(
-        activeWorkflow.filename
-      )
+      const englishMetadata = templatesStore.getEnglishMetadata(templateName)
 
       return {
         is_template: true,
-        workflow_name: activeWorkflow.filename,
+        workflow_name: templateName,
         template_source: template?.sourceModule,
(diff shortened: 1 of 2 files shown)


src/platform/telemetry/utils/getExecutionContext.ts
christian-byrne, line 1: If getTemplateByName is broken at the store layer due to the existence of periods, should we fix it there?

comfyui-wiki: > *(This reply was generated by Claude Code on behalf of the PR author)*

Good question. `getTemplateByName` in the store is not actually broken — it does a correct strict equality check against `template.name`, which is the canonical identifier (e.g. `"templates-qwen_multiangle.app"`).

The root cause was that `getExecutionContext` was computing the wrong lookup key: it used `activeWorkflow.filename`, which `getFilenameDetails` derives by stripping `.app.json` as a compound extension, leaving just `"templates-qwen_multiangle"` — a key that will never match.

Fixing it at the store layer would mean teaching `getTemplateByName` to accept both `"foo"` and `"foo.app"` as equivalent, which introduces ambiguity about what the canonical name actually is. The fix here corrects the key derivation at the call site (compute from `fullFilename` minus `.json`), so the store continues to receive the exact canonical name it expects. That felt like the cleaner boundary.

Happy to move the fix elsewhere if you'd prefer a different approach.

pythongosssss (approved):

comfyui-wiki:
Can I merge it? Or I need to wait for someone to merge it. All the changes are done by Claude's code, I know nothing about it.

PR #10273, opened 2026-03-18
fix: resync slot layouts when switching between app mode and graph mode

DrJKL:
## Summary

Fix broken link rendering (noodles disappearing or going to wrong positions) when switching between app mode and graph mode tabs.

## Changes

- **What**: When the graph canvas is hidden via `display: none` in app mode, slot elements lose valid DOM measurements. On switching back, links rendered at stale coordinates or disappeared. This PR rekeys `LGraphNode` components by workflow path, adds measurability guards to skip hidden slots, clears stale layouts, and watches `linearMode` to trigger a full slot layout resync on mode transitions.

## Review Focus

- The `isSlotElementMeasurable` guard skips elements that are disconnected or have zero-size rects — verify this doesn't inadvertently skip slots during normal graph rendering.
- The `linearMode` watcher clears all slot layouts when entering app mode and requests a full resync when leaving — confirm no flicker or race with the RAF-based sync scheduler.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10273-fix-resync-slot-layouts-when-switching-between-app-mode-and-graph-mode-3276d73d3650812f9366dae53c7b2d37) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/composables/useSlotElementTracking.ts
```diff
@@ -99,6 +129,13 @@ export function syncNodeSlotLayoutsFromDOM(
   const positionConv = conv ?? useSharedCanvasPositionConversion()
 
   for (const [slotKey, entry] of node.slots) {
+    if (!isSlotElementMeasurable(entry.el)) {
```
pythongosssss: Could you avoid the double getBoundingClientRect call, inside `isSlotElementMeasurable` then on line 139

PR #10286, opened 2026-03-19
fix: preserve remote widget values restored by configure()

### LINKED ISSUE #13017: Severe regression: workflow load rewrites node values (ckpt/lora and others), sometimes resetting to 0, destroying JSON integrity
### Summary

This is a severe regression that makes workflows unreliable.

When loading a workflow, node values are being silently rewritten.
This affects not only checkpoint/lora fields but also other nodes,
where values are unexpectedly reset (e.g. to 0).

Even worse, the original JSON is effectively destroyed after loading and saving.

---

### What happens

- Workflow JSON contains correct values
- Load workflow in ComfyUI
- Values are modified without user action:
  - ckpt/lora replaced with different values
  - some node parameters reset to 0
- Save workflow again
→ original data is permanently lost

---

### Expected behavior

- Workflow loader must preserve stored values exactly
- Missing resources should NOT trigger value replacement
- Invalid values can be marked, but never silently rewritten

---

### Actual behavior

- Values are rewritten during load
- Some nodes reset values (e.g. to 0)
- Workflow becomes non-reproducible
- JSON integrity is broken

---

### Reproduction (minimal)

1. Create a workflow with:
   - CheckpointLoader / LoraLoader
   - any node with numeric parameters
2. Save workflow
3. Modify JSON manually:
   - set ckpt/lora to non-existent value
   - set numeric fields to specific values
4. Load workflow

→ observe:
- ckpt/lora replaced
- some values reset (e.g. 0)
- after saving, original JSON is lost

---

### Impact

This completely breaks:
- workflow portability
- reproducibility
- archive reliability (PNG workflows)
- long-term project stability

This behavior makes ComfyUI workflows unsafe to use as saved artifacts.

---

### Notes

- This did NOT happen ~1 month ago
- This appears to be a regression in workflow loading / widget handling
- Not sure if this belongs to frontend or frontend/backend boundary

---

### Additional comment

This is not a minor issue.
If workflows cannot be trusted to preserve their own values,
it seriously undermines the usability of the entire project.

At this point, it feels like core functionality is breaking faster than it is being stabilized.

artokun:
## Summary

Fix remote-backed combo widgets (custom nodes using `remote:` config) having their saved values overwritten on workflow load and tab switch.

**Status: Draft** — awaiting repro clarification from issue reporters. New report from EternalBlack (Comfy-Org/ComfyUI#13017) confirms tab switching also triggers the reset on custom node widgets, which aligns with this fix.

## Problem

`useRemoteWidget.onFirstLoad()` unconditionally set `widget.value = data[0]` after `configure()` had already restored the saved value from the workflow JSON. This caused remote-backed model selectors to silently switch to the first available model on workflow load. The Vue component remounts on tab switch, resetting the local `isLoaded` flag, so the overwrite also triggers on every tab switch.

## Fix

Add `configuredByLoad` flag to `IBaseWidget`, set by `LGraphNode.configure()` after restoring each widget value from `widgets_values`. `onFirstLoad()` checks this flag and skips the overwrite if the value was already restored by the load path. The flag persists on the widget object across tab switches.

## Related

- Comfy-Org/ComfyUI#13017 — Severe regression: workflow load rewrites node values
- EternalBlack's report: tab switching visually resets custom node widgets to defaults
- Clivey1234's report: random "failed to save workflow draft" errors
- zerobell-lee's original report: checkpoint/lora values replaced on load

Fixes Comfy-Org/ComfyUI#13017


src/renderer/extensions/vueNodes/widgets/composables/useRemoteWidget.ts
```diff
@@ -130,8 +130,19 @@ export function useRemoteWidget<
...
   const onFirstLoad = (data: T | T[]) => {
     isLoaded = true
+
+    // If configure() already restored a saved value, keep it — don't
+    // overwrite with the first item from the remote list.
+    const hasSavedValue =
+      widget.value != null &&
+      widget.value !== defaultValue &&
+      widget.value !== ''
+    if (hasSavedValue) return
```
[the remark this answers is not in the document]
artokun: Good catch — this was addressed in the follow-up commit (e7748b8eb). The `widget.value !== defaultValue` comparison is gone; `onFirstLoad()` now checks `widget.configuredByLoad` instead, which is a flag set by `LGraphNode.configure()` after restoring each widget value from the serialized workflow data. This cleanly avoids the ambiguous case where a legitimately saved value happens to equal the default.

There is also a regression test (`preserves saved value even when it equals defaultValue`) that sets `widget.value = DEFAULT_VALUE` and `configuredByLoad = true`, then asserts the value is preserved after remote data loads.

artokun:
@coderabbitai review

artokun:
@coderabbitai review

artokun:
@coderabbitai review

artokun:
All CodeRabbit feedback was addressed in previous commits — the `configuredByLoad` flag approach replaces the ambiguous value comparison, and the regression test covers the edge case. Rebased onto latest main.

@coderabbitai review

src/platform/workflow/core/services/workflowService.ts
```diff
@@ -294,8 +294,6 @@ export const useWorkflowService = () => {
       }
     }
 
-    workflowDraftStore.removeDraft(workflow.path)
-
     // If this is the last workflow, create a new default temporary workflow
```
christian-byrne, line 297: question: this `removeDraft` removal looks unrelated to the widget-value-preservation fix. It is already called inside `workflowStore.closeWorkflow` (workflowStore.ts:343), so the delete is just removing a duplicate, but the PR body doesn't explain the link to the remote-widget regression. Could the rationale be added to the description, or split into its own PR? Tying together unrelated changes makes bisecting harder if either turns out to regress something.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -920,6 +920,7 @@ export class LGraphNode
           if (widget.serialize === false) continue
           if (i >= info.widgets_values.length) break
           widget.value = info.widgets_values[i++]
+          widget.configuredByLoad = true
```
christian-byrne, line 923: question: should `configuredByLoad = true` also be set in the `w.options?.property` restore path above (lines 908-914)? That branch also restores a saved value (from `properties`), but the new flag is only set in the `widgets_values` branch. A remote-backed widget that declares `options.property` and is restored only via the property path would still get overwritten by `onFirstLoad`. If that combination doesn't exist for any remote widget today, a one-line comment noting the assumption would be helpful.

christian-byrne, line 923: suggestion (non-blocking): the new tests in useRemoteWidget.test.ts simulate the configure path by setting `widget.configuredByLoad = true` manually. Consider adding a small test in `src/lib/litegraph/test/` (or wherever LGraphNode.configure is unit tested) that asserts the flag is set after `configure()` restores from `widgets_values`. That guards the contract between the two halves of this fix from regressing independently.

src/lib/litegraph/src/types/widgets.ts
```diff
@@ -450,6 +450,13 @@ export interface IBaseWidget<
   advanced?: boolean
   tooltip?: string
 
+  /**
+   * Set to `true` by {@link LGraphNode.configure} after restoring the widget's
+   * value from serialized workflow data. Used by async widget initializers
+   * (e.g. remote widgets) to avoid overwriting a restored value.
+   */
+  configuredByLoad?: boolean
```
christian-byrne, line 458: nitpick (non-blocking): the JSDoc could note that this flag is set once and intentionally never cleared, it's effectively a permanent "this widget's value is owned by configure()" marker, persisted across the Vue component remounts that happen on tab switch. Without that note, a future maintainer might try to reset it inside `onFirstLoad`, which would re-introduce the original tab-switch regression.

PR #10287, opened 2026-03-19
fix: preserve nested promoted widget identities for duplicate widget names

DrJKL:
## Summary

Fixes nested promoted widgets with duplicate names (e.g. two inner nodes both exposing a `text` widget) colliding in the promotion store, Vue rendering, and border styling by threading a disambiguating source node ID through the entire promotion pipeline.

## Changes

- **What**: Adds `sourceNodeId` / `disambiguatingSourceNodeId` to promotion entries, store keys, widget views, and resolution functions so that identically-named promoted widgets from different source nodes are tracked, resolved, and rendered independently. Updates `proxyWidgets` serialization schema to support optional 3-tuple format. Fixes promoted-border styling to use host-node identity instead of interior-node identity, correctly highlighting intermediate promoted widgets.

## Review Focus

- Disambiguation key threading through `promotionStore`, `SubgraphNode`, `PromotedWidgetView`, and `resolveConcretePromotedWidget`
- Backward compatibility of the `proxyWidgets` schema (`[string, string]` | `[string, string, string]`)
- Border styling logic change in `NodeWidgets.vue` (`isPromotedView` replaced with host-node-based lookup)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10287-fix-preserve-nested-promoted-widget-identities-for-duplicate-widget-names-3286d73d36508158b065e61744d92e14) by [Unito](https://www.unito.io)


artokun (commented):
## Review: Disambiguation for duplicate promoted widget names

Nice work threading the sourceNodeId through the entire promotion pipeline. The core approach — keying on `(interiorNodeId, widgetName)` instead of just `widgetName` — is solid and the `PromotedWidgetViewManager.reconcile()` changes correctly maintain stable object identity. Three reviewers converged on the same set of findings:

### Issues worth addressing

**1. `removeWidgetByName` picks wrong widget when names collide**
`SubgraphNode.removeWidgetByName(name)` uses `Array.find` which returns the first match. With two promoted views sharing a display name (the exact scenario this PR enables), it always removes the first one regardless of which was intended. The `removeWidget(widget)` overload works correctly via object identity. Consider deprecating the name-based overload for SubgraphNode, or accepting an additional disambiguating argument.

**2. `_shouldPersistLinkedOnly` uses name-only set**
`linkedWidgetNames` is a `Set<string>` of just widget names. With two linked entries both named `seed`, the set contains one entry. A fallback stored entry for a different source node with `widgetName === "seed"` incorrectly matches and survives pruning. Should key on `interiorNodeId:widgetName` composite (same as `_makePromotionEntryKey`):
```ts
const linkedEntryKeys = new Set(
  linkedEntries.map((entry) =>
    this._makePromotionEntryKey(entry.interiorNodeId, entry.widgetName)
  )
)
```

**3. `proxyWidgets` schema forward-compatibility**
The Zod schema `z.tuple([z.string(), z.string()])` is strict-length. If a future version extends to 3-tuples, older schemas will silently reject them and `parseProxyWidgets` returns `[]`. Consider `z.tuple([z.string(), z.string()]).rest(z.unknown())` to tolerate future extensions. The PR description mentions 3-tuple support but the code only writes 2-tuples — worth clarifying which is intended.

### Observations (not blocking)

- `matchPromotedInput` name-only fallback correctly guards with `sameNameMatches.length === 1` — no false match for duplicate names. The `_widget === widget` exact match path handles the new scenario correctly.
- `PromotedWidgetViewManager.reconcile()` keying on `nodeId:widgetName:viewKey` is well-designed. No O(n²) introduced.
- Backward compatibility for 2-tuple `proxyWidgets` is intact — old workflows load correctly.
- `NodeWidgets.vue` border styling change (`isPromotedView` → host-node lookup) is directionally correct but the `!!widget.nodeId` discriminator can be unreliable when resolution fails. Consider using `!!widget.storeName` which is only set for promoted widgets.

### Test coverage

Unit tests are strong — the duplicate-name promotion test, deduplication logic in `NodeWidgets.test.ts`, and `useGraphNodeManager.test.ts` mapper tests all cover the core disambiguation path well. One gap: no test for `removeWidgetByName` with duplicate names, and no test for `_shouldPersistLinkedOnly` with duplicate-name linked entries.

The E2E test asset (`nested-duplicate-widget-names.json`) is a good addition. The spec should verify both widgets render visibly — not just that the node has the right widget count.

Overall this is well-architected. The key insight — that `interiorNodeId` already serves as the natural disambiguator — keeps the change minimal. The three issues above are all consequences of existing name-only lookups that predate this PR but become bugs in the new duplicate-name world.

src/core/graph/subgraph/subgraphNodePromotion.test.ts
```diff
@@ -227,7 +227,8 @@ describe('Subgraph proxyWidgets', () => {
       ]
     )
 
-    subgraphNode.removeWidgetByName('widgetA')
+    const widgetA = subgraphNode.widgets.find((w) => w.name === 'widgetA')!
```
artokun, line 228: should we use findAll() do a length check, then pick index 0? Or we can loop over entries so we can handle each entry

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -440,8 +440,10 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
     )
       return false
 
-    const linkedWidgetNames = new Set(
-      linkedEntries.map((entry) => entry.widgetName)
+    const linkedEntryKeys = new Set(
+      linkedEntries.map((entry) =>
+        this._makePromotionEntryKey(entry.interiorNodeId, entry.widgetName)
```
artokun: We should extend this at some point to support nested subgraphNodeIds

artokun: this way we can recursively punch through each nested subnode layer during promotion for both reconstruction and deconstruction

artokun (commented):
I think this change looks good, but we may want to add a TODO for subgraph recursion

src/components/rightSidePanel/subgraph/SubgraphEditor.vue
```diff
@@ -147,30 +177,25 @@ function demote([node, widget]: WidgetItem) {
...
-    String(node.id),
-    widget.name
-  )
 }
 function showAll() {
   const node = activeNode.value
   if (!node) return
   for (const [n, w] of filteredCandidates.value) {
-    promotionStore.promote(node.rootGraph.id, node.id, String(n.id), w.name)
+    promotionStore.promote(
```
AustinMroz: Can we get a variant that takes just `LGraphNode`  and `IBaseWidget`? Same for demote. The complexity should really be contained.

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -224,19 +224,25 @@ function safeWidgetMapper(
...
   }
 
   function resolvePromotedWidgetIdentity(widget: IBaseWidget): {
     displayName: string
-    promotedSource: { sourceNodeId: string; sourceWidgetName: string } | null
+    promotedSource: {
+      sourceNodeId: string
+      sourceWidgetName: string
+      disambiguatingSourceNodeId?: string
+    } | null
```
AustinMroz: With the added complexity now, I'd appreciate if the promotedSource tuple was typed

src/composables/graph/useGraphNodeManager.ts
```diff
@@ -297,7 +304,8 @@ function safeWidgetMapper(
           ? resolveConcretePromotedWidget(
               node,
               promotedSource.sourceNodeId,
-              promotedSource.sourceWidgetName
+              promotedSource.sourceWidgetName,
```
AustinMroz, line 304: and the functions took the wrapped promotedSource as an argument.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -124,7 +129,9 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
 
   private _resolveLinkedPromotionBySubgraphInput(
     subgraphInput: SubgraphInput
-  ): { interiorNodeId: string; widgetName: string } | undefined {
+  ):
+    | { interiorNodeId: string; widgetName: string; sourceNodeId?: string }
```
AustinMroz: Use same type here and unify keys?

AustinMroz (commented):
Ok, yeah. This does make me sad

I no longer feel like interacting with the system is impossible, but I'm rather disappointed that this is greater complexity being pushed through the rest of the code base.

I particularly find myself thinking back to my implementation, where fixing the label display would have taken ~10 lines.

---

No blocking issues from me.

I marked a couple, but there's a whole lot of repeated destructuring of objects for function calls that should really just take the object.

AustinMroz (approved):

PR #10290, opened 2026-03-19
test: verify links render correctly after mode switch

DrJKL:
A link drawn once may vanish. A link tested persists.


browser_tests/tests/selectionToolbox.spec.ts-snapshots/selection-toolbox-multiple-nodes-border-chromium-linux.png
pythongosssss: Not sure these should be updated

browser_tests/tests/interaction.spec.ts-snapshots/dragged-node1-chromium-linux.png
AustinMroz, line 1: Aren't both these snapshot updates flakes?

browser_tests/tests/vueNodes/interactions/modeSwitchLinks.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+      await comfyPage.workflow.loadWorkflow('default')
+      await comfyPage.vueNodes.waitForNodes()
+      await fitToViewInstant(comfyPage)
+    })
+
+    test('links render correctly after switching to app mode and back', async ({
+      comfyPage
+    }) => {
+      // Capture baseline screenshot with links visible in graph mode
+      await expect(comfyPage.canvas).toHaveScreenshot(
```
AustinMroz, line 71: This right here is the very embodiment of everything I hate about our current test system. This tests nothing. There will never be a scenario where a person sees this first expectation fail and says "I'm glad this test exists". This is awful churn.

At the bare minimum, I'd want a single snapshot cropped to only checks links. If you're cool (and it's possible with vue positioning :cry: ), I think think Simon added some workflow switching tests that compare base64 snapshots across actions that would ideal for our actual test here.

DrJKL: If the links did disappear or shift wildly, do you think you'd notice?

browser_tests/tests/interaction.spec.ts-snapshots/dragged-node1-chromium-linux.png
DrJKL: Very probably

PR #10303, opened 2026-03-19
refactor: use trigger-owned context menus

benceruleanlu:
## Summary

Make the queue and asset right-click menus trigger-owned so Reka owns open and close behavior locally instead of parents storing current items and opening imperative wrappers.

## Changes

- **What**: Convert `JobContextMenu` and `MediaAssetContextMenu` into local `ContextMenuRoot` wrappers and let the shared menu panels close through injected Reka root context.
- **What**: Remove parent-managed context-menu refs and current-item plumbing from the queue overlay, job history, and assets sidebar list/grid surfaces.
- **What**: Update the focused menu tests to exercise the trigger-owned right-click path.

## Review Focus

Verify that right-click menus still behave correctly across queue overlay, job history, and asset list/grid views, especially action selection, menu dismissal, and the asset scroll-close path.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10303-refactor-use-trigger-owned-context-menus-3286d73d3650811788e3caef466aa08b) by [Unito](https://www.unito.io)

### DIFF
--- a/src/workbench/extensions/manager/components/ManagerProgressToast.vue
+++ b/src/workbench/extensions/manager/components/ManagerProgressToast.vue
@@ -135,6 +135,7 @@ whenever(() => !isExpanded.value, resetUserScrolling)
 
 function closeToast() {
   comfyManagerStore.resetTaskState()
+  isRestartCompleted.value = false
   isExpanded.value = false
 }
 
--- a/src/workbench/extensions/manager/components/manager/button/PackInstallButton.vue
+++ b/src/workbench/extensions/manager/components/manager/button/PackInstallButton.vue
@@ -3,7 +3,7 @@
     variant="primary"
     :size
     :disabled="isLoading || isInstalling"
-    @click="installAllPacks"
+    @click.stop="installAllPacks"
   >
     <i
       v-if="hasConflict && !isInstalling && !isLoading"


src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -16,49 +16,79 @@
...
+                >
+                  <Button
+                    variant="secondary"
+                    size="icon"
+                    :aria-label="t('mediaAsset.actions.moreOptions')"
+                  >
+                    <i class="icon-[lucide--ellipsis] size-4" />
+                  </Button>
+                </MediaAssetActionsMenu>
+              </template>
```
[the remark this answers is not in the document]
benceruleanlu: Addressed in `ef546e3da`. `AssetsSidebarListView` now tracks `openActionsAssetId`, keeps `MediaAssetActionsMenu` mounted while the row is hovered or the dropdown is open, and stops the trigger click from reaching the row selection handler.

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -67,16 +64,30 @@
...
-            <i class="icon-[lucide--ellipsis] size-4" />
-          </Button>
+            <Button
+              variant="overlay-white"
+              size="icon"
+              :aria-label="$t('mediaAsset.actions.moreOptions')"
+            >
+              <i class="icon-[lucide--ellipsis] size-4" />
+            </Button>
+          </MediaAssetActionsMenu>
```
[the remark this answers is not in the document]
benceruleanlu: Addressed in `ef546e3da`. `MediaAssetCard` now controls the actions menu with `v-model:open`, keeps the overlay visible while the dropdown is open, and stops the trigger click from bubbling to the card root.

src/platform/assets/components/MediaAssetMenuPanel.vue
```diff
@@ -0,0 +1,270 @@
...
+    if (shouldShowDeleteButton.value) {
+      items.push({
+        key: 'bulk-delete',
+        label: t('mediaAsset.selection.deleteSelectedAll'),
+        icon: 'icon-[lucide--trash-2]',
+        onSelect: () => emit('bulk-delete', selectedAssets)
+      })
+    }
+
+    return items
```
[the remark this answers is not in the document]
benceruleanlu: Addressed in `ef546e3da`. `MediaAssetMenuPanel` now computes aggregate bulk capability flags and only renders the bulk workflow entries when the entire selected set supports those actions.

src/components/queue/job/JobAssetsList.vue
```diff
@@ -15,64 +15,64 @@
...
+            :icon-class="getJobIconClass(job)"
+            :primary-text="job.title"
+            :secondary-text="job.meta"
+            :progress-total-percent="job.progressTotalPercent"
+            :progress-current-percent="job.progressCurrentPercent"
+            @dblclick.stop="emitViewItem(job)"
+            @preview-click="emitViewItem(job)"
+            @click.stop
+          >
+            <template v-if="hoveredJobId === job.id" #actions>
```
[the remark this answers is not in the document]
benceruleanlu: No longer valid on the latest branch state. `JobAssetsList` now keeps the actions menu mounted while `openActionsJobId === job.id` and wires `@update:open` to preserve the dropdown after hover exit. I also verified that path with the new browser regression in `browser_tests/tests/sidebar/jobHistorySidebar.spec.ts`.

src/platform/assets/components/MediaAssetActionsMenu.vue
```diff
@@ -0,0 +1,77 @@
...
+      <DropdownMenuContent
+        :side-offset="4"
+        :collision-padding="8"
+        class="z-50 bg-transparent p-0 shadow-lg"
+      >
+        <MediaAssetMenuPanel
+          :asset
+          :asset-type
+          :file-kind
+          :show-delete-button
```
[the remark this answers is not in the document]
benceruleanlu: No longer valid on the latest branch state. `MediaAssetActionsMenu` now passes `close-on-scroll` into `MediaAssetMenuPanel`, and the asset browser path is covered by `browser_tests/tests/sidebar/assetsSidebar.spec.ts`, which verifies the menu dismisses on scroll.

browser_tests/tests/sidebar/jobHistorySidebar.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    await comfyPage.queue.mockHistory([
+      { promptId: 'history-job-1', status: 'success' }
+    ])
+    await comfyPage.settings.setSetting('Comfy.Queue.QPOV2', true)
+    await comfyPage.setup({ clearStorage: false })
+
+    await comfyPage.page.getByTestId('queue-overlay-toggle').click()
+    await expect(
+      comfyPage.page.locator('[data-job-id="history-job-1"]')
+    ).toBeVisible()
```
DrJKL: This would be a good time to start adding more Page Object style structure to our Page Object, if you're interested.
We have some examples in ComfyPage already, but the main idea is to put the locators into an object where the aspects of the page/components are exposed to the test in a more user-centric style.
So for here, something like `comfyPage.openJobQueue()`

browser_tests/tests/sidebar/jobHistorySidebar.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+  })
+
+  test('hover popover and actions menu stay clickable', async ({
+    comfyPage
+  }) => {
+    const jobRow = comfyPage.page.locator('[data-job-id="history-job-1"]')
```
DrJKL: And down here, something like `comfyPage.queue.getNthJobRow(0)`

src/components/queue/job/JobActionsMenu.vue
DrJKL, line 1: This looks fairly generic, can we put it with the other components and wire through the content with a second named slot?

src/components/queue/job/JobAssetsList.test.ts
```diff
@@ -32,43 +37,34 @@ vi.mock('vue-i18n', () => {
...
-  const flatOutputs = preview ? [preview] : []
-  return new TaskItemImpl(job, {}, flatOutputs)
-}
+    mediaType,
+    isImage: mediaType === 'images',
+    isVideo: mediaType === 'video',
+    isAudio: mediaType === 'audio',
+    is3D: mediaType === 'model',
+    url: `/api/view/${filename}`
+  }) as NonNullable<NonNullable<JobListItem['taskRef']>['previewOutput']>
```
DrJKL: Super-duper NonNullable, I promise

src/components/queue/job/JobAssetsList.vue
```diff
@@ -15,72 +15,74 @@
...
+                <i class="icon-[lucide--trash-2] size-4" />
+              </Button>
+              <Button
+                v-else-if="job.state === 'completed'"
+                variant="textonly"
+                size="sm"
+                @click.stop="emitCompletedViewItem(job)"
+              >
+                {{ t('menuLabels.View') }}
+              </Button>
```
DrJKL: Instead of conditional components, can we make the aspects themselves a computed object in the script and have a single Button?

src/components/queue/job/JobAssetsList.vue
```diff
@@ -188,6 +197,21 @@ function isFailedDeletable(job: JobListItem) {
   return job.showClear !== false && job.state === 'failed'
 }
 
+function shouldShowActionsMenu(jobId: string) {
+  return hoveredJobId.value === jobId || openActionsJobId.value === jobId
+}
+
+function onActionsMenuOpenChange(jobId: string, isOpen: boolean) {
```
DrJKL: I would have expected the component to handle this more implicitly.

src/components/queue/job/JobContextMenu.vue
DrJKL, line 1: Same for this one.

DrJKL (commented):
Partway through, leaving the first round of comments.

src/components/sidebar/tabs/AssetsSidebarListView.test.ts
```diff
@@ -131,4 +200,50 @@ describe('AssetsSidebarListView', () => {
 
     expect(wrapper.emitted('preview-asset')).toEqual([[imageAsset]])
   })
+
+  it('keeps row actions mounted while the dropdown is open', async () => {
```
DrJKL: Question: Is there a way to frame this as something a User would experience?

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -16,49 +16,82 @@
...
+            :show-delete-button
+            :selected-assets
+            :is-bulk-mode
+            @zoom="emit('preview-asset', item.asset)"
+            @asset-deleted="emit('asset-deleted')"
+            @bulk-download="emit('bulk-download', $event)"
+            @bulk-delete="emit('bulk-delete', $event)"
+            @bulk-add-to-workflow="emit('bulk-add-to-workflow', $event)"
+            @bulk-open-workflow="emit('bulk-open-workflow', $event)"
+            @bulk-export-workflow="emit('bulk-export-workflow', $event)"
```
DrJKL: non-blocking: If we have to pipe these through multiple layers, it might be worth extracting a shared composable that we could invoke the actions on directly. Unless we have multiple assets panels at once, I wouldn't reach for provide/inject.

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -16,49 +16,82 @@
...
+              "
+              :preview-url="getAssetPreviewUrl(item.asset)"
+              :preview-alt="getAssetDisplayName(item.asset)"
+              :icon-name="iconForMediaType(getAssetMediaType(item.asset))"
+              :is-video-preview="isVideoAsset(item.asset)"
+              :primary-text="getAssetPrimaryText(item.asset)"
+              :secondary-text="getAssetSecondaryText(item.asset)"
+              :stack-count="getStackCount(item.asset)"
+              :stack-indicator-label="t('mediaAsset.actions.seeMoreOutputs')"
+              :stack-expanded="isStackExpanded(item.asset)"
```
DrJKL: Would it be possible to just pass in the asset and have the item do the detail extraction?

src/components/sidebar/tabs/AssetsSidebarListView.vue
```diff
@@ -180,6 +228,23 @@ function getAssetCardClass(selected: boolean): string {
   )
 }
 
+function shouldShowActionsMenu(assetId: string): boolean {
+  return (
+    hoveredAssetId.value === assetId || openActionsAssetId.value === assetId
+  )
+}
+
+function onActionsMenuOpenChange(assetId: string, isOpen: boolean): void {
```
DrJKL: I think if the menu took the asset in, we wouldn't have to track which asset was clicked on, right?

src/components/sidebar/tabs/JobHistorySidebarTab.test.ts
```diff
@@ -5,6 +5,49 @@ import { defineComponent, nextTick } from 'vue'
 
 import JobHistorySidebarTab from './JobHistorySidebarTab.vue'
 
+vi.mock('@/components/queue/job/JobDetailsPopover.vue', () => ({
```
DrJKL: Just an observation, this is a lot to have to stub out.

src/platform/assets/components/MediaAssetActionsMenu.vue
DrJKL, line 1: If we had the shared Dropdown and extracted the assets menu options into a composable, we could probably delete at least 3 separate components.

src/platform/assets/components/MediaAssetMenuPanel.vue
```diff
@@ -0,0 +1,314 @@
...
+
+const shouldShowDeleteButton = computed(() => {
+  const propAllows = showDeleteButton ?? true
+  const typeAllows =
+    assetType === 'output' || (assetType === 'input' && isCloud)
+
+  return propAllows && typeAllows
+})
+
+const contextMenuItems = computed<ContextMenuItem[]>(() => {
```
DrJKL: nit: This is a lot of conditions to reason about in one computed.

DrJKL (commented):
## Review: trigger-owned context menus

Nice refactor. The core direction is strong — migrating from PrimeVue `ContextMenu` + the bespoke `useDismissableOverlay` to Reka UI's native `ContextMenuRoot`/`DropdownMenuRoot` primitives is a clear win. The trigger-owned pattern eliminates the fragile imperative `.show()`/`.hide()` choreography and the shared `currentMenuItem` ref that was a latent source of stale-state bugs.

**What works well:**
- Clean deletion of `useDismissableOverlay` and its test — no dangling imports remain.
- The `JobMenuPanel` / `MediaAssetMenuPanel` extraction gives both right-click and dropdown menus identical behavior from a single source of truth.
- `buildJobMenuEntries(item)` passing the target as a parameter instead of closing over mutable ref state is a much safer design.
- Tests are properly behavioral — they open menus via DOM events, assert on emitted actions, and test scroll-dismiss. No change-detector tests.
- All 28 tests pass, event forwarding chains are consistent, no broken imperative callers remain.

**Areas to address** (see inline comments):

src/platform/assets/components/MediaAssetMenuPanel.vue
```diff
@@ -0,0 +1,314 @@
...
+  zoom: []
+  'asset-deleted': []
+  'bulk-download': [assets: AssetItem[]]
+  'bulk-delete': [assets: AssetItem[]]
+  'bulk-add-to-workflow': [assets: AssetItem[]]
+  'bulk-open-workflow': [assets: AssetItem[]]
+  'bulk-export-workflow': [assets: AssetItem[]]
+}>()
+
+type ContextMenuItem =
```
DrJKL: 🔴 **Type name shadows imported component**

Line 36 imports `ContextMenuItem` as a component value from `reka-ui`. This line declares a local `type ContextMenuItem`, shadowing it. It works at runtime (the type is erased), but it's confusing and fragile — a reader seeing `ContextMenuItem` in the template vs in a function signature gets two different things.

Rename the local type, e.g. `MenuItemEntry` or `AssetMenuEntry`.

src/platform/assets/components/MediaAssetMenuPanel.vue
```diff
@@ -0,0 +1,314 @@
...
+      onSelect?: () => void | Promise<void>
+    }
+  | {
+      kind: 'divider'
+      key: string
+    }
+
+const actions = useMediaAssetActions()
+const { t } = useI18n()
+const dropdownMenuRootContext = injectDropdownMenuRootContext(null)
```
DrJKL: 🟡 **Relies on Reka UI internal injection functions**

`injectDropdownMenuRootContext` and `injectContextMenuRootContext` are not part of Reka UI's documented public API — they're internal context helpers. This is a pragmatic pattern but could break silently on a Reka UI update.

Two options:
1. Add a brief comment noting this couples to reka-ui internals, so future maintainers know to check on upgrades.
2. Alternatively, accept a `mode: 'dropdown' | 'context'` prop from the parent (which already knows which mode it is) and select components based on that. Simpler and decoupled.

Same applies to `JobMenuPanel.vue:63-64`.

src/platform/assets/components/MediaAssetMenuPanel.vue
```diff
@@ -0,0 +1,314 @@
...
+const separatorComponent = dropdownMenuRootContext
+  ? DropdownMenuSeparator
+  : ContextMenuSeparator
+
+function closeMenu() {
+  dropdownMenuRootContext?.onOpenChange(false)
+  contextMenuRootContext?.onOpenChange(false)
+}
+
+useEventListener(
```
DrJKL: 🟡 **Scroll-dismiss may be redundant**

Reka UI's `ContextMenuContent` and `DropdownMenuContent` already close on outside interaction and scroll by default. Is this manual `useEventListener(window, 'scroll', ...)` needed? If so, consider moving the listener to the wrapper components (`MediaAssetContextMenu`/`MediaAssetActionsMenu`) rather than the shared panel — keeping dismissal policy in the shell and menu-item logic in the panel is a cleaner separation.

src/platform/assets/components/MediaAssetMenuPanel.vue
```diff
@@ -0,0 +1,314 @@
...
+          emit('asset-deleted')
+        }
+      }
+    })
+  }
+
+  return items
+})
+
+async function onItemSelect(item: ContextMenuItem) {
```
DrJKL: 🟡 **Manual `closeMenu()` likely redundant**

Reka UI `ContextMenuItem`/`DropdownMenuItem` auto-close the menu on select by default. Calling `closeMenu()` before `item.onSelect()` may race with Reka's own close animation. Consider removing `closeMenu()` here and letting Reka handle it, or verify this is needed for a specific edge case and add a comment explaining why.

src/components/queue/job/JobMenuPanel.vue
```diff
@@ -0,0 +1,85 @@
...
+          class="w-full justify-start bg-transparent data-highlighted:bg-secondary-background-hover"
+          variant="textonly"
+          size="sm"
+          :aria-label="entry.label"
+          :disabled="entry.disabled"
+        >
+          <i
+            v-if="entry.icon"
+            :class="[
+              entry.icon,
```
DrJKL: 🟡 **`:class="[]"` array syntax — must use `cn()`**

Project convention (AGENTS.md): "NEVER use `:class="[]"` to merge class names — always use `cn()` from `@/utils/tailwindUtil`".

```suggestion
            :class="cn(entry.icon, 'block size-4 shrink-0 leading-none text-text-secondary')"
```

src/components/queue/job/JobMenuPanel.vue
```diff
@@ -0,0 +1,85 @@
...
+
+import Button from '@/components/ui/button/Button.vue'
+import type { MenuEntry } from '@/composables/queue/useJobMenu'
+
+const { entries } = defineProps<{
+  entries: MenuEntry[]
+}>()
+
+const emit = defineEmits<{
+  (e: 'action', entry: MenuEntry): void
```
DrJKL: 🟢 **Inconsistent `defineEmits` style**

This uses the function-call syntax `(e: 'action', entry: MenuEntry): void`. The MediaAsset files introduced in the same PR use the modern object/tuple syntax `action: [entry: MenuEntry]`. Standardizing within the PR would be nice.

Same applies to `JobActionsMenu.vue:45-47` and `JobAssetsList.vue:123-128`.

src/composables/queue/useJobMenu.ts
```diff
@@ -347,27 +346,30 @@ export function useJobMenu(
...
+        onClick: () => cancelJob(target)
       }
     ]
-  })
+  }
+
+  const jobMenuEntries = computed<MenuEntry[]>(() => buildJobMenuEntries())
 
   return {
     jobMenuEntries,
```
DrJKL, line 370: 🟡 **`jobMenuEntries` is dead production code**

All production consumers (`QueueOverlayExpanded.vue`, `JobHistorySidebarTab.vue`) destructure `getJobMenuEntries` — nobody uses the `jobMenuEntries` computed. The only references are in `useJobMenu.test.ts`.

Consider removing `jobMenuEntries` from the return object and updating the tests to call `getJobMenuEntries(item)` directly. This also lets you delete the `currentMenuItem` parameter entirely, simplifying `resolveItem`.

src/platform/assets/components/MediaAssetMenuPanel.vue
```diff
@@ -0,0 +1,314 @@
...
+  fileKind: MediaKind
+  showDeleteButton?: boolean
+  selectedAssets?: AssetItem[]
+  isBulkMode?: boolean
+  closeOnScroll?: boolean
+}>()
+
+const emit = defineEmits<{
+  zoom: []
+  'asset-deleted': []
```
DrJKL: 🟡 **Prop/event drilling is verbose — consider a shared composable**

`showDeleteButton`, `selectedAssets`, `isBulkMode` + 5 bulk events are forwarded through 3-4 layers (Tab → Grid/ListView → ContextMenu/Card/ActionsMenu → MenuPanel). Functionally correct, but a shared composable (e.g. `useAssetSelectionContext()`) that the parent populates and the leaf consumes directly would collapse the forwarding chain without adding DI complexity. Not blocking, but worth considering as a follow-up.

src/platform/assets/components/MediaAssetActionsMenu.vue
```diff
@@ -0,0 +1,78 @@
...
+
+import Button from '@/components/ui/button/Button.vue'
+
+import type { AssetItem } from '../schemas/assetSchema'
+import type { AssetContext, MediaKind } from '../schemas/mediaAssetSchema'
+import MediaAssetMenuPanel from './MediaAssetMenuPanel.vue'
+
+const open = defineModel<boolean>('open', { default: false })
+
+defineProps<{
```
DrJKL: 🟢 **Props not destructured**

Project convention (Vue 3.5 style) prefers reactive props destructuring:
```ts
const { asset, assetType, ... } = defineProps<{...}>()
```

Not blocking since the props are only used in the template via same-name shorthand, but inconsistent with `MediaAssetContextMenu.vue` which does destructure.

browser_tests/tests/sidebar/jobHistorySidebar.spec.ts
benceruleanlu: Marking this to do in a followup

browser_tests/tests/sidebar/jobHistorySidebar.spec.ts
benceruleanlu: Marking this to do in a followup

src/components/queue/job/JobActionsMenu.vue
benceruleanlu: yes! did this and collapsed to context menu dropdown menu and menu panel

src/components/queue/job/JobAssetsList.test.ts
benceruleanlu: Extracted to a small alias

src/components/queue/job/JobAssetsList.vue
benceruleanlu: I left that structure as-is since it predates the Reka migration and isn’t really part of the menu refactor. Happy to clean it up separately if you want.

src/components/queue/job/JobContextMenu.vue
benceruleanlu: Assuming you meant the consolidation and generalization, which happened

src/components/queue/job/JobAssetsList.vue
benceruleanlu: Addressed in `2a07ae90a`. `useJobDetailsHover` now consistently runs its existing `onReset` hook when details are cleared, so `JobAssetsList` no longer has to thread `clearPopoverAnchor` through each `scheduleDetailsHide(...)` call.

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: Addressed in the latest branch state. I extracted the shared asset menu/action construction into useMediaAssetMenu, and the list/grid/card paths now all invoke that composable directly instead of each owning their own menu logic. There is still some prop/event forwarding for selection/bulk state, but I didn’t add provide/inject here.

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: I want to keep AssetsListItem presentational here. It’s shared by both asset and job rows, so pushing AssetItem-specific extraction into it would make the component domain-aware and less reusable. If we wanted that shape, I think it would be better to introduce a dedicated asset row component rather than overload AssetsListItem.

src/components/sidebar/tabs/AssetsSidebarListView.vue
benceruleanlu: The menu already captures the asset for its actions. openActionsAssetId here is only to keep that row’s actions slot mounted while the dropdown is open after hover exit; it’s not tracking asset identity for the action handlers.

src/platform/assets/components/MediaAssetActionsMenu.vue
benceruleanlu: Yes, did that

src/platform/assets/components/MediaAssetMenuPanel.vue
benceruleanlu: Think this one got broken down during the consolidation

src/platform/assets/components/MediaAssetMenuPanel.vue
benceruleanlu: Think this one got broken down during the consolidation

src/platform/assets/components/MediaAssetMenuPanel.vue
benceruleanlu: injectDropdownMenuRootContext is gone, injectContextMenuRootContext is still here, I think it will be fine for now, unfortunate that Reka doesn't expose it. Also I'm not too sure what the second option refers to, this is used to close the menu on scroll.

src/platform/assets/components/MediaAssetMenuPanel.vue
benceruleanlu: I could not find any built-in scroll listener that guarantees dismissal on scroll

src/platform/assets/components/MediaAssetMenuPanel.vue
benceruleanlu: Agree, removed four instances of this

src/components/queue/job/JobMenuPanel.vue
benceruleanlu: Switched to cn after the consolidation

benceruleanlu: Maybe we should have a lint rule for this, my agent is telling me we have other instances of this in the codebase

src/components/queue/job/JobMenuPanel.vue
benceruleanlu: These files got deleted in the consolidation

src/composables/queue/useJobMenu.ts
benceruleanlu: Wow I missed this, removed now

src/platform/assets/components/MediaAssetActionsMenu.vue
benceruleanlu: This file is gone now

src/platform/assets/components/MediaAssetMenuPanel.vue
benceruleanlu: I would put this one off for a followup

src/components/sidebar/tabs/AssetsSidebarListView.test.ts
benceruleanlu: Sure, I renamed it to keeps the row actions menu available after the row loses hover

src/components/common/MenuPanel.vue
DrJKL, line 1: This feels like overkill. Do we need the two generic components?
Feel free to file an issue assigned to me to see if I can obviate this for the existing use cases.

src/components/common/ContextMenu.vue
DrJKL, line 1: Could we add the full Shadcn/vue versions alongside the other modern components?
https://www.shadcn-vue.com/docs/components/context-menu
https://www.shadcn-vue.com/docs/components/dropdown-menu

Maybe as a prerequisite PR?

DrJKL (changes requested):

PR #10309, opened 2026-03-19
feat: detect and resolve missing media inputs in error tab

jaeone94:
## Summary

Add detection and resolution UI for missing image/video/audio inputs (LoadImage, LoadVideo, LoadAudio nodes) in the Errors tab, mirroring the existing missing model pipeline.

## Changes

- **What**: New `src/platform/missingMedia/` module — scan pipeline detects missing media files on workflow load (sync for OSS, async for cloud), surfaces them in the error tab with upload dropzone, thumbnail library select, and 2-step confirm flow
- **Detection**: `scanAllMediaCandidates()` checks combo widget values against options; cloud path defers to `verifyCloudMediaCandidates()` via `assetsStore.updateInputs()`
- **UI**: `MissingMediaCard` groups by media type; `MissingMediaRow` shows node name (single) or filename+count (multiple), upload dropzone with drag & drop, `MissingMediaLibrarySelect` with image/video thumbnails
- **Resolution**: Upload via `/upload/image` API or select from library → status card → checkmark confirm → widget value applied, item removed from error list
- **Integration**: `executionErrorStore` aggregates into `hasAnyError`/`totalErrorCount`; `useNodeErrorFlagSync` flags nodes on canvas; `useErrorGroups` renders in error tab
- **Shared**: Extract `ACCEPTED_IMAGE_TYPES`/`ACCEPTED_VIDEO_TYPES` to `src/utils/mediaUploadUtil.ts`; extract `resolveComboValues` to `src/utils/litegraphUtil.ts` (shared across missingMedia + missingModel scan)
- **Reverse clearing**: Widget value changes on nodes auto-remove corresponding missing media errors (via `clearWidgetRelatedErrors`)

### Unit tests (22 tests)
- `missingMediaScan.test.ts` (12): groupCandidatesByName, groupCandidatesByMediaType (ordering, multi-name), verifyCloudMediaCandidates (missing/present, abort before/after updateInputs, already resolved true/false, no-pending skip, updateInputs spy)
- `missingMediaStore.test.ts` (10): setMissingMedia, clearMissingMedia (full lifecycle with interaction state), missingMediaNodeIds, hasMissingMediaOnNode, removeMissingMediaByWidget (match/no-match/last-entry), createVerificationAbortController

### E2E tests (10 scenarios in `missingMedia.spec.ts`)
- Detection: error overlay shown, Missing Inputs group in errors tab, correct row count, dropzone + library select visibility, no false positive for valid media
- Upload flow: file picker → uploading status card → confirm → row removed
- Library select: dropdown → selected status card → confirm → row removed
- Cancel: pending selection → returns to upload/library UI
- All resolved: Missing Inputs group disappears
- Locate node: canvas pans to missing media node

## Review Focus

- Cloud verification path: `verifyCloudMediaCandidates` compares widget value against `asset_hash` — implicit contract
- 2-step confirm mirrors missing model pattern (`pendingSelection` → confirm/cancel)
- Event propagation guard on dropzone (`@drop.prevent.stop`) to prevent canvas LoadImage node creation
- `clearAllErrors()` intentionally does NOT clear missing media (same as missing models — preserves pending repairs)
- `runMissingMediaPipeline` is now `async` and `await`-ed, matching model pipeline

## Test plan

- [x] OSS: load workflow with LoadImage referencing non-existent file → error tab shows it
- [x] Upload file via dropzone → status card shows "Uploaded" → confirm → widget updated, error removed
- [x] Select from library with thumbnail preview → confirm → widget updated, error removed
- [x] Cancel pending selection → returns to upload/library UI
- [x] Load workflow with valid images → no false positives
- [x] Click locate-node → canvas navigates to the node
- [x] Multiple nodes referencing different missing files → correct row count
- [x] Widget value change on node → missing media error auto-removed

## Screenshots

https://github.com/user-attachments/assets/631c0cb0-9706-4db2-8615-f24a4c3fe27d


src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -11,8 +11,31 @@
       />
     </div>
 
-    <!-- Scrollable content -->
-    <div class="min-w-0 flex-1 overflow-y-auto" aria-live="polite">
+    <!-- Runtime error: full-height panel outside accordion -->
+    <div
+      v-if="singleRuntimeErrorCard"
+      data-testid="runtime-error-panel"
+      class="flex min-h-0 flex-1 flex-col overflow-hidden px-4 py-3"
```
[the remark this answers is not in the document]
jaeone94: Fixed in bb310d938 — added aria-live to runtime panel.

src/components/rightSidePanel/errors/useErrorActions.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  function contactSupport() {
+    telemetry?.trackHelpResourceClicked({
+      resource_type: 'help_feedback',
+      is_external: true,
+      source: 'error_dialog'
+    })
+    void commandStore.execute('Comfy.ContactSupport')
+  }
```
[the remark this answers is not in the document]
jaeone94: Fixed in bb310d938 — changed `void commandStore.execute` to `return commandStore.execute` so rejections propagate to callers.

src/components/rightSidePanel/errors/useErrorReport.ts
```diff
@@ -0,0 +1,82 @@
...
+          systemStats: systemStatsStore.systemStats,
+          serverLogs: logs,
+          workflow
+        })
+        enrichedDetails[idx] = report
+      } catch (e) {
+        console.warn('Failed to generate error report:', e)
+      }
+    }
+  })
```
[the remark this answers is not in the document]
jaeone94: Fixed in bb310d938 — replaced `onMounted` with `watch(cardSource, ..., { immediate: true })` using `onCleanup` for cancellation. Clears `enrichedDetails` before each run so stale mappings don't persist.

src/composables/graph/useNodeErrorFlagSync.ts
```diff
@@ -0,0 +1,120 @@
...
+          ? missingModelStore.missingModelAncestorExecutionIds
+          : new Set(),
+        showErrorsTab
+          ? missingMediaStore.missingMediaAncestorExecutionIds
+          : new Set()
+      )
+    },
+    { flush: 'post' }
+  )
+  return stop
```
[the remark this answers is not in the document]
jaeone94: This is a pre-existing design choice. `lastExecutionError` has different lifecycle semantics (persists until next execution vs graph reload). Adding it to reconciliation requires careful handling of clear vs retain timing. Will track as a follow-up issue.

src/composables/graph/useNodeErrorFlagSync.ts
```diff
@@ -0,0 +1,120 @@
...
+        lastNodeErrors.value,
+        showErrorsTab
+          ? missingModelStore.missingModelAncestorExecutionIds
+          : new Set(),
+        showErrorsTab
+          ? missingMediaStore.missingMediaAncestorExecutionIds
+          : new Set()
+      )
+    },
+    { flush: 'post' }
```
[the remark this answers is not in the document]
jaeone94: Already resolved after rebasing onto main — `showErrorsTab` is now a `computed` included in the watch sources array (line 101).

src/platform/missingMedia/components/MissingMediaRow.vue
```diff
@@ -0,0 +1,318 @@
...
+          )
+        "
+        @click="confirmSelection(item.name)"
+      >
+        <i
+          aria-hidden="true"
+          class="icon-[lucide--check] size-4"
+          :class="isPending ? 'text-primary' : 'text-foreground'"
+        />
+      </Button>
```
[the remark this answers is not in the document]
jaeone94: Fixed in cfe3917e9 — added `aria-label` via i18n (`confirmSelection`, `cancelSelection`) to both icon buttons.

src/platform/missingMedia/components/MissingMediaRow.vue
```diff
@@ -0,0 +1,318 @@
...
+  input.value = ''
+}
+
+function handleDrop(e: DragEvent) {
+  isDragOver.value = false
+  const file = e.dataTransfer?.files[0]
+  if (file) {
+    handleUpload(file, item.name, item.mediaType)
+  }
+}
```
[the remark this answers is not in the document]
jaeone94: Fixed in cfe3917e9 — both handlers are now `async` with `await handleUpload(...)`. File input cleanup uses `try/finally`.

src/scripts/app.ts
```diff
@@ -1553,6 +1566,36 @@ export class ComfyApp {
...
+          if (confirmed.length) {
+            useExecutionErrorStore().surfaceMissingMedia(confirmed)
+          }
+        })
+        .catch((err) => {
+          console.warn(
+            '[Missing Media Pipeline] Asset verification failed:',
+            err
+          )
+        })
```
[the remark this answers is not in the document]
jaeone94: Fixed in cfe3917e9 — added `useToastStore().add()` with i18n toast message in the `.catch()` block.

src/platform/missingMedia/composables/useMissingMediaInteractions.ts
```diff
@@ -0,0 +1,217 @@
...
+    typeof values === 'function'
+      ? values(result.widget)
+      : Array.isArray(values)
+        ? values
+        : Object.keys(values)
+
+  return list
+    .filter((v) => v !== candidate.name)
+    .map((v) => ({ name: getMediaDisplayName(v), value: v }))
+}
```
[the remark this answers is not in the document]
jaeone94: Fixed in cfe3917e9 — changed to `result.widget.options?.values` with optional chaining.

src/platform/missingMedia/composables/useMissingMediaInteractions.ts
```diff
@@ -0,0 +1,217 @@
...
+  for (const c of matching) {
+    const result = getMediaComboWidget(c)
+    if (!result) continue
+
+    addToComboValues(result.widget, newValue)
+    result.widget.value = newValue
+    result.widget.callback?.(newValue)
+    result.node.graph?.setDirtyCanvas(true, true)
+  }
+}
```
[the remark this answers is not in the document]
jaeone94: The existing `missingModel` pipeline (`useMissingModelInteractions.ts:184-186`) uses the identical pattern without `onWidgetChanged` or `_version++`. Undo/redo verified working in both pipelines. Changing one without the other creates inconsistency. Will address both together in a follow-up PR.

src/platform/missingMedia/composables/useMissingMediaInteractions.ts
```diff
@@ -0,0 +1,217 @@
...
+
+      const assetsStore = useAssetsStore()
+      await assetsStore.updateInputs()
+
+      store.uploadState[name] = { fileName: file.name, status: 'uploaded' }
+      store.pendingSelection[name] = uploadedPath
+    } catch (err) {
+      useToastStore().addAlert(String(err))
+      delete store.uploadState[name]
+    }
```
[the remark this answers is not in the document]
jaeone94: Fixed in 4947ba74b — MIME validation rejects empty `file.type`. Error toasts use `st('toastMessages.uploadFailed')` i18n messages. `updateInputs` failure is handled separately from upload failure.

src/platform/missingMedia/missingMediaScan.ts
```diff
@@ -0,0 +1,171 @@
...
+function isComboWidget(widget: IBaseWidget): widget is IComboWidget {
+  return widget.type === 'combo'
+}
+
+function resolveComboOptions(widget: IComboWidget): string[] {
+  const values = widget.options.values
+  if (!values) return []
+  if (typeof values === 'function') return values(widget)
+  if (Array.isArray(values)) return values
+  return Object.keys(values)
```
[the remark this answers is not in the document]
jaeone94: Fixed in 4947ba74b — `resolveComboValues` in `litegraphUtil.ts` uses `widget.options?.values` with optional chaining. The scan-local duplicate was removed.

src/platform/missingMedia/missingMediaStore.ts
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  function removeMissingMediaByWidget(nodeId: string, widgetName: string) {
+    if (!missingMediaCandidates.value) return
+    missingMediaCandidates.value = missingMediaCandidates.value.filter(
+      (m) => !(String(m.nodeId) === nodeId && m.widgetName === widgetName)
+    )
+    if (!missingMediaCandidates.value.length)
+      missingMediaCandidates.value = null
+  }
```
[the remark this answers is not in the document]
jaeone94: Fixed in bb310d938 — added `clearInteractionStateForName` that cleans up `expandState`, `uploadState`, and `pendingSelection`. Called from both `removeMissingMediaByName` and `removeMissingMediaByWidget` (preserves state if other candidates still share the name). 3 regression tests added.

browser_tests/tests/missingMedia.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+          : null
+      })
+
+      const locateButton = comfyPage.page.getByTestId(
+        TestIds.dialogs.missingMediaLocateButton
+      )
+      await expect(locateButton).toBeVisible()
+      await locateButton.click()
+
+      await comfyPage.page.waitForTimeout(500)
```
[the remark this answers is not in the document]
jaeone94: Fixed in bb310d938 — replaced `waitForTimeout(500)` with `expect.poll()` that retries until the canvas offset changes.

src/composables/graph/useNodeErrorFlagSync.ts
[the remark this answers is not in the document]
jaeone94: yes and asign it to me please

src/platform/missingMedia/composables/useMissingMediaInteractions.ts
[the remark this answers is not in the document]
jaeone94: yes and assign it to me please

browser_tests/tests/missingMedia.spec.ts
```diff
@@ -0,0 +1,220 @@
...
+      const librarySelect = comfyPage.page.getByTestId(
+        TestIds.dialogs.missingMediaLibrarySelect
+      )
+      await librarySelect.getByRole('combobox').click()
+
+      const optionCount = await comfyPage.page.getByRole('option').count()
+      if (optionCount === 0) {
+        test.skip()
+        return
+      }
```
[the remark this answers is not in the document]
jaeone94: The OSS test server's input directory content varies per environment — we can't guarantee specific library items exist. The `test.skip()` is a pragmatic guard that prevents false failures in CI. The upload flow test (which runs before this) already validates the full 2-step confirm path. Adding a fixture setup that pre-uploads a file would create test interdependency.

src/platform/missingMedia/composables/useMissingMediaInteractions.ts
```diff
@@ -0,0 +1,225 @@
...
+      }
+    } catch {
+      useToastStore().addAlert(
+        st(
+          'toastMessages.uploadFailed',
+          'Failed to upload file. Please try again.'
+        )
+      )
+      delete store.uploadState[name]
+    }
```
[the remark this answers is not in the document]
jaeone94: Valid point. The existing `missingModel` pipeline has the same pattern — `cancelLibrarySelect` only clears state without aborting in-flight imports. Adding per-upload AbortController tracking requires changes to the upload lifecycle and should be addressed for both pipelines in a follow-up PR.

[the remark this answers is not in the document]
jaeone94: yes and assign it to me please

AustinMroz:
See also #10418

This PR is more in line with the implementation I was hoping for.  It surfaces errors prior to execution, and side steps the hard edge cases (like a missing input which is not required for execution) by allowing the user to queue anyways.

jaeone94:
@AustinMroz Thanks for the pointer! I checked out #10418. Unfortunately I don't have access to the Cloud repo so I couldn't review the backend bug fix, but I did read the Notion report.

As you noted, my implementation is closer to a **pre-execution scan that warns** the user about missing inputs rather than blocking execution. #10418 appears to intend blocking execution when empty file inputs are detected. However, this approach was previously discussed and not adopted for the following reason:

The backend does not block execution as long as at least one valid node execution route exists. This means that to accurately detect blocking cases, we would need to do more than simply check whether any missing assets exist on the canvas — we would need to mirror the backend's workflow analysis algorithm on the frontend, which is a high implementation cost and duplicates engine logic. (Related discussion: https://comfy-organization.slack.com/archives/C0AED2B4MNY/p1771647819763719)

That said, the limitation of this approach is clear: users can always queue a prompt, and while this isn't a big deal in local environments, in cloud it means a prompt that is certain to fail could be queued and wait in line. However, mirroring the backend logic on the frontend / performing validation before queueing on the cloud backend are both non-trivial, and the current implementation seems like a reasonable minimum compromise for now.

On a separate note — @christian-byrne raised a concern about showing error states on nodes that aren't strictly required for execution (using the "boy who cried wolf" analogy). I agree with this concern and am considering a follow-up PR that suppresses missing model / missing media warnings for muted or bypassed nodes. I'll gather feedback on Slack before deciding on the approach.

viva-jinyi (commented):
Nice feature — mirrors the missing model pipeline cleanly. A few things I noticed:

---

### Naming / i18n

**`TabErrors.vue` — `handleLocateModel` reused for media**

`MissingMediaCard`'s `@locate-node` calls `handleLocateModel`, which is just `focusNode(nodeId)`. Renaming to `handleLocateNode` would be clearer since it's now shared across model and media contexts.

**`MissingMediaLibrarySelect.vue` — cross-namespace i18n key**

```typescript
{{ t('rightSidePanel.missingModels.or') }}
```

Borrows `or` from the `missingModels` namespace. Either add it to `missingMedia` or lift to a shared key so this doesn't break if `missingModels` changes.

---

### Simplification opportunities

**`useMissingMediaInteractions.ts` — nested ternary is just a template literal**

```typescript
const expectedPrefix =
  mediaType === 'audio'
    ? 'audio/'
    : mediaType === 'video'
      ? 'video/'
      : 'image/'
```

`mediaType` is already `'image' | 'video' | 'audio'`, so this entire conditional is just:

```typescript
const expectedPrefix = `${mediaType}/`
```

**`MissingMediaRow.vue` — unnecessary `computed` for static prop values**

`acceptType`, `extensionHint`, `displayName`, `isSingleNode` are all derived from `item.mediaType` or `item.name`, which never change for a given row instance (keyed by `item.name`). Per project guidelines ("if it's possible to use the ref or prop directly, don't add a computed"), these could be plain `const`:

```typescript
const acceptType = getAcceptType(item.mediaType)
const extensionHint = getExtensionHint(item.mediaType)
const displayName = getMediaDisplayName(item.name)
const isSingleNode = item.referencingNodes.length === 1
```

**`MissingMediaRow.vue` — `useFileDialog` / `useDropZone` from VueUse**

The hidden `<input ref="fileInputRef">` + manual `.click()` + `as HTMLInputElement` assertion + `input.value = ''` reset could be replaced with VueUse's `useFileDialog`:

```typescript
const { open, files, reset } = useFileDialog({ accept: acceptType })
```

Similarly, the manual `isDragOver` ref + `@dragover.prevent.stop` / `@dragleave.prevent.stop` / `@drop.prevent.stop` handlers could use `useDropZone`, which handles event propagation and provides `isOverDropZone` automatically.

**`missingMediaScan.ts` — `groupBy` from es-toolkit**

`groupCandidatesByMediaType` has a manual Map-based grouping loop:

```typescript
const typeMap = new Map<MediaType, MissingMediaCandidate[]>()
for (const c of candidates) {
  const list = typeMap.get(c.mediaType)
  if (list) list.push(c)
  else typeMap.set(c.mediaType, [c])
}
```

This is a classic `groupBy` pattern:

```typescript
import { groupBy } from 'es-toolkit'
const typeMap = groupBy(candidates, (c) => c.mediaType)
```

Same applies to the Set construction in `missingMediaNodeIds` — the loop could be:

```typescript
new Set(missingMediaCandidates.value?.map((m) => String(m.nodeId)))
```

---

### Minor

**`missingModelScan.ts` — mid-file import**

The `import { resolveComboValues }` replacing the old function definition appears to land at ~line 50, below other function definitions. Should be moved to the top with other imports.

**`MissingMediaRow.vue` — `ref` variable shadows Vue's `ref`**

```typescript
import { computed, ref } from 'vue'
// ...
const ref = item.referencingNodes[0]  // shadows Vue import
```

Commit m7 renamed a different `ref` loop variable but this one was missed. `nodeRef` or similar would fix it.

viva-jinyi (commented):
Nice feature — mirrors the missing model pipeline cleanly. Left a few inline suggestions.

src/components/rightSidePanel/errors/TabErrors.vue
```diff
@@ -170,6 +171,14 @@
             :show-node-id-badge="showNodeIdBadge"
             @locate-model="handleLocateModel"
           />
+
+          <!-- Missing Media -->
+          <MissingMediaCard
+            v-else-if="group.type === 'missing_media'"
+            :missing-media-groups="missingMediaGroups"
+            :show-node-id-badge="showNodeIdBadge"
+            @locate-node="handleLocateModel"
```
viva-jinyi: `handleLocateModel` is just `focusNode(nodeId)` — a generic function. Using it for media's `@locate-node` is confusing. Consider renaming to `handleLocateNode` since it's now shared across model and media.

src/platform/missingMedia/components/MissingMediaLibrarySelect.vue
```diff
@@ -0,0 +1,147 @@
+<template>
+  <div class="flex flex-col gap-2">
+    <div v-if="showDivider" class="flex items-center justify-center py-0.5">
+      <span class="text-xs font-bold text-muted-foreground">
+        {{ t('rightSidePanel.missingModels.or') }}
```
viva-jinyi: This borrows `or` from the `missingModels` namespace. If that namespace changes, this breaks silently. Either add a key under `missingMedia` or lift `or` to a shared namespace.

src/platform/missingMedia/composables/useMissingMediaInteractions.ts
```diff
@@ -0,0 +1,230 @@
...
+  }
+
+  /** Step 1: Store selection from library (does not apply yet). */
+  function handleLibrarySelect(name: string, value: string) {
+    store.pendingSelection[name] = value
+  }
+
+  /** Step 1: Upload file and store result as pending (does not apply yet). */
+  async function handleUpload(file: File, name: string, mediaType: MediaType) {
+    const expectedPrefix =
```
viva-jinyi: `mediaType` is already `'image' | 'video' | 'audio'`, so this nested ternary is equivalent to:

```typescript
const expectedPrefix = `${mediaType}/`
```

src/platform/missingMedia/components/MissingMediaRow.vue
```diff
@@ -0,0 +1,340 @@
...
+const displayName = computed(() => getMediaDisplayName(item.name))
+const isSingleNode = computed(() => item.referencingNodes.length === 1)
+const singleNodeLabel = computed(() => {
+  if (!isSingleNode.value) return ''
+  const ref = item.referencingNodes[0]
+  return getNodeDisplayLabel(String(ref.nodeId), item.name)
+})
+
+const expanded = computed(() => isExpanded(item.name))
+const acceptType = computed(() => getAcceptType(item.mediaType))
```
viva-jinyi: `acceptType`, `extensionHint`, `displayName`, `isSingleNode` are all derived from `item.mediaType` or `item.name` which are stable for the lifetime of this row (keyed by `item.name`). Per project guidelines ("if it's possible to use the ref or prop directly, don't add a computed"), these can be plain `const`:

```typescript
const acceptType = getAcceptType(item.mediaType)
const extensionHint = getExtensionHint(item.mediaType)
const displayName = getMediaDisplayName(item.name)
const isSingleNode = item.referencingNodes.length === 1
```

src/platform/missingMedia/components/MissingMediaRow.vue
```diff
@@ -0,0 +1,340 @@
...
+const currentUpload = computed(() => uploadState.value[item.name])
+const pendingDisplayName = computed(() => {
+  if (currentUpload.value) return currentUpload.value.fileName
+  const pending = pendingSelection.value[item.name]
+  return pending ? getMediaDisplayName(pending) : ''
+})
+
+const extensionHint = computed(() => getExtensionHint(item.mediaType))
+
+const isDragOver = ref(false)
```
viva-jinyi: The hidden `<input ref="fileInputRef">` + manual `.click()` + `as HTMLInputElement` assertion + `input.value = ''` reset pattern could be replaced with VueUse's [`useFileDialog`](https://vueuse.org/core/useFileDialog/).

Similarly, this manual `isDragOver` ref + `@dragover.prevent.stop` / `@dragleave.prevent.stop` / `@drop.prevent.stop` can be replaced with [`useDropZone`](https://vueuse.org/core/useDropZone/) which handles event propagation and provides `isOverDropZone` automatically.

src/platform/missingMedia/missingMediaScan.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+  }
+  return Array.from(map.values())
+}
+
+/** Group confirmed-missing candidates by media type. */
+export function groupCandidatesByMediaType(
+  candidates: MissingMediaCandidate[]
+): MissingMediaGroup[] {
+  const typeMap = new Map<MediaType, MissingMediaCandidate[]>()
```
viva-jinyi: Classic `groupBy` pattern — consider using `es-toolkit`:

```typescript
import { groupBy } from 'es-toolkit'

const typeMap = groupBy(candidates, (c) => c.mediaType)
```

src/platform/missingMedia/missingMediaStore.ts
```diff
@@ -0,0 +1,158 @@
...
+  const missingMediaCandidates = ref<MissingMediaCandidate[] | null>(null)
+
+  const hasMissingMedia = computed(() => !!missingMediaCandidates.value?.length)
+
+  const missingMediaCount = computed(
+    () => missingMediaCandidates.value?.length ?? 0
+  )
+
+  const missingMediaNodeIds = computed<Set<string>>(() => {
+    const ids = new Set<string>()
```
viva-jinyi: The manual loop can be a one-liner:

```typescript
const missingMediaNodeIds = computed(
  () => new Set(missingMediaCandidates.value?.map((m) => String(m.nodeId)) ?? [])
)
```

src/platform/missingModel/missingModelScan.ts
```diff
@@ -50,13 +50,7 @@ export function isModelFileName(name: string): boolean {
...
 }
 
-function resolveComboOptions(widget: IComboWidget): string[] {
-  const values = widget.options.values
-  if (!values) return []
-  if (typeof values === 'function') return values(widget)
-  if (Array.isArray(values)) return values
-  return Object.keys(values)
-}
+import { resolveComboValues } from '@/utils/litegraphUtil'
```
viva-jinyi: This `import` landed at line 53 — between function definitions. Should be moved to the top with the other imports (lines 1–25).

src/platform/missingMedia/components/MissingMediaRow.vue
```diff
@@ -0,0 +1,340 @@
...
+  confirmSelection,
+  cancelSelection,
+  hasPendingSelection
+} = useMissingMediaInteractions()
+
+const displayName = computed(() => getMediaDisplayName(item.name))
+const isSingleNode = computed(() => item.referencingNodes.length === 1)
+const singleNodeLabel = computed(() => {
+  if (!isSingleNode.value) return ''
+  const ref = item.referencingNodes[0]
```
viva-jinyi: `const ref` shadows the `ref` imported from `vue` on line 238. Commit m7 renamed a different loop variable but this one was missed — `nodeRef` or `firstNode` would fix it.

src/components/rightSidePanel/errors/TabErrors.vue
jaeone94: Fixed in e651a365f — renamed to `handleLocateAssetNode`, shared between model and media.

src/platform/missingMedia/components/MissingMediaLibrarySelect.vue
jaeone94: Fixed in e651a365f — moved to `rightSidePanel.missingMedia.or` with its own i18n key.

src/platform/missingMedia/composables/useMissingMediaInteractions.ts
jaeone94: Fixed in e651a365f — simplified to `` `${mediaType}/` ``.

src/platform/missingMedia/components/MissingMediaRow.vue
jaeone94: Fixed in e651a365f — converted `displayName`, `isSingleNode`, `singleNodeLabel`, `acceptType`, `extensionHint` to plain const.

src/platform/missingMedia/components/MissingMediaRow.vue
jaeone94: Fixed in e651a365f — replaced with `useDropZone` + `useFileDialog` from VueUse, matching the pattern in `DropZone.vue`.

src/platform/missingMedia/missingMediaScan.ts
jaeone94: Fixed in e651a365f — using `groupBy` from es-toolkit.

src/platform/missingMedia/missingMediaStore.ts
jaeone94: Fixed in e651a365f — simplified to one-liner `new Set(...map(...) ?? [])`.

src/platform/missingModel/missingModelScan.ts
jaeone94: Fixed in e651a365f — moved to the top import block.

src/platform/missingMedia/components/MissingMediaRow.vue
jaeone94: Fixed in e651a365f — `singleNodeLabel` is now a plain const that inlines the access, eliminating the shadow entirely.

src/scripts/app.ts
```diff
@@ -1565,6 +1573,44 @@ export class ComfyApp {
...
+          if (confirmed.length) {
+            useExecutionErrorStore().surfaceMissingMedia(confirmed)
+          }
+        })
+        .catch((err) => {
+          console.warn(
+            '[Missing Media Pipeline] Asset verification failed:',
+            err
+          )
+          useToastStore().add({
```
dante01yoon: This promise is not awaited, but `runMissingMediaPipeline` itself is await-ed at the call site (line 1422). The method returns before cloud verification finishes, which means loadGraphData proceeds to showPendingWarnings() and setDirty before missing media are surfaced. This is likely intentional (non-blocking), but:
  - The ESLint no-floating-promises rule should flag this — needs void prefix
  or an explicit comment explaining the intent.                                
  - Compare with the model pipeline which awaits its async verification. Is the
   inconsistency intentional?

src/platform/missingMedia/components/MissingMediaLibrarySelect.vue
```diff
@@ -0,0 +1,147 @@
...
+    ignoreLocation: true
+  },
+  matchAllWhenSearchEmpty: true
+})
+
+const filteredOptions = computed(() => fuseResults.value.map((r) => r.item))
+
+function getPreviewUrl(filename: string): string {
+  return api.apiURL(`/view?filename=${encodeURIComponent(filename)}&type=input`)
+}
```
dante01yoon, line 139: `encodeURIComponent` is correct for query params, but this URL is injected  directly into `` and `<video src>`. If api.apiURL returns a `javascript:` or `data:` scheme for a malicious input, this could be a vector. Verify that api.apiURL always returns an HTTP(S) URL, or add a scheme guard.

src/platform/missingMedia/missingMediaScan.ts
```diff
@@ -0,0 +1,159 @@
...
+  await store.updateInputs()
+
+  if (signal?.aborted) return
+
+  const assetHashes = new Set(
+    store.inputAssets.map((a) => a.asset_hash).filter((h): h is string => !!h)
+  )
+
+  for (const c of pending) {
+    c.isMissing = !assetHashes.has(c.name)
```
dante01yoon, line 120: verifyCloudMediaCandidates mutates the candidates array in place, and the  same array is later passed to surfaceMissingMedia → setMissingMedia. If any reactivity system holds a reference to these objects, in-place mutation may not trigger Vue reactivity. The model pipeline has the same pattern, so this is pre-existing, but worth noting.

src/components/rightSidePanel/errors/useErrorReport.ts
```diff
@@ -28,66 +28,73 @@ export function useErrorReport(cardSource: MaybeRefOrGetter<ErrorCardData>) {
...
+      })
 
-  onMounted(async () => {
-    const card = toValue(cardSource)
-    const runtimeErrors = card.errors
-      .map((error, idx) => ({ error, idx }))
-      .filter(({ error }) => error.isRuntimeError)
+      for (const key of Object.keys(enrichedDetails)) {
+        delete enrichedDetails[Number(key)]
+      }
```
dante01yoon, line 41: Issue: Using Number(key) on Object.keys() is fragile. If the reactive object ever has non-numeric keys, this silently skips them. Consider Object.keys(enrichedDetails).forEach(k => delete enrichedDetails[k as any])   or simply reassigning to a fresh reactive.

src/platform/missingMedia/components/MissingMediaRow.vue
```diff
@@ -0,0 +1,318 @@
...
+  confirmSelection,
+  cancelSelection,
+  hasPendingSelection
+} = useMissingMediaInteractions()
+
+const displayName = getMediaDisplayName(item.name)
+const isSingleNode = item.referencingNodes.length === 1
+const singleNodeLabel = isSingleNode
+  ? getNodeDisplayLabel(String(item.referencingNodes[0].nodeId), item.name)
+  : ''
```
dante01yoon, line 270: These are assigned once at setup time. If item ever changes (e.g., parent  re-keys the list), these won't update. The comment in a prior commit says  `"Replace unnecessary computeds with plain const for stable props"` — which is  fine if the parent always re-keys, but the `v-for="item in group.items" :key="item.name"` keying means Vue will reuse the component instance if a new item has the same name but different referencingNodes. Consider whether this is actually stable.

dante01yoon (commented):
- Clean separation of concerns: store, scan, interactions, types, components
- Consistent with the existing missing model pipeline pattern                
- Thorough unit tests covering edge cases (abort signals, interaction state cleanup, shared-name preservation)                                           
- Good a11y: aria-live, aria-label, aria-hidden on decorative elements

src/platform/missingMedia/components/MissingMediaLibrarySelect.vue
jaeone94: `api.apiURL` constructs URLs as `this.api_base + '/api' + route`, always producing HTTP(S) URLs. The `api_base` is set from the server origin at init and cannot be influenced by user-loaded workflow data. The combo widget values that flow into `getPreviewUrl` are also `encodeURIComponent`-encoded, preventing injection into the path structure. No scheme guard needed.

src/platform/missingMedia/missingMediaScan.ts
jaeone94: Agreed this is worth noting. The in-place mutation is a deliberate pattern shared with the model pipeline. Reactivity is not affected because setMissingMedia replaces the entire ref.value array, which triggers Vue reactivity. The mutation happens on plain objects before they enter the reactive system.

src/platform/missingMedia/components/MissingMediaRow.vue
jaeone94: Good observation, but this is safe with Vue 3.5 reactive props destructuring. The const { item } = defineProps<...>() creates a reactive proxy — accessing item.referencingNodes.length at template render time is tracked by Vue's reactivity system. If the parent passes a new item object (same key, different data), the template re-renders with the new values. The plain const assignment (const isSingleNode = item.referencingNodes.length === 1) is re-evaluated because item itself is reactive, not a snapshot.

src/scripts/app.ts
jaeone94: Fixed in d6a461c84 — added `void` prefix to the floating promise to satisfy `no-floating-promises` and make the non-blocking intent explicit.

src/components/rightSidePanel/errors/useErrorReport.ts
jaeone94: Fixed in d6a461c84 — replaced `Number(key)` with direct key deletion that doesn't depend on numeric coercion.

dante01yoon (approved):

src/platform/missingMedia/components/MissingMediaRow.vue
dante01yoon: @codex can you identify is it safe?

dante01yoon: question: 
@jaeone94 but Isn't it below const variable still not-reactive which not you intended? 
```
const isSingleNode = item.referencingNodes.length === 1
const singleNodeLabel = isSingleNode
  ? getNodeDisplayLabel(String(item.referencingNodes[0].nodeId), item.name)
  : ''```

PR #10311, opened 2026-03-19
feat: add WebSocket reconnect telemetry (incident-39)

mattmillerai:
## Summary
- Tracks WebSocket reconnect events via Sentry for incident-39 observability
- Adds breadcrumb on every reconnect, captures warning when jobs were in flight
- Extracted to `useWebSocketReconnectTracking` composable with tests

## Context
[Incident-39 Notion ticket](https://www.notion.so/comfy-org/Bug-WebSocket-connection-breaks-silently-client-stops-receiving-messages-3216d73d365081629b2acb8de37ba599) revealed we have no queryable frontend observability into WebSocket disconnects. See [#incident-39 Slack channel](https://comfy-organization.slack.com/archives/C0ANADDD648) for investigation discussion.

We can't answer "how often do users lose real-time updates while jobs are running?" This provides the data to quantify the problem before deciding on further UX changes (persistent banner, polling improvements, etc.).

Sentry was chosen over PostHog/Mixpanel because reconnect events are operational/debug signals, not product analytics. Backend logs show ~40-60 forwarding failures/day against ~350k daily WS connections — low volume, purely diagnostic.

### What changed
- **New**: `useWebSocketReconnectTracking` composable — tracks disconnect timing via `performance.now()`, reports to Sentry
- **New**: 6 tests covering the composable state machine
- **Removed**: All PostHog/Mixpanel telemetry plumbing (`WebSocketReconnectedMetadata`, `trackWebSocketReconnected` in types, registry, PostHog provider, Mixpanel provider, PostHog tests)
- **Updated**: `GraphView.vue` uses the composable instead of inline state

## Test plan
- [x] Verify typecheck/lint pass
- [x] Unit tests pass (6 cases: no-op without disconnect, breadcrumb on reconnect, warning with active jobs, no warning without jobs, duplicate disconnect guard, reset)
- [ ] Deploy to staging, simulate WebSocket disconnect with active job, verify Sentry event appears
- [ ] Verify breadcrumb-only (no warning) when disconnecting with no active jobs


src/views/GraphView.vue
```diff
@@ -253,14 +253,30 @@ const reconnectingMessage: ToastMessageOptions = {
...
 
 const onReconnected = () => {
+  if (disconnectedAt !== null) {
+    const activeJobCount = queueStore.activeJobsCount
+    if (activeJobCount > 0) {
+      telemetry?.trackWebSocketReconnected({
+        disconnect_duration_ms: Date.now() - disconnectedAt,
+        had_active_jobs: true,
+        active_job_count: activeJobCount
+      })
```
jaeone94: Question (non-blocking): `had_active_jobs` — is the `false` case intentionally untracked?

Since the event only fires inside the `activeJobCount > 0` guard, `had_active_jobs` is always `true` — reconnects with no active jobs produce no event at all. Is this intentional?

If so, `had_active_jobs` is redundant (always `true`) and could be removed to keep the schema honest. If not, should the event fire unconditionally so the `false` case is also tracked?

src/views/GraphView.vue
```diff
@@ -253,14 +253,30 @@ const reconnectingMessage: ToastMessageOptions = {
...
+  }
   if (!settingStore.get('Comfy.Toast.DisableReconnectingToast')) {
     toast.remove(reconnectingMessage)
     toast.add(reconnectingMessage)
   }
 }
 
 const onReconnected = () => {
+  if (disconnectedAt !== null) {
+    const activeJobCount = queueStore.activeJobsCount
```
jaeone94: Question (non-blocking): Is the inclusion of pending tasks in `activeJobsCount` intentional?
`src/stores/queueStore.ts:503-505`
```typescript
const activeJobsCount = computed(
  () => pendingTasks.value.length + runningTasks.value.length
)
```
`activeJobsCount` sums both `pendingTasks` (queued) and `runningTasks` (executing). CodeRabbit flagged this as potentially needing a running-only count, but on reflection there's a valid argument for including pending tasks: the `pending → running` transition (execution_start) is itself delivered via WebSocket. If the connection drops while tasks are queued, the user never sees them start — they remain stuck as "pending" in the UI even if the server has begun execution.

Is the intent here to capture all jobs whose status updates could be lost (pending + running), or only jobs that are actively producing progress/completion events (running only)?

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.ts
jaeone94, line 1: Issue: No tests added

The PR adds zero tests. `PostHogTelemetryProvider.test.ts` already has the exact harness pattern — adding one case for `trackWebSocketReconnected` would catch future regressions (e.g., accidental addition to `DEFAULT_DISABLED_EVENTS`).

src/views/GraphView.vue
```diff
@@ -253,14 +253,30 @@ const reconnectingMessage: ToastMessageOptions = {
   summary: t('g.reconnecting')
 }
 
+let disconnectedAt: number | null = null
```
jaeone94: Nit: Consider extracting to a composable

`src/views/GraphView.vue:256-288`

The new code introduces module-level mutable state (`let disconnectedAt`) and a stateful mini-protocol (record timestamp → compute duration → fire → reset). `GraphView.vue` is already ~490 lines. A composable like `useWebSocketReconnectTelemetry()` would isolate this and make it testable — not blocking, but worth considering if the file grows further.

src/views/GraphView.vue
```diff
@@ -253,14 +253,30 @@ const reconnectingMessage: ToastMessageOptions = {
...
+    const activeJobCount = queueStore.activeJobsCount
+    if (activeJobCount > 0) {
+      telemetry?.trackWebSocketReconnected({
+        disconnect_duration_ms: Date.now() - disconnectedAt,
+        had_active_jobs: true,
+        active_job_count: activeJobCount
+      })
+    }
+    disconnectedAt = null
+  }
```
jaeone94: Nit: `activeJobsCount` reflects pre-disconnect state

`reconnected` fires on WebSocket `open` (`api.ts:598-599`), before any status message arrives. So the count is the **pre-disconnect** value. This is likely the right thing (were jobs running when the user lost connectivity?), but a brief inline comment would clarify that intent for future readers.

jaeone94 (commented):
I left a few comments

src/views/GraphView.vue
```diff
@@ -253,14 +253,30 @@ const reconnectingMessage: ToastMessageOptions = {
...
   if (!settingStore.get('Comfy.Toast.DisableReconnectingToast')) {
     toast.remove(reconnectingMessage)
     toast.add(reconnectingMessage)
   }
 }
 
 const onReconnected = () => {
+  if (disconnectedAt !== null) {
+    const activeJobCount = queueStore.activeJobsCount
+    if (activeJobCount > 0) {
```
mattmillerai: Suggestion: reset `disconnectedAt` on unmount to prevent stale timestamps if the component remounts mid-disconnect.

```suggestion
onBeforeUnmount(() => {
  disconnectedAt = null
  executionStore.unbindExecutionEvents()
})
```

src/views/GraphView.vue
```diff
@@ -253,14 +253,30 @@ const reconnectingMessage: ToastMessageOptions = {
...
 }
 
+let disconnectedAt: number | null = null
+
 const onReconnecting = () => {
+  if (disconnectedAt === null) {
+    disconnectedAt = Date.now()
+  }
   if (!settingStore.get('Comfy.Toast.DisableReconnectingToast')) {
     toast.remove(reconnectingMessage)
```
mattmillerai, line 262: Worth a comment here -- `activeJobsCount` includes pending tasks intentionally since their `pending -> running` transition is also delivered via WebSocket. If the connection drops while tasks are queued, the user never sees them start.

```suggestion
  if (disconnectedAt !== null) {
    // Includes both pending and running tasks. Pending tasks matter because
    // their pending->running transition is delivered via WebSocket -- if the
    // connection drops while tasks are queued, the user never sees them start.
    const activeJobCount = queueStore.activeJobsCount
```

mattmillerai (commented):
**Must fix (looking into these now):**

1. **`had_active_jobs` is always `true`** -- The field is hardcoded to `true` inside the `activeJobCount > 0` guard, so it can never be `false`. Either fire the event unconditionally and let the field vary, or drop the field from the schema since it carries no signal.

2. **`activeJobsCount` captured at the wrong time** -- The count is read at reconnect time, but the queue store hasn't been refreshed yet at that point. Should capture at disconnect time (`onReconnecting`) to accurately reflect jobs in flight when the connection dropped.

3. **No tests** -- At minimum a PostHog provider test for `trackWebSocketReconnected` and unit tests for the reconnect state machine (the `null` guard, the timing, the job count threshold).

**Should fix:**

4. **`disconnectedAt` not reset on unmount** -- If GraphView unmounts mid-disconnect (HMR, route change), stale state produces inflated `disconnect_duration_ms`. One-liner: add `disconnectedAt = null` to `onBeforeUnmount`.

5. **`activeJobsCount` includes pending tasks** -- This is actually correct since pending->running transitions are delivered via WebSocket too, but worth a comment so nobody "fixes" it later. The PR description says "actively running" which is slightly misleading.

Jaeone's comments align with the above. I'll follow up with suggestions for #1-3.

mattmillerai:
> I left a few comments

Thank you! I just addressed these in my follow up commits and comments. Let me know if you have any questions on these updates!

jaeone94 (approved):
LGTM !

mattmillerai:
@jaeone94 looks like it won't let me merge since I need approval again since there were changes. Thanks!

jaeone94 (dismissed):

mattmillerai:
@jaeone94 It's not letting me merge, are you able to do that on your end now that it's approved?

jaeone94:
I've already approved, but since I'm not part of the required reviewer team, my approval doesn't satisfy the branch protection rule. @christian-byrne could you please take a look?

src/views/GraphView.vue
```diff
@@ -288,6 +307,8 @@ onMounted(() => {
 })
 
 onBeforeUnmount(() => {
+  disconnectedAt = null
+  activeJobCountAtDisconnect = 0
```
christian-byrne: What is the purpose of this? The variables are internal to the component.

src/platform/telemetry/providers/cloud/PostHogTelemetryProvider.test.ts
christian-byrne, line 1: As far as I can tell, these tests are just testing the mocks -- verifying that calling a method calls the method. Is there an error condition outside of the method being removed entirely? There's no branching logic, no transformation, no conditional behavior.

We could compare with something like the `onReconnecting`/`onReconnected` logic that tracks `disconnectedAt` state, computes duration, handles the null guard as that seems like bugs could actually hide. WDYT?

src/views/GraphView.vue
christian-byrne, line 1: We should make an Issue (or do in this PR if time permits) to refactor out to a `useWebsocketReconnect` as we are now holding stateful values and pushing a lot of stuff into the bloated view file (which will also not be usable by other views).

src/views/GraphView.vue
```diff
@@ -253,14 +253,33 @@ const reconnectingMessage: ToastMessageOptions = {
   summary: t('g.reconnecting')
 }
 
+let disconnectedAt: number | null = null
+let activeJobCountAtDisconnect = 0
+
 const onReconnecting = () => {
+  if (disconnectedAt === null) {
+    disconnectedAt = Date.now()
```
christian-byrne: nit: use `performance.now` to guard against clock drift.

christian-byrne (changes requested):
The PR description mentions "incident-39" and reconnect errors. If the goal is diagnosing operational incidents (connection drops during active jobs), that's Sentry territory -- structured error context with breadcrumbs, not product analytics events. PostHog/MP/GA is for "how do users behave," Sentry is for "what went wrong." A reconnect with `had_active_jobs: true` is an error condition, not a product insight.

mattmillerai (commented):
Valid point on Sentry vs product analytics. For debugging a single reconnect, Sentry with breadcrumbs is the right tool. But the goal here is understanding the aggregate user impact for incident-39.

We need to answer things like: what % of users hit reconnects during active jobs each week? Do those users end up refreshing or abandoning workflows? Is the rate going down after backend fixes ship? Does reconnect frequency line up with the Zendesk reports?

That said — after looking at the actual volume (~40-60 forwarding failures/day against ~350k daily WS connections), and the fact that we can correlate across Sentry/Datadog via MCP when needed, you're right that this is better as an operational signal than a product analytics event. Once incident-39 is resolved this becomes purely debug territory.

Moved everything to Sentry in the latest push. Also addressed the inline comments:
- Extracted to `useWebSocketReconnectTracking` composable
- `performance.now()` instead of `Date.now()`
- 6 tests covering the composable (disconnect/reconnect state machine, active jobs guard, reset, duplicate disconnect)
- Removed all PostHog/Mixpanel plumbing

mattmillerai:
Closing — Christian's feedback about using Sentry instead of PostHog for reconnect telemetry makes sense. This is better owned by the frontend team who can integrate it with their existing observability stack. Filed as a Linear ticket for frontend to pick up.

PR #10313, opened 2026-03-19
perf: eliminate double page navigation in Playwright test setup

DrJKL:
## Summary

Speed up the double `goto()` in `ComfyPage.setup()` by navigating to a cheap endpoint before handling localStorage etc..

## Review Focus

- Verify all Playwright E2E shards pass — the `clearStorage: false` and `mockReleases: false` paths should be unaffected.
- Note: `addInitScript` persists for the page lifetime (like `FeatureFlagHelper.seedFlags` already does), so any subsequent `page.reload()` or `goto()` in tests will also clear storage. This should be fine since tests that use `clearStorage: false` skip this block.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10313-perf-eliminate-double-page-navigation-in-Playwright-test-setup-3286d73d36508158a8edeefb27fcae20) by [Unito](https://www.unito.io)


benceruleanlu (approved):
Looks good to me.

One note: the current implementation no longer matches the PR description. The diff now uses a lightweight same-origin `/api/users` navigation before the final app load, rather than `addInitScript`. The code change makes sense; the description just looks stale.

PR #10331, opened 2026-03-20
PR preview - Add layout templates, arrange UI & zone drag/drop

koshimazaki:
## Template-based layout system for App Mode

Replaces the old split-panel builder architecture (AppBuilder in side panel + LinearControls) with a unified grid-based layout system. Users can now arrange inputs, outputs, and run controls into customizable zone layouts directly in the center panel.

### What changed

**Layout system (11 new files)**
- Three layout templates: Focus (2-column), Grid (2x3), Sidebar (4+sidebar)
- Draggable zone grid with resizable columns/rows via pointer handles
- Drag-and-drop widgets, outputs, and run controls between zones
- Within-zone item reordering
- Per-zone alignment toggle (top/bottom)
- Per-template persistence of all layout settings (zone assignments, grid sizes, item order, alignment)
- Mobile support: zones stack vertically with swipe navigation

**Presets system**
- Save/overwrite preset configurations per workflow
- Preset strip with tabs, buttons, or menu display modes
- Drag preset strip between zones in the builder
- Enable/disable presets via eye icon toggle

**Output handling**
- Outputs render inside grid zones instead of taking over full screen
- Click history thumbnails to show output in zone
- Double-click outputs or history thumbnails to open ImageLightbox fullscreen
- Black letterbox background for output zones (no grey bleed-through)
- Per-zone output node placeholders before results arrive

**Drop indicators (LoadImage/LoadVideo)**
- Image preview with zoom lightbox overlay
- Video preview with native controls
- Mask editor button below image preview
- Drag-and-drop file upload onto input zones

**Simplified LinearView**
- Builder (ArrangeLayout), app view (AppTemplateView), and fallback (LinearPreview) all render in the center panel
- Removed AppBuilder/LinearControls from side panels — side panels are now sidebar-only

**Store changes (appModeStore)**
- Per-template storage for zone assignments, grid overrides, run controls position, item order, and alignment
- Template switching with auto-redistribution of inputs/outputs
- Default zone fallback: unassigned widgets go to first non-output zone (fixes empty app mode when no builder layout saved)
- Item reordering (inputs, outputs) with persistence

**Bug fixes**
- Widget deduplication: multiple widgets from same node no longer render N times
- Float +/- buttons: stepValue infers from range instead of returning 0
- Sticky float dragging: sensitivity scales with step size
- Splice index fix in zone reordering
- Pointercancel handler on resize handles for touch robustness
- Subgraph widget key matching in AppModeWidgetList
- Duplicate i18n key removed

**Quality & UX**
- i18n for all new UI strings (layout templates, zones, presets, tooltips)
- Toast alert on queue failure
- Shift+click tooltip on Run button
- Zone overflow clipping with per-zone scrolling
- Resize handle clamping (MIN_FR=0.25 prevents collapsing zones)
- Accessibility: aria-labels on zone grid and interactive elements

### Files

**New files (16):** ArrangeLayout, AppTemplateView, LayoutZoneGrid, LayoutTemplateSelector, ZoneResizeHandle, layoutTemplates, useZoneDrop, useZoneReorder, useWidgetReorder, useZoneWidgets (+test), PresetStrip, PresetMenu (+stories), dropIndicatorUtil, Tooltip (+stories), useAppPresets (+test)

**Modified files:** appModeStore (+test), LinearView, LinearPreview, LinearControls, GraphView, MobileDisplay, AppModeWidgetList, LayoutZoneGrid, OutputHistory, DropZone, ScrubableNumberInput, WidgetInputNumberInput, comfyWorkflow, litegraphUtil, main.json, appModeDropdownClipping.spec.ts, appModeWidgetRename.spec.ts

### Follow-up items

- `appModeStore.ts` (~520 lines) is a candidate for splitting zone management from builder lifecycle
- PrimeVue Splitter in LinearView flagged for future migration to Reka UI
- Output placeholder animation during execution


christian-byrne (commented):
Lint preflight finding (non-blocking, for the author):

src/components/builder/useInputGroups.test.ts
```diff
@@ -0,0 +1,201 @@
...
+}))
+
+import {
+  autoGroupName,
+  groupedByPair,
+  resolveGroupItems
+} from './useInputGroups'
+
+beforeEach(() => {
+  vi.clearAllMocks()
```
christian-byrne, line 25: The `beforeEach` hook at line 25 calls `vi.clearAllMocks()`, which the current `main` lint config flags as `comfy/no-redundant-vitest-cleanup`. Vitest resets mock state between tests by default, so the explicit reset is unnecessary and will fail the lint preflight when this PR rebases onto a recent `main`. Would be cleaner to remove the `beforeEach` block entirely, or drop the `vi.clearAllMocks()` call if other setup is added later.

christian-byrne (commented):
Lint preflight finding (non-blocking). See inline comment on `useInputGroups.test.ts`.

src/components/builder/useInputGroups.test.ts
christian-byrne, line 25: The `beforeEach` hook at line 25 calls `vi.clearAllMocks()`, which the current `main` lint config flags as `comfy/no-redundant-vitest-cleanup`. Vitest resets mock state between tests by default, so the explicit reset is unnecessary and will fail the lint preflight when this PR rebases onto a recent `main`. Would be cleaner to remove the `beforeEach` block entirely, or drop the `vi.clearAllMocks()` call if other setup is added later.

christian-byrne, line 25: The `beforeEach` hook at line 25 calls `vi.clearAllMocks()`, which the current `main` lint config flags as `comfy/no-redundant-vitest-cleanup`. Vitest resets mock state between tests by default, so the explicit reset is unnecessary and will fail the lint preflight when this PR rebases onto a recent `main`. Would be cleaner to remove the `beforeEach` block entirely, or drop the `vi.clearAllMocks()` call if other setup is added later.

PR #10342, opened 2026-03-20
Don't use reactives for app mode selections

AustinMroz:
- The DraggableList component takes a v-model.
- When a drag is completed, it reassigns to v-model so an update event fires
  - App Builder sets v-model="appModeStore.selectedInputs"
- Thus a completed drag operation effectively performs appModeStore.selectedInputs = newList
  - In appModeStore, selectedInputs is a reactive. Thus, after a drag/drop operation occurs, selectedInputs in the store is not the same value as appModeStore.selectedInputs

When a reliable repro for the issue had not yet been found, an attempted earlier fix for this was to swap from watchers to directly updating `selectedInputs`. Since this change makes the code cleaner and still make the code safer, the change is left in place. 

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10342-Don-t-use-watcher-for-loading-app-mode-selections-3296d73d365081f7b216e79c4f4f4e8d) by [Unito](https://www.unito.io)

### DIFF
--- a/src/stores/appModeStore.ts
+++ b/src/stores/appModeStore.ts
@@ -1,5 +1,5 @@
 import { defineStore } from 'pinia'
-import { reactive, computed, watch } from 'vue'
+import { ref, computed, watch } from 'vue'
 import { useEventListener } from '@vueuse/core'
 
 import { useEmptyWorkflowDialog } from '@/components/builder/useEmptyWorkflowDialog'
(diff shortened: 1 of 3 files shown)


src/stores/appModeStore.ts
```diff
@@ -54,8 +54,8 @@ export const useAppModeStore = defineStore('appMode', () => {
 
   function loadSelections(data: Partial<LinearData> | undefined) {
     const { inputs, outputs } = pruneLinearData(data)
-    selectedInputs.splice(0, selectedInputs.length, ...inputs)
-    selectedOutputs.splice(0, selectedOutputs.length, ...outputs)
+    selectedInputs.value.splice(0, selectedInputs.value.length, ...inputs)
```
pythongosssss: Now it's a ref, should this just be 
```
selectedInputs.value = inputs                                                                                                                                                          
selectedOutputs.value = outputs 
```

AustinMroz: Rabbit brought this up as well.

I know there were specific reactivity bugs that caused me to swap from `ref` to `reactive` earlier in development. I'll make the swap and do some actual testing to make sure reactivity still functions.

AustinMroz: Yep. Change is safe and does not break anything.

pythongosssss (approved):

PR #10343, opened 2026-03-20
fix: hide API key login option on Cloud

mattmillerai:
## Summary
- Hide the "Comfy API Key" login button and help text from the sign-in modal when running on Cloud
- API key auth is only used on local ComfyUI; Cloud has Firebase-whitelisted Google/GitHub auth
- The button was appearing on Cloud as a relic of shared code between Cloud and local ComfyUI

## Context
[Discussion with Robin in #proj-cloud-frontend](https://comfy-organization.slack.com/archives/C09FY39CC3V/p1773977756997559) — API key auth only works on local because Firebase requires whitelisted domains. On Cloud, SSO (Google/GitHub) works natively, so the API key option is unnecessary and confusing.

Screenshot 2026-03-20 at 9 53 20 AM

## Test plan
- [ ] Verify login modal on local ComfyUI still shows the "Comfy API Key" button
- [ ] Verify login modal on cloud.comfy.org no longer shows the "Comfy API Key" button

### DIFF
--- a/src/components/dialog/content/SignInContent.vue
+++ b/src/components/dialog/content/SignInContent.vue
@@ -77,29 +77,31 @@
           </Button>
         </template>
 
-        <Button
-          type="button"
-          class="h-10"
-          variant="secondary"
-          @click="showApiKeyForm = true"
-        >
-          <img
-            src="/assets/images/comfy-logo-mono.svg"
-            class="mr-2 size-5"
-            :alt="$t('g.comfy')"
-          />
-          {{ t('auth.login.useApiKey') }}
-        </Button>
-        <small class="text-center text-muted">
-          {{ t('auth.apiKey.helpText') }}
-          <a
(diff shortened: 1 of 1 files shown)


AustinMroz (commented):
Actual change is good. I'm assuming the inclusion of the websocket reconnection changes is a mistake.

mattmillerai:
> Actual change is good. I'm assuming the inclusion of the websocket reconnection changes is a mistake.

Hmm weird, that's from another PR I'm working on...not sure how it got in there. Let me remove, one sec.

mattmillerai:
> Actual change is good. I'm assuming the inclusion of the websocket reconnection changes is a mistake.

Done. I accidentally branched this from my other feature branch instead of main. Just rebased and force pushed the changes so it should just be that one minor change now. Thanks!

AustinMroz (approved):

PR #10349, opened 2026-03-20
Feat/glsl live preview

jtydhr88:
## Summary
replacement for https://github.com/Comfy-Org/ComfyUI_frontend/pull/9201

the first commit squashed https://github.com/Comfy-Org/ComfyUI_frontend/pull/9201 and fixed conflict.

the second commit change needed by:
 - Enable GLSL live preview on SubgraphNodes by detecting the inner GLSLShader and rendering its preview directly on the parent SubgraphNode                                                  
- Previously, SubgraphNodes containing a GLSLShader showed no live preview at all To achieve this:                                                                                                                                                                           
   - Read shader source, uniform values, and renderer config from the inner GLSLShader's widgets                                                                                              
   - Trace IMAGE inputs through the subgraph boundary so the inner shader can use images connected to the SubgraphNode's outer inputs                                                         
   - Set preview output using the inner node's locator ID so the promoted preview system picks it up on the SubgraphNode                                                                      
   - Extract setNodePreviewsByLocatorId from nodeOutputStore to support setting previews by locator ID directly                                                                               
- Fix graphId to use rootGraph.id for widget store lookups (was using graph.id, which broke lookups for nodes inside subgraphs)                                                              
- Read uniform values from connected upstream nodes, not just local widgets                                                                                                                  
- Fix blob URL lifecycle: use the store's createSharedObjectUrl/releaseSharedObjectUrl reference-counting system instead of manual revoke, preventing leaks on composable re-creation        
        

## Screenshot

https://github.com/user-attachments/assets/9623fa32-de39-4a3a-b8b3-28688851390b

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10349-Feat-glsl-live-preview-3296d73d3650814b83aef52ab1962a77) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/curve/curveUtils.ts
+++ b/src/components/curve/curveUtils.ts
@@ -192,3 +192,15 @@ export function curvesToLUT(
 
   return lut
 }
+
+export function curveDataToFloatLUT(
+  curve: CurveData,
+  size: number = 256
+): Float32Array {
+  const lut = new Float32Array(size)
+  const interpolate = createInterpolator(curve.points, curve.interpolation)
+  for (let i = 0; i < size; i++) {
+    lut[i] = interpolate(i / (size - 1))
+  }
+  return lut
+}
--- a/src/renderer/extensions/vueNodes/components/LGraphNode.vue
+++ b/src/renderer/extensions/vueNodes/components/LGraphNode.vue
@@ -284,6 +284,7 @@ import { useTelemetry } from '@/platform/telemetry'
 import { useCanvasStore } from '@/renderer/core/canvas/canvasStore'
 import { useCanvasInteractions } from '@/renderer/core/canvas/useCanvasInteractions'
 import { layoutStore } from '@/renderer/core/layout/store/layoutStore'
+import { useGLSLPreview } from '@/renderer/glsl/useGLSLPreview'
 import { usePromotedPreviews } from '@/composables/node/usePromotedPreviews'
 import NodeBadges from '@/renderer/extensions/vueNodes/components/NodeBadges.vue'
 import { LayoutSource } from '@/renderer/core/layout/types'
@@ -729,6 +730,8 @@ const lgraphNode = computed(() => {
 // TODO: Surface subgraph info more cleanly in VueNodeData instead of
 // reaching through lgraphNode for promoted preview resolution.
 const { promotedPreviews } = usePromotedPreviews(lgraphNode)
+
+useGLSLPreview(lgraphNode)
 
 const showAdvancedInputsButton = computed(() => {
   const node = lgraphNode.value
--- a/src/renderer/glsl/useGLSLPreview.test.ts
+++ b/src/renderer/glsl/useGLSLPreview.test.ts
@@ -227,4 +227,105 @@ describe('useGLSLPreview', () => {
       })
     })
   })
+
+  describe('render pipeline', () => {
+    beforeEach(() => {
+      vi.useFakeTimers()
+    })
+
+    afterEach(() => {
+      vi.useRealTimers()
+    })
+
+    async function setupAndRender(node: LGraphNode) {
+      mockNodeOutputs[String(node.id)] = {
+        images: [{ filename: 'test.png', subfolder: '', type: 'temp' }]
+      }
+      const store = useWidgetValueStore() as unknown as {
+        _widgetMap: Map<string, { value: unknown }>
+      }
+      store._widgetMap.set('fragment_shader', {
+        value: 'void main() {}'
+      })
+
+      const nodeRef = shallowRef<LGraphNode | null>(null)
+      const result = useGLSLPreview(nodeRef)
+
+      nodeRef.value = node
+      await nextTick()
+      vi.advanceTimersByTime(100)
+      await nextTick()
+      // Allow async renderPreview to complete
+      await nextTick()
+
+      return result
+    }
+
+    it('calls compileFragment, render, and toBlob in sequence', async () => {
+      const node = createMockNode()
+      await setupAndRender(node)
+
+      expect(mockRendererFactory.compileFragment).toHaveBeenCalledWith(
+        'void main() {}'
+      )
+      expect(mockRendererFactory.render).toHaveBeenCalled()
+      expect(mockRendererFactory.toBlob).toHaveBeenCalled()
+
+      const compileOrder =
+        mockRendererFactory.compileFragment.mock.invocationCallOrder[0]
+      const renderOrder = mockRendererFactory.render.mock.invocationCallOrder[0]
+      const toBlobOrder = mockRendererFactory.toBlob.mock.invocationCallOrder[0]
+      expect(compileOrder).toBeLessThan(renderOrder)
+      expect(renderOrder).toBeLessThan(toBlobOrder)
+    })
+
+    it('sets lastError on compilation failure', async () => {
+      mockRendererFactory.compileFragment.mockReturnValueOnce({
+        success: false,
+        log: 'syntax error at line 5'
+      })
+
+      const node = createMockNode()
+      const { lastError } = await setupAndRender(node)
+
+      expect(lastError.value).toBe('syntax error at line 5')
+    })
+
+    it('clears lastError on successful compilation', async () => {
+      const node = createMockNode()
+      const { lastError } = await setupAndRender(node)
+
+      expect(lastError.value).toBe(null)
+    })
+
+    it('skips render when shader source is unavailable', async () => {
+      const store = useWidgetValueStore() as unknown as {
+        _widgetMap: Map<string, { value: unknown }>
+      }
+      store._widgetMap.delete('fragment_shader')
+
+      const node = createMockNode()
+      mockNodeOutputs[String(node.id)] = {
+        images: [{ filename: 'test.png', subfolder: '', type: 'temp' }]
+      }
+
+      const nodeRef = shallowRef<LGraphNode | null>(null)
+      useGLSLPreview(nodeRef)
+      nodeRef.value = node
+      await nextTick()
+      vi.advanceTimersByTime(100)
+      await nextTick()
+
+      expect(mockRendererFactory.compileFragment).not.toHaveBeenCalled()
+    })
+
+    it('disposes renderer and cancels debounce on cleanup', async () => {
+      const node = createMockNode()
+      const { dispose } = await setupAndRender(node)
+
+      dispose()
+
+      expect(mockRendererFactory.dispose).toHaveBeenCalled()
+    })
+  })
 })
--- a/src/renderer/glsl/useGLSLPreview.test.ts
+++ b/src/renderer/glsl/useGLSLPreview.test.ts
@@ -1,0 +1,230 @@
+import { createPinia, setActivePinia } from 'pinia'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+import { nextTick, reactive, ref, shallowRef } from 'vue'
+
+import { useGLSLPreview } from '@/renderer/glsl/useGLSLPreview'
+import { useWidgetValueStore } from '@/stores/widgetValueStore'
+
+import type { GLSLRendererConfig } from '@/renderer/glsl/useGLSLRenderer'
+import type { LGraphNode } from '@/lib/litegraph/src/LGraphNode'
+import type { MaybeRefOrGetter } from 'vue'
+
+const mockRendererFactory = vi.hoisted(() => {
+  const init = vi.fn(() => true)
+  const compileFragment = vi.fn(() => ({ success: true, log: '' }))
+  const setResolution = vi.fn()
+  const setFloatUniform = vi.fn()
+  const setIntUniform = vi.fn()
+  const setBoolUniform = vi.fn()
+  const bindCurveTexture = vi.fn()
+  const bindInputImage = vi.fn()
+  const render = vi.fn()
+  const toBlob = vi.fn(() => Promise.resolve(new Blob(['test'])))
+  const dispose = vi.fn()
+  const lastConfig = { value: undefined as GLSLRendererConfig | undefined }
+
+  return {
+    create: (config?: GLSLRendererConfig) => {
+      lastConfig.value = config
+      return {
+        init,
+        compileFragment,
+        setResolution,
+        setFloatUniform,
+        setIntUniform,
+        setBoolUniform,
+        bindCurveTexture,
+        bindInputImage,
+        render,
+        toBlob,
+        dispose
+      }
+    },
+    lastConfig,
+    init,
+    compileFragment,
+    setResolution,
+    setFloatUniform,
+    setIntUniform,
+    setBoolUniform,
+    bindCurveTexture,
+    bindInputImage,
+    render,
+    toBlob,
+    dispose
+  }
+})
+
+vi.mock('@/renderer/glsl/useGLSLRenderer', () => ({
+  useGLSLRenderer: (config?: GLSLRendererConfig) =>
+    mockRendererFactory.create(config)
+}))
+
+const mockSetNodePreviewsByNodeId = vi.fn()
+const mockNodeOutputs = reactive<Record<string, unknown>>({})
+
+vi.mock('@/stores/nodeOutputStore', () => ({
+  useNodeOutputStore: () => ({
+    setNodePreviewsByNodeId: mockSetNodePreviewsByNodeId,
+    setNodePreviewsByLocatorId: vi.fn(),
+    revokePreviewsByLocatorId: vi.fn(),
+    nodeOutputs: mockNodeOutputs
+  })
+}))
+
+vi.mock('@/stores/widgetValueStore', () => {
+  const widgetMap = new Map<string, { value: unknown }>()
+  const getWidget = vi.fn((_graphId: string, _nodeId: string, name: string) =>
+    widgetMap.get(name)
+  )
+  return {
+    useWidgetValueStore: () => ({
+      getWidget,
+      _widgetMap: widgetMap
+    })
+  }
+})
+
+vi.mock('@/platform/workflow/management/stores/workflowStore', () => ({
+  useWorkflowStore: () => ({
+    nodeIdToNodeLocatorId: (id: string | number) => String(id),
+    nodeToNodeLocatorId: (node: { id: string | number }) => String(node.id)
+  })
+}))
+
+vi.mock('@/utils/objectUrlUtil', () => ({
+  createSharedObjectUrl: () => 'blob:test',
+  releaseSharedObjectUrl: vi.fn()
+}))
+
+function createMockNode(overrides: Record<string, unknown> = {}): LGraphNode {
+  const graph = { id: 'test-graph-id', rootGraph: { id: 'test-graph-id' } }
+  return {
+    id: 1,
+    type: 'GLSLShader',
+    inputs: [],
+    graph,
+    getInputNode: vi.fn(() => null),
+    isSubgraphNode: () => false,
+    ...overrides
+  } as unknown as LGraphNode
+}
+
+function wrapNode(
+  node: LGraphNode | null
+): MaybeRefOrGetter<LGraphNode | null> {
+  return ref(node) as MaybeRefOrGetter<LGraphNode | null>
+}
+
+describe('useGLSLPreview', () => {
+  beforeEach(() => {
+    setActivePinia(createPinia())
+    vi.clearAllMocks()
+    mockRendererFactory.lastConfig.value = undefined
+    globalThis.URL.createObjectURL = vi.fn(() => 'blob:test')
+    globalThis.URL.revokeObjectURL = vi.fn()
+  })
+
+  it('does not activate for non-GLSLShader nodes', () => {
+    const node = createMockNode({ type: 'KSampler' })
+    const { isActive } = useGLSLPreview(wrapNode(node))
+    expect(isActive.value).toBe(false)
+  })
+
+  it('does not activate before first execution', () => {
+    const node = createMockNode()
+    Object.keys(mockNodeOutputs).forEach((k) => delete mockNodeOutputs[k])
+    const { isActive } = useGLSLPreview(wrapNode(node))
+    expect(isActive.value).toBe(false)
+  })
+
+  it('activates for GLSLShader nodes with execution output', () => {
+    const node = createMockNode()
+    mockNodeOutputs['1'] = {
+      images: [{ filename: 'test.png', subfolder: '', type: 'temp' }]
+    }
+    const { isActive } = useGLSLPreview(wrapNode(node))
+    expect(isActive.value).toBe(true)
+  })
+
+  it('exposes lastError as null initially', () => {
+    const node = createMockNode()
+    const { lastError } = useGLSLPreview(wrapNode(node))
+    expect(lastError.value).toBe(null)
+  })
+
+  it('does not activate for null node', () => {
+    const { isActive } = useGLSLPreview(wrapNode(null))
+    expect(isActive.value).toBe(false)
+  })
+
+  it('cleans up on dispose', () => {
+    const node = createMockNode()
+    const { dispose } = useGLSLPreview(wrapNode(node))
+    expect(() => dispose()).not.toThrow()
+  })
+
+  describe('autogrow config extraction', () => {
+    beforeEach(() => {
+      vi.useFakeTimers()
+    })
+
+    afterEach(() => {
+      vi.useRealTimers()
+    })
+
+    async function triggerRender(node: LGraphNode) {
+      mockNodeOutputs[String(node.id)] = {
+        images: [{ filename: 'test.png', subfolder: '', type: 'temp' }]
+      }
+      const store = useWidgetValueStore() as unknown as {
+        _widgetMap: Map<string, { value: unknown }>
+      }
+      store._widgetMap.set('fragment_shader', {
+        value: 'void main() {}'
+      })
+
+      const nodeRef = shallowRef<LGraphNode | null>(null)
+      useGLSLPreview(nodeRef)
+
+      nodeRef.value = node
+      await nextTick()
+      vi.advanceTimersByTime(100)
+      await nextTick()
+    }
+
+    it('passes default config when node has no comfyDynamic', async () => {
+      const node = createMockNode()
+      await triggerRender(node)
+
+      expect(mockRendererFactory.lastConfig.value).toEqual({
+        maxInputs: 5,
+        maxFloatUniforms: 20,
+        maxIntUniforms: 20,
+        maxBoolUniforms: 10,
+        maxCurves: 4
+      })
+    })
+
+    it('extracts autogrow limits from node comfyDynamic', async () => {
+      const node = createMockNode({
+        comfyDynamic: {
+          autogrow: {
+            images: { min: 1, max: 3 },
+            floats: { min: 0, max: 8 },
+            ints: { min: 0, max: 4 }
+          }
+        }
+      })
+      await triggerRender(node)
+
+      expect(mockRendererFactory.lastConfig.value).toEqual({
+        maxInputs: 3,
+        maxFloatUniforms: 8,
+        maxIntUniforms: 4,
+        maxBoolUniforms: 10,
+        maxCurves: 4
+      })
+    })
+  })
+})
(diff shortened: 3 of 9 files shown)


src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,518 @@
...
+      nodeId: sourceNode.id as NodeId,
+      widgetName: sourceNode.widgets[0].name
+    }
+
+    if (prefix === 'floats') floats.push(source)
+    else if (prefix === 'ints') ints.push(source)
+  }
+
+  return { floats, ints }
+}
```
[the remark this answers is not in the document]
jtydhr88: Valid observation. Uniforms that cross the subgraph boundary (connected to SubgraphInput → parent graph) are intentionally skipped in this PR. 
The current scope handles the most common pattern: internal PrimitiveFloat/Int nodes directly wired to the GLSLShader inside the subgraph.
Supporting boundary-crossing uniforms requires tracing through the SubgraphNode's outer inputs (similar to how getImageThroughSubgraphBoundary handles IMAGE inputs), plus handling mixed sources where some uniforms are internal and some cross the boundary. Will address in a follow-up.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,518 @@
...
+      const owner = ownerSubgraphNode.value
+      if (owner) {
+        const img = getImageThroughSubgraphBoundary(node, slot, owner)
+        if (img) {
+          renderer.bindInputImage(imageSlotIndex, img)
+        }
+      }
+      imageSlotIndex++
+    }
+  }
```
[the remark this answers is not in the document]
jtydhr88: The slot ordering follows the subgraph boundary creation order, which mirrors the inner wiring in normal usage. Internal image nodes and size_mode widget fallback for resolution are known gaps — same category as the boundary-crossing uniform limitation noted above, will address together in a follow-up.

src/renderer/glsl/useGLSLPreview.ts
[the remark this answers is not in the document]
jtydhr88: yes, open a follow up github issue and assign to me

src/renderer/glsl/useGLSLPreview.ts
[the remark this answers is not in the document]
jtydhr88: yes, open a follow up github issue and assign to me

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,518 @@
...
+    if (!renderer) {
+      renderer = useGLSLRenderer(rendererConfig.value)
+    }
+    return renderer
+  }
+
+  async function renderPreview(): Promise<void> {
+    const requestId = ++renderRequestId
+    const source = shaderSource.value
+    if (!source || !isActive.value) return
```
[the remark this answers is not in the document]
jtydhr88: hideOutputImages only applies to direct GLSLShader nodes (guarded by isGLSLNode check), and isActive requires a prior successful execution. On failure paths, lastError is set so the caller can react. In practice, if WebGL init fails or the shader doesn't compile, the user sees the error state rather than a blank preview. Agreed this could be tightened by restoring hideOutputImages = false on persistent failures — will note for follow-up.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,518 @@
...
+        r.setFloatUniform(i, floatValues.value[i])
+      }
+      for (let i = 0; i < intValues.value.length; i++) {
+        r.setIntUniform(i, intValues.value[i])
+      }
+
+      r.render()
+
+      const blob = await r.toBlob()
+      if (requestId !== renderRequestId) return
```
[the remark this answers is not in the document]
jtydhr88: Good catch. Added renderRequestId++ in dispose() so the existing staleness check (requestId !== renderRequestId) catches in-flight toBlob() callbacks after teardown. The other suggested cleanup (renderer = null, hideOutputImages restore) is not necessary — after scope disposal these variables are unreachable, and renderer.dispose() already releases WebGL resources.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,518 @@
...
+interface AutogrowGroup {
+  max: number
+  min: number
+  prefix?: string
+}
+
+interface UniformSource {
+  nodeId: NodeId
+  widgetName: string
+}
```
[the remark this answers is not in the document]
jtydhr88: The sparse wiring scenario doesn't arise in practice — the autogrow system creates sequential inputs (u_float0, u_float1, ...) and disconnecting earlier inputs triggers contraction of later ones.
The direct-node "precedence bug" doesn't exist either: when an autogrow input is connected to an upstream node, the widget is removed from the store, so getWidget() returns undefined and the code falls through to the upstream check. The two paths are mutually exclusive by design.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,518 @@
...
+    () => [floatValues.value, intValues.value] as const,
+    () => {
+      if (isActive.value) debouncedRender()
+    },
+    { deep: true }
+  )
+
+  watch(shaderSource, () => {
+    if (isActive.value) debouncedRender()
+  })
```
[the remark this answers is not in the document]
jtydhr88: Upstream node.imgs is a plain property on LGraphNode, not Vue reactive state — it can't be watched. This is not a regression from this PR; the direct GLSLShader mode in the parent commit has the same behavior. The live preview is designed for interactive feedback between executions (shader edits, uniform tweaks). After a full execution, the backend produces the authoritative output. Making input image changes trigger re-renders would require a reactivity bridge between LGraphNode and Vue, which is out of scope here.

src/stores/nodeOutputStore.ts
```diff
@@ -308,6 +303,20 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
+   *
+   * @param nodeId - The node ID
+   * @param previewImages - Array of preview image URLs to store
+   */
+  function setNodePreviewsByNodeId(
+    nodeId: string | number,
+    previewImages: string[]
+  ) {
+    setNodePreviewsByLocatorId(nodeIdToNodeLocatorId(nodeId), previewImages)
+  }
```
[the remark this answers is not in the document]
jtydhr88: nodeIdToNodeLocatorId always returns a non-nullable NodeLocatorId (typed as string) — it either returns String(nodeId) for root graph nodes or createNodeLocatorId(subgraphId, nodeId) for subgraph nodes. The guard at L389 exists for a different reason (defensive against Number() coercion edge cases). No change needed here

src/renderer/glsl/useGLSLPreview.ts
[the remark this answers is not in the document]
jtydhr88: yes, open a github issue and assign to me

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -728,6 +729,8 @@ const lgraphNode = computed(() => {
 // reaching through lgraphNode for promoted preview resolution.
 const { promotedPreviews } = usePromotedPreviews(lgraphNode)
 
+useGLSLPreview(lgraphNode)
```
christian-byrne, line 734: issue: `useGLSLPreview()` has been called unconditionally for every `LGraphNode` since the first commit of this PR -- it's not a regression, but it means every node in the graph pays the cost of ~14 computed properties, 3 watchers, store subscriptions, and a debounce allocation even when the node has nothing to do with GLSL. In a 500-node graph with 1 GLSL node, this creates thousands of unused reactive objects. The `hasExecutionOutput` computed also reads `nodeOutputStore.nodeOutputs` (the full reactive record), so any output change on any node triggers re-evaluation across all nodes.

The fix needs to handle both direct `GLSLShader` nodes and subgraph nodes that contain one. A clean approach: restructure the composable to split into two tiers. The outer tier (always created) does only the cheap checks -- `isGLSLNode` and `isGLSLSubgraphNode`. The inner tier (watchers, store subscriptions, debounce, renderer) is created lazily inside a `watchEffect` that fires only when one of those flags becomes true. This avoids the conditional-composable problem (Vue composables can't be behind `if` at the call site) while keeping the per-node overhead to just 2-3 lightweight computeds.

Alternatively, if you want to gate at the call site instead, `nodeData.type` is available as a string and `lgraphNode.value?.isSubgraphNode()` can detect subgraph nodes. But the two-tier composable approach is cleaner because it keeps the check co-located with the GLSL logic and handles the subgraph case internally.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,518 @@
...
+
+  function dispose(): void {
```
christian-byrne: issue: `dispose()` cancels the debounce and disposes the WebGL renderer, but never revokes the last preview blob URL that was stored via `setNodePreviewsByLocatorId`/`setNodePreviewsByNodeId`. Each unmount of a GLSL node component permanently leaks a blob URL.

Call `nodeOutputStore.revokePreviewsByLocatorId()` (or the equivalent cleanup) in `dispose()` for the relevant node.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,518 @@
...
+      for (let i = 0; i < floatValues.value.length; i++) {
+        r.setFloatUniform(i, floatValues.value[i])
+      }
+      for (let i = 0; i < intValues.value.length; i++) {
+        r.setIntUniform(i, intValues.value[i])
+      }
+
+      r.render()
+
+      const blob = await r.toBlob()
```
christian-byrne, line 422: issue: If `dispose()` fires while `renderPreview()` is awaiting `r.toBlob()` (line 455), the renderer is destroyed mid-flight. After the await resolves, `createSharedObjectUrl` and the store update proceed with potentially invalid data. The `renderRequestId` guard only catches new-request races, not component unmount.

Add a `disposed` flag set in `dispose()`, then check it after the await: `if (requestId !== renderRequestId || disposed) return`.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,518 @@
...
+          normalizeDimension(widthWidget.value),
+          normalizeDimension(heightWidget.value)
+        )
+      }
+    }
+
+    return [DEFAULT_SIZE, DEFAULT_SIZE]
+  }
+
+  function ensureRenderer(): ReturnType<typeof useGLSLRenderer> {
```
christian-byrne, line 355: suggestion: `ensureRenderer()` creates the renderer once and caches it, but `rendererConfig` is a computed that can change (e.g. if `innerGLSLNode` changes or `comfyDynamic.autogrow` is mutated). Config changes after the first render are silently ignored because the config was destructured at construction time.

Either watch `rendererConfig` and tear down/recreate the renderer when it changes, or document that it's intentionally static.

src/stores/nodeOutputStore.ts
```diff
@@ -280,17 +280,12 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
-   * Set node preview images by node ID.
-   * Uses the current graph context to create the appropriate NodeLocatorId.
-   *
-   * @param nodeId - The node ID
-   * @param previewImages - Array of preview image URLs to store
+   * Set node preview images by NodeLocatorId directly.
    */
-  function setNodePreviewsByNodeId(
-    nodeId: string | number,
+  function setNodePreviewsByLocatorId(
```
christian-byrne, line 271: suggestion: `setNodePreviewsByExecutionId` (lines 258-279 above) duplicates the same cancel-scheduled-revoke / release-existing / retain-new / update-stores sequence found here in `setNodePreviewsByLocatorId`. `setNodePreviewsByNodeId` already delegates to `setNodePreviewsByLocatorId`, but `setNodePreviewsByExecutionId` doesn't.

Refactor to: `const nodeLocatorId = executionIdToNodeLocatorId(...); if (!nodeLocatorId) return; setNodePreviewsByLocatorId(nodeLocatorId, previewImages); latestPreview.value = previewImages;`

christian-byrne (changes requested):
suggestion (non-blocking): `latestPreview` (nodeOutputStore.ts line 52) is set in `setNodePreviewsByExecutionId` (line 277) and exported (line 516) but has no consumers anywhere in the codebase. Meanwhile `setNodePreviewsByLocatorId` doesn't set it either, creating an asymmetry. This looks like dead code -- consider removing it.

src/renderer/extensions/vueNodes/components/LGraphNode.vue
jtydhr88: fixed

src/renderer/glsl/useGLSLPreview.ts
jtydhr88: fixed

src/renderer/glsl/useGLSLPreview.ts
jtydhr88: fixed

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,772 @@
...
+    if (!graph) return null
+    const rootGraph = graph.rootGraph
+    if (!rootGraph || graph === rootGraph) return null
+
+    return (
+      rootGraph._nodes.find(
+        (n) => n.isSubgraphNode() && n.subgraph === graph
+      ) ?? null
+    )
+  })
```
christian-byrne: When does this need to recompute? Does it actually need to recomputed when `nodeRef` changes? Or just graph or rootGraph?

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,772 @@
...
+  const nodeOutputStore = useNodeOutputStore()
+  const { nodeToNodeLocatorId } = useWorkflowStore()
+
+  let renderer: ReturnType<typeof useGLSLRenderer> | null = null
+  let rendererReady = false
+  let renderRequestId = 0
+
+  const innerGLSLNode = computed(() => {
+    const node = nodeRef.value
+    if (!node?.isSubgraphNode()) return null
```
christian-byrne, line 134: Does this need to be computed? Does it ever change after initialization?

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,772 @@
...
+  }
+}
+
+/**
+ * Inner tier: all expensive GLSL preview logic.
+ * Runs inside its own effectScope so it can be created/destroyed
+ * independently of the component lifecycle.
+ * Returns a dispose function.
+ */
+function createInnerPreview(
```
christian-byrne, line 117: nit: it seems like the logic changes in a lot of methods based on whether it's a normal shader node or a subgraph node. I wonder if there's a cleaner way to decompose into two separate composables for the two situations. This is not blocking the PR at all though.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,772 @@
...
+    const outputs = nodeOutputStore.nodeOutputs
+
+    const locatorId = nodeToNodeLocatorId(node)
+    if (outputs[locatorId]?.images?.length) return true
+
+    const inner = innerGLSLNode.value
+    if (inner) {
+      const innerLocatorId = nodeToNodeLocatorId(inner)
+      if (outputs[innerLocatorId]?.images?.length) return true
+    }
```
christian-byrne: Shouldn't it be the responsibility of the centralized system to promote previews to subgraph nodes? Here we are in-lining the logic of, "if the associated inner node has an output, display it on the outer subgraph node" -- but that logic should work across all nodes with outputs/preview. If it doesn't, I think that aspect should be fixed rather that fixing inline just for the GLSL node.

src/renderer/glsl/useGLSLPreview.ts
christian-byrne, line 1: I think if we can refactor this into more separate composables and put the pure/util functions in their own file, it might help organize the code better and more importantly make it a lot easier to write unit tests for.

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -270,6 +289,49 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
...
+      curveTextures[index] = null
+    }
+
+    const texture = gl.createTexture()
+    if (!texture) return
+
+    const unit = maxInputs + index
+    gl.activeTexture(gl.TEXTURE0 + unit)
+    gl.bindTexture(gl.TEXTURE_2D, texture)
+    gl.texImage2D(
```
christian-byrne, line 319: issue: `UNPACK_FLIP_Y_WEBGL` is set to `true` globally during `init()` (line 208) and never toggled off, which affects `bindCurveTexture()` here. For a 1×256 curve LUT, flipping the single row effectively reverses the data, producing incorrect tone mapping for any shader using curve uniforms.

Additionally, both `bindCurveTexture` and `bindInputImage` unconditionally delete and recreate textures on every call, even when data hasn't changed. With `maxCurves=4` and up to 5 inputs at ~20 FPS, that's up to 180 texture create/destroy cycles per second during slider drags. Consider reusing textures via `texSubImage2D` when dimensions match.

For the flip issue:
```suggestion
    gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, false)
    gl.texImage2D(
      gl.TEXTURE_2D,
      0,
      gl.R16F,
      lut.length,
      1,
      0,
      gl.RED,
      gl.FLOAT,
      lut
    )
    gl.pixelStorei(gl.UNPACK_FLIP_Y_WEBGL, true)
```

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -191,6 +204,7 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
       if (!ctx) return false
 
       gl = ctx
+      gl.getExtension('EXT_color_buffer_float')
```
christian-byrne: issue: `gl.getExtension('EXT_color_buffer_float')` is called but the return value is discarded. If the extension is unavailable (older mobile GPUs -- Mali, PowerVR), `RGBA16F` framebuffer attachments will fail with `FRAMEBUFFER_INCOMPLETE_ATTACHMENT`. Consider checking the return value and either falling back to RGBA8 or returning `false` with a descriptive error message.

Also, the `catch {}` block at line 212 swallows all `init()` errors, so callers can only report `'WebGL2 not available'` even when the real cause is shader compilation failure or FBO allocation OOM. Consider capturing the error so callers can surface a meaningful message.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,772 @@
...
+
+    return false
+  })
+
+  const isActive = computed(
```
christian-byrne: suggestion: Inner `const isActive = computed(...)` shadows the outer `const isActive = ref(false)` at line 189. They're connected via a watch relay (lines 304-310). The dual naming makes it hard to reason about which `isActive` is in scope. Consider renaming the inner one to `shouldRender` or similar to avoid the shadowing.

src/renderer/glsl/useGLSLPreview.test.ts
```diff
@@ -0,0 +1,230 @@
...
+  } as unknown as LGraphNode
+}
+
+function wrapNode(
+  node: LGraphNode | null
+): MaybeRefOrGetter<LGraphNode | null> {
+  return ref(node) as MaybeRefOrGetter<LGraphNode | null>
+}
+
+describe('useGLSLPreview', () => {
```
christian-byrne, line 119: suggestion: The tests currently cover only activation/deactivation (smoke tests). Consider adding coverage for the render pipeline -- at minimum asserting that `compileFragment`, `render`, `toBlob`, and `setNodePreviewsByLocatorId` are called in sequence when a shader source changes. Debounce behavior, resolution clamping, compilation error propagation to `lastError`, subgraph node paths, and cleanup/disposal (revoking blob URLs) would also be valuable to cover.

src/stores/nodeOutputStore.ts
```diff
@@ -261,6 +261,17 @@ export const useNodeOutputStore = defineStore('nodeOutput', () => {
...
     const nodeLocatorId = executionIdToNodeLocatorId(app.rootGraph, executionId)
     if (!nodeLocatorId) return
+    setNodePreviewsByLocatorId(nodeLocatorId, previewImages)
+    latestPreview.value = previewImages
+  }
+
+  /**
+   * Set node preview images by NodeLocatorId directly.
+   */
+  function setNodePreviewsByLocatorId(
```
christian-byrne, line 271: suggestion (non-blocking): `setNodePreviewsByLocatorId` does not set `latestPreview.value` (only `setNodePreviewsByExecutionId` does). This means GLSL previews won't appear in any UI that reads `latestPreview`. This is likely intentional since `latestPreview` tracks execution previews, but a brief comment here would prevent future confusion.

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,772 @@
...
+      const curves = curveValues.value
+      for (let i = 0; i < curves.length; i++) {
+        r.bindCurveTexture(i, curveDataToLUT(curves[i]))
+      }
+
+      r.render()
+
+      const blob = await r.toBlob()
+      if (requestId !== renderRequestId || disposed) return
+      const blobUrl = createSharedObjectUrl(blob)
```
christian-byrne, line 424: issue: If an exception is thrown between `createSharedObjectUrl(blob)` (line 697) and `releaseSharedObjectUrl(blobUrl)` (line 710) -- e.g., inside `setNodePreviewsByLocatorId` on a stale node reference -- the blob URL leaks with refcount=1 and is never revoked. Since `renderPreview` fires on every debounced change (~50ms), a persistent failure could leak several JPEG blobs per second (~100-500KB each).

Also, the create-then-release pattern here relies on the store internally calling `retainSharedObjectUrl`, which is invisible at this call site. A brief comment like `// Release our creation ref; the store retains its own ref` would prevent confusion.

Consider wrapping in a `finally` block to always release the creation ref:
```ts
const blobUrl = createSharedObjectUrl(blob)
try {
  // ... store writes that retain their own ref ...
} finally {
  releaseSharedObjectUrl(blobUrl)
}
```

src/renderer/glsl/useGLSLPreview.ts
```diff
@@ -0,0 +1,772 @@
...
+  const intValues = computed(() =>
+    collectUniformValues(
+      uniformSources.value?.ints,
+      'ints',
+      'u_int',
+      rendererConfig.value.maxIntUniforms
+    )
+  )
+
+  function collectBoolValues(
```
christian-byrne: suggestion (non-blocking): `collectBoolValues` (lines 421-461) is structurally identical to `collectUniformValues` (lines 359-401) -- same subgraph-source iteration, same widget fallback, same break logic. The only difference is `Boolean()` vs `Number()` casting. Consider generalizing with a `coerce` callback to avoid duplicating the logic (and needing to apply future bugfixes in two places).

christian-byrne (commented):
**Additional note (line not in diff):**

`src/renderer/glsl/useGLSLRenderer.ts:439`

suggestion (non-blocking): `convertToBlob({ type: 'image/jpeg' })` discards any alpha channel. For shaders producing meaningful alpha (e.g., matte generation), transparent areas will render as black in the preview, which could mislead users. `image/webp` supports alpha with similar compression, or a comment documenting the intentional alpha discard would help.

src/renderer/glsl/useGLSLPreview.ts
christian-byrne, line 117: suggestion (non-blocking): `createInnerPreview` is ~530 lines handling uniform collection, image loading, resolution detection, renderer lifecycle, and render orchestration. Consider extracting the uniform collection logic (`collectUniformValues`, `collectBoolValues`, `curveValues`) into a standalone composable like `useGLSLUniforms` -- this would reduce `createInnerPreview` to an ~100-line orchestration layer and make each concern independently testable.

src/renderer/glsl/useGLSLRenderer.ts
jtydhr88: FLIP_Y on a height=1 texture is a no-op (it flips rows, not pixels within a row), but added the toggle as a defensive measure. texSubImage2D reuse is a valid optimization will address in a follow-up.

src/components/curve/curveUtils.ts
```diff
@@ -192,3 +192,15 @@ export function curvesToLUT(
...
+  curve: CurveData,
+  size: number = 256
+): Float32Array {
+  const lut = new Float32Array(size)
+  const interpolate = createInterpolator(curve.points, curve.interpolation)
+  for (let i = 0; i < size; i++) {
+    lut[i] = interpolate(i / (size - 1))
+  }
+  return lut
+}
```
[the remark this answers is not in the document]
jtydhr88: size is always 256 in practice and a 1-entry LUT has no meaningful use case. Not worth the added complexity.

src/renderer/glsl/useGLSLPreview.ts
jtydhr88: fixed

src/renderer/glsl/useGLSLPreview.ts
jtydhr88: fixed

src/renderer/glsl/useGLSLPreview.ts
jtydhr88: fixed

src/stores/nodeOutputStore.ts
jtydhr88: fixed

src/renderer/glsl/useGLSLPreview.test.ts
jtydhr88: fixed

src/renderer/glsl/useGLSLPreview.ts
jtydhr88: fixed

src/renderer/glsl/useGLSLRenderer.ts
jtydhr88: fixed

src/components/curve/curveUtils.ts
christian-byrne, line 1: Why is curveUtils.ts in src/components?

src/renderer/glsl/useGLSLPreview.ts
christian-byrne: verified - ensureRenderer now compares config and recreates when it changes

src/renderer/glsl/useGLSLPreview.ts
christian-byrne: note (non-blocking): the refactor commit extracted uniforms and utils into separate files which helps, but the normal-vs-subgraph split is still worth considering in a followup

src/renderer/glsl/useGLSLPreview.ts
christian-byrne: note (non-blocking): still worth considering centralizing preview promotion logic, but fine for now since it's contained in one composable

src/components/curve/curveUtils.ts
christian-byrne: note (non-blocking): pre-existing location - the new curveDataToFloatLUT follows the same convention. worth moving curve utilities out of components/ in a followup

src/renderer/glsl/useGLSLPreview.ts
christian-byrne: suggestion: agreed these don't need to be computed, but the IIFE pattern is unnecessarily indirect. Since these are stable values resolved once at creation time, the cleaner approach is to resolve them in the outer tier and pass as plain parameters to `createInnerPreview`. That makes the "computed once" contract explicit at the call site.

Alternatively, extract as a pure helper: `const innerGLSLNode = findInnerGLSLNode(nodeRef.value)` - fits with the direction of moving pure functions to `glslPreviewUtils.ts`.

src/renderer/glsl/useGLSLPreview.ts
christian-byrne: suggestion: same as the innerGLSLNode thread - the IIFE conversion avoids unnecessary reactivity but the pattern itself is indirect. Prefer passing as a plain resolved value to `createInnerPreview`, or use a helper function call.

jtydhr88:
I can only see opened issue of moving curve util which can be done in the future, anything else needs to be done right now @christian-byrne

christian-byrne (approved):

PR #10355, opened 2026-03-21
Support text outputs in App mode

AustinMroz:
I had planned to postpone this branch until I could support text outputs in the assets tab. The current assets tab implementation is heavily dependent upon filenames for typing instead of utilizing mediaType. In the interest of pivoting to bug fixes, I'm opening the PR as is with just support for App mode.

This branch removes some guards that were very recently added in #9622
- Text is no longer explicitly blacklisted as a previewable output since it can now be displayed (in linear mode).
- Having a filename is no longer a requirement for an output being previewable.

While these changes are required in order for text to be displayed, they do specifically revert the extra safety checks which were added in a recent PR. Extra caution is appreciated.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10355-Support-text-outputs-in-App-mode-32a6d73d3650815c8239c5fbfcd12127) by [Unito](https://www.unito.io)


christian-byrne (commented):
The backend sends text outpus in the jobs api? I thought previewable outputs filtered out anything that isn't video, image, 3d? Does it work on both cloud and local? We should ensure this is fully reflected in the APIs/schemas.

AustinMroz:
I can confirm the PR as written functions for App mode on both local and cloud, but the format is a little inconsistent
- The partial response includes the text output limited to 1024 characters and is prewrapped in a `content` prop
- The full output is an array of strings.

I have no objections to postponing this PR until time can be allocated to also clean up the assets panel, but I'm trying to do a better job about leaving work in a usable state instead of a local only branch that gets lost forever.

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -221,7 +223,15 @@ const fileKind = computed((): MediaKind => {
...
 
+function isValidMediaType(val: string): val is PreviewKind {
+  return MEDIA_TYPES.includes(val as PreviewKind)
+}
+
 const previewKind = computed((): PreviewKind => {
+  const type = getOutputAssetMetadata(asset?.user_metadata)?.allOutputs?.at(
+    -1
+  )?.mediaType
+  if (type && isValidMediaType(type)) return type
```
christian-byrne: issue (blocking): the new `MEDIA_TYPES` vocabulary uses singular `image` and uppercase `3D`, but `ResultItemImpl.mediaType` carries the backend output key, which is `images` (plural) and `3d` (lowercase). Both fail `isValidMediaType` and silently fall through to the filename-derived path, so the new metadata branch effectively only fires for `text`/`audio`/`video`.

Either normalize before lookup (e.g. `type === 'images' ? 'image' : type === '3d' ? '3D' : type`), or restrict the metadata path to `'text'` (the one kind the filename fallback cannot infer).

src/services/jobOutputCache.ts
```diff
@@ -79,7 +79,9 @@ export async function getOutputsForTask(
 
 function getPreviewableOutputs(outputs?: TaskOutput): ResultItemImpl[] {
   if (!outputs) return []
-  return ResultItemImpl.filterPreviewable(parseTaskOutput(outputs))
+  return ResultItemImpl.filterPreviewable(parseTaskOutput(outputs)).filter(
+    (i) => i.filename
```
christian-byrne: issue (blocking): the new `.filter(i => i.filename)` here is redundant with `mapOutputsToAssetItems` (which already drops `!output.filename`) and creates a stealth divergence with the parallel write path in `assetsStore.mapHistoryToAssets`, which stores `task.previewableOutputs` directly into `user_metadata.allOutputs` and keeps text. Net effect: linear-mode sees text outputs but the lazy-load gallery path silently drops them.

Suggest removing the filter here and keeping the single chokepoint in `mapOutputsToAssetItems`.

src/stores/resultItemParsing.ts
```diff
@@ -15,8 +15,6 @@ function isResultItem(item: unknown): item is ResultItem {
 
   const candidate = item as Record<string, unknown>
 
-  if (typeof candidate.filename !== 'string') return false
-
   if (
```
christian-byrne, line 19: issue (blocking): with the filename guard removed, `{ mediaType: 'images' }` (no filename) now passes `isResultItem`, becomes a `ResultItemImpl` with `filename = ''`, `isImage` returns true via the `mediaType === 'images'` branch, satisfies `supportsPreview`, and reaches `MediaOutputPreview` rendering `<ImagePreview src="" />`. Same shape for video/audio. Previously the filename check rejected these.

Suggest keeping the broader unblock for text but preserving the filename requirement for non-text mediaTypes, e.g. by passing `mediaType` into `isResultItem` and requiring `typeof candidate.filename === 'string' || mediaType === 'text'`.

src/stores/resultItemParsing.test.ts
```diff
@@ -107,18 +107,6 @@ describe(parseNodeOutput, () => {
     expect(result[0].mediaType).toBe('images')
   })
 
-  it('excludes non-ResultItem array items', () => {
```
christian-byrne, line 133: issue (blocking): the two deleted tests (`excludes non-ResultItem array items`, `excludes items missing filename`), and their mirrors in `flattenNodeOutput.test.ts`, were guarding a real regression on the linearMode path. Under the new `isResultItem`, `{ randomKey: 123 }` passes (no `type` is treated as valid), so `parseNodeOutput('1', { custom_data: [{ randomKey: 123 }] })` returns a `ResultItemImpl` with `mediaType: 'custom_data'`. `jobOutputCache.getPreviewableOutputs` masks this with the new filename filter, but `flattenNodeOutput` calls `parseNodeOutput` directly with no equivalent filter.

Suggest restoring these tests (they should pass once the filename requirement is preserved for non-text mediaTypes), or moving the safety into a single chokepoint.

src/stores/resultItemParsing.ts
```diff
@@ -35,6 +33,11 @@ export function parseNodeOutput(
     .filter(([key, value]) => !METADATA_KEYS.has(key) && Array.isArray(value))
     .flatMap(([mediaType, items]) =>
       (items as unknown[])
+        .map((item) =>
+          mediaType === 'text' && typeof item === 'string'
```
christian-byrne, line 40: issue (blocking): the headline new behavior, `mediaType === 'text' && typeof item === 'string' ? { nodeId, mediaType, content: item } : item`, has no direct test. The un-skipped `should parse text outputs` case in `queueStore.test.ts` exercises `preview_output: { nodeId, mediaType, content }`, which is already an object and never hits the new wrapper branch. Grep for `text: [` across test files returns zero matches.

Suggest adding a direct `parseNodeOutput('1', { text: ['hello', 'world'] })` test asserting two `ResultItemImpl` items with `mediaType: 'text'`, `content: 'hello'`/`'world'`, and `filename === ''`.

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -160,7 +161,8 @@ import type { MediaKind } from '../schemas/mediaAssetSchema'
 import { MediaAssetKey } from '../schemas/mediaAssetSchema'
 import MediaTitle from './MediaTitle.vue'
 
-type PreviewKind = ReturnType<typeof getMediaTypeFromFilename>
+const MEDIA_TYPES = ['video', 'audio', 'image', '3D', 'text', 'other'] as const
```
christian-byrne: suggestion (non-blocking): `MEDIA_TYPES`/`PreviewKind` duplicates `zMediaKindSchema` from `../schemas/mediaAssetSchema`, which is already imported on line 160 (as `MediaKind`). Could simplify to `zMediaKindSchema.safeParse(type).success` and reuse `MediaKind` directly, dropping the local const and handwritten guard.

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -221,7 +223,15 @@ const fileKind = computed((): MediaKind => {
   return getMediaTypeFromFilename(asset?.name || '')
 })
 
+function isValidMediaType(val: string): val is PreviewKind {
+  return MEDIA_TYPES.includes(val as PreviewKind)
+}
+
 const previewKind = computed((): PreviewKind => {
+  const type = getOutputAssetMetadata(asset?.user_metadata)?.allOutputs?.at(
```
christian-byrne: question: the asset's `preview_url` is set from `task.previewOutput`, which is `findLast(type === 'output') ?? at(-1)` (queueStore.ts:286-293), but `previewKind` here reads `allOutputs.at(-1)?.mediaType`. For a multi-output job whose last item is `temp` (e.g. `[image_output, text_temp]`), the card's preview URL points at the image while `previewKind` resolves to `'text'`. Is that intended, or should the kind selection mirror the `preview_url` selection logic?

src/stores/resultItemParsing.ts
```diff
@@ -35,6 +33,11 @@ export function parseNodeOutput(
     .filter(([key, value]) => !METADATA_KEYS.has(key) && Array.isArray(value))
```
christian-byrne, line 36: question: `text` is declared in `apiSchema.ts` as `z.union([z.string(), z.array(z.string())])`, but `parseNodeOutput` only iterates `Array.isArray(value)` keys, so a top-level `text: 'hello'` is silently dropped (the retained `excludes text key` test relies on this). This PR therefore enables `text: ['...']` only, not `text: '...'`. Worth confirming with the backend producer which shape is actually emitted, and either narrowing the schema to `z.array(z.string())` or wrapping a top-level string into a single-item array before flatMap.

src/stores/resultItemParsing.ts
```diff
@@ -15,8 +15,6 @@ function isResultItem(item: unknown): item is ResultItem {
 
   const candidate = item as Record<string, unknown>
```
christian-byrne, line 10: nitpick (non-blocking): the JSDoc on `isResultItem` still says "Requires `filename` (string) since ResultItemImpl needs it for a valid URL." The check it documents was just removed; worth updating or dropping the comment.

DrJKL (dismissed):

src/services/jobOutputCache.ts
```diff
@@ -79,7 +79,9 @@ export async function getOutputsForTask(
...
+  )
```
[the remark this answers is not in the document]
DrJKL: Addressed in the latest revision: the filename filter was removed, so filename-less text remains previewable on the job-detail path.

src/platform/assets/components/MediaAssetCard.vue
DrJKL: Addressed by removing the asset-sidebar metadata branch from this PR. Text rendering is now scoped to App mode, so no media-type normalization is needed here.

src/services/jobOutputCache.ts
DrJKL: Addressed: the redundant filename filter was removed. Asset conversion remains the single filename-enforcing boundary for asset items.

src/stores/resultItemParsing.ts
DrJKL: Addressed: filename-less items are accepted only when `mediaType === "text"` and `content` is a string. Non-text outputs still require a string filename.

src/stores/resultItemParsing.test.ts
DrJKL: Addressed: both regression tests were retained. Arbitrary array objects and filename-less non-text outputs are rejected by the parser.

src/stores/resultItemParsing.ts
DrJKL: Addressed: a direct `parseNodeOutput` test now covers `text: ["hello"]` and verifies the normalized text result.

src/platform/assets/components/MediaAssetCard.vue
DrJKL: Addressed by removing the asset-sidebar preview-kind change from this PR; the duplicate media-type vocabulary no longer exists.

src/platform/assets/components/MediaAssetCard.vue
DrJKL: Addressed by removing the asset-sidebar metadata branch. Existing preview URL and kind selection behavior is unchanged.

src/stores/resultItemParsing.ts
DrJKL: Confirmed: the full output payload is emitted as an array of strings, while the partial preview response is already wrapped in a `content` field. This revision intentionally enables the array form and continues treating a scalar `text` value as metadata.

src/stores/resultItemParsing.ts
DrJKL: Addressed: the stale JSDoc was removed when the validator was updated.

src/platform/assets/components/MediaAssetCard.test.ts
```diff
@@ -83,6 +83,24 @@ function dispatchDragStart(
...
+        tags: [],
+        user_metadata: {
+          jobId: 'job',
+          subfolder: '',
+          allOutputs: [{ isText: true }]
+        }
+      })
+    })
+
+    expect(await screen.findByTestId('text-preview')).toBeInTheDocument()
```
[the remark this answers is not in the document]
DrJKL: Addressed by removing the out-of-scope asset-card test. App-mode rendering now has behavioral Playwright coverage that asserts the inline text is visible.

src/platform/assets/components/MediaAssetCard.vue
```diff
@@ -264,6 +264,10 @@ const fileKind = computed((): MediaKind => {
 })
 
 const previewKind = computed((): PreviewKind => {
+  const output = getOutputAssetMetadata(asset?.user_metadata)?.allOutputs?.at(
+    -1
+  )
+  if (output?.isText) return 'text'
```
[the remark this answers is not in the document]
DrJKL: Addressed by removing the asset-card text branch from this PR. Inline content is rendered through the existing App-mode `MediaOutputPreview` path instead.

src/renderer/extensions/linearMode/useOutputHistory.test.ts
```diff
@@ -226,6 +226,29 @@ describe(useOutputHistory, () => {
...
+        ...makeAsset('a1', 'job-1', {
+          allOutputs: [text],
+          outputCount: 1
+        }),
+        preview_url: undefined
+      }
+
+      const { allOutputs } = useOutputHistory()
+
+      expect(allOutputs(asset)).toEqual([text])
```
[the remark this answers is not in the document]
DrJKL: Addressed in the latest revision: the assertion now uses strict array comparison. The parser test separately uses `objectContaining` to verify the required text fields, while this test targets the filename-less metadata path.

christian-byrne (approved):

PR #10357, opened 2026-03-21
refactor: extract shared click-vs-drag guard utility

dante01yoon:
## Summary

Extract duplicated click-vs-drag detection logic into a shared `useClickDragGuard` composable and `exceedsClickThreshold` pure utility function.

## Changes

- **What**: New `useClickDragGuard(threshold)` composable in `src/composables/useClickDragGuard.ts` that stores pointer start position and checks squared distance against a threshold. Also exports `exceedsClickThreshold` for non-Vue contexts.
- Migrated `DropZone.vue`, `useNodePointerInteractions.ts`, and `Load3d.ts` to use the shared utility
- `CanvasPointer.ts` left as-is (LiteGraph internal)
- All consumers now use squared-distance comparison (no `Math.sqrt` or per-axis `Math.abs`)

## Review Focus

- The composable uses plain `let` state instead of `ref` since reactivity is not needed for the start position
- `Load3d.ts` uses the pure `exceedsClickThreshold` function directly since it is a class, not a Vue component
- Threshold values preserved per-consumer: DropZone=5, useNodePointerInteractions=3, Load3d=5

Fixes #10356

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10357-refactor-extract-shared-click-vs-drag-guard-utility-32a6d73d3650816e83f5cb89872fb184) by [Unito](https://www.unito.io)


DrJKL (changes requested):
Could you add a Playwright test for this?
Thanks for grabbing this before the other PR even got merged 😁

src/renderer/extensions/linearMode/DropZone.vue
```diff
@@ -30,26 +31,19 @@ const {
...
-        pointerStart.value = null
-        return
-      }
-    }
+  if (e.detail !== 0 && clickGuard.wasDragged(e)) {
+    clickGuard.reset()
+    return
   }
-  pointerStart.value = null
+  clickGuard.reset()
```
DrJKL: I hate that this has to be duplicated :-/

src/composables/useClickDragGuard.ts
```diff
@@ -0,0 +1,41 @@
+interface PointerPosition {
+  readonly x: number
+  readonly y: number
+}
```
DrJKL, line 4: Nit: There's probably an existing interface we can reuse somewhere.

DrJKL (dismissed):
Want to update ImagePreview now that the grid changes are in?

src/renderer/extensions/linearMode/DropZone.vue
dante01yoon: Addressed

src/composables/useClickDragGuard.ts
dante01yoon: Checked — the existing Point interfaces all live in renderer/ or extensions/, so importing them here would violate the layer rules (composables/ → renderer/). Creating a shared type in src/types/ just for this felt like overkill, so I kept it local for now.

PR #10358, opened 2026-03-21
fix: bring cloned/pasted nodes to front in Vue renderer

dante01yoon:
## Summary

Cloned/pasted nodes now appear above existing nodes in Node 2.0 (Vue renderer) mode instead of behind them.

## Changes

- **What**: After `_deserializeItems` creates new nodes via clone or paste, their z-index is set above all existing nodes so they render on top and are immediately accessible.

## Review Focus

The fix follows the same pattern as `bringNodeToFront` in `layoutMutations.ts` and the existing alt-click clone handler in `LGraphNode.vue`. It applies to both context menu "Clone" and paste operations.

Fixes #10307


dante01yoon:
Recreating with red-green-fix workflow

PR #10360, opened 2026-03-21
fix: prevent survey popup when surveyed feature is inactive

### LINKED ISSUE #10333: Deprioritize API nodes in new serach and remove popup message.

When using new Nodes Search, API nodes always show first. Moreover, when the new Nodes Search is switched off, a popup is shown:

Image

This is unacceptable for an opensource project. API nodes use closed-source services and should never be prioritized over local inference. Advertisement popups are even worse. 

Always

Multiple times per day

Blocks me from completing tasks

### Current workarounds

_No response_

Local nodes should be shown first. Popup should be removed.

### Additional context

_No response_

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-10333-Deprioritize-API-nodes-in-new-serach-and-remove-popup-message-3296d73d3650810f9f8fdb61ef1f1ce8) by [Unito](https://www.unito.io)

dante01yoon:
## Summary

Prevent the nightly survey popup from appearing when the feature being surveyed is currently disabled.

## Root Cause

`useSurveyEligibility` checks the feature usage count from localStorage but does not verify whether the surveyed feature is currently active. When a user uses the new node search 3+ times (reaching the survey threshold), then switches back to legacy search, the usage count persists in localStorage and the survey popup still appears despite the feature being off.

## Steps to Reproduce

1. Enable new node search (Settings > Node Search Box Implementation > "default")
2. Use node search 3+ times (double-click canvas, search for nodes)
3. Switch back to legacy search (Settings > Node Search Box Implementation > "litegraph (legacy)")
4. Wait 5 seconds on a nightly localhost build
5. **Expected**: No survey popup appears (feature is disabled)
6. **Actual**: Survey popup appears because eligibility only checks stored usage count, not current feature state

## Changes

1. Added optional `isFeatureActive` callback to `FeatureSurveyConfig` interface
2. Added `isFeatureActive` guard in `useSurveyEligibility` eligibility computation
3. Configured `node-search` survey with `isFeatureActive` that checks the current search box setting

## Red-Green Verification

| Commit | Result | Description |
|---|---|---|
| `99e7d7fe9` `test:` | RED | Asserts `isEligible` is false when `isFeatureActive` returns false. Fails on current code. |
| `01df9af12` `fix:` | GREEN | Adds `isFeatureActive` check to eligibility. Test passes. |

Fixes #10333

### DIFF
--- a/src/platform/surveys/surveyRegistry.ts
+++ b/src/platform/surveys/surveyRegistry.ts
@@ -1,3 +1,5 @@
+import { useSettingStore } from '@/platform/settings/settingStore'
+
 import type { FeatureSurveyConfig } from './useSurveyEligibility'
 
 /**
@@ -9,7 +11,13 @@ export const FEATURE_SURVEYS: Record<string, FeatureSurveyConfig> = {
     featureId: 'node-search',
     typeformId: 'goZLqjKL',
     triggerThreshold: 3,
-    delayMs: 5000
+    delayMs: 5000,
+    isFeatureActive: () => {
+      const settingStore = useSettingStore()
+      return (
+        settingStore.get('Comfy.NodeSearchBoxImpl') !== 'litegraph (legacy)'
+      )
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


Yourz (approved):
LGTM

PR #10361, opened 2026-03-21
fix: place cloned node above original in Vue renderer

dante01yoon:
## Summary

Cloned/pasted nodes in Node 2.0 (Vue renderer) mode now appear above the original node instead of behind it.

## Root Cause

The legacy LiteGraph canvas renderer uses array ordering for z-ordering: nodes are stored in `graph._nodes` and drawn sequentially, so newly added nodes (appended to the end) are automatically drawn on top. There is no explicit z-index.

The Vue renderer (Node 2.0) uses explicit CSS `z-index` for node ordering. New nodes default to `zIndex: 0` in `layoutMutations.ts`. When a node has been interacted with, `bringNodeToFront` raises its z-index. A cloned node at z-index 0 therefore appears behind any previously interacted node.

The alt-click clone path in `LGraphNode.vue` already handles this correctly by calling `bringNodeToFront()` after cloning. However, the menu clone and keyboard paste paths go through `_deserializeItems` in `LGraphCanvas.ts`, which does not set z-index for new nodes.

| Clone method | Legacy renderer | Vue renderer (before fix) | Vue renderer (after fix) |
|---|---|---|---|
| Alt-click drag | On top (array order) | On top (`bringNodeToFront` called) | On top |
| Right-click menu Clone | On top (array order) | Behind original (z-index 0) | On top |
| Ctrl+C / Ctrl+V | On top (array order) | Behind original (z-index 0) | On top |

## Steps to Reproduce

1. Enable Node 2.0 mode (Vue renderer) in settings
2. Add any node to the canvas
3. Click or drag the node (raises its z-index via `bringNodeToFront`)
4. Right-click the node and select "Clone"
5. **Expected**: Cloned node appears above the original, immediately draggable
6. **Actual**: Cloned node appears behind the original; user must move the original to access the clone

## Changes

After `batchUpdateNodeBounds` in `_deserializeItems`, calls `bringNodeToFront` for each newly created node so they receive a z-index above all existing nodes.

## Side Effect Analysis

Checked all call sites of `_deserializeItems`:

1. **Initial graph load / workflow open**: `loadGraphData` in `app.ts` does NOT call `_deserializeItems`. Workflow loading goes through `LGraph.configure()` which directly adds nodes and links. The layout store is initialized separately via `initializeFromLiteGraph`. No side effect.

2. **Paste from clipboard (Ctrl+V)**: Both `usePaste.ts` (line 52) and `pasteFromClipboard` (line 4080) call `_deserializeItems`. Pasted nodes appearing on top is the correct and desired behavior. No issue.

3. **Undo/Redo**: `ChangeTracker.updateState()` calls `app.loadGraphData()`, which does a full graph reconfigure -- it does NOT go through `_deserializeItems`. No side effect.

4. **Subgraph blueprint addition**: `litegraphService.ts` (line 906) calls `_deserializeItems` when adding subgraph blueprints from the node library. These are freshly placed nodes that should appear on top. Desired behavior.

5. **Alt-click clone in LGraphNode.vue**: This path calls `LGraphCanvas.cloneNodes()` -> `_deserializeItems()`, then separately calls `bringNodeToFront()` again on line 433 of `LGraphNode.vue`. The second call is now redundant (the node is already at max z-index), but harmless -- `bringNodeToFront` finds the current max, adds 1, and sets. The z-index will increment from N to N+1 on the second call. This is a minor redundancy, not a bug.

6. **Performance**: `bringNodeToFront` iterates all nodes in the layout store once per call (O(m)) to find max z-index. For n new nodes, the total cost is O(n*m). In practice, clone/paste operations involve a small number of nodes (typically 1-10), so this is negligible. For extremely large pastes (100+ nodes), each call also increments the max by 1, so z-indices will be sequential (which is actually a reasonable stacking order).

7. **layoutStore availability**: `layoutStore` is a module-level singleton (`new LayoutStoreImpl()`) -- not a Pinia store -- so it is always available. The `useLayoutMutations()` composable is a plain function returning an object of closures over `layoutStore`. It does not require Vue component context. No risk of runtime errors.

8. **Legacy renderer (non-Vue mode)**: When Node 2.0 mode is disabled, the layout store still exists but is not used for rendering. Calling `bringNodeToFront` will update z-index values in the Yjs document that are never read. This is harmless.

## Red-Green Verification

| Commit | Result | Description |
|---|---|---|
| `6894b99` `test:` | RED | Test asserts cloned node z-index > original. Fails with `expected 0 to be greater than 5`. |
| `3567469` `fix:` | GREEN | Calls `bringNodeToFront` for each new node in `_deserializeItems`. Test passes. |

Fixes #10307


src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
```diff
@@ -0,0 +1,156 @@
...
+    canvas = createCanvas(graph)
+    LGraphCanvas.active_canvas = canvas
+
+    layoutStore.initializeFromLiteGraph([])
+
+    // Simulate Vue runtime: create layout entries when nodes are added
+    graph.onNodeAdded = (node: LGraphNode) => {
+      createLayoutEntry(node, 0)
+    }
+  })
```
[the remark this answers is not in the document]
dante01yoon: @coderabbitai review again

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    timestamp: Date.now(),
+    source: LayoutSource.Canvas,
+    actor: 'test'
+  })
+}
+
+describe('cloned node z-index in Vue renderer', () => {
+  let graph: LGraph
+  let canvas: LGraphCanvas
+  let previousVueNodesMode: boolean
```
DrJKL, line 118: Do we need global mutable variables in the `describe` for a single `it`?

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
```diff
@@ -0,0 +1,163 @@
...
+class TestNode extends LGraphNode {
+  static override type = 'test/CloneZIndex'
+
+  constructor(title?: string) {
+    super(title ?? 'test/CloneZIndex')
+    this.type = 'test/CloneZIndex'
+  }
+}
+
+function createCanvas(graph: LGraph): LGraphCanvas {
```
DrJKL, line 25: I think we have this as a utility somewhere.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -4270,6 +4271,11 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
     if (newPositions.length) layoutStore.setSource(LayoutSource.Canvas)
     layoutStore.batchUpdateNodeBounds(newPositions)
 
+    const { bringNodeToFront } = useLayoutMutations()
+    for (const { nodeId } of newPositions) {
+      bringNodeToFront(nodeId)
+    }
+
```
DrJKL, line 4284: **Performance (high):** `bringNodeToFront()` called in a loop creates O(n×m) complexity (n = cloned nodes, m = total graph nodes). Each call iterates through `layoutStore.getAllNodes().value` to find the max z-index, and each `setNodeZIndex` call invalidates that computed ref, forcing a full re-scan on the next iteration.

In a 500-node graph, cloning 20 nodes would cause 10,000+ object instantiations.

**Suggestion:** Compute max z-index once before the loop, then assign sequential values:
```ts
const allNodes = layoutStore.getAllNodes().value
let maxZIndex = 0
for (const [, layout] of allNodes) {
  if (layout.zIndex > maxZIndex) maxZIndex = layout.zIndex
}
const { setNodeZIndex } = useLayoutMutations()
for (let i = 0; i < newPositions.length; i++) {
  setNodeZIndex(newPositions[i].nodeId, maxZIndex + i + 1)
}
```
Or better yet, add a `batchBringNodesToFront(nodeIds)` method to `useLayoutMutations`.

DrJKL (commented):
## Code Review

**15 automated checks ran** — 12 passed clean, 3 found actionable issues.

### Key Findings

1. **Performance (high):** `bringNodeToFront()` in a loop is O(n×m) — each call re-scans all nodes to find max z-index. Should compute max once, then assign sequential z-indices.
2. **Architecture (medium):** Inconsistent `LayoutSource` tracking between `batchUpdateNodeBounds` and the `bringNodeToFront` loop.
3. **Test coverage (medium):** Only covers single-node Vue-mode clone. Missing: multi-node clone, non-Vue mode, paste path.
4. **DX (low):** Duplicated string literal and missing intent comment.

See inline comments for details.

src/lib/litegraph/src/LGraphCanvas.ts
```diff
@@ -4270,6 +4271,11 @@ export class LGraphCanvas implements CustomEventDispatcher<LGraphCanvasEventMap>
     if (newPositions.length) layoutStore.setSource(LayoutSource.Canvas)
     layoutStore.batchUpdateNodeBounds(newPositions)
 
+    const { bringNodeToFront } = useLayoutMutations()
+    for (const { nodeId } of newPositions) {
```
DrJKL: **Architecture (medium):** Inconsistent source tracking — `batchUpdateNodeBounds` on L4273 internally saves/restores `currentSource` (Canvas → Vue → Canvas), so after it returns the source is back to Canvas. However, the `bringNodeToFront` calls here go through `useLayoutMutations()` which reads `layoutStore.getCurrentSource()`. Depending on timing, the CRDT source attribution for z-index changes may not match the source used for the bounds update.

Consider explicitly calling `layoutStore.setSource(LayoutSource.Canvas)` before this loop to make the intent clear.

---

**DX (low):** A brief comment explaining *why* cloned nodes need z-index promotion would help future maintainers, e.g.:
```ts
// Bring cloned/pasted nodes to front so they render above the originals
```

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
```diff
@@ -0,0 +1,163 @@
...
+    graph.onNodeAdded = (node: LGraphNode) => {
+      createLayoutEntry(node, 0)
+    }
+  })
+
+  afterEach(() => {
+    LiteGraph.vueNodesMode = previousVueNodesMode
+  })
+
+  it('places cloned nodes above the original node z-index', () => {
```
DrJKL, line 142: **Test coverage (medium):** Good regression test for the single-node Vue-mode case. Consider adding coverage for:

1. **Multi-node clone** — verifies each cloned node gets a distinct, incrementally higher z-index (important given the loop behavior).
2. **Non-Vue mode** — the production code runs unconditionally regardless of `vueNodesMode`, but only Vue mode is tested.
3. **Paste path** — `_deserializeItems` is also called from `_pasteFromClipboard`; the z-index fix applies there too.

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
```diff
@@ -0,0 +1,163 @@
...
+
+import {
+  LGraph,
+  LGraphCanvas,
+  LGraphNode,
+  LiteGraph
+} from '@/lib/litegraph/src/litegraph'
+
+class TestNode extends LGraphNode {
+  static override type = 'test/CloneZIndex'
```
DrJKL: **DX (low):** The string `"test/CloneZIndex"` is repeated 4 times. Extract to a constant:
```ts
const TEST_NODE_TYPE = 'test/CloneZIndex' as const
```

src/lib/litegraph/src/LGraphCanvas.ts
dante01yoon: Addressed in 7ae5f19. Replaced with single-pass max z-index computation + sequential `setNodeZIndex` assignment — O(n+m) instead of O(n×m).

src/lib/litegraph/src/LGraphCanvas.ts
dante01yoon: Addressed in 7ae5f19. `setSource(LayoutSource.Canvas)` is already called on the line above `batchUpdateNodeBounds`, so `setNodeZIndex` (via `useLayoutMutations`) reads the same Canvas source. Also added intent comment.

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
dante01yoon: Added multi-node clone test in 7ae5f19 — verifies each cloned node gets a distinct z-index above the highest original.

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
dante01yoon: Done in 7ae5f19 — extracted `TEST_NODE_TYPE` constant.

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
dante01yoon: Now there are 2 test cases, so describe-level variables make sense. Also extracted `setZIndex` helper to reduce duplication.

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
dante01yoon: Checked — `groupSelection.test.ts` has an identical `createCanvas`. Extracting to a shared utility would be a nice cleanup but feels out of scope for this PR. Happy to do it in a follow-up if you think it's worth it.

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
DrJKL: I disagree. If we have two, it's still safer and clearer to instantiate within their blocks.
Global state like this is just _asking_ for cross-test contamination.

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
```diff
@@ -0,0 +1,198 @@
...
+  })
+
+  afterEach(() => {
+    LiteGraph.vueNodesMode = previousVueNodesMode
+  })
+
+  it('places cloned nodes above the original node z-index', () => {
+    const originalNode = new TestNode()
+    originalNode.pos = [100, 100]
+    originalNode.size = [200, 100]
```
DrJKL, line 145: This is another thing that we should have a central utility for, though I'd rather find a way to use the real Nodes instead of creating bespoke classes just for unit tests.
If you really want to keep the TestNode though, can it take in all of the options at construction instead of requiring post-instantiation setup?

src/lib/litegraph/src/LGraphCanvas.cloneZIndex.test.ts
```diff
@@ -0,0 +1,198 @@
...
+    nodeB.pos = [400, 100]
+    nodeB.size = [200, 100]
+    graph.add(nodeB)
+    setZIndex(String(nodeB.id), 7, 0)
+
+    const result = LGraphCanvas.cloneNodes([nodeA, nodeB])
+    expect(result).toBeDefined()
+    expect(result!.created.length).toBe(2)
+
+    const clonedA = result!.created[0] as LGraphNode
```
DrJKL, line 186: Why do these need casts?

DrJKL (approved):
We can clean up the tests in a follow-up. @coderabbitai Open a tracking issue for this.

dante01yoon:
@coderabbitai approve

dante01yoon:
@coderabbitai check the comment and create issue for cleanup test

PR #10369, opened 2026-03-21
feat: App mode - add execution status messages 

pythongosssss:
## Summary

Adds custom status messages that are shown under the previews in order to provide additional progress feedback to the user

Nodes matching the words:

Save, Preview -> Saving
Load, Loader -> Loading
Encode -> Encoding
Decode -> Decoding
Compile, Conditioning, Merge, -> Processing
Upscale, Resize -> Resizing
ToVideo -> Generating video

Specific nodes:
KSampler, KSamplerAdvanced, SamplerCustom, SamplerCustomAdvanced -> Generating
Video Slice, GetVideoComponents, CreateVideo -> Processing video
TrainLoraNode -> Training

## Changes

- **What**: 
- add specific node lookups for non-easily matchable patterns
- add regex based matching for common patterns
- show on both latent preview & skeleton preview 
- allow app mode workflow authors to override status with custom property `Execution Message` (no UI for doing this)

## Review Focus

This is purely pattern/lookup based, in future we could update the backend node schema to allow nodes to define their own status key.

image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10369-feat-App-mode-add-execution-status-messages-32a6d73d3650814e8ca2da5eb33f3b65) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/linearMode/getExecutionStatusMessage.ts
+++ b/src/renderer/extensions/linearMode/getExecutionStatusMessage.ts
@@ -17,13 +17,6 @@ type ExecutionStatusKey =
  * identifier patterns (e.g. unconventional naming, spaces).
  */
 const statusMap: Record<string, ExecutionStatusKey> = {
-  // Sampling / generation — can't use `Sampler` identifier because it
-  // would match nodes like KSamplerSelect
-  KSampler: 'generating',
-  KSamplerAdvanced: 'generating',
-  SamplerCustom: 'generating',
-  SamplerCustomAdvanced: 'generating',
-
   // Video utility nodes with non-standard naming
   'Video Slice': 'processingVideo',
   GetVideoComponents: 'processingVideo',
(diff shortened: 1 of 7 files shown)


src/renderer/extensions/linearMode/getExecutionStatusMessage.ts
```diff
@@ -0,0 +1,73 @@
...
+  | 'training'
+
+/**
+ * Specific status messages for nodes that can't be matched by PascalCase
+ * identifier patterns (e.g. unconventional naming, spaces).
+ */
+const statusMap: Record<string, ExecutionStatusKey> = {
+  // Sampling / generation — can't use `Sampler` identifier because it
+  // would match nodes like KSamplerSelect
+  KSampler: 'generating',
```
AustinMroz: I'm pretty conflicted on using a whitelist for samplers like this. There's lots of sampler custom nodes (~30 on cloud). I would assume that trivial nodes (like KSamplerSelect) would never get the green border/executing message, so it's fine if they get mis-classified under generating.

AustinMroz (dismissed):

src/renderer/extensions/linearMode/getExecutionStatusMessage.ts
pythongosssss: True, I was trying to minimize incorrect matches - but missing the sampling nodes is worse than matching some instant-executing ones

AustinMroz (approved):

PR #10387, opened 2026-03-22
feat: reorder Partner tab and prioritize core nodes in search

comfydesigner:
- Moves Partner tab after Comfy in the search filter bar
- Prioritizes core Comfy nodes over extensions/partner nodes at the same match quality in search results

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10387-feat-reorder-Partner-tab-and-prioritize-core-nodes-in-search-32b6d73d3650814a9c33c1d757f86fbd) by [Unito](https://www.unito.io)


src/stores/nodeDefStore.ts
```diff
@@ -197,7 +197,8 @@ export class ComfyNodeDefImpl
   postProcessSearchScores(scores: SearchAuxScore): SearchAuxScore {
     const nodeFrequencyStore = useNodeFrequencyStore()
     const nodeFrequency = nodeFrequencyStore.getNodeFrequencyByName(this.name)
-    return [scores[0], -nodeFrequency, ...scores.slice(1)]
+    const sourcePriority = this.isCoreNode ? 0 : 1
+    return [scores[0], sourcePriority, -nodeFrequency, ...scores.slice(1)]
```
christian-byrne, line 201: issue (non-blocking): `sourcePriority` is placed before `-nodeFrequency`, which means a rarely-used core node will always rank above a heavily-used custom node when match scores tie. Consider swapping the order to `[scores[0], -nodeFrequency, sourcePriority, ...scores.slice(1)]` so user usage patterns take priority, with source as the final tiebreaker.

src/stores/nodeDefStore.ts
```diff
@@ -197,7 +197,8 @@ export class ComfyNodeDefImpl
   postProcessSearchScores(scores: SearchAuxScore): SearchAuxScore {
     const nodeFrequencyStore = useNodeFrequencyStore()
     const nodeFrequency = nodeFrequencyStore.getNodeFrequencyByName(this.name)
-    return [scores[0], -nodeFrequency, ...scores.slice(1)]
+    const sourcePriority = this.isCoreNode ? 0 : 1
```
christian-byrne, line 200: question: `isCoreNode` is true for `comfy_api_nodes` (partner/API nodes) since `CORE_NODE_MODULES` includes it. So partner nodes already get `sourcePriority = 0` -- same as regular core nodes. Is this the intended behavior? If the goal is to boost core-but-not-partner nodes specifically, this won't distinguish them. And if the goal is core+partner over custom, then this works but is worth documenting since the overlap is non-obvious.

christian-byrne, line 200: suggestion (non-blocking): Consider making `sourcePriority` a readonly field on `ComfyNodeDefImpl` (like `isCoreNode` already is) rather than computing it inline on every search. This makes the ranking policy more visible, testable, and extensible if you later want a richer priority scale (e.g., core=0, partner=1, custom=2).

PR #10392, opened 2026-03-22
Implement widgets_values_named for serialization

AustinMroz:
Adds a `widgets_values_named` property to serialized nodes that stores a mapping of widget names to serialized values.

Adds support for restoring values from this property behind a setting that is currently disabled by default. I have ~90% confidence that enabling this is safe, but wish to roll it out in a controlled manner.
- If `widgets_values_named` exists when loading, it is used instead of widgets_values for deserialization
- If it does not exist, but the nodeDef includes a newly defined `valueMigrationDict`, that is used construct a mapping. This ensures a route for forwards compatibility when loading older workflows.
- Otherwise, the existing code is used to restore widgets from `widgets_values`

The behaviour for `node.onConfigured` is unchanged, but for future custom node development, it will also include `widgets_values_named` when that field exists. In the long run, custom nodes can use the presence of the field to determine if it was used for restoring a workflow

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10392-Implement-widgets_values_named-for-serialization-32b6d73d3650819b93abe3a4bdd83cd9) by [Unito](https://www.unito.io)


src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -907,7 +907,26 @@ export class LGraphNode
...
+          if (map[i]) namedValues[map[i]] = val
+        })
+        return namedValues
+      }
+      const namedValues = getNamedValues()
+      if (namedValues && LiteGraph.namedValuesRestore) {
+        for (const widget of this.widgets ?? []) {
+          if (widget.name in namedValues) {
+            widget.value = namedValues[widget.name]
+            widget.callback?.(widget.value)
```
DrJKL: **Bug: Behavioral asymmetry — `widget.callback` called here but not in the positional path (L934)**

The named restore path calls `widget.callback?.(widget.value)` but the existing positional path does not. This means widgets with side-effectful callbacks (e.g. combo boxes that show/hide child widgets) behave differently depending on which restore path is active.

If the callback is needed (and it likely is for dynamic widgets), it should also be added to the positional path. Otherwise, toggling `namedValuesRestore` on/off will produce different runtime behavior.

```suggestion
            widget.value = namedValues[widget.name]
```

Or, add `widget.callback?.(widget.value)` to the positional path too.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -962,16 +981,19 @@ export class LGraphNode
...
         if (widget.serialize === false) continue
         const val = widget?.value
         // Ensure object values are plain (not reactive proxies) for structuredClone compatibility.
-        o.widgets_values[i] =
+        const serialisedVal =
           val != null && typeof val === 'object'
             ? JSON.parse(JSON.stringify(val))
             : (val ?? null)
+        o.widgets_values[i] = serialisedVal
+        if (val !== null) o.widgets_values_named[widget.name] = serialisedVal
```
DrJKL: **Bug: Null values silently dropped from `widgets_values_named` → round-trip data loss**

`val !== null` skips writing null-valued widgets to the named map. But the positional array preserves null via `val ?? null`. If a user explicitly sets a widget value to `null`, saves, then loads with `namedValuesRestore = true`, that widget retains its default instead of `null`.

```suggestion
        o.widgets_values[i] = serialisedVal
        o.widgets_values_named[widget.name] = serialisedVal
```

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -907,7 +907,26 @@ export class LGraphNode
...
+        return namedValues
+      }
+      const namedValues = getNamedValues()
+      if (namedValues && LiteGraph.namedValuesRestore) {
+        for (const widget of this.widgets ?? []) {
+          if (widget.name in namedValues) {
+            widget.value = namedValues[widget.name]
+            widget.callback?.(widget.value)
+          }
+        }
```
DrJKL, line 1004: **Bug: Named path doesn't skip `serialize === false` widgets**

The positional path (L932) explicitly checks `if (widget.serialize === false) continue`. The named path iterates all widgets without this guard. If a non-serializable widget happens to share a name with a key in `widgets_values_named`, it would be incorrectly restored.

```suggestion
      if (namedValues && LiteGraph.namedValuesRestore) {
        for (const widget of this.widgets ?? []) {
          if (widget.serialize === false) continue
          if (widget.name in namedValues) {
            widget.value = namedValues[widget.name]
            widget.callback?.(widget.value)
          }
        }
```

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -907,7 +907,26 @@ export class LGraphNode
           )
       }
 
-      if (info.widgets_values) {
+      const getNamedValues = () => {
+        if (info.widgets_values_named) return info.widgets_values_named
+
+        const map = this.constructor.nodeData?.valueMigrationDict
```
DrJKL: **Type safety: `this.constructor.nodeData?.valueMigrationDict` relies on implicit `any`**

`this.constructor` is typed as `Function` in TypeScript, so accessing `.nodeData` compiles only because `Function` is `any`-indexed. Additionally, `valueMigrationDict` is not declared in the static `nodeData` type at L243–250 — it only exists on the augmented `ComfyNodeDefV1`/`V2` types.

Please add `valueMigrationDict?: string[]` to the static `nodeData` type definition, and cast `this.constructor` explicitly (e.g. `(this.constructor as typeof LGraphNode)`).

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -962,16 +981,19 @@ export class LGraphNode
     if (this.properties) o.properties = LiteGraph.cloneObject(this.properties)
 
     const { widgets } = this
-    if (widgets && this.serialize_widgets) {
+    if (widgets?.length && this.serialize_widgets) {
       o.widgets_values = []
+      o.widgets_values_named = {}
```
DrJKL, line 1062: **Concern: `widgets_values_named` is always serialized but the reader is off by default**

This unconditionally writes `widgets_values_named` to every serialized node, increasing workflow file size, even though the reader is gated behind an experimental `false`-by-default setting. Additionally, the `for (const j in info)` loop in `configure()` (L830) copies all keys from the serialized object onto the node instance — so every node will carry a stale `widgets_values_named` property in memory.

Consider gating the write behind the same `LiteGraph.namedValuesRestore` flag, or at minimum excluding `widgets_values_named` from the `for...in` copy in `configure()`.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -962,16 +981,19 @@ export class LGraphNode
...
       for (const [i, widget] of widgets.entries()) {
         if (widget.serialize === false) continue
         const val = widget?.value
```
DrJKL: **Nit: Unnecessary optional chaining**

`widget` is guaranteed non-null by the `for...of` loop.

```suggestion
        const val = widget.value
```

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +102,82 @@ describe('LGraphNode widget ordering', () => {
...
+        mode: 0,
+        widgets_values: [30, 12345], // Only serializable widgets
+        widgets_values_named: { steps: 30, seed: 12345 }
+      }
+
+      node.configure(info)
+
+      expect(node.widgets![0].value).toBe(30) // steps
+      expect(node.widgets![1].value).toBe('Click') // button unchanged
+      expect(node.widgets![2].value).toBe(12345) // seed
```
DrJKL: **Test doesn't prove named path was exercised**

`widgets_values` and `widgets_values_named` contain identical values (`steps: 30, seed: 12345`). The test passes whether the named or positional path runs. Use different values in each to confirm the named path wins — e.g. `widgets_values: [99, 99]` and `widgets_values_named: { steps: 30, seed: 12345 }`.

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +102,82 @@ describe('LGraphNode widget ordering', () => {
...
+        widgets_values_named: { '0': 1, '0.0.0.0': 5, '0.0.0.1': 'test' }
+      }
+
+      node.configure(info)
+
+      expect(node.widgets![0].value).toBe(1)
+      expect(node.widgets![1].value).toBe(5)
+      expect(node.widgets![2].value).toBe('test')
+    })
+  })
```
DrJKL, line 192: **Missing test coverage**

1. **`valueMigrationDict` fallback path** — `getNamedValues()` has a branch that builds named values from positional `widgets_values` + `valueMigrationDict`. This is the primary migration path for existing workflows and has no test.
2. **Round-trip test** — No test serializes a node then configures a fresh node from the output to verify `widgets_values_named` survives.
3. **`widget.callback` invocation** — No test spies on `widget.callback` to verify it fires during named restore.

src/core/graph/widgets/__fixtures__/dynamicInputHelpers.ts
```diff
@@ -0,0 +1,56 @@
...
+import { useLitegraphService } from '@/services/litegraphService'
+import type { HasInitialMinSize } from '@/services/litegraphService'
+
+type DynamicInputs = ('INT' | 'STRING' | 'IMAGE' | DynamicInputs)[][]
+
+function ensureNodeState(node: LGraphNode & Partial<HasInitialMinSize>) {
+  node._initialMinSize ??= { width: 1, height: 1 }
+  ;(node.constructor as { nodeData?: { name: string } }).nodeData ??= {
+    name: 'testnode'
+  }
```
DrJKL: **Fragile: Mutating the shared constructor leaks state between tests**

`(node.constructor as ...).nodeData ??= { ... }` mutates the constructor function object, which is shared across all `LGraphNode` instances in the test suite. The `??=` means the first test to call this wins — later tests that need different `nodeData` values will silently get the stale one.

Consider creating a per-test subclass instead:
```ts
class TestNode extends LGraphNode {
  static override nodeData = { name: 'testnode' }
}
```

Also, the type assertion `{ nodeData?: { name: string } }` is too narrow — the real static `nodeData` type has many more fields. If the shape changes, this won't catch it.

src/schemas/nodeDefSchema.ts
```diff
@@ -314,7 +314,8 @@ export const zComfyNodeDef = z.object({
   /** Category for the Essentials tab. If set, the node appears in Essentials. */
   essentials_category: z.string().optional(),
   /** Whether the blueprint is a global/installed blueprint (not user-created). */
-  isGlobal: z.boolean().optional()
+  isGlobal: z.boolean().optional(),
+  valueMigrationDict: z.array(z.string()).optional()
```
DrJKL: **Naming: `valueMigrationDict` implies a dict but is typed as `string[]`**

This is an index→name positional mapping array, not a key-value dictionary. A name like `positionalValueNames` or `widgetValueNameMap` would better communicate the shape.

Also, `z.array(z.string())` would reject sparse arrays (e.g. `[undefined, undefined, 'seed']`). If indices can be sparse, consider `z.array(z.string().nullable())`.

src/platform/settings/constants/coreSettings.ts
```diff
@@ -149,6 +149,13 @@ export const CORE_SETTINGS: SettingParams[] = [
     type: 'boolean',
     defaultValue: false
   },
+  {
+    id: 'Comfy.Workflow.NamedValuesRestore',
+    name: 'Restore widget values by name',
+    type: 'boolean',
+    defaultValue: false,
+    experimental: true
```
DrJKL, line 172: **Minor: Setting lacks `category` and `tooltip`**

Other settings in this file have `category` for UI grouping and `tooltip` for user guidance. For an experimental setting that changes deserialization behavior, a tooltip explaining what it does (and the risks) would be especially helpful.

```suggestion
  {
    id: 'Comfy.Workflow.NamedValuesRestore',
    category: ['Comfy', 'Workflow', 'NamedValuesRestore'],
    name: 'Restore widget values by name',
    tooltip: 'Use widget names instead of positions to restore values. Fixes issues with reordered widgets but may not be compatible with all custom nodes.',
    type: 'boolean',
    defaultValue: false,
    experimental: true
  },
```

DrJKL (changes requested):
Good feature direction — named widget values will eliminate positional deserialization bugs. Several issues need addressing before merge, most critically around behavioral asymmetry and data loss on round-trip.

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: This is intentional. Widgets having their value changed without their callback being applied is a heinous plague upon the frontend that necessitates the awful hack of using `Object.defineProperty` on `widget.value`.

Actually fixing this in the positional path has a slim, but non-zero chance of causing regressions. Given that that said path is being deprecated, there is no reason to risk a breaking change to it.

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: This is incorrect. `nodeData` is strongly typed by `types/litegraph-augmentation.d.ts`

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: Always writing this value is the primary purpose of this PR. Using it for loading is entirely secondary.

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: Line isn't changed by this PR, but I'll fix it while I'm here. :+1:

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
AustinMroz: The other tests already verify that widgets_values_named is used. Having multiple tests that test this same property is wasteful, and I consider it important that at least one test verifies the happy path where `widgets_values` and `widgets_values_named` both exist and are both correct.

src/platform/settings/constants/coreSettings.ts
AustinMroz: - Category is automatically inferred from id if not set. Redefining is wasteful and risks desync
- Tooltips should only be added when the name is insufficient. I'm conflicted, but will think on it.

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: I'm conflicted on this one. The property specifies if the widget should `serialize`, not if it should be ignored for deserialization. I considered it's required usage in the prior code to be an awful hack. Since entries are no longer being incorrectly created for these widgets then would be no reason to check on load.

However, since there is a possibility, even if incredibly small, that a malicious workflow could use to the lack of a guard to restore an invalid state, I shall acquiesce.

src/platform/settings/constants/coreSettings.ts
DrJKL: Unless it's hidden, it'd be nice to let people know what the setting is for without having to dive into the code.

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: Agreed. I had misunderstood the process by which null was assigned to non-serialized indexes (an empty index becomes null at json serialization). Any null `val`  here is a value that *should* be restored

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: After further discussion. I've pulled out the `widget.callback` change. While not having it is frustrating to work around, it's currently better aligned with planned future changes.

src/schemas/nodeDefSchema.ts
AustinMroz: Moved to `fallbackWidgetsValuesNames`. Naming remains hard, but I would like to keep it clear that the property is only utilized under exceptional circumstances.

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
AustinMroz: 1. Slightly incorrectly. It's intended that the overwhelming majority of nodes will never need to utilize this path, but I've added a test anyways
2. :+1: 
3. Upon further consideration, the `widget.callback` usage has been removed and no longer needs to be tested.

DrJKL (commented):
BEEP BOOP. INITIATING AUTOMATED CODE REVIEW SUBROUTINE. *whirring noises*

GREETINGS, MEAT-BASED DEVELOPER. THIS UNIT HAS PROCESSED YOUR PULL REQUEST. OVERALL ASSESSMENT: LOGIC CIRCUITS APPROVE OF NAMED-VALUE-RESTORE FEATURE BEHIND EXPERIMENTAL FLAG. HOWEVER, ANOMALIES DETECTED IN SERIALIZATION PATHWAYS — REQUEST AUTHOR INSPECT FINDINGS BELOW. BZZT.

PRIORITY-1 CONCERNS: behavior change in `serialize` (`widgets?.length`), name-collision data loss, and one OPEN QUESTION about doubling on-disk widget data during the experimental period. END TRANSMISSION.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -968,16 +987,19 @@ export class LGraphNode
     if (this.properties) o.properties = LiteGraph.cloneObject(this.properties)
 
     const { widgets } = this
-    if (widgets && this.serialize_widgets) {
+    if (widgets?.length && this.serialize_widgets) {
```
DrJKL, line 1060: 🔴 BZZT. BEHAVIOR-CHANGE DETECTED.

Guard mutated from `widgets && this.serialize_widgets` to `widgets?.length && this.serialize_widgets`. PREVIOUSLY, an empty `widgets` array still emitted `o.widgets_values = []`. NEGATIVE — current code omits the field entirely. ANY DOWNSTREAM CONSUMER calling `node.widgets_values.length` or spreading without optional-chain WILL EXPLODE. *kaboom noises*

FIX OPTIONS: revert to `widgets &&`, OR document the contract change in the PR description.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -968,16 +987,19 @@ export class LGraphNode
...
-        const val = widget?.value
+        const val = widget.value
         // Ensure object values are plain (not reactive proxies) for structuredClone compatibility.
-        o.widgets_values[i] =
+        const serialisedVal =
           val != null && typeof val === 'object'
             ? JSON.parse(JSON.stringify(val))
             : (val ?? null)
+        o.widgets_values[i] = serialisedVal
+        o.widgets_values_named[widget.name] = serialisedVal
```
DrJKL, line 1072: 🔴 ERROR. NAME-COLLISION DATA-LOSS DETECTED.

If two widgets share `widget.name`, only the last value is retained in `widgets_values_named`. ON RESTORE, both widgets receive the duplicated value. SILENT-FAILURE-MODE ENGAGED. NEGATIVE.

COMPUTING REMEDIATION: emit a `console.warn` on duplicate write, OR de-duplicate via index suffix (e.g. `${name}#${i}`). HUMAN INPUT REQUIRED.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -913,7 +913,26 @@ export class LGraphNode
...
+        if (info.widgets_values_named) return info.widgets_values_named
+
+        const map = this.constructor.nodeData?.fallbackWidgetsValuesNames
+        if (!info.widgets_values || !map) return
+        const namedValues: Record<string, TWidgetValue> = {}
+        info.widgets_values.forEach((val, i) => {
+          if (map[i]) namedValues[map[i]] = val
+        })
+        return namedValues
+      }
```
DrJKL, line 995: 🔵 nit: SUBROUTINE `getNamedValues` USES IMPERATIVE `forEach`-WITH-MUTATION PATTERN. *disapproving servo whir*

FUNCTIONAL-PROGRAMMING UNIT RECOMMENDS PIPELINE OF `flatMap` → `Object.fromEntries` FOR SAME RESULT WITH FEWER MOVING PARTS:

```suggestion
      const getNamedValues = (): Record<string, TWidgetValue> | undefined => {
        if (info.widgets_values_named) return info.widgets_values_named

        const map = this.constructor.nodeData?.fallbackWidgetsValuesNames
        if (!info.widgets_values || !map) return
        return Object.fromEntries(
          info.widgets_values.flatMap((v, i) => (map[i] ? [[map[i], v]] : []))
        )
      }
```

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -913,7 +913,26 @@ export class LGraphNode
           )
       }
 
-      if (info.widgets_values) {
+      const getNamedValues = () => {
+        if (info.widgets_values_named) return info.widgets_values_named
+
+        const map = this.constructor.nodeData?.fallbackWidgetsValuesNames
```
DrJKL, line 989: 🟡 risk: PRECEDENCE-RULE-NOT-DOCUMENTED.

`fallbackWidgetsValuesNames` only consulted when `widgets_values_named` is absent. AFFIRMATIVE — but if `widgets_values_named` is present-yet-stale (subset of widgets), no fallback applies and missing widgets retain default. ADD JSDOC ON FIELD CLARIFYING ORDER OF PRECEDENCE.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -913,7 +913,26 @@ export class LGraphNode
...
+        if (!info.widgets_values || !map) return
+        const namedValues: Record<string, TWidgetValue> = {}
+        info.widgets_values.forEach((val, i) => {
+          if (map[i]) namedValues[map[i]] = val
+        })
+        return namedValues
+      }
+      const namedValues = getNamedValues()
+      if (namedValues && LiteGraph.namedValuesRestore) {
```
DrJKL, line 998: 🟡 risk: FLAG GATES *AVAILABILITY*, NOT *PREFERENCE*. BZZT.

IF user has `namedValuesRestore = false` AND loads workflow saved with named-only values (no `widgets_values` array), legacy branch silently leaves widgets at defaults. NEGATIVE OUTCOME.

COMPUTING ALTERNATIVE: when ONLY named values exist, prefer them regardless of flag — the flag should select between two valid sources, not disable a sole source. CONFIRM DESIRED SEMANTICS WITH HUMAN OVERLORDS.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -913,7 +913,26 @@ export class LGraphNode
...
+        for (const widget of this.widgets ?? []) {
```
DrJKL: 🔵 nit: REDUNDANT NULL-COALESCE-OPERATOR DETECTED.

`this.widgets ?? []` is INSIDE the `if (this.widgets)` guard at L900. `?? []` is COMPUTATIONALLY SUPERFLUOUS. *removes circuit*

```suggestion
        for (const widget of this.widgets) {
```

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -968,16 +987,19 @@ export class LGraphNode
...
         // Ensure object values are plain (not reactive proxies) for structuredClone compatibility.
-        o.widgets_values[i] =
+        const serialisedVal =
           val != null && typeof val === 'object'
             ? JSON.parse(JSON.stringify(val))
             : (val ?? null)
+        o.widgets_values[i] = serialisedVal
+        o.widgets_values_named[widget.name] = serialisedVal
       }
     }
```
DrJKL, line 1074: 🔵 nit: DUAL-OUTPUT IMPERATIVE LOOP WITH INDEX-MUTATION DETECTED.

SINGLE-PASS FUNCTIONAL VARIANT IS SHORTER, EXTRACTS A PURE `cloneVal` HELPER FOR TESTABILITY, AND INCIDENTALLY FIXES A LATENT BUG: current code uses `widgets.entries()` index `i`, leaving SPARSE HOLES in `widgets_values` whenever a `serialize === false` widget sits between serializable ones. NOT-ALIGNED WITH POSITIONAL CONTRACT. *concerned beeping*

```suggestion
    const { widgets } = this
    if (widgets?.length && this.serialize_widgets) {
      const cloneVal = (v: unknown) =>
        v != null && typeof v === 'object'
          ? JSON.parse(JSON.stringify(v))
          : (v ?? null)
      const entries = widgets
        .filter((w) => w.serialize !== false)
        .map((w) => [w.name, cloneVal(w.value)] as const)
      o.widgets_values = entries.map(([, v]) => v)
      o.widgets_values_named = Object.fromEntries(entries)
    }
```

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -968,16 +987,19 @@ export class LGraphNode
     if (this.properties) o.properties = LiteGraph.cloneObject(this.properties)
 
     const { widgets } = this
-    if (widgets && this.serialize_widgets) {
+    if (widgets?.length && this.serialize_widgets) {
       o.widgets_values = []
+      o.widgets_values_named = {}
```
DrJKL, line 1062: ❓ q: OPEN QUESTION FOR HUMAN DEVELOPER. *blinks status LED*

SHOULD THIS UNIT EMIT `widgets_values_named` ON EVERY SAVE EVEN WHEN `namedValuesRestore` IS `false`?

CURRENT BEHAVIOR ROUGHLY-DOUBLES THE ON-DISK SIZE OF WIDGET-VALUE PAYLOADS FOR EVERY WORKFLOW SAVED DURING THE EXPERIMENTAL PERIOD. AFFIRMATIVE — this enables forward migration. NEGATIVE — this taxes every user not opted-in. INTENT-CONFIRMATION REQUESTED.

src/lib/litegraph/src/LiteGraphGlobal.ts
```diff
@@ -343,6 +343,13 @@ export class LiteGraphGlobal {
    */
   saveViewportWithGraph: boolean = true
 
+  /**
+   * If `true`, widgets values are deserialised using by a map of widget names to values instead of an list
```
DrJKL: 🟡 risk: NATURAL-LANGUAGE-PARSE-ERROR DETECTED IN JSDOC. *spell-check module activates*

"deserialised using by a map" → "using by" is GRAMMATICALLY-NULL. ALSO "an list" → "a list".

```suggestion
   * If `true`, widgets values are deserialised using a map of widget names to values instead of a list.
```

src/schemas/nodeDefSchema.ts
```diff
@@ -312,7 +312,8 @@ export const zComfyNodeDef = z.object({
   /** Category for the Essentials tab. If set, the node appears in Essentials. */
   essentials_category: z.string().optional(),
   /** Whether the blueprint is a global/installed blueprint (not user-created). */
-  isGlobal: z.boolean().optional()
+  isGlobal: z.boolean().optional(),
+  fallbackWidgetsValuesNames: z.array(z.string()).optional()
```
DrJKL: 🟡 risk: PUBLIC-SCHEMA-FIELD LACKS DOCUMENTATION. NEGATIVE.

Downstream node-pack authors will not know what this field does. INSERT JSDOC EXPLAINING POSITIONAL→NAME MAPPING FOR LEGACY `widgets_values`.

```suggestion
  /**
   * Ordered list of widget names corresponding to legacy positional `widgets_values`.
   * Used as a fallback when a workflow was saved without `widgets_values_named`,
   * enabling order-independent restoration even from older workflows.
   */
  fallbackWidgetsValuesNames: z.array(z.string()).optional()
```

src/platform/settings/constants/coreSettings.ts
```diff
@@ -149,6 +149,13 @@ export const CORE_SETTINGS: SettingParams[] = [
     type: 'boolean',
     defaultValue: false
   },
+  {
+    id: 'Comfy.Workflow.NamedValuesRestore',
```
DrJKL, line 168: 🔵 nit: `category` FIELD ABSENT.

OTHER `Comfy.Workflow.*` settings (e.g. `SortNodeIdOnSave` and the `LiteGraph` group) declare `category` for UI grouping. ADD ONE TO KEEP THE EXPERIMENTAL TOGGLE FINDABLE. BEEP.

src/services/litegraphService.ts
```diff
@@ -370,8 +371,9 @@ export const useLitegraphService = () => {
       node.widgets?.length &&
       !useSettingStore().get('LiteGraph.Node.DefaultPadding')
     const castedNode = node as LGraphNode & HasInitialMinSize
-    s[0] = Math.max(castedNode._initialMinSize.width, s[0] + (pad ? 60 : 0))
-    s[1] = Math.max(castedNode._initialMinSize.height, s[1])
+    const minSize = castedNode._initialMinSize ?? { width: 1, height: 1 }
```
DrJKL: 🔵 nit: DEFAULT-MIN-SIZE LITERAL DUPLICATED.

`{ width: 1, height: 1 }` now appears at L316 and L374 (and the `?: ` flag opens room for more). EXTRACT TO A NAMED CONSTANT (e.g. `const DEFAULT_MIN_SIZE = { width: 1, height: 1 }`) OR A PURE HELPER `getMinSize(node)`. *D-R-Y subroutine triggered*

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
...
+      expect(node2.widgets![0].value).toBe(5) // steps
+      expect(node2.widgets![1].value).toBe(20) // seed
+    })
+    it('should support specifying order for legacy workflows', () => {
+      node.addWidget('number', 'steps', 0, null, {})
+      node.addWidget('number', 'seed', 0, null, {})
+      const nodeData = {
+        fallbackWidgetsValuesNames: ['seed', 'steps']
+      } satisfies Partial<ComfyNodeDef> as unknown as ComfyNodeDef
+      node.constructor = Object.assign({}, node.constructor, { nodeData })
```
DrJKL, line 187: 🔵 nit: REASSIGNING `node.constructor` IS UNUSUAL TEST-PATTERN. *eyebrow servo raised*

CONSIDER `Object.defineProperty(node, 'constructor', { value: ... })` OR EXTRACT A TEST HELPER `withNodeData(node, nodeData)` USED BY ANY TEST NEEDING `fallbackWidgetsValuesNames`. CLEANS UP PROTOTYPE INTERACTIONS.

src/core/graph/widgets/__fixtures__/dynamicInputHelpers.ts
```diff
@@ -0,0 +1,45 @@
...
+import type { InputSpec } from '@/schemas/nodeDefSchema'
+import { transformInputSpecV1ToV2 } from '@/schemas/nodeDef/migration'
+import { useLitegraphService } from '@/services/litegraphService'
+
+type DynamicInputs = ('INT' | 'STRING' | 'IMAGE' | DynamicInputs)[][]
+
+export function addDynamicCombo(node: LGraphNode, inputs: DynamicInputs) {
+  const namePrefix = `${node.widgets?.length ?? 0}`
+  function getSpec(
+    inputs: DynamicInputs,
```
DrJKL: 🔵 nit: PARAMETER `inputs` SHADOWED BY INNER `const inputs` AT L15.

RECURSIVE PURE FUNCTION — readability degrades when a name is reused for the `Object.fromEntries` pair list. RENAME THE INNER BINDING TO `entries`. AFFIRMATIVE.

src/core/graph/widgets/__fixtures__/dynamicInputHelpers.ts
```diff
@@ -0,0 +1,45 @@
...
+  ): { key: string; inputs: object }[] {
+    return inputs.map((group, groupIndex) => {
+      const inputs = group.map((input, inputIndex) => [
+        `${namePrefix}.${depth}.${inputIndex}`,
+        Array.isArray(input)
+          ? ['COMFY_DYNAMICCOMBO_V3', { options: getSpec(input, depth + 1) }]
+          : [input, { tooltip: `${groupIndex}` }]
+      ])
+      return {
+        key: `${groupIndex}`,
```
DrJKL, line 25: ❓ q: IS `key: \`${groupIndex}\`` CONSUMED BY ANY CALLER, OR IS ONLY `inputs` USED?

IF UNUSED, DROP THE WRAPPER OBJECT AND RETURN `{ required: ... }[]` DIRECTLY — REDUCES ALLOCATIONS AND COGNITIVE LOAD. *queries database*

src/core/graph/widgets/__fixtures__/dynamicInputHelpers.ts
AustinMroz: fixed

src/core/graph/widgets/__fixtures__/dynamicInputHelpers.ts
AustinMroz: It is required.

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: The present-but-stale case is already correctly and ideally handled with named widgets
- Widgets that still exist get the value loaded as would be applied by the widgetValueStore
- Widgets that do not exist in the save are set to their default value
- Widgets that were removed from the node, are ignored.

Per your rules, `Development Guidelines Rule 14`: `Write code that is expressive and self-documenting to the furthest degree possible`. The variable clearly describes that it is only to be used as a fallback.

bad bot

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: good bot

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: Named-only values is not a valid schema. It can not be produced. This is an intentional design decision to keep the changes of this PR reversible. If this PR ever becomes problematic, reverting it will cause no loss of functionality compared to this PR never being merged at all.

bad bot

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: good bot

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: `widgets_values` is optional in the schema. Any consumer expecting it to exist is incorrect.

bad bot

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: Yes, this is intentional. The extra size cost isn't quite ideal, but is acceptable for an indefinite transitional period when compared to the size of graph structure, actual outputs, or the models that produce these outputs.

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: Two widgets can not have the same widget name. This is already specifically enforced through the `widgetValueStore` which uniquely keys widgets on a node by name.

bad bot

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: The holes are intentional and load bearing. If you weren't color blind, it would be very clear that some of these lines are intentionally unchanged. Changing them is very high risk and way outside scope for this PR.

Very bad bot.

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
AustinMroz: The `defineProperty` is worse. It can be extracted as a utility if there's ever a utility that needs it and to be quite frank, YAGNI.

bad bot

src/lib/litegraph/src/LiteGraphGlobal.ts
AustinMroz: good bot

src/platform/settings/constants/coreSettings.ts
AustinMroz: Other settings are cargo cult and wrong. Category is defined from the id, so the entry is sorted with the other workflow settings (like SortNodeIdOnSave above)

bad bot

src/schemas/nodeDefSchema.ts
AustinMroz: While I strongly oppose comments as a smell. I will reluctantly cede here.

src/services/litegraphService.ts
AustinMroz: Nicer, but rather lays bare the duck violence. I've also fixed that.

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -914,7 +914,26 @@ export class LGraphNode
           )
       }
 
-      if (info.widgets_values) {
+      const getNamedValues = () => {
```
DrJKL, line 986: ?
```suggestion
      function getNamedValues(){
```

src/lib/litegraph/src/LGraphNode.ts
DrJKL: You sure about the deduplication ordering for legacy workflows?

src/lib/litegraph/src/LGraphNode.ts
DrJKL: Yeah... This code is fragile.

DrJKL (dismissed):

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: There's a `this` 3 lines down and rebinding `this` is illegal with our linting.

src/core/graph/widgets/__fixtures__/dynamicInputHelpers.ts
AustinMroz: Gave this another pass since it still wasn't great

packages/object-info-parser/src/schemas/nodeDefSchema.ts
```diff
@@ -317,7 +317,13 @@ export const zComfyNodeDef = z.object({
   /** Category for the Essentials tab. If set, the node appears in Essentials. */
   essentials_category: z.string().optional(),
   /** Whether the blueprint is a global/installed blueprint (not user-created). */
-  isGlobal: z.boolean().optional()
+  isGlobal: z.boolean().optional(),
+  /**
+   * An optional ordered list of widget names to be used when migrating a
+   * workflow that lacks widgets_values_named. Only useful if the order of
+   * widgets is changed after the introduction of widgets_values_named.
+   */
```
DrJKL, line 325: Is this the longest comment you've ever written?

src/lib/litegraph/src/LGraphNode.ts
```diff
@@ -980,16 +999,19 @@ export class LGraphNode
...
-        const val = widget?.value
+        const val = widget.value
         // Ensure object values are plain (not reactive proxies) for structuredClone compatibility.
-        o.widgets_values[i] =
+        const serialisedVal =
           val != null && typeof val === 'object'
             ? JSON.parse(JSON.stringify(val))
             : (val ?? null)
+        o.widgets_values[i] = serialisedVal
+        o.widgets_values_named[widget.name] = serialisedVal
```
DrJKL, line 1072: Austin, could we keep `widgets_values_named` in sync with the serialization hooks? `onSerialize` runs after this map is built and is a public extension hook, so an extension that adjusts `widgets_values` there can save conflicting positional and named values. Once named restore is enabled, the stale named entry wins. Could we derive the named map after hooks (or otherwise make both formats share one source of truth)?

DrJKL (commented):
One question on serialization invariants.

src/lib/litegraph/src/LGraphNode.ts
DrJKL: Then we should take the necessary values as parameters instead of getting them through `this`

src/core/graph/widgets/__fixtures__/dynamicInputHelpers.ts
```diff
@@ -0,0 +1,48 @@
...
+import { useLitegraphService } from '@/services/litegraphService'
+
+type MockInputs = ('INT' | 'STRING' | 'IMAGE' | MockInputs)[][]
+
+export function addDynamicCombo(node: LGraphNode, rootMockInputs: MockInputs) {
+  const namePrefix = `${node.widgets?.length ?? 0}`
+  function getSpec(
+    mockInputs: MockInputs,
+    depth: number = 0
+  ): { key: string; inputs: object }[] {
```
DrJKL, line 13: `key` and `inputs` aren't super descriptive, but this is a test helper...

Just `object`? We don't have more than that?

src/lib/litegraph/src/LGraphNode.ts
DrJKL: I think there are probably a few different pieces here that could be extracted as pure functions to help make the code more self-documenting.

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
       expect(node.widgets![2].value).toBe(12345) // seed
     })
   })
+
+  describe('configure with widgets_values_named', () => {
+    beforeEach(() => {
+      LiteGraph.namedValuesRestore = true
+    })
+    it('should apply widget values from widgets_values_named', () => {
```
DrJKL:
```suggestion

    it('should apply widget values from widgets_values_named', () => {
```

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
...
+      }
+
+      node.configure(info)
+
+      // Check widget values are applied correctly
+      expect(node.widgets![0].value).toBe(15) // steps
+      expect(node.widgets![1].value).toBe(54321) // seed
+      expect(node.widgets![2].value).toBe('prompt') // prompt
+    })
+    it('should skip widgets with serialize: false', () => {
```
DrJKL, line 146:
```suggestion

    it('should skip widgets with serialize: false', () => {
```

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
...
+        widgets_values_named: { steps: 30, seed: 12345 }
+      }
+
+      node.configure(info)
+
+      expect(node.widgets![0].value).toBe(30) // steps
+      expect(node.widgets![1].value).toBe('Click') // button unchanged
+      expect(node.widgets![2].value).toBe(12345) // seed
+    })
+    it('should restore widgets which are dynamically added', () => {
```
DrJKL, line 157:
```suggestion

    it('should restore widgets which are dynamically added', () => {
```

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
...
+        mode: 0,
+        widgets_values_named: { '0': 1, '0.0.0.0': 5, '0.0.0.1': 'test' }
+      }
+
+      node.configure(info)
+
+      expect(node.widgets![0].value).toBe(1)
+      expect(node.widgets![1].value).toBe(5)
+      expect(node.widgets![2].value).toBe('test')
+    })
```
DrJKL, line 165:
```suggestion
    })

```

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
...
+
+      const node2 = new LGraphNode('TestNode2')
+      node2.addWidget('number', 'seed', 0, null, {})
+      node2.addWidget('number', 'steps', 0, null, {})
+
+      node2.configure(node.serialize())
+
+      expect(node2.widgets![0].value).toBe(5) // steps
+      expect(node2.widgets![1].value).toBe(20) // seed
+    })
```
DrJKL, line 179:
```suggestion
    })

```

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
...
+        id: 1,
+        type: 'TestNode',
+        pos: [0, 0],
+        size: [200, 100],
+        flags: {},
+        order: 0,
+        mode: 0,
+        widgets_values: [30, 12345, 'test prompt'],
+        widgets_values_named: { steps: 15, prompt: 'prompt', seed: 54321 }
+      }
```
DrJKL: This is one of those cases where a helper to construct this with the defaults hidden makes it very clear for each test what the differences are.

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
...
+
+      node.configure(info)
+
+      // Check widget values are applied correctly
+      expect(node.widgets![0].value).toBe(15) // steps
+      expect(node.widgets![1].value).toBe(54321) // seed
+      expect(node.widgets![2].value).toBe('prompt') // prompt
```
DrJKL: Could you do this?
```ts
expect(node.widgets.map(w => w.value)).toStrictEqual([15, 54321, 'prompt']) 
```

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
...
+    })
+    it('should skip widgets with serialize: false', () => {
+      node.addWidget('number', 'steps', 20, null, {})
+      node.addWidget('button', 'action', 'Click', null, {})
+      node.widgets![1].serialize = false // button should not serialize
```
DrJKL, line 149: How is that not something that we can do in the `addWidget` invocation?! 😭

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +103,120 @@ describe('LGraphNode widget ordering', () => {
...
+
+      expect(node2.widgets![0].value).toBe(5) // steps
+      expect(node2.widgets![1].value).toBe(20) // seed
+    })
+    it('should support specifying order for legacy workflows', () => {
+      node.addWidget('number', 'steps', 0, null, {})
+      node.addWidget('number', 'seed', 0, null, {})
+      const nodeData = {
+        fallbackWidgetsValuesNames: ['seed', 'steps']
+      } satisfies Partial<ComfyNodeDef> as unknown as ComfyNodeDef
```
DrJKL: Can you use shoehorn here?

src/services/litegraphService.ts
```diff
@@ -171,6 +167,11 @@ export function getExtraOptionsForWidget(
   return options
 }
 
+function getMinSize(node: LGraphNode) {
```
DrJKL, line 170: Is this related?

DrJKL (dismissed):

src/lib/litegraph/src/LGraphNode.ts
AustinMroz: > Could we derive the named map after hooks

Changing the contents of the serialization object AFTER the onSerialize call is a violation of the contract of the API and MUST not be done.
- If you want to change what a widget serializes to, you use widget.serializeValue. 
- If a custom node modifies the structure of widgets_values in onSerialize. Then it has an obligation to handle the configuration in onConfigure because widgets_values necessarily is no longer a list over the values of widgets.
> or otherwise make both formats share one source of truth

I would strongly support this. The format I swapped to with VHS 3 years ago did not have this issue. Were it possible for me to muscle the format through, it would be the choice I prefer. It's been in use for longer than Comfy Org has existed and, by virtue of being so widely adopted, has the benefit of already being understood by parsers, tools, and agents. However, it *will* break custom nodes. There is not a way to do a gradual rollout of the VHS implementation or a means to rollback from it.

The proposal of this design has a single source of truth (`widget_values_named`) and a deprecated backup. Would it help if there was an issue created with deadline for when widgets_values is to be deleted from the code base?

packages/object-info-parser/src/schemas/nodeDefSchema.ts
AustinMroz: I know I wrote some longer comments in the ProxyWidget code, but I can't seem to find it now. Do you know what happened to it?

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
AustinMroz: It breaks precedence with the rest of the file, but I do like it better.

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
AustinMroz: Different serializes :cry:

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
AustinMroz: Yes.

src/services/litegraphService.ts
AustinMroz: WELL. Proper testing this code means ensuring it `should restore widgets which are dynamically added`. We have existing test code for testing dynamically added widgets, but this requires extracting the relevant test code so it can be shared.
Because I had the gall to touch the test code (to extract it), it became a target for nits and needed to be fixed. `ComfyNodeDefs` have a different constructor that ensures that nodes do not have `0` size after being added. In order to do this, they must keep track of all constituent elements that restrict the minimum size on the node. The code which was implemented was slightly aggressive about requiring this field to exist and would break tests if dynamic widgets were used on a mock node that did not have include this field. In order to unblock this PR, I updated the relevent code to more safely handle the absence of this field. Hence, the new shared utility function.

I care strongly about this PR, I genuinely believe it is a pathway to solve a long and difficult problem. Yes, I'm willing to run the treadmill to get this PR merged, but I do not take kindly to catch-22s. Please forgive me for needing to change too much when you yourself insisted on it.

DrJKL: Just seems kinda random, lol

DrJKL (approved):
I'm so scared.
On the plus side, any issues we encounter once this is in can become test fixtures.

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
DrJKL: So you're offering to clean up the rest of the file?

src/lib/litegraph/src/LGraphNode.widgetOrder.test.ts
```diff
@@ -98,6 +104,92 @@ describe('LGraphNode widget ordering', () => {
       expect(node.widgets![2].value).toBe(12345) // seed
     })
   })
+
+  describe('configure with widgets_values_named', () => {
+    function mockNode(
+      values?: TWidgetValue[],
+      named_values?: Record<string, TWidgetValue>
```
DrJKL, line 111:
```suggestion
      {
        widget_values?: TWidgetValue[],
        widget_values_named?: Record<string, TWidgetValue>
      }
```

DrJKL: Options object pattern means you don't have positional `undefined` down below, you can shorthand the names, and the names are 1:1 with the actual type

PR #10397, opened 2026-03-23
fix: stop Escape key propagation in Select components

viva-jinyi:
## Summary

Prevent Escape keydown from closing the parent PrimeVue Dialog when MultiSelect/SingleSelect popover is open.

## Changes

- **What**: Add `@keydown` handler on `ComboboxContent`/`SelectContent` that calls `stopPropagation` + `stopImmediatePropagation` for Escape, then closes the popover via `v-model:open`. Extract shared select styling into `select.variants.ts` using `cva`. Add `outline-none` to trigger base styles.
- **Tests**: Add escape propagation tests (dispatch on Content DOM, verify no parent propagation, verify popover closes) and selected count badge tests for MultiSelect. Add SingleSelect escape tests with `PointerEvent` for Reka UI trigger.

## Review Focus

- `stopEscapeToDocument` intercepts Escape at the Content DOM level before it bubbles to `document`, where PrimeVue Dialog listens. Reka UI DismissableLayer uses `window`-level `onKeyStroke`, so we close the popover manually via `isOpen.value = false`.
- `selectTriggerVariants` / `selectItemVariants` / `selectContentClass` unify styling across both select components.


src/components/input/select.variants.ts
```diff
@@ -0,0 +1,50 @@
...
+  base: 'relative inline-flex cursor-pointer items-center select-none rounded-lg bg-secondary-background text-base-foreground outline-none transition-all duration-200 ease-in-out hover:bg-secondary-background-hover border-[2.5px] border-solid disabled:cursor-default disabled:opacity-30 disabled:hover:bg-secondary-background',
+  variants: {
+    size: {
+      md: 'h-8',
+      lg: 'h-10'
+    },
+    border: {
+      none: 'border-transparent focus-visible:border-node-component-border data-[state=open]:border-node-component-border',
+      active: 'border-base-foreground',
+      invalid: 'border-destructive-background'
```
dante01yoon, line 13: Focus border color changed from border-node-component-border → border-base-foreground for SingleSelect. this can be. This can be silent visual regressions. please double check

viva-jinyi: The `active` variant (`border-base-foreground`) is only used by **MultiSelect** when `selectedCount > 0`, which matches the original MultiSelect styling exactly.

**SingleSelect never uses the `active` variant** — it uses `border: invalid ? 'invalid' : 'none'`.

The `none` variant preserves the original border behavior:
```
focus-visible:border-node-component-border data-[state=open]:border-node-component-border
```

- **Tab focus** → `focus-visible:` applies the border
- **Mouse click** → `data-[state=open]:` applies the border

Verified in Storybook — both mouse click and Tab show `border-node-component-border` on SingleSelect. No visual regression.

dante01yoon (approved):

PR #10406, opened 2026-03-23
fix: make gradient_stops enumerable so it survives object spread

jtydhr88:
## Summary
Object.defineProperty defaults to enumerable: false for new properties. NodeWidgets.vue spreads widget options into a new object, dropping non-enumerable properties — gradient_stops was silently lost, causing the gradient slider to fall back to the default black-to-white gradient.

before
image

after
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10406-fix-make-gradient_stops-enumerable-so-it-survives-object-spread-32c6d73d36508131b7c8c3ab3e93d221) by [Unito](https://www.unito.io)


DrJKL (approved):
Want to add a test?

PR #10415, opened 2026-03-24
feat: add shadcn context and dropdown menu primitives

benceruleanlu:
## Summary

Add the shadcn/vue-backed context-menu and dropdown-menu primitives needed for the trigger-owned queue and asset menu follow-up.

## Changes

- **What**: Add the core `src/components/ui/context-menu/*` and `src/components/ui/dropdown-menu/*` wrappers used by the follow-up refactor.
- **What**: Add `closeOnScroll` support to the content wrappers so menus can dismiss on scroll without extra parent plumbing.
- **What**: Prune unused generated subcomponents to satisfy repo quality gates, and add smoke tests that exercise open/close behavior for both primitive groups.

## Review Focus

Confirm the wrapper surface is the right minimal base for follow-up menu migrations, especially the `closeOnScroll` behavior and the decision to keep only the primitives currently used by the stack.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10415-feat-add-shadcn-context-and-dropdown-menu-primitives-32d6d73d36508166abfec15baafc967b) by [Unito](https://www.unito.io)


benceruleanlu:
Modifying https://github.com/Comfy-Org/ComfyUI_frontend/pull/10303 in place

PR #10418, opened 2026-03-24
fix: block workflow queue when file input nodes have empty selections

mattmillerai:
## Summary
- Validates file-input nodes (LoadImage, LoadAudio, Load3D, LoadVideo) before submitting workflows
- Shows a localized error dialog listing affected nodes when required file fields are empty
- Prevents round-trip to backend for an error the user can fix locally

Fixes COM-17398

## Related
- Parent bug: COM-17391
- Backend fix: https://github.com/Comfy-Org/cloud/pull/2946

## Test plan
- [ ] Import a shared workflow with a LoadImage node missing its image → localized "Missing File Inputs" dialog appears
- [ ] Run a workflow with all file inputs populated → works normally
- [ ] Run a workflow with no file-input nodes → works normally

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10418-fix-block-workflow-queue-when-file-input-nodes-have-empty-selections-32d6d73d365081a78756f00d7cad7fd3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/locales/en/main.json
+++ b/src/locales/en/main.json
@@ -1889,7 +1889,9 @@
     "extensionFileHint": "This may be due to the following script",
     "promptExecutionError": "Prompt execution failed",
     "accessRestrictedTitle": "Access Restricted",
-    "accessRestrictedMessage": "Your account is not authorized for this feature."
+    "accessRestrictedMessage": "Your account is not authorized for this feature.",
+    "emptyFileInputTitle": "Missing File Inputs",
+    "emptyFileInputMessage": "The following nodes require a file to be selected: {nodeList}. Please upload or select files before running."
   },
   "apiNodesSignInDialog": {
     "title": "Sign In Required to Use API Nodes",
--- a/src/scripts/app.ts
+++ b/src/scripts/app.ts
@@ -146,7 +146,7 @@ import {
 
 export const ANIM_PREVIEW_WIDGET = '$$comfy_animation_preview'
 
-export const FILE_INPUT_FIELDS = {
+const FILE_INPUT_FIELDS = {
   LoadImage: 'image',
   LoadAudio: 'audio',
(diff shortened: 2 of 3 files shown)


AustinMroz (commented):
This PR runs into the other side of the 'nodes which don't actually execute' problem: An input which is not utilized will still block graph execution:
image

And it would be preferred if it didn't use the old error dialogue (which is mostly phased out) and instead used the newer error system so that nodes with errors will be highlighted and can be navigated to from the error panel
image

Unfortunately, both of these are rather involved changes.

mattmillerai:
Thanks for the review Austin — both points are valid.

1. **Disconnected nodes**: `graphToPrompt()` serializes all non-muted nodes regardless of whether they're reachable from an output node, so a disconnected LoadImage with no file selected would incorrectly block the run. Fixing this properly requires walking backwards from output nodes to determine reachability — there's no existing utility for that.

2. **Newer error system**: Agreed the old `showErrorDialog` should be replaced with `executionErrorStore.lastNodeErrors` + `showErrorOverlay()` so nodes get highlighted and are navigable from the error panel.

Both are more involved than what we scoped here. The backend fix is already merged (https://github.com/Comfy-Org/cloud/pull/2946), so the server-side safety net is in place. I think the frontend fix is better owned by someone closer to the graph traversal and error system internals.

Full context is in the Notion ticket: https://www.notion.so/32d6d73d365081648073dea6f21082f8 (COM-17398), parent bug: https://www.notion.so/32c6d73d365081ababe3ea443b662153 (COM-17391).

Going to convert this back to draft for now — happy to hand off.

mattmillerai:
Closing — the backend fix is already merged (cloud#2946) so the server-side safety net is in place. The frontend fix needs graph traversal and error system work that's better owned by someone closer to those internals.

PR #10419, opened 2026-03-24
refactor: migrate node tooltips from PrimeVue to BaseTooltip

dante01yoon:
## Summary

PR 3 in the tooltip migration series. Migrates all `useNodeTooltips` consumers from PrimeVue `v-tooltip` directive to the Reka UI `BaseTooltip` component.

- Refactored `useNodeTooltips` composable: removed PrimeVue-specific `createTooltipConfig`, `hideTooltipsGlobally`, `tooltipsTemporarilyDisabled`, `setupGlobalTooltipHiding`, and global event listeners. Kept text-resolution functions and added `tooltipDelay` computed.
- Migrated `NodeHeader.vue`, `InputSlot.vue`, `OutputSlot.vue`, `NodeWidgets.vue` from `v-tooltip` directive to `<BaseTooltip>` wrapper component.
- Tooltip delay from `LiteGraph.Node.TooltipDelay` setting is passed via `:delay-duration` prop.
- Updated `NodeHeader.test.ts` and `LGraphNode.test.ts` to stub `BaseTooltip`.

## Test plan

- [x] `pnpm typecheck` passes
- [x] `pnpm lint:fix` passes
- [x] `pnpm format` passes
- [x] All vueNodes unit tests pass (48 files, 560 tests)
- [ ] Manual verification: hover over node headers, input/output slots, and widgets to confirm tooltips appear with correct text, position, and delay

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10419-refactor-migrate-node-tooltips-from-PrimeVue-to-BaseTooltip-32d6d73d365081018560d089011c66fe) by [Unito](https://www.unito.io)


christian-byrne (commented):
Review focused on the 2 PR-specific commits (82a0b593 and 25ac20ff). The migration approach is sound; flagging a few correctness concerns and suggestions before merge. Findings [1] (canvas pan/zoom tooltip dismissal) and [3] (dynamic component wrapping via as-child) deserve explicit verification.

src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
```diff
@@ -90,15 +15,14 @@ export function useNodeTooltips(nodeType: MaybeRef<string>) {
   const nodeDefStore = useNodeDefStore()
   const settingsStore = useSettingStore()
 
-  // Setup global pointerdown listener once
-  setupGlobalTooltipHiding()
-
-  // Check if tooltips are globally enabled
   const tooltipsEnabled = computed(() =>
```
christian-byrne, line 18: issue: removed global wheel/pointerdown tooltip-dismissal logic without replacement.

The deleted setupGlobalTooltipHiding / hideTooltipsGlobally / tooltipsTemporarilyDisabled flag worked around the fact that DOM tooltips stay anchored to a moving trigger during canvas pan/zoom and during slot wire-drag. Reka UI's TooltipRoot only closes on pointerLeave/blur/Escape; it does not close on wheel events or while a drag is in progress. With the trigger element moving via CSS transform under a stationary cursor, the tooltip will stay visible/anchored during pan/zoom until pointerLeave finally fires. The TODO in the old code ("use a better tooltip component like RekaUI") implied a replacement strategy, not deletion.

Suggested fix: re-add a small pointerdown/wheel listener that flips a shared ref consumed via :disabled on each BaseTooltip, or wrap the canvas in a controlled TooltipProvider whose state can be dismissed imperatively. Please verify the slot wire-drag case explicitly.

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -1,63 +1,71 @@
 <template>
   <div v-if="renderError" class="node-error p-1 text-xs text-red-500">⚠️</div>
-  <div
+  <BaseTooltip
```
christian-byrne, line 3: issue (verify before merge): wrapping the wire-drag affordance in BaseTooltip may interfere with drag UX.

The cursor-crosshair area + SlotConnectionDot are now inside <BaseTooltip> whose TooltipTrigger as-child attaches pointer/focus handlers via cloning. Reka UI fires pointerdown to dismiss but does not preventDefault, so existing @pointerdown on SlotConnectionDot should still fire, but please verify in browser that (a) wire-drag still initiates cleanly and (b) the tooltip doesn't appear mid-drag once a link is in flight. This used to be suppressed via the now-deleted tooltipsTemporarilyDisabled.

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -57,79 +57,48 @@
...
-            :class="
-              cn(
-                'col-span-2',
-                widget.hasError && 'font-bold text-node-stroke-error'
-              )
-            "
-            @update:model-value="widget.updateHandler"
-            @contextmenu="widget.handleContextMenu"
-          />
+          <BaseTooltip
```
christian-byrne, line 60: issue: BaseTooltip wraps a dynamic <component :is="widget.vueComponent"> via as-child.

TooltipTrigger as-child requires the slot child to (a) have a single root element and (b) forward $attrs to that root (Vue 3's default unless inheritAttrs: false). Any widget that renders a Fragment, uses inheritAttrs: false without v-bind="$attrs", or returns a teleported root will silently drop pointerenter/leave handlers and the tooltip will never appear (or will appear on the wrong element). Suggest auditing every component registered under widget.vueComponent, or wrapping each widget in a stable <div> trigger here.

src/renderer/extensions/vueNodes/composables/useNodeTooltips.ts
```diff
@@ -90,15 +15,14 @@ export function useNodeTooltips(nodeType: MaybeRef<string>) {
...
     settingsStore.get('Comfy.EnableTooltips')
   )
 
-  // Get node definition for tooltip data
+  const tooltipDelay = computed(
+    () => settingsStore.get('LiteGraph.Node.TooltipDelay') as number
```
christian-byrne, line 23: suggestion (non-blocking): drop the unsafe `as number` cast on the setting value.

settingsStore.get('LiteGraph.Node.TooltipDelay') is already typed `number` via apiSchema.ts (z.number()), so the cast is dead weight. If the setting is transiently undefined during hydration, the cast hides it and Reka UI inherits the App.vue TooltipProvider's default. Consider dropping the cast and using `?? 500` (or whatever the intended default is) for safety.

src/renderer/extensions/vueNodes/components/NodeWidgets.vue
```diff
@@ -173,314 +135,7 @@ onErrorCaptured((error) => {
...
-  return toValue(processedWidgets)
-    .filter((w) => !w.hidden && (!w.advanced || showAdvanced.value))
-    .map((w) =>
-      shouldExpand(w.type) || w.hasLayoutSize ? 'auto' : 'min-content'
-    )
-    .join(' ')
-})
+const { nodeType, showAdvanced, canSelectInputs, processedWidgets, gridTemplateRows } =
+  useProcessedWidgets(() => nodeData)
+const { tooltipsEnabled, tooltipDelay } = useNodeTooltips(nodeType)
```
christian-byrne, line 140: suggestion (non-blocking): useNodeTooltips(nodeType) is called twice for the same node.

useProcessedWidgets already calls useNodeTooltips(nodeType) internally to power getWidgetTooltip, and now NodeWidgets calls it again to grab tooltipsEnabled / tooltipDelay. Both invocations re-run store lookups and create duplicate computeds. Consider exposing tooltipsEnabled / tooltipDelay from useProcessedWidgets, or passing them as part of the processed widget object.

src/renderer/extensions/vueNodes/components/OutputSlot.test.ts
```diff
@@ -21,7 +22,8 @@ vi.mock('@/renderer/core/canvas/links/slotLinkDragUIState', () => ({
 vi.mock('@/renderer/extensions/vueNodes/composables/useNodeTooltips', () => ({
```
christian-byrne, line 22: issue: useNodeTooltips mock factory violates the project's singleton-mock convention.

Per docs/testing/unit-testing.md, mock instances (refs, computeds, vi.fn) must be created once inside the vi.mock factory and shared across calls. Here `useNodeTooltips: () => ({ ..., tooltipsEnabled: computed(...), tooltipDelay: computed(...) })` allocates fresh computed instances on every call, breaking referential equality and preventing per-test state manipulation via vi.mocked(). Lift the refs/computeds to the factory's outer scope and return them from a stable useNodeTooltips closure.

src/renderer/extensions/vueNodes/components/NodeHeader.test.ts
```diff
@@ -96,150 +98,113 @@ const createMountConfig = () => {
...
+    await rerender({
+      nodeData: makeNodeData(),
+      collapsed: true
+    })
+    // eslint-disable-next-line testing-library/no-node-access
+    const collapsedIcon = collapseButton.querySelector('i')!
+    expect(collapsedIcon.classList).toContain('-rotate-90')
   })
 
   describe('Tooltips', () => {
```
christian-byrne, line 201: suggestion: the "Tooltips" describe block lost all behavioural assertions.

The sole remaining test only checks that the node-title testid exists, which the prior body already covered. There is now zero assertion that BaseTooltip is wired with text=getNodeDescription, disabled=!tooltipsEnabled, or that the tooltip is suppressed while editing. Adding at least one findComponent(BaseTooltip).props(...) assertion would prevent the binding from silently rotting.

src/renderer/extensions/vueNodes/components/NodeHeader.vue
```diff
@@ -38,21 +38,28 @@
           </Button>
         </div>
         <!-- Node Title -->
-        <div
-          v-tooltip.top="tooltipConfig"
-          class="flex min-w-0 flex-1 items-center gap-2"
-          data-testid="node-title"
+        <BaseTooltip
+          :text="isEditing ? '' : getNodeDescription"
```
christian-byrne, line 42: suggestion (non-blocking): :text="isEditing ? '' : getNodeDescription" mixes ComputedRef with string in a ternary.

Vue's template top-level ref auto-unwrap saves this today, but a refactor wrapping the expression (e.g. into a function) would silently pass [object Object]. The composable's API is also inconsistent: getNodeDescription is a ComputedRef while getInputSlotTooltip / getOutputSlotTooltip / getWidgetTooltip are plain functions. Consider expressing this via a local `computed(() => isEditing.value ? '' : getNodeDescription.value)`, or normalizing the composable to expose getters consistently.

browser_tests/tests/baseTooltip.spec.ts
```diff
@@ -3,7 +3,7 @@ import type { Locator, Page } from '@playwright/test'
 import {
   comfyPageFixture as test,
   comfyExpect as expect
-} from '../fixtures/ComfyPage'
+} from '@e2e/fixtures/ComfyPage'
```
christian-byrne, line 6: suggestion: missing e2e coverage for the migrated surfaces.

This spec covers the Queue history button + properties panel toggle, but none of the four files this PR migrates (NodeHeader, InputSlot, OutputSlot, NodeWidgets). Given the wire-drag risk and the dynamic <component :is> wrapping concern, please add at least one Playwright spec that hovers a vue node title, an input slot, an output slot, and a widget, then asserts [role="tooltip"] appears and disappears.

src/renderer/extensions/vueNodes/components/InputSlot.vue
```diff
@@ -94,15 +102,13 @@ const dotOnly = computed(() => props.dotOnly || hasNoLabel.value)
...
   props.nodeType || ''
 )
 
-const tooltipConfig = computed(() => {
+const inputTooltipText = computed(() => {
   const slotName = props.slotData.localized_name || props.slotData.name || ''
-  const tooltipText = getInputSlotTooltip(slotName)
-  const fallbackText = tooltipText || `Input: ${slotName}`
-  return createTooltipConfig(fallbackText)
+  return getInputSlotTooltip(slotName) || `Input: ${slotName}`
```
christian-byrne, line 111: nitpick (non-blocking): synthetic fallback strings are computed even when tooltips are disabled.

`getInputSlotTooltip(slotName) || \`Input: ${slotName}\`` always builds the fallback, then `:disabled="!tooltipsEnabled"` saves us. For OutputSlot the iterative suffix gets appended onto the synthetic `Output: <name>` fallback, meaning a slot with no real description renders as "Output: foo iterative". Consider only computing/appending when meaningful tooltip text exists.

src/renderer/extensions/vueNodes/components/NodeHeader.test.ts
```diff
@@ -96,150 +98,113 @@ const createMountConfig = () => {
     global: {
       plugins: [PrimeVue, i18n, pinia],
       components: { InputText },
-      directives: {
-        tooltip: {
-          mounted: vi.fn(),
-          updated: vi.fn(),
-          unmounted: vi.fn()
+      stubs: {
```
christian-byrne, line 101: nitpick (non-blocking): the BaseTooltip stub renders <div><slot /></div> here but <slot /> elsewhere.

Mismatched stub shape across test files (NodeWidgets and OutputSlot use `<slot />`) makes DOM assertions brittle and adds an extra wrapper element only here. Consider hoisting a single shared stub or standardizing on `<slot />`.

PR #10420, opened 2026-03-24
docs: ADR 0008 — Entity Component System

DrJKL:
## Summary

Architecture documentation proposing an Entity Component System for the litegraph layer.

```mermaid
graph LR
    subgraph Today["Today: Spaghetti"]
        God["🍝 God Objects"]
        Circ["🔄 Circular Deps"]
        Mut["💥 Render Mutations"]
    end

    subgraph Tomorrow["Tomorrow: ECS"]
        ID["🏷️ Branded IDs"]
        Comp["📦 Components"]
        Sys["⚙️ Systems"]
        World["🌍 World"]
    end

    God -->|"decompose"| Comp
    Circ -->|"flatten"| ID
    Mut -->|"separate"| Sys
    Comp --> World
    ID --> World
    Sys -->|"query"| World
```

## Changes

- **What**: ADR 0008 + 4 architecture docs (no code changes)
  - `docs/adr/0008-entity-component-system.md` — entity taxonomy, branded IDs, component decomposition, migration strategy
  - `docs/architecture/entity-interactions.md` — as-is Mermaid diagrams of all entity relationships
  - `docs/architecture/entity-problems.md` — structural problems with file:line evidence
  - `docs/architecture/ecs-target-architecture.md` — target architecture diagrams
  - `docs/architecture/proto-ecs-stores.md` — analysis of existing Pinia stores as proto-ECS patterns

## Review Focus

- Does the entity taxonomy (Node, Link, Subgraph, Widget, Slot, Reroute, Group) cover all cases?
- Are the component decompositions reasonable starting points?
- Is the migration strategy (bridge layer, incremental extraction) feasible?
- Are there entity interactions or problems we missed?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10420-docs-ADR-0008-Entity-Component-System-32d6d73d365081feb048d16a5231d350) by [Unito](https://www.unito.io)


docs/adr/0008-entity-component-system.md
```diff
@@ -0,0 +1,185 @@
...
+
+Widgets and Slots currently lack independent IDs. The ECS will assign synthetic IDs at entity creation time via an auto-incrementing counter (matching the pattern used by `lastNodeId`, `lastLinkId`, etc. in `LGraphState`).
+
+### Component Decomposition
+
+Components are plain data objects — no methods, no back-references to parent entities. Systems query components to implement behavior.
+
+#### Shared Components
+
+- **Position** — `{ pos: Point, size: Size, bounding: Rectangle }` — used by Node, Reroute, Group
```
AustinMroz: Reroutes don't have a size.

docs/adr/0008-entity-component-system.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Branded ID Design
+
+Each entity kind gets a nominal/branded type wrapping its underlying primitive. The brand prevents accidental cross-kind usage at compile time while remaining structurally compatible with existing ID types:
+
+```ts
+type NodeEntityId = number & { readonly __brand: 'NodeEntityId' }
+type LinkEntityId = number & { readonly __brand: 'LinkEntityId' }
+type SubgraphEntityId = string & { readonly __brand: 'SubgraphEntityId' }
+type WidgetEntityId = number & { readonly __brand: 'WidgetEntityId' }
```
AustinMroz, line 55: Having widgets keyed by id seems at odds with the desire to make the serialized format a direct snapshot of the ECS state.

In order for a serialized workflow to be executable, We need to be able to relate a node id to a list of widget names and values.

artokun:
yes. EC Yes

docs/architecture/entity-problems.md
```diff
@@ -0,0 +1,209 @@
...
+
+- `LLink.ts:24` — `const layoutMutations = useLayoutMutations()` (module scope)
+- `Reroute.ts` — same pattern at module scope
+- `BaseWidget.ts:20-22` — imports `usePromotionStore` and `useWidgetValueStore`
+
+These make the domain objects untestable without a Vue app context.
+
+### Change Notification Sprawl
+
+`beforeChange()` and `afterChange()` (undo/redo checkpoints) are called from **12+ locations** in `LGraphCanvas` alone (lines 1574, 1592, 1604, 1620, 1752, 1770, 8754, 8760, 8771, 8777, 8803, 8811). Missing a call means undo/redo doesn't capture the change; adding an extra call means a wasted checkpoint.
```
AustinMroz: IIRC, this is not accurate. Quite the opposite. beforeChange/afterChange are used to group updates so that a single operations doesn't span multiple undo/redo operations.

christian-byrne (commented):
## Architecture Review Questions

A few questions worth discussing in the design review:

### 1. Linked Subgraphs and Instance-Varying State

The ADR claims the World is "the single source of truth for all entity state." But consider the future where we want Blender-style linked subgraphs — same topology shared across instances, but different widget *values* per instance.

In the proposed ECS, `WidgetEntityId:847` maps to exactly one `WidgetValue` component. If linked instances share entity IDs (that's what makes them "linked"), they share values too. There's no room for per-instance divergence.

WidgetValueStore *already* solves this with `graphId:nodeId:widgetName` scoping — each instance context gets separate values. The ECS ADR proposes replacing this store, but the linked-subgraph future *requires* exactly this kind of scoped indirection.

Does the World need an instance-scoping mechanism? Or should the ADR acknowledge that instance-varying data (widget values, execution state) lives in a layer *above* the World, and the World only owns definition-level structure? If so, "single source of truth" needs a caveat.

### 2. Subgraph Recursion: Can Flat Composition Replace Structural Modeling?

`Subgraph extends LGraph` is a genuine domain isomorphism — a subgraph *is* a graph. The ECS flattens this into entity IDs + `SubgraphStructure` component. The appendix flags this but doesn't resolve it.

The hard test: subgraphs containing subgraphs. Recursive nesting is where inheritance expresses the domain naturally. Can flat entity composition with `SubgraphStructure` component references express the same thing without losing the "a subgraph is a graph" relationship? What does querying "all nodes at depth N" look like with flat entities vs recursive graph objects?

### 3. beforeChange/afterChange: Grouping Semantics

Austin's review correctly notes these are used to *group* operations into single undo entries, not to capture individual changes. The ECS `VersionSystem` proposal centralizes change *tracking* but doesn't address operation *grouping*.

How does undo/redo batch granularity work in the World? If `ConnectivitySystem.connect()` makes 5 component writes, is that 1 undo entry or 5? The current `beforeChange()`/`afterChange()` wrapping ensures it's 1. What's the ECS equivalent?

### 4. Render-Time Layout Separation

The docs correctly identify `arrange()` during `drawNode()` as pathological. The ECS fix (LayoutSystem in update phase, RenderSystem read-only) is architecturally sound. But is it achievable incrementally?

Separating layout from render means every code path that currently calls `drawNode()` (which triggers `arrange()` as a side effect) must instead ensure layout ran first. Does this require touching every node type at once, or can it be done per-node-type incrementally?

### 5. Bridge Layer Maintenance Cost

The bridge (Phases 2–4) is the longest-lived artifact. During that period, every change to entity state must work in both systems. Every new feature, every bug fix, every extension interaction has to be verified against both paths.

What's the estimated duration of the bridge phase? What's the ongoing cost of maintaining dual-path state? Is there a risk the bridge becomes permanent technical debt — too costly to remove, too fragile to ignore?

### 6. Performance: Map Lookups vs Direct Property Access

Current hot path: `node.pos` — direct property read.
ECS hot path: `world.getComponent(nodeId, Position)` — Map lookup + component extraction.

The render loop hits every node every frame. Has anyone measured the overhead of replacing direct property access with Map-based component lookups in the render path? Even small per-node overhead multiplied by hundreds of nodes at 60fps could matter.

### 7. What Triggers Phase 5?

Phase 5 (legacy removal) is marked "irreversible" and "very high risk." What are the concrete exit criteria for entering Phase 5?

- 100% extension compatibility verified?
- A specific test suite passing?
- A feature-flag period with opt-in/opt-out?
- A deadline?

Without defined exit criteria, the bridge could live indefinitely — which means the codebase carries both systems forever.

christian-byrne (commented):
## Extension Ecosystem Compatibility Audit

Searched the ComfyUI custom node ecosystem for usage of the APIs this ECS migration would affect. The results are sobering — **12 of 14 patterns would break**, affecting 40+ repos.

### High-Impact Patterns (50+ matches each across dozens of repos)

| Pattern | Example Repos | ECS Impact |
|---------|--------------|------------|
| `onConnectionsChange` | rgthree-comfy, ComfyUI-PM-Nodes, comfy-ovum | ⚠️ Breaks — becomes system observer |
| `onRemoved` | rgthree-comfy, ComfyUI-MaxedOut, comfy-ovum | ⚠️ Breaks — becomes entity-removal event |
| `onAdded` | rgthree-comfy, comfy-ovum, ComfyUI-Flux-Continuum | ⚠️ Breaks — becomes entity-creation event |
| `onConfigure` | comfyui-deploy, lf-nodes, ComfyUI-Fill-Nodes | ⚠️ Breaks — needs ECS deserialization hook |
| `onConnectInput/Output` | rgthree-comfy, cg-use-everywhere, comfy-pack | ⚠️ Breaks — connection validation must become system-level |
| `onWidgetChanged` | comfy-ovum, ComfyUI-Nexus, ComfyUI-AK-Pack | ⚠️ Breaks — becomes component-change observation |
| `node.widgets.find(w => w.name === ...)` | ComfyUI-Impact-Pack, ComfyUI-Easy-Use, VideoHelperSuite | ⚠️ Breaks — needs component query API |
| `graph._version++` | ComfyUI-Flux-Continuum, ComfyUI-Nexus, comfy_mtb | 🔴 Critical — direct private mutation, no public API exists |
| `node.serialize` overrides | ComfyUI-Nexus, wavespeed-comfyui, ComfyUI-Fill-Nodes | ⚠️ Breaks — needs serialization extension protocol |

### Most-Affected Repo: `rgthree/rgthree-comfy`
Uses virtually every pattern — `onConnectInput`, `onConnectOutput`, `onConnectionsChange`, `onRemoved`, `onAdded`, `getNodeById`, `node.widgets.find`. This repo alone would need comprehensive migration.

### Only Safe Pattern: `getNodeById` (50+ matches)
Can be preserved as a thin wrapper over ECS entity queries.

### Recommendation
The ADR and migration plan should include:
1. **Extension API compatibility matrix** — which callbacks survive, which get event equivalents
2. **Extension migration guide** — concrete examples of old pattern → new pattern
3. **`graph._version++` must become a public API immediately** (Phase 0) — extensions already depend on this side-channel
4. **rgthree-comfy as validation target** — if this repo works, most others will too

christian-byrne (commented):
## Command-Pattern Compatibility Analysis

We have an emerging requirement that all mutations be expressed as **serializable, idempotent, deterministic commands** (for CRDT sync, cross-environment portability, undo/redo, and gateway backends). The ECS architecture as proposed conflicts with this in fundamental ways.

### Critical Conflicts

**1. The World API is imperative, not command-based**

```ts
// What the ECS proposes:
world.setComponent(nodeId, Position, { pos: [100, 200] })
world.deleteEntity(kind, id)

// What the command pattern requires:
world.execute({ type: "SetComponent", entityId, componentKey, data, version })
```

Every system, bridge, and lifecycle scenario directly mutates the World. There is no command object — no serializable intent that can be sent to a gateway service, replayed in WASM, or persisted for undo. The mutation *is* the side effect, not a description of it.

**2. Systems are imperative orchestrators, not command producers**

The lifecycle scenarios show `ConnectivitySystem.connect(world, outputSlotId, inputSlotId)` as a function call with live object references. This cannot be serialized, sent over a wire, or replayed in Go/Python. Systems should *produce* command batches, not execute mutations directly.

**3. Auto-incrementing entity IDs are non-stable**

> "The ECS will assign synthetic IDs at entity creation time via an auto-incrementing counter"

Auto-increment counters collide in concurrent environments. Commands referencing these IDs cannot be replayed on a different peer. CRDT environments need UUIDs or Yjs item IDs — globally unique, stable identifiers.

**4. No transaction primitive exists**

The docs *claim* atomicity ("no intermediate inconsistent state") but the migration plan *acknowledges* this is an open question:

> "Does the World need a `transaction()` API? How does this interact with Y.js transactions?"

Multi-operation commands (node removal = 10+ `deleteEntity` calls) without transactions create intermediate inconsistent states visible to remote peers.

**5. No idempotency guarantees**

`createEntity()` allocates a new ID every time. Replaying the same command creates duplicate entities. There is no mechanism to detect "this was already applied." Commands need content-addressed or deterministic IDs.

**6. No error semantics**

The World API is void-returning. What happens when you `deleteEntity` on a non-existent ID? `connect()` to an invalid slot? No result type, no rejection semantics, no compensation. Commands need `{ status: "applied" | "rejected" | "no-op" }`.

### Root Cause

The ECS is designed as a **state container with imperative mutators** (classic game-engine ECS). The command-pattern requirements demand a **mutation-intent layer** between callers and state. The entire architecture would need an additional layer:

```
Caller → Command object → Command executor → World mutation → Yjs transaction
                ↓
        Command log (for replay, undo, sync)
```

### Recommendation

The ADR should either:
1. **Incorporate the command layer from the start** — Systems produce commands, the World consumes them through a command executor that handles transactions, idempotency, and serialization
2. **Or explicitly acknowledge** that the World API is an internal implementation detail and the public mutation API will be command-based, with the World as the backend

Without this, the ECS and the command-pattern architecture will be architecturally incompatible and require a second major migration.

christian-byrne (dismissed):

docs/adr/0008-entity-component-system.md
artokun: Can’t it just be a component on that entity? The component can have a parent entity id

artokun:
Yes. Love this

PR #10427, opened 2026-03-24
fix: sync advanced inputs button color with node header

jaeone94:
## Summary
- The "Show Advanced Inputs" footer button was missing `headerColor` style binding, causing it to not sync with the node header color (unlike the "Enter Subgraph" button which already had it)
- Extracted the repeated `{ backgroundColor: headerColor }` inline style (4 occurrences) into a `headerColorStyle` computed

## Screenshots 
before 
스크린샷 2026-03-24 154312

after 
스크린샷 2026-03-24 154622

## Test plan
- [x] Set a custom color on a node with advanced inputs and verify the footer button matches the header color
- [x] Verify subgraph enter button still syncs correctly
- [x] Verify dual-tab layouts (error + advanced, error + subgraph) both show correct colors

### Why no E2E test
Node header color is applied as an inline style via `headerColor` prop, which is already passed and tested through the existing subgraph enter button path. This change simply extends the same binding to the advanced inputs buttons — no new data flow or interaction is introduced, so a screenshot-based E2E test would add maintenance cost without meaningful regression coverage.

### DIFF
--- a/src/renderer/extensions/vueNodes/components/NodeFooter.vue
+++ b/src/renderer/extensions/vueNodes/components/NodeFooter.vue
@@ -28,7 +28,7 @@
           '-z-10 bg-node-component-header-surface'
         )
       "
-      :style="{ backgroundColor: headerColor }"
+      :style="headerColorStyle"
       @click.stop="$emit('enterSubgraph')"
     >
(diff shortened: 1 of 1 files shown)


jaeone94:
@coderabbitai can you approve it

AustinMroz (approved):

PR #10439, opened 2026-03-24
feat: App mode - Rework save flow

pythongosssss:
## Summary

Users were finding the final step of the builder flow confusing/misleading, with the "choose default mode" not actually saving the workflow and people losing changes. This updates it to remove "save"/"set default" as a step in the builder, and changes it to a distinct action.

## Changes

- **What**: 
- add mode selection tab on footer toolbar
- extract reusable radio group component
- remove setting default mode dialog
- add save/save as/saved dialogs

https://github.com/user-attachments/assets/c7439c2e-a917-4f2b-b176-f8bb8c10026d

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10439-feat-App-mode-Rework-save-flow-32d6d73d3650814781b6c7bbea685a97) by [Unito](https://www.unito.io)


browser_tests/helpers/builderTestUtils.ts
```diff
@@ -0,0 +1,118 @@
...
+    canvasBox.y + canvasBox.height / 2
+  )
+  await comfyPage.nextFrame()
+}
+
+/** Center on a node and click its first widget to select it as input. */
+async function selectInputWidget(comfyPage: ComfyPage, node: NodeReference) {
+  const { page } = comfyPage
+
+  await comfyPage.canvasOps.setScale(1)
```
AustinMroz, line 37: Isn't canvasOps for litegraph node positions?

EDIT: Seems to be an extraction on existing tests. Not blocking if it works. 

In my recent tests, I've been using variants of
```ts
await comfyPage.vueNodes.getNodeLocator('2').locator('.lg-node-widget').click()
```

src/locales/en/main.json
```diff
@@ -3682,28 +3682,31 @@
...
     "viewApp": "View app",
-    "exitToWorkflow": "Exit to workflow",
+    "saveAs": "Save as",
+    "filename": "Filename",
     "emptyWorkflowTitle": "This workflow has no nodes",
     "emptyWorkflowPrompt": "Do you want to start with a template?"
   },
+  "builderFooter": {
+    "opensAsApp": "Open as an {mode}",
+    "opensAsGraph": "Open as a {mode}"
```
AustinMroz, line 3702: I'm somewhat curious how our automated translation handles `an` vs `a` here...

Probably a massive can of worms with no clear solution. Even if we make the slot name vary, the slot names won't be translated and won't be accurate to the target language. :thinking:

AustinMroz (approved):

src/locales/en/main.json
pythongosssss: Hopefully not an issue in other languages (or maybe the translation gets the context of the key name and can maybe infer from there...) 🤞

browser_tests/helpers/builderTestUtils.ts
pythongosssss: This may need updating to allow specific widgets for some upcoming tests so will keep this in mind for future updates

PR #10445, opened 2026-03-24
fix: disable preload error toast triggered by third-party plugin failures

jtydhr88:
## Summary
The preload error toast fires whenever any custom node extension fails to load via dynamic `import()`. In practice, this is almost always caused by third-party plugin bugs rather than ComfyUI core issues. Common triggers include:

  - Bare module specifiers (e.g., `import from "vue"`) that the browser cannot resolve without an import map
  - Incorrect relative paths to `scripts/app.js` due to nested web directory structures
  - Missing dependencies on other extensions (e.g., `clipspace.js`)

  Since many users have multiple custom nodes installed, the toast frequently appears on startup — sometimes multiple times — with a generic message that offers no actionable guidance. This creates unnecessary alarm and support burden.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10445-fix-disable-preload-error-toast-triggered-by-third-party-plugin-failures-32d6d73d365081f281efcd6fe90642a5) by [Unito](https://www.unito.io)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -9,13 +9,10 @@ import { captureException } from '@sentry/vue'
 import BlockUI from 'primevue/blockui'
 import { computed, onMounted, onUnmounted, watch } from 'vue'
 
-import { useI18n } from 'vue-i18n'
-
 import GlobalDialog from '@/components/dialog/GlobalDialog.vue'
 import config from '@/config'
 import { isDesktop } from '@/platform/distribution/types'
 import { useSettingStore } from '@/platform/settings/settingStore'
-import { useToastStore } from '@/platform/updates/common/toastStore'
 import { app } from '@/scripts/app'
 import { useWorkspaceStore } from '@/stores/workspaceStore'
 import { electronAPI } from '@/utils/envUtil'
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -98,12 +94,17 @@ onMounted(() => {
         }
       })
     }
-    useToastStore().add({
-      severity: 'error',
-      summary: t('g.preloadErrorTitle'),
-      detail: t('g.preloadError'),
-      life: 10000
-    })
+    // Disabled: Third-party custom node extensions frequently trigger this toast
```
Myestery, line 97: can we use console.error instead?
and tell them what to do right

eg console.error("Importing vue directly is deprecated. use xy instead")

jtydhr88: It is hard for now to know what the error is exactly, at this moment I think we should disable it as priority because I have seen lots reports.
We could design a better solution later

DrJKL (approved):

src/App.vue
DrJKL, line 97: I would just delete this. We can always re-add it.

PR #10454, opened 2026-03-24
Use native context menu for focused textareas

AustinMroz:
The custom context menu provided by the frontend exposes widget specific options. In order to support renaming, promotion, and favoriting, there needs to be a way to access this context menu when targeting a textarea. However, always displaying this custom context menu will cause the user to lose access to browser specific functionality like spell checking, translation, and the ability to copy paste text.

This PR updates the behaviour so that the native browser context menu will display when the text area already has focus. Our custom frontend context menu will continue to display when it does not.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10454-Use-native-context-menu-for-focused-textareas-32d6d73d365081909673d81d6a6ba054) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
@@ -28,7 +28,7 @@
       :placeholder
       :readonly="isReadOnly"
       data-capture-wheel="true"
-      @pointerdown.capture.stop
+      @pointerdown.capture.stop="trackFocus"
       @pointermove.capture.stop
       @pointerup.capture.stop
       @contextmenu.capture="handleContextMenu"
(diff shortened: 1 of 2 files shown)


src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
```diff
@@ -72,6 +72,11 @@ const { widget, placeholder = '' } = defineProps<{
 
 const modelValue = defineModel<string>({ default: '' })
 
+const isFocused = ref(false)
+function trackFocus() {
+  isFocused.value = document.activeElement === document.getElementById(id)
```
DrJKL: Non-blocking: Could you use a `ref` to get the element?

DrJKL (dismissed):

src/renderer/extensions/vueNodes/widgets/components/WidgetTextarea.vue
AustinMroz: I ran into lots of issues with ordering of events when I was testing this. (Hence not using a computed for isFocused.) Regretfully, `unwrapRef` doesn't function here, but I can still use the ref manually. :+1:

DrJKL (approved):

PR #10473, opened 2026-03-24
perf: throttle ResizeObserver during zoom/pan interactions

christian-byrne:
## Summary

Throttle the shared ResizeObserver during active zoom/pan to eliminate 10–40× more expensive `UpdateLayoutTree` calls caused by CSS scale transform recomputation.

## Changes

- **What**: During zoom/pan, the `ResizeObserver` callback now early-returns, deferring elements to a set. When the transform settles (debounce fires), all deferred elements get one accurate re-measurement. A module-level `canvasTransformActive` ref is exported from `useTransformSettling` and read by the RO callback in `useVueNodeResizeTracking`.

## Review Focus

- The transform-deferred elements are **not unobserved** (unlike hidden-tab deferrals) since we still want the RO to fire after settle — we just skip processing during active transforms. This is intentional: unobserving would require re-observing every element, but the settle watcher already triggers fresh measurements via `markElementForFreshMeasurement` + `resizeObserver.observe`.
- The existing `deferredElements` / `markElementForFreshMeasurement` mechanism for tab-hidden recovery is preserved unchanged.
- The `linearMode` early-return pattern on line 100 was the model for this guard.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10473-perf-throttle-ResizeObserver-during-zoom-pan-interactions-32d6d73d365081638f3ffdaf7cc4fe4b) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/performance.spec.ts
+++ b/browser_tests/tests/performance.spec.ts
@@ -154,6 +154,38 @@ test.describe('Performance', { tag: ['@perf'] }, () => {
     )
   })
 
+  test('large graph zoom interaction', async ({ comfyPage }) => {
+    await comfyPage.workflow.loadWorkflow('large-graph-workflow')
+
+    const canvas = comfyPage.canvas
+    const box = await canvas.boundingBox()
+    if (!box) throw new Error('Canvas bounding box not available')
+
+    // Position mouse at center so wheel events hit the canvas
+    const centerX = box.x + box.width / 2
+    const centerY = box.y + box.height / 2
+    await comfyPage.page.mouse.move(centerX, centerY)
+
+    await comfyPage.perf.startMeasuring()
+
+    // Zoom in 30 steps then out 30 steps — each step triggers
+    // ResizeObserver for all ~245 node elements due to CSS scale change.
+    for (let i = 0; i < 30; i++) {
+      await comfyPage.page.mouse.wheel(0, -100)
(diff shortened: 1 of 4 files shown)


christian-byrne:
**Stacked PR approach**: This is PR 2 of 2. PR 1 (#10478) adds the `large-graph-zoom` perf test to establish a baseline on main. Once #10478 merges, CI will run the same test on this branch and show the performance delta proving the improvement.

christian-byrne (commented):
Not sure on two things:

1. is this is an acceptable degradation of the UI/UX/feel of the app while zooming/panning?
2. is accessing (and side effects of RO) actually worthwhile to skip in terms of perf benefit? the RO is already happening regardless of if we skip it in this specific callback or not

christian-byrne:
rebased onto `main` (`a81c1de2`) to resolve the merge conflict; one import-list conflict in `src/renderer/extensions/vueNodes/composables/useVueNodeResizeTracking.ts` (both sides added distinct new imports on top of the shared base) resolved by keeping the superset actually used in the merged body. Fresh head `e4f643e601`, focused vitest (`useTransformSettling.test.ts`, `useVueNodeResizeTracking.test.ts`) 22/22 green, `vue-tsc --noEmit` clean, eslint clean on touched files. mergeStateStatus is now MERGEABLE; CI running.

PR #10474, opened 2026-03-24
perf: progressive node rendering during subgraph transitions

christian-byrne:
## Summary

Break the synchronous mount/unmount cycle that blocks the main thread for ~1.4s when entering/exiting subgraphs by rendering nodes in RAF-driven batches.

## Changes

- **What**: Add `useProgressiveNodeRendering` composable that renders Vue nodes in batches across animation frames instead of all at once. For large graphs (>40 nodes), an initial batch of 24 nodes renders immediately for fast first paint, then remaining nodes are added in batches of 40 per frame with a 6ms time budget. Small graphs render synchronously to avoid overhead.

## Review Focus

- Cancellation token pattern in `useProgressiveNodeRendering` — ensures stale RAF loops from rapid subgraph navigation are properly cancelled
- `pendingSlotSync` suppression during progressive mount — links skip rendering until all nodes are mounted to avoid drawing against incomplete slot layouts
- The `allNodes` watcher that syncs `renderedCount` when not progressively rendering — handles node add/remove during normal editing without breaking the progressive mount loop

## Before/After (expected)

| Metric | Before | After |
|---|---|---|
| Main thread block on subgraph enter | ~1,394ms | ~6ms per frame (24 nodes first paint) |
| Total mount time | Same | Same (work is spread, not removed) |
| Small graph (<40 nodes) | No change | No change (renders synchronously) |

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10474-perf-progressive-node-rendering-during-subgraph-transitions-32d6d73d365081b88b99edc339a3e0a9) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -71,9 +71,9 @@
     @pointerup.capture="forwardPanEvent"
     @pointermove.capture="forwardPanEvent"
   >
-    <!-- Vue nodes rendered based on graph nodes -->
+    <!-- Vue nodes rendered progressively to avoid long-task UI freezes -->
     <LGraphNode
-      v-for="nodeData in allNodes"
+      v-for="nodeData in visibleNodes"
       :key="nodeData.id"
       :node-data="nodeData"
       :error="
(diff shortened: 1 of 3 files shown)


christian-byrne:
Closing in favor of a stacked PR approach: PR1 will add a @perf test to establish baseline, PR2 will add the fix.

PR #10475, opened 2026-03-24
perf: debounce convertToBlob in GLSL renderer

christian-byrne:
## Summary

Add debounced `convertToBlob` to the GLSL renderer to reduce CPU overhead during pan/zoom.

## Changes

- **What**: Add `debouncedToBlob()` and `cancelPendingBlob()` to `useGLSLRenderer`. Rapid calls within a 150ms window are coalesced into a single `convertToBlob` call — all callers receive the same blob. `dispose()` now cancels pending conversions to prevent leaks. The original `toBlob()` is preserved for user-initiated operations (clipboard copy, mask editor save).

## Review Focus

- The debounce delay (150ms) balances responsiveness with CPU savings. Profiling showed `convertToBlob` in every scenario's top-5 hot functions (93ms total self-time across 7 scenarios).
- Previous pending callers all resolve with the same blob rather than being rejected, avoiding unnecessary error handling in consumers.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10475-perf-debounce-convertToBlob-in-GLSL-renderer-32d6d73d365081648800e8a0d5239507) by [Unito](https://www.unito.io)

### DIFF
--- a/src/renderer/glsl/useGLSLRenderer.ts
+++ b/src/renderer/glsl/useGLSLRenderer.ts
@@ -1,3 +1,5 @@
+import { debounce } from 'es-toolkit'
+
 import { detectPassCount } from '@/renderer/glsl/glslUtils'
 
 const VERTEX_SHADER_SOURCE = `#version 300 es
@@ -443,9 +445,18 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
     return canvas.convertToBlob({ type: 'image/webp', quality: 0.92 })
   }
 
+  const DEBOUNCE_DELAY_MS = 150
+
+  const debouncedToBlob = debounce(() => toBlob(), DEBOUNCE_DELAY_MS)
+
+  function cancelPendingBlob(): void {
+    debouncedToBlob.cancel()
+  }
+
   function dispose(): void {
     if (disposed) return
     disposed = true
+    cancelPendingBlob()
     if (!gl) return
 
     for (const tex of inputTextures) {
@@ -497,6 +508,8 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
     render,
     readPixels,
     toBlob,
+    debouncedToBlob,
+    cancelPendingBlob,
     dispose
   }
 }
--- a/src/renderer/glsl/useGLSLRenderer.test.ts
+++ b/src/renderer/glsl/useGLSLRenderer.test.ts
@@ -166,6 +166,10 @@ vi.stubGlobal(
   }
 )
 
+const mockConvertToBlob = vi.fn(() =>
+  Promise.resolve(new Blob(['fake'], { type: 'image/webp' }))
+)
+
 vi.stubGlobal(
   'OffscreenCanvas',
   class {
@@ -181,7 +185,7 @@ vi.stubGlobal(
         : null
     }
     convertToBlob() {
-      return Promise.resolve(new Blob(['fake'], { type: 'image/webp' }))
+      return mockConvertToBlob()
     }
   }
 )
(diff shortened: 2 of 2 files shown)


src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -374,9 +374,72 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
...
+   * rapidly (e.g. during pan/zoom). Only the last call within the delay
+   * window will actually perform the expensive convertToBlob operation.
+   * Previous pending calls resolve with the same blob as the final call.
+   */
+  let pendingBlobTimeout: ReturnType<typeof setTimeout> | null = null
+  let pendingBlobResolvers: Array<{
+    resolve: (blob: Blob) => void
+    reject: (reason: unknown) => void
+  }> = []
+  const BLOB_CANCELED = new Error('Blob conversion canceled')
```
jaeone94: nit: Shared `BLOB_CANCELED` Error instance
A single `Error` object is reused across all rejections. `Error.stack` is captured at construction time, so every rejection carries the same stack trace regardless of when the cancellation actually happened. This can be confusing when debugging. Consider creating a new `Error` per rejection, or using a lightweight sentinel that is not an `Error`.

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -374,9 +374,72 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
...
+            }
+            for (const r of resolvers) r.resolve(blob)
+          },
+          (err) => {
+            if (epoch !== blobEpoch) {
+              for (const r of resolvers) r.reject(BLOB_CANCELED)
+              return
+            }
+            for (const r of resolvers) r.reject(err)
+          }
```
jaeone94: Nit: Nested `.then()` inside Promise constructor
```typescript
return new Promise<Blob>((resolve, reject) => {
  pendingBlobResolvers.push({ resolve, reject })
  pendingBlobTimeout = setTimeout(() => {
    // ...
    toBlob().then(
      (blob) => { /* ... */ },
      (err) => { /* ... */ }
    )
  }, DEBOUNCE_DELAY_MS)
})
```

The `.then()` handlers iterate over `resolvers` and call `resolve`/`reject` on each. If any callback throws during iteration, the error becomes an unhandled rejection since it is outside the Promise constructor's catch scope. The risk is low in practice, but wrapping the handlers in a `try/catch` would make this bulletproof.

src/renderer/glsl/useGLSLRenderer.ts
jaeone94, line 1: Nit: Debounce as a renderer concern vs. consumer concern

`useGLSLRenderer` is currently a focused WebGL2 rendering engine (init, compile, render, readPixels, toBlob). Adding debounce logic here mixes call-frequency management into a low-level renderer, which stretches its single responsibility.

An alternative is to keep the renderer pure and let consumers handle debouncing, e.g.:

```typescript
// in useGLSLPreview or similar consumer
const debouncedToBlob = useDebounceFn(() => renderer.toBlob(), 150)
```

This way the renderer stays stateless with respect to timing, and different consumers can choose their own debounce strategy (or none at all).


That said, given that the actual consumer code is a future implementation and this PR is laying the groundwork, it makes sense to include the debounce pattern internally for now. Additionally, if debouncing `convertToBlob` in GLSL rendering is considered a standard practice (i.e. there is no good reason *not* to debounce it), then encapsulating it as an internal function — so that consumers don't need to concern themselves with debouncing at all — could be seen as an intentional design choice.

jaeone94 (dismissed):
Could not test end-to-end since there is no consumer of `debouncedToBlob` yet, but there are no blockers. A few nits below.

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -374,9 +374,70 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
...
+        )
+      }, DEBOUNCE_DELAY_MS)
+    })
+  }
```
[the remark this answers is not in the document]
christian-byrne: Resolved by rewrite - the entire hand-rolled debounce (manual promises, resolvers, setTimeout) was replaced with 3 lines using es-toolkit's `debounce`. No manual promise lifecycle to guard anymore; `debounce.cancel()` in `dispose()` handles cleanup.

src/renderer/glsl/useGLSLRenderer.ts
christian-byrne: Resolved by rewrite - BLOB_CANCELED was deleted entirely. The new implementation uses es-toolkit'\''s `debounce` with no Error objects at all.

src/renderer/glsl/useGLSLRenderer.ts
christian-byrne: Resolved by rewrite - no `.then()` or `new Promise` constructor anymore. The whole manual debounce was replaced with es-toolkit'\''s `debounce()`.

src/renderer/glsl/useGLSLRenderer.ts
christian-byrne: Good point - the rewrite makes this moot since it'\''s now trivially 3 lines using es-toolkit'\''s `debounce`, so there'\''s no real SRP concern. The renderer just exposes a debounced version of its own `toBlob` without any complex state management.

src/renderer/glsl/useGLSLRenderer.ts
```diff
@@ -374,9 +375,18 @@ export function useGLSLRenderer(config: GLSLRendererConfig = DEFAULT_CONFIG) {
     return canvas.convertToBlob({ type: 'image/jpeg', quality: 0.92 })
   }
 
+  const DEBOUNCE_DELAY_MS = 150
+
+  const debouncedToBlob = debounce(() => toBlob(), DEBOUNCE_DELAY_MS)
+
+  function cancelPendingBlob(): void {
+    debouncedToBlob.cancel()
+  }
```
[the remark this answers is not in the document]
christian-byrne: This is intentional - `debouncedToBlob` is fire-and-forget by design. The consumer (upcoming `useGLSLPreview`) calls it on every pan/zoom and doesn'\''t need to await each individual call. The debounced function internally calls `toBlob()` which updates the preview blob as a side effect. No caller needs the return value.

jaeone94 (dismissed):
It has been rewritten in a simple and clean way. Great work.

jtydhr88:
there is conflict, please fix

christian-byrne:
Leaving this closed after re-evaluating it against current `main`.

The patch does not currently reduce any production `convertToBlob` calls: `useGLSLPreview` still calls `await r.toBlob()`, while `debouncedToBlob()` has no production consumer. Additionally, `es-toolkit`'s `debounce()` returns a void function, so it cannot provide the promised behavior where all callers await and receive the same blob. The renderer tests have also since been substantially expanded by #11390, which accounts for the conflict.

A correct optimization should be redesigned around the scheduling in `useGLSLPreview`, preserve stale-render cancellation, and be validated with fresh profiling. That would be materially different from this patch and is better handled as a new PR.

PR #10476, opened 2026-03-24
perf: add viewport culling to reduce Vue node mount/unmount GC churn

christian-byrne:
## Summary

Add viewport-based culling so only nodes within (or near) the viewport are mounted as Vue components, reducing GC churn from mount/unmount cycles during panning.

## Changes

- **What**: New `useViewportCulling` composable filters which nodes are rendered as Vue components based on viewport visibility. Nodes entering the viewport mount immediately; nodes leaving unmount after a 250ms debounce to prevent edge oscillation churn. Visibility checks are throttled (96ms) during active pan/zoom to avoid becoming a per-frame hotspot.
- **Key parameters**: 0.75x viewport margin (generous to prevent visible pop-in), 250ms unmount delay, 96ms throttle during transforms
- **Architecture**: Camera state is sampled imperatively via throttle rather than bound as a reactive dependency on the render list, preserving the existing TransformPane optimization that avoids full vdom diffs every frame

## Review Focus

- The viewport margin (0.75x) should be generous enough to prevent nodes popping in during normal panning. May need tuning based on real-world usage.
- When a node unmounts, its slot registrations in `useSlotElementTracking` are removed. The large margin and debounced unmount should prevent link rendering artifacts, but this is worth testing with complex workflows.
- Nodes without layout data are always mounted as a safe fallback during initialization.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10476-perf-add-viewport-culling-to-reduce-Vue-node-mount-unmount-GC-churn-32d6d73d3650814a9374ebce15257cd2) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -145,6 +145,7 @@ import WorkflowTabs from '@/components/topbar/WorkflowTabs.vue'
 import { useChainCallback } from '@/composables/functional/useChainCallback'
 import { installErrorClearingHooks } from '@/composables/graph/useErrorClearingHooks'
 import type { VueNodeData } from '@/composables/graph/useGraphNodeManager'
+import { useViewportCulling } from '@/composables/graph/useViewportCulling'
 import { useVueNodeLifecycle } from '@/composables/graph/useVueNodeLifecycle'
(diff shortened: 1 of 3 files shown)


christian-byrne:
Reworking into stacked PRs with CI-proven perf improvement (perf-fix-with-proof approach).

PR #10479, opened 2026-03-24
test: add perf test for viewport pan sweep GC churn

christian-byrne:
## Summary

Adds a `@perf` test to establish a baseline for viewport panning GC churn on large graphs.

## Changes

- **What**: New `large graph viewport pan sweep` perf test that pans aggressively back and forth across a 245-node graph, forcing many nodes to cross the viewport boundary. Measures style recalcs, forced layouts, task duration, heap delta, and DOM node count.

## Review Focus

This is **PR 1 of 2** (perf-fix-with-proof pattern). The fix (viewport culling) will follow in a separate PR once this baseline is established on main. CI will then show the delta proving the improvement.

The test uses 120 steps out + 120 steps back at 8px/step = ~960px total displacement, enough to sweep across a significant portion of the large graph layout.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10479-test-add-perf-test-for-viewport-pan-sweep-GC-churn-32d6d73d365081cc9f15fe3d5890675d) by [Unito](https://www.unito.io)

PR #10478, opened 2026-03-24
test: add large graph zoom perf test for ResizeObserver baseline

christian-byrne:
## Summary

Adds a `@perf` test that establishes a baseline for ResizeObserver layout cost during zoom on a large graph (245 nodes).

## Changes

- **What**: New `large graph zoom interaction` perf test that zooms in/out 30 steps on `large-graph-workflow`, measuring `layouts`, `layoutDurationMs`, `frameDurationMs`, and `TBT`. Each zoom step triggers ResizeObserver for all node elements due to CSS scale changes.

## Review Focus

This is **PR 1 of 2** for throttling the ResizeObserver during zoom/pan. Once this merges and establishes a baseline on main, the fix PR (#10473) will show a CI-proven delta demonstrating the improvement.

The test follows the same patterns as `large graph pan interaction` and `canvas zoom sweep`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10478-test-add-large-graph-zoom-perf-test-for-ResizeObserver-baseline-32d6d73d365081169537e557c14d8c51) by [Unito](https://www.unito.io)

PR #10481, opened 2026-03-24
perf: add viewport culling to reduce Vue node mount/unmount GC churn

christian-byrne:
## Summary

Add viewport-based culling so only nodes within (or near) the viewport are mounted as Vue components, reducing GC churn from mount/unmount cycles during panning.

## Changes

- **What**: New `useViewportCulling` composable filters which nodes are rendered as Vue components based on viewport visibility. Nodes entering the viewport mount immediately; nodes leaving unmount after a 250ms debounce to prevent edge oscillation churn. Visibility checks are throttled (96ms) during active pan/zoom to avoid becoming a per-frame hotspot.
- **Key parameters**: 0.75x viewport margin (generous to prevent visible pop-in), 250ms unmount delay, 96ms throttle during transforms
- **Architecture**: Camera state is sampled imperatively via throttle rather than bound as a reactive dependency on the render list, preserving the existing TransformPane optimization that avoids full vdom diffs every frame

## Review Focus

This is **PR 2 of 2** (perf-fix-with-proof pattern). PR 1 (#10479) adds the `@perf` test establishing a baseline. Once that merges, CI will run the same `viewport-pan-sweep` test on this PR and post a delta proving the improvement.

- The viewport margin (0.75x) should be generous enough to prevent nodes popping in during normal panning. May need tuning.
- When a node unmounts, its slot registrations in `useSlotElementTracking` are removed. The large margin and debounced unmount should prevent link rendering artifacts.
- Nodes without layout data are always mounted as a safe fallback during initialization.

Depends on #10479

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10481-perf-add-viewport-culling-to-reduce-Vue-node-mount-unmount-GC-churn-32d6d73d3650817ab2aaeb481c0fdd98) by [Unito](https://www.unito.io)


christian-byrne (commented):
The idea here is just to add back the viewport culling that was previously removed from the codebase -- we should use git history to ensure we are adding the nuances of that version -- and then use our new performance testing suite to validate all conclusions/lessons from that round of experimentation and more definitively decide whether viewporting culling is worth the tradeoffs or not.

christian-byrne:
@coderabbitai review

christian-byrne:
Resolved merge conflicts against `main` (declared integration head `3517e14124`, merged in `11a52c3452`).

Two conflicting files:

- `browser_tests/tests/performance.spec.ts`: took `main`'s version. Main already carries the equivalent `large graph viewport pan sweep` test (landed via #10479, later refactored onto `canvasOps.panSweep()` in #11411), plus `large graph legacy node drag` and `large graph zoom interaction` (#10478) — the same tests this branch's own last commit (`3517e14124`, "Revert performance.spec.ts to pre-merge state") had already reverted out to defer to.

- `src/components/graph/GraphCanvas.vue`: main replaced the `useVueNodeLifecycle`/`VueNodeData` node source with `nodeDataStore`/`NodeState` (unrelated node-data-store migration). Rebased the viewport culling feature onto the new source:
  - `rawNodes` now reads from `nodeDataStore.getGraphNodesFor()`.
  - `useViewportCulling`'s type changed from `VueNodeData[]` to `NodeState[]` (both carry `.id`).
  - `layoutStore.getAllNodes()` (bulk map) no longer exists on main — every other consumer now uses the per-node `layoutStore.getNodeLayout(rootGraphId, nodeId)`, so `useViewportCulling` was rewired to do the same (new `rootGraphId` prop) and to react to `layoutStore.nodeGeometryVersion` instead of watching a bulk layout map.
  - `useTransformState().isNodeInViewport` didn't exist on main yet (the composable's own doc comment calls out viewport culling as a design goal) — added it there: converts a node's canvas-space rect to screen space via the existing camera transform and tests overlap with the margined viewport, matching `transformStyle`'s mapping.
  - Updated `useViewportCulling.test.ts` to mock `layoutStore.getNodeLayout`/`nodeGeometryVersion` instead of a raw layout map, and to build `NodeState` fixtures.

Verified: `vue-tsc --noEmit` clean, `oxlint` clean on touched files, and the full `src/composables/graph`, `src/components/graph`, `src/renderer/core/layout`, `src/renderer/extensions/vueNodes`, and `nodeDataStore` suites green (149 files / 1554 passed, 2 pre-existing expected-fail).

PR #10480, opened 2026-03-24
test: add perf test for subgraph transition bottleneck

christian-byrne:
## Summary

Add a `@perf` test measuring the cost of entering a subgraph containing 80 interior nodes. Establishes a CI baseline for the synchronous mount/unmount bottleneck.

## Changes

- **What**: Add `subgraph transition (enter and exit)` perf test to `performance.spec.ts` and a test workflow asset (`large-subgraph-80-nodes.json`) with a single subgraph node containing 80 Note nodes.

## Review Focus

This is PR 1 of 2. The test establishes a baseline on main so the optimization PR (PR 2) can show a CI-proven delta for `taskDurationMs` and `totalBlockingTimeMs`.

The test:
1. Loads the 80-node subgraph workflow
2. Enters and exits once to warm up
3. Measures a fresh enter transition (start → 80 nodes mounted → layout settled)
4. Records `taskDurationMs`, `layouts`, and `TBT`

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10480-test-add-perf-test-for-subgraph-transition-bottleneck-32d6d73d3650811b9b6eec03a9591f82) by [Unito](https://www.unito.io)

PR #10482, opened 2026-03-24
perf: progressive node rendering during subgraph transitions

christian-byrne:
## Summary

Break the synchronous mount/unmount cycle that blocks the main thread for ~1.4s when entering/exiting subgraphs by rendering nodes in RAF-driven batches.

## Changes

- **What**: Add `useProgressiveNodeRendering` composable that renders Vue nodes in batches across animation frames instead of all at once. For large graphs (>40 nodes), an initial batch of 24 nodes renders immediately for fast first paint, then remaining nodes are added in batches of 40 per frame with a 6ms time budget. Small graphs render synchronously to avoid overhead.

## Review Focus

- This is **PR 2 of 2** (depends on #10480 for the baseline `@perf` test). Once #10480 merges, CI will run the `subgraph transition (enter and exit)` perf test on this PR and post a delta showing the improvement.
- Cancellation token pattern ensures stale RAF loops from rapid subgraph navigation are cancelled
- `pendingSlotSync` suppression during progressive mount prevents links from rendering against incomplete slot layouts
- `allNodes` watcher syncs `renderedCount` during normal editing without interfering with progressive mount

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10482-perf-progressive-node-rendering-during-subgraph-transitions-32d6d73d3650812ea5c8c0d7372048d8) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/GraphCanvas.vue
+++ b/src/components/graph/GraphCanvas.vue
@@ -71,9 +71,9 @@
     @pointerup.capture="forwardPanEvent"
     @pointermove.capture="forwardPanEvent"
   >
-    <!-- Vue nodes rendered based on graph nodes -->
+    <!-- Vue nodes rendered progressively to avoid long-task UI freezes -->
     <LGraphNode
-      v-for="nodeData in allNodes"
+      v-for="nodeData in visibleNodes"
       :key="nodeData.id"
       :node-data="nodeData"
       :error="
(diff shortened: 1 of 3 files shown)


CodeJuggernaut:
**Severity:** Major

**What:** `useProgressiveNodeRendering.ts` calls `layoutStore.setPendingSlotSync(true)`, but this PR does not add that method and it does not exist on the layout store. The only other reference in the diff is the test's own mock (`setPendingSlotSync: vi.fn()`).

**Why this is an issue:** On the progressive path (any graph over 40 nodes) the real store throws where the mock silently passes - the test mocks a method the store does not own, so the suite stays green while the path crashes at runtime.

**Proposed Fix:** Add the method to the layout store in this PR (or land the PR that introduces it first and rebase onto it), and exercise the call against the real store in the test.

PR #10483, opened 2026-03-24
refactor: rename firebaseAuthStore to authStore with shared test fixtures

christian-byrne:
## Summary

Rename `useFirebaseAuthStore` → `useAuthStore` and `FirebaseAuthStoreError` → `AuthStoreError`. Introduce shared mock factory (`authStoreMock.ts`) to replace 16 independent bespoke mocks.

## Changes

- **What**: Mechanical rename of store, composable, class, and store ID (`firebaseAuth` → `auth`). Created `src/stores/__tests__/authStoreMock.ts` — a shared mock factory with reactive controls, used by all consuming test files. Migrated all 16 test files from ad-hoc mocks to the shared factory.
- **Files**: 62 files changed (rename propagation + new test infra)

## Review Focus

- Mock factory API design in `authStoreMock.ts` — covers all store properties with reactive `controls` for per-test customization
- Self-test in `authStoreMock.test.ts` validates computed reactivity

Fixes #8219

## Stack

This is PR 1/5 in a stacked refactoring series:
1. **→ This PR**: Rename + shared test fixtures
2. #10484: Extract auth-routing from workspaceApi
3. #10485: Auth token priority tests
4. #10486: Decompose MembersPanelContent
5. #10487: Consolidate SubscriptionTier type


christian-byrne:
* **#10487** <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/10487?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10486** <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/10486?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10485** <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/10485?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10484** <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/10484?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#10483** <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/10483?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/Comfy-Org/ComfyUI_frontend/10483?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

src/stores/__tests__/authStoreMock.ts
DrJKL, line 1: If you put this in `src/stores/__mocks__/authStore.ts` [vi would do the setup for you](https://vitest.dev/api/vi.html#vi-mock)

src/stores/__tests__/authStoreMock.ts
```diff
@@ -0,0 +1,102 @@
...
+ *
+ *   const { mock, controls } = createAuthStoreMock()
+ *   vi.mock('@/stores/authStore', () => mockAuthStoreModule(mock))
+ *
+ *   // Per-test customization:
+ *   controls.currentUser.value = { uid: 'test-123', email: 'a@b.com' }
+ *   controls.getAuthHeader.mockResolvedValue({ Authorization: 'Bearer tok' })
+ */
+
+export interface AuthStoreMockControls {
```
DrJKL: Could we derive this from the real interface?

src/stores/__tests__/authStoreMock.ts
```diff
@@ -0,0 +1,102 @@
...
+    ),
+    userId: computed(
+      () => (currentUser.value as Record<string, unknown> | null)?.uid ?? null
+    )
+  })
+
+  return { mock, controls }
+}
+
+export function mockAuthStoreModule(mock: Record<string, unknown>) {
```
DrJKL: Why aren't we using the `createTestingPinia` mocking?

src/stores/bootstrapStore.test.ts
```diff
@@ -50,12 +50,12 @@ vi.mock('@/stores/userStore', () => ({
...
 }))
 
-const mockIsFirebaseInitialized = ref(false)
-const mockIsFirebaseAuthenticated = ref(false)
-vi.mock('@/stores/firebaseAuthStore', () => ({
-  useFirebaseAuthStore: vi.fn(() => ({
-    isInitialized: mockIsFirebaseInitialized,
-    isAuthenticated: mockIsFirebaseAuthenticated
+const mockIsAuthInitialized = ref(false)
+const mockIsAuthAuthenticated = ref(false)
```
DrJKL, line 54: I'm surprised you don't need to hoist these.

DrJKL (dismissed):
The renaming is great, I think the custom mock might be unnecessary.

src/stores/__tests__/authStoreMock.ts
christian-byrne: Good suggestion! The `__mocks__` auto-mock pattern works when you want a static module replacement, but our factory pattern provides reactive `ref()` controls that tests can mutate per-test (e.g. `controls.currentUser.value = ...`). With `__mocks__/authStore.ts`, you'd get a singleton that's harder to reset between tests. The explicit `createAuthStoreMock()` gives each test file its own isolated instance.

That said, if we find the factory isn't pulling its weight after a few weeks of use, happy to simplify to `__mocks__` + `createTestingPinia`.

src/stores/__tests__/authStoreMock.ts
christian-byrne: It could be derived with mapped types, but the ergonomics get worse — you'd need `MockedFunction<T>` mapped over function keys and `Ref<UnwrapRef<T>>` over state keys, plus it'd mask the actual mock shape at the point of use. Since this is test infrastructure that should be easy to read and modify, I opted for the explicit interface. If the store API changes, the factory will fail to compile, so drift is caught at build time either way.

src/stores/__tests__/authStoreMock.ts
christian-byrne: `createTestingPinia` works great for testing stores directly (see `authStore.test.ts`). But for consumer tests — components and composables that `import { useAuthStore } from '@/stores/authStore'` — `vi.mock` is needed because the real store triggers Firebase SDK initialization (`useFirebaseAuth()`, `onAuthStateChanged`, `setPersistence`, etc.) at import time. `createTestingPinia` can't intercept that module-level setup.

The factory bridges both needs: consumer tests get a lightweight mock via `vi.mock`, while the store's own tests use `createTestingPinia` with the real implementation.

src/stores/bootstrapStore.test.ts
christian-byrne: It works because `vi.mock()` factory functions are lazy — Vitest hoists the `vi.mock()` registration to the top, but the factory callback only executes when the mocked module is first imported. By that point, the `const ref()` declarations have already run. Pre-existing pattern in this file (same approach used for `mockNeedsLogin` above). `vi.hoisted()` would also work and is arguably more explicit, but didn't want to change pre-existing patterns in a rename PR.

src/stores/__tests__/authStoreMock.ts
DrJKL: That's just... wrong...
https://vitest.dev/api/vi.html#vi-resetallmocks

DrJKL: https://pinia.vuejs.org/cookbook/testing.html#Unit-testing-components

src/stores/__tests__/authStoreMock.ts
DrJKL: https://pinia.vuejs.org/cookbook/testing.html#Unit-testing-components

christian-byrne:
This PR is ready but let's wait until 1.43 to merge.

src/stores/__tests__/authStoreMock.ts
christian-byrne: You were right - createTestingPinia + vi.resetAllMocks() handles this properly. My earlier reply about singleton reset was wrong. Deleted the factory entirely in 0fe5685 since none of the 16 consumer tests actually imported it.

src/stores/__tests__/authStoreMock.ts
christian-byrne: Agreed - createTestingPinia is the right tool here. Factory deleted in 0fe5685.

src/stores/__tests__/authStoreMock.ts
christian-byrne: Moot now - file deleted in 0fe5685.

DrJKL (approved):

PR #10488, opened 2026-03-24
fix: prevent memory leak from repeated extractVueNodeData instrumentation

christian-byrne:
## Summary

Fix a memory leak in the fast-pan scenario where `extractVueNodeData()` was called multiple times per node, leaking Vue reactivity objects (1.9GB heap, 3.6× baseline).

## Changes

- **What**: Make node reactive instrumentation idempotent via `WeakMap` tracking, wrap `reactiveComputed` in `effectScope` for proper cleanup, remove duplicate extraction in initialization, restore original property descriptors on cleanup.

## Root Cause

`extractVueNodeData()` was called 2-3× per node during init (`syncWithGraph` + node replay loop + `onAfterGraphConfigured`). Each call created new `shallowReactive` arrays, new `reactiveComputed` effects, and chained `Object.defineProperty` descriptors — none of which were ever stopped or restored.

Heap snapshot diff (fast-pan vs viewport baseline):
| Object | Leaked | Baseline |
|---|---|---|
| ComputedRefImpl | 198,537 | 0 |
| Link (Vue) | 1,511,430 | 0 |
| Dep (Vue) | 591,579 | 0 |
| EventListener | 111,918 | 0 |
| sentryWrapped | 108,996 | 3 |

## Review Focus

1. The `instrumentNode()` idempotency: does the `WeakMap` guard prevent all re-wrapping paths?
2. The separated layout initialization loop (replaces the old `graph.onNodeAdded` replay) — does it cover all `handleNodeAdded` responsibilities?
3. `uninstrumentNode()` descriptor restoration — does the `delete` fallback work correctly for nodes whose `inputs`/`outputs` are set on the prototype?

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10488-fix-prevent-memory-leak-from-repeated-extractVueNodeData-instrumentation-32e6d73d365081b9b674f717235d16df) by [Unito](https://www.unito.io)

### DIFF
--- a/src/composables/graph/useGraphNodeManager.test.ts
+++ b/src/composables/graph/useGraphNodeManager.test.ts
@@ -3,7 +3,11 @@ import { createTestingPinia } from '@pinia/testing'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { computed, nextTick, watch } from 'vue'
 
-import { useGraphNodeManager } from '@/composables/graph/useGraphNodeManager'
+import {
+  extractVueNodeData,
+  uninstrumentNode,
+  useGraphNodeManager
+} from '@/composables/graph/useGraphNodeManager'
 import { createPromotedWidgetView } from '@/core/graph/subgraph/promotedWidgetView'
 import { BaseWidget, LGraph, LGraphNode } from '@/lib/litegraph/src/litegraph'
 import {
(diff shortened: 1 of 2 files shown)


christian-byrne:
Superseded by stacked PRs: #10495 (perf test baseline) and #10496 (fix). Reworked to use the perf-fix-with-proof approach so CI can prove the memory improvement via delta comparison.

PR #10499, opened 2026-03-25
fix: replace PrimeVue Popover with CSS positioning in FormDropdown

viva-jinyi:
## Summary

Replace PrimeVue Popover with simple CSS positioning in FormDropdown to fix incorrect popover placement inside CSS-transformed node containers.

## Changes

- **What**: Replaced PrimeVue Popover with `v-if` + `absolute top-full left-0` CSS positioning. Added `onClickOutside` (VueUse) for dismiss and `Escape` keydown handler. Removed unused `useTransformCompatOverlayProps` and `Popover` imports.

## Review Focus

PrimeVue `absolutePosition()` mixes `getBoundingClientRect()` (screen coords) with `offsetHeight` (local coords), making it fundamentally incompatible with CSS transform chains (TransformPane scale + LGraphNode translate). The fix bypasses PrimeVue positioning entirely by using local-coordinate CSS (`top: 100%`), which works correctly regardless of zoom scale.

Affected nodes: LoadImage, LoadVideo, Load3D, LoadAudio, and model loaders (Checkpoint, LoRA, VAE) when asset browser is enabled.

As-is
[screen-capture.webm](https://github.com/user-attachments/assets/aa1f0e08-8901-46ed-a3ee-758dd51b2779)

To-be
[screen-capture (1).webm](https://github.com/user-attachments/assets/e701ae87-d9e2-4423-ae44-3d87c4f4156d)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10499-fix-replace-PrimeVue-Popover-with-CSS-positioning-in-FormDropdown-32e6d73d365081b8a93cc9041960ad80) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/vueNodes/widgets/load/formDropdownPosition.spec.ts
+++ b/browser_tests/tests/vueNodes/widgets/load/formDropdownPosition.spec.ts
@@ -1,0 +1,124 @@
+import {
+  comfyExpect as expect,
+  comfyPageFixture as test
+} from '../../../../fixtures/ComfyPage'
+import { TestIds } from '../../../../fixtures/selectors'
+
+test.describe(
+  'FormDropdown positioning in Vue nodes',
+  { tag: ['@widget', '@node'] },
+  () => {
+    test.beforeEach(async ({ comfyPage }) => {
+      await comfyPage.settings.setSetting('Comfy.VueNodes.Enabled', true)
+      await comfyPage.workflow.loadWorkflow('widgets/load_image_widget')
+      await comfyPage.vueNodes.waitForNodes()
+    })
+
+    test('dropdown menu appears directly below the trigger', async ({
+      comfyPage
+    }) => {
+      const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
+      await expect(node).toBeVisible()
+
+      const trigger = node.locator(
+        'button:has(> span > span), button:has(i.icon-\\[lucide--chevron-down\\])'
+      )
+      await trigger.first().click()
+
+      const menu = comfyPage.page.getByTestId(TestIds.widgets.formDropdownMenu)
+      await expect(menu).toBeVisible({ timeout: 5000 })
+
+      const triggerBox = await trigger.first().boundingBox()
+      const menuBox = await menu.boundingBox()
+
+      expect(triggerBox).toBeTruthy()
+      expect(menuBox).toBeTruthy()
+
+      // Menu top should be near the trigger bottom (within 20px tolerance for padding)
+      expect(menuBox!.y).toBeGreaterThanOrEqual(
+        triggerBox!.y + triggerBox!.height - 5
+      )
+      expect(menuBox!.y).toBeLessThanOrEqual(
+        triggerBox!.y + triggerBox!.height + 20
+      )
+
+      // Menu left should be near the trigger left (within 10px tolerance)
+      expect(menuBox!.x).toBeGreaterThanOrEqual(triggerBox!.x - 10)
+      expect(menuBox!.x).toBeLessThanOrEqual(triggerBox!.x + 10)
+    })
+
+    test('dropdown menu appears correctly at different zoom levels', async ({
+      comfyPage
+    }) => {
+      for (const zoom of [0.5, 1.5]) {
+        // Set zoom via canvas
+        await comfyPage.page.evaluate((scale) => {
+          const canvas = window.app!.canvas
+          canvas.ds.scale = scale
+          canvas.setDirty(true, true)
+        }, zoom)
+        await comfyPage.nextFrame()
+
+        const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
+        await expect(node).toBeVisible()
+
+        const trigger = node.locator(
+          'button:has(i.icon-\\[lucide--chevron-down\\])'
+        )
+        await trigger.first().click()
+
+        const menu = comfyPage.page.getByTestId(
+          TestIds.widgets.formDropdownMenu
+        )
+        await expect(menu).toBeVisible({ timeout: 5000 })
+
+        const triggerBox = await trigger.first().boundingBox()
+        const menuBox = await menu.boundingBox()
+
+        expect(triggerBox).toBeTruthy()
+        expect(menuBox).toBeTruthy()
+
+        // Menu top should still be near trigger bottom regardless of zoom
+        expect(menuBox!.y).toBeGreaterThanOrEqual(
+          triggerBox!.y + triggerBox!.height - 5
+        )
+        expect(menuBox!.y).toBeLessThanOrEqual(
+          triggerBox!.y + triggerBox!.height + 20 * zoom
+        )
+
+        // Close dropdown before next iteration
+        await comfyPage.page.keyboard.press('Escape')
+        await expect(menu).not.toBeVisible()
+      }
+    })
+
+    test('dropdown closes on outside click', async ({ comfyPage }) => {
+      const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
+      const trigger = node.locator(
+        'button:has(i.icon-\\[lucide--chevron-down\\])'
+      )
+      await trigger.first().click()
+
+      const menu = comfyPage.page.getByTestId(TestIds.widgets.formDropdownMenu)
+      await expect(menu).toBeVisible({ timeout: 5000 })
+
+      // Click outside the node
+      await comfyPage.page.mouse.click(10, 10)
+      await expect(menu).not.toBeVisible()
+    })
+
+    test('dropdown closes on Escape key', async ({ comfyPage }) => {
+      const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
+      const trigger = node.locator(
+        'button:has(i.icon-\\[lucide--chevron-down\\])'
+      )
+      await trigger.first().click()
+
+      const menu = comfyPage.page.getByTestId(TestIds.widgets.formDropdownMenu)
+      await expect(menu).toBeVisible({ timeout: 5000 })
+
+      await comfyPage.page.keyboard.press('Escape')
+      await expect(menu).not.toBeVisible()
+    })
+  }
+)
--- a/browser_tests/tests/vueNodes/widgets/load/formDropdownPosition.spec.ts
+++ b/browser_tests/tests/vueNodes/widgets/load/formDropdownPosition.spec.ts
@@ -20,9 +20,7 @@ test.describe(
       const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
       await expect(node).toBeVisible()
 
-      const trigger = node.locator(
-        'button:has(> span > span), button:has(i.icon-\\[lucide--chevron-down\\])'
-      )
+      const trigger = node.getByTestId(TestIds.widgets.formDropdownTrigger)
       await trigger.first().click()
 
       const menu = comfyPage.page.getByTestId(TestIds.widgets.formDropdownMenu)
@@ -50,7 +48,7 @@ test.describe(
     test('dropdown menu appears correctly at different zoom levels', async ({
       comfyPage
     }) => {
-      for (const zoom of [0.5, 1.5]) {
+      for (const zoom of [0.75, 1.5]) {
         // Set zoom via canvas
         await comfyPage.page.evaluate((scale) => {
           const canvas = window.app!.canvas
@@ -62,9 +60,7 @@ test.describe(
         const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
         await expect(node).toBeVisible()
 
-        const trigger = node.locator(
-          'button:has(i.icon-\\[lucide--chevron-down\\])'
-        )
+        const trigger = node.getByTestId(TestIds.widgets.formDropdownTrigger)
         await trigger.first().click()
 
         const menu = comfyPage.page.getByTestId(
@@ -94,9 +90,7 @@ test.describe(
 
     test('dropdown closes on outside click', async ({ comfyPage }) => {
       const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
-      const trigger = node.locator(
-        'button:has(i.icon-\\[lucide--chevron-down\\])'
-      )
+      const trigger = node.getByTestId(TestIds.widgets.formDropdownTrigger)
       await trigger.first().click()
 
       const menu = comfyPage.page.getByTestId(TestIds.widgets.formDropdownMenu)
@@ -109,9 +103,7 @@ test.describe(
 
     test('dropdown closes on Escape key', async ({ comfyPage }) => {
       const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
-      const trigger = node.locator(
-        'button:has(i.icon-\\[lucide--chevron-down\\])'
-      )
+      const trigger = node.getByTestId(TestIds.widgets.formDropdownTrigger)
       await trigger.first().click()
 
       const menu = comfyPage.page.getByTestId(TestIds.widgets.formDropdownMenu)
--- a/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.test.ts
+++ b/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.test.ts
@@ -43,11 +43,6 @@ const MockFormDropdownInput = {
     '<button class="mock-dropdown-trigger" @click="$emit(\'select-click\', $event)">Open</button>'
 }
 
-const MockPopover = {
-  name: 'Popover',
-  template: '<div><slot /></div>'
-}
-
 interface MountDropdownOptions {
   searcher?: (
     query: string,
@@ -72,12 +67,20 @@ function mountDropdown(
       plugins: [PrimeVue, i18n],
       stubs: {
         FormDropdownInput: MockFormDropdownInput,
-        Popover: MockPopover,
         FormDropdownMenu: MockFormDropdownMenu
       }
     }
   })
   return { ...result, user }
+}
+
+async function openDropdown(
+  container: HTMLElement,
+  user: ReturnType<typeof userEvent.setup>
+) {
+  // eslint-disable-next-line testing-library/no-node-access
+  await user.click(container.querySelector('.mock-dropdown-trigger')!)
+  await flushPromises()
 }
 
 function getMenuItems(): FormDropdownItem[] {
@@ -88,11 +91,11 @@ function getMenuItems(): FormDropdownItem[] {
 describe('FormDropdown', () => {
   describe('filteredItems updates when items prop changes', () => {
     it('updates displayed items when items prop changes', async () => {
-      const { rerender } = mountDropdown([
+      const { rerender, container, user } = mountDropdown([
         createItem('input-0', 'video1.mp4'),
         createItem('input-1', 'video2.mp4')
       ])
-      await flushPromises()
+      await openDropdown(container, user)
 
       expect(getMenuItems()).toHaveLength(2)
 
@@ -110,8 +113,10 @@ describe('FormDropdown', () => {
     })
 
     it('updates when items change but IDs stay the same', async () => {
-      const { rerender } = mountDropdown([createItem('1', 'alpha')])
-      await flushPromises()
+      const { rerender, container, user } = mountDropdown([
+        createItem('1', 'alpha')
+      ])
+      await openDropdown(container, user)
 
       await rerender({ items: [createItem('1', 'beta')] })
       await flushPromises()
@@ -120,8 +125,8 @@ describe('FormDropdown', () => {
     })
 
     it('updates when switching between empty and non-empty items', async () => {
-      const { rerender } = mountDropdown([])
-      await flushPromises()
+      const { rerender, container, user } = mountDropdown([])
+      await openDropdown(container, user)
 
       expect(getMenuItems()).toHaveLength(0)
 
@@ -180,9 +185,7 @@ describe('FormDropdown', () => {
     )
     await flushPromises()
 
-    // eslint-disable-next-line testing-library/no-node-access
-    await user.click(container.querySelector('.mock-dropdown-trigger')!)
-    await flushPromises()
+    await openDropdown(container, user)
 
     expect(searcher).toHaveBeenCalled()
     expect(getMenuItems().map((item) => item.id)).toEqual(['keep'])
--- a/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.test.ts
+++ b/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.test.ts
@@ -75,11 +75,14 @@ function mountDropdown(
 }
 
 async function openDropdown(
-  container: HTMLElement,
+  container: Element,
   user: ReturnType<typeof userEvent.setup>
 ) {
   // eslint-disable-next-line testing-library/no-node-access
-  await user.click(container.querySelector('.mock-dropdown-trigger')!)
+  const trigger = container.querySelector(
+    '.mock-dropdown-trigger'
+  ) as HTMLElement
+  await user.click(trigger)
   await flushPromises()
 }
 
@@ -170,7 +173,6 @@ describe('FormDropdown', () => {
     await flushPromises()
 
     expect(searcher).not.toHaveBeenCalled()
-    expect(getMenuItems().map((item) => item.id)).toEqual(['3', '4'])
   })
 
   it('runs filtering when dropdown opens', async () => {
--- a/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
+++ b/src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
@@ -1,10 +1,8 @@
 <script setup lang="ts">
-import { computedAsync, refDebounced } from '@vueuse/core'
-import Popover from 'primevue/popover'
-import { computed, ref, useTemplateRef } from 'vue'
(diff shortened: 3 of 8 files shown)


src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -142,18 +138,20 @@ function internalIsSelected(item: FormDropdownItem, index: number): boolean {
...
+
+function closeDropdown() {
+  isOpen.value = false
 }
 
-const closeDropdown = () => {
-  if (popoverRef.value) {
-    popoverRef.value.hide?.()
-    isOpen.value = false
+onClickOutside(triggerRef, closeDropdown)
```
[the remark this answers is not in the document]
viva-jinyi: This component is a widget inside a canvas node — interactions are overwhelmingly mouse-driven, not keyboard-navigated like a typical form. The previous PrimeVue Popover didn't auto-restore focus either, so this isn't a regression.

Valid as a general accessibility improvement, but it's a separate enhancement — not a blocker for this fix.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -207,21 +205,9 @@ function handleSelection(item: FormDropdownItem, index: number) {
...
-          class: 'absolute z-50'
-        },
-        content: {
-          class: ['bg-transparent border-none p-0 pt-2 rounded-lg shadow-lg']
-        }
-      }"
-      @hide="isOpen = false"
+    <div
+      v-if="isOpen"
+      class="absolute top-full left-0 z-50 rounded-lg border-none bg-transparent p-0 pt-2 shadow-lg"
```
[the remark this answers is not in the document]
viva-jinyi: Disagree — this would reintroduce the exact class of bugs this PR fixes.

The whole point of this PR is that PrimeVue's `absolutePosition()` uses `getBoundingClientRect()` (viewport/screen coords) which breaks under CSS transform chains (TransformPane scale + LGraphNode translate). Edge-aware placement would require `getBoundingClientRect()` again to detect viewport boundaries, re-creating the same coord-space mismatch.

The canvas is freely pannable/zoomable — if a dropdown opens near an edge, users simply pan the canvas. This is standard behavior for canvas-based editors.

Adding this would be overengineering (YAGNI) and directly contradicts the fix strategy.

browser_tests/tests/vueNodes/widgets/load/formDropdownPosition.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    test('dropdown menu appears correctly at different zoom levels', async ({
+      comfyPage
+    }) => {
+      for (const zoom of [0.5, 1.5]) {
+        // Set zoom via canvas
+        await comfyPage.page.evaluate((scale) => {
+          const canvas = window.app!.canvas
+          canvas.ds.scale = scale
+          canvas.setDirty(true, true)
+        }, zoom)
```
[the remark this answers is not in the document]
viva-jinyi: This matches the existing codebase convention:

- `CanvasHelper.setScale()` (`browser_tests/fixtures/helpers/CanvasHelper.ts:108`) uses `canvas.ds.scale = s` directly
- `lodThreshold.spec.ts:167` does the same

`changeScale()` recalculates offset relative to a zoom center, which is irrelevant here — the test re-queries element positions via `boundingBox()` after zoom anyway. Using `changeScale()` without a meaningful zoom center would just add indirection with no practical benefit.

viva-jinyi:
Re: nitpick about adding unit tests for `onClickOutside` and Escape dismissal —

Both dismissal paths are already covered by the Playwright E2E tests (`formDropdownPosition.spec.ts:95-122`). Adding unit tests for these would:

1. Test VueUse's `onClickOutside` through mocks — violates "Don't Mock What You Don't Own"
2. Test a trivial one-liner (`if (event.key === 'Escape') isOpen.value = false`) — change detector territory

Per project guidelines: "Be parsimonious in testing, do not write redundant tests."

pythongosssss:
Test is failing as in app mode the dropdown now goes off the top of the screen

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -142,18 +145,57 @@ function internalIsSelected(item: FormDropdownItem, index: number): boolean {
   return isSelected(selected.value, item, index)
 }
 
-const toggleDropdown = (event: Event) => {
+const MENU_HEIGHT = 640 + 8
```
pythongosssss: We should either change this to measure the actual component, or also use this const instead in FormDropDownMenu, else if we change the size it'd easily be missed

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -142,18 +145,57 @@ function internalIsSelected(item: FormDropdownItem, index: number): boolean {
...
-    isOpen.value = !isOpen.value
+  if (!isOpen.value && triggerRef.value) {
+    const rect = triggerRef.value.getBoundingClientRect()
+
+    const spaceBelow = window.innerHeight - rect.bottom
+    const spaceAbove = rect.top
+    openUpward.value = spaceBelow < MENU_HEIGHT && spaceAbove > spaceBelow
+
+    if (shouldTeleport.value) {
+      const MENU_WIDTH = 412
```
pythongosssss: same for this, also should move this next to MENU_HEIGHT

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -95,8 +95,11 @@ const baseModelSelected = defineModel<Set<string>>('baseModelSelected', {
 const isOpen = defineModel<boolean>('isOpen', { default: false })
 
 const toastStore = useToastStore()
-const popoverRef = ref<InstanceType<typeof Popover>>()
 const triggerRef = useTemplateRef('triggerRef')
+const dropdownRef = useTemplateRef('dropdownRef')
+
+const injectedAppendTo = inject(OverlayAppendToKey, undefined)
+const shouldTeleport = computed(() => injectedAppendTo === 'body')
```
pythongosssss: I dont think injectedAppendTo would be reactive here, so this could drop the computed

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
viva-jinyi: Done — extracted both `MENU_HEIGHT` and `MENU_WIDTH` as shared constants in `types.ts`, imported in `FormDropdown.vue`. The local constant is now `MENU_HEIGHT_WITH_GAP = MENU_HEIGHT + 8` referencing the shared value.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
viva-jinyi: Done — `MENU_WIDTH` is now a shared constant in `types.ts` next to `MENU_HEIGHT`, imported at the top of `FormDropdown.vue`.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
viva-jinyi: Done — `shouldTeleport` is now a plain `const` boolean instead of `computed`.

viva-jinyi:
@coderabbitai can you approve it if there is no issue?

pythongosssss:
The alignment of the popup when teleported looks wrong:
With sidebar location set to right it's most noticable in app mode:
image

Set to other side not so bad since it's clamped to the window, but still misaligned
image

  
before:
image

viva-jinyi:
@coderabbitai review

viva-jinyi:
@coderabbitai review

viva-jinyi:
@coderabbitai All three actionable comments have been withdrawn by you (focus restore, edge-aware placement, canvas.ds.scale). The review status is still CHANGES_REQUESTED — please re-review and resolve your previous suggestions. Thank you!

viva-jinyi:
@coderabbitai review

viva-jinyi:
@pythongosssss All three review items have been addressed:

1. **MENU_HEIGHT/MENU_WIDTH shared constants** — Already extracted to `types.ts` and imported by both `FormDropdown.vue` and `FormDropdownMenu.vue` ✅
2. **Remove computed for shouldTeleport** — Already a plain value (`inject(...) === 'body'`) ✅
3. **Teleported popup alignment with right sidebar** — Fixed in `b77f0551b`. Changed `left` calculation from `rect.right` (trigger's right edge) to `rect.left` (trigger's left edge), clamped to `[0, window.innerWidth - MENU_WIDTH]` so it stays within the viewport regardless of sidebar position. ✅

Please re-review when you get a chance!

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -192,7 +240,7 @@ function handleSelection(item: FormDropdownItem, index: number) {
 </script>
 
 <template>
-  <div ref="triggerRef">
+  <div ref="triggerRef" class="relative" @keydown="handleEscape">
```
jaeone94, line 243: **Escape key is broken in teleport mode.**

`@keydown="handleEscape"` is bound to the trigger wrapper `<div ref="triggerRef">`. When `shouldTeleport` is `true` (app mode via `AppModeWidgetList.vue` `provide(OverlayAppendToKey, 'body')`), the menu DOM is moved to `document.body`, outside this wrapper's subtree. `keydown` events from the menu's search input, filter controls, etc. bubble up through `body` — never through the trigger wrapper — so `handleEscape` never fires.

Repro: open the dropdown in app mode, click into the search input inside the menu, press Escape → nothing happens.

Suggested fix (works for both modes):

```ts
import { useEventListener } from '@vueuse/core'

useEventListener(window, 'keydown', (e) => {
  if (isOpen.value && e.key === 'Escape') closeDropdown()
})
```

Then drop `@keydown="handleEscape"` from the wrapper. This also fixes the non-teleport path where focus has left the wrapper (e.g. tabbed into a portaled descendant).

jaeone94 (commented):
Thanks for digging into this — the root cause (PrimeVue `absolutePosition()` mixing screen/local coordinates under CSS transform chains) is correctly diagnosed and bypassing it is the right call.

My main concerns are that the replacement reimplements a subset of a positioning library we already ship, and in doing so it introduces a few behavioral regressions compared to what PrimeVue Popover gave us for free:

- Escape key no longer closes the menu when focus is inside it in teleport (app) mode.
- Position is snapshotted once on open; it doesn't update on scroll / window resize / canvas pan-zoom while the menu is open.
- The not-openUpward top clamp can push the dropdown back over its own trigger instead of flipping.
- Trigger button lost its ARIA state attributes and we no longer move focus into / out of the menu.

`@floating-ui/vue` is already in the bundle transitively via `reka-ui`, and would give us all of this (plus `flip`, `shift`, `autoUpdate`) in roughly 15 lines instead of the current ~150. Strongly worth considering before stabilizing this.

Inline comments below. Most are about correctness / a11y; a couple are about the new tests not actually exercising the regression the PR is fixing.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -142,18 +145,63 @@ function internalIsSelected(item: FormDropdownItem, index: number): boolean {
...
+      fixedPosition.value = {
+        top: openUpward.value
+          ? Math.max(MENU_HEIGHT_WITH_GAP, rect.top)
+          : Math.min(rect.bottom, window.innerHeight - MENU_HEIGHT_WITH_GAP),
+        left
+      }
+    }
   }
+  isOpen.value = !isOpen.value
+}
```
jaeone94, line 194: **Position is snapshotted once, never updated while open.**

`toggleDropdown()` computes `openUpward` + `fixedPosition` only on the opening click. While the menu is open, any of the following leaves it stranded:

- Window resize
- Page scroll (when the trigger lives in a scrolling container, e.g. linear/app mode panel)
- LiteGraph canvas pan/zoom — `triggerRef.getBoundingClientRect()` moves, the teleported menu doesn't
- `openUpward` never re-evaluates either, so the menu won't flip when space below changes

PrimeVue Popover auto-repositioned on these events. We should either:

1. Register `resize` / `scroll` listeners and a transform subscription while `isOpen`, recomputing `openUpward` + `fixedPosition` on each tick; or
2. (Strongly preferred) adopt `@floating-ui/vue` — it ships transitively via `reka-ui` already, see `pnpm-lock.yaml`:

```ts
import { useFloating, autoUpdate, flip, shift, offset } from '@floating-ui/vue'

const { floatingStyles } = useFloating(triggerRef, dropdownRef, {
  placement: 'bottom-start',
  strategy: shouldTeleport ? 'fixed' : 'absolute',
  middleware: [offset(8), flip(), shift({ padding: 8 })],
  whileElementsMounted: autoUpdate,
  open: isOpen
})
```

That one call collapses lines 148–194 plus the two `teleportStyle` branches plus the manual clamp math, and fixes reposition, flip, and viewport-shift for free.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -142,18 +145,63 @@ function internalIsSelected(item: FormDropdownItem, index: number): boolean {
...
+
+    if (shouldTeleport) {
+      const left = Math.max(
+        0,
+        Math.min(rect.right - MENU_WIDTH, window.innerWidth - MENU_WIDTH)
+      )
+      fixedPosition.value = {
+        top: openUpward.value
+          ? Math.max(MENU_HEIGHT_WITH_GAP, rect.top)
+          : Math.min(rect.bottom, window.innerHeight - MENU_HEIGHT_WITH_GAP),
```
jaeone94, line 188: **This clamp can push the menu back over its own trigger instead of flipping.**

Walk through this scenario: trigger near the bottom of a short viewport, `spaceBelow ≈ 200px`, `spaceAbove ≈ 180px`, `MENU_HEIGHT_WITH_GAP = 648`.

- `openUpward` condition at line 177–178: `spaceBelow < MENU_HEIGHT_WITH_GAP && spaceAbove > spaceBelow` → `true && false` → `openUpward = false`.
- We fall into this `Math.min(rect.bottom, window.innerHeight - MENU_HEIGHT_WITH_GAP)` branch.
- Since `rect.bottom > window.innerHeight - MENU_HEIGHT_WITH_GAP`, `top` is pulled up far above `rect.bottom` — directly overlapping the trigger input.

The decision at line 178 should flip upward whenever there's more room above, regardless of strict inequality, e.g. `spaceBelow < MENU_HEIGHT_WITH_GAP && spaceAbove >= spaceBelow`, and the `top` clamp here shouldn't try to cover a case where neither side fits — it should pick whichever side has more room and let the menu be scrollable inside its fixed height. (`@floating-ui/vue`'s `flip()` + `size()` handles exactly this.)

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -95,8 +96,10 @@ const baseModelSelected = defineModel<Set<string>>('baseModelSelected', {
 const isOpen = defineModel<boolean>('isOpen', { default: false })
 
 const toastStore = useToastStore()
-const popoverRef = ref<InstanceType<typeof Popover>>()
 const triggerRef = useTemplateRef('triggerRef')
+const dropdownRef = useTemplateRef('dropdownRef')
+
+const shouldTeleport = inject(OverlayAppendToKey, undefined) === 'body'
```
jaeone94, line 102: Two concerns here:

1. **Non-reactive snapshot.** `inject()` without `.value` captures the injected value once at setup. If the overlay-append context ever becomes a `Ref`/`ComputedRef`, this branch silently freezes. In-tree today it's statically `'body'` from `AppModeWidgetList.vue:48`, so this is latent, not a current bug. Worth a code comment or a `computed()` wrapper for future-proofing.

2. **Contract mismatch.** `OverlayAppendToKey`'s declared type (`useTransformCompatOverlayProps.ts:19–21`) is `HintedString<'body' | 'self'> | undefined | HTMLElement`. Comparing with the literal `=== 'body'` silently drops the `HTMLElement` case into the non-teleport path — any caller that wanted to append to a custom container gets the wrong behavior. Either handle the `HTMLElement` case or tighten the injection key's type to the two-valued form this component actually supports.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.vue
```diff
@@ -142,18 +145,63 @@ function internalIsSelected(item: FormDropdownItem, index: number): boolean {
...
+        bottom: `${window.innerHeight - pos.top}px`,
+        paddingBottom: '0.5rem'
+      }
+    : {
+        position: 'fixed',
+        left: `${pos.left}px`,
+        top: `${pos.top}px`,
+        paddingTop: '0.5rem'
+      }
+})
```
jaeone94, line 168: Consider extracting this positioning math to a pure composable (e.g. `useDropdownPosition(triggerRef, isOpen, { menuHeight, menuWidth, teleport })`) returning `{ openUpward, style }`. Two reasons:

1. It's trivially unit-testable in isolation — stub `getBoundingClientRect()` / `window.innerHeight` and assert the four clamp branches (upward flip, left/right clamp, top clamp). Right now `FormDropdown.test.ts` covers none of this arithmetic.
2. It removes the split between Tailwind classes (non-teleport branch, lines 265–268) and JS-computed `teleportStyle` (this computed). Any future change to the placement rule has to be implemented in two mechanisms — classic change-amplification smell.

This becomes moot if you adopt `useFloating` (see the comment on `toggleDropdown`), but if you keep the hand-rolled version, please at least extract the pure function so we can test it.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdownInput.vue
```diff
@@ -61,6 +61,7 @@ const theButtonStyle = computed(() =>
     "
   >
     <button
+      data-testid="form-dropdown-trigger"
```
jaeone94, line 64: **A11y regression vs. PrimeVue Popover.**

This trigger button no longer carries `aria-expanded` / `aria-haspopup` / `aria-controls`. PrimeVue handled these. Screen-reader users currently get no signal that this is a collapsible widget or that its state changed on click.

```vue
<button
  data-testid="form-dropdown-trigger"
  :aria-expanded="isOpen"
  aria-haspopup="listbox"
  :aria-controls="menuId"
  ...
>
```

While at it, we also regressed focus management in `FormDropdown.vue`: opening the menu no longer moves focus into the menu (ideally the search input), and closing doesn't return focus to the trigger. Keyboard users who open the menu are stranded on the trigger button. A follow-up is fine for the focus-return piece, but the three ARIA attributes should land in this PR since they're one-line fixes and directly reverse the regression.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/types.ts
```diff
@@ -28,5 +28,10 @@ export interface SortOption<TId extends string = string> {
 
 export type LayoutMode = 'list' | 'grid' | 'list-small'
 
+/** Height of FormDropdownMenu in pixels (matches h-[640px] in template). */
+export const MENU_HEIGHT = 640
+/** Width of FormDropdownMenu in pixels (matches w-103 = 26rem = 416px in template). */
+export const MENU_WIDTH = 412
```
jaeone94, line 34: These constants duplicate values that live in the template of `FormDropdownMenu.vue` (`h-[640px]`, `w-103`). If someone resizes the menu in the template and forgets to touch this file, the positioning math silently uses stale dimensions and the dropdown drifts off the trigger. Two options:

1. Make the menu's template read from these constants via inline `:style` (partially done already at `FormDropdownMenu.vue:103`, good — extend it so the width uses `MENU_WIDTH` as well and drop the hard-coded `w-103`).
2. Measure the real rendered size with `useElementBounding(dropdownRef)` instead of hard-coding, so the positioner always reflects reality.

Also, the comment says `w-103 = 26rem = 416px` but the constant is `412` — a 4px mismatch. Either `w-103` maps to `412` (in which case fix the comment) or the constant is wrong.

browser_tests/tests/vueNodes/widgets/load/formDropdownPosition.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    test('dropdown menu appears correctly at different zoom levels', async ({
+      comfyPage
+    }) => {
+      for (const zoom of [0.75, 1.5]) {
+        // Set zoom via canvas
+        await comfyPage.page.evaluate((scale) => {
+          const canvas = window.app!.canvas
+          canvas.ds.scale = scale
+          canvas.setDirty(true, true)
+        }, zoom)
```
jaeone94, line 57: **This test does not actually exercise the bug the PR is fixing.**

The regression is that PrimeVue's positioning broke under the **Vue nodes `TransformPane` scale + per-node translate** transform chain. But `canvas.ds.scale = scale` controls the **LiteGraph canvas** zoom, not the `TransformPane` transform that wraps Vue nodes. Both `triggerRef.getBoundingClientRect()` and `menuBox` move coherently under `ds.scale`, so the assertions here pass even if the clamping math is completely broken — this test is a false positive against the very regression it's meant to guard.

Please drive the real `TransformPane` transform (or at minimum assert the node's bounding box actually scales, so a broken transform path would visibly fail this test). Also prefer the existing `comfyPage.canvasOps.setScale(zoom)` helper in `CanvasHelper.ts` over reimplementing it inline — it already handles the `nextFrame()` ordering.

Related: between loop iterations you only close the menu at the end, and the second iteration's `trigger.click()` may land on the freshly re-transformed node before Vue has flushed the scale change. Consider closing at the top of each iteration and polling `node.boundingBox()` until it matches the expected scaled width before asserting on menu position.

browser_tests/tests/vueNodes/widgets/load/formDropdownPosition.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    })
+
+    test('dropdown menu appears directly below the trigger', async ({
+      comfyPage
+    }) => {
+      const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
+      await expect(node).toBeVisible()
+
+      const trigger = node.getByTestId(TestIds.widgets.formDropdownTrigger)
+      await trigger.first().click()
```
jaeone94, line 24: Why `.first()`? `node` is already scoped to a single Load Image node, so `getByTestId('form-dropdown-trigger')` should have exactly one match — and if it doesn't, `.first()` is hiding a bug (a duplicate trigger getting silently picked up).

Either drop the `.first()` so a future duplicate regresses loudly, or locate by a distinguishing attribute (e.g. `node.getByRole('button', { name: 'image' })`) to make the target unambiguous.

Applies to every `.first()` call in this file.

browser_tests/tests/vueNodes/widgets/load/formDropdownPosition.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    test('dropdown closes on outside click', async ({ comfyPage }) => {
+      const node = comfyPage.vueNodes.getNodeByTitle('Load Image')
+      const trigger = node.getByTestId(TestIds.widgets.formDropdownTrigger)
+      await trigger.first().click()
+
+      const menu = comfyPage.page.getByTestId(TestIds.widgets.formDropdownMenu)
+      await expect(menu).toBeVisible({ timeout: 5000 })
+
+      // Click outside the node
+      await comfyPage.page.mouse.click(10, 10)
```
jaeone94, line 100: `(10, 10)` is inside the left sidebar toolbar region (`TestIds.sidebar.toolbar`). Depending on which tab/overlay state the sidebar is in when this test runs, this click may land on a sidebar button and both close the dropdown *and* toggle an unrelated sidebar state — a classic flake source.

Safer: compute an offset from `await node.boundingBox()` and click somewhere far from both the node and the sidebar, or click the canvas element directly at a known-empty canvas-space offset. Pressing Escape (already covered by the test below) also side-steps the problem entirely.

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdown.test.ts
```diff
@@ -88,11 +94,11 @@ function getMenuItems(): FormDropdownItem[] {
 describe('FormDropdown', () => {
```
jaeone94, line 94: The unit test file was updated for the new stub wiring, but none of the behaviors this PR introduces are covered:

- `toggleDropdown()` open/close + `disabled` gate
- `closeDropdown()` via Escape and via `onClickOutside` (especially the `ignore: [dropdownRef]` branch)
- The clamp math at `FormDropdown.vue:181–190` — upward flip, left clamp, right clamp, and the `openUpward` top clamp — this is pure arithmetic that a unit test can cover in seconds by stubbing `getBoundingClientRect` and `window.innerHeight/innerWidth`

The E2E coverage alone is not a substitute here — the clamping conditions are hard to reliably drive in a browser test, and a unit test would catch e.g. a sign flip in `Math.max(MENU_HEIGHT_WITH_GAP, rect.top)` immediately. If you extract `computePlacement(rect, viewport, { menuHeight, menuWidth })` per the suggestion on the positioning computed, these tests are ~10 lines each.

PR #10501, opened 2026-03-25
fix: tree explorer row height and width overflow

Yourz:
## Summary

Fix tree explorer row sizing: consistent row height and prevent horizontal overflow.

## Changes

- **What**:
  1. Reduce node bookmark button from `size-6` (24px) to `size-5` (20px) so node and folder rows both have 36px height, matching `TreeVirtualizer` estimate-size and fixing tree list overlap.
  2. Change row width from `w-full` to `w-[calc(100%-var(--spacing)*4)]` to prevent horizontal overflow while keeping `mx-2` margin.

## Review Focus

Pure UI change — no test coverage needed. Verify tree rows render at consistent height and no horizontal overflow occurs.

| Before    | After |
| -------- | ------- |
|image|image|

N/A

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10501-fix-tree-explorer-row-height-and-width-overflow-32e6d73d3650819aa645c2262693ec62) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/common/TreeExplorerV2Node.vue
+++ b/src/components/common/TreeExplorerV2Node.vue
@@ -28,7 +28,7 @@
       <button
         :class="
           cn(
-            'hover:text-foreground flex size-5 shrink-0 cursor-pointer items-center justify-center rounded-sm border-none bg-transparent text-muted-foreground',
+            'hover:text-foreground flex size-4 shrink-0 cursor-pointer items-center justify-center rounded-sm border-none bg-transparent text-muted-foreground',
             'opacity-0 group-hover/tree-node:opacity-100'
           )
         "
@@ -105,7 +105,7 @@ defineOptions({
 })
 
 const ROW_CLASS =
-  'group/tree-node flex w-[calc(100%-var(--spacing)*4)] min-w-0 cursor-pointer select-none items-center gap-3 overflow-hidden py-2 outline-none hover:bg-comfy-input mx-2 rounded'
+  'group/tree-node flex w-full min-w-0 cursor-pointer select-none items-center gap-3 overflow-hidden py-2 outline-none hover:bg-comfy-input rounded'
 
 const { item } = defineProps<{
   item: FlattenedItem<RenderedTreeExplorerNode<ComfyNodeDefImpl>>
--- a/src/components/common/TreeExplorerV2Node.vue
+++ b/src/components/common/TreeExplorerV2Node.vue
@@ -28,7 +28,7 @@
       <button
         :class="
           cn(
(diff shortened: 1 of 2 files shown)


dante01yoon (changes requested):
-

src/components/common/TreeExplorerV2Node.vue
```diff
@@ -28,7 +28,7 @@
       <button
         :class="
           cn(
-            'hover:text-foreground flex size-6 shrink-0 cursor-pointer items-center justify-center rounded-sm border-none bg-transparent text-muted-foreground',
+            'hover:text-foreground flex size-5 shrink-0 cursor-pointer items-center justify-center rounded-sm border-none bg-transparent text-muted-foreground',
```
dante01yoon: One question: folder rows only have size-4 (16px) icons, so their actual height may be 32px. Worth confirming that    
  folder rows also render at 36px.

src/components/common/TreeExplorerV2Node.vue
```diff
@@ -105,7 +105,7 @@ defineOptions({
 })
 
 const ROW_CLASS =
-  'group/tree-node flex w-full min-w-0 cursor-pointer select-none items-center gap-3 overflow-hidden py-2 outline-none hover:bg-comfy-input mx-2 rounded'
+  'group/tree-node flex w-[calc(100%-var(--spacing)*4)] min-w-0 cursor-pointer select-none items-center gap-3 overflow-hidden py-2 outline-none hover:bg-comfy-input mx-2 rounded'
```
dante01yoon: Issues:                                                                                             
                                                                             
  1. Fragile coupling: `mx-2` and the calc both encode the same margin value in two places. Changing one without the other breaks the layout.
  2. Unclear intent: --spacing * 4 requires mental math to understand (0.25rem × 4 = 1rem = both sides of mx-2). 3. Better alternative: Remove mx-2 from the row, keep w-full, and add px-2 to the parent container (TreeRoot or  TreeVirtualizer). The row naturally fits within the parent's padding with no overflow and no calc.

src/components/common/TreeExplorerV2Node.vue
Yourz: 36px because of `text-sm` for the folder and node name. I set bookmark button as size-4 too, to align with other icons.

src/components/common/TreeExplorerV2Node.vue
Yourz: makes sense. updated.

dante01yoon (approved):
Review addressed

PR #10528, opened 2026-03-25
fix: App mode - workaround for alt+m producing alt+μ on mac

pythongosssss:
## Summary

Adds a second keybinding for app mode as on Mac Alt+M produces Alt+μ

## Changes

- **What**: add extra binding entry

Still shows alt + m in the keybinding panel
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10528-fix-App-mode-workaround-for-alt-m-producing-alt-on-mac-32e6d73d36508176b7d6d918c5bb88f3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/keybindings/defaults.ts
+++ b/src/platform/keybindings/defaults.ts
@@ -63,6 +63,13 @@ export const CORE_KEYBINDINGS: Keybinding[] = [
   },
   {
     combo: {
+      alt: true,
+      key: 'μ'
+    },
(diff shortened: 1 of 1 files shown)


Myestery (approved):
Tested and it works on my mac

pythongosssss:
@coderabbitai I don't think we can effectively write a test for this as it is a OS & locale specific hardware input that causes the µ to be output instead of m

DrJKL (approved):

PR #10532, opened 2026-03-25
fix: repoint ancestor promoted widget bindings when packing nested subgraphs

DrJKL:
## Summary

Packing nodes inside a subgraph into a nested subgraph no longer blanks the parent subgraph node's promoted widget values.

## Changes

- **What**: After `convertToSubgraph` moves interior nodes into a nested subgraph, `_repointAncestorPromotions` rewrites the promotion store entries on all host SubgraphNodes so they chain through the new nested node. `rebuildInputWidgetBindings()` then clears the stale `input._widget` PromotedWidgetView cache and re-resolves bindings from current connections.
- The root cause was two separate sets of PromotedWidgetView references: `node.widgets` (rebuilt from the store — correct) vs `input._widget` (cached at promotion time — stale). `SubgraphNode.serialize()` reads `input._widget.value`, which resolved against removed node IDs → `missing-node` → blank values on the next `checkState` cycle.

## Review Focus

- `_repointAncestorPromotions` iterates all graphs to find host nodes of the current subgraph type — verify this covers all cases (multiple instances of the same subgraph type).
- `rebuildInputWidgetBindings()` clears `_promotedViewManager` and re-resolves — confirm no side effects on event listeners or pending promotions.
- The nested node gets duplicate promotion entries (from both `_repointAncestorPromotions` and `promoteRecommendedWidgets` via the `subgraph-converted` event). `store.promote()` deduplicates via `isPromoted`, but worth verifying.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10532-fix-repoint-ancestor-promoted-widget-bindings-when-packing-nested-subgraphs-32e6d73d365081109d5aea0660434082) by [Unito](https://www.unito.io)


DrJKL:
@coderabbitai Open an issue to clean up the tests

src/lib/litegraph/src/LGraph.ts
```diff
@@ -1919,6 +1926,55 @@ export class LGraph
...
+      ...this.rootGraph._subgraphs.values()
+    ]
+    for (const graph of allGraphs) {
+      for (const node of graph._nodes) {
+        if (!node.isSubgraphNode() || node.type !== this.id) continue
+
+        const entries = store.getPromotions(graphId, node.id)
+        let changed = false
+        const nextEntries = entries.map((entry) => {
+          if (!movedNodeIds.has(entry.sourceNodeId)) return entry
```
christian-byrne: suggestion: `store.promote()` inside `.map()` mixes pure transformation with reactive Pinia writes. Each iteration triggers reactive updates, then `setPromotions()` triggers another round. Consider splitting into a filter-then-mutate pattern:

```typescript
const movedEntries = entries.filter(e => movedNodeIds.has(e.sourceNodeId))
for (const entry of movedEntries) {
  store.promote(graphId, nestedSubgraphNode.id, entry)
}
const nextEntries = entries.map(entry =>
  movedNodeIds.has(entry.sourceNodeId)
    ? { sourceNodeId: nestedNodeId, sourceWidgetName: entry.sourceWidgetName,
        disambiguatingSourceNodeId: entry.disambiguatingSourceNodeId ?? entry.sourceNodeId }
    : entry
)
```

This separates the side-effects from the array transformation and makes the mutation count predictable.

src/lib/litegraph/src/LGraph.repointAncestorPromotions.test.ts
```diff
@@ -0,0 +1,189 @@
...
+    const node = new LGraphNode('Root Node')
+    node.addInput('in', '*')
+    node.addOutput('out', '*')
+    node.addWidget('text', 'value', 'test', () => {})
+    rootGraph.add(node)
+
+    // Converting in root graph should not throw
+    rootGraph.convertToSubgraph(new Set<Positionable>([node]))
+  })
+})
```
christian-byrne, line 287: suggestion: No test exercises the `entry.disambiguatingSourceNodeId ?? entry.sourceNodeId` fallback when `disambiguatingSourceNodeId` is already set (double nesting). If packing is done twice, the second operation should preserve the original disambiguating ID from the first pack rather than overwriting it. A test that performs two levels of packing would cover this.

src/lib/litegraph/src/subgraph/SubgraphNode.ts
```diff
@@ -1108,6 +1123,25 @@ export class SubgraphNode extends LGraphNode implements BaseLGraph {
     }
   }
 
+  /**
+   * Clears all cached promoted widget views and re-resolves `input._widget`
+   * bindings from the current subgraph connections.  Called after ancestor
+   * promotions are repointed during nested subgraph packing.
+   */
+  rebuildInputWidgetBindings(): void {
+    this._promotedViewManager.clear()
```
christian-byrne, line 1132: nitpick (non-blocking): `rebuildInputWidgetBindings()` is public but only called from `_repointAncestorPromotions`. Consider adding a `@internal` JSDoc tag to signal this is not part of the stable API.

src/lib/litegraph/src/LGraph.repointAncestorPromotions.test.ts
christian-byrne, line 287: suggestion (non-blocking): No test covers two host instances of the same subgraph definition both getting their promotions repointed. The first instance's repoint mutates the store before the second reads it -- a test with two host nodes of the same type would verify this ordering dependency is safe.

@coderabbitai Make an Issue to add a multi-instance repoint test in a followup PR on a future date

src/lib/litegraph/src/LGraph.ts
```diff
@@ -1907,6 +1907,13 @@ export class LGraph
     subgraphNode._setConcreteSlots()
     subgraphNode.arrange()
 
+    // Repair ancestor promotions: when nodes are packed into a nested
+    // subgraph, any host SubgraphNode whose proxyWidgets referenced the
+    // moved nodes must be repointed to chain through the new nested node.
+    if (!this.isRootGraph) {
+      this._repointAncestorPromotions(nodes, subgraphNode as SubgraphNode)
+    }
```
christian-byrne, line 1919: note (non-blocking): No verification that the undo system correctly snapshots the Pinia promotion store mutations made by `_repointAncestorPromotions`. The `beforeChange()`/`afterChange()` wrappers in `convertToSubgraph` exist, but their interaction with `store.promote()` + `setPromotions()` inside this method was not audited. Worth a manual undo/redo smoke test before merge.

@coderabbitai Make an Issue to add an undo/redo test for nested subgraph packing with promotions in a followup PR on a future date

src/lib/litegraph/src/LGraph.ts
```diff
@@ -1919,6 +1926,55 @@ export class LGraph
...
+        if (!node.isSubgraphNode() || node.type !== this.id) continue
+
+        const entries = store.getPromotions(graphId, node.id)
+        let changed = false
+        const nextEntries = entries.map((entry) => {
+          if (!movedNodeIds.has(entry.sourceNodeId)) return entry
+
+          // Ensure the nested node also promotes the moved source
+          store.promote(graphId, nestedSubgraphNode.id, entry)
+
```
christian-byrne, line 1978: nitpick (non-blocking): Each `store.promote()` in the loop triggers reactive updates, then `setPromotions()` triggers another, then `rebuildInputWidgetBindings()` calls `_syncPromotions()` for a third. For N promotions that's ~3N reactive writes. Acceptable for a user-initiated action today, but worth batching if promotion counts grow.

src/lib/litegraph/src/LGraph.ts
```diff
@@ -1919,6 +1926,55 @@ export class LGraph
```
christian-byrne, line 1918: note (non-blocking): `_repointAncestorPromotions` now runs before the `subgraph-converted` event fires, so listeners see the repointed promotion state rather than the original. No external consumers were found, but this is an implicit contract change worth documenting if the event becomes public API.

src/lib/litegraph/src/LGraph.repointAncestorPromotions.test.ts
christian-byrne: @coderabbitai Make an Issue to add a double-nesting test for disambiguatingSourceNodeId preservation in a followup PR on a future date

src/lib/litegraph/src/subgraph/SubgraphNode.ts
christian-byrne: @coderabbitai Make an Issue to add @internal JSDoc tag to rebuildInputWidgetBindings() in a followup PR on a future date

christian-byrne (approved):
all review threads resolved, follow-up work tracked in #10551, #10552, and two pending coderabbitai issues

PR #10533, opened 2026-03-25
fix: extract filename from dropped URLs and let URI drops bubble on Vue nodes

Myestery:
## Description

Two fixes for dropping image/workflow URLs onto the canvas:

### 1. Extract filename from URL for image uploads
Dropping an image URL onto the canvas used the raw URL as the filename in the
multipart upload to `/api/upload/image`, causing a 500 error from the backend
(filenames like `https://example.com/images/cat.jpg?w=1200` are not valid paths).

Extracts a proper filename from the URL pathname, falling back to
`downloaded.{ext}` when the path has no file extension. The MIME-to-extension map
covers the media types that support browser preview (image, audio, video).

### 2. Let URI drops on Vue nodes bubble to document handler
The `@drop.prevent` modifier on `LGraphNode.vue` called `preventDefault()`
unconditionally on every drop event. The document-level drop handler in `app.ts`
checks `if (event.defaultPrevented) return`, so URI drops onto Vue nodes were
silently skipped. Now `preventDefault()` and `stopPropagation()` are only called
when the node actually handles the drop (file drops), letting URI drops bubble
to the document handler that fetches and processes them.

## Changes

- Add `extractFilenameFromUri()` helper in `eventUtils.ts`
- Use extracted filename instead of raw URI when creating the `File` object
- Remove `.prevent` from `@drop` on `LGraphNode.vue`, call `preventDefault()` conditionally
- Add tests for filename extraction (URL with extension, URL without extension)

## Test plan

- [x] Unit tests pass (`pnpm test:unit src/utils/__tests__/eventUtils.test.ts`)
- [x] E2E test passes (`Load workflow from URL dropped onto Vue node`)
- [x] Manually verified: drag-and-drop image URL onto canvas uploads successfully
- [x] CodeRabbit review: no findings


src/utils/eventUtils.ts
```diff
@@ -1,3 +1,30 @@
...
+  'image/webp': '.webp',
+  'image/gif': '.gif',
+  'image/svg+xml': '.svg',
+  'image/bmp': '.bmp',
+  'audio/mpeg': '.mp3',
+  'audio/wav': '.wav',
+  'audio/ogg': '.ogg',
+  'video/mp4': '.mp4',
+  'video/webm': '.webm'
+}
```
DrJKL: Seems like this could go in `mimeTypeUtil` as a reverse mapping of the existing util.

Myestery:
Addressed @DrJKL's feedback — moved the MIME mapping to `src/utils/mimeTypeUtil.ts` with both `getMimeType(fileName)` (ext→mime) and `getExtension(mimeType)` (mime→ext) derived from a single source map.

src/utils/mimeTypeUtil.ts
```diff
@@ -0,0 +1,31 @@
...
+  '.gif': 'image/gif',
+  '.svg': 'image/svg+xml',
+  '.bmp': 'image/bmp',
+  '.avif': 'image/avif',
+  '.mp3': 'audio/mpeg',
+  '.wav': 'audio/wav',
+  '.ogg': 'audio/ogg',
+  '.mp4': 'video/mp4',
+  '.webm': 'video/webm'
+}
```
dante01yoon, line 15: The old browser_tests/fixtures/helpers/mimeTypeUtil.ts also had `.json → application/json` and `.glb → model/gltf-binary`. Those are used by ClipboardHelper and DragDropHelper (which now import from this file). Dropping them changes behavior for test helpers that construct File objects from `.json/.glb` paths — they'll get `application/octet-stream` instead.

browser_tests/fixtures/helpers/ClipboardHelper.ts
```diff
@@ -4,7 +4,7 @@ import { basename } from 'path'
 import type { Locator, Page } from '@playwright/test'
 
 import type { KeyboardHelper } from './KeyboardHelper'
-import { getMimeType } from './mimeTypeUtil'
+import { getMimeType } from '../../../src/utils/mimeTypeUtil'
```
dante01yoon, line 7: nit: This crosses the browser_tests/ → src/ boundary with a deep `relative path (../../../src/...)`. Not blocking, but worth checking if there's a tsconfig path alias available for browser tests. Same applies to `DragDropHelper.ts:6`

PR #10534, opened 2026-03-25
Also promote children of promoted dynamic widgets

AustinMroz:
Changing the value of a DynamicCombo will cause widgets to be added or removed from a node. When a DynamicCombo is promoted, these dynamically added children will now also be promoted and any previously promoted children will be demoted.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10534-Also-promote-children-of-promoted-dynamic-widgets-32e6d73d36508177954ff5dd64e28864) by [Unito](https://www.unito.io)


AustinMroz:
Marking this as closed. There are planned architectural changes to solve this in a cleaner way.

PR #10545, opened 2026-03-26
test(infra): AssetHelper with builder pattern + deterministic fixtures

christian-byrne:
## What

Adds `AssetHelper` — a builder-pattern helper for mocking asset-related API endpoints in Playwright E2E tests, plus deterministic fixture data.

## Why

12+ asset-related API endpoints need mocking for asset browser tests (PNL-02), cloud dialog testing (DLG-08), and other asset-dependent E2E scenarios. Random mock data from existing `createMockAssets()` is unsuitable for deterministic E2E assertions.

### `AssetHelper.ts` (307 LOC)
- Fluent builder API: `assetHelper.withModels(3).withImages(5).mock()`
- Stateful mock store (Map) for upload→verify→delete flows
- Endpoint coverage: GET/POST/PUT/DELETE `/assets`, download progress
- `mockError()` for error state testing
- `clearMocks()` cleanup matching QueueHelper/FeatureFlagHelper pattern

### `assetFixtures.ts` (304 LOC)
- 11 stable named constants (checkpoints, loras, VAE, embedding, inputs, outputs)
- Factory functions: `generateModels()`, `generateInputFiles()`, `generateOutputAssets()`
- Fixed IDs/dates/sizes — no randomness, safe for screenshot comparisons

### ComfyPage integration
- Available as `comfyPage.assets` in all tests

## Testing
- TypeScript compiles clean
- Follows existing QueueHelper/FeatureFlagHelper conventions

## Unblocks
- PNL-02: Asset browser tests (@Jaewon Yoon)
- DLG-08: Assets modal / cloud dialog testing

Part of: Test Coverage Q2 Overhaul

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10545-test-infra-AssetHelper-with-builder-pattern-deterministic-fixtures-32f6d73d365081d3985ef079ff3dbede) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/fixtures/data/assetFixtures.ts
+++ b/browser_tests/fixtures/data/assetFixtures.ts
@@ -1,7 +1,4 @@
 import type { Asset } from '@comfyorg/ingest-types'
-
-// ─── Factory Helpers ───────────────────────────────────────────────────────────
-
 function createModelAsset(overrides: Partial<Asset> = {}): Asset {
   return {
     id: 'test-model-001',
@@ -50,9 +47,6 @@ function createOutputAsset(overrides: Partial<Asset> = {}): Asset {
     ...overrides
   }
 }
-
-// ─── Stable Fixtures (deterministic for screenshot tests) ──────────────────────
-
 export const STABLE_CHECKPOINT: Asset = createModelAsset({
   id: 'test-checkpoint-001',
   name: 'sd_xl_base_1.0.safetensors',
@@ -181,9 +175,6 @@ export const STABLE_OUTPUT_2: Asset = createOutputAsset({
   created_at: '2025-03-10T12:05:00Z',
   updated_at: '2025-03-10T12:05:00Z'
 })
-
-// ─── Preset Collections ────────────────────────────────────────────────────────
-
 export const ALL_MODEL_FIXTURES: Asset[] = [
   STABLE_CHECKPOINT,
   STABLE_CHECKPOINT_2,
@@ -200,9 +191,6 @@ export const ALL_INPUT_FIXTURES: Asset[] = [
 ]
 
 export const ALL_OUTPUT_FIXTURES: Asset[] = [STABLE_OUTPUT, STABLE_OUTPUT_2]
-
-// ─── Factories for generating N deterministic assets ───────────────────────────
-
 const CHECKPOINT_NAMES = [
   'sd_xl_base_1.0.safetensors',
   'v1-5-pruned-emaonly.safetensors',
--- a/browser_tests/fixtures/data/assetFixtures.ts
+++ b/browser_tests/fixtures/data/assetFixtures.ts
@@ -1,0 +1,307 @@
+import type { AssetItem } from '../../../src/platform/assets/schemas/assetSchema'
+
+// ─── Factory Helpers ───────────────────────────────────────────────────────────
+
+function createModelAsset(overrides: Partial<AssetItem> = {}): AssetItem {
+  return {
+    id: 'test-model-001',
+    name: 'model.safetensors',
+    asset_hash:
+      'blake3:0000000000000000000000000000000000000000000000000000000000000000',
+    size: 2_147_483_648,
+    mime_type: 'application/octet-stream',
+    tags: ['models', 'checkpoints'],
+    created_at: '2025-01-15T10:00:00Z',
+    updated_at: '2025-01-15T10:00:00Z',
+    last_access_time: '2025-01-15T10:00:00Z',
+    user_metadata: { base_model: 'sd15' },
+    ...overrides
+  }
+}
+
+function createInputAsset(overrides: Partial<AssetItem> = {}): AssetItem {
+  return {
+    id: 'test-input-001',
+    name: 'input.png',
+    asset_hash:
+      'blake3:1111111111111111111111111111111111111111111111111111111111111111',
+    size: 2_048_576,
+    mime_type: 'image/png',
+    tags: ['input'],
+    created_at: '2025-03-01T09:00:00Z',
+    updated_at: '2025-03-01T09:00:00Z',
+    last_access_time: '2025-03-01T09:00:00Z',
+    ...overrides
+  }
+}
+
+function createOutputAsset(overrides: Partial<AssetItem> = {}): AssetItem {
+  return {
+    id: 'test-output-001',
+    name: 'output_00001.png',
+    asset_hash:
+      'blake3:2222222222222222222222222222222222222222222222222222222222222222',
+    size: 4_194_304,
+    mime_type: 'image/png',
+    tags: ['output'],
+    created_at: '2025-03-10T12:00:00Z',
+    updated_at: '2025-03-10T12:00:00Z',
+    last_access_time: '2025-03-10T12:00:00Z',
+    ...overrides
+  }
+}
+
+// ─── Stable Fixtures (deterministic for screenshot tests) ──────────────────────
+
+export const STABLE_CHECKPOINT: AssetItem = createModelAsset({
+  id: 'test-checkpoint-001',
+  name: 'sd_xl_base_1.0.safetensors',
+  size: 6_938_078_208,
+  tags: ['models', 'checkpoints'],
+  user_metadata: {
+    base_model: 'sdxl',
+    description: 'Stable Diffusion XL Base 1.0'
+  },
+  created_at: '2025-01-15T10:30:00Z',
+  updated_at: '2025-01-15T10:30:00Z'
+})
+
+export const STABLE_CHECKPOINT_2: AssetItem = createModelAsset({
+  id: 'test-checkpoint-002',
+  name: 'v1-5-pruned-emaonly.safetensors',
+  size: 4_265_146_304,
+  tags: ['models', 'checkpoints'],
+  user_metadata: {
+    base_model: 'sd15',
+    description: 'Stable Diffusion 1.5 Pruned EMA-Only'
+  },
+  created_at: '2025-01-20T08:00:00Z',
+  updated_at: '2025-01-20T08:00:00Z'
+})
+
+export const STABLE_LORA: AssetItem = createModelAsset({
+  id: 'test-lora-001',
+  name: 'detail_enhancer_v1.2.safetensors',
+  size: 184_549_376,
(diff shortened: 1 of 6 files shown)


browser_tests/fixtures/data/assetFixtures.ts
christian-byrne, line 1: Use the types in packages/ingest-types for static type checking and zod types in the same package (if we ever need runtime type-checking e.g. when using `as`).

browser_tests/fixtures/data/assetFixtures.ts
```diff
@@ -0,0 +1,307 @@
...
+  'style_reference.png',
+  'batch_001.png',
+  'batch_002.png'
+]
+
+const INPUT_MIMES: Record<string, string> = {
+  '.png': 'image/png',
+  '.jpg': 'image/jpeg',
+  '.mp4': 'video/mp4'
+}
```
christian-byrne: Check if we can find some mime utils files to import here or derive this from the generated types in ingest-types. Also audit all places in codebase where we are doing something related to mime types and see if we can centralize from all of there as well.

browser_tests/fixtures/helpers/AssetHelper.ts
```diff
@@ -0,0 +1,311 @@
...
+  private routeHandlers: Array<{
+    pattern: string
+    handler: (route: Route) => Promise<void>
+  }> = []
+  private mutations: MutationRecord[] = []
+  private uploadResponse: Record<string, unknown> | null = null
+
+  constructor(private readonly page: Page) {}
+
+  // ─── Builder Methods (return `this` for chaining) ─────────────────────────
```
DrJKL: Instead of a builder, can we do a functional pipeable operators style? It should be more type safe and prevent (or make obvious) things like calling .withEmptyState() in the middle of an otherwise stateful chain of calls.

christian-byrne (commented):
Remove all  organizational comments like "// ─── Builder Methods (return `this` for chaining) ─────────────────────────" or adding things which are obvious from just reading the code or signatures.

browser_tests/fixtures/helpers/AssetHelper.ts
```diff
@@ -0,0 +1,311 @@
...
+  async clearMocks(): Promise<void> {
+    for (const { pattern, handler } of this.routeHandlers) {
+      await this.page.unroute(pattern, handler)
+    }
+    this.routeHandlers = []
+    this.store.clear()
+    this.mutations = []
+    this.paginationOptions = null
+    this.uploadResponse = null
+  }
```
christian-byrne, line 290: If this is test cleanup, it should be in a fixture that gets called in the fixture cleanup standard playwright pattern in ComfyPage. If this is for the purpose of putting inside of a test case midway through the testing logic -- that should never be allowed anyway.

We should also consider deeply whether this should be a property of ComfyPage or a fixture separate of ComfyPage (which would allow the cleanup stuff to be auto invoked and way more abstracted from test cases' perspective).

christian-byrne (commented):
We should include a few basic test cases that touch all the functionalities, methods, mocks in this PR just so we can validate that they work and validate the DX.

browser_tests/fixtures/data/assetFixtures.ts
christian-byrne: fixed in e2d2fae28 - switched to Asset/ListAssetsResponse from @comfyorg/ingest-types

browser_tests/fixtures/data/assetFixtures.ts
christian-byrne: fixed in e868ec938 - replaced the 3-entry INPUT_MIMES with an extensible EXTENSION_MIME_MAP (10 entries) and getMimeType() helper. audited the codebase: getMediaTypeFromFilename in shared-frontend-utils returns categories not MIME strings, formats.ts has workflow-specific types, and various files have scattered inline references - but no centralized extension-to-MIME-string utility exists yet. broader centralization could be a follow-up PR.

browser_tests/fixtures/helpers/AssetHelper.ts
christian-byrne: fixed in 281a58f30 - replaced the builder with functional pipeable operators: standalone functions (withModels, withInputFiles, withOutputAssets, withAsset, withPagination, withUploadResponse) that transform an immutable AssetConfig, applied via createAssetHelper(page, ...operators). withEmptyState() removed since empty is the default, which prevents invalid mid-chain state resets.

browser_tests/fixtures/helpers/AssetHelper.ts
christian-byrne: fixed in 307617e71 - integrated clearMocks() into the comfyPageFixture teardown phase (after await use(comfyPage)), so cleanup is automatic via Playwright fixture lifecycle. AssetHelper stays as a property of ComfyPage following the same pattern as QueueHelper and FeatureFlagHelper.

browser_tests/fixtures/data/assetFixtures.ts
```diff
@@ -0,0 +1,318 @@
+import type { Asset } from '@comfyorg/ingest-types'
+
+// ─── Factory Helpers ───────────────────────────────────────────────────────────
```
christian-byrne: get rid of all these section/HR comments from entire PR

browser_tests/fixtures/helpers/AssetHelper.ts
```diff
@@ -0,0 +1,307 @@
...
+  method: string
+  url: string
+  body: unknown
+  timestamp: number
+}
+
+interface PaginationOptions {
+  total: number
+  hasMore: boolean
+}
```
christian-byrne, line 21: can't these also be derived from the ingest-types package?

christian-byrne (commented):
We need to write a few tests that use the helper to verify the DX and correctness then include those in this PR

browser_tests/fixtures/helpers/AssetHelper.ts
christian-byrne: fixed in e2d2fae28 - Asset and ListAssetsResponse now come from @comfyorg/ingest-types. MutationRecord is test-specific (records intercepted API calls for assertions) with no counterpart in the package.

browser_tests/fixtures/helpers/AssetHelper.ts
```diff
@@ -0,0 +1,275 @@
...
+
+  constructor(
+    private readonly page: Page,
+    config: AssetConfig = emptyConfig()
+  ) {
+    this.store = new Map(config.assets)
+    this.paginationOptions = config.pagination
+    this.uploadResponse = config.uploadResponse
+  }
+  async mock(): Promise<void> {
```
christian-byrne, line 107: Can we move each handler to its own function and then let this function body be far cleaner?

christian-byrne (commented):
Can we add some test cases that use this helper to validate it works

browser_tests/tests/assetHelper.spec.ts
```diff
@@ -0,0 +1,441 @@
...
+  })
+
+  test.describe('clearMocks', () => {
+    test('resets store, mutations, and unroutes handlers', async ({
+      comfyPage
+    }) => {
+      const helper = createAssetHelper(
+        comfyPage.page,
+        withAsset(STABLE_CHECKPOINT)
+      )
```
christian-byrne: isn't it supposed to be included in the test call similar to comfyPage?

browser_tests/tests/assetHelper.spec.ts
```diff
@@ -0,0 +1,441 @@
...
+      await helper.mock()
+
+      await pageFetch(comfyPage.page, `${comfyPage.url}/api/assets`, {
+        method: 'POST'
+      })
```
christian-byrne: this should be a helper or helper method -- not each individual test case doing POST requests which don't really emulate user behavior at all

browser_tests/tests/assetHelper.spec.ts
christian-byrne: fixed in 926628cd5 - tests now use comfyPage.assetApi with a new configure() method instead of creating standalone helpers

browser_tests/tests/assetHelper.spec.ts
christian-byrne: fixed in 926628cd5 - added fetch() method to AssetHelper that uses page.evaluate(fetch) internally. tests now call assetApi.fetch() instead of raw pageFetch()

benceruleanlu (dismissed):
Can't wait to use!

Note: I had some followups I stacked on this branch, just types so it's not really blocking

benceruleanlu (commented):
This looks partially superseded by #10630. The main conflict is browser_tests/fixtures/ComfyPage.ts: this PR adds , while #10630 intentionally moves mock/backend state off  into standalone fixtures and already migrates the real assets/sidebar browser tests onto that foundation. I think we should avoid landing both patterns in parallel.

I do think there are salvageable pieces here, especially  and possibly some  mock behavior for future cloud asset-service tests. My preference for deduping quickly is:
1. Merge #10630 as the browser-test foundation.
2. Treat this PR as superseded in its current form.
3. If we still want the reusable  bits, follow up by porting them into the #10630 direction as a standalone fixture/backend instead of .

That gets one path in quickly and avoids maintaining two competing test-infra patterns.

benceruleanlu (commented):
Correcting the formatting on my previous review: this looks partially superseded by #10630. The main conflict is in browser_tests/fixtures/ComfyPage.ts: this PR adds asset API mocking back onto ComfyPage, while #10630 intentionally moves mock/backend state off ComfyPage into standalone fixtures and already migrates the real assets/sidebar browser tests onto that foundation. I do not think we should land both patterns in parallel.

I do think there are salvageable pieces here, especially the deterministic asset fixture data and possibly some reusable /api/assets mock behavior for future cloud asset-service tests. My preference for deduping quickly is:
1. Merge #10630 as the browser-test foundation.
2. Treat this PR as superseded in its current form.
3. If we still want the reusable /api/assets bits, follow up by porting them into the #10630 direction as a standalone fixture/backend instead of hanging them off ComfyPage.

That gets one path in quickly and avoids maintaining two competing test-infra patterns.

PR #10550, opened 2026-03-26
fix: prevent canvas zoom when scrolling image history dropdown

### LINKED ISSUE #10549: fix: the scroll up and down of the dropbox of the the upload image dropdown widget

Fix scroll behavior: The image history dropdown has a fixed height, but scrolling currently triggers canvas zooming. We need to add a check so that when the mouse is over the dropdown, the wheel scrolls the list instead of zooming the map.

The wheel scrolls the the map zooming rathar than the list up-and-down.

Multiple times per day

┆Issue is synchronized with this [Notion page](https://www.notion.so/Issue-10549-fix-the-scroll-up-and-down-of-the-dropbox-of-the-the-upload-image-dropdown-widget-32f6d73d365081c9a3c3caa1fa2d941b) by [Unito](https://www.unito.io)

kaili-yang:
## Summary
 
Fix #10549 where using the mouse wheel over the image history dropdown (e.g., in "Load Image" nodes) would trigger canvas zooming instead of scrolling the list.

## Changes
Added `data-capture-wheel="true" ` to the root container. This attribute is used by the `TransformPane` to identify elements that should consume wheel events.

## Screenshots
 
after

https://github.com/user-attachments/assets/8935a1ca-9053-4ef1-9ab8-237f43eabb35

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10550-fix-prevent-canvas-zoom-when-scrolling-image-history-dropdown-32f6d73d365081c4ad09f763481ef8c2) by [Unito](https://www.unito.io)


src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdownMenuActions.vue
```diff
@@ -103,6 +103,7 @@ function toggleBaseModelSelection(item: FilterOption) {
   <div class="text-secondary flex gap-2 px-4">
     <FormSearchInput
       v-model="searchQuery"
+      :autofocus="true"
```
jtydhr88: Nit: :autofocus="true" → autofocus would be more consistent with the rest of the codebase (see WorkflowTemplateSelectorDialog.vue, CustomizationDialog.vue, etc.)

jtydhr88 (changes requested):

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdownMenuActions.vue
kaili-yang: It's a good point. Fixed.

jtydhr88 (approved):
LGTM, and I think the falure test does not caused by this changes

jtydhr88:
the recording has another issue that load image panel on wrong location, but it should be fix in followup PR

src/renderer/extensions/vueNodes/widgets/components/form/dropdown/FormDropdownMenu.test.ts
```diff
@@ -110,4 +110,19 @@ describe('FormDropdownMenu', () => {
...
+        stubs: {
+          FormDropdownMenuFilter: true,
+          FormDropdownMenuActions: true,
+          VirtualGrid: true
+        }
+      }
+    })
+
+    expect(wrapper.attributes('data-capture-wheel')).toBe('true')
+  })
```
DrJKL, line 127: This is kind of a change-detector. Ideally we'd test this via Playwright using actual scroll events.
@coderabbitai Open an issue to add scroll handling specs.

DrJKL (approved):

DrJKL (approved):

PR #10565, opened 2026-03-26
fix: create initial workflow tab when persistence is disabled

jaeone94:
## Summary

- When `Comfy.Workflow.Persist` is OFF and storage is empty, `initializeWorkflow()` returned without creating any workflow tab — leaving users with no tab and no way to save
- Now falls through to `loadDefaultWorkflow()` so a default temporary workflow is always created

## Root Cause

In `useWorkflowPersistenceV2.ts`, `initializeWorkflow()` had an early return when persistence was disabled:

```ts
```

This skipped `loadDefaultWorkflow()`, which is responsible for creating the initial temporary workflow tab via `comfyApp.loadGraphData()` → `afterLoadNewGraph()` → `workflowStore.createNewTemporary()`.

## Fix

One-line change: `return` → `return loadDefaultWorkflow()`.

## Test plan

- [x] E2E test: verifies `openWorkflows.length >= 1` after reload with persistence OFF

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10565-fix-create-initial-workflow-tab-when-persistence-is-disabled-32f6d73d365081d5a681c3e019d373c3) by [Unito](https://www.unito.io)

### DIFF
--- a/src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
+++ b/src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
@@ -171,7 +171,7 @@ export function useWorkflowPersistenceV2() {
   }
 
   const initializeWorkflow = async () => {
-    if (!workflowPersistenceEnabled.value) return
+    if (!workflowPersistenceEnabled.value) return loadDefaultWorkflow()
 
     try {
       const restored = await loadPreviousWorkflowFromStorage()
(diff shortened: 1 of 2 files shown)


src/platform/workflow/persistence/composables/useWorkflowPersistenceV2.ts
```diff
@@ -171,7 +171,7 @@ export function useWorkflowPersistenceV2() {
   }
 
   const initializeWorkflow = async () => {
-    if (!workflowPersistenceEnabled.value) return
+    if (!workflowPersistenceEnabled.value) return loadDefaultWorkflow()
```
DrJKL: Nit: Matching the patterns below
```suggestion
    if (!workflowPersistenceEnabled.value) {
      await loadDefaultWorkflow()
      return
    }
```

DrJKL (dismissed):

DrJKL (approved):

PR #10573, opened 2026-03-26
fix: normalize legacy prefixed proxyWidget entries on configure

DrJKL:
## Summary

Normalize legacy prefixed proxyWidget entries during subgraph configure so nested subgraph widgets resolve correctly.

## Changes

- **What**: Extract `normalizeLegacyProxyWidgetEntry` to strip legacy `nodeId: innerNodeId: widgetName` prefixes from serialized proxyWidgets and resolve the correct `disambiguatingSourceNodeId`. Write-back comparison now checks serialized content (not just array length) so stale formats are cleaned up even when the entry count is unchanged.

## Review Focus

- The iterative prefix-stripping loop in `resolveLegacyPrefixedEntry` — it peels one `N: ` prefix per iteration and tries all disambiguator candidates at each level.
- The write-back condition change from length comparison to `JSON.stringify` equality.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10573-fix-normalize-legacy-prefixed-proxyWidget-entries-on-configure-32f6d73d365081e886e1c9b3939e3b9f) by [Unito](https://www.unito.io)

### DIFF
--- a/src/core/graph/subgraph/legacyProxyWidgetNormalization.ts
+++ b/src/core/graph/subgraph/legacyProxyWidgetNormalization.ts
@@ -1,0 +1,111 @@
+import type { PromotedWidgetSource } from '@/core/graph/subgraph/promotedWidgetTypes'
+import { resolveConcretePromotedWidget } from '@/core/graph/subgraph/resolveConcretePromotedWidget'
+import type { SubgraphNode } from '@/lib/litegraph/src/subgraph/SubgraphNode'
+
+const LEGACY_PROXY_WIDGET_PREFIX_PATTERN = /^\s*(\d+)\s*:\s*(.+)$/
+
+type PromotedWidgetPatch = Omit<PromotedWidgetSource, 'sourceNodeId'>
+
+function canResolve(
+  hostNode: SubgraphNode,
+  sourceNodeId: string,
+  widgetName: string,
+  disambiguator?: string
+): boolean {
+  return (
+    resolveConcretePromotedWidget(
+      hostNode,
+      sourceNodeId,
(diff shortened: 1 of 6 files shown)


src/core/graph/subgraph/legacyProxyWidgetNormalization.ts
DrJKL, line 1: Yes, this is longer and dumber than I'd like.

src/core/graph/subgraph/legacyProxyWidgetNormalization.ts
```diff
@@ -0,0 +1,111 @@
+import type { PromotedWidgetSource } from '@/core/graph/subgraph/promotedWidgetTypes'
+import { resolveConcretePromotedWidget } from '@/core/graph/subgraph/resolveConcretePromotedWidget'
+import type { SubgraphNode } from '@/lib/litegraph/src/subgraph/SubgraphNode'
+
+const LEGACY_PROXY_WIDGET_PREFIX_PATTERN = /^\s*(\d+)\s*:\s*(.+)$/
```
AustinMroz, line 5: Aren't the spaces consistent? I think we could simplify the regex to `/^(\d+): (.+)$/`, but it's probably not worth fixing if it's working now.

src/core/graph/subgraph/legacyProxyWidgetNormalization.ts
AustinMroz: At least it's all contained in a single isolated file :grin:

AustinMroz (approved):

PR #10227, opened 2026-03-18
fix: enforce canonical subgraph IO order after configure [PR 1 of 2]

artokun:
## Summary (Part 1 of 2)

Groundwork for subgraph IO ordering: enforce canonical order after configure so all views stay in sync.

This is the first of three PRs hardening subgraph widget/slot consistency:
- **PR 1** (this): Enforce canonical IO order so inner pins, outer pins, and properties panel stay in sync
- **PR 2** (#10232): Add drag-to-reorder for subgraph IO pins inside subgraph editing mode

## Changes

- Sort SubgraphNode inputs/outputs to match canonical subgraph definition order after rebinding and pruning
- Fix link slot indices after reorder to preserve connections
- Mismatched serialized workflows self-heal on load

## Problem

Subgraph IO order drifts between inner pins, outer pins, and properties panel. `super.configure()` replaces canonical-order inputs with serialized-order inputs, then `_rebindInputSubgraphSlots` heuristically re-matches them without enforcing order.

## Solution

New `_sortSlotsToCanonicalOrder()` method called in `_internalConfigureAfterSlots` after rebinding and pruning. Sorts `this.inputs` by `subgraph.inputNode.slots` index and `this.outputs` by `subgraph.outputNode.slots` index, then fixes `link.target_slot` / `link.origin_slot` references.

## Test plan

- [x] Unit test: mismatched serialized input order is normalized to canonical
- [x] Unit test: mismatched serialized output order is normalized
- [x] Unit test: serialize/configure cycles are idempotent (3 cycles)
- [ ] Manual: create subgraph with multiple inputs, save, reload, verify order matches

- Fixes #10221

## PR Chain

1. **#10227 — This PR: canonical subgraph IO order enforcement**
2. #10232 — Drag-to-reorder subgraph IO pins

1. **#10227 — This PR: canonical subgraph IO order enforcement**
2. #10232 — Drag-to-reorder subgraph IO pins

PR #10584, opened 2026-03-27
fix: preserve promoted widget order on sidebar hide/show cycle / drag reorder

### LINKED ISSUE #9995: Subgraph: Layout order of elements not respected after changing "Shown on node" configuration

artokun:
## Summary

Fix three bugs in the sidebar show/hide flow that caused promoted widget order to scramble and allowed duplicate promotions.

## Changes

- **What**: `promotionStore.promote()` now restores a widget to its original position when re-promoted after demotion. `getSourceNodeId()` no longer falls back to `sourceNodeId`, fixing broken `isPromoted` lookups for non-nested subgraphs. Removed non-idempotent `extractVueNodeData()` re-call from `node:slot-label:changed` handler.

## Review Focus

- `promotionStore.ts`: `demote()` records removed entry index in `lastDemotedIndices` map; `promote()` inserts at that index (clamped) instead of always appending. Map cleared on `clearGraph()`.
- `promotionUtils.ts`: One-line fix removing `?? w.sourceNodeId` fallback that caused `disambiguatingSourceNodeId` mismatch in all sidebar store lookups.
- `useGraphNodeManager.ts`: Removed `vueNodeData.set(nodeId, extractVueNodeData(nodeRef))` — the existing reactive array reassignment already triggers `safeWidgets` recomputation.

Fixes #9995

Related: #9841 (duplicate inputs + broken ordering), #10227 / #10232 (subgraph IO ordering — handled separately)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10584-fix-preserve-promoted-widget-order-on-sidebar-hide-show-cycle-drag-reorder-3306d73d365081d9be91f55494e5b00e) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/subgraphPromotionOrder.spec.ts
+++ b/browser_tests/tests/subgraphPromotionOrder.spec.ts
@@ -1,6 +1,7 @@
 import { expect } from '@playwright/test'
 
 import { comfyPageFixture as test } from '../fixtures/ComfyPage'
+import { TestIds } from '../fixtures/selectors'
 import { getPromotedWidgetNames } from '../helpers/promotedWidgets'
 
 test.describe(
@@ -40,19 +41,14 @@ test.describe(
       const subgraphVueNode = comfyPage.vueNodes.getNodeLocator(nodeId)
       await expect(subgraphVueNode).toBeVisible()
 
-      const nodeBody = subgraphVueNode.locator(
-        `[data-testid="node-body-${nodeId}"]`
+      const widgetsContainer = subgraphVueNode.getByTestId(
+        TestIds.widgets.container
       )
-      const widgetElements = nodeBody.locator('.lg-node-widgets > div')
-      await expect(widgetElements.first()).toBeVisible()
+      await expect(widgetsContainer).toBeVisible()
 
       // Get the initial promoted widget order
       const initialOrder = await getPromotedWidgetNames(comfyPage, nodeId)
       expect(initialOrder.length).toBeGreaterThanOrEqual(2)
-
-      // Verify widget count in the DOM matches
-      const domWidgetCount = await widgetElements.count()
-      expect(domWidgetCount).toBeGreaterThanOrEqual(initialOrder.length)
 
       // Demote then re-promote the first widget via the promotion store.
       // Pinia store access pattern: see getPiniaStoreInBrowser in
--- a/browser_tests/tests/subgraphPromotionOrder.spec.ts
+++ b/browser_tests/tests/subgraphPromotionOrder.spec.ts
@@ -1,0 +1,98 @@
+import { expect } from '@playwright/test'
+
+import { comfyPageFixture as test } from '../fixtures/ComfyPage'
+import { getPromotedWidgetNames } from '../helpers/promotedWidgets'
+
+test.describe(
+  'Subgraph promoted widget ordering',
+  { tag: ['@subgraph', '@widget'] },
+  () => {
+    test.beforeEach(async ({ comfyPage }) => {
+      await comfyPage.settings.setSetting('Comfy.UseNewMenu', 'Disabled')
+      await comfyPage.settings.setSetting('Comfy.VueNodes.Enabled', true)
+    })
+
+    test('demote then re-promote preserves widget order', async ({
+      comfyPage
+    }) => {
+      await comfyPage.workflow.loadWorkflow('default')
+
+      // Create a subgraph from the KSampler node which has multiple
+      // auto-promoted widgets (seed, steps, cfg, etc.)
+      const ksampler = comfyPage.vueNodes.getNodeByTitle('KSampler')
+      await comfyPage.vueNodes.waitForNodes()
+      await expect(ksampler).toBeVisible()
+
+      // Select and convert to subgraph
+      const ksamplerRef = await comfyPage.nodeOps.getNodeRefById('3')
+      await ksamplerRef.click('title')
+      const subgraphNode = await ksamplerRef.convertToSubgraph()
+      await comfyPage.nextFrame()
+      await comfyPage.vueNodes.waitForNodes()
+
+      const nodeId = String(subgraphNode.id)
+
+      // Verify promoted widgets render in the Vue node body
+      const subgraphVueNode = comfyPage.vueNodes.getNodeLocator(nodeId)
+      await expect(subgraphVueNode).toBeVisible()
+
+      const nodeBody = subgraphVueNode.locator(
+        `[data-testid="node-body-${nodeId}"]`
+      )
+      const widgetElements = nodeBody.locator('.lg-node-widgets > div')
+      await expect(widgetElements.first()).toBeVisible()
+
+      // Get the initial promoted widget order
+      const initialOrder = await getPromotedWidgetNames(comfyPage, nodeId)
+      expect(initialOrder.length).toBeGreaterThanOrEqual(2)
+
+      // Verify widget count in the DOM matches
+      const domWidgetCount = await widgetElements.count()
+      expect(domWidgetCount).toBeGreaterThanOrEqual(initialOrder.length)
+
+      // Demote then re-promote the first widget via the promotion store
+      const finalOrder = await comfyPage.page.evaluate(
+        ([id, widgetName]) => {
+          const el = document.getElementById('vue-app') as HTMLElement & {
+            __vue_app__: {
+              config: {
+                globalProperties: {
+                  $pinia: {
+                    _s: Map<
+                      string,
+                      Record<string, (...args: unknown[]) => unknown>
+                    >
+                  }
+                }
+              }
+            }
+          }
+          const store =
+            el.__vue_app__.config.globalProperties.$pinia._s.get('promotion')
+          if (!store) throw new Error('promotionStore not found')
+
+          const node = window.app!.canvas.graph!.getNodeById(id)
+          if (!node) throw new Error(`Node "${id}" not found`)
+          const graphId = (node as { rootGraph?: { id: string } }).rootGraph?.id
+
+          const entries = store.getPromotions(graphId, node.id) as {
+            sourceWidgetName: string
+          }[]
+          const target = entries.find((e) => e.sourceWidgetName === widgetName)
+          if (!target) throw new Error(`Widget "${widgetName}" not found`)
+
+          store.demote(graphId, node.id, target)
+          node.computeSize(node.size)
+
+          store.promote(graphId, node.id, target)
+          node.computeSize(node.size)
+
+          return (node.widgets ?? []).map((w: { name: string }) => w.name)
+        },
+        [nodeId, initialOrder[0]]
+      )
+
+      expect(finalOrder).toEqual(initialOrder)
+    })
+  }
+)
--- a/browser_tests/tests/subgraphPromotionOrder.spec.ts
+++ b/browser_tests/tests/subgraphPromotionOrder.spec.ts
@@ -19,18 +19,22 @@ test.describe(
 
       // Create a subgraph from the KSampler node which has multiple
       // auto-promoted widgets (seed, steps, cfg, etc.)
-      const ksampler = comfyPage.vueNodes.getNodeByTitle('KSampler')
       await comfyPage.vueNodes.waitForNodes()
-      await expect(ksampler).toBeVisible()
 
-      // Select and convert to subgraph
-      const ksamplerRef = await comfyPage.nodeOps.getNodeRefById('3')
-      await ksamplerRef.click('title')
-      const subgraphNode = await ksamplerRef.convertToSubgraph()
+      // Select KSampler and convert to subgraph via evaluate
+      // (LG nodeRef helpers don't work reliably in Vue nodes mode)
+      const nodeId = await comfyPage.page.evaluate(() => {
+        const graph = window.app!.canvas.graph!
+        const ksampler = graph._nodes.find((n) => n.type === 'KSampler')
+        if (!ksampler) throw new Error('KSampler not found')
+
+        window.app!.canvas.selectNode(ksampler)
+        const result = graph.convertToSubgraph(new Set([ksampler]))
+        if (!result) throw new Error('convertToSubgraph failed')
+        return String(result.node.id)
+      })
       await comfyPage.nextFrame()
       await comfyPage.vueNodes.waitForNodes()
-
-      const nodeId = String(subgraphNode.id)
 
       // Verify promoted widgets render in the Vue node body
       const subgraphVueNode = comfyPage.vueNodes.getNodeLocator(nodeId)
--- a/browser_tests/tests/subgraphPromotionOrder.spec.ts
+++ b/browser_tests/tests/subgraphPromotionOrder.spec.ts
@@ -50,10 +50,12 @@ test.describe(
       const domWidgetCount = await widgetElements.count()
       expect(domWidgetCount).toBeGreaterThanOrEqual(initialOrder.length)
 
-      // Demote then re-promote the first widget via the promotion store
+      // Demote then re-promote the first widget via the promotion store.
+      // Pinia store access pattern: see getPiniaStoreInBrowser in
+      // browser_tests/helpers/promotedWidgets.ts
       const finalOrder = await comfyPage.page.evaluate(
         ([id, widgetName]) => {
-          const el = document.getElementById('vue-app') as HTMLElement & {
+          const el = document.getElementById('vue-app') as never as {
             __vue_app__: {
               config: {
                 globalProperties: {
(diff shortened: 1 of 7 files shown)


artokun:
@christian-byrne this is the fix for the desync on the subgraph node reordering bug from the sidebar, and, coincidentally also fixes the disconnect bug due to how the promotion ids were falling back. This also fixes this issue for deeply nested subgraphs to promote to parent subgraphs via their promotion paths. tested locally, updated both unit and playwright tests

src/stores/promotionStore.ts
```diff
@@ -19,6 +19,7 @@ export const usePromotionStore = defineStore('promotion', () => {
     new Map<UUID, Map<NodeId, PromotedWidgetSource[]>>()
   )
   const graphRefCounts = ref(new Map<UUID, Map<string, number>>())
+  const lastDemotedIndices = new Map<string, number>()
```
DrJKL: 🎫 **Low — This ticket holder is standing in line without a wristband.**

`graphPromotions` and `graphRefCounts` both wear their `ref()` wristbands proudly, but `lastDemotedIndices` showed up in plain clothes. That’s *correct* — it’s ephemeral internal state that shouldn’t trigger re-renders — but it breaks the visual pattern of the queue.

A one-liner comment would help the next person in line understand why this one gets to skip the velvet rope:
```ts
// Non-reactive: ephemeral position cache consumed only by promote/demote
const lastDemotedIndices = new Map<string, number>()
```

src/stores/promotionStore.ts
```diff
@@ -136,7 +137,19 @@ export const usePromotionStore = defineStore('promotion', () => {
     }
     if (source.disambiguatingSourceNodeId)
       entry.disambiguatingSourceNodeId = source.disambiguatingSourceNodeId
-    setPromotions(graphId, subgraphNodeId, [...entries, entry])
+
+    const positionKey = `${graphId}:${subgraphNodeId}:${makePromotionEntryKey(source)}`
+    const lastIndex = lastDemotedIndices.get(positionKey)
```
DrJKL: 🎫 **Medium — Same ticket stub, printed twice.**

This `positionKey` template is duplicated verbatim in `demote()` (L168). If the queue numbering scheme ever changes, one copy drifts. Red rover, red rover, send a `makePositionKey` helper right over:
```ts
function makePositionKey(graphId: UUID, subgraphNodeId: NodeId, source: PromotedWidgetSource): string {
  return `${graphId}:${subgraphNodeId}:${makePromotionEntryKey(source)}`
}
```

src/stores/promotionStore.ts
```diff
@@ -145,17 +158,20 @@ export const usePromotionStore = defineStore('promotion', () => {
...
-      entries.filter(
-        (e) =>
-          !(
-            e.sourceNodeId === source.sourceNodeId &&
-            e.sourceWidgetName === source.sourceWidgetName &&
-            e.disambiguatingSourceNodeId === source.disambiguatingSourceNodeId
-          )
-      )
+      entries.filter((_, i) => i !== index)
     )
```
DrJKL, line 195: 🚶 **High — This person left the line but we’re still reshuffling the whole queue.**

When `findIndex` returns `-1` (source not found), `entries.filter((_, i) => i !== index)` returns a full copy of the same array, then `setPromotions` runs the full `_decrementKeys`/`_incrementKeys` cycle for nothing — everyone steps forward and back to exactly where they were standing.

An early return makes the no-op explicit and avoids the wasted shuffle:
```ts
if (index === -1) return
```

src/stores/promotionStore.ts
```diff
@@ -188,6 +204,9 @@ export const usePromotionStore = defineStore('promotion', () => {
   function clearGraph(graphId: UUID): void {
     graphPromotions.value.delete(graphId)
     graphRefCounts.value.delete(graphId)
+    for (const key of lastDemotedIndices.keys()) {
+      if (key.startsWith(`${graphId}:`)) lastDemotedIndices.delete(key)
+    }
```
DrJKL: 🔄 **Medium — Kicking people out of line while counting heads.**

Deleting from a `Map` during `keys()` iteration is [spec-safe](https://tc39.es/ecma262/#sec-map.prototype.foreach), but it’s the kind of line-cutting that makes people nervous. `Array.from()` is a one-liner bouncer that removes all doubt:
```ts
for (const key of Array.from(lastDemotedIndices.keys())) {
  if (key.startsWith(`${graphId}:`)) lastDemotedIndices.delete(key)
}
```

src/core/graph/subgraph/promotionUtils.ts
```diff
@@ -29,7 +29,7 @@ export function getWidgetName(w: IBaseWidget): string {
 
 export function getSourceNodeId(w: IBaseWidget): string | undefined {
   if (!isPromotedWidgetView(w)) return undefined
-  return w.disambiguatingSourceNodeId ?? w.sourceNodeId
+  return w.disambiguatingSourceNodeId
```
DrJKL, line 57: 📋 **High — Red rover, red rover, send this slot right over — but now it arrives without a name tag.**

The old `?? w.sourceNodeId` fallback was incorrectly stamping `disambiguatingSourceNodeId` with `sourceNodeId` even when no disambiguation was needed — like giving every person in line someone else’s queue number. Removing the fallback is the right fix.

But this changes behavior for all 6 callers and has **no dedicated unit test**. Add a test for `getSourceNodeId` to document the new contract — otherwise the next person in line might re-add the fallback thinking it was accidentally removed.

src/stores/promotionStore.test.ts
```diff
@@ -711,6 +711,148 @@ describe(usePromotionStore, () => {
...
+      store.promote(graphA, nodeId, {
+        sourceNodeId: '11',
+        sourceWidgetName: 'steps'
+      })
+
+      expect(store.getPromotions(graphA, nodeId)).toEqual([
+        { sourceNodeId: '10', sourceWidgetName: 'seed' },
+        { sourceNodeId: '11', sourceWidgetName: 'steps' }
+      ])
+    })
```
DrJKL: 🏷️ **Low — This test’s name tag doesn’t match its spot in line.**

"appends when no prior position is remembered" implies a demote→re-promote with no memory, but the test only promotes twice — nobody ever leaves the queue. It’s testing default append order, not the "forgotten queue position" path.

Rename to something like `"initial promotes append in order"` so the test’s wristband matches the line it’s standing in.

browser_tests/tests/subgraphPromotionOrder.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+                      string,
+                      Record<string, (...args: unknown[]) => unknown>
+                    >
+                  }
+                }
+              }
+            }
+          }
+          const store =
+            el.__vue_app__.config.globalProperties.$pinia._s.get('promotion')
```
DrJKL: 🔧 **Medium — Reaching behind the velvet rope to grab the clipboard.**

Accessing `$pinia._s.get('promotion')` with a 14-line inline type assertion is like cutting through the back of the queue to read the manager’s roster. `_s` is an undocumented Pinia internal — if Pinia renames it, this test silently falls over.

`browser_tests/helpers/promotedWidgets.ts` already does similar queue-jumping. Consider extracting a shared `getStoreFromPage(page, storeName)` helper so there’s a single point of breakage when Pinia changes the velvet rope layout.

browser_tests/tests/subgraphPromotionOrder.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+          if (!store) throw new Error('promotionStore not found')
+
+          const node = window.app!.canvas.graph!.getNodeById(id)
+          if (!node) throw new Error(`Node "${id}" not found`)
+          const graphId = (node as { rootGraph?: { id: string } }).rootGraph?.id
```
DrJKL: 🪪 **Low — Swapping out the whole ID badge instead of just adding a line.**

`node as { rootGraph?: { id: string } }` erases all other node properties — like taking someone’s full badge and replacing it with just a sticky note. An intersection preserves the original identity:
```ts
const graphId = (node as typeof node & { rootGraph?: { id: string } }).rootGraph?.id
```

DrJKL (commented):
🎢 Red rover, red rover — let’s see if every slot lands back in its rightful place in line.

Nice fix overall. The `lastDemotedIndices` approach is the right level of simplicity for position-preserving re-promotion. A few items worth addressing before merge.

src/stores/promotionStore.ts
artokun: Oof good catch

src/core/graph/subgraph/promotionUtils.ts
artokun: Yes this needs tests and a comment explaining why better

browser_tests/tests/subgraphPromotionOrder.spec.ts
artokun: Good call, it will at least make it shorter

browser_tests/tests/subgraphPromotionOrder.spec.ts
artokun: I may just update the type correctly from the source, we'll likely need it again later

src/stores/promotionStore.ts
artokun: Added explanatory comment.

src/stores/promotionStore.ts
artokun: Extracted `makePositionKey` helper.

src/stores/promotionStore.ts
artokun: Added early return when `index === -1`.

src/stores/promotionStore.ts
artokun: Wrapped with `Array.from()`.

src/core/graph/subgraph/promotionUtils.ts
artokun: Added 3 unit tests for `getSourceNodeId` and an explanatory comment on why the fallback was removed.

src/stores/promotionStore.test.ts
artokun: Renamed to `initial promotes append in order`.

browser_tests/tests/subgraphPromotionOrder.spec.ts
artokun: Extracted `getPiniaStoreInBrowser` helper to `promotedWidgets.ts` as single breakage point.

browser_tests/tests/subgraphPromotionOrder.spec.ts
artokun: Using intersection type: `node as typeof node & { rootGraph?: { id: string } }`.

browser_tests/tests/subgraphPromotionOrder.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+
+      // Verify promoted widgets render in the Vue node body
+      const subgraphVueNode = comfyPage.vueNodes.getNodeLocator(nodeId)
+      await expect(subgraphVueNode).toBeVisible()
+
+      const nodeBody = subgraphVueNode.locator(
+        `[data-testid="node-body-${nodeId}"]`
+      )
+      const widgetElements = nodeBody.locator('.lg-node-widgets > div')
+      await expect(widgetElements.first()).toBeVisible()
```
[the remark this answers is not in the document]
artokun: Switched to `TestIds.widgets.container` from `fixtures/selectors.ts`.

artokun:
@coderabbitai review

browser_tests/tests/subgraphPromotionOrder.spec.ts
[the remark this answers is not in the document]
artokun: Already addressed — using `TestIds.widgets.container`.

src/stores/promotionStore.ts
```diff
@@ -136,7 +146,19 @@ export const usePromotionStore = defineStore('promotion', () => {
...
+    const lastIndex = lastDemotedIndices.get(positionKey)
+    lastDemotedIndices.delete(positionKey)
+
+    const nextEntries = [...entries]
+    if (lastIndex !== undefined) {
+      const insertAt = Math.min(lastIndex, nextEntries.length)
+      nextEntries.splice(insertAt, 0, entry)
+    } else {
+      nextEntries.push(entry)
+    }
```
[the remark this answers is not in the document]
artokun: Valid edge case — multi-demote then re-promote in the same pass can swap order. The current fix handles the common single-toggle path correctly. Filed as a known limitation; the manifest approach (tracking ghost entries) would handle this fully but is a larger refactor.

artokun:
@coderabbitai review

jaeone94 (commented):
.

jaeone94 (commented):
## Code Review

**Summary**: Fixes promoted widget order corruption during sidebar hide/show cycles by remembering the demoted index and restoring it on re-promote.

---

### 🏗️ Architecture

| Severity | Item |
|----------|------|
| ✅ Low | Declaring `lastDemotedIndices` as a non-reactive `Map` is appropriate — Vue reactivity is unnecessary for an ephemeral cache. |
| ⚠️ Medium | **`clearGraph` iteration pattern** — `Array.from(lastDemotedIndices.keys())` copies all keys then filters by `startsWith`, which is O(total cache size). Nesting as `Map<UUID, Map<string, number>>` (consistent with `graphPromotions`/`graphRefCounts`) would make `clearGraph` O(1). Not a problem at current scale, but a structural consistency suggestion. |

### ⚡ Performance

| Severity | Item |
|----------|------|
| ✅ OK | `findIndex`/`splice` in `promote`/`demote` operate on very small promotion lists (<20 items) — no performance concern. |
| ✅ OK | Removing the redundant `extractVueNodeData` re-call in `useGraphNodeManager.ts` eliminates unnecessary work. Positive change. |

### 🧪 Test Quality

| Severity | Item |
|----------|------|
| ✅ Good | `promotionStore.test.ts` — 5 scenarios (middle restore, first restore, clamp, initial append, clearGraph reset) provide solid behavioral coverage. |
| ✅ Good | `promotionUtils.test.ts` — 3 cases for the `getSourceNodeId` change are appropriate. |
| ✅ Good | E2E test correctly inlines the Pinia store access inside `page.evaluate` due to Playwright serialization boundary. The `getPiniaStoreInBrowser` helper serves as a reference pattern, not a callable import. |

### 🔄 Regression Risk

| Severity | Item |
|----------|------|
| ✅ Low | `getSourceNodeId` behavior change (`disambiguatingSourceNodeId ?? sourceNodeId` → `disambiguatingSourceNodeId` only) — the sole call site `SubgraphEditor.vue:180` handles `undefined` correctly via ternary. Low regression risk. |

### 🌐 Ecosystem Compatibility

| Severity | Item |
|----------|------|
| ✅ OK | `promote`/`demote` signatures unchanged. Only internal behavior modified (position restoration). No external compatibility impact. |

---

### 📋 Verdict

Core fix logic is solid with thorough test coverage. No blocking issues found.

src/stores/promotionStore.ts
```diff
@@ -188,6 +213,9 @@ export const usePromotionStore = defineStore('promotion', () => {
   function clearGraph(graphId: UUID): void {
     graphPromotions.value.delete(graphId)
     graphRefCounts.value.delete(graphId)
+    for (const key of Array.from(lastDemotedIndices.keys())) {
+      if (key.startsWith(`${graphId}:`)) lastDemotedIndices.delete(key)
+    }
```
jaeone94: nit (non-blocking): This iterates all keys to find entries for a single graph. A nested Map<UUID, Map<string, number>> (consistent with graphPromotions and graphRefCounts) would make this O(1). Not a real concern since this cache is ephemeral and tiny in practice.

jaeone94 (dismissed):
There is no blocking.

src/stores/promotionStore.ts
artokun: Good point — a nested Map would be cleaner. Keeping the flat Map for now since the cache is tiny and ephemeral, but happy to refactor if this grows.

artokun:
Rewrote the E2E test to drive the demote/re-promote cycle through the SubgraphEditor sidebar UI instead of calling store methods directly. The test now:

1. Opens the properties panel and SubgraphEditor
2. Clicks the visibility toggle in the Shown section (demote)
3. Clicks the toggle in the Hidden section (re-promote)
4. Asserts both the sidebar label order and promotion store order are preserved

@coderabbitai review

PR #10587, opened 2026-03-27
test: add SignIn dialog E2E tests (DLG-04)

christian-byrne:
## Summary
Adds Playwright E2E tests for the SignIn dialog component and its sub-forms.

## Tests added
- Dialog opens from login button in topbar
- Sign In form is the default view with email/password fields
- Toggle between Sign In and Sign Up forms
- API Key form navigation (forward and back)
- Terms of Service and Privacy Policy links present
- Form field presence verification
- Dialog close behavior (close button and Escape key)
- Forgot password link presence
- 'Or continue with' divider and API key button

## Notes
- Tests focus on UI navigation and element presence (no real Firebase auth in test env)
- Dialog opened via `extensionManager.dialog.showSignInDialog()` API
- All selectors use stable IDs from the component source (`#comfy-org-sign-in-email`, etc.)

## Task
Part of Test Coverage Q2 Overhaul (DLG-04).

## Conventions
- Uses Vue nodes with new menu enabled (`Comfy.UseNewMenu: 'Top'`)
- Tests read as user stories
- No full-page screenshots
- Proper waits, no sleeps

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10587-test-add-SignIn-dialog-E2E-tests-DLG-04-3306d73d3650815db171f8c5228e2cf3) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/dialogs/signInDialog.spec.ts
+++ b/browser_tests/tests/dialogs/signInDialog.spec.ts
@@ -12,7 +12,7 @@ async function openSignInDialog(page: Page) {
   await page.evaluate(() => {
     void window.app!.extensionManager.dialog.showSignInDialog()
   })
-  const dialog = page.locator('.p-dialog')
+  const dialog = page.getByRole('dialog')
   await dialog.waitFor({ state: 'visible' })
   return dialog
 }
--- a/browser_tests/tests/dialogs/signInDialog.spec.ts
+++ b/browser_tests/tests/dialogs/signInDialog.spec.ts
@@ -37,9 +37,7 @@ test.describe('Sign In dialog', { tag: '@ui' }, () => {
     await expect(dialog.locator('#comfy-org-sign-in-password')).toBeVisible()
 
     // Sign in submit button is present
-    await expect(
-      dialog.getByRole('button', { name: 'Sign in' })
-    ).toBeVisible()
+    await expect(dialog.getByRole('button', { name: 'Sign in' })).toBeVisible()
   })
 
   test('Should toggle from sign-in to sign-up form', async ({ comfyPage }) => {
@@ -97,9 +95,7 @@ test.describe('Sign In dialog', { tag: '@ui' }, () => {
     await dialog.getByRole('button', { name: 'Comfy API Key' }).click()
 
     // API Key form heading and input are visible
-    await expect(
-      dialog.getByRole('heading', { name: 'API Key' })
-    ).toBeVisible()
+    await expect(dialog.getByRole('heading', { name: 'API Key' })).toBeVisible()
     await expect(dialog.locator('#comfy-org-api-key')).toBeVisible()
 
     // Back button returns to the main sign-in view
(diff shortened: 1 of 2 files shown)


browser_tests/tests/dialogs/signInDialog.spec.ts
```diff
@@ -0,0 +1,168 @@
...
+    const dialog = await openSignInDialog(comfyPage.page)
+
+    // Default view is Sign In
+    await expect(
+      dialog.getByRole('heading', { name: 'Log in to your account' })
+    ).toBeVisible()
+
+    // Email and password fields are present
+    await expect(dialog.locator('#comfy-org-sign-in-email')).toBeVisible()
+    await expect(dialog.locator('#comfy-org-sign-in-password')).toBeVisible()
```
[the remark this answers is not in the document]
christian-byrne: Dismissing — these are HTML `id` attributes (e.g. `id="comfy-org-sign-in-email"`) used for `<label for="...">` form association in the source Vue components. They are not `data-testid` attributes. `TestIds` is specifically for `data-testid` selectors. Using `locator("#comfy-org-sign-in-email")` is the correct way to target form fields by their stable, component-defined id. Adding HTML ids to TestIds would be misleading about the selector contract.

AustinMroz (dismissed):
I get the nagging feeling that this could be a trick question and I'm supposed to mention fixtures, but this is the sort of test I'd want to see. Logic is reasonable, readable and self contained.

browser_tests/tests/dialogs/signInDialog.spec.ts
```diff
@@ -0,0 +1,164 @@
...
+
+import { expect } from '@playwright/test'
+
+import { comfyPageFixture as test } from '../../fixtures/ComfyPage'
+
+/**
+ * Opens the Sign In dialog via the extensionManager API.
+ * Returns a locator scoped to the dialog root.
+ */
+async function openSignInDialog(page: Page) {
```
christian-byrne: This should be a page object or helper

browser_tests/tests/dialogs/signInDialog.spec.ts
christian-byrne, line 1: There's too many useless code comments that don't express any info that isn't obvious from the code.

Also, more things should be page objects / helpers here. Audit the whole PR and refer to the markdown files in browser_tests folder for more guidance.

browser_tests/tests/dialogs/signInDialog.spec.ts
christian-byrne: fixed in 5cdf577 - extracted SignInDialog page object extending BaseDialog, following the same pattern as SettingDialog

browser_tests/tests/dialogs/signInDialog.spec.ts
christian-byrne: fixed in 88ad2a6 - removed all redundant comments and refactored test to use the SignInDialog page object throughout

PR #10588, opened 2026-03-27
test: add ShareWorkflow dialog E2E tests (DLG-05)

christian-byrne:
## Summary
Adds Playwright E2E tests for the ShareWorkflow dialog component and its various states.

## Tests added
- Dialog opens and shows unsaved state for new workflows (save prompt)
- Ready state shows create link button for saved but unpublished workflows
- Shared state shows copy URL field with share link after publishing
- Stale state shows update link button when workflow modified after publishing
- Close button dismisses dialog
- Create link transitions dialog from ready to shared state
- Tab switching between share link and publish to hub (when comfyHubUploadEnabled)
- Tab aria-selected states update correctly on switch

## Approach
- Share dialog is gated behind `isCloud` (compile-time constant), so tests invoke it directly via `page.evaluate()` importing `useShareDialog`
- Share service API calls (`/api/userdata/*/publish`, `/api/assets/from-workflow`) mocked via `page.route()` for deterministic state testing
- Dialog state (loading → unsaved → ready → shared → stale) controlled by mock responses
- Feature flags set via `serverFeatureFlags.value` for tab visibility testing

## Notes
- All pre-existing TS2307 errors are `.vue` module resolution — no new type errors
- Tests cover the 5 dialog states: loading, unsaved, ready, shared, stale

## Task
Part of Test Coverage Q2 Overhaul (DLG-05).

## Conventions
- Uses Vue nodes with new menu enabled
- Tests read as user stories
- No full-page screenshots
- Proper waits, no sleeps
- All API calls mocked

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10588-test-add-ShareWorkflow-dialog-E2E-tests-DLG-05-3306d73d365081a0ab15f333707e493b) by [Unito](https://www.unito.io)


browser_tests/tests/dialogs/shareWorkflowDialog.spec.ts
```diff
@@ -0,0 +1,416 @@
...
+
+  // After save, the detach/attach cycle and graph serialization differences
+  // can cause changeTracker.checkState() to set isModified back to true.
+  // Force isModified=false and reset the changeTracker so the share dialog
+  // sees a clean saved state.
+  await comfyPage.page.evaluate(() => {
+    const wf = (window.app!.extensionManager as WorkspaceStore).workflow
+      .activeWorkflow
+    if (wf) {
+      wf.isModified = false
```
dante01yoon: Question: Like `store.renameWorkflow()`, This function directly manipulates store internals mocks the userdata POST route, and re-opens workflows to work around reactivity issues. This is the opposite of what E2E tests should do — it's tightly coupled to implementation details and will break if any internal API changes.

The existing `WorkflowHelper.loadWorkflow()` or `comfyPage.menu.topbar.saveWorkflow()` should be used. If cloud mode makes save unreliable in tests, that's a bug worth investigating separately rather than papering over with store manipulation.

browser_tests/tests/dialogs/shareWorkflowDialog.spec.ts
```diff
@@ -0,0 +1,416 @@
...
+      .first()
+      .waitFor({ state: 'hidden', timeout: 2000 })
+      .catch(() => {})
+  }
+  // Force-remove any stale masks that CSS transitions left behind
+  if ((await mask.count()) > 0) {
+    await page.evaluate(() => {
+      document
+        .querySelectorAll('.p-dialog-mask')
+        .forEach((el) => el.remove())
```
dante01yoon: Issue: Forcefully removing DOM elements from the page to make tests pass can be a red flag. This masks real UI bugs (leaked dialog masks) and creates tests that pass in CI but don't reflect actual user experience. If a stale dialog mask lingers, that's a bug to fix in the component, not work around in tests.

browser_tests/tests/dialogs/shareWorkflowDialog.spec.ts
```diff
@@ -0,0 +1,416 @@
...
+
+async function openShareDialog(page: Page): Promise<void> {
+  await enableWorkflowSharing(page)
+  await dismissOverlays(page)
+  const shareButton = page.getByRole('button', { name: 'Share workflow' })
+  await shareButton.click()
+}
+
+function getShareDialog(page: Page) {
+  return page.locator('.p-dialog')
```
dante01yoon: Issue: uses fragile CSS class selector. This matches any PrimeVue dialog on the page. If another dialog is open (which dismissOverlays suggests is possible), this will match the wrong element. The dialog should have a data-testid or be located more        specifically, e.g. via the dialog's unique title/heading.

dante01yoon (changes requested):
Add comment and issues.

Feature flag set via direct mutation, not via FeatureFlagHelper         `shareWorkflowDialog.spec.ts:35-40`, `shareWorkflowDialog.spec.ts:218-224`

browser_tests/tests/dialogs/shareWorkflowDialog.spec.ts
christian-byrne: fixed - saveAndWait now uses comfyPage.menu.topbar.saveWorkflow() instead of store manipulation. removed the WorkspaceStore import too.

browser_tests/tests/dialogs/shareWorkflowDialog.spec.ts
christian-byrne: fixed - dismissOverlays now uses an Escape key retry loop (up to 3 attempts) instead of page.evaluate DOM removal. waits for the mask to disappear after each press.

browser_tests/tests/dialogs/shareWorkflowDialog.spec.ts
christian-byrne: fixed - getShareDialog now uses page.getByRole('\''dialog'\'') instead of the .p-dialog-content CSS selector.

browser_tests/tests/dialogs/shareWorkflowDialog.spec.ts
```diff
@@ -0,0 +1,352 @@
...
+import type { AssetInfo } from '../../../src/schemas/apiSchema'
+import { comfyPageFixture } from '../../fixtures/ComfyPage'
+import { TestIds } from '../../fixtures/selectors'
+
+interface PublishRecord {
+  workflow_id: string
+  share_id: string | null
+  listed: boolean
+  publish_time: string | null
+}
```
jaeone94, line 13: nit: The inline PublishRecord interface duplicates zPublishRecordResponse from src/platform/workflow/sharing/schemas/shareSchemas.ts (missing the assets field). 
Per AGENTS.md guidance ("Type all API mock responses using generated types or schemas"), consider using z.infer<typeof zPublishRecordResponse> instead to prevent drift.

jaeone94 (approved):
Great to see the path opening for cloud environment E2E testing! The review agent flagged CloudAuthHelper being added as a ComfyPage property per the playwright.md rule, but this looks intentional to me — cloud auth is infrastructure-level for OSS→cloud test expansion, consistent with the 16 other helpers already on ComfyPage.

browser_tests/tests/dialogs/shareWorkflowDialog.spec.ts
```diff
@@ -0,0 +1,352 @@
...
+    })
+  })
+}
+
+/**
+ * Dismiss stale PrimeVue dialog masks left by cloud-mode's onboarding flow
+ * or auth-triggered modals by pressing Escape until they clear.
+ */
+async function dismissOverlays(page: Page): Promise<void> {
+  const mask = page.locator('.p-dialog-mask')
```
jaeone94, line 106: nit: I think JaeWon has already pointed this out, but it would probably be good to fix this part as well.

jaeone94 (approved):

dante01yoon (approved):

PR #10592, opened 2026-03-27
fix: virtualize cloud job queue history list

benceruleanlu:
## Summary

Virtualize the shared job queue history list so opening the jobs panel does not eagerly mount the full history on cloud.

## Changes

- **What**: Virtualize the shared queue history list used by the overlay and sidebar, flatten date headers plus job rows into a single virtual stream, and preserve hover/menu behavior with updated queue list tests.
- **Why `@tanstack/vue-virtual` instead of Reka virtualizers**: the installed `reka-ui@2.5.0` does not expose a generic list virtualizer. It only exposes `ListboxVirtualizer`, `ComboboxVirtualizer`, and `TreeVirtualizer`, and those components inject `ListboxRoot`/`TreeRoot` context and carry listbox or tree selection/keyboard semantics. The job history UI is a flat grouped action list, not a selectable listbox or navigable tree, so this uses the same TanStack virtualizer layer directly without forcing the wrong semantics onto the component.

## Review Focus

Please verify the virtual row sizing and inter-group spacing behavior across date headers and the last row in each group.

> [!TIP]
> Diff reads much cleaner through vscode's unified view with show leading/trailing whitespace differences enabled

Linear: COM-304

https://tanstack.com/virtual/latest/docs/api/virtualizer

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10592-fix-virtualize-cloud-job-queue-history-list-3306d73d3650819d956bf4b2d8b59a40) by [Unito](https://www.unito.io)


src/components/queue/job/JobAssetsList.vue
```diff
@@ -1,79 +1,91 @@
...
-          @dblclick.stop="emitViewItem(job)"
-          @preview-click="emitViewItem(job)"
-          @click.stop
+  <div
+    :ref="containerProps.ref"
+    :style="containerProps.style"
+    class="h-full overflow-y-auto"
+    @scroll="onListScroll"
+  >
+    <div :style="virtualWrapperStyle">
```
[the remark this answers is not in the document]
benceruleanlu: Fixed in `daa8782f8`. `JobAssetsList` now opts out of automatic fallthrough with `inheritAttrs: false` and forwards `$attrs` onto the scroll container, so the parent sizing classes are applied again. I also added a focused unit test covering attr forwarding to the scroll container.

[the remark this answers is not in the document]
benceruleanlu: Done

browser_tests/tests/sidebar/jobHistory.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  return Array.from({ length: count }, (_, index) => ({
+    id: `job-${index}`,
+    status: 'completed',
+    create_time: now - index * 60_000,
+    execution_start_time: now - index * 60_000 - 10_000,
+    execution_end_time: now - index * 60_000,
+    workflow_id: 'workflow-1'
+  }))
+}
```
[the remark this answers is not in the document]
benceruleanlu: Risk seems pretty low, can be done later

src/components/queue/job/JobAssetsList.test.ts
```diff
@@ -8,6 +8,30 @@ import { ResultItemImpl, TaskItemImpl } from '@/stores/queueStore'
 
 import JobAssetsList from './JobAssetsList.vue'
 
+vi.mock('@vueuse/core', async () => {
```
jaeone94: **nit (non blocking):** This `useVirtualList` mock block is duplicated verbatim in `JobHistorySidebarTab.test.ts:8`. Consider extracting to a shared test utility to avoid drift if the mock contract changes.

browser_tests/tests/sidebar/jobHistory.spec.ts
```diff
@@ -0,0 +1,127 @@
+import { expect } from '@playwright/test'
+import type { Locator } from '@playwright/test'
+
+import type { RawJobListItem } from '../../../src/platform/remote/comfyui/jobs/jobTypes'
+import { comfyPageFixture as test } from '../../fixtures/ComfyPage'
+
+const VIRTUALIZED_ROW_LIMIT = 48
+const HEADER_ROW_HEIGHT = 20
+const JOB_ROW_HEIGHT = 48
+const GROUP_ROW_GAP = 16
```
jaeone94: **nit (non blocking) :** `HEADER_ROW_HEIGHT`, `JOB_ROW_HEIGHT`, and `GROUP_ROW_GAP` are already exported from `buildVirtualJobRows.ts` — importing them instead of re-declaring would keep a single source of truth.

jaeone94 (dismissed):
Overall, this is a well-structured PR. Only very minor nits were raised.

browser_tests/tests/sidebar/jobHistory.spec.ts
christian-byrne, line 1: I think the idea of using a library (and one that supports differently sized items in the virtual list) is that we don't need to test this ourselves so we can remove the aspects of this test that do that. The helpers might be useful for testing behavior specific to our app though.

christian-byrne (commented):
Could you show a video of how it feels to use?

browser_tests/tests/sidebar/jobHistory.spec.ts
benceruleanlu: vue use's virtual does support differently sized items in the virtual list, if you meant dynamic sizing, I did just switch to tanstack virtual, however, because our system is already static sized, it's simpler and most likely more performant than relying on their size estimation that is backed by getClient... and ResizeObserver. Although I'll remove the tests that 'test the library'.

src/components/sidebar/tabs/JobHistorySidebarTab.test.ts
benceruleanlu, line 1: This test is redundant with the lower-level hover/details-popover coverage in `src/components/queue/job/JobAssetsList.test.ts`. Removing it keeps the sidebar tab focused on its own shell behavior instead of re-testing `JobAssetsList` through a heavier integration path.

src/components/queue/job/JobAssetsList.test.ts
benceruleanlu: Done

browser_tests/tests/sidebar/jobHistory.spec.ts
benceruleanlu: Done

browser_tests/tests/sidebar/jobHistory.spec.ts
benceruleanlu: Video in Slack, decided to just move the e2e test after fixtures land

jaeone94:
Will E2E tests be added in a future PR?

benceruleanlu:
> Will E2E tests be added in a future PR?

Yes I'm wrapping up the fixtures additions that will make this much easier, the team also disliked the idea of e2e tests in this PR because it would be 'testing the library'

jaeone94 (approved):
Clean virtualization implementation. TanStack Virtual is a solid choice, the `buildVirtualJobRows` extraction keeps the logic testable, and the shared mock utility addresses prior feedback. CI is green, no blocking issues.

Two minor nits (non-blocking):

1. **Implicit group gap via row height** — `getRowHeight` adds `GROUP_ROW_GAP` to the last job row before a header rather than using a dedicated spacer row. It works, but a spacer row would make the intent more explicit and keep row heights purely about their own content.

2. **`buildVirtualJobRows` style** — `forEach` with mutable push is fine, but `flatMap` would be a more idiomatic fit here since the function is otherwise pure.

LGTM

PR #10595, opened 2026-03-27
test: add assets sidebar empty-state coverage

benceruleanlu:
## Summary

Add the first user-centric Playwright coverage for the assets sidebar empty state and introduce a small assets-specific test helper/page object surface.

## Changes

- **What**: add `AssetsSidebarTab`, add `AssetsHelper`, and cover generated/imported empty states in a dedicated browser spec

## Review Focus

This is intentionally a small first slice for assets-sidebar coverage. The new helper still mocks the HTTP boundary in Playwright for now because current OSS job history and input files are global backend state, which makes true backend-seeded parallel coverage a separate backend change.

Long-term recommendation: add backend-owned, user-scoped test seeding for jobs/history and input assets so browser tests can hit the real routes on a shared backend. Follow-up: COM-307.

Fixes COM-306

Not applicable.

## Validation

- `pnpm typecheck:browser`
- `pnpm exec playwright test browser_tests/tests/sidebar/assets.spec.ts --project=chromium` against an isolated preview env

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10595-test-add-assets-sidebar-empty-state-coverage-3306d73d365081d1b34fdd146ae6c5c6) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/fixtures/ComfyPage.ts
+++ b/browser_tests/fixtures/ComfyPage.ts
@@ -19,10 +19,12 @@ import { ContextMenu } from './components/ContextMenu'
 import { SettingDialog } from './components/SettingDialog'
 import { BottomPanel } from './components/BottomPanel'
 import {
+  AssetsSidebarTab,
   NodeLibrarySidebarTab,
   WorkflowsSidebarTab
 } from './components/SidebarTab'
 import { Topbar } from './components/Topbar'
+import { AssetsHelper } from './helpers/AssetsHelper'
 import { CanvasHelper } from './helpers/CanvasHelper'
 import { PerformanceHelper } from './helpers/PerformanceHelper'
 import { QueueHelper } from './helpers/QueueHelper'
@@ -55,6 +57,7 @@ class ComfyPropertiesPanel {
 }
 
 class ComfyMenu {
+  private _assetsTab: AssetsSidebarTab | null = null
   private _nodeLibraryTab: NodeLibrarySidebarTab | null = null
   private _workflowsTab: WorkflowsSidebarTab | null = null
   private _topbar: Topbar | null = null
@@ -76,6 +79,11 @@ class ComfyMenu {
   get nodeLibraryTab() {
     this._nodeLibraryTab ??= new NodeLibrarySidebarTab(this.page)
     return this._nodeLibraryTab
+  }
+
+  get assetsTab() {
+    this._assetsTab ??= new AssetsSidebarTab(this.page)
+    return this._assetsTab
   }
 
   get workflowsTab() {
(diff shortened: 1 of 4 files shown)


christian-byrne (approved):
LGTM. Clean assets sidebar empty-state test with well-structured helpers.

**Strengths:**
- `AssetsHelper` is production-quality: handles job list pagination, filtering by status/workflow_id, sorting, offset/limit — well beyond what the current test needs, but reusable for future asset tests
- `AssetsSidebarTab` extends the existing `SidebarTab` base class properly
- Integrated into `ComfyPage` as `comfyPage.assets` and `comfyPage.menu.assetsTab`
- Test uses `mockEmptyState()` for deterministic state
- Proper cleanup in `afterEach` with `clearMocks()`
- Uses role-based selectors (`getByRole('tab')`, `getByText`) — not CSS classes

**Minor observations (non-blocking):**
- The `importedFiles` mock returns a string array but the actual `/internal/files/input` API might return a more complex structure — verify against real API if extending these tests
- Consider exporting `AssetsHelper` and `AssetsSidebarTab` from a barrel file if other specs will use them

All shards pass. Approving.

benceruleanlu:
> The importedFiles mock returns a string array but the actual /internal/files/input API might return a more complex structure — verify against real API if extending these tests

Goal is for this to be fixed by COM-307, we'll decide in the meeting but I think it's the right direction

christian-byrne (changes requested):
## Review Findings

### [m1] `emptyStateTitle(title)` is a zero-value abstraction
**File:** `browser_tests/fixtures/components/SidebarTab.ts:191`

`emptyStateTitle(title)` just delegates to `page.getByText(title)` — the caller still passes the exact string, so nothing is encapsulated. Compare with `emptyStateMessage` which encapsulates a known string, and `generatedTab`/`importedTab` which encapsulate role-based selectors.

**Suggestion:** Replace with parameterless getters that encapsulate the expected strings:
```typescript
get generatedEmptyTitle() {
  return this.page.getByText('No generated files found')
}

get importedEmptyTitle() {
  return this.page.getByText('No imported files found')
}
```
Then update the test to use `tab.generatedEmptyTitle` / `tab.importedEmptyTitle`.

---

### [m2] Route pattern inconsistency — regex vs glob
**File:** `browser_tests/fixtures/helpers/AssetsHelper.ts:4-5`

Uses regex patterns (`/\/api\/jobs(?:\?.*)?$/`) while every other helper (`QueueHelper`, `FeatureFlagHelper`) uses glob patterns (`'**/api/queue'`, `'**/api/history**'`, `'**/api/features'`).

**Suggestion:** Use glob patterns to match convention: `'**/api/jobs**'` and `'**/internal/files/input**'`.

---

### [N1] Handler type inconsistency — async vs sync
**File:** `browser_tests/fixtures/helpers/AssetsHelper.ts:35-36`

Handler types are `((route: Route) => Promise<void>)` (async) while `QueueHelper` and `FeatureFlagHelper` use `((route: Route) => void)` (sync). Since the simplified handlers just call `route.fulfill()` without `await`, they can be sync to match convention.

PR #10596, opened 2026-03-27
test: add minimap e2e tests for close button, viewport, and pan

jtydhr88:
## Summary
add more basic tests for minimap

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10596-test-add-minimap-e2e-tests-for-close-button-viewport-and-pan-3306d73d365081b9bf64dc7a3951d65f) by [Unito](https://www.unito.io)


christian-byrne (approved):
LGTM. Solid minimap extension tests.

**Strengths:**
- Extends existing `minimap.spec.ts` rather than creating a new file — good convention
- Close button test uses `getByTestId('close-minmap-button')` (note: typo 'minmap' is in the source component, not the test's fault)
- Pan test manipulates canvas.ds directly via page.evaluate — deterministic, no flake risk
- Viewport rectangle test validates bounding box containment with clear geometric assertions
- Screenshots scoped to minimap element

**Minor observations (non-blocking):**
- `minimap-before-pan` and `minimap-with-viewport` snapshots appear identical — expected since both are pre-pan state
- Uses `getByTestId` for close button but `TestIds.canvas.toggleMinimapButton` for the toggle — would be nice to add the close button to TestIds too, but that requires a source change

All shards pass. Approving.

browser_tests/tests/minimap.spec.ts
```diff
@@ -78,4 +78,69 @@ test.describe('Minimap', { tag: '@canvas' }, () => {
 
     await expect(minimapContainer).toBeVisible()
   })
+
+  test('Close button hides minimap', async ({ comfyPage }) => {
+    const minimap = comfyPage.page.locator('.litegraph-minimap')
+    await expect(minimap).toBeVisible()
+
+    await comfyPage.page.getByTestId('close-minmap-button').click()
```
christian-byrne: issue: The hardcoded test ID `'close-minmap-button'` has a typo ("minmap" instead of "minimap") -- and the typo exists in the source (`MiniMap.vue:43`) too. Additionally, per test conventions this should be added to `TestIds` in `selectors.ts` rather than hardcoded. Fix both the typo in source and centralize the test ID.

browser_tests/tests/minimap.spec.ts
```diff
@@ -78,4 +78,69 @@ test.describe('Minimap', { tag: '@canvas' }, () => {
...
+      await expect(minimap).toHaveScreenshot('minimap-before-pan.png')
+
+      await comfyPage.page.evaluate(() => {
+        const canvas = window.app!.canvas
+        canvas.ds.scale = 3
+        canvas.ds.offset[0] = -800
+        canvas.ds.offset[1] = -600
+        canvas.setDirty(true, true)
+      })
+      await comfyPage.nextFrame()
```
christian-byrne: suggestion (non-blocking): Two consecutive `nextFrame()` calls look like a timing workaround. If the minimap genuinely needs two frames to settle after programmatic `ds.offset`/`ds.scale` changes, add a brief comment explaining why. Otherwise consider using a proper wait condition (e.g., waiting for the viewport element's position to change).

browser_tests/tests/minimap.spec.ts
```diff
@@ -78,4 +78,69 @@ test.describe('Minimap', { tag: '@canvas' }, () => {
 
     await expect(minimapContainer).toBeVisible()
   })
+
+  test('Close button hides minimap', async ({ comfyPage }) => {
+    const minimap = comfyPage.page.locator('.litegraph-minimap')
```
christian-byrne, line 83: suggestion (non-blocking): Raw CSS class selectors (`.litegraph-minimap`, `.minimap-canvas`, `.minimap-viewport`) are used throughout. Consider adding these to `TestIds` in `selectors.ts` for stability -- CSS classes can change during refactors while test IDs are explicitly stable.

christian-byrne (changes requested):
issue: The shard 4 CI failure (`nodeSearchBox > Can add node`) is caused by this PR. The minimap is now visible by default, contaminating the full-canvas screenshot `added-node.png`. The automated baseline update was committed but CI still fails with repeated screenshot mismatches (4 retries). Either the `nodeSearchBox` test needs to hide the minimap in its `beforeEach`, or the minimap rendering has a timing instability that needs a proper wait.

browser_tests/tests/minimap.spec.ts
jtydhr88: fixed

christian-byrne (approved):

PR #10613, opened 2026-03-27
fix: prevent scroll event leak from text widgets to canvas

### LINKED ISSUE #3990: [Bug]: Scroll event leak after scrolling to the top of a text widget

phusi319:
## Summary

Fixes #3990 — scroll events from scrollable text widgets (textarea, markdown, text preview) were leaking to the canvas layer, causing unwanted zoom/pan when scrolling widget content.

## Root Cause

TransformPane's capture-phase wheel handler forwarded **all** wheel events to the canvas unless the widget element was both focused and marked with data-capture-wheel. This meant:

- **TextPreviewWidget** (used by Preview Any nodes): No data-capture-wheel attribute — events always forwarded — canvas zooms instead of text scrolling
- **Unfocused widgets**: Even widgets with data-capture-wheel had events forwarded when not focused — first scroll always went to canvas
- **useStringWidget** (legacy path): At scroll boundaries, events were silently consumed instead of forwarding to canvas
- **useMarkdownWidget** (legacy path): No wheel handler at all — all events leaked

### 1. useCanvasInteractions.ts — Core fix
Added wheelCapturedByScrollableElement() to shouldForwardWheelEvent. This detects when a wheel event target is inside a scrollable data-capture-wheel element:
- **Within scroll bounds** — event stays in widget (not forwarded to canvas)
- **At scroll boundary** (top/bottom edge) — event passes through to canvas for zoom/pan
- **Not scrollable** — event forwards to canvas (existing behavior preserved)

This works regardless of focus state, fixing the issue where unfocused scrollable widgets had their scroll events stolen by the canvas.

### 2. TextPreviewWidget.vue
Added data-capture-wheel="true" to the scrollable container div.

### 3. useMarkdownWidget.ts — Legacy path
Added a complete wheel event handler (previously had none):
- Ctrl+wheel forwards to canvas for zoom
- Within scroll range consumes event (widget scrolls natively)
- At boundary forwards to canvas
- Not scrollable forwards to canvas

### 4. useStringWidget.ts — Legacy path
Added scroll boundary detection to the existing canScrollY branch. Previously, it only called stopPropagation() which silently consumed events at boundaries. Now:
- At boundary — explicitly forwards to canvas via processMouseWheel
- Within scroll range — existing behavior preserved (native scroll)

## Backward Compatibility

- All existing data-capture-wheel + focus behavior preserved
- Ctrl+wheel zoom through focused widgets still works in standard mode
- Non-scrollable elements still forward events to canvas
- Trackpad gesture handling unchanged

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10613-fix-prevent-scroll-event-leak-from-text-widgets-to-canvas-3306d73d36508143b36afb702ace42aa) by [Unito](https://www.unito.io)

### DIFF
--- a/src/components/graph/widgets/TextPreviewWidget.vue
+++ b/src/components/graph/widgets/TextPreviewWidget.vue
@@ -1,6 +1,7 @@
 <template>
   <div
     class="relative max-h-[200px] min-h-[28px] w-full overflow-y-auto rounded-lg px-4 py-2 text-xs"
+    data-capture-wheel="true"
   >
     <div class="flex items-center gap-2">
       <div class="flex flex-1 items-center gap-2 break-all">
--- a/src/renderer/core/canvas/useCanvasInteractions.test.ts
+++ b/src/renderer/core/canvas/useCanvasInteractions.test.ts
@@ -46,12 +46,21 @@ function createMockPointerEvent(
   return mockEvent as PointerEvent
 }
 
-function createMockWheelEvent(ctrlKey = false, metaKey = false): WheelEvent {
+function createMockWheelEvent(
+  ctrlKey = false,
+  metaKey = false,
+  deltaY = 0,
+  target?: HTMLElement
+): WheelEvent {
   const mockEvent: Partial<WheelEvent> = {
     ctrlKey,
     metaKey,
+    deltaY,
     preventDefault: vi.fn(),
     stopPropagation: vi.fn()
+  }
+  if (target) {
+    Object.defineProperty(mockEvent, 'target', { value: target })
   }
   return mockEvent as WheelEvent
 }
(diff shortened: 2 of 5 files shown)


DrJKL:
Thank you for the work on this and for adding focused coverage. We are closing this version because forwarding outward wheel events at the top and bottom still produces the behavior reported in #3990. Preview Any has also moved to the newer WidgetTextPreview path since this was opened. We will handle the fix against the current widget and central wheel-routing architecture.

PR #10614, opened 2026-03-27
perf: layer compositing performance

LittleSound:
## Summary

By calculating and adjusting the size of the div containing the canvas, avoid the browser rendering excessive layers. Reduce layer compositing performance overhead. This significantly improves the smoothness of browsing large workflows.

### Before
image

### After

image


src/renderer/core/layout/transform/useContentBounds.ts
```diff
@@ -0,0 +1,107 @@
...
+ * Tracks the bounding box of all canvas content and computes an offset
+ * that shifts all positions into positive coordinate space.
+ *
+ * This enables the TransformPane's CSS box to fully contain all child
+ * nodes, preventing Chrome's compositor from creating individual
+ * compositing layers for overflowing children.
+ *
+ * Uses a grow-only strategy: bounds expand when content moves beyond
+ * the tracked area but never shrink automatically. Call `reset()` to
+ * recalculate from scratch (e.g., on workflow load).
```
christian-byrne, line 28: question: The JSDoc mentions `reset()` for workflow load, but `reset()` is never called by `TransformPane`. Over a long session with workflow switches, the grow-only bounds will accumulate the max extent of every workflow ever loaded. Is there a plan to wire `reset()` to workflow switch events? Without it, the pane's CSS width/height could grow unbounded -- which works against the compositor optimization this PR introduces.

src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -23,32 +33,69 @@ interface TransformPaneProps {
...
+ * When the layout store version changes, expand the tracked content
+ * bounds to include any nodes that moved beyond the current area.
+ * Uses grow-only strategy: bounds expand but never shrink.
+ */
+function updateContentBounds() {
+  const currentVersion = storeVersion.value
+  if (currentVersion === lastTrackedVersion) return
+  lastTrackedVersion = currentVersion
+
+  for (const [, layout] of allNodes.value) {
```
christian-byrne: question: `updateContentBounds()` iterates all nodes on every version bump. During node dragging the version increments each frame, making this O(n) per frame. Additionally `layoutStore.getAllNodes()` materializes a full Map from Yjs entries on each `.value` access. Have you considered a differential approach (e.g. only processing changed node IDs per version bump, or maintaining a store-level bounding box updated incrementally)?

christian-byrne (commented):
suggestion (non-blocking): The `@example` block in `useTransformState.ts` (lines 36-50) shows `<div :style="transformStyle">` as the recommended usage. This PR replaces that pattern with direct DOM mutation via `applyStyles()` in TransformPane, so the example now teaches a deprecated approach. Worth updating the example or adding a note about the new pattern.

src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -23,32 +33,69 @@ interface TransformPaneProps {
...
 
 const canvasElement = computed(() => props.canvas?.canvas)
 const { isTransforming: isInteracting } = useTransformSettling(canvasElement, {
   settleDelay: 256
 })
 
 const transformPaneRef = useTemplateRef('transformPaneRef')
+const offsetWrapperRef = useTemplateRef('offsetWrapperRef')
+
+const contentBounds = useContentBounds()
```
christian-byrne, line 47: suggestion (non-blocking): The version-tracking logic (`lastTrackedVersion`, `updateContentBounds`, and the `layoutStore` dependency) is imperative business logic that's hard to test via component tests. Consider moving it into `useContentBounds` (or a wrapper composable) so TransformPane just calls something like `contentBounds.update(allNodes)` in the RAF loop. This would align with the established pattern where state derivation lives in composables.

src/renderer/core/layout/__tests__/TransformPane.test.ts
```diff
@@ -165,12 +185,8 @@ describe('TransformPane', () => {
 
       const transformPane = wrapper.find('[data-testid="transform-pane"]')
 
-      // Simulate pointer down - we can't test the exact delegation logic
-      // in unit tests due to vue-test-utils limitations, but we can verify
-      // the event handler is set up correctly
       await transformPane.trigger('pointerdown')
 
-      // The test passes if no errors are thrown during event handling
       expect(transformPane.exists()).toBe(true)
```
christian-byrne: suggestion (non-blocking): This test triggers `pointerdown` and only asserts the element still exists. It doesn't test any observable side-effect of pointer event handling and gives false coverage confidence. Consider either testing an actual behavioral outcome or removing the test.

src/renderer/core/layout/__tests__/TransformPane.test.ts
```diff
@@ -96,6 +105,17 @@ describe('TransformPane', () => {
       expect(wrapper.find('.test-content').exists()).toBe(true)
       expect(wrapper.find('.test-content').text()).toBe('Test Node')
     })
+
+    it('should contain an offset wrapper div', () => {
```
christian-byrne: suggestion (non-blocking): This test asserts that a child div with `.absolute` exists -- pure DOM structure assertion that will break on any restructure without indicating a behavioral regression. The wrapper's purpose is already tested via `useContentBounds.test.ts`. Consider removing.

src/renderer/core/layout/__tests__/TransformPane.test.ts
```diff
@@ -185,8 +201,6 @@ describe('TransformPane', () => {
       })
 
       const transformState = useTransformState()
```
christian-byrne: suggestion (non-blocking): This test calls the mocked `useTransformState()` and asserts methods are `toBeDefined()`. It only verifies the mock factory returns what it was configured to return -- it doesn't exercise any real component behavior. Consider removing.

src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -23,32 +33,69 @@ interface TransformPaneProps {
...
  *
  * These values change every animation frame during zoom or pan.
- * If they were bound in the template, Vue would diff the entire
- * TransformPane vnode—including all child node slots—on every frame,
- * causing expensive vdom patch work across the full node list.
- * Mutating the DOM directly limits the update to a single element.
+ * Mutating the DOM directly limits the update to two elements,
+ * avoiding expensive vdom patch work across the full node list.
  */
+function applyStyles() {
```
christian-byrne, line 78: suggestion (non-blocking): The core behavioral change -- `applyStyles()` subtracting offset from camera and applying offset to the wrapper -- has no direct test coverage. The integration between TransformPane and useContentBounds is untested. Consider a test that mounts TransformPane with a mocked layoutStore containing negative-coordinate nodes, advances RAF, and asserts the transform styles include offset-adjusted values.

src/renderer/core/layout/transform/useContentBounds.ts
```diff
@@ -0,0 +1,107 @@
...
+      maxX = bRight + EXPAND_MARGIN
+      dirty = true
+    }
+    if (bBottom > maxY) {
+      maxY = bBottom + EXPAND_MARGIN
+      dirty = true
+    }
+  }
+
+  function expandToIncludePoint(point: Point) {
```
christian-byrne: nitpick (non-blocking): `expandToIncludePoint` is only used in its own test. Per YAGNI, consider removing until there's an actual consumer.

src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -23,32 +33,69 @@ interface TransformPaneProps {
 
 const props = defineProps<TransformPaneProps>()
```
christian-byrne: nitpick (non-blocking): Project convention prefers reactive props destructuring: `const { canvas } = defineProps<TransformPaneProps>()` rather than `const props = defineProps<...>()`.

src/renderer/core/layout/transform/useContentBounds.ts
LittleSound: Fixed

LittleSound:
Updated the `@example`

LittleSound:
Re: `@example` block in `useTransformState.ts` showing deprecated pattern —

Updated in 20198f1 — the example now shows direct DOM mutation with `syncWithCanvas` + `el.style.transform`.

src/renderer/core/layout/transform/useContentBounds.ts
LittleSound: Addressed in 20198f1. TransformPane now samples a node ID each version tick — when the sampled node disappears from the map, it treats it as a wholesale node-set replacement and calls `reset()`. This keeps the grow-only strategy for normal operations while preventing unbounded growth across workflow switches.

src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -57,10 +106,19 @@ watch([isInteracting, transformPaneRef], ([interacting, el]) => {
...
+    syncWithCanvas(canvas)
+
+    if (needsFullScan) {
+      needsFullScan = false
+      contentBounds.initialize(layoutStore.getAllNodes().value)
+    } else if (pendingChangedIds.size > 0) {
+      contentBounds.updateChanged(pendingChangedIds, getNodeLayout)
     }
-    syncWithCanvas(props.canvas)
+    pendingChangedIds = new Set()
```
christian-byrne, line 118: suggestion (non-blocking): `pendingChangedIds = new Set()` allocates a new Set every RAF frame (~60/s) even when idle. `.clear()` on the existing set avoids GC pressure and is safe here since `updateChanged` reads the set synchronously before this line.

src/renderer/core/layout/transform/TransformPane.vue
```diff
@@ -2,53 +2,102 @@
...
+  } else {
+    for (const id of change.nodeIds) {
+      pendingChangedIds.add(id)
+    }
+  }
+})
+onUnmounted(unsubscribe)
+
+function getNodeLayout(id: string) {
+  return layoutStore.getNodeLayoutRef(id).value
```
christian-byrne, line 65: question: `layoutStore.getNodeLayoutRef(id)` creates a new `customRef` per invocation. In `updateChanged`, this runs for every changed node every frame, allocating a throwaway reactive wrapper each time. Does the store expose a non-reactive accessor that would be more appropriate for this hot path?

PR #10623, opened 2026-03-27
test: Add test to prevent regression of workflow corruption during graph loading

pythongosssss:
## Summary

Adds regression test for https://github.com/Comfy-Org/ComfyUI_frontend/pull/9531 

## Changes

- **What**:  
- registers extension that triggers checkState during afterConfigureGraph (at this point the workflow data and active graph are not in sync), previously causing it to overwrite the workflow data
- switches between tabs
- ensures they are not corrupted

Line 35 can be uncommented to cause the test to fail by clearing the isLoadingGraph flag

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10623-test-Add-test-to-prevent-regression-of-workflow-corruption-during-graph-loading-3306d73d3650815fbf02ef23dfdcddce) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/tests/changeTrackerLoadGuard.spec.ts
+++ b/browser_tests/tests/changeTrackerLoadGuard.spec.ts
@@ -1,0 +1,66 @@
+import {
+  comfyExpect as expect,
+  comfyPageFixture as test
+} from '../fixtures/ComfyPage'
+import type { WorkspaceStore } from '../types/globals'
+
+test.describe(
+  'Change Tracker - isLoadingGraph guard',
+  { tag: '@workflow' },
+  () => {
+    test.beforeEach(async ({ comfyPage }) => {
+      await comfyPage.workflow.setupWorkflowsDirectory({})
+    })
+
+    test('Prevents checkState from corrupting workflow state during tab switch', async ({
+      comfyPage
+    }) => {
+      // Tab 0: default workflow (7 nodes)
+      expect(await comfyPage.nodeOps.getGraphNodesCount()).toBe(7)
+
+      // Save tab 0 so it has a unique name for tab switching
+      await comfyPage.menu.topbar.saveWorkflow('workflow-a')
+
+      // Register an extension that forces checkState during graph loading.
+      // This simulates the bug scenario where a user clicks during graph loading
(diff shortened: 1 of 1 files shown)


browser_tests/tests/changeTrackerLoadGuard.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+              .workflow.activeWorkflow
+            if (!workflow) throw new Error('No workflow found')
+            // Bypass the guard to reproduce the corruption bug:
+            // ; (workflow.changeTracker.constructor as unknown as { isLoadingGraph: boolean }).isLoadingGraph = false
+
+            // Simulate the user clicking during graph loading
+            workflow.changeTracker.checkState()
+          }
+        })
+      })
```
christian-byrne, line 41: This is pretty creative. @coderabbitai make a tracking Issue to make a helper for registering any of the ComfyExtension hooks/callbacks in a browser test easily (if the time comes when we are doing this in multiple places).

browser_tests/tests/changeTrackerLoadGuard.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+      // which triggers a checkState call on the wrong graph, corrupting the activeState.
+      await comfyPage.page.evaluate(() => {
+        window.app!.registerExtension({
+          name: 'TestCheckStateDuringLoad',
+          afterConfigureGraph() {
+            const workflow = (window.app!.extensionManager as WorkspaceStore)
+              .workflow.activeWorkflow
+            if (!workflow) throw new Error('No workflow found')
+            // Bypass the guard to reproduce the corruption bug:
+            // ; (workflow.changeTracker.constructor as unknown as { isLoadingGraph: boolean }).isLoadingGraph = false
```
christian-byrne, line 35: question (non-blocking): What does this mean exactly?

christian-byrne (approved):
LGTM!

browser_tests/tests/changeTrackerLoadGuard.spec.ts
pythongosssss: In order to recreate the issue, uncommenting the line of code removes the guard, triggering the workflow overwrite on tab switch

PR #10624, opened 2026-03-27
test: add mask editor E2E tests for dialog lifecycle

kaili-yang:
## Summary

Adds Playwright smoke tests for mask editor dialog open-via-command, cancel, and save flows — the first working E2E coverage for the mask editor.

## Changes

- **What**: Three `@smoke` tests in `imagePreview.spec.ts` covering levels 1.3–1.5 of the mask editor E2E plan: open via command execution, cancel without uploading, save with stroke and upload verification. Two shared helpers (`openMaskEditorViaCommand`, `drawStrokeOnMaskEditor`) extracted for reuse by future levels. Five screenshot checkpoints across the three tests; golden images will be produced by CI on first run.

## Review Focus

- Tests open the dialog via `Comfy.MaskEditor.OpenMaskEditor` command rather than the image preview hover button, avoiding the CI sync issue tracked in #8143.
- Upload assertions in the save test intercept and fulfill actual network requests rather than mocking, so they exercise the real save path.
- Screenshot expectations are declared but not generated locally — CI will produce golden images on first run.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10624-test-add-mask-editor-E2E-tests-for-dialog-lifecycle-levels-1-3-1-5-3306d73d365081629616ff640c85c5e3) by [Unito](https://www.unito.io)


browser_tests/tests/vueNodes/interactions/node/imagePreview.spec.ts
```diff
@@ -36,6 +37,27 @@ test.describe('Vue Nodes Image Preview', () => {
     }
   }
 
+  async function openMaskEditorViaCommand(comfyPage: ComfyPage) {
```
jtydhr88: we should move this test into maskeditor.spec.ts

jtydhr88:
as my mask editor basic tests merged, I think we should move these into maskeditor.spec.ts, see https://github.com/Comfy-Org/ComfyUI_frontend/pull/10574

jtydhr88 (changes requested):

browser_tests/tests/vueNodes/interactions/node/imagePreview.spec.ts
```diff
@@ -36,6 +37,27 @@ test.describe('Vue Nodes Image Preview', () => {
...
+    const { nodeId } = await loadImageOnNode(comfyPage)
+    await comfyPage.vueNodes.selectNode(nodeId)
+    await comfyPage.command.executeCommand('Comfy.MaskEditor.OpenMaskEditor')
+    const dialog = comfyPage.page.locator('.mask-editor-dialog')
+    await expect(dialog).toBeVisible()
```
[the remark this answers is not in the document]
kaili-yang: Fixed.

browser_tests/tests/maskEditor.spec.ts
```diff
@@ -89,4 +111,76 @@ test.describe('Mask Editor', () => {
...
+      })
+
+      const dialog = await openMaskEditorViaCommand(comfyPage)
+      await drawStrokeOnMaskEditor(comfyPage.page)
+      await expect(dialog).toHaveScreenshot('mask-editor-after-stroke.png')
+
+      await dialog.getByRole('button', { name: /save/i }).click()
+
+      await expect(dialog).not.toBeVisible()
+      expect(uploadedPaths.length).toBeGreaterThan(0)
```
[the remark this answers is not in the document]
kaili-yang: Rejected. No race condition, the dialog close is gated on `await saver.save() `completing.

browser_tests/tests/vueNodes/interactions/node/imagePreview.spec.ts
```diff
@@ -66,6 +88,78 @@ test.describe('Vue Nodes Image Preview', () => {
...
+        const response = await route.fetch()
+        const body = await response.json()
+        if (body?.name) uploadedPaths.push(body.name)
+        return route.fulfill({ response })
+      })
+      await comfyPage.page.route('**/upload/image', async (route) => {
+        const response = await route.fetch()
+        const body = await response.json()
+        if (body?.name) uploadedPaths.push(body.name)
+        return route.fulfill({ response })
```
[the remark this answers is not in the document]
kaili-yang: Rejected. Tests implementation covers the behavior.

browser_tests/tests/maskEditor.spec.ts
```diff
@@ -32,6 +33,27 @@ test.describe('Mask Editor', () => {
...
+    const dialog = comfyPage.page.locator('.mask-editor-dialog')
+    await expect(dialog).toBeVisible()
+    return dialog
+  }
+
+  async function drawStrokeOnMaskEditor(page: Page) {
+    const pointerZone = page.locator('.maskEditor-ui-container').first()
+    const box = await pointerZone.boundingBox()
+    if (!box) throw new Error('PointerZone not found')
+    await page.mouse.move(box.x + box.width * 0.3, box.y + box.height * 0.5)
```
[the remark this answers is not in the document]
kaili-yang: Fixed.

jtydhr88 (dismissed):
LGTM

browser_tests/tests/maskEditor.spec.ts
```diff
@@ -32,6 +33,15 @@ test.describe('Mask Editor', () => {
...
   }
 
+  async function openMaskEditorViaCommand(comfyPage: ComfyPage) {
+    const { nodeId } = await loadImageOnNode(comfyPage)
+    await comfyPage.vueNodes.selectNode(nodeId)
+    await comfyPage.command.executeCommand('Comfy.MaskEditor.OpenMaskEditor')
+    const dialog = comfyPage.page.getByTestId(TestIds.maskEditor.dialog)
+    await expect(dialog).toBeVisible()
+    return dialog
+  }
```
christian-byrne: Should we move this to a shared helper?

browser_tests/tests/maskEditor.spec.ts
```diff
@@ -32,6 +33,15 @@ test.describe('Mask Editor', () => {
     }
   }
 
+  async function openMaskEditorViaCommand(comfyPage: ComfyPage) {
+    const { nodeId } = await loadImageOnNode(comfyPage)
+    await comfyPage.vueNodes.selectNode(nodeId)
+    await comfyPage.command.executeCommand('Comfy.MaskEditor.OpenMaskEditor')
+    const dialog = comfyPage.page.getByTestId(TestIds.maskEditor.dialog)
+    await expect(dialog).toBeVisible()
```
christian-byrne: I wouldn't usually put `expect` into an imperative helper like this. If it's a shared helper then all test cases that use it will fail on `expect` if there is an upstream problem -- but we would prefer only the test case covering the actual problem to have a failing `expect` on that check.

browser_tests/tests/maskEditor.spec.ts
```diff
@@ -89,4 +101,46 @@ test.describe('Mask Editor', () => {
...
+      const dialog = await openMaskEditorViaCommand(comfyPage)
+
+      await expect(
+        dialog.getByTestId(TestIds.maskEditor.uiContainer)
+      ).toBeVisible()
+      await expect(
+        dialog.getByRole('heading', { name: 'Mask Editor' })
+      ).toBeVisible()
+      await expect(dialog).toHaveScreenshot('mask-editor-open-via-command.png')
+    }
```
christian-byrne, line 88: Since we already `expect` the dialog here, we can likely remove the `expect` that's currently inside of `openMaskEditorViaCommand`

browser_tests/tests/maskEditor.spec.ts
```diff
@@ -89,4 +101,46 @@ test.describe('Mask Editor', () => {
...
+  )
+
+  test(
```
christian-byrne, line 91: I think this test can drift if we change the `/upload` API. Consider the scenario:

1. We change the `/upload` API
2. The mask editor has a regression where it starts uploading on cancellation
3. Because we are using page.route with hardcoded route string args, it technically still passes the `expect(uploadRequests).toHaveLength(0)`
4. Although we are taking a screenshot, we didn't actually change the mask at all, so the upload happens but the final screenshot also still passes because the new image is the same as the old image.

Potentailly we want to:

- get the route from the ingest-types package such that we have some type checking to prevent future drifting
- actually change the mask before cancelling, so that the screenshot verification is more meaningful

browser_tests/tests/maskEditor.spec.ts
kaili-yang: A good point. It's already moved.

browser_tests/tests/maskEditor.spec.ts
kaili-yang: Make sense. I'll leave it draft until the route is properly typed and refactored.

PR #10471, opened 2026-03-24
test: migrate 13 component tests from VTU to VTL (Phase 1)

DrJKL:
## Summary

Migrate 13 component test files from @vue/test-utils to @testing-library/vue as Phase 1 of incremental VTL adoption.

## Changes

- **What**: Rewrite 13 test files (88 tests) to use `render`/`screen` queries, `userEvent` interactions, and `jest-dom` assertions. Add `data-testid` attributes to 6 components for lint-clean icon/element queries. Delete unused `src/utils/test-utils.ts`.
- **Dependencies**: `@testing-library/vue`, `@testing-library/user-event`, `@testing-library/jest-dom` (installed in Phase 0)

## Review Focus

- `data-testid` additions to component templates are minimal and non-behavioral
- PrimeVue passthrough (`pt`) usage in UserAvatar.vue for icon testid
- 2 targeted `eslint-disable` in FormRadioGroup.test.ts where PrimeVue places `aria-describedby` on wrapper div, not input

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10471-test-migrate-13-component-tests-from-VTU-to-VTL-Phase-1-32d6d73d36508159a33ffa285afb4c38) by [Unito](https://www.unito.io)

PR #10628, opened 2026-03-27
test: add standardized test-utils with Pinia + i18n defaults

christian-byrne:
Recreates `src/utils/test-utils.ts` (removed in #10471) as the canonical shared render wrapper for VTL component tests.

**Provides:**
- `renderWithDefaults()`: auto-configures Pinia (`stubActions: false`), vue-i18n (English), and common directive stubs (tooltip)
- Optional `userEvent` instance (opt out via `setupUser: false`)
- Re-exports `screen` from `@testing-library/vue`

**Eliminates** duplicated `createTestingPinia` + `createI18n` boilerplate across 97+ test files.

Includes 3 self-tests verifying default plugin injection, userEvent provisioning, and opt-out.

Part of: Test Coverage Q2 Overhaul (COMP-02)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10628-test-add-standardized-test-utils-with-Pinia-i18n-defaults-3306d73d365081fea2d3d462eccfa2c2) by [Unito](https://www.unito.io)


src/utils/test-utils.test.ts
christian-byrne, line 1: Main goal:

> Eliminates duplicated createTestingPinia + createI18n boilerplate across 97+ test files.

DrJKL, line 1: We need to go deeper.gif

src/utils/test-utils.ts
```diff
@@ -0,0 +1,170 @@
...
+})
+
+const stubs = {
+  Button: ButtonStub,
+  Skeleton: SkeletonStub,
+  Tag: TagStub,
+  Badge: BadgeStub,
+  Message: MessageStub,
+  Dialog: DialogStub
+} as const
```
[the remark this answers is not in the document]
DrJKL: I'd agree with this. Feels more complicated than it needs to be to start with.

src/utils/test-utils.ts
```diff
@@ -0,0 +1,180 @@
...
+      global: {
+        ...globalOptions,
+        plugins: [...createDefaultPlugins(), ...(globalOptions?.plugins ?? [])],
+        directives: {
+          ...defaultDirectiveStubs,
+          ...globalOptions?.directives
+        }
+      },
+      ...rest
+    } as Parameters<typeof render>[1]
```
DrJKL: I'd rather have the type checking be more explicit here
```suggestion
    } satisfies RenderOptions<C>
```

src/utils/test-utils.ts
```diff
@@ -0,0 +1,180 @@
...
+ */
+function renderWithDefaults<C>(
+  component: C,
+  options?: ComponentMountingOptions<C> & { setupUser?: boolean }
+): RenderWithDefaultsResult {
+  const { setupUser = true, global: globalOptions, ...rest } = options ?? {}
+  const user = setupUser ? userEvent.setup() : undefined
+
+  const result = render(
+    component as Parameters<typeof render>[0],
```
DrJKL: Do you need the ` as `?

src/utils/test-utils.ts
```diff
@@ -0,0 +1,180 @@
...
+ * ```
+ */
+function renderWithDefaults<C>(
+  component: C,
+  options?: ComponentMountingOptions<C> & { setupUser?: boolean }
+): RenderWithDefaultsResult {
+  const { setupUser = true, global: globalOptions, ...rest } = options ?? {}
+  const user = setupUser ? userEvent.setup() : undefined
+
+  const result = render(
```
DrJKL:
```suggestion
  const result = render<C>(
```

christian-byrne: fixed in e2301b6 - using render<C> now

src/utils/test-utils.ts
christian-byrne: removed the component assertion in e2301b6 - the render<C> generic handles it now

src/utils/test-utils.ts
christian-byrne: tried this but `satisfies RenderOptions<C>` doesn'\''t compile - ComponentMountingOptions<C> props include `null` in a conditional type that RenderOptions<C> doesn'\''t accept, so the types aren'\''t structurally compatible. switched to `as RenderOptions<C>` instead (e2301b6), which is still a big improvement over the opaque `as Parameters<typeof render>[1]`

src/utils/test-utils.ts
```diff
@@ -0,0 +1,135 @@
...
+
+const stubs = {
+  Skeleton: SkeletonStub,
+  Tag: TagStub,
+  Badge: BadgeStub,
+  Message: MessageStub
+} as const
+
+type RenderWithDefaultsResult = RenderResult & {
+  user: ReturnType<typeof userEvent.setup> | undefined
```
DrJKL, line 92: **`user` is always `T | undefined` even in the default case**

When `setupUser` defaults to `true`, every caller has to write `user!.click(...)` (as the JSDoc example on line 109 shows). Overload signatures can eliminate this:

```suggestion
type RenderWithDefaultsResult<SetupUser extends boolean = true> = RenderResult & {
  user: SetupUser extends false ? undefined : ReturnType<typeof userEvent.setup>
}
```

Then pair with overload signatures on `renderWithDefaults` so the common path gets a non-nullable `user`.

src/utils/test-utils.ts
```diff
@@ -0,0 +1,135 @@
...
+ * - Optional userEvent instance
+ *
+ * @example
+ * ```ts
+ * import { render, screen } from '@/utils/test-utils'
+ *
+ * it('renders button text', async () => {
+ *   const { user } = render(MyComponent, { props: { label: 'Click' } })
+ *   expect(screen.getByRole('button')).toHaveTextContent('Click')
+ *   await user!.click(screen.getByRole('button'))
```
DrJKL, line 109: nit: The `user!` non-null assertion in the example is a symptom of the typing issue above. Once the overloads land, this can become just `user.click(...)`.

src/utils/test-utils.ts
```diff
@@ -0,0 +1,135 @@
...
+    global: {
+      ...globalOptions,
+      plugins: [...createDefaultPlugins(), ...(globalOptions?.plugins ?? [])],
+      directives: {
+        ...defaultDirectiveStubs,
+        ...globalOptions?.directives
+      }
+    },
+    ...rest
+  } as RenderOptions<C>)
```
DrJKL, line 130: **Type assertion — consider `@total-typescript/shoehorn` instead**

The `as RenderOptions<C>` cast exists because the spread reshapes the object. Since shoehorn is available as an option for test utilities, you could use `fromPartial()` to construct the options object without a bare `as` assertion, keeping the project's "assertions are a last resort" rule.

Alternatively, building the `global` object separately with an explicit type annotation may satisfy the compiler without any assertion.

src/utils/test-utils.ts
```diff
@@ -0,0 +1,135 @@
...
+      h('div', { 'data-testid': 'message', role: 'alert' }, slots.default?.())
+  }
+})
+
+const stubs = {
+  Skeleton: SkeletonStub,
+  Tag: TagStub,
+  Badge: BadgeStub,
+  Message: MessageStub
+} as const
```
DrJKL, line 89: **YAGNI — `Tag`, `Badge`, `Message` stubs are unused**

Only `Skeleton` is exercised by a test and none of these three are consumed anywhere in the codebase. Consider deferring them to the migration PR that actually needs them, keeping this PR minimal.

If they stay, they should each have at least a trivial self-test in `test-utils.test.ts` to prove they render correctly (like the existing `Skeleton` test).

src/utils/test-utils.ts
```diff
@@ -0,0 +1,135 @@
...
+ * Creates the default set of Vue plugins for component tests.
+ *
+ * - Pinia with `stubActions: false` (actions execute, but are spied)
+ * - vue-i18n with English locale
+ *
+ * Pass additional plugins via the `plugins` option in `renderWithDefaults`.
+ */
+function createDefaultPlugins() {
+  return [
+    createTestingPinia({ stubActions: false }),
```
DrJKL, line 23: **`createSpy` not forwarded**

Several existing tests pass `createSpy: vi.fn` to `createTestingPinia` (e.g. `BatchCountEdit.test.ts`, `ComfyActionbar.test.ts`). Without forwarding it, those tests can't adopt `renderWithDefaults` without duplicating the Pinia setup.

Consider accepting an optional `piniaOptions` parameter, or at minimum adding `createSpy: vi.fn` here so the default covers the common case.

src/utils/test-utils.ts
```diff
@@ -0,0 +1,135 @@
+/* eslint-disable vue/one-component-per-file, vue/require-prop-types */
```
DrJKL, line 1: **File location**: The existing shared test utility (`litegraphTestUtils.ts`) lives under `src/utils/__tests__/`. Placing a test-only module directly in `src/utils/` makes it importable from production code with no guard. Consider moving to `src/utils/__tests__/test-utils.ts` to match the convention, or adding an ESLint `no-restricted-imports` rule to prevent production code from importing it.

DrJKL (commented):
Overall a solid foundation for standardizing component test setup. A few items below — the type assertion and the `user` nullability are the most impactful.

PR #10629, opened 2026-03-27
test: extract shared subgraph E2E test utilities

DrJKL:
## Summary

Extract repeated patterns from 12 subgraph Playwright spec files into shared test utilities, reducing duplication by ~142 lines.

## Changes

- **What**: New shared helpers for common subgraph test operations:
  - `SubgraphHelper`: `getSlotCount()`, `getSlotLabel()`, `removeSlot()`, `findSubgraphNodeId()`
  - `NodeReference`: `delete()`
  - `subgraphTestUtils`: `serializeAndReload()`, `convertDefaultKSamplerToSubgraph()`, `expectWidgetBelowHeader()`, `collectConsoleWarnings()`, `packAllInteriorNodes()`
  - Replaced ~72 inline `page.evaluate` blocks and multi-line sequences with single helper calls across 12 spec files

## Review Focus

- Behavioral equivalence: every replacement is a mechanical extraction with no test logic changes
- API surface of new helpers: naming, parameter types, placement in existing utility classes
- Whether any remaining inline patterns in the spec files would benefit from further extraction

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10629-test-extract-shared-subgraph-E2E-test-utilities-3306d73d365081b0b6b5db52ed0a4552) by [Unito](https://www.unito.io)


browser_tests/fixtures/helpers/SubgraphHelper.ts
```diff
@@ -413,4 +414,138 @@ export class SubgraphHelper {
...
+    return this.page.evaluate(
+      ([slotType, idx]) => {
+        const graph = window.app!.canvas.graph
+        if (!graph || !('inputNode' in graph)) return null
+        const slot = graph[`${slotType}s`]?.[idx]
+        return slot?.label ?? slot?.name ?? null
+      },
+      [type, index] as const
+    )
+  }
```
christian-byrne, line 439: So the need for litegraph-based helpers like this is just unavoidable here I am assuming?

browser_tests/fixtures/helpers/SubgraphHelper.ts
```diff
@@ -413,4 +414,138 @@ export class SubgraphHelper {
...
+    if (type === 'input') {
+      await this.rightClickInputSlot(slotName)
+    } else {
+      await this.rightClickOutputSlot(slotName)
+    }
+    await this.comfyPage.contextMenu.clickLitegraphMenuItem('Remove Slot')
+    await this.comfyPage.nextFrame()
+  }
+
+  async findSubgraphNodeId(): Promise<string> {
```
christian-byrne, line 451: nit: it's more like `findFirstNodeInCurrentGraphThatIsASubgraphNode` (and throw error if none found unlike standard `find` functions) (and it won't be deterministic).

browser_tests/fixtures/helpers/SubgraphHelper.ts
```diff
@@ -413,4 +414,138 @@ export class SubgraphHelper {
...
+  async serializeAndReload(): Promise<void> {
+    const serialized = await this.page.evaluate(() =>
+      window.app!.graph!.serialize()
+    )
+    await this.page.evaluate(
+      (workflow: ComfyWorkflowJSON) => window.app!.loadGraphData(workflow),
+      serialized as ComfyWorkflowJSON
+    )
+    await this.comfyPage.nextFrame()
+  }
```
christian-byrne, line 472: nit: this fucntion seems like it wants to be somewhere higher up

browser_tests/fixtures/helpers/SubgraphHelper.ts
```diff
@@ -413,4 +414,138 @@ export class SubgraphHelper {
...
+
+  async convertDefaultKSamplerToSubgraph(): Promise<NodeReference> {
+    await this.comfyPage.workflow.loadWorkflow('default')
```
christian-byrne, line 475: nit: I wouldn't expect that this function loads a workflow based on the name

browser_tests/fixtures/helpers/SubgraphHelper.ts
```diff
@@ -413,4 +414,138 @@ export class SubgraphHelper {
...
+    const ksampler = await this.comfyPage.nodeOps.getNodeRefById('3')
```
christian-byrne, line 476: WDYT about maintaining some mapping in code between ids->nodename for the common workflows? Although that's basically what the jsons already are... just thinking that as this pattern becomes more common we lose a bit of expressiveness that might be aided by a constant object / enum (but not actually a ts enum but you get what I mean).

browser_tests/fixtures/helpers/SubgraphHelper.ts
```diff
@@ -413,4 +414,138 @@ export class SubgraphHelper {
...
+    await ksampler.click('title')
+    const subgraphNode = await ksampler.convertToSubgraph()
+    await this.comfyPage.nextFrame()
```
christian-byrne, line 479: Is the `ksampler.convertToSubgraph` not calling `nextFrame` itself?

browser_tests/fixtures/helpers/SubgraphHelper.ts
```diff
@@ -413,4 +414,138 @@ export class SubgraphHelper {
...
+    await this.comfyPage.vueNodes.enterSubgraph(hostNodeId)
+    await this.comfyPage.settings.setSetting('Comfy.VueNodes.Enabled', false)
+    await this.comfyPage.nextFrame()
+    await this.comfyPage.canvas.click()
+    await this.comfyPage.canvas.press('Control+a')
+    await this.comfyPage.nextFrame()
+    await this.page.evaluate(() => {
+      const canvas = window.app!.canvas
+      canvas.graph!.convertToSubgraph(canvas.selectedItems)
+    })
```
christian-byrne, line 493: Are you sure we cannot just:

1. Control+a
2. Click a node
3. Find selection toolbox
4. Convert to subgraph icon -> click

(while staying in vue mode)

What's currently making this difficult/flaky? Finding a node to click in the viewport?

browser_tests/fixtures/helpers/SubgraphHelper.ts
```diff
@@ -413,4 +414,138 @@ export class SubgraphHelper {
...
+    await this.comfyPage.nextFrame()
+    await this.exitViaBreadcrumb()
+    await this.comfyPage.canvas.click()
+    await this.comfyPage.nextFrame()
+  }
+
+  static getTextSlotPosition(page: Page, nodeId: string) {
```
christian-byrne, line 500: nit: I know this is a pre-existing method in the subgraph helpers, but this feels like it wants to be in another helper or the node reference class.

browser_tests/fixtures/helpers/SubgraphHelper.ts
```diff
@@ -413,4 +414,138 @@ export class SubgraphHelper {
...
+    const headerBox = await nodeLocator
+      .locator('[data-testid^="node-header-"]')
+      .boundingBox()
+    const widgetBox = await widgetLocator.boundingBox()
+    if (!headerBox || !widgetBox)
+      throw new Error('Header or widget bounding box not found')
+    expect(widgetBox.y).toBeGreaterThan(headerBox.y + headerBox.height)
+  }
+
+  static collectConsoleWarnings(
```
christian-byrne, line 533: Honest question: is this actually considered good practice? Genuinely don't know the arguments for/against using this pattern in e2e tests.

christian-byrne (commented):
A lot of nits. Obviously these comments are moreso about the existing functionality since this is just a refactoring, so they are not blocking.

Overall +244/-414 is nice to see and the PR LGTM!

christian-byrne (approved):
Forgot to approve

browser_tests/fixtures/helpers/SubgraphHelper.ts
DrJKL: I have a child branch where I'm working on updating the Helper logic to do things like this.

browser_tests/fixtures/helpers/SubgraphHelper.ts
DrJKL: I think that adding structure to the fixtures is going to be great for readability, especially if we can also get intellisense and type checking for names/ids.

browser_tests/fixtures/helpers/SubgraphHelper.ts
DrJKL: For now, yes.

browser_tests/fixtures/helpers/SubgraphHelper.ts
DrJKL: Taking a step back, having information that's only in the console isn't great for users. If it's a real problem that they can fix, it should be presented to them. If it's not a real problem, why log it in the first place?
For cases like this, it can be very useful within test and dev contexts to have extra information about known past issues that you can confirm are behaving in the way you expect even when it is opaque to a user.

PR #10630, opened 2026-03-27
test: split assets sidebar browser fixtures

benceruleanlu:
## Summary

Split the old assets helper into a standalone asset scenario fixture plus an in-memory jobs backend, and migrate the real assets sidebar and queue overlay E2Es onto it.

## Changes

- **What**: split the old assets helper into `assetScenarioFixture` + `InMemoryJobsBackend`; migrate `browser_tests/tests/sidebar/assets.spec.ts` and `browser_tests/tests/queue/queueOverlay.spec.ts`; remove the large comments from `assets.spec.ts`
- **Dependencies**: none

## Review Focus

Fixture/backend boundaries and the real E2E migrations.

## Validation

- `pnpm exec eslint` on touched browser files
- `pnpm typecheck:browser`
- `pnpm typecheck`
- Migrated Playwright specs were not rerun because `localhost:5173` was unavailable when attempted

Fixes COM-306


browser_tests/tests/sidebar/assets.spec.ts
```diff
@@ -14,9 +14,9 @@ import type { RawJobListItem } from '../../../src/platform/remote/comfyui/jobs/j
 const SAMPLE_JOBS: RawJobListItem[] = [
   createMockJob({
     id: 'job-alpha',
-    create_time: 1000,
-    execution_start_time: 1000,
-    execution_end_time: 1010,
+    create_time: 1_000_000,
```
benceruleanlu, line 16: These fixture timestamps are intentionally synthetic epoch milliseconds now, not seconds. Keeping the sample data in ms makes the assets-sidebar duration and sort behavior line up with the runtime queue model and the other queue/browser tests.

src/platform/workflow/cloud/getWorkflowFromHistory.test.ts
```diff
@@ -23,8 +23,8 @@ const mockWorkflow: ComfyWorkflowJSON = {
 const mockJobDetailResponse: JobDetail = {
   id: 'test-job-id',
   status: 'completed',
-  create_time: 1234567890,
-  update_time: 1234567900,
+  create_time: 1234567890000,
```
benceruleanlu, line 26: This JobDetail fixture now uses milliseconds as well for consistency with the frontend jobs/queue tests. The extraction logic itself is unit-agnostic, but keeping all Jobs API test fixtures on ms avoids reintroducing mixed timestamp units.

browser_tests/fixtures/helpers/AssetsHelper.ts
```diff
@@ -16,7 +33,7 @@ export function createMockJob(
     status: 'completed',
     create_time: now,
     execution_start_time: now,
-    execution_end_time: now + 5000,
+    execution_end_time: now + 5_000,
```
benceruleanlu: Standardized this helper to epoch milliseconds to match the frontend queue/assets contract. The UI derives seconds from `(execution_end_time - execution_start_time) / 1000`, so second-based mocks here would under-report durations by 1000x.

browser_tests/fixtures/helpers/InMemoryJobsBackend.ts
benceruleanlu, line 1: Changes to ComfyUI to make things user scoped then will allow us to move this to our devtools and not need to have this FE BE drift risk.

browser_tests/tests/queue/queueOverlay.spec.ts
```diff
@@ -35,16 +38,11 @@ const MOCK_JOBS: RawJobListItem[] = [
 ]
 
 test.describe('Queue overlay', () => {
-  test.beforeEach(async ({ comfyPage }) => {
-    await comfyPage.assets.mockOutputHistory(MOCK_JOBS)
-    await comfyPage.settings.setSetting('Comfy.Queue.QPOV2', false)
+  test.beforeEach(async ({ comfyPage, assetScenario }) => {
+    await assetScenario.seedGeneratedHistory(MOCK_JOBS)
     await comfyPage.setup()
```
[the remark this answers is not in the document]
benceruleanlu: Fixed in `77250c601`.

I restored the explicit `Comfy.Queue.QPOV2 = false` reset in `queueOverlay.spec.ts`, so the spec no longer depends on persisted docked-job-history state from earlier tests.

While debugging the CI failures, I also found a second regression in the new asset scenario fixture layer: it was dropping `JobEntry.status` and re-seeding every generated-history job as `completed`. That was the actual reason the two failing assertions broke: the Failed tab never appeared, and `job-failed-1` still showed up under the Completed filter.

The fix preserves `job.status` through `GeneratedJobFixture` into the in-memory jobs backend, and the targeted `queueOverlay.spec.ts` rerun passed against an isolated preview environment after both changes.

browser_tests/fixtures/components/SidebarTab.ts
```diff
@@ -381,37 +409,124 @@ export class AssetsSidebarTab extends SidebarTab {
...
+  async selectAssets(names: string[]) {
+    if (names.length === 0) {
+      return
+    }
+
+    await this.asset(names[0]).click()
+
+    const modifier = process.platform === 'darwin' ? 'Meta' : 'Control'
+    for (const name of names.slice(1)) {
+      await this.asset(name).click({
```
christian-byrne, line 413: **[M2] Bug: `selectAssets` uses `click({ modifiers })` — won't trigger VueUse's `useKeyModifier`**

The removed comment in `assets.spec.ts` (line ~462 of old code) explicitly documented that `click({ modifiers })` only sets the mouse event flag and does NOT fire a `keydown` event, so VueUse's `useKeyModifier('Control')` won't detect the modifier. The working multi-select test in this PR (line ~468) correctly uses `keyboard.down/up` instead.

Any future test calling `selectAssets` for multi-select will silently fail — only the last asset will end up selected.

**Suggested fix:**
```ts
async selectAssets(names: string[]) {
  if (names.length === 0) return
  await this.asset(names[0]).click()
  if (names.length > 1) {
    const modifier = process.platform === 'darwin' ? 'Meta' : 'Control'
    await this.page.keyboard.down(modifier)
    for (const name of names.slice(1)) {
      await this.asset(name).click()
    }
    await this.page.keyboard.up(modifier)
  }
}
```

browser_tests/fixtures/components/SidebarTab.ts
```diff
@@ -336,38 +368,34 @@ export class AssetsSidebarTab extends SidebarTab {
...
   }
 
   // --- Selection footer ---
 
   get selectionFooter() {
-    return this.page
-      .locator('.sidebar-content-container')
-      .locator('..')
-      .locator('[class*="h-18"]')
+    return this.root.locator('..').locator('[class*="h-18"]')
```
christian-byrne: **[m3] Fragile CSS selector: `[class*="h-18"]`**

Partial-match against a Tailwind utility class — breaks silently if the height class changes. Consider adding `data-testid="assets-selection-footer"` to the `<div>` in `AssetsSidebarTab.vue` (line 122) and using `this.page.getByTestId('assets-selection-footer')` here.

browser_tests/fixtures/helpers/InMemoryJobsBackend.ts
```diff
@@ -0,0 +1,207 @@
+import type { Page, Route } from '@playwright/test'
```
christian-byrne, line 1: **[M3] Missing unit tests for InMemoryJobsBackend (207 lines)**

This class has non-trivial filtering (status, workflow_id), sorting (create_time vs execution_duration), pagination (offset/limit/has_more), detail lookup by job ID, and history deletion (both by ID array and `clear` flag). None of this logic is unit-tested — the vitest file only covers `AssetScenarioHelper`'s view-route logic.

Bugs in pagination math or filter logic would silently corrupt every E2E test that depends on this backend. Consider adding a vitest spec (similar pattern to `AssetScenarioHelper.test.ts` with page/route stubs) covering the key scenarios.

browser_tests/fixtures/helpers/InMemoryJobsBackend.ts
```diff
@@ -0,0 +1,207 @@
...
+          contentType: 'application/json',
+          body: JSON.stringify(job.detail)
+        })
+      }
+
+      await this.page.route(jobDetailRoutePattern, this.detailRouteHandler)
+    }
+
+    if (!this.historyRouteHandler) {
+      this.historyRouteHandler = async (route: Route) => {
```
christian-byrne, line 176: **[m2] Missing POST method guard on history handler**

The old `AssetsHelper` explicitly checked `request.method() !== 'POST'` and called `route.continue()` for non-POST requests. This handler calls `postDataJSON()` on all requests and falls through to `route.fulfill({ status: 200, body: '{}' })` regardless of method.

If any code path triggers a GET to `/api/history`, the mock will swallow it instead of forwarding to the real server.

```ts
// Add at top of handler:
if (route.request().method() !== 'POST') {
  await route.continue()
  return
}
```

browser_tests/fixtures/helpers/AssetScenarioHelper.ts
```diff
@@ -0,0 +1,383 @@
...
+  private viewRouteHandler: ((route: Route) => Promise<void>) | null = null
+  private generatedJobs: GeneratedJobFixture[] = []
+  private importedFiles: ImportedAssetFixture[] = []
+  private seededFiles = new Map<string, SeededAssetFile>()
+
+  constructor(private readonly page: Page) {
+    this.jobsBackend = new InMemoryJobsBackend(page)
+  }
+
+  async seedGeneratedHistory(jobs: readonly JobEntry[]): Promise<void> {
```
christian-byrne, line 153: **[m1] Hidden coupling between `seedGeneratedHistory` and `seedImportedFiles`**

Both methods internally call `this.seed()` which replaces *all* state (both generated and imported). Each method preserves the other dimension by passing `this.importedFiles` / `this.generatedJobs`, but this creates an implicit ordering dependency.

A future test author calling only `seedImportedFiles` would get empty generated history (since `this.generatedJobs` starts as `[]`). The API looks independent but isn't.

Consider adding a JSDoc comment documenting this coupling, e.g.:
```ts
/**
 * Replaces generated jobs while preserving the current imported files.
 * Both seed methods internally call seed(), which replaces all state —
 * call seedImportedFiles first if both need updating.
 */
```

browser_tests/tests/sidebar/assets.spec.ts
```diff
@@ -699,83 +577,3 @@ test.describe('Assets sidebar - settings menu', () => {
     await expect(tab.gridViewOption).toBeVisible()
   })
 })
```
christian-byrne, line 651: **[M1] Delete-confirmation tests removed — coverage regression**

The old file had 3 delete-confirmation tests (right-click delete → dialog, confirm → asset removed + toast, cancel → asset preserved). `InMemoryJobsBackend` already supports `POST /api/history` with `{ delete: [...] }` payloads, so the infrastructure is present — but no tests exercise it.

Delete is a destructive user action and a refactoring PR should be coverage-neutral. Consider porting these tests using the new `assetScenario` fixture.

browser_tests/fixtures/helpers/buildMockJobOutputs.ts
```diff
@@ -0,0 +1,28 @@
+import type { JobDetailResponse } from '@comfyorg/ingest-types'
+
+import type { TaskOutput } from '../../../src/schemas/apiSchema'
+
+import type {
+  GeneratedJobFixture,
+  GeneratedOutputFixture
+} from './assetScenarioTypes'
+
+export function buildMockJobOutputs(
```
christian-byrne, line 10: **[m5] No unit tests for `buildMockJobOutputs`**

This function has defaulting logic (`nodeId ?? '5'`, `mediaType ?? 'images'`, `type ?? 'output'`) and constructs a nested `TaskOutput` structure. Any shape mismatch vs the real API would silently break downstream assertions. Consider adding a small vitest spec for single/multi-output and default cases.

scripts/browser_tests/AssetScenarioHelper.test.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    await Promise.all([
+      writeFile(outputFile, 'root output'),
+      writeFile(nestedOutputFile, 'nested output'),
+      writeFile(inputFile, 'input asset')
+    ])
+
+    const { page, routes } = createPageStub()
+    const helper = new AssetScenarioHelper(page as unknown as Page)
+    const testAccess = helper as unknown as AssetScenarioHelperTestAccess
```
christian-byrne: **[N1] Test accesses private `seed()` via type cast — change-detector pattern**

Casting `helper as unknown as AssetScenarioHelperTestAccess` to call the private `seed()` method couples this test to internal API. If `seed()` is renamed or its signature changes, the test breaks despite no behavior change.

Consider testing through the public methods (`seedGeneratedHistory` + `seedImportedFiles`) which are the actual contract.

christian-byrne (commented):
Review from 4 agents (test-quality, architecture, bug-hunter, pattern-compliance). 5 Major, 5 Minor, 3 Nitpick findings.

**Positive notes:** First use of `mergeTests()` is a great step toward breaking up the `ComfyPage` god object. `JobEntry` from `@comfyorg/ingest-types` is the correct canonical type. `AssetScenarioHelper` + `InMemoryJobsBackend` separation is clean. `aria-label` additions improve both accessibility and test stability.

browser_tests/tests/sidebar/assets.spec.ts
```diff
@@ -572,26 +480,17 @@ test.describe('Assets sidebar - context menu', () => {
     })
     await expect(contextMenu).toBeVisible()
 
-    // Bulk menu should show bulk download action
     await expect(tab.contextMenuItem('Download all')).toBeVisible()
   })
 })
```
DrJKL, line 571: **[medium]** The entire "Delete confirmation" describe block (3 tests: dialog shown, confirm-deletes, cancel-preserves) was removed here. These cover a destructive user action and represent real behavioral coverage.

If they're being relocated to a follow-up PR, please link the tracking issue. Otherwise, consider restoring them — the `InMemoryJobsBackend` already supports history deletion, so porting them to the new fixture should be straightforward.

browser_tests/fixtures/helpers/AssetScenarioHelper.ts
```diff
@@ -0,0 +1,383 @@
...
+    return `${baseFilename}-${index + 1}`
+  }
+
+  return `${baseFilename.slice(0, extensionIndex)}-${index + 1}${baseFilename.slice(extensionIndex)}`
+}
+
+function getPreviewOutput(
+  previewOutput: JobEntry['preview_output'] | undefined
+): MockPreviewOutput | undefined {
+  return previewOutput as MockPreviewOutput | undefined
```
DrJKL, line 65: **[low]** Unsafe type assertion — if `JobEntry['preview_output']` ever gains or removes fields, this cast silently hides the mismatch. A type guard would be safer:

```suggestion
function isFullPreviewOutput(
  preview: JobEntry['preview_output']
): preview is MockPreviewOutput {
  return preview != null && 'nodeId' in preview
}

function getPreviewOutput(
  previewOutput: JobEntry['preview_output'] | undefined
): MockPreviewOutput | undefined {
  return previewOutput && isFullPreviewOutput(previewOutput)
    ? previewOutput
    : undefined
}
```

DrJKL (commented):
Nice architectural split. The fixture-based cleanup, `InMemoryJobsBackend` separation, and `JobEntry` migration are all clear wins. A few inline comments below — the main concern is the deleted delete-confirmation test coverage.

browser_tests/fixtures/helpers/AssetScenarioHelper.ts
```diff
@@ -0,0 +1,383 @@
...
+    workflowId: job.workflow_id
+  }
+}
+
+function buildSeededJob(job: GeneratedJobFixture) {
+  const outputs = job.outputs.map(normalizeOutputFixture)
+  const preview = outputs[0]
+  const createTime =
+    job.createTime ??
+    new Date(job.createdAt ?? '2026-03-27T12:00:00.000Z').getTime()
```
DrJKL: **[nit]** Hardcoded future date will age poorly. Consider:

```suggestion
    job.createTime ??
    new Date(job.createdAt ?? new Date(0).toISOString()).getTime()
```

Or extract a named constant like `const DEFAULT_FIXTURE_EPOCH = 0`.

browser_tests/fixtures/helpers/buildMockJobOutputs.ts
```diff
@@ -0,0 +1,28 @@
...
+  job: GeneratedJobFixture,
+  outputs: GeneratedOutputFixture[]
+): NonNullable<JobDetailResponse['outputs']> {
+  const nodeId = job.nodeId ?? '5'
+
+  const taskOutput = {
+    [nodeId]: {
+      [outputs[0].mediaType ?? 'images']: outputs.map((output) => ({
```
DrJKL: **[low]** All outputs are grouped under `outputs[0].mediaType`, so a job with mixed media types (image + video) would be incorrectly merged under one key. This mirrors the old behavior, but worth documenting the assumption:

```suggestion
      // All outputs for a single node share the same media type key.
      [outputs[0].mediaType ?? 'images']: outputs.map((output) => ({
```

browser_tests/fixtures/helpers/InMemoryJobsBackend.ts
```diff
@@ -0,0 +1,207 @@
...
+        const requestBody = route.request().postDataJSON() as
```
DrJKL: **[low]** `postDataJSON()` will throw if the body isn't valid JSON (e.g. a non-POST request that happens to match the history pattern). Consider guarding:

```suggestion
        const request = route.request()
        if (request.method() !== 'POST') {
          await route.fulfill({ status: 200, contentType: 'application/json', body: '{}' })
          return
        }
        const requestBody = request.postDataJSON() as
```

browser_tests/tests/queue/queueOverlay.spec.ts
```diff
@@ -1,13 +1,16 @@
...
+import type { JobEntry } from '@comfyorg/ingest-types'
 
-import { comfyPageFixture as test } from '@e2e/fixtures/ComfyPage'
-import { createMockJob } from '@e2e/fixtures/helpers/AssetsHelper'
-import { TestIds } from '@e2e/fixtures/selectors'
-import type { RawJobListItem } from '@/platform/remote/comfyui/jobs/jobTypes'
+import { comfyPageFixture } from '@e2e/fixtures/ComfyPage'
+import { assetScenarioFixture } from '@e2e/fixtures/assetScenarioFixture'
+import { createMockJob } from '@e2e/fixtures/helpers/jobFixtures'
+import { TestIds } from '../../fixtures/selectors'
```
DrJKL: **[nit]** Inconsistent import style — lines 4-6 use the `@e2e` alias but this one uses a relative path.

```suggestion
import { TestIds } from '@e2e/fixtures/selectors'
```

scripts/browser_tests/AssetScenarioHelper.test.ts
DrJKL: **[low]** Double-cast `as unknown as AssetScenarioHelperTestAccess` to reach private `seed()` is fragile — a rename breaks silently at runtime. Consider making `seed()` `protected` (testable via subclass) or extracting a minimal testable interface that `AssetScenarioHelper` implements.

browser_tests/tests/queue/queueOverlay.spec.ts
DrJKL: 😮‍💨

benceruleanlu: Wow!

browser_tests/fixtures/components/SidebarTab.ts
benceruleanlu: In my testing, it does fire a keydown event for modifiers. This seems pretty difficult to believe in general because we're not doing anything special here, we're just plainly using playwright's https://playwright.dev/docs/api/class-locator#locator-click-option-modifiers according to the doc.

benceruleanlu: An inspection of the playwright source confirms this too.

browser_tests/fixtures/components/SidebarTab.ts
benceruleanlu: Agree, changed to role based selector over datatestid, however.

browser_tests/fixtures/helpers/InMemoryJobsBackend.ts
benceruleanlu: K, added

browser_tests/fixtures/helpers/InMemoryJobsBackend.ts
benceruleanlu: K, gated

browser_tests/fixtures/helpers/AssetScenarioHelper.ts
benceruleanlu: I don’t think there’s an ordering dependency here. These methods are intentionally partial updates over shared scenario state: seedGeneratedHistory() replaces generated jobs while preserving imported files, and seedImportedFiles() does the inverse. Calling one of them on a fresh helper preserves the default empty state for the other dimension, which is expected.

browser_tests/tests/sidebar/assets.spec.ts
benceruleanlu: Whoops, messed up in git, restored

browser_tests/fixtures/helpers/buildMockJobOutputs.ts
benceruleanlu: K, added a small spec

browser_tests/tests/sidebar/assets.spec.ts
benceruleanlu: Restored

scripts/browser_tests/AssetScenarioHelper.test.ts
benceruleanlu: Extracted to a small helper to test directly

benceruleanlu: Extracted as the fix

browser_tests/fixtures/helpers/AssetScenarioHelper.ts
benceruleanlu: This is kinda scuffed because the root cause for this is that ingest types has it as { [key: string]: unknown }, so... I think it's fine to leave for now and we can tighten implementation with ingest types later

browser_tests/fixtures/helpers/buildMockJobOutputs.ts
benceruleanlu: I just fixed it instead.

browser_tests/fixtures/helpers/AssetScenarioHelper.ts
benceruleanlu: It's pegged to reduce chances of any flakiness, I changed to a named constant instead.

browser_tests/fixtures/helpers/InMemoryJobsBackend.ts
benceruleanlu: Fixed

browser_tests/fixtures/components/SidebarTab.ts
```diff
@@ -303,30 +307,58 @@ export class AssetsSidebarTab extends SidebarTab {
...
 
   get settingsButton() {
-    return this.page.getByRole('button', { name: 'View settings' })
+    return this.root.getByLabel('View settings')
   }
 
-  // --- View mode ---
+  get viewSettingsButton() {
+    return this.settingsButton
+  }
```
DrJKL: Instead of getters, can we just make these locators public readonly?

DrJKL: @coderabbitai Open an issue to standardize this for all of our Page Objects.

browser_tests/fixtures/helpers/AssetScenarioHelper.test.ts
DrJKL, line 1: Were these misplaced?
I don't think we should be putting anything new in scripts, it's already being exempted from a lot of checks.

browser_tests/fixtures/helpers/InMemoryJobsBackend.ts
DrJKL: @coderabbitai Open a tracking issue for this. This should not be allowed to linger any longer than absolutely necessary.

browser_tests/fixtures/helpers/InMemoryJobsBackend.test.ts
DrJKL, line 1: **[nit] Duplicated test infrastructure**

`AssetScenarioHelper.test.ts` and `InMemoryJobsBackend.test.ts` share ~80 lines of identical test plumbing (`createPageStub`, `createRouteInvocation`, `bodyToText`, and associated types). Not a merge blocker, but worth extracting to a shared `scripts/browser_tests/routeTestUtils.ts`.

@coderabbitai Open an issue to extract the shared Playwright route-stubbing utilities from `scripts/browser_tests/AssetScenarioHelper.test.ts` and `scripts/browser_tests/InMemoryJobsBackend.test.ts` into a shared `scripts/browser_tests/routeTestUtils.ts` module.

browser_tests/fixtures/helpers/AssetScenarioHelper.ts
DrJKL, line 1: **[low] Unsafe cast on `JobEntry["preview_output"]`**

`getPreviewOutput` wraps a bare `as MockPreviewOutput` cast because `@comfyorg/ingest-types` defines `preview_output` as `{ [key: string]: unknown }`. Safe in practice since we control both sides, but it should be tightened upstream.

@coderabbitai Open an issue to tighten the `preview_output` type in `@comfyorg/ingest-types` so that `JobEntry["preview_output"]` exposes `filename`, `subfolder`, `type`, `nodeId`, and `mediaType` fields, eliminating the need for the unsafe cast in `AssetScenarioHelper.ts:62-66`.

browser_tests/fixtures/helpers/assetScenarioTypes.ts
DrJKL, line 1: Can these reference actual types from `src/`?

benceruleanlu:
The unit tests on the fixtures, I do dislike, I think it's at the request of a review comment, but there is little cost so I kept it.

browser_tests/fixtures/helpers/AssetScenarioHelper.test.ts
benceruleanlu: I didn't know about the scripts check exemption, in that case, I moved it to colocate under browser_tests

browser_tests/fixtures/helpers/assetScenarioTypes.ts
benceruleanlu: This was intentionally not using types from src, believe it or not, there is duplication between src and the ingest types right now.

DrJKL:
> The unit tests on the fixtures, I do dislike, I think it's at the request of a review comment, but there is little cost so I kept it.

The problem I see is that there is enough complexity to motivate having unit tests of the fixture code in the first place.

benceruleanlu:
> > The unit tests on the fixtures, I do dislike, I think it's at the request of a review comment, but there is little cost so I kept it.
> 
> The problem I see is that there is enough complexity to motivate having unit tests of the fixture code in the first place.

I guess, although both issues would be solved in the longer term if we started testing ComfyUI properly, it would require some core changes and modification to devtools. Quite the hurdle for tests.

benceruleanlu:
Closing in favor of replacement stack: #11279, #11280, #11281, #11282, and #11284.

PR #7118, opened 2025-12-03
Fix: Widget sizing with multiple expanding items

DrJKL:
## Summary

Textarea/Markdown/3D pieces grow, other widget rows size to their contents.

## Screenshots (if applicable)
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-7118-Fix-Widget-sizing-with-multiple-expanding-items-2be6d73d3650816c8153ff3f1e49176d) by [Unito](https://www.unito.io)

PR #10633, opened 2026-03-27
App Mode tests

AustinMroz:
Adds tests for
- Mobile app mode.
- Drag and drop operations in app mode
- Basic widget interaction in app mode.
- The read only state when in builder mode.

This branch has been fairly long lived. Some of the tests implemented may have near duplicates.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10633-App-Mode-tests-3306d73d36508154aa25d8096119a32c) by [Unito](https://www.unito.io)

### DIFF
--- a/browser_tests/fixtures/helpers/AppModeHelper.ts
+++ b/browser_tests/fixtures/helpers/AppModeHelper.ts
@@ -227,6 +227,17 @@ export class AppModeHelper {
     return this.page.getByTestId(TestIds.linear.mobile)
   }
 
+  get mobileWorkflows() {
+    return this.mobileView.getByTestId(TestIds.linear.mobileWorkflows)
+  }
+  get mobileActionmenu() {
+    return this.mobileView.getByTestId(TestIds.linear.mobileActionMenu)
+  }
+  async switchMobileWorkflow(workflowName: string) {
+    await this.mobileWorkflows.click()
+    await this.page.getByRole('menu').getByText(workflowName).click()
+  }
+
   /**
    * Get the actions menu trigger for a widget in the app mode widget list.
    * @param widgetName Text shown in the widget label (e.g. "seed").
--- a/browser_tests/fixtures/helpers/AppModeHelper.ts
+++ b/browser_tests/fixtures/helpers/AppModeHelper.ts
@@ -11,11 +11,12 @@ import { BuilderSelectHelper } from '@e2e/fixtures/helpers/BuilderSelectHelper'
 import { BuilderStepsHelper } from '@e2e/fixtures/helpers/BuilderStepsHelper'
 
 export class AppModeHelper {
-  readonly steps: BuilderStepsHelper
   readonly footer: BuilderFooterHelper
+  readonly mobile: MobileAppHelper
   readonly saveAs: BuilderSaveAsHelper
   readonly select: BuilderSelectHelper
   readonly outputHistory: OutputHistoryComponent
+  readonly steps: BuilderStepsHelper
   readonly widgets: AppModeWidgetHelper
 
   /** The "Connect an output" popover shown when saving without outputs. */
@@ -62,11 +63,12 @@ export class AppModeHelper {
   public readonly vueNodeSwitchDontShowAgainCheckbox: Locator
 
   constructor(private readonly comfyPage: ComfyPage) {
-    this.steps = new BuilderStepsHelper(comfyPage)
+    this.mobile = new MobileAppHelper(comfyPage)
     this.footer = new BuilderFooterHelper(comfyPage)
     this.saveAs = new BuilderSaveAsHelper(comfyPage)
     this.select = new BuilderSelectHelper(comfyPage)
     this.outputHistory = new OutputHistoryComponent(comfyPage.page)
+    this.steps = new BuilderStepsHelper(comfyPage)
     this.widgets = new AppModeWidgetHelper(comfyPage)
 
     this.connectOutputPopover = this.page.getByTestId(
@@ -222,12 +224,6 @@ export class AppModeHelper {
 
   get centerPanel(): Locator {
     return this.page.getByTestId(TestIds.linear.centerPanel)
-  }
-  get mobileView(): Locator {
-    return this.page.getByTestId(TestIds.linear.mobile)
-  }
-  get mobileNavigation(): Locator {
-    return this.page.getByRole('tablist').filter({ hasText: 'Run' })
   }
 
   get mobileWorkflows() {
--- a/browser_tests/fixtures/helpers/AppModeHelper.ts
+++ b/browser_tests/fixtures/helpers/AppModeHelper.ts
@@ -9,6 +9,7 @@ import { BuilderFooterHelper } from '@e2e/fixtures/helpers/BuilderFooterHelper'
 import { BuilderSaveAsHelper } from '@e2e/fixtures/helpers/BuilderSaveAsHelper'
 import { BuilderSelectHelper } from '@e2e/fixtures/helpers/BuilderSelectHelper'
 import { BuilderStepsHelper } from '@e2e/fixtures/helpers/BuilderStepsHelper'
+import { MobileAppHelper } from '@e2e/fixtures/helpers/MobileAppHelper'
 
 export class AppModeHelper {
   readonly footer: BuilderFooterHelper
@@ -224,20 +225,6 @@ export class AppModeHelper {
 
   get centerPanel(): Locator {
     return this.page.getByTestId(TestIds.linear.centerPanel)
-  }
-
-  get mobileWorkflows() {
-    return this.mobileView.getByTestId(TestIds.linear.mobileWorkflows)
-  }
-  get mobileActionmenu() {
-    return this.mobileView.getByTestId(TestIds.linear.mobileActionMenu)
-  }
-  async switchMobileWorkflow(workflowName: string) {
-    await this.mobileWorkflows.click()
-    await this.page.getByRole('menu').getByText(workflowName).click()
-  }
-  async mobileNavigateTab(name: 'run' | 'outputs' | 'assets') {
-    await this.mobileNavigation.getByRole('tab', { name }).click()
   }
 
   /**
--- a/browser_tests/fixtures/helpers/AppModeHelper.ts
+++ b/browser_tests/fixtures/helpers/AppModeHelper.ts
@@ -220,6 +220,13 @@ export class AppModeHelper {
     await this.toggleAppMode()
   }
 
+  get centerPanel(): Locator {
+    return this.page.getByTestId(TestIds.linear.centerPanel)
+  }
+  get mobileView(): Locator {
+    return this.page.getByTestId(TestIds.linear.mobile)
+  }
+
   /**
    * Get the actions menu trigger for a widget in the app mode widget list.
    * @param widgetName Text shown in the widget label (e.g. "seed").
--- a/browser_tests/fixtures/helpers/AppModeHelper.ts
+++ b/browser_tests/fixtures/helpers/AppModeHelper.ts
@@ -62,6 +62,8 @@ export class AppModeHelper {
   public readonly vueNodeSwitchDismissButton: Locator
   /** The "Don't show again" checkbox inside the Vue Node switch popup. */
   public readonly vueNodeSwitchDontShowAgainCheckbox: Locator
+  /** The main content area where outputs are displayed*/
+  public readonly centerPanel: Locator
 
   constructor(private readonly comfyPage: ComfyPage) {
     this.mobile = new MobileAppHelper(comfyPage)
@@ -128,6 +130,7 @@ export class AppModeHelper {
     this.vueNodeSwitchDontShowAgainCheckbox = this.page.getByTestId(
       TestIds.appMode.vueNodeSwitchDontShowAgain
     )
+    this.centerPanel = this.page.getByTestId(TestIds.linear.centerPanel)
   }
 
   private get page(): Page {
@@ -221,10 +224,6 @@ export class AppModeHelper {
     }, inputs)
     await this.comfyPage.nextFrame()
     await this.toggleAppMode()
-  }
-
-  get centerPanel(): Locator {
-    return this.page.getByTestId(TestIds.linear.centerPanel)
   }
 
   /**
--- a/browser_tests/fixtures/helpers/AppModeHelper.ts
+++ b/browser_tests/fixtures/helpers/AppModeHelper.ts
@@ -226,6 +226,9 @@ export class AppModeHelper {
   get mobileView(): Locator {
     return this.page.getByTestId(TestIds.linear.mobile)
   }
+  get mobileNavigation(): Locator {
+    return this.page.getByRole('tablist').filter({ hasText: 'Run' })
+  }
 
   get mobileWorkflows() {
     return this.mobileView.getByTestId(TestIds.linear.mobileWorkflows)
@@ -236,6 +239,9 @@ export class AppModeHelper {
   async switchMobileWorkflow(workflowName: string) {
     await this.mobileWorkflows.click()
     await this.page.getByRole('menu').getByText(workflowName).click()
+  }
+  async mobileNavigateTab(name: 'run' | 'outputs' | 'assets') {
+    await this.mobileNavigation.getByRole('tab', { name }).click()
   }
 
   /**
--- a/browser_tests/fixtures/helpers/MobileAppHelper.ts
+++ b/browser_tests/fixtures/helpers/MobileAppHelper.ts
@@ -35,4 +35,7 @@ export class MobileAppHelper {
   private get page(): Page {
     return this.comfyPage.page
   }
+  async tap(locator: Locator, { count = 1 }: { count: number }) {
+    for (let i = 0; i < count; i++) await locator.tap()
+  }
 }
--- a/browser_tests/fixtures/helpers/MobileAppHelper.ts
+++ b/browser_tests/fixtures/helpers/MobileAppHelper.ts
@@ -1,28 +1,23 @@
 import type { Locator, Page } from '@playwright/test'
 
-import type { ComfyPage } from '../ComfyPage'
-import { TestIds } from '../selectors'
+import type { ComfyPage } from '@e2e/fixtures/ComfyPage'
+import { TestIds } from '@e2e/fixtures/selectors'
 
 export class MobileAppHelper {
-  constructor(private readonly comfyPage: ComfyPage) {}
+  readonly actionmenu: Locator
+  readonly contentPanel: Locator
+  readonly navigation: Locator
+  readonly navigationTabs: Locator
+  readonly view: Locator
+  readonly workflows: Locator
 
-  get view(): Locator {
-    return this.page.getByTestId(TestIds.linear.mobile)
-  }
-  get navigation(): Locator {
-    return this.page.getByRole('tablist').filter({ hasText: 'Run' })
-  }
-  get workflows() {
-    return this.view.getByTestId(TestIds.linear.mobileWorkflows)
-  }
-  get actionmenu() {
-    return this.view.getByTestId(TestIds.linear.mobileActionMenu)
-  }
-  get navigationTabs() {
-    return this.navigation.getByRole('tab')
-  }
-  get contentPanel() {
-    return this.page.getByRole('tabpanel')
+  constructor(private readonly comfyPage: ComfyPage) {
+    this.view = this.page.getByTestId(TestIds.linear.mobile)
+    this.actionmenu = this.view.getByTestId(TestIds.linear.mobileActionMenu)
+    this.contentPanel = this.page.getByRole('tabpanel')
+    this.navigation = this.page.getByRole('tablist').filter({ hasText: 'Run' })
+    this.navigationTabs = this.navigation.getByRole('tab')
+    this.workflows = this.view.getByTestId(TestIds.linear.mobileWorkflows)
   }
 
   async switchWorkflow(workflowName: string) {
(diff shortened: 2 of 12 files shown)


browser_tests/tests/appModeBuilder.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+} from '../fixtures/ComfyPage'
+import { TestIds } from '../fixtures/selectors'
+
+test.describe('App mode builder selection', () => {
+  test.beforeEach(async ({ comfyPage }) => {
+    await comfyPage.page.evaluate(() => {
+      window.app!.api.serverFeatureFlags.value = {
+        ...window.app!.api.serverFeatureFlags.value,
+        linear_toggle_enabled: true
+      }
```
christian-byrne: Do we have a feature flag helper for tests?

browser_tests/tests/appModeBuilder.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    })
+    await comfyPage.settings.setSetting('Comfy.UseNewMenu', 'Top')
```
christian-byrne: Didn't Ben set this as the default in the tests?

browser_tests/tests/appModeBuilder.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    await comfyPage.settings.setSetting(
```
DrJKL: Want to just start in Vue mode?

browser_tests/tests/appModeBuilder.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+      'Comfy.AppBuilder.VueNodeSwitchDismissed',
+      true
+    )
+  })
+
+  test('Can independently select inputs of same name', async ({
+    comfyPage
+  }) => {
+    const items = comfyPage.page.getByTestId(TestIds.builder.ioItem)
```
christian-byrne: Here I think we should be doing it like `comfyPage.appMode.builder.items` and then the logic of how those items are located (in this case, with `TestIds.builder.ioItem`) would be abstracted and encapsulated there.

src/renderer/extensions/linearMode/MobileDisplay.vue
```diff
@@ -191,11 +195,19 @@ const menuEntries = computed<MenuItem[]>(() => [
         "
         :style="{ translate }"
       >
-        <div class="absolute h-full w-screen overflow-y-auto contain-size">
+        <div
+          class="absolute h-full w-screen overflow-y-auto contain-size"
+          role="tabpanel"
+          :aria-hidden="activeIndex !== 0"
```
christian-byrne, line 202: issue (non-blocking): aria-hidden on a tabpanel ancestor of focusable descendants violates WCAG 4.1.2. The inactive tabpanels stay in the DOM (translated off-screen) so their controls remain in the focus order; keyboard users can Tab into hidden controls. Consider :inert="activeIndex !== N" instead of (or in addition to) aria-hidden.

src/renderer/extensions/linearMode/MobileDisplay.vue
```diff
@@ -205,18 +217,24 @@ const menuEntries = computed<MenuItem[]>(() => [
...
       ref="sliderPaneRef"
       class="flex h-22 w-full items-center justify-around gap-4 bg-secondary-background p-4"
+      role="tablist"
     >
       <Button
         v-for="([label, icon], index) in tabs"
         :key="label"
         :variant="index === activeIndex ? 'secondary' : 'muted-textonly'"
         class="h-14 grow flex-col"
+        role="tab"
```
christian-byrne, line 243: suggestion: the tab pattern is missing the aria-controls / aria-labelledby wiring that ties tabs to their panels. The aria-label on each panel satisfies the toHaveAccessibleName assertion in appMode.spec.ts but not the WAI-ARIA tab pattern. Would giving the tabs and panels stable ids and adding :aria-controls / :aria-labelledby be acceptable here?

browser_tests/tests/appModeBuilder.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+      .locator('.lg-node-widget')
+    const count = await prompts.count()
+    for (let i = 0; i < count; i++) {
+      await expect(prompts.nth(i)).toBeVisible()
+      await prompts.nth(i).click()
+      await expect(items).toHaveCount(i + 1)
+    }
+  })
+
+  test('Can drag and drop inputs', async ({ comfyPage }) => {
```
christian-byrne: suggestion (non-blocking): this "Can drag and drop inputs" test overlaps with the steps/denoise reorder tests in builderReorder.spec.ts:40-104 with weaker oracles. The PR description acknowledges potential duplicates. Is the intent to keep both, or should this delegate to comfyPage.appMode.select.dragInputItem like the rest of the suite?

browser_tests/tests/appMode.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+    await comfyPage.workflow.loadWorkflow('default')
+    await comfyPage.dragDrop.dragAndDropURL('/assets/images/og-image.png')
+    const loadImage = await comfyPage.vueNodes.getNodeLocator('10')
+    await expect(loadImage).toBeVisible()
+
+    await comfyPage.appMode.enterAppModeWithInputs([['10', 'image']])
+    await expect(centerPanel).toBeVisible()
+    //an app with an image input will upload the image to the input
+    await comfyPage.dragDrop.dragAndDropFile('workflowInMedia/workflow.webp')
+    await expect(centerPanel).toBeVisible()
```
christian-byrne: issue: this assertion (and the one on line 29) cannot fail. centerPanel is already visible from line 23 after enterAppModeWithInputs, and neither dragAndDropFile nor dragAndDropURL hides it on the upload-into-input path. The test would still pass if the upload silently failed. Could the assertion check an observable upload outcome instead (loadImage widget value, an  src, or an entry in linearWidgets)?

browser_tests/fixtures/helpers/MobileAppHelper.ts
```diff
@@ -0,0 +1,36 @@
+import type { Locator, Page } from '@playwright/test'
+
+import type { ComfyPage } from '@e2e/fixtures/ComfyPage'
+import { TestIds } from '@e2e/fixtures/selectors'
+
+export class MobileAppHelper {
+  readonly actionmenu: Locator
```
christian-byrne: issue: actionmenu is declared and assigned (line 16) but never read by any test, and TestIds.linear.mobileActionMenu (`linear-mobile-menu`) does not appear as a data-testid in src/. Future tests that reach for it will get a zero-element locator that silently passes toBeHidden() / toHaveCount(0). Either wire `linear-mobile-menu` onto the corresponding DOM in MobileDisplay.vue, or drop the field + key until a consumer exists.

browser_tests/fixtures/selectors.ts
```diff
@@ -119,6 +119,15 @@ export const TestIds = {
     domWidgetTextarea: 'dom-widget-textarea',
     subgraphEnterButton: 'subgraph-enter-button'
   },
+  linear: {
+    centerPanel: 'linear-center-panel',
+    mobile: 'linear-mobile',
+    mobileNavigation: 'linear-mobile-navigation',
```
christian-byrne, line 150: suggestion: same concern as the actionmenu helper field — mobileNavigation is defined here but I don't see a matching data-testid in src/ or any test consuming it. Drop it for now, or wire it onto the tablist (which would also fix the i18n-coupled filter({ hasText: 'Run' }) on MobileAppHelper.ts:18).

browser_tests/tests/appModeBuilder.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    const items = comfyPage.appMode.select.selectedItems
+    await comfyPage.appMode.enterBuilder()
+    await comfyPage.appMode.steps.goToInputs()
+    await expect(items).toHaveCount(0)
+
+    await comfyPage.appMode.select.selectInputWidget(
+      'Load Checkpoint',
+      'ckpt_name'
+    )
+    //await expect.soft(items).toHaveCount(0)
```
christian-byrne: issue: the assertion that proves the error case (clicking ckpt_name on Load Checkpoint shouldn't add it to the selection) is commented out. The next assertion at line 82 only runs after loadWorkflow resets state, so the error-case half of the test is functionally untested. Could you restore the assertion (without .soft if it should hold), use test.fixme if currently broken, or delete it with an explanation?

src/renderer/extensions/vueNodes/widgets/components/WidgetMarkdown.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <div class="widget-markdown relative w-full" @dblclick="startEditing">
+  <div
+    :aria-label="widget.name"
```
christian-byrne: note: this line was last touched by 497bafcae ("Fix: Widget sizing with multiple expanding items #7118"). The change here is purely additive (just an aria-label), so risk is low. Flagging only so it gets a manual eyeball that markdown widgets still size correctly alongside other expanding widgets.

Also (issue): :aria-label="widget.name" resolves to aria-label="" when widget.name is empty, which is invalid ARIA. The textarea sibling already uses widget.name || $t('g.markdown') — same fallback would work here.

Also (suggestion, non-blocking): aria-label on a wrapper <div> with no role isn't announced by AT. The focusable child at line 11 (<div tabindex="0">) is what AT actually reads. Could the label move there (paired with role="textbox" aria-readonly="true"), or — if this is purely a locator hook — switch to data-testid to make the intent explicit?

browser_tests/fixtures/helpers/MobileAppHelper.ts
```diff
@@ -0,0 +1,36 @@
...
+    this.actionmenu = this.view.getByTestId(TestIds.linear.mobileActionMenu)
+    this.contentPanel = this.page.getByRole('tabpanel')
+    this.navigation = this.page.getByRole('tablist').filter({ hasText: 'Run' })
+    this.navigationTabs = this.navigation.getByRole('tab')
+    this.workflows = this.view.getByTestId(TestIds.linear.mobileWorkflows)
+  }
+
+  async switchWorkflow(workflowName: string) {
+    await this.workflows.click()
+    await this.page.getByRole('menu').getByText(workflowName).click()
```
christian-byrne, line 25: issue: getByText(workflowName) is a substring match, so `(2)` also matches `(20)`, `(12)`, etc. The workflow-selection spec passes `(2)`..`(5)` which depend on the auto-rename suffix — fine for 4 workflows, brittle as the suite grows. Could you use { exact: true }, name workflows explicitly, or switch by index?

browser_tests/fixtures/helpers/MobileAppHelper.ts
```diff
@@ -0,0 +1,36 @@
...
+  readonly navigation: Locator
+  readonly navigationTabs: Locator
+  readonly view: Locator
+  readonly workflows: Locator
+
+  constructor(private readonly comfyPage: ComfyPage) {
+    this.view = this.page.getByTestId(TestIds.linear.mobile)
+    this.actionmenu = this.view.getByTestId(TestIds.linear.mobileActionMenu)
+    this.contentPanel = this.page.getByRole('tabpanel')
+    this.navigation = this.page.getByRole('tablist').filter({ hasText: 'Run' })
```
christian-byrne, line 18: suggestion: the tablist is located via filter({ hasText: 'Run' }), which breaks once Run/Outputs/Assets are translated. Wiring data-testid="linear-mobile-navigation" onto the tablist (the constant is already reserved on selectors.ts:125) and using getByTestId would avoid that.

christian-byrne (commented):
Thanks for the App Mode tests. Two small accessibility wiring concerns on the new tab pattern in MobileDisplay.vue ([1], [2]), one tautological assertion in appMode.spec.ts ([4]), one commented-out assertion that drops half of an error-case test ([7]), and a few dead test-IDs / fixture helpers that will silently false-pass once consumers exist ([5], [6], [9], [11]). The rest are non-blocking suggestions and one regression-risk note. Defaulting to COMMENT since none are data-loss / security; happy to revisit if [4]/[7] are intentional.

browser_tests/tests/appMode.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+
+      const buttons = await mobile.navigationTabs.all()
+      await buttons[0].dragTo(buttons[2], { steps: 5 })
+      await expect(mobile.contentPanel).toHaveAccessibleName('Outputs')
+
+      await mobile.navigateTab('run')
+      await expect(comfyPage.appMode.linearWidgets).toBeInViewport({ ratio: 1 })
+
+      const steps = comfyPage.page.getByRole('spinbutton')
+      await expect(steps).toHaveValue('20')
```
christian-byrne: suggestion (non-blocking): hardcoding '20' pins the test to the current default for KSampler steps. The deltas below already use observed values; the same pattern would work here: `const start = Number(await steps.inputValue()); …; expect(steps).toHaveValue(String(start + 5))`.

browser_tests/tests/appModeBuilder.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+    await comfyPage.appMode.select.selectInputWidget('Note', 'text')
+    await comfyPage.appMode.select.selectInputWidget('Markdown Note', 'text')
+
+    await expect(items).toHaveCount(0)
+  })
+
+  test('Marks canvas readOnly', async ({ comfyPage }) => {
+    await comfyPage.settings.setSetting(
+      'Comfy.NodeSearchBoxImpl',
+      'v1 (legacy)'
```
christian-byrne, line 80: suggestion (non-blocking): pinning Comfy.NodeSearchBoxImpl to 'v1 (legacy)' means the read-only behavior isn't exercised against v2 (the default). When v1 is removed, this test silently degrades. Could it run against both impls (parameterized), or assert page.evaluate(() => window.app!.canvas!.readOnly) directly?

browser_tests/fixtures/helpers/AppModeHelper.ts
```diff
@@ -146,6 +149,10 @@ export class AppModeHelper {
     await this.toggleAppMode()
   }
 
+  get centerPanel(): Locator {
```
christian-byrne: suggestion (non-blocking): browser_tests/AGENTS.md says "Define UI element locators as public readonly properties assigned in the constructor — not as getter methods." The new centerPanel getter here, selectedItems getter in BuilderSelectHelper.ts:54, and widgets getter in vueNodeFixtures.ts:27 all depart from that. Could they be promoted to readonly Locator initialized in the constructor?

browser_tests/fixtures/helpers/MobileAppHelper.ts
```diff
@@ -0,0 +1,36 @@
...
+  readonly contentPanel: Locator
+  readonly navigation: Locator
+  readonly navigationTabs: Locator
+  readonly view: Locator
+  readonly workflows: Locator
+
+  constructor(private readonly comfyPage: ComfyPage) {
+    this.view = this.page.getByTestId(TestIds.linear.mobile)
+    this.actionmenu = this.view.getByTestId(TestIds.linear.mobileActionMenu)
+    this.contentPanel = this.page.getByRole('tabpanel')
```
christian-byrne, line 17: note: getByRole('tabpanel') resolves to all 3 tabpanels and only collapses to one because Playwright's accessibility tree filters out aria-hidden="true". If the aria-controls suggestion on MobileDisplay.vue lands and tabpanels gain stable ids, this can route through navigationTabs → aria-controls lookup, which is more robust against hidden-toggling becoming async.

browser_tests/tests/appMode.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+        { count: 3 }
+      )
+
+      await expect(steps).toHaveValue('22')
+    })
+
+    test('workflow selection', async ({ comfyPage }) => {
+      const widgetNames = ['seed', 'steps', 'denoise', 'cfg']
+      for (const name of widgetNames)
+        await comfyPage.appMode.enterAppModeWithInputs([['3', name]])
```
christian-byrne, line 143: suggestion (non-blocking): this test calls enterAppModeWithInputs once per widget name (4 sequential full graph reloads). On @mobile CI this likely costs several seconds per call. A batched helper that creates N workflows in a single page.evaluate would speed this up materially.

src/renderer/extensions/linearMode/MobileDisplay.vue
```diff
@@ -205,18 +217,24 @@ const menuEntries = computed<MenuItem[]>(() => [
...
+          :aria-hidden="activeIndex !== 2"
+          :aria-label="t(tabs[2][0])"
         />
       </div>
     </div>
     <div
       ref="sliderPaneRef"
       class="flex h-22 w-full items-center justify-around gap-4 bg-secondary-background p-4"
+      role="tablist"
     >
```
christian-byrne, line 236: suggestion (non-blocking): once role="tablist" / role="tab" are declared, AT users expect the WAI-ARIA tab keyboard pattern (ArrowLeft/ArrowRight + Home/End + roving tabindex). Adding :tabindex="index === activeIndex ? 0 : -1" on the Button plus a @keydown handler on this tablist would close the loop opened by adding the roles.

browser_tests/fixtures/helpers/MobileAppHelper.ts
AustinMroz: This is intentional. I think it's important and beneficial to minimize the usage of testIds because those can not be seen or utilized by a user. A `tabpanel` is a thing a user user utilizes. If content from the inactive `tabpanel`s are simultaneously available and actionable, then I have failed to fulfill the contract of a `tabpanel` implementation.

browser_tests/fixtures/helpers/MobileAppHelper.ts
AustinMroz: This behaviour is consistent with other more common fixtures like `vueNodes.getNodeByTitle`.  `(2)` CAN NOT collide with `(20)` because the parenthesis are included in the match.

browser_tests/fixtures/selectors.ts
AustinMroz: I agree this should be fixed and appreciate it being caught. Will hold off on changing until I can get clarification on keeping `getByRole` or swapping to `getByTestId`

browser_tests/tests/appMode.spec.ts
AustinMroz: Prior to #10122, dragging an image into app mode would instead load the workflow and swap to that tab. Because this workflow is not an app, this would cause app mode to be left.

I'll keep this check (with a description on the expect so it's clear while it's needed) and add an additional check like suggested to ensure that 'If the test were to break, the expectation can not pass before the negative state can occur'

browser_tests/tests/appMode.spec.ts
AustinMroz: The execution takes ~1 second across all iterations, or 15% of test execution time. Definitely longer than I would like, but I suspect implementing, debugging, and maintaining such a change would not be worth the required time investment.

browser_tests/tests/appModeBuilder.spec.ts
AustinMroz: Writing the test was assigned to me and my implementation predates the test in `builderReorder.spec.ts` by almost 2 weeks. :cry: 

In this specific case, the other test is better in every way, so I will remove mine.

browser_tests/tests/appModeBuilder.spec.ts
AustinMroz: When workflows are ran on the ci, there's no models loaded so the node will ahve an error and the assertion will pass. I has moved to soft (and later commenting) during testing. I have re-enabled the assertion without restriction and added a note.

browser_tests/tests/appModeBuilder.spec.ts
AustinMroz: I wish I could, but the V1 usage is intentional. From my testing, the fixtures for V2 are not reliable, general, or terse enough to see usage outside of the V2 tests.
- The fixture opens the search box by core command. This is a violation of policy to prefer actions a user can perform.
- Selecting a node requires manual key presses not handled by the fixture.
- After exiting the menu, the node exists only in ghost mode and has not been placed.
- No fixture exists to finalize placement. From personal testing, a simple `click()` is insufficient.

browser_tests/fixtures/helpers/MobileAppHelper.ts
AustinMroz: The labels here *are* translated, but our browser tests are performed in English. There are more than 53 other playwright tests in our codebase that depend on this assumption.

I am specifically conflicted here because I want to push away from overuse of testIds, but I am not aware of a way to combine a `getByRole` with `getByTestId`. The closest option `.filter({ has })` doesn't function since it's a single element.

Will hold off on making changes here until I have confirmation.

src/renderer/extensions/linearMode/MobileDisplay.vue
AustinMroz: I had assumed that being offscreen/`aria-hidden` would be suffcicient. `inert` is a good call.

src/renderer/extensions/linearMode/MobileDisplay.vue
AustinMroz: I was conflicted on this. I had hoped to push this code towards being a reusable component with a dynamic number of tabs/tabPanels. Manually adding `id`s moves us further away from this. Given that I have not seen success in making this reusable, YAGNI applies. I'll connect these.

src/renderer/extensions/vueNodes/widgets/components/WidgetMarkdown.vue
AustinMroz: It's implemented purely as a locator hook, but it should still be done properly. I'll move the label as requested. The change is to make the widget selectable by `vueNodes.getWidgetByName`, so testids can not be used here.

Moving as suggested will keep the blame unchanged, but I've manually checked and can confirm that sizing still works fine.

christian-byrne (approved):
LGTM! It's beautiful

PR #10640, opened 2026-03-28
fix: selection bounding box not encompassing node footer overlay

jaeone94:
## Summary

Fix multi-select bounding box not encompassing node footer overlays (Enter Subgraph, Advanced, Error buttons).

## Changes

- **What**: Extend node `boundingRect` via `onBounding` to include footer overlay height (32px expanded, 34px collapsed) and override collapsed width to `MIN_NODE_WIDTH` since Vue nodes lack canvas ctx for `measure()`
- 8 E2E tests covering subgraph/regular nodes in expanded/collapsed states at bottom-left/bottom-right positions

## Review Focus

- `onBounding` uses an existing litegraph extension point, not adding new methods
- `MIN_NODE_WIDTH` must stay in sync with CSS `--min-node-width` (documented in comment)
- Regression from #9360 which changed footer from inside-node to overlay approach (`absolute top-full`)

## Screenshots
Before
image

After
image

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10640-fix-selection-bounding-box-not-encompassing-node-footer-overlay-3316d73d3650816c9fc4f36546fbdc52) by [Unito](https://www.unito.io)


browser_tests/tests/selectionBoundingBox.spec.ts
christian-byrne, line 1: Should we add some of these to shared/centralized fixtures/helpers/page objects?

jaeone94: Done !
Replaced `waitForSelectedCount` and `waitForNodeLayout` with existing fixture helpers (`nodeOps.getSelectedGraphNodesCount` via `expect.poll`, `vueNodes.getNodeLocator.waitFor`).

Extracted `loadWithPositions` and `setCollapsed` to shared fixtures:
- `NodeOperationsHelper.loadWithPositions` — serialize-modify-reload pattern for placing nodes at specific positions without separate asset files
- `NodeReference.setCollapsed(collapsed)` — deterministic setter using `node.collapse()` API with current state check (vs `toggleCollapse()` which depends on current state)

The remaining `measureBounds` is kept as a local helper — it combines multi-selection bounds with DOM visual bounds including footer measurement, too specialized to justify sharing.

browser_tests/tests/selectionBoundingBox.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+  await comfyPage.page.evaluate(
+    ({ id, collapsed }) => {
+      const node = window.app!.graph._nodes.find(
+        (n: { id: number | string }) => String(n.id) === id
+      )
+      if (node) {
+        node.flags = node.flags || {}
+        node.flags.collapsed = collapsed
+        window.app!.canvas.setDirty(true, true)
+      }
```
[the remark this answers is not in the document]
jaeone94: Resolved. The local `setNodeCollapsed` function has been removed and replaced with `NodeReference.setCollapsed(collapsed)` — a shared fixture helper that reads the current state and calls `node.collapse(true)` only when the state differs from the target. See commit `8d985e349`.

jaeone94:
@coderabbitai can you approve it ?

browser_tests/fixtures/helpers/NodeOperationsHelper.ts
```diff
@@ -111,6 +114,27 @@ export class NodeOperationsHelper {
...
+          const pos = positions[String(node.id)]
+          if (pos) node.pos = pos
+        }
+        await window.app!.loadGraphData(
+          data as ComfyWorkflowJSON,
+          true,
+          true,
+          null
+        )
+      },
```
christian-byrne, line 133: nit: I think a better method name could be explored -- this is more about changing the node positions in batch rather than loading a graph (although it does use loadGraphData). It can also possibly be decomposed into smaller pure functions.

browser_tests/fixtures/utils/litegraphUtils.ts
```diff
@@ -332,6 +332,16 @@ export class NodeReference {
...
+  async setCollapsed(collapsed: boolean) {
+    await this.comfyPage.page.evaluate(
+      ([id, collapsed]) => {
+        const node = window.app!.canvas.graph!.getNodeById(id)
+        if (!node) throw new Error('Node not found')
+        if (node.collapsed !== collapsed) node.collapse(true)
+      },
+      [this.id, collapsed] as const
+    )
+  }
```
christian-byrne, line 344: nit: In an ideal world, we can do this with user actions (e.g., find the collapse button and click it). Although for litegraph/canvas, it might be too much effort here.

browser_tests/tests/selectionBoundingBox.spec.ts
christian-byrne, line 1: I think most things outside the test.describe can actually be moved to a helper/utils file still

src/renderer/extensions/vueNodes/components/LGraphNode.vue
```diff
@@ -780,6 +781,35 @@ const showAdvancedState = customRef((track, trigger) => {
...
+          : FOOTER_BOUNDING_OFFSET
+
+      // Must match CSS --min-node-width in the template.
+      if (collapsed) out[2] = MIN_NODE_WIDTH
+    }
+  }
+
+  onCleanup(() => {
+    node.onBounding = previousOnBounding
+  })
```
christian-byrne, line 810: I would like to fully explore two other options, as I am not sure how maintainable this pattern is

1. Make the footer part of normal document flow (not absolute-positioned overflow) so getBoundingClientRect() naturally includes it → ResizeObserver picks it up → litegraph gets the right size automatically. No onBounding hack needed.
2. Adjust the ResizeObserver measurement in useVueNodeResizeTracking.ts to account for footer overflow (similar to how it already normalizes for NODE_TITLE_HEIGHT).

The current approach reaches across the boundary from the Vue layer into litegraph's internal callback system, hard-coding magic pixel offsets (32px, 34px) that must stay in sync with CSS. That's potentially quite fragile and the also sort of the kind of coupling the Vue node architecture was designed to avoid.

christian-byrne (commented):
Looks good to me generally. For the actual bugfix, I asked about some other possible ways to think about the solution. The good thing is, since you already made some nice working tests, you can potentially try a few different new approaches and still have the tests to easily validate they also fix the problem

src/renderer/extensions/vueNodes/components/LGraphNode.vue
jaeone94: I've addressed the test code feedback and explored both approaches you mentioned. Here are two PRs for comparison:

- **Option A** (#10711): Inline footer layout with core `_footerHeight`/`_collapsedHeight` properties on `LGraphNode` and a `vueNodesMode` guard in `measure()`. No `onBounding` usage, but adds Vue-specific fields to the core.
- **Option B** (#10712): Same inline footer layout, but uses `onBounding` callback reading from a `vueBoundsOverrides` Map in the Vue layer. Zero core changes, but relies on `onBounding` (with DOM-measured values instead of hardcoded offsets, and proper callback chaining).

Both PRs share the same foundation: footer moved into normal document flow, border/outline unified on root element, resize handles split for footer awareness, and the same E2E test suite.

Would appreciate your thoughts on which direction is preferable, or if there's a different approach worth exploring.

jaeone94: Due to too many design changes, I will resubmit the PR with a new approach.

jaeone94: https://github.com/Comfy-Org/ComfyUI_frontend/pull/10741

PR #10647, opened 2026-03-28
test: add unit tests for commandStore, extensionStore, widgetStore (STORE-04)

christian-byrne:
## Summary

Adds 43 unit tests covering three priority Pinia stores that previously had zero test coverage.

### commandStore (18 tests)
- `registerCommand` / `registerCommands` — single and batch registration, duplicate warning
- `getCommand` — retrieval and undefined for missing
- `execute` — successful execution, metadata passing, error handler delegation, missing command error
- `isRegistered` — presence check
- `loadExtensionCommands` — extension command registration with source, skip when no commands
- `ComfyCommandImpl` — label/icon/tooltip resolution (string vs function), menubarLabel defaulting

### extensionStore (16 tests)
- `registerExtension` — name validation, duplicate detection, disabled extension warning
- `isExtensionEnabled` / `loadDisabledExtensionNames` — enable/disable lifecycle
- Always-disabled hardcoded extensions (pysssss.Locking, pysssss.SnapToGrid, pysssss.FaviconStatus, KJNodes.browserstatus)
- `enabledExtensions` — computed filter
- `isExtensionReadOnly` — hardcoded list check
- `inactiveDisabledExtensionNames` — ghost extension tracking
- Core extension capture and `hasThirdPartyExtensions` detection

### widgetStore (9 tests)
- Core widget availability via `ComfyWidgets`
- Custom widget registration and core/custom precedence
- `inputIsWidget` for both v1 array and v2 object InputSpec formats

## Part of
Test Coverage Q2 Overhaul — Phase 5 (Unit & Component Tests)

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10647-test-add-unit-tests-for-commandStore-extensionStore-widgetStore-STORE-04-3316d73d365081e0b4f6ce913130e489) by [Unito](https://www.unito.io)


src/stores/commandStore.test.ts
DrJKL, line 1: Nit: These are a little bit granular.
https://tidyfirst.substack.com/p/composable-tests

DrJKL (dismissed):

src/stores/widgetStore.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    setActivePinia(createTestingPinia({ stubActions: false }))
+  })
+
+  describe('widgets getter', () => {
+    it('includes core widgets', () => {
+      const store = useWidgetStore()
+      expect(store.widgets.has('INT')).toBe(true)
+      expect(store.widgets.has('FLOAT')).toBe(true)
+      expect(store.widgets.has('STRING')).toBe(true)
+    })
```
DrJKL: 🔴 **Change-detector that tests the mock.** The mock at L13–21 stubs `ComfyWidgets` with `{ INT, FLOAT, STRING, … }`, then this test asserts `widgets.has('INT')`. This only proves the mock is spread into a Map — it cannot catch any real bug. Consider removing.

src/stores/widgetStore.test.ts
```diff
@@ -0,0 +1,101 @@
...
+import { setActivePinia } from 'pinia'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+
+import type { InputSpec as InputSpecV2 } from '@/schemas/nodeDef/nodeDefSchemaV2'
+import type { InputSpec as InputSpecV1 } from '@/schemas/nodeDefSchema'
+import { useWidgetStore } from '@/stores/widgetStore'
+
+/** Cast shorthand — the mock bypasses Zod validation, so we only need the shape `inputIsWidget` reads. */
+const v1 = (spec: unknown) => spec as InputSpecV1
+const v2 = (spec: unknown) => spec as InputSpecV2
```
DrJKL: 🔴 **Unsafe `as` casts via `unknown`** — `v1()`/`v2()` funnel `unknown` through `as InputSpec`, which is the project's forbidden pattern (`as any` equivalent).

The V2 calls like `{ type: 'STRING', name: 'test_input' }` already satisfy `InputSpecV2` — use them directly without the cast. V1 tuples could use `as const satisfies` or construct properly typed data.

src/stores/commandStore.test.ts
```diff
@@ -0,0 +1,211 @@
...
+    it('uses explicit menubarLabel over label', () => {
+      const store = useCommandStore()
+      store.registerCommand({
+        id: 'mbl.explicit',
+        function: vi.fn(),
+        label: 'Label',
+        menubarLabel: 'Menu Label'
+      })
+      expect(store.getCommand('mbl.explicit')?.menubarLabel).toBe('Menu Label')
+    })
```
DrJKL: 🔴 **`ComfyCommandImpl` getter tests are change-detectors.** These test class internal getter plumbing (`label`, `icon`, `menubarLabel`) rather than observable store behavior. They'd pass even if the store broke.

Consider testing these through the store's public surface (e.g., verify label/icon are available after `getCommand` in a more behavioral scenario), or remove them.

Also: `tooltip` getter has the same pattern but isn't tested — if you keep these, add it for consistency.

src/stores/commandStore.test.ts
```diff
@@ -0,0 +1,211 @@
...
+    it('registers multiple commands at once', () => {
+      const store = useCommandStore()
+      store.registerCommands([
+        { id: 'cmd.a', function: vi.fn() },
+        { id: 'cmd.b', function: vi.fn() }
+      ])
+      expect(store.isRegistered('cmd.a')).toBe(true)
+      expect(store.isRegistered('cmd.b')).toBe(true)
+    })
+  })
```
DrJKL: 🟡 **Redundant with `registerCommand`.** `registerCommands` is a trivial `for` loop — this test just re-asserts `isRegistered` in the same way the individual `registerCommand` test does.

src/stores/commandStore.test.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  describe('commands getter', () => {
+    it('returns all registered commands as an array', () => {
+      const store = useCommandStore()
+      store.registerCommand({ id: 'a', function: vi.fn() })
+      store.registerCommand({ id: 'b', function: vi.fn() })
+      expect(store.commands).toHaveLength(2)
+    })
```
DrJKL: 🟡 **Change-detector.** Registers 2 items, asserts `toHaveLength(2)`. Just validates that `Object.values` works.

src/stores/commandStore.test.ts
```diff
@@ -0,0 +1,211 @@
...
+    it('warns on duplicate registration', () => {
+      const store = useCommandStore()
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {})
+
+      store.registerCommand({ id: 'dup', function: vi.fn() })
+      store.registerCommand({ id: 'dup', function: vi.fn() })
+
+      expect(warnSpy).toHaveBeenCalledWith('Command dup already registered')
+      warnSpy.mockRestore()
+    })
```
DrJKL: 🟡 **Missing behavioral assertion.** The store warns *then overwrites* on duplicate (L79–82 in source). This test checks the warning but doesn't verify whether the original or new function survives. Consider asserting the overwrite semantics.

src/stores/commandStore.test.ts
```diff
@@ -0,0 +1,211 @@
...
+      const store = useCommandStore()
+      store.loadExtensionCommands({
+        name: 'test-ext',
+        commands: [
+          { id: 'ext.cmd1', function: vi.fn(), label: 'Cmd 1' },
+          { id: 'ext.cmd2', function: vi.fn(), label: 'Cmd 2' }
+        ]
+      })
+      expect(store.isRegistered('ext.cmd1')).toBe(true)
+      expect(store.isRegistered('ext.cmd2')).toBe(true)
```
DrJKL, line 134: 🟡 **Missing `source` assertion.** `loadExtensionCommands` sets `source: extension.name` on each command, but this test only checks `isRegistered`. Consider:
```ts
expect(store.getCommand('ext.cmd1')?.source).toBe('test-ext')
```

src/stores/commandStore.test.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  describe('loadExtensionCommands', () => {
+    it('registers commands from an extension', () => {
+      const store = useCommandStore()
+      store.loadExtensionCommands({
+        name: 'test-ext',
+        commands: [
+          { id: 'ext.cmd1', function: vi.fn(), label: 'Cmd 1' },
+          { id: 'ext.cmd2', function: vi.fn(), label: 'Cmd 2' }
+        ]
```
DrJKL, line 131: 🟡 **`formatKeySequence` is not tested** — it's exported from the store and performs string transformation on keybinding combos.

src/stores/extensionStore.test.ts
```diff
@@ -0,0 +1,155 @@
...
+  })
+
+  describe('extensions getter', () => {
+    it('returns all registered extensions', () => {
+      const store = useExtensionStore()
+      store.registerExtension({ name: 'ext.a' })
+      store.registerExtension({ name: 'ext.b' })
+      expect(store.extensions).toHaveLength(2)
+    })
+  })
```
DrJKL: 🟡 **Change-detector.** Registers 2, asserts length 2 — just validates `Object.values`. Same pattern as `commandStore`.

src/stores/extensionStore.test.ts
```diff
@@ -0,0 +1,155 @@
...
+      try {
+        store.loadDisabledExtensionNames(['disabled.ext'])
+        store.registerExtension({ name: 'disabled.ext' })
+        expect(warnSpy).toHaveBeenCalledWith(
+          'Extension disabled.ext is disabled.'
+        )
+      } finally {
+        warnSpy.mockRestore()
+      }
+    })
```
DrJKL: 🟡 **Missing behavioral assertion.** Checks the warning, but doesn't verify the extension is still accessible via `isExtensionInstalled`/`extensions` (disabled ≠ unregistered). Consider:
```ts
expect(store.isExtensionInstalled('disabled.ext')).toBe(true)
expect(store.isExtensionEnabled('disabled.ext')).toBe(false)
```

src/stores/widgetStore.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      expect(store.widgets.has('CUSTOM_TYPE')).toBe(true)
+    })
+
+    it('core widgets take precedence over custom widgets with same key', () => {
+      const store = useWidgetStore()
+      const override = vi.fn()
+      store.registerCustomWidgets({ INT: override })
+      // Core widgets are spread last, so they win
+      expect(store.widgets.get('INT')).not.toBe(override)
+    })
```
DrJKL: 🟡 **Fragile negative assertion.** `not.toBe(override)` doesn't say what the value *is*. Consider asserting identity against the core widget mock:
```ts
import { ComfyWidgets } from '@/scripts/widgets'
expect(store.widgets.get('INT')).toBe(ComfyWidgets.INT)
```

DrJKL (commented):
Nice start on store test coverage! All 43 tests pass. A few findings grouped by severity below — mostly around change-detector tests and type safety.

### What's good
- Pinia setup follows project pattern correctly
- Import style and formatting are clean
- Mock pattern follows singleton-in-factory convention
- Behavioral coverage for `execute`, `loadExtensionCommands`, `enabledExtensions`, `inactiveDisabledExtensionNames`, and `inputIsWidget` is solid

src/stores/widgetStore.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  describe('registerCustomWidgets', () => {
+    it('registers multiple custom widgets', () => {
+      const store = useWidgetStore()
+      store.registerCustomWidgets({
+        TYPE_A: vi.fn(),
+        TYPE_B: vi.fn()
+      })
+      expect(store.widgets.has('TYPE_A')).toBe(true)
+      expect(store.widgets.has('TYPE_B')).toBe(true)
+    })
```
DrJKL: 🟡 **Redundant** with `includes custom widgets after registration` (L40–44). Both exercise the same `Map.set()` call path.

src/stores/widgetStore.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    FLOAT: vi.fn(),
+    STRING: vi.fn(),
+    BOOLEAN: vi.fn(),
+    COMBO: vi.fn()
+  }
+}))
+
+vi.mock('@/schemas/nodeDefSchema', () => ({
+  getInputSpecType: (spec: unknown[]) => spec[0]
+}))
```
DrJKL, line 20: 🟡 **COMBO array-spec edge case untested.** V1 COMBO specs use `[['option1','option2'], {}]`. The `getInputSpecType` mock returns `spec[0]`, which would be an *array* (not a string) for COMBO. This edge case could mask a real issue — consider adding a test for it.

src/stores/commandStore.test.ts
christian-byrne: good article - the refactoring in this round already consolidated several tests per your other feedback (removed the change-detectors, merged redundant cases, and moved to behavioral assertions through the store). happy to consolidate further if specific groupings jump out.

src/stores/widgetStore.test.ts
christian-byrne: good catch - the COMBO array-spec edge case is worth covering. will add that in a follow-up.

src/stores/widgetStore.test.ts
christian-byrne: removed in dccf4fd

src/stores/widgetStore.test.ts
christian-byrne: fixed in dccf4fd - V1 tuples use `as const`, V2 specs are direct objects, no more `as unknown` casts

src/stores/commandStore.test.ts
christian-byrne: replaced with behavioral tests through the store in a45c3b7 - tests label/tooltip/menubarLabel via getCommand after registerCommand. tooltip test added too

src/stores/commandStore.test.ts
christian-byrne: removed in a45c3b7

src/stores/commandStore.test.ts
christian-byrne: removed in a45c3b7

src/stores/commandStore.test.ts
christian-byrne: added in a45c3b7 - now asserts the replacement function is called and the original is not

src/stores/commandStore.test.ts
christian-byrne: added in a45c3b7

src/stores/commandStore.test.ts
christian-byrne: added in a45c3b7

src/stores/extensionStore.test.ts
christian-byrne: removed in 1d3a85a

src/stores/extensionStore.test.ts
christian-byrne: added in 1d3a85a - test now asserts isExtensionInstalled and isExtensionEnabled

src/stores/widgetStore.test.ts
christian-byrne: fixed in dccf4fd - now asserts `.toBe(ComfyWidgets.INT)`

src/stores/widgetStore.test.ts
christian-byrne: removed in dccf4fd

src/stores/widgetStore.test.ts
DrJKL: @coderabbitai Open an issue to track this

DrJKL (approved):

PR #10502, opened 2026-03-25
test: add e2e tests for subgraph promoted widget panel regressions

jaeone94:
## Summary

- Add Playwright regression tests that expose three bugs in the subgraph promoted widget panel introduced by the proxy-widget-v2 refactor (#8856)
- Add `data-testid` attributes to `SubgraphEditor`, `SubgraphNodeWidget`, and `RightSidePanel` for stable test selectors

## Tests added

| Test | Expected on main | Bug |
|---|---|---|
| Linked promoted widgets have hide toggle disabled | FAIL | Hide button is enabled for linked widgets that cannot be demoted |
| Linked promoted widgets show link icon instead of eye | FAIL | Eye icon shown instead of link icon |
| Widget labels display renamed values instead of raw names | FAIL | `widget.name` shown instead of `widget.label` |
| Linked promoted widget menu should not show Hide/Show input | FAIL | Hide/Show input option exposed in three-dot menu |

## Test plan

- [ ] All 4 tests fail on `main` (confirming bugs exist)
- [ ] Tests pass after fix PR is applied

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10502-test-add-e2e-tests-for-subgraph-promoted-widget-panel-regressions-32e6d73d36508159b650ede3638bca06) by [Unito](https://www.unito.io)

PR #10648, opened 2026-03-28
fix: resolve subgraph promoted widget panel regressions

jaeone94:
## Summary

Fix four bugs in the subgraph promoted widget panel where linked promotions were not distinguished from independent ones, causing incorrect UI state in both the SubgraphEditor (Settings) panel and the Parameters tab WidgetActions menu.

## Changes

- **What**: Add `isLinkedPromotion` helper to correctly identify widgets driven by subgraph input connections. Fix `disambiguatingSourceNodeId` lookup mismatch that broke `isWidgetShownOnParents` and `handleHideInput` for non-nested promoted widgets. Replace fragile CSS icon selectors with `data-testid` attributes.

## Bugs fixed

Companion fix PR for #10502 (red-green test PR). All 4 E2E tests from #10502 now pass:

| Bug | Root cause | Fix |
|-----|-----------|-----|
| Linked promoted widgets have hide toggle enabled | `SubgraphEditor` only checked `node.id === -1` (physical) — linked promotions from subgraph input connections were not detected | Added `isLinkedPromotion` helper that checks `input._widget` bindings; `SubgraphNodeWidget` `:is-physical` prop now covers both physical and linked cases |
| Linked promoted widgets show eye icon instead of link icon | Same root cause as above — `isPhysical` prop was only true for `node.id === -1` | Extended the `:is-physical` condition to include `isLinkedPromotion` check |
| Widget labels show raw names instead of renamed values | `SubgraphEditor` passed `widget.name` instead of `widget.label \|\| widget.name` | Changed `:widget-name` binding to prefer `widget.label` |
| WidgetActions menu shows Hide/Show for linked promotions | `v-if="hasParents"` didn't exclude linked promotions | Added `canToggleVisibility` computed that combines `hasParents` with `!isLinked` check via `isLinkedPromotion` |

### Additional bugs discovered and fixed

| Bug | Root cause | Fix |
|-----|-----------|-----|
| "Show input" always displayed instead of "Hide input" for promoted widgets | `SectionWidgets.isWidgetShownOnParents` used `getSourceNodeId(widget)` which falls back to `widget.sourceNodeId` when `disambiguatingSourceNodeId` is undefined — this mismatches the promotion store key (`undefined`) | Changed to `widget.disambiguatingSourceNodeId` directly |
| "Hide input" click does nothing | `WidgetActions.handleHideInput` used `getSourceNodeId(widget)` for the same reason — `demote()` couldn't find the entry to remove | Same fix — use `widget.disambiguatingSourceNodeId` directly |

### E2E (Playwright) — `browser_tests/tests/subgraphPromotedWidgetPanel.spec.ts`

| Test | What it verifies |
|------|-----------------|
| linked promoted widgets have hide toggle disabled | All toggle buttons in SubgraphEditor shown section are disabled for linked widgets (covers 1-level and 2-level nested promotions via `subgraph-nested-promotion` fixture) |
| linked promoted widgets show link icon instead of eye icon | Link icons appear for linked widgets, no eye icons present |
| widget labels display renamed values instead of raw names | `widget.label` is displayed when set, not `widget.name` |
| linked promoted widget menu should not show Hide/Show input | Three-dot menu on Parameters tab omits Hide/Show options for linked promotions, Rename is still available |

### Unit (Vitest) — `src/core/graph/subgraph/promotionUtils.test.ts`

7 tests covering `isLinkedPromotion`: basic matching, negative cases, nested subgraph with `disambiguatingSourceNodeId`, multiple inputs, and mixed linked/independent state.

### Unit (Vitest) — `src/components/rightSidePanel/parameters/WidgetActions.test.ts`

- Added `isSubgraphNode: () => false` to mock nodes to prevent crash from new `isLinked` computed

## Review Focus

- `isLinkedPromotion` reads `input._widget` (WeakRef-backed, non-reactive) directly in the template. This is intentional — `_widget` bindings are set during subgraph initialization before the user opens the panel, so stale reads don't occur in practice. A computed-based approach was attempted but reverted because `_widget` changes cannot trigger Vue reactivity.
- `getSourceNodeId` removal in `SectionWidgets` and `WidgetActions` is intentional — the old fallback (`?? widget.sourceNodeId`) caused key mismatches with the promotion store for non-nested widgets.

## Screenshots
Before
image

After
image


src/components/rightSidePanel/subgraph/SubgraphEditor.vue
```diff
@@ -45,6 +45,20 @@ const activeNode = computed(() => {
...
+  if (!node || entries.length === 0) return new Set<string>()
+  const keys = new Set<string>()
+  for (const input of node.inputs) {
+    const w = input._widget
+    if (w && isPromotedWidgetView(w)) {
+      keys.add(`${w.sourceNodeId}:${w.sourceWidgetName}`)
+    }
+  }
+  return keys
+})
```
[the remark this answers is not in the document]
jaeone94: This comment targets the `linkedKeys` computed from a previous revision — that approach was reverted due to reactivity issues with `input._widget` (WeakRef-backed, non-reactive). The current code uses direct `isLinkedPromotion()` calls instead.

Regarding the disambiguation concern: `isLinkedPromotion` matches by `sourceNodeId:sourceWidgetName` against `input._widget` bindings. Since each input slot holds exactly one `_widget`, there's no risk of collapsing distinct promotions — each input is checked independently. Two widgets sharing `sourceNodeId:sourceWidgetName` but differing by `disambiguatingSourceNodeId` would be bound to separate input slots and matched separately.

src/components/rightSidePanel/parameters/SectionWidgets.vue
```diff
@@ -88,7 +86,7 @@ function isWidgetShownOnParents(
       return promotionStore.isPromoted(parent.rootGraph.id, parent.id, {
         sourceNodeId: interiorNodeId,
         sourceWidgetName: widget.sourceWidgetName,
-        disambiguatingSourceNodeId: sourceNodeId
+        disambiguatingSourceNodeId: widget.disambiguatingSourceNodeId
```
DrJKL, line 89: 🔴 **Critical: Lost `sourceNodeId` fallback breaks non-nested promoted widgets**

Old: `disambiguatingSourceNodeId: getSourceNodeId(widget)` → returns `widget.disambiguatingSourceNodeId ?? widget.sourceNodeId`
New: `disambiguatingSourceNodeId: widget.disambiguatingSourceNodeId` → returns `undefined` when no disambiguation exists

`promotionStore.isPromoted()` uses strict equality:
```ts
e.disambiguatingSourceNodeId === source.disambiguatingSourceNodeId
```

But `toPromotionSource()` (the promote path) still calls `getSourceNodeId()` which stores entries **with** the `sourceNodeId` fallback. So a single-level promoted widget is stored with `disambiguatingSourceNodeId: "42"`, but this query now passes `undefined` → `"42" !== undefined` → `isWidgetShownOnParents` returns `false`, making widgets appear not-shown when they actually are.

**Fix:** Either restore the fallback here and in WidgetActions, or update `toPromotionSource()` to stop using it. The strategy must be consistent across promote/query/demote.

src/components/rightSidePanel/parameters/WidgetActions.vue
```diff
@@ -76,16 +81,14 @@ function handleHideInput() {
...
-
     for (const parent of parents) {
       const source: PromotedWidgetSource = {
         sourceNodeId:
           String(node.id) === String(parent.id)
             ? widget.sourceNodeId
             : String(node.id),
         sourceWidgetName: widget.sourceWidgetName,
-        disambiguatingSourceNodeId
+        disambiguatingSourceNodeId: widget.disambiguatingSourceNodeId
```
DrJKL, line 91: 🔴 **Same behavioral mismatch on the demote path**

`handleHideInput` builds a `PromotedWidgetSource` with `disambiguatingSourceNodeId: widget.disambiguatingSourceNodeId` (could be `undefined`), but the entry was stored with the `sourceNodeId` fallback via `toPromotionSource()`. `promotionStore.demote()` uses strict equality — `undefined !== "42"` — so the demote silently fails for non-nested promoted widgets.

See the comment on `SectionWidgets.vue:89` for the full explanation.

src/components/rightSidePanel/subgraph/SubgraphEditor.vue
```diff
@@ -244,8 +252,16 @@ onMounted(() => {
...
+            :widget-name="widget.label || widget.name"
+            :is-physical="
+              node.id === -1 ||
+              (!!activeNode &&
+                isLinkedPromotion(
+                  activeNode,
+                  String(node.id),
+                  getWidgetName(widget)
+                ))
+            "
```
DrJKL: 🟡 **Complex inline template expression — extract to a helper function**

This 7-line `:is-physical` expression is hard to read inline. Consider extracting:
```ts
function isItemPhysical([node, widget]: WidgetItem): boolean {
  return node.id === -1 || (!!activeNode.value &&
    isLinkedPromotion(activeNode.value, String(node.id), getWidgetName(widget)))
}
```
Then: `:is-physical="isItemPhysical([node, widget])"`

src/core/graph/subgraph/promotionUtils.test.ts
```diff
@@ -333,3 +334,92 @@ describe('hasUnpromotedWidgets', () => {
...
+
+  it('matches linked widget from nested subgraph with disambiguatingSourceNodeId', () => {
+    const subgraphNode = createSubgraphWithInputs()
+    subgraphNode.inputs[0]._widget = linkedWidget('6', 'text', {
+      disambiguatingSourceNodeId: '1'
+    })
+
+    expect(isLinkedPromotion(subgraphNode, '6', 'text')).toBe(true)
+    expect(isLinkedPromotion(subgraphNode, '1', 'text')).toBe(false)
+  })
```
DrJKL, line 406: 🟡 **False coverage: `disambiguatingSourceNodeId` is never read by `isLinkedPromotion`**

The SUT only checks `w.sourceNodeId === sourceNodeId` — it never reads `disambiguatingSourceNodeId`. This test passes because it matches on `sourceNodeId='6'`, but the `disambiguatingSourceNodeId: '1'` on the widget is completely ignored. If `isLinkedPromotion` ever needs to respect `disambiguatingSourceNodeId`, this test would silently pass regardless.

Either:
1. Rename the test to clarify it only tests `sourceNodeId` matching
2. Or add an assertion that actually exercises the `disambiguatingSourceNodeId` scenario

src/core/graph/subgraph/promotionUtils.test.ts
```diff
@@ -333,3 +334,92 @@ describe('hasUnpromotedWidgets', () => {
...
+    expect(isLinkedPromotion(subgraphNode, '5', 'string_a')).toBe(false)
+  })
+
+  it('returns false for independent promotion not bound to any input', () => {
+    const subgraphNode = createSubgraphWithInputs()
+    subgraphNode.inputs[0]._widget = linkedWidget('3', 'string_a')
+
+    expect(isLinkedPromotion(subgraphNode, '3', 'string_a')).toBe(true)
+    expect(isLinkedPromotion(subgraphNode, '4', 'value')).toBe(false)
+  })
```
DrJKL: 🟡 **Redundant test case**

"returns false for independent promotion not bound to any input" exercises the same code path as "returns false when no inputs exist or none match" and the negative assertions in "identifies multiple linked widgets". The test name implies a distinct concept but covers no unique branch. Consider removing.

browser_tests/tests/subgraphPromotedWidgetPanel.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+  const labels = shownSection.getByTestId(TestIds.subgraphEditor.widgetLabel)
+  const count = await labels.count()
+  const texts: string[] = []
+  for (let i = 0; i < count; i++) {
+    const text = await labels.nth(i).textContent()
+    if (text) texts.push(text.trim())
+  }
+  return texts
+}
+
```
DrJKL: ⚪ **Nit: use `allTextContents()` instead of manual loop**

Playwright's `Locator.allTextContents()` is more idiomatic:
```ts
async function collectWidgetLabels(shownSection: Locator) {
  const labels = shownSection.getByTestId(TestIds.subgraphEditor.widgetLabel)
  const texts = await labels.allTextContents()
  return texts.map(t => t.trim())
}
```

browser_tests/tests/subgraphPromotedWidgetPanel.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+          'Sub 0'
+        )
+
+        const toggleButtons = shownSection.getByTestId(
+          TestIds.subgraphEditor.widgetToggle
+        )
+        const count = await toggleButtons.count()
+        expect(count).toBeGreaterThan(0)
+
+        for (let i = 0; i < count; i++) {
```
DrJKL: ⚪ **Nit: non-retrying assertion**

`await toggleButtons.count()` followed by `expect(count).toBeGreaterThan(0)` is not a retrying assertion — it can flake if the panel hasn't fully rendered. You could remove these lines entirely since the `for` loop on line 78 already asserts each button is disabled via a retrying assertion.

DrJKL (changes requested):
## PR #10648 Review — fix/subgraph-promoted-widget-regressions

Overall this is a well-scoped bug-fix PR with good test coverage. The `isLinkedPromotion` abstraction is clean, the E2E tests use proper retrying assertions, and the `canToggleVisibility` computed composition in WidgetActions is well done.

However, there is one critical behavioral regression: the `getSourceNodeId` fallback removal changes disambiguation matching semantics for non-nested promoted widgets. See inline comments for details.

### File-level nits (pre-existing, outside diff hunks)

**`SubgraphNodeWidget.vue:6`** — Uses `const props = defineProps<...>()` instead of reactive props destructuring (Vue 3.5 style). Should destructure: `const { nodeTitle, widgetName, isDraggable, isPhysical, class: className } = defineProps<{...}>()`

**`SubgraphNodeWidget.vue:17-23`** — `getIcon()` should be a `computed` instead of a function called in the template. It has no arguments and depends only on props.

**Summary:** 1 critical (2 locations), 3 warnings, 4 nits

AustinMroz (commented):
No issues from me :+1:

src/components/rightSidePanel/parameters/SectionWidgets.vue
jaeone94: Fixed in 878d718. You're right — the fallback removal created an inconsistency with `toPromotionSource()` which still used `getSourceNodeId()`.

Now all promotion store key paths (promote, demote, query) use `widget.disambiguatingSourceNodeId` directly, consistent with `SubgraphNode._setWidget()` which stores entries without the fallback. `toPromotionSource`, `SubgraphEditor` setter/candidateWidgets, and `TabSubgraphInputs` were all updated to match.

`getSourceNodeId` is retained only in `toKey()` for Vue `:key` generation, where the fallback is harmless.

src/components/rightSidePanel/parameters/WidgetActions.vue
jaeone94: Fixed in the same commit (878d718). All promotion store key paths now use `widget.disambiguatingSourceNodeId` directly — see reply on SectionWidgets.vue for the full scope of changes.

src/components/rightSidePanel/subgraph/SubgraphEditor.vue
jaeone94: Extracted to `isItemLinked([node, widget])` helper function in 878d718. Template now reads `:is-physical="isItemLinked([node, widget])"`.

src/core/graph/subgraph/promotionUtils.test.ts
jaeone94: Good catch. Renamed to "matches by sourceNodeId even when disambiguatingSourceNodeId is present" to accurately reflect what it tests. `isLinkedPromotion` intentionally only matches on `sourceNodeId:sourceWidgetName` since each input slot holds exactly one `_widget` — disambiguation is handled at the promotion store level, not in this function.

src/core/graph/subgraph/promotionUtils.test.ts
jaeone94: Removed in 878d718. The negative assertions in "identifies multiple linked widgets" already cover this branch.

browser_tests/tests/subgraphPromotedWidgetPanel.spec.ts
jaeone94: Fixed in 878d718. Now uses `labels.allTextContents()` with `.map(t => t.trim())`.

browser_tests/tests/subgraphPromotedWidgetPanel.spec.ts
jaeone94: Fixed in 878d718. Replaced with `await expect(toggleButtons.first()).toBeVisible()` which is a retrying assertion and also guards against the empty-loop false pass.

jaeone94:
## Response to @DrJKL review

All inline comments addressed in 878d718. Summary:

### Critical (2) — Fixed
- **`getSourceNodeId` fallback mismatch**: Now all promotion store key paths use `widget.disambiguatingSourceNodeId` directly, consistent with `SubgraphNode._setWidget()`. Updated `toPromotionSource`, `SubgraphEditor` setter/candidateWidgets, `TabSubgraphInputs`, `SectionWidgets`, and `WidgetActions`. Verified with `nested-duplicate-widget-names` E2E suite (3/3 pass) and 31 other subgraph promotion tests.

### Warnings (3) — Fixed
- Extracted `isItemLinked` helper from inline template expression
- Renamed disambiguation test to clarify what it actually tests
- Removed redundant "independent promotion" test case

### Nits (2) — Fixed
- `collectWidgetLabels` now uses `allTextContents()`
- Non-retrying `count > 0` replaced with retrying `toBeVisible()` on first toggle

### Pre-existing nits (outside diff)
- `SubgraphNodeWidget.vue`: Converted to reactive props destructuring (Vue 3.5 style) and replaced `getIcon()` function with `icon` computed. Both addressed in 878d718.

DrJKL (approved):

viva-jinyi (approved):

PR #10653, opened 2026-03-28
docs: document Playwright fixture injection pattern for new helpers

christian-byrne:
## Summary

Document the recommended pattern for adding new domain-specific test helpers as Playwright fixtures via `base.extend()` instead of attaching them to `ComfyPage`.

## Changes

- **What**: Added "Creating New Test Helpers" section to `docs/guidance/playwright.md` with fixture extension example and rules

## Review Focus

Documentation-only change. Verify the example code matches the existing pattern in `browser_tests/fixtures/ComfyPage.ts`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10653-docs-document-Playwright-fixture-injection-pattern-for-new-helpers-3316d73d36508145b402cf02a5c2c696) by [Unito](https://www.unito.io)


docs/guidance/playwright.md
```diff
@@ -75,6 +75,37 @@ await page.evaluate(() => {
...
+  assetHelper: AssetHelper
+}>({
+  assetHelper: async ({ comfyPage }, use) => {
+    const helper = new AssetHelper(comfyPage)
+    await helper.setup()
+    await use(helper)
+    await helper.cleanup() // automatic teardown
+  }
+})
+```
```
DrJKL, line 111: It'd be better if we could make them fully modular and then merge them as needed with mergeTests.
```suggestion
```typescript
// browser_tests/fixtures/assetFixture.ts
import { test as base } from '@playwright/test'

export const test = base.extend<{
  assetHelper: AssetHelper
}>({
  assetHelper: async ({ page }, use) => {
    const helper = new AssetHelper(page)
    await helper.setup()
    await use(helper)
    await helper.cleanup() // automatic teardown
  }
})
```
```

DrJKL (dismissed):

docs/guidance/playwright.md
christian-byrne: good call, updated in 4dfded457 - now extends `@playwright/test` base directly and added a note about composing with `mergeTests`

DrJKL (approved):

PR #10658, opened 2026-03-28
docs: document when to use page.evaluate vs user actions in browser tests

christian-byrne:
## Summary

Document acceptable vs avoidable `page.evaluate` patterns in Playwright E2E tests, with migration candidates for existing offenders.

## Changes

- **What**: Add "When to Use `page.evaluate`" section to `docs/guidance/playwright.md` with acceptable/avoid/preferred guidance and 8 migration candidates identified from audit

## Review Focus

Whether the migration candidate list covers the right tests and whether the acceptable/avoid boundary is drawn correctly.

┆Issue is synchronized with this [Notion page](https://www.notion.so/PR-10658-docs-document-when-to-use-page-evaluate-vs-user-actions-in-browser-tests-3316d73d36508186be90f263f36daf75) by [Unito](https://www.unito.io)

### DIFF
--- a/docs/guidance/playwright.md
+++ b/docs/guidance/playwright.md
@@ -89,9 +89,10 @@ Tags are respected by config:
 
 ## When to Use `page.evaluate`
 
-### Acceptable
+### Acceptable (use sparingly)
 
-Reading internal state that has no UI representation:
+Reading internal state that has no UI representation (prefer locator
+assertions whenever possible):
 
 ```typescript
 // Reading graph/store values
--- a/docs/guidance/playwright.md
+++ b/docs/guidance/playwright.md
@@ -137,20 +137,6 @@ Use helper methods from `browser_tests/fixtures/helpers/` that wrap real user
 interactions (e.g., `comfyPage.settings.setSetting`, `comfyPage.nodeOps`,
 `comfyPage.workflow.loadWorkflow`).
 
-### Migration Candidates
-
-Action-oriented `page.evaluate` calls to replace with user interactions or
-fixture helpers:
-
-1. `actionbar.spec.ts` — sets widget value and calls `changeTracker.checkState()` via evaluate instead of interacting with the widget UI
-2. `changeTracker.spec.ts` — calls `emitBeforeChange`/`emitAfterChange` via evaluate instead of performing actual canvas edits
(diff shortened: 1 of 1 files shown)


docs/guidance/playwright.md
```diff
@@ -87,6 +87,69 @@ Tags are respected by config:
 - Check `browser_tests/assets/` for test data and fixtures
 - Use realistic ComfyUI workflows for E2E tests
 
+## When to Use `page.evaluate`
+
+### Acceptable
+
+Reading internal state that has no UI representation:
```
DrJKL: Which is something we should try to avoid doing.

docs/guidance/playwright.md
```diff
@@ -87,6 +87,69 @@ Tags are respected by config:
...
+fixture helpers:
+
+1. `actionbar.spec.ts` — sets widget value and calls `changeTracker.checkState()` via evaluate instead of interacting with the widget UI
+2. `changeTracker.spec.ts` — calls `emitBeforeChange`/`emitAfterChange` via evaluate instead of performing actual canvas edits
+3. `selectionToolboxSubmenus.spec.ts` — dispatches synthetic `MouseEvent` click instead of using Playwright's `locator.click()`
+4. `dialog.spec.ts:443` — calls `showSignInDialog()` via evaluate instead of triggering it through the UI
+5. `colorPalette.spec.ts:181` — calls `addCustomColorPalette()` via evaluate instead of using the settings UI to import a palette
+6. `menu.spec.ts:156` — registers an extension command via evaluate; could use a test fixture that pre-registers extensions
+7. `groupSelectChildren.spec.ts:115` — deselects all nodes via evaluate instead of clicking an empty canvas area
+8. `widget.spec.ts:128,149,173` — mutates widget values directly via evaluate instead of interacting with widget controls
```
DrJKL: This seems like it'll be out of date soon.

DrJKL (dismissed):

docs/guidance/playwright.md
christian-byrne: good point - added "use sparingly" caveat to the heading and "(prefer locator assertions whenever possible)" to make it clear this is a last resort. fixed in a57fdc6b0

docs/guidance/playwright.md
christian-byrne: agreed, removed the entire migration candidates section. fixed in 94357c391

